PR #2090, opened 2021-09-10
Test that trace and span id are present where they are expected by the product

vladkluev:
# Context
We currently store trace_id and span_id in two places on the errors table:

* As columns
* As keys and values on the contexts table

The product expects both of these to be present

# Solution
Add tests to make sure that if someone removes these it's intentional

### DIFF
--- a/snuba/datasets/errors_processor.py
+++ b/snuba/datasets/errors_processor.py
@@ -104,12 +104,13 @@ class ErrorsProcessor(EventsProcessorBase):
         tags: Mapping[str, Any],
     ) -> None:
         transaction_ctx = contexts.get("trace") or {}
-        if transaction_ctx.get("trace_id", None):
-            output["trace_id"] = str(uuid.UUID(transaction_ctx["trace_id"]))
-        if transaction_ctx.get("span_id", None):
-            output["span_id"] = int(transaction_ctx["span_id"], 16)
         # We store trace_id and span_id as promoted columns and on the query level
         # we make sure that all queries on contexts[trace.trace_id/span_id] use those promoted
         # columns instead. So we don't need to store them
-        transaction_ctx.pop("trace_id", None)
-        transaction_ctx.pop("span_id", None)
+        trace_id = transaction_ctx.pop("trace_id", None)
+        span_id = transaction_ctx.pop("span_id", None)
+
+        if trace_id:
+            output["trace_id"] = str(uuid.UUID(trace_id))
+        if span_id:
+            output["span_id"] = int(span_id, 16)
--- a/snuba/datasets/errors_processor.py
+++ b/snuba/datasets/errors_processor.py
@@ -104,11 +104,8 @@ class ErrorsProcessor(EventsProcessorBase):
         tags: Mapping[str, Any],
     ) -> None:
         transaction_ctx = contexts.get("trace") or {}
-        # We store trace_id and span_id as promoted columns and on the query level
-        # we make sure that all queries on contexts[trace.trace_id/span_id] use those promoted
-        # columns instead. So we don't need to store them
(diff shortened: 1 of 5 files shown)


snuba/datasets/errors_processor.py
```diff
@@ -113,3 +108,8 @@ def extract_promoted_contexts(
             output["trace_id"] = str(uuid.UUID(transaction_ctx["trace_id"]))
         if transaction_ctx.get("span_id", None):
             output["span_id"] = int(transaction_ctx["span_id"], 16)
+        # We store trace_id and span_id as promoted columns and on the query level
+        # we make sure that all queries on contexts[trace.trace_id/span_id] use those promoted
+        # columns instead. So we don't need to store them
+        transaction_ctx.pop("trace_id", None)
+        transaction_ctx.pop("span_id", None)
```
lynnagara, line 115: I'd write this as
```
trace_id = transaction_ctx.pop("span_id", None)
if trace_id is not None:
   output["trace_id"] = str(uuid.UUID(trace_id))
```

lynnagara (commented):
This seems good to me. Just one question - do you know whether we are returning trace ID data from contexts anywhere? I noticed we have this in Sentry and I'm wondering if this change is going to alter some of the results we're sending back https://github.com/getsentry/sentry/blob/f718040568a10f566db096e4b6202be7d133fd44/src/sentry/snuba/events.py#L260-L265

vladkluev:
That URL is for the `events` dataset which is not altered.

vladkluev:
But also I don't really have good context for the sentry codebase. As far as I know, clickhouse is only used for querying. We don't return clickhouse rows, we just return event ids to look up in nodestore right?

lynnagara (approved):
The events dataset is the main way we query the Clickhouse errors table, and we can return many columns not just event ID. If we change the data we're writing to some columns then queries to that dataset will return slightly different results. I don't think that particular column is important though so this is not a blocker to me. I think there is a chance this may slightly change some results we send to Sentry in some edge cases though.

fpacifici (changes requested):
Please do not remove these columns from the errors table without having ensured nobody is using them. We validated that nobody is using them via the querylog only for transactions. Unless somebody else did the test for you we would have to test the queries against the errors table as well. 

Specifically, issue search allows customers to issue arbitrary queries on the contexts column (I suspect this is basically unintended tech debt, but it is now in use). 

The task was aimed at reducing the size of contexts on the transactions table. The errors table is not a problem at present. So I would focus first on transactions, measure the improvement and then we could get to the errors table if needed after testing that we are never querying those contexts via arrayJoin as well.

vladkluev:
@fpacifici since it actually is important I'll remove the storage changes but I'll keep the tests that make sure they're there.

snuba/datasets/errors_processor.py
```diff
@@ -109,7 +104,10 @@ def extract_promoted_contexts(
...
-            output["trace_id"] = str(uuid.UUID(transaction_ctx["trace_id"]))
-        if transaction_ctx.get("span_id", None):
-            output["span_id"] = int(transaction_ctx["span_id"], 16)
+        trace_id = transaction_ctx.get("trace_id", None)
+        span_id = transaction_ctx.get("span_id", None)
+
+        if trace_id:
+            output["trace_id"] = str(uuid.UUID(trace_id))
+        if span_id:
+            output["span_id"] = int(span_id, 16)
```
fpacifici, line 113: You do not need this change anymore. Up to you.

fpacifici (approved):
thanks

snuba/datasets/errors_processor.py
vladkluev: technically it's a performance improvement 😂

PR #2091, opened 2021-09-10
feat(suspect-spans): Add spans table

Zylphrex:
This change adds a new span table to be able to support span search. This includes

- The table schema for the table
- The migration group for spans
- The basic consumer for spans

### DIFF
--- a/snuba/datasets/storages/__init__.py
+++ b/snuba/datasets/storages/__init__.py
@@ -26,4 +26,5 @@ class StorageKey(Enum):
     SESSIONS_HOURLY = "sessions_hourly"
     ORG_SESSIONS = "org_sessions"
     SPANS = "spans"
+    SPANS_V2 = "spans_v2"
     TRANSACTIONS = "transactions"
--- a/snuba/datasets/storages/spans_v2.py
+++ b/snuba/datasets/storages/spans_v2.py
@@ -1,0 +1,63 @@
+from snuba.clickhouse.columns import UUID, ColumnSet, DateTime, Float
+from snuba.clickhouse.columns import SchemaModifiers as Modifiers
+from snuba.clickhouse.columns import String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.datasets.schemas.tables import WritableTableSchema
+from snuba.datasets.spans_processor import SpansMessageProcessorV2
+from snuba.datasets.storage import WritableTableStorage
+from snuba.datasets.storages import StorageKey
+from snuba.datasets.table_storage import build_kafka_stream_loader_from_settings
+from snuba.query.processors.type_converters.hexint_column_processor import (
+    HexIntColumnProcessor,
+)
+from snuba.utils.streams.topics import Topic
+from snuba.web.split import TimeSplitQueryStrategy
+
+columns = ColumnSet(
+    [
+        ("project_id", UInt(64)),
+        ("transaction_id", UUID()),
+        ("trace_id", UUID()),
+        ("transaction_span_id", UInt(64)),
+        ("span_id", UInt(64)),
+        ("parent_span_id", UInt(64, Modifiers(nullable=True))),
+        ("transaction_name", String()),
+        ("span_group_id", UUID()),
+        ("op", String()),
+        ("status", UInt(8)),
+        # transaction start/finish times
+        ("start_ts", DateTime()),
+        ("start_ns", UInt(32)),
+        ("finish_ts", DateTime()),
+        ("finish_ns", UInt(32)),
+        # span start/finish times
+        ("span_start_ts", DateTime()),
+        ("span_start_ns", UInt(32)),
+        ("span_finish_ts", DateTime()),
+        ("span_finish_ns", UInt(32)),
+        ("duration_ms", UInt(32)),
+        ("exclusive_time_ms", Float(64)),
+        ("retention_days", UInt(16)),
+        ("deleted", UInt(8)),
+    ]
+)
+
+schema = WritableTableSchema(
+    columns=columns,
+    local_table_name="spans_v2_local",
+    dist_table_name="spans_v2_dist",
+    storage_set_key=StorageSetKey.TRANSACTIONS,
+)
+
+storage = WritableTableStorage(
+    storage_key=StorageKey.SPANS_V2,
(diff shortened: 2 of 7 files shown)


vladkluev:
I know this is a draft PR however please separate the migrations from all other code changes. See the [clickhouse migration guide](https://www.notion.so/sentry/Running-Clickhouse-Migrations-5faaf0564fd947acb0640536f49c0b5e) for more info

snuba/datasets/storages/spans_v2.py
```diff
@@ -0,0 +1,63 @@
...
+columns = ColumnSet(
+    [
+        ("project_id", UInt(64)),
+        ("transaction_id", UUID()),
+        ("trace_id", UUID()),
+        ("transaction_span_id", UInt(64)),
+        ("span_id", UInt(64)),
+        ("parent_span_id", UInt(64, Modifiers(nullable=True))),
+        ("transaction_name", String()),
+        ("span_group_id", UUID()),
```
Zylphrex, line 25: Clickhouse doesn't have an `UInt128` type so I'm using `UUID` here because the `span_group_id` is currently an `MD5` hash. Though we could potentially change this to a `UInt64` and choose a different hash function.

snuba/migrations/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,82 @@
...
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="spans_v2_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    storage_set=StorageSetKey.TRANSACTIONS,
+                    version_column="deleted",
+                    order_by="(project_id, toStartOfDay(finish_ts), transaction_name, cityHash64(transaction_span_id)), cityHash64(span_id)",
```
Zylphrex, line 48: The order by here starts the same way as the [order by on the transactions table](https://github.com/getsentry/snuba/blob/a0d15215b4c017c4f1d02c385b39ff5a9177db73/snuba/migrations/snuba_migrations/transactions/0001_transactions.py#L73). The differences are
- `span_id` on the transactions table is `transaction_span_id` on the span table
- the span table is additionally sorting by the individual span's `span_id`

snuba/datasets/storages/spans_v2.py
```diff
@@ -0,0 +1,63 @@
...
+        ("op", String()),
+        ("status", UInt(8)),
+        # transaction start/finish times
+        ("start_ts", DateTime()),
+        ("start_ns", UInt(32)),
+        ("finish_ts", DateTime()),
+        ("finish_ns", UInt(32)),
```
Zylphrex, line 32: We require the transaction's start/finish times here as it needs to be in the order by clause to help with the merge join.

snuba/migrations/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,82 @@
...
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="spans_v2_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    storage_set=StorageSetKey.TRANSACTIONS,
+                    version_column="deleted",
+                    order_by="(project_id, toStartOfDay(finish_ts), transaction_name, cityHash64(transaction_span_id), cityHash64(span_id))",
```
Zylphrex, line 48: The order by here starts the same way as the [order by on the transactions table](https://github.com/getsentry/snuba/blob/a0d15215b4c017c4f1d02c385b39ff5a9177db73/snuba/migrations/snuba_migrations/transactions/0001_transactions.py#L73). The differences are
- `span_id` on the transactions table is `transaction_span_id` on the span table
- the span table is additionally sorting by the individual span's `span_id`

Also, will we need to order by the `trace_id` as that'll likely be in the `JOIN`?

evanh: Not sure about that, @fpacifici is this what you were asking the Clickhouse team?

snuba/datasets/storages/spans_v2.py
evanh: UUID makes sense to me. This is the grouping hash that's based on description amongst other things?

Zylphrex: This grouping hash is currently based on a normalized description. Of course, we can add more to it to refine the grouping as needed. The relevant PR is getsentry/sentry#28184

snuba/datasets/storages/__init__.py
```diff
@@ -26,4 +26,5 @@ class StorageKey(Enum):
     SESSIONS_HOURLY = "sessions_hourly"
     ORG_SESSIONS = "org_sessions"
     SPANS = "spans"
+    SPANS_V2 = "spans_v2"
```
fpacifici, line 29: Just replace the existing one. We are never going to use that one again.

snuba/datasets/storages/spans_v2.py
fpacifici: Why do we need a full 128 bit? 
https://en.wikipedia.org/wiki/Birthday_attack
With 64 bits you will have less than 0.1% chances of collision with a 10^8 different values. And even if it happened, this is a similarity indicator, occasional  collision would likely be harmless.

Why is this important: 128 bit  makes it a column twice as big as a 64 bit one. Which means it has a 2x impact on the data to scan, directly impacting the response time of your query.

snuba/migrations/snuba_migrations/spans_v2/0001_spans_v2.py
fpacifici: Sadly it turns out we cannot use merge joins.
https://stackoverflow.com/questions/69182344/partial-merge-join-behavior-if-the-right-table-is-already-sorted-according-to-th/69183060#69183060

We need to rethink this. 
I will revisit this tomorrow morning. Promised

Zylphrex:
Switching strategies from span table to array columns, different PR coming up.

PR #2092, opened 2021-09-10
ref(latency) Add metrics to track various ingest latencies

evanh:
Add a metric that will track the delay between our processing and the event
timestamp, the time we received the event and the time the event was written
to nodestore. This should help debug a potential ingestion delay problem.

### DIFF
--- a/snuba/datasets/events_processor_base.py
+++ b/snuba/datasets/events_processor_base.py
@@ -3,7 +3,7 @@ from abc import ABC, abstractmethod
 from datetime import datetime
 from typing import Any, Mapping, MutableMapping, Optional, Sequence, Tuple, TypedDict
 
-from snuba import settings
+from snuba import environment, settings
 from snuba.consumers.types import KafkaMessageMetadata
 from snuba.datasets.events_format import (
     EventTooOld,
@@ -25,8 +25,12 @@ from snuba.processor import (
     _ensure_valid_date,
     _unicodify,
 )
+from snuba.utils.metrics.wrapper import MetricsWrapper
 
 logger = logging.getLogger(__name__)
+
+
+metrics = MetricsWrapper(environment.metrics, "events.base.processor")
 
 
 REPLACEMENT_EVENT_TYPES = frozenset(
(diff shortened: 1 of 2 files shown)


snuba/datasets/events_processor_base.py
```diff
@@ -108,6 +112,32 @@ def extract_promoted_contexts(
...
+            metrics.timing(
+                "process_timestamp_latency",
+                (now - output["timestamp"].timestamp()) * 1000,
+            )
+        if output.get("received") is not None:
+            metrics.timing(
+                "process_received_latency",
+                (now - output["received"].timestamp()) * 1000,
+            )
+        if data.get("nodestore_insert") is not None:
```
lynnagara, line 132: Wouldn't it be more straightforward to just track the latency between received time and nodestore_insert time from Sentry and not need pass extra metadata about nodestore (a system Snuba knows nothing about) to Snuba? What additional information does doing it this way provide us?

evanh: I have already compared nodestore with received, and it's negligible. The question I'm trying to answer now is what is the discrepancy between received, nodestore and the Clickhouse write. It seems like sometimes that lag is on the order of several minutes and I want to see if I can track that down.

snuba/datasets/events_processor_base.py
```diff
@@ -108,6 +112,32 @@ def extract_promoted_contexts(
...
+    ) -> None:
+        # XXX: For now this function is purely for debugging purposes. While
+        # it would be nice to have these kind of statistics for our ingest
+        # system I don't think this hack is the correct solution.
+        now = datetime.utcnow().timestamp()
+        if "timestamp" in output:
+            metrics.timing(
+                "process_timestamp_latency",
+                (now - output["timestamp"].timestamp()) * 1000,
+            )
```
fpacifici, line 126: I think we have something more reliable for this:
https://github.com/getsentry/arroyo/blob/b0a28268925347645974ca5174dd0369ff58fd2a/arroyo/processing/strategies/batching.py#L139-L146

Which is the time between when the event was inserted in the Kafka topic (events) and the time we were writing to clickhouse. It is what we use for our alerts and SLO.

A metric on event time is not supposed to be useful as the event tiestamp is self reported by the user. It can be in the past or even in the future, so unrelated from our ingestion pipeline.

fpacifici (changes requested):
I would not track the ingestion latency based on the event timestamp as that is reported by the user and unrelated with our ingestion pipeline.

fpacifici:
The root cause of the issue Evan was investigating has been identified. 
Let's resurrect this when/if we start working again on tracking events end to end

PR #2093, opened 2021-09-13
feat: Produce original message timestamp to commit log headers

lynnagara:
This change writes the timestamp of the original Kafka message
in the header of its commit log topic with the key `orig_message_ts`.
This timestamp is used as the basis for subscription scheduling and
having this in the commit log will enable subscriptions to no longer
need to consume the original topic in future, only the commit log.

This depends on https://github.com/getsentry/arroyo/pull/28,
https://github.com/getsentry/arroyo/pull/29 and a
new release of arroyo.

### DIFF
--- a/snuba/utils/streams/kafka_consumer_with_commit_log.py
+++ b/snuba/utils/streams/kafka_consumer_with_commit_log.py
@@ -1,12 +1,16 @@
+from datetime import datetime
 from typing import Any, Mapping, Optional
 
 from arroyo import Message, Partition, Topic
 from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
 from arroyo.synchronized import Commit, commit_codec
+from arroyo.types import Offset
 from arroyo.utils.retries import RetryPolicy
 from confluent_kafka import KafkaError
 from confluent_kafka import Message as ConfluentMessage
 from confluent_kafka import Producer as ConfluentProducer
+
+from snuba.settings import PAYLOAD_DATETIME_FORMAT
 
 
 class KafkaConsumerWithCommitLog(KafkaConsumer):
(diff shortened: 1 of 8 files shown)


evanh (approved):
Looks good but I would update the commit message to be `orig_message_ts` to match the code.

Are the tests failing because the arroyo version is wrong?

snuba/utils/streams/kafka_consumer_with_commit_log.py
```diff
@@ -33,16 +37,21 @@ def __commit_message_delivery_callback(
...
             payload = commit_codec.encode(commit)
             self.__producer.produce(
                 self.__commit_log_topic.name,
                 key=payload.key,
                 value=payload.value,
+                headers={
+                    "orig_message_ts": datetime.strftime(
+                        offset.timestamp, PAYLOAD_DATETIME_FORMAT
+                    ),
+                },
```
fpacifici, line 54: What is the reason to use headers for this? 
Headers should be used for message metadata (basically stuff that would not break your application if they disappeared), or observability/routing information. 
If it is done in order to avoid having to change the consumers, I think we should do that investment and have a clean interface.
Is it about kafka compaction of this topic? 

Also doing this breaks the idea that we abstract the logic to generate payloads through a codec, thus increasing the coupling between the producing and consuming code.
If we need to store them as header, I think you should still add the timestamp to the Commit object and change the codec to extract headers

lynnagara: Apart from putting it into the headers the other options would be to put it in the key or the body
- Key: I don't think we should put it in the key as we lose the benefits of log compaction.
- Body: The payload is relatively simple today (it's just the commit offset) and I was reluctant to make it more complex. If we put it the payload we probably want to convert that to json format. Unlike the header solution, he post process forwarder would then need to be aware of this and deserialise the json. To avoid this I have a slight preference to put the timestamp in the headers but this is not a strong opinion and I think there are good arguments for both ways.

I didn't update the commit object and codec because I didn't want arroyo to need to be aware of this but I can make that change.

fpacifici: re: commit object. I would definitively do that. 

re: body. Mh, ok, I think eventually we will have to change the body structure anyway. It is ok if we do not do it right away considering the impact on consumers.

PR #2094, opened 2021-09-14
feat(subscription): Add support for sessions subscriptions

ahmedetefy:
This PR:
- Modifies `LegacySubscriptionData`, `DelegateSubscriptionData`, `SnqlSubscriptionData` to include a `entity_subscription ` parameter
- Creates classes `EntitySubscription`, `SessionsSubscription`, `EventsBaseSubscription`, `TransactionSubscription` and `EventSubscription` to decouple the dataset specific logic from `SnqlSubscriptionData`, `LegacySubscriptionData` and `DelegateSubscriptionData`
- Modifies `SessionsQueryStorageSelector.select_storage` to check if it is a crash rate alert subscription and if the time_window is less than or equal to 1h and if so uses raw storage otherwise use materialized storage
- Adds intellij .idea to gitignore

Note:
As a follow up task, we should investigate refactoring out the `dataset` use in Subscription classes by not sending the `dataset` to `SnqlSubscription.build_request` and instead have the entity sent in the Subscription class constructor and just store it on the class instance. 
It seems that the dataset is only needed because `SnqlSubsctription.build_request` calls `build_request` which needs the dataset to call `parse_snql_query` function, but the function uses the dataset to extract the entity -> https://github.com/getsentry/snuba/blob/master/snuba/query/snql/parser.py#L1316


snuba/subscriptions/data.py
```diff
@@ -129,6 +141,57 @@ class LegacySubscriptionData(SubscriptionData):
...
+    def has_crash_rate_alert_aggregation(aggregations: Sequence[Aggregation]) -> bool:
+        for aggregation in aggregations:
+            try:
+                if aggregation[2] and bool(
+                    CRASH_RATE_ALERT_PATTERN.match(aggregation[2])
+                ):
+                    return True
+            except IndexError:
+                continue
+        return False
```
lynnagara, line 159: Doesn't that mean anyone who just uses an alias containing `crash_free_percentage` in any subscription query will match the sessions behavior? It doesn't seem to check the dataset so it could even be applying this logic to events/transactions. This seems a rather unreliable way to determine whether how to process a subscription since this is something that may even be user defined in the future. I don't think I quite get why we need to do this at all?

snuba/subscriptions/data.py
```diff
@@ -129,6 +141,57 @@ class LegacySubscriptionData(SubscriptionData):
...
+        dictionary used to create an instance from this class.
+        """
+        if cls.has_crash_rate_alert_aggregation(data.get("aggregations", [])):
+            if not data.get("organization"):
+                raise InvalidQueryException(
+                    "No organization provided for Sessions subscription"
+                )
+
+            # Querying over the sessions dataset requires adding the following required fields:
+            # organization, offset and limit.
```
lynnagara, line 177: Why do you say offset is a required field in sessions? I don't think it's there at all.

snuba/subscriptions/data.py
```diff
@@ -129,6 +141,57 @@ class LegacySubscriptionData(SubscriptionData):
...
+                if aggregation[2] and bool(
+                    CRASH_RATE_ALERT_PATTERN.match(aggregation[2])
+                ):
+                    return True
+            except IndexError:
+                continue
+        return False
+
+    @classmethod
+    def update_sessions_subscriptions_with_session_params(
```
lynnagara, line 162: `subscriptions.data` module should be fairly generic and work for any datasets we want to have subscriptions for in the future. I don't think you should be adding sessions specific methods here as this does not seem to be the right place for them. More concerning to me is the fact that we now seem to be calling these functions on all subscriptions even those that have nothing to do with sessions. Perhaps all of this logic would be better as defined on the dataset/entity or in a query processor instead?

snuba/subscriptions/data.py
```diff
@@ -173,28 +255,47 @@ def from_dict(cls, data: Mapping[str, Any]) -> LegacySubscriptionData:
...
+            "project_id": data["project_id"],
+            "conditions": data["conditions"],
+            "aggregations": data["aggregations"],
+            "time_window": timedelta(seconds=data["time_window"]),
+            "resolution": timedelta(seconds=data["resolution"]),
+        }
+        legacy_subs_dict = cls.update_sessions_subscriptions_with_session_params(
+            data_dict=legacy_subs_dict, data=data
         )
+        return LegacySubscriptionData(**legacy_subs_dict)
```
lynnagara, line 268: @evanh Do you know the status of LegacySubscriptionData and if it can be removed? I think it would help to simplify this implementation if we do not have to support both types of subscription data.

snuba/datasets/entities/sessions.py
```diff
@@ -211,7 +212,18 @@ def __init__(self) -> None:
...
+        # This checks if we are creating this query as a consequence of a Crash Rate Alert and if so
+        # we set the granularity sent by Sentry.
+        # For time windows <= 1h, we expect to use granularity of 60s, and 3600s for time windows
+        # greater than that
+        for col in query.get_selected_columns():
+            if col.name and bool(CRASH_RATE_ALERT_PATTERN.match(col.name)):
+                granularity = query.get_granularity()
+
+        if granularity is None:
+            granularity = extract_granularity_from_query(query, "started") or 3600
```
lynnagara, line 226: Why can't we just use the granularity specified in the query to begin with rather than infer it from some random alias used in the query? Under what scenarios would we expect those values to be different anyway and why?

lynnagara (commented):
A few comments inline. To summarise, my main concerns are:
- The method of detecting the alert condition is unreliable
- We shouldn't make the subscriptions module highly specific to sessions. It should be generic and dataset/entity/storage related implementation should live in those classes instead
- It seems we would now be running the sessions specific code on all types of subscriptions not just sessions ones and we should not be doing this

snuba/datasets/entities/sessions.py
fpacifici: granularity is not a reliable indicator at this point of the query in the ast + snql world with entities. 
If a query processor on the entity makes changes to the time condition, granularity is not kept up to date. Te only source of truth is the actual timestamp condition.
The entire concept of query granaularity should be rethought.

ahmedetefy:
> A few comments inline. To summarise, my main concerns are:
> 
> * The method of detecting the alert condition is unreliable
> * We shouldn't make the subscriptions module highly specific to sessions. It should be generic and dataset/entity/storage related implementation should live in those classes instead
> * It seems we would now be running the sessions specific code on all types of subscriptions not just sessions ones and we should not be doing this

Thanks for your review! I have made changes in accordance with your comments.. please let me know if my changes cover it

snuba/subscriptions/data.py
ahmedetefy: That makes sense!

snuba/subscriptions/data.py
ahmedetefy: I mistook an error message

snuba/subscriptions/data.py
ahmedetefy: I changed the alias to a more specific name `_crash_rate_alert_aggregate` and now I ensured I only call it from within the Sessions entity .. does that resolve it?

snuba/datasets/entities/sessions.py
ahmedetefy: So I set the granularity now in the query processor, and I do not call the `extract_granularity_from_query` if i did set it and it is for a crash rate alert subscription.. is that a good option

snuba/datasets/entities/sessions.py
```diff
@@ -253,10 +264,11 @@ def __init__(self) -> None:
 
     def get_extensions(self) -> Mapping[str, QueryExtension]:
         return {
-            "timeseries": TimeSeriesExtension(
+            "timeseries": SessionsTimeSeriesExtension(
```
evanh, line 267: Extensions are deprecated and are not being used anymore in prod. All this logic is done in the `TimeSeriesProcessor` now.

snuba/datasets/entities/sessions.py
evanh: The `granularity` field is not a source of truth, since the `TimeSeriesProcessor` will change the conditions of the query but not the granularity field. Is there a reason to change this at all? Why not extract the granularity here?

snuba/query/timeseries_extension.py
```diff
@@ -122,3 +122,48 @@ def __init__(
...
+
+
+class SessionsTimeSeriesExtensionProcessor(TimeSeriesExtensionProcessor):
+    def __init__(
+        self,
+        timstamp_column: str,
+        granularity_check_func: Optional[Callable[[Query], Any]],
+    ) -> None:
+        super(SessionsTimeSeriesExtensionProcessor, self).__init__(timstamp_column)
+        if granularity_check_func is not None:
```
evanh, line 134: Why is this optional? The process_query function will fail if this is None.

snuba/datasets/entities/sessions.py
```diff
@@ -304,3 +316,14 @@ def get_query_processors(self) -> Sequence[QueryProcessor]:
...
         ]
+
+
+def check_if_crash_rate_alert_subscription(query: Query) -> bool:
+    """
+    Checks if the provided query is for a Crash Rate Alert subscription based on the if one of
+    the aggregations matches the CRASH_RATE_ALERT_PATTERN
+    """
+    for col in query.get_selected_columns():
+        if col.name and bool(CRASH_RATE_ALERT_PATTERN.match(col.name)):
```
evanh, line 327: Relying on a specific alias to make this work seems extremely brittle.

snuba/query/timeseries_extension.py
```diff
@@ -122,3 +122,48 @@ def __init__(
             ),
             processor=TimeSeriesExtensionProcessor(timestamp_column),
         )
+
+
+class SessionsTimeSeriesExtensionProcessor(TimeSeriesExtensionProcessor):
+    def __init__(
+        self,
+        timstamp_column: str,
```
evanh, line 130:
```suggestion
        timestamp_column: str,
```

snuba/query/timeseries_extension.py
```diff
@@ -122,3 +122,48 @@ def __init__(
             ),
             processor=TimeSeriesExtensionProcessor(timestamp_column),
         )
+
+
+class SessionsTimeSeriesExtensionProcessor(TimeSeriesExtensionProcessor):
```
evanh, line 127: Again extensions are deprecated, so you'll need to subclass `TimeSeriesProcessor` (the query processor) and move this logic there.

snuba/subscriptions/data.py
evanh: I wouldn't bother with updating anything Legacy related. It is not used in prod at all and the only reason it's not outright deleted is to support onprem/oss that are not migrated yet.

snuba/subscriptions/data.py
```diff
@@ -230,18 +272,20 @@ def add_conditions(
             ),
         ]
```
evanh, line 274: You'll need to add an organization condition here as well.

evanh (commented):
I'm not sure I understand why granularity is even used. The Subscription doesn't store the query granularity anywhere. It returns aggregate values for a single span of time. Are subscriptions supposed to return multiple rows now?

snuba/query/timeseries_extension.py
ahmedetefy: I don't understand what you mean by `granularity` field is not a source of truth

I need the granularity field because, for `time_window` <= 1h, I want to use the raw sessions table and for >1h I want to use the materialized.  I extract it in the Processor because I need the time window to decide on that and I don't have the time_window in the `select_storage` function

ahmedetefy: >Why is this optional? The process_query function will fail if this is None.

Sure I will add a noop

snuba/query/timeseries_extension.py
ahmedetefy: Thanks for letting me know

snuba/datasets/entities/sessions.py
ahmedetefy: Will do

snuba/subscriptions/data.py
ahmedetefy: > I wouldn't bother with updating anything Legacy related. It is not used in prod at all and the only reason it's not outright deleted is to support onprem/oss that are not migrated yet

There is initial validation that will fail if I don't update the legacy stuff

snuba/datasets/entities/sessions.py
ahmedetefy: I agree.. I will think of a better solution

evanh:
> I need the granularity field because, for time_window <= 1h, I want to use the raw sessions table and for >1h I want to use the materialized. I extract it in the Processor because I need the time window to decide on that and I don't have the time_window in the select_storage function

@ahmedetefy 

The granularity of a query determines how we group the rows returned (e.g. one row for each hour/day etc.). Subscriptions only return one row so this is not applicable. My point is that you are using granularity as a flag to change the storage selection logic for subscriptions and I don't think that's correct.

What I would actually do is in the storage selector, check the type of `RequestSettings` since that will be a `SubscriptionRequestSettings` for subscriptions. Then you can extract the time window using `snuba.clickhouse.query_dsl.accessors.get_time_range` and select the appropriate storage using that information.

snuba/subscriptions/data.py
lynnagara: It was more about using an alias at all, it is not a reliable method of detecting a subscription even if we made it a more obscure name, we should find another way.

snuba/subscriptions/data.py
evanh: Extensions are deprecated. The `SnQLSubscriptionData` needs to check for an organization ID and add a condition for it, in the same way that it adds a project_id.

snuba/subscriptions/data.py
```diff
@@ -299,21 +341,29 @@ def from_dict(cls, data: Mapping[str, Any]) -> SnQLSubscriptionData:
...
-            resolution=timedelta(seconds=data["resolution"]),
-            query=data["query"],
-        )
+        snql_subs_dict = {
+            "project_id": data["project_id"],
+            "time_window": timedelta(seconds=data["time_window"]),
+            "resolution": timedelta(seconds=data["resolution"]),
+            "query": data["query"],
+        }
+        organization = data.get("organization")
```
evanh, line 350: Don't need a separate `if` statement for this.

snuba/subscriptions/data.py
```diff
@@ -229,19 +261,29 @@ def add_conditions(
...
+            conditions_to_add += [
+                binary_condition(
+                    ConditionFunctions.EQ,
+                    Column("_snuba_org_id", None, "org_id"),
+                    Literal(None, self.organization),
+                ),
+            ]
 
+        # This check is necessary because the `sessions` dataset does not have an `offset` column
         if offset is not None:
```
evanh, line 274: Nit:
```suggestion
        if offset is not None and "offset" in entity.get_data_model().get_column_names():
```

snuba/datasets/entities/sessions.py
```diff
@@ -211,8 +212,24 @@ def __init__(self) -> None:
...
+        # If the passed in `request_settings` arg is an instance of `SubscriptionRequestSettings`,
+        # then it is a crash rate alert subscription, and hence we decide on whether to use the
+        # materialized storage or the raw storage by examining the time_window.
+        # If the `time_window` <=1h, then select the raw storage otherwise select materialized
+        # storage
+        # NOTE: If we were to support other types of subscriptions over the sessions dataset that
+        # do not follow this method used to identify which storage to use, we would need to
+        # match functions used with a regex.
+        if isinstance(request_settings, SubscriptionRequestSettings):
+            from_date, to_date = get_time_range(query, "started")  # type: ignore
```
evanh, line 226: Why did you add `# type: ignore`?

snuba/clickhouse/columns.py
```diff
@@ -539,6 +540,12 @@ def get(
         except KeyError:
             return default
 
+    def get_column_names(self) -> Set[str]:
```
evanh, line 543: You don't need this, the class already has a `__contains__` function. You can do `"offset" in <ColumnSet>` directly. Please remove this.

snuba/datasets/entities/sessions.py
```diff
@@ -211,8 +212,24 @@ def __init__(self) -> None:
...
+        use_materialized_storage: Optional[bool] = None
+
+        # If the passed in `request_settings` arg is an instance of `SubscriptionRequestSettings`,
+        # then it is a crash rate alert subscription, and hence we decide on whether to use the
+        # materialized storage or the raw storage by examining the time_window.
+        # If the `time_window` <=1h, then select the raw storage otherwise select materialized
+        # storage
+        # NOTE: If we were to support other types of subscriptions over the sessions dataset that
+        # do not follow this method used to identify which storage to use, we would need to
+        # match functions used with a regex.
```
evanh, line 224:
```suggestion
        # find a different way to distinguish them.
```

snuba/subscriptions/data.py
```diff
@@ -173,27 +197,35 @@ def from_dict(cls, data: Mapping[str, Any]) -> LegacySubscriptionData:
...
-            resolution=timedelta(seconds=data["resolution"]),
-        )
+        legacy_subs_dict = {
+            "project_id": data["project_id"],
+            "conditions": data["conditions"],
+            "aggregations": data["aggregations"],
+            "time_window": timedelta(seconds=data["time_window"]),
+            "resolution": timedelta(seconds=data["resolution"]),
+        }
+        organization = data.get("organization")
```
evanh, line 207: I don't think you need add `organization` separately. You can add it to the original call.
```
LegacySubscriptionData(
    project_id=data["project_id"],
    conditions=data["conditions"],
    aggregations=data["aggregations"],
    time_window=timedelta(seconds=data["time_window"]),
    resolution=timedelta(seconds=data["resolution"]),
    organization=data.get("organization),
)
```

This is true of all the times you add `organization` in a separate `if`.

snuba/subscriptions/data.py
```diff
@@ -173,27 +197,35 @@ def from_dict(cls, data: Mapping[str, Any]) -> LegacySubscriptionData:
...
     def to_dict(self) -> Mapping[str, Any]:
-        return {
+        legacy_subs_dict = {
             "project_id": self.project_id,
             "conditions": self.conditions,
             "aggregations": self.aggregations,
             "time_window": int(self.time_window.total_seconds()),
             "resolution": int(self.resolution.total_seconds()),
         }
+        if self.organization:
```
evanh, line 220: Don't need a separate `if` statement for this.

snuba/subscriptions/data.py
```diff
@@ -299,21 +341,29 @@ def from_dict(cls, data: Mapping[str, Any]) -> SnQLSubscriptionData:
...
     def to_dict(self) -> Mapping[str, Any]:
-        return {
+        snql_subs_dict = {
             self.TYPE_FIELD: SubscriptionType.SNQL.value,
             "project_id": self.project_id,
             "time_window": int(self.time_window.total_seconds()),
             "resolution": int(self.resolution.total_seconds()),
             "query": self.query,
         }
+        if self.organization:
```
evanh, line 364: Don't need a separate `if` statement for this.

snuba/subscriptions/data.py
```diff
@@ -372,17 +425,23 @@ def from_dict(cls, data: Mapping[str, Any]) -> DelegateSubscriptionData:
...
+        delegate_subs_dict = {
+            "project_id": data["project_id"],
+            "time_window": timedelta(seconds=data["time_window"]),
+            "resolution": timedelta(seconds=data["resolution"]),
+            "conditions": data["conditions"],
+            "aggregations": data["aggregations"],
+            "query": data["query"],
+        }
+
+        organization = data.get("organization")
```
evanh, line 437: Don't need a separate `if` statement for this.

snuba/subscriptions/data.py
```diff
@@ -392,6 +451,10 @@ def to_dict(self) -> Mapping[str, Any]:
             "query": self.query,
         }
 
+        if self.organization:
```
evanh, line 454: Don't need a separate `if` statement for this.

tests/subscriptions/test_codecs.py
```diff
@@ -28,53 +28,79 @@
...
-    )
+def build_legacy_subscription_data(organization=None) -> LegacySubscriptionData:
+    legacy_subs_dict = {
+        "project_id": 5,
+        "conditions": [["platform", "IN", ["a"]]],
+        "aggregations": [["count()", "", "count"]],
+        "time_window": timedelta(minutes=500),
+        "resolution": timedelta(minutes=1),
+    }
+    if organization:
```
evanh, line 39: Don't need a separate `if` statement for this.

tests/subscriptions/test_codecs.py
```diff
@@ -28,53 +28,79 @@
...
-        query="MATCH events SELECT count() WHERE in(platform, 'a')",
-    )
+def build_snql_subscription_data(organization=None) -> SnQLSubscriptionData:
+    snql_subs_dict = {
+        "project_id": 5,
+        "time_window": timedelta(minutes=500),
+        "resolution": timedelta(minutes=1),
+        "query": "MATCH events SELECT count() WHERE in(platform, 'a')",
+    }
+    if organization:
```
evanh, line 51: Don't need a separate `if` statement for this.

tests/subscriptions/test_codecs.py
```diff
@@ -28,53 +28,79 @@
...
+def build_delegate_subscription_data(organization=None) -> DelegateSubscriptionData:
+    delegate_subs_dict = {
+        "project_id": 5,
+        "time_window": timedelta(minutes=500),
+        "resolution": timedelta(minutes=1),
+        "conditions": [["platform", "IN", ["a"]]],
+        "aggregations": [["count()", "", "count"]],
+        "query": "MATCH events SELECT count() WHERE in(platform, 'a')",
+    }
+    if organization:
```
evanh, line 65: Don't need a separate `if` statement for this.

tests/subscriptions/test_codecs.py
```diff
@@ -84,10 +110,17 @@ def test_encode() -> None:
     assert data["time_window"] == int(subscription.time_window.total_seconds())
     assert data["resolution"] == int(subscription.resolution.total_seconds())
 
+    if organization:
+        assert data["organization"] == subscription.organization
```
evanh, line 114: You should be able to assert this regardless of `organization`, since both values should be `None`.

tests/subscriptions/test_data.py
```diff
@@ -91,29 +94,103 @@
...
+            resolution=timedelta(minutes=1),
+        ),
+        JsonSchemaValidationException,
+        id="No organization provided for sessions dataset subscription",
+    ),
+]
+
 
-class TestBuildRequest(BaseSubscriptionTest):
+class TestBuildRequestBase:
```
evanh, line 146: Why did you change the class inheritance here?

tests/test_sessions_api.py
```diff
@@ -74,7 +49,7 @@ def generate_manual_session_events(self, project_id: int) -> None:
                 "session_id": session_1,
                 "distinct_id": user_1,
                 "seq": 123,
-                "status": 1,
+                "status": 2,
```
evanh, line 52: Why did this change?

evanh (commented):
Looking better. Still something to address.

tests/test_sessions_api.py
ahmedetefy: Oh this is just me changing the dummy data to have one crashed session (status=2) as well just to ensure crash_free_percentage is calculated as expected.. I could change it back if it shouldn't be there?

snuba/datasets/entities/sessions.py
ahmedetefy: mypy was failing because this query is not an instance of `ProcessableQuery[Table]` .. I could alternatively cast the type if it makes more sense?

tests/subscriptions/test_data.py
ahmedetefy: Because `BaseSubscriptionTest` basically creates a bunch of dummy events in `setup_method` which I really don't need for the the crash rate alerts tests but I need functions like `compare_conditions`, and so I just subclass `BaseSubscriptionTest` when I am testing for events

snuba/subscriptions/data.py
```diff
@@ -150,16 +151,26 @@ def build_request(
...
+            extra_conditions = [[["ifnull", ["offset", 0]], "<=", offset]]
+
+        build_request_dict = {
+            "project": self.project_id,
+            "conditions": [*self.conditions, *extra_conditions],
+            "aggregations": self.aggregations,
+            "from_date": (timestamp - self.time_window).isoformat(),
+            "to_date": timestamp.isoformat(),
+        }
+        if self.organization:
```
ahmedetefy, line 170: @evanh if I remove this I get a `JsonValidationError` for entities that are not sessions

snuba/datasets/entities/sessions.py
lynnagara: We shouldn't ignore or cast it to the wrong type. `get_time_range` currently expects to be called with `ProcessableQuery[Table]` which is a type of `Query` later in the processing pipeline. If that function is expected to work with any kind of query we can change the type of the `get_time_range` function to allow that.

snuba/subscriptions/data.py
lynnagara, line 170: There is a lot of dataset specific code here. `self.organization` is only relevant to sessions and it should not be in the shared subscriptions code.

I think we should just skip providing this LegacySubscriptionData implementation altogether if we can as we will not have any legacy items for sessions and this needs to be deprecated anyway.

snuba/subscriptions/data.py
```diff
@@ -229,8 +243,17 @@ def add_conditions(
                 Literal(None, timestamp),
             ),
         ]
-
-        if offset is not None:
+        if self.organization:
+            conditions_to_add += [
+                binary_condition(
+                    ConditionFunctions.EQ,
+                    Column("_snuba_org_id", None, "org_id"),
```
lynnagara, line 250: Why did you assign the `_snuba_org_id` alias?

snuba/subscriptions/data.py
```diff
@@ -229,8 +243,17 @@ def add_conditions(
                 Literal(None, timestamp),
             ),
         ]
-
-        if offset is not None:
+        if self.organization:
```
lynnagara, line 246: I know this is something that was already like this before your PR but personally I don't think we should be putting code that can handle every single dataset here. It will get more convoluted as we add more datasets. If building the SubscriptionData is specific for each entity, then why not  define a function on the entity itself that knows how to build this and just call the relevant function for the entity we are concerned with from here.

snuba/subscriptions/data.py
ahmedetefy: So we can't currently because there is still some request validation that would fail. However, @evanh told me that 
1- We are not using `LegacySubscriptionData` in production anyways 
2- He will be removing class entirely soon enough 

So I suggest I could either not add anything at all to the `LegacySubscriptionData` class to support sessions and just remove the request validation, or just leave these changes as is and when @evanh deletes this class, they will be removed anyways

snuba/subscriptions/data.py
ahmedetefy: I basically mirrored this https://github.com/getsentry/snuba/blob/5390cf11ef086385be696e54d10d604a6a62de0a/snuba/query/organization_extension.py#L27

ahmedetefy:
@lynnagara Thanks for your review 
I have addressed your comments in the following ways:
- I have removed any changes to `LegacySubscriptionData` class, and hence I had to remove the request validation for it so that it does not error https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/subscription.py#L36 
I think this should be fine because afaik @evanh will be removing this class soon enough anyways and it is not used in production atm
- I followed your suggestion by extracting dataset specific logic into the entity class and just called it from within data subscription class 
- I have changed the type of `get_time_range` to be `Union[ProcessableQuery[Entity], ProcessableQuery[Table]]` rather than only `ProcessableQuery[Table]`

snuba/clickhouse/query_dsl/accessors.py
```diff
@@ -134,7 +134,8 @@ def get_time_range_expressions(
 
 
 def get_time_range(
-    query: ProcessableQuery[Table], timestamp_field: str
+    query: Union[ProcessableQuery[Table], ProcessableQuery[Entity]],
```
evanh, line 137: I don't think mypy runs on these feature branches, so I would run it manually to make sure this works.

snuba/datasets/entity.py
```diff
@@ -149,3 +150,13 @@ def describe(self) -> Description:
                 Description(header="Relationships", content=relationships),
             ],
         )
+
+    @staticmethod
+    def get_extra_conditions_for_subscriptions_query(
```
evanh, line 155: You're messing up bounded contexts here. Entities are meant to not have any knowledge of how they're used, subscriptions vs. queries etc. So having subscription specific functions in the entity is not correct. Subscription logic should stay in the subscription code.

snuba/subscriptions/subscription.py
```diff
@@ -33,7 +33,8 @@ def create(self, data: SubscriptionData, timer: Timer) -> SubscriptionIdentifier
         # If there is a delegate subscription, we need to run both the SnQL and Legacy validator
         if isinstance(data, DelegateSubscriptionData):
             self._test_request(data.to_snql(), timer)
-            self._test_request(data.to_legacy(), timer)
+            # Commented this out because `LegacySubscriptionData` is deprecated
+            # self._test_request(data.to_legacy(), timer)
```
evanh, line 37: What fails if you leave this in? `LegacySubscriptionData` is deprecated for us (Sentry) but not for OSS or Onprem users. If we remove this now we might let those customers put their alerts into a bad state.

tests/test_sessions_api.py
evanh: Just making sure this was a deliberate change.

evanh (changes requested):
Entities should not have subscription code in them. See my comments.

snuba/clickhouse/query_dsl/accessors.py
ahmedetefy: It does, it is a pre-commit hook

snuba/subscriptions/subscription.py
ahmedetefy: It fails if I am trying to build a sessions query without an `organization` param

snuba/subscriptions/data.py
evanh: I don't think we want to encode knowledge in the entity about how it's being used (subscriptions vs regular queries). Also I don't think this is dataset specific, if any dataset sends us an organization we will correctly add it to the conditions. That doesn't seem wrong.

lynnagara: This is why I think it's dataset specific:

- `organization_id` is relevant for sessions but not for events or transactions
- `offset` is the same, relevant for events and transactions not sessoins
- Today `project_id` is the same on the datasets but it seems plausible that some datasets support cross product queries in future and not others

Yes there are a few fields that might overlap but ultimately this just seems a co-incidence to me and not something that we should enforce. Basically every time we need to add a subscription for a dataset with slightly different fields we would need to change the function that gets run on *every* dataset rather than the part only specific to that kind of subscription and that doesn't seem right to me.

snuba/clickhouse/query_dsl/accessors.py
evanh: The pre-commit hook I think only runs on the changed files. I would run it manually on the whole codebase just in case.

snuba/subscriptions/dataset_subscription.py
```diff
@@ -0,0 +1,98 @@
+from typing import Any, List, Mapping, Optional, Type, Union
+
+from snuba.query.conditions import ConditionFunctions, binary_condition
+from snuba.query.exceptions import InvalidQueryException
+from snuba.query.expressions import Column, FunctionCall, Literal
+
+
+class DatasetSubscription:
```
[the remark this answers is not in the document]
lynnagara: Can we make this an abstract base class? I don't think we want anyone to use it directly.

snuba/subscriptions/dataset_subscription.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    Function that returns the correct DatasetSubscription class based on the provided dictionary
+    of data
+    """
+    # At the moment checking if organization is in the data_dict suffices to identify that
+    #  SessionsSubscription class should be returned. However, If we were to extend the datasets
+    #  that subscriptions support, we might need a better approach of checking.
+    return (
+        SessionsSubscription
+        if "organization" in data_dict
```
lynnagara, line 96: You shouldn't determine the dataset from the presence of some key in the data dictionary. We already know the entity, why not pass that in here?

ahmedetefy: We don’t have an entity actually at that point 
So the way to do would be to extract it from the query 
Why is it not a good idea to check for organization .. I validate that the organization exists in the init function for sessions class ?

ahmedetefy: @lynnagara that key is only present on creation but then when the worker is going through the subscriptions, the key is no longer there

lynnagara: It is not a good idea to check for organization because in future we will have other subscriptions with organization like metrics. Then we will need a new way to tell the subscriptions apart.

I think we always have the dataset at least when using the subscription codec like here https://github.com/getsentry/snuba/blob/master/snuba/web/views.py#L479-L480. How about passing it to the subscription codec which can then pass it to the SuscriptionData.from_dict method?

snuba/subscriptions/data.py
lynnagara: Do you know what's involved in turning off that request validation? (at least for this dataset). If it's straightforward enough it may be worth just doing that rather than providing an implementation that won't be used. It may be worth it since it seems we would need to do that eventually anyway.

snuba/subscriptions/data.py
lynnagara: Hmm, I think that's specific to extensions because of where they happen in the query pipeline (after aliases assigned). I don't think you need to manually assign any alias here, just leave it as None.

snuba/subscriptions/dataset_subscription.py
```diff
@@ -0,0 +1,98 @@
+from typing import Any, List, Mapping, Optional, Type, Union
+
+from snuba.query.conditions import ConditionFunctions, binary_condition
+from snuba.query.exceptions import InvalidQueryException
+from snuba.query.expressions import Column, FunctionCall, Literal
+
+
+class DatasetSubscription:
+    def __init__(self, subscription_type: str, data_dict: Mapping[str, Any]) -> None:
```
lynnagara, line 9: Ideally we don't support legacy at all, but if it's unavoidable we should probably use the enum here rather than a string

ahmedetefy: I don't think we can at the moment.. as far as i know some on prem customers still use it, and so turning off the validation for legacy would permit customers creating faulty subscriptions

snuba/subscriptions/dataset_subscription.py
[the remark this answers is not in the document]
ahmedetefy: Sure

lynnagara (commented):
Are you planning to merge this PR into master? It is currently based on another branch. Wondering what is the plan for that.

snuba/datasets/entities/sessions.py
```diff
@@ -211,8 +212,24 @@ def __init__(self) -> None:
...
+        # then it is a crash rate alert subscription, and hence we decide on whether to use the
+        # materialized storage or the raw storage by examining the time_window.
+        # If the `time_window` <=1h, then select the raw storage otherwise select materialized
+        # storage
+        # NOTE: If we were to support other types of subscriptions over the sessions dataset that
+        # do not follow this method used to identify which storage to use, we would need to
+        # find a different way to distinguish them.
+        if isinstance(request_settings, SubscriptionRequestSettings):
+            from_date, to_date = get_time_range(query, "started")
+            if from_date and to_date:
```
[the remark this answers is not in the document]
ahmedetefy: No, because we will just fall to the logic we already have which will set granularity to 3600 i.e. materialised storage

snuba/datasets/entities/sessions.py
```diff
@@ -211,8 +212,24 @@ def __init__(self) -> None:
...
+                use_materialized_storage = to_date - from_date > timedelta(hours=1)
```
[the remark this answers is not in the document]
ahmedetefy: Yes, and we are okay with that .. its a tradeoff decision made for in favor of running the subscription logic every minute

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+                    Literal(None, offset),
+                )
+            )
+        elif self.subscription_type == SubscriptionType.LEGACY:
+            conditions_to_add = [[["ifnull", ["offset", 0]], "<=", offset]]
+        return conditions_to_add
+
+
+class TransactionsSubscription(EventsSubscription):
+    ...
```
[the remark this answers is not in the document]
ahmedetefy: Why not? you said that the transactions and events subscription logic will change and for me logically it makes sense to separate these classes

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+    ) -> None:
+        self.subscription_type = subscription_type
+
+    def get_entity_subscription_conditions(
+        self, offset: Optional[int] = None
+    ) -> Sequence[Union[Expression, Condition]]:
+        return []
+
+    def to_dict(self) -> Mapping[str, Any]:
+        return {}
```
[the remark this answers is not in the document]
ahmedetefy: Why though? I don't want this function to be mandatory and I want this to be the default behaviour if I don't define the function

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+class EntitySubscription(ABC):
+    def __init__(
+        self, subscription_type: SubscriptionType, data_dict: Mapping[str, Any]
+    ) -> None:
+        self.subscription_type = subscription_type
+
+    def get_entity_subscription_conditions(
+        self, offset: Optional[int] = None
+    ) -> Sequence[Union[Expression, Condition]]:
+        return []
```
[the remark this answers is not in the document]
ahmedetefy: Same as previous comment, I'd rather keep the default

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def to_dict(self) -> Mapping[str, Any]:
+        return {}
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, EntitySubscription):
+            return NotImplemented
+        return self.to_dict() == other.to_dict() and isinstance(other, type(self))
```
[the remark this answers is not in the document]
ahmedetefy: Its necessary in case I want to compare subscriptions .. subscriptions could be practically equivalent but won't be `==` because the entity_subscriptions defined represent different objects

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+
+class EntitySubscription(ABC):
+    def __init__(
+        self, subscription_type: SubscriptionType, data_dict: Mapping[str, Any]
+    ) -> None:
+        self.subscription_type = subscription_type
+
+    def get_entity_subscription_conditions(
+        self, offset: Optional[int] = None
+    ) -> Sequence[Union[Expression, Condition]]:
```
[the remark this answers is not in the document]
ahmedetefy: Sure I think that's a good idea

ahmedetefy:
>Are you planning to merge this PR into master? It is currently based on another branch. Wondering what is the plan for that.

Yeah my plan is initially to merge into the feature branch because it has a small PR that I would like to merge together with this, and then to merge into master

snuba/subscriptions/codecs.py
```diff
@@ -2,20 +2,28 @@
...
 from snuba.subscriptions.worker import SubscriptionTaskResult
 from snuba.utils.codecs import Codec, Encoder
 
 
 class SubscriptionDataCodec(Codec[bytes, SubscriptionData]):
+    dataset_name: str
+
+    def __init__(self, dataset: Dataset):
+        super().__init__()
+        self.dataset_name = get_dataset_name(dataset)
```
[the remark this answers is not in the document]
ahmedetefy: I am not sure why I would want to pass the `Entity` itself, all I need is actually the EntityKey but I am not sure how to get that

snuba/datasets/entities/sessions.py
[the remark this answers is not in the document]
lynnagara: It looks like it would only fall to materialised storage if `extract_granularity_from_query(query, "started")` is None or a multiple of 3600. What value are we expecting to get there for subscriptions?

snuba/datasets/entities/sessions.py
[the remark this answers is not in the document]
lynnagara: Ok, that sounds fine. I guess I'm confused why it's still possible to fall back to the non materialised storage later if either `to_date` or `from_date` is missing here then.

snuba/subscriptions/codecs.py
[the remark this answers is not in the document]
lynnagara: You can get the entity key from the entity from this mapping https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/factory.py#L12

snuba/subscriptions/entity_subscription.py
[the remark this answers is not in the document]
lynnagara: I'm actually not currently aware of any need for these to diverge in the future.

I think this pattern of inheritance is unexpected. If you really do want to keep them separate I would suggest having a `BaseEventsSubscription` abstract class and `EventsSubscription` and `TransactionSubscription` classes that extend that.

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+                    FunctionCall(
+                        None,
+                        "ifNull",
+                        (Column(None, None, "offset"), Literal(None, 0)),
+                    ),
+                    Literal(None, offset),
+                )
+            )
+        elif self.subscription_type == SubscriptionType.LEGACY:
+            conditions_to_add = [[["ifnull", ["offset", 0]], "<=", offset]]
```
lynnagara, line 99: Please use `ifNull` rather than `ifnull`. This list of functions will be enforced soon https://github.com/getsentry/snuba/blob/master/snuba/query/functions.py#L173

tests/subscriptions/test_entity_subscriptions.py
```diff
@@ -0,0 +1,162 @@
...
+                ),
+                Literal(None, 5),
+            )
+        ],
+        True,
+        id="Events subscription with offset of type SNQL",
+    ),
+    pytest.param(
+        EventsSubscription(subscription_type=SubscriptionType.LEGACY, data_dict={}),
+        [[["ifnull", ["offset", 0]], "<=", 5]],
```
lynnagara, line 61:
```suggestion
        [[["ifNull", ["offset", 0]], "<=", 5]],
```

tests/subscriptions/test_entity_subscriptions.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+        ],
+        True,
+        id="Transactions subscription with offset of type SNQL",
+    ),
+    pytest.param(
+        TransactionsSubscription(
+            subscription_type=SubscriptionType.LEGACY, data_dict={}
+        ),
+        [[["ifnull", ["offset", 0]], "<=", 5]],
```
lynnagara, line 95:
```suggestion
        [[["ifNull", ["offset", 0]], "<=", 5]],
```

snuba/subscriptions/entity_subscription.py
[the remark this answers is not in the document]
lynnagara: I guess it just seems unlikely to me that in practice we will have any entities which actually have that default behaviour, which defeats the point of making it default. To me, the downside of the default is that it makes it less clear to the next person providing a new EntitySubscription subclass whether they need to provide an implementation or use the default, and it's easier to just forget to provide it since it won't throw an error.

snuba/subscriptions/entity_subscription.py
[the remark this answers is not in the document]
lynnagara: You mention `in case`. Are you actually using this today?

snuba/datasets/entities/sessions.py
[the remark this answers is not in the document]
ahmedetefy: It should never actually not have `from_date` and `to_date` in the subscription queries, I just added the check as a safe guard but if it does fail for whatever reason and fallsback to the old logic then `extract_granularity_from_query(query, "started")` will always return `None` because we don't group by any column (time column in this case) and there is validation against any grouping logic actually for subscription queries.

snuba/datasets/entities/sessions.py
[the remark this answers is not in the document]
ahmedetefy: It is not possible in subscription queries because there is no group by time column so if it doesn't actually go into line 228, it will always use the materialised view

snuba/subscriptions/entity_subscription.py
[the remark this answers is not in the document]
ahmedetefy: Yes but only in tests, it is genuinely a pita if the equality function is not there because all tests basically compare some form of subscription objects or subscription objects embedded within task objects, and I have tried to create some test utility function to compare subscription objects but I would need to create one of these for every kind of object that has a subscription in it, which is why I ended up just adding this `__eq__` function to the class.
However, if you have strong opinions against adding this `__eq__` function when it is not required in a non-test environment, I can always just monkey patch the `EntitySubscription` class in tests to have a `__eq__` function

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,133 @@
...
+    entity_subscription_type: Type[EntitySubscription],
+) -> str:
+    return ENTITY_SUBSCRIPTION_TO_KEY_MAPPER[entity_subscription_type].value
+
+
+ENTITY_SUBSCRIPTION_TO_KEY_MAPPER: Mapping[Type[EntitySubscription], EntityKey] = {
+    SessionsSubscription: EntityKey.SESSIONS,
+    EventsSubscription: EntityKey.EVENTS,
+    TransactionsSubscription: EntityKey.TRANSACTIONS,
+}
```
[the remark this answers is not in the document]
ahmedetefy: To be honest, I do not want to store it in the class but I modified so that the reverse mapping is generated from the original mapping so that we wouldn't have to edit both mappings everytime we want to support a new entity subscription .. hope that is okay

snuba/subscriptions/entity_subscription.py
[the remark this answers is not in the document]
ahmedetefy: Actually scratch that, I couldn't justify adding the `__eq__` just for tests so I just monkey patched the `EntitySubscription` class in tests

ahmedetefy:
@lynnagara 
So an update on this PR:
- I changed the base branch here to master
- I would like to merge this new relax aggregation validation PR https://github.com/getsentry/snuba/pull/2112 into this branch 
- Merge this into master

snuba/datasets/entities/sessions.py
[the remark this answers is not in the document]
lynnagara: Ok, in that case I think you can rewrite in a way that makes it clearer that the fallback is always materialised and we don't attempt to look for non existent "started" value. Like:
```
        if isinstance(request_settings, SubscriptionRequestSettings):
            from_date, to_date = get_time_range(query, "started")
            if from_date and to_date:
                use_materialized_storage = to_date - from_date > timedelta(hours=1)
            else:
                use_materialized_storage = True
        else:
            granularity = extract_granularity_from_query(query, "started") or 3600
            use_materialized_storage = granularity >= 3600 and (granularity % 3600) == 0
```

However, I would question whether we should be falling back at all. If there is no `from_date` or `to_date` wouldn't we be better off raising an invalid subscription rather than attempting to run a broken query. I think it is better to always reject a incorrectly formatted subscription than try to make it valid which may cause us problems down the line.

snuba/subscriptions/codecs.py
```diff
@@ -2,20 +2,27 @@
...
     SubscriptionData,
-    SubscriptionType,
 )
+from snuba.subscriptions.entity_subscription import SubscriptionType
 from snuba.subscriptions.worker import SubscriptionTaskResult
 from snuba.utils.codecs import Codec, Encoder
 
 
 class SubscriptionDataCodec(Codec[bytes, SubscriptionData]):
+    dataset_name: str
```
lynnagara, line 21: Was this left over from a previous implementation? I don't think it's needed anymore

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,122 @@
...
+            return []
+
+        return [[["ifNull", ["offset", 0]], "<=", offset]]
+
+    def to_dict(self) -> Mapping[str, Any]:
+        return {}
+
+
+class EventsSubscription(BaseEventsSubscription):
+    ...
```
lynnagara, line 107: I am not familiar with ellipsis in this context. I've always seen this written as `pass`. Do you know what is the difference?

snuba/subscriptions/codecs.py
ahmedetefy: oops missed that

snuba/subscriptions/entity_subscription.py
ahmedetefy: Yeah its just a built in singleton object. A cuter way of writing `pass` imo
Its just that statement needs to be there after the function def if no logic will be defined

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,122 @@
...
+
+class SubscriptionType(Enum):
+    LEGACY = "legacy"
+    SNQL = "snql"
+    DELEGATE = "delegate"
+
+
+class EntitySubscription(ABC):
+    def __init__(self, data_dict: Mapping[str, Any]) -> None:
+        ...
```
lynnagara, line 20: This isn't necessary

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,122 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def to_dict(self) -> Mapping[str, Any]:
+        raise NotImplementedError
+
+
+class SessionsSubscription(EntitySubscription):
+    def __init__(self, data_dict: Mapping[str, Any]) -> None:
+        super().__init__(data_dict)
```
lynnagara, line 49: No need for this

ahmedetefy: the linter complains if I don't define this

snuba/subscriptions/entity_subscription.py
ahmedetefy: It actually is though, because I will get an error if I do for example `EventsSubscription(data_dict={})` because it will say unexpected keyword

snuba/subscriptions/entity_subscription.py
lynnagara: This needs to be applied together with this other comment https://github.com/getsentry/snuba/pull/2094#discussion_r715083833

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,122 @@
...
+                ConditionFunctions.EQ,
+                Column(None, None, "org_id"),
+                Literal(None, self.organization),
+            ),
+        ]
+
+    def get_entity_subscription_conditions_for_legacy(
+        self, offset: Optional[int] = None
+    ) -> Sequence[Condition]:
+        return []
```
lynnagara, line 71: How come you don't provide the equivalent org_id condition here that the SnQL one has? Aren't they always supposed to match? Or is it because we don't want to support legacy on this entity?

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def to_dict(self) -> Mapping[str, Any]:
+        return {"organization": self.organization}
```
lynnagara, line 74: Where does this key `organization` end up getting used? I noticed we use the name `project_id` for the project ID. Should we follow a convention?

ahmedetefy: It will fail because I think there is still an extension there that validates on `organization`

snuba/subscriptions/entity_subscription.py
ahmedetefy: For legacy, it is already there in the conditions sent from sentry

snuba/subscriptions/entity_subscription.py
ahmedetefy: I cannot remove it because of this https://github.com/getsentry/snuba/pull/2094#discussion_r715089069

snuba/subscriptions/entity_subscription.py
lynnagara: Why would only legacy send it in the conditions from Sentry not SnQL? That seems fairly inconsistent. I think it's a bit dangerous for Snuba to not apply it here and just assume that the condition got added by the client. What if someone tries to make a subscription on multiple orgs? Applying the condition here would help prevent that.

snuba/subscriptions/entity_subscription.py
lynnagara: What does this have to do with organization extensions though? Isn't it just the key used for encoding and decoding the subscription data in redis?

snuba/subscriptions/entity_subscription.py
lynnagara: Ah, forgot you're always passing the dict to the constructor

snuba/subscriptions/entity_subscription.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class BaseEventsSubscription(EntitySubscription, ABC):
+    def get_entity_subscription_conditions_for_snql(
+        self, offset: Optional[int] = None
+    ) -> List[Expression]:
```
lynnagara, line 80:
```suggestion
    ) -> Sequence[Expression]:
```

ahmedetefy: I don't think that's a good idea because you can't really use th `+` with `Sequence` here -> https://github.com/getsentry/snuba/blob/3ea9573691182f29541b15b8c42c558827bfe775/snuba/subscriptions/data.py#L246

snuba/subscriptions/entity_subscription.py
ahmedetefy: Not really because there is extension validation that runs and validates against the `organization` param in Session entity right over here -> https://github.com/getsentry/snuba/blob/master/snuba/query/organization_extension.py#L8

snuba/datasets/entities/sessions.py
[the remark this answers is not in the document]
ahmedetefy: >However, I would question whether we should be falling back at all. If there is no from_date or to_date wouldn't we be better off raising an invalid subscription/query error rather than attempting to run a broken query

I don't agree here because there is a required `time_window` param that is passed to Subscription classes, and from that the `from_date`, and `to_date` are generated so if we get to this point, we will always have `from_date`, and `to_date`. However, there is something to be said for the readability of this code segment and not having to go through the extracting the granularity from time columns that we know don't exist in a subscription query, so I will use your snippet

snuba/subscriptions/entity_subscription.py
ahmedetefy: Because `SnqlSubscription` class does not require `conditions` so basically sentry sends all the parameters required for `SnqlSubscription` and `LegacySubscription` through `DelegateSubscription` class which then basically routes to the class instance whether for Snql or Legacy, and so Legacy requires `conditions` and `aggregations` while Snql just requires a query basically.. so on Sentry side we send those conditions, but we also construct a Snql query from them and send them to snuba .. so if you look at the LegacySubscription class conditions, the organization condition will already be there as it is sent from sentry

snuba/subscriptions/entity_subscription.py
lynnagara: Hmm I see. How about:
```
conditions_to_add = [
  condition1,
  condiiton2,
  *self.entity_subscription.get_entity_subscription_conditions_for_snql(offset)
]
```
That is more similar to what you do above for the legacy subscription data version and I think there is no reason to give snql and legacy different types just because of using splat operator in one place and += in the other.

snuba/clickhouse/query_dsl/accessors.py
lynnagara, line 137: Do you need this change anymore? I think we ended up moving the storage selection to later in the pipeline where we already have a table.

snuba/subscriptions/data.py
```diff
@@ -261,7 +265,9 @@ def validate_subscription(
             raise InvalidSubscriptionError("Only simple queries are supported")
         entity = get_entity(from_clause.key)
 
-        SubscriptionAllowedClausesValidator().validate(query)
+        SubscriptionAllowedClausesValidator(
+            self.entity_subscription.MAX_ALLOWED_AGGREGATIONS
+        ).validate(query)
```
lynnagara, line 270: I think this would be more flexible if the validate function was actually on the entity subscription itself rather then just getting the MAX_ALLOWED_AGGREGATIONS from there and passing it to the SubscriptionAllowedClausesValidator. e.g. `self.entity_subscription.validate(query)`

It would allow us to add any kind of custom validations on an entity, not just enforce the maximum number of aggregations. It would allow to add much stricter validations on the subscription data and immediately reject subscriptions that don't meet that criteria rather than have them fail somewhere later in the pipeline.

Obviously this works fine now as the only validation we have is on max aggregations anyway so I'll leave it up to you whether you want to take this approach or not.

snuba/subscriptions/entity_subscription.py
ahmedetefy: Agreed

snuba/clickhouse/query_dsl/accessors.py
ahmedetefy: Yes we still need it because this instance of `Query` -> https://github.com/getsentry/snuba/blob/6cd74121601df0f4dbef6590a3b04899847a7559/snuba/datasets/entities/sessions.py#L212 
is an instance of `class Query(AbstractQuery[Entity]):`

snuba/subscriptions/data.py
ahmedetefy: I agree with you.. logically it makes sense to just move the `validate` to `EntitySubscription`

snuba/clickhouse/query_dsl/accessors.py
lynnagara: Ah, right

lynnagara (approved):

PR #2096, opened 2021-09-15
waiting until alter query to be executed on replicas

asteny:
Hi!

In this PR I fixed 2 problems for distributed clickhouse:
1) ADD INDEX **IF NOT EXISTS** When migration runs on the second node of clickhouse I get an error that index already exists
2) **SETTINGS replication_alter_partitions_sync = 2** I tried to understand why I can't finish migrations on clickhouse cluster and got an error "Metadata on replica is not up to date with common metadata in Zookeeper. Cannot alter". My colleague from Yandex Clickhouse helped me to understand, that migration runs a lot of alter operations and they don't have time to finish. He [improved a little bit error message](https://github.com/ClickHouse/ClickHouse/pull/29009/files). For clickhouse cluster we need [wait for actions to be executed on replicas](https://clickhouse.tech/docs/en/operations/settings/settings/#replication-alter-partitions-sync), and should use replication_alter_partitions_sync = 2

Plz, take a look for PR, what do you think?

### DIFF
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -32,7 +32,10 @@ class ClickhouseClientSettingsType(NamedTuple):
 class ClickhouseClientSettings(Enum):
     CLEANUP = ClickhouseClientSettingsType({}, None)
     INSERT = ClickhouseClientSettingsType({}, None)
-    MIGRATE = ClickhouseClientSettingsType({"load_balancing": "in_order"}, 10000)
+    MIGRATE = ClickhouseClientSettingsType(
+        {"load_balancing": "in_order", "replication_alter_partitions_sync": 2},
+        10000
+    )
     OPTIMIZE = ClickhouseClientSettingsType({}, 10000)
     QUERY = ClickhouseClientSettingsType({"readonly": 1}, None)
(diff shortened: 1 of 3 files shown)


lynnagara:
Hey @asteny, this is great thanks for finding this solution. 

Looking at the docs on `replication_alter_partitions_sync`, it seems this should apply to all alter, optimize and truncate operations in migrations.

How about adding the setting to the migration client settings here instead? https://github.com/getsentry/snuba/blob/3317622fecf16a9435a4ae26300552db5fe968a5/snuba/clusters/cluster.py#L35
Then we would not need to apply it to each of the operations individually and potentially miss some cases. It seems to me there isn't a migration scenario where we wouldn't want this setting.

asteny:
> How about adding the setting to the migration client settings here instead?
> 
> https://github.com/getsentry/snuba/blob/3317622fecf16a9435a4ae26300552db5fe968a5/snuba/clusters/cluster.py#L35

Hello @lynnagara !)
Sure, you right. You suggested the right way
Plz check my changes

lynnagara (approved):

lynnagara:
@asteny Perfect, thanks a lot for your contribution!

PR #2101, opened 2021-09-16
feat(web): Entity-specific insertion endpoint for tests

jjbayer:
The existing API endpoint `/tests/<dataset:dataset>/insert` allows writing data to the default entity of each dataset. In the case of metrics, this means that only `metrics_sets` can be written.

This PR introduces an additional endpoint `/tests/entities/<entity:entity>/insert`, which allows the client to define the target entity explicitly.

This is required to run tests in https://github.com/getsentry/sentry/pull/28598.

### DIFF
--- a/snuba/datasets/metrics.py
+++ b/snuba/datasets/metrics.py
@@ -1,7 +1,18 @@
+from typing import Sequence
+
 from snuba.datasets.dataset import Dataset
 from snuba.datasets.entities import EntityKey
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.entity import Entity
 
 
 class MetricsDataset(Dataset):
     def __init__(self) -> None:
         super().__init__(default_entity=EntityKey.METRICS_SETS)
+
+    def get_all_entities(self) -> Sequence[Entity]:
+        return (
+            get_entity(EntityKey.METRICS_COUNTERS),
+            get_entity(EntityKey.METRICS_DISTRIBUTIONS),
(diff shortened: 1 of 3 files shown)


snuba/datasets/metrics.py
```diff
@@ -1,7 +1,18 @@
...
 from snuba.datasets.entities import EntityKey
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.entity import Entity
 
 
 class MetricsDataset(Dataset):
     def __init__(self) -> None:
         super().__init__(default_entity=EntityKey.METRICS_SETS)
+
+    def get_all_entities(self) -> Sequence[Entity]:
```
jjbayer, line 13: I added this such that `/tests/metrics/drop` truncates all entities.

snuba/web/views.py
```diff
@@ -37,14 +37,13 @@
...
     InvalidDatasetError,
-    enforce_table_writer,
     get_dataset,
     get_dataset_name,
     get_enabled_dataset_names,
 )
 from snuba.datasets.schemas.tables import TableSchema
 from snuba.query.composite import CompositeQuery
-from snuba.query.data_source.simple import Entity
+from snuba.query.data_source.simple import Entity as EntityMarker
```
evanh, line 46: This is a nit, but I would rather leave these imports the same and have the aliases in the testing code.

evanh (approved):
Looks good to me, just the one nit.

PR #2102, opened 2021-09-16
Communicate read-through cache exceptions to waiting clients

vladkluev:
### Problem
We have a lot of ExecutionErrors happening for reasons that we don't really know. This PR reports those errors back to clients waiting for results of duplicate queries

### Proposed Solution

If an exception happens serialize the exception using SerializableException and put that in redis. Deserialize it on the other side and re-raise it

### Feedback Desired

* The code has unit tests but I'm wondering if there is a larger integration test I can/should write. Are there larger tests that I can take inspiration from?

### DIFF
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -28,23 +28,33 @@ RESULT_WAIT = 2
 
 # ---- prototype stuff ----
 JsonStrBytes = bytes
-import json
+from typing import Any, cast
+
+import rapidjson as json
 
 ERROR_HEADER = b"__ERR__"
 
 
 def serialize_exception(exception: Exception) -> JsonStrBytes:
+    def _is_jsonable(item: Any) -> bool:
+        try:
+            json.dumps(item)
+            return True
+        except Exception:
+            return False
+
     edict = {
         "__name__": exception.__class__.__name__,
-        "__message__": getattr(exception, "message", str(exception)),
+        "__args__": [a for a in exception.args if _is_jsonable(a)],
     }
-    return json.dumps(edict).encode("UTF-8")
+    # I don't know why this cast is necessary
+    return cast(JsonStrBytes, json.dumps(edict).encode("UTF-8"))
 
 
 def deserialize_exception(s: JsonStrBytes) -> Exception:
     edict = json.loads(s)
-    res = type(edict["__name__"], (Exception,), {"args": (edict["__message"])})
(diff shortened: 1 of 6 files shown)


snuba/state/cache/redis/backend.py
```diff
@@ -215,14 +254,17 @@ def build_notify_queue_key(task_ident: str) -> str:
...
 
                 # If there is no value, that means that the client responsible
                 # for generating the cache value errored while generating it.
                 if raw_value is None:
-                    # TODO: If we wanted to get clever, this could include the
-                    # error message from the other client, or a Sentry ID or
-                    # something.
-                    raise ExecutionError("no value at key")
+                    raise ExecutionError(
+                        "no value at key: if we're here that means the original process executing the query crashed before we could handle the exception"
```
evanh, line 266: Please no pronouns in our error messages.

snuba/state/cache/redis/backend.py
```diff
@@ -26,6 +26,41 @@
...
+
+import rapidjson as json
+
+ERROR_HEADER = b"__ERR__"
+
+
+def serialize_exception(exception: Exception) -> JsonStrBytes:
+    def _is_jsonable(item: Any) -> bool:
+        try:
+            json.dumps(item)
```
evanh, line 41: This would be better if it returned the JSON value, so we don't JSON encode everything twice.

evanh:
> 1.    I think what I'm doing breaks the type safety of the cache class a little bit. All functions expect a return type of TValue however a serialized exception is not of type TValue is there a better way to do what I'm doing?

The TValue ultimately trickles down to the Codec class used to encode/decode values. I would create a new codec class that correctly handles exceptions and actual results.

> 2.    Are we okay with JSON as a serialization format? Does anyone have other suggestions?

That's what the codec currently uses. I would get rid of the header you're appending though. You can check if the value is an error based on the keys of the JSON stored in Redis (e.g. `if "error" in decoded...`). Also I don't think we need every field of the exception. We can encode only the exception type and probably error message and that's it.

> 3.    Do we want the ability to reraise the same Exception class? Currently we are just re-raising something that looks like it.

I think this is the ideal solution. That way we can be sure all the error handling at the higher levels will serve the correct responses. I'm not sure what the best way to do that would be.

> 4.    What other information do we want to convey in the error that we serialize that we are not serializing right now?

I think you got everything. The two important fields are the exception class and the message of that exception.

evanh (approved):

PR #2103, opened 2021-09-17
feat(subscription): relax validation on aggregations

ahmedetefy:
Increases the allowed number of aggregations in the select column for `sessions` subscription queries to two since in crash rate alerts, we'd like to be able to return the total count along side the aggregate

Corresponding change in Sentry https://github.com/getsentry/sentry/pull/28675

### DIFF
--- a/snuba/query/validation/validators.py
+++ b/snuba/query/validation/validators.py
@@ -97,8 +97,13 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
 
     def validate(self, query: Query, alias: Optional[str] = None) -> None:
         selected = query.get_selected_columns()
-        if len(selected) != 1:
-            raise InvalidQueryException("only one aggregation in the select allowed")
+
+        if len(selected) > 2:
+            # In datasets like `sessions` dataset, we might also want to return the total count
+            # along with the aggregate value
+            raise InvalidQueryException(
+                "a maximum of two aggregations in the select are allowed"
+            )
 
         disallowed = ["groupby", "having", "orderby"]
         for field in disallowed:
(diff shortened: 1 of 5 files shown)


evanh (commented):
This change looks good to me. Please add a test to `tests/subscriptions/test_worker.py` so we can get an E2E test.

Is there an associated Sentry change that needs to happen as well?

ahmedetefy:
Hey @evanh I have made some changes so that the checks are entity specific and added tests for that 

> This change looks good to me. Please add a test to `tests/subscriptions/test_worker.py` so we can get an E2E test.

I will add an E2E test in a separate PR to the feature branch I am working on if that's okay with you because to me it doesn't feel like an E2E makes sense here especially that the logic that accepts sessions subscription is in a different PR

> Is there an associated Sentry change that needs to happen as well?

I am still working on it.. will put up a PR soon and link it here

snuba/query/validation/validators.py
```diff
@@ -95,10 +98,14 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
     clauses are being used in the query, and that those clauses are in the correct structure.
     """
 
+    def __init__(self, entity: Optional["Entity"] = None) -> None:
+        self.entity = entity
```
evanh, line 102: This is causing an import loop because entities are supposed to contain validators, not the other way around. However, the query knows which entity it is being called on, you can get it using this class: https://github.com/getsentry/snuba/blob/master/snuba/query/parser/validation/functions.py#L33

You can then do whatever check you need from there.

evanh (changes requested):
Fix the import loop please, but otherwise makes sense.

evanh (approved):
Please ensure CI is passing before merging.

PR #2105, opened 2021-09-20
feat(suspect-spans): Schema changes for suspect spans

Zylphrex:
Adds the nested columns for op, group, and exclusive span for suspect spans.

# Forwards

```SQL
Local operations:
ALTER TABLE transactions_local ADD COLUMN IF NOT EXISTS spans Nested(op LowCardinality(String), group UInt64, exclusive_time Float64) AFTER span_op_breakdowns.value;
ALTER TABLE transactions_local MODIFY COLUMN spans.op Array(LowCardinality(String)) TTL finish_ts + INTERVAL 1 MONTH;
ALTER TABLE transactions_local MODIFY COLUMN spans.group Array(UInt64) TTL finish_ts + INTERVAL 1 MONTH;
ALTER TABLE transactions_local MODIFY COLUMN spans.exclusive_time Array(Float64) TTL finish_ts + INTERVAL 1 MONTH;
ALTER TABLE transactions_local ADD INDEX IF NOT EXISTS bf_spans_op spans.op TYPE bloom_filter() GRANULARITY 1;
ALTER TABLE transactions_local ADD INDEX IF NOT EXISTS bf_spans_group spans.group TYPE bloom_filter() GRANULARITY 1;


Dist operations:
ALTER TABLE transactions_dist ADD COLUMN IF NOT EXISTS spans Nested(op LowCardinality(String), group UInt64, exclusive_time Float64) AFTER span_op_breakdowns.value;
```

# Backwards

```SQL
Local operations:
ALTER TABLE transactions_local DROP INDEX IF EXISTS bf_spans_group;
ALTER TABLE transactions_local DROP INDEX IF EXISTS bf_spans_op;
ALTER TABLE transactions_local DROP COLUMN IF EXISTS spans;


Dist operations:
ALTER TABLE transactions_dist DROP COLUMN IF EXISTS spans;
```

### DIFF
--- a/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
+++ b/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
@@ -50,7 +50,7 @@ class Migration(migration.ClickhouseNodeMigration):
             operations.DropIndex(
                 storage_set=StorageSetKey.TRANSACTIONS,
                 table_name="transactions_local",
-                index_name="bf_spans_op",
+                index_name="bf_spans_group",
             ),
             operations.DropIndex(
                 storage_set=StorageSetKey.TRANSACTIONS,
--- a/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
+++ b/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
@@ -45,8 +45,9 @@ class Migration(migration.ClickhouseNodeMigration):
                     "spans",
                     Nested(
                         [
-                            ("key", String(Modifiers(low_cardinality=True))),
-                            ("value", Float(64)),
+                            ("op", String(Modifiers(low_cardinality=True))),
+                            ("group", UInt(64)),
+                            ("exclusive_time", Float(64)),
                         ]
                     ),
                 ),
--- a/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
+++ b/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
@@ -27,10 +27,36 @@ class Migration(migration.ClickhouseNodeMigration):
                 ),
                 after="span_op_breakdowns.value",
             ),
+            operations.AddIndex(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="transactions_local",
+                index_name="bf_spans_op",
+                index_expression="spans.op",
+                index_type="bloom_filter()",
+                granularity=1,
+            ),
+            operations.AddIndex(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="transactions_local",
+                index_name="bf_spans_group",
+                index_expression="spans.group",
+                index_type="bloom_filter()",
+                granularity=1,
+            ),
         ]
 
     def backwards_local(self) -> Sequence[operations.SqlOperation]:
         return [
+            operations.DropIndex(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="transactions_local",
+                index_name="bf_spans_op",
+            ),
+            operations.DropIndex(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="transactions_local",
+                index_name="bf_spans_op",
+            ),
             operations.DropColumn(
                 StorageSetKey.TRANSACTIONS, "transactions_local", "spans"
             ),
--- a/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
+++ b/snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
@@ -1,0 +1,62 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, Float, Nested, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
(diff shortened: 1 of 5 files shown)


snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
```diff
@@ -0,0 +1,62 @@
...
+        ]
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name="transactions_dist",
+                column=Column(
+                    "spans",
+                    Nested(
```
evanh, line 46: Why is this different from above?

Zylphrex: Shoot, this is a typo, will fix.

snuba/migrations/snuba_migrations/transactions/0012_transactions_add_spans.py
```diff
@@ -0,0 +1,63 @@
...
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
```
fpacifici, line 14: You will also have to add a Bloomfilter index to the `spans.op` column and to the `spans.group` column.

See
https://github.com/getsentry/snuba/blob/ae6aaae1b815273945a2c6c73c427921f87320d7/snuba/migrations/snuba_migrations/metrics/templates.py#L133-L140

Zylphrex: According to the [documentation](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/#available-types-of-indices), a bloom filter index is only used with the functions `equals/notEquals/in/notIn/has`.

If I'm understanding this correctly, this means that we can use it for a search like

```sql
has(spans.op, 'db')
```

However, if we want a search for multiple span op/group pairs like `[('db', 1234), ('http.client', 5678)]`, would we'd have to use `AND`/`OR`s like

```sql
(has(spans.op, 'db') OR has(spans.op, 'http.client')) AND (has(spans.group, 1234) OR has(spans.group, 5678))
```

in order to leverage the bloom filter index right?

fpacifici: Not exactly. 
You would use the bloomfilter to filter down the transactions row before the arrayJoin. 
So 
```
(has(spans.op, 'db') OR has(spans.op ....)) AND (has(spans.group ....))
```
Then, on the result of the arrayjoin you would run the equality as you would do today with measurements or span_ops to narrow down the results to the exact rows you want out of the array join.

The bloom filter would operate early in the pipeline narrowing down the blocks to scan, then within the blocks you are scanning, if you do an arrayjoin, you sill have to pick only the rows you care for.

evanh (approved):
This looks good to me. I think we can move ahead on this.

fpacifici:
One last thing to check before merging. I remember we ran into issues with something like this on the querylog. 
Could you test that we can add new columns to the spans nested structure on a distributed clickhoouse (@evanh  could help you on this). 
I think I remember that, on querylog, we wanted to add columns to the `clickhouse_queries` nested column. In order to do this we would add array columns named something like `clickhouse_queries.something`, though when inserting new rows that did not contain that field clickhoouse would reject the insert because the size of the array on the new column (empty) did not have the same length as the other fields. 

Could you verify we can add new columns to the spans nested column before merging ?

evanh:
@fpacifici How did was that problem solved before?

fpacifici:
Unfortunately I am not sure whether that problem was solved with no downtime. On transactions we could not afford a downtime.

Zylphrex:
After some testing, there are some issues with adding a new nested array column but I believe there is a viable strategy to do so with no downtime.

## Migration

I added a new column with the following migration

```sql
ALTER TABLE transactions_local ADD COLUMN IF NOT EXISTS spans.duration Array(Float64) AFTER spans.exclusive_time;
```

## Problem

If we naively apply the migration without any change to the transaction processor, we begin encountering the following error:

```
snuba.clickhouse.errors.ClickhouseWriterError: [190] Elements 'spans.op' and 'spans.duration' of Nested data structure 'spans' (Array columns) have different array sizes. (version 20.3.9.70 (official build))
```

This is due to the fact that the processor is not emitting a corresponding array of the same length for the new column. This error looks to be the expected behaviour from ClickHouse as it was a check added to make sure the data being inserted is of the expected shape because subsequent queries may assume the shape of the data. Relevant PR at ClickHouse/ClickHouse#3118.

## Proposed Migration Strategy (Tested and Verified Locally)

1. Enable the [input_format_skip_unknown_fields](https://clickhouse.com/docs/en/operations/settings/settings/#settings-input-format-skip-unknown-fields) setting in ClickHouse so that unknown values during insertion are skipped rather than raise an error.
2. Modify the transaction processor to being emitting the array values for the new columns, this value will be skipped as the migration has not occurred.
3. Apply the migration to being storing the newly emitted values.
4. [Optional] Disable the setting again.

fpacifici:
Looks like a good plan to me. 
Please provide the SQL for the distributed migration as well. and then it is good to go.

Zylphrex:
> Wait though, I think we are missing one last thing. We discussed keeping spans around for less than 90 days. Can we add a 30 days TTL to the new column?

@fpacifici Yes, we can change the TTL to 30 days instead of 90 days. I'll update that.

fpacifici:
LGTM

evanh:
@Zylphrex Can you also add the rollback SQL commands into the comment just in case?

evanh (approved):
Let's not merge until we've tested this in staging.

PR #2107, opened 2021-09-20
feat(subscription): Support Crash Rate Alerts

ahmedetefy:
* Add intellij .idea to gitignore

* feat(subscription): Relax validation on aggregations

Increases the allowed number of aggregations in the
select column for sessions subscription queries to
two since in crash rate alerts, we'd like to be able
to return the total count along side the aggregate

### DIFF
--- a/snuba/query/validation/validators.py
+++ b/snuba/query/validation/validators.py
@@ -97,8 +97,13 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
 
     def validate(self, query: Query, alias: Optional[str] = None) -> None:
         selected = query.get_selected_columns()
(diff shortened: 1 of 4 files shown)


lynnagara:
I am not opposed to this implementation which loosens the validations everywhere to accomodate sessions but I think it would be more ideal for each entity to have its own validators.

Since you are introducing the EntitySubscription class in https://github.com/getsentry/snuba/pull/2094, it seems like that could be a logical place to define the validator for each entity.

ahmedetefy:
@lynnagara I agree with you.. I was actually thinking the same since we have specialized `EntitySubscription` classes now, I will close this in favor of a new PR

PR #2109, opened 2021-09-20
SnubaException: A new exception class for all your exception needs

vladkluev:
### Context

Following up from #2102 this PR implements a serializable exception class which can be serialized and reraised upon deserialization

### Solution

SerializableException is an extension of the Exception class that intends to be the base exceptions for all exceptions in Snuba. All Exception subclasses have already been replaced in this PR. In addition to serializability, SerializableException also adds whether the error should be reported to sentry as a first class citizen.

### DIFF
--- a/snuba/datasets/entities/metrics.py
+++ b/snuba/datasets/entities/metrics.py
@@ -48,8 +48,10 @@ class TagsTypeTransformer(QueryProcessor):
 
             key = exp.key
             if not isinstance(key.value, str) or not key.value.isdigit():
-                raise InvalidExpressionException(
-                    exp, "Expected a string key containing an integer in subscriptable."
+                raise InvalidExpressionException.from_args(
+                    exp,
+                    "Expected a string key containing an integer in subscriptable.",
+                    should_report=True,
                 )
 
             return SubscriptableReference(
--- a/snuba/datasets/entities/metrics.py
+++ b/snuba/datasets/entities/metrics.py
@@ -51,7 +51,6 @@ class TagsTypeTransformer(QueryProcessor):
                 raise InvalidExpressionException.from_args(
                     exp,
                     "Expected a string key containing an integer in subscriptable.",
-                    should_report=True,
                 )
 
             return SubscriptableReference(
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -16,9 +16,8 @@ class ValidationException(InvalidQueryException):
 
 
 class InvalidExpressionException(ValidationException):
-    # TODO (figure out what do do with this)
     @classmethod
     def from_args(
-        cls, expression: Expression, message: str, should_report: bool,
+        cls, expression: Expression, message: str, should_report: bool = True,
     ) -> "InvalidExpressionException":
         return cls(message, should_report=should_report, expression=repr(expression))
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -1,3 +1,4 @@
+from snuba.utils.snuba_exception import SnubaException
 from snuba.query.expressions import Expression
 
 
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -1,9 +1,8 @@
 from snuba.query.expressions import Expression
-from snuba.utils.snuba_exception import JsonSerializable, SnubaException
+from snuba.utils.snuba_exception import SnubaException
 
 
 class InvalidQueryException(SnubaException):
-    # TODO (figure out what do do with this)
     """
     Common parent class used for invalid queries during parsing
     and validation.
@@ -13,12 +12,6 @@ class InvalidQueryException(SnubaException):
         message: Message of the exception
         report: Should we report the exception to Sentry or not
     """
-
-    @classmethod
-    def from_args(
-        cls, *, message: str, report: bool = True, **kwargs: JsonSerializable
-    ) -> "InvalidQueryException":
-        return cls(message=message, report=report, **kwargs)
 
     def __str__(self) -> str:
         return f"{self.message}"
@@ -30,12 +23,8 @@ class ValidationException(InvalidQueryException):
 
 class InvalidExpressionException(ValidationException):
     # TODO (figure out what do do with this)
-    def __init__(
-        self, expression: Expression, message: str, report: bool = True
-    ) -> None:
-        self.expression = expression
-        self.message = message
-        super().__init__(message, report=report)
-
-    def __str__(self) -> str:
-        return f"Invalid Expression {self.expression}: {self.message}"
+    @classmethod
+    def from_args(
+        cls, message: str, expression: Expression, should_report: bool,
+    ) -> "InvalidExpressionException":
+        return cls(message, should_report=should_report, expression=repr(Expression))
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -8,13 +8,7 @@ class InvalidQueryException(SnubaException):
     and validation.
     This should not be used for system errors.
 
-    Attributes:
-        message: Message of the exception
-        report: Should we report the exception to Sentry or not
     """
-
-    def __str__(self) -> str:
-        return f"{self.message}"
 
 
 class ValidationException(InvalidQueryException):
@@ -25,6 +19,6 @@ class InvalidExpressionException(ValidationException):
     # TODO (figure out what do do with this)
     @classmethod
     def from_args(
-        cls, message: str, expression: Expression, should_report: bool,
+        cls, expression: Expression, message: str, should_report: bool,
     ) -> "InvalidExpressionException":
-        return cls(message, should_report=should_report, expression=repr(Expression))
+        return cls(message, should_report=should_report, expression=repr(expression))
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -1,8 +1,8 @@
 from snuba.query.expressions import Expression
-from snuba.utils.snuba_exception import SnubaException
+from snuba.utils.serializable_exception import SerializableException
 
 
-class InvalidQueryException(SnubaException):
+class InvalidQueryException(SerializableException):
     """
     Common parent class used for invalid queries during parsing
     and validation.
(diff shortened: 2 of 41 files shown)


tests/utils/test_snuba_exception.py
```diff
@@ -0,0 +1,42 @@
...
+        assert snubs_exc.message == message
+        assert snubs_exc.extra_data["from_standard_exception"]
+        assert snubs_exc.should_report
+
+    # if we create a SnubaException from a standard exception
+    # before, we'll create the same one again
+    try:
+        raise Exception("Other message")
+    except Exception as e:
+        new_snubs_exc = SnubaException.from_standard_exception_instance(e)
```
evanh, line 41: What is this testing exactly? Why would these be different, the only difference between them is the message you're passing in.

snuba/utils/snuba_exception.py
```diff
@@ -0,0 +1,99 @@
...
+    def get_class_by_name(self, cls_name: str) -> Optional[Type["SnubaException"]]:
+        return self.__mapping.get(cls_name)
+
+
+_REGISTRY = None
+
+
+def _get_registry() -> _ExceptionRegistry:
+    global _REGISTRY
+    # NOTE (Should this be protected by a mutex? I'm thinking it should be fine)
```
evanh, line 35: I don't think this is thread safe. You would need a lock here to avoid multiple threads modifying the registry at the same time, since this will be called every time a `SnubaException` is thrown.

evanh: I don't know if that's a performance problem since in theory exceptions happen very rarely. In a disaster scenario though this could make the disaster worse since not only are we throwing a lot of exceptions but those exceptions are bottlenecking on this class.

I think if you changed this a bit so the registry was created initially (instead of being set to None) and exception classes were added only once it would be better.

So if an exception happens, the registry is checked to see if that class has been added (which would be a threadsafe read). If the exception is not in the registry, then the lock is acquired and the class is added.

snuba/query/exceptions.py
```diff
@@ -1,37 +1,24 @@
...
-    def __init__(
-        self, expression: Expression, message: str, report: bool = True
-    ) -> None:
-        self.expression = expression
-        self.message = message
-        super().__init__(message, report=report)
-
-    def __str__(self) -> str:
-        return f"Invalid Expression {self.expression}: {self.message}"
+    # TODO (figure out what do do with this)
```
evanh, line 19: ?

snuba/datasets/entities/metrics.py
```diff
@@ -48,8 +48,10 @@ def transform_expression(exp: Expression) -> Expression:
 
             key = exp.key
             if not isinstance(key.value, str) or not key.value.isdigit():
-                raise InvalidExpressionException(
-                    exp, "Expected a string key containing an integer in subscriptable."
+                raise InvalidExpressionException.from_args(
+                    exp,
+                    "Expected a string key containing an integer in subscriptable.",
+                    should_report=True,
```
evanh, line 54: Why did you add `should_report=True`? That's the default already. Is this going to cause the exceptions to be reported when they weren't before?

evanh (changes requested):

snuba/utils/snuba_exception.py
vladkluev: This happens at the time the python AST is loaded into memory. The only time this would be modified by threads is if threads are defining a new exception locally inside themselves

snuba/datasets/entities/metrics.py
vladkluev: good point, not needed

evanh (approved):
Except for the couple TODOs that need explaining this looks good. Well done, I did not know about __init_subclass.

tests/utils/test_snuba_exception.py
vladkluev: This is testing the codepath of going from standard exception (e.g. AttributeError) to a SnubaException and that the message is preserved. The different messages are confusing so I'll fix that

vladkluev: > Why would these be different, the only difference between them is the message you're passing in.

I just reread your question. Basically this is testing that any other time a bare `Exception` is raised, if we create a `SnubaException` from it, it will be the same type as the one raised on this machine before (even if the message is different)

fpacifici (commented):
I am quite skeptic about this approach. 
The only scenario where being able to serialize and deserialize an exception is relevant is query execution after the cache has been applied.
The approach we take here is to force every new snuba exception to extend this specific one, which basically couples them with the caching scenario.
There are multiple downsides I can see:
- somebody will for sure forget to extend SnubaException when creating an exception that is thrown by the execution pipeline. Thus we will be turning that exception into a standard SnubaException.
- any exception thrown by an external library will have to be wrapped in a SnubaException. While this is common, I don't think we can rely on that for our exceptions to work. 

If serializing and deserializing exception was fundamental for most of our use cases, like being able to return them on the network to our clients, I would be more supportive, but in this scenario , as I mentioned above, it seems we are introducing a constraint everywhere, that will be hard to enforce, for the sake of a very specific use case.

I would suggest to apply this approach only to exception related with the last steps of the execution, but we are about to move caching earlier in the pipeline so that would not help. We would still have to keep this issue in mind for all the exceptions in the query pipeline.

Pickling is probably not a good idea either because the pickeld object can be big, but it may be an easier option to be able to serialize an object that is not simply json serializable. Have you considered that one?

vladkluev:
> The only scenario where being able to serialize and deserialize an exception is relevant is query execution after the cache has been applied.

The only scenario so far

> The approach we take here is to force every new snuba exception to extend this specific one, which basically couples them with the caching scenario.

I would argue that: 

* caching is decently central to the application
* I'm not sure how this necessarily "couples" exceptions to the caching scenario. It does couple them to JSON though so maybe that's not ideal? Although I'd argue that having everything in an error be JSON serializable makes it so that when it gets bubbled up, we can actually inspect whatever we put in it

> somebody will for sure forget to extend SnubaException when creating an exception that is thrown by the execution pipeline. Thus we will be turning that exception into a standard SnubaException.

That's not what happens. A new Exception type will be created with the same name and message. I also don't really understand what the actual downside of this is? A non-snuba exception will be defined, it'll get thrown and bubbled up to sentry?

> any exception thrown by an external library will have to be wrapped in a SnubaException. While this is common, I don't think we can rely on that for our exceptions to work.

What do you mean by "rely on that for our exceptions to work."? What would or would not work about exceptions?

>  it seems we are introducing a constraint everywhere, that will be hard to enforce, for the sake of a very specific use case.

Is the constraint you're talking about the necessity of extending `SnubaException`? I don't think it's a constraint necessarily. Nothing bad will happen if you don't do it. I think there are good things we can do if all of the exceptions we define are derived from a base class we control. For example: 

* being able to define whether an exception should be reported to sentry
* putting default tags on an exception for easier triage
* having different messages for the client and for internal logging

> pickling

Aside from size and debuggability, pickling doesn't gracefully handle the case of different processes having different versions of the code at a point in time but this approach does.

fpacifici:
About the coupling: this approach does couple each scenario that needs to raise an exception to caching logically. I know there is no direct code dependency, but the need for your exception to be json serializable is in itself coupling. I know that caching is important in the query pipeline, but we break down the query pipeline into phases so that each phase can be thought of independently from what happens before and after. You should not care whether caching exists or not when you write a translator or a query processor.
The coupling is caused by you having to take into account the need of a separate component when writing your component. 
Now, I could see this more justifiable if we restricted the idea to the exceptions raised during the execution pipeline. Let's say `QueryExecutionException`. We kind of already do something like that (see `QueryException`, it carries some specific data used to build a response). Generalizing that one leaving the other areas of Snuba alone would make more sense to me. You could start with those that are used after caching (which are very few).

> The only scenario so far
This argument tends not to convince me :D at least not if we cannot articulate a viable, plausible scenario that will show up in a reasonably short period of time.


> That's not what happens. A new Exception type will be created with the same name and message. I also don't really understand what the actual downside of this is? A non-snuba exception will be defined, it'll get thrown and bubbled up to sentry?

I assume you are talking about this:
https://github.com/getsentry/snuba/pull/2109/files#diff-7f215c7c643d28b180c87a203565ed73631f3407919e2bf9ac7aea6cf96a50d2R121-R130

That would be ok to me, but then why do we need to extend SnubaException at all ?

vladkluev:
> This argument tends not to convince me :D

Good heuristic

> but we break down the query pipeline into phases so that each phase can be thought of independently from what happens before and after. 

I have some opinions on whether this is actually the case but I'll save that for tech debt days.

> You should not care whether caching exists or not when you write a translator or a query processor.

As someone writing a translator or a query processor. If I'm writing a custom exception and I do:

```python3
class MyCustomException(SnubaException):
    pass
 
# instead of

class MyCustomException(Exception):
    pass 
```

At which point have I thought about caching existing? I've thought about Snuba existing but I'm working on Snuba. 

It seems to me that the main objection is the JSON. Am I correct in that?

> That would be ok to me, but then why do we need to extend SnubaException at all ?

So that you can [catch the same exception on  a different machine](https://github.com/getsentry/snuba/pull/2109/files#diff-7f215c7c643d28b180c87a203565ed73631f3407919e2bf9ac7aea6cf96a50d2R8-R14)

evanh:
I agree with Vlad about the JSON encoding. Making our internal exceptions JSON encodable seems more like a best practice than coupling every component to caching. The fact that caching leverages the fact that exceptions are JSON encodable is an independent concern.

fpacifici:
I have no issue with the JSON encoding. I think that is a good format for this.

I was thinking at `SnubaException` as an "Exception the cache module can load from redis" and not as a more generic "serializable exception", my bad. I think the idea of a generic serializable exception is more interesting and not inherently coupled with the cache. 
Could we change the naming though (something like `SerializableException`) as `SnubaException` does not tell anything of why that exception exists and about what it does?

As a separate issue, about the query pipeline, the whole system is supposed to resemble a pipe and filter architecture, which I believe has a great value when reusing components in different pipelines and in breaking down the logic into units that can be thought of in isolation. 
It is a common pattern in query planners and compilers where the whole processing is extremely complex. 
I do agree that there is a ton of tech debt and that architecture is not properly followed everywhere but that is still the goal.
If you plan to change this premise in your refactoring work, we should probably have an architecture discussion before. If, instead, you plan to make it more compliant with a pipe and filter architecture, that's great.

tests/utils/test_snuba_exception.py
evanh: I think I understand, but this might be clearer if you used something other than a bare Exception, like a custom exception class or ValueError or something.

PR #2111, opened 2021-09-21
feat(metrics): 10 second granularity

jjbayer:
This PR adds a migration which drops the materialized views of all existing metrics types, and recreates them with a 10 second granularity.

No data migration is performed.

### DIFF
--- a/snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
+++ b/snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
@@ -1,0 +1,97 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import AggregateFunction, Column, Float, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.snuba_migrations.metrics.templates import (
+    get_forward_view_migration_local,
+)
+
+
+def drop_views() -> List[operations.SqlOperation]:
+    return [
+        operations.DropTable(
+            storage_set=StorageSetKey.METRICS,
+            table_name=f"metrics_{metric_type}_mv_local",
+        )
+        for metric_type in ("sets", "counters", "distributions")
+    ]
+
+
+def recreate_views(granularity: int) -> List[operations.SqlOperation]:
+    return [
+        get_forward_view_migration_local(
+            source_table_name="metrics_buckets_local",
+            table_name="metrics_sets_local",
+            mv_name="metrics_sets_mv_local",
+            aggregation_col_schema=[
+                Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
+            ],
+            aggregation_states="uniqCombined64State(arrayJoin(set_values)) as value",
+            granularity=granularity,
+        ),
+        get_forward_view_migration_local(
+            source_table_name="metrics_counters_buckets_local",
+            table_name="metrics_counters_local",
+            mv_name="metrics_counters_mv_local",
+            aggregation_col_schema=[
+                Column("value", AggregateFunction("sum", [Float(64)])),
+            ],
+            aggregation_states="sumState(value) as value",
+            granularity=granularity,
+        ),
+        get_forward_view_migration_local(
+            source_table_name="metrics_distributions_buckets_local",
+            table_name="metrics_distributions_local",
+            mv_name="metrics_distributions_mv_local",
+            aggregation_col_schema=[
+                Column(
+                    "percentiles",
+                    AggregateFunction(
+                        "quantiles(0.5, 0.75, 0.9, 0.95, 0.99)", [Float(64)]
+                    ),
+                ),
+                Column("min", AggregateFunction("min", [Float(64)])),
+                Column("max", AggregateFunction("max", [Float(64)])),
+                Column("avg", AggregateFunction("avg", [Float(64)])),
+                Column("sum", AggregateFunction("sum", [Float(64)])),
+                Column("count", AggregateFunction("count", [Float(64)])),
+            ],
+            aggregation_states=(
+                "quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)((arrayJoin(values) AS values_rows)) as percentiles, "
+                "minState(values_rows) as min, "
+                "maxState(values_rows) as max, "
+                "avgState(values_rows) as avg, "
+                "sumState(values_rows) as sum, "
+                "countState(values_rows) as count"
+            ),
+            granularity=granularity,
+        ),
+    ]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Re-create materialized views for metrics with a 10 second granularity
+    instead of 60 seconds.
+
+    The forward migration keeps the data rounded to start of minute, as there
+    is no way of recovering granularity for existing entries.
+
+    The backward migration could theoretically round existing entries to start of minute,
+    but that is omitted for the sake of simplicity.
+    """
+
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views() + recreate_views(granularity=10)
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views() + recreate_views(granularity=60)
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
+
+    def backwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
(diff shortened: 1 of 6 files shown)


snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
```diff
@@ -0,0 +1,97 @@
...
+                Column("count", AggregateFunction("count", [Float(64)])),
+            ],
+            aggregation_states=(
+                "quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)((arrayJoin(values) AS values_rows)) as percentiles, "
+                "minState(values_rows) as min, "
+                "maxState(values_rows) as max, "
+                "avgState(values_rows) as avg, "
+                "sumState(values_rows) as sum, "
+                "countState(values_rows) as count"
+            ),
```
jjbayer, line 67: There is quite some code duplication here. We could define the schema and states as constants somewhere, but I did not want to create a larger diff than necessary in the existing migrations.

snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
```diff
@@ -0,0 +1,97 @@
...
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views() + recreate_views(granularity=10)
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views() + recreate_views(granularity=60)
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
```
jjbayer, line 94: Not 100% sure if anything has to be done for "dist".

lynnagara (commented):
The migration looks good to me, though I can't comment on the substance of the change (60 -> 10)

snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
```diff
@@ -0,0 +1,97 @@
...
+    but that is omitted for the sake of simplicity.
+    """
+
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views() + recreate_views(granularity=10)
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views() + recreate_views(granularity=60)
```
fpacifici, line 91: Why are you deleting the 60 seconds materialized views? I thought we wanted both of them.

jjbayer: I did not want to get into the business of handling multiple storages for reading. Would keeping both of them mean writing a `QueryStorageSelector` for metrics?

fpacifici: We will definitively need to have multiple granularities to deal with. I don't think we will ever be able to get away with reasonable query performance. It will be unreasonable to run a month long query over 10 seconds buckets. 
We are likely to have at least 10 seconds, 1 hour, 1 day. Then maybe more. Not sure about the 1 minute aggregation.

Now since they are all in the same table we have the choice of having multiple "virtual storages" on top of the same table or having only one storage and deal with the granularity in the query processing phase. But this does not change the problem that multiple granularities have to exist. 

Since the format of different granularities here does not change (it does on sessions and outcomes) we could even make it simpler and have that done at Entity level where the `granularity` field is still visible.

fpacifici: I think we can keep it simple at first and Have a query processor in the Entity that picks up the granularity field from the query and applies the proper condition to the entity itself. No need for multiple storages. Still the application needs to deal with multiple granularities.

fpacifici: @jjbayer, do you need further assistance with this PR ?

jjbayer: @fpacifici Thanks, I think I might be able to write the query processor as you suggested, depending on how desperately we need this change. @jan-auer what do you think?

fpacifici: So multiple granularities are a must to run this data model in production. 
If you want to have only one for now, I am ok with that, but we are never going to put production load on this without having the hourly and daily aggregation, so it will have to be done anyway. Up to you.

If you build the processor. Please do that as an entity processor not as a storage selector. There you do not need to find patterns in the query. You will instead have access to the granularity field in the query and can process that.

snuba/query/processors/granularity_processor.py
```diff
@@ -0,0 +1,35 @@
+from snuba.datasets.metrics import DEFAULT_GRANULARITY
+from snuba.query.conditions import ConditionFunctions, binary_condition
+from snuba.query.expressions import Column, Literal
+from snuba.query.logical import Query
+from snuba.query.processors import QueryProcessor
+from snuba.request.request_settings import RequestSettings
+
+#: Granularities for which a materialized view exist, in ascending order
+GRANULARITIES_AVAILABLE = (10, 60)
```
jjbayer, line 9: Not sure if this list could be derived automatically somehow.

snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
fpacifici: So I assume the way you plan to run your queries is generally something like:

```
SELECT bucketed_time
GROUP BY bucketed_time
GRANULARITY something like 10 - 3600 - etc.
```

Is that correct? 
If yes I would expect you plan to use multiple types of granularities always multiple of the minimum one that is 10 seconds. Is that correct?

Then do you ever need the bare timestamp column ?

snuba/migrations/snuba_migrations/metrics/0007_granularity_10s.py
```diff
@@ -0,0 +1,27 @@
...
+    The backward migration does *not* delete any data from the destination tables.
+    """
+
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return create_views(granularity=10)
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return drop_views(granularity=10)
```
fpacifici, line 21: I don't think adding all 10s granularities is a good idea as it creates a three views in one migraiton.
The issue is that the atomic unit, when running the migration framework is the individual migration.
The framework does not expose individual statements in a migration. Multiple statements should be in the same migration only if they are supposed to be atomic. Either they all succeed or they are all reversed. 

Three views are largely independent. If one view creation fails the framework won't tell you which one did. It will just tell you this entire migration failed. 

Please split it into three, one per table, each creates a materialized view. Yes it is more verbose but that is  to allow the framework to get into a consistent state after each migration. 

This way will also make the templates simpler to understand.

Also, we are likely one day to split tables into separate clusters. So keeping the migrations separated by use cases (as they previous migrations are) will make things easier there.

snuba/query/processors/granularity_processor.py
```diff
@@ -0,0 +1,35 @@
...
+        if requested_granularity:
+            for granularity in reversed(GRANULARITIES_AVAILABLE):
+                if (requested_granularity % granularity) == 0:
+
+                    return granularity
+
+        return DEFAULT_GRANULARITY
+
+    def process_query(self, query: Query, request_settings: RequestSettings) -> None:
+        granularity = self.__get_granularity(query)
```
fpacifici, line 28: Shouldn't we reject queries that have a granularity that is not a multiple of 10 seconds or a multiple of one of the granularities we stored.

Please create a subclass of https://github.com/getsentry/snuba/blob/master/snuba/query/exceptions.py#L5 for that.

Actually, better, create an entity validator for that 
https://github.com/getsentry/snuba/blob/master/snuba/datasets/entity.py#L103-L110

snuba/datasets/entities/metrics.py
```diff
@@ -104,6 +105,7 @@ def get_extensions(self) -> Mapping[str, QueryExtension]:
     def get_query_processors(self) -> Sequence[QueryProcessor]:
         return [
             BasicFunctionsProcessor(),
+            GranularityProcessor(),
```
fpacifici, line 108: So, if my understanding is correct, only a limited amount of granularities will be allowed in the query.
The ones we have in the DB and all their multiples. 
This processor picks the granularity and sets a condition as they are all in the same storage, then The TimeSeriesProcessor applies the proper time transformation.
You will never use the bare `timestamp` column. 

Is this correct ?

fpacifici (changes requested):
Please split the migration into one per materialized view so each of them can be applied atomically.
Otherwise just some minor issues.

PR #2112, opened 2021-09-23
feat(subscriptions): Support entity max aggregations

ahmedetefy:
This PR:
- Adds the ability to set an entity specific number of allowed maximum aggregations for subscriptions in the `EntitySubscription` class

### DIFF
--- a/snuba/query/validation/validators.py
+++ b/snuba/query/validation/validators.py
@@ -101,14 +101,14 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
     def validate(self, query: Query, alias: Optional[str] = None) -> None:
         selected = query.get_selected_columns()
         if len(selected) > self.max_allowed_aggregations:
-            if self.max_allowed_aggregations == 1:
-                error_text = "A maximum of 1 aggregation is allowed in the select"
-            else:
-                error_text = (
-                    f"A maximum of {self.max_allowed_aggregations} aggregations are allowed in "
-                    f"the select"
-                )
-            raise InvalidQueryException(error_text)
+            aggregation_error_text = (
+                "1 aggregation is"
+                if self.max_allowed_aggregations == 1
+                else f"{self.max_allowed_aggregations} aggregations are"
+            )
+            raise InvalidQueryException(
+                f"A maximum of {aggregation_error_text} allowed in the select"
+            )
 
         disallowed = ["groupby", "having", "orderby"]
         for field in disallowed:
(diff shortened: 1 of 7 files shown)


snuba/query/validation/validators.py
```diff
@@ -95,10 +95,20 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
...
-            raise InvalidQueryException("only one aggregation in the select allowed")
+        if len(selected) > self.max_allowed_aggregations:
+            if self.max_allowed_aggregations == 1:
+                error_text = "A maximum of 1 aggregation is allowed in the select"
+            else:
+                error_text = (
+                    f"A maximum of {self.max_allowed_aggregations} aggregations are allowed in "
+                    f"the select"
+                )
+            raise InvalidQueryException(error_text)
```
lynnagara, line 111:
```suggestion
        max_aggregations = "1 aggregation" if self.max_allowed_aggregations == 1 else f"{self.max_allowed_aggregations} aggregations"
        raise InvalidQueryException(f"A maximum of {max_aggregations} are allowed in the select")
```

tests/subscriptions/test_data.py
```diff
@@ -128,6 +128,26 @@
...
+                WHERE org_id = 1 AND project_id IN tuple(1) LIMIT 1
+                OFFSET 0 GRANULARITY 3600
+                """
+            ),
+            time_window=timedelta(minutes=120),
+            resolution=timedelta(minutes=1),
+            entity_subscription=create_entity_subscription("sessions"),
+        ),
+        InvalidQueryException,
+        id="Delegate subscription",
```
lynnagara, line 149: Isn't it a SnQL subscription not a delegate subscription?

tests/test_api.py
```diff
@@ -2083,6 +2083,48 @@ def test_delegate(self) -> None:
             "subscription_id": f"0/{expected_uuid.hex}",
         }
 
+    def test_delegate_with_sessions_entity_subscription(self) -> None:
```
lynnagara, line 2086: I think `test_api.py` is generally for testing events, there is `test_sessions_api.py` for sessions API tests.

tests/subscriptions/test_data.py
ahmedetefy: oops

tests/test_sessions_api.py
```diff
@@ -281,3 +283,91 @@ def test_minute_granularity_range(self, get_project_id: Callable[[], int]) -> No
             "type": "invalid_query",
             "message": "Minute-resolution queries are restricted to a 7-hour time window.",
         }
+
+    def test_delegate_with_sessions_entity_subscription(self) -> None:
```
lynnagara, line 287: Would you mind putting the subscriptions tests on a separate class. This TestSessionsApi is for testing the query API. Basically let's follow the same pattern as https://github.com/getsentry/snuba/blob/master/tests/test_api.py#L2010

lynnagara (approved):
Thanks

PR #2113, opened 2021-09-23
fix: Fix queries on columns with different underlying types

lynnagara:
Queries involving columns with different internal and external
types did not behave properly when it included a condition involving
one of the >, >=, <. <=, LIKE or NOT LIKE operators.

The BaseTypeConverter contained an optimization that attempted to use
the internal column for the comparison however this is not correct
as the comparison returns completely different results depending on
the type.

This change drops the optimizations on all comparisons except =, !=,
IS NULL and IS NOT NULL. Queries containing a mix of optimizable and
non optimizable conditions are treated as unoptimized and the
external type is used.

### DIFF
--- a/snuba/query/processors/type_converters/__init__.py
+++ b/snuba/query/processors/type_converters/__init__.py
@@ -3,7 +3,7 @@ from typing import Set
 
 from snuba.clickhouse.processors import QueryProcessor
 from snuba.clickhouse.query import Query
-from snuba.query.conditions import FUNCTION_TO_OPERATOR, ConditionFunctions
+from snuba.query.conditions import ConditionFunctions
 from snuba.query.exceptions import ValidationException
 from snuba.query.expressions import Column, Expression, FunctionCall, Literal
 from snuba.query.matchers import Any as AnyMatch
@@ -30,8 +30,29 @@ class BaseTypeConverter(QueryProcessor, ABC):
             Or(
                 [
                     String(op)
-                    for op in FUNCTION_TO_OPERATOR
-                    if op not in (ConditionFunctions.IN, ConditionFunctions.NOT_IN)
+                    for op in (
+                        ConditionFunctions.EQ,
+                        ConditionFunctions.NEQ,
+                        ConditionFunctions.IS_NULL,
+                        ConditionFunctions.IS_NOT_NULL,
+                    )
+                ]
+            ),
+        )
+
+        unoptimizable_operator = Param(
+            "operator",
+            Or(
+                [
+                    String(op)
+                    for op in (
+                        ConditionFunctions.GT,
+                        ConditionFunctions.GTE,
+                        ConditionFunctions.LT,
+                        ConditionFunctions.LTE,
+                        ConditionFunctions.LIKE,
+                        ConditionFunctions.NOT_LIKE,
+                    )
                 ]
             ),
         )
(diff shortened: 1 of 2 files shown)


untitaker (approved):
I'm trying to follow the PR, I didn't know before what BaseTypeConverter really does. Is the problem that BaseTypeConverter, for `less(event_id, 'a1b4c3c5a39040509eeb13da2a511898'))`, either

1. converts column to be a string (i.e. call `_process_expressions` of `UUIDArrayColumnProcessor` which produces a slow query)
2. converts literal to be a UUID (i.e. `_replace_literal` which produces a fast query)

and now you're forcing to only go with (2) if the operator is about equality or null-checks, but no longer ordering?

Assuming "yes", makes sense so far, some questions:

1. the problem arises in a query like this:
    ```
    SELECT
	    (replaceAll(toString(event_id), '-', '') AS _snuba_event_id),
	    (group_id AS _snuba_group_id),
	    (project_id AS _snuba_project_id),
	    (timestamp AS _snuba_timestamp)
    FROM errors_local
    PREWHERE in(_snuba_group_id, tuple(1))
    WHERE equals(deleted, 0) AND
	    greaterOrEquals(_snuba_timestamp, toDateTime('2021-09-21T05:02:56', 'Universal')) AND
	    less(_snuba_timestamp, toDateTime('2021-09-21T15:02:56', 'Universal')) AND
	    in(_snuba_project_id, tuple(2)) AND
	    lessOrEquals(_snuba_timestamp, toDateTime('2021-09-20T08:17:10', 'Universal')) AND
	    (less(_snuba_timestamp, toDateTime('2021-09-20T08:17:10', 'Universal')) OR less(event_id, 'a1b4c3c5-a390-4050-9eeb-13da2a511898')) AND
	    in(_snuba_project_id, tuple(2))
    ORDER BY _snuba_timestamp DESC, _snuba_event_id DESC
    LIMIT 10 OFFSET 0
    ```

	It would be faster to optimize orderby here. Do we not want to do that because Sentry already expects string-looking columns to be ordered lexicographically? Could we redefine/"set back" those expectations and say that ordering is defined per-column and Sentry should not rely on anything except for it being consistent in all clauses? We can then still write down somewhere that e.g. specifically the title column is not only a string, but a "real string" with "real string ordering".
	
	I think establishing a rule that Sentry should not generally assume ordering to be anything except consistent within Snuba (with exemptions granted per-column) will give you more freedom in the future. Over-the-top example: What if the next clickhouse version orders integers and timestamps by reversed bit order? Then we're screwed perf-wise.

2. will other type conversions be affected by this? this makes sense for string <-> UUID, but I wonder if instead we should apply this behavior specifically for UUIDArrayColumnProcessor based on some flag that is overridden in the subclass. If we have other type conversions that do not affect ordering behavior it would be a shame if we forego an optimization because of this.

evanh (approved):
Markus makes a good point about this not being applied to all UUID columns. Things like trace_id for example.

lynnagara:
I've updated this to use the optimised ordering for hexint columns as the ordering on the actual integer should still match the external representation for that type, so we can still take advantage of the optimisation. Now the changes involving ordering operators only apply to UUIDs. I kept LIKE and NOT LIKE as always unoptimised across all the types as I doubt these will ever work properly without converting the type.

We had a discussion yesterday about lexicographical ordering being the only correct one for UUIDs. It is possible (even likely) this will cause performance regression on certain types of queries but at this point we are prioritising correctness over speed. With that in mind I don't think we want to make a blanket exceptions now for certain column names such as `trace_id` unless there is a very strong reason to do so. However if we can justify it we can easily add the option to the UUID processor later to switch between the two modes.

untitaker:
>We had a discussion yesterday about lexicographical ordering being the only correct one for UUIDs.

Can you summarize the reasoning for this conclusion? This is what I am currently unsure about myself.

My argument for lexicographical ordering being the only right one is: If it looks like a string to Sentry, it should sort like a string. But you can also just say "Sentry should not make this expectation." Maybe it's deeper than that though.

lynnagara:
@untitaker My reasoning is basically the same as yours. Not only does it look like a string, Snuba actually returns the type "string" for this column in the metadata sent back to the client. So there should be no difference to a regular string. UUID storage of event_id and trace_id is considered an internal optimisation only and Sentry should not know anything about it. It's possible one day we will need to relax this expectation and add some other UUIDString type or so but today it's just considered a regular string.

PR #2114, opened 2021-09-24
fix(ci): Run Sentry integration tests with Python 3.6

armenzg:
Snuba's Python version is 3.8, however, Sentry uses 3.6, thus, fixing it.

There's currently some integration tests failing to complete because it is
installing a newer version of psycopg2-binary which makes the tests fail.

See [this commit][1] for details.

[1]: https://github.com/getsentry/sentry/commit/3f71db8928bf6e20eb821026ad6713b4b043f2bf

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,7 +182,7 @@ jobs:
         with:
           workdir: sentry
           cache-files-hash: ${{ hashFiles('sentry/requirements**.txt') }}
-          python-version: 3.8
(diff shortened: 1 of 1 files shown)


armenzg:
This does not directly contribute to the work for Apple M1, however, I'm tracking it as part of the side effects of the work.

armenzg:
CC @lynnagara

billyvg (approved):

evanh (approved):

PR #2117, opened 2021-09-24
feat(suspect-spans): Add array join optimizations on spans queries

Zylphrex:
Similar to the array join optimization on tags, this change rewrites array join
queries on spans as a single array join with the optional array filter
optimization. It also adds the necessary conditions in order to take advantage
of the bloom filter index available on the spans.op and spans.group columns.


Zylphrex:
> This all makes sense and is well done. I suspect you tried this, but did you consider trying to abstract the current array join optimizer to support a more arbitrary nesting structure? That might end up being the same size change as adding this one, and then we wouldn't have to maintain the same functionality in two places.

So I tried that at first, but that quickly got very complicated as the filtering capability would need to work on all possible combinations of the key columns. I'll give this another try now that I have a clearer idea what is common across the two optimizers, but see below for where the complexity is.

Take the nested structure for tags for example, it only uses `tags.keys` as the key column so the filtering is only done one that one column.

However, with the nested structures for spans, we need to support filtering on the 3 combinations of the 2 key columns.

So to support arbitrarily nested structures, it'd get more complex as it would need to support all possible combinations of the `n` key columns.

snuba/query/processors/arrayjoin_optimizer.py
```diff
@@ -0,0 +1,648 @@
...
+                    )
+
+            # No viable optimization
+            return expr
+
+        query.transform_expressions(replace_expression)
+
+        # If there is a bloom filter index present on the key columns, we want
+        # to take advantage of it by adding in the necessary `has` conditions.
+        if self.__use_bf_index:
```
evanh, line 173: I think I would make the bloom filter check a separate query processor. It feels like it's just tacked on here and would simplify this processor a bit. Up to you but the arrayjoin optimization and the bloom filter optimization feel like separate concepts.

Zylphrex: Split out in de2187c

snuba/datasets/storages/transactions.py
```diff
@@ -130,9 +131,11 @@
         ArrayJoinKeyValueOptimizer("tags"),
         ArrayJoinKeyValueOptimizer("measurements"),
         ArrayJoinKeyValueOptimizer("span_op_breakdowns"),
-        ArrayJoinOptimizer(
-            "spans", ["op", "group"], ["exclusive_time"], use_bf_index=True
-        ),
+        # the bloom filter optimizer should occur before the array join optimizer
```
evanh, line 134: For the record you could have had the bloom filter look for the conditions created by the arrayjoin. But not a big deal.

PR #2121, opened 2021-09-29
feat(subscriptions): Subscription data store uses entity not dataset name

lynnagara:
The RedisSubscriptionDataStore now stores subscriptions by entity name instead
of dataset name. This is a step towards making all subscriptions based on
an entity rather than a dataset to align with our API. Since multiple
entities (such as events and transactions) are likely to eventually become
part of the same dataset, we should ensure that subscriptions are stored separately
for each entity even if they later become part of the same dataset.

Since all dataset and entity names for our current subscriptions are the same,
no migration is required.

### DIFF
--- a/snuba/cli/subscriptions.py
+++ b/snuba/cli/subscriptions.py
@@ -14,6 +14,7 @@ from arroyo.processing.strategies.batching import BatchProcessingStrategyFactory
 from arroyo.synchronized import SynchronizedConsumer
 
 from snuba import environment, settings
+from snuba.datasets.entities.factory import ENTITY_NAME_LOOKUP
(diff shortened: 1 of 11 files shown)


evanh:
> Since all dataset and entity names for our current subscriptions are the same,
no migration is required.

This is true for events -> errors as well?

lynnagara:
> This is true for events -> errors as well?
Yes it is. Only the name of the storage was changed in the errors storage migration. The entity didn't change at all.

evanh (approved):
I think this is a good change. Per our offline discussion, `get_default_entity` is a hack to facilitate the move from dataset -> entity, so getting rid of it is only a good thing.

PR #2122, opened 2021-09-29
feat(replacements) Add a simple rate limiter to cap the amount of INSERT queries to clickhouse

fpacifici:
The replacer, today, does not cap in any way the amount of INSERT queries sent to clickhouse. This was not too dangerous before re-processing was built as replacement operations were scattered and rare. 
Now we can have a constant amount of replacements being issued for reprocessing over a long period of time. These were used to be batched to reduce the impact on clickhouse but now the batch size has been reduced to only 10 and we can have large amount of replacements issued at high rate over a long period of time.
This can actually cause clickhouse to accumulate inactive parts and to be unable to keep up with merge. Which would cause an outage.
This adds a simple safe guard by capping the amount of INSERT per second you can run on any replacer so that if the issue happens, in the worst case scenario, we would backlog on Kafka, which is still a lot better than a Clickhouse outage.

In the meantime the ingest team is working on re-introducing the batching.


snuba/datasets/replacements/rate_limiter.py
```diff
@@ -0,0 +1,66 @@
...
+
+from snuba import state
+
+RATE_LIMIT_PER_SEC = "mem_rate_limit_per_sec_"
+
+# Stores the rate limit current state.
+# The key is the bucket to keep multiple instances logically apart
+# The value contains the epoch and the number of attempt in that
+# epoch
+BUCKETS: MutableMapping[str, Tuple[int, int]] = {}
```
vladkluev, line 14: Two things about this variable:

1. I don't like having mutable global state without at least a level of indirection. I would prefer it if this was put behind an accessor like the [SerializableExceptionRegistry](https://github.com/getsentry/snuba/blob/master/snuba/utils/serializable_exception.py#L78-L85). 
2. If this isn't thread-safe it should be made threadlocal

vladkluev (changes requested):

evanh (commented):
Is there a reason you didn't want to use Redis for this?

snuba/datasets/replacements/rate_limiter.py
fpacifici: I removed the entire global state, made the rate limiter module reusable (it is in snuba/utils now) and it is now threadsafe.

fpacifici:
@evanh  I did not see the reason for using redis as this is only relevant for one process at a time. There is no need to synchronize between processes and replicas so everything can be in memory. 
If anything redis in this case would introduce additional latency that we do not need.

fpacifici:
Reorganized the code:
- The rate limiter is not in snuba/utils and reusable
- there is no more global state. Each user manages the scope on its own. The instance of the rate limiter carries the state which means the life cycle and the validity of the state is decided by who instantiates it 
- it is now threadsafe

vladkluev (approved):

PR #2124, opened 2021-09-30
Fix deserialized exception message being swallowed by sentry

vladkluev:
Sentry sdk uses `str` to get the `value` field of an exception. SerializableException was not calling the `super` constructor and thus its `str` method was misbehaving. This PR fixes that


fpacifici (approved):
thanks

PR #2127, opened 2021-10-01
feat(subscriptions): Add codec for scheduled subscription task

lynnagara:
Currently subscription scheduling and executing are done together
in a single process by the subscription worker. Soon this will be
split into two separate phases with the scheduled subscriptions published
to a new Kafka topic to be subsequently picked up by the executors.
We will need to be able to serialize the scheduled subscriptions into
bytes so they can be produced to Kafka.

This PR introduces a SubscriptionScheduledTaskEncoder codec which can
encode a ScheduledTask[Subscription] into bytes and those bytes back
to a ScheduledTask[Subscription]. The encoder is not being used yet.

Note that the encoder does not support non SnQL subscriptions.
This is to avoid unnecessary complexity as legacy subscriptions
are expected to be deprecated and completely removed from our
codebase soon.

Also fixes the typing issues in test_codec.py.

### DIFF
--- a/snuba/subscriptions/codecs.py
+++ b/snuba/subscriptions/codecs.py
@@ -1,7 +1,11 @@
 import json
+from datetime import datetime
+from typing import cast
 
+import rapidjson
 from arroyo.backends.kafka import KafkaPayload
 
+from snuba.datasets.entities import EntityKey
 from snuba.datasets.entities.factory import ENTITY_NAME_LOOKUP
 from snuba.datasets.entity import Entity
 from snuba.query.exceptions import InvalidQueryException
@@ -9,7 +13,9 @@ from snuba.subscriptions.data import (
     DelegateSubscriptionData,
     LegacySubscriptionData,
     SnQLSubscriptionData,
+    Subscription,
     SubscriptionData,
+    SubscriptionIdentifier,
 )
 from snuba.subscriptions.entity_subscription import (
     InvalidSubscriptionError,
@@ -17,6 +23,7 @@ from snuba.subscriptions.entity_subscription import (
 )
 from snuba.subscriptions.worker import SubscriptionTaskResult
 from snuba.utils.codecs import Codec, Encoder
+from snuba.utils.scheduler import ScheduledTask
 
 
 class SubscriptionDataCodec(Codec[bytes, SubscriptionData]):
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/codecs.py
```diff
@@ -62,3 +69,44 @@ def encode(self, value: SubscriptionTaskResult) -> KafkaPayload:
             ).encode("utf-8"),
             [],
         )
+
+
+class SubscriptionScheduledTaskEncoder(Encoder[bytes, ScheduledTask[Subscription]]):
```
fpacifici, line 74: `SubscriptionScheduledTaskCodec` ? Why doesn't this extend `Decoder` as well ?

snuba/subscriptions/codecs.py
```diff
@@ -62,3 +69,44 @@ def encode(self, value: SubscriptionTaskResult) -> KafkaPayload:
...
+        self.__entity_key = entity_key
+
+    def encode(self, value: ScheduledTask[Subscription]) -> bytes:
+        assert isinstance(value.task.data, SnQLSubscriptionData)
+
+        return cast(
+            bytes,
+            rapidjson.dumps(
+                {
+                    "timestamp": value.timestamp.isoformat(),
```
fpacifici, line 90: Why do we need to format the date every time? Can't we pass the timestamp as integer instead ? The date format will have an impact when you will try to schedule tons of these at the same time.

snuba/subscriptions/codecs.py
```diff
@@ -62,3 +69,44 @@ def encode(self, value: SubscriptionTaskResult) -> KafkaPayload:
...
+                    "task": {
+                        "identifier": str(value.task.identifier),
+                        "data": value.task.data.to_dict(),
```
fpacifici, line 93: I wonder if we even need to bother serializing the task data instead of just providing the subscription id and the timestamp and let the executor figure out the storage.  Let's keep an eye on the impact of this when you will profile the scheduler.

snuba/subscriptions/codecs.py
lynnagara: Oops, this was supposed to extend Codec` not `Encoder`. Thanks.

snuba/subscriptions/codecs.py
lynnagara: `value.timestamp` is not an integer to begin with, it's a Python datetime object so it would to be formatted each time we schedule. We could format the date as an integer but that wouldn't be any faster than this. The other subscription codec uses isoformat() for timestamp so I think it's better to keep it consistent.

fpacifici: Considering the internal representation (https://github.com/python/cpython/blob/v3.7.0/Include/datetime.h#L10-L22) , I'd bet it will be more efficient to encode and decode into timestamp than into formatted string. 
Though I guess it would have a much bigger impact on decoding (parsing the string to the internal representation), that part is less critical as it would be done on the executor, so maybe we can get away on this. 

Still we should revisit this when we profile the scheduler and we see the actual impact.

fpacifici (approved):

PR #2128, opened 2021-10-01
Communicate read through cache errors in separate key 

vladkluev:
### Context
 
The read through cache was changed to communicate errors to waiting clients by storing the error value in place of the result. The result however has a TTL of 7 minutes in production meaning that if a query hits an error, the same error will be propagated for the next 7 minutes.

### Solution

Store error values in a key of their own. If a waiting process receives a notification and there is nothing in the value key, it checks the error key. The value key only holds true result values.

### Test plan

There is already a test for transient errors not being re-raised introduced in #2125 but now the time between tasks can be much smaller as we're not replying on the TTL to handle that invalidation logic for us.

### DIFF
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -94,6 +94,11 @@ class RedisCache(Cache[TValue]):
         # (or lack thereof.)
         result_key = self.__build_key(key)
 
+        # if we hit an error, we want to communicate that to waiting clients
+        # but we do not want it to be considered a true value. hence we store
+        # the error info in a different key
+        error_key = self.__build_key(key, "error")
+
         # The wait queue (a Redis list) is used to identify clients that are
         # currently "subscribed" to the evaluation of the function and awaiting
         # its result. The first member of this queue is a special case -- it is
@@ -153,6 +158,7 @@ class RedisCache(Cache[TValue]):
                 task_ident,
                 task_timeout,
             )
+            redis_key_to_write_to = result_key
 
             argv = [task_ident, 60]
             try:
@@ -169,11 +175,15 @@ class RedisCache(Cache[TValue]):
                 argv.extend(
                     [
                         self.__codec.encode_exception(error_value),
-                        # NOTE (Vlad): this is only temporary so we don't cache
-                        # error values for too long
+                        # the error data only needs to be present for long enough such that
+                        # the waiting clients know that they all have their queries rejected.
+                        # thus we set it to only one second
                         1,
                     ]
                 )
+                # we want the result key to only store real query results in it as the TTL
+                # of a cached query can be fairly long (minutes).
+                redis_key_to_write_to = error_key
                 raise e
             finally:
                 # Regardless of whether the function succeeded or failed, we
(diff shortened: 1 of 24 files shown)


snuba/state/cache/redis/scripts/set.lua
```diff
@@ -11,22 +20,22 @@
 -- responsible for executing. If it doesn't exist or does not match the current
 -- task ID, we must have overrun the timeout.
 -- TODO: This may still be able to safely set the cache value?
-local task_id = redis.call('GET', KEYS[3])
-if not task_id or task_id ~= ARGV[1] then
+local task_id = redis.call('GET', task_id_key)
+if not task_id or task_id ~= task_id then
```
evanh, line 24: Isn't this variable shadowing, having two locals called `task_id`? My lua is not great.

snuba/state/cache/redis/scripts/set.lua
```diff
@@ -11,22 +20,22 @@
...
-    redis.call('SETEX', KEYS[1], ARGV[4], ARGV[3])
+if value ~= nil then
+    redis.call('SETEX', value_key, value_ttl, value)
 end
 
 -- Move the data from the waiting queue to the notify queue.
-redis.call('RENAME', KEYS[2], KEYS[4])
-redis.call('EXPIRE', KEYS[4], ARGV[2])
+redis.call('RENAME', wait_queue_key, notify_queue_key)
+redis.call('PEXPIRE', notify_queue_key, notify_queue_ttl)
```
evanh, line 35: Do we still need `PEXPIRE` here now that the TTL for exceptions is 1 second?

evanh (commented):
Couple questions.

snuba/state/cache/redis/backend.py
```diff
@@ -231,8 +241,12 @@ def build_notify_queue_key(task_ident: str) -> str:
                 # If there is no value, that means that the client responsible
                 # for generating the cache value errored while generating it.
                 if raw_value is None:
+                    upsteam_error_payload = self.__client.get(error_key)
+                    if upsteam_error_payload:
+                        return self.__codec.decode(upsteam_error_payload)
+
                     raise ExecutionError(
-                        "no value at key: this means the original process executing the query crashed before the exception could be handled"
+                        "no value at key: this means the original process executing the query crashed before the exception could be handled or we were not able to set the cache result"
```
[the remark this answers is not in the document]
vladkluev: dammit, tried so hard to avoid those 😅

snuba/state/cache/redis/scripts/set.lua
vladkluev: yes you're right, great catch

snuba/state/cache/redis/scripts/set.lua
vladkluev: another great catch. Weird that the tests didn't catch this. I will add a test before re-submitting for review

tests/state/test_cache.py
```diff
@@ -219,3 +219,71 @@ def functioning_query() -> bytes:
         setter.result()
 
     assert waiter.result() == b"hello"
+
+
+def test_notify_queue_ttl() -> None:
+    # Tests that waiting clients can be notified of the cache status
+    # even with network delays. This test will break if the notify queue
+    # TTL is set to a very low number
```
evanh, line 227: I would clarify what "a very low number" is here.

evanh (approved):
Small nit, otherwise looks good.

fpacifici (commented):
Please make the ttl configurable before merging.

PR #2130, opened 2021-10-04
feat(subscriptions): Enable Sessions Subscriptions settings

ahmedetefy:
Since the corresponding PR has been merged in Sentry https://github.com/getsentry/sentry/pull/28526, Would like to get this enabled to be able to EA the feature

Corresponding Ops ticket: https://github.com/getsentry/ops/pull/3980

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -146,7 +146,7 @@ TRANSACT_SKIP_CONTEXT_STORE: Mapping[int, Set[str]] = {}
 CLICKHOUSE_ZOOKEEPER_OVERRIDE: Mapping[str, str] = {}
 
 # Metric Alerts Subscription Options
-ENABLE_SESSIONS_SUBSCRIPTIONS = os.environ.get("ENABLE_SESSIONS_SUBSCRIPTIONS", False)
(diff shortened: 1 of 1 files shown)


lynnagara:
Isn't this feature supposed to be only for early adopters? Why removing the flag entirely? This seems like it will start writing the sessions commit log in all environments now. Is that your intention? Your comment doesn't say anything about how this will affect prod, onpremise, etc.

ahmedetefy:
> Isn't this feature supposed to be only for early adopters? Why removing the flag entirely? This seems like it will start writing the sessions commit log in all environments now. Is that your intention? Your comment doesn't say anything about how this will affect prod, onpremise, etc.

Yeah now thinking of it .. that probably makes more sense to just enable it for the time being

lynnagara (changes requested):
This change turns on sessions subscriptions by default in all environments. Specifically we will be writing a commit log everywhere. Are you sure about doing that?

I was under the impression that sessions subscriptions was a temporary product being tested mostly internally and will be superseded soon by metrics subscriptions. If that is true (I could be wrong here) I don't think we necessarily want to add additional burden to single tenant or onpremise for that.

Even if we really do eventually want it in all environments, please let's not do it all in one hit like this. How about turning on the feature flag in production first, run it for a while and only once we are confident about that mainline the feature like you are doing here.

PR #2134, opened 2021-10-06
feat(subscriptions): Add tick buffer processing strategy

lynnagara:
Adds a processing strategy that buffers ticks before submitting them
to the next processing step in the pipeline. This is not currently
being used anywhere but will be used by the future subscription scheduler.

The tick buffer has two modes, which correspond to the two different
ways we will be doing subscription scheduling.

- Partition mode immediately submits a tick to the next step when
we consume it. The use case for this is for subscriptions on topics
that are semantically partitioned by project ID. When we get a tick,
we can safely schedule it immediately since all the project's messages
are in the same partition and order is already guaranteed within a
single project.

- Global mode buffers ticks in memory and only
submits a tick to the next step once the upper timestamp of that tick
has been reached on every partition. The use case for this mode is for
subscriptions on topics that are partitioned randomly or by something other
than the project ID. Waiting for all partitions to reach a timestamp provides
a guarantee that all data up to that point has been written to storage.
This can also be used to run queries spanning multiple projects, as waiting
ensures our data up to that timestamp is complete before running the
subscription query. An integer value for `max_ticks_buffered_per_partition`
must be provided for this strategy. Since we cannot infinitely buffer ticks,
once this max value is reached, ticks will be submitted to the next step even
if not all partitions have reached that point yet.

The TickBuffer strategy currently records the lag between the slowest and
fastest partition (similarly to what is done in the temporary `MeasurePartitionLag`
strategy). In future, we will likely want to record more metrics here, such
as the count of buffered ticks. This strategy will likely also need to
be responsible for committing offsets.

### DIFF
--- a/snuba/datasets/entities/metrics.py
+++ b/snuba/datasets/entities/metrics.py
@@ -34,9 +34,13 @@ from snuba.query.extensions import QueryExtension
 from snuba.query.logical import Query
 from snuba.query.processors import QueryProcessor
 from snuba.query.processors.basic_functions import BasicFunctionsProcessor
+from snuba.query.processors.granularity_processor import GranularityProcessor
 from snuba.query.processors.project_rate_limiter import ProjectRateLimiterProcessor
 from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
-from snuba.query.validation.validators import EntityRequiredColumnValidator
+from snuba.query.validation.validators import (
+    EntityRequiredColumnValidator,
+    GranularityValidator,
+)
 from snuba.request.request_settings import RequestSettings
 
 
@@ -95,7 +99,10 @@ class MetricsEntity(Entity, ABC):
             ),
             join_relationships={},
             writable_storage=writable_storage,
-            validators=[EntityRequiredColumnValidator({"org_id", "project_id"})],
+            validators=[
+                EntityRequiredColumnValidator({"org_id", "project_id"}),
+                GranularityValidator(minimum=10),
+            ],
             required_time_column="timestamp",
         )
 
@@ -105,6 +112,7 @@ class MetricsEntity(Entity, ABC):
     def get_query_processors(self) -> Sequence[QueryProcessor]:
         return [
             BasicFunctionsProcessor(),
+            GranularityProcessor(),
             TimeSeriesProcessor({"bucketed_time": "timestamp"}, ("timestamp",)),
             ProjectRateLimiterProcessor(project_column="project_id"),
             TagsTypeTransformer(),
--- a/snuba/datasets/metrics.py
+++ b/snuba/datasets/metrics.py
@@ -5,6 +5,8 @@ from snuba.datasets.entities import EntityKey
 from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.entity import Entity
 
+DEFAULT_GRANULARITY = 60
+
 
 class MetricsDataset(Dataset):
     def __init__(self) -> None:
--- a/snuba/datasets/storages/discover.py
+++ b/snuba/datasets/storages/discover.py
@@ -13,8 +13,10 @@ from snuba.clusters.storage_sets import StorageSetKey
 from snuba.datasets.schemas.tables import TableSchema
 from snuba.datasets.storage import ReadableTableStorage
 from snuba.datasets.storages import StorageKey
+from snuba.datasets.storages.errors import storage as error_storage
 from snuba.datasets.storages.errors_common import mandatory_conditions
 from snuba.datasets.storages.events_bool_contexts import EventsBooleanContextsProcessor
+from snuba.datasets.storages.transactions import storage as transactions_storage
 from snuba.query.processors.arrayjoin_keyvalue_optimizer import (
     ArrayJoinKeyValueOptimizer,
 )
@@ -23,6 +25,7 @@ from snuba.query.processors.empty_tag_condition_processor import (
 )
 from snuba.query.processors.mapping_optimizer import MappingOptimizer
 from snuba.query.processors.mapping_promoter import MappingColumnPromoter
+from snuba.query.processors.null_column_caster import NullColumnCaster
 from snuba.query.processors.prewhere import PrewhereProcessor
 from snuba.query.processors.type_converters.hexint_column_processor import (
     HexIntColumnProcessor,
@@ -105,6 +108,7 @@ storage = ReadableTableStorage(
                 "project_id",
             ]
         ),
+        NullColumnCaster([transactions_storage, error_storage]),
     ],
     query_splitters=[
         ColumnSplitQueryStrategy(
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -165,6 +165,9 @@ class MetricsLoader(DirectoryLoader):
             "0004_metrics_counters",
             "0005_metrics_distributions_buckets",
             "0006_metrics_distributions",
+            "0007_metrics_sets_granularity_10",
+            "0008_metrics_counters_granularity_10",
+            "0009_metrics_distributions_granularity_10",
         ]
 
 
--- a/snuba/migrations/snuba_migrations/metrics/0002_metrics_sets.py
+++ b/snuba/migrations/snuba_migrations/metrics/0002_metrics_sets.py
@@ -5,6 +5,7 @@ from snuba.migrations import migration, operations
 from snuba.migrations.snuba_migrations.metrics.templates import (
     get_forward_migrations_dist,
     get_forward_migrations_local,
+    get_migration_args_for_sets,
     get_reverse_table_migration,
 )
 
@@ -13,15 +14,7 @@ class Migration(migration.ClickhouseNodeMigration):
     blocking = False
 
     def forwards_local(self) -> Sequence[operations.SqlOperation]:
-        return get_forward_migrations_local(
-            source_table_name="metrics_buckets_local",
-            table_name="metrics_sets_local",
-            mv_name="metrics_sets_mv_local",
-            aggregation_col_schema=[
-                Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
-            ],
-            aggregation_states="uniqCombined64State(arrayJoin(set_values)) as value",
-        )
+        return get_forward_migrations_local(**get_migration_args_for_sets(),)
 
     def backwards_local(self) -> Sequence[operations.SqlOperation]:
         return [
--- a/snuba/migrations/snuba_migrations/metrics/0004_metrics_counters.py
+++ b/snuba/migrations/snuba_migrations/metrics/0004_metrics_counters.py
@@ -5,6 +5,7 @@ from snuba.migrations import migration, operations
 from snuba.migrations.snuba_migrations.metrics.templates import (
     get_forward_migrations_dist,
     get_forward_migrations_local,
+    get_migration_args_for_counters,
     get_reverse_table_migration,
 )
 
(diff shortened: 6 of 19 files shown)


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+        self.__buffers: Mapping[int, Deque[Message[Tick]]] = {
+            index: deque() for index in range(self.__partitions)
+        }
+
+        # Stores the latest timestamp we received for any partition. This is
+        # just for recording the partition lag.
+        self.__latest_ts: Optional[datetime] = None
+
+    def poll(self) -> None:
+        pass
```
fpacifici, line 68: I think you need to poll from on `self.__next_step`

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def submit(self, message: Message[Tick]) -> None:
+        # If the scheduler mode is immediate or there is only one partition
+        # immediately submit message to the next step.
+        # We don't keep any latest_ts values as it is not relevant.
+        if self.__mode == SchedulerMode.IMMEDIATE or self.__partitions == 1:
+            self.__next_step.submit(message)
+            return
```
fpacifici, line 76: Why not having the SchedulerMode.IMMEDIATE as a separate step (or even no step at all) and pick it when we build the strategy factory at initialization? It seems there is no reason for fitting both strategies (very different responsibilities) in the same class when the processingStrategy is already designed to make composition easy.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+            self.__latest_ts = message.payload.timestamps.upper
+
+        tick_partition = message.payload.partition
+        self.__buffers[tick_partition].append(message)
+
+        # If the buffer length exceeds `max_ticks_buffered_per_partition`
+        # immediately submit the earliest message in that buffer to the next step.
+        if len(self.__buffers[tick_partition]) > cast(
+            int, self.__max_ticks_buffered_per_partition
+        ):
```
fpacifici, line 92: Is there any advantage tracking the maximum amount of ticks per partition in terms of number of commit messages instead of time? IT seems that if we set a limit in terms of time, the behavior here would be independent from the granularity of the commit messages. With this current strategy, instead, the bigger the batch we set the more sparse the commit messages, thus the longer it would take to identify a late partition.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+
+                if partition_ts < earliest_ts:
+                    earliest_ts = tick.timestamps.upper
+                    earliest_ts_partitions = {tick.partition}
+
+                elif partition_ts == earliest_ts:
+                    earliest_ts_partitions.add(tick.partition)
+
+            for partition_index in earliest_ts_partitions:
+                self.__next_step.submit(self.__buffers[partition_index].popleft())
```
fpacifici, line 127: So this means that you are producing all the same ticks you received in input, just you produce them when all partitions reached that timestamp. Which means you are producing twice the same tick (for two different partitions) if you had to ticks with the same `upper` timestamp.   What was the reason for going this way instead of producing a single watermark tick valid across partitions when that timestamp is reached ?

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+            break
+
+    def close(self) -> None:
+        pass
+
+    def terminate(self) -> None:
+        pass
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        pass
```
fpacifici, line 150: I think you need to propagate the calls to these methods to the next step.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+            return
+
+        # If there are any empty buffers, we can't submit anything yet.
+        # Otherwise if all the buffers have ticks then we look for the partition/s
+        # with the earliest tick (i.e. the tick with the earliest upper timestamp
+        # interval value) and submit it to the next step.
+        if len(self.__buffers[tick_partition]) > 1:
+            return
+
+        while True:
```
fpacifici, line 103: Can't you replace this `while True:` (which always makes me nervous on a loop that has nested loops and that is fairly long) with a loop like this:
```
# Keep doing it as long as there is a tick per partition in the buffer.
while all(len(buffer) > 0 for buffer in self.__buffers.values):
     # populate earliest_ts_partitions
     # send the ticks in earliest_ts_paritions
     # record metrics
```

It will be shorter and won;t need break nor continue

fpacifici (changes requested):
I like this idea. There are some comments to address.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(
+        self,
+        mode: SchedulerMode,
+        partitions: int,
+        max_ticks_buffered_per_partition: Optional[int],
+        next_step: ProcessingStrategy[Tick],
+        metrics: MetricsBackend,
+    ) -> None:
+        if mode == SchedulerMode.WAIT_FOR_SLOWEST_PARTITION:
+            assert max_ticks_buffered_per_partition is not None
```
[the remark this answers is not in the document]
lynnagara: I think making this optional gives the false impression that is is ok to infinitely buffer ticks when it is not. The problem if this starts to error is that even on a restart it will continue to error and won't be able to make any progress. If we have an incident where a single partition falls far behind the others (i.e. we hit some buffer limit and start erroring) we would not be able to recover without manual intervention probably from someone who has at least some context about what this code is doing. I'd worry that it can compound our problems in the middle of an incident. Now it may be that we consider this scenario unlikely and don't want to worry about it, but this was my reason for forcing a value to be provided.

We can add other type of alerting on this later that does not actually error so hitting the buffer max still happens loudly enough for us to know something is wrong.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+
+        self.__mode = mode
+        self.__partitions = partitions
```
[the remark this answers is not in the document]
lynnagara: I think the subject of topic and partition management is actually quite separate to subscriptions, and there are other impacts that are not just about subscriptions, so I don't know that this is easily resolved here.

Today the way of dealing with Kafka topics in Snuba is a bit different depending on the environment.

In on-premise, single tenant and dev at least, the topic configuration comes from Snuba, and the topic gets created with the number of partitions specified via settings during bootstrap. The Snuba setting TOPIC_PARTITION_COUNTS acts as the source of truth for the partition configurations. This is the assumption on which the scheduler works today, the current subscriptions also works, and other parts of our codebase do.

The implication of your suggestion seems to be that Snuba shouldn't be responsible for managing topics at all, it should just read live configuration from Kafka on startup or even periodically and use that anywhere it needs to know the partition count. Basically someone else should be responsible for managing these configurations.

I think there are good arguments for pursuing something like what you suggest but we've never really discussed it before and it seems quite different to how our codebase works today. So I am pretty hesitant to go ahead and build something just for subscriptions that is different from how we do things elsewhere. I think there is really no pressing need or immediate benefit from adding this to subscriptions immediately, and it's kind of tangential to how the scheduler works.

Anyway it seems to me if we go down this route of using configuration from the admin API then we will also want to make some other changes to Snuba including deprecating settings.TOPIC_PARTITION_COUNTS. At that point it would probably make sense to just have `KafkaTopicSpec.partitions_number` (which in turn checks settings today) handle fetching the value and not necessarily even changing the scheduler implementation at all since that already calls that same function.

Ultimately I see this as a mostly separate problem to subscription scheduling, since it might affect other things as well. Until there is a clearer long term direction I would prefer to keep using `KafkaTopicSpec.partitions_number` not because I'm advocating for doing it this way longer term but mostly because I'd rather preserve the status quo rather than introduce a new way of doing things in one place, at least until we have some more agreement on the overall direction.

Either way, I don't think it makes any difference to this strategy as it will most likely always get passed a partition count in the constructor from someone else outside who will be doing the fetching.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Ticks seemed the most straightforward to me since it's number of ticks (not time between them) that are the ones actually filling are buffers and potentially causing problems if we have too many. I think having a max time as well might also make sense but to me that more of a product concern about when we want to call a partition "too late". My primary motivation in limiting the number of ticks is only to prevent infinite growth of buffers, and not really about the product behaviour which I think having the max in terms of time addresses.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Perhaps this is overthinking things but I deliberately did `while True` in order to avoid the unnecessary check of the 64 (well... 64 in our specific case right now) buffers on each iteration of this loop. I made a change to get rid of some of the earlier break/continue stuff that was a bit confusing. Not sure the current implementation makes you any less nervous though.

snuba/subscriptions/scheduler_processing_strategy.py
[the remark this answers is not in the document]
fpacifici: What if we had a topic where consistency across partition was so important that we did not want to allow it to proceed and where the manual intervention was preferrable. 
It should not be the common practice as most topics would not need it, but still allowed. 
I am ok adding it separately, though please still add warnings logged when we decide to move ahead anyway.

snuba/subscriptions/scheduler_processing_strategy.py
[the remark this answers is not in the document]
fpacifici: I agree this may not impact the strategy (wrote it above).
I forgot about bootstrap, though I have some vague memories about us discussing whether we should have a separate set of parameters for initialization of a topic (for bootstrap) from a way to access the current states of some parameters.

In practice, you are right, the situation now is ambiguous. We do not make a distinction between the two cases.
I think the root cause is exactly this. There should be an initial configuration somewhere (settings can be ok), but then Kafka settings are generally modified outside of snuba and snuba needs to adapt to those changes instead of driving them. At least this is what would happen today in all environment as we do not provide a way (through snuba) to manage the topics after creation.

If Anton needed to add partitions to the events topic in prod today, he would use the kafka console and snuba would know nothing about it, which would be disastrous. 
Let's ensure we fix this process in the next quarter. While adding partitions to events is reasonably unlikely, other topics we are adding alerts onto are more likely to change the partitions number.

snuba/subscriptions/scheduler_processing_strategy.py
fpacifici: So if the concern is the buffer size (fair), the maximum should be a setting and not an argument we pass to the consumer as it is not related to the product functionality, or to ensure we make progress. It is meant to just preserve the integrity of the consumer itself.

snuba/subscriptions/scheduler_processing_strategy.py
fpacifici: I would not freak out about 64 (or even 128 or 256) iterations over the buffer. That is not going to be your bottleneck in this consumer by a long shot as you will have Kafka consumption, production, message serialization and generation of a message per subscription (which are 2 orders of magnitude more).

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -0,0 +1,156 @@
...
+
+        tick_partition = message.payload.partition
+        self.__buffers[tick_partition].append(message)
+
+        # If the buffer length exceeds `max_ticks_buffered_per_partition`
+        # immediately submit the earliest message in that buffer to the next step.
+        if len(self.__buffers[tick_partition]) > cast(
+            int, self.__max_ticks_buffered_per_partition
+        ):
+            self.__next_step.submit(self.__buffers[tick_partition].popleft())
```
fpacifici, line 96: If you get here please log a warning.

fpacifici (commented):
I don't think you need to be concerned about scanning the 64 elements of the buffer. See comment inline. And you can make that code more intuitive.

Other than that, there are just a couple of comments otherwise it is good.

snuba/subscriptions/scheduler_processing_strategy.py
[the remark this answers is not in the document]
lynnagara: Yeah if that requirement ever comes up, it can be added as a 3rd strategy

fpacifici (approved):

PR #2135, opened 2021-10-06
build(freight): Add "events-subscriptions-scheduler" to Freight deployment

tonyo:
Companion to https://github.com/getsentry/ops/pull/3966
Freight needs this to find and update image for the corresponding container.


lynnagara (approved):
Thanks

PR #2136, opened 2021-10-06
feat(subscriptions): Use scheduler command to debug commit log issues (temporarily)

lynnagara:
Upon turning on the scheduler process in production for testing, we experienced
a lot of issues with invalid intervals. The lower bound of the interval constructed
by the CommitLogTickConsumer was frequently higher than the upper bound, which is invalid.

One potential cause of this could be messages being out of their expected order in
the commit log. Though this is known to be possible due to consumer crashes, rebalancing, etc
the frequency with whicih this appeared to happen seemed unusual.

This PR aims to change the scheduler consumer to simply collect more information about
commit log ordering to understand how common out of order messages might be.
For each message consumed, we simply record whether it is earlier or later than the
last one we received for that partition.

In order to avoid introducing other bugs or complexities to interval calculations,
it tries to be as minimalistic as possible and uses just a standard consumer, not the
tick consumer.

### DIFF
--- a/snuba/subscriptions/scheduler_consumer.py
+++ b/snuba/subscriptions/scheduler_consumer.py
@@ -282,7 +282,7 @@ class MeasureCommitLogOrderMetrics(ProcessingStrategy[KafkaPayload]):
         current_message = MessageDetails(commit.offset, commit.orig_message_ts)
 
         # Record metrics about in order vs out of order messages
-        last_message = self.__previous_messages[commit.partition]
+        last_message = self.__previous_messages.get(commit.partition)
         if last_message is not None:
             current_message = MessageDetails(commit.offset, commit.orig_message_ts)
             try:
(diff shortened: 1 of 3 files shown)


snuba/subscriptions/scheduler_consumer.py
```diff
@@ -243,6 +249,70 @@ def __build_tick_consumer(self) -> CommitLogTickConsumer:
...
+        self.__commit({message.partition: Position(message.offset, message.timestamp)})
+
+    def close(self) -> None:
+        pass
+
+    def terminate(self) -> None:
+        pass
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        pass
```
fpacifici, line 313: You may want to call the next step methods even if this is temporary code, just to be consistent.

PR #2138, opened 2021-10-07
Perf/avoid final based on query time range

rahul-kumar-saini:
## Overview
- Queries are set as FINAL if they touch projects that currently have unmerged replacements
- This change introduces logic to avoid setting a query as FINAL if the projects it touches have replacements that have a time range before this query's time range
- This is checked using the latest "exclude groups" replacement per project for the query, finding the latest replacement overall and comparing it to the time range of the query

## Changes
- Created new function in `errors_replacer.py` to get the latest timestamp for a replacement given project ids
- Tested said function
- Added a check to see if the query overlaps replacements before setting the query as final

#### Closes SNS-771

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,8 @@ RUN set -ex; \
     '; \
     apt-get update; \
     apt-get install -y $buildDeps --no-install-recommends; \
+    # Since there's no confluent-kafka wheel for aarch64, remove when there is
+    [ $(uname -m) = "aarch64" ] && apt-get install -y librdkafka-dev --no-install-recommends; \
     \
     pip install -r requirements.txt; \
     \
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,7 +43,7 @@ python-rapidjson==1.4
 redis==2.10.6
 redis-py-cluster==1.3.5
 sentry-arroyo==0.0.4
-sentry-relay==0.8.7
+sentry-relay==0.8.9
 sentry-sdk==1.1.0
 simplejson==3.15.0
 urllib3==1.26.5
--- a/snuba/datasets/transactions_processor.py
+++ b/snuba/datasets/transactions_processor.py
@@ -27,7 +27,7 @@ from snuba.processor import (
     _ensure_valid_ip,
     _unicodify,
 )
-from snuba.state import is_project_in_rollout_group
+from snuba.state import get_config, is_project_in_rollout_group
 from snuba.utils.metrics.wrapper import MetricsWrapper
 
 logger = logging.getLogger(__name__)
@@ -292,20 +292,38 @@ class TransactionsMessageProcessor(MessageProcessor):
         trace_context = data["contexts"]["trace"]
 
         try:
+            max_spans_per_transaction = get_config("max_spans_per_transaction", 2000)
+            assert isinstance(max_spans_per_transaction, (int, float))
+        except Exception:
+            metrics.increment("bad_config.max_spans_per_transaction")
+            max_spans_per_transaction = 2000
+
+        try:
             if not is_project_in_rollout_group(
                 "write_span_columns_projects", processed["project_id"]
             ):
                 return
 
+            num_processed = 0
             processed_spans = []
 
             processed_root_span = self._process_span(trace_context)
             if processed_root_span is not None:
+                num_processed += 1
                 processed_spans.append(processed_root_span)
 
             for span in data.get("spans", []):
+                # The number of spans should not exceed 1000 as enforced by SDKs.
+                # As a safety precaution, enforce a hard limit on the number of
+                # spans we actually store .
+                if num_processed >= max_spans_per_transaction:
+                    metrics.increment("too_many_spans")
+                    break
+
                 processed_span = self._process_span(span)
+
                 if processed_span is not None:
+                    num_processed += 1
                     processed_spans.append(processed_span)
 
             processed["spans.op"] = []
--- a/tests/datasets/test_transaction_processor.py
+++ b/tests/datasets/test_transaction_processor.py
@@ -352,3 +352,52 @@ class TestTransactionsProcessor:
             message.serialize(), meta
         ) == InsertBatch([message.build_result(meta)], None)
         settings.TRANSACT_SKIP_CONTEXT_STORE = old_skip_context
+
+    def test_too_many_spans(self) -> None:
+        old_skip_context = settings.TRANSACT_SKIP_CONTEXT_STORE
+        settings.TRANSACT_SKIP_CONTEXT_STORE = {1: {"experiments"}}
+        set_config("write_span_columns_projects", "[1]")
+        set_config("max_spans_per_transaction", 1)
+
+        start, finish = self.__get_timestamps()
+        message = TransactionEvent(
+            event_id="e5e062bf2e1d4afd96fd2f90b6770431",
+            trace_id="7400045b25c443b885914600aa83ad04",
+            span_id="8841662216cc598b",
+            transaction_name="/organizations/:orgId/issues/",
+            status="cancelled",
+            op="navigation",
+            timestamp=finish,
+            start_timestamp=start,
+            platform="python",
+            dist="",
+            user_name="me",
+            user_id="myself",
+            user_email="me@myself.com",
+            ipv4="127.0.0.1",
+            ipv6=None,
+            environment="prod",
+            release="34a554c14b68285d8a8eb6c5c4c56dfc1db9a83a",
+            sdk_name="sentry.python",
+            sdk_version="0.9.0",
+            http_method="POST",
+            http_referer="tagstore.something",
+            geo={"country_code": "XY", "region": "fake_region", "city": "fake_city"},
+        )
+        meta = KafkaMessageMetadata(
+            offset=1, partition=2, timestamp=datetime(1970, 1, 1)
+        )
+
+        payload = message.serialize()
+
+        # there are 2 spans in the transaction but only 1
+        # will be inserted because of the limit set above
+        result = message.build_result(meta)
+        result["spans.op"] = ["navigation"]
+        result["spans.group"] = [int("a" * 16, 16)]
+        result["spans.exclusive_time"] = [1.2345]
+
+        assert TransactionsMessageProcessor().process_message(
+            payload, meta
+        ) == InsertBatch([result], None)
+        settings.TRANSACT_SKIP_CONTEXT_STORE = old_skip_context


snuba/datasets/errors_replacer.py
```diff
@@ -314,6 +314,32 @@ def process_exclude_groups_and_replacement_types_results(
     return needs_final, exclude_groups, replacement_types
 
 
+def get_latest_replacement_time_by_projects(
+    project_ids: Sequence[int], state_name: Optional[ReplacerState]
+) -> Optional[float]:
```
fpacifici, line 319: please return a datetime. It will be much easier and safe to manage than a bare float. 
It also makes the interface a lot more intuitive. If you see a float in the return value you don't know what it is an you need to read through the code.
If you return a date time you immediately know what it is.

snuba/datasets/errors_replacer.py
```diff
@@ -314,6 +314,32 @@ def process_exclude_groups_and_replacement_types_results(
...
+    p = redis_client.pipeline()
+
+    for project_id in project_ids:
+        key, _ = get_project_exclude_groups_key_and_type_key(project_id, state_name)
+        # Gets the item with the highest score for a key, along with the score itself
+        p.zrevrangebyscore(
+            key, float("inf"), float("-inf"), withscores=True, start=0, num=1
+        )
+    # results are in the form [[(str, float)], [(str, float)], ...]
+    results = p.execute()
```
fpacifici, line 333: I don't think this is going to be enough to provide the effect you want.
If we add the condition to exclude groups to a query even if that does not overlap with the replacement, the performance issue is minimal with the only exception of queries that are demoted to FINAL because the number of groups to exclude is too high.

What we would like to avoid is to apply FINAL to a query that does not overlap with the replacement.
Which means the redis key you are interested into the is FINAL key (which will need to be provided with a timestamp) not the group exclusion key.

Were you planning to do this as an intermediate step? That would be fine. Otherwise remember there is an additional step to do.

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -43,40 +51,72 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
         set_final = False
         if project_ids:
-            final, exclude_group_ids, replacement_types = get_projects_query_flags(
-                list(project_ids), self.__replacer_state_name,
-            )
+            (
+                query_is_final,
+                exclude_group_ids,
+                replacement_types,
+            ) = get_projects_query_flags(list(project_ids), self.__replacer_state_name,)
```
fpacifici, line 58: please remove the comma

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -43,40 +51,72 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
+                    if len(exclude_group_ids) > max_group_ids_exclude:
+                        metrics.increment(
+                            "final",
+                            tags={
+                                "cause": "max_groups",
+                                "referrer": request_settings.referrer,
+                                "parent_api": request_settings.get_parent_api(),
+                            },
+                        )
+                        set_final = True
```
fpacifici, line 92: This is the second condition twice.

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -43,40 +51,72 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
-                max_group_ids_exclude = get_config(
-                    "max_group_ids_exclude", settings.REPLACER_MAX_GROUP_IDS_TO_EXCLUDE
-                )
-                assert isinstance(max_group_ids_exclude, int)
-                if len(exclude_group_ids) > max_group_ids_exclude:
-                    tags["cause"] = "max_groups"
-                    metrics.increment(
-                        "final", tags=tags,
+            if not query_is_final and exclude_group_ids:
+                if self._query_overlaps_replacements(query, project_ids):
```
fpacifici, line 69: Please invert the condition, the code will be less nested and easier to read sequentially. 
```
if not self._query_overlaps_replacements(...):
    return
```

fpacifici (changes requested):
Please see comments inline. Also update the tests for the ReplacedGroupProcessor as well.

snuba/datasets/storages/processors/replaced_groups.py
rahul-kumar-saini: Since that last line of the function still needs to run, it would look like:
```
if not self._query_overlaps_replacements(...):
    query.set_from_clause(replace(query.get_from_clause(), final=False))
    return
...
```
It'll reduce 1 nesting level but it means that `query.set_from_clause()` would need to be duplicated there.

snuba/datasets/errors_replacer.py
```diff
@@ -314,6 +314,55 @@ def process_exclude_groups_and_replacement_types_results(
...
+
+    for project_id in project_ids:
+        exclude_groups_key, _ = get_project_exclude_groups_key_and_type_key(
+            project_id, state_name
+        )
+        project_needs_final_key, _ = get_project_needs_final_key_and_type_key(
+            project_id, state_name
+        )
+        # Gets the item with the highest score for a key, along with the score itself
+        p.zrevrangebyscore(
```
evanh, line 335: In theory you could do `zrevrange` with `0, 0`. Since our data is sorted by timestamp, the value at the end of the set is by definition the value with the highest timestamp.

However I think it would be better to maybe try to move this logic behind the existing redis call.

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -71,12 +83,27 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
 
         query.set_from_clause(replace(query.get_from_clause(), final=set_final))
+
+    def _query_overlaps_replacements(self, query: Query, project_ids: Set[int]) -> bool:
+        """
+        Given a Query and a set of project ids the query "touches", returns whether
+        or not this Query's time range overlaps with any replacement for any of the
+        projects.
+        """
+        query_from, _ = get_time_range(query, "timestamp")
```
evanh, line 102: I don't think this logic is correct. Shouldn't we be comparing `query_to` instead of `query_from`? Then the check would be `return latest_replacement_time <= query_to` or "did the replacement happen before or after the time period the query is querying".

rahul-kumar-saini: From my understanding, a replacement could be touching data an unknown amount of time back. So if a replacement happens at a certain time, any query looking at data before that time would be "overlapping". For now, a query doesn't need final if it's looking at data that has no replacements yet.

`.......Time ->....`


Query doesn't overlap:

`...Rep..From..To..` <- This case would say the query does overlap if we `return latest <= query_to`


Query overlaps:

`..From..Rep..To...` 
`.From..To..Rep....` <- This case would say the query doesn't overlap if we `return latest <= query_to`

If my understanding is incorrect, I can change the logic.

snuba/datasets/errors_replacer.py
fpacifici: No, you will have to add a timestamp and keep it up to date to the FINAL case as well, to make it work.
If you rely on the TTL, your change won't have any effect as, when the TTL is passed, the key will be removed and FINAL will not be applied anyway.

snuba/datasets/errors_replacer.py
```diff
@@ -259,6 +266,14 @@ def get_projects_query_flags(
         p.zremrangebyscore(type_key, float("-inf"), now - settings.REPLACER_KEY_TTL)
         p.zrevrangebyscore(type_key, float("inf"), now - settings.REPLACER_KEY_TTL)
 
+    for exclude_groups_key, _ in exclude_groups_keys_and_types:
+        p.zrevrange(
```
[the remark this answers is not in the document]
rahul-kumar-saini: This `zrevrange` gets the latest item only and has `withscores=True` so we can get its score. The other call is getting all items but without the scores. I could change it so that call also responds with the scores but it would be a lot of extra data we don't need since only the latest score matters.

rahul-kumar-saini:
In order to keep track of when a "project needs final" replacement occurs, I swapped out `<key>:True` for `<key>:<timestamp>` in `set_project_needs_final()`. I've added a backwards compatibility check in the code which can be removed later.

I've also moved the overlap check to the beginning of `process_query()`.

snuba/datasets/errors_replacer.py
```diff
@@ -206,93 +213,129 @@ def set_project_needs_final(
...
+        latest_exclude_groups_replacements: Sequence[Optional[Tuple[group_id, datetime]]]...
     ]
     - `needs_final` slice is `len_projects` long
     - `excludes_groups` slice is `len_projects * 2` long
     - `needs_final_replacement_types` slice is `len_projects` long
     - `groups_replacement_types` slice is `len_projects * 2` long
+    - `latest_exclude_groups_replacements` slice is `len_projects` long
+
+    [needs_final   |exclude_groups    |...]
+    [ len(projects)|len(projects) * 2 |   ]
```
vladkluev, line 324: either remove this explanation or replace the above one with what it actually is

snuba/datasets/errors_replacer.py
```diff
@@ -206,93 +213,129 @@ def set_project_needs_final(
...
         p.get(needs_final_key)
 
     exclude_groups_keys_and_types = [
-        get_project_exclude_groups_key_and_type_key(project_id, state_name)
+        build_project_exclude_groups_key_and_type_key(project_id, state_name)
         for project_id in s_project_ids
     ]
-
     for exclude_groups_key, _ in exclude_groups_keys_and_types:
+        # remove stale excluded groups
```
evanh, line 267: Nit: some of these comments of seem unnecessary.

snuba/datasets/errors_replacer.py
```diff
@@ -206,93 +213,124 @@ def set_project_needs_final(
...
     4. Fetch replacement types for each Project
+    5. Fetch latest exclude groups replacement type for each Project
 
-    Returns (needs_final, group_ids_to_exclude, replacement_types)
+    Returns (
+        needs_final,
+        group_ids_to_exclude,
+        replacement_types,
+        latest_replacement_time
+    )
```
fpacifici, line 247: This does not add any information with respect to the return type structure. You do not need to explain that here, and it can be counter productive if we update the type and forget to update the comment.

Remove please.

snuba/datasets/errors_replacer.py
```diff
@@ -311,7 +349,45 @@ def process_exclude_groups_and_replacement_types_results(
...
+    latest_exclude_groups_result: List[Any],
+) -> Optional[datetime]:
+    latest_replacements = set()
+    if needs_final:
+        latest_need_final_replacement_times = [
+            # Backwards compatibility: Before it was simply "True" at each key,
+            # now it's the timestamp at which the key was added.
+            # This should stay for at least 24h after this change is deployed.
+            # `and timestamp != b"True"` can be removed when there are no
+            # longer any "key:True" entries
```
fpacifici, line 373: This has to stay indefinitely as you do not know when single tenant and self hosted deployment will upgrade and use this version. It may happen a year from now.

snuba/datasets/errors_replacer.py
```diff
@@ -168,25 +168,32 @@ def set_project_exclude_groups(
     project_id: int,
     group_ids: Sequence[int],
     state_name: Optional[ReplacerState],
+    #  replacement type is just for metrics, not necessary for functionality
     replacement_type: ReplacementType,
 ) -> None:
     """
+    This method is called when a replacement comes in. For a specific project, record
+    the group ids which were deleted as a result of this replacement
+
```
fpacifici, line 177: ok, this time, though, going on, I would discourage you from mixing cosmetics that are unrelated with your PR with the logic of your PR.
That makes reviewing harder because it means spending time to understand why a given change is there.

snuba/datasets/errors_replacer.py
fpacifici: to echo what Evan said, commenting to explain what the code does generally means your code is not readable enough. Fixing the code is the preferred option instead of adding comments. So if you feel you needed to explain that, probably refactoring is needed. Here I doubt that comment is needed entirely.

Commenting on the "why" the code does something is different and can be useful a lot. Basically the idea is that the comment provides context, it does not explains what we can already read in the code.

In this specific case, I think you are explaining everything in the docstring. No need to pollute the code, if you want to add more context, please feel free to beef up the docstring

snuba/datasets/errors_replacer.py
```diff
@@ -206,93 +213,124 @@ def set_project_needs_final(
...
     results = p.execute()
 
-    return process_exclude_groups_and_replacement_types_results(
+    return _process_exclude_groups_and_replacement_types_results(
         results, len(s_project_ids)
     )
 
 
-def process_exclude_groups_and_replacement_types_results(
+def _process_exclude_groups_and_replacement_types_results(
```
fpacifici, line 302: This method and the following seem to have the only goal of producing the the `ProjectsQueryFlags` instance from the bare results you got from Redis. 

Now the `ProjectQueryFlags` data class is just a bag of data. It seems these methods should be part of that class as a static factory method. 

```
class ProjectsQueryFlags:
....
@classmethod
def load_from_redis(cls, project_ids: Sequence[int], state_name: Optional[ReplacerState]):  ProjectsQueryFlags:
   ...the whole logic ...
```

This encapsulates these methods into a class that is the only place concerned with them.

Please do that in a separate PR

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -43,40 +51,64 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
 
         set_final = False
         if project_ids:
-            final, exclude_group_ids, replacement_types = get_projects_query_flags(
-                list(project_ids), self.__replacer_state_name,
+            flags: ProjectsQueryFlags = get_projects_query_flags(
```
fpacifici, line 54: the PR does not test the change to this query processor. Please let's cover the new case
https://github.com/getsentry/snuba/blob/master/tests/datasets/storages/processors/test_replaced_groups.py

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -43,40 +51,64 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
-            if not final and exclude_group_ids:
-                # If the number of groups to exclude exceeds our limit, the query
-                # should just use final instead of the exclusion set.
-                max_group_ids_exclude = get_config(
-                    "max_group_ids_exclude", settings.REPLACER_MAX_GROUP_IDS_TO_EXCLUDE
-                )
-                assert isinstance(max_group_ids_exclude, int)
-                if len(exclude_group_ids) > max_group_ids_exclude:
-                    tags["cause"] = "max_groups"
+            # Continue process if the query's time range overlaps with any replacements
```
fpacifici, line 57: The code is clear here. Please remove the comment. If you want you can add some context in the docstring.

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -43,40 +51,64 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
+            if self._query_overlaps_replacements(query, flags.latest_replacement_time):
```
fpacifici, line 58: Please invert this condition and exit early if the query does not overlap the replacements. This reduces the indentation of the code making it more readable.

```
if not self._query_overlaps_replacements(query, flags.latest_replacement_time):
      query.set_from_clause(replace(query.get_from_clause(), final=False))
      return

# Now add your logic
```

This also makes the code easier to review as the entire if block is not part of the changeset.

snuba/datasets/errors_replacer.py
```diff
@@ -206,93 +213,124 @@ def set_project_needs_final(
...
         p.zremrangebyscore(
             exclude_groups_key, float("-inf"), now - settings.REPLACER_KEY_TTL
         )
+        # TODO: put this into another for loop to make post processing easier
+        # get up to date excluded groups
         p.zrevrangebyscore(
             exclude_groups_key, float("inf"), now - settings.REPLACER_KEY_TTL
         )
 
+    # get the replacement types for metrics purposes
```
fpacifici, line 277: please remove

fpacifici (changes requested):
Please add tests for the query processor changes. 

Otherwise just minor comments

snuba/datasets/errors_replacer.py
vladkluev: why? it's communicating intent

snuba/datasets/errors_replacer.py
vladkluev: I think this comment is better here than in the docstring as it actually tells me what it's doing when it's doing it. I also think that comment is actually communicating intent. Consider this line:

```python3
p.zremrangebyscore(
    exclude_groups_key, float("-inf"), now - settings.REPLACER_KEY_TTL
)
``` 

In order to understand what it's doing you need to:
* Know what `remrangebyscore` does
* its arguments and their order
* how this sorted set is scored

I understand that somebody with a lot of context in the codebase can see this pattern clearly but processing the above without the comment is much harder than with.

snuba/datasets/errors_replacer.py
fpacifici: Then in such case there should be first a PR with the refactoring and then a PR with the logic change.

snuba/datasets/errors_replacer.py
rahul-kumar-saini: #2155

tests/datasets/storages/processors/test_replaced_groups.py
```diff
@@ -32,17 +33,68 @@ def build_in(project_column: str, projects: Sequence[int]) -> Expression:
...
 
 
+def test_query_set_final(query: ClickhouseQuery) -> None:
+    enforcer = PostReplacementConsistencyEnforcer("project_id", None)
+
+    enforcer._set_query_final(query, True)
+    assert query.get_from_clause().final
+
+    enforcer._set_query_final(query, False)
+    assert not query.get_from_clause().final
```
fpacifici, line 95: No need to test what is not a public method. 
Private/protected methods are considered internal implementation details. 
You generally rely on unit test to ensure no regression if you do a refactoring that keeps the interface unchanged. 
You run the test as it is and it must pass after the refactoring as well as before.

If your unit test tests individual private/protected methods, that cannot be done, because you will have to make changes to the test if you refactor the internals of the class. 
And when you need to make changes to the test to test for regressions you may not have introduced regressions in the test itself.

tests/datasets/storages/processors/test_replaced_groups.py
```diff
@@ -127,3 +173,53 @@ def test_too_many_groups_to_exclude(query: ClickhouseQuery) -> None:
...
+
+    # overlaps replacement and should be final due to too many groups to exclude
+    state.set_config("max_group_ids_exclude", 2)
+    set_project_exclude_groups(
+        2,
+        [100, 101, 102],
+        ReplacerState.EVENTS,
+        ReplacementType.EXCLUDE_GROUPS,  # Arbitrary replacement type, no impact on tests
+    )
+    enforcer._set_query_final(query_with_timestamp, True)
```
fpacifici, line 218: Please do not set the flag to True before the test. This way you are not testing the behavior of the `process_query` method anymore as you don't know if the final flag is on because it was set by the process_query method or by yourself on this line.

tests/datasets/storages/processors/test_replaced_groups.py
```diff
@@ -127,3 +173,53 @@ def test_too_many_groups_to_exclude(query: ClickhouseQuery) -> None:
...
+        query_with_timestamp, datetime(2021, 2, 1)
+    )
+    # latest replacement time < query from
+    assert not enforcer._query_overlaps_replacements(
+        query_with_timestamp, datetime(2020, 1, 1)
+    )
+    # replacement time unknown, default to overlaps
+    assert enforcer._query_overlaps_replacements(query_with_timestamp, None)
+    # query time range unknown, default to overlaps
+    assert enforcer._query_overlaps_replacements(query, datetime(2021, 2, 1))
```
fpacifici, line 197: You can skip this test as well for the same reason above. 
If you think the method is interesting enough to be tested on its own, take it out of the class as a stand alone function and let it be exposed. For as long as it is a protected method, it should be considered implementation detail of the class and its hierarchy.

fpacifici (approved):
Please address the comments on the test before merging together with those on the comments.

PR #2140, opened 2021-10-08
feat(subscriptions): Add scheduler mode to stream loader

lynnagara:
The Kafka stream loader defines the scheduler watermark mode that applies to
the subscriptions on a specific storage. The scheduler mode is
associated in code with the storage / topic and should not need
to be changed by users when running subscriptions. Generally speaking,
"partition" mode will apply to semantically partitioned topics and
"global" mode applies for non semantically partitioned topics
where we will wait for all partitions to reach a timestamp before
executing the subscriptions associated with that particular timestamp.
The mode must be specified if subscriptions are supported for that storage.

### DIFF
--- a/snuba/subscriptions/utils.py
+++ b/snuba/subscriptions/utils.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+
+class SchedulerMode(Enum):
+    IMMEDIATE = "immediate"
+    WAIT_FOR_SLOWEST_PARTITION = "wait-for-slowest"
--- a/snuba/subscriptions/utils.py
+++ b/snuba/subscriptions/utils.py
@@ -1,6 +1,6 @@
 from enum import Enum
 
 
-class SchedulerMode(Enum):
-    IMMEDIATE = "immediate"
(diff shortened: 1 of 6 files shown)


vladkluev:
I see there is code to reference immediate mode but the PR description seems to indicate that it will change the way the loader actually works. Is the code for that being written at a later point? Has it already been written?

snuba/subscriptions/utils.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class SchedulerMode(Enum):
+    IMMEDIATE = "immediate"
+    WAIT_FOR_SLOWEST_PARTITION = "wait-for-slowest"
```
fpacifici, line 6: Let's rename these to 
```
class WatermarkMode:    # Or SchedulingWatermarkMode
       PARTITION = "partition"
       GLOBAL = "global"
```

This concept has a specific name in streaming system, using its own proper name will make the code easier to understand.

fpacifici (approved):
One comment inline. No other issues

PR #2141, opened 2021-10-08
feat(subscriptions): Switch the scheduler back to use the tick consumer

lynnagara:
Now that we are properly filtering commit log messages by the consumer
group being followed, let's try the tick consumer again. We shouldn't
see the invalid intervals anymore.

The --followed-consumer-group option is now required.


fpacifici (approved):
Thanks.

By the way did you figure out what was causing the commit log messages to be non monotonic ?

PR #2144, opened 2021-10-11
bug(mismatched nulls) allow aggregations across mismatched null columns

vladkluev:
### Context

Some discover queries are failing because:

1. Discover is a merge table
2. On one side of the merge, a column is nullable, on the other it is not
3. Clickhouse gets confused running aggregation on such columns

```
Conversion from AggregateFunction(uniq, LowCardinality(String)) to AggregateFunction(uniq, LowCardinality(Nullable(String))) is not supported
```

### Solution

Clickhouse provides a `cast`operator such that the mismatched fields can be cast to null. This PR introduces a query processor which:

* Will pick up the mismatched null fields
* anytime they are being aggregated in the query, wrap them in a cast for their own datatype

### Test Plan

* Unit tests for the query processor
* The actual casting operation has been verified to fix the problem:

```sql
-- Should fail
SELECT (ifNull(uniq(sdk_version AS _snuba_sdk_version), 0) AS _snuba_count_unique_sdk_version) 
FROM discover_dist
PREWHERE in((project_id AS _snuba_project_id), tuple(5433960)) 
WHERE 
    equals(deleted, 0) AND 
    greaterOrEquals((timestamp AS _snuba_timestamp), toDateTime('2021-08-18T18:34:04', 'Universal')) AND 
    less(_snuba_timestamp, toDateTime('2021-09-01T18:34:04', 'Universal')) AND 
    equals(_snuba_project_id, 5433960) 
LIMIT 1 
OFFSET 0

-- Should work
SELECT ifNull(uniq(CAST(sdk_version, 'Nullable(String)') AS _snuba_sdk_version), 0) AS _snuba_count_unique_sdk_version
FROM discover_dist
PREWHERE (project_id AS _snuba_project_id) IN tuple(5433960)
WHERE (deleted = 0) AND ((timestamp AS _snuba_timestamp) >= toDateTime('2021-08-18T18:34:04', 'Universal')) AND (_snuba_timestamp < toDateTime('2021-09-01T18:34:04', 'Universal')) AND (_snuba_project_id = 5433960)
LIMIT 0, 1
```
```
First query
↘ Progress: 77.16 thousand rows, 933.80 KB (727.83 thousand rows/s., 8.81 MB/s.)  60%Received exception from server (version 20.3.9):
Code: 70. DB::Exception: Received from localhost:9000. DB::Exception: Conversion from AggregateFunction(uniq, LowCardinality(String)) to AggregateFunction(uniq, LowCardinality(Nullable(String))) is not supported: While executing SourceFromInputStream. 

0 rows in set. Elapsed: 0.207 sec. Processed 77.16 thousand rows, 933.80 KB (372.94 thousand rows/s., 4.51 MB/s.) 
```

```
Second Query:

SELECT ifNull(uniq(CAST(sdk_version, 'Nullable(String)') AS _snuba_sdk_version), 0) AS _snuba_count_unique_sdk_version
FROM discover_dist
PREWHERE (project_id AS _snuba_project_id) IN tuple(5433960)
WHERE (deleted = 0) AND ((timestamp AS _snuba_timestamp) >= toDateTime('2021-08-18T18:34:04', 'Universal')) AND (_snuba_timestamp < toDateTime('2021-09-01T18:34:04', 'Universal')) AND (_snuba_project_id = 5433960)
LIMIT 0, 1

┌─_snuba_count_unique_sdk_version─┐
│                               2 │
```


snuba/query/processors/null_column_caster.py
```diff
@@ -0,0 +1,125 @@
...
+                    return FunctionCall(
+                        exp.alias,
+                        "cast",
+                        (
+                            # move the alias up to the cast function
+                            Column(
+                                None,
+                                table_name=exp.table_name,
+                                column_name=exp.column_name,
+                            ),
```
vladkluev, line 107: I am not sure about this move here and want to make sure the reviewer sees this line. I do not know how our system uses aliases and what I should do with the existing alias (if anything)

evanh: What you're doing is correct. We want to ensure that the alias of the expression going in to your function is the alias of the expression coming out. Those aliases might be used elsewhere in the query. You don't really need to alias the inner `Column` so that's also fine.

PR #2145, opened 2021-10-12
perf(subscriptions): Add minimum tick interval option for subscriptions

lynnagara:
This is an experiment to see if we can improve the performance of the subscription
worker by creating fewer ticks which span larger intervals, which will result in
less calls to later parts of the subscriptions pipeline and messages submitted
to the processing strategy.

The subscriptions CLI now supports a --min-tick-interval-ms option, which can
be a value between 1 and 1000. If a value is passed, a tick will only be
created when we get messages spanning an interval larger than this range. This
results in fewer tick message but each covering a larger range.

We should only attempt to use this option (at least for now) on the experimental
sessions subscriptions consumer, and not events or transactions.

### DIFF
--- a/snuba/subscriptions/consumer.py
+++ b/snuba/subscriptions/consumer.py
@@ -92,11 +92,15 @@ class TickConsumer(Consumer[Tick]):
     # the consumer.
 
     def __init__(
-        self, consumer: Consumer[Any], time_shift: Optional[timedelta] = None
+        self,
+        consumer: Consumer[Any],
+        time_shift: Optional[timedelta] = None,
+        min_interval: Optional[timedelta] = None,
     ) -> None:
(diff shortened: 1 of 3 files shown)


snuba/subscriptions/consumer.py
```diff
@@ -92,11 +92,15 @@ class TickConsumer(Consumer[Tick]):
     # the consumer.
 
     def __init__(
-        self, consumer: Consumer[Any], time_shift: Optional[timedelta] = None
+        self,
+        consumer: Consumer[Any],
+        time_shift: Optional[timedelta] = None,
+        min_interval: Optional[timedelta] = None,
```
fpacifici, line 98: Mind adding a comment about what this is for?

PR #2148, opened 2021-10-15
feat(metrics): Add metrics consumer to devserver

jjbayer:
Add a metrics consumer to devserver if dev features are enabled.

Open question: Is the multistorage consumer suitable for this purpose?

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -137,6 +137,23 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
         ),
     ]
 
+    if settings.ENABLE_DEV_FEATURES:
+        daemons += [
+            (
+                "metrics-consumer",
+                [
+                    "snuba",
+                    "multistorage-consumer",
+                    "--storage=metrics_counters_buckets",
+                    "--storage=metrics_distributions_buckets",
+                    "--storage=metrics_buckets",
+                    "--auto-offset-reset=latest",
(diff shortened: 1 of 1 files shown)


snuba/cli/devserver.py
```diff
@@ -137,6 +137,23 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
         ),
     ]
 
+    if settings.ENABLE_DEV_FEATURES:
+        daemons += [
+            (
+                "metrics-consumer",
+                [
+                    "snuba",
+                    "multistorage-consumer",
```
jjbayer, line 146: Should I use a multi-storage consumer here, or rather multiple consumers with different consumer groups?

fpacifici: A multi storage consumer is the right option. The topic is the same, there is no need for multiple consumers in dev.

snuba/cli/devserver.py
```diff
@@ -137,6 +137,23 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
         ),
     ]
 
+    if settings.ENABLE_DEV_FEATURES:
```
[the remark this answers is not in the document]
fpacifici: You should also check that the dataset is enabled 
```
DISABLED_DATASETS
```

fpacifici (changes requested):
Please add the disabled dataset check, otherwise it is good, thanks

PR #2149, opened 2021-10-15
feat(subscriptions): Add consumer queue options to subscriptions

lynnagara:
This adds the ability to override the default queued.min.messages and
queued.max.messages.kbytes librdkafka settings on the subscriptions consumer,
allowing us to provide different values for each subscriptions consumer.

The immediate motivation for this change is to try and reduce the consumer group
lag on the sessions consumer by increasing the buffer size.

We already support these two options in the main snuba consumer and replacer
processes, this change adds the same settings to subscriptions.

### DIFF
--- a/snuba/cli/subscriptions.py
+++ b/snuba/cli/subscriptions.py
@@ -83,13 +83,11 @@ logger = logging.getLogger(__name__)
 )
 @click.option(
     "--queued-max-messages-kbytes",
-    default=settings.DEFAULT_QUEUED_MAX_MESSAGE_KBYTES,
     type=int,
     help="Maximum number of kilobytes per topic+partition in the local consumer queue.",
 )
 @click.option(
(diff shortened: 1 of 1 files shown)


snuba/cli/subscriptions.py
```diff
@@ -146,6 +158,8 @@ def subscriptions(
                     consumer_group,
                     auto_offset_reset=auto_offset_reset,
                     bootstrap_servers=bootstrap_servers,
+                    queued_max_messages_kbytes=queued_max_messages_kbytes,
+                    queued_min_messages=queued_min_messages,
```
fpacifici, line 162: I wonder what happens to librdkafka if we create two kafka consumers with the confluent library (this one and the commit log one) to the same broker but with different settings. 
Did you by chance try ?

fpacifici (approved):

PR #2150, opened 2021-10-18
feat(ratelimit) Add an organization rate limiter

evanh:
We have some entities that don't have project ID as a required key, and as such
are only being rate limited on the table and global rate limits. However these
tables do have an org ID so we can rate limit on the org ID.

The project rate limiter was abstracted into an object rate limiter that can
search for IDs in a query based on a column name, and the organization and
project rate limiters extend it.

Currently there is no per org limit set, there will have to be some research
to tune the settings correctly.

### DIFF
--- a/snuba/cli/api.py
+++ b/snuba/cli/api.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Union
 
 import os
 import click
@@ -7,12 +7,30 @@ from snuba.environment import setup_logging
 
 
 @click.command()
+@click.option("--bind", help="Address to listen on.")
 @click.option("--debug", is_flag=True)
 @click.option("--log-level", help="Logging level to use.")
 @click.option("--processes", default=1)
 @click.option("--threads", default=1)
-def api(*, debug: bool, log_level: Optional[str], processes: int, threads: int) -> None:
+def api(
+    *,
+    bind: Optional[str],
+    debug: bool,
+    log_level: Optional[str],
+    processes: int,
+    threads: int
+) -> None:
     from snuba import settings
+
+    port: Union[int, str]
+    if bind:
+        if ":" in bind:
+            host, port = bind.split(":", 1)
+            port = int(port)
+        else:
+            raise click.ClickException("bind can only be in the format <host>:<port>")
+    else:
+        host, port = settings.HOST, settings.PORT
 
     if debug:
         if processes > 1 or threads > 1:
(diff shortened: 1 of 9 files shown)


tests/query/processors/test_org_rate_limiter.py
```diff
@@ -0,0 +1,95 @@
...
+    settings = HTTPRequestSettings()
+
+    num_before = len(settings.get_rate_limit_params())
+    OrganizationRateLimiterProcessor("org_id").process_query(query, settings)
+    assert len(settings.get_rate_limit_params()) == num_before + 1
+    rate_limiter = settings.get_rate_limit_params()[-1]
+    assert rate_limiter.rate_limit_name == ORGANIZATION_RATE_LIMIT_NAME
+    assert rate_limiter.bucket == str(org_id)
+    assert rate_limiter.per_second_limit == 1000
+    assert rate_limiter.concurrent_limit == 1000
```
vladkluev, line 72: where does the number 1000 come from?

snuba/query/processors/object_id_rate_limiter.py
```diff
@@ -0,0 +1,96 @@
...
+                return
+
+        obj_ids = get_object_ids_in_query_ast(query, self.object_column)
+        if not obj_ids:
+            return
+
+        # TODO: Add logic for multiple IDs
+        obj_id = obj_ids.pop()
+
+        orl, ocl = get_configs(
```
vladkluev, line 45: nit: I would prefer to just have the full name of these variables. 
```suggestion
        object_per_second_limit, object_concurrent_limit = get_configs(
```

snuba/datasets/entities/sessions.py
```diff
@@ -285,6 +288,7 @@ def get_query_processors(self) -> Sequence[QueryProcessor]:
             TimeSeriesProcessor(
                 {"bucketed_started": "started"}, ("started", "received")
             ),
+            OrganizationRateLimiterProcessor(org_column="org_id"),
```
vladkluev, line 291: we also want this on outcomes right? This is the only reason I'm requesting changes.

vladkluev (changes requested):

snuba/datasets/entities/sessions.py
evanh: I did add this to outcomes. See `outcomes.py` and `outcomes_raw.py`.

vladkluev: oh weird, for some reason when I looked at this yesterday I didn't see it. my bad

vladkluev (approved):
some nits, but LGTM

PR #2151, opened 2021-10-18
feat(subscriptions): Use tick buffer in scheduler, properly commit offsets

lynnagara:
This change adds a new ProvideCommitStrategy processing step that is
responsible for determining whether the offset of a message can be
safely committed or not. It is designed to be chained together with the tick
buffer strategy - it get passed a `Tick` and submits that message to the next
step as a `CommittableTick` - basically that same tick with a True or False
`should_commit` value indicating whether it can be safely committed or not.

In order to guarantee that all subscriptions are executed at least once, we can
only safely commit offsets based on the slowest partition. This ensures that even
if the scheduler is restarted and loses it's state we do not skip any intervals
on any partitions.

The subscription scheduler has also been updated to use the new tick
buffer processing strategy together with the commit strategy. The mode
is temporarily hardcoded to 'global' just for testing and experimentation.
The correct mode value should be provided by the stream loader.
The buffer size can be configured via the 
`SUBSCRIPTIONS_DEFAULT_BUFFER_SIZE` and
`SUBSCRIPTIONS_ENTITY_BUFFER_SIZE` settings. The default is 10000.

We are monitoring this in production via the `partition_lag_ms` and consumer
group lag metrics.


vladkluev:
Can we get a better understanding of the test plan? What is going to be tested and monitored once this rolls out?

lynnagara:
@vladkluev The metrics to be monitored are:
- `partition_lag_ms` metric which has now moved into the tick buffer strategy
- Consumer group lag, which may increase a bit since we are committing offsets later now

tests/subscriptions/test_scheduler_processing_strategy.py
```diff
@@ -200,3 +204,124 @@ def test_tick_buffer_wait_slowest() -> None:
...
+            ),
+        ),
+        epoch,
+        3,
+    )
+
+    strategy.submit(message_p1)
+
+    assert next_step.submit.call_args_list == [
+        mock.call(make_message_for_next_step(message_p1, True))
```
fpacifici, line 269: I'd beef up this test as it test one simple use case.
Two messages for two partitions with timestamps that are perfectly aligned.
I think there are quite a few interesting cases to test.
- multiple messages for one partitions before you see the first message of the second
- time stamps not aligned. Like partition1: 1-5,6-9,10-11 and partition2: 1-2,3-5,6-12
- Multiple updates of the watermark causing more than one commit only. Like starting with the case you are already testing, then keep submitting till you see the second commit to ensure watermarks are updated properly.

tests/subscriptions/test_scheduler_processing_strategy.py
```diff
@@ -200,3 +204,124 @@ def test_tick_buffer_wait_slowest() -> None:
...
+        ),
+        now,
+        7,
+    )
+    strategy.submit(message_1_0)
+
+    assert next_step.submit.call_count == 2
+    assert next_step.submit.call_args_list == [
+        mock.call(make_message_for_next_step(message_0_0, False)),
+        mock.call(make_message_for_next_step(message_1_0, True)),
```
fpacifici, line 326: Same for this test. I would add a few less trivial cases.

fpacifici (approved):
Seems good.
Please beef up the unit tests before merging. 
This algorithm is tricky to get right. Let's have a good test coverage.
Also the tests will be example people can refer to.

PR #2153, opened 2021-10-20
feat(suspect-spans): Enforce limit on number of spans per txn

Zylphrex:
The SDKs already enforce a limit of 1000 spans per transaction. This adds an
additional check so we do not store an arbitrary number of spans for each
transaction. The limit is current set at 2000 spans per transaction which should
be more than enough.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,8 @@ RUN set -ex; \
     '; \
     apt-get update; \
     apt-get install -y $buildDeps --no-install-recommends; \
+    # Since there's no confluent-kafka wheel for aarch64, remove when there is
+    [ $(uname -m) = "aarch64" ] && apt-get install -y librdkafka-dev --no-install-recommends; \
     \
     pip install -r requirements.txt; \
     \
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,7 +43,7 @@ python-rapidjson==1.4
 redis==2.10.6
 redis-py-cluster==1.3.5
 sentry-arroyo==0.0.4
-sentry-relay==0.8.7
+sentry-relay==0.8.9
 sentry-sdk==1.1.0
 simplejson==3.15.0
 urllib3==1.26.5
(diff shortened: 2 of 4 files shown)


snuba/settings/__init__.py
```diff
@@ -149,6 +149,9 @@
 # Metric Alerts Subscription Options
 ENABLE_SESSIONS_SUBSCRIPTIONS = os.environ.get("ENABLE_SESSIONS_SUBSCRIPTIONS", False)
 
+# The maximum number of spans to store in the array columns for each transaction.
+MAX_SPANS_PER_TRANSACTION = 2000
```
vladkluev, line 153: why not just make this 1000?

Zylphrex: 2000 was chosen after some discussion with @evanh

snuba/datasets/transactions_processor.py
```diff
@@ -297,15 +297,37 @@ def _process_spans(
...
+                    spans_dropped = settings.MAX_SPANS_PER_TRANSACTION - num_spans
+
+                    logger.warning(
+                        "Too many spans found",
+                        extra={
+                            "trace_context": trace_context,
+                            "num_spans": num_spans,
+                            "spans_dropped": spans_dropped,
+                        },
+                    )
```
fpacifici, line 324: pelase do not use a the logger for an event that can happen often on a project. Each one of these message will be a new error to ingest for sentry.
We do not ever generate errors/warnings in the ingestion pipeline itself, unless we are sure the volume will be small. 

use a datadog metric instead.

snuba/datasets/transactions_processor.py
```diff
@@ -297,15 +297,37 @@ def _process_spans(
...
             processed_root_span = self._process_span(trace_context)
             if processed_root_span is not None:
+                num_processed += 1
                 processed_spans.append(processed_root_span)
 
             for span in data.get("spans", []):
+                # The number of spans should not exceed 1000 as enforced by SDKs.
+                # As a safety precaution, enforce a hard limit on the number of
+                # spans we actually store .
+                if num_processed >= settings.MAX_SPANS_PER_TRANSACTION:
```
fpacifici, line 312: Let's make this a runtime flag so we can change it quickly if/when needed. With a default of 2000. We will keep the default most of the time.

fpacifici (dismissed):

PR #2154, opened 2021-10-20
build(feat): Support aarch64 image building

armenzg:
In order to support building a Docker image that can run on `arm64/aarch64` we need to install or build Python binary wheels.

`sentry-relay` version `0.8.9` has Python wheels for `aarch64`, thus, we need to bump the version.

In order to build confluent-kafka from source, we need to add `librdkafka-dev` since there's no `aarch64` wheel for it.
In order for the code to find the library, we need to leave it on disk, otherwise, the image won't build.

> #13 2.834   File "/usr/local/lib/python3.8/site-packages/confluent_kafka/deserializing_consumer.py", line 19, in <module>
> #13 2.834     from confluent_kafka.cimpl import Consumer as _ConsumerImpl
> #13 2.834 ImportError: librdkafka.so.1: cannot open shared object file: No such file or directory

### DIFF
--- a/snuba/subscriptions/codecs.py
+++ b/snuba/subscriptions/codecs.py
@@ -69,40 +69,46 @@ class SubscriptionTaskResultEncoder(Encoder[KafkaPayload, SubscriptionTaskResult
         )
 
 
-class SubscriptionScheduledTaskEncoder(Codec[bytes, ScheduledTask[Subscription]]):
+class SubscriptionScheduledTaskEncoder(
+    Codec[KafkaPayload, ScheduledTask[Subscription]]
+):
     """
-    Encodes/decodes a scheduled subscription to bytes for Kafka.
+    Encodes/decodes a scheduled subscription to Kafka payload.
     Does not support non SnQL subscriptions.
     """
 
     def __init__(self, entity_key: EntityKey) -> None:
         self.__entity_key = entity_key
 
-    def encode(self, value: ScheduledTask[Subscription]) -> bytes:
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -44,6 +44,8 @@ RUN set -ex; \
     '; \
     apt-get update; \
     apt-get install -y $buildDeps --no-install-recommends; \
+    # Since there's no confluent-kafka wheel for aarch64, remove when there is
+    [ $(uname -m) = "aarch64" ] && apt-get install -y librdkafka-dev --no-install-recommends; \
```
armenzg, line 48: For the Intel build, this library is not necessary since there's an Intel wheel.
Hopefully one day `confluent-kafka` will have a wheel and this will not be required.
There's also a chance I might be able to produce it but for now let's use this.

requirements.txt
```diff
@@ -43,7 +43,7 @@ python-rapidjson==1.4
 redis==2.10.6
 redis-py-cluster==1.3.5
 sentry-arroyo==0.0.4
-sentry-relay==0.8.7
+sentry-relay==0.8.9
```
armenzg, line 46: The new release includes `aarch64` wheels, thus, removing the need to build from source and installing relay build tools & libraries.

armenzg (commented):
This does not yet the problem that we don't have automation to produce the image but it gets us ready for when we do.

armenzg:
The CI changes is related to the MAX_INSTANCES changes by @joshuarli . Unrelated to this change, I'm off tomorrow. I can merge when I come back (or feel free if you like).

PR #2156, opened 2021-10-20
ref(sentry-metrics): Enable metrics dataset

MeredithAnya:
In preparation for sentry metrics to be tested out in production, we need to remove it from the `DISABLED_DATASETS` and other dev only settings. 

**BEFORE MERGING**
* This should be merged after we have the necessary provisions (https://github.com/getsentry/ops/pull/3977 and https://github.com/getsentry/ops/pull/4020)
* In addition, we need to make sure the clickhouse cluster is in the prod config (as well as the single tenant one: https://github.com/getsentry/single-tenant/pull/426) before merging.

### DIFF
--- a/snuba/datasets/factory.py
+++ b/snuba/datasets/factory.py
@@ -9,13 +9,14 @@ from snuba.utils.serializable_exception import SerializableException
 DATASETS_IMPL: MutableMapping[str, Dataset] = {}
 DATASETS_NAME_LOOKUP: MutableMapping[Dataset, str] = {}
 
-DEV_DATASET_NAMES: Set[str] = {"metrics"}
+DEV_DATASET_NAMES: Set[str] = set()
 
 DATASET_NAMES: Set[str] = {
     "discover",
     "events",
     "groupassignee",
     "groupedmessage",
+    "metrics",
     "outcomes",
     "outcomes_raw",
     "sessions",
--- a/snuba/datasets/storages/factory.py
+++ b/snuba/datasets/storages/factory.py
@@ -46,7 +46,9 @@ CDC_STORAGES: Mapping[StorageKey, CdcStorage] = {
     **(DEV_CDC_STORAGES if settings.ENABLE_DEV_FEATURES else {}),
 }
 
-DEV_WRITABLE_STORAGES: Mapping[StorageKey, WritableTableStorage] = {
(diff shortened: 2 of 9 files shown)


snuba/datasets/factory.py
```diff
@@ -9,13 +9,14 @@
...
 
-DEV_DATASET_NAMES: Set[str] = {"metrics"}
+DEV_DATASET_NAMES: Set[str] = set()
 
 DATASET_NAMES: Set[str] = {
     "discover",
     "events",
     "groupassignee",
     "groupedmessage",
+    "metrics",
```
fpacifici, line 19: This will not work until we have cluster declared in the prod and single tenant config.
Snuba will refuse to run healthchecks because it will not find a cluster for the storages associated to the dataset.
So we need the cluster in prod and accessible first.

snuba/datasets/storages/factory.py
```diff
@@ -69,13 +72,16 @@
     **(DEV_WRITABLE_STORAGES if settings.ENABLE_DEV_FEATURES else {}),
 }
 
-DEV_NON_WRITABLE_STORAGES: Mapping[StorageKey, ReadableTableStorage] = {
+DEV_NON_WRITABLE_STORAGES: Mapping[StorageKey, ReadableTableStorage] = {}
+
+METRICS_NON_WRITABLE_STORAGES: Mapping[StorageKey, ReadableTableStorage] = {
```
fpacifici, line 77: I don't think you need to declare metrics storages in a separate variable anymore. You can just add them to the list at line 88.

snuba/settings/__init__.py
```diff
@@ -153,6 +153,9 @@
 # Map the Zookeeper path for the replicated merge tree to something else
 CLICKHOUSE_ZOOKEEPER_OVERRIDE: Mapping[str, str] = {}
 
+# Enable Sentry Metrics (used for the snuba metrics consumer)
+ENABLE_SENTRY_METRICS = os.environ.get("ENABLE_SENTRY_METRICS", False)
```
fpacifici, line 157: Maybe `ENABLE_SENTRY_METRICS_DEV` as that is meant only to affect dev

fpacifici (approved):
Looks good. 
Please wait for the dependencies to be deployed first

PR #2162, opened 2021-10-25
ref(errors_replacer): Refactor get_projects_query_flags() into a Class

rahul-kumar-saini:
## Overview
- `errors_replacer.get_projects_query_flags()` getting bloated, requires multiple helper functions to be readable
- Refactored the function and helpers into a class called `ProjectsQueryFlags`  as suggested [here](https://github.com/getsentry/snuba/pull/2138#discussion_r730154582)

## Changes
- Removed `errors_replacer.get_projects_query_flags()`
- Added `errors_replacer.ProjectsQueryFlags` with appropriate documentation in function docstrings instead of inline comments
- Updated tests


rahul-kumar-saini:
The `ProjectQueryFlags` data class is just a bag of data, these methods should all be in a single class that handles everything abstractly. Although it doesn't do a _great_ job at abstracting Redis, it's still cleaner than the current state of the code. This way at least we fully encapsulate the Redis usage into a single class. Also, creating new helper functions going forward can also be encapsulated within the class as opposed to cluttering the rest of the file.

evanh (approved):
OK this is an improvement. Having a class to hold the results is nicer for sure.

PR #2173, opened 2021-10-27
ref(legacy) Remove legacy query endpoint

evanh:
The legacy query endpoint is no longer being used. Remove it, and make the SnQL
parser the default. This also removes the concept of the query Language, since
now there is only one API language. This also slightly restructures the
build_request by pushing custom processing up a level, which cleans up the code
a bit.

This also should speed up our API tests since we no longer have to run tests as both
a legacy query and a snql query.


fpacifici (commented):
How long ago was the dependency removed from Sentry? 
If it is within this release I don't think we can do this just yet. If we patch the snuba release generating a new one, while sentry did not yet cut a release, we may on premise customers running a sentry version that relies on the old endpoint, but a snuba version that does not provide it.

evanh:
The dependency was removed from Sentry multiple releases ago. Also snuba and sentry releases run in lockstep, so if someone is running an older sentry with a newer snuba that is something we have stated we don't support.

fpacifici:
>  Also snuba and sentry releases run in lockstep

They do, but we may always cut a patch release for an individual component. This is unlikely to happen far from the monthly release time, but still possible.

evanh:
> They do, but we may always cut a patch release for an individual component.

That's a good point. I think we should be OK either way since the legacy API on the query side has been deprecated for at least 2 releases at this point.

fpacifici (approved):
Can we throw away the entire legacy parser or do we still need it for the subscriptions ?

evanh:
> Can we throw away the entire legacy parser or do we still need it for subscriptions ?

We can throw it all away, subscriptions are now using SnQL as well. Those deletions will happen in separate PRs.

PR #2178, opened 2021-10-30
feat(subscriptions): Add processing strategy to schedule subscriptions

lynnagara:
Adds a processing strategy that takes a `CommittableTick` - that is - a tick
message with a boolean value indicating whether its offset can be safely
committed or not, and does two things:
- schedule all of the subscriptions that need to be run given the tick's interval and partition
- commit the offset if the `CommittableTick` specifies should_commit = True

This strategy is supposed to come directly after the TickBuffer and ProvideCommitStrategy
strategies and actually schedule the subscriptions.

This code is not used yet and will be integrated into the scheduler as a follow up
step after https://github.com/getsentry/snuba/pull/2151.


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -151,3 +165,72 @@ def terminate(self) -> None:
...
+        tick = message.payload.tick
+
+        tasks = self.__schedulers[tick.partition].find(tick.timestamps)
+        futures = [
+            self.__producer.produce(self.__scheduled_topic, self.__encoder.encode(task))
+            for task in tasks
+        ]
+
+        for future in as_completed(futures):
+            future.result()
```
fpacifici, line 221: I don't think you need to wait for this at this point.
I would at least wait for the callbacks only when you know you are about to commit. So you can keep processing and producing till a checkpoint where you need to be sure all messages have been sent before going on.

You can even go on eagerly processing further messages till you are done with the callbacks for the batch before committing:
- produce all the messages for the ticks until you get a `should_commit` = `true`
- At that point, if all futures have completed (likely not), keep processing and keep the commit offset aside
- While you process new messages keep checking if all callbacks for the first batch have completed. 
- When they are commit and continue.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -151,3 +165,72 @@ def terminate(self) -> None:
...
+    This strategy is responsible for scheduling all of the subscriptions
+    for a given tick.
+
+    The subscriptions to be scheduled are those that:
+    - Have the same partition index as the tick received, and
+    - Are scheduled to be run within the time interval defined by the tick
+
+    For backward compatibility, this assumes that the number of partitions of
+    the subscription storage matches the number of partitions of the main topic
+    being subscribed to.
```
fpacifici, line 181: What do we gain by keeping the storage partitioned ini the same way as the topic partitioning?
It seem coupling the two is a limitation as there is no safe way to increase the number of topic partition without any sort of outage.

Can we keep track here of the last timestamp we scheduled, and as soon as we see a tick with an upper timestamp that is higher than the watermark, schedule all the subscriptions across partitions ?

fpacifici (changes requested):
See my comments in line

snuba/subscriptions/scheduler_processing_strategy.py
fpacifici: Adding some context. This does not require any major redesign:
See the StreamProcessor class from arroyo and the ProcessingStrategy docstring.
What the StreamProcessor does is:
- poll from the consumer
- poll on the processing strategy 
- submit to the processing strategy
- repeat

The rationale behind that is:
- all these methods are asynchronous (they return nothing), which means potentially none of them needs to wait on anything. This does not need to have any batching step where we stop for a specified amount of time every specified amount of time. We could be able to continue processing elements one by one. Each step can tell if it is falling behind by raising `MessageRejected`
- When you call submit (the processor calls submit), that is supposed to be an immediate operation, it schedules work to do, it is not required to do anything, it should not make the caller wait. It can also exert back-pressure by raising MessageRejected if there are too many items (messages) to be processed.
- At every step poll is called to actually trigger work. Poll should not wait either but it may be impractical to avoid.

This actually can help you in avoiding the wait and optimizing the time you spend executing.
In theory in order to follow the abstraction you could do it this way:
- submit stores the tick in a list and does nothing else. Before that it checks if you have pending messages you are waiting the callback from. If that list is too long raise MessageRejected to exert backpressure.
- poll will do what this method actually does except for waiting. Though it checks which futures are completed (which is not an operation you wait onto). It does not wait. As soon as it finds a committable tick to be completed it calls commit. 
- The code here never waits on anything. It is the StreamProcessor that interprets `MessageRejected` as "please wait".

I don't think we ever used this system that way (which is how it is supposed to be used), but this seems a valid use case. In practice it may cause a tight loop as we never wait on the future and keep iterating trying to submitting, that would increase CPU usage, though I think it is worth trying, we can always add a waiting operation on the futures in the poll method to avoid that.

lynnagara: Thanks, this was very helpful advice. I rewrote like you suggested to make both poll() and submit() non blocking and raise the MessageRejected to slow down the consumer.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: I removed this comment - it wasn't actually accurate, and there is no need for the partitioning strategies to be the same.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -151,3 +165,72 @@ def terminate(self) -> None:
...
+
+    def poll(self) -> None:
+        pass
+
+    def submit(self, message: Message[CommittableTick]) -> None:
+        assert not self.__closed
+
+        tick = message.payload.tick
+
+        tasks = self.__schedulers[tick.partition].find(tick.timestamps)
```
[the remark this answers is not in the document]
lynnagara: Yes it does, updated docstring to be explicit about that

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -151,3 +165,72 @@ def terminate(self) -> None:
...
+    ) -> None:
+        self.__schedulers = schedulers
+        self.__encoder = SubscriptionScheduledTaskEncoder(entity_key)
+        self.__producer = producer
+        self.__scheduled_topic = Topic(scheduled_topic_spec.topic_name)
+        self.__commit = commit
+        self.__closed = False
+
+    def poll(self) -> None:
+        pass
```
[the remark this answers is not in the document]
lynnagara: It was originally because submit() was waiting for all the futures to finish before. Since we're no longer waiting there, now updated poll() to evict completed futures from the queue and commit offsets.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -257,3 +278,167 @@ def terminate(self) -> None:
...
+        self.__producer = producer
+        self.__scheduled_topic = Topic(scheduled_topic_spec.topic_name)
+        self.__commit = commit
+        self.__closed = False
+
+        # Stores each tick with it's futures
+        self.__queue = ScheduledSubscriptionQueue()
+
+        # Not a hard max
+        self.__max_buffer_size = 10000
```
fpacifici, line 375: Setting ?

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -257,3 +278,167 @@ def terminate(self) -> None:
...
+                    }
+                )
+
+    def submit(self, message: Message[CommittableTick]) -> None:
+        assert not self.__closed
+
+        # If queue is full, raise MessageRejected to tell the stream
+        # processor to pause consuming
+        if len(self.__queue) >= self.__max_buffer_size:
+            raise MessageRejected
```
fpacifici, line 410: nice. 
Please let's ensure we add some metrics about this in arroyo as we are barely using this exception today and we are not monitoring how long we spend sending back pressure

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -257,3 +278,167 @@ def terminate(self) -> None:
...
+        start = time.time()
+
+        while self.__queue:
+            remaining = timeout - (time.time() - start) if timeout is not None else None
+
+            if remaining is not None and remaining <= 0:
+                logger.warning(f"Timed out with {len(self.__queue)} futures in queue")
+                break
+
+            self.__queue.popleft().subscription_future.result(remaining)
```
fpacifici, line 444: You are not committing here. I think you should still be committing all  ticks that have `should_commit` set to true

tests/subscriptions/test_scheduler_processing_strategy.py
```diff
@@ -418,3 +438,130 @@ def test_tick_buffer_with_commit_strategy() -> None:
...
+    )
+
+    future: Future[Message[KafkaPayload]] = Future()
+
+    queue.append(tick_message, deque([future]))
+
+    assert len(queue) == 1
+    assert queue.peek() == TickSubscription(tick_message, future, should_commit=True)
+    assert queue.popleft() == TickSubscription(tick_message, future, should_commit=True)
+    assert len(queue) == 0
```
fpacifici, line 475: Please test the `popleft` method with more than one tick and with more than one future per tick. That is the interesting part of this class.

tests/subscriptions/test_scheduler_processing_strategy.py
```diff
@@ -418,3 +438,130 @@ def test_tick_buffer_with_commit_strategy() -> None:
...
+    assert codec.decode(second_message.payload).timestamp == epoch + timedelta(
+        minutes=1
+    )
+
+    # No 3rd message
+    assert broker_storage.consume(partition, 2) is None
+
+    # Close the strategy
+    strategy.close()
+    strategy.join()
```
fpacifici, line 567: SO this is a good integration test, but it is missing a few cases to test:
- verify commit happens
- test two subscriptions to ensure the right one is produced.

I think a better way to do this is test multiple use cases with the in memory storage (multiple subscriptions, multiple ticks, multiple partitions, etc.)
Then run one single test on the real Kafka so you can see if the future work properly.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -257,3 +278,167 @@ def terminate(self) -> None:
...
+    def poll(self) -> None:
+        # Remove completed tasks from the queue and raise if an exception occurred.
+        # This method does not attempt to recover from any exception.
+        # Also commits any offsets required.
+        while True:
+            tick_subscription = self.__queue.peek()
+
+            if tick_subscription is None:
+                break
+
```
fpacifici, line 386:
```
   while self.__queue:
```
So you avoid the risk of a while true and you will have an explicit exit condition that will always be met in some way.

fpacifici (approved):
I think you are missing a commit in the `join` method. Other than that please beef up the tests as this logic will be very critical and hard to troubleshoot.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -257,3 +278,167 @@ def terminate(self) -> None:
...
+        # the queue
+        tick = message.payload.tick
+        tasks = self.__schedulers[tick.partition].find(tick.timestamps)
+        self.__queue.append(
+            message,
+            deque(
+                [
+                    self.__producer.produce(
+                        self.__scheduled_topic, self.__encoder.encode(task)
+                    )
```
fpacifici, line 422: I think we should partition the subscription by subscription_id so that the same subscription id is always going to the same executor and they are executed sequentially (so the sentry consumer does not receive data points out of order).

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: I'm not convinced we actually need to make this configurable- I think we can probably get away with hardcoding a reasonable value. I can't think of a reason why this would need to differ for different datasets right now. I think we can move it to settings later if there is a strong justification for being able to override it.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Thanks, added committing here to join as well

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: It is partitioned by subscription_id. It's done by the subscription task codec here https://github.com/getsentry/snuba/blob/badf1a0f2126af1da29712cf27abb5aabe8d521c/snuba/subscriptions/codecs.py#L73

PR #2179, opened 2021-10-31
perf(consumer) Add a mock version of the consumers for load test

fpacifici:
This introduces a mock batch writer and a mock replacer writer to allow us to run load tests of our consumers on production data in an idempotent way without writing anything on Clickhouse.

The idea is to create a dedicated consumer cluster. 
Let the consumer group backlog for some time 
Then start the consumer and measure how long it takes to the consumer to burn the backlog. 
This can be done to evaluate the current headroom (if we do not burn the backlog we have a problem) as well as to compare parameters (processes, batch size, replicas count)

The mock batch writer adds a configurable delay to simulate the clickhouse write delay (which we already measure). We do not do the same for replacement as they are so rare compared to batch writes that the replacement production delay would not impact the test.


snuba/consumers/consumer_builder.py
```diff
@@ -202,6 +208,10 @@ def __build_streaming_strategy_factory(
                     self.producer if self.replacements_topic is not None else None
                 ),
                 replacements_topic=self.replacements_topic,
+            )
+            if not self.__mock
+            else build_mock_batch_writer(
+                self.storage, bool(self.replacements_topic), self.metrics
```
vladkluev, line 214: This isn't actionable for this PR but I do want to raise something here that I've seen throughout the codebase. 

We have a lot of `Factory` classes and `build_something` functions and that makes it really hard to follow the code unless you're running it in the debugger. We have a big pattern of injecting things that are functions (or look a lot like functions) into other things that wrap them and return other functions. I understand the desire for levels of indirection (as it makes things theoretically extensible) but the ability to understand the code as written goes down.  

Again, not actionable but since I saw this and went down a rabbit hole to understand what these things did it made me remember this again. I don't know what the answer is yet except maybe a more unifying abstraction but this is something that's often in the back of my head.

vladkluev (approved):

snuba/consumers/consumer_builder.py
fpacifici: I think there are multiple concerns that you are raising:
- factories and builders. (objects instantiated whoes only goal is to instantiate another object)
- using functions as first class concepts and passing them around. 
The concerns produced by each and the reason for using vs not using them are very different IMO, and they are not just about adding indirection. 

I don't think there is anything wrong with the concept of either, especially passing around functions is a well established functional programming pattern and can produce very readable and elegant code. The issue can rise in the individual use case, if too much indirection is added. 
So I would not claim that these practices are either wrong or right in absolute in our code base.

In the specific use case. Building a `ProcessingStrategy` for the consumer to use it, the need of a factory is mandated by the fact that the `Consumer` needs to be able to create, destroy and recreate a `ProcessingStrategy` (which is a sequence of processing steps to simplify) when it needs. Examples are during partition assignment and revoking. 
We separated the `Consumer` concern (which infrastructure and specific to how Kafka works) from the processing concern (the `ProcessingStrategy`, which is application specific). I would argue the idea is good as we touch the consumer much less than we touch the processing logic. Though it has the downside that the consumer needs to be able to create and destroy the processing logic when it wants, thus the need of a factory there. 
Can this be organized in a way that makes the processing strategy code more readable. Probably. Never tried. If you have an idea, please feel free to file a backlog task so we can move ahead on that.

snuba/consumers/consumer.py
```diff
@@ -179,13 +179,37 @@ def join(self, timeout: Optional[float] = None) -> None:
         )
 
 
+class MockReplacementBatchWriter(ProcessingStep[ReplacementBatch]):
+    """
+    Fake replacement writer used to run consumer load tests.
+    Contrarily to the mock writer. THis one does not introduce a delay.
+    This is because replacement are rare enough that adding the
```
evanh, line 186: I'm not sure I understand why there's no delay here. Replacements take non-zero time to process, and isn't the intent here to test what happens when we have a lot of replacements?

evanh (approved):
One question, otherwise LGTM.

snuba/consumers/consumer.py
fpacifici: the main consumer only produces a replacement message on the replacement topic. 
The consumer that processes the replacement is another one and it does not use this code path.
This code is the one that takes the replacement message on the events topic, and sends it on the replacement topic.

snuba/settings/__init__.py
```diff
@@ -149,6 +149,12 @@
 # Metric Alerts Subscription Options
 ENABLE_SESSIONS_SUBSCRIPTIONS = os.environ.get("ENABLE_SESSIONS_SUBSCRIPTIONS", False)
 
+# The mock write latency to be applied when "writing" with the mock writer.
+# This is expressed in milliseconds
+MOCK_WRITER_LATENCIES = {"default": 100}
+# Deviation to apply to the above. In milliseconds
+MOCK_LATENCY_STD_DEVIATTION = 50
+
```
lynnagara, line 157: I'd question the usefulness of putting this in settings. I think settings ideally is fairly permanent and is mostly useful so we can run Snuba differently across different environments. In this case, it seems more these settings support a test we are only planning to run in prod and nowhere else. I think it'd be better to just pass the values on the CLI and not pollute settings (and possibly confuse operators of Snuba)

fpacifici: Isn't it worse if we add these permanent parameters to the command line interface that are only usable if another option is set to true and during a test?

lynnagara: Btw, when I wrote this comment I sort of assumed there was a separate API for the mock test, not just an option on the regular consumer, hence not concerned about adding additional options. Any thoughts on also doing that?

lynnagara: I think we we typing at the same time. Yeah I assumed it was a separate command.. any downside to making it so?

fpacifici: I did think about that. IT seemed it would introduce a massive duplication. But I think it may have been only relevant for the multi storage consumer. Let me try again.

lynnagara: What is the use case for ever needing to load test the multistorage consumer?

fpacifici: If a consumer in production runs as a multistorage one (like the metrics one) we need to load test it in the same way

snuba/consumers/consumer.py
evanh: That doesn't really answer my question. Sending the replacement on takes a non-zero amount of time. Why would we test with no delay?

You also  say "replacement are rare enough..." but isn't the point of this code to test what happens when replacements are not rare?

fpacifici: >  isn't the point of this code to test what happens when replacements are not rare?

Not really, this is not about testing replacements. This tests the main consumer on production load (where replacement messages are very rare). The main consumer needs a replacement writer where to produce replacement messages. If we do not provide a mock, the consumer we are testing will produce replacement on the real replacements topic so the test will not be idempotent. 

> Why would we test with no delay?

Because replacements messages are exceedingly rare for the consumer, we are going to test in production like conditions. It is really unlikely that ignoring that delay could impact the overall results.

fpacifici: > we are going to test in production like conditions.

We are going to test the main errors consumer under production conditions. 
Compared with the amount of messages it receives containing events to write, the replacements messages are a tiny fraction. 
Each one of the replacement messages is just produced on a kafka topic basically with no processing.
Producing on Kafka is largely an asynchronous operation. When we close the batch we just ensure that all the callbacks have been executed. 
Since the replacements are scattered along the batch and they are a tiny fraction, by the time we close a batch all callbacks have generally been executed already. The callbacks do almost nothing. So The impact of this process on the consumer workload is very likely negligible.

snuba/settings/__init__.py
fpacifici: I removed the settings, and also the support for the multistorage consumer. 
Will take care of that in another PR because of the amount of refactoring needed. We are only using it for Change data capture as of now

lynnagara (approved):
Thanks!

PR #2017, opened 2021-07-20
build(ci): Push a CI-only snuba image to make use of docker caching

billyvg:
Docker builds do not seem to be using cache.  Instead, we push (and pull) a CI-only image to GitHub's Container Registry that will be used to make use of docker layer caching.

Changed the test jobs to depend on a new job that builds the snuba CI image. We tag this image with `latest`, `<commit sha>`, and `<branch>`. This should mean that the `Tests and code coverage` jobs and Sentry integration test will have all layers cached when they build the snuba image.

PR #2180, opened 2021-11-01
feat(ci): Fix Docker caching & arm64 image name

armenzg:
Fix Docker caching by adding the --build-arg `BUILDKIT_INLINE_CACHE=1`. See [docs](https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources)

Running `sentry devservices up snuba` on an Apple arm64 host, pulls the `amd64` image, thus, let's publish it under a different name

We should only publish to latest when pushing from master, thus, keeping PRs running side by side from interfering with each other.


.github/workflows/ci.yml
```diff
@@ -65,32 +67,28 @@ jobs:
...
             --target testing
 
       - name: Publish images for cache
         if: github.repository_owner == 'getsentry'
         run: |
-          docker push ghcr.io/getsentry/snuba-ci:${{ steps.branch.outputs.branch }}
-          docker push ghcr.io/getsentry/snuba-ci:${{ github.sha }}
-          docker push ghcr.io/getsentry/snuba-ci:latest
+          # Useful to speed up CI on re-runs
+          docker push $SNUBA_IMAGE:${{ steps.branch.outputs.branch }}
```
armenzg, line 85: For PRs, we only need to push to the same branch.

.github/workflows/ci.yml
```diff
@@ -65,32 +67,28 @@ jobs:
...
+          if [ "${branch}" == 'refs/head/master']; then
+            # Useful for regression testing on Sentry's checkout
+            docker push $SNUBA_IMAGE:${{ github.sha }}
+            # The latest tag should only be publish on `master`
```
armenzg, line 89:
```suggestion
            # The latest tag should only be published on `master`
```

.github/workflows/ci.yml
```diff
@@ -65,32 +67,28 @@ jobs:
...
+            docker push $SNUBA_IMAGE:latest
```
armenzg, line 90: We only need to push to `latest` on `master` for *all* CI runs to have `latest` as the base image to cache from.

For instance, this is useful for the 1st CI run on a PR.

.github/workflows/ci.yml
```diff
@@ -65,32 +67,28 @@ jobs:
...
         if: github.repository_owner == 'getsentry'
         run: |
-          docker push ghcr.io/getsentry/snuba-ci:${{ steps.branch.outputs.branch }}
-          docker push ghcr.io/getsentry/snuba-ci:${{ github.sha }}
-          docker push ghcr.io/getsentry/snuba-ci:latest
+          # Useful to speed up CI on re-runs
+          docker push $SNUBA_IMAGE:${{ steps.branch.outputs.branch }}
+          if [ "${branch}" == 'refs/head/master']; then
+            # Useful for regression testing on Sentry's checkout
+            docker push $SNUBA_IMAGE:${{ github.sha }}
```
armenzg, line 88:
```suggestion
```

I don't think we need to push the sha tag since we don't use this CI images as part of the dev services.

.github/workflows/ci.yml
```diff
@@ -103,11 +101,21 @@ jobs:
     name: "Docker Arm64 build"
     runs-on: macos-11
     env:
-      SNUBA_IMAGE: ghcr.io/getsentry/snuba-ci
+      # `sentry devservices up snuba` picks up the `amd64` version of the image
+      # For now, we will publish to a different name until we want to officially
+      # support the image on Docker Hub
+      # NOTE: This image has only been tested for local development
+      SNUBA_IMAGE: ghcr.io/getsentry/snuba-arm64-dev
```
armenzg, line 108: This overwrites the top-level environment variable.

The day we want to officially support `arm64` images (publish on Docker Hub) and we want to use those for local dev, we can investigate why docker-py does not use the host's platform as the tag.

armenzg: In this PR, this is the actual fix for using the `arm64` image on Apple M1s.

.github/workflows/ci.yml
```diff
@@ -123,18 +131,24 @@ jobs:
...
+            -t $SNUBA_IMAGE:latest \
+            -t $SNUBA_IMAGE:${{ steps.branch.outputs.branch }} \
+            -t $SNUBA_IMAGE:${{ github.sha }} \
+            --load .
 
       - name: Publish
         # Forks cannot push to the getsentry org
         if: github.repository_owner == 'getsentry'
-        env:
-          branch: ${{ github.ref }}
```
armenzg, line 134: Not used anymore.

.github/workflows/ci.yml
```diff
@@ -123,18 +131,24 @@ jobs:
...
         run: |
-          docker push $SNUBA_IMAGE:${{ github.sha }}
-          if [ "${branch}" == 'refs/head/master']; then
+          # Useful to speed up CI on re-runs
```
armenzg, line 146:
```suggestion
          # Useful to speed up CI
```

.github/workflows/ci.yml
```diff
@@ -123,18 +131,24 @@ jobs:
...
+          docker push $SNUBA_IMAGE::${{ steps.branch.outputs.branch }}
+          if [ "${{ steps.branch.outputs.branch }}" == 'master']; then
+            # Useful for regression testing on Sentry's checkout
+            docker push $SNUBA_IMAGE:${{ github.sha }}
+            # The latest tag should only be publish on `master`
```
armenzg, line 151:
```suggestion
            # The latest tag should only be published on `master`
```

.github/workflows/ci.yml
```diff
@@ -150,19 +164,12 @@ jobs:
       - uses: actions/checkout@v2
         name: Checkout code
 
-      - name: Pull snuba CI images
-        run: |
-          docker pull ghcr.io/getsentry/snuba-ci:${{ github.sha }} || \
-            docker pull ghcr.io/getsentry/snuba-ci:${{ needs.snuba-image.outputs.branch }} || \
-            docker pull ghcr.io/getsentry/snuba-ci:latest || true
```
armenzg, line 157: Unnecessary since `--cache-from` will pull them.

.github/workflows/ci.yml
```diff
@@ -150,19 +164,12 @@ jobs:
...
-
-      - name: Build snuba docker image for CI
+      - name: Build testing image
```
armenzg, line 167: Making it more clear that this is different than the Snuba image.

.github/workflows/ci.yml
```diff
@@ -150,19 +164,12 @@ jobs:
...
         run: |
           docker build . \
             -t snuba-test \
-            --cache-from ghcr.io/getsentry/snuba-ci:${{ github.sha }} \
```
armenzg, line 163: Removing. We don't need to use a sha tag since the branch tag does the job.

.github/workflows/ci.yml
```diff
@@ -197,19 +204,12 @@ jobs:
       - name: Checkout snuba
         uses: actions/checkout@v2
 
-      - name: Pull snuba CI images
-        run: |
-          docker pull ghcr.io/getsentry/snuba-ci:${{ github.sha }} || \
-            docker pull ghcr.io/getsentry/snuba-ci:${{ needs.snuba-image.outputs.branch }} || \
-            docker pull ghcr.io/getsentry/snuba-ci:latest || true
```
armenzg, line 204: Not needed.

evanh (approved):

.github/workflows/ci.yml
```diff
@@ -65,32 +67,28 @@ jobs:
         run: |
           echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u $GITHUB_ACTOR --password-stdin
 
-      # These are pulled in order to be able to use docker layer caching
-      - name: Pull snuba CI images
-        if: github.repository_owner == 'getsentry'
-        run: |
-          docker pull ghcr.io/getsentry/snuba-ci:${{ github.sha }} || \
-            docker pull ghcr.io/getsentry/snuba-ci:${{ steps.branch.outputs.branch }} || \
-            docker pull ghcr.io/getsentry/snuba-ci:latest || true
```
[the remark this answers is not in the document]
billyvg: I don't think that `--cache-from` actually does take care of this unfortunately e.g. these two runs are a run + a re-run, the latter should be cached, but it doesn't look like it is:

https://github.com/getsentry/snuba/runs/4071609247?check_suite_focus=true#step:5:1

(re-run) https://github.com/getsentry/snuba/runs/4072434335?check_suite_focus=true#step:5:745

.github/workflows/ci.yml
```diff
@@ -65,32 +67,28 @@ jobs:
...
-            -t ghcr.io/getsentry/snuba-ci:${{ github.sha }} \
-            -t ghcr.io/getsentry/snuba-ci:${{ steps.branch.outputs.branch }} \
-            --cache-from ghcr.io/getsentry/snuba-ci:latest \
-            --cache-from ghcr.io/getsentry/snuba-ci:${{ steps.branch.outputs.branch }} \
-            --cache-from ghcr.io/getsentry/snuba-ci:${{ github.sha }} \
+            --cache-from $SNUBA_IMAGE:latest \
+            --cache-from $SNUBA_IMAGE:${{ steps.branch.outputs.branch }} \
+            -t $SNUBA_IMAGE:latest \
+            -t $SNUBA_IMAGE:${{ steps.branch.outputs.branch }} \
+            -t $SNUBA_IMAGE:${{ github.sha }} \
```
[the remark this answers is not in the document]
billyvg: `github.sha` for a PR will be the commit sha of the merge commit, however we generally squash our commits to master so on master, it will be a different sha.

.github/workflows/ci.yml
billyvg: Should we push `master` tag?

.github/workflows/ci.yml
billyvg: Do we want to also tag the commit sha?

.github/workflows/ci.yml
```diff
@@ -103,11 +99,21 @@ jobs:
     name: "Docker Arm64 build"
```
billyvg, line 99: Do we want this in the `ci` workflow where the other jobs are for testing? And here we are producing images to be used in local development for sentry. This also takes 45+ minutes, which is way too long imo for CI. It might make more sense to have this in its own workflow that runs only on pushes to master.

.github/workflows/ci.yml
billyvg: isn't this a local dev image?

.github/workflows/ci.yml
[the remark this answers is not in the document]
armenzg: Thanks for the catch 😥 
I'm surprised `--cached-from` would *not* do pulling of the images first.
Adding this back helps.

Looking at the jobs on `master`, it seems we don't have the caching working properly
https://github.com/getsentry/snuba/actions/workflows/ci.yml?query=branch%3Amaster
image

I thought we were using [BUILDKIT_INLINE_CACHE](https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources) but we were not.

I've added it back and we're back to a good state:
https://github.com/getsentry/snuba/runs/4080200190?check_suite_focus=true
image

.github/workflows/arm64-dev-env.yml
```diff
@@ -0,0 +1,71 @@
+name: arm64
```
armenzg: Moving the `arm64` build into its own workflow.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,7 @@ jobs:
         if: github.repository_owner == 'getsentry'
         run: |
           docker build . \
+            --build-arg BUILDKIT_INLINE_CACHE=1 \
```
armenzg, line 80: This argument is the one that fixes the Docker caching not working from properly. See [docs](https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources).

We removed `BUILDKIT_INLINE_CACHE=1` back in [here](https://github.com/getsentry/snuba/pull/2017/files#r674222449). If an image was not built with that argument then we don't have metadata that would allow it to be used with `--cache-from`, thus, when I fixed #2175 we started having slow Docker builds.

.github/workflows/ci.yml
```diff
@@ -88,9 +89,13 @@ jobs:
       - name: Publish images for cache
         if: github.repository_owner == 'getsentry'
         run: |
+          # Useful to speed up CI
           docker push ghcr.io/getsentry/snuba-ci:${{ steps.branch.outputs.branch }}
           docker push ghcr.io/getsentry/snuba-ci:${{ github.sha }}
-          docker push ghcr.io/getsentry/snuba-ci:latest
+          if [ "${{ steps.branch.outputs.branch }}" == 'master' ]; then
```
armenzg, line 95: This prevents a PR affecting the execution of another PR or a CI run on `master`.

.github/workflows/ci.yml
armenzg: I've fixed the caching and it now takes less than 15 minutes.

.github/workflows/ci.yml
```diff
@@ -120,22 +135,34 @@ jobs:
         run: |
           echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u $GITHUB_ACTOR --password-stdin
 
+      - name: Pull snuba CI images
+        run: |
+          docker pull ${SNUBA_IMAGE}:${{ steps.branch.outputs.branch }} || \
+            docker pull ${SNUBA_IMAGE}:latest || true
```
armenzg, line 141: This *and* using `BUILDKIT_INLINE_CACHE=1` fixes the caching for this build.

.github/workflows/ci.yml
```diff
@@ -120,22 +135,34 @@ jobs:
...
-          docker buildx build --platform linux/arm64 --cache-from $SNUBA_IMAGE:latest \
-            -t $SNUBA_IMAGE:${{ github.sha }} -t $SNUBA_IMAGE:latest --load .
+          docker buildx build --platform linux/arm64 \
+            --build-arg BUILDKIT_INLINE_CACHE=1 \
+            --cache-from ${SNUBA_IMAGE}:latest \
+            --cache-from ${SNUBA_IMAGE}:${{ steps.branch.outputs.branch }} \
+            -t ${SNUBA_IMAGE}:latest \
+            -t ${SNUBA_IMAGE}:${{ steps.branch.outputs.branch }} \
+            -t ${SNUBA_IMAGE}:${{ github.sha }} \
+            --load .
```
armenzg, line 153: This change is moving each argument into a new line, adding `--build-arg BUILDKIT_INLINE_CACHE=1` and using `--cache-from` for a branch.

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
-  - repo: git://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
```
armenzg, line 6: It suddenly [started failing](https://github.com/getsentry/snuba/runs/4080712275?check_suite_focus=true):

```
pre-commit install --install-hooks
pre-commit installed at .git/hooks/pre-commit
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for git://github.com/pre-commit/pre-commit-hooks.
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: remote error: 
      The unauthenticated git protocol on port 9418 is no longer supported.
    Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
    
Check the log at /home/runner/.cache/pre-commit/pre-commit.log
make: *** [Makefile:11: setup-git] Error 3
```

armenzg:
@billyvg I've simplified this PR to only be focusing on the fixes, thus, the diff will look smaller than what you looked at yesterday.

billyvg (approved):

.pre-commit-config.yaml
evanh: You can revert this I think, i have a separate PR that upgrades the pre-commit version and that fixes this. https://github.com/getsentry/snuba/pull/2183

evanh (approved):
LGTM minus the pre-commit change, that is not needed. Not sure if you want Billy or someone more knowledgeable to confirm though.

PR #2181, opened 2021-11-01
feat(suspect-spans): Allow enabling ingestion by percentage

Zylphrex:
Previously, suspect span ingestion can only be enabled on a per project basis.
This change allows us to enable it for a percentage of projects at a time. See
getsentry/getsentry#6531 and getsentry/sentry#29692 for the sentry PRs.


evanh (approved):
Well done, thanks!

PR #2182, opened 2021-11-01
feat(admin): Snuba admin prototype

lynnagara:
Snuba admin prototype


snuba/admin/clickhouse/system_queries.py
```diff
@@ -0,0 +1,100 @@
...
+    assert query
+    clickhouse_port = 9000
+    # TODO: don't create a new pool every time, that's no good
+    storage_key = StorageKey(storage_name)
+    storage = get_storage(storage_key)
+    (clickhouse_user, clickhouse_password) = storage.get_cluster().get_credentials()
+    database = storage.get_cluster().get_database()
+    connection = ClickhousePool(
+        clickhouse_host, clickhouse_port, clickhouse_user, clickhouse_password, database
+    )
```
fpacifici, line 96: Are you planning to keep a connection pool instance around and reuse it?
I think it may not be that useful to do that since we are going to run queries very rarely and over a large amount of hosts. So keeping all those connections open to use them rarely is not useful.

snuba/admin/views.py
```diff
@@ -0,0 +1,49 @@
...
+@application.route("/run_clickhouse_query", methods=["POST"])
+def clickhouse() -> str:
+    # TODO: You can do something like this to get all the hosts:
+    # SELECT * FROM system.clusters
+    req = request.get_json()
+    print("REQUEST: ", request, req)
+    results, columns = run_query(
+        req.get("host", "localhost"),
+        req.get("storage", "transactions"),
+        req.get("query_name"),
```
fpacifici, line 33: I think you should be already able to use the Clusters here. If you open the clusters from the snuba settings, a cluster can give you the list of hosts.
https://github.com/getsentry/snuba/blob/master/snuba/clusters/cluster.py#L336

PR #2185, opened 2021-11-03
Runtime changes slack notifications

MeredithAnya:
Most basic of basic slack app to send notifications about changes to runtime config:

**updated config**: 
> Screen Shot 2021-11-02 at 5 00 32 PM

**new config**:
> Screen Shot 2021-11-02 at 5 10 32 PM

**removed config**:
>Screen Shot 2021-11-02 at 5 16 41 PM

To test locally:
1. Have a test slack workspace you want to build the app on
1. Go to https://api.slack.com/apps and click on **Create New App** (create it on your test workspace)
1. Give your app a bot scope of `chat.write` (under **OAuth & Permissions** section)
1. Under settings click **Install App**
1. Add bot app to your desired channel (only one for now)
1. export `SLACK_API_TOKEN` in your terminal (token is under **OAuth & Permissions** section)
1. export `SNUBA_SLACK_CHANNEL_ID` in your terminal (see https://github.com/getsentry/snuba/pull/2185#issuecomment-958545503)

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -6,3 +6,22 @@ application = Flask(__name__, static_url_path="", static_folder="dist")
 @application.route("/")
 def root() -> Response:
     return application.send_static_file("index.html")
+
+
+@application.route("/test-slack")
+def test_slack() -> Response:
+    from typing import Dict, Union
+
+    from snuba.admin.slack.client import runtime_config_slack_client
+
+    data: Dict[str, Union[str, float, int]] = {
+        "option": "enable_events_read_only_table",
+        "old": 0,
+        "new": 1,
+    }
+
+    runtime_config_slack_client.notify(
+        action="Removed", option_data=data, user="meredith@sentry.io"
+    )
+
+    return application.send_static_file("index.html")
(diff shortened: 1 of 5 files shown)


MeredithAnya:
To get channel details you just go the channel and while you are in it click at the top, should say "Get channel details". Open that up and the channel id will be at the bottom 

Screen Shot 2021-11-02 at 5 39 20 PM

snuba/admin/views.py
```diff
@@ -6,3 +6,22 @@
...
+
+    from snuba.admin.slack.client import runtime_config_slack_client
+
+    data: Dict[str, Union[str, float, int]] = {
+        "option": "enable_events_read_only_table",
+        "old": 0,
+        "new": 1,
+    }
+
+    runtime_config_slack_client.notify(
```
MeredithAnya, line 23: @nikhars I left this up just to show an example of what I was thinking we could do. I think we probs should change the action to be an enum but yeah anyway some sort of action arg

snuba/settings/settings_test.py
```diff
@@ -19,3 +19,7 @@
 # something and you're at your wit's end, try setting this to False
 # to explore the unrefined Expression structure
 PRETTY_FORMAT_EXPRESSIONS = True
+
+# Admin
+SLACK_API_TOKEN = os.getenv("SLACK_API_TOKEN")
+SNUBA_SLACK_CHANNEL_ID = os.getenv("SNUBA_SLACK_CHANNEL_ID")
```
lynnagara, line 25: I don't think these lines are needed. I think this is only if we are running with different values in tests. As far as I can tell these aren't being used in tests and are not different from the original ones.

MeredithAnya: ahhhh okays yeah i'll take it out

PR #2186, opened 2021-11-03
feat(admin): Minimal Snuba admin

lynnagara:
This an extremely minimalistic version of https://github.com/getsentry/snuba/pull/2182. The goal is to merge a stripped back version of Snuba admin so that we can test the deployment process while finishing development, as well as hopefully making it easier to review.

Most functionality has been removed except static files and the health check endpoint.


vladkluev (approved):

nikhars (commented):
LGTM

nikhars (approved):

PR #2187, opened 2021-11-03
feat(config): check value types

nikhars:
The current config system does not check the type of the values being set
for keys. This allows for user mistakes where the type of the value could
be changed unknowingly. Added type checking for values and an option to
override it in case the existing value type was incorrectly set.

### Tests
Tests using CLI which is the primary method of setting configs today

```
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config set abc 1                                                                                       ok | direnv  3.8.11 py  10:33:39 AM
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config set abc 2                                                                                       ok | direnv  3.8.11 py  10:40:51 AM
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config set abc 0.5                                                                                     ok | direnv  3.8.11 py  10:41:00 AM
Mismatched types for abc: Original type: <class 'int'>, New type: <class 'float'>
  File "/Users/nikharsaxena/workspace/snuba/snuba/state/__init__.py", line 131, in set_config
    raise MismatchedTypeException(key, type(original_value), type(value))
snuba.state.MismatchedTypeException: ('abc', <class 'int'>, <class 'float'>)
The new value type <class 'float'> does not match the old value type <class 'int'>. Use the force option to disable this check
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config set abc 0.5 --force                                                                             ok | direnv  3.8.11 py  10:41:20 AM
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config set abc some-string                                                                             ok | direnv  3.8.11 py  10:41:32 AM
Mismatched types for abc: Original type: <class 'float'>, New type: <class 'str'>
  File "/Users/nikharsaxena/workspace/snuba/snuba/state/__init__.py", line 131, in set_config
    raise MismatchedTypeException(key, type(original_value), type(value))
snuba.state.MismatchedTypeException: ('abc', <class 'float'>, <class 'str'>)
The new value type <class 'str'> does not match the old value type <class 'float'>. Use the force option to disable this check
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config set abc some-string --force                                                                     ok | direnv  3.8.11 py  10:41:46 AM
 ~/workspace/snuba  nikhar/type-check-values *8 ?2  snuba config get-all                                                                                         ok | direnv  3.8.11 py  10:41:53 AM
test-float = 0.5 (float)
test-str = 'some-string' (str)
abc = 'some-string' (str)
```


snuba/state/__init__.py
```diff
@@ -130,24 +138,40 @@ def abtest(value: Optional[Any]) -> Optional[Any]:
...
 
-def set_config(key: str, value: Optional[Any], user: Optional[str] = None) -> None:
+def set_config(
+    key: str, value: Optional[Any], user: Optional[str] = None, force: bool = False
+) -> None:
     if value is not None:
-        value = "{}".format(value).encode("utf-8")
+        enc_value = "{}".format(value).encode("utf-8")
+    else:
+        enc_value = b""
```
lynnagara, line 147: Is handling of None values different to before?

snuba/state/__init__.py
```diff
@@ -130,24 +138,40 @@ def abtest(value: Optional[Any]) -> Optional[Any]:
...
 
     try:
-        original_value = rds.hget(config_hash, key)
-        if value == original_value:
-            return
+        enc_original_value = rds.hget(config_hash, key)
+        if enc_original_value is not None and value is not None:
+            original_value = numeric(enc_original_value.decode("utf-8"))
```
lynnagara, line 152: Do we need to cast to number? How about just enforcing the new value is the same as the original raw one? My motivation for asking is that I personally think it would be better to do away with `numeric` completely and just return whatever type the user actually asked for. What do you think?

snuba/state/__init__.py
```diff
@@ -130,24 +138,40 @@ def abtest(value: Optional[Any]) -> Optional[Any]:
...
             rds.hdel(config_hash, key)
             rds.hdel(config_history_hash, key)
         else:
-            rds.hset(config_hash, key, value)
+            rds.hset(config_hash, key, enc_value)
             rds.hset(config_history_hash, key, json.dumps(change_record))
         rds.lpush(config_changes_list, json.dumps((key, change_record)))
         rds.ltrim(config_changes_list, 0, config_changes_list_limit)
+        logger.info(f"Successfully changed option {key} to {value}")
+        # TODO: Link with the notification system
```
lynnagara, line 169: I'm not sure if config system should depend on the notification system (slack, etc). Config sort of seems like it should be a lower level thing to me. Ideally the notification system could even use configs if we wanted.

tests/state/test_state.py
```diff
@@ -33,6 +35,41 @@ def test_config(self) -> None:
...
 
+    def test_config_types(self) -> None:
+        # Tests for ints
+        state.set_config("test_int", 1)
+        state.set_config("test_int", 2)
+        with pytest.raises(MismatchedTypeException):
+            state.set_config("test_int", 0.1)
+        with pytest.raises(MismatchedTypeException):
+            state.set_config("test_int", "some_string")
+        state.set_config("test_int", None)
```
lynnagara, line 46: can we add assertions for get_config for each of these cases too?

snuba/state/__init__.py
nikhars: It's not. A none value is basically a delete operation. Right now, when a delete happens, we delete the key also from the config history. 

That's not great since we don't have history of the delete operation. I am going to make the change wherein the delete is also in the history. In that case I intend to show an empty byte.

snuba/state/__init__.py
nikhars: `numeric` name is a bit misleading. It basically converts a string value to the proper typed value. So in essence, it is comparing that the 2 values are of the same type.

We shouldn't get rid of `numeric` just yet. In redis, we are only storing the encoded values. There is no type information in redis. All type information is inferred. Until we get to a point where we start storing types in redis as well, it needs to stay

snuba/state/__init__.py
nikhars: It's a placeholder for now. I can build an observer pattern for notifications so there's no dependency. Will work on it during integration

snuba/state/__init__.py
```diff
@@ -93,14 +100,20 @@ def wrapper() -> Any:
 
 
 def numeric(value: Any) -> Any:
+    # Return the given value based on its correct type
+    # It supports the following types: int, float, bool, string
     try:
-        assert isinstance(value, (str, SupportsInt))
+        assert isinstance(value, (str, int))
```
nikhars, line 106: Changed it to int to differentiate between int's and bools

snuba/state/__init__.py
```diff
@@ -93,14 +100,20 @@ def wrapper() -> Any:
...
         return int(value)
     except (ValueError, AssertionError):
         try:
             assert isinstance(value, (str, SupportsFloat))
             return float(value)
         except (ValueError, AssertionError):
+            if value == "True" or value == "true":
+                return True
+            elif value == "False" or value == "false":
+                return False
```
nikhars, line 116: Added to support boolean's. In redis, booleans are stored as strings "True" or "False"

snuba/state/__init__.py
nikhars: Now delete's are also saved in the log history rather than deleting the key itself.

snuba/state/__init__.py
```diff
@@ -130,33 +143,48 @@ def abtest(value: Optional[Any]) -> Optional[Any]:
...
+                return
 
-        change_record = (time.time(), user, original_value, value)
+            if not force and type(value) != type(original_value):
+                raise MismatchedTypeException(type(original_value), type(value))
+
+        change_record = (time.time(), user, enc_original_value, enc_value)
         if value is None:
             rds.hdel(config_hash, key)
-            rds.hdel(config_history_hash, key)
```
nikhars, line 145: Removed since we want to maintain history of changes of keys which are deleted as well.

tests/conftest.py
```diff
@@ -118,6 +118,6 @@ def disable_query_cache() -> Generator[None, None, None]:
     cache, readthrough = state.get_configs(
         [("use_cache", settings.USE_RESULT_CACHE), ("use_readthrough_query_cache", 1)]
     )
-    state.set_configs({"use_cache": 0, "use_readthrough_query_cache": 0})
+    state.set_configs({"use_cache": False, "use_readthrough_query_cache": 0})
```
nikhars, line 121: The default value in settings file is boolean. So changed it to boolean

tests/state/test_state.py
```diff
@@ -33,6 +35,41 @@ def test_config(self) -> None:
...
+        state.set_config("test_str", "some_other_string")
+        with pytest.raises(MismatchedTypeException):
+            state.set_config("test_str", 1)
+        with pytest.raises(MismatchedTypeException):
+            state.set_config("test_str", 0.1)
+        state.set_config("test_str", None)
+
+        # Tests with force option
+        state.set_config("some_key", 1)
+        state.set_config("some_key", 0.1)
```
evanh: Don't you have to pass `force=True` here? I feel like I'm missing something.

snuba/cli/config.py
```diff
@@ -55,18 +57,33 @@ def get(*, key: str, format: str) -> None:
...
+@click.option(
+    "--force", is_flag=True, default=False, help="Override type checking of values"
+)
+def set(*, key: str, value: str, force: bool) -> None:
     "Set a single key."
-
-    state.set_config(key, value, user=get_user())
+    try:
+        typed_value = numeric(value)
+        state.set_config(key, typed_value, user=get_user(), force=force)
```
lynnagara, line 67: Do you need to call `numeric(value)` and pass that to set_config here? Why not have `state.set_config()` handle everything?

snuba/state/__init__.py
lynnagara: I wonder if there was a reason why that was happening in the first place. Will they ever be deleted from log history or do they stay there forever? Do you know what happens if we re-add the previously deleted key?

nikhars: > I wonder if there was a reason why that was happening in the first place. Will they ever be deleted from log history or do they stay there forever? Do you know what happens if we re-add the previously deleted key?
I was wrong about this. There is the audit log and and every config's last history. I have reverted that change to be like the original one

snuba/state/__init__.py
nikhars: I have removed the TODO comment from here

snuba/cli/config.py
nikhars: Moved the conversion to typed value in state.set_config

snuba/state/__init__.py
```diff
@@ -92,9 +98,13 @@ def wrapper() -> Any:
         return wrapper
 
 
-def numeric(value: Any) -> Any:
+def get_typed_value(value: Any) -> Any:
```
lynnagara, line 101: Ideally I think this should be:
```suggestion
def get_typed_value(value: str) -> Optional[str, int, float]:
```
I'm not sure why most of this file is just typed as `Any`.

snuba/state/__init__.py
```diff
@@ -104,59 +114,46 @@ def numeric(value: Any) -> Any:
...
-                return numeric(v)
-
-    return value
-
-
-def set_config(key: str, value: Optional[Any], user: Optional[str] = None) -> None:
-    if value is not None:
-        value = "{}".format(value).encode("utf-8")
+def set_config(
+    key: str, value: Optional[Any], user: Optional[str] = None, force: bool = False
```
lynnagara, line 118: Is value now just `Optional[str]` since we are doing the type conversion in here?

snuba/cli/config.py
```diff
@@ -55,18 +57,32 @@ def get(*, key: str, format: str) -> None:
 @config.command()
 @click.argument("key")
 @click.argument("value")
-def set(*, key: str, value: str) -> None:
+@click.option(
+    "--force", is_flag=True, default=False, help="Override type checking of values"
```
fpacifici, line 61: I would call this `force-type` since it is meant for a specific case that you want to force.

snuba/state/__init__.py
```diff
@@ -104,59 +114,46 @@ def numeric(value: Any) -> Any:
...
 
     try:
-        original_value = rds.hget(config_hash, key)
-        if value == original_value:
-            return
+        enc_original_value = rds.hget(config_hash, key)
+        if enc_original_value is not None and value is not None:
+            original_value = get_typed_value(enc_original_value.decode("utf-8"))
+            if value == original_value:
+                return
```
fpacifici, line 128: Is this intentional ?
```
>>> a = 1
>>> type(a)
<class 'int'>
>>> b = 1.0
>>> type(b)
<class 'float'>
>>> a == b
True
```

Basically an int can be equal to a float if they represent the same number but in terms of types they would still be different.

snuba/state/__init__.py
```diff
@@ -92,9 +98,13 @@ def wrapper() -> Any:
...
 
-def numeric(value: Any) -> Any:
+def get_typed_value(value: Any) -> Any:
+    # Return the given value based on its correct type
+    # It supports the following types: int, float, string
+    if value is None:
+        return None
     try:
-        assert isinstance(value, (str, SupportsInt))
+        assert isinstance(value, (str, int))
```
fpacifici, line 107: Any reason for this change ?

snuba/cli/config.py
lynnagara: Updated

snuba/state/__init__.py
lynnagara: Added a check for the type on this line as well

snuba/state/__init__.py
lynnagara: I assume to avoid a float being cast to an int here, as float implements __int__

lynnagara:
Made a bunch of the suggested changes, I think this is ready for another review.

snuba/web/views.py
```diff
@@ -298,16 +299,33 @@ def config(fmt: str = "html") -> Union[Response, RespTuple]:
...
+                    200,
+                    {"Content-Type": "application/json"},
+                )
+            except MismatchedTypeException as exc:
+                return (
+                    json.dumps(
+                        {
+                            "error": {
+                                "type": "client_error",
+                                "message": "Existing value and New value have different types. Use option force to override check",
```
evanh, line 318: I don't see a way to force it using the web API.

```suggestion
                                "message": "Existing value and New value have different types.",
```

evanh (approved):
One small change.

snuba/web/views.py
lynnagara: Good catch, will add it. That endpoint isn't used anywhere I believe, just something we can move over to admin.

lynnagara: Hmmm, this doesn't fit easily without changing the format of this API. I'm going to omit this since it's getting deprecated and not being used.

PR #2191, opened 2021-11-04
feat(admin): Add ability to view current configurations

lynnagara:
This change adds:
- A `/configs` endpoint in Snuba admin that returns the current config values (including type information)
- A table in the `Runtime config` page of Snuba admin app that displays all of the current configurations fetched from that API

There is currently no ability to modify the configurations in any way.


snuba/admin/static/runtime_config/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  // Load data if it was not previously loaded
+  if (data === null) {
+    fetchData();
+  }
+
+  function fetchData() {
+    api.getConfigs().then((res) => {
+      setData(res);
+    });
+  }
```
vladkluev, line 25: I don't think it really matters for this but just for my education:

Does this make it so that the render is blocked until data is fetched because this is a synchronous function

snuba/admin/static/table.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { ReactNode } from "react";
+
+import { COLORS } from "./theme";
+
+type TableProps = {
+  headerData: ReactNode[];
+  rowData: ReactNode[][];
```
vladkluev, line 7: Does this mean that a table element can also be a component?

vladkluev: So I can but buttons and other wacky stuff in there?

snuba/admin/views.py
```diff
@@ -11,3 +16,23 @@ def root() -> Response:
...
+
+
+def get_config_type(value: Union[str, int, float]) -> str:
+    if isinstance(value, str):
+        return "string"
+    if isinstance(value, int):
+        return "int"
+    if isinstance(value, float):
+        return "float"
+    raise ValueError("Unexpected config type")
```
vladkluev, line 38: I would prefer this be in its own file. I have a feeling this one would get polluted fairly quickly

vladkluev (approved):
A few nits and questions but looks great overall. Really glad we have someone with frontend expertise on this team. The table component is :chefskiss:

snuba/admin/static/runtime_config/index.tsx
lynnagara: We don't actually block rendering of the component. Basically we are just rendering an empty component until some data is received from the API then we show the table. React automatically triggers re-render soon after the state is updated by the `setData` hook, but it happens asynchronously.

snuba/admin/static/table.tsx
lynnagara: Yeah, the table element is a component. The idea was that we can just pass anything with any kind of formatting and style into the rowData array and it will stick it all into the table cells.

snuba/admin/views.py
lynnagara: Good point, this should probably move. Dumping it here for now as I'm not sure how it will fit with Nikhar's API changes yet.

lynnagara: Actually probably better to add just that module now, moved it over there.

nikhars (approved):

PR #2192, opened 2021-11-05
feat(admin): add system queries APIs to admin tool

vladkluev:
### Testing
ran with: `CLICKHOUSE_READONLY_USER=default snuba admin`

```
$ curl -d '{"query_name": "ActivePartitions", "host": "localhost", "port": 9000, "storage": "transactions" }' http://localhost:1219/run_clickhouse_system_query -H 'Content-Type: application/json'
{"column_names":["active","count()"],"rows":[["0","78"],["1","14"]]}

$ curl -d '{"query_name": "InactivePartitions", "storage": "errors"}' http://localhost:1219/run_clickhouse_system_query -H 'Content-Type: application/json'
{"data":{"extra_data":{"query_name":"InactivePartitions"}},"error":"NonExistentSystemQuery"}

$ curl -d '{"query_name": "ActivePartitions", "port": 4000 }' http://localhost:1219/run_clickhouse_system_query -H 'Content-Type: application/json'    
\{"data":{"extra_data":{"host":"localhost","port":4000}},"error":"InvalidNodeError"}
```
(status is 400 for validation errors)


snuba/admin/views.py
```diff
@@ -17,6 +18,31 @@ def health() -> Response:
...
+    return make_response(jsonify(res), 200)
+
+
+@application.route("/run_clickhouse_system_query", methods=["POST"])
+def clickhouse_system_query() -> Response:
+    # TODO: You can do something like this to get all the hosts:
+    # SELECT * FROM system.clusters
+    req = request.get_json()
+    results, columns = run_system_query(
+        req.get("host", "localhost"),
```
lynnagara, line 33: I think we should avoid letting the client pick any host they like and running the query there. Ideally the host and port values get picked up from the cluster that is configured in settings.

snuba/admin/clickhouse/system_queries.py
```diff
@@ -0,0 +1,109 @@
...
+    clickhouse_host: str, storage_name: str, system_query: Union[str, SystemQuery]
+) -> Tuple[Sequence[Any], Sequence[Tuple[str, str]]]:
+    query = (
+        SystemQuery.from_name(system_query)
+        if isinstance(system_query, str)
+        else system_query
+    )
+    if not query:
+        raise NonExistentSystemQuery
+    clickhouse_port = 9000
```
lynnagara, line 97: Should also come from the settings cluster configuration, as it does not necessarily have to be 9000

snuba/admin/views.py
vladkluev: But we want to be able to do system queries on different nodes right?

snuba/admin/clickhouse/system_queries.py
```diff
@@ -0,0 +1,109 @@
...
+    sql = """ SELECT
+        active,
+        count()
+    FROM system.parts
+    GROUP BY active
+    """
+
+
+def run_system_query(
+    clickhouse_host: str, storage_name: str, system_query: Union[str, SystemQuery]
```
fpacifici, line 88: It seems we should pass only a cluster to this function. 
Now in production, it is true that we do not have one cluster configured for clickhouse cluster, but that is something we can do and should do if useful for this. 
Now sure why we should take host, port and user from different places.

snuba/admin/views.py
fpacifici: I don't think we should let the client pick the host either.
Why cannot the client pick the storage or storage set and we resolve the cluster from there.
Of course we would have to fix our config to have all clusters properly defined. Which is something we should do anyway.
Then in order to identify the host you can pass shard and replica number. Or even the name at that point, so we can validate.

snuba/settings/__init__.py
```diff
@@ -56,6 +56,9 @@
     "subscriptions.process_message": 0.1,
 }
 
+CLICKHOUSE_READONLY_USER = os.environ.get("CLICKHOUSE_READONLY_USER", "readonly")
```
lynnagara, line 59: I think we should leave this as `default` and set to `readonly` in those environments where we have explicitly added a user with this name otherwise it will fail

snuba/admin/views.py
```diff
@@ -23,6 +30,41 @@ def health() -> Response:
...
+#  -H 'Content-Type: application/json' \
+#  http://localhost:1219/run_clickhouse_system_query
+@application.route("/run_clickhouse_system_query", methods=["POST"])
+def clickhouse_system_query() -> Response:
+    req = request.get_json()
+    try:
+        results, columns = run_system_query_on_host_by_name(
+            req.get("host", "localhost"),
+            req.get("port", 9000),
+            req.get("storage", "transactions"),
```
lynnagara, line 52: I think we should omit the defaults and make these required, these seem like fairly arbitrary values

snuba/admin/clickhouse/system_queries.py
```diff
@@ -0,0 +1,132 @@
...
+    sql = """ SELECT
+        active,
+        count()
+    FROM system.parts
+    GROUP BY active
+    """
+
+
+def _is_valid_node(host: str, port: int) -> bool:
+    return (host == "localhost" or host == "127.0.0.1") and port == 9000
```
lynnagara, line 93: I guess this is temporary. If we want to start with just one node, I'd suggest just validating it against the query node so it works in other environments not just dev.

snuba/admin/views.py
```diff
@@ -23,6 +30,41 @@ def health() -> Response:
...
+            req.get("query_name"),
+        )
+        res = {}
+        res["column_names"] = [name for name, _ in columns]
+        res["rows"] = []
+        for row in results:
+            res["rows"].append([str(col) for col in row])
```
lynnagara, line 59: Might be nice to have a column formatter that keeps the original types intact where possible

snuba/admin/clickhouse/system_queries.py
```diff
@@ -0,0 +1,109 @@
...
+        _QUERY_REGISTRY.register_class(cls)
+        return super().__init_subclass__()
+
+    @classmethod
+    def from_name(cls, name: str) -> Optional[Type["SystemQuery"]]:
+        return _QUERY_REGISTRY.get_class_by_name(name)
+
+    @classmethod
+    def all_queries(cls) -> Sequence[Type["SystemQuery"]]:
+        return _QUERY_REGISTRY.all_queries
```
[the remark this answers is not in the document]
lynnagara: Agree with this, seems a bit odd to expose the rest of the system queries here. Is there a reason why it's needed?

lynnagara (commented):
Just a few questions

snuba/settings/__init__.py
onewland: Sure, does this mean we have to wait on ops to test the tool, though?

snuba/admin/views.py
onewland: I'm fine with that

snuba/settings/__init__.py
lynnagara: We'll need to open a PR in ops to set `CLICKHOUSE_READONLY_USER="readonly"`. In the meantime it would use the default user.

snuba/admin/clickhouse/system_queries.py
onewland: Are you assuming I would fetch the cluster by storage name and use https://github.com/getsentry/snuba/blob/ea463d9d35ebdcd6957fb0055d68fad06b2b85df/snuba/clusters/cluster.py#L184-L185 or something and check that it matches the request input host/port?

snuba/admin/clickhouse/system_queries.py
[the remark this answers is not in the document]
onewland: Didn't write this but it looks like `_QueryRegistry` is a class intended to be private to this file. By making it a class method on SystemQuery we don't expose the existence of the registry to files outside this package.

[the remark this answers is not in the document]
lynnagara: I see, ok then

lynnagara (approved):
Looks good to me

PR #2193, opened 2021-11-05
feat(admin): Add runtime config audit log

lynnagara:
This change adds audit log funtionality to the admin app. Only
changes to runtime config are included in the audit log.

- There is a new endpoint /runtime_config that returns the list of
recent config changes
- There is a new tab in the UI which fetches the list from that endpoint
and displays the results in a tab

Screen Shot 2021-11-10 at 11 28 51 am


snuba/admin/static/runtime_config/types.tsx
```diff
@@ -8,4 +8,12 @@ type Config = { key: ConfigKey; value: ConfigValue; type: ConfigType };
 
 type RowData = [ReactNode, ReactNode, ReactNode];
 
-export { Config, ConfigKey, ConfigValue, ConfigType, RowData };
+type ConfigChange = {
+  key: ConfigKey;
+  user: string | null;
+  timestamp: number;
+  before: ConfigValue | null;
```
onewland, line 15: I think the canonical optional syntax in TS uses the `?`: https://stackoverflow.com/a/47942746

```suggestion
  before?: ConfigValue;
```

this lets you use the nice null-safe accesses like `configChangeInstance.before?.configValueProperty` which will be null if before is null

snuba/admin/static/runtime_config/auditlog.tsx
```diff
@@ -0,0 +1,104 @@
...
+        headerData={[`Timestamp (${usersTimeZone})`, "User", "Key", "Action"]}
+        rowData={rowData}
+        columnWidths={[1, 1, 1, 1]}
+      ></Table>
+    </div>
+  );
+}
+
+function getActionDetail(
+  before: ConfigValue | null,
```
onewland, line 66: My comment below about optionals applies here

onewland (commented):
Code looks good to me though I'm not 100% clear on the requirements

snuba/admin/static/runtime_config/auditlog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+  }
+
+  if (data === null) {
+    fetchData();
+  }
+
+  if (!data) {
+    return null;
+  }
```
onewland, line 24: should this be in `constructor`? I find the control flow a bit confusing here but maybe this is a react style that I'm not accustomed to.

snuba/admin/static/runtime_config/types.tsx
lynnagara: This is the typing of the data that is returned from the API. The keys `before` and `after` are always present but they can be null. The question mark implies it can be missing, which I don't think is the case, so I prefer the stricter way.

snuba/admin/static/runtime_config/auditlog.tsx
lynnagara: In class based React, I think this is normally in `componentDidMount` as async things don't normally happen in the constructor. In hooks based React, I think this is often inside a useEffect hook. But this seemed the easiest way to only run fetch once.

snuba/admin/static/runtime_config/auditlog.tsx
lynnagara: If we made them optional instead of nullable, then I don't think we could pass an after to this function without a before, as that comes later in the list of arguments.

lynnagara:
@onewland Appreciate the React review, I know not a lot of people in this team would be comfortable reviewing front end code.

snuba/admin/static/runtime_config/auditlog.tsx
```diff
@@ -0,0 +1,104 @@
...
+  before: ConfigValue | null,
+  after: ConfigValue | null
+): ReactNode {
+  const formattedBefore = typeof before === "string" ? `"${before}"` : before;
+  const formatttedAfter = typeof after === "string" ? `"${after}"` : after;
+
+  if (before === null && after !== null) {
+    if (after === null) {
+      // Value was null before and after the change, something went wrong
+      throw new Error("Unkknown action");
```
vladkluev, line 75: should this really throw or just make the error text something like 

❗ Value was null before and after the change, something went wrong ❗

vladkluev: because if we throw the whole page won't render right?

vladkluev (changes requested):
A few small changes.

Also for frontend changes what do you think of uploading a gif or image with the PR to show your changes visually?

snuba/admin/static/runtime_config/auditlog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    </div>
+  );
+}
+
+function getActionDetail(
+  before: ConfigValue | null,
+  after: ConfigValue | null
+): ReactNode {
+  const formattedBefore = typeof before === "string" ? `"${before}"` : before;
+  const formatttedAfter = typeof after === "string" ? `"${after}"` : after;
```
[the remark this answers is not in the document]
lynnagara: It just formats them without the quotes around it. In the PR stacked on top of this one https://github.com/getsentry/snuba/pull/2194 I fixed all the types and made them strings so we don't get weird formatting even for big numbers and things.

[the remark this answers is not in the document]
vladkluev: I see

snuba/admin/static/runtime_config/auditlog.tsx
lynnagara: Yeah, I want to add a component to render the error text properly on all of our pages if the API can't load or comes back with totally invalid data. I didn't get to it yet but still thought it's better to fail hard than quietly for now.

snuba/admin/static/runtime_config/auditlog.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { ReactNode, useState } from "react";
+
+import Client from "../api_client";
+import { containerStyle, paragraphStyle } from "./styles";
+import { Table } from "../table";
+import { ConfigValue } from "./types";
+
+function AuditLog(props: { api: Client }) {
+  const { api } = props;
+  const [data, setData] = useState<any[] | null>(null);
```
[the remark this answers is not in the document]
lynnagara: Yes, will fix.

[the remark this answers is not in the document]
lynnagara: Actually this is fixed in the other PR that comes after this one which addresses a bunch of type issues on both the UI and API https://github.com/getsentry/snuba/pull/2194. Will leave it as any here to avoid merging issues.

snuba/admin/static/runtime_config/auditlog.tsx
vladkluev: oh also, this is dead code,

```javascript
  if (before === null && after !== null) {
    if (after === null) {
      // dead code
    }
  }

lynnagara: Oops, thanks. Fixed.

vladkluev (approved):

PR #2194, opened 2021-11-08
feat(admin): Fix the typing of runtime config values

lynnagara:
All config values are now returned from the admin APIs as strings
along with their associated type infomration. The main reason for this
change is to avoid JavaScript quirks with numbers, such as the inability
to handle large integers.

Also updates the `state.get_config_changes()` function to return correctly
typed values for runtime config change history. The values are now in line
with what is returned by the other functions (instead of everything being
returned as a string).

Depends on https://github.com/getsentry/snuba/pull/2193


snuba/cli/config.py
```diff
@@ -1,7 +1,8 @@
-import click
 import json
 from typing import Any, Mapping
 
+import click
```
vladkluev, line 4: should we put a disclaimer at the top of the file that it is being deprecated in favor of the snuba admin portal?

vladkluev (approved):

PR #2195, opened 2021-11-08
Add more sentry logging to query parser

MeredithAnya:
This is to get more visibility into the breakdowns with the `build_request` span in Sentry, that sometimes seems to take a pretty long time without obvious reasons. 

What these added spans will look like in the Sentry product: 

> Screen Shot 2021-11-09 at 10 12 37 AM

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -16,6 +16,7 @@ from typing import (
     Union,
 )
 
+import sentry_sdk
 from parsimonious.exceptions import IncompleteParseError
 from parsimonious.grammar import Grammar
 from parsimonious.nodes import Node, NodeVisitor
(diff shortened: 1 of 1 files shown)


snuba/query/snql/parser.py
```diff
@@ -1300,23 +1301,28 @@ def _post_process(
...
-    query = parse_snql_query_initial(body)
+    with sentry_sdk.start_span(op="parser", description="parse_snql_query_initial"):
+        query = parse_snql_query_initial(body)
 
-    _post_process(
-        query, POST_PROCESSORS,
-    )
+    with sentry_sdk.start_span(op="processor", description="post_processors"):
+        _post_process(
+            query, POST_PROCESSORS,
```
evanh, line 1309: I would add a span for each processor as well. You could probably do it programatically in the `_post_process` function doing `start_span(... description=func.__name__)`. If you add it to `_post_process` it will also have the benefit of capturing all the other post processing functions below.

evanh (commented):
See my comment inline. Otherwise looks good.

evanh (approved):

PR #2197, opened 2021-11-09
feat: Add entity data model and wildcard columns

lynnagara:
This change adds an EntityColumnSet class which should be used for the logical entity data model. The original ClickHouse column set is still used by the physical ClickHouse data model. The entity column set shares the same core functionality as the ClickHouse column set, however it adds support for wildcard columns, that is columns without a fixed name but match some pattern such as `tags[...]` or `contexts[...]`.

This will allow us to provide more accurate entity schema definitions and add column based validations to queries before they are go through query processing and are run on ClickHouse.


evanh (approved):
LGTM.

PR #2203, opened 2021-11-10
log(replaced_groups): Add tags PostReplacementConsistencyEnforcer span

rahul-kumar-saini:
## Overview
- Span tells us redis pipeline execution is the bottleneck
- Need to figure out why that is
- Possibly a correlation between size of results and time taken to execute

## Changes
- Added a bunch of tags just to log the results

### DIFF
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -286,6 +286,7 @@ def get_projects_query_flags(
         op="function", description="execute_redis_pipeline"
     ) as span:
         results = p.execute()
+
         # getting size of str(results) since sys.getsizeof() doesn't count recursively
         span.set_tag("results_size", sys.getsizeof(str(results)))
 
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import random
+import sys
 import time
 import uuid
 from abc import abstractmethod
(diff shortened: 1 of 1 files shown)


snuba/datasets/errors_replacer.py
```diff
@@ -281,13 +282,23 @@ def get_projects_query_flags(
                 exclude_groups_key, 0, 0, withscores=True,
             )
 
-    with sentry_sdk.start_span(op="function", description="execute_redis_pipeline"):
+    with sentry_sdk.start_span(
+        op="function", description="execute_redis_pipeline"
+    ) as span:
         results = p.execute()
+        span.set_tag("results", results)
+        span.set_tag("results_size", sys.getsizeof(str(results)))
```
rahul-kumar-saini, line 290: Got the size of `str(results))` since `sys.getsizeof()` does not count recursively.

snuba/datasets/errors_replacer.py
```diff
@@ -281,13 +282,23 @@ def get_projects_query_flags(
                 exclude_groups_key, 0, 0, withscores=True,
             )
 
-    with sentry_sdk.start_span(op="function", description="execute_redis_pipeline"):
+    with sentry_sdk.start_span(
+        op="function", description="execute_redis_pipeline"
+    ) as span:
         results = p.execute()
+        span.set_tag("results", results)
```
vladkluev, line 289: do you want to be sending all the results to sentry? what would this give you?

snuba/datasets/errors_replacer.py
vladkluev: Put that in a comment

snuba/datasets/errors_replacer.py
rahul-kumar-saini: Yeah I guess I don't need to, I'm tagging the processed results a few lines down anyway.

vladkluev (approved):

PR #2204, opened 2021-11-11
feat(validation): Add Query Column Validation rollout plan

rahul-kumar-saini:
## Overview
- Queries don't have their columns validated against entity schema's aside from the required columns
- Queries can query for columns that do not exist and nothing validates it till ClickHouse itself
- Added a way to rollout query column validation per entity as Entities are updated with their new ColumnSet's

## Rollout plan
- At first, no column validation will actually happen, but the validator will exist
- Every time an Entity is updated to use the new `EntityDataModel` (#2197), we can flip the validation mode to `WARN` for the column validator for that Entity so we start getting warnings in Sentry when queries query for invalid columns
- We would monitor Sentry for warnings per Entity, and verify the validation would work as expected
- Once we are satisfied with the way the validator works, we can flip the validation mode to `ERROR` so the API starts sending back `400` responses with an appropriate message

## Changes
- Created a new `QueryValidator` called `EntityContainsColumnsValidator` which will do one of the following based on the validation mode set for the validator for the Entity it is defined for:
    - Do nothing
    - Send warnings to Sentry when a Query contains invalid columns for the Entity
    - Raise an error
- Updated the `Entity` base class to accept this new validator, with a `DO_NOTHING` mode by default
- Added a bunch of tests for this new functionality
- Set `Outcomes` Entity to `WARN` mode since the Columns for this Entity will look the same as the new Columns created in the new `EntityColumnSet` change


snuba/datasets/entities/outcomes.py
```diff
@@ -42,11 +47,12 @@ def __init__(self) -> None:
                     storage=materialized_storage,
                 ),
             ),
-            abstract_column_set=read_schema.get_columns(),
+            abstract_column_set=data_model,
             join_relationships={},
             writable_storage=writable_storage,
             validators=[EntityRequiredColumnValidator({"org_id"})],
             required_time_column="timestamp",
+            validate_data_model=ColumnValidationMode.WARN,
```
rahul-kumar-saini, line 55: Outcomes data model will still look the same after using the new `EntityDataModel` so we can start sending warnings to Sentry for invalid queries now

snuba/query/snql/parser.py
```diff
@@ -1198,7 +1198,7 @@ def validate_entities_with_query(
                 entity = get_entity(node.data_source.key)
                 try:
                     for v in entity.get_validators():
-                        v.validate(query)
+                        v.validate(query, alias)
```
rahul-kumar-saini, line 1201: Previously, no validators made use of the `alias` parameter. Now we need to use it for the Column validator in case of composite queries.

snuba/query/validation/validators.py
```diff
@@ -10,6 +13,14 @@
 )
 from snuba.query.exceptions import InvalidExpressionException, InvalidQueryException
 
+logger = logging.getLogger(__name__)
+
+
+class ColumnValidationMode(Enum):
+    DO_NOTHING = 0
+    WARN = 1
+    ERROR = 2
```
rahul-kumar-saini, line 22: Rollout model for what to do when queries are invalid.

tests/test_snql_api.py
```diff
@@ -549,3 +550,121 @@ def test_nullable_query(self):
             ),
         )
         assert response.status_code == 200
+
+    def test_invalid_column(self) -> None:
+        response = self.post(
+            "/discover/snql",
```
lynnagara, line 556: Outcomes queries should go to `/outcomes/snql` not `/discover/sqnl` -> it works because datasets are never checked today, we just honour whatever entity is asked for in the `MATCH` clause, but shouldn't really. Same for the other tests in this file.
```suggestion
            "/outcomes/snql",
```

lynnagara (approved):
Small nit in the test case, overall this seems good to me

snuba/query/validation/validators.py
```diff
@@ -55,6 +71,38 @@ def validate(self, query: Query, alias: Optional[str] = None) -> None:
...
+    def validate(self, query: Query, alias: Optional[str] = None) -> None:
+        if self.validation_mode == ColumnValidationMode.DO_NOTHING:
+            return
+
+        query_columns = query.get_all_ast_referenced_columns()
+
+        missing = set()
+        for column in query_columns:
+            if column.table_name == alias and column.column_name not in {
+                c.name for c in self.entity_data_model.columns
```
evanh, line 94: Can you move `{c.name for c in self.entity_data_model.columns}` outside the loop so we don't have an n^2 thing going on here. Also I think columnsets already implement a `__contains__` function, is there a reason you're not leveraging that?

snuba/query/validation/validators.py
```diff
@@ -55,6 +71,38 @@ def validate(self, query: Query, alias: Optional[str] = None) -> None:
...
+            }:
+                missing.add(column.column_name)
+
+        if missing:
+            error_message = f"query column(s) {', '.join(missing)} do not exist"
+            if self.validation_mode == ColumnValidationMode.ERROR:
+                raise InvalidQueryException(error_message)
+            elif self.validation_mode == ColumnValidationMode.WARN:
+                logger.warning(error_message)
```
evanh, line 103: I'm pretty sure you have to have `exc_info=True` in the logging call to get them sent to Sentry.

tests/test_snql_api.py
```diff
@@ -549,3 +550,121 @@ def test_nullable_query(self):
...
+            ),
+        )
+        assert response.status_code == 200
+
+    MATCH = "MATCH (e: events) -[grouped]-> (gm: groupedmessage)"
+    SELECT = "SELECT e.group_id, gm.status, avg(e.retention_days) AS avg BY e.group_id, gm.status"
+    WHERE = "WHERE e.project_id = 1 AND gm.project_id = 1"
+    TIMESTAMPS = "AND e.timestamp >= toDateTime('2021-01-01') AND e.timestamp < toDateTime('2021-01-02')"
+
+    invalid_columns_composite_query_tests = [
```
evanh, line 604: Probably worth adding a nested column here too. `SELECT uniq(toString(e.fdsdsf))...`

PR #2208, opened 2021-11-11
chore(kafka): Add a log message of librdkafka log level

nikhars:
We want to know what level the librdkafka is logging at in production.
Added a log message to find it out


snuba/consumers/consumer_builder.py
```diff
@@ -90,6 +93,7 @@ def __init__(
         self.broker_config = get_default_kafka_configuration(
             topic, bootstrap_servers=kafka_params.bootstrap_servers
         )
+        logger.info(f"librdkafka log level: {self.broker_config.get('log_level', 6)}")
```
vladkluev, line 96: How do you know the value is `6` if it's not set on the broker config?

nikhars: Looking at https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md, if you look at the default value for `log_level`, it is set to 6

vladkluev (approved):

snuba/consumers/consumer_builder.py
fpacifici: I don't think you need to log this to know.
We are setting that config here 
https://github.com/getsentry/arroyo/blob/c1765a28364f020426b4afde8c3cf3eae197b912/arroyo/backends/kafka/configuration.py#L61
So it is based on the log level we run the consumer at (which is INFO)

nikhars: Yes, I had looked at that code before. Even when we run locally, the default is `INFO` as per https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L6. So I was assuming librdkafka's log level to be `6`. But when I added the log message and ran a consumer, I got the level as `7` which stumped me
`10:29:31 consumer                            | 2021-11-12 10:29:31,582 librdkafka log level: 7`

PR #2211, opened 2021-11-11
remove test from `make test` that requires CI resources

onewland:
## Description
`tests/migrations/test_runner_add_node.py::test_add_node` fails when a user doesn't have a multi-node Clickhouse cluster set up, which exists on CI but not on developer local environments.

Had to force push after my initial PR to rebase

## Testing
The test continues to run (and passes) on CI: https://github.com/getsentry/snuba/runs/4183977329?check_suite_focus=true

but is skipped when I run `make test` locally


onewland:
mypy got cancelled right after finishing, so we can treat that as a success

fpacifici (approved):
Looks good to me. Please verify why the CI is not clean (it is not your change)

onewland:
I'll rerun, I think it was a transient fleet issue in CI.

PR #2212, opened 2021-11-12
fix(validation): Disable warnings on outcomes entity

rahul-kumar-saini:
## Overview
- Enabling Sentry warnings for invalid queries on the outcomes entity caused a flood of warnings ~12k events in 10 minutes

## Changes
- Disabled warnings on outcomes till we can figure out a way to sample or reduce number of events we log

### DIFF
--- a/snuba/datasets/entities/outcomes.py
+++ b/snuba/datasets/entities/outcomes.py
@@ -17,10 +17,7 @@ from snuba.query.processors.object_id_rate_limiter import (
 )
 from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
 from snuba.query.timeseries_extension import TimeSeriesExtension
-from snuba.query.validation.validators import (
-    ColumnValidationMode,
-    EntityRequiredColumnValidator,
-)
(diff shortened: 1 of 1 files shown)


snuba/datasets/entities/outcomes.py
```diff
@@ -66,7 +63,8 @@ def __init__(self) -> None:
             writable_storage=writable_storage,
             validators=[EntityRequiredColumnValidator({"org_id"})],
             required_time_column="timestamp",
-            validate_data_model=ColumnValidationMode.WARN,
```
nikhars, line 69: Why don't you just use the `DO_NOTHING` value here?

rahul-kumar-saini: The `Entity` base class defaults to `DO_NOTHING`. Setting it to `WARN` or `ERROR` here is to override it on an entity by entity basis.

https://github.com/getsentry/snuba/blob/655a05079fee0828f72e2f05fa2cf8946f11826b/snuba/datasets/entity.py#L37

nikhars (approved):

PR #2220, opened 2021-11-15
feat(suspect-spans): Start emitting spans.exclusive_time as float32

Zylphrex:
Begin emitting `spans.exclusive_time_32` with `input_format_skip_unknown_fields`
set to 1. This is step 1 and 2 of the migration process as explained at
 https://github.com/getsentry/snuba/pull/2105#issuecomment-928220731


evanh (commented):
Could you wrap the new column in a setting so we can toggle it off just in case we start getting errors? I don't want to have to wait for a rollback if this doesn't work for some reason.

PR #2222, opened 2021-11-16
feat(metric-alerts): Adds subscription consumer for metrics [INGEST-623]

ahmedetefy:
This PR adds subscription consumer for metrics (hidden behind a feature flag) to power crash rate alerts. Also, modifies the kafka stream loader for `metrics_counters` and `metrics_sets` to write to the same commit log topic and the same subscription results topic. 
Upon conversations with @fpacifici , we decided that this is the approach to support consumers for a dataset with multiple entities


lynnagara:
Not sure I get the motivation for making this change now. Why doing alert/subscriptions part before we even have the metrics data to run queries on and before doing the load testing? Ideally we can avoid ever putting metrics on the old subscriptions infrastructure and go straight to the new one.

snuba/datasets/storages/metrics.py
```diff
@@ -62,6 +63,17 @@
     ),
 )
 
+if settings.ENABLE_METRICS_SUBSCRIPTIONS:
```
fpacifici, line 66: No need to make a difference here. Just change the existing kafka stream loader, metrics are not in prod, no need to maintain two ways

fpacifici (commented):
@lynnagara, Likely your change will be ready before this will work in prod anyway. I would not be too concerned of having this in the code base as it won't be in prod any time soon.

@ahmedetefy, please hold off on this PR until you have a the whole pipeline working locally.

ahmedetefy:
@fpacifici I have validated that the entire pipeline works locally with some changes ofcourse

snuba/datasets/storages/metrics.py
```diff
@@ -76,7 +76,12 @@
     ),
     query_processors=[],
     stream_loader=build_kafka_stream_loader_from_settings(
-        processor=CounterMetricsProcessor(), default_topic=Topic.METRICS,
+        processor=CounterMetricsProcessor(),
```
jjbayer, line 79: @ahmedetefy just to double check, you said that making this conditional as with sessions is not necessary, right?

https://github.com/getsentry/snuba/blob/b55fe5027878533f065da8d4b3025cb55c8607eb/snuba/datasets/storages/sessions.py#L120-L132

jjbayer (approved):
Looks good to me, but I'll leave final approval to the code owners.

snuba/datasets/storages/metrics.py
ahmedetefy: Yes since it is not in production yet according to @fpacifici

untitaker (approved):

fpacifici (approved):

PR #2223, opened 2021-11-16
feat(metric-alerts): Set granularity for subscriptions [INGEST-626]

ahmedetefy:
Adds logic that sets the granularity of subscription queries based on the time interval passed in query


snuba/query/processors/granularity_processor.py
```diff
@@ -31,7 +34,20 @@ def __get_granularity(query: Query) -> int:
         )
 
     def process_query(self, query: Query, request_settings: RequestSettings) -> None:
-        granularity = self.__get_granularity(query)
+        granularity = None
+        if isinstance(request_settings, SubscriptionRequestSettings):
+            from_date, to_date = get_time_range(query, "timestamp")
```
evanh, line 39: Even though this processor is currently only used in metrics, you could fetch the time column programmatically so that you don't have to hardcode it. See my example below (you might have to add extra type checking).


```suggestion
            entity = get_entity(query.get_from_clause().key)
            from_date, to_date = get_time_range(query, entity.required_time_column)
```

ahmedetefy: sounds good

lynnagara:
What does setting this field `granularity` in a subscription actually do? I'm confused about what granularity means in this context since subscription queries never return any raw data to be grouped, just an aggregated number. As far as I know we never combined the concepts of granularity together with subscriptions before as they weren't very compatible.

ahmedetefy:
> What does setting this field `granularity` in a subscription actually do? I'm confused about what granularity means in this context since subscription queries never return any raw data to be grouped, just an aggregated number. As far as I know we never combined the concepts of granularity together with subscriptions before as they weren't very compatible.

In this particular context, granularity is just a column on the table, so it basically just filters according to this granularity value

Screenshot 2021-11-17 at 08 54 17

ahmedetefy:
@fpacifici This is how I addressed the granularity topic we discussed in the last meeting

snuba/query/processors/granularity_processor.py
```diff
@@ -30,8 +35,32 @@ def __get_granularity(query: Query) -> int:
...
+                granularity = 60
+            else:
+                granularity = 60 * 60
+        return granularity
+
     def process_query(self, query: Query, request_settings: RequestSettings) -> None:
-        granularity = self.__get_granularity(query)
+        granularity = None
+        if isinstance(request_settings, SubscriptionRequestSettings):
+            granularity = self.__get_granularity_for_subscription(query)
```
jjbayer, line 60: Should we raise an exception in this branch if `query.get_granularity()` is not None? It might be surprising to get a different granularity than expected.

ahmedetefy: For subscriptions? there is no way for a user to define a granularity on a crash rate alert anyway

untitaker (changes requested):
I think before we merge this we need to have a discussion with sns on where we actually want to set granularity.

* This PR overrides granularity for all subscriptions even if Sentry sets one. I'd rather have it only set one if Sentry doesn't send one.
* I am not sure we want such logic in Snuba at all? We have similar logic in Sentry already for determining rollup based on timeframe.
* See also `__get_granularity` a couple lines above the function introduced in this PR, could potentially be merged with this one. @jjbayer said in a meeting that this is just a more sophisticated version of the other.

fpacifici:
I think I need some additional details to review this. 
How would a metric subscription look like?
- Is there any boundary we want to set in time range ? I see you are setting the granularity for time ranges longer than one hour. 
- Are we allowing all aggregations from metrics ?

snuba/query/processors/granularity_processor.py
```diff
@@ -30,8 +35,32 @@ def __get_granularity(query: Query) -> int:
...
+            elif time_interval <= timedelta(hours=4):
+                granularity = 60
+            else:
+                granularity = 60 * 60
+        return granularity
+
     def process_query(self, query: Query, request_settings: RequestSettings) -> None:
-        granularity = self.__get_granularity(query)
+        granularity = None
+        if isinstance(request_settings, SubscriptionRequestSettings):
```
fpacifici, line 59: I don;t think this is the right place to add subscriptions specific logic.
My understanding of the reasoning for adding `EntitySubscription` was to allow customizations of the query that were specific of an entity in the subscription scenario. 

Also you cannot rely on the instance of RequestSetting for the correctness. That can change at any moment for any reason.

I think the right way to approach this is to make `EntitySubscription` more expressive. Was there any reason to allow `EntitySubscription` to only be able to return static conditions like `get_entity_subscription_conditions_for_snql` instead of returning processors to be executed on the query (@evanh do you remember ?)

evanh: That level of modification wasn't necessary at the time. Can I ask, why do you even need granularity with a subscription? That affects how the results are grouped, but there is only one result with a subscription right?

ahmedetefy: Nope with metrics the returned json looks something like this 
```
[{‘project_id’: 56, ‘tags[320]’: 321, ‘value’: 2.0}, {‘project_id’: 56, ‘tags[320]’: 328, ‘value’: 1.0}]
```

> That affects how the results are grouped

How so?

As far as I know, granularity is merely a column on the dataset

ahmedetefy: @fpacifici 

I am not entirely sure what you are suggesting with this 

>I think the right way to approach this is to make EntitySubscription more expressive. Was there any reason to allow EntitySubscription to only be able to return static conditions like get_entity_subscription_conditions_for_snql instead of returning processors to be executed on the query (@evanh do you remember ?)

fwiw I only added the `EntitySubscription` recently for sessions alerts because we wanted to keep the subscription data classes as abstract as possible, and some entities had specific conditions that need to be added 

I think a possible solution is to just add the granularity condition on `EntitySubscription` class like we do for project or org 
https://github.com/getsentry/snuba/blob/0c64d26865f21c68aa4f7f20f432ca0f3e9097f5/snuba/subscriptions/entity_subscription.py#L83-L92

evanh: Ah OK there are two concepts being conflated here. The `granularity` property of a query determines how we roll up the time field in queries (`toStartofDay`, `toStartOfHour` etc.)

If metrics have a `granularity` column, that wouldn't be handled using this, it would be used like a normal column in the query.

fpacifici: > If metrics have a granularity column, that wouldn't be handled using this, it would be used like a normal column in the query.

They do have it, so technically we could require the product to manually provide a granularity condition. I doubt that would be a good idea as it makes the entity very counter intuitive and the Snuba query has a granularity attribute, which is supposed to work even if we know it has limitations. 

That's the reason to set that condition in the logical layer with a processor till we redesign the `granularity` attribute 
https://github.com/getsentry/snuba/blob/master/snuba/query/processors/granularity_processor.py

Now that was before crash free alerts. That approach obviously does not work for subscriptions.

Back to @ahmedetefy's question. I thought a processing like this should be done by a query processor (or any ither kind of logic) provided by the EntitySubscription as we do not want to customize the format of the subscription. 
Though, thinking over this again, @evanh  with SnQL subscriptions are we not able to provide the granularity attribute as it is part of the query to be parsed ?
https://github.com/getsentry/snuba/blob/master/snuba/query/snql/parser.py#L116

evanh: You can absolutely do that. However I don't think we should be using the granularity field in the query as an alternative for the column in the metrics entity. That field means something specific, and if we do as your suggesting that means `query.get_granularity` will mean different things depending on the entity being queried.

fpacifici: I think we have two different concerns here.
Regarding subscriptions. @ahmedetefy I think we should keep the subscription consistent with how we run the query against the dataset whether that is the right or wrong abstraction. The dataset queries rely on the granularity field to identify the granularity so let's pass that field in the same way into subscriptions. You should be able to do that with SnQL.

Regarding the usage of granularity. @evanh, conceptually I think the way the granularity field is used in erros/transactions is the same as the way we are using it on metrics even though the implementation is radically different.
Specifically, all queries on metrics are on rollup pre-aggregated data we do not have raw data.
The granularity field in the errors/transactions tweaks the expressions that contain the timestamp so we can group by time buckets. 
In metrics this is not needed because we already stored pre-aggregated with multiple discrete granularities (each bucket  is multiplied a number of times equivalent to the number  of granularities we support). So when we want to have a hourly aggregation, we do not tweak the timestamp column according to the granularity field, we just filter out all the rows that are not rolled up hourly. 
The timestamp field is already aligned to the hour. 
So fundamentally setting the granularity field in a query on metrics has the same effect as setting it on errors (with the exception that metrics only support a number of pre-defined granularities). The implementation is very different as we do not deal with raw data.
Does this make sense ?

evanh: Yes, that makes sense. I'm fine with this then.

ahmedetefy: @fpacifici 
>Regarding subscriptions. @ahmedetefy I think we should keep the subscription consistent with how we run the query against the dataset whether that is the right or wrong abstraction. The dataset queries rely on the granularity field to identify the granularity so let's pass that field in the same way into subscriptions. You should be able to do that with SnQL.

Are you referring to passing it from the sentry side or adding it in the `SnqlSubscription` data class?

fpacifici: You should be able to pass it from the Sentry side as part of the query.
The create subscription POST endpoint allows you to pass a SnQL query
https://github.com/getsentry/snuba/blob/master/snuba/web/views.py#L438-L449
The SnQL query string can specify the `granularity` attribute 
https://getsentry.github.io/snuba/language/snql.html

ahmedetefy: Ah okay got you 
I will close this in favor of handling it on the sentry side. Thanks

PR #2226, opened 2021-11-17
perf(replaced_groups): Avoid max_groups final if Query has group_id's

rahul-kumar-saini:
## Overview
- Many Queries specifically query for certain group id's like `WHERE group_id = x` or `WHERE group_id IN [x, y, z, ...]`
- A Query that does this should be dealt with smartly for the case of replaced groups needing exclusion from the query
- - A query looking for group_id `x` should not be set to final if there are too many groups to exclude, since they wouldn't have been queried anyway
- The actual groups to exclude should be the intersection of the original groups to exclude coming from Redis and the groups the Query is specifically looking for

#### Examples (pseudocode and non exhaustive)
```python
query.condition = "WHERE group_id = 1"
query_groups = {1}
groups_to_exclude = {1, 2, 3}  # from redis
actual_groups_to_exclude = {1}
>>> replaced_groups.process()
>>> query.condition
"WHERE group_id = 1 AND group_id NOT IN [1]" 
>>> query.condition_before_this_change
"WHERE group_id = 1 AND group_id NOT IN [1, 2, 3]"
```
```python
query.condition = "WHERE group_id IN [1, 2, 3]"
query_groups = {1, 2, 3}
groups_to_exclude = {2, 3, 4} 
actual_groups_to_exclude = {2, 3}
>>> replaced_groups.process()
>>> query.condition
"WHERE group_id IN [1, 2, 3] AND group_id NOT IN [2, 3]" 
>>> query.condition_before_this_change
"WHERE group_id IN [1, 2, 3] AND group_id NOT IN [2, 3, 4]" 
```
```python
query.condition = "WHERE group_id IN [1, 2, 3]"
query_groups = {1, 2, 3}
max_groups_to_exclude = 5
groups_to_exclude = {5, 6, 7, 8, 9, 10, 11, 12}  #  > max
actual_groups_to_exclude = {}
>>> replaced_groups.process()
>>> query.condition
"WHERE group_id IN [1, 2, 3]"
>>> query.condition_before_this_change
"FINAL WHERE group_id IN [1, 2, 3]"
```

## Changes
- Updated `PostReplacementConsistencyEnforcer` to now find this intersection of group ids to exclude and continue with those instead of the original right from Redis

## Testing
- Added lots of new tests to test this functionality in `test_replaced_groups.py`

## Notes
- Removed 3 unnecessary sentry spans added previously to pinpoint processor performance issue in #2198 
- Most important sentry spans/tags for this are still intact


vladkluev:
Thanks for the great PR description. Do we know how much this is expected to impact performance?

rahul-kumar-saini:
> Thanks for the great PR description. Do we know how much this is expected to impact performance?

Thanks I try 😄 

And I'm not sure how much, but we definitely consistently receive queries looking for specific group ids, and I'm sure a percentage of those queries are set to final. The `avoided_max_groups` value for the `cause` tag I added at the end should give us insight into how much this change impacts performance.

tests/datasets/storages/processors/test_replaced_groups.py
```diff
@@ -198,3 +231,155 @@ def test_query_overlaps_replacements_processor(
...
+    assert not query_with_single_group_id.get_from_clause().final
+
+    set_project_exclude_groups(
+        2,
+        [100, 101, 102],
+        ReplacerState.EVENTS,
+        ReplacementType.EXCLUDE_GROUPS,  # Arbitrary replacement type, no impact on tests
+    )
+
+    # Too many groups to exclude, but query only needs one of them
```
evanh, line 263: I would split each of these cases into a separate test. That way you don't have to keep resetting the query object.

tests/datasets/storages/processors/test_replaced_groups.py
```diff
@@ -198,3 +231,155 @@ def test_query_overlaps_replacements_processor(
...
+    assert not query_with_multiple_group_ids.get_from_clause().final
+
+    set_project_exclude_groups(
+        2,
+        [100, 101, 102],
+        ReplacerState.EVENTS,
+        ReplacementType.EXCLUDE_GROUPS,  # Arbitrary replacement type, no impact on tests
+    )
+
+    # Too many groups to exclude, but query only needs a few of them
```
evanh, line 332: Same thing here, separate tests.

snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -45,10 +48,7 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
         if request_settings.get_turbo():
             return
 
-        with sentry_sdk.start_span(
```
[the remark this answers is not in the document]
rahul-kumar-saini: These functions take < 1 ms to execute and that won't change significantly across different queries, so these spans just clutter the performance view and our code. I'd say the only span we really care about is `execute_redis_pipeline` to track the performance of the Redis call over time.

evanh (approved):
LGTM, nice work!

PR #2227, opened 2021-11-17
feat(metric-alerts): Set default entity to METRICS_COUNTERS [INGEST-627]

ahmedetefy:
Modifies the default entity to metrics dataset to be metrics counter as the `get_default_entity` method is used for exclusively used for subscriptions and currently we are only making subscriptions work for crash rate alerts which use metrics counters

PS As a follow up to this PR, the subscription classes and web views need to be refactored to accept entities rather than datasets 
For example,
- Assumes one entity per dataset will be required for subscriptions which might not be the case with alerts on metrics data. Currently we will be supporting `METRICS_COUNTERS` but we might want to support alerts over `METRICS_DISTRIBUTIONS` and `METRICS_SETS`
https://github.com/getsentry/snuba/blob/badf1a0f2126af1da29712cf27abb5aabe8d521c/snuba/subscriptions/subscription.py#L26-L27
- Same here as ^ 
https://github.com/getsentry/snuba/blob/badf1a0f2126af1da29712cf27abb5aabe8d521c/snuba/subscriptions/subscription.py#L51-L59
- For creation of subscriptions as well 
https://github.com/getsentry/snuba/blob/267e48554c34933e2075dd9e72f76c20b085e10b/snuba/web/views.py#L438-L449


fpacifici:
I don't think we should approach the problem this way. 
While the first use case is indeed crash free alerts, mimicking that is not the goal of the project.
The goal of the project is to build a metric platform, so there is no gain in adding a hack to subscriptions in order to make it work in that specific case.

The right way to approach this is to actually fix the subscriptions generation pipeline. 
Said that, we should be already capable to create SnQL subscriptions, which embed the entity name. What specific problem did you run into ?

PR #2228, opened 2021-11-17
Anonimize querylog aliases

vladkluev:
### Context

In our queries, there ca be aliases that have user inputs in them, for example `snuba_tags[something]`. That `something` has the potential to be PII which we do not want to leak.

### Solution

In the Clickhouse query anonymizer, find aliases with square brackets and anonymize what is between them.


snuba/clickhouse/formatter/expression.py
```diff
@@ -35,22 +36,22 @@ class ClickhouseExpressionFormatterBase(ExpressionVisitor[str], ABC):
...
+        elif self._parsing_context.is_alias_present(alias):
             ret = escape_alias(alias)
             # This is for the type checker. escape_alias can return None if
             # we pass None. But here we do not pass None so a None return value
             # is not valid.
             assert ret is not None
             return ret
         else:
-            self.__parsing_context.add_alias(alias)
+            self._parsing_context.add_alias(alias)
```
vladkluev, line 54: had to change this to `_parsing_context` so that child classes could use it.

snuba/clickhouse/formatter/expression.py
```diff
@@ -219,3 +220,31 @@ def _format_datetime_literal(self, exp: Literal) -> str:
...
     def _format_date_literal(self, exp: Literal) -> str:
         return "$D"
+
+    def _anonimize_alias(self, alias: str) -> str:
+        # there may be an alias that looks like `snuba_tags[something]`
+        # the string between the square brackets has the potential to be PII
+        # This function will anonimize that which is between the brackets.
+        # this may erroneously anonimize aliases with square brackets
+        # if they are input by the user, but that is better than leaking PII
+        between_square_brackets_regex = re.compile(r"(?<=\[).*?(?=\])")
```
evanh, line 230: Compiling regex is expensive, please do this outside the function.

evanh: Also why is this using lookaheads/behinds? Why not `\[(.+)\]`? Then you would capture the string to replace directly.

snuba/clickhouse/formatter/expression.py
```diff
@@ -219,3 +220,31 @@ def _format_datetime_literal(self, exp: Literal) -> str:
...
+        alias_as_list = list(alias)
+        for match in between_square_brackets_regex.finditer(alias):
+            start_i, end_i = match.span()
+            alias_as_list[start_i:end_i] = "$" + "A" * (end_i - start_i - 1)
```
evanh, line 234: It would be a lot better if this was replaced with a single "$A" instead of an arbitrary number of them. It would make the anonymized queries more comparable.

snuba/clickhouse/formatter/expression.py
```diff
@@ -219,3 +220,31 @@ def _format_datetime_literal(self, exp: Literal) -> str:
...
+        return "".join(alias_as_list)
+
+    def _alias(self, formatted_exp: str, alias: Optional[str]) -> str:
+        if not alias:
+            return formatted_exp
+        elif self._parsing_context.is_alias_present(alias):
+            # ret = escape_alias(self._anonimize_alias(alias))
```
evanh, line 241: 💀

snuba/clickhouse/formatter/expression.py
vladkluev: That's an interesting point. If we want comparability, should we do a hash of the tag and put that in instead? That way we would know if it was the same one. Otherwise the comparability would be fake since we'd be saying certain aliases are the same when they aren't

snuba/clickhouse/formatter/expression.py
vladkluev: > Also why is this using lookaheads/behinds? Why not \[(.+)\]? Then you would capture the string to replace directly.

two reasons:

1. The lookaheads/behinds returns just the string within the square brackets
2. This way you can support arbitrary indexing. E.g. `_snuba_tags[something][something_else]` I know we don't currently do that but it seemed the more future proof solution

snuba/clickhouse/formatter/expression.py
evanh: That's an interesting idea. However I think what we want to be able to do with the anonymized querylog right now is look at the structure of the query, rather than the specific values of the query. Most of the times when we're optimizing we want to know how a certain column was queried, not the specific values it was queried for.  If we keep to the single `$A` then we can analyze how often particular patterns of query are used.

That's how the rest of the query is anonymized, with generic `$S` for strings etc. I do like the idea of having hashed values though. Maybe we should add a backlog to task to store a "hashed anonymized" version of the query that hashes the values so we can also compare between different values. It might be interesting to see queries that have the same structure, but perform differently with different values.

snuba/clickhouse/formatter/expression.py
evanh: That makes sense. However regex already has support for what you're trying to do, by wrapping things with brackets to make capture groups.

```suggestion
        between_square_brackets_regex = re.compile(r"(?<=\[)(.*?)(?=\])") # this should still be outside the function
        anonymized_alias = regex.sub("$A", alias)
```

If the input is `_snuba_tags[something][something_else]`, then the output of this should be `_snuba_tags[$A][$A]`.

snuba/clickhouse/formatter/expression.py
vladkluev: So I tried to do the `$A` solution but it actually made the code look much worse. Essentially when the alias that's substituted into the string is not of the same length, you either have to re-search the string (because it has changed) or but in some logic to re-calculate where the new start/end indexes are. Either way it doesn't look pretty. I implemented the hash version instead because that was actually easier and I think the code looks better. Let me know what you think

vladkluev: never mind, I forgot about re.sub

vladkluev: ok now it should do exactly what you wanted.

evanh (approved):

PR #2229, opened 2021-11-17
feat(dev): Add direnv script to set up Python venv

armenzg:
This creates a venv automatically if missing (if `direnv` is used) and uses a newer Python version to avoid installing a different version than Sentry's.


armenzg:
This is not anymore focused on fixing confluent-kafka failing for pre-commit hooks since it is handled in https://github.com/getsentry/sentry/pull/30094

evanh (approved):
Yes, finally. I like this.

PR #2231, opened 2021-11-17
feat(suspect-spans): Change exclusive time to Float32

Zylphrex:
Using Float32 is enough precision for the exclusive time column. Dropping the
original column and starting over. This column is only internal at the moment so
it is safe to drop any existing data. Before dropping the column, make sure the
`input_format_skip_unknown_fields` option is enabled and/or the processors are
not emitting the span columns to avoid errors in the transaction processors.
Additionally, make sure that `materialize_ttl_after_modify` is disabled prior
modifying the TTL to avoid massive merges.

# Forwards

```sql
Local operations:
ALTER TABLE transactions_local ADD COLUMN IF NOT EXISTS spans.exclusive_time_32 Array(Float32) AFTER spans.group;
ALTER TABLE transactions_local MODIFY COLUMN spans.exclusive_time_32 Array(Float32) TTL finish_ts + INTERVAL 1 MONTH;


Dist operations:
ALTER TABLE transactions_dist ADD COLUMN IF NOT EXISTS spans.exclusive_time_32 Array(Float32) AFTER spans.group;
```

# Backwards

```sql
Local operations:
ALTER TABLE transactions_local DROP COLUMN IF EXISTS spans.exclusive_time_32;


Dist operations:
ALTER TABLE transactions_dist DROP COLUMN IF EXISTS spans.exclusive_time_32;
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -40,27 +40,27 @@ steps:
       - "-v"
       - "--remove-orphans"
   - name: "gcr.io/$PROJECT_ID/docker-compose"
-    id: get-onpremise-repo
+    id: get-self-hosted-repo
     waitFor:
       # We need to wait until unit-tests finish as `pytest` tries
-      # to run tests under onpremise repo too
+      # to run tests under self-hosted repo too
       - unit-tests
     entrypoint: "bash"
     args:
       - "-e"
       - "-c"
       - |
-        mkdir onpremise && cd onpremise
-        curl -L "https://github.com/getsentry/onpremise/archive/master.tar.gz" | tar xzf - --strip-components=1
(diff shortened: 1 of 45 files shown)


Zylphrex:
Unsure how we typically enable these options, but the `input_format_skip_unknown_fields` option can be enabled via #2232 2232

evanh:
@fpacifici I see https://github.com/getsentry/ops/pull/4092 is still open. I think we disabled `materialize_ttl_after_modify` manually though right?

fpacifici:
We did not, we have to manually disable it every time we run a DDL operation

evanh (approved):
This all looks good to me. We'll have to wait until next week to deploy though.

PR #2233, opened 2021-11-18
fix(consumers): Allow bypassing header based routing for small percentage of messages

nikhars:
### Context
While introducing kafka header based routing helped us reclaim CPU cycles on consumers which use the same shared topic, it has a subtle side effect. In case there are only one type of messages on the shared topic, the consumer group lag for the consumers of other type of messages keeps increasing since they never get a chance to commit their offset. This PR
introduces a kafka header filter with bypass which basically allows a small number of messages which would have been dropped by the original rule to bypass the checks.

### Tests
- Added a test case with 100% bypass
- Local tests with client sending only errors

1. Initial consumer group data
[appuser@e6b0fda824dc ~]$ kafka-consumer-groups --bootstrap-server localhost:9092 --group transactions_group --describe

Consumer group 'transactions_group' has no active members.

GROUP              TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID     HOST            CLIENT-ID
transactions_group events          0          25314           25314           0               -               -               -
[appuser@e6b0fda824dc ~]$ kafka-consumer-groups --bootstrap-server localhost:9092 --group snuba-consumers --describe

Consumer group 'snuba-consumers' has no active members.

GROUP           TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID     HOST            CLIENT-ID
snuba-consumers events          0          25314           25314           0               -               -               -

Lag is 0 and End offsets are same for both error and transaction consumers

2. Start sending error data with consumers using existing filter. Not the one with bypass.
[appuser@e6b0fda824dc ~]$ kafka-consumer-groups --bootstrap-server localhost:9092 --group snuba-consumers --describe

GROUP           TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID                                  HOST            CLIENT-ID
snuba-consumers events          0          29274           29303           29              rdkafka-cf8ca1ed-0be0-4b22-8298-553cb00c0c91 /172.21.0.1     rdkafka
[appuser@e6b0fda824dc ~]$ kafka-consumer-groups --bootstrap-server localhost:9092 --group transactions_group --describe

Consumer group 'transactions_group' has no active members.

GROUP              TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID     HOST            CLIENT-ID
transactions_group events          0          27016           29704           2688            -               -               -

Since only errors are being sent, the lag is pretty low for errors consumer, around 29. For transactions consumer, no data is being sent hence the Lag is large 2688. This confirms the existing behavior.

3. Add the change to use the new bypass filter with 1 percent of messages.

4. Start sending new error data
[appuser@e6b0fda824dc ~]$ kafka-consumer-groups --bootstrap-server localhost:9092 --group transactions_group --describe

GROUP              TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID                                  HOST            CLIENT-ID
transactions_group events          0          40346           40496           150             rdkafka-992ff3d7-c699-4c8f-8aec-bbc2fc0b24b4 /172.21.0.1     rdkafka
[appuser@e6b0fda824dc ~]$ kafka-consumer-groups --bootstrap-server localhost:9092 --group snuba-consumers --describe

GROUP           TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID                                  HOST            CLIENT-ID
snuba-consumers events          0          40639           40666           27              rdkafka-5f6e7c39-c0e3-4818-b46f-d6a5fde61c88 /172.21.0.1     rdkafka

Lag for transactions consumer is now decreasing as well because of the bypass.


evanh (approved):
This works, but out of curiosity could we not change the consumer to commit messages it skips without doing any other processing? So it would see that a message should be skipped, and then go straight to committing the offset?

It seems like with your solution we will end up doing a lot of extra work for 1% of messages.

nikhars:
> This works, but out of curiosity could we not change the consumer to commit messages it skips without doing any other processing? So it would see that a message should be skipped, and then go straight to committing the offset?

It's hard to do that today because of the way arroyo processors are set up. The amount of work required to get to the ideal fix would take some time and refactoring.

> It seems like with your solution we will end up doing a lot of extra work for 1% of messages.
It used to do that work before the performance improvement was made. Doesn't cost much in terms of CPU processing to do the work for 1% of the messages.

snuba/datasets/message_filters.py
```diff
@@ -23,17 +25,48 @@ def should_drop(self, message: Message[KafkaPayload]) -> bool:
...
+    and avoiding unnecessary increase in consumer group lag in edge cases.
+    """
+
+    bypass_limit: float
+
+    def should_drop(self, message: Message[KafkaPayload]) -> bool:
+        if super().should_drop(message) == False:
+            return False
+
+        return False if random.random() <= self.bypass_limit else True
```
fpacifici, line 72: Why, instead of making this random, don't we count the number of consecutive skipped messages and and just send one through after X messages consistently? So we would have a more deterministic result

nikhars: Changed the logic to allow bypass based on consecutive drops.

PR #2234, opened 2021-11-18
add a per-project-referrer rate limiter

vladkluev:
We're getting abused on a by a specific project on a specific API referrer. Add a rate limiter for that.

### DIFF
--- a/snuba/datasets/storages/transactions.py
+++ b/snuba/datasets/storages/transactions.py
@@ -30,7 +30,6 @@ from snuba.query.processors.empty_tag_condition_processor import (
 )
 from snuba.query.processors.mapping_optimizer import MappingOptimizer
 from snuba.query.processors.mapping_promoter import MappingColumnPromoter
-from snuba.query.processors.object_id_rate_limiter import ProjectReferrerRateLimiter
 from snuba.query.processors.prewhere import PrewhereProcessor
 from snuba.query.processors.table_rate_limit import TableRateLimit
 from snuba.query.processors.type_converters.hexint_column_processor import (
@@ -149,7 +148,6 @@ storage = WritableTableStorage(
                 "title",
             ]
         ),
-        ProjectReferrerRateLimiter("project_id"),
         TableRateLimit(),
     ],
     stream_loader=build_kafka_stream_loader_from_settings(
--- a/snuba/datasets/storages/transactions.py
+++ b/snuba/datasets/storages/transactions.py
@@ -30,6 +30,7 @@ from snuba.query.processors.empty_tag_condition_processor import (
 )
 from snuba.query.processors.mapping_optimizer import MappingOptimizer
 from snuba.query.processors.mapping_promoter import MappingColumnPromoter
+from snuba.query.processors.object_id_rate_limiter import ProjectReferrerRateLimiter
 from snuba.query.processors.prewhere import PrewhereProcessor
 from snuba.query.processors.table_rate_limit import TableRateLimit
 from snuba.query.processors.type_converters.hexint_column_processor import (
@@ -148,6 +149,7 @@ storage = WritableTableStorage(
                 "title",
             ]
         ),
+        ProjectReferrerRateLimiter("project_id"),
         TableRateLimit(),
     ],
     stream_loader=build_kafka_stream_loader_from_settings(
--- a/snuba/query/processors/object_id_rate_limiter.py
+++ b/snuba/query/processors/object_id_rate_limiter.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from snuba.clickhouse.query_dsl.accessors import get_object_ids_in_query_ast
 from snuba.query.logical import Query
 from snuba.query.processors import QueryProcessor
@@ -6,6 +8,7 @@ from snuba.state import get_configs
 from snuba.state.rate_limit import (
     ORGANIZATION_RATE_LIMIT_NAME,
     PROJECT_RATE_LIMIT_NAME,
+    PROJECT_REFERRER_RATE_LIMIT_NAME,
     RateLimitParameters,
 )
 
@@ -24,11 +27,13 @@ class ObjectIDRateLimiterProcessor(QueryProcessor):
         rate_limit_name: str,
         per_second_name: str,
         concurrent_name: str,
+        request_settings_field: Optional[str] = None,
     ) -> None:
         self.object_column = object_column
         self.rate_limit_name = rate_limit_name
         self.per_second_name = per_second_name
         self.concurrent_name = concurrent_name
+        self.request_settings_field = request_settings_field
 
     def process_query(self, query: Query, request_settings: RequestSettings) -> None:
         # If the settings don't already have an object rate limit, add one
@@ -42,7 +47,13 @@ class ObjectIDRateLimiterProcessor(QueryProcessor):
             return
 
         # TODO: Add logic for multiple IDs
-        obj_id = obj_ids.pop()
+        obj_id = str(obj_ids.pop())
+        if self.request_settings_field is not None:
+            request_settings_field_val = getattr(
+                request_settings, self.request_settings_field, None
+            )
+            if request_settings_field_val is not None:
+                obj_id = f"{obj_id}_{request_settings_field_val}"
 
         object_rate_limit, object_concurrent_limit = get_configs(
             [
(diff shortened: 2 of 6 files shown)


snuba/datasets/storages/transactions.py
```diff
@@ -148,6 +149,7 @@
                 "title",
             ]
         ),
+        ProjectReferrerRateLimiter("project_id"),
```
vladkluev, line 152: Help: mypy gives me a type error here and I have no idea why

Is this not a query processor?

```
snuba/datasets/storages/transactions.py:152: error: List item 13 has incompatible type "ProjectReferrerRateLimiter"; expected "QueryProcessor"
```

snuba/query/processors/object_id_rate_limiter.py
```diff
@@ -24,11 +27,30 @@ def __init__(
...
+        self.request_settings_field = request_settings_field
+
+    def get_object_id(
+        self, query: Query, request_settings: RequestSettings
+    ) -> Optional[str]:
+        obj_ids = get_object_ids_in_query_ast(query, self.object_column)
+        if not obj_ids:
+            return None
+
+        # TODO: Add logic for multiple IDs
```
onewland, line 45: If `obj_id` is a generic string could this be a string join of `obj_ids` with ";". Or would that not group properly?

onewland:
Seems fine to me, did you test that it would effectively shed the bad load in your local environment?

snuba/query/processors/object_id_rate_limiter.py
vladkluev: that's not a bad idea, the only reason I can think of that being bad is that we hit the redis key size limit. For example observe this query:

```
SELECT
  event_id AS `_snuba_event_id` |> event_id,
  group_id AS `_snuba_group_id` |> group_id,
  project_id AS `_snuba_project_id` |> project_id,
  timestamp AS `_snuba_timestamp` |> timestamp
FROM
  Entity(events)

ORDER_BY
  timestamp AS `_snuba_timestamp` DESC,
  event_id AS `_snuba_event_id` DESC
WHERE
  and(
    greaterOrEquals(
      timestamp AS `_snuba_timestamp`,
      datetime(2021-08-20T21:25:06)
    ),
    and(
      less(
        timestamp AS `_snuba_timestamp`,
        datetime(2021-11-18T21:25:07)
      ),
      and(
        in(
          project_id AS `_snuba_project_id`,
          tuple(
            1,
            54785,
            5992455,
            11276,
            2783252,
            2053674,
            2721840,
            5250619,
            1419836,
            5250626,
            5250627,
            5250630,
            230989,
            1280593,
            5711442,
            3487317,
            155735,
            1492057,
            133215,
            282218,
            5662316,
            1932911,
            200821,
            5600888,
            5738630,
            300688,
            5176468,
            76439,
            5898904,
            5324467,
            5748916,
            5953206,
            1473210,
            5953213,
            1371326,
            5645511,
            4846280,
            1267915,
            3487436,
            1297620,
            1297621,
            5941975,
            5266138,
            1297627,
            1360617,
            1447659,
            5613292,
            5350637,
            1187568,
            37617,
            2783478,
            239368,
            15626,
            221969,
            1489174,
            191772,
            5905698,
            1492261,
            2423079,
            5246761,
            5493568,
            1886021,
            1505606,
            5641555,
            1465686,
            191841,
            1492835,
            1492842,
            5397869,
            5723503,
            5397873,
            162676,
            5397878,
            1460611,
            4857230,
            1466766,
            185747,
            5823380,
            1934741,
            5496745,
            1321390,
            1192882,
            1269704,
            5637584,
            1492949,
            5540827,
            5683165,
            5683166,
            5420004,
            1455593,
            5726700,
            1367021,
            2450415,
            5480435,
            47606,
            1292798
          )
        ),
        and(
          notEquals(
            type AS `_snuba_type`,
            'transaction'
          ),
          and(
            in(
              project_id AS `_snuba_project_id`,
              tuple(
                5898904,
                5905698,
                5992455,
                5941975,
                5953206,
                5953213,
                1886021,
                300688,
                2053674,
                5613292,
                5645511,
                5683165,
                5683166,
                11276,
                1267915,
                155735,
                200821,
                5266138,
                1,
                162676,
                5748916,
                5823380,
                239368,
                5738630,
                1934741,
                5246761,
                1465686,
                4857230,
                5726700,
                5723503,
                1492057,
                5711442,
                5662316,
                5641555,
                5637584,
                1292798,
                5600888,
                5480435,
                5350637,
                5324467,
                2423079,
                230989,
                76439,
                1489174,
                1492835,
                3487317,
                2783478,
                2721840,
                2450415,
                1321390,
                5540827,
                5250626,
                5496745,
                5420004,
                5397878,
                5397873,
                5397869,
                5250619,
                5250627,
                5250630,
                1280593,
                1269704,
                1192882,
                191841,
                54785,
                1187568,
                191772,
                185747,
                47606,
                37617,
                2783252,
                1505606,
                1473210,
                1460611,
                1455593,
                5176468,
                221969,
                282218,
                1371326,
                1419836,
                1492949,
                1447659,
                15626,
                1492842,
                1492261,
                1360617,
                1466766,
                1367021,
                133215,
                1297627,
                1297620,
                1297621,
                3487436,
                5493568,
                1932911,
                4846280
              )
            ),
            in(
              event_id AS `_snuba_event_id`,
              tuple(
                '821bf18dabfe479481df7773995e2051'
              )
            )
          )
        )
      )
    )
  )
  LIMIT 1

```

vladkluev:
> did you test that it would effectively shed the bad load in your local environment?

I did not because:

1. That is not something we've done before so setting up the environment to test that would take too long
2. This follows a very commonly used pattern within the codebase so I'm pretty confident this will work

onewland:
Not trying to block an urgent fix but I do think it's a bit concerning that it would take a long time to do an e2e validation of a new rate limiter. 

What if we deploy new rate limiting logic and find that it sheds 100% of traffic?

vladkluev:
Rate limiting can be configured without a redeploy through runtime config. We can turn off the rate limiter very quickly if we wanted to

snuba/query/processors/object_id_rate_limiter.py
onewland: sounds good, was just wondering if we could wipe out a TODO before merging to master

fpacifici (approved):

fpacifici (changes requested):
Wait though. 
Let's not return a rate limit parameter if the project is not the one we look for and the referrer is not the one we look for.

snuba/query/processors/object_id_rate_limiter.py
```diff
@@ -37,20 +59,15 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
...
         object_rate_limit, object_concurrent_limit = get_configs(
             [
                 (self.per_second_name, DEFAULT_LIMIT),
                 (self.concurrent_name, DEFAULT_LIMIT),
             ]
         )
-
+        obj_id = self.get_object_id(query, request_settings)
+        if obj_id is None:
+            return
```
fpacifici, line 70: This way we are adding this RateLimitParameters class to the rate limitier for all project_id+referrer pairs, which is a huge amount. Each one of them will write a key in Redis. 
We do not need that. 
Let's return a rate limit object only if there is a valid config for that specific referrer and that specific project without adding a default. 
That will ignore this rule for the majority of projects and referrer

vladkluev:
@fpacifici 

you only want to apply this for a specific project and referrer? So I should do a `snuba.state.get_config("rate_limited_project_referrers")` which should return a list of `["{project_id}__{referrer}"]` and we should check that?

fpacifici:
No, not that. 
Yes we do want to apply this only to specific referrers and projects. 
Though you do not need to do anything that fancy. What you have there already solves 99% of the problem.
This line will ensure a ProjectReferrerRateLimiter is always added. 
https://github.com/getsentry/snuba/pull/2234/files#diff-88e988806f2ba4d0fa6bd3d8fc090d9f1fa19c8cba994a6541c672d880a9cd08R86

Most cases that will contain the default values loaded from here:
https://github.com/getsentry/snuba/pull/2234/files#diff-88e988806f2ba4d0fa6bd3d8fc090d9f1fa19c8cba994a6541c672d880a9cd08R77

You just need to avoid applying the default if there is no specific proejct/referrer config setup and that step. Which means you will have to make this rate limiter behave slightly differently from others

fpacifici (approved):

PR #2236, opened 2021-11-19
feat: Subscription scheduler produces scheduled task messages

lynnagara:
This change integrates the ProduceScheduledSubscriptionMessage
step into the scheduler. Now the scheduler produces messages
for every scheduled task. This completes the core functionality of
the subscriptions scheduler.

A temporary `load_factor` option is also supported. The number
of subscriptions will be multiplied by the provided `load_factor`
to simulate a larger number of subscriptions for testing.

Thhe "scheduled-subscriptions-events" topic must be created in
production before this change can be merged and deployed.


snuba/cli/subscriptions_scheduler.py
```diff
@@ -56,11 +60,42 @@ def subscriptions_scheduler(
...
+
+    - The tick consumer consumes the commit log and reads the "orig_message_ts" header.
+    It constructs a new `Tick` message representing the intervals between each of the
+    original messages, which gets passed to the processing strategy. Note: A tick always
+    corresponds to a single partition on the original topic (not the commit log topic
+    as that is never partitioned).
+
+    - The first processing step is a tick buffer. It buffers ticks where needed and
+    determines when to submit them to the rest of the pipeline. The tick buffer behavior
+    depends on the watermark mode specsified by the entity. In PARTITION mode, ticks are
```
fpacifici, line 82: typo: specified

snuba/cli/subscriptions_scheduler.py
```diff
@@ -56,11 +60,42 @@ def subscriptions_scheduler(
...
+    never buffered and immediately submitted to the next step. In GLOBAL mode we wait
+    (filling the buffer) until the timestamp of a tick has been reached on every
+    partition before eventually submitting a tick to the next step. This guarantees that
+    a subscription is never be scheduled before data on every partition up to that
```
fpacifici, line 86: ... is never scheduled  ...

snuba/cli/subscriptions_scheduler.py
```diff
@@ -56,11 +60,42 @@ def subscriptions_scheduler(
...
+    timestamp is written to storage.
+
+    - The second processing step provides the strategy for committing offsets. Ticks are
+    only marked as `should_commit` if every partition has already reached the timestamp
+    of the tick. Only the offset of the slowest partition will get committed. This
+    guarantees at least once scheduling of subscriptions.
```
fpacifici, line 92: You may want to clarify that it is only the offset "on the commit log" of the slowest partition "on the main topic" that gets committed.

snuba/cli/subscriptions_scheduler.py
```diff
@@ -72,11 +107,36 @@ def subscriptions_scheduler(
...
+
+    scheduled_topic_spec = stream_loader.get_subscription_scheduled_topic_spec()
+    assert scheduled_topic_spec is not None
+
+    producer = KafkaProducer(
+        build_kafka_producer_configuration(
+            scheduled_topic_spec.topic,
+            override_params={
+                "partitioner": "consistent",
+                "message.max.bytes": 50000000,  # 50MB, default is 1MB
```
fpacifici, line 128: I don't think a scheduled subscription can get even remotely close to something 3 order of magnitude smaller. I think you can leave it to the default.

fpacifici (approved):
I think we should have some metrics in place in arroyo to track how often we reject messages and in the sheduler to keep track of the length of the queue of messages being produced.

PR #2237, opened 2021-11-23
[sns-800] support multiple columns in LIMIT BY

onewland:
### Functional Changes
- made internal representation of column[s] in LimitBy into a Sequence of Expression rather than an atomic Expression
- added grammar for optional list of comma-separated columns to follow single required column
- updated output formatters to emit all columns from input 

### Non-Functional Changes
- small fix in parameter order in documents
- updated documentation to reflect multi-column support

### Testing
- verified no regression in unit tests
- added a new E2E test to verify that LIMIT BY was using all columns (and limiting results)

[SNS-800]: https://getsentry.atlassian.net/browse/SNS-800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snuba/query/snql/parser.py
```diff
@@ -617,11 +618,25 @@ def visit_totals_clause(
     def visit_limit_by_clause(
         self,
         node: Node,
-        visited_children: Tuple[Any, Any, Any, Literal, Any, Any, Any, Column],
+        visited_children: Tuple[
+            Any, Any, Any, Literal, Any, Any, Any, Column, Sequence[Column]
```
evanh, line 622: This typing isn't quite right.
```suggestion
            Any, Any, Any, Literal, Any, Any, Any, Column, Optional[Sequence[Column]]
```

tests/query/snql/test_query.py
```diff
@@ -133,11 +162,11 @@ def build_cond(tn: str) -> str:
                 SelectedExpression("c", Column("_snuba_c", None, "c")),
             ],
             condition=required_condition,
-            limitby=LimitBy(5, Column("_snuba_c", None, "c")),
+            limitby=LimitBy(5, [Column("_snuba_c", None, "c")],),
```
evanh, line 165:
```suggestion
            limitby=LimitBy(5, [Column("_snuba_c", None, "c")]),
```

evanh (commented):
Well done, this looks great! Just a couple very small nits.

Also, could you please add a test to one of the `_api.py` tests, so we can test the full pipeline? Since this is going to be used by transactions ultimately, probably `test_transactions_api.py`.

tests/test_transactions_api.py
```diff
@@ -626,3 +630,42 @@ def test_span_id(self) -> None:
...
+        )
+        response = self.app.post(SNQL_ROUTE, data=json.dumps({"query": query_str}))
+        parsed_data = json.loads(response.data)
+
+        assert response.status_code == 200
+        # Note that the following assertions will be wrong if there isn't a sufficient cross-product
+        # of generated test data (meaning each environment needs to have at least
+        # LIMIT_BY_COUNT records present for each platform). The number of environments
+        # had to be made unequal to the number of platforms to get the test data generator
+        # to write suitable input data
```
onewland, line 657: Let me know if we think this is too brittle. I wanted to avoid mucking around with the definition of `generate_fizzbuzz_events`

tests/test_transactions_api.py
```diff
@@ -626,3 +630,42 @@ def test_span_id(self) -> None:
...
+                    WHERE project_id = 1
+                    AND finish_ts >= toDateTime('{start_time}')
+                    AND finish_ts < toDateTime('{end_time}')
+                    LIMIT {limit_by_count} BY environment, platform
+                    """.format(
+            start_time=(self.base_time - self.skew).isoformat(),
+            end_time=(self.base_time + self.skew).isoformat(),
+            limit_by_count=LIMIT_BY_COUNT,
+        )
+        response = self.app.post(SNQL_ROUTE, data=json.dumps({"query": query_str}))
```
evanh, line 649:
```suggestion
        response = self.app.post(SNQL_ROUTE, data=json.dumps({"query": query_str, "dataset": "transactions"}))
```

Obviously this isn't strictly necessary but is best practice.

tests/test_transactions_api.py
evanh: This mostly makes sense, but I'm not sure I understand the last part. Can you elaborate on why adding another environment was necessary?

tests/test_transactions_api.py
```diff
@@ -626,3 +630,42 @@ def test_span_id(self) -> None:
...
+
+        records_by_limit_columns = defaultdict(list)
+
+        for datum in parsed_data["data"]:
+            records_by_limit_columns[(datum["platform"], datum["environment"])].append(
+                datum
+            )
+
+        for key in records_by_limit_columns.keys():
+            assert len(records_by_limit_columns[key]) == LIMIT_BY_COUNT
```
evanh, line 671:
```suggestion
            assert len(records_by_limit_columns[key]) == LIMIT_BY_COUNT, key
```
This way if it fails we'll at least know which key it failed for.

docs/source/language/snql.rst
```diff
@@ -13,7 +13,7 @@ This is the query structure.::
     WHERE condition [[AND | OR] condition]*
     HAVING condition [[AND | OR] condition]*
     ORDER BY expressions ASC|DESC [, expressions ASC|DESC]*
-    LIMIT expression BY n
+    LIMIT n BY [expressions]
```
evanh, line 16: Good catch!

evanh (approved):
Just a couple small nits before you merge. Well done though!

tests/test_transactions_api.py
onewland: If you look at the data generator, both platform and environment are set from an input array that's cycled through via the value of the variable `tock`:

```
"platform": self.platforms[
    (tock * p) % len(self.platforms)
],
```

```
"environment": self.environments[
    (tock * p) % len(self.environments)
]
```

If both of these are the same length, then every environment gets matched to one platform in every record (e.g. "a" is always "prod" and "b" is always "test"). This means that if you try to `LIMIT BY platform, environment` you can't test that individual column values are being used in the LIMIT BY (it's equivalent to `LIMIT BY platform` or `LIMIT BY environment`).

Does that make sense?

evanh: Ah I see, yes that makes sense. Thank you.

PR #2241, opened 2021-11-24
fix(metric): Restore final flag in Queries metrics

rahul-kumar-saini:
## Overview
- `snuba.api.query.count` metric had its `final` tag always set to `false`
- This is probably due to Snuba query `__final` attribute and `get_final()` being hardcoded to `False` and `set_final()` was never used

## Changes
- After a Query is run through the pipeline, the results come with a `stats` dictionary that contains a `final` flag which tells us whether or not the Query will be final when passed to ClickHouse
- Updated the Query classes to appropriately use `set_final()` and `get_final()`, which should fix the metric issue

### DIFF
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -28,7 +28,12 @@ from snuba.util import with_span
 from snuba.utils.metrics.gauge import Gauge
 from snuba.utils.metrics.timer import Timer
 from snuba.utils.metrics.wrapper import MetricsWrapper
-from snuba.web import QueryException, QueryResult, transform_column_names
+from snuba.web import (
+    QueryException,
+    QueryExtraData,
+    QueryResult,
+    transform_column_names,
+)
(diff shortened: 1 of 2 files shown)


snuba/web/query.py
```diff
@@ -119,15 +124,22 @@ def parse_and_run_query(
...
     except QueryException as error:
+        _set_query_final(request, error.extra)
         record_query(request, timer, query_metadata, error.extra)
         raise error
 
     return result
 
 
+def _set_query_final(request: Request, extra: QueryExtraData) -> None:
+    if "final" in extra["stats"]:
```
rahul-kumar-saini, line 139: Not sure if this check is necessary but `QueryExtraData.stats` isn't a `TypedDict` so I'm not sure the `"final"` key always exists.

evanh (approved):

PR #2243, opened 2021-11-25
feat(metric-alerts): Add MetricsCountersSubscription [INGEST-625]

ahmedetefy:
This PR:-
- adds MetricsCountersEntitySubscription class to create crash rate alerts for metrics which accepts a maximum of 3 aggregations and accepts a group by clause 

Group by clause is required here because:
- This is the Snql query that currently gets crash free percentage from metrics
https://github.com/getsentry/sentry/blob/1564d0e3c6a421721d211b3e87fdd786ec7959f3/src/sentry/release_health/metrics.py#L217-L232

And this query returns a json that looks like this 
```
[{‘project_id’: 56, ‘tags[320]’: 321, ‘value’: 2.0}, {‘project_id’: 56, ‘tags[320]’: 328, ‘value’: 1.0}]
```

If we were to remove the group by clause, we would merely get something that looks like this 
```
[{'value': 10.0}]
```
Which doesn't really help us determine the crash free percentage as this is merely the total sum of the sessions 
- Adds validation that ensures that any field in the `groupby` clause has a matching condition (either `IN` or `=`)

### DIFF
--- a/snuba/query/conditions.py
+++ b/snuba/query/conditions.py
@@ -1,12 +1,7 @@
 from typing import Any, Mapping, Optional, Sequence, Set, Union
 
 from snuba.query.dsl import literals_tuple
-from snuba.query.expressions import (
-    Expression,
-    FunctionCall,
-    Literal,
-    OptionalScalarType,
-)
+from snuba.query.expressions import Expression, FunctionCall, Literal
 from snuba.query.matchers import Any as AnyPattern
 from snuba.query.matchers import AnyExpression, AnyOptionalString
 from snuba.query.matchers import Column as ColumnPattern
@@ -304,7 +299,7 @@ def build_match(
     ops: Sequence[str],
     param_type: Any,
     alias: Optional[str] = None,
-    key: OptionalScalarType = None,
+    key: Optional[str] = None,
 ) -> Or[Expression]:
     # The IN condition has to be checked separately since each parameter
     # has to be checked individually.
@@ -312,7 +307,7 @@ def build_match(
     pattern: Union[ColumnPattern, SubscriptableReferencePattern]
     if key is not None:
         pattern = SubscriptableReferencePattern(
-            table_name=alias_match, column_name=String(col), key=AnyPattern(str)
+            table_name=alias_match, column_name=String(col), key=String(key)
         )
     else:
         pattern = ColumnPattern(table_name=alias_match, column_name=String(col))
--- a/snuba/query/conditions.py
+++ b/snuba/query/conditions.py
@@ -1,7 +1,12 @@
-from typing import Any, Mapping, Optional, Sequence, Set
+from typing import Any, Mapping, Optional, Sequence, Set, Union
 
 from snuba.query.dsl import literals_tuple
-from snuba.query.expressions import Expression, FunctionCall, Literal
+from snuba.query.expressions import (
+    Expression,
+    FunctionCall,
+    Literal,
+    OptionalScalarType,
+)
 from snuba.query.matchers import Any as AnyPattern
 from snuba.query.matchers import AnyExpression, AnyOptionalString
 from snuba.query.matchers import Column as ColumnPattern
@@ -9,6 +14,7 @@ from snuba.query.matchers import FunctionCall as FunctionCallPattern
 from snuba.query.matchers import Integer
 from snuba.query.matchers import Literal as LiteralPattern
 from snuba.query.matchers import Or, Param, Pattern, String
+from snuba.query.matchers import SubscriptableReference as SubscriptableReferencePattern
 
 
 class ConditionFunctions:
@@ -294,12 +300,25 @@ def is_condition(exp: Expression) -> bool:
 
 
 def build_match(
-    col: str, ops: Sequence[str], param_type: Any, alias: Optional[str] = None
+    col: str,
+    ops: Sequence[str],
+    param_type: Any,
+    alias: Optional[str] = None,
+    key: OptionalScalarType = None,
 ) -> Or[Expression]:
     # The IN condition has to be checked separately since each parameter
     # has to be checked individually.
     alias_match = AnyOptionalString() if alias is None else String(alias)
-    column_match = Param("column", ColumnPattern(alias_match, String(col)))
+    pattern: Union[ColumnPattern, SubscriptableReferencePattern]
+    if key is not None:
+        pattern = SubscriptableReferencePattern(
+            table_name=alias_match, column_name=String(col), key=AnyPattern(str)
+        )
+    else:
+        pattern = ColumnPattern(table_name=alias_match, column_name=String(col))
+
+    column_match = Param("column", pattern)
+
     return Or(
         [
             FunctionCallPattern(
--- a/snuba/query/matchers.py
+++ b/snuba/query/matchers.py
@@ -357,6 +357,7 @@ class SubscriptableReference(Pattern[SubscriptableReferenceExpr]):
     If column_name and key arguments are provided, they have to match, otherwise they are ignored.
     """
 
+    table_name: Optional[Pattern[Optional[str]]] = None
     column_name: Optional[Pattern[str]] = None
     key: Optional[Pattern[str]] = None
 
@@ -365,6 +366,12 @@ class SubscriptableReference(Pattern[SubscriptableReferenceExpr]):
             return None
 
         result = MatchResult()
+
+        if self.table_name is not None:
+            partial_result = self.table_name.match(node.column.table_name)
+            if partial_result is None:
+                return None
+            result = result.merge(partial_result)
 
         if self.column_name is not None:
             partial_result = self.column_name.match(node.column.column_name)
--- a/snuba/query/validation/validators.py
+++ b/snuba/query/validation/validators.py
@@ -2,7 +2,7 @@ import logging
 from abc import ABC, abstractmethod
 from datetime import datetime
 from enum import Enum
-from typing import Optional, Set
+from typing import Optional, Sequence, Set
 
 from snuba.datasets.entities.entity_data_model import EntityColumnSet
 from snuba.query import Query
@@ -147,7 +147,16 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
     def __init__(self, max_allowed_aggregations: int) -> None:
         self.max_allowed_aggregations = max_allowed_aggregations
 
-    def validate(self, query: Query, alias: Optional[str] = None) -> None:
+    def validate(
+        self,
+        query: Query,
+        alias: Optional[str] = None,
+        disallowed: Optional[Sequence[str]] = None,
+    ) -> None:
+        if disallowed is None:
+            raise InvalidQueryException(
+                "Disallowed clauses list is a required attribute"
+            )
         selected = query.get_selected_columns()
         if len(selected) > self.max_allowed_aggregations:
             aggregation_error_text = (
@@ -159,7 +168,6 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
                 f"A maximum of {aggregation_error_text} allowed in the select"
             )
 
-        disallowed = ["groupby", "having", "orderby"]
         for field in disallowed:
             if getattr(query, f"get_{field}")():
                 raise InvalidQueryException(
--- a/snuba/query/validation/validators.py
+++ b/snuba/query/validation/validators.py
@@ -144,19 +144,13 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
     clauses are being used in the query, and that those clauses are in the correct structure.
     """
 
-    def __init__(self, max_allowed_aggregations: int) -> None:
-        self.max_allowed_aggregations = max_allowed_aggregations
-
-    def validate(
-        self,
-        query: Query,
-        alias: Optional[str] = None,
-        disallowed: Optional[Sequence[str]] = None,
+    def __init__(
+        self, max_allowed_aggregations: int, disallowed_aggregations: Sequence[str]
     ) -> None:
-        if disallowed is None:
-            raise InvalidQueryException(
-                "Disallowed clauses list is a required attribute"
-            )
+        self.max_allowed_aggregations = max_allowed_aggregations
+        self.disallowed_aggregations = disallowed_aggregations
+
+    def validate(self, query: Query, alias: Optional[str] = None,) -> None:
         selected = query.get_selected_columns()
         if len(selected) > self.max_allowed_aggregations:
             aggregation_error_text = (
@@ -168,7 +162,7 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
                 f"A maximum of {aggregation_error_text} allowed in the select"
             )
 
-        for field in disallowed:
+        for field in self.disallowed_aggregations:
             if getattr(query, f"get_{field}")():
                 raise InvalidQueryException(
                     f"invalid clause {field} in subscription query"
--- a/snuba/query/validation/validators.py
+++ b/snuba/query/validation/validators.py
@@ -12,6 +12,8 @@ from snuba.query.conditions import (
     get_first_level_and_conditions,
 )
 from snuba.query.exceptions import InvalidExpressionException, InvalidQueryException
+from snuba.query.expressions import Column
+from snuba.query.expressions import SubscriptableReference as SubscriptableReferenceExpr
 
 logger = logging.getLogger(__name__)
 
@@ -60,7 +62,9 @@ class EntityRequiredColumnValidator(QueryValidator):
         missing = set()
         if self.required_columns:
             for col in self.required_columns:
-                match = build_match(col, [ConditionFunctions.EQ], int, alias)
+                match = build_match(
+                    col=col, ops=[ConditionFunctions.EQ], param_type=int, alias=alias
+                )
                 found = any(match.match(cond) for cond in top_level)
                 if not found:
                     missing.add(col)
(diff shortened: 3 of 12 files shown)


snuba/query/validation/validators.py
```diff
@@ -147,7 +147,16 @@ class SubscriptionAllowedClausesValidator(QueryValidator):
     def __init__(self, max_allowed_aggregations: int) -> None:
         self.max_allowed_aggregations = max_allowed_aggregations
 
-    def validate(self, query: Query, alias: Optional[str] = None) -> None:
+    def validate(
+        self,
+        query: Query,
+        alias: Optional[str] = None,
+        disallowed: Optional[Sequence[str]] = None,
```
fpacifici, line 154: Does this actually type check ?
The validate method comes from the parent class 
`def validate(self, query: Query, alias: Optional[str] = None,) -> None:` 
Adding a parameter violates the parent interface.

If you want to provide a parameter used by the implementation but remaining compatible with the parent class interface, please pass the list of disallowed aggregation to the constructor, store it as an attribute and use it in the validate method.
Also please give the parameter a name that reflects more accurately what this argument is for (like `disallowed_aggregations`).

snuba/subscriptions/entity_subscription.py
```diff
@@ -123,10 +124,16 @@ class TransactionsSubscription(BaseEventsSubscription):
     ...
 
 
+class MetricsCountersSubscription(SessionsSubscription):
+    MAX_ALLOWED_AGGREGATIONS: int = 3
+    disallowed = ["having", "orderby"]
```
fpacifici, line 129: why do you need ato allow the group by ?

fpacifici (changes requested):
Please see my comment inline. 
At a high level I am not sure why you need to allow the group by aggregation. Could you please add more color to the PR description explaining which problem this PR is solving ?

snuba/subscriptions/entity_subscription.py
```diff
@@ -123,10 +124,16 @@ class TransactionsSubscription(BaseEventsSubscription):
     ...
 
 
+class MetricsCountersSubscription(SessionsSubscription):
+    MAX_ALLOWED_AGGREGATIONS: int = 3
+    disallowed_aggregations = ["having", "orderby"]
```
fpacifici, line 129: Resurfacing my original question:

> At a high level I am not sure why you need to allow the group by aggregation. Could you please add more color to the PR description explaining which problem this PR is solving ?

snuba/subscriptions/subscription.py
```diff
@@ -22,12 +22,13 @@ class SubscriptionCreator:
     the resulting query is valid.
     """
 
-    def __init__(self, dataset: Dataset):
+    def __init__(self, dataset: Dataset, entity_key: EntityKey):
         self.dataset = dataset
-        self.entity_key = ENTITY_NAME_LOOKUP[dataset.get_default_entity()]
+        self.entity_key = entity_key
```
fpacifici, line 27: This is going to be a problem and a much larger refactoring.
The entire subscription infra works only with one default entity per dataset. Now you need to create subscriptions on multiple entities on the same dataset. 
The problems you are encountering is that: there is no in code relationship between entity and dataset. We still did not build that. So you may create a subscription for entity `metrics_counter` and dataset `discover` and this will be fine. Except that everything will break when you try to execute the query. If you are lucky that would be noticed by the `_test_request` method here, but you should not rely on that. @evanh  do we have a  task anywhere to store this relationship in code eventually ? I think it will have to be done before this PR.

snuba/web/views.py
```diff
@@ -435,25 +438,39 @@ def handle_subscription_error(exception: InvalidSubscriptionError) -> Response:
...
 @application.route("/<dataset:dataset>/subscriptions", methods=["POST"])
 @util.time_request("subscription")
-def create_subscription(*, dataset: Dataset, timer: Timer) -> RespTuple:
-    entity_key = ENTITY_NAME_LOOKUP[dataset.get_default_entity()]
+def create_subscription(
+    *, dataset: Dataset, timer: Timer, entity: Optional[Entity] = None
+) -> RespTuple:
+    if not entity:
+        entity = dataset.get_default_entity()
+    entity_key = ENTITY_NAME_LOOKUP[entity]
```
fpacifici, line 449: This surfaces the problem I mentioned above. You can pass an entity and a dataset that do not correspond to each other.

fpacifici (commented):
I think there are a couple of fundamental issues in the subscription data model that would prevent you from moving ahead without doing some larger refactoring first. 
Specifically connecting an entity to its own dataset and managing multiple subscriptions storages in the same subscriptions consumer. 
Subscriptions were never designed with multiple entities in mind, if we need to support it now we will have to do some additional refactoring before moving ahead. I added comments for Lyn and Evan to provide input on.

snuba/subscriptions/subscription.py
lynnagara: I think at a minimum you could just validate that the entity passed corresponds to one that is in `dataset.get_all_entities()`

snuba/cli/subscriptions.py
```diff
@@ -43,6 +48,9 @@
     type=click.Choice(DATASET_NAMES),
     help="The dataset to target",
 )
+@click.option(
+    "--entity", "entity_name", help="The entity to target",
+)
```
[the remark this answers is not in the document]
lynnagara: I think at least the executor should be simple enough that a shared executor pool for multiple entities in a dataset should be reasonably straightforward

snuba/subscriptions/entity_subscription.py
ahmedetefy: @fpacifici I updated the PR description already to explain why we need the group by column. Did you miss it or is it still not clear and I should I try to make it more expressive?

snuba/subscriptions/subscription.py
ahmedetefy: fwiw I did create that relationship between dataset and entity on the sentry side so technically its not possible atm to `discover` dataset and `metrics_counters` entity.
Here i a link to that: https://github.com/getsentry/sentry/pull/30400/files#diff-b3812aba29e6dd5fb20c8e98e849b988faf38564c8f3dcd5526bab213c8f7853R238-R274

However, I agree with @lynnagara I should probably also validate this using `dataset.get_all_entities()` so I added that change

snuba/cli/subscriptions.py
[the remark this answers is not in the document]
ahmedetefy: @fpacifici 

>Does this mean we will have to run one subscriptions consumer cluster per metrics entity ?

I thought that was already the case given that the default entity was always extracted as per https://github.com/getsentry/snuba/pull/2243/files#diff-11204424e2e0a0540fa9eadb5dbaf07391016ab4d94ad0818abdfb2e9a5bfaebL132
I just gave the choice to use a specific entity

@lynnagara 
> since metrics never even writes a commit log now

There is a PR open for that https://github.com/getsentry/snuba/pull/2222/files#diff-93e400870e85f66f3db6afceb3d7ea01c4cf66788c389e1c5a50eca0eb0b78c9R78-R84

snuba/web/views.py
ahmedetefy: This should be resolved.. See above comment

fpacifici:
Thanks for the context around the group by. A few high level questions about the PR description, specifically about allowing the group by:
Is there a tech spec or list of requirements we want to support for alerts on metrics? The changes this PR makes are going to have to be supported for all alerts on metrics as the metrics dataset is not specific for release health. So every change should be thought that way and not with the goal of supporting a specific use case (crash free alerts in this case).

The issue with allowing the group by is that we are now changing the response of a subscription query from a single time series to an unspecified and variable number of time series (as the group by can return any number of results for each query). The second scenario (unspecified amount of time series returned), is a lot more complex conceptually and drops several design assumptions that we will have to support for all metrics. Specifically: 
- a query could go out of memory, and cause an outage on alerts. Subscription queries must be extremely deterministic in cpu and memory performance. 
- Data points for the consumer in sentry may be missing if a given group is not present for a time bucket. 
- This locks us into using a SQL style DB to run subscriptions queries (that supports group by). We cannot replace the alerts storage with one that needs us to know the specific time series the query will return before running the query. Before taking this direction, we need to think this thoroughly, thus the need of a tech spec/list of requirements.

ahmedetefy:
@fpacifici 

 >Is there a tech spec or list of requirements we want to support for alerts on metrics?
 
 That's not clearly defined yet.. However, I believe the primary goal was to port crash rate alerts to metrics as an extension of release health porting. It's a good point however, I will start for a doc for that and I will bring it up with the team

> The changes this PR makes are going to have to be supported for all alerts on metrics as the metrics dataset is not specific for release health

I disagree here. Firstly, These changes are only applied to the `metrics_counters` entity atm, and I believe grouping by tag and project will not be exclusively a release health subscription requirement, given how the data model i.e. specifically tags.

As an example, here is a sample of all the release health queries called from sentry and they have require a group by clause 
https://github.com/getsentry/sentry/blob/6f0870e63da17050508a2b66b76a3885c745954f/src/sentry/release_health/metrics.py

>The issue with allowing the group by is that we are now changing the response of a subscription query from a single time series to an unspecified and variable number of time series (as the group by can return any number of results for each query). The second scenario (unspecified amount of time series returned), is a lot more complex conceptually and drops several design assumptions that we will have to support for all metrics. Specifically:
>- a query could go out of memory, and cause an outage on alerts. Subscription queries must be extremely deterministic in cpu and memory performance.
>- Data points for the consumer in sentry may be missing if a given group is not present for a time bucket.
>- This locks us into using a SQL style DB to run subscriptions queries (that supports group by). We cannot replace the alerts storage with one that needs us to know the specific time series the query will return before running the query. Before taking this direction, we need to think this thoroughly, thus the need of a tech spec/list of requirements.

I agree with this, and I will kickstart the tech spec and share with SNS. 
However, I do not feel like this is a blocker for crash rate alerts over metrics because currently metrics crash rate alerts require grouping by the `session.status` tag which will always return a maximum of 4 buckets and project_id, and a subscription is per a single project_id (I just realised that the response snippet I added in the PR description contains time series of two different projects but that is inaccurate because subscription queries always filter on a project_id), so there will always be a maximum of 4 buckets.
Knowing this, do you reckon we can move forward with crash rate alerts, while realising that we need to explore solutions for future metrics alerts that we want to support?

fpacifici:
> However, I do not feel like this is a blocker for crash rate alerts over metrics because currently metrics crash rate alerts require grouping by the session.status tag which will always return a maximum of 4 buckets (I just realised that the response snippet I added in the PR description contains time series of two different projects but that is inaccurate because subscription queries always filter on a project_id so I updated the response snippet).

From what I hear from @jan-auer I don't think I can agree with this statement.
Per conversations I had with Jan, this is not a release health cluster or a release health feature. This is a cluster and data model that will support custom metrics as well, where we have a lot less control on the metrics and dimensionality and queries we will receive. 
If that is the case, the fact that in this specific scenario we have only 4 buckets does not mean another use case will.

Now we have three options here:
- isolate the release health feature. Accept that what we are building is release health specific, then we can introduce product specific features and move on
- aim at building a generic metrics platform that will be shared by multiple use cases, which is my understanding, then we need to think through carefully before allowing this use case.
- Allow multiple use cases on the same deployment/cluster but partition them so that each use case can have different restrictions. 

But we cannot build the second and fast track specific product features that have looser constraints than others.

@jan-auer how would the product want to proceed ?

fpacifici:
There is also a fourth option. 
Enable groupby but validate that the column in the group by also appears in a `=` or `IN` condition so to cap the number of groups to what the client knows. Though this should be the outcome of the technical spec discussion.

snuba/cli/subscriptions.py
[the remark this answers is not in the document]
ahmedetefy: Moving this to https://github.com/getsentry/snuba/pull/2258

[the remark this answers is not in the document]
fpacifici: > What I don't understand is how can all metrics share the same steam loader when they have different processors?

What matters for subscriptions consumer is whether they share the same commit log, which they do as they are in the same multi storage consumer. Two consumers that share the same topic and commit log can, in principle, share the subscriptions consumer. The problem is the subscriptions storage which is partitioned by entity, but that is an implementation issue not a conceptual issue.

> Why don't we want to allow one subscriptions consumer cluster per metrics entity?
Because it means allocating, maintaining and troubleshooting  4 clusters instead of one, while 1 will be able to deal with this for a long time. 

> "load multiple storages for multiple entities in the same dataset" Not sure I understand what this means
The subscriptions storage is partitioned by entity. The subscription consumer can only work with one at a time now. This is an implementation detail and not a conceptual issue. So we would have to make that change to run on one cluster. 

Now @lynnagara made the change to allow this to work in the new architecture (multiple schedulers but one executor cluster). So the situation will be better there. 
Do you need alerts on all types of metrics as of now? We can deploy only the cluster for the entities we need if you are not supporting alerts on all metrics types yet.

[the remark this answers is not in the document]
lynnagara: I made a fairly small change `SubscriptionWithTick` so that it also stores the entity name (as that will no longer be clear when multiple entities are being passed in via the CLI. You may need to rebase this PR to get those tests and mypy passing here.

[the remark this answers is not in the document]
ahmedetefy: @lynnagara 
Sounds good thanks.
After reading this https://github.com/getsentry/snuba/pull/2255, I am assuming this is no longer required -> https://github.com/getsentry/snuba/pull/2222/ because its the old infra, and instead I should put up a new PR where the subscription scheduler supports the new metrics entities?

@fpacifici 
Thanks for clarification.

> Do you need alerts on all types of metrics as of now? We can deploy only the cluster for the entities we need if you are not supporting alerts on all metrics types yet.

Yes that's fine for now. For Crash Rate alerts, we only need to support `metrics_counters` (for sessions crash free rate) and `metrics_sets` (for users crash free rate). That being said, like I mentioned in our slack thread I am meeting with product early next week and I will kick start a tech spec for future metric alerts to share with you then

ahmedetefy:
Moving this out of a slack convo with @fpacifici 

> There is also a fourth option.
Enable groupby but validate that the column in the group by also appears in a = or IN condition so to cap the number of groups to what the client knows. Though this should be the outcome of the technical spec discussion.

We have decided to add this validation

evanh:
@ahmedetefy I know I'm coming to this late, but is it possible to fetch the percentage by returning two results in the same query? Say you were trying to find the percentage of failed sessions, you could write a query along the lines of `SELECT countIf(tags[320] = failed) / count()...` and then do the division that way? I'm not clear on why the GROUP BY is necessary in the subscription query.

fpacifici:
If you decide to return two results in the same query without GROUP BY you will still have to valdiate that the condition in the countIf is also present as a top level condition of the query. I don't think the countIf would use the primary key or the bloom filter index as all aggregation functions are applied after the filtering of the blocks has been done. 
Though this would have to be confirmed on Clickhouse prod in our version.

In principle if you run 
`count() WHERE tag[123] = 234` 
First clickhosue would use the bloom filter index on the tags to exclude the blocks that do not contains `tags[123] = 234` then it will count the rows on the resulting blocks
The countIf instead is likely to not filter blocks, scann all the rows and check the condition on each one, making the query a lot less efficient which is a problem for subscriptions where queries need to be lightening fast and predictable

evanh:
> you will still have to valdiate that the condition in the countIf is also present as a top level condition of the query.

OK that makes sense. I guess it's equal work either way.

tests/datasets/validation/test_subscription_clauses_validator.py
```diff
@@ -29,12 +45,63 @@
...
+                        Literal(None, 2),
+                    ),
+                ),
+            ),
+            groupby=[
+                Column("_snuba_project_id", None, "project_id"),
+                Column("_snuba_tags[3]", None, "tags[3]"),
+            ],
+        ),
+        id="groupby is allowed in metrics counters subscriptions",
```
fpacifici, line 89: Please test the case where you have a `tags[asd] IN (1,2,3)` case

fpacifici (changes requested):
see comments in line

snuba/query/matchers.py
```diff
@@ -357,6 +357,7 @@ class SubscriptableReference(Pattern[SubscriptableReferenceExpr]):
     If column_name and key arguments are provided, they have to match, otherwise they are ignored.
     """
 
+    table_name: Optional[Pattern[Optional[str]]] = None
```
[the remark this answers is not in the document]
ahmedetefy: fwiw, I don't have much context on what's going on here 
but I mirrored what the `ColumnPattern` expects which is a `table_name` and a `column_name`
so I don't understand how I could pass an `alias` for the table name if I don't add this field here ->
https://github.com/getsentry/snuba/pull/2243/files/26ac38ccc3c9da76d5af19c049b69857cc013f67#diff-637ef9b4fedb7d04975dee35c884be9c46741ae38423cb07ed7261c85a2ddd05R315

snuba/query/conditions.py
```diff
@@ -294,12 +300,25 @@ def is_condition(exp: Expression) -> bool:
...
 ) -> Or[Expression]:
     # The IN condition has to be checked separately since each parameter
     # has to be checked individually.
     alias_match = AnyOptionalString() if alias is None else String(alias)
-    column_match = Param("column", ColumnPattern(alias_match, String(col)))
+    pattern: Union[ColumnPattern, SubscriptableReferencePattern]
+    if key is not None:
+        pattern = SubscriptableReferencePattern(
+            table_name=alias_match, column_name=String(col), key=AnyPattern(str)
+        )
```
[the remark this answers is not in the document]
ahmedetefy: Oh I see why that is problematic.. will fix

snuba/query/matchers.py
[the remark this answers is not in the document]
ahmedetefy: Is it better if I change `SubscriptibleReferencePattern` to accept a column rather than a column name, and then just pass the alias to the column?

snuba/query/conditions.py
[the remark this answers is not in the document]
ahmedetefy: This should be fixed

tests/datasets/validation/test_subscription_clauses_validator.py
ahmedetefy: This as well should be resolved

fpacifici (approved):
Please ensure somebody from Search and Storage is present when deploying as this is a large change.
Preferably somebody with context on metrics or on crash free alerts.

fpacifici:
please fix the type errors before merging

PR #2244, opened 2021-11-26
ref(subscriptions): Decouple subscription logic from datasets [INGEST-627]

ahmedetefy:
Decouples subscription creation and deletion logic from datasets to also consider entities since dataset to entity is not a 1-to-1 relation

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -2182,7 +2182,7 @@ class TestCreateSubscriptionApi(BaseApiTest):
         with patch("snuba.subscriptions.subscription.uuid1") as uuid4:
             uuid4.return_value = expected_uuid
             resp = self.app.post(
-                "{}/subscriptions".format(self.dataset_name),
+                url,
                 data=json.dumps(
                     {
                         "project_id": 1,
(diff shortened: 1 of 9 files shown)


tests/test_api.py
```diff
@@ -2169,7 +2169,14 @@ def test_error_handler(self, pipeline_mock: MagicMock) -> None:
...
 
-    def test(self) -> None:
+    @pytest.mark.parametrize(
+        "url",
+        [
+            "events/subscriptions",  # Only dataset in url
+            "events/events/subscriptions",  # dataset and entity in url
+        ],
+    )
+    def test(self, url: str) -> None:
```
evanh, line 2179: Where is `url` used in this test?

evanh (commented):
Can you add a test that ensures that the entity being specified is the one being used, and not just the default for the dataset? So for example use dataset `discover` but entity `transactions` or something like that.

tests/test_api.py
ahmedetefy: oops .. I missed that

ahmedetefy:
@evanh I have addressed the PR comments.. Let me know if there is anything else :)

evanh (approved):
LGTM. In order to use the `--entity` option in prod you'll need to make an ops PR as well.

ahmedetefy:
> LGTM. In order to use the `--entity` option in prod you'll need to make an ops PR as well.

Thanks! Can you tell me where to edit that?

PR #2245, opened 2021-11-29
feat(consumer-util): Expose metric of consumer utilization

nikhars:
librdkafka exposes the metric called replyq which tells the number of operations which are awaiting in the queue for application to poll. Documentation of the metric can be found here https://github.com/edenhill/librdkafka/blob/master/STATISTICS.md#top-level

The metric is a gauge so it gives us an indication of point in time size of the queue. Exposing this metric would allow us to know if the queue is ever empty and if yes, we could start fine tuning some parameters.

The current PR just adds the option to be able to capture the metrics. To enable the metrics, the CLI option would need to be added in the K8s configuration.

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -2,6 +2,7 @@ import signal
 from typing import Any, Optional, Sequence
 
 import click
+import rapidjson
 from arroyo import configure_metrics
 
 from snuba import environment, settings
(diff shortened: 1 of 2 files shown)


fpacifici:
From the librdkafka config 

> replyq = Number of ops (callbacks, events, etc) waiting in queue for application to serve with rd_kafka_poll()

I don;t think that is what we want. That is the number of callback in queue, not the size of the consume buffer.

fpacifici:
I think these are the metrics we care for 
```

fetchq_cnt | int gauge |   | Number of pre-fetched messages in fetch queue
-- | -- | -- | --
fetchq_size | int gauge |   | Bytes in fetchq
fetch_state | string | "active" | Consumer fetch state for this partition (none, stopping, stopped, offset-query, offset-wait, active).

```

PR #2246, opened 2021-11-30
ref(subscriptions): Deleting Legacy and Delegate Subscription logic

ahmedetefy:
Hacks out logic that has to do with legacy subscription and delegate subscription data classes since these have been removed

Follow up on #2169


evanh (approved):
I completely missed this, nice catch.

PR #2247, opened 2021-11-30
ref(sentry-metrics): Rename metrics consumer group for dev

MeredithAnya:
Updated the snuba metrics consumer in https://github.com/getsentry/ops/pull/4171, so doing the same here


evanh (approved):

MeredithAnya:
This got changed in https://github.com/getsentry/snuba/pull/2588

PR #2249, opened 2021-11-30
Update k8s deployments containing `metrics-consumer` containers

oioki:
Make Freight deploy metrics consumers too.


.freight.yml
```diff
@@ -44,6 +44,8 @@ steps:
     name: querylog-consumer
   - image: us.gcr.io/sentryio/snuba:{sha}
     name: snuba-admin
+  - image: us.gcr.io/sentryio/snuba:{sha}
+    name: metrics-consumer
```
MeredithAnya, line 48: the `metrics-consumer` corresponds to the yaml file name right? aka `k8s/services/snuba/_deployment.metrics-consumer.yaml `

oioki: I believe it corresponds to the container name in k8s deployment:
https://github.com/getsentry/freight/blob/master/freight/providers/pipeline.py#L337-L340
which is `metrics-consumer` in our case.

MeredithAnya (approved):

fpacifici (approved):
weird that meredith's acceptance was not enough

PR #2250, opened 2021-11-30
feat(suspect-spans): Drop old exclusive time column

Zylphrex:
Will be using the Float32 column moving forwards, so no need to keep the Float64
column. Also map the original name to the Float32 column for compatibility.

# Forwards

```sql
Local operations:
ALTER TABLE transactions_local DROP COLUMN IF EXISTS spans.exclusive_time;


Dist operations:
ALTER TABLE transactions_dist DROP COLUMN IF EXISTS spans.exclusive_time;
```

# Backwards

```sql
Local operations:
ALTER TABLE transactions_local ADD COLUMN IF NOT EXISTS spans.exclusive_time Array(Float64) AFTER spans.group;
ALTER TABLE transactions_local MODIFY COLUMN spans.exclusive_time Array(Float64) TTL finish_ts + INTERVAL 1 MONTH;


Dist operations:
ALTER TABLE transactions_dist ADD COLUMN IF NOT EXISTS spans.exclusive_time Array(Float64) AFTER spans.group;
```


evanh (approved):
Can you please add the SQL to the PR description?

You know this already but let's sync before merging/running this.

lynnagara:
This hasn't moved for some time, please reopen if still needed

PR #2251, opened 2021-12-01
feat(query): multi column array join

onewland:
https://getsentry.atlassian.net/browse/SNS-799

Functional changes:
- Add support for multiple columns in ARRAY JOIN in SnQL that basically just pass through the expressions to Clickhouse SQL

Non-Functional changes:
- because it was depending on a conflicting version of snuba-sdk for this model change and we considered the use of snuba-sdk there a hack, I rewrote test logic from `tests/query/parser/test_query.py` from parsing queries from a (snuba-sdk specific) JSON representation to the actual SnQL query that the JSON was serialized as.
- additionally, I added a single test case in `tests/query/parser/test_query.py` for multi-column array join

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -738,14 +738,24 @@ class SnQLVisitor(NodeVisitor):  # type: ignore
     def visit_arrayjoin_clause(
         self,
         node: Node,
-        visited_children: Tuple[
-            Any, Any, Any, Expression, Optional[Sequence[Expression]]
-        ],
+        visited_children: Tuple[Any, Any, Any, Expression, Optional[List[Expression]]],
     ) -> Sequence[Expression]:
         _, _, _, join_first, join_rest = visited_children
         exprs = [join_first]
+
         if join_rest is not None:
             exprs.extend(join_rest)
+
+        return exprs
+
+    def visit_arrayjoin_optional(
+        self, node: Node, visited_children: List[Tuple[Any, Any, Any, Expression]],
+    ) -> List[Expression]:
+        exprs: List[Expression] = list()
+        if visited_children is not None:
+            for child in visited_children:
+                _, _, _, exp = child
+                exprs.append(exp)
         return exprs
 
     def visit_parameter(
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -106,7 +106,7 @@ snql_grammar = Grammar(
     match_clause          = space* "MATCH" space+ (relationships / subquery / entity_single )
     select_clause         = space+ "SELECT" space+ select_list
     group_by_clause       = space+ "BY" space+ group_list
-    arrayjoin_clause      = space+ "ARRAY JOIN" space+ (tag_column / subscriptable / simple_term)
+    arrayjoin_clause      = space+ "ARRAY JOIN" space+ arrayjoin_entity arrayjoin_optional
     where_clause          = space+ "WHERE" space+ or_expression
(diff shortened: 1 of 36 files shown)


tests/query/parser/test_query.py
```diff
@@ -64,13 +63,30 @@ def with_required(condition: Optional[Expression] = None) -> Expression:
     return required
 
 
+DEFAULT_TEST_QUERY_CONDITIONS = [
+    "timestamp >= toDateTime('2021-01-01T00:00:00')",
```
onewland, line 67: We could extract out more aspects of the datetime and use the constants in the assertions above in `with_required`. But I don't know if we get much benefit from that

snuba/query/snql/parser.py
```diff
@@ -1011,18 +1031,24 @@ def _unpack_array_conditions(
...
     entity_alias = f"{entity_alias}." if entity_alias is not None else ""
     for column in schema:
         if isinstance(column.type, Array):
             aliased_base_name = f"{entity_alias}{column.base_name}"
             aliased_flattened = f"{entity_alias}{column.flattened}"
-            if aliased_base_name == array_join or aliased_flattened == array_join:
+            if (
+                aliased_base_name in array_join_columns
+                or aliased_flattened in array_join_columns
+            ):
```
onewland, line 1051: I am not super confident about my logic changes here so it would be good if somebody could think through potential implications

evanh: This should work fine. Unpacking array conditions is syntactic sugar we added to make writing queries on the product side easier. If they provide a condition on an array column we will automatically convert the condition to be correct for an array column.

For example, a condition like this can be specified in the query: `tags.key = 'foo'`
In this case, the type of the column `tags.key` is an Array(String). If we passed this condition directly to Clickhouse, it would complain that you can't compare a String to an Array(String). So we automatically convert it to something of the form `arrayExists(x -> assumeNotNull(equals(x, 'foo')), tags.key)`, which is equivalent to `'foo' in tags.key`.

However, if the query is also array joining on `tags.key`, then the array column is already being unpacked, so we can pass the condition `tags.key = 'foo'` directly to Clickhouse.

tests/query/parser/test_query.py
```diff
@@ -64,13 +63,30 @@ def with_required(condition: Optional[Expression] = None) -> Expression:
...
+    "timestamp < toDateTime('2021-01-02T00:00:00')",
+    "project_id = 1",
+]
+
+# If these are the only WHERE conditions we can do the simplest thing
+PRECOMPUTED_TEST_QUERY_CONDITIONS_STR = " AND ".join(DEFAULT_TEST_QUERY_CONDITIONS)
+
+
+def prepend_condition_str_to_default(condition: str) -> str:
```
evanh, line 76: This is a style nit, but I would change this function to take an arbitrary number of conditions, and then for the base case you could call it with no arguments.

```
def prepend_condition_str_to_default(*conditions: str) -> str:
    return " AND ".join(list(conditions) + DEFAULT_TEST_QUERY_CONDITIONS)
```

Then you don't need `PRECOMPUTED_TEST_QUERY_CONDITIONS_STR` and every test can use `prepend_condition_str_to_default`

You don't have to change this if you don't want, it feels like it simplifies things a bit.

evanh (approved):
This looks good to me.

tests/query/parser/test_query.py
onewland: made this change, also changed the name of the method to `snql_conditions_with_default`

PR #2252, opened 2021-12-02
feat(subscriptions): Also produce tick to the scheduled subscriptions topic

lynnagara:
We need to also produce the tick along with the scheduled subscription data
to the scheduled subscriptions topic. This is required by the executor
in order to actually build the request and execute the query. It cannot be
omitted as many subscriptions will later modify the query to add an offset
condition during execution. We are likely to also use it to run comparisons
between the new and existing pipeline during roll out.

An alternative would be to have the scheduler build the entire request
and produce it and have the executor only run it, but I did not go down this
route as it:
- puts more work on the scheduler (which is harder to scale than executors)
- added complexity of serializing and deserializing the entire request object
(in comparison scheduled task + tick are more simple structures)

Since we don't have an executor that is decoding any of the messages yet,
this change should be able to safely merge and deploy.


snuba/subscriptions/scheduler.py
```diff
@@ -318,3 +385,24 @@ def find(
...
+        ):
+            for subscription in subscriptions:
+                task = self.__builder.get_task_v2((subscription, tick), timestamp)
+                if task is not None:
+                    yield task
+
+        metrics = self.__builder.reset_metrics()
+        if any(metric for metric in metrics if metric[1] > 0):
+            for metric in metrics:
+                self.__metrics.increment(metric[0], metric[1], tags=metric[2])
```
fpacifici, line 408: Looking at how the `find` method is executed, it seems to me you could just replace the old `find` method with this one and pass the Tick in all cases. Where the `find` method is called you have both the interval and the Tick so you should be fine.
That would remove all this annoying code duplication. Basically all find methods are fully duplicated.

lynnagara: Updated to remove the code duplication and use the new methods everywhere. Previously the `Tick` in the old pipeline was different to the new one, I had to make them the same - there are quite a few changes to the existing subscriptions consumer to make this work.

snuba/subscriptions/data.py
```diff
@@ -223,3 +233,40 @@ def to_dict(self) -> Mapping[str, Any]:
...
+
+    # The task that should be executed.
+    task: SubscriptionWithTick
+
+
+class SubscriptionScheduler(ABC):
+    """
+    The scheduler maintains the scheduling state for subscription tasks and
+    provides the ability to query the schedule to find tasks that were
+    scheduled between the time interval of a tick.
```
fpacifici, line 261: why `to find tasks that were scheduled between the time interval of a tick` ?
Should this say `find tasks that should be scheduled between ....` 
If I am getting this right you are using this to schedule subscription, not to fetch what was already scheduled

snuba/subscriptions/scheduler.py
```diff
@@ -35,16 +40,16 @@ class TaskBuilder(ABC, Generic[TSubscription]):
 
     @abstractmethod
     def get_task(
-        self, subscription: Subscription, timestamp: int
-    ) -> Optional[ScheduledTask[TSubscription]]:
+        self, subscription_with_tick: SubscriptionWithTick, timestamp: int
+    ) -> Optional[ScheduledSubscriptionTask]:
```
fpacifici, line 44: Unrelated. I wonder if we should move the subscription scheduling logic into arroyo.

fpacifici (approved):

snuba/subscriptions/data.py
lynnagara: Yea it was sloppy wording, will update

snuba/subscriptions/scheduler.py
lynnagara: Curious why you say that. Do you have an actual use case in mind where scheduling would be useful outside Snuba?

fpacifici: I think the idea of having a consumer that follows the watermark produced by another and then "does something" can be reused. Same for the synchronized consumer which could be used for the post processing as well.

PR #2256, opened 2021-12-06
feat(perf) Optimize away redundant tag conditions

vladkluev:
### Context
We occasionally fail to apply the tags hashmap optimization because there is a redundant clause in there. 

The problem boils down to this:
```
        valueOf('my_tag') != '' AND valueOf('my_tag') == "something"
                          ^                            ^
                          |                            |
                      existence check               value check
```

The tags hash map is stored with elements `k=v` ([here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/tags_hash_map.py#L7-L11)) meaning that checking `exists("my_tag")` is not possible with the hashmap as it exists today. 

The mapping optimizer will not apply the hashmap optimization if the `tags` column has to be loaded anyway. This is because the cost of scanning an extra column in clickhouse is higher than the speedup gained from using the optimization

### Solution

Within the mapping optimizer, write a function that removes the redundant conditions first, then apply the mapping optimzer as before

### Test plan

* Unit test coverage of new optimizer functionality
* end to end query pipeline test on problematic snql query to see that the optimization is applied
* Experiment configured (to be verified in discover)


snuba/query/processors/mapping_optimizer.py
```diff
@@ -159,20 +185,110 @@ def __replace_with_hash(self, condition: Expression) -> Expression:
...
+                                )
+                                continue
+                            else:
+                                query.add_experiment(
+                                    "tags_redundant_optimizer_enabled", 0
+                                )
+                        else:
+                            # when in doubt, apply the optimization
+                            query.add_experiment("tags_redundant_optimizer_enabled", 1)
+                            continue
```
fpacifici, line 263: This means the configuration is broken. 
Let's not fix it silently, either log an error or just raise an exception.

snuba/query/processors/mapping_optimizer.py
```diff
@@ -159,20 +185,110 @@ def __replace_with_hash(self, condition: Expression) -> Expression:
             ),
         )
 
+    def _get_condition_without_redundant_checks(
+        self, condition: Expression, query: Query
+    ) -> Expression:
+        """Optimizes the case where the query condition contains the following:
+
+        valueOf('my_tag') != '' AND valueOf('my_tag') == "something"
```
fpacifici, line 193: I think this function is missing one use case that we actually instructed the product team to prefer:
`has(tags.key, 'blabla')` and `NOT has(tags.key, 'blabla')`.
We are supposed to see more of these and fewer `valeOf(tags. values, indexOf(tags.key, 'blabla') != ''`.
So we should support both cases now.

snuba/query/processors/mapping_optimizer.py
```diff
@@ -159,20 +185,110 @@ def __replace_with_hash(self, condition: Expression) -> Expression:
...
+            return combine_or_conditions(pruned_conditions)
+        elif condition.function_name == BooleanFunctions.AND:
+            sub_conditions = get_first_level_and_conditions(condition)
+            tag_eq_match_strings = set()
+            matched_tag_exists_conditions = {}
+            for condition_id, cond in enumerate(sub_conditions):
+                tag_exist_match = _tag_exists_pattern.match(cond)
+                if tag_exist_match:
+                    matched_tag_exists_conditions[condition_id] = tag_exist_match
+                eq_match = self.__optimizable_pattern.match(cond)
```
fpacifici, line 237: put this into an else? if `tag_exists_match` is true this cannot be true so you can skip the check. This may have an impact as it happens really often.

snuba/query/processors/mapping_optimizer.py
```diff
@@ -159,20 +185,110 @@ def __replace_with_hash(self, condition: Expression) -> Expression:
...
+                if eq_match:
+                    tag_eq_match_strings.add(eq_match.string("key"))
```
fpacifici, line 239: nit: when manipulating the result of a `mapping_expression`, the names of the parameters are defined in some constants https://github.com/getsentry/snuba/blob/ee7c41a6729508efa05a8f2f2d2a18414e858bdb/snuba/clickhouse/translators/snuba/mappers.py#L237-L256. You can reuse them

fpacifici (commented):
Please see my comment on the alternative way to formulate the tags existence condition `has(tags.key, 'asd')`

snuba/query/processors/mapping_optimizer.py
```diff
@@ -159,20 +185,110 @@ def __replace_with_hash(self, condition: Expression) -> Expression:
...
+            for condition_id, cond in enumerate(sub_conditions):
+                tag_exist_match = matched_tag_exists_conditions.get(condition_id, None)
+                if tag_exist_match:
+                    requested_tag = tag_exist_match.string("key")
+                    if requested_tag in tag_eq_match_strings:
+                        # how often we should skip this optimization
+                        skip_rate = state.get_config(
+                            "tags_redundant_optimizer_skip_rate", 0
+                        )
+                        if isinstance(skip_rate, float) or isinstance(skip_rate, int):
```
evanh, line 250: This is going to suffer from a mistake I made: you want to optimize the entire query or none of it. In this case, if someone has a check on two different tag keys like `(has(tags, A) AND tags[A] = 1) AND (has(tags, B) AND tags[B] = 1)` you might optimize the condition on A but not the one on B and end up with a query that is half optimized. This should also apply across the WHERE and HAVING clause so that either both are optimized or neither are.

snuba/query/processors/mapping_optimizer.py
evanh: There is also a different processor (EmptyTagConditionProcessor) that will convert `valueOf(tags. values, indexOf(tags.key, 'blabla') != ''` into `has(tags.key, 'blabla')`. Right now that processor happens after this one, but better to be safe than sorry.

snuba/query/processors/mapping_optimizer.py
```diff
@@ -83,6 +97,33 @@ def __init__(self, column_name: str, hash_map_name: str, killswitch: str) -> Non
                 Param("right_hand_side", Literal(Any(str))),
             ),
         )
+        self.__tag_exists_patterns = [
+            FunctionCall(
```
vladkluev, line 101: Note: it may be a better idea to run the EmptyTagProcessor as part of this processor so we only have to worry about/match one case. Thoughts?

evanh: I have mixed feelings about it. On the one hand it makes sense because these processors are working on similar things, but on the other hand I feel like we should be striving to keep processors as simple as possible for readability and debugging.

snuba/query/processors/mapping_optimizer.py
```diff
@@ -159,20 +200,111 @@ def __replace_with_hash(self, condition: Expression) -> Expression:
...
+                ├──┬┘         └┬──┤
+                │  │           │  │
+             ┌──┴┐ c           a ┌┴──┐
+             │AND│    becomes    │AND│
+             └┬─┬┘               ├──┬┘
+              │ │                │  │
+              a b                b  c
+        """
+        # how often we should skip this optimization
+        skip_rate = state.get_config("tags_redundant_optimizer_skip_rate", 0)
```
evanh, line 236: This function is recursive, so this might get called multiple times. I would suggest fetching this value in the `process_query` function and passing it in to this function.

vladkluev: good point. Actually even better we this whole function can be skipped outside of itself

snuba/query/processors/mapping_optimizer.py
```diff
@@ -59,6 +67,12 @@ class MappingOptimizer(QueryProcessor):
     - `ifNull('tags.value[indexOf(tags.key, 'my_tag')]', '') = ''`
        this condition is equivalent to looking whether a tag is
        missing, which cannot be done with the hash map.
+        - There is a special case of this where the query will have
+            (
+                ifNull('tags.value[indexOf(tags.key, 'my_tag')]', '') != '' AND
```
onewland, line 72: Just curious: why are we unable to apply this optimization if `ifNull` is absent?

onewland:
It seems like we're recording the percentage of requests that enter into the new optimization flow, but are we recording the number of times the optimization is applied and the query to Clickhouse is simplified from the user's input in a metric somewhere?

onewland (commented):
Seems reasonable to me, though I'm not super confident about its correctness.

snuba/query/processors/mapping_optimizer.py
vladkluev: I've updated the comment. Let me know if there's any confusion left about how the optimization works.

vladkluev:
@onewland thanks for the feedback, there is now a metric being output for when the query is being reduced

PR #2258, opened 2021-12-07
ref(subscriptions): Decouple subscription logic from datasets [INGEST-627]

ahmedetefy:
Replica of #2244 
Re-opening because I accidentally merged it into the main branch

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -469,7 +469,12 @@ def handle_subscription_error(exception: InvalidSubscriptionError) -> Response:
 def create_subscription(
     *, dataset: Dataset, timer: Timer, entity: Optional[Entity] = None
 ) -> RespTuple:
-    if not entity:
+    if entity:
+        if entity not in dataset.get_all_entities():
+            raise InvalidSubscriptionError(
+                "Invalid subscription dataset and entity combination"
+            )
+    else:
         entity = dataset.get_default_entity()
     entity_key = ENTITY_NAME_LOOKUP[entity]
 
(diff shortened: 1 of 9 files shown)


snuba/web/views.py
```diff
@@ -435,25 +438,49 @@ def handle_subscription_error(exception: InvalidSubscriptionError) -> Response:
     )
 
 
+@application.route("/<dataset:dataset>/<entity:entity>/subscriptions", methods=["POST"])
 @application.route("/<dataset:dataset>/subscriptions", methods=["POST"])
```
lynnagara, line 442: Are you planning to remove this one as a follow up? I think we should

snuba/web/views.py
```diff
@@ -435,25 +438,49 @@ def handle_subscription_error(exception: InvalidSubscriptionError) -> Response:
...
     )
 
 
+@application.route(
+    "/<dataset:dataset>/<entity:entity>/subscriptions/<int:partition>/<key>",
+    methods=["DELETE"],
+)
 @application.route(
     "/<dataset:dataset>/subscriptions/<int:partition>/<key>", methods=["DELETE"]
 )
```
lynnagara, line 471: Same here, let's deprecate this one once it is no longer being used?

lynnagara:
Just some questions about which endpoints we'll be deprecating, otherwise LGTM

snuba/web/views.py
ahmedetefy: Yep that's the plan, I am planning to remove line 442 and 470 i.e. for creation and deletion as a follow up

lynnagara: Sounds good, it could be useful to mark or comment these in code so it's clear what will happen to the old routes

lynnagara (approved):

PR #2259, opened 2021-12-07
feat(subscriptions): Add execute query processing strategy

lynnagara:
Adds a processing strategy that decodes a subscription task from the
Kafka payload and executes the ClickHouse query.

Not used by the subscriptions executor yet (i.e. we are not running
any actual queries on ClickHouse).


snuba/subscriptions/executor_consumer.py
```diff
@@ -90,3 +102,127 @@ def create(
...
+        next_step: ProcessingStrategy[KafkaPayload],
+    ) -> None:
+        self.__dataset = dataset
+        self.__max_concurrent_queries = max_concurrent_queries
+        self.__metrics = metrics
+        self.__next_step = next_step
+
+        self.__encoder = SubscriptionScheduledTaskEncoder()
+
+        self.__executor = ThreadPoolExecutor()
```
fpacifici, line 132: You should be able to create the executor at module level. Having it in the constructor means it will be recreated every time the strategy is recreated (including all rebalancing).

snuba/subscriptions/executor_consumer.py
```diff
@@ -90,3 +102,127 @@ def create(
...
+            self.__next_step.submit(message)
+
+        self.__next_step.poll()
+
+    def submit(self, message: Message[KafkaPayload]) -> None:
+        assert not self.__closed
+
+        # Tell the consumer to pause until we have removed some futures from
+        # the queue
+        if len(self.__queue) >= self.__max_concurrent_queries:
```
fpacifici, line 187: I don't think you need to implement this check by yourself. 
Just create an executor pool with a number of thread with the maximum amount of concurrent threads you want. 
That enforces the maximum concurrently. 
Then you may also add a higher max queue size to apply back pressure when needed.

snuba/subscriptions/executor_consumer.py
```diff
@@ -90,3 +102,127 @@ def create(
...
+    def poll(self) -> None:
+        while self.__queue:
+            if not self.__queue[0].future.done():
+                break
+
+            result_future = self.__queue.popleft()
+
+            message = result_future.message
+
+            self.__next_step.submit(message)
```
fpacifici, line 178: So the next step is the one that produces the results ?

snuba/subscriptions/executor_consumer.py
```diff
@@ -90,3 +102,127 @@ def create(
...
+            self.__metrics, "executor.concurrent.clickhouse"
+        )
+
+    def __execute_query(
+        self, task: ScheduledSubscriptionTask, tick: Tick
+    ) -> Tuple[Request, Result]:
+        timer = Timer("query")
+
+        with self.__concurrent_gauge:
+            request = task.task.subscription.data.build_request(
```
fpacifici, line 151: We will need to track the subscription queries executed by a subscriptions consumer separately from these ones. This way they would have the same referrer and be indistinguishable.
I think we should either ensure the metrics we use to track query timings have a dedicated tag or that we use a different referrer.

fpacifici (changes requested):
A few comments in line

snuba/subscriptions/executor_consumer.py
lynnagara: All the executor metrics are namespaced under `subscriptions.executor` so it's actually a different metric that just follows a similar naming convention to the existing one. (well it should be after this bugfix at least https://github.com/getsentry/snuba/pull/2265)

snuba/subscriptions/executor_consumer.py
lynnagara: Yes!

snuba/subscriptions/executor_consumer.py
lynnagara: Alright

snuba/subscriptions/executor_consumer.py
fpacifici: Everything recorded via the Timer will still go into `api` metrics without a distinction between the existing subscriptions and these ones. 
This is going to be a bit tricky, though we should definitively separate the metrics. Either by changing the referrer or by finding a way to add tags to the Timer (maybe changing the referrer is appropriate)

lynnagara: Ah I see, thought you were talking about the concurrent gauge

lynnagara: Changed the referrer to `subscriptions_executor`. Regular worker is still `subscriptions`.

fpacifici (approved):
cool, thanks

PR #2260, opened 2021-12-08
feat(subscriptions): Executor command takes multiple entities

lynnagara:
The executor command can now take multiple entities as long as those entities:
- match the dataset value which is passed
- have the same scheduled topic defined
- have the same result topic defined

In the future we may want to make this more flexible by providing a way to override
the scheduled and result topics - but for now they must all be the same in the code.

The entity name is not being used by the executor process yet. Later we will filter
scheduled subscriptions with entity names that do not match any of the values provided.

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -67,7 +67,7 @@ def build_executor_consumer(
         assert get_topics_for_entity(entity_name) == (
             scheduled_topic_spec,
             result_topic_spec,
-        )
+        ), "All entities must have same scheduled and result topics"
 
     return StreamProcessor(
         KafkaConsumer(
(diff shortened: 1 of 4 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -32,21 +33,41 @@ def build_executor_consumer(
...
+
+        return scheduled_topic_spec, result_topic_spec
+
+    scheduled_topic_spec, result_topic_spec = get_topics_for_entity(entity_names[0])
+
+    for entity_name in entity_names[1:]:
+        assert get_topics_for_entity(entity_name) == (
+            scheduled_topic_spec,
+            result_topic_spec,
+        )
```
fpacifici, line 70: Please add an error message

fpacifici (approved):

PR #2263, opened 2021-12-08
reduce health check "SHOW TABLES" queries to once per cluster

onewland:
https://getsentry.atlassian.net/browse/SNS-866

Collect list of tables at the start of health check, collect a list of unique clusters, call SHOW TABLES once per cluster and check that against the table list from the start.

Verified with logging that it changed the behavior on my local machine from checking my local cluster only once vs. once per table.

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -1,4 +1,5 @@
 import functools
+import itertools
 import logging
 import os
 import time
@@ -14,6 +15,7 @@ from typing import (
     Text,
     Tuple,
     Union,
+    cast,
 )
 from uuid import UUID
 
@@ -92,23 +94,33 @@ def check_clickhouse() -> bool:
     Checks if all the tables in all the enabled datasets exist in ClickHouse
     """
     try:
-        for name in get_enabled_dataset_names():
-            dataset = get_dataset(name)
-            for entity in dataset.get_all_entities():
-                for storage in entity.get_all_storages():
-                    clickhouse = storage.get_cluster().get_query_connection(
-                        ClickhouseClientSettings.QUERY
-                    )
-                    clickhouse_tables = clickhouse.execute("show tables")
-                    source = storage.get_schema()
-                    if isinstance(source, TableSchema):
-                        table_name = source.get_table_name()
-                        if (table_name,) not in clickhouse_tables:
-                            return False
+        datasets = [get_dataset(name) for name in get_enabled_dataset_names()]
+        entities = itertools.chain(
+            *[dataset.get_all_entities() for dataset in datasets]
+        )
+        storages = list(
+            itertools.chain(*[entity.get_all_storages() for entity in entities])
+        )
+        known_table_names = [
+            cast(TableSchema, storage.get_schema()).get_table_name()
+            for storage in storages
+            if isinstance(storage.get_schema(), TableSchema)
+        ]
+        unique_clusters = set([storage.get_cluster() for storage in storages])
+
+        logger.debug(f"checking for {known_table_names} on {unique_clusters}")
+        for cluster in unique_clusters:
+            clickhouse = cluster.get_query_connection(ClickhouseClientSettings.QUERY)
+            clickhouse_tables = clickhouse.execute("show tables")
+            for table in known_table_names:
+                if (table,) not in clickhouse_tables:
+                    logger.error(f"{table} not present in cluster {cluster}")
+                    return False
 
         return True
 
-    except Exception:
+    except Exception as err:
+        logger.error(err)
         return False
 
 


snuba/web/views.py
```diff
@@ -92,23 +94,33 @@ def check_clickhouse() -> bool:
...
+        )
+        storages = list(
+            itertools.chain(*[entity.get_all_storages() for entity in entities])
+        )
+        known_table_names = [
+            cast(TableSchema, storage.get_schema()).get_table_name()
+            for storage in storages
+            if isinstance(storage.get_schema(), TableSchema)
+        ]
+        unique_clusters = set([storage.get_cluster() for storage in storages])
```
onewland, line 109: I doubt it causes an issue but these seem like they could be memoized as I wouldn't expect any of these to change without a deploy

snuba/web/views.py
```diff
@@ -92,23 +94,33 @@ def check_clickhouse() -> bool:
...
-                            return False
+        datasets = [get_dataset(name) for name in get_enabled_dataset_names()]
+        entities = itertools.chain(
+            *[dataset.get_all_entities() for dataset in datasets]
+        )
+        storages = list(
+            itertools.chain(*[entity.get_all_storages() for entity in entities])
+        )
+        known_table_names = [
+            cast(TableSchema, storage.get_schema()).get_table_name()
```
vladkluev, line 105: how do you know these will all be TableSchemas?

snuba/web/views.py
vladkluev: Definitely could.

```
import functools

@functools.lru_cache
def get_unique_clickhouse_clusters():
    # your function
    
```

But yes the performance issue is not an issue

snuba/web/views.py
```diff
@@ -92,23 +94,33 @@ def check_clickhouse() -> bool:
...
-                    clickhouse = storage.get_cluster().get_query_connection(
-                        ClickhouseClientSettings.QUERY
-                    )
-                    clickhouse_tables = clickhouse.execute("show tables")
-                    source = storage.get_schema()
-                    if isinstance(source, TableSchema):
-                        table_name = source.get_table_name()
-                        if (table_name,) not in clickhouse_tables:
-                            return False
+        datasets = [get_dataset(name) for name in get_enabled_dataset_names()]
```
vladkluev, line 97: I think the retrieval of the unique cluster to table mapping could be made into its own function and that function could have a test.

vladkluev (approved):
Looks good! I would recommend a test but I'll leave that up to you

snuba/web/views.py
onewland: If you look two lines down, it's in the conditional of the list comprehension (e.g. it's filtered by the time it gets here)

vladkluev: 🤦 got it lol

snuba/web/views.py
fpacifici: This is not enough to identify uniquer clusters. 
You may have separate cluster config that use the same proxy host on Envoy.
For example these clusters: `erorrs_ro`, `errors`, `cdc` are all accessed through `localhost:WHATEVER`. 

So what you really want to deduplicate are host:port pairs and not the entire cluster config. I can point you to the prod config offline

fpacifici (changes requested):
please see my comment inline

snuba/clusters/cluster.py
```diff
@@ -274,6 +273,9 @@ def get_distributed_nodes(self) -> Sequence[ClickhouseNode]:
         ), "distributed_cluster_name must be set"
         return self.__get_cluster_nodes(self.__distributed_cluster_name)
 
+    def connection_tuple(self) -> Tuple[str, int]:
+        return (self.__query_node.host_name, self.__http_port)
```
fpacifici, line 277: This is not the right port. We use the TCP one for queries. 
If you want to return the connection tuple from the cluster you should have two methods, one for TCP and one for HTTP

snuba/web/views.py
```diff
@@ -92,23 +94,37 @@ def check_clickhouse() -> bool:
...
+        }.values()
+
+        logger.debug(f"checking for {known_table_names} on {unique_clusters}")
+        for cluster in unique_clusters:
+            clickhouse = cluster.get_query_connection(ClickhouseClientSettings.QUERY)
+            clickhouse_tables = clickhouse.execute("show tables")
+            for table in known_table_names:
+                if (table,) not in clickhouse_tables:
+                    logger.error(f"{table} not present in cluster {cluster}")
+                    return False
```
fpacifici, line 122: This does not seem right.
For each cluster you are extracting the clickhosue tables on that cluster.
Then you check that all known tables would be in `clickhouse_tables` which will fail because some tables can be only on some clusters. 
Like the metrics tables would be on the metrics cluster but not in the others, when you run show tables on the  errors cluster, the metrics tables would not be there and the check will fail.

fpacifici (changes requested):
Please see my comments inline

snuba/clusters/cluster.py
onewland: Makes sense, and we don't care about the HTTP port, or should that be part of the connection info tuple?

snuba/web/views.py
onewland: Sounds good, I'll construct the mapping of tables to clusters and make the check more granular.

snuba/web/views.py
```diff
@@ -92,23 +96,42 @@ def check_clickhouse() -> bool:
...
-                            return False
+        datasets = [get_dataset(name) for name in get_enabled_dataset_names()]
+        entities = itertools.chain(
+            *[dataset.get_all_entities() for dataset in datasets]
+        )
+        storages = list(
+            itertools.chain(*[entity.get_all_storages() for entity in entities])
+        )
+
+        cluster_grouped_table_names: Dict[ConnectionId, List[str]] = defaultdict(list)
```
fpacifici, line 107: This can be a `MutableMapping[ConnectionId, Set[str]]`
- MutableMapping is more abstract than Dict, it provides a more restrictive interface that does not surface too many implementation details.
- Set instead of list because if the same table was here twice (which could in theory happen) you don't care about dealing with it twice.

snuba/web/views.py
```diff
@@ -92,23 +96,42 @@ def check_clickhouse() -> bool:
...
+            storage.get_cluster().get_connection_id(): storage.get_cluster()
+            for storage in storages
+        }
+
+        for (cluster_key, cluster) in unique_clusters.items():
+            clickhouse = cluster.get_query_connection(ClickhouseClientSettings.QUERY)
+            clickhouse_tables = clickhouse.execute("show tables")
+            known_table_names = cluster_grouped_table_names[cluster_key]
+            logger.debug(f"checking for {known_table_names} on {cluster_key}")
+            for table in known_table_names:
```
fpacifici, line 126: Sorry, I think we forgot one dimension.
On the same host:port pair you can have multiple databases, which will be on separate clusters and will require a separate `SHOW TABLES` command.
So doing one `SHOW TABLES` per host:port is not enough. it has to be per host:port:database

fpacifici (changes requested):
I think we need to add the database name as a field in the connectionId tuple

snuba/web/views.py
```diff
@@ -92,23 +96,44 @@ def check_clickhouse() -> bool:
...
-                            return False
+        datasets = [get_dataset(name) for name in get_enabled_dataset_names()]
+        entities = itertools.chain(
+            *[dataset.get_all_entities() for dataset in datasets]
+        )
+        storages = list(
+            itertools.chain(*[entity.get_all_storages() for entity in entities])
+        )
+
+        cluster_grouped_table_names: MutableMapping[
```
fpacifici, line 107: rename to `connection_grouped_table_names`

snuba/web/views.py
```diff
@@ -92,23 +96,44 @@ def check_clickhouse() -> bool:
...
+            ConnectionId, Set[str]
+        ] = defaultdict(set)
+        for storage in storages:
+            if isinstance(storage.get_schema(), TableSchema):
+                cluster = storage.get_cluster()
+                cluster_grouped_table_names[cluster.get_connection_id()].add(
+                    cast(TableSchema, storage.get_schema()).get_table_name()
+                )
+
+        # De-dupe clusters by host:port pairs
```
fpacifici, line 117: host-port-database

fpacifici (approved):

PR #2264, opened 2021-12-08
feat(subscriptions): Default load factor comes from settings

lynnagara:
This change uses settings.SUBSCRIPTIONS_SCHEDULER_LOAD_FACTOR
as the default value for the load test. This allows us to
scale the load test up and down via code changes (and without
involvement from ops).

### DIFF
--- a/snuba/cli/subscriptions_scheduler.py
+++ b/snuba/cli/subscriptions_scheduler.py
@@ -6,7 +6,7 @@ import click
 from arroyo import configure_metrics
 from arroyo.backends.kafka import KafkaProducer
 
-from snuba import environment
+from snuba import environment, settings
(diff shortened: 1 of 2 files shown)


vladkluev (commented):
I'd like to get some more context.

1. How is additional load simulated?
2. What is the load factor measured in?
3. What are the value ranges for the load factor?

lynnagara:
The load factor option is temporary and basically just multiples the amount of actual subscriptions in the store by the load factor provided. It is an integer - we will start by increasing to 2 then work our way up until things start falling over / backlogging. The goal is to see how far we can get with this architecture / single scheduler process. This functionality already exists, this is just to change the values without ops intervention.

fpacifici (approved):

PR #2266, opened 2021-12-09
perf: Subscription scheduler load test

lynnagara:
Increase load_factor of the subscription scheduler to 2 for testing.
The scheduler will duplicate each subscription, essentially behaving
like there are double the number of subscriptions that actually exist.
Scheduler isn't used anywhere product facing yet so this is just for testing.


fpacifici (approved):
simpsons-villain

PR #2267, opened 2021-12-09
fix(reqs) - fix tests broken by requirements-test.txt allow new snuba-sdk

onewland:
fix requirements-test.txt to be more restrictive about incompatible snuba-sdk version


lynnagara (approved):
Thanks

evanh (approved):

PR #2268, opened 2021-12-10
perf: Omit redundant information from scheduled subscription message

lynnagara:
This change attempts to improve the performance of the scheduler by
reducing the size of the scheduled message and minimizing the amount
of work that the encoder needs to do. Previously we were encoding
and producing the entire tick to the scheduled subscription topic.
However the only part we actually need in order to build the request
is the upper offset.

### DIFF
--- a/snuba/subscriptions/codecs.py
+++ b/snuba/subscriptions/codecs.py
@@ -13,9 +13,8 @@ from snuba.subscriptions.data import (
     Subscription,
     SubscriptionData,
     SubscriptionIdentifier,
-    SubscriptionWithTick,
+    SubscriptionWithMetadata,
 )
-from snuba.subscriptions.utils import Tick
 from snuba.subscriptions.worker import SubscriptionTaskResult
 from snuba.utils.codecs import Codec, Encoder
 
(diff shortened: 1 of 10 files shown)


fpacifici:
I'd suggest we do more metrics gathering and some profiling before attempting fixes on this problem.
While this may help a bit, we do not really need what the culprit is.

lynnagara:
> I'd suggest we do more metrics gathering and some profiling before attempting fixes on this problem.
> While this may help a bit, we do not really need what the culprit is.

Yes, but I still think we should do this no matter what as there's no value in having the extra information there

snuba/subscriptions/codecs.py
```diff
@@ -97,14 +96,14 @@ def decode(self, value: KafkaPayload) -> ScheduledSubscriptionTask:
...
+            SubscriptionWithMetadata(
                 entity_key,
                 Subscription(
                     SubscriptionIdentifier.from_string(subscription_identifier),
                     SnQLSubscriptionData.from_dict(
                         scheduled_subscription_dict["task"]["data"], entity_key
                     ),
                 ),
-                Tick.from_dict(scheduled_subscription_dict["tick"]),
+                scheduled_subscription_dict["tick_upper_offset"],
```
fpacifici, line 107: This is push safe because the executor is not yet running right ?

fpacifici (approved):

snuba/subscriptions/codecs.py
lynnagara: Yes

PR #2269, opened 2021-12-10
perf: Collect metric for time spent encoding scheduled task message

lynnagara:
Track the time spent encoding scheduled task messages for the load test.
This is intended to be temporary and removed once we are done optimizing
this function.

### DIFF
--- a/snuba/subscriptions/scheduler_processing_strategy.py
+++ b/snuba/subscriptions/scheduler_processing_strategy.py
@@ -437,14 +437,19 @@ class ProduceScheduledSubscriptionMessage(ProcessingStrategy[CommittableTick]):
         tick = message.payload.tick
         assert tick.partition is not None
         tasks = self.__schedulers[tick.partition].find(tick)
+
+        start_encoding = time.time()
+        encoded_tasks = [self.__encoder.encode(task) for task in tasks]
+        self.__metrics.timing(
+            "ScheduledSubscriptionTask.encode", (time.time() - start_encoding) * 1000
+        )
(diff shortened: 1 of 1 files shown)


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -437,14 +437,19 @@ def submit(self, message: Message[CommittableTick]) -> None:
         tick = message.payload.tick
         assert tick.partition is not None
         tasks = self.__schedulers[tick.partition].find(tick)
+
+        start_encoding = time.time()
+        encoded_tasks = [self.__encoder.encode(task) for task in tasks]
+        self.__metrics.timing(
+            "ScheduledSubscriptionTask.encode", (time.time() - start_encoding) * 1000
+        )
```
fpacifici, line 445: The issue with this approach is that it does not tell you where we are spending time comparatively. It only tells you the absolute amount of time we spend in encoding, which may or may not be the culprit. 
You are right. It is likely to be the culprit, but it should be compared to the time we spend finding ticks at line 439 and the time spent producing, which is supposed to be neglibgible, but if we actually misconfigured the buffer size may be the culprit. 
So please collect the metric for the three steps

lynnagara: We have the total time of the function so this gives us time spent encoding vs (producing + finding ticks)

fpacifici (approved):
well, ok, if we will have to break down produce and find we will do as a follow up

PR #2272, opened 2021-12-13
feat(subscriptions): Use float for timestamp instead of date time string

lynnagara:
Smaller and faster to decode. This change is safe to deploy since we are
not running an executor (and hence decoding) yet.


evanh:
Looks good to me, except for the test failure.

PR #2274, opened 2021-12-14
feat(admin): Return the list of nodes for each storage

lynnagara:
Adds an API endpoint `/clickhouse_nodes` to Snuba admin. The new
endpoint returns the list of valid storages, their table names and the
ClickHouse host each table exists on. This will allow the client application
to later select which host they want to run a specific ClickHouse query on.


snuba/admin/clickhouse/nodes.py
```diff
@@ -0,0 +1,37 @@
...
+    schema = get_storage(storage_key).get_schema()
+    assert isinstance(schema, TableSchema)
+    return schema.get_table_name()
+
+
+def _get_local_nodes(storage_key: StorageKey) -> Sequence[str]:
+    storage = get_storage(storage_key)
+    return [
+        f"{node.host_name}:{node.port}"
+        for node in storage.get_cluster().get_local_nodes()
```
fpacifici, line 25: This will fail in production as we have clusters that are not single node but that have no cluster name set 
https://github.com/getsentry/snuba/blob/master/snuba/clusters/cluster.py#L272

fpacifici (changes requested):
You may want to catch that exception. So now we cannot run this without changing the production config .
I think we have two possible ways. 
1) we only return the hosts for the properly configured clusters. 
2) we return the hosts from the properly configured clusters, and we complement that with a temporary list we store in config till we do not fix the cluster config by declaring all the clusters we use individually.

snuba/admin/clickhouse/nodes.py
lynnagara: I forgot about that, thanks

onewland (approved):
Looks good to me

PR #2275, opened 2021-12-14
feat: Sample queries run by subscriptions executor

lynnagara:
The subscriptions executor samples queries according to the rate
provided by `executor_sample_rate`. This is to avoid overloading
ClickHouse with queries while we are running both the old and new
pipelines in parallel. Sampling is based on a hash of the subscription
ID so that the same subscriptions are always being executed. The default
is 0.0, i.e. no queries are passed through the executor pipeline or run
on ClickHouse.

This depends on https://github.com/getsentry/snuba/pull/2270

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -4,7 +4,8 @@ import logging
 import time
 from collections import deque
 from concurrent.futures import ThreadPoolExecutor
-from typing import Callable, Deque, Mapping, Optional, Sequence, Tuple
+from typing import Callable, Deque, Mapping, Optional, Sequence, Tuple, cast
+from zlib import crc32
 
 from arroyo import Message, Partition, Topic
 from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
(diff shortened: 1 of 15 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -222,17 +224,28 @@ def submit(self, message: Message[KafkaPayload]) -> None:
...
-                    self.__executor.submit(
-                        self.__execute_query, task, tick_upper_offset
+        # We need to sample queries to ClickHouse while this is being rolled out
+        # as we don't want to duplicate every subscription query
+        executor_sample_rate = cast(
+            float, state.get_config("executor_sample_rate", 0.0)
+        )
+        subscription_id = str(task.task.subscription.identifier)
+        should_execute = (
+            (crc32(subscription_id.encode("utf-8")) & 0xFFFFFFFF) / 2 ** 32
```
fpacifici, line 234: Why the bitwise AND ? `& 0xFFFFFFFF` should not be needed.

fpacifici (approved):

snuba/subscriptions/executor_consumer.py
lynnagara: Probably not needed but it was suggested in the docs https://docs.python.org/3/library/zlib.html#zlib.crc32 to make it compatible across all platforms and python versions

PR #2279, opened 2021-12-14
feat(tracing) Change the return type of the native execute 

evanh:
It is very difficult to capture new data from Clickhouse with the current return
value from the function. The current wrapper around the clickhouse-driver
library returns a simple list of results, that gets overloaded with column types
when that meta information is returned. Change the return type to be an object,
so that any data that we might want to propagate up from Clickhouse becomes easy
to add without having to go through this process again.


lynnagara (approved):

snuba/clickhouse/native.py
```diff
@@ -269,9 +299,14 @@ def __transform_result(self, result: Sequence[Any], with_totals: bool) -> Result
...
-            new_result = {"data": data, "meta": meta, "totals": totals}
+            new_result = {
+                "data": data,
+                "meta": meta,
+                "totals": totals,
+                "profile": profile,
+            }
         else:
-            new_result = {"data": data, "meta": meta}
+            new_result = {"data": data, "meta": meta, "profile": profile}
```
lynnagara, line 309: Does that mean our API is returning "profile" information now?

snuba/clickhouse/native.py
```diff
@@ -23,6 +26,13 @@
 metrics = MetricsWrapper(environment.metrics, "clickhouse.native")
 
 
+@dataclass(frozen=True)
+class ClickhouseResult:
+    results: Sequence[Any] = field(default_factory=list)
+    meta: Sequence[Any] | None = None
+    profile: Mapping[str, Any] | None = None
```
fpacifici, line 33: Can we provide something more specific ?
Like having a datclass for the profile information. I am not sure what we gain from leaving it as a free form mapping knowing that the profile information is static and well known.

snuba/clickhouse/native.py
evanh: Yes, the data is there, just not being used yet. I am going to make use of it in separate PR(s).

snuba/clickhouse/native.py
evanh: I removed the profiling data from this PR, it's not really this is about. I'll add it back in a separate PR.

PR #2280, opened 2021-12-14
feat(subscriptions): Executor uses the ExecuteQuery strategy

lynnagara:
Integrates the ExecuteQuery strategy in the subscriptions executor.
Note that currently the `executor_sample_rate` option is not set
so all scheduled subscriptions will be filtered at the start of the
pipeline, and even their offsets will not be committed.

This is to avoid running duplicate queries on ClickHouse until we
are ready to begin testing.


snuba/subscriptions/executor_consumer.py
```diff
@@ -124,10 +124,27 @@ def join(self, timeout: Optional[float] = None) -> None:
...
+    ) -> None:
+        self.__dataset = dataset
+        self.__max_concurrent_queries = max_concurrent_queries
+        self.__metrics = metrics
+
     def create(
         self, commit: Callable[[Mapping[Partition, Position]], None]
     ) -> ProcessingStrategy[KafkaPayload]:
-        return Noop(commit)
+        executor = ThreadPoolExecutor(self.__max_concurrent_queries)
```
fpacifici, line 137: You do not need to create a threadpool each time you create the step, which can happen often. Please keep it static.

lynnagara: Moved this line out of the factory and passing it in now so it doesn't get recreated on partition assignment, etc

fpacifici (approved):

PR #2054, opened 2021-08-12
feat(hackweek): Anonymized Query Formatter for logical queries

JoshFerge:
adds a new query formatter to format logical queries into a basic string with literals stripped. This PR is ~70% of the way there, todos below:

- [ ] Add more test coverage, probably fix some things based on that
- [ ] Organize code, attempt to reduce duplication

PR #2282, opened 2021-12-15
fix(pipeline): Add Anonymized SnQL to BigQuery

rahul-kumar-saini:
## Overview
- The incoming query to Snuba is not passed correctly to BigQuery
- Added anonymizer to anonymize snql at parser level and enabled passing an anonymized snql query to BigQuery

## Notes
- Code structure inspired from #2054


snuba/query/snql/parser.py
```diff
@@ -1347,10 +1349,14 @@ def _post_process(
...
 def parse_snql_query(
     body: str, dataset: Dataset, custom_processing: Optional[CustomProcessors] = None,
-) -> Union[CompositeQuery[QueryEntity], LogicalQuery]:
+) -> Tuple[Union[CompositeQuery[QueryEntity], LogicalQuery], str]:
     with sentry_sdk.start_span(op="parser", description="parse_snql_query_initial"):
         query = parse_snql_query_initial(body)
 
+    with sentry_sdk.start_span(op="parser", description="anonymize_snql_query"):
+        query_copy = copy.deepcopy(query)
+        snql_anonymized = format_snql_anonymized(query_copy).get_sql()
```
rahul-kumar-saini, line 1358: For reasons I could not figure out, the anonymize formatter modifies the query AST, so a deep copy is required.

tests/query/snql/test_snql_anonymizer.py
```diff
@@ -0,0 +1,385 @@
...
+    )
+  )
+)""",
+        id="aliased columns in select and group by",
+    ),
+]
+
+
+@pytest.mark.parametrize("query_body, expected_snql_anonymized", test_cases)
+def test_format_expressions(query_body: str, expected_snql_anonymized: str) -> None:
```
rahul-kumar-saini, line 360: These tests are just a subset of tests in tests/query/snql/test_query.py. If the tests should just be in that file I can put them there, it just adds a lot of text.

snuba/query/snql/parser.py
rahul-kumar-saini: Nevermind, figured it out: #2284. I'll update this when that is merged.

snuba/query/snql/anonymize.py
```diff
@@ -0,0 +1,215 @@
...
+            _format_limit(query, formatter),
+            _format_offset(query, formatter),
+            _format_granularity(query, formatter),
+            _format_totals(query, formatter),
+        ]
+        if v is not None
+    ]
+
+
+def _format_select(
```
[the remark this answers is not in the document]
fpacifici: I think you should be able to simply change the signature of the methods in formatter/query to take an `ExpressionVisitor[str]`. There should be no specific dependency on Clickhouse related logic. So you should be able to get rid of all these methods.

snuba/query/snql/anonymize_visitor.py
```diff
@@ -0,0 +1,22 @@
+from datetime import date, datetime
+
+from snuba.query.expressions import Literal, StringifyVisitor
+
+
+class AnonymizeAndStringifySnQLVisitor(StringifyVisitor):
```
fpacifici, line 6: You do not need to rebuild this.
https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/formatter/expression.py#L205

Now it relies on ClickhouseExpressionFormatterBase, but that is not a blocker for you. 
In every case you can turn ClickhouseExpressionFormatterBase into `ExpressionFormatterBase` as there is really not anything clickhouse specific in that class.

snuba/query/snql/parser.py
fpacifici: This means that, after #2284  you will remove this change ?

snuba/request/__init__.py
```diff
@@ -14,6 +14,7 @@ class Request:
     id: str
     body: Mapping[str, Any]
     query: Union[Query, CompositeQuery[Entity]]
+    snql_anonymized: str
```
fpacifici, line 17: Why do we need this in the request? 
If you fix https://github.com/getsentry/snuba/pull/2282/files#r769854871 this change should go as well, right ?

snuba/subscriptions/worker.py
```diff
@@ -96,6 +96,7 @@ def run_consistent() -> Result:
                     id=request.id,
                     body=copy.deepcopy(request.body),
                     query=copy.deepcopy(request.query),
+                    snql_anonymized=request.snql_anonymized,
```
fpacifici, line 99: Same, I think we should be able to generate the formatted query from the original query right before this statement
https://github.com/getsentry/snuba/blob/master/snuba/web/query.py#L122-L130

fpacifici (changes requested):
Thanks. 
See my comments in line

snuba/query/snql/parser.py
rahul-kumar-saini: I removed the deep copy logic, the anonymizing still happens here

snuba/clickhouse/formatter/query.py
```diff
@@ -201,22 +195,29 @@ def _format_arrayjoin(
...
+        if ast_limit is not None and ast_limit != 1000
         else None
     )
 
 
+def _format_offset(
+    query: AbstractQuery, formatter: ExpressionVisitor[str]
+) -> Optional[StringNode]:
+    ast_offset = query.get_offset()
+    return StringNode(f"OFFSET {ast_offset}") if ast_offset != 0 else None
```
rahul-kumar-saini, line 212: I've split up the limit and offset format functions, and made it so they don't show up if they're just the default values. Since the formatted queries would only be read by humans, this shouldn't be a problem correct? I can revert the default value check if they should be displayed regardless.

snuba/subscriptions/worker.py
rahul-kumar-saini: If we anonymize here, we won't be anonymizing the "original" version of the Query we receive. The parser itself runs a bunch of [post processors](https://github.com/getsentry/snuba/blob/d9c91a606f70c05e2d998e985a897d354fc97f93/snuba/query/snql/parser.py#L1361-L1381) on the Query which modify the AST, so we have to anonymize right after creating the initial AST. This just means it happens in `query/snql/parser.py::parse_snql_query()` and it has to be propagated through the Request.

snuba/request/__init__.py
rahul-kumar-saini: That fix was referring only to avoiding creating a deep copy, the comment below addresses why it's in the Request.

snuba/clickhouse/formatter/query.py
evanh: I think that's fine.

snuba/query/snql/parser.py
```diff
@@ -1350,10 +1351,13 @@ def _post_process(
...
 def parse_snql_query(
     body: str, dataset: Dataset, custom_processing: Optional[CustomProcessors] = None,
-) -> Union[CompositeQuery[QueryEntity], LogicalQuery]:
+) -> Tuple[Union[CompositeQuery[QueryEntity], LogicalQuery], str]:
     with sentry_sdk.start_span(op="parser", description="parse_snql_query_initial"):
         query = parse_snql_query_initial(body)
 
+    with sentry_sdk.start_span(op="parser", description="anonymize_snql_query"):
+        snql_anonymized = format_snql_anonymized(query).get_sql()
+
```
fpacifici, line 1360: Are you returning the anonymized query from the parser in order to format it before applying the post processing steps ?
I am not convinced this is the right approach. We are now coupling the parser with the concept of query log (that needs anonymization). It is very arbitrary that the parser needs to concern itself with how we anonymize the query for logging. That should be a concern internal to the logger. 
In principle, in order to keep your design sane, you want each component, class, function, to depend on as few things as possible, and each dependency is supposed to have a logical meaning. This allows you to understand the design much better and to architect changes. A dependency here on logic that is only concerned with logging is not a dependency that is functional or logically related with parsing.

The issue is that separating these two concerns is going to be tricky. @evanh  do you remember why we do post processing inside the `parse_snql_query` function instead of attaching them to the execution pipeline once the request is built ? Is there a reason or is it there for historical reasons, or was it an arbitrary distinction between what is a "second pass of the parser" vs what is query processing ?

snuba/clickhouse/formatter/query.py
```diff
@@ -201,22 +195,29 @@ def _format_arrayjoin(
...
 def _format_limit(
-    query: AbstractQuery, formatter: ClickhouseExpressionFormatterBase
+    query: AbstractQuery, formatter: ExpressionVisitor[str]
 ) -> Optional[StringNode]:
     ast_limit = query.get_limit()
     return (
-        StringNode(f"LIMIT {ast_limit} OFFSET {query.get_offset()}")
-        if ast_limit is not None
+        StringNode(f"LIMIT {ast_limit}")
+        if ast_limit is not None and ast_limit != 1000
```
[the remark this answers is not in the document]
evanh: 1000 is the default limit the parser adds if the user didn't supply one.

snuba/query/snql/parser.py
evanh: I believe it was a mix of both. For example some of the post processing is validation, which we need to do in the parser since subsequent steps don't have a mechanism for rejecting queries. Also the post processing includes steps such as picking the entity for Discover queries and aligning the date conditions, which have an impact on downstream pipeline steps. Some of the post processing was definitely me deciding that it made more sense in the parser instead of adding another query processor.

Conceptually right now we have an idea that at every step of the pipeline, the query is correct and valid. If we split the logic into two steps, the parsing and the post processing, then we will be breaking that contract.

We do want the anonymized query to be directly after parsing to the AST, in order to keep as close as possible to what is actually being sent over the wire. Inserting the logging in here means we don't have to parse the SnQL twice and ensures that what is logged hasn't been transformed by the post processors at all.

snuba/clickhouse/formatter/query.py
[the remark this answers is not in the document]
rahul-kumar-saini: It would always just add "LIMIT 1000" to the query (unless a different limit was given by the user), I could remove that it just means every snql query we see on bigquery will end with `LIMIT 1000`. It's just slightly cleaner this way and the only downside is we wouldn't see the limit if the user happens to supply a limit of 1000.

snuba/clickhouse/formatter/expression.py
```diff
@@ -202,7 +202,12 @@ def _format_date_literal(self, exp: Literal) -> str:
         )
 
 
-class ExpressionFormatterAnonymized(ExpressionFormatterBase):
+def _gen_random_number() -> int:
+    # https://xkcd.com/221/
+    return -1337
```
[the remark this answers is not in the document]
rahul-kumar-saini: Both changes you've commented on are already in the codebase and were introduced by @vladkluev in this PR:
https://github.com/getsentry/snuba/pull/2308

I believe the single quotes are meant to show that it is indeed a string, and the number to show it is a number instead of `$S` and `$N`?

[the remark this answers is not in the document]
vladkluev: yes, this way you can copy and paste the query to run/profile it without needing to find and replace in it.

snuba/query/snql/parser.py
fpacifici: mh ok, I don't think mixing the two concepts is a great idea, though probably not worth blocking this. 
Can we record a task to revisit these dependencies ?

fpacifici (changes requested):
I think we cannot just remove the limit if it is the default value as that is not a default value for Clickhouse. That would change the production behavior.

snuba/clickhouse/formatter/query.py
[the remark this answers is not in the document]
rahul-kumar-saini: I hadn't realized this formatter produces the Query sent directly to ClickHouse. I'll revert this.

fpacifici (approved):

snuba/clickhouse/formatter/expression.py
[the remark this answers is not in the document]
evanh: Gotcha, that makes sense.

PR #2284, opened 2021-12-15
fix(parsing): Group By clause in AST is now a List

rahul-kumar-saini:
## Overview
- Parser building the AST from a query creates a `map()` for the "group by" clause of a query
- This map is a kind of generator, which means if turned into a list somewhere down the line, the generator is drained and `query.get_group_by()` returns an empty map

### Example
```python
>>> from snuba.query.snql.parser import parse_snql_query_initial
>>> body = "MATCH (...) SELECT ... BY xyz ..."
>>> query = parse_snql_query_initial(body)
>>> print(query.get_groupby())
<map object at 0x110d21a60>
>>> print(list(query.get_groupby()))
[xyz]
>>> print(list(query.get_groupby()))
[]
```
- If at any time the `map` gets cast to a `list`, we lose the group by clause entirely in the AST

## Changes
- Replaced `map` with `list` in the SnQL visitor when parsing group by clause


vladkluev (commented):
can you add a test that would've broken before this change?

evanh (approved):

PR #2287, opened 2021-12-15
feat(subscriptions): Add the strategy to produce a subscription result

lynnagara:
This change adds the ProduceResult strategy. It's job is to take a
SubscriptionTaskResult, encode the message to a Kafka payload and produce
it to the provided result topic. Once producing is done, offsets are
committed. This strategy is designed to be chained together with the
ExecuteQuery strategy (which provides the SubscriptionTaskResult) in the
subscriptions executor.

The approach taken is very similar the ProducedScheduledSubscriptionMessage
strategy in the subscription scheduler which encodes and produces the
messages for the scheduled topic.

This change only adds the strategy - it is not being used anywhere yet
and will be integrated into the executor as a follow up.

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import logging
 import time
 from collections import deque
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import Future, ThreadPoolExecutor
 from typing import Callable, Deque, Mapping, Optional, Sequence, Tuple, cast
 from zlib import crc32
 
 from arroyo import Message, Partition, Topic
+from arroyo.backends.abstract import Producer
 from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -286,3 +290,74 @@ def join(self, timeout: Optional[float] = None) -> None:
...
+        self.__result_topic = Topic(result_topic)
+        self.__commit = commit
+
+        self.__encoder = SubscriptionTaskResultEncoder()
+
+        self.__queue: Deque[
+            Tuple[Message[SubscriptionTaskResult], Future[Message[KafkaPayload]]]
+        ] = deque()
+
+        self.__max_buffer_size = 10000
```
fpacifici, line 317: this is something we will want to tweak. please let's ensure we eventually move it to CLI or settings.

fpacifici (approved):

snuba/subscriptions/executor_consumer.py
lynnagara: 👍

PR #2288, opened 2021-12-15
feat(admin): Add ability to delete runtime config items

lynnagara:
Add API and UI for deleting runtime config items. Includes a confirmation
step in the UI.

Demo:

https://user-images.githubusercontent.com/1779792/146277009-17e57ea8-31b3-482c-b395-5fec752fe600.mov


snuba/admin/views.py
```diff
@@ -129,6 +129,24 @@ def configs() -> Response:
...
 
+@application.route("/configs/<config_key>", methods=["DELETE"])
+def config(config_key: str) -> Response:
+    if request.method == "DELETE":
+        user = request.headers.get("X-Goog-Authenticated-User-Email")
+        state.delete_config(config_key, user=user)
+
+        notification_client.notify(
+            RuntimeConfigAction.ADDED,
+            {"option": config_key, "old": None, "new": None},
```
evanh, line 140: It might be worth logging the old value, in case we need to restore it.

evanh (approved):

snuba/admin/views.py
lynnagara: Yeah, unfortunately we don't get it back from the config API so we'd need to make a get call first for it. I decided to omit that for now.

evanh: Is it a problem to fetch the value? This isn't a high traffic endpoint.

lynnagara: No it's not, will add it

PR #2291, opened 2021-12-16
fix(cra-metrics): Send `entity` in subscription payload

ahmedetefy:
Adds `entity` in the subscription results payload and updates the schema version to `3`

- When snuba sends a subscription update for a subscription that has been deleted from sentry, sentry responds by sending a `DELETE` request to snuba to delete that subscription. 
- Previously only the `dataset` was required in the request because the `dataset` to `entity` relation was 1-to-1. 
- Now however, since the introduction of `metrics` dataset and its multiple entities, there is no "clean way" to figure out on sentry side the `entity_key` since sentry no longer has a record of the subscription.
- However, I can temporarily parse the entity from the query sent in the subscription update  https://github.com/getsentry/sentry/pull/30432/files#diff-b75c61e51cc94ac9306008d19e42a36839745e9a981faf6c06f069d2c686056dR285-R301

Pending the new schema version: https://github.com/getsentry/sentry/pull/30736

As a follow up, once this is merged, [I will hack out that diff shown in the sentry PR link](https://github.com/getsentry/sentry/pull/30720)


evanh (approved):

fpacifici:
Aren't `subscription_id` and `entity` redundant? If you have the subscription id you should be able to know the entity in Sentry, or am I missing something ?
If they are, we should not add the entity explicitly as its own field as that couples the protocol to the query syntax and to the snuba data model. 
The idea of subscriptions is fairly abstract from the snuba data model in terms of interface (the way you create, delete the subscription, the way queries are executed and the way you receive the results). Which would make it easier to adapt to a different data model. 
Adding an explicit dependency in the protocol (which already happened when porting to snql unfortunately) couples the subscriptions mechanism more to the snuba data model, thus it is not ideal.

So, if adding this field is not necessary it should be avoided, thus the question whether it is actually redundant when you already have the `subscription_id`. If it is required from a functional standpoint it is a different matter.

ahmedetefy:
@fpacifici 
> Aren't subscription_id and entity redundant? If you have the subscription id you should be able to know the entity in Sentry, or am I missing something ?

Not really, because I am handling the case where the subscription was deleted from sentry but it wasn't deleted from snuba yet, so basically this branch https://github.com/getsentry/sentry/blob/01ec97f2c73067b99e26f483a35c662ecb3a7aac/src/sentry/snuba/query_subscription_consumer.py#L271
and so getting the `subscription_id` won't really help because there is no corresponding record of it

fpacifici:
Nah, not worth it. Please go ahead and merge. Fitting it into the request would be a worse hack.
Though another question. Why do we end in the situation where the snuba subscription was not deleted by the sentry query was ? This seems wrong.

ahmedetefy:
@fpacifici 
>Nah, not worth it. Please go ahead and merge. Fitting it into the request would be a worse hack.
Though another question. Why do we end in the situation where the snuba subscription was not deleted by the sentry query was ? This seems wrong.

I think its there just there as a failsafe solution. I think in reality what actually happens is that some subscription updates are still sent that were queued before the subscription was deleted.
Although I guess it could also happen for whatever reason that snuba deletion fails since its not like the deletion from sentry and snuba are in a transaction somehow.

lynnagara (approved):

PR #2292, opened 2021-12-16
ref(api): Removes deprecated routes

ahmedetefy:
Removes deprecated routes for creation and deletion of subscriptions that only accept dataset as url params, in favor of routes that take both a dataset and an entity key url params

Now that the routes called for subscriptions have been updated on sentry side, it makes sense to deprecate these 
https://github.com/getsentry/sentry/pull/30400
https://github.com/getsentry/sentry/pull/30432


evanh (commented):
Question: are these endpoints completely gone in Sentry? Or are they being controlled by a feature flag?

evanh:
Also please make sure the tests are all passing.

ahmedetefy:
> Question: are these endpoints completely gone in Sentry? Or are they being controlled by a feature flag?

Completely gone

evanh (approved):
👍

lynnagara (approved):
Yay

PR #2293, opened 2021-12-16
feat(pipeline): Capture Trace flag

rahul-kumar-saini:
## Overview
- We need a flag to know whether or not to capture the ClickHouse trace
- This comes in through `RequestSettings` and gets propagated through to query execution
- The flag currently isn't used to do anything but it is required for future changes to capture and return ClickHouse tracing data when needed through the admin portal

## Changes
- Added `capture_trace` flag to `RequestSettings` and schema
- Updated execution pipeline to propagate flag through to ClickHouse `execute()` methods


vladkluev:
So this means that technically any snuba API call can request a trace. Is that what we want? Are there any dangers in exposing this functionality to all callers?

onewland:
Is there any reason to separate the change where it gets wired through?

rahul-kumar-saini:
Right, the flag needs to be passed via the API request but I'm unsure how to make it so only the admin portal can use it. 

@onewland This doesn't _need_ to be a separate PR but I'd like to figure out how to limit the exposure of the flag in the API and we can work on using it later.

evanh (commented):
Could you add a test for this? Maybe add a test to one of the *_api.py tests to make sure that adding the field doesn't break anything.

evanh (approved):

snuba/request/schema.py
```diff
@@ -133,6 +133,10 @@ def generate_template(self) -> Any:
             "dry_run": {"type": "boolean", "default": False},
             # Flags if this a legacy query that was automatically generated by the SnQL SDK
             "legacy": {"type": "boolean", "default": False},
+            # Flag to retrieve only ClickHouse tracing data. If this is True,
+            # the query execution should not return the results of the query, but
+            # return the tracing data of the query execution within ClickHouse
+            "capture_trace": {"type": "boolean", "default": False},
```
fpacifici, line 139: I am not sure we want to expose this in the snuba api.
We cannot change the log handler there if that impacts the system logs or snuba logs, which means we would not be using this in that circumstance. 
Also this may not work in HTTP and we want to move away from the TCP protocol for the Snuba queries, which may mean this will not be able to work on the snuba api.

I would suggest we remove it from there till we do not decide to expose it.

PR #2294, opened 2021-12-16
feat(admin) - wire in queries and rendering results

onewland:
Construct queries and send them to the API, and display the query result history in a table


snuba/admin/static/clickhouse_queries/types.tsx
```diff
@@ -9,4 +9,27 @@ type ClickhouseNodeData = {
   local_nodes: ClickhouseNode[];
 };
 
-export { ClickhouseNodeData };
+type ClickhouseCannedQuery = {
+  description?: string;
+  name: string;
+  sql?: string;
```
lynnagara, line 15: How come description and sql are optional, shouldn't we always get this from the API?

snuba/admin/static/api_client.tsx
```diff
@@ -57,6 +64,25 @@ function Client() {
...
+          method: "POST",
+          body: new Blob([JSON.stringify(query)], { type: "application/json" }),
+        })
+      )
+        .then((resp) => resp.json())
+        .then((resp: QueryResult) => {
+          // used for keying in history
+          resp.timestamp = Date.now();
+          return resp;
+        });
```
lynnagara, line 84: Can you remove this? Seems like it's not being used

snuba/admin/static/api_client.tsx
```diff
@@ -57,6 +64,25 @@ function Client() {
...
+    getClickhouseCannedQueries: () => {
+      const url = baseUrl + "clickhouse_queries";
+      return fetch(url).then((resp) => resp.json());
+    },
+    executeQuery: (query: QueryRequest) => {
+      const url = baseUrl + "run_clickhouse_system_query";
+      return fetch(
+        new Request(url, {
+          method: "POST",
+          body: new Blob([JSON.stringify(query)], { type: "application/json" }),
```
lynnagara, line 76: Why blob? Why not just post the JSON?

snuba/admin/static/clickhouse_queries/index.tsx
```diff
@@ -44,9 +51,27 @@ function ClickhouseQueries(props: { api: Client }) {
...
+        query_name: queryName,
+      };
+    });
+  }
+
+  function executeQuery() {
+    props.api.executeQuery(query as QueryRequest).then((result) => {
+      result.input_query = `${query.query_name}(${query.storage},${query.host}:${query.port})`;
+      setQueryResultHistory((prevHistory) => [result, ...prevHistory]);
+      console.log(queryResultHistory);
```
lynnagara, line 67: nit: Remove this

snuba/admin/static/table.tsx
```diff
@@ -5,13 +5,15 @@ import { COLORS } from "./theme";
 type TableProps = {
   headerData: ReactNode[];
   rowData: ReactNode[][];
-  columnWidths: number[];
+  columnWidths?: number[];
 };
 
 function Table(props: TableProps) {
-  const { headerData, rowData, columnWidths } = props;
+  let { headerData, rowData, columnWidths } = props;
```
lynnagara, line 12: nit: Why changing this? I normally prefer const unless reassigning later

snuba/admin/static/table.tsx
```diff
@@ -5,13 +5,15 @@ import { COLORS } from "./theme";
...
 
-  const sumColumnWidths = columnWidths.reduce((acc, i) => acc + i, 0);
+  const autoColumnWidths = Array(headerData.length).fill(1);
+  const notEmptyColumnWidths = columnWidths ?? autoColumnWidths;
```
lynnagara, line 15: TIL about ??

lynnagara (commented):
For some reason this does not work for me. I get 500 errors and `TypeError: cannot unpack non-iterable ClickhouseResult object` error when making requests via this UI.

snuba/admin/static/api_client.tsx
onewland: I just copied this example from MDN: https://developer.mozilla.org/en-US/docs/Web/API/Request/blob#examples

snuba/admin/static/table.tsx
onewland: in my first version, i reassigned columnWidths. will update

snuba/admin/static/clickhouse_queries/types.tsx
onewland: description is not present for one of the predefined queries.

sql should always be, not sure why I did that

onewland:
> For some reason this does not work for me. I get 500 errors and `TypeError: cannot unpack non-iterable ClickhouseResult object` error when making requests via this UI.

This seemed to be caused by a merge from master, I'll look into it

snuba/admin/static/api_client.tsx
lynnagara: I think we should just send JSON like the rest of the API
```suggestion
return fetch(
  headers: { "Content-Type": "application/json" },
  url,
  method: "POST",
  body: JSON.stringify(query)
)
```

snuba/admin/static/clickhouse_queries/types.tsx
lynnagara: It seems like the description key is actually always there but it is nullable: https://github.com/getsentry/snuba/blob/master/snuba/admin/clickhouse/system_queries.py#L54.
So the type should maybe be `string | null`

lynnagara (approved):

PR #2295, opened 2021-12-16
change confusing warning message

vladkluev:
This message is confusing because there is nothing about this function which makes it have to be a write. In fact subscriptions use this function to read.

### DIFF
--- a/snuba/clickhouse/native.py
+++ b/snuba/clickhouse/native.py
@@ -168,7 +168,7 @@ class ClickhousePool(object):
             except (errors.NetworkError, errors.SocketTimeoutError, EOFError) as e:
                 # Try 3 times on connection issues.
                 logger.warning(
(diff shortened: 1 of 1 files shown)


snuba/clickhouse/native.py
```diff
@@ -168,7 +168,7 @@ def execute_robust(
             except (errors.NetworkError, errors.SocketTimeoutError, EOFError) as e:
                 # Try 3 times on connection issues.
                 logger.warning(
-                    "Write to ClickHouse failed: %s (%d tries left)",
+                    "Robust ClickHouse query execution failed: %s (%d tries left)",
```
onewland, line 171: I would drop "robust" from the message, but I'd rather just get this merged than slow you down

onewland (approved):

PR #2298, opened 2021-12-17
feat(subscriptions): Only execute scheduled tasks that match entity

lynnagara:
Only execute the scheduled tasks on a topic that match the entity name
passed via the CLI. Also records a metric counting the number of skipped
executions.


tests/subscriptions/test_executor_consumer.py
```diff
@@ -155,13 +155,14 @@ def generate_message(
...
     dataset = get_dataset("events")
+    entity_names = ["events"]
     max_concurrent_queries = 2
     executor = ThreadPoolExecutor(max_concurrent_queries)
     metrics = TestingMetricsBackend()
     next_step = mock.Mock()
 
     strategy = ExecuteQuery(
-        dataset, executor, max_concurrent_queries, metrics, next_step
+        dataset, entity_names, executor, max_concurrent_queries, metrics, next_step
```
fpacifici, line 165: please let's test the negative case when a query is not executed because it is not in the list of entity names provided.

evanh (commented):
Will this have the same problem we had with the errors/transactions, where the backlog would jump up because there were no messages coming in and the commit offset wasn't being updated?

lynnagara:
> Will this have the same problem we had with the errors/transactions, where the backlog would jump up because there were no messages coming in and the commit offset wasn't being updated?

If we have a lot of one type of subscription and not the other AND we choose to run them on separate executor pools it's possible the one with less subscriptions will have a higher lag. But the numbers won't be as high because we are not dealing with nearly as many messages on this topic. It's also more likely we want to run all the entities for a topic together (hence adding this functionality) so we are less likely to have that problem.

fpacifici:
@evanh  also this is not a synchronized consumer. It will never be accumulating errors while waiting for the first transaciton. It just consumes whatever shows up on the topic.

tests/subscriptions/test_executor_consumer.py
lynnagara: Sounds good, added the test

fpacifici (approved):

PR #2299, opened 2021-12-17
feat(subscriptions): Integrate ProduceResult step into executor

lynnagara:
Subscription executor now produces results to the result topic
after executing the ClickHouse query and before committing offsets.
The result topic temporarily comes from either the hardcoded value
or the --override-result-topic flag passed to the CLI to avoid
any chance of writing to the actual result topic that is used
in production currently.


snuba/cli/subscriptions_executor.py
```diff
@@ -79,10 +84,32 @@ def subscriptions_executor(
...
+
+    storage = get_entity(entity_key).get_writable_storage()
+    assert storage is not None
+    stream_loader = storage.get_table_writer().get_stream_loader()
+    result_topic_spec = stream_loader.get_subscription_scheduled_topic_spec()
+    assert result_topic_spec is not None
+
+    producer = KafkaProducer(
+        build_kafka_producer_configuration(
+            result_topic_spec.topic, override_params={"partitioner": "consistent"},
```
fpacifici, line 104: where does this parameter come from ?

tests/subscriptions/test_executor_consumer.py
```diff
@@ -107,8 +110,7 @@ def test_executor_consumer() -> None:
 
     encoded_task = encoder.encode(task)
 
-    fut = producer.produce(topic, payload=encoded_task)
-    fut.result()
+    producer.produce(scheduled_topic, payload=encoded_task).result()
```
fpacifici, line 113: Shouldn't we be validating that the query happens, and we have proper results ?

fpacifici (commented):
I think we can make the executor test better to verify that we receive a result, now that you added the producer there.

snuba/cli/subscriptions_executor.py
fpacifici: Ok, it is from the existing subscriptions consumer

tests/subscriptions/test_executor_consumer.py
fpacifici: Now we do

fpacifici:
With this I added the test I asked for in the previous review and fixed one bug in the executor consumer

snuba/subscriptions/executor_consumer.py
```diff
@@ -307,7 +295,6 @@ def submit(self, message: Message[KafkaPayload]) -> None:
 
     def close(self) -> None:
         self.__closed = True
-        self.__next_step.close()
```
fpacifici, line 310: This was wrong. The close method should not close the following ones as the partition revoke calls join which can cause messages to be submitted.

PR #2300, opened 2021-12-17
perf(subscriptions): Multiprocessing experiment

lynnagara:
Just a quick experiment to see what happens if we use multiprocessing
for the step that encodes all of the subscription tasks.

Number of processes used comes from the `scheduler_processes` config
and defaults to 1 (i.e. no multiprocessing).

### DIFF
--- a/snuba/subscriptions/scheduler_processing_strategy.py
+++ b/snuba/subscriptions/scheduler_processing_strategy.py
@@ -5,6 +5,7 @@ import time
 from collections import deque
 from concurrent.futures import Future
 from datetime import datetime
+from multiprocessing import Pool
 from typing import (
     Callable,
     Deque,
@@ -376,6 +377,7 @@ class ProduceScheduledSubscriptionMessage(ProcessingStrategy[CommittableTick]):
         scheduled_topic_spec: KafkaTopicSpec,
         commit: Callable[[Mapping[Partition, Position]], None],
         metrics: MetricsBackend,
+        processes: int = 1,
     ) -> None:
         self.__schedulers = schedulers
         self.__encoder = SubscriptionScheduledTaskEncoder()
@@ -383,6 +385,7 @@ class ProduceScheduledSubscriptionMessage(ProcessingStrategy[CommittableTick]):
         self.__scheduled_topic = Topic(scheduled_topic_spec.topic_name)
         self.__commit = commit
         self.__metrics = metrics
+        self.__processes = processes
         self.__closed = False
 
         # Stores each tick with it's futures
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -437,7 +444,14 @@ def submit(self, message: Message[CommittableTick]) -> None:
         tasks = self.__schedulers[tick.partition].find(tick)
 
         start_encoding = time.time()
-        encoded_tasks = [self.__encoder.encode(task) for task in tasks]
+
+        if self.__processes > 1:
+            encoded_tasks = self.__multiprocessing_pool.map(
```
evanh, line 449: Should this have a timeout on it? And what happens if one task fails?

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -437,7 +444,14 @@ def submit(self, message: Message[CommittableTick]) -> None:
...
+                self.__encoder.encode, tasks
```
fpacifici, line 450: This is very unlikely to give you any improvements framed this way:
Every subscription will be pickled, then send on a socket to the process, then unpickled and same for the response. 
The pickling unpickling and lock contention will likely make things worse. 

A way you can try is to batch them into relatively small groups of messages and send batches to each process. 
More effective would be the idea of using the parallel transform step we have in arroyo that avoid the communication via socket. 

Or alternative. Send a tick to a process. The process loads the the tasks and generates all the subscriptions for that tick.

Fundamentally the idea is to reduce the inter process communication and increase the size of the task each process has to do every time. With threads things would be different (to an extent) because they all share the same memory so they can access the same data structures, but processes cannot.

fpacifici (approved):
See my comment inline, I don't think this will produce an improvement, though if you want to try out anyway to assess the overhead, feel free to go ahead. 

By the way, the CPU usage metrics will likely go down because you will be allocating more processes thus the requested millicores will increase while your usage will not increase that much. Though that won't mean that you increased the capacity.

fpacifici:
Actually I think there is a smarter way that you can try quickly 
It is inspired to how Redis does sharding: https://redis.io/topics/cluster-tutorial
You divide your subscription id space in logical partitions. One per subprocess.
Then every time you receive a tick, you schedule a job per partition. 
Each job loads the subscriptions and only encodes those of the partition it is supposed to handle.

This still does not cover the serialization of all the encoded strings back to the main process without shared memory but it could be interesting to see the results.

lynnagara:
> Actually I think there is a smarter way that you can try quickly It is inspired to how Redis does sharding: https://redis.io/topics/cluster-tutorial You divide your subscription id space in logical partitions. One per subprocess. Then every time you receive a tick, you schedule a job per partition. Each job loads the subscriptions and only encodes those of the partition it is supposed to handle.
> 
> This still does not cover the serialization of all the encoded strings back to the main process without shared memory but it could be interesting to see the results.

I like this idea, will come back to it. The other thought I had was just to do the producing in the threads as well and return nothing. The produced messages would not be in order anymore.

PR #2301, opened 2021-12-17
feat(admin) Add authorization provider in snuba admin

fpacifici:
The admin relies on a proxy to perform authentication and authorization.
This is not great for two reasons:
- it would by default allow everybody if the proxy was misconfigured while failing open (blocking everybody) would be safer.
- it does not allow to decide which features to show depending on roles.

This adds an abstraction to performa authorization before each server request on the admin UI. 
This abstraction has multiple implementations for multiple authorization providers. The default is NOOP and allows everything through. The next to be implemented is IAP, then we can add one for basic HTTP auth useful on prem.


evanh (approved):

snuba/admin/auth.py
```diff
@@ -0,0 +1,39 @@
...
+auth_provider = Callable[[], str]
+
+# This function takes the Flask request and authorizes it.
+# If the request is valid it would return the user id.
+# If not it will raise UnauthorizedException
+#
+# TODO: provide a more structured representation of the User that
+# includes the role at least.
+def authorize_request() -> str:
+    provider_id = state.settings.ADMIN_AUTH_PROVIDER
```
lynnagara, line 21: Don't think you need to go through state module for this
```suggestion
    provider_id = snuba.settings.ADMIN_AUTH_PROVIDER
```

fpacifici: right I changed it from runtime config to settings sand forgot to fix that

lynnagara (approved):
Thanks, this will be useful

PR #2302, opened 2021-12-17
feat(subscriptions): Update command args to accept metrics entities

ahmedetefy:
Adds to the subscriptions_executor and subscriptions_scheduler
commands the supported metrics entity choices


lynnagara:
@ahmedetefy This is not really ready for use yet, do you need these for a particular reason? Or just making sure these options are eventually available.

ahmedetefy:
@lynnagara 
Yep, just making sure these options are available eventually.

lynnagara (approved):

PR #2304, opened 2021-12-17
feat(tracing) Capture trace output in clickhouse result

evanh:
Using the existing capture trace field, add a log handler to the clickhouse driver
log file to capture the trace for the query and return it in the result. Since
this is a debug feature only, also discard the results of the query to avoid
exposing any PII.


snuba/clickhouse/native.py
```diff
@@ -20,16 +23,39 @@
...
 logger = logging.getLogger("snuba.clickhouse")
+trace_logger = logging.getLogger("clickhouse_driver.log")
+trace_logger.setLevel("INFO")
 
 Params = Optional[Union[Sequence[Any], Mapping[str, Any]]]
 
 metrics = MetricsWrapper(environment.metrics, "clickhouse.native")
 
 
+ClickhouseProfile = TypedDict(
```
vladkluev, line 34: nit: why not a dataclass?

snuba/clickhouse/native.py
```diff
@@ -103,14 +130,42 @@ def execute(
...
+                            else:
+                                settings = {"send_logs_level": "trace"}
+                            result_data = query_execute()
+                            # In order to avoid exposing PII the results are discarded
+                            result_data = [[], []] if with_column_types else []
+                            trace_output = buffer.getvalue()
+                    else:
+                        result_data = query_execute()
+
+                    profile_data: ClickhouseProfile = {
```
vladkluev, line 148: so this is always available whether we capture the trace or not?

tests/subscriptions/test_worker.py
```diff
@@ -244,6 +244,7 @@ def test_subscription_worker(
         future_result = request, result = future.result()
         assert message.payload.task.timestamp == timestamp
         assert message.payload == SubscriptionTaskResult(task, future_result)
+        del result["profile"]
```
vladkluev, line 247: why did this test need to be modified?

vladkluev (commented):
Can we add a test?

tests/subscriptions/test_worker.py
evanh: Because it checks the result exactly on line 270. Although I had to change it there anyways because there's also trace output now.

snuba/clickhouse/native.py
evanh: Because it gets converted to a dictionary in `NativeReader` and it felt like a lot of unnecessary conversion.

snuba/clickhouse/native.py
evanh: Yes because I want to use tracing data for attribution on all queries.

vladkluev (approved):

PR #2305, opened 2021-12-17
fix(pipeline): Remove capture trace flag from Request

rahul-kumar-saini:
Removed the flag from RequestSettings, now you must directly use the ClickHouse methods to use the flag.


vladkluev:
so what's the way to do it directly with clickhouse methods?

rahul-kumar-saini:
We can use the ClickHouse driver, specifically when we call [`snuba/clickhouse/native.py::NativeDriveReader.execute()`](https://github.com/getsentry/snuba/blob/1c7d46ed6551a7e01287c4d85bd110c46a497ec8/snuba/clickhouse/native.py#L307-L315), and pass the `capture_trace` flag here.

evanh (approved):

PR #2308, opened 2021-12-17
Relax bigquery querylog anonymization

vladkluev:
### Problem
The querylog is too anonymized. Dates and datetimes do not need to be scrubbed from the querylog. Integers are maybe sensitive but we shouldn't replace them with a string, we should replace them with a number to anonymize them

### Solution

Remove unnecessary anonymization code, modify/add tests


snuba/clickhouse/formatter/expression.py
```diff
@@ -212,16 +217,7 @@ def _format_string_literal(self, exp: Literal) -> str:
...
-        return "$N"
-
-    def _format_boolean_literal(self, exp: Literal) -> str:
-        return "$B"
-
-    def _format_datetime_literal(self, exp: Literal) -> str:
-        return "$DT"
-
-    def _format_date_literal(self, exp: Literal) -> str:
-        return "$D"
```
vladkluev, line 224: this diff is misleading, it looks like ` return str(_gen_random_number())` is the implementation for `format_date_literal` but I just removed `format_date_literal` and made it inherit form the non-anonymized clickhouse expression formatter.

snuba/clickhouse/formatter/expression.py
```diff
@@ -212,16 +217,7 @@ def _format_string_literal(self, exp: Literal) -> str:
...
+        return str(_gen_random_number())
```
fpacifici, line 220: What about hardcoding 0 or -1 ? Basically something that tells us for sure that this is a value we tweaked.
A random number could mislead people in thinking that it is the real value provided by the user.

snuba/clickhouse/formatter/expression.py
```diff
@@ -202,7 +202,12 @@ def _format_date_literal(self, exp: Literal) -> str:
         )
 
 
-class ClickHouseExpressionFormatterAnonymized(ClickhouseExpressionFormatterBase):
+def _gen_random_number() -> int:
+    # https://xkcd.com/221/
+    return 4
```
fpacifici, line 207: I think we should have a design meeting and set up a committee to define a process to choose this number.

fpacifici (approved):

fpacifici:
Please fix the broken tests

snuba/clickhouse/formatter/expression.py
onewland: I may have missed something but do we want anonymized queries to be runnable? 

Because if we don't care about that it seems to me like it may be more clear to see a $N rather than an arbitrary number (like 4) that you're looking at an anonymized value.

vladkluev: Short answer is: yes we do want to be able to run anonymized queries. 

The reason is that a lot of times the reason you're looking at a query is:

* Clickhouse said it was invalid
* It timed out for some reason

When that happens, I want to either:

* Copy and paste it into my local clickhouse and see if the error is reproducible
* Run an EXPLAIN on it and see what it gives me

Either way I want to be able to run it

PR #2309, opened 2021-12-17
feat(tracing) Add a UI and backend for trace data

evanh:
Add a (very rudimentary) UI that allows a query to be run and displays the trace
for that query. It will also display any error that may have arrived from
running the query.


snuba/admin/static/tracing/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+              ?.local_nodes.map((node, nodeIndex) => (
+                <option
+                  key={`${node.host}:${node.port}`}
+                  value={`${node.host}:${node.port}`}
+                >
+                  {node.host}:{node.port}
+                </option>
+              ))}
+          </select>
+        )}
```
onewland, line 100: If this is totally common to the system query selector, we should probably try extracting that to a common file/component. Maybe Lyn knows more about the right way to do this in React

snuba/clickhouse/native.py
```diff
@@ -109,6 +109,9 @@ def execute(
         failures.
         """
         try:
+            # if capture_trace:
+            #     conn = self._create_conn(send_logs_level="trace")
+            # else:
```
lynnagara, line 114: nit: Remove

snuba/admin/views.py
```diff
@@ -116,6 +121,54 @@ def clickhouse_system_query() -> Response:
...
+    try:
+        host = req["host"]
+        port = req["port"]
+        storage = req["storage"]
+    except KeyError:
+        return make_response(jsonify({"error": "Invalid request"}), 400)
+
+    try:
+        raw_sql = req["sql"]
+    except KeyError:
```
onewland, line 141: why have a second `except keyError` here that has the same error response?

tests/admin/test_api.py
```diff
@@ -71,3 +71,20 @@ def test_post_configs(admin_api: Any) -> None:
...
+        "/clickhouse_trace_query",
+        data=json.dumps(
+            {
+                "host": "localhost",
+                "port": 9000,
+                "storage": "errors_ro",
+                "sql": "SELECT count() FROM errors_local",
+            }
+        ),
+    )
```
onewland, line 87: Nice job adding a test, I didn't realize these endpoints were in the tests

snuba/admin/static/tracing/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <button onClick={(_) => executeQuery()}>Execute query</button>
+        )}
+      </form>
+      <div>
+        <h2>Query results</h2>
+        <Table
+          headerData={['Query', 'Response', 'Error']}
+          rowData={queryResultHistory.map((queryResult) => [
+            <span>{queryResult.input_query}</span>,
+            <span>{queryResult.trace_output || ''}</span>,
```
onewland, line 117: Is the trace output "\n" delimited? Because we may want to replace that with `<br/>`

snuba/clickhouse/native.py
```diff
@@ -256,7 +261,14 @@ def execute_robust(
...
                 raise ClickhouseError(e.message, code=e.code) from e
 
-    def _create_conn(self) -> Client:
+    def _create_conn(self, **kwargs: str) -> Client:
+        if kwargs:
+            client_settings: Mapping[str, str] = {
+                **self.client_settings,
+                **kwargs,
+            }
+        else:
```
lynnagara, line 270: Why this change? I don't see where we are passing any kwargs into here

snuba/admin/static/api_client.tsx
```diff
@@ -95,14 +95,17 @@ function Client() {
...
-      const url = baseUrl + "clickhouse_queries";
+      const url = baseUrl + 'clickhouse_queries';
       return fetch(url).then((resp) => resp.json());
     },
-    executeQuery: (query: QueryRequest) => {
-      const url = baseUrl + "run_clickhouse_system_query";
+    executeQuery: (query: QueryRequest, endpoint?: string) => {
+      const endpoint_to_use = endpoint
+        ? endpoint
+        : 'run_clickhouse_system_query';
```
onewland, line 104: It's not super important but I'd almost prefer this to require the argument (and make the change in the other file)

snuba/admin/views.py
```diff
@@ -116,6 +121,54 @@ def clickhouse_system_query() -> Response:
...
+#  http://localhost:1219/clickhouse_trace_query?query=SELECT+count()+FROM+errors_local
+@application.route("/clickhouse_trace_query", methods=["POST"])
+def clickhouse_trace_query() -> Response:
+    req = json.loads(request.data)
+    try:
+        host = req["host"]
+        port = req["port"]
+        storage = req["storage"]
+    except KeyError:
+        return make_response(jsonify({"error": "Invalid request"}), 400)
```
onewland, line 137: For these `KeyError`s, could we say which key is missing?

snuba/admin/clickhouse/common.py
```diff
@@ -0,0 +1,62 @@
...
+class InvalidStorageError(SerializableException):
+    pass
+
+
+def is_valid_node(host: str, port: int, cluster: ClickhouseCluster) -> bool:
+    connection_id = cluster.get_connection_id()
+    return host == connection_id.hostname and port == connection_id.tcp_port
+
+
+CONNECTIONS: MutableMapping[StorageKey, ClickhousePool] = {}
```
lynnagara, line 28: Is there anything that has changed here apart from holding onto the connections here?

lynnagara (commented):
Just a couple of questions

onewland (commented):
Looks good to me, outside of a couple comments.

Were you able to test this locally e.g. with the curl command in the comment?

snuba/admin/clickhouse/common.py
lynnagara: Do we actually want to hold onto these connections here?

snuba/admin/static/api_client.tsx
onewland: also for this pattern you can shorten it to `endpoint ?? 'run_clickhouse_system_query'`

snuba/admin/clickhouse/common.py
fpacifici: It is ok, now we have metrics about the connections number and we know that flask will kill the process (and thus the connections) periodically. So it is not an issue.

snuba/admin/static/tracing/index.tsx
rahul-kumar-saini: Done 💯

snuba/admin/static/components/query_display/query_display.tsx
```diff
@@ -0,0 +1,198 @@
...
+import { Table } from "../../table";
+import { COLORS } from "../../theme";
+
+import { ClickhouseNodeData, QueryRequest, QueryResult } from "./types";
+
+type QueryState = Partial<QueryRequest>;
+
+function QueriesDisplay(props: {
+  api: Client;
+  endpoint: string;
```
onewland, line 12: nice job moving more things to props

snuba/admin/static/components/query_display/types.tsx
```diff
@@ -24,6 +24,8 @@ type QueryResult = {
   timestamp: number;
   column_names: QueryResultColumnMetadata;
   rows: [QueryResultRow];
+  trace_output?: string;
+  error?: string;
```
lynnagara, line 28: I would not try and make every kind of API response fit into this QueryResult type by making lots of stuff optional.
I think it is better to have 3 separate and stricter types.
- SystemQueryResult
- TracingQueryResult (this one has the trace_output)
- QueryResultError (this one only has the error key)

snuba/admin/static/tracing/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (queryResult.error === undefined) {
+      elements = { Trace: [queryResult.trace_output, 400] };
+    } else {
+      elements = { Error: [queryResult.error, 200] };
+    }
+    return tracingOutput(elements);
+  }
+
+  function tracingOutput(elements: Object) {
+    console.log(elements);
```
lynnagara, line 18: Remove

snuba/admin/static/tracing/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import Client from "../api_client";
+import QueriesDisplay from "../components/query_display/query_display";
+import { QueryResult } from "../components/query_display/types";
+
+function TracingQueries(props: { api: Client }) {
+  function tablePopulator(queryResult: QueryResult) {
+    var elements = {};
+    if (queryResult.error === undefined) {
```
lynnagara, line 9: The proper way to check for undefined in javascript is:
`typeof(query.result.error) === "undefined"`
Or if you only need to know whether it's falsy can just be the easier:
`if (queryResult.error)`

snuba/admin/static/tracing/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+      elements = { Trace: [queryResult.trace_output, 400] };
```
lynnagara, line 10: I think we shouldn't be checking the API response code in the components like this. It's supposed to be the job of `api_client` to check the response code and reject with the right message on error

lynnagara (changes requested):
This PR is quite large and involves a lot of refactoring that touches other functionality including core ClickHouse driver functionality. It's also hard to figure out what's changed and what's just moved around. Please consider splitting it up.

Currently you must run `make build-admin` manually before pushing the bundle.js file as we do not build it in CI for prod and rely on the checked in version of this file. Currently you have pushed the dev version of the bundle.

snuba/admin/views.py
```diff
@@ -116,6 +121,52 @@ def clickhouse_system_query() -> Response:
...
+        result = run_query_and_get_trace(host, port, storage, raw_sql)
+        trace_output = result.trace_output
+        return make_response(jsonify({"trace_output": trace_output}), 200)
+    except InvalidCustomQuery as err:
+        return make_response(jsonify({"error": err.message or "Invalid query"}), 400)
+    except QueryException as err:
+        status = 500
+        details: Mapping[str, Any]
+
+        cause = err.__cause__
```
lynnagara, line 152: Are you sure this how QueryException is meant to work? It doesn't seem right to get the message out of this private attribute

snuba/admin/views.py
```diff
@@ -116,6 +121,52 @@ def clickhouse_system_query() -> Response:
...
+
+    try:
+        result = run_query_and_get_trace(host, port, storage, raw_sql)
+        trace_output = result.trace_output
+        return make_response(jsonify({"trace_output": trace_output}), 200)
+    except InvalidCustomQuery as err:
+        return make_response(jsonify({"error": err.message or "Invalid query"}), 400)
+    except QueryException as err:
+        status = 500
+        details: Mapping[str, Any]
```
lynnagara, line 150: Why Any? It could be Mapping[str, str] or even a typed dict

snuba/admin/clickhouse/tracing.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Raises InvalidCustomQuery if query is invalid or not allowed.
+    """
+    sql_query = " ".join(sql_query.split())
+    lowered = sql_query.lower().strip()
+
+    if not lowered.startswith("select"):
+        raise InvalidCustomQuery("Only SELECT queries are allowed")
+
+    disallowed_keywords = ["insert", "join", ";"]
```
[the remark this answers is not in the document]
lynnagara: I also think we should allow join for tracing

snuba/admin/static/components/query_display/types.tsx
evanh: I don't think I agree to this. If we make different return types for different functions, then the common QueryDisplay component has to also know which function is being called and the type it returns, which means that has to be passed in the props and is going to make the nice common component Rahul created a lot messier. 

Also these types are not independent, a system query result could have a tracing output and/or an error. This type maps to the response from Snuba, which can have any combination of these fields. Saying that one call to Snuba will necessarily only have a subset of fields isn't strictly true.

snuba/admin/static/tracing/index.tsx
evanh: I think Lyn is right in that error checking can happen in the QueryDisplay component. I don't think tracing and system queries should handle that differently.

snuba/admin/clickhouse/common.py
```diff
@@ -0,0 +1,66 @@
...
+
+def is_valid_node(host: str, port: int, cluster: ClickhouseCluster) -> bool:
+    nodes = cluster.get_local_nodes()
+    return any(node.host_name == host and node.port == port for node in nodes)
+
+
+CONNECTIONS: MutableMapping[StorageKey, ClickhousePool] = {}
+
+
+def get_clickhouse_connection(
```
nikhars, line 31:
```suggestion
def get_clickhouse_readonly_connection(
```
It makes it clear that noone can do do inserts/alters when they use this API

snuba/admin/clickhouse/common.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+    database = cluster.get_database()
+    connection = ClickhousePool(
+        clickhouse_host,
+        clickhouse_port,
+        settings.CLICKHOUSE_READONLY_USER,
+        settings.CLICKHOUSE_READONLY_PASSWORD,
+        database,
+        # force read-only
```
nikhars, line 62: You might also want to set `max_pool_size` parameter to something smaller than used in production. snuba-admin won't need so many connections.

nikhars (commented):
Left some minor comments

snuba/admin/clickhouse/common.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        storage_key = StorageKey(storage_name)
+    except ValueError:
+        raise InvalidStorageError(
+            f"storage {storage_name} is not a valid storage name",
+            extra_data={"storage_name": storage_name},
+        )
+
+    if storage_key in CONNECTIONS:
+        return CONNECTIONS[storage_key]
```
fpacifici, line 44: I don't think this will work. 

The issue is that there are two separate use cases, you cannot use the same logic to retrieve a connection for the tracing tool (that needs to talk to a query node through envoy) and for the system tables tool (which needs to access individual nodes and uses the storage key just to fetch the right cluster to validate the host name/port).

Specifically:
Here the system table tool will create a connection to the required host the first time and cache it. Subsequently it will always return the same connection to the same host for the same storage, no matter which host/port the user wants. Which is wrong. If you want to cache connections for the system tables tool, the key must contain at least the host name as well.

I think you want two methods, one to return a query connection to a cluster given a storage key. And another one to take host/port and validates they are part of a storage and then return a connection to those. 
The system tables tool would use the first. The tracing tool would use the second (or both if you want to get a trace from a query ran on a storage node)

fpacifici (changes requested):

snuba/admin/clickhouse/common.py
evanh: I split the connection functions into two, one for system queries that cache by hostname, and one for tracing queries that cache by cluster.

nikhars:
LGTM

snuba/admin/clickhouse/tracing.py
```diff
@@ -0,0 +1,27 @@
...
+    disallowed_keywords = ["insert", ";"]
+    for kw in disallowed_keywords:
+        if kw in lowered:
+            raise InvalidCustomQuery(f"{kw} is not allowed in the query")
+
+
+def run_query_and_get_trace(storage_name: str, query: str) -> ClickhouseResult:
+    validate_trace_query(query)
+    connection = get_ro_cluster_connection(storage_name)
+    query_result = connection.execute(query=query, capture_trace=True)
```
fpacifici, line 26: This will run the query on one single thread. Which is good to start with.
Though later we will want to add a switch to use 1 thread vs 2 or 4 to test how the query parallelizes.

fpacifici (approved):
Fyi: the readonly user is not present on the query nodes. So you will need to send a request to create it. @nikhars  could help you with this.

PR #2310, opened 2021-12-17
feat(admin): Replace nested table with raw JSON and copy to clipboard button

lynnagara:
This is what the UI looks like after this change

Screen Shot 2021-12-17 at 3 14 17 pm


onewland (approved):
Would be nice if the JSON was indented/highlighted but not that important since there's a quick copy/paste

PR #2311, opened 2021-12-18
feat(admin): Send SQL queries from the UI to system tables

lynnagara:
Adds UI components to send freeform SQL queries from the UI to
any system table. Removes UI components related to the canned queries.

Looks like this now:
Screen Shot 2021-12-17 at 4 14 44 pm


snuba/admin/static/api_client.tsx
```diff
@@ -94,17 +92,19 @@ function Client() {
...
         method: "POST",
         body: JSON.stringify(query),
-      }).then((resp) => resp.json());
+      }).then((resp) => {
+        if (resp.ok) {
+          return resp.json();
+        } else {
+          return resp.json().then(Promise.reject.bind(Promise));
+        }
+      });
```
lynnagara, line 107: This looks terrible but it seems to work. Should probably all be async/await

onewland: I think we need to add some UI container for displaying errors in general, and then make HTTP request logic more generic so every API call writes to the same error `div` or whatever.

This is perfectly good for what we already have, though

snuba/admin/static/clickhouse_queries/index.tsx
```diff
@@ -156,6 +152,40 @@ const jsonStyle = {
...
+  onChange: (nextValue: string) => void;
+}) {
+  const { value, onChange } = props;
+  return (
+    <textarea
+      spellCheck={false}
+      value={value}
+      onChange={(evt) => onChange(evt.target.value)}
+      style={{ width: "100%", height: 100 }}
+      placeholder={"Write your query here"}
```
onewland, line 186: maybe this placeholder could be a valid query? like the one in your screenshot

onewland (approved):

snuba/admin/static/clickhouse_queries/index.tsx
lynnagara: I'm just going to merge this as-is. Let's build canned queries soon 🙂

PR #2315, opened 2021-12-20
fix: Prevent all subscriptions from being logged as non-matching

lynnagara:
We added "profile" to the ClickHouse result which contains information
about the query execution that will never actually be the same. Strip this
key before diffing the two queries and logging discrepancies to Sentry.


snuba/subscriptions/worker.py
```diff
@@ -252,4 +253,8 @@ def handle_nan(result: Result) -> Result:
             except TypeError:
                 pass
 
+    # Strip the profile information as timings will not be the same
+    if "profile" in result_copy:
+        del result_copy["profile"]
```
vladkluev, line 258: this seems the wrong way to go about this. We're massaging the return payload so we can `==` it. Why not just write the comparison function to do what we want and keep the return payload the same?

lynnagara: This is what we are doing. We are making a copy of `result` - not modifying the original

vladkluev (approved):

PR #2318, opened 2021-12-21
add retries to redis initialization

vladkluev:
This makes our CI fail fairly often. Add some retries to give time for the cluster to come up


snuba/redis.py
```diff
@@ -31,24 +39,44 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
-    )
+def _retry(func: Callable[[], RedisClientType]) -> Callable[[], RedisClientType]:
+    @wraps(func)
+    def wrapper() -> RedisClientType:
+        retry_counter = 3
+        while retry_counter > 0:
+            try:
+                return func()
+            except RedisClusterException as e:
+                if "All slots are not" in str(e) and retry_counter >= 0:
```
onewland, line 50: I don't love text matching in an error message but the stakes are low here if we can't check a class type or against a constant

snuba/redis.py
```diff
@@ -31,24 +39,44 @@ def execute_command(self, *args: Any, **kwargs: Any) -> Any:
...
-        host=settings.REDIS_HOST,
-        port=settings.REDIS_PORT,
-        password=settings.REDIS_PASSWORD,
-        db=settings.REDIS_DB,
-        socket_keepalive=True,
-    )
+def _retry(func: Callable[[], RedisClientType]) -> Callable[[], RedisClientType]:
+    @wraps(func)
+    def wrapper() -> RedisClientType:
+        retry_counter = 3
```
onewland, line 45: nit: it's nice if magic numbers/constants are split out from logic (and even methods).

`MAX_RETRIES=3`
`KNOWN_REDIS_INITIALIZATION_NOT_FINISHED_MESSAGE = "All slots are not"`
`RANDOM_SLEEP_MAX=50`

Additionally, it would be cool if _retry could be parameterized so the annotation is something like

`@_retry(3)`

onewland (approved):
Despite my nit-picking, totally fine merging this as-is

snuba/redis.py
fpacifici: Having the amount of retries in settings also allows us to have a different behavior between self hosted/dev and prod, so please let's do that way.

docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
     ulimits:
       nofile: 262144
   redis:
-    image: redis:3.2-alpine
+    image: redis:5.0-alpine
```
fpacifici, line 33: Any specific reason for upgrading? Does this match what we run in prod/single tenant ?

vladkluev: this matches what we run in CI

PR #2319, opened 2021-12-21
Render system query results in a table

vladkluev:
Before:
Screen Shot 2021-12-20 at 4 51 03 PM

After:
Screen Shot 2021-12-20 at 4 51 12 PM


fpacifici (approved):
Screen Shot 2021-12-21 at 11 09 23 AM

onewland:
looks fine, I wonder if the ideal version is a toggle if we find both modes useful.

vladkluev:
copy to clipboard still copies the json

onewland (approved):

PR #2321, opened 2021-12-21
fix(admin): Don't log invalid system table or column errors

lynnagara:
Since the ClickHouse admin tool allows freeform SQL queries to system tables
we get a lot of noise in Sentry due to invalid columns or table names
being entered. This turns off error logging for those specific errors only.

### DIFF
--- a/snuba/admin/clickhouse/system_queries.py
+++ b/snuba/admin/clickhouse/system_queries.py
@@ -2,6 +2,8 @@ import re
 from dataclasses import dataclass
 from typing import Dict, Optional, Sequence, Type
 
+from clickhouse_driver.errors import ErrorCodes
+
 from snuba import settings
 from snuba.clickhouse.errors import ClickhouseError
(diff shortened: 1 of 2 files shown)


snuba/admin/clickhouse/system_queries.py
```diff
@@ -178,9 +179,18 @@ def run_system_query_on_host_with_sql(
...
-    return _run_sql_query_on_host(
-        clickhouse_host, clickhouse_port, storage_name, system_query_sql
-    )
+    try:
+        return _run_sql_query_on_host(
+            clickhouse_host, clickhouse_port, storage_name, system_query_sql
+        )
+    except ClickhouseError as exc:
+        TABLE_DOES_NOT_EXIST = 60
+        MISSING_COLUMN = 47
```
nikhars, line 188: Can we use the actual error codes definition from clickhouse driver instead of redefining them?

lynnagara: yes, fixed

evanh (approved):

PR #2325, opened 2021-12-22
feat(metric-alerts): Adds subscription consumer for metrics [INGEST-623]

ahmedetefy:
This PR adds subscription consumer for metrics (hidden behind a feature flag) to power crash rate alerts. Also, modifies the kafka stream loader for metrics_counters and metrics_sets to write to the same commit log topic and the same subscription results topic.

~~Pending the creation of the topics in production https://getsentry.atlassian.net/browse/OPS-1258~~ 
Topics have been created in production


snuba/datasets/storages/metrics.py
```diff
@@ -57,11 +73,22 @@
...
+        default_topic=Topic.METRICS,
+        commit_log_topic=Topic.METRICS_COMMIT_LOG,
+        subscription_scheduler_mode=SchedulingWatermarkMode.GLOBAL,
+        subscription_scheduled_topic=Topic.SUBSCRIPTION_SCHEDULED_METRICS,
+        subscription_result_topic=Topic.SUBSCRIPTION_RESULTS_METRICS,
+    )
+else:
+    counters_kafka_stream_loader = build_kafka_stream_loader_from_settings(
+        processor=CounterMetricsProcessor(), default_topic=Topic.METRICS,
+    )
```
lynnagara, line 91: Why would we want to support metrics without a commit log? This is a requirement for subscriptions

lynnagara (changes requested):
I'm not sure I get what the goal of this PR is. Why would we want to support metrics consumers without a commit log even for dev/testing? AFAIK metrics should always have a commit log associated with them so we can run subscriptions.

I think having the option to turn metrics on and off via ENABLE_METRICS_SUBSCRIPTIONS flag is useful but I don't get why we should add the third option of supporting it without a commit log. That seems like it should be a core part of metrics functionality as it is required for subscriptions.

evanh:
@fpacifici Are these changes to the topic OK to ship? @ahmedetefy Can you give some context on why you're having counters and sets use the same commit log?

ahmedetefy:
> @fpacifici Are these changes to the topic OK to ship? @ahmedetefy Can you give some context on why you're having counters and sets use the same commit log?

@evanh `metrics_sets` consumer is required for users crash rate alerts while `metric_counters` is used for sessions crash rate alerts

fpacifici (commented):
Please double check with @oioki that the commit log topic has been created and ported to Krapfen. As you merge this, the consumer will start producing messages on the commit log, so if the topic is not on the right cluster the consuemr will crash, we will backlog, etc. etc.

oioki:
> Please double check with @oioki that the commit log topic has been created and ported to Krapfen. As you merge this, the consumer will start producing messages on the commit log, so if the topic is not on the right cluster the consuemr will crash, we will backlog, etc. etc.

I can confirm that new topics are created on the Krapfen cluster:
```
$ kafkactl config use-context krapfen
$ kafkactl list topics -o wide
TOPIC                               PARTITIONS     CONFIGS
__confluent.support.metrics         1              segment.bytes=1073741824,retention.ms=31536000000
__consumer_offsets                  50             compression.type=producer,cleanup.policy=compact,segment.bytes=104857600
ingest-metrics                      64             segment.bytes=1073741824,retention.ms=86400000,max.message.bytes=10000000
metrics-subscription-results        8              segment.bytes=1073741824
scheduled-subscriptions-metrics     16             segment.bytes=1073741824,retention.ms=86400000
snuba-metrics                       64             segment.bytes=1073741824,retention.ms=86400000,max.message.bytes=10000000,message.timestamp.type=LogAppendTime
snuba-metrics-commit-log            1              compression.type=producer,min.insync.replicas=1,cleanup.policy=compact,segment.bytes=104857600,retention.ms=86400000,min.compaction.lag.ms=3600000,segment.ms=3600000
```

fpacifici (approved):

snuba/cli/devserver.py
```diff
@@ -153,6 +153,45 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
...
+                        "--dataset=metrics",
+                        "--entity=metrics_sets",
+                        "--commit-log-topic=snuba-metrics-commit-log",
+                        "--commit-log-group=metrics_group",
+                        "--delay-seconds=1",
+                        "--schedule-ttl=10",
+                        "--max-query-workers=1",
+                    ],
+                ),
+            ]
```
lynnagara, line 194: What's the relationship between these 2 consumers? They are both consuming from the same topic, same consumer group, same commit log topic, same commit log group. Isn't that going to cause issues? Should this be run as a single multistorage consumer instead?

lynnagara (approved):

PR #2326, opened 2021-12-22
feat(subscriptions): Adds metrics topics to topics validation

ahmedetefy:
Reverts #2313

### DIFF
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -1,5 +1,7 @@
 from typing import Any, Mapping, MutableMapping
 
+from snuba import settings
+
 
 def _validate_settings(locals: Mapping[str, Any]) -> None:
     if locals.get("QUERIES_TOPIC"):
@@ -36,6 +38,14 @@ def _validate_settings(locals: Mapping[str, Any]) -> None:
         "transactions-subscription-results",
         "sessions-subscription-results",
     }
+    if settings.ENABLE_METRICS_SUBSCRIPTIONS:
+        topic_names.update(
+            {
+                "snuba-metrics-commit-log",
+                "scheduled-subscriptions-metrics",
+                "metrics-subscription-results",
+            }
+        )
 
     for key in locals["KAFKA_TOPIC_MAP"].keys():
         if key not in topic_names:
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -1,7 +1,5 @@
 from typing import Any, Mapping, MutableMapping
 
-from snuba import settings
-
 
 def _validate_settings(locals: Mapping[str, Any]) -> None:
     if locals.get("QUERIES_TOPIC"):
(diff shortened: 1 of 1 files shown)


snuba/settings/validation.py
```diff
@@ -36,6 +38,14 @@ def _validate_settings(locals: Mapping[str, Any]) -> None:
         "transactions-subscription-results",
         "sessions-subscription-results",
     }
+    if settings.ENABLE_METRICS_SUBSCRIPTIONS:
```
lynnagara, line 41: You don't need to check this setting here, you can just add it to the main `topic_names` dictionary. Putting it on this list doesn't really have any effect, just allows those topics to be configured via KAFKA_TOPIC_MAP and KAFKA_BROKER_CONFIG without erroring.

lynnagara:
Nit: This PR is not actually a revert of 2313 right now since you have changed stuff so I would not describe it as such.

snuba/settings/validation.py
lynnagara: Fwiw I'm not sure why the original PR was reverted earlier - it should be harmless

ahmedetefy: I am not entirely sure as well, but it seemed that the consumers kept restarting according to what @fpacifici told me until I reverted that PR so I assumed it must be that its checking for topic names that don't exist in prod somehow?

lynnagara: I don't think so. Can you link/paste the actual error message you were seeing to confirm?

AFAIK the issue the other day was actually to do with this PR https://github.com/getsentry/snuba/pull/2222/files

If you pass a commit log topic to the consumer like that does it will automatically start publishing offsets to that topic. Is it possible the metrics commit log didn't exist in prod yet?

ahmedetefy: @lynnagara Do you reckon I should try to deploy this without the feature flag check since these are not production consumers?

lynnagara: https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py#L40-L46
These are the only lines where `topic_names` is used.. you can see there is no need for the actual topics to exist.

lynnagara: Let's confirm what happened last week. I don't think we should be putting feature flags around everything "just in case"

lynnagara:
@ahmedetefy This PR has changed quite a lot since earlier reviewed. Is there a reason it now does a lot more than changing the validation.py file?

lynnagara:
Also how come it's not based on master?

ahmedetefy:
@lynnagara Yeah it just needed to be rebased

fpacifici (commented):
Are these topics present in production now ?

oioki:
> Are these topics present in production now ?

They have been created:
https://getsentry.atlassian.net/browse/OPS-1258?focusedCommentId=38881

fpacifici (approved):
Please only deploy this change with the supervision of somebody in Searh and Storage.

PR #2327, opened 2021-12-22
Fix redundant clause optimizer not capturing experiment data

vladkluev:
### Problem
The redundant clause optimizer was capturing when it was applying the optimization but not when it wasn't. This made analyzing its effects difficult

### Solution
At the moment the optimization is supposed to be applied, add an experiment value for when it is not.

### DIFF
--- a/snuba/query/processors/mapping_optimizer.py
+++ b/snuba/query/processors/mapping_optimizer.py
@@ -268,6 +268,8 @@ class MappingOptimizer(QueryProcessor):
                     if requested_tag in tag_eq_match_strings:
                         if should_apply_optimization:
                             query.add_experiment("redundant_clause_removed", 1)
+                            # the clause is redundant, thus we continue the loop
+                            # and do not add it to useful_conditions
                             continue
                         else:
                             query.add_experiment("redundant_clause_removed", 0)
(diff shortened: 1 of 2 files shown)


snuba/query/processors/mapping_optimizer.py
```diff
@@ -264,10 +266,17 @@ def _get_condition_without_redundant_checks(
...
+                            query.add_experiment("redundant_clause_removed", 1)
+                            # the clause is redundant, thus we continue the loop
+                            # and do not add it to useful_conditions
+                            continue
+                        else:
+                            query.add_experiment("redundant_clause_removed", 0)
                 useful_conditions.append(
-                    self._get_condition_without_redundant_checks(cond, query)
+                    self._get_condition_without_redundant_checks(
+                        cond, query, should_apply_optimization
```
onewland, line 278: Not suggesting this should change now but I think if one more argument were added (especially a bool, Query, or Expression), it might be good to specify the argument names in this call, since the order isn't really semantic in any way.

onewland (approved):

snuba/query/processors/mapping_optimizer.py
vladkluev: agreed. Hopefully the experiment will be successful and I'll remove this entirely!

PR #2328, opened 2021-12-22
feat(admin): Improve admin notifications

lynnagara:
A few improvements to admin notifications
- Updating an existing config now sends a notification
- Creating a new config now includes the old value in the nofication message
- Add Sentry logging for Slack client

### DIFF
--- a/snuba/admin/notifications/slack/client.py
+++ b/snuba/admin/notifications/slack/client.py
@@ -31,8 +31,19 @@ class SlackClient(object):
         except Exception as exc:
             logger.error(exc, exc_info=True)
 
-        # todo: slack is annoying be a 200 could still be a failed case
-        # you have to check the "ok" param in the response, so we should
-        # check for ok: False and log those failures too.
-        if resp.status_code != 200:
-            logger.error("Slack error: {resp.content}")
+            # Slack error handling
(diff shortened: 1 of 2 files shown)


snuba/admin/notifications/slack/client.py
```diff
@@ -25,12 +28,11 @@ def post_message(
...
-            return
+        except Exception as exc:
+            logger.error(exc, exc_info=True)
 
         # todo: slack is annoying be a 200 could still be a failed case
         # you have to check the "ok" param in the response, so we should
         # check for ok: False and log those failures too.
-        print(resp.status_code)
-        print(resp.content)
+        if resp.status_code != 200:
```
MeredithAnya, line 37: So slack is annoying in that they return 200s even where there is an error. So you have to check th `ok` param to know whether there was an error or not. I think we wanna have some logic that is similar to: https://github.com/getsentry/sentry/blob/601f829c9246ae73c8169510140fd7f47fc6dfc3/src/sentry/integrations/slack/client.py#L36-L53
so that we can log the actual error

vladkluev: ☝️  That's great context to put in a comment!

lynnagara: Meh. Ok copied that code snippet over.

vladkluev (commented):
The change looks good to me as long as Meredith's comment is a addressed

vladkluev (approved):
oop I see it was

MeredithAnya (approved):

PR #2329, opened 2022-01-03
fix(arm64) Fix tests so they work on arm64 as well

evanh:
This is mostly fixing a bug with our current tests, where in the old version of
Clickhouse the events are not being merged before the tests run. In the new
version of Clickhouse the events are merged before the tests, and since the
fixtures all use the same span_id, they are actually only sending one event.

### DIFF
--- a/tests/test_transactions_api.py
+++ b/tests/test_transactions_api.py
@@ -70,7 +70,7 @@ class TestTransactionsApi(BaseApiTest):
                 # project N sends an event every Nth minute
                 if tock % p == 0:
                     trace_id = "7400045b25c443b885914600aa83ad04"
-                    span_id = "8841662216cc598b"
+                    span_id = f"8841662216cc598b{tock}"[-16:]
                     processed = (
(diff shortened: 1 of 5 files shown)


tests/test_transactions_api.py
```diff
@@ -303,7 +304,7 @@ def test_column_formatting(self) -> None:
         assert response.status_code == 200, response.data
         assert len(data["data"]) == 1
         assert data["data"][0]["event_id"] == first_event_id
-        assert data["data"][0]["span_id"] == "8841662216cc598b"
+        assert data["data"][0]["span_id"] == "841662216cc598b1"
```
onewland, line 307: could we maybe extract this to a constant, to make it more straightforward if we want to update it again?

onewland (approved):

tests/test_transactions_api.py
evanh: Since the number actually changes (this happens to be the first) it's not possible to make it a true constant.

PR #2330, opened 2022-01-06
Roll out redundant tag optimization to all queries

vladkluev:
This optimization has shown to be effective. Full report [here](https://www.notion.so/sentry/Removing-Redundant-Tag-Checks-From-Queries-7267cac97649414d9fee4b102d55c8a7)

This PR removes the experiment code and adds a tag to the transaction when the tags hashmap optimization is applied.


evanh (approved):
Nicely done!

PR #2331, opened 2022-01-07
feat(ops): add transactions_ro storage support

onewland:
Add a read-only mode to transactions via a new `transactions_ro` storage.

It should be settable via a global flag or by a referrer list in `snuba.settings`.

https://getsentry.atlassian.net/browse/SNS-990

### DIFF
--- a/snuba/datasets/storages/transactions_ro.py
+++ b/snuba/datasets/storages/transactions_ro.py
@@ -12,7 +12,7 @@ from snuba.datasets.storages.transactions import (
 schema = TableSchema(
     columns=columns,
     local_table_name="transactions_local",
-    dist_table_name="transactions_dist_ro",
+    dist_table_name="transactions_dist",
     storage_set_key=StorageSetKey.TRANSACTIONS_RO,
     mandatory_conditions=[],
 )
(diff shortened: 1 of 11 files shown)


snuba/datasets/storages/transactions_ro.py
```diff
@@ -0,0 +1,26 @@
...
+    mandatory_conditions=mandatory_conditions,
+)
+
+storage = ReadableTableStorage(
+    storage_key=StorageKey.TRANSACTIONS_RO,
+    storage_set_key=StorageSetKey.TRANSACTIONS_RO,
+    schema=schema,
+    query_processors=query_processors,
+    query_splitters=query_splitters,
+)
```
fpacifici, line 26: These are not the right parameters for the transactions storage.
They need to match those in `storages/transactions.py`.
These seem to match those in errors_ro.

snuba/datasets/storages/transactions_ro.py
```diff
@@ -0,0 +1,27 @@
...
+    columns,
+    mandatory_condition_checkers,
+    query_processors,
+    query_splitters,
+)
+
+schema = TableSchema(
+    columns=columns,
+    local_table_name="transactions_local",
+    dist_table_name="transactions_dist_ro",
```
fpacifici, line 15: Let's keep the same name for now.  `transactions_dist`. So it will work independently on which cluster we put the storage on.

fpacifici (commented):
Please fix the table name. 
Now this must be merged after the change to the prod configuration to declare the storage set.

fpacifici (approved):
Thanks, 
I guess https://github.com/getsentry/snuba/pull/2334 goes first, right ? (or at the same time).

PR #2333, opened 2022-01-08
feat(snql) Support higher order functions in SnQL

evanh:
SnQL now supports using lambdas in its clauses. Lambdas are defined as a tuple
of identifiers, followed by an arrow, followed by a function that may or may not
use those identifiers. 

```
(`x`, `y`) -> foo(`x`, `y`)
```

An identifier in SnQL is a set of characters enclosed in
backticks. The distinction to use backticks was to make parsing and validating
easier, since the parser doesn't have to try and determine if a string is a
column or an identifier.

This feature set is slightly more limited than what is available in Clickhouse.
The major difference is that a lambda can only be used as a parameter in a
function, and the transformation of a lambda must always be a function itself.
```
-- Not valid since lambdas must be inside functions
SELECT (`x`) -> foo(`x`) ... 

-- Not valid since lambdas must have functions
SELECT arrayMap((`x`) -> `x`, ...
```

This change makes using higher order functions such as arrayMap possible, since
those functions require a lambda to be defined to use them.

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 from dataclasses import replace
 from datetime import datetime, timedelta
@@ -159,10 +161,11 @@ snql_grammar = Grammar(
 
     low_pri_op            = "+" / "-"
     high_pri_op           = "/" / "*"
-    param_expression      = low_pri_arithmetic / quoted_literal
+    param_expression      = low_pri_arithmetic / quoted_literal / identifier
     parameters_list       = parameter* (param_expression)
-    parameter             = param_expression space* comma space*
+    parameter             = (lambda / param_expression) space* comma space*
     function_call         = function_name open_paren parameters_list? close_paren (open_paren parameters_list? close_paren)? (space+ "AS" space+ alias_literal)?
+    lambda                = open_paren space* identifier (comma space* identifier)* space* close_paren space* arrow space* function_call
 
     aliased_tag_column    = tag_column space+ "AS" space+ alias_literal
     aliased_subscriptable = subscriptable space+ "AS" space+ alias_literal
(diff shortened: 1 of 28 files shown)


evanh:
@vladkluev I added examples, hope that makes it clearer.

tests/fixtures.py
```diff
@@ -253,6 +253,8 @@ def get_raw_transaction(span_id: str | None = None) -> InsertEvent:
                     "description": "SELECT * FROM users",
                     "data": {},
                     "timestamp": calendar.timegm(end_timestamp.timetuple()),
+                    "hash": "5029609156d8133",
+                    "exclusive_time": 1.2,
```
vladkluev, line 257: is this related to your change?

vladkluev (approved):
2 questions but not blocking. well done!

tests/fixtures.py
evanh: Yes, I needed this data in here to test the spans query, which is the use case for this change.

tests/query/snql/test_invalid_queries.py
```diff
@@ -80,6 +80,16 @@
...
         id="aliases are only in the select",
     ),
+    pytest.param(
+        "MATCH (discover_events) SELECT arrayMap((x) -> identity(`x`), sdk_integrations) AS sdks WHERE project_id = 1 AND timestamp >= toDateTime('2021-01-01') AND timestamp < toDateTime('2021-01-02')",
+        "Parsing error on line 1 at 'ap((x) -> ide'",
+        id="identifiers have backticks",
+    ),
+    pytest.param(
+        "MATCH (discover_events) SELECT arrayMap((`x`) -> `x`, sdk_integrations) AS sdks WHERE project_id = 1 AND timestamp >= toDateTime('2021-01-01') AND timestamp < toDateTime('2021-01-02')",
+        "Parsing error on line 1 at 'ap((`x`) -> `'",
```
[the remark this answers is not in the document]
evanh:
```
arrayMap((`x`) -> `x`
      {          }^
           ^      |
prefix to error   error happens here since the identifier is not a function
```

The parsing error is based on parsimonious (which doesn't have great messages) and is appending an arbitrary prefix from the original string to provide context. The `ap` is the end of `arrayMap`.

snuba/query/snql/parser.py
```diff
@@ -1119,6 +1161,33 @@ def mangle_column_value(exp: Expression) -> Expression:
         query.transform_expressions(mangle_column_value)
 
 
+def validate_identifiers_in_lambda(
```
vladkluev, line 1164: one thing I remembered before you merge. You don' test the recursive case of this function, you just have the [one test](https://github.com/getsentry/snuba/pull/2333/files#diff-4876cb16271701ca5a31e45e8df4ccaccc894b2898686d5184b756d1125dd4a9R20-R24). You may want to do that

evanh: Thanks for the suggestion, I added a new test and tweaked the code as a result.

PR #2337, opened 2022-01-11
On Freight deploy, update k8s deployments containing `events-subscriptions-executor` container

oioki:
Similar to https://github.com/getsentry/snuba/pull/2278 https://github.com/getsentry/snuba/pull/2249

### DIFF
--- a/.freight.yml
+++ b/.freight.yml
@@ -56,6 +56,8 @@ steps:
     name: loadtest-outcomes-consumer
   - image: us.gcr.io/sentryio/snuba:{sha}
     name: loadtest-loadbalancer-outcomes-consumer
+  - image: us.gcr.io/sentryio/snuba:{sha}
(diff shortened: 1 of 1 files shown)


evanh (approved):
This is fine, but as Filippo mentioned in Slack this isn't used in prod yet.

fpacifici:
The executor is in a crash loop now, I believe it is because it fails to commit on kafka and we went out of retention. 
If we merge this it will prevent us from deploying. So let's keep it open so it is not redeployed and I will find some time to clean it up this week.

fpacifici (changes requested):
See comment above. Let's not merge this until the executor is fixed.

lynnagara:
Sorry, I didn't see this PR earlier before making a duplicate. Closing because it's been added in https://github.com/getsentry/snuba/pull/2350

PR #2339, opened 2022-01-12
Enforce uniq in HAVING and SELECT

vladkluev:
### Context
This is a hacky query processor put in for the purpose of
mitigating clickhouse crashes. When a query is sent to clickhouse
which has a `uniq` in the `HAVING` clause but not in the `SELECT`
clause, the query node will crash. This may be fixed with future versions
of clickhouse but we have not upgraded yet. This exists as a check so that we
return an exception to the user but don't crash the clickhouse query node

### Solution

* add query processor to catch this case, make it throw an exception
* Add query processor to transaction storage 

### Feedback desired
* General approach, is it acceptable?
* Any other cases I should test for?
* What's a good way to dry run it? My thought is to send the error to sentry when it happens, hide this behind a runtime config

### DIFF
--- a/snuba/query/expressions.py
+++ b/snuba/query/expressions.py
@@ -79,6 +79,12 @@ class Expression(_Expression, ABC):
         else:
             return super().__repr__()
 
+    def functional_eq(self, other: Expression) -> bool:
+        """Returns if an expression is functionally equivalent to the other. i.e. performs an equality
+        operation ignoring aliases
+        """
+        raise NotImplementedError
+
 
 class ExpressionVisitor(ABC, Generic[TVisited]):
     """
@@ -270,6 +276,11 @@ class Literal(Expression):
     def accept(self, visitor: ExpressionVisitor[TVisited]) -> TVisited:
         return visitor.visit_literal(self)
 
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        return self.value == other.value
+
 
 @dataclass(frozen=True, repr=_AUTO_REPR)
 class Column(Expression):
@@ -288,6 +299,14 @@ class Column(Expression):
 
     def accept(self, visitor: ExpressionVisitor[TVisited]) -> TVisited:
         return visitor.visit_column(self)
+
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        return (
+            self.table_name == other.table_name
+            and self.column_name == other.column_name
+        )
 
 
 @dataclass(frozen=True, repr=_AUTO_REPR)
@@ -325,6 +344,13 @@ class SubscriptableReference(Expression):
         for sub in self.key:
             yield sub
         yield self
+
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        return self.column.functional_eq(other.column) and self.key.functional_eq(
+            other.key
+        )
 
 
 @dataclass(frozen=True, repr=_AUTO_REPR)
@@ -375,6 +401,20 @@ class FunctionCall(Expression):
     def accept(self, visitor: ExpressionVisitor[TVisited]) -> TVisited:
         return visitor.visit_function_call(self)
 
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        if self.function_name != other.function_name:
+            return False
+        if len(self.parameters) != len(other.parameters):
+            return False
+        for i, param in enumerate(self.parameters):
+            other_parameter = other.parameters[i]
+            params_functionally_equivalent = param.functional_eq(other_parameter)
+            if not params_functionally_equivalent:
+                return False
+        return True
+
 
 @dataclass(frozen=True, repr=_AUTO_REPR)
 class CurriedFunctionCall(Expression):
@@ -422,6 +462,18 @@ class CurriedFunctionCall(Expression):
     def accept(self, visitor: ExpressionVisitor[TVisited]) -> TVisited:
         return visitor.visit_curried_function_call(self)
 
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        if not self.internal_function.functional_eq(other.internal_function):
+            return False
+        for i, param in enumerate(self.parameters):
+            other_parameter = other.parameters[i]
+            params_functionally_equivalent = param.functional_eq(other_parameter)
+            if not params_functionally_equivalent:
+                return False
+        return True
+
 
 @dataclass(frozen=True, repr=_AUTO_REPR)
 class Argument(Expression):
@@ -440,6 +492,11 @@ class Argument(Expression):
 
     def accept(self, visitor: ExpressionVisitor[TVisited]) -> TVisited:
         return visitor.visit_argument(self)
+
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        return self.name == other.name
 
 
 @dataclass(frozen=True, repr=_AUTO_REPR)
@@ -472,3 +529,12 @@ class Lambda(Expression):
 
     def accept(self, visitor: ExpressionVisitor[TVisited]) -> TVisited:
         return visitor.visit_lambda(self)
+
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        if self.parameters != other.parameters:
+            return False
+        if not self.transformation.functional_eq(other.transformation):
+            return False
+        return True
--- a/snuba/query/expressions.py
+++ b/snuba/query/expressions.py
@@ -122,6 +122,37 @@ class ExpressionVisitor(ABC, Generic[TVisited]):
         raise NotImplementedError
 
 
+class NoopVisitor(ExpressionVisitor[None]):
+    """A noop visitor that will traverse every node but will not
+    return anything. Good for extending for search purposes
+    """
+
+    def visit_literal(self, exp: Literal) -> None:
+        return None
+
+    def visit_column(self, exp: Column) -> None:
+        return None
+
+    def visit_subscriptable_reference(self, exp: SubscriptableReference) -> None:
+        return exp.column.accept(self)
+
+    def visit_function_call(self, exp: FunctionCall) -> None:
+        for param in exp.parameters:
+            param.accept(self)
+        return None
+
+    def visit_curried_function_call(self, exp: CurriedFunctionCall) -> None:
+        for param in exp.parameters:
+            param.accept(self)
+        return exp.internal_function.accept(self)
+
+    def visit_argument(self, exp: Argument) -> None:
+        return None
+
+    def visit_lambda(self, exp: Lambda) -> None:
+        return exp.transformation.accept(self)
+
+
 class StringifyVisitor(ExpressionVisitor[str]):
     """Visitor implementation to turn an expression into a string format
     Usage:
--- a/snuba/query/processors/uniq_in_select_and_having.py
+++ b/snuba/query/processors/uniq_in_select_and_having.py
@@ -6,6 +6,7 @@ of clickhouse but we have not upgraded yet. This exists as a check so that we
 return an exception to the user but don't crash the clickhouse query node
 """
 
+import logging
 from dataclasses import fields
 from typing import List, Sequence
 
@@ -14,6 +15,7 @@ from snuba.clickhouse.query import Query
 from snuba.query.exceptions import InvalidQueryException
 from snuba.query.expressions import Column, Expression, FunctionCall, NoopVisitor
 from snuba.request.request_settings import RequestSettings
+from snuba.state import get_config
 
 
 class MismatchedAggregationException(InvalidQueryException):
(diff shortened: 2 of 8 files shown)


snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,97 @@
...
+            return None
+        selected_columns = query.get_selected_columns()
+        uniq_finder = _FunctionNameFindingVisitor("uniq")
+        having_clause.accept(uniq_finder)
+        if uniq_finder.found_functions:
+            matcher = _ExpressionOrAliasMatcher(uniq_finder.found_functions)
+            for col in selected_columns:
+                col.expression.accept(matcher)
+            if not all(matcher.found_expressions):
+                should_throw = get_config("throw_on_uniq_select_and_having", False)
```
fpacifici, line 90: Regarding the dry run, I would simply log a warning in Sentry. We should not have too many of these queries to begin with, but you can do a quick estimate of how many queries have a uniq in the HAVING clause from the querylog to ensure they are not hundreds per second.

snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,97 @@
...
+            if exp_to_match.alias is not None and exp_to_match.alias == exp.column_name:
+                self.found_expressions[i] = True
+
+    def visit_function_call(self, exp: FunctionCall) -> None:
+        for param in exp.parameters:
+            param.accept(self)
+        for i, exp_to_match in enumerate(self.expressions_to_match):
+            if (
+                exp_to_match.function_name == exp.function_name
+                and _expressions_equal_ignore_aliases(exp_to_match, exp)
```
fpacifici, line 72: Is this the reason why you were looking for a way to generate a matcher from an expression ?

snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,97 @@
...
+
+class _ExpressionOrAliasMatcher(NoopVisitor):
+    def __init__(self, expressions_to_match: Sequence[FunctionCall]):
+        self.expressions_to_match = expressions_to_match
+        self.found_expressions = [False] * len(expressions_to_match)
+
+    def visit_column(self, exp: Column) -> None:
+        for i, exp_to_match in enumerate(self.expressions_to_match):
+            if exp_to_match.alias is not None and exp_to_match.alias == exp.column_name:
+                self.found_expressions[i] = True
```
fpacifici, line 64: This will never happen. You can remove this method entirely. 
The parsing operation expands all the aliases by design. After parsing alias references do not exist anymore in the query. 
That was intentional so that the query processing does not need to deal with them. We also tweak them (`_snuba_aliasname`) so that we would not have shadowing of real columns.

This means that if you had 
```
SELECT u_user
...
HAVING (uniq(user) as u_user) > x
```

In the AST, at every step of query processing you will have 
```
SELECT (uniq(user) as u_user)
....
HAVING (uniq(user) as u_user) > x
```

So you do not have to resolve aliases, you just need to be consistent with alias assignment if you apply changes that are not across the board. Changes are generally applied across the board as the transformation functions in the AST api have very little context of what they are transforming. 
I can point you to the document where we discussed multiple options for this case.

fpacifici (changes requested):
I think the approach is fine. 
please see my comment inline around alias expansion.

snuba/query/processors/uniq_in_select_and_having.py
vladkluev: yes

snuba/query/processors/uniq_in_select_and_having.py
vladkluev: I'm okay to remove this however I'm wondering if this breaks the modularity of the pipeline. In order to make this assumption, this stage of the pipeline has to know about the behavior of the parsing stage. Is that a good thing?

vladkluev: got it

snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class UniqInSelectAndHavingProcessor(QueryProcessor):
+    def process_query(self, query: Query, request_settings: RequestSettings) -> None:
+        having_clause = query.get_having()
+        if not having_clause:
+            return None
+        selected_columns = query.get_selected_columns()
+        uniq_finder = _FunctionNameFindingVisitor("uniq")
```
evanh, line 78: You can replace _FunctionNameFindingVisitor with a matcher:

```
matcher = Param("function", FunctionCallMatch(String("uniq")))
for exp in having_clause:
    match = matcher.match(exp)
    if match is not None:
        found_function = match.expression("function")
```

snuba/query/processors/uniq_in_select_and_having.py
evanh: I should have responded to your question in Slack, but you can also programmatically build matchers. I don't necessarily think that is the right solution here, but just FYI.

```
for exp in expressions_to_match:
    matcher = FunctionCallMatch(String(exp.function_name))
```

evanh (commented):
I know that this was happening only on transactions, but presumably it could also happen on errors, right?

Also, can you clarify how this is intended to work with aliases? If there is an alias of the `uniq` expression in the select, is that sufficient?

snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,92 @@
...
+                col.expression.accept(matcher)
+            if not all(matcher.found_expressions):
+                should_throw = get_config("throw_on_uniq_select_and_having", False)
+                error = MismatchedAggregationException(
+                    f"Aggregation is in HAVING clause but not SELECT: {uniq_finder.found_functions}"
+                )
+                if should_throw:
+                    raise error
+                else:
+                    logging.warning(str(error))
```
fpacifici, line 92: If you log the warning this way, every message with a different expression will cause a new group. Which can be a lot and also unreadable.

Please:
- use the `extra` field to provide metadata for your error, you can pass a dictionary.
- set `exc_info` = True so you have the stack trace and function call parameters in sentry
- Only write the error message as log message. Do not serialize the parts of the AST inside the message

fpacifici (changes requested):

vladkluev:
@evanh 

> I know that this was happening only on transactions, but presumably it could also happen on errors, right?

Good point will add it

> Also, can you clarify how this is intended to work with aliases? If there is an alias of the uniq expression in the select, is that sufficient?

It is sufficient for clickhouse yes. This will not crash it for example:

```sql
SELECT 
    (project_id AS _snuba_project_id), transaction_name,
    _snuba_count_unique_user -- <----- only the alias will work.
FROM transactions_dist_PROD 
WHERE 
    greaterOrEquals((finish_ts AS _snuba_finish_ts), toDateTime('2021-12-18T19:20:14', 'Universal')) AND 
    less(_snuba_finish_ts, toDateTime('2021-12-29T21:55:02', 'Universal')) AND 
    in((project_id AS _snuba_project_id), [173225]) 
GROUP BY transaction_name, _snuba_project_id 
HAVING 
    less((divide(countIf(notEquals(transaction_status, 0) AND
    notEquals(transaction_status, 1) AND notEquals(transaction_status, 2)), count()) AS _snuba_failure_rate), 1.0) AND 
    greater((uniq(user) AS _snuba_count_unique_user), 1.0)
ORDER BY _snuba_count_unique_user DESC LIMIT 3 OFFSET 0
```

I originally even had this as a case that was handled but @fpacifici requested I remove it because the parser expands all aliases

https://github.com/getsentry/snuba/pull/2339#discussion_r782660867

snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self, expressions_to_match: Sequence[FunctionCall]):
+        self.expressions_to_match = expressions_to_match
+        self.found_expressions = [False] * len(expressions_to_match)
+
+    def visit_function_call(self, exp: FunctionCall) -> None:
+        for param in exp.parameters:
+            param.accept(self)
+        for i, exp_to_match in enumerate(self.expressions_to_match):
+            if (
+                exp_to_match.function_name == exp.function_name
```
fpacifici, line 66: Isn't this condition already covered by line 50 ?

snuba/query/processors/uniq_in_select_and_having.py
```diff
@@ -0,0 +1,92 @@
...
+    fields_equal = True
+    for field in fields(exp1):
+        if field.name == "alias":
+            continue
+        field_val1 = getattr(exp1, field.name)
+        field_val2 = getattr(exp2, field.name)
+        if isinstance(field_val1, Expression):
+            fields_equal &= _expressions_equal_ignore_aliases(field_val1, field_val2)
+        else:
+            fields_equal &= field_val1 == field_val2
```
fpacifici, line 50: Not sure this works for the parameters of a function:
`FunctionCall` has three fields:
```
alias: string
name: string
parameters: Tuple(expression....)
```
When you check for the equality if the parameters you will fall to line 50, where the two tuples will be compared to each other. And the elements of the tuples will be compared with `==` which takes the alias into account.

snuba/query/processors/uniq_in_select_and_having.py
vladkluev: yes it is, it's just a small performance optimization. No need to inspect the object and do the rest of the processing if the function call is not the same

vladkluev: If it's confusing I can remove it (or comment it)

snuba/query/processors/uniq_in_select_and_having.py
vladkluev: Good catch! What do you think of me adding this functionality to the Expression class?

fpacifici: Do you mean equality without aliases ? If yes, let me think if there was any reason for not doing it.

vladkluev: yes that is what I mean. Functional equivalency

fpacifici: Well the idea was to solve this very problem with matchers, but the use case of taking an expression and find an identical one was not very common. I have nothing against adding it to the expression. Maybe you can build something more flexible by generating the matcher from the expression, so that the result can be used to build a larger matcher, etc.

vladkluev: ok let me give the matcher path a shot

vladkluev: Actually, while I think this is a cool thing to build. I think it may be too much overhead for testing it at this time given that this is an active production issue. There's also nothing preventing us from implementing the matcher solution later and swapping the functional equivalency for it.

tests/query/test_expressions.py
```diff
@@ -199,97 +202,101 @@ def test_hash() -> None:
...
-    [
-        (Column(None, "t1", "c1"), "t1.c1"),
-        (Column(None, None, "c1"), "c1"),
-        (Literal(None, "meowmeow"), "'meowmeow'"),
-        (Literal(None, 123), "123"),
-        (Literal(None, False), "False"),
-        (
-            Literal(None, datetime(2020, 4, 20, 16, 20)),
-            "datetime(2020-04-20T16:20:00)",
+TEST_CASES = [
```
vladkluev, line 205: the cases have stayed the same, I just pulled them out into a list

vladkluev:
@fpacifici @evanh I've added functional equivalency as a function on Expression. I think this is the easier thing to do than generating matchers from expressions. please take a look when you can

snuba/query/expressions.py
```diff
@@ -344,6 +401,20 @@ def __iter__(self) -> Iterator[Expression]:
...
         return visitor.visit_function_call(self)
 
+    def functional_eq(self, other: Expression) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        if self.function_name != other.function_name:
+            return False
+        if len(self.parameters) != len(other.parameters):
+            return False
+        for i, param in enumerate(self.parameters):
```
evanh, line 411: You don't need to change this, but you could shave one line off this by doing

```
for other, param in zip(self.parameters, other.parameters):
```

tests/query/processors/test_uniq_in_select_and_having.py
```diff
@@ -0,0 +1,97 @@
...
+            ],
+            condition=None,
+            having=uniq_expression(alias="my_alias"),
+        ),
+        id="same aggregation, different alias",
+    ),
+]
+
+
+@pytest.mark.parametrize("input_query", deepcopy(INVALID_QUERY_CASES))
```
evanh, line 82: Why do you do a `deepcopy` here?

evanh (commented):
This works fine, but I wonder if the logic of `function_eq` should be the default behaviour. I  think generally speaking when we do equality the alias shouldn't be part of that.

tests/query/processors/test_uniq_in_select_and_having.py
vladkluev: It's a habit from the Query tests. Since it's just a python list it can be modified but I guess I don't need to do it here because Expressions are immutable.

evanh: OK I was just curious if there was a reason. The list isn't modified anywhere either.

fpacifici (approved):

PR #2340, opened 2022-01-13
fix(snql) Update to the latest snuba SDK

evanh:
Update to the latest version, which had some breaking changes that needed to be
fixed. Also added an E2E test for using lambdas to support suspect spans.


evanh:
This will fail until 0.1.1 is released, just wanted to get this PR open so it can be reviewed.

PR #2342, opened 2022-01-19
feat: Add attribution fields to the API and log them in Datadog

evanh:
Add two fields, `team` and `feature` that are logged in Datadog along with the
bytes scanned for a given query, so we can see (in Datadog at least) how teams
are using Clickhouse.


snuba/querylog/__init__.py
```diff
@@ -12,6 +12,41 @@
...
+) -> None:
+    tags = {
+        "team": request.settings.get_team(),
+        "feature": request.settings.get_feature(),
+        "parent_api": request.settings.get_parent_api(),
+        "referrer": request.referrer,
+    }
+    profile = extra_data["profile"] if extra_data["profile"] else {}
+    bytes_scanned = profile.get("bytes", 0.0)
+    metrics.increment("snuba.attribution", bytes_scanned, tags=tags)
```
vladkluev, line 47: if there's no profile or bytes scanned, do we still want to record the metric?

snuba/querylog/__init__.py
```diff
@@ -28,24 +63,8 @@ def record_query(
...
-            },
-            mark_tags={
-                "final": final,
-                "referrer": referrer,
-                "parent_api": request.settings.get_parent_api(),
-            },
-        )
-
+        _record_timer_metrics(request, timer, query_metadata)
+        _record_attribution_metrics(request, extra_data)
```
vladkluev, line 67: How does extra_data get populated? There are expectations of it having specific fields but its type is `Mapping[str, Any]` without any explanation in the docstring

snuba/clickhouse/native.py
```diff
@@ -147,12 +147,12 @@ def execute(
...
-                        "blocks": conn.last_query.profile_info.blocks,
-                        "rows": conn.last_query.profile_info.rows,
-                        "elapsed": conn.last_query.elapsed,
-                    }
+                    profile_data = ClickhouseProfile(
+                        bytes=conn.last_query.profile_info.bytes or 0,
+                        blocks=conn.last_query.profile_info.blocks or 0,
+                        rows=conn.last_query.profile_info.rows or 0,
+                        elapsed=conn.last_query.elapsed or 0.0,
+                    )
```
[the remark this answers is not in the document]
evanh: I think it could be 0, yes, if there is no data to scan it should return 0. I don't think that will mess up the metrics for us though.

snuba/querylog/__init__.py
evanh: Yes, because we want to record all the teams/features/referrers even if the numbers are 0. Otherwise there might be some features that we aren't aware of.

snuba/querylog/__init__.py
evanh: It's actually a TypedDict elsewhere in the code (https://github.com/getsentry/snuba/pull/2342/files#diff-9523648ba19de16486059a3c7602cf46b28ed7065d2db716f81838e356d7d412R13)

I think it's not typed here to avoid dependencies.

snuba/request/request_settings.py
```diff
@@ -132,6 +150,12 @@ def get_dry_run(self) -> bool:
     def get_legacy(self) -> bool:
         return False
 
+    def get_team(self) -> str:
+        return "subscription"
+
+    def get_feature(self) -> str:
+        return "subscription"
+
```
vladkluev, line 158: I'm not sure how I feel about this being hardcoded. Can it not be populated further up the chain?

tests/test_snql_api.py
```diff
@@ -580,6 +580,24 @@ def test_suspect_spans_data(self) -> None:
...
+                    WHERE timestamp >= toDateTime('{self.base_time.isoformat()}')
+                    AND timestamp < toDateTime('{self.next_time.isoformat()}')
+                    AND project_id IN tuple({self.project_id})
+                    """,
+                    "team": "sns",
+                    "feature": "test",
+                }
+            ),
+        )
+        assert response.status_code == 200
```
vladkluev, line 599: Is there any way for us to test the metrics we output? I'm not saying you should do it in this PR if we don't but I think we probably should?

vladkluev (approved):
I'm okay for this to be merged but did have some questions. You can decide what you want to do about them

tests/test_snql_api.py
evanh: I checked and right now we don't. I don't think it would be hard to add/expand the `DummyMetrics` class to facilitate those types of tests, but I didn't want to do that in this PR.

snuba/request/request_settings.py
evanh: For subscriptions? Can I ask why you don't like this?

vladkluev: Are all subscriptions always the same? Can't you subscribe to different things?

evanh: OK I think I see. Right now we lump all these queries into one thing. Every subscription even has the same referrer. Since one team owns subscriptions, and they are part of a single feature, we hardcoded things this way. I can move this logic up a level so that if we want to start separate metrics/release health alerts etc. we can do that.

vladkluev: Yeah I would prefer that

snuba/request/schema.py
```diff
@@ -133,6 +133,9 @@ def generate_template(self) -> Any:
             "dry_run": {"type": "boolean", "default": False},
             # Flags if this a legacy query that was automatically generated by the SnQL SDK
             "legacy": {"type": "boolean", "default": False},
+            # Team and feature are used for resource attribution
+            "team": {"type": "string", "default": "<unknown>"},
```
fpacifici, line 137: Food for thoughts. 
I think we will have to decide how we allocate capacity to teams. There should be two ways:
- A team asks for a quota. We assign that, we give it an id like an app key (which is what we would see here) and somewhere else (out of the snuba query path) we keep track of a data model that assigns quota ids to teams/features
- We identify a quota with the feature-team pair. This univocally assign a quota to a team directly. Though it becomes impractical during reorgs as teams can merge and split for reasons that are not technical. When that happens we would have to adjust the allocation and wreck our data.
- We identify a qutoa with a product-feature pair. That gives the same level of granularity as the team-feature pair but it is agnostic to team reorganizations. The mapping between product and team happens in a data model outside of snuba. The data consumption we store does not change.

I think the first and third are not substantially different. The third maybe can make some analysis quicker to do. The second may become hard to manage.

evanh: I agree. I think ultimately using an API key makes more sense because it's easier to automate and lets teams use their quota how they want, instead of being tied to a specific product/feature name. I think this first step will let us get an MVP of how things are being used and we can iterate on the api key concept and how quotas are defined.

PR #2343, opened 2022-01-19
Add explanation of query cost to docs

volokluev:
Ideally we can get to the point that someone asks a question and we just link them the docs.

### DIFF
--- a/docs/source/query/overview.rst
+++ b/docs/source/query/overview.rst
@@ -5,6 +5,57 @@ Querying Snuba
 This guide drives you through the process of authoring and testing a
 Snuba query.
 
+
+Understanding the cost of queries
+=================================
+The cost of a snuba query has many factors but these are key to keep in mind when constructing queries:
+
+* Amount of rows scanned
+* Amount of columns scanned
+* Type of data scanned
+
+Let's take this query as an example::
+
+    MATCH events
+    SELECT count() AS count
+    WHERE 
+        timestamp > "2021-01-01" AND 
+        timestamp <= "2021-01-02" AND
+        project_id = 1
+ 
+The query above will scan two columns: timestamp and project id over one day. Timestamps and integers are also easy comparisons for a computer to make (can be done in one instruction). Thus this query is fairly cheap. The only way to make this query cheaper is to reduce the time window and scan fewer rows.
+
+Now let's see how we can make this query more expensive.
+
+Expanding the time frame::
+
+    MATCH events
+    SELECT count() AS count
+    WHERE 
+        timestamp > "2021-01-01" AND 
+        timestamp <= "2021-03-01" AND
+        project_id = 1
+ 
+
+Adding more fields to scan. Any extra fields that are scanned are extra files which have to be opened by clickhouse::
+
+    MATCH events
+    SELECT count() AS count
+    WHERE 
+        timestamp > "2021-01-01" AND 
+        timestamp <= "2021-03-01" AND
+        project_id = 1 AND
+        browser = "firefox" AND
+        os = "ios" AND
+        url = "https://santry.io/some_path" AND
+        tags[duration_group] = IN tuple("<10s", "<30s")
+        
+
+Nested fields are especially expensive to unnest and compare.
+
+
+
+
 Exploring the Snuba data model
 ==============================
 


docs/source/query/overview.rst
```diff
@@ -5,6 +5,57 @@ Querying Snuba
...
+        timestamp > "2021-01-01" AND 
+        timestamp <= "2021-03-01" AND
+        project_id = 1 AND
+        browser = "firefox" AND
+        os = "ios" AND
+        url = "https://santry.io/some_path" AND
+        tags[duration_group] = IN tuple("<10s", "<30s")
+        
+
+Nested fields are especially expensive to unnest and compare.
```
evanh, line 54: You should clarify what a "nested" field is here.

docs/source/query/overview.rst
```diff
@@ -5,6 +5,57 @@ Querying Snuba
...
+
+Adding more fields to scan. Any extra fields that are scanned are extra files which have to be opened by clickhouse::
+
+    MATCH events
+    SELECT count() AS count
+    WHERE 
+        timestamp > "2021-01-01" AND 
+        timestamp <= "2021-03-01" AND
+        project_id = 1 AND
+        browser = "firefox" AND
```
evanh, line 48: browser/os/url are not good examples, because they are actually context fields, so as far as the product is concerned they are also nested (`contexts[browser][name]`).

evanh (commented):
There are a couple other quick examples I would give of high cost: 

- Regex match on a column: `url LIKE '*santry.io/*'`
- Filtering on array columns: `has(tags.key, 'browser')`

docs/source/query/overview.rst
```diff
@@ -5,6 +5,57 @@ Querying Snuba
...
+* Amount of rows scanned
+* Amount of columns scanned
+* Type of data scanned
+
+Let's take this query as an example::
+
+    MATCH events
+    SELECT count() AS count
+    WHERE 
+        timestamp > "2021-01-01" AND
```
evanh, line 22:
```suggestion
        timestamp > toDateTime('2021-01-01') AND 
```
Dates are wrapped in `toDateTime('x')`

docs/source/query/overview.rst
evanh, line 48:
```suggestion
        browser = 'firefox' AND
```
SnQL uses single quotes.

docs/source/query/overview.rst
fpacifici: Nested fields are fairly complicated, I am not sure I would just brush them off as "especially expensive" as it is not exactly accurate,

The reason for this is that I saw a lot of scenarios with product teams accepting that something is "universally too expensive" and drawing the wrong conclusion that such feature is never to be used.

Specific cases:
- load tags for your results (even if you arrayJoin them) without filtering on them or aggregating a large amount of rows. This is not a problem as it happens after filtering. So you already restricted your dataset
- Aggregate over an arrayJoin of a nested column on a relatively short period of time. Not an issue. It largely depends if the dataset has enough shards.
- Search for the values of a specific tag (has(tags.key, 'asd')). This is generally good (at least unless every event has that field), That is because we have a bloomfilter index on the tags keys column
- Search on a specific tag key-value pair. Same there is a bloomfitler index. This is manageable over large period of times (unless all rows have it).
- regex on tags. This is deadly. we cannot use the index we need to unpack everything and then run a regex over every row.

So fundamentally are nested columns good or bad? it depends. I don't think we should take a universal position one way or another. Unfortunately for all practical purposes there is no simple one line rule that applies everything. In order to assess the impact of using them we need to know how they work and how they would work in the specific use case.

docs/source/query/overview.rst
```diff
@@ -5,6 +5,57 @@ Querying Snuba
...
+    MATCH events
+    SELECT count() AS count
+    WHERE 
+        timestamp > "2021-01-01" AND 
+        timestamp <= "2021-03-01" AND
+        project_id = 1 AND
+        browser = "firefox" AND
+        os = "ios" AND
+        url = "https://santry.io/some_path" AND
+        tags[duration_group] = IN tuple("<10s", "<30s")
```
fpacifici, line 51: This would be a reasonable usage of tags. The query should become:
`has(_tags_hashmap, cityHash64("duration_group:<10s")) OR has(_tags_hashmap, cityHash64("duration_group:<30s"))`. Which uses the index. It is ok but it gets worse if all rows have that specific tag with that specific value so the index does not exclude anything.

See my message below for details on how to deal with nested columns. This is a more complicated matter.

docs/source/query/overview.rst
```diff
@@ -5,6 +5,57 @@ Querying Snuba
...
 Snuba query.
 
+
+Understanding the cost of queries
+=================================
+The cost of a snuba query has many factors but these are key to keep in mind when constructing queries:
+
+* Amount of rows scanned
+* Amount of columns scanned
+* Type of data scanned
```
fpacifici, line 15: I think this is a bit of an oversimplification. Do you plan to expand this section or do you plan to keep this level of detail? (I think we should expand and provide a lot more details over time. Sadly when addressing the performance of clickhouse queries there are very few areas where we can have simple absolute rules, for all practical purposes it is critical to understand what clickhouse would do in the specific case).

On sharded datasets the type of aggregtions that happen on the distributed table are equally important because those do not parallelize well.

The type of conditions to run are also critical: equality is one thing, regex is very different.

docs/source/query/overview.rst
vladkluev: > has(_tags_hashmap, cityHash64("duration_group:<10s")) OR has(_tags_hashmap, cityHash64("duration_group:<30s"))

Are you sure it would become that? Is there a query processor that expands `IN` tag to `OR`s?

fpacifici: would it become like that if you provide an `IN`. No
That query should be expressed (as of today) as a series of OR conditions. 
We could write a processor to optimize that
The problem for the reader though is that I would interpret the paragraph and the example as "conditions on tags are bad" while it is more nuanced: 
- equality conditions on tags are ok
- IN can be expressed as a series of equalities in OR
- if your tag has extremely low selectivity the result is still bad. as the index does not exclude anything.

docs/source/query/overview.rst
fpacifici: You may want to be aware of this problem before going with that strategy.
Out of experience explaining query performance to product teams, one constant I saw is that, if we provide a simple rule ignoring the corner cases, the product team takes it literally. The result is that they will derive their strategy from that simple rule and more often than not they will run into corner case where the rule does not apply.
Example: 2 years ago the message was "conditions on tags are bad". That became the universal rule and we still did not manage to get the nuance around "which conditions on tags are bad and which ones are ok" through.

If I were you I would try to cover a smaller area but providing details for it. 
Just as an example: You could focus on rows scanned first. And provide a longer list of ways to reduce the rows scanned. Then in a follow up talk about reducing columns scanned with its details. 
But, your call.

PR #2344, opened 2022-01-20
approach check: Metrics direct write to aggregateFunction column

onewland:
I know this sucks in its current state so don't worry, it will be polished before merge or a real approval process

Basically I want to validate that this is, at a very high level and ignoring the ugly details, the right way to split out processing for metrics if we want to write with the format `VALUES()`. That is:
- creating a new type of `InsertBatchInterpretExpressions` to indicate to the next step to use the `VALUES` format for insertion (or a format that allows for execution of things like `arrayReduce('quantilesState(0.5,0.75,0.9,0.95,0.99)', [1.0, 2.0, ...])` on insert)
- creating a new type of `ValuesInsertBatch` for direct use by the writer
- creating a new insert batch writer for the VALUES format
- adding `ValuesInsertBatch` handling to the generic multi-storage consumer message processor
- changing the MetricsProcessor to return `InsertBatchInterpretExpressions` for direct writes

This successfully writes to the AggregateFunction column, though it's not handling things like Datetime(s) or being smart about escaping

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -152,6 +152,17 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--consumer-group=metrics_group",
                 ],
             ),
+            (
+                "metrics-direct-write-consumer",
+                [
+                    "snuba",
+                    "multistorage-consumer",
+                    "--storage=metrics_distributions",
+                    "--auto-offset-reset=latest",
+                    "--log-level=debug",
+                    "--consumer-group=metrics_direct_write_group",
+                ],
+            ),
         ]
 
     if settings.ENABLE_SESSIONS_SUBSCRIPTIONS:
--- a/snuba/clickhouse/http.py
+++ b/snuba/clickhouse/http.py
@@ -75,7 +75,8 @@ class InsertStatement:
         columns_statement = (
             f"({','.join(self.__column_names)}) " if self.__column_names else ""
         )
-        format_statement = f" FORMAT {self.__format}" if self.__format else ""
+        # FIXME maybe but idk what other use empty format has
+        format_statement = f" FORMAT {self.__format}" if self.__format else "VALUES"
         return f"INSERT INTO {self.__database}.{self.__table_name} {columns_statement} {format_statement}"
 
 
@@ -138,7 +139,7 @@ class HTTPWriteBatch:
             "Connection": "keep-alive",
             "Accept-Encoding": "gzip,deflate",
         }
-        if password != '':
+        if password != "":
             headers["X-ClickHouse-Key"] = password
         if encoding:
             headers["Content-Encoding"] = encoding
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -251,6 +251,28 @@ class ClickhouseCluster(Cluster[ClickhouseWriterOptions]):
             buffer_size=buffer_size,
         )
 
+    def get_values_batch_writer(
+        self,
+        metrics: MetricsBackend,
+        insert_statement: InsertStatement,
+        encoding: Optional[str],
+        options: ClickhouseWriterOptions,
+        chunk_size: Optional[int],
+        buffer_size: int,
+    ) -> BatchWriter[bytes]:
+        return HTTPBatchWriter(
+            host=self.__query_node.host_name,
+            port=self.__http_port,
+            user=self.__user,
+            password=self.__password,
+            metrics=metrics,
+            statement=insert_statement.with_database(self.__database),
+            encoding=encoding,
+            options=options,
+            chunk_size=chunk_size,
+            buffer_size=buffer_size,
+        )
+
     def is_single_node(self) -> bool:
         """
         This will be used to determine:
@@ -334,6 +356,10 @@ expected_storage_sets = {
     for s in StorageSetKey
     if (s not in DEV_STORAGE_SETS or settings.ENABLE_DEV_FEATURES)
 }
+
+assert (
+    not expected_storage_sets - _unique_registered_storage_sets
+), "All storage sets must be assigned to a cluster"
 
 # Map all storages to clusters via storage sets
 _STORAGE_SET_CLUSTER_MAP = {
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -37,7 +37,12 @@ from snuba.datasets.storage import WritableTableStorage
 from snuba.datasets.storages import StorageKey
 from snuba.datasets.table_storage import TableWriter
 from snuba.environment import setup_sentry
-from snuba.processor import InsertBatch, MessageProcessor, ReplacementBatch
+from snuba.processor import (
+    InsertBatch,
+    InsertBatchInterpretExpressions,
+    MessageProcessor,
+    ReplacementBatch,
+)
 from snuba.utils.metrics import MetricsBackend
 from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.streams.configuration_builder import build_kafka_producer_configuration
@@ -60,6 +65,19 @@ class JSONRowInsertBatch(NamedTuple):
             )
         else:
             return type(self), (self.rows, self.origin_timestamp)
+
+
+class ValuesInsertBatchWriter(ProcessingStep[InsertBatchInterpretExpressions]):
+    def __init__(
+        self,
+        writer: BatchWriter[InsertBatchInterpretExpressions],
+        metrics: MetricsBackend,
+    ) -> None:
+        self.writer = writer
+        self.metrics = metrics
+
+    def submit(self, message: Message[InsertBatchInterpretExpressions]) -> None:
+        assert False
 
 
 class InsertBatchWriter(ProcessingStep[JSONRowInsertBatch]):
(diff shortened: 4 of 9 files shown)


snuba/consumers/consumer.py
```diff
@@ -62,6 +67,95 @@ def __reduce_ex__(
...
+        self.__metrics.timing("batch_write_ms", write_finish - write_start)
+        rows = sum(len(message.payload.rows) for message in self.__messages)
+        self.__metrics.increment("batch_write_msgs", rows)
+
+        logger.debug(
+            "Waited %0.4f seconds for %r rows to be written to %r.",
+            write_finish - write_start,
+            rows,
+            self.__writer,
+        )
```
onewland, line 150: This is a copy/paste of `InsertBatchWriter.close`

onewland:
My commit hook `mypy` is succeeding but running the command manually and the CI version is failing with 

```
snuba/consumers/consumer_builder.py:207: error: Cannot infer type argument 1 of "KafkaConsumerStrategyFactory"
snuba/web/views.py:594: error: Cannot infer type argument 1 of "KafkaConsumerStrategyFactory"
```

As a general note, verifying the type checking around this section of the code is very treacherous

snuba/consumers/consumer.py
```diff
@@ -62,6 +67,95 @@ def __reduce_ex__(
...
+        if isinstance(value, str):
+            return value
+
+        if isinstance(value, int):
+            return str(value)
+
+        if isinstance(value, list):
+            for x in value:
+                if not isinstance(x, int):
+                    raise TypeError("ints should only contain arrays")
```
MeredithAnya, line 85: nit: arrays should only contain ints

snuba/cli/devserver.py
```diff
@@ -152,6 +152,17 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--consumer-group=metrics_group",
                 ],
             ),
+            (
+                "metrics-direct-write-consumer",
+                [
+                    "snuba",
+                    "multistorage-consumer",
```
nikhars, line 159: You don't need multistorage-consumer if you are writing to a singe storage. The regular consumer should work.

snuba/clusters/cluster.py
```diff
@@ -335,6 +357,10 @@ def __get_cluster_nodes(self, cluster_name: str) -> Sequence[ClickhouseNode]:
     if (s not in DEV_STORAGE_SETS or settings.ENABLE_DEV_FEATURES)
 }
 
+assert (
+    not expected_storage_sets - _unique_registered_storage_sets
+), "All storage sets must be assigned to a cluster"
```
nikhars, line 362: Why do you need to add this back? I thought we recently removed it for some reason.

MeredithAnya:
@onewland I'm not as familiar with how the batch writer was originally working but to overall this makes sense to me. You'd just be added a couple more processors like `DistributionsMetricsProcessorDirectWriter` for the sets and counters yeah?

snuba/consumers/consumer.py
```diff
@@ -62,6 +67,95 @@ def __reduce_ex__(
...
+            self.__metrics.timing(
+                "latency_ms", (write_finish - message.timestamp.timestamp()) * 1000
+            )
+            if message.payload.origin_timestamp is not None:
+                self.__metrics.timing(
+                    "end_to_end_latency_ms",
+                    (write_finish - message.payload.origin_timestamp.timestamp())
+                    * 1000,
+                )
+        self.__metrics.timing("batch_write_ms", write_finish - write_start)
```
nikhars, line 141: It would be a good idea to differentiate the name of the metrics for value based writes v/s JSON writes. That way you could compare the two if you wanted to.

onewland:
Yeah, that's right, and they would also have to generate multiple rows each to account for the separate granularities. That may be a requirement we should add to to this proof-of-concept, though I'll defer to others on how complex that sounds (1 input row mapping to multiple output rows)

snuba/consumers/consumer.py
```diff
@@ -204,15 +298,17 @@ def join(self, timeout: Optional[float] = None) -> None:
 
 
 class ProcessedMessageBatchWriter(
-    ProcessingStep[Union[None, JSONRowInsertBatch, ReplacementBatch]]
+    ProcessingStep[Union[None, JSONRowInsertBatch, ReplacementBatch, ValuesInsertBatch]]
 ):
     def __init__(
         self,
         insert_batch_writer: InsertBatchWriter,
```
nikhars, line 305: I guess this should also be made optional, since the insert with values doesn't need to set this up

snuba/consumers/consumer.py
```diff
@@ -428,7 +553,12 @@ def process_message(
 
 def process_message_multistorage(
     message: Message[MultistorageKafkaPayload],
-) -> Sequence[Tuple[StorageKey, Union[None, JSONRowInsertBatch, ReplacementBatch]]]:
+) -> Sequence[
```
nikhars, line 556: All these changes which have been made for multistorage will also be needed for the single storage

snuba/cli/devserver.py
onewland: Sure, though eventually this will write to 3 tables.

snuba/clusters/cluster.py
onewland: I think I branched off an old master maybe -- I'll keep in mind we don't want to restore it

snuba/clickhouse/http.py
```diff
@@ -75,7 +75,8 @@ def build_statement(self) -> str:
         columns_statement = (
             f"({','.join(self.__column_names)}) " if self.__column_names else ""
         )
-        format_statement = f" FORMAT {self.__format}" if self.__format else ""
+        # FIXME maybe but idk what other use empty format has
```
fpacifici, line 78: None that I can think of.

snuba/clusters/cluster.py
```diff
@@ -251,6 +251,28 @@ def get_batch_writer(
...
+            user=self.__user,
+            password=self.__password,
+            metrics=metrics,
+            statement=insert_statement.with_database(self.__database),
+            encoding=encoding,
+            options=options,
+            chunk_size=chunk_size,
+            buffer_size=buffer_size,
+        )
+
```
fpacifici, line 275: This seems identical to the one above ?

fpacifici (commented):
I think there may be a simpler way to approach this problem that may even be cleaner.
What if we started from the assumption that the INSERT statement format is a setting of a storage class. 
Context:
while we define storages in code, these are meant to be configuration, so the code (tech debt apart) is supposed to be fully static so that it can be moved to a config file instead.

Example, the storage for metrics is defined here 
https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/metrics.py#L152-L177. This could be expressed in a yaml file 
Now, in order to write on that table we need to make it a writable storage like this 
https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/metrics.py#L83-L102

What if the writer format was an init parameter of a WritableStorage (just another storage config parameter, like table name or topic name).

Then the writable storage provides the batch writer (`get_batch_writer`) which is what is ultimately used to write on Clickhouse no matter if you are in a single storage consumer or multi storage consumer. The BatchWriter is the object that knows which cluster to write to and which table to write onto. The consumer does not know these details. It relies on the BatchWriter that comes from the storage to perform the write.
This separation of concerns allows us to separate the consumer logic from the details of the DB we want to write onto.

https://github.com/getsentry/snuba/blob/master/snuba/datasets/storage.py#L181-L191

Here we could initialize the TableWriter with the format (the TableWriter creates the InsertStatement, thus it can set the format).

At this point you should not need to touch anything else around the consumer or the initialization. No matter how you try to write on the aggregated table, the batch writer will always be initialized with the proper INSERT format. 

This basically restrict the concerns regarding the format to the BatchWriter (which is the only component that knows and should know about the DB), nobody else knows about it and makes it fully storage specific (which is kind of what we want).

What do you think ?

snuba/clusters/cluster.py
onewland: It's true that this is equivalent though that assumes `JSONRow=bytes` (in terms of return typing). I think I went down a path that didn't yield anything here by separating it out.

onewland:
@fpacifici I think I agree at a high level -- I was assuming that by the time this reached its final form that the writer would have to be fairly significantly refactored to avoid a bunch of duplication. 

I'll parse through the individual details on Monday and make sure I agree with the approach you're suggesting (or one similar).

onewland:
@fpacifici I think your comment makes sense, but I'd like to run one thing by you. The `process_message` and `process_message_multistorage` seem to have a coupling to the table format (JSONInsertRow) by doing the serializing directly:

https://github.com/getsentry/snuba/blob/69862db3ad224b48810ac1bb3001e4c446bf0aff/snuba/consumers/consumer.py#L453-L462

https://github.com/getsentry/snuba/blob/69862db3ad224b48810ac1bb3001e4c446bf0aff/snuba/consumers/consumer.py#L421-L424

Is there a good reason for this or is it vestigial? Because I'd like to move the encoding to the batch writer (and it doesn't seem like that logic belongs at the high-level consumer flow). Maybe we could call the processing finished something like a `FinalizedRowInsertBatch`?

fpacifici:
I don't think there is a deliberate reason to make the processor aware of the Json serialization.
Ideally I think it should not. In practice though, we know that a consumer that uses the "reduce" expression cannot be used with json serialization so probably considering the processor completely independent from the serialization is an oversimplification as the object returned by the processor only works with a VALUES format and not JSON format.

Knowing that the processor is provided by the storage definition and the format would be provided by the storage definition as well, I would not be against letting the processor produce, in this case a `ValuesRowInsertBatch` instead of `JSONRowInsertBatch`, make the high level consumer infra blissfully unaware of that and making the batch writer validate they receive the right type.

I don't think it would be viable to enforce that a storage whose processor produces ValueBatch must have a Value format through type checking. Likely you would have to make the entire storage definition generic. It is a huge refactoring for very little value. A runtime check should be enough.

onewland:
Got the feedback I needed from this, actual change will be in a separate PR/branch

PR #2345, opened 2022-01-21
docs(granularity): Add more description

nikhars:
Add more description to the granularity clause. The content has been taken from the previous README.md

### DIFF
--- a/docs/source/language/snql.rst
+++ b/docs/source/language/snql.rst
@@ -165,6 +165,12 @@ GRANULARITY
 
 An integer representing the granularity to group time based results.
 
+Snuba provides a magic column `time`, that you can use in groupby. This column gives a floored time value for each event so that events in the same minute/hour/day/etc. can be grouped.
+
+granularity determines the number of seconds in each of these time buckets. Eg, to count the number of events by hour, you would do
+
+``BY time GRANULARITY 3600``
+
 TOTALS
 ======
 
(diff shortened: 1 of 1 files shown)


docs/source/language/snql.rst
```diff
@@ -165,6 +165,12 @@ GRANULARITY
 
 An integer representing the granularity to group time based results.
 
+Snuba provides a magic column `time`, that you can use in groupby. This column gives a floored time value for each event so that events in the same minute/hour/day/etc. can be grouped.
+
+granularity determines the number of seconds in each of these time buckets. Eg, to count the number of events by hour, you would do
+
+``BY time GRANULARITY 3600``
```
vladkluev, line 172: Can we provide a full query? Otherwise it looks good to me

docs/source/language/snql.rst
```diff
@@ -165,6 +165,17 @@ GRANULARITY
 
 An integer representing the granularity to group time based results.
 
+Snuba provides a magic column `time`, that you can use in groupby. This column gives a floored time value for each event so that events in the same minute/hour/day/etc. can be grouped.
```
fpacifici, line 168: I don't think this is accurate.
The existence of the `time` column is part of the schema of a specific entities. Entities that define it will have it, others may not have it. The granularity field is relevant only when this is present.
Which makes the data model inconsistent because you can provide a granularity for a query on an entity that does not have a bucketed time column. There is a tech debt task to revamp the meaning of the granularity column entirely.

This is where the bucketed time column is defined for errors 
https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/events.py#L186-L188

nikhars: Thanks for providing the additional context. I have added documentation around how to find the magic column from the code with an example for errors.

fpacifici (approved):

PR #2346, opened 2022-01-24
feat(metrics) - write directly to aggregate tables

onewland:
Add support for writing directly to aggregate tables for metrics (`metrics_distributions`, `metrics_sets`, and `metrics_counters`)

Functional changes:
- support for inserting using the VALUES format (https://clickhouse.com/docs/en/interfaces/formats/#data-format-values) at the WritableTableStorage level
- support in the multistorage consumer for serializing to the VALUES format (including a new ValuesRowEncoder)
- added a new type of return from processing called `AggregateInsertBatch` to indicate to callers that the data being returned needs to be interpreted by Clickhouse (e.g. it can contain function call expressions)
- added a new consumer type on startup for `metrics-aggregates-consumer`

Non-functional changes:
- changed the storages for metrics aggregate tables to be writeable
- renamed original processors to indicate that they are for writing to buckets (to distinguish from aggregates)

Testing:
- added unit tests for the new metrics aggregate processors
- added a new e2e test to make sure that `AggregateInsertBatch` flows correctly
- did some ad hoc testing on my local machine to verify that data coming from Sentry is written to the tables properly 
  - This was only possible for distributions and counters as far as I know. I didn't see any sets flowing into Kafka but I didn't spend a lot of time seeing if there was a path for this.


snuba/clickhouse/http.py
```diff
@@ -39,12 +39,38 @@ def __default(self, value: Any) -> Any:
...
+                    raise TypeError("ints should only contain arrays")
+
+            return "[" + ",".join([str(x) for x in value]) + "]"
+
+        return "0"
+
+    def encode(self, row: WriterTableRow) -> bytes:
+        ordered_columns = [self.encode_value(row[column]) for column in self.__columns]
+        ordered_columns_str = ",".join(ordered_columns)
+        return f"({ordered_columns_str})".encode("utf-8")
```
onewland, line 71: This should be a lot more fleshed out before it's an active code path but I copied it from my other PR to make sure we had enough data flowing through (e.g. with the table schema's columns in order) for it to work

snuba/clickhouse/http.py
```diff
@@ -39,12 +39,38 @@ def __default(self, value: Any) -> Any:
         else:
             raise TypeError
 
-    def encode(self, value: WriterTableRow) -> JSONRow:
+    def encode(self, value: WriterTableRow) -> bytes:
```
onewland, line 42: These types are equivalent, but I didn't like seeing JSONRow referring to byte strings that are not JSON

snuba/consumers/consumer.py
```diff
@@ -41,11 +41,21 @@
...
+    BatchWriterEncoderWrapper,
+    MockBatchWriter,
+    MockEncoder,
+    WriterTableRow,
+)
 
 logger = logging.getLogger("snuba.consumer")
 
 
+class WriterTableRowBatch(NamedTuple):
```
fpacifici, line 54: I think you will need the same `__reduce_ex__` implementation you have in JSONRowInsertBatch. This is (maybe was) needed for the multi process consumer that needs to pickle rows generate by the processor method that runs in the subprocess and transfer them back to the main process via shared memory.

snuba/datasets/schemas/tables.py
```diff
@@ -47,13 +53,15 @@ def __init__(
...
         part_format: Optional[Sequence[util.PartSegment]] = None,
+        write_format: WriteFormat = WriteFormat.JSON
     ):
         self.__local_table_name = local_table_name
         self.__dist_table_name = dist_table_name
         self.__storage_set_key = storage_set_key
         self.__columns = columns
         self.__mandatory_conditions = mandatory_conditions
         self.__part_format = part_format
+        self.__write_format = write_format
```
fpacifici, line 64: The TableSchema is meant to provide the table config for both readable and writable storages. The write format is only relevant for the WritableStorage. Was there a way to fit this in the WritableStorage class ?

snuba/writer.py
```diff
@@ -95,3 +95,8 @@ def write(self, values: Iterable[bytes]) -> None:
                 / 1000
             )
         )
+
+
+class MockEncoder(Encoder[bytes, TDecoded]):
+    def encode(self, x: TDecoded) -> bytes:
+        return str(x).encode("utf-8")
```
fpacifici, line 102: What is this for.

fpacifici (commented):
The approach seems fine to me. See comments inline

snuba/clickhouse/http.py
```diff
@@ -39,12 +39,38 @@ def __default(self, value: Any) -> Any:
...
+            return str(value)
+
+        if isinstance(value, list):
+            for x in value:
+                if not isinstance(x, int):
+                    raise TypeError("ints should only contain arrays")
+
+            return "[" + ",".join([str(x) for x in value]) + "]"
+
+        return "0"
```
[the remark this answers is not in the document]
onewland: yeah, I'll probably replace this with `raise NotImplementedError` before putting it up for real review (nothing will use VALUES format with just the refactor)

snuba/writer.py
onewland: Because I changed the signature of `InsertBatchWriter` to take a `BatchWriterEncoderWrapper[WriterTableRow]` instead of a `BatchWriter[JsonRow]` I needed to create a BatchWriterEncoderWrapper object in the `build_mock_batch_writer` method at https://github.com/getsentry/snuba/blob/8c5732968106e690db0d5cc39580be3a5dd7ae80/snuba/consumers/consumer.py#L325

fpacifici: ohhhh that mock. Right, ignore my comment.

snuba/consumers/consumer.py
onewland: I think this comment actually reveals why the consumer was serializing to the insert format and those concerns weren’t separated. I bet whoever did that implementation was trying to avoid the flow of:

1. serialize on process 1 to send data to process 2 
2. deserialize the message on process 2
3. re-serialize the message for process 2 to insert

By serializing to JSON, process 2 can skip a couple steps. My intuition is that this efficiency doesn’t buy us much (it could reduce latency but I don't think these writes are extremely latency-sensitive). Do you want to disagree, though? It’s kind of fundamental to this approach

fpacifici: Ah yes, definitively we cannot afford to do the JSON serialization before sending to the subprocess. If we did that will be an instant production incident.

onewland: Spoke offline: we're rejecting this approach

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,137 @@
...
+from snuba.processor import (
+    AggregateInsertBatch,
+    MessageProcessor,
+    ProcessedMessage,
+    _ensure_valid_date,
+)
+
+
+class MetricsAggregateProcessor(MessageProcessor, ABC):
+    GRANULARITIES = [10, 3600, 3600 * 24]
```
nikhars, line 20: Could you define constants for the different granularities?

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,137 @@
...
+                "granularity": ClickhouseInt(granularity),
+            }
+            for granularity in self.GRANULARITIES
+        ]
+        return AggregateInsertBatch(processed, None)
+
+
+class SetsAggregateProcessor(MetricsAggregateProcessor):
+    def _should_process(self, message: Mapping[str, Any]) -> bool:
+        return message["type"] is not None and message["type"] == "s"
```
nikhars, line 74: These message types look really arbitrary. "s" for sets. "c" for counters. "d" for distributions

Can you define all the metrics message types somewhere and refer to them in the processing logic.

Also, what was the reason to not have the message type be "set" or "counters"?

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,137 @@
...
+class DistributionsAggregateProcessor(MetricsAggregateProcessor):
+    def _should_process(self, message: Mapping[str, Any]) -> bool:
+        return message["type"] is not None and message["type"] == "d"
+
+    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+        values = message["value"]
+        for v in values:
+            assert isinstance(
+                v, (int, float)
+            ), "Illegal value in set. Int expected: {v}"
```
nikhars, line 115: The assert is testing against both int and float but the assert message says that only Int is expected.

snuba/datasets/table_storage.py
```diff
@@ -208,18 +211,36 @@ def get_batch_writer(
         chunk_size: int = settings.CLICKHOUSE_HTTP_CHUNK_SIZE,
     ) -> BatchWriter[JSONRow]:
         table_name = table_name or self.__table_schema.get_table_name()
-
+        if self.__write_format == WriteFormat.JSON:
+            insert_statement = InsertStatement(table_name).with_format("JSONEachRow")
+        elif self.__write_format == WriteFormat.VALUES:
+            column_names = self.get_writeable_columns()
+            print(column_names)
```
nikhars, line 218: Nit: Remove print

snuba/datasets/metrics_aggregate_processor.py
onewland: This is something I've inherited. @MeredithAnya do you know if we documented the rationale for this somewhere?

I'm down to extract these magic constants out, though

snuba/datasets/metrics_aggregate_processor.py
onewland: Given that this is a constant, I'm wondering why it's more helpful to subdivide it into more constants?

I can if people want too, though

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,135 @@
...
+from snuba.processor import (
+    AggregateInsertBatch,
+    MessageProcessor,
+    ProcessedMessage,
+    _ensure_valid_date,
+)
+
+
+class MetricsAggregateProcessor(MessageProcessor, ABC):
+    GRANULARITIES_SECONDS = [10, 3600, 3600 * 24]
```
nikhars, line 20: I was probably not clear with my comments the first time around. These are Granularity buckets in seconds. I was suggesting a change along the following lines where the time periods were defined as constants so that the code is easy to read. But it is not a blocker for me.
`
GRANULARITY = [10SEC, 1HOUR, 1DAY]
`

snuba/datasets/metrics_aggregate_processor.py
MeredithAnya: @onewland @nikhars yeah this was set in relay I believe, I also inherited this. @jan-auer maybe you have context on the rationale?

snuba/datasets/metrics_aggregate_processor.py
onewland: I don't feel strongly, but I would like for one more person to agree with you before making this change as I don't think it increases clarity for me

snuba/cli/devserver.py
```diff
@@ -152,6 +152,19 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
...
+                    "snuba",
+                    "multistorage-consumer",
+                    "--storage=metrics_distributions",
+                    "--storage=metrics_counters",
+                    "--storage=metrics_sets",
+                    "--auto-offset-reset=latest",
+                    "--log-level=debug",
+                    "--consumer-group=metrics_aggregates_group",
+                ],
+            ),
```
fpacifici, line 167: I think we should be duplicating data this way as two consumers would be consuming the same topic and write on the same tables (via matview for the first one). 
Feel free to swap the old one with the new one.
Also you will want to use the same consumer group. So you pick up from the last committed offset.

snuba/clickhouse/http.py
```diff
@@ -39,12 +44,34 @@ def __default(self, value: Any) -> Any:
...
+    def __init__(self, columns: Iterable[str]) -> None:
+        self.__columns = columns
+
+    def encode_value(self, value: Any) -> str:
+        if isinstance(value, ClickhouseUnguardedExpression):
+            return value.val
+        elif isinstance(value, ClickhouseInt):
+            return str(value.val)
+        elif isinstance(value, ClickhouseNumArray):
+            return "[" + ",".join([str(x) for x in value.val]) + "]"
```
fpacifici, line 63: Two issues here:
- We are not escaping strings. If I received `asdasd'asdasd` This will produce an invalid statement for Clickhouse
- It is true that, on metrics we only have integers, but the encoder needs to work in a generic environment, we need to support floats and arrays of string and floats.

Also let's have a unit test of the encoder, with pytest parametrization you can write in a compact a way a large amount of test cases, which is a good scenario for the encoder.

snuba/clickhouse/value_types.py
```diff
@@ -0,0 +1,32 @@
...
+
+@dataclass
+class ClickhouseUnguardedExpression:
+    """
+    This is unescaped text that will be sent to Clickhouse and evaluated
+    as an expression so we want to make it clear from naming
+    that using this can be unsafe
+    """
+
+    val: str
```
fpacifici, line 32: Not sure I agree with this approach. I think the encoder should jsut properly escape what it gets, thus work on any string. No need to care if the string is unsafe.

fpacifici (changes requested):

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+        value = message["value"]
+        assert isinstance(
+            value, (int, float)
+        ), "Illegal value for counter value. Int/Float expected {value}"
+
+        return {
+            "value": ClickhouseUnguardedExpression(
+                f"arrayReduce('sumState', [{value}])"
```
fpacifici, line 99: Pretty sure `arrayReduce(sumState, [1,2,3,4,5])` equals `sum([1,2,3,4,5])`. Or just compute the sum directly here. I think it may work.

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,135 @@
...
+        return {
+            "percentiles": ClickhouseUnguardedExpression(
+                f"arrayReduce('quantilesState(0.5,0.75,0.9,0.95,0.99)', {value_array})"
+            ),
+            "min": ClickhouseUnguardedExpression(
+                f"arrayReduce('minState', {value_array})"
+            ),
+            "max": ClickhouseUnguardedExpression(
+                f"arrayReduce('maxState', {value_array})"
+            ),
```
fpacifici, line 125: min, max, count and sum may not need to be calculated with the state function. You may just be able to do the math in the consumer and reduce the size of the batch. Though I am not sure, it would be worth trying out.

fpacifici (commented):
I liked the idea of having two separate PR, one with the infra and one with the specific feature. Is there any specific reason for merging everything into one ?

snuba/clickhouse/http.py
onewland: OK, though for this use case we do need to be able to push unescaped expressions to Clickhouse? Note also that I'm checking that at least the inputs to this method claim to be ClickhouseInt or ClickhouseNumArray, though as I think about it I guess python won't runtime enforce those types on the `val` property of those classes.

Are you thinking that this line:
```python
"percentiles": ClickhouseUnguardedExpression(
      f"arrayReduce('quantilesState(0.5,0.75,0.9,0.95,0.99)', {value_array})"
)
```

should be something like:
```python
# value_array is a ClickhouseNumArray
# encoding of ClickhouseStr would wrap with single quotes and ensure that single quotes inside are properly escaped
ClickhouseFunctionCall('arrayReduce', ClickhouseStr("quantilesState(0.5,0.75,0.9,0.95,0.99)"), value_array)
```

snuba/datasets/metrics_aggregate_processor.py
onewland: I think it would equal `sumState(15)` but that just inserting 15 directly into the table will cause a datatype error (float64 vs. aggregatefunction). Will confirm locally, but I agree it makes sense to do the sum internally

snuba/cli/devserver.py
onewland: am I correct that your first sentence should read "we should not be"?

snuba/clickhouse/value_types.py
onewland: See my comment above. This isn't meant to represent a clickhouse string literal but rather a clickhouse raw expression

snuba/consumers/consumer.py
```diff
@@ -273,6 +277,20 @@ def join(self, timeout: Optional[float] = None) -> None:
 
 json_row_encoder = JSONRowEncoder()
 
+values_row_encoders: MutableMapping[StorageKey, ValuesRowEncoder] = dict()
```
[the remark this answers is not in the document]
onewland: I guess I thought it was inefficient to keep calling `table_writer.get_writeable_columns()`, but I'm happy to remove the memoization

onewland:
> I liked the idea of having two separate PR, one with the infra and one with the specific feature. Is there any specific reason for merging everything into one ?

How would you like this split?

I thought it was somewhat important that I get the values row encoding with the metrics aggregate processors to be sure that it met our requirements. But now that I know it all functions together, I can split this up for easier reading if you think that would be helpful

snuba/clickhouse/http.py
onewland: > It is true that, on metrics we only have integers, but the encoder needs to work in a generic environment, we need to support floats and arrays of string and floats.

Is this actually true? I think all input should be handled properly in the sense of "throw an error for input we can't properly encode" but I don't see why we need to make this generalized if we only intend to use it for this one case (which itself is a bit of an experiment)

snuba/cli/devserver.py
fpacifici: "I think we would be duplicating data this way ..."

snuba/clickhouse/http.py
fpacifici: > I think all input should be handled properly in the sense of "throw an error for input we can't properly encode"

A function that takes a string `f('asdasdasdasd')` would not fail to encode if the string needed escaping `asdasd'asdasd`. The insert in clickhouse would fail.
The problem, when that happens, is that the clickhouse INSERT would fail but since the INSERT is happening in batch, the batch itself will be too long to find the specific bad field (it gets truncated) and we won't know where the issue is. This already happened.

Anyway, if you want to implement only what is needed for this specific case for the sake of the experiment, that's ok, please add a TODO so you can provide a full implementation pf what is needed if the experiment works. 
Still please provide a unit test for the encoder.

Also we have a module with escaping function ready for use 
https://github.com/getsentry/snuba/blob/0429dc30390cc70221aa121b57057bf4ba324c98/snuba/clickhouse/escaping.py#L20-L42

snuba/clickhouse/value_types.py
fpacifici: Why not just using what we already have to format queries?
Create an AST expression:
`FunctionCall(whatever whatever)`
And then, in the encoder, just format it with the formatter we already have
https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/formatter/expression.py#L172

It would take care of any sort of escaping and formatting for you and it is thoroughly tested as we use it to format the AST into a clickhosue query.

The downside is that it could increase the CPU usage, we would have to try. I think it would be fine to do that as a follow up

snuba/consumers/consumer.py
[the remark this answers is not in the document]
fpacifici: mh, ok. Yes the schema is large. Feel free to leave it here

snuba/clickhouse/value_types.py
onewland: Sure, I'm fine with that

snuba/datasets/metrics_aggregate_processor.py
onewland: The arrayReduce is actually necessary or I get an error like 

```
INSERT INTO metrics_distributions_local (org_id, project_id, metric_id, granularity, timestamp, retention_days, tags.key, tags.value, percentiles, min, max, avg, sum, count) VALUES


Exception on client:
Code: 184. DB::Exception: Aggregate function minState(24.) is found in wrong place in query
```

for
```
insert into metrics_distributions_local (org_id,project_id,metric_id,granularity,timestamp,retention_days,tags.key,tags.value,percentiles,min,max,avg,sum,count)values (1, 2, 8, 10, toStartOfInterval(toDateTime('2022-01-05T21:35:10'), toIntervalSecond(10)), 90, [4, 5, 6, 9], [117, 7, 91, 134], arrayReduce('quantilesState(0.5,0.75,0.9,0.95,0.99)', [24.0, 80.0, 119.0, 146.0, 182.0]), minState(24.0), maxState(182.0), arrayReduce('avgState', [24.0, 80.0, 119.0, 146.0, 182.0]), sumState(551.0), countState(5))
```

I think the array reduction makes the data types line up for Clickhouse internally (didn't see a lot of good results when Googling this error message).

snuba/cli/devserver.py
onewland: I thought we were trying to keep the buckets table writes as a diagnostic thing for a while. Was I wrong on that?

snuba/datasets/metrics_aggregate_processor.py
onewland: Made this change in my latest commits

onewland:
Closing this in favor of two smaller changes (with the exact same content when combined):

https://github.com/getsentry/snuba/pull/2360
https://github.com/getsentry/snuba/pull/2361

PR #2348, opened 2022-01-25
feat(suspect-spans): Add array has processor

Zylphrex:
SnQL only allows for conditions of the form `LHS OPERATOR RHS`. This means that
conditions like `has(column, value)` are currently illegal, and has to be
written as `has(column, value) = 1` to satisfy this constraint. This causes
problems with Clickhouse when it's not able to use the bloomfilter index for
these conditions. This change rewrites such conditions to remove the surrounding
`... = 1` so that Clickhouse is able to properly leverage the bloomfilter index.

### DIFF
--- a/snuba/query/processors/array_has_optimizer.py
+++ b/snuba/query/processors/array_has_optimizer.py
@@ -2,10 +2,7 @@ from typing import Sequence
 
 from snuba.clickhouse.processors import QueryProcessor
 from snuba.clickhouse.query import Query
-from snuba.query.expressions import Column as ColumnExpr
 from snuba.query.expressions import Expression
-from snuba.query.expressions import FunctionCall as FunctionCallExpr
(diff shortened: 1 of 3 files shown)


snuba/query/processors/array_has_optimizer.py
```diff
@@ -0,0 +1,62 @@
...
+
+            # The arrayJoins we are looking for are not present, so skip this entirely
+            if match is None:
+                return expr
+
+            col = match.string("col")
+            val = match.string("val")
+
+            return FunctionCallExpr(
+                None, "has", (ColumnExpr(None, None, col), LiteralExpr(None, val)),
```
evanh, line 59: You shouldn't discard the alias and the table name of the column here. I would capture the entire Column expression instead of just the column name.

evanh (approved):
One small change, otherwise looks good.

PR #2349, opened 2022-01-26
feat(errors): Add a new errors_v2 storage option

nikhars:
### Context
We would like to introduce a new storage which would be used to write to the new errors clusters in production.

### Testing
Tested on my local cluster with the new storage pointing to a different cluster and changing errors consumer to be multi storage consumer. Verified that the count of errors on both the clusters were the same


snuba/datasets/storages/errors_v2.py
```diff
@@ -0,0 +1,56 @@
...
+    local_table_name="errors_local",
+    dist_table_name="errors_dist",
+    storage_set_key=StorageSetKey.ERRORS_V2,
+    mandatory_conditions=mandatory_conditions,
+    part_format=[util.PartSegment.RETENTION_DAYS, util.PartSegment.DATE],
+)
+
+storage = WritableTableStorage(
+    storage_key=StorageKey.ERRORS_V2,
+    storage_set_key=StorageSetKey.ERRORS_V2,
```
nikhars, line 33: This file is similar to https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/errors.py except for the storage keys over here and over here https://github.com/getsentry/snuba/pull/2349/files#diff-85d48fb2f4bfd3f7b16a317b13b37953c66c95ef61ff41544af53e23536f8178R26

lynnagara:
Is the new storage set intended to be temporary?

nikhars:
> Is the new storage set intended to be temporary?

Yes it would be temporary until we migrate completely to the new cluster.

lynnagara:
All storage sets need to be assigned to a cluster... unless you want to special case this and make it optional

fpacifici:
> All storage sets need to be assigned to a cluster... unless you want to special case this and make it optional

Not anymore. We had a deploy safety issue:
- The validation of settings, expected every storage set declared in the settings to be present in the code base
- This check expected that each declared storage set to be present in settings as assigned to a cluster.

https://github.com/getsentry/snuba/pull/2334/files#diff-7df145bde25ca970ceb9ba5c6fd95515b040310b905d4bfc9fc1e09b6e3eafa2L338-L341

This would make it impossible to deploy in prod where we cannot change settings and code atomically. 
Now we deploy the storage set in the code base first
And then we change the setting to assign the storage set to a cluster.

snuba/datasets/storages/errors_v2.py
```diff
@@ -0,0 +1,56 @@
...
+    schema=schema,
+    query_processors=query_processors,
+    query_splitters=query_splitters,
+    mandatory_condition_checkers=[ProjectIdEnforcer()],
+    stream_loader=build_kafka_stream_loader_from_settings(
+        processor=ErrorsProcessor(promoted_tag_columns),
```
fpacifici, line 39: Please ensure that the processor itself does not mutate the original message. That would be a problem with the multi storage consumer as you saw for transactions.

snuba/datasets/storages/errors_v2.py
```diff
@@ -0,0 +1,56 @@
...
+        subscription_scheduler_mode=SchedulingWatermarkMode.PARTITION,
+        subscription_scheduled_topic=Topic.SUBSCRIPTION_SCHEDULED_EVENTS,
+        subscription_result_topic=Topic.SUBSCRIPTION_RESULTS_EVENTS,
+    ),
+    replacer_processor=ErrorsReplacer(
+        schema=schema,
+        required_columns=required_columns,
+        tag_column_map={"tags": promoted_tag_columns, "contexts": {}},
+        promoted_tags={"tags": list(promoted_tag_columns.keys()), "contexts": []},
+        state_name=ReplacerState.ERRORS,
```
fpacifici, line 53: You need to provide a different value for this enum.
The ReplacerState is what we use in the replacements to separate the key space in redis between multiple replacers (we used it in the migration from events to errors).
So, in order to run two independent replacers you need two separate states here.

fpacifici (changes requested):
Please fix the replacer state, otherwise it looks good.

lynnagara:
Just to confirm do we really want to go down this route of duplicating the storage and filling them both just to do a ClickHouse upgrade? As opposed to just running the upgrade on the existing servers after testing it all in a staging environment

fpacifici:
This time yes. 
We need to upgrade in a timely manner because the version we have is out of support, then we will build what we need to test it in staging for the following upgrade

PR #2351, opened 2022-01-26
feat(admin) Expose tracing endpoint in the UI

evanh:
Expose the tracing endpoint with a UI. It follows the same format as the
Clickhouse system queries form, but it can run any arbitrary query. The result
is the raw trace logs that Clickhouse produces when running the query.

Screen Shot 2022-01-26 at 4 10 54 PM


vladkluev:
For frontend changes, can you post a screenshot or screenrecording of the changes?

snuba/admin/static/components/query_display/types.tsx
```diff
@@ -0,0 +1,31 @@
+type ClickhouseNode = {
```
onewland, line 1: Was this file copied from somewhere and we forgot to delete the old one?

snuba/admin/static/components/query_display/query_display.tsx
```diff
@@ -0,0 +1,199 @@
+import React, { useEffect, useState } from "react";
+import Client from "../../api_client";
+import { Table } from "../../table";
+import { COLORS } from "../../theme";
+
+import { ClickhouseNodeData, QueryRequest, QueryResult } from "./types";
+
+type QueryState = Partial<QueryRequest>;
+
+function QueriesDisplay(props: {
```
onewland, line 10: nit: can we make the class name QueryDisplay or the filename queries_display.tsx?

onewland (approved):
Looks good to me

PR #2353, opened 2022-01-27
feat(storage) New transactions_v2 storage

nikhars:
Add a new storage `TRANSACTIONS_V2` which would be used to populate the new cluster. Tested it out locally and verified that I am able to write to 2 storages using the multi storage consumer.


snuba/datasets/transactions_processor.py
```diff
@@ -231,12 +231,17 @@ def _process_contexts_and_user(
...
                 del contexts[context]
 
-        transaction_ctx = contexts.get("trace", {})
         # We store trace_id and span_id as promoted columns and on the query level
         # we make sure that all queries on contexts[trace.trace_id/span_id] use those promoted
         # columns instead. So we don't need to store them in the contexts array as well
-        transaction_ctx.pop("trace_id", None)
-        transaction_ctx.pop("span_id", None)
+        #
+        # When we remove the trace_id/span_id from transaction_ctx, it breaks the consumer if the
```
evanh, line 238: We don't want to leave these fields in the context either, they take up an enormous amount of space that we don't need. I think you need to come up with a different solution here.

evanh (changes requested):
Mostly looks good but we need to come up with a different solution for the processor changes. We pop those values out of the context so that they don't get stored in the context array in Clickhouse. They take up a lot of space.

evanh (approved):
OK this looks good. Thanks!

PR #2354, opened 2022-01-27
fix(jobs) Add sentry SDK to the cleanup and optimize jobs

evanh:
Add the SDK to the jobs so we hopefully get more meaningful error messages.


onewland (approved):
Any way to test this before deploy to prod? Maybe staging?

evanh:
> Any way to test this before deploy to prod? Maybe staging?

There's no staging snuba to my knowledge, and even if there was it would still be called against the prod Clickhouse. So unfortunately we have no good way to test this.

PR #2355, opened 2022-01-28
feat(subscriptions): Cache crc32 subscription ID

lynnagara:
crc32 can be a bit slow. Since we are constantly running this function
on the same set of subscription IDs, lets cache the results.

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -4,6 +4,7 @@ import logging
 import time
 from collections import deque
 from concurrent.futures import Future, ThreadPoolExecutor
+from functools import lru_cache
 from typing import Callable, Deque, Mapping, Optional, Sequence, Tuple, cast
 from zlib import crc32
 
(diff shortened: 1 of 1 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -274,8 +282,8 @@ def submit(self, message: Message[KafkaPayload]) -> None:
 
         subscription_id = str(task.task.subscription.identifier)
         should_execute = (
-            (crc32(subscription_id.encode("utf-8")) & 0xFFFFFFFF) / 2 ** 32
-        ) < executor_sample_rate
+            subscription_id_to_float(subscription_id) < executor_sample_rate
+        )
```
fpacifici, line 286: THere is something wrong here. 
We see the metric `skipped_execution` being there 10-20k times per second. 
But that would be the number of messages coming from the scheduler.
With 10k subscriptions registered, we should never have more than 100ish queries scheduled per second. 

I think we may be scheduling way too much.

fpacifici (approved):
Please see my comment on the metric we are recording. It does not match with the load we should have but it seems 2 orders of magnitude higher

snuba/subscriptions/executor_consumer.py
lynnagara: Will look into it

lynnagara: btw, we are intentionally scheduling double - we are running a load test that does multiply the number of subscriptions to simulate a higher load

PR #2358, opened 2022-01-29
fix(replacer): Replacer gets stuck

rahul-kumar-saini:
## Overview

**Problem:**

- Replacements are executed in batches, the consumer does not poll for new messages while the batch is being executed
- The broker will wait for the consumer to poll again for a certain amount of time before kicking the consumer
- Sometimes the batch of replacements takes longer than the consumer timeout to execute
- When the consumer restarts, it does not know what to process and the entire replacer gets stuck

**Plan:**

- After a long INSERT is successful, write the Kafka offset on Redis
- The next time a message is processed, first read the offset from Redis
    - This covers the case where the consumer fails and restarts, since it would try to process the next message and it will continue ignoring messages till a new one is presented
- If the offset is the same or greater than the message we are about to process, skip the message
- This ensures that we do not get stuck as long as ClickHouse succeeds in the query

**Details:**

- The idea here is to maintain a single key value pair (per topic, consumer group, partition) in Redis to track the offset that was processed if it took greater than a certain amount of time to execute
- The format would be like:
    - `topic:consumer_group:paritition` → `offset`
- In order to know when the replacement starts/stops being processed, we can use the `process_message()`, `pre_replacement()`, and `post_replacement()` methods of the [`ReplacerProcessor`](https://github.com/getsentry/snuba/blob/master/snuba/replacers/replacer_processor.py#L57-L68)
- `process_message()` will be called every time a message is received, and it will add to a batch of Replacements
- `pre/post_replacement()` are called per replacement when the batch is being executed

**Workflow:**

- `process_message()` looks into Redis and finds one of the following:
    - `topic:consumer_group:partition` → `x`
        - In this case we do not process if the current offset is ≤ `x`
    - Nothing
        - In this case we continue processing messages as normal
- `pre_replacement()` records the time before executing the Replacement
- `post_replacement()` compares the time before and after executing the Replacement
    - If the difference between the times is ≥ 2 mins, write the offset to Redis with a TTL
    - Overwrite any existing entry if it has not expired already

#### Visualization

Some set of steps that may occur:
```python
m1 = Message(offset=1)
m2 = Message(offset=2)

batch: Sequence[Replacement] = []
batch.append(replacer.process_message(m1))
batch.append(replacer.process_message(m2))

replacer_worker.flush_batch(batch)
```
Here's what would be going on in `flush_batch()`:
```python
replacer.pre_replacement(r1)
query_executer.execute(r1.get_insert_query())
replacer.post_replacement(r1)

replacer.pre_replacement(r2)
query_executer.execute(r2.get_insert_query())
replacer.post_replacement(r2)
```
Now if any of the replacements, let's say `r2`, takes too long to execute, `post_replacement()` would write the offset to Redis. If the consumer restarts since it took too long, it will start processing messages again from the beginning of the batch, but will skip along since it knows which ones were already processed.

## Changes
- `Replacement` class now has a field to store `ReplacementMessageMetadata`, this object contains info from the original Kafka message
- `process_message()`, `pre_replacement()`, and `post_replacement()` updated to reflect stuff written in the **Workflow** section above

## Testing
- Added a couple of tests to verify messages are ignored correctly
- Tested messages being ignored locally by spinning up Sentry + Snuba and monitoring messages


snuba/settings/__init__.py
```diff
@@ -133,6 +133,8 @@
 REPLACER_KEY_TTL = 12 * 60 * 60
 REPLACER_MAX_GROUP_IDS_TO_EXCLUDE = 256
 REPLACER_IMMEDIATE_OPTIMIZE = False
+REPLACER_PROCESSING_TIMEOUT_THRESHOLD = 2 * 60 * 1000  # 2 minutes in ms
+REPLACER_PROCESSING_TIMEOUT_THRESHOLD_KEY_TTL = 60 * 60  # 1 hour in s
```
onewland, line 137: I don't disagree with it, but what is the negative result if we don't set a key TTL here. Just trying to understand the logic (and why we'd want it to be an hour)

onewland (approved):
I think this looks good overall, but I would recommend considering adding a prefix constant

snuba/datasets/errors_replacer.py
```diff
@@ -565,6 +614,22 @@ def pre_replacement(self, replacement: Replacement, matching_records: int) -> bo
...
         return False
 
+    def post_replacement(self, replacement: Replacement, matching_records: int) -> None:
+        """
+        Write the offset just processed to Redis if execution took longer than the threshold.
+        """
+        if (
+            time.time() - self.__start_time
+        ) < settings.REPLACER_PROCESSING_TIMEOUT_THRESHOLD:
+            return
```
fpacifici, line 624: This time is in seconds, not milliseconds, but you set your threshold in milliseconds.

fpacifici (changes requested):
Please see my comment inline

snuba/datasets/errors_replacer.py
```diff
@@ -526,8 +541,42 @@ def process_message(self, message: ReplacementMessage) -> Optional[Replacement]:
 
         return processed
 
+    def _message_already_processed(self, message: ReplacementMessage) -> bool:
```
fpacifici, line 544: WHat is the reasoning for putting this login in `errors_replacer.py` in stead of the more abstract `replacer.py`?
https://github.com/getsentry/snuba/blob/master/snuba/replacer.py#L334-L383

I think there is a case for moving it in the `ReplacerWorker`. 
- errors_replacer is independent from Kafka, the replacement object does not need to know about offsets or partitions. So if you move it to replacer.py you would not have to introduce these concerns here. replacer.py instead knows about kafka
- if you move it to replacer.py you will be able to reuse this logic in other replacements (beyond errors).

snuba/datasets/errors_replacer.py
```diff
@@ -526,8 +541,42 @@ def process_message(self, message: ReplacementMessage) -> Optional[Replacement]:
...
+
+        Check whether there is an entry in Redis for this specific
+        topic, consumer group, and partition. If there is, we can conclude whether
+        or not to process the incoming message by comparing the incoming message
+        to the offset in Redis.
+        """
+        key = self._build_topic_group_index_key(message.metadata)
+        processed_offset = redis_client.get(key)
+        if processed_offset is not None:
+            offset = int(processed_offset)
```
fpacifici, line 556: If this fails you can clean up the redis key (so the replacement does not get stuck). If somebody corrupted the data in Redis and the content is not an int anymore, there is no point in keeping the state around as there is nothing we will be able to do to fix it.  Just clean it up, log an error and go on.

snuba/datasets/errors_replacer.py
```diff
@@ -526,8 +541,42 @@ def process_message(self, message: ReplacementMessage) -> Optional[Replacement]:
...
+        """
+        Figure out whether or not the message was already processed.
+
+        Check whether there is an entry in Redis for this specific
+        topic, consumer group, and partition. If there is, we can conclude whether
+        or not to process the incoming message by comparing the incoming message
+        to the offset in Redis.
+        """
+        key = self._build_topic_group_index_key(message.metadata)
+        processed_offset = redis_client.get(key)
```
fpacifici, line 554: You do not need to read from redis at every iteration. 
This value could be useful for the current consumer only right after the start. 
Until we are caught up with messages that need to be dropped.
If an insert is successful, and the commit is successful, the following iteration does not care about what is in redis, it cannot be a message to skip. 
I think you should read it at the first iteration, keep the value in a variable, and avoid reading from redis if the variable is already populated.

snuba/datasets/errors_replacer.py
```diff
@@ -526,8 +541,42 @@ def process_message(self, message: ReplacementMessage) -> Optional[Replacement]:
...
+    ) -> str:
+        """
+        Builds a unique key for a message being processed for a specific
+        topic, consumer group, and partition.
+        """
+        return ":".join(
+            [
+                message_metadata.topic_name,
+                self.__state_name.value,
+                str(message_metadata.partition_index),
```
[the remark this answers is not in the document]
rahul-kumar-saini: Sure, I'll add something like `processed:` since the info really means this offset was processed.

snuba/settings/__init__.py
rahul-kumar-saini: Just to prune old data since after a certain amount of time, that data isn't useful anymore cause a consumer would have picked up or passed the offset. I don't think the offset number itself ever resets to 0 but if for some reason it does,  having an offset existing in Redis would be really bad since it would skip messages till the number in Redis.

The TTL is an hour in case the consumer takes a long time to restart or if we're troubleshooting an issue.

snuba/datasets/errors_replacer.py
rahul-kumar-saini: I had this logic written in `ReplacerWorker` to start with actually, but at the worker level I do not have access to the Replacer state (events/errors). I moved the logic into `errors_replacer.py` to get the state to use as part of the Redis key.

I think I can create method like `get_state()` on the base `ReplacerProcessor` class to get around this, if that would be a better way to do it.

fpacifici: Yes, I think exposing the state to the parent class will be better than exposing Kafka details to the worker and make it errors specific.

snuba/replacer.py
```diff
@@ -381,3 +403,64 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+    def _build_topic_group_index_key(
+        self, message_metadata: ReplacementMessageMetadata
+    ) -> str:
+        """
+        Builds a unique key for a message being processed for a specific
+        topic, consumer group, and partition.
+        """
+        return ":".join(
+            [
+                "processed",
```
fpacifici, line 461: `processed` can mean anything.
Please prefix this with `replacement`

snuba/replacer.py
```diff
@@ -381,3 +403,64 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+                message_metadata.topic_name,
```
fpacifici, line 462: Why does the topic name have to be in the replacement metadata? The individual replacement does not need to know about topics and consumer groups.
Also I think you will have to add the consumer group to the cache key as well.

snuba/replacer.py
```diff
@@ -381,3 +403,64 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+    def _message_already_processed(self, metadata: ReplacementMessageMetadata) -> bool:
+        """
+        Figure out whether or not the message was already processed.
+
+        Check if there exists a recorded offset that took too long to execute.
+        If there is, we can conclude whether or not to process the incoming
+        message by comparing the incoming message to the recorded offset.
+        """
+        # float("-inf") implies this is the first message this worker is processing since startup
+        if self.__prolonged_offset == float("-inf"):
```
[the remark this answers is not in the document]
rahul-kumar-saini: I had `None` to begin with but mypy complained about setting an int to a Nonetype. I could use any other holding int like -2, I chose `float("-inf")` since it's a common way of denoting a default low number.

snuba/replacer.py
rahul-kumar-saini: The topic is lost after `process_message()` since it just returns a None or Replacement. It needs to be in the metadata for the replacement itself unless there's another way to save that context for the offset. Same goes for the partition.

I can store the topic and partition as variables local to the worker itself, but that assumes the worker will always be reading from the same topic+partition. I'm not sure if this is the case.

As for the consumer group, I didn't see a way to get that info from a Kafka message itself but I believe the replacer processor's state is enough to identify a difference in consumer groups, so I added that state to the cache key.

snuba/replacers/replacer_processor.py
```diff
@@ -1,10 +1,26 @@
...
 from snuba.datasets.events_processor_base import ReplacementType
 from snuba.datasets.schemas.tables import WritableTableSchema
 
 
+class ReplacerState(Enum):
+    EVENTS = "events"
+    ERRORS = "errors"
+
+
+class ReplacementMessageMetadata(NamedTuple):
```
nikhars, line 14: This should contain the ReplacerState as well.

I intend to have another consumer group running with a different ReplacerState for the new errors cluster. The existing parameters would overlap for both the consumer groups. So we need to be able to separate their processing logic.

snuba/replacer.py
nikhars: You can get rid of topic from here. Replacements would always be written to the same topic. Partition assignments for a consumer could change.

Having consumer group in the key would be good since what you are tracking is consumer group's progression across partitions. So it seems logical to have that in the key.

You should be able to use consumer group from https://github.com/getsentry/snuba/blob/master/snuba/datasets/events_processor_base.py#L180-L182 and pass it to https://github.com/getsentry/snuba/blob/master/snuba/cli/replacer.py#L119

nikhars:
Please add a setting/option to the replacer to bypass the redis checks for specific consumer groups.

Use case is that we would want to be able to reset the offset of the replacer so that it starts off from the beginning. We need a mechanism to bypass these checks which you have added here temporarily

snuba/replacer.py
rahul-kumar-saini: Alright, I've got the consumer group and topic passed to the Worker on instantiation. ReplacerWorker now requires these params:
```python
storage: WritableTableStorage,
replacements_topic: str,
consumer_group: str,
metrics: MetricsBackend
```

The 2 new strings are just saved locally to the worker and are used when building the cache key.

snuba/replacer.py
```diff
@@ -454,12 +460,13 @@ def _build_topic_group_index_key(
...
         Builds a unique key for a message being processed for a specific
-        topic, consumer group, and partition.
+        consumer group and partition.
         """
         return ":".join(
             [
                 "replacement",
-                message_metadata.topic_name,
+                self.__consumer_group,
+                self.__replacements_topic,
```
nikhars, line 469: You don't need a replacements topic in the key. The reason being there wouldn't be multiple replacement topics. All the consumer groups would work on the same topic. 

You can remove the storing/passing of topics from the PR altogether.

snuba/replacer.py
```diff
@@ -30,6 +37,8 @@
 
 NODES_REFRESH_PERIOD = 10
 
+RESET_CHECK_CONFIG = "consumer_groups_to_reset_offset_check"
```
rahul-kumar-saini, line 40: Open to suggestions if anyone opposes this name for the runtime config.

snuba/replacer.py
```diff
@@ -381,3 +410,85 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+
+        There exists a config which tells us which consumer groups require their
+        replacer(s) reset. Ideally this config is populated with consumer groups
+        temporarily, then cleared once relevant consumers restart.
+        """
+        # expected format is "[consumer_group1, consumer_group2, ..]"
+        consumer_groups = (get_config(RESET_CHECK_CONFIG) or "[]")[1:-1].split(",")
+        if self.__consumer_group in consumer_groups:
+            self.__prolonged_offset = -1
+            redis_client.delete(key)
```
rahul-kumar-saini, line 457: Expected format of the runtime config is a string due to the way runtime configs are stored (string, int, or float).

If this consumer group requires the offset to be reset, the replacer clears the offset it compares incoming messages to. We can then clear the config.

rahul-kumar-saini (commented):
I'm introducing a runtime config to address the point Nikhar brings up about resetting Kafka offsets manually.

Workflow would be something like:
- `consumer_groups_to... = "[]"`     (runtime config)
- Now let's say we need to reset offset for some topic, consumer groups “a” and “b” read from that topic
- Set  `consumer_groups_to... = "[a, b]"`
- Now we manually restart consumer groups “a” and “b” (or send at least one message to each of the consumers to trigger the check)
- Set `consumer_groups_to... = "[]"`

snuba/replacer.py
```diff
@@ -381,3 +410,85 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+
+        Eg. The offset is manually reset in Kafka to start processing messages
+        again from an older offset. The replacer will ignore this and just skip
+        messages till it's back to the offset stored in Redis.
+
+        There exists a config which tells us which consumer groups require their
+        replacer(s) reset. Ideally this config is populated with consumer groups
+        temporarily, then cleared once relevant consumers restart.
+        """
+        # expected format is "[consumer_group1, consumer_group2, ..]"
```
evanh, line 453: This is a bit of a nit but the comment here has spaces after the commas. I assume that's not accurate?

rahul-kumar-saini: You're right, force of habit 🤦‍♂️

snuba/replacer.py
```diff
@@ -267,6 +281,9 @@ def __init__(self, storage: WritableTableStorage, metrics: MetricsBackend) -> No
         self.__sharded_pool = RoundRobinConnectionPool(self.__storage.get_cluster())
         self.__rate_limiter = RateLimiter("replacements")
 
+        self.__prolonged_offset = float("-inf")
```
nikhars, line 284: You could just use the value "-1" here like you do when bypassing checks.

nikhars (approved):
Left a small comment. Otherwise looks good.

snuba/replacer.py
rahul-kumar-saini: `float("-inf")` is necessary to tell the replacer to look in Redis. `-1` means we know there is nothing in Redis. This allows us to only read from Redis once and then maintain the state locally as well as in Redis.

snuba/replacer.py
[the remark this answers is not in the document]
fpacifici: Back to this comment, sorry for the delay.
Please do not use the float. Offsets are integers.
If you need that to be nullable, make it Optional[int] instead of int.
Turning the offset into a float while it is an int in order to avoid the nullability issue introduces a hack to solve a problem that can be solved properly with nullability.

snuba/replacer.py
fpacifici: Could we renamed this as, "last_offset_processed"?

snuba/replacer.py
```diff
@@ -332,12 +349,26 @@ def run_query(
...
+        )
+        if self._message_already_processed(metadata):
+            logger.warning(
+                f"Replacer ignored a message, consumer group: {self.__consumer_group}",
+                extra={
+                    "partition": metadata.partition_index,
+                    "offset": metadata.offset,
+                },
+            )
+            return None
```
fpacifici, line 363: You should rollout this behind a feature flag just in case.
Add a runtime flag here, that, when you rollout, will be disabled, if the flag is disabled jsut log and go on. If the flag is enabled, actually return None.
This allows you to do a safe deployment in case of bugs. Then you can activate and remove the flag once you are sure you are not logging these messages.

snuba/replacer.py
```diff
@@ -381,3 +417,85 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+        key = self._build_topic_group_index_key(metadata)
+        self._reset_offset_check(key)
+
+        # float("-inf") implies this is the first message this worker is processing since startup
+        if self.__prolonged_offset == float("-inf"):
+            processed_offset = redis_client.get(key)
+            try:
+                self.__prolonged_offset = (
+                    -1 if processed_offset is None else int(processed_offset)
+                )
```
fpacifici, line 438: I don't think this will work as you want.
Different partitions of the same topic can have reached very different offsets. Each partitions has a monotonic offset, these offsets are independent from each other. 
So on partition 1 the last offset may be 5000 while on partition 2 the last offset can be 200,000. 
This means you will have to keep track the last offset from all partitions, you cannto have a single `self.__prolonged offset`.

Let's say you have an issue when committing partition 2 (offset 200,000).
Let's say you load the offset from redis for partition 2: that yields 200,000. Now you load that in memory, from redis when you restart. If you get a message from partition 1 with offset 5001 you will discard it while you never processed it.
Basically you cannot compare offsets across partitions. You can only compare offsets of the same partition.

snuba/replacer.py
```diff
@@ -381,3 +417,85 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+    def _message_already_processed(self, metadata: ReplacementMessageMetadata) -> bool:
+        """
+        Figure out whether or not the message was already processed.
+
+        Check if there exists a recorded offset that took too long to execute.
+        If there is, we can conclude whether or not to process the incoming
+        message by comparing the incoming message to the recorded offset.
+        """
+        key = self._build_topic_group_index_key(metadata)
+        self._reset_offset_check(key)
```
fpacifici, line 430: You actually cannot delete it just yet at this point. 
You would have to wait for a successful commit before deleting. If the consumer restarts and fails again at committing you would have removed the key. 
So you would have to delete only after the commit.
I think that may be more complicated than letting the keys expire, probably not worth trying to delete.

snuba/replacer.py
```diff
@@ -381,3 +417,85 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
...
+                self.__prolonged_offset = (
+                    -1 if processed_offset is None else int(processed_offset)
+                )
+            except ValueError as e:
+                redis_client.delete(key)
+                logger.warning(
+                    "Unexpected value found for an offset in Redis", exc_info=e,
+                )
+
+        return metadata.offset <= self.__prolonged_offset
```
fpacifici, line 445: Here you will have to match the offset of the partition of the message you received with the stored offset for that partition. See the comment above for more details on the issue.

fpacifici (changes requested):
I think there are still a few issues. Please see the comments inline

snuba/replacer.py
rahul-kumar-saini: This method isn't mean to remove the key every time, it's only for the runtime config fix for when we need to manually roll back the offset in Kafka.

If you take a look at the method you'll see it only resets if necessary:
https://github.com/getsentry/snuba/blob/3737fe7a89f78c3882f351be1377c0f4a278f1e9/snuba/replacer.py#L464-L468

snuba/replacer.py
rahul-kumar-saini: I've renamed it but I feel it may be a bit misleading, it's the last offset processed _that took more than 2 min to process_ which is why I went with "prolonged" offset initially.

fpacifici (approved):

PR #2359, opened 2022-01-31
feat: Share ClickHouse connections to the same nodes between clusters

lynnagara:
Previously ClickHouse connection pools were cached in the ClickhouseCluster class so they would be reused when attempting to access a ClickHouse node via a cluster. However if the same nodes are present in multiple clusters (as is the case with the way Sentry's query nodes are currently configured) we would create a separate and unnecessary connection pool to those same nodes from each cluster.

This change addresses that issue by moving the connection cache outside of the ClickHouseCluster class and reusing the same cache for all clusters.


snuba/clusters/cluster.py
```diff
@@ -134,6 +134,31 @@ def get_batch_writer(
...
+
+    def __setitem__(
+        self,
+        key: Tuple[ClickhouseNode, ClickhouseClientSettings],
+        value: ClickhousePool,
+    ) -> None:
+        self.__cache[key] = value
+
+
+connection_cache = ConnectionCache()
```
vladkluev, line 159: My only concern with this global variable is that we use threading throughout the application. This object is not thread safe, are the ClickousePools threadsafe?

snuba/clusters/cluster.py
```diff
@@ -134,6 +134,31 @@ def get_batch_writer(
 ClickhouseWriterOptions = Optional[Mapping[str, Any]]
 
 
+class ConnectionCache:
```
fpacifici, line 137: The maximum number of connections we store in a `ClickhousePool` is 20. After that the client waits for one to be free. 
If now we are going to reuse them we will have fewer `ClickhousePool` instances active, each one with a higher number of active connections. We need to ensure before hand that the total amount is not close to 20 otherwise clients will get stuck waiting for a connection.

snuba/clusters/cluster.py
lynnagara: I guess it is possible for us to set the cache key/value simultaneously in separate threads instead of reusing the same connection pool. Though this is also the case today and it seems relatively harmless. Are there more serious concerns I am not considering?

snuba/clusters/cluster.py
lynnagara: Are you talking about `settings.CLICKHOUSE_MAX_POOL_SIZE`? We have the default set to 25 and don't seem to get close to that currently. Though it may make sense to ultimately allow this to be configured per process.

fpacifici: > don't seem to get close to that currently.
How did you find this out?

My concern is that, with this change, all the connections for all the clusters defined in the prod config will share the same `ClickhousePool` (now they are not, they are separate). Are we going to hit that 25 when we merge the connections for all clusters in the same `ColickhousePool` ?

lynnagara: I looked at the metric `snuba.clickhouse.native.connections`. It normally hovers around 3 right now but it can spike up. Let me know if you think we need to bump it up a bit.

evanh: I don't think that metric gives an accurate picture. When I `sum by` that metric, it goes up as high as 90. I think it needs to be tagged with a storage or something so we can see: if we dedupe the connections, and have fewer pools, will we start maxing out the connections to those pools.

lynnagara: Sum by is not a meaningful aggregation to do on this metric since we might be running on hundreds of hosts (e.g. all the api, replacer, consumer, subscription processes). If you look at the max by host it's not often above 3 or 4.

tests/clusters/test_cluster.py
```diff
@@ -142,6 +142,10 @@ def test_cache_connections() -> None:
         "localhost", 8000, "default", "", "default", 8001, {"events"}, True
     )
 
+    cluster_2 = cluster.ClickhouseCluster(
+        "localhost", 8000, "default", "", "default", 8001, {"transsactions"}, True
```
MeredithAnya, line 146: not sure if it matters really but `transsactions` -> `transactions`

lynnagara: updated, thanks

snuba/clusters/cluster.py
fpacifici: ClickhousePool is thread safe. It uses `queue.LifoQueue` to store connections and it prevents them from being shared. This class indeed is not thread safe. I would synchronize the access just in case to avoid opening too many connections in case of race conditions.

fpacifici (commented):
I think it is worth using a thread safe structure for the connection cache. ClickhousePool is already safe.

snuba/clusters/cluster.py
lynnagara: Ok updated to avoid this scenario by making the connection cache manage the connections. In doing so I realised there was another issue where it would assume the user/password/database was the same across clusters and reuse those connections even if that wasn't the case. That should now be fixed as part of this.

fpacifici: This is still not thread safe as you can create more connections than you need, but better than before.

fpacifici (approved):

snuba/clusters/cluster.py
vladkluev: seems pretty easy to put a `threading.Lock` around either this or the `__cache` member. Why not just do that?

Also, can we add a test to make sure that when the username/password are different, a different connection is made. I know it's obvious from reading the code now but we should enshrine that behavior with a test

vladkluev (approved):

PR #2360, opened 2022-02-01
feat(consumer): Add support for VALUES format when writing to ClickHouse

onewland:
This is necessary for an upcoming change which will write aggregate expressions directly in the insert.

Functional changes:
- support for inserting using the VALUES format (https://clickhouse.com/docs/en/interfaces/formats/#data-format-values) at the WritableTableStorage level
- support in the multistorage consumer for serializing to the VALUES format (including a new ValuesRowEncoder)
- added a new type of return from processing called `AggregateInsertBatch` to indicate to callers that the data being returned needs to be interpreted by Clickhouse (e.g. it can contain function call expressions)

This was split out of https://github.com/getsentry/snuba/pull/2346 to be more readable (without the additional metrics changes). There will be an upcoming PR based off this one to support direct metrics aggregate writing

### DIFF
--- a/snuba/clickhouse/http.py
+++ b/snuba/clickhouse/http.py
@@ -15,6 +15,8 @@ from urllib3.exceptions import HTTPError
 from snuba import settings
 from snuba.clickhouse import DATETIME_FORMAT
 from snuba.clickhouse.errors import ClickhouseWriterError
+from snuba.clickhouse.formatter.expression import ClickhouseExpressionFormatter
+from snuba.clickhouse.query import Expression
 from snuba.utils.codecs import Encoder
 from snuba.utils.iterators import chunked
 from snuba.utils.metrics import MetricsBackend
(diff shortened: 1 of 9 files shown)


fpacifici (approved):
Seems good to me.

snuba/clickhouse/http.py
```diff
@@ -39,12 +41,31 @@ def __default(self, value: Any) -> Any:
...
             raise TypeError
 
-    def encode(self, value: WriterTableRow) -> JSONRow:
+    def encode(self, value: WriterTableRow) -> bytes:
         return cast(
             bytes, rapidjson.dumps(value, default=self.__default).encode("utf-8")
         )
 
 
+class ValuesRowEncoder(Encoder[bytes, WriterTableRow]):
```
fpacifici, line 50: PLease add some unit tests for this before merging.

onewland: I've added a few unit tests. Let me know if you'd like coverage of anything else.

fpacifici: It is ok to me.
Optional:
I would add some more complex expression(like function containing a function). 
Also what happens if you define an alias in an expression in the INSERT? I never tried.

PR #2361, opened 2022-02-01
feat(metrics): directly write aggregates to metrics table

onewland:
Add support for writing directly to aggregate tables for metrics (`metrics_distributions`, `metrics_sets`, and `metrics_counters`)

Functional changes:
- added support for direct aggregate writes in the existing metrics multistorage consumer
- new setting to switch between existing materialized view supported behavior and direct writes to aggregate tables

Non-functional changes:
- changed the storages for metrics aggregate tables to be writeable
- renamed original processors to indicate that they are for writing to buckets (to distinguish from aggregates)

Testing:
- added unit tests for the new metrics aggregate processors
- added a new e2e test to make sure that `AggregateInsertBatch` flows correctly
- did some ad hoc testing on my local machine to verify that data coming from Sentry is written to the tables properly 
  - This was only possible for distributions and counters as far as I know. I didn't see any sets flowing into Kafka but I didn't spend a lot of time seeing if there was a path for this.

This was split out of #2346 to be more readable (without the ValuesRowEncoder changes). It assumes that https://github.com/getsentry/snuba/pull/2360 has been merged or will be merged at the same time


snuba/cli/devserver.py
```diff
@@ -152,6 +152,19 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
...
+                    "snuba",
+                    "multistorage-consumer",
+                    "--storage=metrics_distributions",
+                    "--storage=metrics_counters",
+                    "--storage=metrics_sets",
+                    "--auto-offset-reset=latest",
+                    "--log-level=debug",
+                    "--consumer-group=metrics_aggregates_group",
+                ],
+            ),
```
fpacifici, line 167: Back to the original comment here. 
If you spin up two independent consumers you will have two independent consumer groups with independent offsets. The raw data and the aggregated ones will not be consistent with each other. 
I think you should jsut add the storages to the existing consumer so they share the same consumer group. 
Also, this way you will be duplicating data in the aggregated table because both the you will be writing both through the aggregate consumer and through the materialzied views that still exist. 

I think there is a larger discussion to decide how to roll this out (and how to roll it back if we decided to go back to mat views) in a consistent way that would not miss or duplicate data.

Could you please come up with a plan (here, no need for more documents) for the steps to achieve this knowing that in production the deployment configuration is not in the snuba code base so config and code cannot be deployed together.

snuba/datasets/storages/factory.py
```diff
@@ -68,6 +68,9 @@
             sessions_raw_storage,
             transactions_storage,
             spans_storage,
+            metrics_distributions_storage,
+            metrics_sets_storage,
+            metrics_counters_storage,
```
fpacifici, line 73: Please add them to the `METRICS_WRITEABLE_STORAGES` constant instead.

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,166 @@
...
+METRICS_SET_TYPE = "s"
+METRICS_DISTRIBUTIONS_TYPE = "d"
+METRICS_COUNTERS_TYPE = "c"
+
+
+class MetricsAggregateProcessor(MessageProcessor, ABC):
+    TEN_SECONDS = 10
+    ONE_HOUR = 3600
+    ONE_DAY = 3600 * 24
+    GRANULARITIES_SECONDS = [TEN_SECONDS, ONE_HOUR, ONE_DAY]
```
fpacifici, line 41: Pretty sure we have the 1 minute granularity as well.
https://github.com/getsentry/snuba/blob/master/snuba/migrations/snuba_migrations/metrics/templates.py#L243

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+        values = message["value"]
+        for v in values:
+            assert isinstance(v, int), "Illegal value in set. Int expected: {v}"
+
+        return {
+            "value": _call(
+                "arrayReduce", (_literal("uniqState"), _array_literal(values)),
+            )
```
fpacifici, line 107: Let's collect a datadog metric (sampled, let's not record a metric for every message) to have a sense of the size of our batches.

This is where you have to declare your metric to have a sampling rate set 
https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L57-L60

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+        value = message["value"]
+        assert isinstance(
+            value, (int, float)
+        ), "Illegal value for counter value. Int/Float expected {value}"
+
+        return {
+            "value": _call(
+                "arrayReduce", (_literal("sumState"), _array_literal([value]),),
```
fpacifici, line 123: What about  summing up all the content of the array and passing `sumState(array(THE_ACTUAL_SUM))` so you would pass a much smaller payload. 
Though, let's build this optimization as a follow up and gate it with a runtime flag so we can see if it improves things or make them worse.

fpacifici (changes requested):
I think it is mostly ok. 
Please see specifically my connect on the rollout plan. We should be able to roll this out and back preserving consistency.

snuba/datasets/metrics_aggregate_processor.py
onewland: I think `value` is a single int or float. So there's nothing to sum unless I'm wrong

snuba/cli/devserver.py
onewland: We spoke offline, and the `materialization_version` is used by the materialized views to only run if it is equal to 0, so I think I agree that we can combine these consumers into one and change the bucket writing code (in `MetricsBucketProcessor`) to set `materialization_version=1` so that the materialized views and the consumer don't both write to the aggregates tables.

Therefore my deploy plan would be:
1. merge this change (also setting materialization_version to 1 in the buckets processors)
2. ask ops to change the multistorage consumer configuration to also use the direct aggregate table storages as arguments
3. stop the consumer group and record the offset
4. do a deploy to restart the consumer group which now writes direct aggregates

I see two different rollback-requiring situations:
1. the aggregate consumer is too slow and can't keep up with the incoming data
2. the aggregate consumer keeps up with incoming data but corrupts data in one or more of the aggregate tables 

For Situation 1, where VALUES-writing is too slow, we can simply roll back the ops change and then the code change. Since the consumer offset won't advance beyond what the writers can process, reverting to the old logic should just "catch up" over time.

Situation 2 is a bit more complex. To fix data in the aggregate tables, we'll need to "undo" all the writes for materialization_version = 1 and rerun all the writes with materialization_version = 0. The process I'm imagining is:
1. remove all data from aggregate tables where the interval start is after the start time of the first record[s] written with materialization_version = 1
2. roll back the change, setting the materialization_version to 0
3. restart the consumer group from the offset where we stopped using the materialized views. This will cause a large number of row replacements but that should cause the aggregates to be rewritten with the original logic.

snuba/datasets/metrics_aggregate_processor.py
onewland: when I run migrations in my local I see:
```
7b701b64404c :) show tables;

SHOW TABLES

┌─name──────────────────────────────────┐
│ discover_local                        │
│ errors_local                          │
│ groupassignee_local                   │
│ groupedmessage_local                  │
│ metrics_buckets_local                 │
│ metrics_counters_buckets_local        │
│ metrics_counters_local                │
│ metrics_counters_mv_10s_local         │
│ metrics_counters_mv_3600s_local       │
│ metrics_counters_mv_86400s_local      │
│ metrics_counters_mv_local             │
│ metrics_distributions_buckets_local   │
│ metrics_distributions_local           │
│ metrics_distributions_mv_10s_local    │
│ metrics_distributions_mv_3600s_local  │
│ metrics_distributions_mv_86400s_local │
│ metrics_distributions_mv_local        │
│ metrics_sets_local                    │
│ metrics_sets_mv_10s_local             │
│ metrics_sets_mv_3600s_local           │
│ metrics_sets_mv_86400s_local          │
│ metrics_sets_mv_local                 │
│ migrations_local                      │
│ outcomes_hourly_local                 │
│ outcomes_mv_hourly_local              │
│ outcomes_raw_local                    │
│ sentry_local                          │
│ sessions_hourly_local                 │
│ sessions_hourly_mv_local              │
│ sessions_raw_local                    │
│ transactions_local                    │
└───────────────────────────────────────┘

31 rows in set. Elapsed: 0.011 sec. 
```

and on prod from the admin tool:
```
{
  "column_names": [
    "name"
  ],
  "rows": [
    [
      "metrics_buckets_dist"
    ],
    [
      "metrics_buckets_local"
    ],
    [
      "metrics_counters_buckets_dist"
    ],
    [
      "metrics_counters_buckets_local"
    ],
    [
      "metrics_counters_dist"
    ],
    [
      "metrics_counters_local"
    ],
    [
      "metrics_counters_mv_10s_local"
    ],
    [
      "metrics_counters_mv_3600s_local"
    ],
    [
      "metrics_counters_mv_86400s_local"
    ],
    [
      "metrics_counters_mv_local"
    ],
    [
      "metrics_distributions_buckets_dist"
    ],
    [
      "metrics_distributions_buckets_local"
    ],
    [
      "metrics_distributions_dist"
    ],
    [
      "metrics_distributions_local"
    ],
    [
      "metrics_distributions_mv_10s_local"
    ],
    [
      "metrics_distributions_mv_3600s_local"
    ],
    [
      "metrics_distributions_mv_86400s_local"
    ],
    [
      "metrics_distributions_mv_local"
    ],
    [
      "metrics_sets_dist"
    ],
    [
      "metrics_sets_local"
    ],
    [
      "metrics_sets_mv_10s_local"
    ],
    [
      "metrics_sets_mv_3600s_local"
    ],
    [
      "metrics_sets_mv_86400s_local"
    ],
    [
      "metrics_sets_mv_local"
    ],
    [
      "asynchronous_metrics"
    ],
    [
      "metrics"
    ]
  ],
  "input_query": "select name from system.tables where name like '%metrics%' (metrics_counters,snuba-metrics-0-1:9000)"
}

```

fpacifici:
```
CREATE MATERIALIZED VIEW default.metrics_sets_mv_local TO default.metrics_sets_local
(
    `org_id` UInt64,
    `project_id` UInt64,
    `metric_id` UInt64,
    `granularity` UInt32,
    `tags.key` Array(UInt64),
    `tags.value` Array(UInt64),
    `timestamp` DateTime,
    `retention_days` UInt16,
    `value` AggregateFunction(uniqCombined64, UInt64)
) AS
SELECT
    org_id,
    project_id,
    metric_id,
    60 AS granularity,
    tags.key,
    tags.value,
    toStartOfInterval(timestamp, toIntervalSecond(60)) AS timestamp,
    retention_days,
    uniqCombined64State(arrayJoin(set_values)) AS value
FROM default.metrics_buckets_local
WHERE materialization_version = 0
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days │
```

metrics_sets_mv_local (the default ones are 60 seconds aggregations)

snuba/cli/devserver.py
fpacifici: After step 1 we would be loosing data as the current matview would not be materializing anything and you need to reset the offset to go back. It also would require downtime.
What about this:
1. make a change to this so that the version is taken from settings (not runtime). Also make the aggregated processor rely on the same setting in the opposite way. If the setting says write to matview, write the current version and drop the messages from aggregated processor 
2. change the multistorage config to start the new processors (the aggregated ones). These would do nothing per the setting above.
3. deploy a change that changes the setting. Setting are loaded at startup. So atomically the raw data stop getting materialized (different version) but keeps getting written in the raw. storage and the aggregated storage start writing. 

Rollback strategy: switch the setting back. and deploy.


> remove all data from aggregate tables where the interval start is after the start time of the first record[s] written with materialization_version = 1

This is impossible to achieve consistently as clickhosue keeps merging buckets. Once data is written in the aggregated storage is permanent and cannot be deleted in a consistent way.

snuba/datasets/metrics_aggregate_processor.py
onewland: got it, I'll add that in

snuba/cli/devserver.py
onewland: In that case we're assuming data corruption is not fixable? I can imagine that is OK for preproduction but just want to be sure we're on the same page since theoretically you could recalculate the aggregates from the bucket data.

snuba/datasets/metrics_aggregate_processor.py
onewland: added, let me know if I did that right

snuba/cli/devserver.py
nikhars: > In that case are we assuming that data corruption is not fixable? I can imagine that is OK for preproduction but just want to be sure we're on the same page since theoretically you could recalculate the aggregates from the bucket data.

I would try very hard to avoid data corruption. When you said theoretically you could build by recalculating the aggregates, were you referring to dropping the whole materialized view and creating it again from scratch? Even if you did that you would have to stop the Kafka consumer till you delete/create the materialized view on all nodes of the cluster. Creating it from scratch would take a lot of time I belive.

fpacifici: Expanding on data corruption. What kind of corruption are you talking about ?
- data is written in a way that clickhouse cannot read. This is a disaster recovery scenario and generally the outcome of a clickhouse bug otherwise it would not allow you to write, depending what happened you may have to restore from backup. You cannot just restore from raw data as the raw data will always have lower retention that aggregated data. If you cannot read parts because they are corrupted on disk, clickhouse does not allow you to access the table entirely. More likely it will exclude broken parts.
- We wrote data we should not have written by mistake. In this case you cannot really redress the result consistently (but it is the same kind of issues in any aggregated store). Once clickhouse merged a bucket (imagine a daily bucket) that contains good and bad data you cannot separate them unless you throw away the entire bucket. 
- We did not write stuff we should have written. In this case you could materialize data from the raw table.

All these scenarios are disaster recovery scenarios and would happen with the database not accessible for the whole operation.

So the goal of being able to revert is basically something we would do in pre-production and just in case keep around in case we start seeing errors later. Imagine INSERTs failing for some reasons. 
The reason for doing it with no downtime is because it makes it a lot easier operationally but also in terms of implementation as you do not have to build anything to deal with offset based synchronization which we do not have and simply. .... the procedure is simpler: switch a setting.

onewland: > * We wrote data we should not have written by mistake. In this case you cannot really redress the result consistently (but it is the same kind of issues in any aggregated store). Once clickhouse merged a bucket (imagine a daily bucket) that contains good and bad data you cannot separate them unless you throw away the entire bucket.

I was thinking of this situation, and specifically a subset case where we wrote the correct data to the bucket table, but we're misunderstanding something about the aggregate merge tree, so only the aggregate table has some kind of bad data. In such a situation, I can imagine in theory using the bucket table to recreate *only* bad data in the aggregate table, but I also believe you when you say that this is not feasible in practice.

I am convinced of using a setting to control which way we write to the aggregate tables -- I might ask, though, why not have it be a runtime config so that we can turn it on/off ourselves without involving ops?

fpacifici: runtime config, being runtime, is not going to be synchronized between processors. 
If each processor (buckets and aggregated) independently load the runtime config, for the same message one could read 0 and the second may read 1 if the value changed in between. So you may end up double writing a batch if both processors think they have to write or not writing it at all if none thinks they need to write.
settings are read at startup, the entire consumer has the same value and it is only re-read upon restart.

onewland: Sounds good, will add this to settings and follow your proposed approach

onewland: I've implemented the changes you suggested.

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,182 @@
...
+)
+from snuba.query.expressions import (
+    Expression,
+    FunctionCall,
+    Literal,
+    OptionalScalarType,
+)
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+snuba_metrics = MetricsWrapper(environment.metrics, "metrics.processor")
```
MeredithAnya, line 21: could we name this `snuba_metrics.processor` just to be explicit?

onewland: I *believe* it already gets prefixed with snuba, but my confidence in that is low:

https://github.com/getsentry/snuba/blob/bab9bc59f334fa8a7a19355dc14791492a9b885d/snuba/environment.py#L43-L46

fpacifici: it does already get prefixed with `snuba`

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,182 @@
...
+            {
+                "org_id": _literal(message["org_id"]),
+                "project_id": _literal(message["project_id"]),
+                "metric_id": _literal(message["metric_id"]),
+                "timestamp": _call(
+                    "toStartOfInterval",
+                    (
+                        _call("toDateTime", (_literal(timestamp.isoformat()),),),
+                        _call("toIntervalSecond", (_literal(granularity),)),
+                    ),
```
fpacifici, line 85: Why don't you just round the date in the python code and just pass `toDateTime(THE_DATE_YOU_ROUNDED)`? 
That would save all the AST processing to create the expression and serialize it and make the whole thing more compact.

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,182 @@
...
+                ),
+                "tags.key": _array_literal(keys),
+                "tags.value": _array_literal(values),
+                **self._process_values(message),
+                "retention_days": _literal(message["retention_days"]),
+                "partition": metadata.partition,
+                "offset": metadata.offset,
```
fpacifici, line 92: These are not supposed to exist in the aggregate table. I guess it still work because you filter only the columns that actually exist.

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -0,0 +1,182 @@
...
+            settings.WRITE_METRICS_AGG_DIRECTLY
+            and message["type"] is not None
+            and message["type"] == METRICS_SET_TYPE
+        )
+
+    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+        values = message["value"]
+        for v in values:
+            assert isinstance(v, int), "Illegal value in set. Int expected: {v}"
+        snuba_metrics.gauge("set.size", len(values))
```
fpacifici, line 112: A gauge is not the metric you want to use.
A gauge is used to measure a single value that grows and shrinks. Example the concurrent queries on Clickhouse.
Here you want a counter. Each message logs the number of values so that we can later extract avg and quantiles, or just the aggregated sum of all messages to know how many values we process per minute/second etc.

`snuba_metrics.increment("set.size", len(values))`

fpacifici (approved):
Looks good to me. Please fix the metrics collection.

snuba/datasets/metrics_aggregate_processor.py
onewland: I'm a bit hesitant to put in logic like https://stackoverflow.com/a/10854034 if we're trying to match the exact same behavior as `toStartOfInterval`. https://clickhouse.com/docs/en/sql-reference/functions/date-time-functions/#tostartofintervaltime-or-data-interval-x-unit-time-zone is sort of unclear on details around interval boundaries. 

I think the key question is: are we going to receive queries looking for the window at a specific timestamp vs. querying within a range? I think these questions of "do you draw the day at the logical boundary vs. round down to `timestamp=(ms since epoch/86400 seconds day)`?" don't matter if we're expecting clients to query for ranges of timestamps, but if they query for `timestamp = {value}` then I think this isn't valuable enough to work on before a sanity test for the overall approach.

fpacifici: > are we going to receive queries looking for the window at a specific timestamp vs. querying within a range

How we query does not matter. Since we are storing time buckets with multiple granularities in the same table, in order  for clickhouse to merge all the data points that represent a specific time bucket, all data points (rows we insert) need to be aligned with the beginning of the time bucket.

Let's say you insert three rows (granularity, one hour), one comes in with timestamp: 12:01, one 12:02, one 12:03, In order for clickhouse to merge all those three in one time bucket, they all must be written with timestamp = 12:00, otherwise they will not be merged in the same one hour bucket.

So since all rows in the table are aligned to the beginning of the time bucket they represent according to their granularity, there is only one way to query data: which is with time condition aligned with the beginning of the time bucket.

MeredithAnya: > are we going to receive queries looking for the window at a specific timestamp vs. querying within a range? 

does this mean the difference for searching last 15min from a given timestamp versus inputing a range of two timestamps?

onewland: Thanks for feedback, I think I talked myself into this being a bit more complicated than it is. Will update with in-process time rounding

onewland:
This has been updated to round the time in python

PR #2362, opened 2022-02-01
perf(subscriptions): Add metrics to SubscriptionScheduler.find()

lynnagara:
This function seems to be taking a really long time relative to other parts
of the scheduler. This change adds a bunch of timings to this function to see
if we can isolate the issue further.

Also cleans up a couple metrics we were previously recording but don't need anymore.


evanh (approved):
We also have the Timer class to help with this style of timing.

PR #2366, opened 2022-02-02
feat(tracing): Use the readonly user for tracing queries

lynnagara:


evanh (commented):
Wasn't the readonly user only just added to the query nodes? Don't you need to use that user somewhere?

evanh (approved):

PR #2368, opened 2022-02-02
output a metric for how long clickhouse subscription queries take

vladkluev:
One of the reasons the subscription consumer's lag can increase is if clickhouse queries take longer. Currently we don't know if they do. Add a metric for that

### DIFF
--- a/snuba/subscriptions/worker.py
+++ b/snuba/subscriptions/worker.py
@@ -88,10 +88,12 @@ class SubscriptionWorker(
     def __execute_query(
         self, request: Request, timer: Timer, task: ScheduledSubscriptionTask
     ) -> Tuple[Request, Result]:
+        partition = task.task.subscription.identifier.partition
         with self.__concurrent_gauge:
             is_consistent_query = request.settings.get_consistent()
 
             def run_consistent() -> Result:
+                start = time.time()
                 request_copy = Request(
                     id=request.id,
                     body=copy.deepcopy(request.body),
@@ -101,7 +103,7 @@ class SubscriptionWorker(
                     ),
                 )
 
-                return parse_and_run_query(
+                res = parse_and_run_query(
                     self.__dataset,
                     request_copy,
                     timer,
@@ -110,8 +112,15 @@ class SubscriptionWorker(
                     if is_consistent_query
                     else None,
                 ).result
+                self.__metrics.timing(
+                    "clickhouse_query.latency",
+                    (time.time() - start) * 1000,
+                    tags={"partition": str(partition), "consistent": "1"},
+                )
+                return res
 
             def run_non_consistent() -> Result:
+                start = time.time()
                 request_copy = Request(
                     id=request.id,
                     body=copy.deepcopy(request.body),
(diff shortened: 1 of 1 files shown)


snuba/subscriptions/worker.py
```diff
@@ -130,6 +139,12 @@ def run_non_consistent() -> Result:
                     if not is_consistent_query
                     else None,
                 ).result
+                self.__metrics.timing(
+                    "clickhouse_query.latency",
+                    (time.time() - start) * 1000,
+                    tags={"partition": str(partition), "consistent": "0"},
+                )
+                return res
```
fpacifici, line 147: I think you are going to have a problem with this metric and it will not be representative. 
These queries are executed concurrently in a ThreadPool. Python being Python they are concurrent, but only the clickhouse part is in parallel, the processing phase is not. So logging the process end to end will give you meaningless results as the time between `start` and this line will include the processing of other queries that the thred scheduler decided to run in the meantime.

Anyway, isn't thi metric equivalent to `snuba.api.query` that already has the breakdown by phase and includes `snuba.query.execute` to see only the clickhouse part ?

fpacifici: Also if you are looking to tie the query response time for queries related to a partition with the latency on that partition, I don't think it will work. The consumer consumer multiple partitions together, then creates batches that contains schedules from all partitions it consumes. If queries for a specific partition where slower than usual, that would impact the latency of all partitions as the whole schedule would take longer thus slowing down the consumption of all partitions assigned to the consumer.

vladkluev: Interesting, does every partition run on every consumer or are they deterministic per host? Because you would see specific partitions going up but not others right?

fpacifici: kafka decides which partitions run where. 
Each partition is assigned to one and only one consumer at a time.
Each consumer run a number of partitions that kafka decides.
In the current schema we have 4 partitions per consumer replica. If you increase the replicas you are going to have fewer if you decrease that you will have more. The number of partitions, and which partitions run where depends on what kafka decides during rebalancing. 
Generally contiguous partitions run on the same host: 0-3 on replica 1, 4-7 on replica 2 etc.

vladkluev: > So logging the process end to end will give you meaningless results as the time between start and this line will include the processing of other queries that the thread scheduler decided to run in the meantime.

I understand that but why would that be useless information? The scheduler is part of the system, why would we not time it? Are you saying there would be too much noise in the metric?

fpacifici: What I am saying is that what you are adding here is already being logged as the `snuba.api.query` which is also broken down by phase. 
The scheduling phase is not included in what you are logging here. It happens before. 
Still, both we have and what you would be adding is not very useful with the execption of the clickhouse execution time which is `snuba.api.query.execute`

fpacifici: What would be useful, would be to break down `snuba.api.query` by either dataset or storage, so that we can break down the subscriptions query execution time by dataset since they all share the same referrer.
I think those metrics are logged by the timer instantiated here https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/worker.py#L81

PR #2369, opened 2022-02-03
Log dataset with query.execute

vladkluev:
This PR adds the ability to add default tags to timers and uses that to specify:

* A partition (for subscriptions)
* An entity (for subscriptions)
* the query dataset

All default tags will be overridden when calling `send_metrics_to`


snuba/querylog/__init__.py
```diff
@@ -24,6 +24,7 @@ def _record_timer_metrics(
             "referrer": referrer,
             "parent_api": request.settings.get_parent_api(),
             "final": final,
+            "dataset": query_metadata.dataset,
```
fpacifici, line 27: You need to add this to the `mark_tags` as well, otherwise you will have the dataset tags only in the top level metric `snuba.api.query` and not in the marks `snuba.api.query.execute`.

fpacifici (approved):
Please fix the mark tags in querylog.py before merging

PR #2370, opened 2022-02-03
ref(admin) Disable the tracing tool

evanh:
Without the readonly user the tracing tool is a big risk. Disable it until we
have the correct users/connections in place.


fpacifici (approved):

evanh:
Disabling this is temporary until the readonly user PR (https://github.com/getsentry/snuba/pull/2366) is merged and deployed.

fpacifici:
> What are the actual risks that are there and are we sure those risks outweigh the cost of not having this too
The readonly user enforces a small quota, so you cannot eat into the default user quota on clickhouse. The default user does not. This tool does not have a rate limiter, which means you can quickly issue enough concurrent heavy queries to saturate small clusters and impact the user experience for real users.

fpacifici:
@evanh  you can close this now.

PR #2371, opened 2022-02-03
feat(admin) - tracing tool readability improvement

onewland:
Turns the tracing tool output into a hierarchy where you can expand/collapse nodes.

https://user-images.githubusercontent.com/130124/152393838-4374e38b-3e3a-47f4-9e10-1fe63de0cccd.mov


snuba/admin/static/tracing/index.tsx
```diff
@@ -1,8 +1,115 @@
-import React from "react";
+import React, { useState } from "react";
 import Client from "../api_client";
 import QueryDisplay from "../components/query_display/query_display";
 import { QueryResult } from "../components/query_display/types";
 
+type LogLine = {
```
evanh, line 6: Could you put all of this logic in a different file inside the tracing/ directory? This file will only get more complex as we change the UI so I'd like to keep things separate if possible.

snuba/admin/static/tracing/index.tsx
```diff
@@ -20,17 +127,23 @@ function TracingQueries(props: { api: Client }) {
...
-              {title}
-              <textarea
-                spellCheck={false}
-                value={value}
-                style={{ width: "100%", height: height }}
-                disabled
-              />
-            </div>
-          );
+          console.log(title);
```
evanh, line 130: 💀

snuba/admin/static/tracing/index.tsx
```diff
@@ -1,8 +1,115 @@
...
+  title?: string;
+  logsBucketed: BucketedLogs;
+}) {
+  const [visible, setVisible] = useState<boolean>(false);
+
+  const nodeKey = props.host + "-" + props.category;
+  return (
+    <li key={nodeKey}>
+      <span onClick={() => setVisible(!visible)}>
+        {props.title} - Click to {visible ? "collapse" : "expand"}
```
evanh, line 93: Could we use a symbol to show this state instead of "Click to X"?

evanh (approved):
This is really awesome. I just have a couple housekeeping comments.

snuba/admin/static/tracing/index.tsx
```diff
@@ -1,8 +1,115 @@
...
+      host: host,
+      pid: pid!,
+      query_id: local_query_id,
+      log_level: log_level!,
+      component: component!,
+      message: message!,
+    };
+    return logLineParsed;
+  } else {
+    return null;
```
lynnagara, line 40: Which are the lines that can't be formatted this way - is it common? I wonder if we should display the output instead of removing them in case they contain something useful.

onewland: I think the trace output contains one empty line at the end

snuba/admin/static/tracing/index.tsx
onewland: Looks like Lyn is creating a new component on another branch -- maybe for now this could be [+] or [-] depending on current visibility state?

lynnagara: I like [+] and [-]. Collapse component was more for history, we could try it out but sure if it would look right here.

evanh: +1 for [+]/[-]

PR #2374, opened 2022-02-03
ref(snuba-admin): Add precanned system queries to UI

MeredithAnya:
**context**
I've added the ability to select from a pre-made list of queries that can be used. Selecting the drop down shows you the formatted query, and allows you to **Use Query** in the text box below. It basically just copies the query and passes it as a prop. 

You'll still need to choose the storage and host in order to actually execute the query, but this allows you to edit the query before executing so we can make base queries that are easy to build from. 

Screen Shot 2022-02-03 at 11 03 17 AM

**adding new queries**
I pulled out the pre-canned queries into their own file - `predefined_system_queries.py` - so it should be easy to add new ones as we go. I know this won't scale forever,  but for now I think this is a really good way to quickly switch between different types of queries.

### DIFF
--- a/snuba/datasets/storages/metrics.py
+++ b/snuba/datasets/storages/metrics.py
@@ -165,8 +165,8 @@ sets_storage = WritableTableStorage(
     storage_key=StorageKey.METRICS_SETS,
     storage_set_key=StorageSetKey.METRICS,
     schema=WritableTableSchema(
-        local_table_name="metrics_sets_local",
-        dist_table_name="metrics_sets_dist",
+        local_table_name="metrics_sets_v2_local",
+        dist_table_name="metrics_sets_v2_dist",
         storage_set_key=StorageSetKey.METRICS,
         columns=ColumnSet(
             [


snuba/admin/clickhouse/system_queries.py
```diff
@@ -124,7 +55,7 @@ def run_system_query_on_host_by_name(
         (FROM|from)
         \s
         system.[a-z_]+
-        (?P<extra>\s[\w\s,=()*+<>'%\-\/]+)?
+        (?P<extra>\s[\w\s,=()*+<>'"%\-\/]+)?
```
MeredithAnya, line 58: needed this to be able to have queries that include things like `partition = '(90,"2022-01-24")' `

snuba/admin/static/components/query_display/query_display.tsx
```diff
@@ -63,7 +70,7 @@ function QueryDisplay(props: {
       })
       .catch((err) => {
         console.log("ERROR", err);
-        window.alert("An error occurred: " + err.error.message);
+        window.alert("An error occurred: " + err.error);
```
MeredithAnya, line 73: `err.error.message` is undefined, we just need the `err.error`

snuba/admin/clickhouse/predefined_system_queries.py
```diff
@@ -0,0 +1,120 @@
+from dataclasses import dataclass
+from typing import Dict, Optional, Sequence, Type
+
+
+class _QueryRegistry:
```
lynnagara, line 5: Is this a copy and paste over from the system_queries file or has it changed in some way? I think we should take the opportunity to clean it up - I believe some of the functionality here may deprecated now.

snuba/admin/static/clickhouse_queries/index.tsx
```diff
@@ -1,20 +1,137 @@
...
+                style={{
+                  fontSize: 14,
+                  marginBottom: 5,
+                  width: "50vw",
+                }}
+              >
+                <p
+                  dangerouslySetInnerHTML={{
+                    __html: predefinedQuery.description || "",
+                  }}
```
lynnagara, line 102: Hmm, this seems a bit odd. Was there a reason we need to resort to this way of setting the content of the element?

snuba/admin/static/clickhouse_queries/index.tsx
```diff
@@ -1,20 +1,137 @@
...
+                ></p>
+              </div>
+            }
+            <textarea
+              readOnly={true}
+              value={predefinedQuery?.sql}
+              spellCheck={false}
+              style={{ width: "50vw", height: 100 }}
+            />
```
lynnagara, line 111: How come we are adding 2 text boxes with the query and extra step to copy it over to the main input? Why not just populate the main input area to begin with?

lynnagara (commented):
Would it be possible to spilt up the API and UI changes into separate PRs? There's quite a lot going on here.

snuba/admin/clickhouse/predefined_system_queries.py
```diff
@@ -0,0 +1,120 @@
...
+    <strong><code>partition = '(90,"2022-01-24")'</code></strong> with the partition you want.
+
+    To get partitions you can use <strong><code>SELECT partition FROM system.parts</code></strong>
+    """
+
+    sql = """
+    SELECT name, formatReadableSize(bytes_on_disk) as size
+    FROM system.parts
+    WHERE
+        partition = '(90,"2022-01-24")'  AND
```
nikhars, line 116: For queries wherein you expect the user to pass in a parameter, I feel we should have a mechanism to clearly identify where a parameter is needed. We could use something similar to how redash does it https://redash.io/help/user-guide/querying/query-parameters.

```suggestion
        partition = {{ partition }} AND
```

onewland: I'll oppose this slightly (or maybe add a caveat). I'd like to see a valid value here, maybe it could be a [comment](https://clickhouse.com/docs/en/sql-reference/syntax/#comments)?

`partition = {{partition}} AND /* ex. '(90,"2022-01-24")' */`

MeredithAnya: yeah the reason I put it in there was so that it was explicit in what the syntax should be, since that was not obvious to me. That's why in the description I mentioned that you should replace it. Worst case someone runs the query and gets zero results.

snuba/admin/static/clickhouse_queries/index.tsx
MeredithAnya: So I think it's nice to be able to click between the precanned queries without them messing up the current query you have. Im thinking of these precanned queries as basically as reference of what you could do. Like you may not know just by the name if it's the _exact_ query you want. I like being able to see what the sql is before I decide I want to use it. 

open to thoughts here if it ends up being annoying for folks. much easier to take away later than to add back in.

snuba/admin/static/clickhouse_queries/index.tsx
MeredithAnya: I just wanted to be able to format the description with some html on the backend and just have it render here. This is def not ideal tho. Might be nice to render markdown here, and use something like https://www.npmjs.com/package/markdown-to-jsx, but i dk how strict we are being with adding extra dependencies

snuba/admin/clickhouse/predefined_system_queries.py
```diff
@@ -0,0 +1,120 @@
...
+    def all_queries(cls) -> Sequence[Type["SystemQuery"]]:
+        return _QUERY_REGISTRY.all_queries
+
+
+class CurrentMerges(SystemQuery):
+    """Currently executing merges"""
+
+    sql = """
+    SELECT
+        count(),
```
fpacifici, line 57: `SELECT table, elapsed, progress, result_part_name`
This will be more useful. The number of merges without knowing what clickhosue is merging is not useful.

snuba/admin/clickhouse/predefined_system_queries.py
```diff
@@ -0,0 +1,120 @@
...
+        is_currently_executing
+    FROM system.replication_queue
```
fpacifici, line 59: `FROM system.merges` not replication queue, if you are looking for merges, replication_queue also has standard replication events.

snuba/admin/clickhouse/predefined_system_queries.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+
+class ActivePartitions(SystemQuery):
+    """Currently active parts"""
+
+    sql = """
+    SELECT
+        active,
+        count()
```
fpacifici, line 70: I think this query will be misleading. I would not add it. 
For all practically purpose number of parts and size of parts are meaningful only across partition (except for counting inactive parts, but there as well, you want to know how they are distributed). 

This is because clickhouse does not merge across partitions, so the total number of inactive or active parts means nothing, and only set limits per partition. So the total number of parts is not enforced in any way.

snuba/admin/clickhouse/predefined_system_queries.py
```diff
@@ -0,0 +1,120 @@
...
+    GROUP BY active
+    """
+
+
+class PartsPerTable(SystemQuery):
+    """Number of parts grouped by table. Parts should not be in the high thousands."""
+
+    sql = """
+    SELECT
+        count(),
```
fpacifici, line 81: Same as above, break down by partition.

snuba/admin/static/clickhouse_queries/index.tsx
evanh: I agree that it is nice to keep the two separate. That way I can do something like copy the query down into the edit box, and keep the precanned version visible so I know how to populate the parameterized values.

I think the Clear button is weird though, since it's up with the precanned queries but erases what is in the editing textbox. I'm not sure that button is useful at all.

lynnagara: I think the additional text box makes the whole system queries page (both when you are using canned queries and without) more visually complicated.

Some thoughts:
- There are now two input boxes on the page - imo the wrong one is emphasised at the top.
- It looks like you can and should edit the predefined queries text box but you actually can't.
- There's a lot more buttons and clicks involved when you are using a predefined query as you need to send it through 2 text boxes to run it instead of just selecting it and hitting Execute which seems much more straightforward to me.

Ultimately I think the double textbox design may be optimizing for what could be a pretty niche use case (wanting to see multiple queries together both predefined and self entered) at the detriment of the regular experience where you just want to type a query or select one predefined query only. I am not opposed to this design if many people feel that they want to use the tool with this workflow, but I don't see myself doing that.

i think that keeping UIs simple is pretty hard even though it looks easy, and I disagree with the notion that we should add just in case because taking away is easier than adding things.

MeredithAnya: @evanh the clear shouldn't change whats in the edited text box, should just clear the top part so its not in the way if you've moved on from using the precanned queries. so if it's doing that it wasn't intended

evanh: Ah OK. I'll have to test again to see if that's what was happening. I would maybe change that from Clear -> Hide or maybe Collapse so it's a little more clear what's happening. Obviously a very small nit though.

snuba/admin/static/clickhouse_queries/index.tsx
lynnagara: I think we avoid sending HTML from the server to the client. Why can't we just send the SQL string?

snuba/admin/static/clickhouse_queries/index.tsx
MeredithAnya: > Ultimately I think the double textbox design may be optimizing for what could be a pretty niche use case (wanting to see multiple queries together both predefined and self entered) at the detriment of the regular experience where you just want to type a query or select one predefined query only. I am not opposed to this design if many people feel that they want to use the tool with this workflow, but I don't see myself doing that.


@lynnagara I think this is a good point - I def built this out in a way that I felt was useful for me. Maybe if I can clarify the goals I was trying to solve by doing it this way, folks may have other ideas on how to achieve that. 

1. Since we won't be able to have every possible query we'd want, my guess was that we'd have a lot of canned queries that people would want to edit. I wanted to make it possible to quickly see the original query without it overriding what ever you had edited, incase you had removed something you wanted back in. It's possible you could get this from the result history, but I find that very hard to use in its current state.
2. We talked about having this page have more info on what options you have to even query by in clickhouse without going to the system.tables documentation. I felt the easiest way to do this was to have a bunch of canned queries that utilized those options (different columns from different tables) to use as a reference. Again, if you are in the middle of building a query, I think I would find it annoying showing a pre-canned query overrode what i was currently working on. 

I guess the overall goal I was trying to do was to make this useful as a reference while building your own queries and not just a plug in and execute functionality.

snuba/admin/static/clickhouse_queries/index.tsx
MeredithAnya: we do send the sql string, this is just for the description of the query

lynnagara: This is because one of the descriptions has bold text? I think we can live without that.

Today, yes it is true that the possibility of say dodgy links or scripts being placed here may be small since we don't have a lot of canned queries and all of them have gone through Snuba code review process. But it's also very possible we might miss something in code review. And if allow users to edit and define these dynamically in the future it will open a can of worms.

I would prefer not to open the door to the possibility of HTML injection even if it seems unlikely today.

snuba/admin/clickhouse/predefined_system_queries.py
MeredithAnya: addressed in https://github.com/getsentry/snuba/pull/2382

PR #2375, opened 2022-02-03
feat(admin) Split tracing UI into separate component

evanh:
Most of what this does is split the tracing UI into its own components so it
isn't sharing UI components with the system query UI any more. Now we can
iterate on the tracing tool without breaking system queries by accident.

There is also a check that disables the execute button when a query is running,
to prevent someone queuing up a bunch of queries by mashing the button before
the previous query had finished.


lynnagara (approved):
LGTM

PR #2378, opened 2022-02-04
fix(admin) Change to readonly user type 2

evanh:
This will allow the user to change settings. See here: https://clickhouse.com/docs/en/operations/settings/permissions-for-queries/#settings_readonly

I suspect this will require an ops PR as well, but I figured it wouldn't hurt to try here just in case.


vladkluev (approved):

lynnagara (changes requested):
Please create separate ADMIN clickhouse query setting and we use that expanded permission only for admin. Let's leave QUERY unchanged as we use that everywhere else.

evanh:
> Please create separate ADMIN clickhouse query setting and we use that expanded permission only for admin. Let's leave QUERY unchanged as we use that everywhere else.

Is that a problem? The other parts of the code aren't changing settings (or running arbitrary user queries).

lynnagara (approved):

PR #2381, opened 2022-02-05
fix(admin): Fix page scrolling on system queries

lynnagara:
Previously query results with many columns or long text inside those columns would
cause the entire page to get very wide and scroll weirdly. This change clamps the
main page to 100% width and only scrolls the inner table element. Later we will probably
want to go a step further and scroll each result separately after moving to the collapsible
history layout in https://github.com/getsentry/snuba/pull/2380


evanh (approved):

lynnagara:
Now that https://github.com/getsentry/snuba/pull/2380 is merged and we no longer have nested tables this became a bit simpler - updated it to just scroll each table independently of the rest

PR #2383, opened 2022-02-07
fix: Remove impossible query warning

evanh:
This warning isn't providing any actionable value. The metrics will stay in
place, and as the data model in the SDK is rolled out these numbers should
drop down.


fpacifici (commented):
Let's keep a datadog metric, just not to be totally blind.

evanh:
> Let's keep a datadog metric, just not to be totally blind.

There already are, I left those.

PR #2386, opened 2022-02-07
fix(admin): Prevent form post

lynnagara:
Sometimes when you run a system query it just clears the page instead
of properly displaying the results. This is due to the default browser
action associated with posting the form and redirecting to the next page.
Since this is a single page app and we don't want to go anywhere just
suppress the default action with event.preventDefault().


onewland (approved):
Looks good though I think we could also just get rid of the `form` container since we don't actually use the form functionality for anything

lynnagara:
Yeah, I thought about that, it seemed slightly more semantically correct to keep the form

PR #2389, opened 2022-02-08
docs(internal): Add some doc comments for MetricsBackend

onewland:
Tried adding some clarification around some misunderstandings I had in another PR

### DIFF
--- a/snuba/utils/metrics/backends/abstract.py
+++ b/snuba/utils/metrics/backends/abstract.py
@@ -13,16 +13,26 @@ class MetricsBackend(ABC):
     def increment(
         self, name: str, value: Union[int, float] = 1, tags: Optional[Tags] = None
     ) -> None:
+        """
+        Emit a single instance of a non-timing metric that can be counted,
+        sliced/diced into percentiles, or summed
+        """
         raise NotImplementedError
 
     @abstractmethod
     def gauge(
         self, name: str, value: Union[int, float], tags: Optional[Tags] = None
     ) -> None:
+        """
+        Emit a metric that is the authoritative value at a point in time
+        """
         raise NotImplementedError
 
     @abstractmethod
     def timing(
         self, name: str, value: Union[int, float], tags: Optional[Tags] = None
     ) -> None:
+        """
+        Emit a metric for the performance of an operation
+        """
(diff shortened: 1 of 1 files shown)


snuba/utils/metrics/backends/abstract.py
```diff
@@ -13,16 +13,26 @@ class MetricsBackend(ABC):
     def increment(
         self, name: str, value: Union[int, float] = 1, tags: Optional[Tags] = None
     ) -> None:
+        """
+        Emit a single instance of a non-timing metric that can be counted,
+        sliced/diced into percentiles, or summed
+        """
```
lynnagara, line 19: This is a real nitpick but I think it's clearer as:
Increments a count metric by the `value` provided. (or something along those lines)

TIL we don't expose `decrement` function in here but otherwise your definition would be the same.

vladkluev (approved):
Thank you, that's very helpful

snuba/utils/metrics/backends/abstract.py
onewland: Yeah, I guess I find the wording around metrics to be clumsy and not super-standardized though it's possible I'm not familiar with the standards.

If I'm in Datadog I'm often much more interested to view the p99 of a metric emitted like this vs. the sum so I don't really think of it as a counter (even if that's their term for it).

lynnagara: I see, I believe this is often referred to as a "counter" metrics (at least in statsd). I think it may helpful to align the wording to that.

onewland: I wonder if it would be helpful to add examples here. I'm fine with using stats/Datadog terminology but I'm thinking that coming in as a new person I didn't understand what you meant.

lynnagara: Fair enough. Examples never hurt, up to you whether to add

onewland:
@lynnagara Let me know what you think of these updates

evanh (approved):

lynnagara (approved):
Thanks!

vladkluev (approved):

PR #2390, opened 2022-02-08
feat(dev): Mock data for subscriptions

lynnagara:
I needed to build this as part of load testing the scheduler
in https://github.com/getsentry/snuba/pull/2387 and although most
of those other changes will be thrown away, I think this part may be
worth keeping. I wanted an easy way for everyone to be able to get
mock subscription data into Snuba for dev and load testing and right
now it's a bit of a hassle.

### DIFF
--- a/bin/mocks/mock-subscriptions
+++ b/bin/mocks/mock-subscriptions
@@ -3,21 +3,18 @@ import argparse
 import random
 from datetime import timedelta
 
-from snuba.datasets.entities import EntityKey
-from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.entities.factory import ENTITY_NAME_LOOKUP
 from snuba.datasets.factory import get_dataset
-from snuba.datasets.table_storage import KafkaTopicSpec
 from snuba.redis import redis_client
 from snuba.subscriptions.data import PartitionId, SnQLSubscriptionData
-from snuba.subscriptions.entity_subscription import EventsSubscription
+from snuba.subscriptions.entity_subscription import ENTITY_KEY_TO_SUBSCRIPTION_MAPPER
 from snuba.subscriptions.store import RedisSubscriptionDataStore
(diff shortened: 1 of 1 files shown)


bin/mocks/mock-subscriptions
```diff
@@ -0,0 +1,65 @@
...
+timer = Timer("mock-subscriptions")
+
+for _ in range(parsed.number):
+    project_id = random.choice(parsed.project)
+
+    subscription_data = SnQLSubscriptionData(
+        query=f"MATCH (events) SELECT count() AS count WHERE project_id = {project_id}",
+        project_id=project_id,
+        time_window=timedelta(minutes=1),
+        resolution=timedelta(minutes=1),
```
onewland, line 47: any reason not to parameterize these?

bin/mocks/mock-subscriptions
```diff
@@ -0,0 +1,65 @@
...
+    nargs="+",
+)
+
+parsed = parser.parse_args()
+
+dataset_name = "events"
+dataset = get_dataset(dataset_name)
+entity_key = EntityKey.EVENTS
+entity = get_entity(EntityKey.EVENTS)
+event_topic_spec = KafkaTopicSpec(Topic.EVENTS)
```
onewland, line 36: could it be useful to parameterize all this?

onewland (approved):
If you say this is useful, I believe you.

bin/mocks/mock-subscriptions
lynnagara: I'll refactor to this as input but will only support events for now as there is a bit of additional logic required to support the others

bin/mocks/mock-subscriptions
lynnagara: It could be, or it could generate randomly different windows but I will not add it yet as it's not something we need right now

PR #2391, opened 2022-02-08
ref(attribution) Add test hook for metrics

evanh:
There are some changes we make that are simply adding/changing metrics. There is
now a hook in DummyMetrics that will record all the calls made to it, so tests
can assert that the metrics they expect were called.

This means also removing the default settings for DOGSTATSD. Since they were
pointing to a non-existent service this should have no functional effect beyond
forcing local development to use DummyMetrics by default.


snuba/utils/metrics/backends/dummy.py
```diff
@@ -1,9 +1,37 @@
-from collections import Mapping
-from typing import Optional, Union
+from __future__ import annotations
+
+from typing import List, Mapping, MutableMapping, Optional, Tuple, Union
 
 from snuba.utils.metrics.backends.abstract import MetricsBackend
 from snuba.utils.metrics.types import Tags
 
+RecordedMetricCalls = List[Tuple[Union[int, float], Tags]]
```
vladkluev, line 8: I think this API could be made more user friendly by making this a dataclass.

```python3
@dataclass
class RecordedMetric:
    value: int | float
    tags: Optional[Tags]
```

That way this line:

```
[c for c in metric_calls if c[1]["team"] == "sns"]
```

becomes:

```
[c for c in metric_calls if c.tags["team"] == "sns"]
```

which is much more readable

snuba/utils/metrics/backends/dummy.py
```diff
@@ -1,9 +1,37 @@
...
+
+    if name not in RECORDED_METRIC_CALLS[mtype]:
+        RECORDED_METRIC_CALLS[mtype][name] = []
+
+    if tags is None:
+        tags = {}
+    RECORDED_METRIC_CALLS[mtype][name].append((value, tags))
+
+
+def get_recorded_metric_calls(mtype: str, name: str) -> RecordedMetricCalls | None:
```
vladkluev, line 29: Having this be one global variable across all tests will make it so that different tests pollute each others' results. I have two different solutions to this problem:

1. Clear the global dictionary in the conftest.py before every test
2. Instead of keeping track of this state ourselves we:

1. Make the DummyMetricsBackend log each metric under a its own logger 
```python3
metrics_logger = logging.get_logger("dummy.metrics")
```
2. Log each metric as a dictionary:
```python3
metrics_logger.info({"name": ...})
```
3. use the caplog fixture in pytest to get the logs specific to a test

evanh: I decided to reset it for now. If this starts to become a pervasive testing strategy we can simplify it.

vladkluev (approved):

PR #2394, opened 2022-02-09
fix(cleanup) Fix rounding error in cleanup script

evanh:
Since the cleanup script wasn't rounding to midnight, it was possible to have a
partition that was still in the retention period, but counted as being out
the script wasn't rounding it's values to midnight.

For example, the script runs at 5 minutes past midnight. It finds a partition
with a last part that is only 2 minutes past midnight. Before, since the
partition date is less than the script date, the partition would be deleted. Now
everything is rounded to midnight so the script will keep the partition.


onewland:
The logic seems right to me. Could this be tested in your local environment with some fake data?

snuba/cleanup.py
```diff
@@ -47,21 +47,25 @@ def get_active_partitions(
...
+    """Returns the current day. Separate function for mocking purposes."""
+    return datetime.utcnow()
+
+
+def filter_stale_partitions(parts: Sequence[util.Part]) -> Sequence[util.Part]:
+    """
+    Filter partitions down to ones that are out of the retention window based on the current time."""
+
+    def midnight(a_date: datetime) -> datetime:
+        return datetime(a_date.year, a_date.month, a_date.day)
```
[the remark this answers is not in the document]
evanh: It's only used by filter_stale_partitions. I don't have a strong opinion about where to put it though.

evanh:
> Could this be tested in your local environment with some fake data?

I don't think there's a better way to reproduce this specific edge case beyond what the test is doing.

snuba/cleanup.py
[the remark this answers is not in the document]
onewland: I'd rather it be top-level, maybe as `_midnight`. This just seems less readable to me. But not a merge-blocking request.

onewland (approved):

snuba/cleanup.py
```diff
@@ -47,21 +47,25 @@ def get_active_partitions(
...
 
 
-def filter_stale_partitions(
-    parts: Sequence[util.Part], as_of: Optional[datetime] = None
-) -> Sequence[util.Part]:
-    """Filter partitions down to ones that are out of the retention
-    window based on `as_of` (default: now)."""
+def current_time() -> datetime:
+    """Returns the current day. Separate function for mocking purposes."""
+    return datetime.utcnow()
```
lynnagara, line 52: Just my $0.02 but I think the way it was written earlier (passing this date as an optional argument to `filter_stale_partitions` so it can be tested more straightforward than making this a separate function then needing to do Mock in tests. Any particular reason for the change?

tests/test_cleanup.py
```diff
@@ -69,15 +70,29 @@ class TestCleanup:
...
+        prebase = datetime.utcnow()
+        base = datetime(
+            1999,
+            12,
+            26,
+            prebase.hour,
+            prebase.minute,
+            prebase.second,
+            prebase.microsecond,
+        )  # a sunday
```
lynnagara, line 94: You could just update the hardcoded base time to include some hours, minutes, seconds. It's less verbose than using combination of hardcoded and actual time and merging them together

tests/test_cleanup.py
```diff
@@ -176,3 +191,42 @@ def to_monday(d: datetime) -> datetime:
...
+        )
+
+        table = storage.get_table_writer().get_schema().get_local_table_name()
+
+        database = storage.get_cluster().get_database()
+
+        parts = cleanup.get_active_partitions(clickhouse, storage, database, table)
+        assert parts == []
+
+        timestamp = datetime(2021, 10, 25)
```
lynnagara, line 224: What's the scenario that's failing exactly currently? Is it the one where this is a Sunday?

snuba/cleanup.py
evanh: I have an opposing opinion. I don't like the pattern of having parameters in a function that exist purely for tests. `as_of` was just a confusing extra parameter if you didn't know it was for test mocking. I don't think main code should be written to accomodate tests if it can be avoided.

tests/test_cleanup.py
lynnagara: Please add the comment. It may also be worth defining the time the script runs as `base = timestamp + timedelta(days=90, minutes=4)` if that is your intention as it's hard to tell the relationship between the 2 dates.

lynnagara (approved):

PR #2396, opened 2022-02-10
feat(profiling): Create the stacktraces table to store profiling data

phacops:
This PR aims to create the table necessary to store profiling data in ClickHouse.

I also added 2 new type of columns to be able to use a `Tuple` type in ClickHouse. This helps to consider the version as one value with 2 components which can't be separated.

### DIFF
--- a/snuba/clickhouse/columns.py
+++ b/snuba/clickhouse/columns.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Sequence, Tuple, Union
+from typing import Sequence, Tuple as PTuple, Union
 
 from snuba.utils.schemas import UUID, AggregateFunction, Any, Array, Column
 from snuba.utils.schemas import ColumnSet as BaseColumnSet
@@ -14,11 +14,13 @@ from snuba.utils.schemas import (
     Float,
     IPv4,
     IPv6,
+    NamedTuple,
     Nested,
     Nullable,
     ReadOnly,
     SchemaModifiers,
     String,
+    Tuple,
     TModifiers,
     TypeModifier,
     TypeModifiers,
@@ -41,12 +43,14 @@ __all__ = (
     "Float",
     "IPv4",
     "IPv6",
+    "NamedTuple",
     "Nested",
     "Nullable",
     "ReadOnly",
     "SchemaModifiers",
     "String",
     "TModifiers",
+    "Tuple",
     "TypeModifier",
     "TypeModifiers",
     "UInt",
@@ -69,7 +73,7 @@ class ColumnSet(BaseColumnSet):
     def __init__(
         self,
         columns: Sequence[
-            Union[Column[SchemaModifiers], Tuple[str, ColumnType[SchemaModifiers]]]
+            Union[Column[SchemaModifiers], PTuple[str, ColumnType[SchemaModifiers]]]
         ],
     ) -> None:
         for column in columns:
@@ -82,7 +86,7 @@ class ColumnSet(BaseColumnSet):
 
     def __add__(
         self,
-        other: Union[ColumnSet, Sequence[Tuple[str, ColumnType[SchemaModifiers]]]],
+        other: Union[ColumnSet, Sequence[PTuple[str, ColumnType[SchemaModifiers]]]],
     ) -> ColumnSet:
         if isinstance(other, ColumnSet):
             return ColumnSet([*self.columns, *other.columns])
--- a/snuba/clickhouse/columns.py
+++ b/snuba/clickhouse/columns.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Sequence, Tuple as PTuple, Union
+from typing import Sequence, Union
 
 from snuba.utils.schemas import UUID, AggregateFunction, Any, Array, Column
 from snuba.utils.schemas import ColumnSet as BaseColumnSet
@@ -73,7 +73,7 @@ class ColumnSet(BaseColumnSet):
     def __init__(
         self,
         columns: Sequence[
-            Union[Column[SchemaModifiers], PTuple[str, ColumnType[SchemaModifiers]]]
+            Union[Column[SchemaModifiers], tuple[str, ColumnType[SchemaModifiers]]]
         ],
     ) -> None:
         for column in columns:
@@ -86,7 +86,7 @@ class ColumnSet(BaseColumnSet):
 
     def __add__(
         self,
-        other: Union[ColumnSet, Sequence[PTuple[str, ColumnType[SchemaModifiers]]]],
+        other: Union[ColumnSet, Sequence[tuple[str, ColumnType[SchemaModifiers]]]],
     ) -> ColumnSet:
         if isinstance(other, ColumnSet):
             return ColumnSet([*self.columns, *other.columns])
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -17,7 +17,7 @@ class MigrationGroup(Enum):
     SESSIONS = "sessions"
     QUERYLOG = "querylog"
     SPANS_EXPERIMENTAL = "spans_experimental"
-    STACKTRACES = "stacktraces"
+    PROFILES = "profiles"
 
 
 # Migration groups are mandatory by default, unless they are on this list
@@ -26,7 +26,7 @@ OPTIONAL_GROUPS = {
     MigrationGroup.SESSIONS,
     MigrationGroup.QUERYLOG,
     MigrationGroup.SPANS_EXPERIMENTAL,
-    MigrationGroup.STACKTRACES,
+    MigrationGroup.PROFILES,
 }
 
 
@@ -203,12 +203,13 @@ class SpansExperimentalLoader(DirectoryLoader):
     def get_migrations(self) -> Sequence[str]:
         return ["0001_spans_experimental"]
 
-class StacktracesLoader(DirectoryLoader):
+
+class ProfilesLoader(DirectoryLoader):
     def __init__(self) -> None:
-        super().__init__("snuba.migrations.snuba_migrations.stacktraces")
+        super().__init__("snuba.migrations.snuba_migrations.profiles")
 
     def get_migrations(self) -> Sequence[str]:
-        return ["0001_stacktraces"]
+        return ["0001_profiles"]
 
 
 _REGISTERED_GROUPS = {
@@ -221,7 +222,7 @@ _REGISTERED_GROUPS = {
     MigrationGroup.SESSIONS: SessionsLoader(),
     MigrationGroup.QUERYLOG: QuerylogLoader(),
     MigrationGroup.SPANS_EXPERIMENTAL: SpansExperimentalLoader(),
-    MigrationGroup.STACKTRACES: StacktracesLoader(),
+    MigrationGroup.PROFILES: ProfilesLoader(),
 }
 
 
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -17,6 +17,7 @@ class MigrationGroup(Enum):
     SESSIONS = "sessions"
     QUERYLOG = "querylog"
     SPANS_EXPERIMENTAL = "spans_experimental"
+    STACKTRACES = "stacktraces"
 
 
 # Migration groups are mandatory by default, unless they are on this list
@@ -25,6 +26,7 @@ OPTIONAL_GROUPS = {
     MigrationGroup.SESSIONS,
     MigrationGroup.QUERYLOG,
     MigrationGroup.SPANS_EXPERIMENTAL,
+    MigrationGroup.STACKTRACES,
 }
 
 
@@ -201,6 +203,13 @@ class SpansExperimentalLoader(DirectoryLoader):
     def get_migrations(self) -> Sequence[str]:
         return ["0001_spans_experimental"]
 
+class StacktracesLoader(DirectoryLoader):
+    def __init__(self) -> None:
+        super().__init__("snuba.migrations.snuba_migrations.stacktraces")
+
+    def get_migrations(self) -> Sequence[str]:
+        return ["0001_stacktraces"]
+
 
 _REGISTERED_GROUPS = {
     MigrationGroup.SYSTEM: SystemLoader(),
@@ -212,6 +221,7 @@ _REGISTERED_GROUPS = {
     MigrationGroup.SESSIONS: SessionsLoader(),
     MigrationGroup.QUERYLOG: QuerylogLoader(),
     MigrationGroup.SPANS_EXPERIMENTAL: SpansExperimentalLoader(),
+    MigrationGroup.STACKTRACES: StacktracesLoader(),
 }
 
 
--- a/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
+++ b/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
@@ -45,11 +45,11 @@ class Migration(migration.ClickhouseNodeMigration):
     def forwards_local(self) -> Sequence[operations.SqlOperation]:
         return [
             operations.CreateTable(
-                storage_set=StorageSetKey.STACKTRACES,
-                table_name="stacktraces_local",
+                storage_set=StorageSetKey.PROFILES,
+                table_name="profiles_local",
                 columns=columns,
                 engine=table_engines.ReplacingMergeTree(
-                    storage_set=StorageSetKey.STACKTRACES,
+                    storage_set=StorageSetKey.PROFILES,
                     order_by="(organization_id, project_id, transaction_id)",
                     ttl="received + toIntervalDay(retention_days)",
                     settings={"index_granularity": "8192"},
@@ -62,18 +62,18 @@ class Migration(migration.ClickhouseNodeMigration):
     def backwards_local(self) -> Sequence[operations.SqlOperation]:
         return [
             operations.DropTable(
-                storage_set=StorageSetKey.STACKTRACES, table_name="stacktraces_local",
+                storage_set=StorageSetKey.PROFILES, table_name="profiles_local",
             )
         ]
 
     def forwards_dist(self) -> Sequence[operations.SqlOperation]:
         return [
             operations.CreateTable(
-                storage_set=StorageSetKey.STACKTRACES,
-                table_name="stacktraces_dist",
+                storage_set=StorageSetKey.PROFILES,
+                table_name="profiles_dist",
                 columns=columns,
                 engine=table_engines.Distributed(
-                    local_table_name="stacktraces_local",
+                    local_table_name="profiles_local",
                     sharding_key="cityHash64(transaction_id)",
                 ),
             )
@@ -82,6 +82,6 @@ class Migration(migration.ClickhouseNodeMigration):
     def backwards_dist(self) -> Sequence[operations.SqlOperation]:
         return [
             operations.DropTable(
-                storage_set=StorageSetKey.STACKTRACES, table_name="stacktraces_dist",
+                storage_set=StorageSetKey.PROFILES, table_name="profiles_dist",
             )
         ]
--- a/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
+++ b/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
@@ -49,7 +49,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 table_name="profiles_local",
                 columns=columns,
                 engine=table_engines.ReplacingMergeTree(
-                    order_by="(organization_id, project_id, toStartOfDay(received), transaction_id)",
+                    order_by="(organization_id, project_id, toStartOfDay(received), cityHash64(transaction_id))",
                     partition_by="(retention_days, toMonday(received))",
                     sample_by="cityHash64(transaction_id)",
                     settings={"index_granularity": "8192"},
--- a/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
+++ b/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
@@ -12,7 +12,7 @@ columns: List[Column[Modifiers]] = [
     Column("transaction_id", UUID()),
     Column("received", DateTime()),
     # profiling data
-    Column("stacktrace", String(Modifiers(codecs=["LZ4HC(9)"]))),
+    Column("profile", String(Modifiers(codecs=["LZ4HC(9)"]))),
     Column("symbols", String(Modifiers(codecs=["LZ4HC(9)"]))),
     # filtering data
     Column("android_api_level", UInt(32, Modifiers(nullable=True))),
--- a/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
+++ b/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
@@ -33,7 +33,6 @@ columns: List[Column[Modifiers]] = [
     Column("version", NamedTuple((("name", String()), ("code", String())))),
     # internal data
     Column("retention_days", UInt(16)),
-    Column("deleted", UInt(8)),
     Column("partition", UInt(16)),
     Column("offset", UInt(64)),
 ]
@@ -55,7 +54,6 @@ class Migration(migration.ClickhouseNodeMigration):
                     settings={"index_granularity": "8192"},
                     storage_set=StorageSetKey.PROFILES,
                     ttl="received + toIntervalDay(retention_days)",
-                    version_column="deleted",
                 ),
             )
         ]
--- a/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
+++ b/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
@@ -49,12 +49,13 @@ class Migration(migration.ClickhouseNodeMigration):
                 table_name="profiles_local",
                 columns=columns,
                 engine=table_engines.ReplacingMergeTree(
-                    storage_set=StorageSetKey.PROFILES,
-                    order_by="(organization_id, project_id, transaction_id)",
-                    ttl="received + toIntervalDay(retention_days)",
+                    order_by="(organization_id, project_id, toStartOfDay(received), transaction_id)",
+                    partition_by="(retention_days, toMonday(received))",
+                    sample_by="cityHash64(transaction_id)",
                     settings={"index_granularity": "8192"},
+                    storage_set=StorageSetKey.PROFILES,
+                    ttl="received + toIntervalDay(retention_days)",
                     version_column="deleted",
-                    partition_by="(organization_id, project_id, toMonday(received))",
                 ),
             )
         ]
--- a/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
+++ b/snuba/migrations/snuba_migrations/profiles/0001_profiles.py
@@ -35,7 +35,7 @@ columns: List[Column[Modifiers]] = [
     Column("retention_days", UInt(16)),
     Column("deleted", UInt(8)),
     Column("partition", UInt(16)),
-    Column("offset", UInt(16)),
+    Column("offset", UInt(64)),
 ]
 
 
--- a/snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
+++ b/snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
@@ -1,0 +1,95 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import (
+    UUID,
+    Column,
+    DateTime,
+    String,
+    UInt,
+    Tuple,
+    NamedTuple,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+columns: List[Column[Modifiers]] = [
+    # primary key
+    Column("project_id", UInt(64)),
+    Column("transaction_id", UUID()),
+
+    # profiling data
+    Column("stacktrace", String(Modifiers(codecs=["LZ4HC(9)"]))),
+    Column("symbols", String(Modifiers(codecs=["LZ4HC(9)"]))),
+
+    # filtering data
+    Column("trace_id", UUID()),
+    Column("transaction_name", String(Modifiers(low_cardinality=True))),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+
+    Column("version", NamedTuple((("name", String), ("code", String)), Modifiers(low_cardinality=True))),
+    Column("platform", String(Modifiers(low_cardinality=True))),
+    Column("android_api_level", UInt(32, Modifiers(nullable=True))),
+    Column("device_classification", String(Modifiers(low_cardinality=True))),
+    Column("device_locale", String(Modifiers(low_cardinality=True))),
+    Column("device_manufacturer", String(Modifiers(low_cardinality=True))),
+    Column("device_model", String(Modifiers(low_cardinality=True))),
+    Column("device_os_build_number", String(Modifiers(low_cardinality=True))),
+    Column("device_os_name", String(Modifiers(low_cardinality=True))),
+    Column("device_os_version", String(Modifiers(low_cardinality=True))),
+
+    Column("error_code", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("error_description", String(Modifiers(low_cardinality=True, nullable=True))),
+
+    Column("duration_ns", UInt(64)),
+    Column("ingested_at_ts", DateTime()),
+
+    # internal data
+    Column("retention_days", UInt(16, Modifiers(default=30))),
+]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    storage_set=StorageSetKey.STACKTRACES,
+                    order_by="(project_id, transaction_id, ingested_at_ts)",
+                    ttl="finish_ts + toIntervalDay(retention_days)",
+                    settings={"index_granularity": "8192"},
+                ),
+            )
+        ]
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.DropTable(
+                storage_set=StorageSetKey.STACKTRACES, table_name="stacktraces_local",
+            )
+        ]
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_dist",
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name="stacktraces_local",
+                    sharding_key="cityHash64(project_id, transaction_name)",
+                ),
+            )
+        ]
+
+    def backwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.DropTable(
+                storage_set=StorageSetKey.STACKTRACES, table_name="stacktraces_dist",
+            )
+        ]
--- a/snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
+++ b/snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
@@ -48,7 +48,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 columns=columns,
                 engine=table_engines.ReplacingMergeTree(
                     storage_set=StorageSetKey.STACKTRACES,
-                    order_by="(organiation_id, project_id, transaction_id, received)",
+                    order_by="(organization_id, project_id, transaction_id, received)",
                     ttl="received + toIntervalDay(retention_days)",
                     settings={"index_granularity": "8192"},
                     version_column="version",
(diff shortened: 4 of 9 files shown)


snuba/clickhouse/columns.py
```diff
@@ -69,7 +73,7 @@ class ColumnSet(BaseColumnSet):
     def __init__(
         self,
         columns: Sequence[
-            Union[Column[SchemaModifiers], Tuple[str, ColumnType[SchemaModifiers]]]
+            Union[Column[SchemaModifiers], PTuple[str, ColumnType[SchemaModifiers]]]
```
evanh, line 76: You actually don't need to import `Tuple` from typing if you have the `future` import in the file. You can type things with the builtin Python type. This is true for list, dict etc. https://www.python.org/dev/peps/pep-0585/
There are a couple edge cases where that isn't true but not here.

```suggestion
            Union[Column[SchemaModifiers], tuple[str, ColumnType[SchemaModifiers]]]
```

phacops: Removed all aliases in favor of the built-in type.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,95 @@
...
+    Column("device_manufacturer", String(Modifiers(low_cardinality=True))),
+    Column("device_model", String(Modifiers(low_cardinality=True))),
+    Column("device_os_build_number", String(Modifiers(low_cardinality=True))),
+    Column("device_os_name", String(Modifiers(low_cardinality=True))),
+    Column("device_os_version", String(Modifiers(low_cardinality=True))),
+
+    Column("error_code", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("error_description", String(Modifiers(low_cardinality=True, nullable=True))),
+
+    Column("duration_ns", UInt(64)),
```
lynnagara, line 44: Does this mean nanoseconds? Is that useful? I believe duration in other tables like transactions is in milliseconds.

lynnagara:
Just a heads up, the new migration/migration group will not automatically get run unless it is added to https://github.com/getsentry/snuba/blob/master/snuba/migrations/groups.py. We also have this OPTIONAL_GROUPS list that you can add it to if you want to be able to run it in dev but not yet roll out to users https://github.com/getsentry/snuba/blob/master/snuba/migrations/groups.py#L23-L28

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,95 @@
...
+    # profiling data
+    Column("stacktrace", String(Modifiers(codecs=["LZ4HC(9)"]))),
+    Column("symbols", String(Modifiers(codecs=["LZ4HC(9)"]))),
+
+    # filtering data
+    Column("trace_id", UUID()),
+    Column("transaction_name", String(Modifiers(low_cardinality=True))),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+
+    Column("version", NamedTuple((("name", String()), ("code", String())), Modifiers(low_cardinality=True))),
```
lynnagara, line 30:
```suggestion
    Column("version", NamedTuple((("name", String()), ("code", String(Modifiers(low_cardinality=True)))))),
```

I think there may be an issue with string nested inside tuple, the SQL generated doesn't look quite right

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: It does mean nanoseconds and it's a carry-over from the Specto SDKs where we would send nanoseconds. I'll check with my team but we can likely change that to milliseconds if you insist.

phacops:
So, I don't know if we need to treat that differently, it's meant to go to a different ClickHouse cluster when deployed in Sentry. Is there a way I can specify/configure that in the `MigrationGroup`?

phacops:
Alright so I made the `MigrationGroup` optional until we decide otherwise and added a new cluster configuration that will default to the same cluster since I assume we won't run multiple in some environments (like for tests).

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
lynnagara: Definitely not insisting, it doesn't really make a ton of difference here.. apart from maybe being a UInt32 vs UInt64. Asking the question more from the product perspective really - do we want to present different granularities for each type of product in the UI or would we want to keep them uniform?

snuba/settings/__init__.py
```diff
@@ -49,6 +55,18 @@
...
+        "port": int(os.environ.get("CLICKHOUSE_PROFILING_PORT", CLICKHOUSE_PORT)),
+        "user": os.environ.get("CLICKHOUSE_PROFILING_USER", CLICKHOUSE_USER),
+        "password": os.environ.get("CLICKHOUSE_PROFILING_PASSWORD", CLICKHOUSE_PASSWORD),
+        "database": os.environ.get("CLICKHOUSE_PROFILING_DATABASE", CLICKHOUSE_DATABASE),
+        "http_port": int(os.environ.get("CLICKHOUSE_PROFILING_HTTP_PORT", CLICKHOUSE_HTTP_PORT)),
+        "storage_sets": {
+            "stacktraces",
+        },
+        "single_node": True,
+    },
```
lynnagara, line 69: Do you mind just keeping these all on the same cluster instead of adding a new one? This default configuration is used in self hosted and is intentionally as simple as possible with all the storage sets assigned to one cluster.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_local",
+                columns=columns,
+                engine=table_engines.MergeTree(
+                    storage_set=StorageSetKey.STACKTRACES,
+                    order_by="(organiation_id, project_id, transaction_id, received)",
```
lynnagara, line 63:
```suggestion
                    order_by="(organization_id, project_id, transaction_id, received)",
```

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,95 @@
...
+                    ttl="finish_ts + toIntervalDay(retention_days)",
```
lynnagara, line 64: There is no finish_ts column

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,95 @@
...
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_local",
+                columns=columns,
+                engine=table_engines.MergeTree(
```
lynnagara, line 61: IIRC you had this as replacing merge tree earlier and changed it to merge tree. Just curious about the reason behind that. It means we would not be able to do any updates or deletions nor deduplicate potentially duplicate inserts.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: We'd be converting to milliseconds for sure before displaying that, I think we kept that to work with integers only until display. I think we'd like to keep that and we need to store it as a `UInt64` otherwise it would represent only max of 4s.

snuba/settings/__init__.py
phacops: For sure, I'll remove that.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: I'm still wondering if I should keep that field. You seem to only work with project IDs anyway so this might be redundant. Thoughts?

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: Ah right, this should be `received`.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: I don't think we need the `ReplacingMergeTree` features as we ingest only 1 profile per transaction and everything is sorted with the transaction ID in the sorting key. But maybe it still be a good idea in case we ingest twice the same data (which wouldn't be needed, just a "bug" or a client wrongly sending twice the same transaction).

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
lynnagara: I don't really have any strong opinion about this. @evanh might have thoughts on how this impacts things on the query side though.

evanh: I think you should keep `organization_id`. We do work with organization IDs in different parts of the product. In fact the only datasets that don't use organization IDs are errors and transactions.

I don't think there is any downside to having organization ID here, besides using slightly more storage. There should never be a case in the product where you have a project ID with no organization ID, and there are times when you want to query across an entire organization (for example if the user has every project selected).

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
lynnagara: I think ReplacingMergeTree might be safer even if are not going to support merging and deletions per se, for the reasons you mentioned. We generally do try to deduplicate events in ingest but some still make their way to ClickHouse. Also for compliance (gdpr?) we may want a way to remove data on customer request prior to the ttl. If the TTL is pretty short anyway it may not be a big deal as data will get automatically removed before long, but might be good to have the option to delete still.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
lynnagara: 👍

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_dist",
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name="stacktraces_local",
+                    sharding_key="cityHash64(project_id, transaction_name)",
```
lynnagara, line 82: I think it may be good to pick a sharding key that distributes the data a bit more evenly between the shards (Not sure what transaction_name represents exactly but it does seem to be a a low cardinality thing), e.g. `cityHash64(transaction_id)` or something like that. @fpacifici @evanh Any opinions here?

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
fpacifici: A couple of remarks around the the order by key. 
This serves two purposes: primary key (for deduplication and merge when you will turn this into a ReplacingMergeTree), and sorting key for queries. 
So first you need to ensure that you are adding enough fields that allow you to deduplicate rows otherwise (on a replacing merge tree) clickhouse will replace those that have the same order by key.

Second it is a matter of search efficiency.
- organization and project first are good. You always have them in your query. Generally we do not need to have org id in snuba unless you need it for some reason (querying across an organization without specifying a project. DO you have such use case?)
- Then you probably want the time range. If queries based on time ranges are a thing. From your scenario they are. So you can put `toStartOfDay(timestamp or whatever time column)`. This narrows down the dataset dramatically when doing aggregation queries.
- Then you want the trace_id/transaction_id. Which one do you plan to search by ? 

Whether we should have timestamp first or transaction_id first depends on what do you want to optimize between search by trace or search by time range.
- If you put transaction_id first, Clickhouse will be extremely efficient at finding one trace. Querying by any other field within a time range will be a nightmare. This is because you will have to scan literally the entire table as timestamps would not be sorted. Clickhouse will not be able to exclude any block based on timestamp since the table is sorted by transaciton.
- If you put timestamp first, you will have much more efficient queries by time range. Clickhosue will be able to narrow down the portion of the table to scan very quickly. (so short time range queries will be more efficient).  The search by transaciton_id will be somewhat less effective as you will have to find the trace _id on every day potentially (you are sorting by toStartOfDay(timestamp). That is not a major issue as the sections of the table to search are relatively few (90 days). This can be made better if we knew somehow a shorter time range in which to look for a a transaction.

fpacifici: If the queries that extract a single transactions are very prominent, I would suggest to store the traces separately (like in nodestore) as we do for events/transacitons to have an effective search by transaction id.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_local",
+                columns=columns,
+                engine=table_engines.MergeTree(
+                    storage_set=StorageSetKey.STACKTRACES,
+                    order_by="(organiation_id, project_id, transaction_id, received)",
+                    ttl="received + toIntervalDay(retention_days)",
+                    settings={"index_granularity": "8192"},
```
fpacifici, line 62: You are missing the partition key (if you expect 90 days, definitively partition by week,  if you expect 30 days or less we can consider whether partitioning by week or by day).

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
fpacifici: Clickhouse is quite efficient at partitioning evenly up at least 10 shards (likely more). If you partition evenly as Lyn suggested you get a few perks for free:
- even distribution when you add shards
- no hot shards, queries on a massive project will be distributed across shards. if you put  a project and a transaction_name in the same shard, one host alone will have to perform the entire scan.

Now this will depend a lot on your query pattern. 
- if you query by project_id,transaction id, and shard by project_id, transaction_name, you do not take advantage of shard locality at all. Clickhouse needs to fan out the query to all shards anyway
- if you query by project_id and any field other than transaction_name you have the same problem. Clickhouse cannot narrow down one shard so it runs the query on all of them.

So it seems that you do not gain a lot by sharding by anything that is not spreading data evenly across shards. 
If you where sharding by project_id and transaciton_id then yes you would be able to narrow down the query to one shard knowing in advance where to find your data. Though you will not gain anything in the queries that are likely to be heavier in terms of load as they have to scan more data: aggregation by any other field.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+                ),
+            )
+        ]
```
fpacifici, line 65: Let's talk about data skipping indices:
MergeTree tables can have data skipping indices https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/#table_engine-mergetree-data_skipping-indexes

To make them useful you need to anticipate which fields are more likely to end in a filter. Let's say that your query has a time range and a number of other fields that are not in the order by key, this means clickhouse will scan all the involved columns within the time range.
If you have fields that have specific features, the index can help: `trace_id` is likely to be fairly unique, which makes it very high selectivity. A bloomfilter index will help you there as it will exclude most blocks before scanning. if you never query by trace_id then don't do that ,it would be a waste of space
Potentially a version field could use a set index as over time versions may be clustered. the most recent version may be more common.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
fpacifici: Have you also considered adding the platform (ios/android) in the order by ? Is there any chance you will know whether you are looking for an android trace or an ios trace before querying ? That would exclude all the rows from the other platform.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+columns: List[Column[Modifiers]] = [
+    # primary key
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("transaction_id", UUID()),
+    Column("received", DateTime()),
+
+    # profiling data
+    Column("stacktrace", String(Modifiers(codecs=["LZ4HC(9)"]))),
+    Column("symbols", String(Modifiers(codecs=["LZ4HC(9)"]))),
```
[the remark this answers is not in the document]
phacops: It's a blob to store processed symbols for iOS and methods for Android. We can choose another name if you prefer.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+columns: List[Column[Modifiers]] = [
+    # primary key
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("transaction_id", UUID()),
```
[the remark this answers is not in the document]
phacops: Yes.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+    Column("received", DateTime()),
+
+    # profiling data
+    Column("stacktrace", String(Modifiers(codecs=["LZ4HC(9)"]))),
```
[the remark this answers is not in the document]
phacops: These 2 columns will store a JSON string and will represent the biggest impact on storage. We compress once and LZ4 is very efficient at decompressing so I'd thought the trade-off with CPU time was acceptable.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+    Column("symbols", String(Modifiers(codecs=["LZ4HC(9)"]))),
+
+    # filtering data
+    Column("android_api_level", UInt(32, Modifiers(nullable=True))),
+    Column("device_classification", String(Modifiers(low_cardinality=True))),
```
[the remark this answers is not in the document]
phacops: Yes, and it actually can be an `Enum` since we classify devices in 3 values (`high`/`mid`/`low`).

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+    DateTime,
+    String,
+    UInt,
+    NamedTuple,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+columns: List[Column[Modifiers]] = [
```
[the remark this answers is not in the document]
phacops: `elapsed_relative_to_start_date_ns`, `thread_id` and `queue_name` are going to be in the JSON object stored in the `stacktrace` column. We chose not to have several rows for each entry (as we were doing previously) since it's simpler to integrate into the Sentry SDKs that way.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,92 @@
...
+    # primary key
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("transaction_id", UUID()),
+    Column("received", DateTime()),
```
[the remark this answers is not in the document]
phacops: Yes. For more context, `start_time` indicated when the trace measurement started, `received` is more about when the measurement was done or even when we received this trace.

All that matters is we have a timestamp to let us order by time and aggregate more recent stack traces instead of old ones.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: I guess my idea was to group data having the same transaction name as we'll aggregate by that a lot but I didn't really give enough thought to it. It sounds like a bad idea.

Given what you said, using the `transaction_id` will provide the most uniform distribution. Let's use that then.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: Would it be useful to partition by `organization_id`, `project_id` as well as by week? We won't do queries which are not scoped by these columns.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
phacops: We'll already have the primary key to select efficiently via `transaction_id`. Do we have to add this index now? I'd rather wait until we need to aggregate to think about that.

snuba/utils/schemas.py
```diff
@@ -558,3 +557,69 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+        super().__init__(modifiers)
+        self.types = types
+
+    def _repr_content(self) -> str:
+        return ", ".join("{}".format(repr(v)) for v in self.types)
+
+    def __eq__(self, other: object) -> bool:
+        return (
+            self.__class__ == other.__class__
+            and self.get_modifiers() == cast(Enum[TModifiers], other).get_modifiers()
```
lynnagara, line 577:
```suggestion
            and self.get_modifiers() == cast(Tuple[TModifiers], other).get_modifiers()
```

snuba/utils/schemas.py
```diff
@@ -558,3 +557,69 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+            for column in self.types
+        ]
+
+    def _repr_content(self) -> str:
+        return repr(self.types)
+
+    def __eq__(self, other: object) -> bool:
+        return (
+            self.__class__ == other.__class__
+            and self.get_modifiers() == cast(Enum[TModifiers], other).get_modifiers()
```
lynnagara, line 612:
```suggestion
            and self.get_modifiers() == cast(NamedTuple[TModifiers], other).get_modifiers()
```

snuba/utils/schemas.py
```diff
@@ -558,3 +557,69 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+            and self.types == cast(Tuple[TModifiers], other).types
+        )
+
+    def _for_schema_impl(self) -> str:
+        return "Tuple({})".format(", ".join([repr(t) for t in self.types]))
```
lynnagara, line 582: The repr is not necessarily the same as the schema representation

```suggestion
        return "Tuple({})".format(", ".join([t.for_schema() for t in self.types]))
```

snuba/utils/schemas.py
```diff
@@ -558,3 +557,69 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+        modifiers: Optional[TModifiers] = None,
+    ) -> None:
+        super().__init__(modifiers)
+        self.types = tuple(Column(*t) for t in types)
+
+    def flatten(self, name: str) -> Sequence[FlattenedColumn]:
+        return [
+            FlattenedColumn(name, column.name, Array(column.type))
+            for column in self.types
+        ]
```
lynnagara, line 604: I don't think you need this, why changing from the default implementation?

lynnagara (commented):
Forgot to mention this earlier, a couple of comments on the new column types

snuba/utils/schemas.py
phacops: Not quite sure, I thought it made sense given how it was done in `Nested` but I can't exactly remember why so it's probably not useful. Will remove.

phacops:
I just want to remind everybody of the 2 use cases we will have:
- select a stack trace by ID
- aggregate hundreds of stack traces together

Also, we will be self-hosted for our beta (own ClickHouse cluster, snuba workers and all) and this schema is likely to evolve before it reaches GA. In addition, we will onboard one customer with low traffic to test this out in the next month but not more. We're not going to receive thousands of stack traces per second just yet.

Our immediate need is to provide the UI with a list of stack traces and then a way to select a stack trace by ID. The very next thing will be to provide a way to query hundreds of those stack traces at once to aggregate them in our service (not in the database). The longer term is to denormalize and moved all the aggregation in the database but we have not figured out all the details for that just yet.

I think storing those stack traces in the Bigtable to satisfy the "select by ID" use case will be inefficient as soon as we're going to aggregate (fetching hundreds of stack traces after fetching IDs in ClickHouse then turning to the nodestore for data will be slow) and we would benefit from having the data when we filter in ClickHouse. That's why I think a good starting point for us is to store everything in ClickHouse and then denormalize from there.

Now, I think the primary/sorting key should be optimized for "selecting by ID" and we should have a secondary index to filter on version/transaction_name/received since they will be our default filters for the list. I'm not sure trying to combine them into the sorting key will be very efficient and that's why 2 indexes seem like the right approach and that's why I made the sorting key this way. Is there a better way to handle both query patterns with 1 index?

I took into account most of the comments and this comment should explain why I didn't take into account the rest. Of course, I probably missed a comment or two that still stands despite those explanations so let me know about that.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
```diff
@@ -0,0 +1,87 @@
...
+                storage_set=StorageSetKey.STACKTRACES,
+                table_name="stacktraces_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    storage_set=StorageSetKey.STACKTRACES,
+                    order_by="(organization_id, project_id, transaction_id)",
+                    ttl="received + toIntervalDay(retention_days)",
+                    settings={"index_granularity": "8192"},
+                    version_column="deleted",
+                    partition_by="(organization_id, project_id, toMonday(received))",
```
lynnagara, line 57: I don't think every project should have its own partition. Let's make this `(retention_days, toMonday(timestamp))` so it will be easy to figure out what they are out of retention and we can safely drop the data.

lynnagara:
Probably not a bad idea to add a sample_by key in case we will be doing sampling on queries for the aggregations use case.

lynnagara:
About the strategy to put transaction_id in the primary key and add a data skipping index for `received`, it sounds reasonable to me, though I'd be interested in Filippo's thoughts here. In other tables we did actually combine them into the primary key with something like `(org_id, project_id, toStartOfDay(received), transaction_id)` and that has seemed to work pretty well for both the sorting by timestamp as well as finding a single row by ID use cases.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
fpacifici: please keep the transaciton_id, not rand. We may be able to query only one shard if we know which transaction_id we look for and we partition by transaction_id.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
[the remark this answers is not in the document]
fpacifici: if it can be an enum, please keep using lowcardinality. The storage underneath performs in the same way for all practical purposes and you do not haveto hardcode values in the schema.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
fpacifici: no please do not partition by org/project. That can generate hundreds of thousands of partitions. Which will generate a stupidly high amount of files on disk (each one will be a directory containing another directory per part). 
The order by key is good enough to restrict the data you are searching when you filter by project id.

snuba/migrations/snuba_migrations/stacktraces/0001_stacktraces.py
fpacifici: yeah we can add them when you want.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,88 @@
...
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("error_code", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("error_description", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("platform", String(Modifiers(low_cardinality=True))),
+    Column("trace_id", UUID()),
+    Column("transaction_name", String(Modifiers(low_cardinality=True))),
+    Column("version", NamedTuple((("name", String()), ("code", String())))),
+    # internal data
+    Column("retention_days", UInt(16)),
+    Column("deleted", UInt(8)),
```
nikhars, line 36: Are the entries which are written to Clickhouse mutable? If not, then you can get rid of the deleted column.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,88 @@
...
+    Column("partition", UInt(16)),
+    Column("offset", UInt(16)),
```
nikhars, line 38: This column should be UInt64

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,88 @@
...
+    Column("device_locale", String(Modifiers(low_cardinality=True))),
+    Column("device_manufacturer", String(Modifiers(low_cardinality=True))),
+    Column("device_model", String(Modifiers(low_cardinality=True))),
+    Column("device_os_build_number", String(Modifiers(low_cardinality=True))),
+    Column("device_os_name", String(Modifiers(low_cardinality=True))),
+    Column("device_os_version", String(Modifiers(low_cardinality=True))),
+    Column("duration_ns", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("error_code", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("error_description", String(Modifiers(low_cardinality=True, nullable=True))),
```
nikhars, line 29: The columns which have `low_cardinality` set to True, are these user input strings or machine generated strings? Just wanted to check. You might not want to have columns which are user input strings

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,88 @@
...
+                table_name="profiles_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(organization_id, project_id, toStartOfDay(received), cityHash64(transaction_id))",
+                    partition_by="(retention_days, toMonday(received))",
+                    sample_by="cityHash64(transaction_id)",
+                    settings={"index_granularity": "8192"},
+                    storage_set=StorageSetKey.PROFILES,
+                    ttl="received + toIntervalDay(retention_days)",
+                    version_column="deleted",
```
nikhars, line 58: If you are going to remove `deleted` column then remove this as well.

snuba/migrations/groups.py
```diff
@@ -17,6 +17,7 @@ class MigrationGroup(Enum):
     SESSIONS = "sessions"
     QUERYLOG = "querylog"
     SPANS_EXPERIMENTAL = "spans_experimental"
+    PROFILES = "profiles"
```
nikhars, line 20: Please add this migration to the skipped migration group for now https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L157. Since we are still experimenting we do not want the migration to be run for Self hosted or On-Prem installations

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,88 @@
...
+
+columns: List[Column[Modifiers]] = [
+    # primary key
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("transaction_id", UUID()),
+    Column("received", DateTime()),
+    # profiling data
+    Column("profile", String(Modifiers(codecs=["LZ4HC(9)"]))),
+    Column("symbols", String(Modifiers(codecs=["LZ4HC(9)"]))),
```
nikhars, line 16: Is it safe to assume that these are non-searchable fields and would only be used for the use case where a user wants to do a lookup by a specific profile?

phacops: They are non-searchable fields. They would be used when a user wants to lookup a specific profile but also when we would do an aggregation (for example, trying to profile a specific transaction on all iPhone 11 in the last week). We'd load those fields into our aggregation service and aggregate.

In the future, we hope to structure that into its own table and run aggregations directly in ClickHouse.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
phacops: I'd remove it since it's actually optional for `ReplacingMergeTree` but `mypy` forces me to have something as `version_column`.

```
snuba/migrations/snuba_migrations/profiles/0001_profiles.py:50: error: Missing positional argument "version_column" in call to "ReplacingMergeTree"
```

nikhars: You can make the `version` parameter optional in https://github.com/getsentry/snuba/blob/master/snuba/migrations/table_engines.py#L96. Clickhouse `ReplacingMergeTree` states that it is optional https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree/

nikhars (dismissed):
LGTM. Approving from my end. I will let @fpacifici give the final approval

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,86 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import UUID, Column, DateTime, NamedTuple, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+columns: List[Column[Modifiers]] = [
+    # primary key
+    Column("organization_id", UInt(64)),
```
fpacifici, line 10: Why do we need the organization id? Do we expect queries across all the projects of an organization ? If not you don't need it here.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.PROFILES,
+                table_name="profiles_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(organization_id, project_id, toStartOfDay(received), cityHash64(transaction_id))",
+                    partition_by="(retention_days, toMonday(received))",
```
fpacifici, line 52: Can we ensure the `retention_days` field will be a field chosen from a list of 2 max 3 values? Each partition is a directory that does not get merged with others, which makes it very effective to be dropped. But if you have a large quantity of granularities that would increase the number of partitions beyond control.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -0,0 +1,86 @@
...
+                    sample_by="cityHash64(transaction_id)",
+                    settings={"index_granularity": "8192"},
+                    storage_set=StorageSetKey.PROFILES,
+                    ttl="received + toIntervalDay(retention_days)",
+                ),
```
fpacifici, line 57: Are you going to add  the bloomfilter indices separately ?

fpacifici (approved):

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
phacops: Likely, the value will stay 30 days for month and probably switch to 90 days one day. So yes, it will be a list of 2 or 3 values max.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
phacops: Yes, when we start doing queries needing more filters.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
phacops: I guess it's not really necessary. Maybe in case we need to delete all data for a given organization, otherwise I believe we'll always have a project ID to select by.

We discussed that earlier and there didn't seem to be a strong opinion to remove it so I kept it https://github.com/getsentry/snuba/pull/2396#discussion_r806061404 .

PR #2398, opened 2022-02-11
Fallback host retry

onewland:
A temporary mitigation for incident-causing query node unavailability.


snuba/clickhouse/native.py
```diff
@@ -111,13 +114,16 @@ def execute(
...
+                1 if USE_FALLBACK_HOST_IN_NATIVE_CONNECTION_POOL else 0
+            )
+            use_fallback_host = False
             while attempts_remaining > 0:
                 attempts_remaining -= 1
                 # Lazily create connection instances
                 if conn is None:
                     self.__gauge.increment()
-                    conn = self._create_conn()
+                    conn = self._create_conn(use_fallback_host)
```
fpacifici, line 126: This is going to be a problem. If you assign the fallback connection to `conn`, this will be put back into the main connections pool at line 202. This means that following queries would end up picking this connection from that pool even at the first attempt to run a query. 

I would suggest keeping this conneciton around separately from the main conneciton pool after you have created it. It can be a separate (smaller) queue. So that when you decide you want to use a fallback connection, you pick it from that queue instead of recreating it new. and during normal operation the main queue will still rely on envoy

snuba/clusters/cluster.py
```diff
@@ -151,6 +151,7 @@ def get_node_connection(
         user: str,
         password: str,
         database: str,
+        fallback_host: Optional[str] = None,
```
fpacifici, line 154: Who is providing the fallback host ?
We cannot rely on it being hardcoded in settings otherwise we would have to change production settings to add/remove hosts, which would defy the purpose of the PR.

I think you should have two runtime config flags: 
- one to activate/deactivate the fallback
- another (string type) with a comma separated list of hosts.

When you instantiate a cluster (which happens at startup), pick one random host from the runtime setting. 

Alternative approach: Avoid making change to cluster.py entirely. and do everything in native.py:
- When you need a fallback host pick a random one from the runtime settings, create the connecitons and store it in the queue.
- If a new query shows up that needs the fallback host, but the existing connection is already in use, pick another host from the runtime setting and add it to the queue. 

THis is just like the current connection pool works. it keeps a queue of connections (thread safe) every time a query is executed, it tries to pop a conneciton from the queue, if there is none, it creates one, once the query is done it puts the conneciton back in the queue.

snuba/settings/__init__.py
```diff
@@ -183,6 +183,9 @@
 # rather than using materialized views
 WRITE_METRICS_AGG_DIRECTLY = False
 
+# Temporary setting to mitigate a production issue where query nodes are unreliable
+USE_FALLBACK_HOST_IN_NATIVE_CONNECTION_POOL: bool = False
```
fpacifici, line 187: Let's use only runtime settings.

snuba/clusters/cluster.py
onewland: In constructing the runtime config key for the list of fallback hosts, do you think it makes sense to use:

`fallback_hosts:<database_name>`

because then I see a clear path to getting the list from within ClickhousePool without needing to pass more information from cluster.py on creation.

snuba/clickhouse/native.py
```diff
@@ -81,12 +82,26 @@ def __init__(
         self.client_settings = client_settings
 
         self.pool: queue.LifoQueue[Optional[Client]] = queue.LifoQueue(max_pool_size)
+        self.fallback_pool: queue.LifoQueue[Optional[Client]] = queue.LifoQueue(3)
         self.__gauge = ThreadSafeGauge(metrics, "connections")
```
onewland, line 86: do we also want a gauge for fallback connections?

snuba/clickhouse/native.py
```diff
@@ -81,12 +84,31 @@ def __init__(
...
         for _ in range(max_pool_size):
             self.pool.put(None)
 
+        for _ in range(self.FALLBACK_POOL_SIZE):
+            self.fallback_pool.put(None)
+
+    def fallback_pool_enabled(self) -> bool:
+        return (
+            state.get_config("use_fallback_host_in_native_connection_pool", "False")
+            == "True"
```
fpacifici, line 102: We use 1 and 0 integers for flags. Please do not use strings, so we keep consistency across the code base.

snuba/clickhouse/native.py
```diff
@@ -269,9 +313,9 @@ def execute_robust(
             except errors.Error as e:
                 raise ClickhouseError(e.message, code=e.code) from e
 
-    def _create_conn(self) -> Client:
+    def _create_conn(self, use_fallback_host: bool = False) -> Client:
         return Client(
-            host=self.host,
+            host=(self.host if not use_fallback_host else self.get_fallback_host()),
             port=self.port,
```
fpacifici, line 321: The port can be different from the one we have on the envoy proxy. So you should also keep track of the fallback node port.

fpacifici (changes requested):

evanh:
Can you elaborate more on when this setting is going to be used? It doesn't seem like it's meant to bet set 24/7.

onewland:
> Can you elaborate more on when this setting is going to be used? It doesn't seem like it's meant to bet set 24/7.

Sure, we have an undiagnosed instability in the query nodes (we think it's likely due to the old version of Clickhouse that we're on) and it's causing some query failures. (see https://app.datadoghq.com/notebook/1955338/connection-error-graphs?start=1644440672119&range=172800000&live=true&view=view-mode). Ops can restart the query nodes when they drop connections but we want to temporarily have the Snuba API hosts pick different equivalent query nodes when this problem occurs so that customers and on-call engineers aren't impacted.

When we find the root cause of query nodes crashing with network issues, we should remove this change.

snuba/clickhouse/native.py
fpacifici: ideally yes, but feel free to do that in a separate PR

fpacifici (approved):

onewland:
> > we want to temporarily have the Snuba API hosts pick different equivalent query nodes when this problem occurs so that customers and on-call engineers aren't impacted.
> 
> Right, so when will this value be set? When we notice the query nodes having problems? Or will we hardcode an arbitrary query node to be the permanent fallback?

We'll be hardcoding some fallbacks in advance, basically overriding the normal load-balancing functions of envoy/clickhouse

snuba/clickhouse/native.py
```diff
@@ -81,12 +94,36 @@ def __init__(
...
+        return state.get_config("use_fallback_host_in_native_connection_pool", 0) == 1
+
+    def get_fallback_host(self) -> Tuple[str, int]:
+        config_hosts_str = state.get_config(
+            f"fallback_hosts:{self.host}:{self.port}", None
+        )
+        assert config_hosts_str, f"no fallback hosts found for {self.host}:{self.port}"
+
+        config_hosts = cast(str, config_hosts_str).split(",")
+        selected_host_port = random.choice(config_hosts).split(":")
```
evanh, line 119: What happens if the fallback host is also the bad host?

onewland: It'll fail the way it does today if a single attempt fails

PR #2399, opened 2022-02-11
perf(subscriptions): Avoid slow delegate task builder where possible

lynnagara:
The delegate task builder is much slower than using the immediate
or jittered one since it has to check which mode to apply for
every subscription within the tick individually. This change uses
the immediate or jittered task builder directly for the whole tick
if we are not in a transition mode. This is checked at the start
of each tick.

### DIFF
--- a/snuba/subscriptions/scheduler.py
+++ b/snuba/subscriptions/scheduler.py
@@ -295,7 +295,31 @@ class SubscriptionScheduler(SubscriptionSchedulerBase):
 
         self.__subscriptions: List[Subscription] = []
         self.__last_refresh: Optional[datetime] = None
-        self.__builder = DelegateTaskBuilder()
+
+        self.__delegate_builder = DelegateTaskBuilder()
+        self.__jittered_builder = JitteredTaskBuilder()
+        self.__immediate_builder = ImmediateTaskBuilder()
+
+        self.__reset_builder()
+
+    def __reset_builder(self) -> None:
+        """
+        Use the jittered or immediate builder directly if we can as it is faster.
+        If we are in transition between the two modes, we must use the delegate builder.
+        This function is called for every tick.
+        """
+        general_mode = TaskBuilderMode(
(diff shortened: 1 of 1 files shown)


lynnagara:
> Can you provide some background on what these builders do?

There are 2 main builders - immediate and jittered. Immediate means that the subscription is scheduled "immediately", i.e. if its resolution is 1 minute queries are run at the start of the minute, if 1 hour at the start of each hour etc. The problem we ran into is that it causes a huge load on ClickHouse at the start of each minute and hour, which is why the jittered builder was introduced. It delays the execution of each subscription by some stable jitter amount based on the subscription ID. The delegate builder supports transitioning between the modes.

snuba/subscriptions/scheduler.py
```diff
@@ -323,6 +347,7 @@ def __get_subscriptions(self) -> List[Subscription]:
         return self.__subscriptions
 
     def find(self, tick: Tick) -> Iterator[ScheduledSubscriptionTask]:
+        self.__reset_builder()
```
fpacifici, line 350: This is executed once per message on the commit log, right ?

fpacifici (approved):

snuba/subscriptions/scheduler.py
lynnagara: Yes, that's how the ticks get created. One per per message.

onewland:
> > Can you provide some background on what these builders do?
> 
> There are 2 main builders - immediate and jittered. Immediate means that the subscription is scheduled "immediately", i.e. if its resolution is 1 minute queries are run at the start of the minute, if 1 hour at the start of each hour etc. The problem we ran into is that it causes a huge load on ClickHouse at the start of each minute and hour, which is why the jittered builder was introduced. It delays the execution of each subscription by some stable jitter amount based on the subscription ID. The delegate builder supports transitioning between the modes.

and what is a builder? what does it do?

PR #2400, opened 2022-02-11
chore(suspect-spans): Remove rollout options

Zylphrex:
Removing the rollout options to make suspect spans enabled by default.


snuba/datasets/transactions_processor.py
```diff
@@ -303,13 +299,6 @@ def _process_spans(
             max_spans_per_transaction = 2000
 
         try:
```
evanh, line 301: Is this feature solid enough to remove the try..except as well?

evanh (approved):

PR #2401, opened 2022-02-11
perf(subscriptions): Increase the load factor to 3

lynnagara:
Since we have added a number of optimizations now, let's try
increasing the scheduler load factor to 3 so we can observe
how scheduler performance would respond to more subscriptions.


snuba/settings/__init__.py
```diff
@@ -175,7 +175,7 @@
 SUBSCRIPTIONS_ENTITY_BUFFER_SIZE: Mapping[str, int] = {}  # (entity name, buffer size)
 
 # Temporary setting for subscription scheduler test
-SUBSCRIPTIONS_SCHEDULER_LOAD_FACTOR = 2
+SUBSCRIPTIONS_SCHEDULER_LOAD_FACTOR = 3
```
fpacifici, line 178: what about 4 or 5 ?

fpacifici (approved):

snuba/settings/__init__.py
lynnagara: Yeah we can probably handle 3 easily, just curious in see how this goes step by step. We will probably not be on 3 for long.

PR #2403, opened 2022-02-11
ref(subscriptions): Remove subscription "type" field

lynnagara:
Whilst we were transitioning from the legacy format to SnQL for subscriptions
we had to support two formats. Now we only have SnQL subscriptions we can
remove the "type" field from subscription data. This should be safe to do
as we are no longer actually relying on this value for anything.


evanh (approved):

evanh:
I'm not sure we need the abstract class anymore either. It's not having an impact to keep it, just a thought I had.

lynnagara:
@evanh Which abstract class? Do you mean SubscriptionData and SnQLSubscriptionData? I was thinking of removing that too (inheritance is a little bit slower in Python) but as a follow up

PR #2406, opened 2022-02-11
perf(subscriptions): Increase scheduler load factor to 5

lynnagara:
Scheduler is doing fine, let's increase it's load to 5x and see what happens


snuba/settings/__init__.py
```diff
@@ -175,7 +175,7 @@
 SUBSCRIPTIONS_ENTITY_BUFFER_SIZE: Mapping[str, int] = {}  # (entity name, buffer size)
 
 # Temporary setting for subscription scheduler test
-SUBSCRIPTIONS_SCHEDULER_LOAD_FACTOR = 3
+SUBSCRIPTIONS_SCHEDULER_LOAD_FACTOR = 5
```
evanh, line 178: 6?

evanh (approved):

snuba/settings/__init__.py
lynnagara: Decided to play it safe first and go to 5, will update the description now

PR #2410, opened 2022-02-14
fix(consumers): Add storage tags to multistorage consumers

nikhars:
This PR fixes the tags which are passed to the processors of multistorage
consumers. It uses the first storage which is provided with the command
line arguments.


evanh (approved):
This looks fine, but I don't think it's correct to say that it corresponds to the first storage in the command line arguments. Dictionaries aren't ordered and `storage_keys` is not sorted after the list is built.

I don't know that it matters, but I just wanted to clarify that.

snuba/cli/multistorage_consumer.py
```diff
@@ -213,7 +213,11 @@ def multistorage_consumer(
             consumer_configuration, producer=producer, commit_log_topic=commit_log,
         )
 
-    metrics = MetricsWrapper(environment.metrics, "consumer")
+    metrics = MetricsWrapper(
+        environment.metrics,
+        "consumer",
+        tags={"group": consumer_group, "storage": storage_keys[0].value},
```
fpacifici, line 219: Want to mark these metrics in a way so that it is recognizable that they are from a multistorage consumer, thus the storage we log is only one of the supported ones? 

Examples:
- another tag that identifies a multi storage consumer
- append `_m` to the storage name, or something like that ?

nikhars: Added _m to the storage name. Used that approach since most of the consumer graphs are based off storages.

PR #2417, opened 2022-02-15
fix(storages): Allow storage definitions in conf which do not exist

nikhars:
### Context
Adding a storage today is like a chicken and egg problem. If the code
dealing with a new storage is introduced first and the settings in prod
do not exist, then because of implicit imports its possible that an
exception could be raised. If the prod settings for new storage
are introduced first then because of validation of settings there would
be an exception since the storage mentioned is not in the code.

### Solution
This PR tries to break this cycle by allowing storages to be mentioned
in the settings without the corresponding storage to be defined in
code. It does that by ensuring that the storage to cluster map is only
populated if the storage is defined in code.

### Tests
Added a  unit test which tests the `to_set` method introduced in this PR.

Also ran Snuba with a non-existent storage to ensure Snuba comes up
The output of the run can be found here https://gist.github.com/nikhars/461b2bc20514d9d3f29d6ca6031df4ad

### DIFF
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -1,7 +1,7 @@
 from typing import Any, Mapping, MutableMapping
 
 
-def _validate_settings(locals: Mapping[str, Any]) -> None:
+def validate_settings(locals: Mapping[str, Any]) -> None:
     if locals.get("QUERIES_TOPIC"):
         raise ValueError("QUERIES_TOPIC is deprecated. Use KAFKA_TOPIC_MAP instead.")
 
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -55,8 +55,12 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
 
     for cluster in locals["CLUSTERS"]:
         for cluster_storage_set in cluster["storage_sets"]:
-            if cluster_storage_set in StorageSetKey.to_set():
+            try:
                 storage_set_to_cluster[StorageSetKey(cluster_storage_set)] = cluster
+            except ValueError:
+                # We allow definition of storage_sets in configuration files
+                # that are not defined in StorageSetKey.
+                pass
 
     for group in JOINABLE_STORAGE_SETS:
         clusters = [storage_set_to_cluster[storage_set] for storage_set in group]
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -55,7 +55,8 @@ def _validate_settings(locals: Mapping[str, Any]) -> None:
 
     for cluster in locals["CLUSTERS"]:
         for cluster_storage_set in cluster["storage_sets"]:
-            storage_set_to_cluster[StorageSetKey(cluster_storage_set)] = cluster
+            if cluster_storage_set in StorageSetKey.to_set():
+                storage_set_to_cluster[StorageSetKey(cluster_storage_set)] = cluster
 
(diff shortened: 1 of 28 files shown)


nikhars:
> The change looks fine but the test doesn't test that the functionality you wanted (the new storage settings can be defined without the storage and pass validation)

The functionality is getting implicitly tested because of the implicit imports. That's not too good though. I will add an explicit test.

nikhars:
> Why don't you just remove those validation lines as it doesn't seem like it's validating anything important now?
> 
> Also can we consider reverting [4631aea](https://github.com/getsentry/snuba/commit/4631aeae677cf9e2fac8df156f65dcdab93a3793) which attempted to solve the problem in the other direction?

The validation which happens over at https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py#L61 requires that the storage_set_to_cluster map be populated. Hence I am not removing it for now.

IMO, we should not revert this PR https://github.com/getsentry/snuba/commit/4631aeae677cf9e2fac8df156f65dcdab93a3793. This is what allowed me from adding `transactions_v2` storage set without any issues. In an ideal scenario, I would like us to be in a state where we can add new storage set keys in code without the production settings being there. But that requires a some effort in refactoring.

snuba/settings/validation.py
```diff
@@ -55,7 +55,8 @@ def _validate_settings(locals: Mapping[str, Any]) -> None:
 
     for cluster in locals["CLUSTERS"]:
         for cluster_storage_set in cluster["storage_sets"]:
-            storage_set_to_cluster[StorageSetKey(cluster_storage_set)] = cluster
+            if cluster_storage_set in StorageSetKey.to_set():
+                storage_set_to_cluster[StorageSetKey(cluster_storage_set)] = cluster
```
lynnagara, line 59:
```suggestion
try:
    storage_set_to_cluster[StorageSetKey(cluster_storage_set)] = cluster
except ValueError:
    pass
```
You could remove StorageSetKey.to_set

lynnagara:
I'm not sure I follow this. This way this PR is proposing is that the storage set is added to cluster in settings before it actually exists in the codebase. So if we are doing this way we should be able to put back the validation that all storage sets are assigned. Sure, it can work fine if we just remove all validation on both sides, but I don't think it's a good idea.

> IMO, we should not revert this PR https://github.com/getsentry/snuba/commit/4631aeae677cf9e2fac8df156f65dcdab93a3793. This is what allowed me from adding transactions_v2 storage set without any issues. In an ideal scenario, I would like us to be in a state where we can add new storage set keys in code without the production settings being there. But that requires some effort of refactoring.

nikhars:
> I'm not sure I follow this. This way this PR is proposing is that the storage set is added to cluster in settings before it actually exists in the codebase. So if we are doing this way we should be able to put back the validation that all storage sets are assigned. Sure, it can work fine if we just remove all validation on both sides, but I don't think it's a good idea.
> 

There is a bigger discussion which has to happen to nail down what we want and how to fix this in the long term. I opened a ticket so we can track it. If you feel very strongly about putting the other validation back, we can do it in another PR.

My primary motive right now is to get unblocked so that we can start writing to the new clusters.

nikhars:
> The change looks fine but the test doesn't test that the functionality you wanted (the new storage settings can be defined without the storage and pass validation)

Test has been added now

lynnagara (approved):
LGTM, thanks. We can tackle the other thing later.

PR #2419, opened 2022-02-15
perf(subscriptions): Simplify SubscriptionData, only validate on creation

lynnagara:
This PR does 2 things:

1. Consolidates the SubscriptionData class into a single one, instead of a SnQLSubscriptionData
class which inherits from a SubscriptionData class which inherits from another _SubscriptionData class.
Since don't need to support non-SnQL subscription data anymore so we don't need the base class.

2. Moves validation out of `__post__init__` into a separate validate() function. The validate() function
is now only called upon initial creation of the subscription and not every single time it's loaded from
the store or decoded from already scheduled task data.

### DIFF
--- a/bin/mocks/mock-subscriptions
+++ b/bin/mocks/mock-subscriptions
@@ -6,7 +6,7 @@ from datetime import timedelta
 from snuba.datasets.entities.factory import ENTITY_NAME_LOOKUP
 from snuba.datasets.factory import get_dataset
 from snuba.redis import redis_client
-from snuba.subscriptions.data import PartitionId, SnQLSubscriptionData
+from snuba.subscriptions.data import PartitionId, SubscriptionData
 from snuba.subscriptions.entity_subscription import ENTITY_KEY_TO_SUBSCRIPTION_MAPPER
 from snuba.subscriptions.store import RedisSubscriptionDataStore
 from snuba.subscriptions.subscription import SubscriptionCreator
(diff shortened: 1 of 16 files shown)


nikhars (approved):
LGTM. Curious to know how much percentage performance gain you got from removing Subclassing. Do you have any numbers or you need to test it out?

fpacifici:
> We create these objects very frequently in the scheduler and executor and subclassing is slower.
Since we don't need to support non-SnQL subscription data anymore, having the base class doesn't add
much value.

Is this something you find out with profiling? Do you have the report that shows there is a performance degradation due to subclassing? This seems a bit weird.

lynnagara:
> > We create these objects very frequently in the scheduler and executor and subclassing is slower.
> > Since we don't need to support non-SnQL subscription data anymore, having the base class doesn't add
> > much value.
> 
> Is this something you find out with profiling? Do you have the report that shows there is a performance degradation due to subclassing? This seems a bit weird.

Nah, I did see creating the object was a slow but did not verify that subclassing actually makes a difference. I always thought that subclassing was slower, but now I am doubting where I got that belief from. I ran some tests locally and it does not actually seem to make much difference either way.

Consolidating the classes was something we discussed doing as part of the SnQL subscriptions cleanup so I wouldn't mind keeping the code anyway, though I can pull that change out of this perf PR into a separate one if you like.

fpacifici:
If consolidation is something that we planned to do anyway, all good. 
Though if it is just a shot in the dark to improve performance (which would reduce code and design quality) we should not do that without profiling that provide evidence. 
In this case it seems it was planned anyway, so no problem with that.

PR #2420, opened 2022-02-16
fix(runtime config): Execute set_config() Redis calls in a pipeline

rahul-kumar-saini:
### Small optimization
`set_config()` has multiple Redis calls that could be put in a pipeline to reduce network calls from 4 to 1.

### DIFF
--- a/snuba/state/__init__.py
+++ b/snuba/state/__init__.py
@@ -131,14 +131,16 @@ def set_config(
                 raise MismatchedTypeException(key, type(original_value), type(value))
 
         change_record = (time.time(), user, enc_original_value, enc_value)
+        p = rds.pipeline()
         if value is None:
-            rds.hdel(config_hash, key)
-            rds.hdel(config_history_hash, key)
+            p.hdel(config_hash, key)
+            p.hdel(config_history_hash, key)
         else:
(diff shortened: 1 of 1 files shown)


vladkluev (approved):

snuba/state/__init__.py
```diff
@@ -131,14 +131,16 @@ def set_config(
...
+            p.hdel(config_history_hash, key)
         else:
-            rds.hset(config_hash, key, enc_value)
-            rds.hset(config_history_hash, key, json.dumps(change_record))
-        rds.lpush(config_changes_list, json.dumps((key, change_record)))
-        rds.ltrim(config_changes_list, 0, config_changes_list_limit)
+            p.hset(config_hash, key, enc_value)
+            p.hset(config_history_hash, key, json.dumps(change_record))
+        p.lpush(config_changes_list, json.dumps((key, change_record)))
+        p.ltrim(config_changes_list, 0, config_changes_list_limit)
```
vladkluev, line 142: how does updating and then modifying a value work with the redis pipeline?

rahul-kumar-saini: Redis pipelined commands are executed in order, so the value would be updated then trimmed like it would have if we made those calls one by one instead of in a pipeline.

PR #2421, opened 2022-02-16
fix(cron): Allow transactions_v2 storage for cron jobs

nikhars:
In order to run the cron jobs on the new storages we need to allow
passing in the new storage

### DIFF
--- a/snuba/cli/optimize.py
+++ b/snuba/cli/optimize.py
@@ -19,7 +19,7 @@ from snuba.environment import setup_logging, setup_sentry
 @click.option(
     "--storage",
     "storage_name",
-    type=click.Choice(["events", "errors", "transactions"]),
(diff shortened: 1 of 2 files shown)


evanh (approved):

snuba/cli/optimize.py
```diff
@@ -19,7 +19,7 @@
 @click.option(
     "--storage",
     "storage_name",
-    type=click.Choice(["events", "errors", "transactions"]),
+    type=click.Choice(["events", "errors", "transactions", "transactions_v2"]),
```
fpacifici, line 22: We do not run optimize on transactions today.

nikhars: I have removed transactions option from optimize job. If we need it in the future we can add it back.

PR #2423, opened 2022-02-18
ref(ratelimits): make project|referrer rate limiter able to be applied across projects

vladkluev:
With the way the ProjectReferrer rate limiter was done before, we would have to specify a limit for every referrer/project pair:

`project_referrer_concurrent_limit_{project}_{referrer}`
`project_referrer_per_second_limit_{project}_{referrer}`

This was fine to protect us against specific abusive customers but not as good when product teams shipped a change that affected all of them. This PR changes that behavior. Now one can specify the same rate limit across referrers:

`project_referrer_concurrent_limit_{referrer}`
`project_referrer_per_second_limit_{referrer}`

This will create a per-project limit across the specified referrer, allowing for even rate limiting of problematic referrers across all customers.

### Breaking changes

Because of the way this key is constructed, old `project_referrer` keys will not work as before:

Keys of this format:

`project_referrer_concurrent_limit_{project}_{referrer}`
`project_referrer_per_second_limit_{project}_{referrer}`

Will have to become:

`project_referrer_concurrent_limit_{referrer}_{project}`
`project_referrer_per_second_limit_{referrer}_{project}`

### DIFF
--- a/snuba/datasets/entities/events.py
+++ b/snuba/datasets/entities/events.py
@@ -25,6 +25,7 @@ from snuba.query.processors.handled_functions import HandledFunctionsProcessor
 from snuba.query.processors.object_id_rate_limiter import (
     ProjectRateLimiterProcessor,
     ProjectReferrerRateLimiter,
+    ReferrerRateLimiterProcessor,
 )
 from snuba.query.processors.tags_expander import TagsExpanderProcessor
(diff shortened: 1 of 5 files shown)


fpacifici (approved):
please fix the test failures before merging.
They are relevant.

vladkluev:
and here I was about to abuse my admin privileges

PR #2424, opened 2022-02-19
perf(quota) Make max thread configurable per referrer and project

fpacifici:
Configuring the rate limiter works to protect the system though it does not help when we have continuous moderate load that is far from being limited by concurrency. By having low concurrency every query uses a large amount of threads.
This is ok for the UI for a short period of time, but it is a problem when it comes from script that hit the api continuously. 
This allows to manually set the max thread config setting by runtime config and reduce quota for specific referrer/projects.

A further improvement would be to support a proper quota system where the max_thread field is calculated based on the past behavior of the project and adapted to enforce a quota over a period of time.

### DIFF
--- a/snuba/request/request_settings.py
+++ b/snuba/request/request_settings.py
@@ -64,7 +64,7 @@ class RequestSettings(ABC):
         pass
 
     @abstractmethod
-    def add_resource_quota(self, quota: ResourceQuota) -> None:
+    def set_resource_quota(self, quota: ResourceQuota) -> None:
         pass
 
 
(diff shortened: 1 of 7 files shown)


snuba/request/request_settings.py
```diff
@@ -119,6 +129,12 @@ def get_rate_limit_params(self) -> Sequence[RateLimitParameters]:
     def add_rate_limit(self, rate_limit_param: RateLimitParameters) -> None:
         self.__rate_limit_params.append(rate_limit_param)
 
+    def get_resource_quota(self) -> Optional[ResourceQuota]:
+        return self.__resource_quota
+
+    def add_resource_quota(self, quota: ResourceQuota) -> None:
```
evanh, line 135: I would change this to `set_resource_quota` since you are overriding any previous settings.

evanh (approved):
Just one small change, otherwise looks good.

PR #2425, opened 2022-02-21
feat(errors upgrade) Set up query path for the errors v2 storage

fpacifici:
This includes most of what is needed to direct a portion of errors queries to the new storage `errors_v2`.
It uses the PipelineDelegator to run two queries in parallel and compare the results.
It introduces `RolloutSelector` to define the rollout policy. So we can control the amount of queries that go to `errors_vs` by referrer and rely on `errors_v2` for the production results when needed.

Still to do:
- add a callback to compare the results
- register errors_v2_ro

### DIFF
--- a/snuba/datasets/entities/clickhouse_upgrade.py
+++ b/snuba/datasets/entities/clickhouse_upgrade.py
@@ -1,8 +1,7 @@
 from enum import Enum
 from random import random
-from typing import Mapping, NamedTuple, Optional
+from typing import NamedTuple, Optional, cast
 
-from snuba import settings
 from snuba.state import get_config
 
 
@@ -31,7 +30,6 @@ class RolloutSelector:
     There are multiple way to configure this rollout:
     First choice: which query should be trusted:
     - check runtime config if the default secondary should be trusted
-    - check settings if the default secondary should be trusted
     - check global rollout if the default secondary should be trusted
     Now we know which query will be trusted. hould we run the second ?
     Check the configs and settings in the same order.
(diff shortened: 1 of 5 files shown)


fpacifici:
@nikhars This should set up a viable query path for `errors_v2`
@lynnagara  for review of the PipelineDelegator usage.

snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -0,0 +1,96 @@
...
+        return random() <= rollout_percentage
+
+    def choose(self, referrer: str) -> Choice:
+        trust_secondary = self.__is_query_rolled_out(
+            referrer,
+            "trust",
+            settings.ERRORS_UPGRADE_TRUST_SECONDARY,
+            f"rollout_upgraded_{self.__config_prefix}_trust",
+            settings.ERRORS_UPGRADE_TRUST_SECONDARY_GLOBAL,
+        )
```
lynnagara, line 75: If this is true, I don't think we should run the other one in the thread, just return `("errors_v2", [])` here. Reason is we shouldn't be doing this until we are really confident and the other one isn't necessary. For similar reasons I also think if we trust the secondary, we shouldn't need to sample anymore.

snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -0,0 +1,96 @@
...
+
+        execute_both = self.__is_query_rolled_out(
+            referrer,
+            "execute",
+            settings.ERRORS_UPGRADE_EXECUTE_BOTH,
+            f"rollout_upgraded_{self.__config_prefix}_execute",
+            settings.ERRORS_UPGRADE_EXECUTE_BOTH_GLOBAL,
+        )
```
lynnagara, line 83: Can we do anything to make setting these more straightforward? It feels like a huge number of knobs to turn, and took me a little while to decipher the interactions between them. On top of that you can also set some of the same items via both settings and config mechanism.

lynnagara (commented):
I got the impression that this is designed for a fairly gradual rollout process since there's so many granular ways to rollout different referrers via config and settings.

Whilst I think it's a good idea to increase traffic to the new cluster 1 referrer at a time, I think it's safer to not even begin using errors v2 as primary anywhere until we are confident and trust all the results from it everywhere (i.e. all referrers).

fpacifici:
@lynnagara , regarding when to move to the secondary.
This use case is a bit different than the one you worked on last year. There are two main difference:
- We know the primary has stability issues. Hopefully it won;'t happen but we may have to quickly move traffic to the new cluster in an emergency. In that case we would take the risk of rolling out results that may not be fully accurate in order to avoid large incidents with the old cluster.
- In practice, the two schemas are identical here. Only difference is the clickhouse version. There are a lot fewer chances of running into discrepancies as you did last year.  So I would expect us to not have to spend too long in validation.

snuba/datasets/entities/clickhouse_upgrade.py
fpacifici: Will try

snuba/datasets/entities/clickhouse_upgrade.py
fpacifici: See my comment below on why we may have to move faster in moving traffic to the new cluster in case of emergencies. We may well be in a place where we have no choice other than move traffic to the new cluster, still we would be able to still get some insight on accuracy with this comparison.

snuba/datasets/entities/clickhouse_upgrade.py
fpacifici: done

PR #2426, opened 2022-02-22
feat(errorsv2) Similarity score algorithm to compare results from multiple Clickhouse versions

fpacifici:
Based on #2425 .

This implements the callback function for the PipelineDelegator. The callback has access to both the primary and secondary query results and tries to calculate a similarity score between the two results.

The idea is to divide each row of the results in a key (strings, UUID, datetime), which are generally what the user put in the GROUP BY, and the numeric fields, which are the results of aggregations. (For simplicity this only relies on the data type to decide what is key and what is value).
If two results are very similar, the keys should match and the values for matching keys should be close to each other.
So the score is the average similarity score for matching rows. Each pair of matching rows (same key between the two results) has a similarity score. 
The similarity score of a pair of matching rows is calculated as the average of the ratios of  matching columns between the two rows.

Any pair of results that do not share the same schema (same columns) get a score of 0.0.

cc @nikhars

### DIFF
--- a/snuba/datasets/entities/clickhouse_upgrade.py
+++ b/snuba/datasets/entities/clickhouse_upgrade.py
@@ -170,8 +170,11 @@ def calculate_score(
         result1_index[row.key].append(row)
 
     row_similarities: MutableSequence[float] = []
+    result2_index: MutableMapping[Key, List[SplitRow]] = defaultdict(list)
     for i in range(min(len(data_row2), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
         row = split_row(data_row2[i], schema)
+        result2_index[row.key].append(row)
+
         if row.key not in result1_index:
             row_similarities.append(0.0)
         else:
@@ -185,17 +188,10 @@ def calculate_score(
             ]
             row_similarities.append(max(key_similarities))
 
-    result2_index: MutableMapping[Key, List[SplitRow]] = defaultdict(list)
-    for i in range(min(len(data_row2), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
-        row = split_row(data_row2[i], schema)
-        result2_index[row.key].append(row)
-
-    for i in range(min(len(data_row1), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
-        row = split_row(data_row1[i], schema)
-        if row.key not in result2_index:
+    for key in result1_index.keys():
+        if key not in result2_index:
             row_similarities.append(0.0)
 
-    print(row_similarities)
     return (
         sum(row_similarities) / len(row_similarities)
         if len(row_similarities) > 0
--- a/snuba/datasets/entities/clickhouse_upgrade.py
+++ b/snuba/datasets/entities/clickhouse_upgrade.py
@@ -6,7 +6,6 @@ from enum import Enum
 from random import random
 from typing import (
     List,
-    Mapping,
     MutableMapping,
     MutableSequence,
     NamedTuple,
@@ -54,7 +53,6 @@ class RolloutSelector:
     There are multiple way to configure this rollout:
     First choice: which query should be trusted:
     - check runtime config if the default secondary should be trusted
-    - check settings if the default secondary should be trusted
     - check global rollout if the default secondary should be trusted
     Now we know which query will be trusted. hould we run the second ?
     Check the configs and settings in the same order.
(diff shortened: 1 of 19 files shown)


snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -94,3 +117,179 @@ def choose(self, referrer: str) -> Choice:
...
+
+    - We match the rows that are present in both results (matching the non decimal
+    fields)
+    - The similarity score is the average of the similarity of each number column.
+      Ex. row1 : (a, 10, 100, 1000) row2 (a, 9, 99, 990). The similarity between
+      these two is the avg([(smaller value / bigger value) for each column])
+    - Calculate the average of the similarity scores of all the matching rows.
+    - Rows that do not have a match have a similarity score = 0.0
+    """
+    if primary_result is None or len(results) == 0:
```
evanh, line 148: It might be worth checking to make sure the secondary result also has 0 rows?

snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -94,3 +117,179 @@ def choose(self, referrer: str) -> Choice:
...
+        return
+
+    secondary_result = results[0]
+    primary_schema = split_metadata(primary_result.result)
+    secondary_schema = split_metadata(secondary_result.result)
+
+    if primary_schema != secondary_schema:
+        metrics.increment("schema_mismatch", tags={"referrer": referrer})
```
evanh, line 156: For some of these error cases I might also log out more information, such as which columns are missing.

snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -94,3 +117,179 @@ def choose(self, referrer: str) -> Choice:
...
+    for i in range(min(len(data_row2), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
+        row = split_row(data_row2[i], schema)
+        result2_index[row.key].append(row)
+
+    for i in range(min(len(data_row1), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
+        row = split_row(data_row1[i], schema)
+        if row.key not in result2_index:
+            row_similarities.append(0.0)
+
+    print(row_similarities)
```
evanh, line 217: 💀

snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -94,3 +117,179 @@ def choose(self, referrer: str) -> Choice:
...
+            # We choose the highest match as that is most likely the row in
+            # result1 that correspond to the row we are examining in reult2.
+            key_similarities = [
+                row_similarity(row, matching_row) for matching_row in matching_rows
+            ]
+            row_similarities.append(max(key_similarities))
+
+    result2_index: MutableMapping[Key, List[SplitRow]] = defaultdict(list)
+    for i in range(min(len(data_row2), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
+        row = split_row(data_row2[i], schema)
```
evanh, line 209: You're doing this work twice, I would build `result2_index` in the loop above.

snuba/datasets/entities/clickhouse_upgrade.py
```diff
@@ -94,3 +117,179 @@ def choose(self, referrer: str) -> Choice:
...
+        result2_index[row.key].append(row)
+
+    for i in range(min(len(data_row1), settings.MAX_ROWS_TO_CHECK_FOR_SIMILARITY)):
+        row = split_row(data_row1[i], schema)
```
evanh, line 213: This key has already been calculated and is available in `result1_index`. You could iterate through that instead.

evanh (commented):
This mostly makes sense, but I wonder if you should factor in ORDER BY in your scoring. I would expect a query that doesn't have ORDER BY to have a lower similarity score, even if the two results are functionally the same.

lynnagara:
I think logging the raw values to Sentry in the case of mismatches is the most useful thing we can do to begin with, as there may be something we can do to actually reduce the discrepancies or improve the similarity calculation. The similarity score on its own is kind of a black box and it will be hard to action or even just improve the calculation if we don't know what the underlying differences actually are.

fpacifici:
@lynnagara  Yes I will as a followup log a percentage of discrepancies in sentry. 
I want to first have a sense on the volume of events we would be logging from the metrics alone

fpacifici:
@evanh 

> I wonder if you should factor in ORDER BY in your scoring. I would expect a query that doesn't have ORDER BY to have a lower similarity score, even if the two results are functionally the same.

In such case (two result sets sorted differently but containing the same data), I would like the similarity score to be 1.0. Unless the order by is provided an not honored by the database. Which I think is such an unlikely case that we may not need to bother checking it.

snuba/datasets/entities/clickhouse_upgrade.py
fpacifici: I will as a follow up. I want to have a sense of the volume I would log first.

evanh:
>  I would like the similarity score to be 1.0.

It would only matter if there were > `MAX_ROWS_TO_CHECK_FOR_SIMILARITY` results, but if you compare the outputs of a query that doesn't have an ORDER BY, you may get a low similarity score despite the results actually being identical.

fpacifici:
> It would only matter if there were > MAX_ROWS_TO_CHECK_FOR_SIMILARITY results,
That is meant to throttle down in case we have performance issues in the callback. In practice I would like to keep it above 10000 so it should not be a problem.

snuba/datasets/entities/clickhouse_upgrade.py
fpacifici: `len(results) == 0` means there are no secondary results and we ran only one query.

evanh (approved):

PR #2430, opened 2022-02-22
fix(snql) Correctly assign aliases to expressions with none

evanh:
If a non-Column expression inside a SelectedExpression does not have an alias
assigned, give it the same expression as the SelectedExpression, which in this
case is the raw string passed into the parser.

This prevents a warning later that an expression is missing an alias when the
results are being processed. Once this is out and working I will remove the
warning here: https://github.com/getsentry/snuba/blob/master/snuba/web/query.py#L256

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -714,6 +714,8 @@ class SnQLVisitor(NodeVisitor):  # type: ignore
             return exp
 
         alias = exp.alias or node.text.strip()
+        if not isinstance(exp, Column) and exp.alias is None:
+            exp = replace(exp, alias=alias)
         return SelectedExpression(alias, exp)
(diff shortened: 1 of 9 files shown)


snuba/query/snql/parser.py
```diff
@@ -1305,11 +1307,6 @@ def validate_entities_with_query(
                 f"validation failed for entity {query.get_from_clause().key.value}: {e}",
                 should_report=e.should_report,
             )
-        except InvalidExpressionException as e:
```
evanh, line 1308: This is unrelated to the change, but I fixed it anyways. This check is redundant since `InvalidExpressionException` is a subclass of `InvalidQueryException`.

volokluev (approved):

PR #2432, opened 2022-02-22
feat(runtime config): Add support for descriptions for configs

rahul-kumar-saini:
### Overview
- Unless you look through the codebase, it's unclear what a runtime config does or what the expected format is (if it's a string)
- Added functionality to be able to add descriptions for runtime configs

descriptions

### Changes
- Added a new hash in Redis to hold descriptions for runtime configs
- Updated UI to add a box for descriptions for a config
- Updated API interface on backend + frontend to reflect changes


snuba/admin/views.py
```diff
@@ -215,6 +220,15 @@ def configs() -> Response:
         )
 
 
+@application.route("/all_config_descriptions", methods=["GET"])
+def all_config_descriptions() -> Response:
```
evanh, line 224: What is this for?

snuba/admin/static/data.tsx
```diff
@@ -8,21 +8,21 @@ function Placeholder(props: any) {
 }
 
 const NAV_ITEMS = [
-  { id: "overview", display: "Overview", component: Placeholder },
-  { id: "config", display: "Runtime config", component: RuntimeConfig },
+  { id: "overview", display: "🤿 Overview", component: Placeholder },
```
evanh, line 11: Moar emojis!

snuba/admin/static/runtime_config/index.tsx
```diff
@@ -44,10 +61,17 @@ function RuntimeConfig(props: { api: Client }) {
     });
   }
 
+  function fetchDescriptions() {
```
evanh, line 64: Can you explain why you need this? I'm not sure why you would fetch descriptions without fetching the associated config value. Is this for "deleted but still described" configs?

evanh (commented):
I think you need a test for the "keep description" deletion case.

snuba/admin/static/runtime_config/index.tsx
rahul-kumar-saini: Yep, this would be called in the "new config" flow.

When we add a new config:
- We fetch all available descriptions (which includes descriptions saved but config was deleted)
- https://github.com/getsentry/snuba/blob/b44edbde95b5f9aaedf26ec5a3fa52599103d264/snuba/admin/static/runtime_config/index.tsx#L70-L75
- As we write the name for the config, we check the map of descriptions to see whether we already have a description for this key
- If it's found, the description box is populated
- https://github.com/getsentry/snuba/blob/b44edbde95b5f9aaedf26ec5a3fa52599103d264/snuba/admin/static/runtime_config/index.tsx#L185-L192

snuba/admin/views.py
rahul-kumar-saini: (Same reason as the other comment)

evanh (approved):
Sorry for the delay in approval. Looks good to me.

PR #2433, opened 2022-02-23
feat(ratelimit): Remove redundant code from object Id ratelimiters

volokluev:
I copy-pasted this code in a hurry and thus violated DRY. as part of tech debt week undo my hastiness and add a docstring.


snuba/query/processors/object_id_rate_limiter.py
```diff
@@ -188,7 +170,7 @@ def get_object_id(
         return str(obj_id)
 
 
-class ReferrerRateLimiterProcessor(OnlyIfConfiguredRateLimitProcessor):
+class ReferrerRateLimiterProcessor(ObjectIDRateLimiterProcessor):
```
evanh, line 173: I don't think this should inherit from the `ObjectIDRateLimiterProcessor`. It isn't filtering by an object ID at all. It can be its own separate class.

PR #2434, opened 2022-02-23
ref(consumer-util): add replyq metric to multistorage consumer

MeredithAnya:
This is basically the same thing as https://github.com/getsentry/snuba/pull/2245, but applied to the multistorage consumers. (again for reference, the stats definitions for librdkafka can be found [here](https://github.com/edenhill/librdkafka/blob/master/STATISTICS.mdl)). 

 
* **cli command vs. runtime config**: ~we could limit by consumer group in the runtime config, but there can be multiple consumers in the same consumer group so for now I just want to be able to test reporting these metrics for specific consumers to make sure things look ok~ Changed to use the runtime config, lets try it out with a low risk consumer group and go from there
* **more stats metrics**: testing locally `replyq` seemed to be the correct metric for queue size. As for other stats metrics, I don't think I understand what the values represent enough at the moment to add them to the list. I think once we fine tune what frequency we want for the callback, we should have no problem adding to the list

### DIFF
--- a/snuba/cli/multistorage_consumer.py
+++ b/snuba/cli/multistorage_consumer.py
@@ -14,6 +14,7 @@ from snuba.consumers.consumer import MultistorageConsumerProcessingStrategyFacto
 from snuba.datasets.storages import StorageKey
 from snuba.datasets.storages.factory import WRITABLE_STORAGES, get_writable_storage
 from snuba.environment import setup_logging, setup_sentry
+from snuba.state import get_config
 from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.streams.configuration_builder import (
     build_kafka_consumer_configuration,
@@ -80,11 +81,6 @@ logger = logging.getLogger(__name__)
     "--output-block-size", type=int,
 )
 @click.option("--log-level")
-@click.option(
-    "--stats-collection-frequency-ms",
-    type=click.IntRange(100, 2000),
-    help="The frequency of collecting statistics from librdkafka.",
-)
 def multistorage_consumer(
     storage_names: Sequence[str],
     consumer_group: str,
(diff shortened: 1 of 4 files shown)


snuba/cli/multistorage_consumer.py
```diff
@@ -192,6 +199,27 @@ def multistorage_consumer(
...
+            "group": consumer_group,
+            "storage": "_".join([storage_keys[0].value, "m"]),
+        },
+    )
+    # Collect metrics from librdkafka if we set stats_collection_frequency_ms
+    if stats_collection_frequency_ms and stats_collection_frequency_ms > 0:
+
+        def stats_callback(stats_json: str) -> None:
+            stats = rapidjson.loads(stats_json)
+            metrics.gauge("total_queue_size", stats.get("replyq", 0))
```
volokluev, line 215: while we're here is there anything else we want to send?

snuba/cli/multistorage_consumer.py
```diff
@@ -79,6 +80,11 @@
     "--output-block-size", type=int,
 )
 @click.option("--log-level")
+@click.option(
```
volokluev, line 83: Why not have this be something set by runtime config? That way we don't need ops to change something for us if we want to tweak this

MeredithAnya: I think @nikhars at first didn't do it that way so that we didn't have to enable it for all consumers just yet

volokluev: We could set it per-consumer though. If the key is not set, don't turn it on?

MeredithAnya: but we can think about using the runtime config and maybe using the consumer group appended to the value to do the look up

fpacifici:
https://github.com/getsentry/snuba/pull/2245#issuecomment-1048491856

I don;t think replyq is the metric we are looking for. See my comment on the comment I left on the other pr.

snuba/cli/multistorage_consumer.py
MeredithAnya: I'm going to hold off for now until I have a better understand of what each of the metrics are actually doing.

volokluev (approved):
I still think we should do this with runtime config but I won't block the PR if you feel strongly about not doing it

fpacifici (commented):
Looks good to me. please get to an agreement with @volokluev  around the runtime flag

PR #2435, opened 2022-02-23
feat(profiling): Read and write data from and to Snuba

phacops:
This is a PR to allow reading and writing from the new `profiles` table in ClickHouse. Snuba should read messages from a Kafka topics containing the processed profiles (after symbolication) and should insert them into ClickHouse. Fields are 1 to 1 mapping to the columns.


snuba/datasets/profiles_processor.py
```diff
@@ -0,0 +1,38 @@
...
+                "duration_ns": message["duration_ns"],
+                "environment": message.get("environment"),
+                "platform": message["platform"],
+                "trace_id": message["trace_id"],
+                "version_name": message["version_name"],
+                "version_code": message["version_code"],
+                "retention_days": 30,
+            }
+        except KeyError:
+            return None
```
evanh, line 37: I would add a metric or a log here so you can debug this in case it breaks. You can look at the transactions processor for an example of how to set up metrics in a processor.

snuba/cli/devserver.py
```diff
@@ -135,6 +135,16 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
...
+        (
+            "profiles",
+            [
+                "snuba",
+                "consumer",
+                "--auto-offset-reset=latest",
+                "--log-level=debug",
+                "--storage=profiles",
+            ],
+        ),
```
lynnagara, line 147: Do you want to turn this on by default for Sentry dev already? The migrations are still skipped so it won't really work unless that switch is flipped too. May be better to keep this line behind the ENABLE_DEV_FEATURES flag or something.

snuba/utils/streams/topics.py
```diff
@@ -25,12 +25,14 @@ class Topic(Enum):
     SUBSCRIPTION_RESULTS_SESSIONS = "sessions-subscription-results"
     SUBSCRIPTION_RESULTS_METRICS = "metrics-subscription-results"
     QUERYLOG = "snuba-queries"
+    PROFILES = "processed-profiles"
```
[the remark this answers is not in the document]
phacops: No one yet, I'm making that PR right now.

snuba/migrations/snuba_migrations/profiles/0001_profiles.py
```diff
@@ -25,12 +25,11 @@
...
     Column("duration_ns", UInt(64)),
     Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
-    Column("error_code", String(Modifiers(low_cardinality=True, nullable=True))),
-    Column("error_description", String(Modifiers(low_cardinality=True, nullable=True))),
     Column("platform", String(Modifiers(low_cardinality=True))),
     Column("trace_id", UUID()),
     Column("transaction_name", String(Modifiers(low_cardinality=True))),
-    Column("version", NamedTuple((("name", String()), ("code", String())))),
+    Column("version_name", String()),
+    Column("version_code", String()),
```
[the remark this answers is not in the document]
phacops: Of course, that's why I allowed myself to do that.

snuba/utils/streams/topics.py
```diff
@@ -25,12 +25,14 @@ class Topic(Enum):
...
 
 
 def get_topic_creation_config(topic: Topic) -> Mapping[str, str]:
     config = {
         Topic.EVENTS: {"message.timestamp.type": "LogAppendTime"},
         Topic.METRICS: {"message.timestamp.type": "LogAppendTime"},
+        Topic.PROFILES: {"message.timestamp.type": "LogAppendTime"},
```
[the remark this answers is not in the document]
phacops: If I understand correctly, subscriptions/alerting is useful to send an email to a user saying something happened, right? I don't think we'll want to do that on individual profiles, I think it's something we'd want to do on aggregations (like, the 95th percentile timing of that function went over X ms when looking at high end phones on iOS). That's why I left it out for now because I don't think that's how I'd do that right now?

snuba/datasets/profiles_processor.py
```diff
@@ -0,0 +1,38 @@
...
+                "device_os_build_number": message.get("device_os_build_number"),
+                "device_os_name": message["device_os_name"],
+                "device_os_version": message["device_os_version"],
+                "duration_ns": message["duration_ns"],
+                "environment": message.get("environment"),
+                "platform": message["platform"],
+                "trace_id": message["trace_id"],
+                "version_name": message["version_name"],
+                "version_code": message["version_code"],
+                "retention_days": 30,
```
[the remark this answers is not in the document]
phacops: It's more of a Specto thing right now, we only kept 30 days of data. I suspect we'll want to change as we shape the business-side of the product or align on whatever Sentry is doing. It seemed like a good default though.

snuba/datasets/storages/profiles.py
```diff
@@ -0,0 +1,117 @@
...
+        ("platform", String()),
+        ("trace_id", UUID()),
+        ("transaction_name", String()),
+        ("version_name", String()),
+        ("version_code", String()),
+        ("retention_days", UInt(16)),
+        ("partition", UInt(16)),
+        ("offset", UInt(64)),
+    ]
+)
```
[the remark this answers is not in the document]
phacops: One random question about `offset`/`partition`: I'm not sure what I'm supposed to do with those. I added them after a comment later one but I don't think I plan to populate them myself in any ways. Should I just remove them?

snuba/utils/streams/topics.py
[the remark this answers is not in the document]
lynnagara: Ok. Random suggestion but how about calling it `profiles` instead of `processed-profiles` since we don't use that prefix elsewhere. Not that we really have any conventions about naming here but in some other cases I believe `ingest-topicname` usually means unprocessed and `topicname` means processed.

snuba/utils/streams/topics.py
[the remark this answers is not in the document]
lynnagara: Yees subscriptions would allow you to run those kinds of aggregate queries over your project data on a regular interval. Not important now, I was just curious, subscription topics are the ones where we need LogAppendTime.

snuba/datasets/profiles_processor.py
```diff
@@ -0,0 +1,45 @@
...
+                "environment": message.get("environment"),
+                "platform": message["platform"],
+                "trace_id": message["trace_id"],
+                "transaction_name": message["transaction_name"],
+                "version_name": message["version_name"],
+                "version_code": message["version_code"],
+                "retention_days": 30,
+            }
+        except KeyError as e:
+            metrics.increment(f"missing_{e.args[0]}")
```
lynnagara, line 43: I think the way you are using this is a bit better suited to a log sent to Sentry than a metric as it may create too many distinct metrics in Datadog.

```suggestion
            logger.warning("Invalid profile data", extra={"missing": e.args[0]})
```

snuba/datasets/profiles_processor.py
```diff
@@ -0,0 +1,38 @@
...
+
+class ProfilesProcessor(MessageProcessor):
+    def process_message(
+        self, message: Mapping[str, Any], metadata: KafkaMessageMetadata
+    ) -> Optional[ProcessedMessage]:
+        try:
+            processed = {
+                "organization_id": message["organization_id"],
+                "project_id": message["project_id"],
+                "transaction_id": message["transaction_id"],
```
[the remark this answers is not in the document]
lynnagara: IIRC the reason we do have this problem in other processors is because Sentry always deals with transaction_ids and other uuids as hexadecimal strings with length of 32 (i.e. without the dashes). So we typically would do something like `str(uuid.UUID(transaction_id))` to add the dashes before inserting to ClickHouse.

snuba/datasets/profiles_processor.py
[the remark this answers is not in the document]
lynnagara: Ok. Should we pass the 30 in through each message anyway like with the other topics, e.g.
`message["retention_days"]`? That way it can just get changed in Sentry later and Snuba will pick up whatever values are passed.

snuba/datasets/storages/profiles.py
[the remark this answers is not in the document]
lynnagara: We do use it to filter subscriptions to the relevant range. Why don't you fill it in in the processor? You can get the values from the second argument to process_message, i.e. `metadata.offset` and `metadata.partition`.

snuba/datasets/storages/profiles.py
```diff
@@ -0,0 +1,92 @@
...
+    UniqInSelectAndHavingProcessor,
+)
+from snuba.utils.streams.topics import Topic
+
+PROFILES_LOCAL_TABLE_NAME = "profiles_local"
+PROFILES_DIST_TABLE_NAME = "profiles_dist"
+
+
+processors = [
+    UniqInSelectAndHavingProcessor(),
```
lynnagara, line 25: This processor was a workaround for a ClickHouse 20.3 bug and will be deprecated. I think you probably don't need it here.

snuba/settings/__init__.py
```diff
@@ -155,7 +155,7 @@
 COLUMN_SPLIT_MAX_RESULTS = 5000
 
 # Migrations in skipped groups will not be run
-SKIPPED_MIGRATION_GROUPS: Set[str] = {"querylog", "spans_experimental", "profiles"}
+SKIPPED_MIGRATION_GROUPS: Set[str] = {"querylog", "spans_experimental"}
```
lynnagara, line 158: Are you sure we are ready to remove from this list already? Remember this will go out to all self hosted users the moment we do this.

phacops: I guess I'm confused. Why do we have `OPTIONAL_GROUPS` and `SKIPPED_MIGRATION_GROUPS`? I want to deploy that in snuba production but not anywhere else.

snuba/utils/streams/topics.py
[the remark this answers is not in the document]
lynnagara: Ah, I assumed we were going to call that earlier one `ingest-profiles`. In that case this seems fine to me.

snuba/settings/__init__.py
lynnagara: Sorry it is a bit confusing. `OPTIONAL_GROUPS` is hardcoded in Snuba's codebase and cannot be changed. Anything in this file (incl. SKIPPED_MIGRATION_GROUPS) can be overridden by the user when they are running snuba. So the groups that are allowed to be optional are hardcoded, but the ones that are actually skipped can be reconfigured by the user. I think you should still skip `profiles` by default here. But you can actually override this value in `ops` if you want to roll it out to Snuba production.

snuba/datasets/factory.py
```diff
@@ -8,7 +8,7 @@
 DATASETS_IMPL: MutableMapping[str, Dataset] = {}
 DATASETS_NAME_LOOKUP: MutableMapping[Dataset, str] = {}
 
-DEV_DATASET_NAMES: Set[str] = set()
+DEV_DATASET_NAMES: Set[str] = set("profiles")
```
lynnagara, line 11: Should be `{"profiles"}` or `set(["profiles"])`

fpacifici (changes requested):
comments inline

snuba/datasets/factory.py
phacops: Indeed, what I wrote ends up a little... weird.

snuba/settings/__init__.py
phacops: Alright, sounds good.

snuba/datasets/profiles.py
```diff
@@ -0,0 +1,7 @@
+from snuba.datasets.dataset import Dataset
+from snuba.datasets.entities import EntityKey
+
+
+class ProfilesDataset(Dataset):
```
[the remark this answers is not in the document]
phacops: I think I'll go the B route.

snuba/settings/__init__.py
lynnagara: Are you planning to put profiles back here? I think it should stay in the default settings until it is ready for roll out in every environment.

lynnagara (commented):
Just one question about the skipped migration group list.

phacops:
Oh right, I need that skipped.

snuba/settings/__init__.py
phacops: Oh right, I need that skipped.

lynnagara (approved):
Thanks

fpacifici (changes requested):
Hold on, this cannot be merged until the `DISABLED_DATASET` change is not applied in production. If this is merged and then deployed, snuba will refuse to start because the health check will not pass

snuba/datasets/profiles_processor.py
```diff
@@ -0,0 +1,63 @@
...
+                "partition": metadata.partition,
+            }
+        except ValueError:
+            logger.warning(
+                "Invalid UUID",
+                extra={
+                    "transaction_id": message["transaction_id"],
+                    "trace_id": message["trace_id"],
+                },
+            )
```
fpacifici, line 59: This is still an issue. We cannot log in consumers unless we have the guarantee that the issue will happen very rarely.
https://github.com/getsentry/snuba/pull/2435#discussion_r817295707

phacops: Not anymore.

snuba/datasets/storages/profiles.py
```diff
@@ -0,0 +1,88 @@
...
+    dist_table_name=PROFILES_DIST_TABLE_NAME,
+    storage_set_key=StorageSetKey.PROFILES,
+)
+
+readable_storage = ReadableTableStorage(
+    storage_key=StorageKey.PROFILES,
+    storage_set_key=StorageSetKey.PROFILES,
+    schema=readable_schema,
+    query_processors=processors,
+)
```
fpacifici, line 68: You do not need these. 
You have one table only, which is read/write. 
`writable_storage` will do for both.
Actually you have to use only one. Otherwise the storage key is assigned to two storages.

fpacifici (approved):
Good for me. 
Adding a new dataset is a bit of a critical operation. 
Please merge and deploy during business hours with the assistance of a search and storage engineer.

snuba/datasets/entities/profiles.py
```diff
@@ -0,0 +1,72 @@
...
+)
+
+
+class ProfilesEntity(Entity, ABC):
+    def __init__(self,) -> None:
+        writable_storage = get_writable_storage(StorageKey.PROFILES)
+        readable_storage = get_storage(StorageKey.PROFILES)
+
+        super().__init__(
+            storages=[writable_storage, readable_storage],
```
fpacifici, line 54: You just need one storage here. No need for a readable one and a writable one.

PR #2436, opened 2022-02-23
ref(querylog) Add the Clickhouse error code to the querylog stats

evanh:
The stats of each query in the query list will now have an "error_code" field if
the query failed in Clickhouse.

### DIFF
--- a/tests/datasets/test_querylog_processor.py
+++ b/tests/datasets/test_querylog_processor.py
@@ -62,7 +62,7 @@ def test_simple() -> None:
                 sql_anonymized="select event_id from sentry_dist sample 0.1 prewhere project_id in ($I) limit 50, 100",
                 start_timestamp=datetime.utcnow() - timedelta(days=3),
                 end_timestamp=datetime.utcnow(),
-                stats={"sample": 10},
+                stats={"sample": 10, "error_code": 386},
                 status=QueryStatus.SUCCESS,
                 profile=ClickhouseQueryProfile(
                     time_range=10,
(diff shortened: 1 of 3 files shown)


tests/test_snql_api.py
```diff
@@ -309,6 +309,29 @@ def test_record_queries(self, record_query_mock: Any) -> None:
...
+            "/events/snql",
+            data=json.dumps(
+                {
+                    "query": f"""MATCH (events)
+                        SELECT event_id, title, transaction, tags[a], tags[b], message, project_id
+                        WHERE timestamp >= toDateTime('{self.base_time.isoformat()}')
+                        AND timestamp < toDateTime('{self.next_time.isoformat()}')
+                        AND project_id IN tuple({self.project_id})
+                        AND tags[a] = 1
+                        LIMIT 5""",
```
volokluev, line 326: out of curiosity, why does this cause a 386?

tests/datasets/test_querylog_processor.py
```diff
@@ -245,3 +245,111 @@ def test_missing_fields() -> None:
             ],
             None,
         )
+
+
+def test_error_code() -> None:
```
volokluev, line 250: this test seems larger than the change, is it testing for more than just the error code?

lynnagara (approved):

tests/test_snql_api.py
evanh: Because we are comparing a tag to an int. I couldn't think of a case that would consistently cause a Clickhouse error, so if/when we change this behaviour we'll have to change/remove the test.

tests/datasets/test_querylog_processor.py
evanh: No, I copied the `simple` test and added the error code.

volokluev: I see, can this be refactored to introduce less code duplication? Asking because from this test it is not clear that what's being tested is the error code. Either the scope of the test should change or the simple test part should be refactored to be reusable

tests/test_snql_api.py
evanh: I rewrote the test to not depend on this behaviour. Should be better long term.

tests/datasets/test_querylog_processor.py
evanh: I collapsed it into the simple test.

volokluev (approved):

PR #2437, opened 2022-02-23
feat(dead-letter-queue): Add dead letter queue for failed inserts

MeredithAnya:
**context**
In the move to upgrade clickhouse, we introduced the option to allow for failures during writes in https://github.com/getsentry/snuba/pull/2395. This was done so that we didn't cause incidents for the main errors and transactions storages if there are issues writing to the new errors_v2 and transactions_v2 storages.

However in order to make the full move over to the new cluster we need to make sure we have consistency between the old and new one and that won't be the case if we drop batches that failed to write. This is why we are adding the dead letter step for failed inserts to the `MultistorageConsumerProcessingStrategyFactory`. 

**failed insert flow**
The `MultistorageCollector` step basically routes messages to be batched up by `storage_key`, and during this step is where we catch a bubbled up exception for if a batch for a specific `storage_key` failed. This PR adds `self.__messages` to the collector step so that in the event of failure, we have a record of which messages were _supposed_ to be written. Now that we have those messages we can do the following:

1. For whichever `storage_key` batch failed, we submit the messages individually to the `DeadLetterStep`
2. The `DeadLetterStep` produces each row[^1] to the `snuba-dead-letter-inserts` topic with the key being the value `storage_key`
3. From here we should be able to create a consumer that just consumes the messages, and has a `CollectStep` almost identical to the collect step already in the `MultistorageConsumerProcessingStrategyFactory`, just this time without catching to the dead letter queue - _note: this isn't added yet but shouldn't be to complicated_

[^1]: The payload we expect in the `submit` method for the `DeadLetterStep` is a tuple of the `storage_key` and [`BytesInsertBatch`](https://github.com/getsentry/snuba/blob/master/snuba/consumers/consumer.py#L57). The `rows` attribute is a sequence of bytes, but seems like most times `rows` is actually only of length 1, which means the numbers of messages should roughly be the number of rows but since it's not guaranteed we can't assume that


snuba/consumers/consumer.py
```diff
@@ -355,6 +357,29 @@ def build_writer() -> ProcessedMessageBatchWriter:
...
+        self.__producer.produce(self.__topic, message)
+
+    def close(self) -> None:
+        pass
+
+    def terminate(self) -> None:
+        pass
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        pass
```
fpacifici, line 380: you'll have to ensure everything gets flushed as you did for the indexer

snuba/consumers/consumer.py
```diff
@@ -384,15 +416,17 @@ def submit(
...
+            storage_message = Message(
+                message.partition,
+                message.offset,
+                payload,
+                message.timestamp,
+                message.next_offset,
             )
+            self.__messages[storage_key] = self.__messages.get(storage_key, []) + [
+                storage_message
+            ]
```
fpacifici, line 428: What is this for ?

snuba/consumers/consumer.py
```diff
@@ -402,6 +436,9 @@ def close(self) -> None:
                 try:
                     step.close()
                 except Exception as e:
+                    if self.__dead_letter_step:
+                        # TODO: self.__dead_letter_step.submit
+                        pass
```
fpacifici, line 441: Are you sure that the exception should be caught here and not in poll?
@nikhars  ?

fpacifici (commented):
Please ensure that, when you use the dead letter topic, each message there is a message of the batch you tried to write on clickhouse. 
You will likely won't be able to fit the entire batch in one single kafka message for the dead letter queue

snuba/cli/multistorage_consumer.py
```diff
@@ -213,6 +218,15 @@ def multistorage_consumer(
             consumer_configuration, producer=producer, commit_log_topic=commit_log,
         )
 
+    dead_letter_producer: Optional[KafkaProducer] = None
+    dead_letter_queue: Optional[Topic] = None
+    if dead_letter_topic:
+        dead_letter_queue = Topic(dead_letter_topic)
+
+        dead_letter_producer = KafkaProducer(
+            build_kafka_producer_configuration(StreamsTopic(dead_letter_topic))
```
MeredithAnya, line 227: this didn't like the arroyo `Topic`, probs need to fix that

snuba/consumers/consumer.py
MeredithAnya: for now I don't have the messages in the `MultistorageCollector` step, so I wasn't sure how to forward them to the next step without collecting them in this one, I could move the produce step further down to like the `InsertBatchWriter` instead

snuba/consumers/consumer.py
nikhars: This is the right place to catch the exception

nikhars (commented):
The messages on the dead letter queue would be of the type (storage_key, batch_of_messages) right? Exceptions can be raised from both errors_v2 and transactions_v2. So we would need to know where to do the insert for a specific batch of messages

snuba/cli/multistorage_consumer.py
lynnagara: What does this mean? It's not supposed to be the arroyo topic. It's the topic for fetching the configuration from settings in Snuba.

snuba/consumers/consumer.py
MeredithAnya: @fpacifici @nikhars does this approach make sense to do? In terms of being able to capture the messages that were attempted to be batch inserted

nikhars: I guess it should be ok. I don't know what you intend to put in the dead letter topic. If it's InsertBatch or KafkaMessage. But you can decide that in the DeadLetterStep.

Ideally we should get payload in the dead letter topic in a form so that its easy to apply like an actual INSERT statement. @fpacifici might have opinion on it.

MeredithAnya: I assumed we could have each message that goes into the dead letter topic be a `KafkaPayload`, where maybe the key can be the `storage_key` and the payload could be the rows from the `BytesInsertBatch`. (maybe the payload also has to have the `origin_timestamp` too) But basically I was thinking you could have a consumer strategy that was basically only a `CollectStep`, and you'd pass in something quite similar to the `MultistorageCollector` as the collector.

snuba/cli/multistorage_consumer.py
MeredithAnya: I guess it was just confusing to have two separate topic objects

snuba/consumers/consumer.py
```diff
@@ -355,6 +360,66 @@ def build_writer() -> ProcessedMessageBatchWriter:
...
+
+    def __format_payload(
+        self,
+        message: Message[
+            Tuple[StorageKey, Union[None, BytesInsertBatch, ReplacementBatch]]
+        ],
+    ) -> Optional[KafkaPayload]:
+        kafka_payload = None
+        storage_key, payload = message.payload
+        if isinstance(payload, BytesInsertBatch):
```
MeredithAnya, line 380: We won't be handling `ReplacementBatch`'s right? since those go to the replacements topic yeah?

snuba/consumers/consumer.py
```diff
@@ -425,6 +521,9 @@ def join(self, timeout: Optional[float] = None) -> None:
 
             step.join(timeout_remaining)
 
+        if self.__dead_letter_step:
+            self.__dead_letter_step.join(5.0)
```
MeredithAnya, line 525: I don't know how much time we wanna leave for `DeadLetterStep.join()` but I'm assuming it should be a rare occasion to have messages in the dead letter topic

snuba/consumers/consumer.py
```diff
@@ -355,6 +360,66 @@ def build_writer() -> ProcessedMessageBatchWriter:
...
+        pass
+
+    def submit(
+        self,
+        message: Message[
+            Tuple[StorageKey, Union[None, BytesInsertBatch, ReplacementBatch]]
+        ],
+    ) -> None:
+        payload = self.__format_payload(message)
+        if payload:
```
MeredithAnya, line 402: Does anyone know when this would actually be `None`?

snuba/consumers/consumer.py
fpacifici: > The flow to me (and correct me if wrong):
collector collects up messages to be batch written -> batch fails -> messages unbatched and produced to dead letter topic -> new consumer collector collects dead letter topic messages to be batch written again -> hopefully batch succeeds

Works for me

snuba/consumers/consumer.py
fpacifici: yep ignore replacements

fpacifici:
@nikhars 

> The messages on the dead letter queue would be of the type (storage_key, batch_of_messages) right?

No, we cannot send an entire batch as a single message. That would likely be beyond the max message size. (transactions batches are big). 
Each row in the batch has to be its own message.

snuba/cli/multistorage_consumer.py
```diff
@@ -213,6 +218,15 @@ def multistorage_consumer(
             consumer_configuration, producer=producer, commit_log_topic=commit_log,
         )
 
+    dead_letter_producer: Optional[KafkaProducer] = None
+    dead_letter_queue: Optional[Topic] = None
+    if dead_letter_topic:
+        dead_letter_queue = Topic(dead_letter_topic)
+
+        dead_letter_producer = KafkaProducer(
```
fpacifici, line 226: Who is closing the KafkaProducer? All producers in Arraoyo need to be closed (method `close`)  at the end of the script, otherwise the producer will hang during shutdown

snuba/consumers/consumer.py
```diff
@@ -355,6 +360,66 @@ def build_writer() -> ProcessedMessageBatchWriter:
...
+    ) -> Optional[KafkaPayload]:
+        kafka_payload = None
+        storage_key, payload = message.payload
+        if isinstance(payload, BytesInsertBatch):
+            rows = [rapidjson.loads(row) for row in payload.rows]
+
+            kafka_payload = KafkaPayload(
+                storage_key.value.encode("utf-8"),
+                rapidjson.dumps(
+                    {"rows": rows, "origin_timestamp": payload.origin_timestamp}
```
fpacifici, line 386: We need to produce this row by row (one message per row). The entire batch will be too big for a kafka message.

snuba/consumers/consumer.py
```diff
@@ -355,6 +360,66 @@ def build_writer() -> ProcessedMessageBatchWriter:
...
+    def __format_payload(
+        self,
+        message: Message[
+            Tuple[StorageKey, Union[None, BytesInsertBatch, ReplacementBatch]]
+        ],
+    ) -> Optional[KafkaPayload]:
+        kafka_payload = None
+        storage_key, payload = message.payload
+        if isinstance(payload, BytesInsertBatch):
+            rows = [rapidjson.loads(row) for row in payload.rows]
```
fpacifici, line 381: You should not need to parse the JSON.
Just take binary data from  `payload.rows` and produce them as they are. No parsing.

snuba/consumers/consumer.py
```diff
@@ -402,6 +492,9 @@ def close(self) -> None:
                 try:
                     step.close()
```
fpacifici, line 493: You do not clean up `self.__messages` because this colelctor is created and destroyed at every batch, is that correct ?

fpacifici (commented):
A few remarks inline

snuba/consumers/consumer.py
MeredithAnya: I _think_ so... but I could also reset `self.__messages` in the join method

snuba/cli/multistorage_consumer.py
MeredithAnya: added here https://github.com/getsentry/snuba/pull/2437/files#diff-fdeca11abbdfdf36e4aa9604b39c17a937f1c35368351d1f20740a129c7b6478R258

snuba/consumers/consumer.py
```diff
@@ -355,6 +360,77 @@ def build_writer() -> ProcessedMessageBatchWriter:
...
+
+    def terminate(self) -> None:
+        self.__closed = True
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        start = time.time()
+
+        while self.__futures:
+            if not timeout or timeout > time.time() - start:
+                break
```
nikhars, line 423: An edge case which can happen here is that the producer is not done writing the failed insert message to the dead letter topic before the timeout is complete. In those cases we will lose data since the consumer will commit the kafka offset. The question then is, whether we can live with that or should we prevent committing the kafka offset?

snuba/consumers/consumer.py
nikhars: Good that there's a timeout passed to the step

snuba/cli/devserver.py
```diff
@@ -153,6 +153,7 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--auto-offset-reset=latest",
                     "--log-level=debug",
                     "--consumer-group=metrics_group",
+                    "--dead-letter-topic=snuba-dead-letter-inserts",
```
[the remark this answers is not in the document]
MeredithAnya: oop yes sorry from my testing - will remove

snuba/consumers/consumer.py
MeredithAnya: Hmm so @nikhars are you suggesting we raise here if we timeout? there is also the case where soemthing failed when producing (I just log a warning below for now). We could log a warning for timeouts as well for now first? Not sure at what point we want to stop ingestion

nikhars: Can we add metrics around how many times we are failing to produce. Either because of timeouts or problems in futures. 

I agree we can not do anything more than that for now.

tests/test_consumer.py
```diff
@@ -204,6 +212,84 @@ def test_multistorage_strategy(
...
+    ]
+
+    for message in messages:
+        strategy.submit(message)
+
+    strategy.close()
+    strategy.join()
+
+    # assert that we submit to the dead letter step for both messages
+    assert len(mock_submit.call_args_list) == 2
```
fpacifici, line 290: There is a better way to test this. 
Pass a mock producer to the dead letter step and check what is produced there.

fpacifici (approved):
Looks good. 

Please still add a unit test for the `DeadLetterStep` in isolation. 
The integration test you added is good, but it covers a lot of code and not a lot of corner cases.
So the coverage of DeadLetterStep is not great, moreover, when that test fails, it takes long to troubleshoot. 
Adding a unit test as well instead, covers DeadLetterStep better and it makes it easier to understand what is broken when the test fails.

PR #2441, opened 2022-02-23
metrics(replacer): Consumer group per replacement processed

rahul-kumar-saini:
- We're adding a new consumer group for replacements
- We'd like to see replacements metrics split by consumer groups and update some alerts

### DIFF
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -485,7 +485,11 @@ class ErrorsReplacer(ReplacerProcessor[Replacement]):
         type_ = message.action_type
 
         if type_ in REPLACEMENT_EVENT_TYPES:
-            metrics.increment("process", 1, tags={"type": type_})
+            metrics.increment(
+                "process",
+                1,
+                tags={"type": type_, "consumer_group": message.metadata.consumer_group},
+            )
 
         if type_ in (
             ReplacementType.START_DELETE_GROUPS,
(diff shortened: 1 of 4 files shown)


snuba/datasets/errors_replacer.py
```diff
@@ -485,7 +485,11 @@ def process_message(self, message: ReplacementMessage) -> Optional[Replacement]:
         type_ = message.action_type
 
         if type_ in REPLACEMENT_EVENT_TYPES:
-            metrics.increment("process", 1, tags={"type": type_})
+            metrics.increment(
```
nikhars, line 488: Is this the only metric which is relevant per consumer group?

nikhars (changes requested):
Can you add the consumer group as a tag over here https://github.com/getsentry/snuba/blob/master/snuba/cli/replacer.py#L102. That way any metrics reported by the ReplacerWorker will have the relevant tags

rahul-kumar-saini:
> Can you add the consumer group as a tag over here https://github.com/getsentry/snuba/blob/master/snuba/cli/replacer.py#L102.

I think this already has the consumer group:
https://github.com/getsentry/snuba/blob/master/snuba/cli/replacer.py#L93

snuba/datasets/errors_replacer.py
rahul-kumar-saini: I believe so. The other metrics already have consumer group data.

nikhars (approved):

PR #2443, opened 2022-02-24
feat: Uniq function killswitch

lynnagara:
Adds a killswitch for the uniq() function, only to be used in cases of emergency.
It can be turned on by table or by referrer (or a combination of both).
Since it returns incorrect results for uniq(), it should not be turned on under
normal circumstances.

### DIFF
--- a/snuba/query/processors/uniq_killswitch.py
+++ b/snuba/query/processors/uniq_killswitch.py
@@ -9,8 +9,22 @@ from snuba.request.request_settings import RequestSettings
 
 class UniqKillswitchProcessor(QueryProcessor):
     """
+    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+    USE ONLY IN CASE OF EMERGENCY
+    THIS PROCESSOR BEING TURNED ON RETURNS INCORRECT
+    RESULTS.
+    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+    Usage:
+        In Runtime config, set the following flags:
+
+        uniq_killswitch_tables=errors_dist,discover_dist,transactions_dist,(etc)
+        uniq_killswitch_referrers=search,tsdb-modelid:4,(etc)
+
     This processor removes uniq() expressions from the query and replaces them with zero.
-    The reason for doing this is because uniq() queries are suspected to be causing ClickHouse to segfault.
+    The reason for doing this is because uniq() queries are suspected to be
+    causing ClickHouse to segfault under very specific consequences we do
+    not understand.
     Since it returns incorrect results, it is only to be used in case of emergency.
 
     Example usage:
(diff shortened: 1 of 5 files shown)


tests/query/processors/test_uniq_killswitch.py
```diff
@@ -0,0 +1,65 @@
...
+                    "alias2",
+                    FunctionCall("alias2", "count", (Column(None, None, "column2"),)),
+                ),
+            ],
+        ),
+    ),
+]
+
+
+@pytest.mark.parametrize("input_query, expected_query", test_data)
```
volokluev, line 43: nit: calling this `expected_query` and then having clauses where it's not expected is confusing. maybe call it `killswitched_query` or something

volokluev (commented):
Can we get an api test where we see the killswitch actually trigger? If this is for emergencies, we don't want to find out that it doesn't work when we need it.

snuba/query/processors/uniq_killswitch.py
```diff
@@ -0,0 +1,48 @@
...
+from snuba.clickhouse.processors import QueryProcessor
+from snuba.clickhouse.query import Query
+from snuba.query.expressions import Expression, FunctionCall, Literal
+from snuba.request.request_settings import RequestSettings
+
+
+class UniqKillswitchProcessor(QueryProcessor):
+    """
+    This processor removes uniq() expressions from the query and replaces them with zero.
+    The reason for doing this is because uniq() queries are suspected to be causing ClickHouse to segfault.
```
fpacifici, line 13: "some uniq() queries under some specific circumstances"

snuba/query/processors/uniq_killswitch.py
```diff
@@ -0,0 +1,48 @@
...
+    - Enable for referrer
+        - uniq_killswitch_referrers=referrer1,referrer2,referrer3
+
+    `uniq_killswitch_tables` and `uniq_killswitch_referrers` are independent
+    and uniq will be removed from all tables and referrers specified.
+    """
+
+    def process_query(self, query: Query, request_settings: RequestSettings) -> None:
+        table_config = state.get_config("uniq_killswitch_tables")
+        referrer_config = state.get_config("uniq_killswitch_referrers")
```
fpacifici, line 28: Can we have a column config and a project config as well? I think it is ok to do in a followup.
It is likely that only one column is the root cause. So most likely uniq(group_id) is fine while uniq(user_id) isn't.

snuba/query/processors/uniq_killswitch.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+        table_name = query.get_from_clause().table_name
+
+        should_transform_uniq = (
+            request_settings.referrer in referrers or table_name in table_names
+        )
+
+        def process_functions(exp: Expression) -> Expression:
+            if isinstance(exp, FunctionCall):
+                if exp.function_name == "uniq":
```
evanh, line 42: You might also need to catch `uniqIf`.

snuba/query/processors/uniq_killswitch.py
```diff
@@ -0,0 +1,48 @@
...
+                    return Literal(exp.alias, 0)
```
evanh, line 43: I think we should default to 1, to avoid possible zero division errors.

snuba/query/processors/uniq_killswitch.py
volokluev: ok, I will do that in a separate PR because that increases the complexity

evanh (approved):

PR #2444, opened 2022-02-24
feat: Add a metric to track how long optimize takes

lynnagara:
We don't have any metrics for `snuba optimize`. This change adds a metric
which records the timing for each part that is optimized.

### DIFF
--- a/snuba/optimize.py
+++ b/snuba/optimize.py
@@ -19,6 +19,7 @@ def run_optimize(
     database: str,
     before: Optional[datetime] = None,
 ) -> int:
+    start = time.time()
     schema = storage.get_schema()
     assert isinstance(schema, TableSchema)
     table = schema.get_local_table_name()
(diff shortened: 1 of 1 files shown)


snuba/optimize.py
```diff
@@ -112,4 +117,6 @@ def optimize_partitions(
 
         query = (query_template % args).strip()
         logger.info(f"Optimizing partition: {part.name}")
+        start = time.time()
         clickhouse.execute(query)
+        metrics.timing("optimized_part", time.time() - start, tags={"table": table})
```
MeredithAnya, line 122: is there any value in tagging the partition as well? I'm not sure if that's needed or not

fpacifici: we cannot tag partitions. Their names are dates. Which means we will be increasing the tags over time.
You can tag partition as the number of weeks from today.

fpacifici (approved):
Could you also please add a metric for the entire job ?

evanh (approved):
Tracking how many parts we optimized might also be beneficial.

lynnagara:
I tthink entire job time and # of optimised parts can be fairly easily inferred, but I added another metric for total job time anyway in case people find it helpful. How many parts we optimised is essentially the same as a count over the `optimized_part` metric.

PR #2447, opened 2022-02-24
ref(makefile): Add backend-typing

MeredithAnya:
sentry's makefile uses [`backend-typing`](https://github.com/getsentry/sentry/blob/master/Makefile#L133) so I thought it might be nice to just keep the same naming even though we don't have much of a frontend except for admin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ api-tests:
 	SNUBA_SETTINGS=test pytest -vv tests/*_api.py
 
 backend-typing:
-	mypy snuba --strict --warn-unreachable --config-file mypy.ini --ignore-missing-imports
+	mypy snuba --strict --warn-unreachable --config-file mypy.ini
 
 install-python-dependencies:
 	pip install -e .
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,6 +18,9 @@ tests: test
 api-tests:
 	SNUBA_SETTINGS=test pytest -vv tests/*_api.py
 
+backend-typing:
+	mypy snuba --strict --warn-unreachable --config-file mypy.ini --ignore-missing-imports
```
fpacifici, line 22: I don't think you should add the `ignore-missing-imports`. We already do it package by package in mypy.ini.
So `ignore-missing-imports` is by default False.

mypy.ini
```diff
@@ -7,6 +7,9 @@ ignore_missing_imports = True
 [mypy-clickhouse_driver]
 ignore_missing_imports = True
 
+[mypy-clickhouse_driver.errors]
```
MeredithAnya, line 10: @fpacifici I was getting missing imports errors for this when I ran it without `--ignore-missing-imports` so that's why I thought I needed it.

PR #2448, opened 2022-02-24
feat(pipeline delegator) Split rate limiter namespace

fpacifici:
In order to rollout to rollout a large amount of queries to errors_v2 we need to split the rate limiter name spaces.
If we do not do that we would double count and reject queries that should be accepted.
The resources to rate limit are disjoint (separate clusters) so there is no issue in having a double quota.

The way to implement this is to leverage the PipelineDelegator so that it provides a a separate instance of RequestSettings to each of the two pipelines to be executed.
This allows us to still make each query pipeline independent and unaware of the pipeline delegator. Keeping each pipeline independent reduces the complexity of the design.

Limitations (which are not a major problem here).
- The callback receives the original request settings. We would need a much bigger redesign to provide the proper settings there 
- This could undercount on the project rate limiter when we add a delegator for transactions as the builder_id would be different. I think we can live with this during the rollout.


snuba/pipeline/settings_delegator.py
```diff
@@ -0,0 +1,70 @@
...
+    into otherwise we would increase the pipeline complexity
+    dramatically.
+
+    Still parallel pipelines on separate clusters should not share the
+    same quota and rate limiter.
+
+    The idea is to provide the query pipeline with delegates over the
+    components they need to execute it that create separate namespaces.
+    Specifically the PipelineDelegator provides a RateLimiterDelegate to
+    the secondary query so that the query pipeline can use a separate
```
evanh, line 21: As far as I can see both the primary and secondary query receive the delegate.

evanh (approved):
I'm going to approve this, because it works, but it feels like this is solving the wrong problem.

The  issue is that rate limits are being configured on the entity, which is unaware that the query will ultimately be sent to two different storages. We can't rate limit at the entity level because we want to check caching first.

Why not move all the rate limiting processors onto the storages? Then the storage name could be prepended to the rate limiters to properly namespace them.

snuba/pipeline/settings_delegator.py
fpacifici: Right will fix

fpacifici:
> The issue is that rate limits are being configured on the entity, which is unaware that the query will ultimately be sent to two different storages. We can't rate limit at the entity level because we want to check caching first.
>
> Why not move all the rate limiting processors onto the storages? Then the storage name could be prepended to the rate limiters to properly namespace them.

If we do it this way, cross-storage rate limiters like the project overall rate limiter would not work anymore as they would be contextualized to a storage.
If we instead keep a single project rate limiter (without introducing namespace) we have a snuplicator problem as each query counts twice. 
Now this implementation still split the global project rate limiter into multiple ones across entities (which is undesirable, but not too bad for the sake of the rollout as it has no adverse impact in quota). It should not be the general design though because we will have quotas that are not broken down by storage.

So the problem is a little different I think and broader, requiring a larger discussion:
- How should an experiment where we duplicate your queries impact your quota? We could easily say it should not. This is not a problem in this case as we run queries in two separate storages but what if we were running experiment within the same storage. There we would be impacting the quota we have.
- How should we break the quota between experiment and primary for a rate limiter that is cross-storage? Like the main project one? I would say so but it should not be broken by storage. It should be broken down by whether we are in the experiment group or not.

evanh:
Firstly, I want to say, feel free to merge this PR in the sake of progressing the roll out. I think we both agree this isn't an ideal solution but is fine for now.

> cross-storage rate limiters like the project overall rate limiter

I think we disagree on something here: the project overall rate limiter is _not_ cross-storage in our system now. Your PR proves that. It is incremented once for every storage that the query is divided between. Since it is already not cross storage, my suggestion is to move it into the storages, so it accurately reflects what is happening now anyways.

I think our problem is that we don't actually have any cross-storage rate limiters. Since the rate limit params are passed through to the storage layer before being evaluated, we are not accurately rate limiting across storages.

PR #2453, opened 2022-02-25
fix(prewhere) Exclude columns from the prewhere candidates if they are in a uniq expression

fpacifici:
We identified some queries that can cause a segfault on the clickhosue server if this pattern is present.
```
SELECT
   uniq(something)
PREWHERE something = 'asd'
```
The issue is probably narrower than that but this is the first step to prevent this from happening.
Basically if a column is in a uniq expression anywhere in the query, it will become ineligible for prewhere

### DIFF
--- a/snuba/query/processors/prewhere.py
+++ b/snuba/query/processors/prewhere.py
@@ -10,8 +10,6 @@ from snuba.query.conditions import (
     get_first_level_and_conditions,
 )
 from snuba.query.expressions import FunctionCall
-from snuba.query.matchers import FunctionCall as FunctionPattern
-from snuba.query.matchers import String
 from snuba.request.request_settings import RequestSettings
 from snuba.utils.metrics.wrapper import MetricsWrapper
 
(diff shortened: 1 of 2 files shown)


snuba/query/processors/prewhere.py
```diff
@@ -46,13 +51,29 @@ def __init__(
...
+        for exp in expressions:
+            match = self.__uniq_matcher.match(exp)
+            if match is not None:
+                columns = get_columns_in_expression(exp)
+                for c in columns:
+                    uniq_cols.add(c.column_name)
+
+        for col in uniq_cols:
+            if col in prewhere_keys:
+                metrics.increment("uniq_col_in_prewhere_candidate")
```
fpacifici, line 73: I suspect this will be logged a lot because `prewhere_keys` are jsut the candidate columns, not necessarily the columns that will be chosen.

lynnagara: Would be good to tag the col so we can decide what to do about those queries

lynnagara (approved):

snuba/query/processors/prewhere.py
fpacifici: ok

PR #2454, opened 2022-02-25
Add read path for upgraded transactions

volokluev:
Basically copy-paste of https://github.com/getsentry/snuba/pull/2425 but for transactions.

TRANSACTIONS_UPGRADE_BEGINNING_OF_TIME was determined by looking at the first insert query in datadog.

Tests are copy and pasted from the errors one. If this was less urgent I'd be more clever and reuse that code but since we only need to do this for three entities I think it's ok.


snuba/datasets/entities/transactions.py
```diff
@@ -111,12 +121,49 @@ def select_storage(
...
+
+
+def v2_selector_function(query: Query, referrer: str) -> Tuple[str, List[str]]:
+    if settings.ERRORS_UPGRADE_BEGINING_OF_TIME is None or not isinstance(
+        query, ProcessableQuery
+    ):
+        return ("transactions_v1", [])
+
+    range = get_time_range(query, "timestamp")
+    if range[0] is None or range[0] < settings.ERRORS_UPGRADE_BEGINING_OF_TIME:
```
evanh, line 145:
```suggestion
    if range[0] is None or range[0] < settings.TRANSACTIONS_UPGRADE_BEGINING_OF_TIME:
```

snuba/datasets/entities/transactions.py
```diff
@@ -111,12 +121,49 @@ def select_storage(
...
+    def __init__(self, mappers: TranslationMappers) -> None:
+        self.__transactions_table = get_writable_storage(StorageKey.TRANSACTIONS_V2)
+        # TODO: Register TRANSACTIONS_V2_RO and then remove comment
+        # self.__transactions_ro_table = get_storage(StorageKey.TRANSACTIONS_V2)
+        self.__mappers = mappers
+
+    def select_storage(
+        self, query: Query, request_settings: RequestSettings
+    ) -> StorageAndMappers:
+        # TODO: Register ERRORS_V2_RO and support the ro storage
```
fpacifici, line 134: This is different for transactions, you do not need a selected a storage for transactions_v2 
just use `SingleStorageQueryPlanBuilder` at line 177. There is only one storage, no selection to be done.

snuba/datasets/entities/transactions.py
```diff
@@ -111,12 +121,49 @@ def select_storage(
...
+        return StorageAndMappers(self.__transactions_table, self.__mappers)
+
+
+def v2_selector_function(query: Query, referrer: str) -> Tuple[str, List[str]]:
+    if settings.ERRORS_UPGRADE_BEGINING_OF_TIME is None or not isinstance(
```
fpacifici, line 139: This references the wrong setting

snuba/datasets/entities/transactions.py
fpacifici, line 145: references the wrong setting. it should be the transactions one

fpacifici (changes requested):

snuba/datasets/entities/transactions.py
volokluev: got it

snuba/datasets/entities/transactions.py
```diff
@@ -111,21 +124,61 @@ def select_storage(
...
-                    if custom_mappers is None
-                    else transaction_translator.concat(custom_mappers)
-                )
+                selector=TransactionsQueryStorageSelector(mappers=mappers)
             ),
         )
 
+        v2_pipeline_builder = SimplePipelineBuilder(
+            query_plan_builder=SingleStorageQueryPlanBuilder(
+                storage=get_writable_storage(StorageKey.TRANSACTIONS_V2),
```
fpacifici, line 168: You just need a readable storage, `get_writable_storage` return a writable one. You can go with `get_storage`

fpacifici (approved):

PR #2456, opened 2022-02-25
fix(prewhere) remove from prewhere columns that appear in countIf as well

fpacifici:
Like 
https://github.com/getsentry/snuba/pull/2453

But countIf has the same problem


evanh (approved):
Can you add a comment here that explains the reasoning a bit and also clarifies that this can be removed once we are on 21.8?

volokluev (approved):

PR #2457, opened 2022-02-25
ref(dead-letter-queue): add dead letter consumer

MeredithAnya:
consumer to be able to consumer messages from the dead letter queue that was added in https://github.com/getsentry/snuba/pull/2437


snuba/cli/dead_letter_consumer.py
```diff
@@ -0,0 +1,124 @@
...
+@click.option(
+    "--max-batch-time-ms",
+    default=settings.DEFAULT_MAX_BATCH_TIME_MS,
+    type=int,
+    help="Max length of time to buffer messages in memory before writing to Kafka.",
+)
+@click.option(
+    "--auto-offset-reset",
+    default="error",
+    type=click.Choice(["error", "earliest", "latest"]),
```
nikhars, line 51: When you deploy this in prod for the first time, you can pass `latest` option.

nikhars (commented):
Some questions:
1. Lets assume a failure happens with a multistorage consumer. And its writing to 3 storages. 1 succeeds and 2 fail. Would the message on the topics contain messages only for the 2 failed ones?
2. Are you considering to roll this consumer out only for the new storages which are on tiger clusters or for all existing writable storages?
3. What happens in case there is a failure in the dead letter consumer while writing to a storage? Should it just dump the failure to another topic and move along? Or should it completely stop and someone has to come in and manually intervene?

snuba/cli/dead_letter_consumer.py
```diff
@@ -0,0 +1,124 @@
...
+@click.command(hidden=True)
+@click.option(
+    "--storage",
+    "storage_names",
+    type=click.Choice([storage_key.value for storage_key in WRITABLE_STORAGES.keys()]),
+    multiple=True,
+    required=True,
+)
+@click.option(
+    "--consumer-group", default="snuba-consumers",
```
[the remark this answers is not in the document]
MeredithAnya: oop yeah sorry left over from copy/pasta

snuba/cli/dead_letter_consumer.py
```diff
@@ -0,0 +1,124 @@
...
+from snuba.utils.streams.topics import Topic as StreamsTopic
+
+logger = logging.getLogger(__name__)
+
+
+@click.command(hidden=True)
+@click.option(
+    "--storage",
+    "storage_names",
+    type=click.Choice([storage_key.value for storage_key in WRITABLE_STORAGES.keys()]),
```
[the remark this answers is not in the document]
MeredithAnya: yeah was only gonna have it be `errors_v2, transactions_v2` I could hard code those choices instead of using the `WRITABLE_STORAGES` if that's better

MeredithAnya:
> Some questions:
> 
> 1. Lets assume a failure happens with a multistorage consumer. And its writing to 3 storages. 1 succeeds and 2 fail. Would the message on the topics contain messages only for the 2 failed ones?

Yes so the messages are collected on a per storage basis so only messages for the storages that failed would be written into the dead letter topic

> 2. Are you considering to roll this consumer out only for the new storages which are on tiger clusters or for all existing writable storages?

I think my assumption was that this was to help with the failures we may be seeing in the new clusters. I'm not quite sure if we want to use this beyond that.

> 3. What happens in case there is a failure in the dead letter consumer while writing to a storage? Should it just dump the failure to another topic and move along? Or should it completely stop and someone has to come in and manually intervene?

I'm hoping that the dead letter queue won't be used that often in general and therefore also hopefully it doesn't fail often either. I think that if it's failing we should probably figure out if there is a bigger issue why the data failed to written. My assumption is that if we have to stop the dead letter consumer for a little it wouldn't be too big a deal since there shouldn't be tons and tons of data coming in anyway... right?

snuba/cli/multistorage_consumer.py
```diff
@@ -79,6 +80,9 @@
     "--output-block-size", type=int,
 )
 @click.option("--log-level")
+@click.option(
+    "--dead-letter-topic", help="Dead letter topic to send failed insert messages."
+)
```
nikhars, line 85: We should add these options to the regular consumer as well. Not only the multi storage consumer. That makes switching from single consumer to multi storage consumer much less likely to fail.

For example, the consumer which fills the profiles dataset would be a single storage consumer but could use this functionality when it first starts off in production.

snuba/cli/dead_letter_consumer.py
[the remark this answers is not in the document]
nikhars: Thinking about this a bit more, I feel we should keep it flexible that any storage can be selected. In prod, we will choose only errors_v2 and transactions_v2.

snuba/consumers/consumer.py
```diff
@@ -402,6 +507,13 @@ def close(self) -> None:
                 try:
                     step.close()
                 except Exception as e:
+                    messages = self.__messages[storage_key]
+                    if self.__dead_letter_step and messages:
```
nikhars, line 511: Good idea to track a metric of how many messages have been written to dead letter topic per storage.

snuba/consumers/consumer.py
```diff
@@ -522,6 +642,8 @@ def __init__(
         input_block_size: Optional[int],
         output_block_size: Optional[int],
         metrics: MetricsBackend,
+        producer: Optional[AbstractProducer[KafkaPayload]],
+        topic: Optional[Topic],
```
nikhars, line 646: Maybe prefix these with `dead_letter` or something to make it explicit. Its possible that in the future we want to produce something on another topic which is not related to dead letter.

snuba/consumers/consumer.py
```diff
@@ -355,6 +360,77 @@ def build_writer() -> ProcessedMessageBatchWriter:
     return build_writer
 
 
+class DeadLetterStep(
```
nikhars, line 363: Can you add some tests for this class?

snuba/cli/dead_letter_consumer.py
```diff
@@ -0,0 +1,124 @@
...
+)
+@click.option(
+    "--queued-min-messages",
+    default=settings.DEFAULT_QUEUED_MIN_MESSAGES,
+    type=int,
+    help="Minimum number of messages per topic+partition librdkafka tries to maintain in the local consumer queue.",
+)
+@click.option(
+    "--dead-letter-topic", help="Dead letter topic to consume failed insert messages."
+)
```
fpacifici, line 68: For a separate PR, 
Since we are going to keep this around, topic names should not be passed as CLI arguments, they are registered in Snuba and provided by the storage via `KafkaStreamLoader` 
https://github.com/getsentry/snuba/blob/11ca1fd9cdb66657be3b82fda85d410a721b9ee9/snuba/datasets/table_storage.py#L104-L120

So using the CLI was ok during the emergency but we will have to express this in the same way as we do for other topics. Which means we need to think over what to do about multiple storages that use the same topic. Do we make one consumer take care of multiple storages or will we have one consumer per storage ?

snuba/cli/devserver.py
```diff
@@ -153,6 +153,7 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--auto-offset-reset=latest",
                     "--log-level=debug",
                     "--consumer-group=metrics_group",
+                    "--dead-letter-topic=snuba-dead-letter-inserts",
```
fpacifici, line 156: Why this ? we are failing the entire batch if there is a failure on the metrics consumer. We should not need a dead letter queue (at least not this dead letter queue) there .

snuba/cli/multistorage_consumer.py
```diff
@@ -234,11 +248,15 @@ def multistorage_consumer(
...
             metrics=metrics,
+            producer=dead_letter_producer,
+            topic=dead_letter_queue,
         ),
     )
 
     def handler(signum: int, frame: Any) -> None:
         processor.signal_shutdown()
+        if dead_letter_producer:
+            dead_letter_producer.close()
```
fpacifici, line 259: Is this not already in https://github.com/getsentry/snuba/pull/2437 (also in the version that works).
Is this PR based on the right branch ?

snuba/cli/multistorage_consumer.py
fpacifici: Why though? The single storage consumer never drops data. If the write fails it does not commit and crashes, so it will retry. This dead letter queue implementation is only useful for multi storage consumer for the secondary storages if we do not crash the consumer upon error

fpacifici (changes requested):

snuba/cli/multistorage_consumer.py
nikhars: I was thinking about the manner in which we introduce new datasets and their consumers in production. When they are introduced, we cannot foresee if and when they could fail. In those cases, should a single instance of write failure cause the consumer to crash and cause backlog of the system? Or should it keep moving along and let dead letter consumer deal with retrying the failures.

The write failures we were seeing while writing to the new errors_v2 cluster were a storage problem. Not a consumer related problem. 

We can go with only multi-storage consumers for now since it increases complexity of the system for little gain for now.

MeredithAnya:
This is pretty stale and we don't actually use the DLQ step that was added in https://github.com/getsentry/snuba/pull/2437, so maybe will revisit this later if we intend to use it

PR #2459, opened 2022-02-25
ref: Change QueryException to show inner error

evanh:
If a QueryException was `str`, it would json encode the extra_data and no
context on what the inner error was. This isn't a big deal locally, but when
looking at these errors in Sentry, the exception is buried in the inner errors
and makes it harder to diagnose and understand errors. I also think it is
causing Sentry to group errors that it shouldn't.

Add the inner exception message to the `str` of the QueryException so it is
easily readable in Sentry and makes debugging easier.

### DIFF
--- a/snuba/web/__init__.py
+++ b/snuba/web/__init__.py
@@ -5,7 +5,6 @@ from typing import Any, Mapping, NamedTuple
 from mypy_extensions import TypedDict
 
 from snuba.reader import Column, Result, Row, transform_rows
-from snuba.utils.serializable_exception import SerializableException
 
 
 class QueryExtraData(TypedDict):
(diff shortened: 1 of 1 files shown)


volokluev:
> Also removed the SerializableException since QueryException is never
serialized

Yes it is, it can be thrown during query execution at which point it has to be serialized and put into redis

evanh:
> Yes it is, it can be thrown during query execution at which point it has to be serialized and put into redis

Where? I couldn't figure out where that might happen. As far as I can tell, the QueryException is created here: https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L523

At this point the caching has already happened I thought.

evanh:
We had some offline discussions about this, which ended up at: the QueryException is not really an Exception, it's a container that passes information about the query up the exception stack. We shouldn't call it an Exception, and pass that data in a different way.

PR #2461, opened 2022-02-28
test(metrics) - metrics end-to-end validation

onewland:
A few basic end-to-end tests for writing to/reading from the metrics tables

Let me know if you think there are edge cases I should be covering. 

It takes about 10s to run, probably due to the large number of row insertions, but that seems on par with `test_transactions_api` which I modeled this after which takes ~30s to run on my laptop


tests/test_metrics_api.py
```diff
@@ -0,0 +1,362 @@
...
+        if not metric_id:
+            metric_id = self.metric_id
+        if not org_id:
+            org_id = self.org_id
+        if not start_time:
+            start_time = (self.base_time - self.skew).isoformat()
+        if not end_time:
+            end_time = (self.base_time + self.skew).isoformat()
+        query_str = """MATCH (metrics_counters)
+                    SELECT sumMerge(value) AS total_seconds BY project_id, org_id
```
MeredithAnya, line 128: does the match query need to use `sumMerge` ? Do the translation mappers https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/metrics.py#L151 handle converting that?

tests/test_metrics_api.py
```diff
@@ -0,0 +1,362 @@
...
+                processed = processor.process_message(
+                    msg, KafkaMessageMetadata(0, 0, self.base_time),
+                )
+                if processed:
+                    events.append(processed)
+        write_processed_messages(self.storage, events)
+
+    def test_sets_basic(self) -> None:
+        query_str = """MATCH (metrics_sets)
+                    SELECT uniqCombined64Merge(value) AS unique_values BY project_id, org_id
```
MeredithAnya, line 237: same question here about using `uniqCombined64Merge` vs just `uniq` given the translation mappers

tests/test_metrics_api.py
onewland: I didn't know those existed, and I think I should definitely use them, as that tests more of the end-to-end for these datasets/entities

tests/test_metrics_api.py
onewland: same answer as above, I'll make that change

onewland:
Looking into the translation mappers, I've realized that this is not yet exercising the min/max/avg/count flows on distributions so any review can probably wait on that

evanh (commented):
One other thing I would maybe test is the different granularities. Maybe each case (Counters/Sets etc.) has a different granularity.

onewland:
This is ready for another look, I think

evanh (approved):

PR #2467, opened 2022-03-01
feat(subscriptions): Make `override-result-topic` required

lynnagara:
This removes the hardcoded default result topic value of "snuba-subscription-result-test"
which was used temporarily, and makes passing a value mandatory.

Note: this depends on applying a currently pending ops change in Sentry first to pass
--override-result-topic to the events subscriptions executor.


evanh (approved):

joshuarli:
https://github.com/getsentry/ops/pull/4406 is now applied.

PR #2468, opened 2022-03-01
fix(subscriptions): Commit all offsets in subscription executor

lynnagara:
Commit offsets for all queries in the executor even if we are temporarily
skipping execution (to avoid putting extra load on ClickHouse).

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -275,16 +275,6 @@ class ExecuteQuery(ProcessingStrategy[KafkaPayload]):
             float, state.get_config("executor_sample_rate", 0.0)
         )
 
-        # HACK: Just commit offsets and return if we haven't started rollout
-        # yet. This is just a temporary workaround to prevent the lag continuously
-        # growing and dwarfing others on op's Kafka dashboard
-        if self.__commit is not None and executor_sample_rate == 0:
-            self.__commit(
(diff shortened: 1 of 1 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -305,6 +295,12 @@ def submit(self, message: Message[KafkaPayload]) -> None:
                 )
             )
         else:
+            # Just commit offset even if we are skipping execution
+            if self.__commit is not None:
+                self.__commit(
```
fpacifici, line 300: This would commit every message, this puts a lot of load on Kafka for no reason. Commit is an expensive and synchronous operation (it writes on kafka and waits for replication).
You should only commit periodically.

fpacifici (changes requested):

snuba/subscriptions/executor_consumer.py
lynnagara: This is intentional. We commit every message that is after it is executed so equally this commits every message that is temporarily skipped. Now it may make sense in the future to try and optimise this, but I don't think it is presently an issue due to the number of subscriptions that we have being rather low.

fpacifici: you have hundreds of messages per second that means hundreds of commits per second. This is not a good number for Kafka. a handful of commits per second are ok.

PR #2469, opened 2022-03-01
perf(metrics) - consolidate metrics materialized views

onewland:
Consolidate materialized views for writing metrics raw data to aggregate tables. This reduces the number of distinct writes to the aggregate tables and hopefully improves performance.

When deployed and executed these migrations will not take immediate effect. They will be put in place when a settings change is deployed to set materialization_version = 2.

Once we've switched that over and verified that all new rows being written are at 2, we should drop the at that point non-functional materialized views that act on `materialization_version = 0`


snuba/migrations/snuba_migrations/metrics/templates.py
```diff
@@ -125,14 +125,38 @@ def get_forward_bucket_table_dist(
...
 
+MATVIEW_STATEMENT_CONSOLIDATED = """
+SELECT
+    org_id,
+    project_id,
+    metric_id,
+    arrayJoin([10,60,3600,86400]) as granularity,
+    tags.key,
+    tags.value,
+    toDateTime(granularity * intDiv(toUnixTimestamp(timestamp), granularity)) as timestamp,
```
onewland, line 136: We were not able to use `toStartOfInterval` here the way we do for the non-consolidated tables because the interval argument only accepts a constant value (it cannot be the result of an arrayJoin or another select).

This logic approximates what I did for direct aggregate writing in Python: https://github.com/getsentry/snuba/blob/a81d90790ef5b24fc4ef7a7dad90548791d21376/snuba/datasets/metrics_aggregate_processor.py#L41-L44

snuba/migrations/snuba_migrations/metrics/templates.py
```diff
@@ -190,21 +214,35 @@ def get_forward_view_migration_local(
...
+            storage_set=StorageSetKey.METRICS,
+            view_name=mv_name,
+            destination_table_name=table_name,
+            columns=aggregated_cols,
+            query=MATVIEW_STATEMENT_CONSOLIDATED
+            % {
+                "raw_table_name": source_table_name,
+                "aggregation_states": aggregation_states,
+            },
+        )
```
fpacifici, line 245: Why not just copying this piece of code in your forward_local methods.
Migrations were meant to be isolated from each other and immutable as they are append only. 
The idea of having complex templates that generate them has a lot of drawbacks as it is easy to change the semantic of an existing migration by mistake by refactoring the template. And it will be very hard to notice.

fpacifici (approved):
Looks good to me. Please see the only comment inline. 
Also before merging ensure it works end to end.

onewland:
Output SQL:

Sets:
```
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_sets_consolidated_mv_local TO metrics_sets_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(uniqCombined64, UInt64)) AS 
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    uniqCombined64State(arrayJoin(set_values)) as value
FROM metrics_buckets_local
WHERE materialization_version = 2
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;
```

Counters:
```
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_counters_consolidated_mv_local TO metrics_counters_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(sum, Float64)) AS 
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    sumState(value) as value
FROM metrics_counters_buckets_local
WHERE materialization_version = 2
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;
```

Distributions:
```
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_distributions_consolidated_mv_local TO metrics_distributions_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, percentiles AggregateFunction(quantiles(0.5, 0.75, 0.9, 0.95, 0.99), Float64), min AggregateFunction(min, Float64), max AggregateFunction(max, Float64), avg AggregateFunction(avg, Float64), sum AggregateFunction(sum, Float64), count AggregateFunction(count, Float64)) AS 
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)((arrayJoin(values) AS values_rows)) as percentiles, minState(values_rows) as min, maxState(values_rows) as max, avgState(values_rows) as avg, sumState(values_rows) as sum, countState(values_rows) as count
FROM metrics_distributions_buckets_local
WHERE materialization_version = 2
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;
```

PR #2473, opened 2022-03-02
feat(metrics) - switch to consolidated materialized views for metrics

onewland:
switch to the logic added in https://github.com/getsentry/snuba/pull/2469 as the new MVs are present now


fpacifici (approved):

onewland:
> tests are failing

yeah, the tests were looking for the exact old value of the setting. pushed an update

fpacifici (approved):

PR #2474, opened 2022-03-03
ref: Remove the ability to run a replacer on the events storage

lynnagara:
Events storage is deprecated, remove the ability to run a replacer
on it and simplify errors replacer code.

Most of the tests in test_replacer.py were already duplicated in
test_errors_replacer.py so could be deleted.

### DIFF
--- a/snuba/utils/streams/topics.py
+++ b/snuba/utils/streams/topics.py
@@ -8,7 +8,7 @@ from snuba import settings
 class Topic(Enum):
     EVENTS = "events"
     EVENT_REPLACEMENTS = "event-replacements"
-    EVENT_REPLACEMENTS_LEGACY = "event-replacements-legacy"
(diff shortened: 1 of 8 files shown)


snuba/utils/streams/topics.py
```diff
@@ -8,7 +8,7 @@
 class Topic(Enum):
     EVENTS = "events"
     EVENT_REPLACEMENTS = "event-replacements"
-    EVENT_REPLACEMENTS_LEGACY = "event-replacements-legacy"
+    # EVENT_REPLACEMENTS_LEGACY = "event-replacements-legacy"
```
evanh, line 11: Is there a reason to keep this?

evanh (approved):

snuba/utils/streams/topics.py
lynnagara: No it was an oversight, removed

PR #2476, opened 2022-03-03
ref: Remove events and events_ro storages

lynnagara:
These are no longer being read from or written to.


evanh (approved):
`+6 −1,066`

That is a nice looking change.

PR #2482, opened 2022-03-04
test: Attempt to fix flaky test

lynnagara:
This test often fails when run on GCP. Try waiting for the
future. Note the sleep is still required since the metric
gets incremented in the callback function that runs in a separate thread.
Also adds 3 retries.


lynnagara:
This passed CI on the first run.

onewland (approved):
Seems fine, maybe we could use something like https://github.com/pytest-dev/pytest-rerunfailures#re-run-individual-failures in the future

rahul-kumar-saini (approved):

MeredithAnya (approved):
seems reasonable to me

PR #2484, opened 2022-03-05
feat: Drop unused events tables

lynnagara:
Remove the original `sentry_local` and `sentry_dist` tables.


evanh (commented):
Question: the 0001 migration doesn't add the `_tags_hash_map` column, that happens in 0005. Does that need to be added here as well to make sure we end up back in the same state?

lynnagara:
> Question: the 0001 migration doesn't add the `_tags_hash_map` column, that happens in 0005. Does that need to be added here as well to make sure we end up back in the same state?

I forgot that we changed these later. Added _tags_hash_map and hierachical_hashes back.

evanh:
One more question: how does this work with self hosted and open source? Is there an edge case where the old table could still be used by someone?

lynnagara:
No, we did that roll out a long time ago, and you can see that table was already truncated in a prior migration. The only reason this code was kept around so long was for single tenant.

snuba/migrations/snuba_migrations/events/0016_drop_legacy_events.py
```diff
@@ -0,0 +1,191 @@
...
+            operations.AddColumn(
+                storage_set=StorageSetKey.EVENTS,
+                table_name="sentry_local",
+                column=Column(
+                    "_tags_hash_map",
+                    Array(UInt(64), Modifiers(materialized=TAGS_HASH_MAP_COLUMN)),
+                ),
+                after="_tags_flattened",
+            ),
+        ]
```
nikhars, line 159: This seems like a lot of effort to support backwards_local. Having to define all the columns again for supporting reverts. And there could be instances where you might miss a column which might have been added by a migration. Is there a mechanism in migrations where you can provide Clickhouse SQL directly? That would be useful in such scenarios

lynnagara: Yes there is. It's `operations.RunSql`. I considered just dumping the strings there but I did it this way because I (originally) thought it was a simple copy and paste.

evanh (approved):
LGTM.

snuba/migrations/snuba_migrations/events/0016_drop_legacy_events.py
nikhars: My primary concern is that we could have added/removed columns, added bloom filter indexes, modified TTL's after the first migration. A copy/paste of just the columns would miss them. Can you verify that the SQL command created by this is similar to the show create command on clickhouse?

lynnagara: Yes, it is the exact same string when reversing as it is before the migration. I've checked all the migrations for the things you described.

PR #2486, opened 2022-03-07
feat: Deploy new metrics subscription schedulers with Freight

lynnagara:


.freight.yml
```diff
@@ -42,6 +42,10 @@ steps:
     name: transactions-subscriptions-consumer
   - image: us.gcr.io/sentryio/snuba:{sha}
     name: sessions-subscriptions-consumer
+  - image: us.gcr.io/sentryio/snuba:{sha}
+    name: metrics-counters-subscriptions-scheduler
+  - image: us.gcr.io/sentryio/snuba:{sha}
+    name: metrics-sets-subscriptions-scheduler
```
nikhars, line 48: Do we need to add metrics-distributions?

lynnagara: Discussed this with the workflow team. They don't plan to have any subscriptions on metrics_distributions right now.

nikhars: Another question: Would it be possible to create an alert with metrics from both `metrics-counters` and `metrics-sets`? Let say, creating an alert when `counterA` from metrics-counters table goes above 100 and `setB` from metrics-sets goes above 10. That means running queries across 2 tables.

Does the subscription consumer need to support that use case?

lynnagara: We don't support anything like this currently (either via join or via 2 separate queries that are synced together in some other way). All subscriptions are a single query on a single table. It's possible we will want to add join support in the future, but that's out of scope here.

fpacifici: Also let's not assume the consumer will always be the same, thus we could eventually have two separate commit logs.

lynnagara: I'm not sure if you are suggesting to change something with the current implementation. Yes we could hypothetically have separate consumers/commit logs. We could support it in the scheduler if we decide to build out that feature. But I don't think it's particularly relevant right now. The scheduler mostly works the way it does currently (operates on a single entity) because we need to share code and stay compatible with the old subscriptions pipeline until that is deprecated. Once that is no longer the case I think other features like the one you describe will be easier.

PR #2487, opened 2022-03-07
feat(subscriptions): Set executor sample rate by dataset

lynnagara:
We are going to roll out the executor one dataset at a time. As a result
we need to be able to set the sample rate by dataset rather than a
shared rate for all executors.


snuba/subscriptions/executor_consumer.py
```diff
@@ -271,8 +271,9 @@ def submit(self, message: Message[KafkaPayload]) -> None:
 
         # We need to sample queries to ClickHouse while this is being rolled out
         # as we don't want to duplicate every subscription query
+        dataset_name = get_dataset_name(self.__dataset)
         executor_sample_rate = cast(
-            float, state.get_config("executor_sample_rate", 0.0)
+            float, state.get_config(f"executor_sample_rate_{dataset_name}", 0.0)
```
nikhars, line 276: Please don't forget to remove the config "executor_sample_rate" from prod config after you merge this change.

nikhars (approved):
Left a minor comment

PR #2488, opened 2022-03-07
perf(consumers): allow parallel collect option consumers

nikhars:
Added an option to pass the paralle collect flag to the consumers which can be used to control whether the consumers should use ParallelCollectStep or not.


MeredithAnya (approved):

snuba/cli/consumer.py
```diff
@@ -93,6 +93,9 @@
 @click.option(
     "--profile-path", type=click.Path(dir_okay=True, file_okay=False, exists=True)
 )
+@click.option(
+    "--parallel-collect", is_flag=True, default=False,
+)
```
lynnagara, line 98: nit: I find this more readable when the order here matches below - can you put this above the optional ones?

nikhars: Done

lynnagara (approved):
Thanks

PR #2489, opened 2022-03-07
perf(consumers): Skip processing similar storages

nikhars:
When the multistorage consumer needs to write to identical storages it does additional work for sanitizing and forming the BytesInsertBatch. This is unnecessary since we know that the data written to both the storages is exactly identical. This PR aims to avoid the unnecessary processing of messages by storing the result in a map and re-using the result in case identical storages are found.


snuba/consumers/consumer.py
```diff
@@ -590,13 +591,35 @@ def process_message(
         return result
 
 
+def process_message_multistorage_work(
```
fpacifici, line 594: Since this is not supposed to be used externally. Let's make it "private" ?
`_process_message_multistorage_work`

snuba/consumers/consumer.py
```diff
@@ -616,40 +639,70 @@ def process_message_multistorage(
...
+    it finds that the storages on which data needs to be written are identical. This is a
+    performance optimization since we remove the message processing time completely for all
+    identical storages like errors and errors_v2.
+
+    It is possible that the storage keys in the message could be a mix of identical and
+    non-identical storages. This method takes into account that scenario as well.
+
+    The reason why this method has been created rather than modifying the existing
+    process_message_multistorage is to avoid doing a check for every message in cases
+    where there are no identical storages like metrics.
```
fpacifici, line 666: Are you sure about doing this optimization? It seems the check is extremely cheap.

snuba/consumers/consumer.py
```diff
@@ -616,40 +639,70 @@ def process_message_multistorage(
...
+
+    intermediate_results: MutableMapping[
+        StorageKey, Union[None, BytesInsertBatch, ReplacementBatch]
+    ] = {}
+
+    for index, storage_key in enumerate(message.payload.storage_keys):
+        skip_processing = False
+        for other_storage_key, insert_batch in intermediate_results.items():
+            if are_writes_identical(storage_key, other_storage_key):
+                intermediate_results[storage_key] = copy.copy(insert_batch)
```
fpacifici, line 681: Do you need to do a copy of the batch? 
A deep copy would probably introduce too much overhead. This copy on the other hand is only a shallow copy, so I am not sure what you are trying to protect against. Is it even possible that the batch is modified after this point ?

snuba/consumers/consumer.py
```diff
@@ -616,40 +639,70 @@ def process_message_multistorage(
...
+
+        if skip_processing:
+            continue
+
+        # The `process_message` api of individual storage processors could modify the content of the payload
+        # in some cases. For example, they can remove redundant data from the payload to decrease size of
+        # storage on clickhouse. Hence, we need to make copies of the payload and send the copies to the api.
+        # The last storage will use the original payload to avoid unnecessary copying.
+        storage_message = (
+            copy.deepcopy(value)
```
fpacifici, line 693: I forgot about the deeopcopy. 
I think this could be one of the reasons we ran into performance issues with the multi storage consumer recently especially for consumers that have many storages (metrics has three) and high volume. 50k messages per sec.
Can we remove it and instead fix the processors? A processor should never modify the message. 
We can always have a test for each processor that verifies that the code does not change the original message

snuba/consumers/consumer.py
```diff
@@ -616,40 +639,70 @@ def process_message_multistorage(
...
+    for index, storage_key in enumerate(message.payload.storage_keys):
+        skip_processing = False
+        for other_storage_key, insert_batch in intermediate_results.items():
+            if are_writes_identical(storage_key, other_storage_key):
+                intermediate_results[storage_key] = copy.copy(insert_batch)
+                skip_processing = True
+                break
+
+        if skip_processing:
+            continue
```
fpacifici, line 686: Nit: I think you should be able to avoid this continue that are counter intuitive by inverting the conditions:
when you find the storage, instead of  setting `skip_processing=True` assign the message to a variable (processed_message).
Then here, if the `processed_message` is None, do the processing, otherwise store processed_message in `intermediate_results`

snuba/consumers/consumer.py
```diff
@@ -683,10 +736,19 @@ def __init__(
...
 
+        self.__process_message_fn = process_message_multistorage
+        for storage1, storage2 in itertools.combinations(self.__storages, 2):
+            if are_writes_identical(
+                storage1.get_storage_key(), storage2.get_storage_key()
+            ):
+                self.__process_message_fn = (
+                    process_message_multistorage_identical_storages
+                )
+                break
```
fpacifici, line 750:
```
if any(are_writes_identical for storage1, storage2 in itertools.combinations(self.__storages, 2)):
   self.__process_message_fn = (
        process_message_multistorage_identical_storages
    )
```

fpacifici (commented):
Some comments inline.
Also we should remove the deepcopy. I think that is one of our performance issue.

snuba/consumers/consumer.py
nikhars: Yes I want to keep this optimization for now. Don't want to impact other storages when only a few would be affected.

snuba/consumers/consumer.py
nikhars: That's right. The batch is not modified after this point. Removed the copy.

tests/consumers/test_multistorage_consumer.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.parametrize("message, expected", test_data)
+def test_process_multistorage_equivalent_write(
+    message: Message[MultistorageKafkaPayload], expected: bool
+) -> None:
+    results = process_message_multistorage_identical_storages(message)
+    assert len(results) == 2
+
+    assert (results[0][1] == results[1][1]) is expected
```
fpacifici, line 78: you may want to assert the two results are the same instance:
`id(result[0][1]) == id(result[1][1])`

fpacifici (approved):

PR #2491, opened 2022-03-08
fix(consumers): Remove deepcopy from multi storage consumers

nikhars:
### Context
We recently added the deepcopy in the process_message_multistorage api since the transaction processor was modifying the message in place. Doing a deepcopy would ensure that functionality would work as expected. But it has a side effect of affecting the consumers performance since copying is a heavy operation.

### Solution
Fix the transaction processor so that the operation where it needs to modify the context does the deepcopy. That way other consumers are not affected.

### Tests
Added a test to check idempotency of messages for errors and transactions processor.


evanh (commented):
Isn't this just moving the performance problem to a different file? Deep copying the contexts could still be a performance issue.

nikhars:
> Isn't this just moving the performance problem to a different file? Deep copying the contexts could still be a performance issue.

Not exactly.
Now the decision of taking a performance impact is on the specific processor. In this particular case, transaction processor would take a hit so the transactions consumer have a performance hit. But errors consumer and metrics consumer don't have to take the performance hit.

For transactions, the offending statement is https://github.com/getsentry/snuba/blob/master/snuba%2Fdatasets%2Ftransactions_processor.py#L230-L235. The tradeoff is saving storage space vs taking a performance penalty. Based on this comment https://github.com/getsentry/snuba/pull/2353#pullrequestreview-866305303, I decided to go with a performance penalty for now.

snuba/datasets/transactions_processor.py
```diff
@@ -206,8 +207,8 @@ def _process_breakdown(
     def _process_contexts_and_user(
         self, processed: MutableMapping[str, Any], event_dict: EventDict,
     ) -> None:
-        contexts: MutableMapping[str, Any] = _as_dict_safe(
-            event_dict["data"].get("contexts", None)
+        contexts: MutableMapping[str, Any] = copy.deepcopy(
+            _as_dict_safe(event_dict["data"].get("contexts", None))
         )
```
fpacifici, line 212: Why is this working?
This pop happens before the call to _process_contexts_and_user
https://github.com/getsentry/snuba/blob/master/snuba%2Fdatasets%2Ftransactions_processor.py#L334-L337

fpacifici:
There is another way, but it may be overkill. Here we would have to profile it to see if there is an improvement: since the mutation code is only removing elements from the contexts dictionary. You could just build a contexts dictionary (to put in the processed event) skipping the contexts you want to remove instead of deepcopying all the others.

snuba/datasets/transactions_processor.py
nikhars: Good catch. You are right that the message was missing those fields.

I have refactored the code so that all modifications of contexts happen just before we fill in the contexts key and value columns. So there should not be any more blindspots.

Also verified that the errors processor does not perform and `del` or `pop` operations.

nikhars:
> There is another way, but it may be overkill. Here we would have to profile it to see if there is an improvement: since the mutation code is only removing elements from the contexts dictionary. You could just build a contexts dictionary (to put in the processed event) skipping the contexts you want to remove instead of deepcopying all the others.

While it is a viable option, I would like to avoid doing this right now. We should be able to move away from multistorage consumers in a couple of months time and then it won't matter.

snuba/datasets/transactions_processor.py
```diff
@@ -331,10 +327,44 @@ def _process_spans(
...
-        trace_context.pop("exclusive_time", None)
+        transaction_ctx.pop("hash", None)
+        transaction_ctx.pop("exclusive_time", None)
+
+        skipped_contexts = settings.TRANSACT_SKIP_CONTEXT_STORE.get(
+            processed["project_id"], set()
+        )
+        for context in skipped_contexts:
+            if context in sanitized_context:
+                del contexts[context]
```
fpacifici, line 365: I think this should be sanitized contexts

fpacifici (changes requested):

fpacifici (approved):

PR #2494, opened 2022-03-09
perf(metrics) - create polymorphic bucket table

onewland:
This is a migration for creating a bucket table for metrics that combines all metric types (counters, sets, distributions) and can therefore reduce the amount of merges when writing data from the `snuba-metrics` topic.

If people want I can add materialized views that actually take data from this table and write to aggregate tables to this PR, but I assumed the preference would be to split that up


snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=self.local_table_name,
+                columns=self.column_list,
+                engine=table_engines.MergeTree(
+                    storage_set=StorageSetKey.METRICS,
+                    order_by="(project_id, metric_id, tags.key, tags.value, timestamp)",
```
evanh, line 43: Is removing org_id intentional? The existing metrics entity requires org_id be supplied in every query, and seems like a fairly efficient filter. I assume this is related to the performance tracing you've been doing.

onewland: The thought process behind that is that there's no filtering that org_id does that project_id doesn't drill down better. If we want to add org_id into the ORDER BY I'm fine with it, but I think it's a less effective ordering column than project_id

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+    column_list: Sequence[Column[Modifiers]] = [
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("timestamp", DateTime()),
+        Column("tags", Nested([Column("key", UInt(64)), Column("value", UInt(64))]),),
+        Column("metric_type", String(Modifiers(low_cardinality=True))),
+        Column("set_values", Array(UInt(64))),
+        Column("count_value", Float(64)),
+        Column("distribution_values", Array(Float(64))),
```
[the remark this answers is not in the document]
onewland: Unless I'm not understanding your question, we don't know how many distribution values we'll get in a single entry. So this has to be something variably sized.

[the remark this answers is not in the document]
evanh: How could that be? If it's an array of variable size, how do we know which values correspond to which distributions?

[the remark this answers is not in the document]
onewland: I think the name of this column is maybe throwing you a bit. 

Let's say I have a client app, project_id=20, that's reporting performance of a database call as metric_id = 5, with tag[1234]=5678 as a distribution metric and it's making 9 database calls per request. The client therefore sends us 9 values in this array.

Every time a batch gets written to `metrics_buckets_consolidated` that has all these "key" values (project_id=20, metric_id=5, tag.key=1234, tag.value=5678, metric_type="distribution"), that's going to be a new row which the materialized view will write to the metrics_distributions aggregate table, transforming these distribution_values into a single datasketch. The aggregating merge tree takes its snapshot of the distribution in a datasketch and merges it with that in-memory datasketch from the materialized view containing the 9 values and writes a new datasketch accounting for those new values to the aggregate table.

Does that make sense?

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+    dist_table_name = "metrics_buckets_consolidated_dist"
+
+    column_list: Sequence[Column[Modifiers]] = [
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("timestamp", DateTime()),
+        Column("tags", Nested([Column("key", UInt(64)), Column("value", UInt(64))]),),
+        Column("metric_type", String(Modifiers(low_cardinality=True))),
+        Column("set_values", Array(UInt(64))),
```
fpacifici, line 26: I wonder whether empty array or null will be more efficient to store.

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+        Column("count_value", Float(64)),
```
fpacifici, line 27: Functionally:
I would assume your materialized views will always filter by metric_type. So whether this is nullable or not does not matter. Could you confirm ? 
Storage:
Could you check whether a column full of 0.0 float64 is bigger than a nullable column full of NULL ?
You can do that locally by creating a table with a ton of values and check the `bytes_on_disk` field in the `system.columns` table.

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
fpacifici: Ok a few changes. 
- `metric_type` should go first. The matview will be much more efficient as it always filter by metric type 
- remove tags.key/tags.value. This is a mergetree, not a RreplacingMergeTree or aggregatingMergeTree, so rows do not get squashed together. Adding the tags.key/tags.value only makes the index unnecessarily big. tags can be somewhere around 4 * 2 * 64 bit
- I would still keep the org_id for as long as it is in the aggregated tables. It is small.

fpacifici (commented):
Mostly good. 
Please see the change requested in the order by key and please do a quick test on the column size (this should be a very quick test, please let me know if you need more context)

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Create a polymorphic bucket table for metrics_sets,
+    metrics_distributions, and metrics_counters
+    """
+
+    blocking = False
+    local_table_name = "metrics_buckets_consolidated"
+    dist_table_name = "metrics_buckets_consolidated_dist"
+
+    column_list: Sequence[Column[Modifiers]] = [
```
fpacifici, line 19: This is missing the use case id to partition the table by use case.

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+        ]
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=self.dist_table_name,
+                columns=self.column_list,
+                engine=table_engines.Distributed(
+                    local_table_name=self.local_table_name, sharding_key=None,
```
nikhars, line 64: Should we add a sharding key for the table? Maybe by the `org_id` or `project_id`? That way the materialized view does not need to go across shards when we increase the number of shards.

nikhars (commented):
> This is a migration for creating a bucket table for metrics that combines all metric types (counters, sets, distributions) and can therefore reduce the amount of merges when writing data from the snuba-metrics topic.

Is the merge process proving to be a bottleneck? Have you considered adding a new shard versus creating the polyglot table? I am trying to reason why this would be better than adding a new shard.

onewland:
> > This is a migration for creating a bucket table for metrics that combines all metric types (counters, sets, distributions) and can therefore reduce the amount of merges when writing data from the snuba-metrics topic.
> 
> Is the merge process proving to be a bottleneck? Have you considered adding a new shard versus creating the polyglot table? I am trying to reason why this would be better than adding a new shard.

I don't think that merging bucket tables and sharding data to separate clusters are mutually exclusive unless we were specifically planning to shard by metric type.

Well, we'd like to reduce the number of merges and the number of inactive parts. Ideally if the storage consumers could cause less merges per write on ClickHouse, then we could be reducing the number of consumers and increasing their commit frequency. We currently have 16 replicas consuming the snuba-metrics topic today, and we believe that number is unreasonably high.

If we shard by anything but metric_type, this will make inserts more efficient.

onewland:
> > This is a migration for creating a bucket table for metrics that combines all metric types (counters, sets, distributions) and can therefore reduce the amount of merges when writing data from the snuba-metrics topic.
> 
> Is the merge process proving to be a bottleneck? Have you considered adding a new shard versus creating the polyglot table? I am trying to reason why this would be better than adding a new shard.

If we can write to one table instead of three, then we should be able to write more frequently as inserts will not cause CH to do as much merging work. If we can write more frequently, then we should be able to scale down our currently overscaled snuba-metrics consumers (currently scaled to 16 concurrent pods).  

Additionally, we carry forward this performance benefit even if we decide to shard later (assuming we did not shard on metric type).

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
onewland: I was sort of copying this forward from another migration: https://github.com/getsentry/snuba/blob/b3d58e642dd1d1dee53a1cbe022e3dde159e38ac/snuba/migrations/snuba_migrations/metrics/0001_metrics_buckets.py#L25

It's not clear to me what the distributed table really does for us here, but maybe @fpacifici could add a comment since he wrote that migration

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
onewland: You cannot make arrays nullable (https://clickhouse.com/docs/en/sql-reference/data-types/nullable/#data_type-nullable) but an empty array seems to take about as much space as a scalar Float64 (see https://gist.github.com/onewland/20799a8feee3317a1a6f6061f28297a9 for some testing I did around NULL)

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
onewland: a column filled with 0.0 (or any float) is definitely much smaller than a column filled with NULLs, see some testing I did with https://gist.github.com/onewland/20799a8feee3317a1a6f6061f28297a9

onewland: This result may have actually been due to bad partitioning, I'll update in a minute

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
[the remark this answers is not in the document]
onewland: Yeah, the persistent data sketches are in the aggregate tables (`metrics_sets_local`, `metrics_counters_local`, `metrics_distributions_local`).

The materialized views (sort of equivalent to stored procedures/triggers in traditional SQL databases) read from the bucket tables and write to the aggregate tables with some light transformation

onewland:
I think I've addressed all comments and I'm going to start working on the materialized view changes that build on top of this. Please let me know if you want more changes

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Create a polymorphic bucket table for metrics_sets,
+    metrics_distributions, and metrics_counters
+    """
+
+    blocking = False
+    local_table_name = "metrics_buckets_consolidated"
```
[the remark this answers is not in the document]
fpacifici: So first you need to add a `_local` suffix for the local table. That's the convention we use everywehere.
Then I would call it `metrics_raw_local`. The aggregated tables do not have a suffix that tell that they are aggregated, while the raw/bucketed one do. So this is a bucketed one I would add a suffix to say that. We cannot call this `metrics_buckets_local` because it is already taken, so maybe `raw` ?

snuba/migrations/snuba_migrations/metrics/0020_polymorphic_buckets_table.py
fpacifici: > Should we add a sharding key for the table? Maybe by the org_id or project_id? That way the materialized view does not need to go across shards when we increase the number of shards.

The materialized view only operates locally on local tables, it would never go across shards.

> It's not distributing application reads which are all going through the aggregate tables, and I assume writes on any local instance causes all the replicas to make the same copies

If you write directly on a storage node the distributed table is not needed, but, since we are likely to reintroduce the query node as we will nto have time to fix the deployment problems before production without the query node, we will need a distributed table for the query node to write onto.

fpacifici (approved):
Please fix the table naming before merging.
Also please add the DDL statement to the PR so I can apply that to prod. 
Also, let's double check this. 
Try to add `SETTINGS ttl_only_dropparts = 1` to the create statement and see if it works on 20.7. I am not 100% sure it was introduced later

lynnagara:
Will you still be keeping the 3 separate storages and entities for each metric type and just mapping them all to this one new table? Curious how that part will work.

onewland:
> Will you still be keeping the 3 separate storages and entities for each metric type and just mapping them all to this one new table? Curious how that part will work.

I was considering writing to a new single storage for the raw datatype, since I think ideally the processor would be shared between the types for this new approach

lynnagara:
What about the entity? That is part of the API and affects how subscriptions for e.g. work. I'm not sure it should depend on the underlying storage.

onewland:
> What about the entity? That is part of the API and affects how subscriptions for e.g. work. I'm not sure it should depend on the underlying storage.

I don't think subscriptions or queries will change as a result of this PR -- they only read from the aggregate tables (which will function the same)

onewland:
@fpacifici 

```
CREATE TABLE IF NOT EXISTS metrics_raw_local (use_case_id LowCardinality(String), org_id UInt64, project_id UInt64, metric_id UInt64, timestamp DateTime, tags Nested(key UInt64, value UInt64), metric_type LowCardinality(String), set_values Array(UInt64), count_value Float64, distribution_values Array(Float64), materialization_version UInt8, retention_days UInt16, partition UInt16, offset UInt64) 
ENGINE MergeTree() 
ORDER BY (metric_type, org_id, project_id, metric_id, timestamp) 
PARTITION BY (use_case_id, toStartOfDay(timestamp)) 
TTL timestamp + toIntervalDay(7)
SETTINGS ttl_only_dropparts = 1; -- added by hand
```

PR #2498, opened 2022-03-10
feat(profiling): Add and remove some fields

phacops:
After some round of testings, we realized a few things:
- we don't need to store the symbols separately as the symbolicated version will be store in the profile column.
- we need to have an ID specific to the profile and not just copy the transaction_id
- `received` is a timestamp in Relay and there's no need to transform it as a RFC3339 string along the way to be parsed here as such, might as well just parse it as a timestamp here


evanh (commented):
This looks good to me, I just want to confirm that if this table exists and is being written too, that this PR will cause a bunch of errors.

phacops:
This table doesn't exist anywhere yet. Migrations are still skipped and the dataset is still disabled.

evanh (approved):
OK this is fine by me.

PR #2500, opened 2022-03-10
fix(subscriptions): Fix committing of offsets in scheduler

lynnagara:
There were a number of issues with the way committing offsets were
being handled in the subscriptions scheduler previously. This resulted
in offsets not actually being committed when they should be.

- The ProvideCommitStrategy was attempting to set the "committable" status
based on the partition with the earliest timestamp (based on the original topic).
This is wrong, it should actually be the partition with the earliest offset
on the commit log topic that can be safely committed.
- The strategy was attempting to mark all messages as either "committable" or
not. This is not the right approach as it a later message cannot necessarily
be immediately committed, it only makes an earlier one safe to commit. Now
we are passing the offset_to_commit (which is always an earlier one) through
the pipeline together with the actual tick message.
- A bug in ScheduledSubscriptionQueue.peek() was actually causing offsets
to be committed even if they were never marked "committable" to begin with,
resulting in the issue not being discovered earlier as long as there are
any subscriptions that are scheduled.


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -33,7 +33,9 @@
 
 class CommittableTick(NamedTuple):
     tick: Tick
-    should_commit: bool
+    # Offset that we can safely committed once the tick is processed.
+    # Not necessarily the same as the tick's offset.
+    offset_to_commit: Optional[int]
```
fpacifici, line 38: Could you please adjust the docstring of `ProvideCommitStrategy` (specifically the last paragraph)? It still reference `should_commit`

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -100,50 +103,56 @@ def submit(self, message: Message[Tick]) -> None:
...
         tick_partition = message.payload.partition
         assert tick_partition is not None
         self.__latest_messages_by_partition[tick_partition] = message
 
-        slowest = message
-        fastest = message
+        # Keep track of the slowest/fastest partition based on the tick's lower timestamp.
+        # This is only used for recording the partition lag metric.
+        slowest = message.payload.timestamps.lower
+        fastest = message.payload.timestamps.lower
```
fpacifici, line 123: Am I missing something or the task of this strategy can be done in a much simpler way by the TickBuffer by flagging a Tick as committable if, when it submits it, that is the oldest offset it knows about? 

Example.
Let's say that TickBuffer, at a point sees that all partitions are at least at time 99.
`P2 T98 O1 P2 T99 O2 P1 T99 O12`
When it submits these three messages it knows that it can only let the next steps commit up to offset 2 because it does not have any offsets below that in its buffer, while it has, in the buffer, all the offsets between offset 2 and 12.

I think it may be easier to follow that way than trying to rebuild the watermark in a separate strategy. It is quite tricky to ensure this works properly.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -100,50 +103,56 @@ def submit(self, message: Message[Tick]) -> None:
...
 
-            if partition_timestamp < slowest.payload.timestamps.lower:
-                slowest = partition_message
+            if partition_timestamp < slowest:
+                slowest = partition_timestamp
 
-            if partition_timestamp > fastest.payload.timestamps.lower:
-                fastest = partition_message
+            if partition_timestamp > fastest:
+                fastest = partition_timestamp
```
fpacifici, line 139: Do you actually need to keep track of `slowest` and `fastest` for anything other than logging the lag ?

fpacifici (commented):
Please see my comments on ProvideCommitStrategy

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: No we don't, that's noted in a comment above where the variables are declared.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Agree this is tricky but I think that TickBuffer works very differently and I think combining this functionality into there is going to be even harder. The TickBuffer's main job is to worry about the differences between "global" and "partition" mode and submit them to the next step and the right time so no one else has to care about that. In order to do this it keeps things ordered by time and submits based on time (but only for the "global" strategy, "partition" just immediately submits). Unlike TickBuffer, this ProvideCommitStrategy is not really concerned with timestamps at all (apart from some metrics) - it only cares about keeping track of the latest offset that it has already exceeded on every partition - that is the one that is safe to commit.

I think the example you gave is more complicated and has more edge cases. 1) It doesn't work for the "partition" strategy so we would need to build something different to work there. 2) We try to empty buffers as much as possible so there are cases where we can't find an offset in every buffer. In the end I think we would just end up with what we have here - a mapping of partition index to latest offset seen.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -45,10 +47,11 @@ class ProvideCommitStrategy(ProcessingStrategy[Tick]):
...
 
-    A tick can be safely committed only if its lower timestamp has also been
-    reached on every other partition. If that condition is not met, the message
-    is still submitted to the next step with a `should_commit` value of false
-    indicating that it's offset is not to be commited.
+    Each time a tick message is received by this strategy we need to figure out
+    if we can advance the offset that is committed. The offset that is considered
+    safe to commit is based on the partition with the earliest offset that we have
+    already seen. It's marked `offset_to_commit` and submitted with the tick
+    to the next step.
```
fpacifici, line 54: Showing an example in the docstring may help

snuba/subscriptions/scheduler_processing_strategy.py
fpacifici: Ok, 

I think there is a problem though. 
I am not sure you can identify the offset to commit by looking only at the offset and without maintaining a watermark in terms of timestamp.
Example:
Let's say this step receives commits in this sequence
`P1 T102 O5.  P0 T100 O1.  P0 T101 O2 P0 T102 O3` (this is a watermark provided by TickBuffer) then `P0 T103 O4 ......`

After `P0 T102 O3` This would insrtuct to commit offset 3. And it seems right. 
Though then we will receive `P0 T103 O4`. Since `P1 T102 O5` is already in `self.__latest_messages_by_partition` and offset 4 is lower than offset 5, we would mark offset 4 to be committed even if we will not have received T103 (or higher) on partition 1.
While I think we should be waiting for that timestamp as well.

fpacifici: What about reconsidering the idea of making the TickBuffer just hold off ticks until the watermark is ready.
Instead what about making TickBuffer produce one message per watermark (every time we want to advance) that contains all the ticks and the time range to process.

This way `ProvideCommitStrategy` will know which ticks correspond to a given watermark, so we should be able to always commit the highest offset of the slowest partition.

lynnagara: It's not a problem to commit offset 4 after we got messages 4 and 5. The sequence in your example where timestamps are received out of order is only possible with the tick buffer in "partition" mode, otherwise they'd be ordered. The timestamps other partitions have reached is never important in partition mode and we never even look at it except to log a metric. The sequence received by this strategy is basically the same (well, accounting for the ticks) as the sequence of those messages in the underlying commit log topic.

lynnagara: About making the tick buffer hold off ticks until watermark is ready - are you are suggesting removing `partition` mode completely and making everything `global` to make it easier to combine the strategies?

lynnagara: In general, I feel we should
- execute as soon as we can (don't fill our buffers with ticks unnecessarily if we don't need to)
- commit the latest possible offset that we can do safely to avoid unnecessary duplication

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Will add it

snuba/subscriptions/scheduler_processing_strategy.py
fpacifici: > About making the tick buffer hold off ticks until watermark is ready - are you are suggesting removing partition mode completely and making everything global?
No, there are two ways to implement partition mode. 
1) have a different path. No `ProvideCommitStrategy` nor `TickBuffer` 
2) `TickBuffer` would produce a watermark per commit log message, without waiting for anything and set that watermark to always be committed. So basically `ProvideCommitStrategy` does nothing.

fpacifici: Let me take a step back and explain the reasoning more. 
My concern with this implementation is that, if we will find a bug in production (let's say the scheduler stop advancing), we will have a really really hard time troubleshooting it.
I think the underlying reasons for this are:
- the actual complexity in trying to rebuild the watermark in `ProvideCommitStrategy`. There are a ton of corner cases  to keep track of. This is not incidental complexity. It is real complexity in the moment we want to rebuild the watermark from the information `ProvideCommitStrategy` receives alone. Ensuring all the combinations are covered is really hard.
- The fact that `TickBuffer` can produce ticks out of order (across partitions. For a given partition they are in order). 
- The fact that we may have to wait before committing an offset because there may be an older offset (for a later time watermark) that cannot yet be committed.

So I think it is worth to go back thinking how to simplify this system conceptually. It does not have to be my idea of making `TickBuffer` produce watermarks, if you have a better one that better.
Do you agree with this?

If we agree on this point. I think what we should try to avoid is to make `ProvideCommitStrategy` having to rebuild a watermark (to find the offset to commit) based on messages that can be out of order without any additional help from TickBuffer. If we break this, I think, `ProvideCommitStrategy` will become much simpler. For simpler I mean easy to assess whether it is correct or not.

What do you think ?

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -45,12 +47,25 @@ class ProvideCommitStrategy(ProcessingStrategy[Tick]):
     guarantees that all subscriptions are scheduled at least once and we do
     not miss any even if the scheduler restarts and loses its state.
 
-    A tick can be safely committed only if its lower timestamp has also been
-    reached on every other partition. If that condition is not met, the message
-    is still submitted to the next step with a `should_commit` value of false
-    indicating that it's offset is not to be commited.
+    Each time a tick message is received by this strategy we need to figure out
+    if we can advance the offset that is committed. The offset that is considered
+    safe to commit is the latest offset that has been seen on every partition.
```
fpacifici, line 52: Could you pelase specify here what you mean by `seen on every partition`.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: I think this strategy and the code to implement it taken by itself is actually fairly simple- the rule applied is this:
- Commit the latest possible offset that every partition has reached

What's hard is proving that this strategy taken together with the TickBuffer is correct. Now what makes this hard? I believe it's because each step reorders messages in a slightly different way. So there are roughly 3 orderings of messages we are dealing with:
1 - There is the original ordering of ticks as they arrive at the strategies
2 - There is the order that the TickBuffer submits to the next step. For "partition" this is the same as 1 but for "global" it is not
3 - The order that we can safely commit the offsets in

Now correct me if I'm wrong here but I believe what you are kind of implying is this: 1 and 3 seem to be the same so why do we re-order to 2 in the middle and why not keep 1 to reuse for 3. But this wouldn't be correct - because 3 actually depends on both 1 and 2 and so must happen afterwards.

To try and make this more concrete - imagine if ticks are sitting in the tick buffer waiting to be submitted but have not been submitted yet. Just because the tick buffer has seen them doesn't make those offsets safe to commit because we haven't done anything with them yet. Doing it as a separate step afterwards guarantees that those ticks have already been submitted for scheduling and are thus safe to commit. We just wouldn't be guaranteed of that if we attempted to move this any earlier or reuse some prior ordering.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: sorry i mean `latest offset that has been reached or exceeded on every partition`.. is that clearer?

fpacifici (approved):

PR #2505, opened 2022-03-14
feat(profiling): Remove profiles from the list of disabled datasets

phacops:


lynnagara:
This depends on `profiles` being present in the cluster configuration in every environment, otherwise it will cause an error when attempting to start Snuba processes.

fpacifici (approved):

PR #2506, opened 2022-03-14
fix(transactions): Remove flattened columns

nikhars:
### Context
The transactions processor currently defines the contexts flattened and tag flattened columns which are not filled in by the consumers and not queried by the api's. These columns have a TTL expiry of 1 day which causes Clickhouse to spend merge TTL on something which we don't use. Hence removing these columns.

### Migration command
Local operations:
ALTER TABLE transactions_local DROP COLUMN IF EXISTS _tags_flattened;
ALTER TABLE transactions_local DROP COLUMN IF EXISTS _contexts_flattened;

Dist operations:
ALTER TABLE transactions_dist DROP COLUMN IF EXISTS _tags_flattened;
ALTER TABLE transactions_dist DROP COLUMN IF EXISTS _contexts_flattened;


snuba/datasets/storages/transactions_common.py
```diff
@@ -68,10 +68,8 @@
         ("http_method", String(Modifiers(nullable=True))),
         ("http_referer", String(Modifiers(nullable=True))),
         ("tags", Nested([("key", String()), ("value", String())])),
-        ("_tags_flattened", String()),
```
onewland, line 71: Does this make deploy complicated? e.g. will we cause errors if the ColumnSet is missing columns returned from the database? Or is it safe to deploy this logic change and then run the migrations later?

onewland (approved):
Looks like a nice, simple improvement.

snuba/datasets/storages/transactions_common.py
onewland: I don't think you have to worry about prod since we don't run migrations automatically. Less certain about single-tenant/open source

nikhars: The reason I would separate these out is that if the consumer is broken because of the change, I just need to add the columns back. If I do it this way, the migrations would need to be reverted and I would want to avoid reverting migrations.

PR #2509, opened 2022-03-14
Move clickhouse queries of death to public docs

volokluev:


docs/source/clickhouse/death_queries.rst
```diff
@@ -0,0 +1,16 @@
+Clickhouse Queries Of Death
+===========================
+
+
+The following queries have been shown to segfault clickhouse on 20.7 (which is what snuba is running). Do not run these queries in the tracing tool, unless you really want to take clickhouse down.
```
onewland, line 5:
```suggestion
The following queries have been shown to segfault clickhouse on 20.7 (which is the minimum Clickhouse version of Snuba). Do not run these queries in the tracing tool, unless you really want to take clickhouse down.
```

for our open source users

onewland (approved):
Seems fine. We should settle on a consistent way of capitalizing ClickHouse (which looks like how they do it on clickhouse.com)

docs/source/clickhouse/death_queries.rst
onewland: Also, did this segfault 20.3? Because I think that's the actual min version in the console

volokluev: I'm pretty sure we run 20.7, but also possibly 20.3?

PR #2510, opened 2022-03-14
[wip] connect writing pipeline for the raw/polymorphic bucket

onewland:


lynnagara:
Subscriptions are stored and managed based on entity so whether that will need to change depends on what your plans are for the read path of this storage. Today we have MetricsSetsEntity, MetricsCountersEntity and MetricsDistributionsEntity. Are you planning to eventually collapse them all into a single entity as well or are you planning to keep all 3 and just route them to the same storage with a different filter applied to each.

snuba/datasets/storages/metrics.py
```diff
@@ -47,6 +49,35 @@
     Column("offset", UInt(64)),
 ]
 
+polymorphic_bucket = WritableTableStorage(
```
fpacifici, line 52: Tied to @lyn's comment. Now each metric entity has a writable storage. Once this is done and we get rid of the other storages, how will that look like? Are we going to have this writable storage as an Entity-less storage (not associated to any entity)? which would be cool.
And are we going to remove the writable storage from the metrics entities? Not sure if that is possible, never tried.

tests/datasets/test_metrics_processor.py
```diff
@@ -106,7 +108,7 @@
                 "tags.key": [10, 20, 30],
                 "tags.value": [11, 22, 33],
                 "values": [324.12, 345.23, 4564.56, 567567],
-                "materialization_version": 2,
+                "materialization_version": MATERIALIZATION_VERSION,
```
fpacifici, line 111: Do you plan to add a test for the new processor ?

fpacifici (commented):
A few questions inline, but, apart from the issue with devserver.py, it seems good.

snuba/cli/devserver.py
```diff
@@ -143,13 +143,8 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
...
                     "snuba",
-                    "multistorage-consumer",
-                    "--storage=metrics_counters_buckets",
-                    "--storage=metrics_distributions_buckets",
-                    "--storage=metrics_buckets",
-                    "--storage=metrics_distributions",
-                    "--storage=metrics_counters",
-                    "--storage=metrics_sets",
+                    "consumer",
+                    "--storage=metrics_raw",
```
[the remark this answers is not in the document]
onewland: This PR is based off the raw materialized view branch but it does look like that may have broken the sentry tests. I can revert for now

tests/datasets/test_metrics_processor.py
onewland: We are in effect doing this in the e2e test_metrics_api.py because we're using the polymorphic bucket as the incoming writable storage, but I can write some unit tests more specific to the processor itself.

snuba/datasets/storages/metrics.py
onewland: I was assuming the first option

tests/datasets/test_metrics_processor.py
fpacifici: I don't think we should rely on the end to end tests only ([test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html)). Everything is covered there, this is true, but that hardly help you in testing a functionality. You cannot test corner cases and if something break, you will just see a large test failing without having any hint on what broke (query path, consumer, processor, etc.).
The unit test for the processor still has the value to validate that the processor is processing the messages appropriately. It only covers that area, it can cover corner cases and if that breaks you know the processor is broken, no need for additional investigation

onewland: yeah, in between my response and your reply I actually added unit tests for the polymorphic processor

snuba/cli/devserver.py
```diff
@@ -147,9 +147,7 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--storage=metrics_counters_buckets",
                     "--storage=metrics_distributions_buckets",
                     "--storage=metrics_buckets",
-                    "--storage=metrics_distributions",
-                    "--storage=metrics_counters",
-                    "--storage=metrics_sets",
+                    "--storage=metrics_raw",
```
lynnagara, line 150: Just to confirm, does this already work end to end in dev with `metrics_raw`?

onewland: yes, it does

snuba/datasets/storages/metrics.py
lynnagara: If I'm not mistaken what Filippo suggested are not separate options but 2 changes that are compatible. I think we should do both.
- separate writable storage (basically this PR)
- remove the other 3 writable storages from the metrics entities so they each only have a readable storage

onewland: Is there a reason they can't share a writable storage? that could let use re-use the sentry snuba integration tests without making changes in that repo

onewland:
Here's a change to update the writable storages that works with the sentry snuba integration tests (or at least a couple I ran quickly that failed before if I tried to change materialization_version to 3): https://github.com/getsentry/snuba/pull/2512

snuba/cli/devserver.py
lynnagara: I think it's better to hold off on changing this line until https://github.com/getsentry/snuba/pull/2512. I believe devserver is used by some test suites  / Sentry dev env and this may break tests. It's probably not going to be a big deal since things are still behind the `ENABLE_SENTRY_METRICS_DEV` flag but imo it's still better to do after all the tests will pass.

lynnagara (approved):

PR #2514, opened 2022-03-16
perf(consumer) Prepare to remove the latency metric at every message

fpacifici:
We are logging a latency metric between clickhouse write an message timestamp at every message. This seems to contribute to up to 1 second latency for every batch, and likely it is the cause of some hard to explain backlog events.

The idea is to only log one latency per batch, as our SLO should be based on the max latency. 
Before replacing the existing metric, this PR starts by recording the max latency and the avg latency per batch to compare how consistent is with the current metric. If it is close enough, we can replace the existing one.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -138,7 +138,7 @@ class InsertBatchWriter(ProcessingStep[BytesInsertBatch]):
                 "max_end_to_end_latency_ms", max_end_to_end_latency * 1000
             )
 
-        self.__metrics.timing("batch_write_ms", write_finish - write_start)
+        self.__metrics.timing("batch_write_ms", (write_finish - write_start) * 1000)
         rows = sum(len(message.payload.rows) for message in self.__messages)
         self.__metrics.increment("batch_write_msgs", rows)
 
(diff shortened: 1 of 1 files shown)


snuba/consumers/consumer.py
```diff
@@ -106,16 +106,38 @@ def close(self) -> None:
...
                 )
+
+        if max_latency is not None:
+            self.__metrics.timing("max_latency_ms", max_latency * 1000)
+            self.__metrics.timing(
+                "avg_latency_ms", (latency_sum / len(self.__messages)) * 1000
+            )
+        if max_end_to_end_latency is not None:
+            self.__metrics.timing(
+                "max_end_to_end_latency_ms", max_end_to_end_latency * 1000
```
nikhars, line 138: Do we care about measuring the average end to end latency when we replace the existing metrics? Or only the max?

snuba/consumers/consumer.py
```diff
@@ -106,16 +106,38 @@ def close(self) -> None:
...
+            )
+
         self.__metrics.timing("batch_write_ms", write_finish - write_start)
```
nikhars, line 141: Since you are here, could you fix the batch_write_ms metric so that all metrics are in `ms`
```suggestion
        self.__metrics.timing("batch_write_ms", (write_finish - write_start) * 1000)
```

nikhars (approved):

PR #2516, opened 2022-03-16
fix(metrics) - fix storage key for metrics_raw

onewland:
The key and value of the enum in StorageKey need to be identical. This caused an issue when trying to deploy an ops change to run the metrics_raw consumer (https://getsentry.atlassian.net/browse/OPS-1481)

Tested locally that `snuba multistorage-consumer --storage=metrics_counters_buckets --storage=metrics_distributions_buckets --storage=metrics_buckets --storage=metrics_raw --consumer-group=metrics_group` starts successfully


fpacifici (approved):

onewland:
verified that `make test` passed locally

PR #2520, opened 2022-03-16
fix(query): Long queries rejected at Snuba level

rahul-kumar-saini:
### Overview
- Currently when a query is too long for ClickHouse to process, we find out when we attempt to process it and ClickHouse tells us `Max query size exceeded.` We log this as an Error in Sentry
- With these changes Snuba will validate the size of the query and respond with a `400` (and an appropriate message) if the query size exceeds the max.

### Changes
- Introduced a `QueryTooLongException`
- Snuba API responds with a 400 when query is too long
- Got rid of `>=100%` query size group span tag in favor of `100%` as the maximum possible size group

### DIFF
--- a/snuba/web/__init__.py
+++ b/snuba/web/__init__.py
@@ -27,6 +27,12 @@ class QueryException(SerializableException):
         self.extra = extra
 
 
+class QueryTooLongException(SerializableException):
+    """
+    Exception thrown when a query is too long for ClickHouse
+    """
+
+
 class QueryResult(NamedTuple):
     result: Result
     extra: QueryExtraData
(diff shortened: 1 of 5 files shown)


snuba/web/__init__.py
```diff
@@ -27,6 +27,12 @@ def __init__(self, extra: QueryExtraData):
         self.extra = extra
 
 
+class QueryTooLongException(SerializableException):
```
onewland, line 30: Is this naming a bit ambiguous? I first read it as "query run-time too long" and when I read "long queries" (as in the PR title) I sort of mentally translate it to "slow queries".

Don't let this block if nobody agrees

volokluev (approved):

onewland (approved):
that get_query_size_group logic seems like a very weird way to do things to me but this is fine

snuba/web/__init__.py
rahul-kumar-saini: I see, I can change it to something like `QueryStringTooLongException`?

PR #2521, opened 2022-03-17
fix(snql) Remove invalid time sentry logging

evanh:
This error comes up frequently and has not been an issue in Snuba for a long
time. Stop logging this error to Sentry and just return a 400.

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -1289,7 +1289,9 @@ def _align_max_days_date_align(
     )
     to_date = to_date - timedelta(seconds=(to_date - to_date.min).seconds % date_align)
     if from_date > to_date:
-        raise ParsingException(f"invalid time conditions on entity {key.value}")
+        raise ParsingException(
+            f"invalid time conditions on entity {key.value}", should_report=False
+        )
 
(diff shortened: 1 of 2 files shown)


rahul-kumar-saini:
> Stop logging this error to Sentry and just return a 400.

Just curious, this is a "Stop reporting to Sentry but _continue_ returning 400" and not "Stop reporting to Sentry and _start_ returning 400", correct?

> has not been an issue in Snuba for a long time

What does this mean, does it not happen much anymore or we just don't care about it anymore?

lynnagara:
@rahul-kumar-saini  I think it implies there was once an issue with this in Snuba but now it's just the client (Sentry) sending us invalid data

lynnagara (approved):

evanh:
> Just curious, this is a "Stop reporting to Sentry but _continue_ returning 400" and not "Stop reporting to Sentry and _start_ returning 400", correct?

It means stop reporting to Sentry. Nothing else changes.
 
> What does this mean, does it not happen much anymore or we just don't care about it anymore?

It is not a Snuba issue any more. We tracked it initially to ensure we were parsing the time conditions correctly but for a while now all the issues have been from the product side.

PR #2507, opened 2022-03-14
feat(metrics) - create materialized views for polymorphic input tables

onewland:
This creates the 3 materialized views necessary for the polymorphic metrics input table.

I tested locally that this creates correct values in the aggregate tables for all 3 types.

If people don't find this to be "interesting" or useful enough on its own, I can do a PR that also includes the application processing logic and therefore uses that polymorphic table end to end

Follow-up of https://github.com/getsentry/snuba/pull/2504

The storage writer that lives on top of this: https://github.com/getsentry/snuba/pull/2510

PR #2504, opened 2022-03-11
feat(metrics) - create polymorphic bucket table

onewland:
Had to create a new PR because Github wasn't recognizing new commits

Old PR, same changes except for table rename: https://github.com/getsentry/snuba/pull/2494

```sql
CREATE TABLE IF NOT EXISTS metrics_raw_local (use_case_id LowCardinality(String), org_id UInt64, project_id UInt64, metric_id UInt64, timestamp DateTime, tags Nested(key UInt64, value UInt64), metric_type LowCardinality(String), set_values Array(UInt64), count_value Float64, distribution_values Array(Float64), materialization_version UInt8, retention_days UInt16, partition UInt16, offset UInt64) 
ENGINE MergeTree() 
ORDER BY (metric_type, org_id, project_id, metric_id, timestamp) 
PARTITION BY (use_case_id, toStartOfDay(timestamp)) 
TTL timestamp + toIntervalDay(7)
SETTINGS ttl_only_dropparts = 1; -- added by hand
```

PR #2524, opened 2022-03-18
feat(metrics) Increase the partition size of the raw metrics table

fpacifici:
Daily partition is appealing if most data is concentrated in the latest partition. Unfortunately it seems the content of an INSERT is more evenly divided across partitions. This generates one part per partition almost at every batch, which creates 5 parts every time increasing the number of inactive parts instead of reducing 
```
┌─partition───────────┬─active─┬───c─┬─formatReadableSize(sum(bytes_on_disk))─┐
│ 2022-03-12 00:00:00 │      1 │   4 │ 7.48 MiB                               │
│ 2022-03-13 00:00:00 │      0 │ 621 │ 72.55 MiB                              │
│ 2022-03-13 00:00:00 │      1 │  11 │ 65.22 MiB                              │
│ 2022-03-14 00:00:00 │      0 │ 810 │ 471.11 MiB                             │
│ 2022-03-14 00:00:00 │      1 │  10 │ 97.29 MiB                              │
│ 2022-03-15 00:00:00 │      0 │ 701 │ 278.61 MiB                             │
│ 2022-03-15 00:00:00 │      1 │   7 │ 175.70 MiB                             │
│ 2022-03-16 00:00:00 │      0 │ 541 │ 85.79 MiB                              │
│ 2022-03-16 00:00:00 │      1 │  11 │ 487.36 MiB                             │
│ 2022-03-17 00:00:00 │      0 │ 529 │ 200.09 MiB                             │
│ 2022-03-17 00:00:00 │      1 │  14 │ 14.73 GiB                              │
│ 2022-03-18 00:00:00 │      0 │ 523 │ 814.98 MiB                             │
│ 2022-03-18 00:00:00 │      1 │  16 │ 5.69 GiB                               │
└─────────────────────┴────────┴─────┴────────────────────────────────────────┘
```

So this pr makes the partitioning key more coarse. We try a 3 days partition.
This PR is almost identical to 
https://github.com/getsentry/snuba/pull/2504
https://github.com/getsentry/snuba/pull/2507
except for the partition key and the table names.

cc @onewland

### DIFF
--- a/snuba/migrations/snuba_migrations/metrics/0022_repartition_polymorphic_table.py
+++ b/snuba/migrations/snuba_migrations/metrics/0022_repartition_polymorphic_table.py
@@ -1,0 +1,76 @@
+from typing import Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.utils.schemas import Array, Column, DateTime, Float, Nested, String, UInt
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Create a polymorphic bucket table for metrics_sets,
+    metrics_distributions, and metrics_counters.
+    This is partitioned by 3 days intervals.
+    """
+
+    blocking = False
+    local_table_name = "metrics_raw_v2_local"
+    dist_table_name = "metrics_raw_v2_dist"
+
+    column_list: Sequence[Column[Modifiers]] = [
+        Column("use_case_id", String(Modifiers(low_cardinality=True))),
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("timestamp", DateTime()),
+        Column("tags", Nested([Column("key", UInt(64)), Column("value", UInt(64))]),),
+        Column("metric_type", String(Modifiers(low_cardinality=True))),
+        Column("set_values", Array(UInt(64))),
+        Column("count_value", Float(64)),
+        Column("distribution_values", Array(Float(64))),
+        Column("materialization_version", UInt(8)),
+        Column("retention_days", UInt(16)),
+        Column("partition", UInt(16)),
+        Column("offset", UInt(64)),
+    ]
+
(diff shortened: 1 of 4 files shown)


onewland (approved):
Looks good to me, but is there a reason the partition size shouldn't be 5 or 7 days? It seems like there's a decent amount of 3-14 activity received on 3-18.

Also you and Markus have a conflict in the numbering of the migration if we care about that: https://github.com/getsentry/snuba/pull/2525

onewland:
Also, not as part of this PR but I wonder if we should emit a [sampled] metric for (message timestamp - system time), so if we see that go very negative we can try to address that with the product team or another partition change

fpacifici:
> Looks good to me, but is there a reason the partition size shouldn't be 5 or 7 days? It seems like there's a decent amount of 3-14 activity received on 3-18.

It seemed a reasonable compromise. 3 days means we are going to write max 2 parts per insert, which is common in all clusters we have. 5 would have been one but then we would have had to keep data around up to 15 days before dropping partitions. This way it is only 9

snuba/migrations/snuba_migrations/metrics/0022_repartition_polymorphic_table.py
```diff
@@ -0,0 +1,76 @@
...
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=self.local_table_name,
+                columns=self.column_list,
+                engine=table_engines.MergeTree(
+                    storage_set=StorageSetKey.METRICS,
+                    order_by="(use_case_id, metric_type, org_id, project_id, metric_id, timestamp)",
+                    partition_by="(toStartOfInterval(timestamp, INTERVAL 3 day))",
```
nikhars, line 46: Question: Do we need to create new tables every time the partition by column changes? What factors made you decide that we should create a new table vs repartitioning the existing table?

onewland: In this case I think it makes more sense to create a new table vs. rewriting data from the old one because we never need to access the old data.

PR #2525, opened 2022-03-18
feat(metrics): Add histogram(250) column to distribution metrics [INGEST-947]

untitaker:
Add a way to query histograms. This requires rollout of polymorphic raw table if that hasn't already happened. I purposefully didn't touch [DistributionsAggregateProcessor](https://github.com/getsentry/snuba/blob/51acb50abb009602f3b42f5a95062e6d99a699f3/snuba/datasets/metrics_aggregate_processor.py#L147) since that appears to be doing something deprecated to aggregate tables that is probably succeeded by the polymorphic raw table.

I was picking 250 as bucket count, the max allowed value, since current product conversations around zooming indicate we will need all the buckets we can get from ClickHouse.


snuba/migrations/snuba_migrations/metrics/0006_metrics_distributions.py
```diff
@@ -2,7 +2,7 @@
 
 from snuba.migrations import migration, operations
 from snuba.migrations.snuba_migrations.metrics.templates import (
-    COL_SCHEMA_DISTRIBUTIONS,
+    COL_SCHEMA_DISTRIBUTIONS_V3,
```
onewland, line 5: as a general rule we try not to modify old migrations, can we just have the version suffix in the new column schema

snuba/migrations/snuba_migrations/metrics/templates.py
```diff
@@ -347,8 +356,12 @@ def get_consolidated_mv_name(metric_type: str) -> str:
     return f"metrics_{metric_type}_consolidated_mv_local"
 
 
-def get_polymorphic_mv_name(metric_type: str) -> str:
-    return f"metrics_{metric_type}_polymorphic_mv_local"
+def get_polymorphic_mv_name(metric_type: str, materialization_version: int) -> str:
```
onewland, line 359: similar to my comment above, I'd rather you make a new method called `get_versioned_polymorphic_mv_name` and leave this method definition static. I don't think you changed anything but it could be very confusing if you inadvertently did

onewland (changes requested):
Mostly fine, but let's refactor so we're not rewriting existing migrations -- also I'm not sure who will merge first but you and Filippo both have a migration numbered 22: https://github.com/getsentry/snuba/pull/2524

snuba/migrations/snuba_migrations/metrics/0021_polymorphic_bucket_materialized_views.py
```diff
@@ -36,6 +36,7 @@ def forwards_local(self) -> Sequence[operations.SqlOperation]:
                     "countState(values_rows) as count"
                 ),
                 metric_type="distribution",
+                materialization_version=3,
```
untitaker, line 39: @onewland I think this also goes against the rule then? But splitting up that function seems quite extreme.

snuba/migrations/snuba_migrations/metrics/templates.py
```diff
@@ -351,6 +361,16 @@ def get_polymorphic_mv_name(metric_type: str) -> str:
...
 
 
+def get_versioned_polymorphic_mv_name(
+    metric_type: str, materialization_version: int
+) -> str:
+    assert materialization_version >= 3
+    materialization_version_str = (
+        "" if materialization_version == 3 else f"_v{materialization_version}"
+    )
+    return f"metrics_{metric_type}{materialization_version_str}_polymorphic_mv_local"
```
evanh, line 371: Out of curiosity, what is the migration plan to start using the new materialized view? And do we drop the old views?

untitaker: the way oliver described it, materialization_version only defines which query is used to insert data into the aggregate table. we still keep the same aggregate table, we just swap out the MV that writes from raw -> aggregate. so essentially there's no data migration here. we run this migration, bump the active materialization_version and should start seeing histograms for new data.

snuba/migrations/snuba_migrations/metrics/0021_polymorphic_bucket_materialized_views.py
untitaker: I am happy to do whatever y'all think should be done. I personally think shared modules like `templates.py` don't make a ton of sense to have in the first place if we want to avoid accidentally breaking old migrations at cost of DRY.

untitaker:
<details><summary>migration sql</summary>

```
Local operations:
ALTER TABLE metrics_distributions_local ADD COLUMN IF NOT EXISTS histogram_buckets AggregateFunction(histogram(250), Float64) AFTER count;
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_distributions_polymorphic_mv_v3_local TO metrics_distributions_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, percentiles AggregateFunction(quantiles(0.5, 0.75, 0.9, 0.95, 0.99), Float64), min AggregateFunction(min, Float64), max AggregateFunction(max, Float64), avg AggregateFunction(avg, Float64), sum AggregateFunction(sum, Float64), count AggregateFunction(count, Float64), histogram AggregateFunction(histogram(250), Float64)) AS
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)((arrayJoin(distribution_values) AS values_rows)) as percentiles, minState(values_rows) as min, maxState(values_rows) as max, avgState(values_rows) as avg, sumState(values_rows) as sum, countState(values_rows) as count, histogramState(250)(values_rows) as histogram_buckets
FROM metrics_raw_local
WHERE materialization_version = 4
  AND metric_type = 'distribution'
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;


Dist operations:
```
</details>

snuba/migrations/snuba_migrations/metrics/0024_metrics_distributions_add_histogram.py
```diff
@@ -0,0 +1,81 @@
...
+)
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Add HdrHistogram to distribution metrics.
+    """
+
+    blocking = False
+    raw_table_name = "metrics_raw_local"
```
untitaker, line 19: I see that in migration 0023 we are referencing metrics_raw_v2_local. Is that the new one I should use?

snuba/migrations/snuba_migrations/metrics/0024_metrics_distributions_add_histogram.py
```diff
@@ -0,0 +1,81 @@
...
+                storage_set=StorageSetKey.METRICS,
+                table_name=table_name,
+                column_name="histogram_buckets",
+            )
+        ]
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            *self.__forward_migrations("metrics_distributions_local"),
+            get_forward_view_migration_polymorphic_table_v2(
```
untitaker, line 50: @onewland pointed out that bumping the enabled materialization version will stop writing into the aggregation tables for everything _but_ distributions. We need to re-create all materialized views for metrics here.

snuba/migrations/snuba_migrations/metrics/0024_metrics_distributions_add_histogram.py
onewland: yes, it is

snuba/migrations/snuba_migrations/metrics/0024_metrics_distributions_add_histogram.py
onewland: yeah, it's a bit of a pain but that's probably the cleanest approach

fpacifici: We should not need this. 
So, if you want to recreate all the matviews, there has to be a migration with the three. 
Like this https://github.com/getsentry/snuba/pull/2524/files#diff-255f09f391ed726eb0533354f3dc4836bb4c1d66958fb1b06a0e118788fc15c0R1

In alternative, you can change the consumer and write a different version per metric type and only create a new matview for distribution that has a different version from the others.

Considering we are still not in prod and making change every day, up to you. The only critical element is that the migrations need to stay consistent and that you need to be able to achieve a stable state when running all with no manual intervention.

fpacifici (commented):
Please ensure to fix the raw_table_name before merging

snuba/migrations/snuba_migrations/metrics/templates.py
```diff
@@ -174,6 +179,31 @@ def get_forward_bucket_table_dist(
...
+    project_id,
+    metric_id,
+    arrayJoin([10,60,3600,86400]) as granularity,
+    tags.key,
+    tags.value,
+    toDateTime(granularity * intDiv(toUnixTimestamp(timestamp), granularity)) as timestamp,
+    retention_days,
+    %(aggregation_states)s
+FROM %(raw_table_name)s
+WHERE materialization_version = 4
```
fpacifici, line 194: Oh wait, the only difference here is the materialization_version? 
Sorry I misunderstood the problem this morning before reading this. 
If the matview template is constant but the only thing we change is the materialization version I think it would make sense to use the template. This is the same change we will make over and over again. 
Basically we already do that for the `%(raw_table_name)s` one line above.

I thought we were making actual functional changes to the matview that would not be backwards compatible. 

I suggest we turn this line into `materialization_version = %(materialization_version)s` on the template. We can create a new template to avoid changing the existing migrations (in hindsight we should have made materialization_version a parameter to begin with).

untitaker:
<details><summary>migration sql</summary>

```
Local operations:
ALTER TABLE metrics_distributions_local ADD COLUMN IF NOT EXISTS histogram_buckets AggregateFunction(histogram(250), Float64) AFTER count;
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_distributions_polymorphic_mv_v3_local TO metrics_distributions_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, percentiles AggregateFunction(quantiles(0.5, 0.75, 0.9, 0.95, 0.99), Float64), min AggregateFunction(min, Float64), max AggregateFunction(max, Float64), avg AggregateFunction(avg, Float64), sum AggregateFunction(sum, Float64), count AggregateFunction(count, Float64), histogram AggregateFunction(histogram(250), Float64)) AS 
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)((arrayJoin(distribution_values) AS values_rows)) as percentiles, minState(values_rows) as min, maxState(values_rows) as max, avgState(values_rows) as avg, sumState(values_rows) as sum, countState(values_rows) as count, histogramState(250)(values_rows) as histogram_buckets
FROM metrics_raw_v2_local
WHERE materialization_version = 4
  AND metric_type = 'distribution'
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_sets_polymorphic_mv_v3_local TO metrics_sets_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(uniqCombined64, UInt64)) AS 
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    uniqCombined64State(arrayJoin(set_values)) as value
FROM metrics_raw_v2_local
WHERE materialization_version = 4
  AND metric_type = 'set'
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_counters_polymorphic_mv_v3_local TO metrics_counters_local (org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(sum, Float64)) AS 
SELECT
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    sumState(count_value) as value
FROM metrics_raw_v2_local
WHERE materialization_version = 4
  AND metric_type = 'counter'
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;


Dist operations:
```
</detail>

PR #2530, opened 2022-03-18
feat(metrics) Switch table name to use the longer partition one

fpacifici:
Swaps metrics_raw_local with metrics_raw_v2_local which has the longer partition size


fpacifici:
Duplicate of https://github.com/getsentry/snuba/pull/2530

PR #2532, opened 2022-03-18
ref: Increase the log level during subscription consumer shutdown

lynnagara:
We need to better understand what is happening during the shutdown
sequence of the subscriptions consumer. Increase the log level
to debug.

### DIFF
--- a/snuba/cli/subscriptions.py
+++ b/snuba/cli/subscriptions.py
@@ -260,6 +260,10 @@ def subscriptions(
         )
 
         def handler(signum: int, frame: Optional[Any]) -> None:
+            # TODO: Temporary code for debugging the shutdown sequence of the subscriptions
+            # consumer without updating arroyo or affecting other consumers.
+            logging.getLogger().setLevel(logging.DEBUG)
+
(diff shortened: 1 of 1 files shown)


volokluev (approved):

snuba/cli/subscriptions.py
```diff
@@ -260,6 +260,10 @@ def subscriptions(
         )
 
         def handler(signum: int, frame: Optional[Any]) -> None:
+            # TODO: Temporary code for debugging the shutdown sequence of the subscriptions
+            # consumer without updating arroyo or affecting other consumers.
+            logging.getLogger().setLevel(logging.DEBUG)
```
fpacifici, line 265: I'd assume you will leave this running in prod for a quite short time. THere are quite a few debug log statement in the query processing, so you are going to log something for each query you run.

lynnagara: Yeah it should not be for long. It's possible it'll exacerbate the problem where shutdown takes too long. But it may help shed some light so I'd like to give a shot.

fpacifici (approved):

PR #2535, opened 2022-03-21
feat(optimize) Ensure the optimize job does not go past 11 pm

fpacifici:
We saw a few times the optimize job being restarted by Kubernetes after a few hours. 
That job is stateless so it restarts from scratch. If k8s restarts it too late, the job may end so late to overlap with the next day instance. This puts unnecessary load on the DB.

If we are about to start a new instance there is no point in finishing the previous one, so this adds a cutoff time to ensure the job does not go past a certain time of the dat provided in settings.

The proper fix would be to make the job stateful so that it picks up from where it left when restarted, though that is a large refactoring. This is at least a safeguard to prevent the overlap.

### DIFF
--- a/snuba/cli/optimize.py
+++ b/snuba/cli/optimize.py
@@ -29,7 +29,6 @@ def optimize(
     clickhouse_host: Optional[str],
     clickhouse_port: Optional[int],
     storage_name: str,
-    ignore_cutoff: Optional[bool] = False,
     log_level: Optional[str] = None,
 ) -> None:
     from datetime import datetime
--- a/snuba/cli/optimize.py
+++ b/snuba/cli/optimize.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from typing import Optional
 
 import click
@@ -80,7 +81,11 @@ def optimize(
         )
 
     if not ignore_cutoff:
-        last_midnight = datetime.now().replace(
+        # Adding 10 minutes to the current time before finding the midnight time
+        # to ensure this keeps working even if the system clock of the host that
+        # starts the pod is slightly ahead of the system clock of the host running
+        # the job. This prevents us from getting the wrong midnight.
+        last_midnight = (datetime.now() + timedelta(minutes=10)).replace(
             hour=0, minute=0, second=0, microsecond=0
         )
         cutoff: Optional[datetime] = last_midnight + settings.OPTIMIZE_JOB_CUTOFF_TIME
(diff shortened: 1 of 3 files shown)


snuba/cli/optimize.py
```diff
@@ -71,5 +80,21 @@ def optimize(
...
-    num_dropped = run_optimize(connection, storage, database, before=today)
+    if not ignore_cutoff:
+        # Adding 10 minutes to the current time before finding the midnight time
+        # to ensure this keeps working even if the system clock of the host that
+        # starts the pod is slightly ahead of the system clock of the host running
+        # the job. This prevents us from getting the wrong midnight.
+        last_midnight = (datetime.now() + timedelta(minutes=10)).replace(
+            hour=0, minute=0, second=0, microsecond=0
+        )
+        cutoff: Optional[datetime] = last_midnight + settings.OPTIMIZE_JOB_CUTOFF_TIME
```
volokluev, line 91: is there a need to plumb this through the whole stack? if we're doing the work of calculating what the cuttoff is here, why not enforce it here as well?

fpacifici: we cannot. 
The goal is not to prevent the entire job from starting past 11 pm. The goal is to ensure that a running job stops by 11 pm in case it is that time and it still has work to do. 
That can only be enforced in the `optimize_partitions` call, because there is when we execute the `OPTIMIZE` statement for each partition one by one. That function takes hours to execute as it runs each OPTIMIZE. So it is the place where to pending work if it is 11 pm

lynnagara:
Can we just pass the optional cutoff time via the CLI rather than achieve it through a combination of settings and CLI args?

fpacifici:
@lynnagara  mostly because I did not want to add the complexity of parsing and validating time intervals from the CLI. Specifically considering that having a cutoff should be the default value. Disabling it should be only for exceptional cases (like testing stuff). I would actually prefer removing the flag entirely and making the behavior always on.

fpacifici:
I removed the flag to disable the cutoff time

fpacifici:
Actually committing at times help.
@lynnagara, now it is gone

snuba/cli/optimize.py
```diff
@@ -29,6 +29,7 @@ def optimize(
     clickhouse_host: Optional[str],
     clickhouse_port: Optional[int],
     storage_name: str,
+    ignore_cutoff: Optional[bool] = False,
```
nikhars, line 32: You can get rid of this now since its not being used.

volokluev (approved):

PR #2537, opened 2022-03-21
feat(dlq): Extend Count policy to save to/load from Redis

rahul-kumar-saini:
Added a new DLQ Policy which extends the Count policy and adds functionality to save the hits in Redis and to be able to load the state upon instantiation.


snuba/state/stateful_count.py
```diff
@@ -0,0 +1,60 @@
...
+        self.__name = redis_hash_name
+        self.__seconds = seconds
+        super().__init__(limit, seconds, self._load_state())
+
+    def handle_invalid_message(self, e: InvalidMessage) -> None:
+        """
+        Asynchronously updates Redis hash with a new hit while
+        the base Policy handles the invalid message
+        """
+        Thread(target=self._add_to_redis).start()
```
rahul-kumar-saini, line 30: Doing this in a new thread since the network calls to Redis don't have any impact to the logic. I'm not sure if this is the right way or if there are any implications of doing this.

rahul-kumar-saini (commented):
Using a Redis Hash instead of a Sorted Set since the values (# of hits in time bucket) will not be unique. There is a better way of doing this using the new Redis TimeSeries type but that is only available in Redis >= 4.0.0.

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class StatefulCountInvalidMessagePolicy(CountInvalidMessagePolicy[TPayload]):
+    """
+    An extension of the CountInvalidMessagePolicy which is able to save and load
+    the state of counted hits in Redis
+    """
+
+    def __init__(self, redis_hash_name: str, limit: int, seconds: int = 60) -> None:
+        self.__name = redis_hash_name
```
MeredithAnya, line 21: just curious is the `redis_hash_name` going to be set by the consumer? like will different consumers have different hash names?

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Increments the current time entry in Redis by 1
+        """
+        now = int(time())
+        p = redis_client.pipeline()
+        p.hincrby(self.__name, str(now), 1)
+        p.hkeys(self.__name)
+        self._prune(now, p.execute()[1])
+
+    def _prune(self, now: int, timestamps: Sequence[Any]) -> None:
```
MeredithAnya, line 43: are the timestamps here also `int`s?

snuba/state/stateful_count.py
rahul-kumar-saini: Yeah each consumer would have its own DLQ, which means a policy would be instantiated per consumer. The name of the hash storing the counts would be set wherever the consumer is instantiated with a DLQ.

snuba/state/stateful_count.py
rahul-kumar-saini: Technically anything Redis returns is a `bytes` type, so it'd look something like `b'123'` which could be casted back into an `int`, but it's not necessary since those bytes can be compared with each other.

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,60 @@
...
+        self.__seconds = seconds
+        super().__init__(limit, seconds, self._load_state())
+
+    def handle_invalid_message(self, e: InvalidMessage) -> None:
+        """
+        Asynchronously updates Redis hash with a new hit while
+        the base Policy handles the invalid message
+        """
+        Thread(target=self._add_to_redis).start()
+        super().handle_invalid_message(e)
```
evanh, line 31: Where does your code actually check the current count to decide if this should throw or not?

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def _add_to_redis(self) -> None:
+        """
+        Increments the current time entry in Redis by 1
+        """
+        now = int(time())
+        p = redis_client.pipeline()
+        p.hincrby(self.__name, str(now), 1)
+        p.hkeys(self.__name)
```
lynnagara, line 40: What is this line for?

snuba/state/stateful_count.py
lynnagara: If it's a `bytes` type, why not annotate it as `bytes` instead of `Any`. I would probably still cast it back into int to avoid unexpected things like the example Evan gave.

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,60 @@
...
+        self._prune(now, p.execute()[1])
+
+    def _prune(self, now: int, timestamps: Sequence[Any]) -> None:
+        """
+        Removes old timestamps from the Redis hash
+        """
+        oldest_time = str(now - self.__seconds).encode("UTF-8")
```
lynnagara, line 47:
```suggestion
        oldest_time = str(now - self.__seconds).encode("utf-8")
```

tests/state/test_stateful_count_policy.py
```diff
@@ -0,0 +1,70 @@
...
+        redis_client.delete(NAME)
+
+    def test_simple(
+        self, policy: StatefulCountInvalidMessagePolicy[KafkaPayload]
+    ) -> None:
+        now = int(time.time())
+        policy.handle_invalid_message(INVALID_MESSAGE_EXCEPTION)
+        # waiting for thread
+        time.sleep(0.1)
+        assert redis_client.hgetall(NAME) == {str(now).encode("UTF-8"): b"1"}
```
lynnagara, line 38:
```suggestion
        assert redis_client.hgetall(NAME) == {str(now).encode("utf-8"): b"1"}
```

snuba/state/stateful_count.py
rahul-kumar-saini: It's to retrieve all timestamps to prune old ones. I could have just done `redis_client.hincrby()` in this method and then `redis_client.hkeys()` in the prune method but this saves 1 network call to Redis since it's done in the pipeline.

rahul-kumar-saini:
> What's the reason for putting `StatefulCountInvalidMessagePolicy` in Snuba and not arroyo?

Redis is only accessible in Snuba and not Arroyo, the [base class](https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/dead_letter_queue/policies/count.py) does this counting anyway but doesn't add anything to Redis like this policy does.

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,59 @@
...
+from arroyo.processing.strategies.dead_letter_queue import CountInvalidMessagePolicy
+from arroyo.processing.strategies.dead_letter_queue.policies.abstract import (
+    InvalidMessage,
+)
+from arroyo.types import TPayload
+
+from snuba.redis import redis_client
+
+
+class StatefulCountInvalidMessagePolicy(CountInvalidMessagePolicy[TPayload]):
```
lynnagara, line 13: I'm not sure I get why this needs to be generic. It seems to only deal with `InvalidMessage` errors and never the message payload itself.

tests/state/test_stateful_count_policy.py
```diff
@@ -0,0 +1,69 @@
...
+        assert redis_client.hgetall(SAVED_NAME) == {str(now).encode("utf-8"): b"1"}
+
+    def test_past_limit(
+        self, policy: StatefulCountInvalidMessagePolicy[KafkaPayload],
+    ) -> None:
+        # limit is 5, should raise on 6th invalid message
+        for _ in range(5):
+            policy.handle_invalid_message(INVALID_MESSAGE_EXCEPTION)
+        with pytest.raises(InvalidMessage):
+            policy.handle_invalid_message(INVALID_MESSAGE_EXCEPTION)
```
lynnagara, line 46: This tests only covers the functionality of the the base class. I think you should at least add an assertion about the redis state.

tests/state/test_stateful_count_policy.py
```diff
@@ -0,0 +1,69 @@
...
+from snuba.state.stateful_count import StatefulCountInvalidMessagePolicy
+
+NAME = "consumer-group"
+SAVED_NAME = f"dlq:{NAME}"
+INVALID_MESSAGE_EXCEPTION = InvalidMessage(
+    Message(Partition(Topic(""), 0), 0, KafkaPayload(None, b"", []), datetime.now())
+)
+
+
+class TestStatefulCountPolicy:
```
lynnagara, line 21: You should have a test case covering the prune functionality too

snuba/state/stateful_count.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    def __init__(self, consumer_group_name: str, limit: int, seconds: int = 60) -> None:
+        self.__name = f"dlq:{consumer_group_name}"
+        self.__seconds = seconds
+        super().__init__(limit, seconds, self._load_state())
+
+    def handle_invalid_message(self, e: InvalidMessage) -> None:
+        """
+        Asynchronously updates Redis hash with a new hit while
```
lynnagara, line 26: Can you update this comment now the thread got removed?

rahul-kumar-saini: I removed it in general since the method is pretty simple

snuba/state/stateful_count.py
lynnagara: @rahul-kumar-saini I think this question still stands. Why are any of the policies generic or would take a message payload?

rahul-kumar-saini: You're right, I didn't realize Policy doesn't need a generic payload anymore:

https://github.com/getsentry/arroyo/pull/49

lynnagara (approved):

PR #2539, opened 2022-03-22
feat(metrics): create new metrics_counters aggregate table with updated granularity …

onewland:
…and use-case ID column


snuba/migrations/snuba_migrations/metrics/0025_metrics_counters_aggregate_v2.py
```diff
@@ -0,0 +1,99 @@
...
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=self.table_name,
+                columns=self.aggregated_cols,
+                engine=table_engines.AggregatingMergeTree(
+                    storage_set=StorageSetKey.METRICS,
+                    order_by="(org_id, project_id, metric_id, granularity, timestamp, tags.key, tags.value)",
```
fpacifici, line 34: We need to separate the ORDER BY key from the Primary key

Like this 
```
PRIMARY KEY (org_id, project_id, metric_id, granularity, timestamp)
ORDER BY (org_id, project_id, metric_id, granularity, timestamp, tags.key, tags.value)
```

This ensures we store value uniquely by ORDER BY and sorted that way but we include only `org_id, project_id, metric_id, granularity, timestamp` in the index, which reduces the index size.

Also, please add the use_case_id to the ORDER BY key as well (either before or after tags) but not to the PRIMARY KEY. The partition key should be enough to prevent Clickhosue from merging rows of different retentions but we never know.

snuba/migrations/snuba_migrations/metrics/0025_metrics_counters_aggregate_v2.py
```diff
@@ -0,0 +1,99 @@
...
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.snuba_migrations.metrics.templates import COMMON_AGGR_COLUMNS
+from snuba.utils.schemas import AggregateFunction, Float, String
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    table_name = "metrics_counters_v2_local"
+    dist_table_name = "metrics_counters_v2_dist"
+    granularity = "1024"
```
fpacifici, line 16: Let's 2048 first ?

snuba/migrations/snuba_migrations/metrics/0025_metrics_counters_aggregate_v2.py
```diff
@@ -0,0 +1,99 @@
...
+        # but doesn't exist in the old templates so we include it here
+        Column("use_case_id", String(Modifiers(low_cardinality=True))),
+        *COMMON_AGGR_COLUMNS,
+        Column("value", AggregateFunction("sum", [Float(64)])),
+    ]
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+
+        return [
+            operations.CreateTable(
```
fpacifici, line 28: This is missing the TTL.

fpacifici (changes requested):
A couple of changes. Please see comments inline

onewland:
New SQL:
```
Local operations:
CREATE TABLE IF NOT EXISTS metrics_counters_v2_local (use_case_id LowCardinality(String), org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(sum, Float64)) ENGINE AggregatingMergeTree()  PRIMARY KEY (org_id, project_id, metric_id, granularity, timestamp) ORDER BY (org_id, project_id, metric_id, granularity, timestamp, use_case_id, tags.key, tags.value) PARTITION BY (retention_days, toMonday(timestamp)) TTL timestamp + toIntervalDay(retention_days) SETTINGS index_granularity=2048;
ALTER TABLE metrics_counters_v2_local ADD COLUMN IF NOT EXISTS _tags_hash Array(UInt64) MATERIALIZED arrayMap((k, v) -> cityHash64(concat(toString(k), '=', toString(v))), tags.key, tags.value) AFTER tags.value;
ALTER TABLE metrics_counters_v2_local ADD INDEX IF NOT EXISTS bf_tags_hash _tags_hash TYPE bloom_filter() GRANULARITY 1;
ALTER TABLE metrics_counters_v2_local ADD INDEX IF NOT EXISTS bf_tags_key_hash tags.key TYPE bloom_filter() GRANULARITY 1;
```

onewland:
Addressed all comments AFAIK

snuba/migrations/snuba_migrations/metrics/0025_metrics_counters_aggregate_v2.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=self.table_name,
+                columns=self.aggregated_cols,
+                engine=table_engines.AggregatingMergeTree(
+                    storage_set=StorageSetKey.METRICS,
+                    order_by="(org_id, project_id, metric_id, granularity, timestamp, use_case_id, tags.key, tags.value)",
```
fpacifici, line 33: retention is still missing.
I thought `use_case_id` would be the first in the key. 
If we start using use_case_id we will always filter by that, so we should put it both in the order by and the primary key. 

The fields we have in the order by but not in the primary key are not part of the index, so we do not get any advantage there

onewland: I'm fine with use_case_id being part of the `primary key` and `order by`, but was this comment above incorrect?

> Also, please add the use_case_id to the ORDER BY key as well (either before or after tags) but not to the PRIMARY KEY. The partition key should be enough to prevent Clickhosue from merging rows of different retentions but we never know.

And can you explain in more depth what `retention is still missing.` means? You want `retention_days` in the ORDER BY so that rows with different timestamps can't be merged (even though they should always be partitioned separately)?

fpacifici:
> And can you explain in more depth what retention is still missing. means? You want retention_days in the ORDER BY so that rows with different timestamps can't be merged (even though they should always be partitioned separately)?

Yes, two buckets have different retention_days value but the same values for all the other fields in the order by key, they will be merged together. I don't know what the resulting retention will be. 

Now we are also partitioning by retention, so this should never happen in practice because each partition can only contain one retention value, though that makes the data model less sound, it would be hard to notice the mistake if we change the partitioning and remove retention_days from there.

onewland:
Added use_case_id to PRIMARY KEY and ORDER BY
Added retention_days to the end of ORDER BY

SQL:
```
Local operations:
CREATE TABLE IF NOT EXISTS metrics_counters_v2_local (use_case_id LowCardinality(String), org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(sum, Float64)) ENGINE AggregatingMergeTree() PRIMARY KEY (use_case_id, org_id, project_id, metric_id, granularity, timestamp) ORDER BY (use_case_id, org_id, project_id, metric_id, granularity, timestamp, tags.key, tags.value, retention_days) PARTITION BY (retention_days, toMonday(timestamp)) TTL timestamp + toIntervalDay(retention_days) SETTINGS index_granularity=2048;
ALTER TABLE metrics_counters_v2_local ADD COLUMN IF NOT EXISTS _tags_hash Array(UInt64) MATERIALIZED arrayMap((k, v) -> cityHash64(concat(toString(k), '=', toString(v))), tags.key, tags.value) AFTER tags.value;
ALTER TABLE metrics_counters_v2_local ADD INDEX IF NOT EXISTS bf_tags_hash _tags_hash TYPE bloom_filter() GRANULARITY 1;
ALTER TABLE metrics_counters_v2_local ADD INDEX IF NOT EXISTS bf_tags_key_hash tags.key TYPE bloom_filter() GRANULARITY 1;


Dist operations:
```

fpacifici (approved):

PR #2540, opened 2022-03-22
feat(metrics) write to new counters_aggregate table

onewland:
Starts writing to metrics_counters_v2_local on `2021-03-24 00:00:00 UTC` (can be updated before merge, this is just a marker for backfilling which doesn't have to be super precise)

Assumes that materialization version is set to 4 following deploy of https://github.com/getsentry/snuba/pull/2525

Also assumes https://github.com/getsentry/snuba/pull/2539 is deployed and the table is created

### DIFF
--- a/snuba/migrations/snuba_migrations/metrics/0026_metrics_counters_v2_writing_matview.py
+++ b/snuba/migrations/snuba_migrations/metrics/0026_metrics_counters_v2_writing_matview.py
@@ -1,0 +1,50 @@
+from typing import Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.snuba_migrations.metrics.templates import (
+    get_forward_view_migration_polymorphic_table_v3,
+    get_polymorphic_mv_v3_name,
+)
+from snuba.utils.schemas import AggregateFunction, Column, Float
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Creates materialized view for all metrics types which writes to 10s, 1m, 1h, 1d granularities
+    The backward migration does *not* delete any data from the destination tables.
+    """
+
+    blocking = False
+    table_name = "metrics_counters_v2_local"
+    raw_table_name = "metrics_raw_v2_local"
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            get_forward_view_migration_polymorphic_table_v3(
+                source_table_name=self.raw_table_name,
+                table_name=self.table_name,
+                aggregation_col_schema=[
+                    Column("value", AggregateFunction("sum", [Float(64)])),
+                ],
+                mv_name=get_polymorphic_mv_v3_name("counters"),
+                aggregation_states="sumState(count_value) as value",
+                metric_type="counter",
+                target_mat_version=4,
+                appended_where_clause="AND timestamp > toDateTime('2021-03-24 00:00:00')",
+            )
+        ]
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.DropTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=get_polymorphic_mv_v3_name("counters"),
+            )
+        ]
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
+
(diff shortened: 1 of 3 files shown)


onewland:
SQL:
```
➜  snuba git:(write-to-new-counters-table) ✗ snuba migrations run --group metrics --migration-id 0026_metrics_counters_v2_writing_matview --dry-run
Local operations:
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_counters_polymorphic_mv_v3_local TO metrics_counters_v2_local (use_case_id LowCardinality(String), org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(sum, Float64)) AS 
SELECT
    use_case_id,
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    sumState(count_value) as value
FROM metrics_raw_v2_local
WHERE materialization_version = 4
  AND metric_type = 'counter'
  AND timestamp > toDateTime('2021-03-24 00:00:00')
GROUP BY
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;

```

onewland:
SQL:
```

**Comment by onewland** [comment]
> This isn't a blocker, but the versions of this are super confusing. You are writing to a `v2` table using a `v3` materialization that has a `target_mat_version=4`. Are these versions all for different things? Are they supposed to eventually line up?

Yeah, it's not great. These are versions of different things. Doing this much iteration via the migration system is a little painful and I think we should consider ways to experiment that only have us write a migration at the end of experimentation (this was my goal for https://www.notion.so/Migration-to-new-aggregate-table-s-781f9517a1b64b90bdf1b86c1714e222#c69635d80be24810a44aaac2795ed0e6 but that wasn't technically feasible)

1. This is v2 of the aggregate table we're writing to. This is the new table from PR #2539. Using that new table as the basis of queries to improve performance is the main functional goal here.
1. The "v3" materialization refers to this being created by the third definition of the materialization template. Maybe this could refer back to the aggregate table version and be something more semantic like `metrics_counters_polymorphic_mv_agg_v2_local` but even so there's an issue with all these having to 
1. The target_mat_version=4 is what the materialized view uses to determine which rows from the input tables to transform and write to the aggregate tables. Rows with materialization_version != 4 will be ignored and this lets us control via the code-base which versions of the materialized views we want to run against the input. The reason 4 was chosen for this PR is that we don't need a smooth migration so I'm just sticking with the prior version. This has functional impact so it's going to reasonably not be in sync with the other versions

File: snuba/migrations/snuba_migrations/metrics/0026_metrics_counters_v2_writing_matview.py [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+                source_table_name=self.raw_table_name,
+                table_name=self.table_name,
+                aggregation_col_schema=[
+                    Column("value", AggregateFunction("sum", [Float(64)])),
+                ],
+                mv_name=get_polymorphic_mv_v3_name("counters"),
+                aggregation_states="sumState(count_value) as value",
+                metric_type="counter",
+                target_mat_version=4,
+                appended_where_clause="AND timestamp > toDateTime('2021-03-24 00:00:00')",
```
fpacifici, line 34: We are in 2022 :D

fpacifici (approved):
Please fix the date

onewland:
Updated the date to be ~9AM PST this morning:
```
CREATE MATERIALIZED VIEW IF NOT EXISTS metrics_counters_polymorphic_mv_v3_local TO metrics_counters_v2_local (use_case_id LowCardinality(String), org_id UInt64, project_id UInt64, metric_id UInt64, granularity UInt32, tags Nested(key UInt64, value UInt64), timestamp DateTime, retention_days UInt16, value AggregateFunction(sum, Float64)) AS 
SELECT
    use_case_id,
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    retention_days,
    sumState(count_value) as value
FROM metrics_raw_v2_local
WHERE materialization_version = 4
  AND metric_type = 'counter'
  AND timestamp > toDateTime('2022-03-24 16:00:00')
GROUP BY
    use_case_id,
    org_id,
    project_id,
    metric_id,
    tags.key,
    tags.value,
    timestamp,
    granularity,
    retention_days
;
```

PR #2542, opened 2022-03-23
feat(subscriptions): Add basic verifier

lynnagara:
Add a subscriptions verifier process that consumes two result topics
and attempts to diff them. Later this should actually compare
the results of each subscription individually to verify that every
result is the same. For the intial version, I kept it simpler. All it does
is compare the count of subscriptions results it got on each topic for
a given scheduled timestamp, and report the difference.


snuba/subscriptions/verifier.py
```diff
@@ -99,3 +112,203 @@ def close(self, timeout: Optional[float] = None) -> None:
...
+
+            # Set next_ts
+            if next_orig_ts is None and next_new_ts is None:
+                next_ts = None
+            elif next_orig_ts is None:
+                next_ts = next_new_ts
+            elif next_new_ts is None:
+                next_ts = next_orig_ts
+            else:
+                next_ts = min([next_orig_ts, next_orig_ts])
```
evanh, line 234:
```suggestion
                next_ts = min([next_orig_ts, next_new_ts])
```

lynnagara: ouch, thanks for noticing this!

snuba/subscriptions/verifier.py
```diff
@@ -99,3 +112,203 @@ def close(self, timeout: Optional[float] = None) -> None:
...
+            else:
+                next_ts = min([next_orig_ts, next_new_ts])
+
+            # Nothing else to remove
+            if next_ts is None:
+                return
+
+            orig_count = self.__stores[ResultTopic.ORIGINAL].pop(next_ts, 0)
+            new_count = self.__stores[ResultTopic.NEW].pop(next_ts, 0)
+            self.__metrics.increment("result_count_diff", new_count - orig_count)
```
onewland, line 242: For this metric, is it correct to say that lower absolute value is better but we probably won't hit 0 that often?

Or should this be 0 most of the time with occasional non-zero values when things can't be kept in sync?

snuba/subscriptions/verifier.py
```diff
@@ -99,3 +112,203 @@ def close(self, timeout: Optional[float] = None) -> None:
...
+            ):
+                self.__metrics.increment("stale_message")
+            return
+
+        # Increment counters
+        store = self.__stores[topic]
+
+        if item.timestamp not in store:
+            store[item.timestamp] = 0
+        store[item.timestamp] += 1
```
onewland, line 204: nit: self.__stores could store two defaultdicts with values initialized to 0

onewland (commented):
Looks good to me assuming my questions have sensible answers

snuba/subscriptions/verifier.py
lynnagara: Yeah, that's correct. The verifier does make an effort to not log anything when things are not quite in sync and wait 30 seconds in general for things to get in sync so we are definitely aiming for 0s all the time.

snuba/subscriptions/verifier.py
```diff
@@ -99,3 +112,203 @@ def close(self, timeout: Optional[float] = None) -> None:
...
+
+        # Record metrics and advance the low watermark
+        low_watermark = self.__timestamp_high_watermark - self.__threshold_sec
+
+        if low_watermark > self.__timestamp_low_watermark:
+            self._record(low_watermark)
+            self.__timestamp_low_watermark = low_watermark
+
+    def _record(self, until: int) -> None:
```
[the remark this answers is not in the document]
lynnagara: I don't think that can happen since the StreamProcessor which manages calling `submit` doesn't use any threading. I don't think we are going to be doing any threading in the verifier processing in general as there is not really IO tasks to to do.

snuba/subscriptions/verifier.py
lynnagara: I always forget about defaultdict

snuba/subscriptions/verifier.py
evanh: I was thinking about this, and the value should be wrapped in an `abs`. Otherwise you could report both a +1 and -1 at basically the same time, and Datadog will add them together.

lynnagara: Yeah, good point. I forgot Datadog adds these together in each interval before sending. Updated.

evanh (approved):
Thanks for the changes, looks good.

snuba/cli/subscriptions_verifier.py
```diff
@@ -0,0 +1,70 @@
...
+    default="snuba-subscription-verifier",
+    help="Consumer group used for consuming the commit log topic.",
+)
+@click.option(
+    "--orig-result-topic", required=True, type=str, help="Original result topic"
+)
+@click.option("--new-result-topic", required=True, type=str, help="New result topic")
+@click.option(
+    "--threshold-sec",
+    default=30,
```
fpacifici, line 26: I think you should give yourself at least a minute considering the jitter we apply is up to one minute

snuba/subscriptions/verifier.py
```diff
@@ -99,3 +113,200 @@ def close(self, timeout: Optional[float] = None) -> None:
...
+        message_topic = message.partition.topic
+        store_key = {
+            self.__orig_result_topic: ResultTopic.ORIGINAL,
+            self.__new_result_topic: ResultTopic.NEW,
+        }[message_topic]
+
+        self.__store.increment(store_key, message.payload)
+
+        # Immediately commit offsets
+        self.__commit({message.partition: Position(message.offset, message.timestamp)})
```
fpacifici, line 278: Are you not committing only one offset here ?
Also you would be committing around 100 times per second per topic as you commit for every query executed. That's too much. Please batch the commits.

fpacifici (commented):
Please fix the commit call.

snuba/subscriptions/verifier.py
lynnagara: Updated to throttle to 1 a second.

PR #2543, opened 2022-03-23
perf(subscriptions): Throttle committing offsets in executor

lynnagara:
We want to avoid committing every offset so that the amount of
committing we need to do does not scale with subscriptions volume.
This change throttles committing offsets in the executor to at most
once every second.

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -5,7 +5,16 @@ import time
 from collections import deque
 from concurrent.futures import Future, ThreadPoolExecutor
 from functools import lru_cache
-from typing import Callable, Deque, Mapping, Optional, Sequence, Tuple, cast
+from typing import (
+    Callable,
+    Deque,
+    Mapping,
+    MutableMapping,
+    Optional,
+    Sequence,
+    Tuple,
+    cast,
+)
 from zlib import crc32
 
 from arroyo import Message, Partition, Topic
@@ -40,6 +49,8 @@ from snuba.utils.streams.configuration_builder import build_kafka_consumer_confi
 from snuba.web.query import parse_and_run_query
 
 logger = logging.getLogger(__name__)
+
+COMMIT_FREQUENCY_SEC = 1
 
 
 def build_executor_consumer(
@@ -363,6 +374,10 @@ class ProduceResult(ProcessingStrategy[SubscriptionTaskResult]):
         self.__producer = producer
         self.__result_topic = Topic(result_topic)
         self.__commit = commit
+        self.__commit_data: MutableMapping[Partition, Position] = {}
+
+        # Time we last called commit
+        self.__last_committed: Optional[float] = None
 
         self.__encoder = SubscriptionTaskResultEncoder()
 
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -373,6 +388,22 @@ def __init__(
...
+    def __throttled_commit(self, force: bool = False) -> None:
+        # Commits all offsets and resets self.__commit_data at most
+        # every COMMIT_FREQUENCY_SEC. If force=True is passed, the
+        # commit frequency is ignored and we immediately commit.
+
+        now = time.time()
+
+        if (
+            self.__last_committed is None
+            or self.__last_committed + COMMIT_FREQUENCY_SEC < now
```
nikhars, line 400: This feels more readable.
```suggestion
            or now - self.__last_committed >= COMMIT_FREQUENCY_SEC
```

snuba/subscriptions/executor_consumer.py
```diff
@@ -404,6 +437,10 @@ def terminate(self) -> None:
 
     def join(self, timeout: Optional[float] = None) -> None:
         start = time.time()
+
+        # Commit all pending offsets
+        self.__throttled_commit(force=True)
```
nikhars, line 442: You might want to batch the individual commits happening here https://github.com/getsentry/snuba/pull/2543/files#diff-b81082d36647653087818cb560a918148759eb0e770ef1c04313589ea32167d7R454-R456 as well and then commit them altogether.

tests/subscriptions/test_executor_consumer.py
```diff
@@ -356,6 +356,11 @@ def test_produce_result() -> None:
     strategy.poll()
     assert commit.call_count == 1
 
+    # Commit is throttled so if we immediately submit another message, the commit count will not change
+    strategy.submit(message)
+    strategy.poll()
+    assert commit.call_count == 1
```
nikhars, line 362: Can you add a strategy.join() and verify that the commit call count goes up

MeredithAnya, line 362: how do we know that the commit count didn't change because the future wasn't done?

lynnagara: Sure

snuba/subscriptions/executor_consumer.py
lynnagara: I thought about this, but prefer to leave it as is. I think it's slightly simpler and it's harmless to call commit on each of the pending futures when we shut down. It's not possible for there to be a large number.

tests/subscriptions/test_executor_consumer.py
lynnagara: The throttled commit function is being once every time we poll regardless of whether the futures are done.

snuba/subscriptions/executor_consumer.py
nikhars: Do you have a rough idea of how large the pending futures might be?

In general, when the consumers are shutting down, we should try to finish shut down as quickly as possible. Think about the subscriptions rebalance problems we started seeing recently. If the complete shutdown happens quickly, in theory we should have fewer rebalances.

lynnagara: The theoretical maximum is the total concurrent queries (which we have defined as 24) divided by the number of executor replicas. So if we have 8 replicas, the max possible is 3.

nikhars: Ok. That's not a big number so we should be fine. I forgot that these were running queries and not processing ingest traffic.

nikhars (approved):

tests/subscriptions/test_executor_consumer.py
MeredithAnya: ah yeah okies `self.__commit_data` could be `{}` but we still call `self.__commit`

PR #2544, opened 2022-03-23
fix(metrics): Bump materialization version [INGEST-947]

untitaker:
In order to finish #2525, bump matversion so we start ingesting
histogram data.


onewland:
This is legitimately breaking a test or two (though it's a very simple fix)

onewland (approved):

PR #2546, opened 2022-03-23
test: Fix typing in tests and run mypy on tests in CI

lynnagara:
The goal of this change is to reduce the likelihood of
introducing new tests that are improperly typed or introducing
new code that breaks tests that were previously typed correctly.

Mypy is now run on all tests by default in CI. Only those provided
on the exclude list will be excluded.


tests/utils/metrics/test_decorators.py
```diff
@@ -1,14 +0,0 @@
-from unittest.mock import sentinel
```
volokluev, line 1: ?

lynnagara: Not used anywhere

volokluev (approved):

PR #2547, opened 2022-03-24
fix(profiling): Add a missing return

phacops:


evanh (approved):

onewland:
@phacops this appears to have caused a crash loop again: https://freight.getsentry.net/deploys/snuba/production/1547

just an FYI that i'm going to remove the profiles-consumer from freight

PR #2548, opened 2022-03-24
fix(pipeline_delegator): Execute logical query enity processing once

nikhars:
### Context
The current pipeline delegator uses a SimplePipelineExecution provided by the entity builder. The SimplePipelineExecution uses the EntityQueryPlanner which runs execute_entity_processors on the logical query. Running the entity processors on the same logical query can cause race conditions for pipelined delegator since the threaded delegator is also modifying the same query object from another thread. This can cause bugs like the one we see where Snuba function `notHandled()` is not translated correctly. The function is supposed to be modified by the `HandledFunctionsProcessor`.

### Solution
Added a new `SkippedLogicalEntityProcessorPipelineBuilder` which creates a `ClickhouseEntityQueryPlanner`. `ClickhouseEntityQueryPlanner` does not run the processors on the logical query. The `execute` method of the pipelinie delegator will run the processors now. The pipeline delegator now also requires a composite query planner. This is because the composite query planner needs to run the entity processors for the logical query.

I have added more description in the class `ClickhouseEntityQueryPlanner`.

Not doing a deepcopy to solve this problem because that would add performance penalty on Snuba query execution times for large query objects.

### Tests
Added tracing locally to prove the race conditions and verification of the fix which shows that processors on the logical query are being executed only once. Also verified all tests pass locally.

### Reviewers
I would like a thorough review on this. Especially from one of @fpacifici , @evanh , @lynnagara since they might have the most context


snuba/datasets/entities/events.py
```diff
@@ -139,8 +142,11 @@ def v2_selector_function(query: Query, referrer: str) -> Tuple[str, List[str]]:
...
         return ("errors_v1", [])
 
-    range = get_time_range(query, "timestamp")
-    if range[0] is None or range[0] < settings.ERRORS_UPGRADE_BEGINING_OF_TIME:
+    time_range = get_time_range(query, "timestamp")
+    if (
+        time_range[0] is None
+        or time_range[0] < settings.ERRORS_UPGRADE_BEGINING_OF_TIME
+    ):
         return ("errors_v1", [])
```
onewland, line 150: Not an aspect of this change, but could we describe the semantic meaning of what v2_selector function is returning somewhere?

Given the vagueness of str, List[str] I wonder if these values could be represented as enums

onewland (approved):
I think these changes make sense and look good

evanh (commented):
I'm a little confused by this fix. While I understand the race condition problem, I'm not sure why that race condition would cause `notHandled` to be missed. Both threads should be transforming it, so regardless of which one transformed it first, it should be transformed. I think you need a test to confirm the fix.

I also think the naming is confusing. The `EntityQueryPlanner` was conceived for subqueries and joins, where there are multiple separate queries, each on different entities and storages. Your case is different, in that you have the same entity, but different Clickhouse storages. Could you maybe rename your planner to `DuplicatedQueryPlanner` or something along those lines? I think that clear up the use case of the new classes.

snuba/datasets/entities/events.py
nikhars: We can make the improvement you are suggesting. But I would prefer to do that in a separate PR since it would touch more files.

nikhars:
> I'm a little confused by this fix. While I understand the race condition problem, I'm not sure why that race condition would cause `notHandled` to be missed. Both threads should be transforming it, so regardless of which one transformed it first, it should be transformed. I think you need a test to confirm the fix.
Let's just assume for now that this PR is handling the race condition. The way I would like to verify is by having this fix in place and seeing if we still see the problem in production.

> I also think the naming is confusing. The `EntityQueryPlanner` was conceived for subqueries and joins, where there are multiple separate queries, each on different entities and storages. Your case is different, in that you have the same entity, but different Clickhouse storages. Could you maybe rename your planner to `DuplicatedQueryPlanner` or something along those lines? I think that clear up the use case of the new classes.
Done. I was a little confused for the name to use.

lynnagara:
This is pretty complicated. Let me start by making some (probably stupid) clarifying questions / remarks.

- The PipelineDelegator interface seems a lot more complicated now. If you are building 2 pipelines, it needs to get passed 3 pipeline builders. Basically you need n + 1 query builders. There's quite a bit of duplication in what you need to pass in. I don't get why the 3rd one is needed.
- Now `build_planner` only builds one of the plans all the time (it's hardcoded) - not the actual one that is "selected" by the selector function like it used to... is this an issue? How is it we getting away with that?
- I found the name "composite" was also a bit confusing here as it is very close to "composite queries" which we use for i.e. joins or those with multiple data sources.
- It seems to me like we used to run `execute_entity_processors` twice in the execution phase and now we are just running it once.. is that right?
- Did you consider making the handled function processor idempotent so we can avoid the issue without changing the delegator?
- You mentioned the main concern with deepcopy is the impact on execution time of a single large query? Any idea what numbers we are talking about there? We are already copying the query over when transforming it to the ClickHouse query... would it really be much more than that other copy we are already doing?

fpacifici:
> You mentioned the main concern with deepcopy is the impact on execution time of a single large query? Any idea what numbers we are talking about there? We are already copying the query over when transforming it to the ClickHouse query... would it really be much more than that other copy we are already doing?

It is likely going to be a problem on subscription, specifically now that we have very little headroom, if we roll this out to a large portion of the queries.
We can probably handle that if we keep subscriptions off and deal with them separately.

nikhars:
Since the subscriptions are being run on the tiger cluster via the new scheduler/executor and the results are going to be verified, I am going to use deepcopy to simplify things and exclude subscriptions from running 2 instances of the queries

PR #2551, opened 2022-03-25
feat: Add app_id field and propogate to Datadog

evanh:
Add the app_id field to the request and propagate it through to the attribution
logging in the querylog module.

Added an attribution module to handle all the attribution specific code, since
this logic will get more complex going forward. Ideally this module is
responsible for tracking everything attribution related so the query pipeline
code can remain largely unchanged. Moved the attribution logging logic from the
querylog file into the attribution module so it can be expanded later.

Created a default file for all the AppIDs. Ultimately this will be augmented by
files stored in ops. There is a default AppID as a placeholder, and an invalid
AppID to make tracking invalid cases a little easier.

This will also stop logging the team/feature/parent_api to Datadog. They are
deprecated and no one is using that data so it is safe to remove.


snuba/attribution/default.py
```diff
@@ -0,0 +1,6 @@
+from datetime import datetime
+
+APPIDS = [
```
evanh, line 3: The file in ops would look similar to this I think.

snuba/web/db_query.py
```diff
@@ -408,10 +408,10 @@ def execute_query_with_caching(
     )
 
     with sentry_sdk.start_span(description="execute", op="db") as span:
+        key = get_query_cache_key(formatted_query)
```
evanh, line 411: There was a test failing that uses this code, and I had to make this change to ensure query_id is populated.

I don't think we actually use this function any more except in tests though.

snuba/web/query.py
```diff
@@ -117,8 +117,11 @@ def parse_and_run_query(
     """
     # from_clause = request.query.get_from_clause()
     start, end = None, None
+    entity_name = "unknown"
     if isinstance(request.query, LogicalQuery):
-        entity = get_entity(request.query.get_from_clause().key)
+        entity_key = request.query.get_from_clause().key
```
evanh, line 122: Not sure yet how this will work with composite queries.

volokluev (commented):
I think this is still too coupled to the query pipeline as it is. The `Request` object that is being used here to carry the app_id has a lot of other things in it not related to attribution/costing. By putting that information in the same place that we configure how the request should be run, we are coupling attribution and the query pipeline. Extracting that information in `build_request` where we parse the query is another mangling of concerns I'm not comfortable with. Here is my idea for how to make these separate:

### Passing the app_id
The app_id should be passed through headers instead of the regular JSON payload. This more closely resembles the pattern of authentication that we see in REST APIs. This way it is also separated from the query at the source. 

### Validating the app_id
This should be done at the flask web app layer. You can write a [middleware component ](https://medium.com/swlh/creating-middlewares-with-python-flask-166bd03f2fd4)that will do this for the request, or just make a decorator for the revelant endpoint (basically the same thing).

### Logging Attribution

The way I would like to maintain these components being separate is the following:

```python3
# current_app_id is passed in to the query view (or it can be put on the request)


query_results = return dataset_query(dataset, body, timer)
attribute_query(query_results, current_app_id)
```

The query pipeline knows nothing about the app_id and cannot use it for anything. The query pipeline surfaces the necessary information about the query that is needed for attribution. 


### Limiting Over-Quota Queries

The purpose of the project is to eventually limit the resources that customers use. Those limits should be decided by the attribution system before the query is parsed.

```
  QueryRequest
  (headers:app_id=discover)

         │
         │
         │
         ▼
┌────────────────────┐
│Attribution         │           ┌────────────────┐
│ Middleware         │           │                │
│                    │  quota    │ Quota          │
│ ┌────────────────┐ │◄──────────┤ Management     │
│ │ determine      │ │           │                │
│ │ query          │ │           │                │
│ │ limits         │ │           └────────────────┘
│ └─────┬──────────┘ │
│       │            │
│       │            │
└───────┼────────────┘
        │
        │app_id, query_limits, query
        │
        ▼
  ┌──────────────┐
  │              │
  │ query        │
  │ pipeline     │
  │              │
  └──────────────┘
  ```

 The query pipeline still does not know anything about app_ids, however it does have the concepts of enforced limits that are injected to it in the beginning of the request to the service.

evanh:
There's a lot to comment on here. 

**Using the web layer for attribution**

1) We should not put any logic into Flask that isn't required. Flask is the thin HTTP wrapper around our API and nothing else.
2) Subscription queries also have AppIDs so we can't put logic in the headers/web layer anyways.

So I'm going to skip over the suggestions about putting the attribution logic in the web layer.

**Where to store the AppID**

I put the AppID in the request since that is the logical place for it. As you correctly pointed out, AppIDs are not a query concern. However they are a request concern, so I added them to that object. They are also a shared concept between web queries and subscription queries.

On a related note, I'm a little confused by this comment because it appears that you also think it should be on the Request.
`# current_app_id is passed in to the query view **(or it can be put on the request)**`

Maybe you meant the HTTP request?

**Where to check attribution**
I agree that attribution should happen before query parsing, which is what I did. Right now it's just a call to `get_app_id` but that can expand as we build out the attribution model.

PR #2553, opened 2022-03-28
generate migrations from CLI

onewland:
This makes the prefix of migrations timestamp-based so two people don't have to coordinate to keep the ordering in sync. Inspired by some ruby on rails design decisions

Example run:
```
➜  snuba git:(migrations-cli-update) ✗ snuba migrations create --group metrics --description "add a column to a thing"                 
generated migration snuba/migrations/snuba_migrations/metrics/20220328102616_add_a_column_to_a_thing.py
add 20220328102616_add_a_column_to_a_thing to snuba/migrations/groups.py when ready
```

Output file from that command:
```
➜  snuba git:(migrations-cli-update) ✗ cat snuba/migrations/snuba_migrations/metrics/20220328102616_add_a_column_to_a_thing.py 
from typing import Sequence

# from snuba.clickhouse.columns import Column, Float
from snuba.migrations import migration, operations

class Migration(migration.ClickhouseNodeMigration):
    """
    add a column to a thing
    """
    blocking = False

    def forwards_local(self) -> Sequence[operations.SqlOperation]:
        return [
            # operations.CreateTable()
            # operations.AddColumn()
        ]

    def backwards_local(self) -> Sequence[operations.SqlOperation]:
        return [
            # operations.DropTable()
            # operations.RemoveColumn()
        ]

    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
        return [
            #operations.CreateTable(
            #   ...,
            #   engine=table_engines.Distributed(
            #        local_table_name=local_table_name, sharding_key=None,
            #    ),
            #),
        ]

    def backwards_dist(self) -> Sequence[operations.SqlOperation]:
        return [
            #operations.DropTable(),
        ]%                                                            
```

Open to ideas for how to improve the template

### DIFF
--- a/snuba/migrations/generation.py
+++ b/snuba/migrations/generation.py
@@ -1,0 +1,29 @@
+import datetime
+from dataclasses import dataclass
+
+from jinja2 import Environment, PackageLoader, select_autoescape
+
+from snuba.migrations.groups import MigrationGroup
+
+env = Environment(
+    loader=PackageLoader("snuba", package_path="migrations"),
+    autoescape=select_autoescape(),
+)
+
+template = env.get_template("migration_template.py.jinja")
+
+
+@dataclass(frozen=True)
+class GenerationResult:
+    filename: str
+    groups_entry: str
+
+
+def generate(group_enum: MigrationGroup, description: str) -> GenerationResult:
+    group = group_enum.value
+    transformed_description = description.lower().replace(" ", "_")
+    filesystem_friendly_timestamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+    entry = f"{filesystem_friendly_timestamp}_{transformed_description}"
+    output_file = f"snuba/migrations/snuba_migrations/{group}/{entry}.py"
+    template.stream({"group": group, "description": description}).dump(output_file)
+    return GenerationResult(filename=output_file, groups_entry=entry)
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -8,6 +8,7 @@ from snuba.clusters.storage_sets import StorageSetKey
 from snuba.environment import setup_logging
 from snuba.migrations.connect import check_clickhouse_connections
 from snuba.migrations.errors import MigrationError
+from snuba.migrations.generation import generate
 from snuba.migrations.groups import MigrationGroup
 from snuba.migrations.runner import MigrationKey, Runner
 from snuba.migrations.status import Status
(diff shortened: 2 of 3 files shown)


onewland:
Some next steps if we want to move forward with this:
1. move the groups definitions into manifest files so that we can automatically insert new migrations without weird copy/pasting
2. some kind of durable storage of the schema so we can detect drift between an environment's schema and the schema generated by running all migrations

snuba/migrations/generation.py
```diff
@@ -0,0 +1,29 @@
...
+@dataclass(frozen=True)
+class GenerationResult:
+    filename: str
+    groups_entry: str
+
+
+def generate(group_enum: MigrationGroup, description: str) -> GenerationResult:
+    group = group_enum.value
+    transformed_description = description.lower().replace(" ", "_")
+    filesystem_friendly_timestamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
```
nikhars, line 25: Do we need the granularity upto seconds. Can't we do with just the date? The number just becomes too big in my opinion with that much granularity.

onewland: Well the goal is that it should generate timestamps in order to create an increasing unique key. So I think one day isn't enough granularity (@untitaker and I created a migration in the same group on the same day recently) but maybe we could go down to the minute.

Also open to other ideas for this, I'm really just copying what Rails does for file naming in its migration system.

nikhars: Could you answer the following question to provide more clarity. 

When you do encounter a conflict on the migrations, are the steps like the following?
1. Rename the file using git.
2. Rename the migration in the migrations list.

Or is there more to it?

onewland: yeah, that's exactly what you do (assuming 2 of the same version never make it to master, and then you can't rename)

lynnagara: My main issue with switching over to the date is that we now have two conventions to contend with in our codebase, since we cannot go back and rename the older migrations to entirely use the new one. I feel that adds a lot of confusion. Fwiw Sentry/Django migrations uses the same naming convention we have here with the incrementing number prefix. An alternative idea to guard against this could be to add test or some other CI check that fails on any migration with a duplicate prefix in the same group.

untitaker: Migrations in Sentry, AFAIK, even go out of their way to force a merge-conflict by having this lockfile around. I wonder whether this was done because we couldn't switch to a different sequence format (like Rails) and had to ensure unique IDs, or because we actually wanted to make the author double-check whether migrations conflict.

fpacifici (commented):
I think we have an issue with this approach.
The timestamp of the migrations merged in master is not guaranteed to be monotonically increasing, which ill make things quite confusing:
- I generate a migration at timestamp 1
- You generate a migration at timestamp 3
- I go on PTO
- You add and merge your migration
- When I come back I add it to groups.py and merge. 

3 is before 1.

Having a monotonically increasing id in your migrations has the advantage of ringing an alarm if you fall behind and you find yourself adding an id that is lower than what is already in master. Example: somebody merged something after you started working on your migration. At that point you need to re-validate if your migration is fine with the new schema.

I like the idea of the commit hook that ensures you do not have duplication. I would go further and make that commit hook ensure that the ids are in order so we catch automatically potential semantic conflicts (which can be very dangerous when verisoning materialized views)

onewland:
Consensus seems to be that this creates more problems than it solves, abandoning this approach for now

PR #2555, opened 2022-03-28
fix(metrics): Fix optimize metrics

nikhars:
The optimize call is made on the one shard of each replica. Added a
host tag to each call of optimization to differentiate between calls
being made to each shard and differentiate tiger cluster from
errors cluster.

### DIFF
--- a/snuba/optimize.py
+++ b/snuba/optimize.py
@@ -1,7 +1,7 @@
 import logging
 import time
 from datetime import datetime, timedelta
-from typing import Optional, Sequence
+from typing import Mapping, Optional, Sequence
 
 from snuba import environment, settings, util
 from snuba.clickhouse.native import ClickhousePool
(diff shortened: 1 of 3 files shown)


snuba/optimize.py
```diff
@@ -22,12 +22,21 @@ class JobTimeoutException(SerializableException):
     pass
 
 
+def _get_metrics_tags(table: str, clickhouse_host: Optional[str]) -> Mapping[str, str]:
+    return (
+        {"table": table, "host": clickhouse_host}
+        if clickhouse_host
```
onewland, line 28: I'm not saying it's wrong but why would `clickhouse_host` ever be nil?

onewland (approved):

snuba/optimize.py
nikhars: Good question. I found out that the replacement consumer can also call run_optimize under certain circumstances. https://github.com/getsentry/snuba/blob/master/snuba/replacer.py#L418-L420. From there, `clickhouse_host` won't be passed.

PR #2556, opened 2022-03-28
ref(discover): Change isHandled to check all the values in the array

wmak:
- Based on this conversation: ([slack](https://sentry.slack.com/archives/CA2V2LBDL/p1646674287100079)|[notion](https://www.notion.so/sentry/error-handled-and-unhandled-logic-f720e11193f6470890ec02d8ac587008))
- This changes isHandled to check that all values in the
  exception_stacks.mechanism_handled array are either Null or 1
  - This is cause currently isHandled() doesn't equal the reverse of
    notHandled() which results in `error.handled:1` and
    `error.unhandled:1` returning the same rows
- This keeps notHandled the same as before where there's at least a 0 in
  the array

### DIFF
--- a/snuba/query/processors/handled_functions.py
+++ b/snuba/query/processors/handled_functions.py
@@ -27,7 +27,7 @@ class HandledFunctionsProcessor(QueryProcessor):
     The implementation of these functions is too complex for clients to provide so
     these wrappers are required.
 
-    - The `isHandled` function searches an array field for null or 1.
+    - The `isHandled` function checks the entire array field is null or 1.
     - The `notHandled` function searches an array field for 0, null
       values will be excluded from the result.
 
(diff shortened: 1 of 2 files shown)


volokluev:
It looks like there is important context for this in the chat links. Since you're modifying this code can you explain what the notHandled functionality is in the docstring? Saying `The isHandled function checks the entire array field is null or 1.` doesn't give me the reader of the code much context as to why it is doing what it is doing.

snuba/query/processors/handled_functions.py
```diff
@@ -27,9 +27,14 @@ class HandledFunctionsProcessor(QueryProcessor):
...
-    - The `isHandled` function searches an array field for null or 1.
-    - The `notHandled` function searches an array field for 0, null
-      values will be excluded from the result.
+    An event is considered unhandled if at least one of its stacktraces are
+    unhandled, regardless if the remaining stacktraces are handled or null. And the
+    inverse should hold true for whether an event is considered handled, that all of
+    its stacktraces are either handled or null.
+
+    - The `isHandled` function checks the entire array field is null or 1.
+    - The `notHandled` function searches an array field for at least one occurance of
```
volokluev, line 36:
```suggestion
    - The `notHandled` function searches an array field for at least one occurrence of
```

volokluev (approved):
just one small nit

PR #2557, opened 2022-03-28
ignore this, testing a github action 

onewland:


snuba/migrations/snuba_migrations/outcomes/0005_outcomes_copy.py
```diff
@@ -0,0 +1,128 @@
+from typing import Sequence
```
onewland, line 1: This is a copy of 0001 in outcomes, so that github would detect a migration change. It will be removed before merge

onewland:
Closing in favor of https://github.com/getsentry/snuba/pull/2567 which will reuse an existing comment vs posting a bunch of dupes

PR #2558, opened 2022-03-29
chore(logging): downgrade warning log to info

volokluev:
This message shows up occasionally in sentry but is not usually actionable. We can keep it in the logs for when we need to debug something but there's no need to surface it to sentry as often as we do.


lynnagara:
This is mostly coming from the new metrics scheduler. There is a task to investigate it in the next couple of weeks.

lynnagara:
I won't paste it here since it's a private resource but I've linked it in Sentry

PR #2559, opened 2022-03-29
test: Fix CLI test

lynnagara:
This test wasn't actually passing because invalid arguments were being passed. The sleep was just shorter than it took to actually start the consumer.


lynnagara:
Damn, still passing in CI though 2 seconds fails locally for me.

PR #2554, opened 2022-03-28
feat(metrics) add migrations to write to v2 sets/distributions tables

onewland:
Same as https://github.com/getsentry/snuba/pull/2540 but for sets and distributions

PR #2560, opened 2022-03-29
feat(metrics) - switch to v2 distributions aggregate table

onewland:
Contains changes from https://github.com/getsentry/snuba/pull/2554 which should be merged first

### DIFF
--- a/snuba/migrations/snuba_migrations/metrics/0030_metrics_distributions_v2_writing_mv.py
+++ b/snuba/migrations/snuba_migrations/metrics/0030_metrics_distributions_v2_writing_mv.py
@@ -39,7 +39,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 ),
                 metric_type="distribution",
                 target_mat_version=4,
-                appended_where_clause="AND timestamp > toDateTime('2022-03-29 00:00:00')",
+                appended_where_clause="AND timestamp >= toDateTime('2022-03-29 00:00:00')",
             )
         ]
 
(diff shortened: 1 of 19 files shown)


snuba/migrations/snuba_migrations/metrics/0030_metrics_distributions_v2_writing_mv.py
```diff
@@ -0,0 +1,60 @@
...
+                    "minState(values_rows) as min, "
+                    "maxState(values_rows) as max, "
+                    "avgState(values_rows) as avg, "
+                    "sumState(values_rows) as sum, "
+                    "countState(values_rows) as count, "
+                    "histogramState(250)(values_rows) as histogram_buckets"
+                ),
+                metric_type="distribution",
+                target_mat_version=4,
+                appended_where_clause="AND timestamp > toDateTime('2022-03-29 00:00:00')",
```
fpacifici, line 42: I think you can remove these conditions from the dev environment. They are really meaningful when we do a migration like the one we did in prod.

fpacifici (approved):

snuba/migrations/snuba_migrations/metrics/0030_metrics_distributions_v2_writing_mv.py
onewland: yeah, that is what was screwing up the tests. I think I made a very insigificant change to work around it

armenzg:
Hello,
I'm trying to understand how we got Sentry's master red.
Is it because we merged this even though there were some checks red?
image

Should we make those checks as required?

I'm trying to understand the situation and see what's needed to prevent future situations like this.

PR #2562, opened 2022-03-29
ref(profiling): Allow to customize the name of the field for the OrgIdEnforcer

phacops:


volokluev:
Add a test please

tests/query/processors/test_mandatory_condition_enforcer.py
```diff
@@ -106,16 +111,53 @@
...
+                binary_condition(
+                    "equals",
+                    Column(None, None, "bad_organization_id"),
+                    Literal(None, 1),
+                ),
+            ),
+        ),
+        False,
+        OrgIdEnforcer("organization_id"),
+        id="Valid query. Only one mandatory column",
```
evanh, line 149:
```suggestion
        id="Invalid query. Only one mandatory column",
```

evanh (approved):
One small change please, otherwise looks good.

PR #2564, opened 2022-03-29
Revert d46a0b8 and add fix for sentry tests

onewland:
Original revert PR https://github.com/getsentry/snuba/pull/2563


volokluev:
How does this fix the sentry tests?

onewland:
> How does this fix the sentry tests?

[This line](https://github.com/getsentry/snuba/pull/2564/files#diff-1177fdef1146fb34687b873f0f151ba90d542fe92e061aafb2b84d0b918b78c7R42) is the relevant change.

The sentry metrics-related tests use a hardcoded date of roughly 1 day ago: [source](https://github.com/getsentry/sentry/blob/7ae9b36b26f455d35bc67b8f9d85275f44d9d4ab/tests/snuba/api/endpoints/test_organization_sessions.py#L24-L30). I refactored these from a constant date to a dynamic one because we have a 90-day TTL on these tables, and this is pulling from ClickHouse. The change in this PR has us reading from a new table but that migration has us only writing to the new table from after 3-29-22 00:00:00 UTC. So the sentry tests are reading from a table where this data flow is not enabled yet.

onewland:
Looks like tests are still failing, investigating

snuba/migrations/snuba_migrations/metrics/0030_metrics_distributions_v2_writing_mv.py
```diff
@@ -39,7 +39,7 @@ def forwards_local(self) -> Sequence[operations.SqlOperation]:
                 ),
                 metric_type="distribution",
                 target_mat_version=4,
-                appended_where_clause="AND timestamp > toDateTime('2022-03-29 00:00:00')",
+                appended_where_clause="AND timestamp >= toDateTime('2022-03-22 00:00:00')",
```
fpacifici, line 42: Why not just removing it. 
Also I think you will have the same issue with sets

lynnagara, line 42: Separately to the tests, can you explain what this date actually means? Like, what is different before and after this date?

fpacifici:
I am not sure this will fix the problem though. 
Since the migration already landed in master and it was applied on everybody's developer environment, changing the timestamp will not have any impact for those that already applied the migration. To fix that you will have to add a new migration that creates a new matview and drops the existing.

snuba/migrations/snuba_migrations/metrics/0030_metrics_distributions_v2_writing_mv.py
onewland: > I am not sure this will fix the problem though. Since the migration already landed in master and it was applied on everybody's developer environment, changing the timestamp will not have any impact for those that already applied the migration. To fix that you will have to add a new migration that creates a new matview and drops the existing.

yeah, I'm doing this now

onewland: It was just a random marker so that Filippo could only backfill to that point and we know any row with a timestamp after that comes from the materialized view

onewland:
Please wait until I explicitly request a review before adding more comments.

onewland:
Ready for review again

snuba/migrations/snuba_migrations/metrics/0032_redo_0030_and_0031_without_timestamps.py
```diff
@@ -0,0 +1,77 @@
...
+                    "minState(values_rows) as min, "
+                    "maxState(values_rows) as max, "
+                    "avgState(values_rows) as avg, "
+                    "sumState(values_rows) as sum, "
+                    "countState(values_rows) as count, "
+                    "histogramState(250)(values_rows) as histogram_buckets"
+                ),
+                metric_type="distribution",
+                target_mat_version=4,
+                appended_where_clause="AND 1=1",
```
fpacifici, line 50: You should not need this. Please remove

snuba/migrations/snuba_migrations/metrics/0032_redo_0030_and_0031_without_timestamps.py
```diff
@@ -0,0 +1,77 @@
...
+                source_table_name=self.raw_table_name,
+                table_name=self.sets_table_name,
+                aggregation_col_schema=[
+                    Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
+                ],
+                aggregation_states="uniqCombined64State(arrayJoin(set_values)) as value",
+                mv_name=get_polymorphic_mv_variant_name("sets", self.mv_version),
+                metric_type="set",
+                target_mat_version=4,
+                appended_where_clause="AND 1=1",
```
fpacifici, line 66: Same, please remove

fpacifici (approved):
Please fix the conditions before merging

PR #2565, opened 2022-03-29
perf(optimize): Allow running 2 clcikhouse optimize queries concurently

nikhars:
### Context
The optimize jobs have started taking longer to complete. Sometimes they complete around noon PST. Which means that the job is running even during peak traffic time.

### Solution
To help with this, we can try running optimize on 2 different partitions simultaneously. This should be ok to do since we start the optimize job at 6 PM PST when the traffic volume is relatively low. I will also first do this only on the tiger cluster to observe the behavior.

### Tests
Added tests around the relevant code. Also ran the optimize command locally with the parallel flag and clickhouse having multiple partitions and did not see any issue.


snuba/cli/optimize.py
```diff
@@ -24,11 +24,15 @@
     required=True,
 )
 @click.option("--log-level", help="Logging level to use.")
+@click.option(
```
volokluev, line 27: are you sure we're going to cap out at 2? Why not just make the amount of parallelism a number?

volokluev: You can even validate the number not going higher than 2, but from an interface perspective I think this makes more sense

evanh (commented):
Looks good to me, besides Volo's comment.

volokluev (approved):

snuba/optimize.py
```diff
@@ -126,15 +128,68 @@ def get_partitions_to_optimize(
...
+                        ignore_cutoff,
+                        clickhouse_host,
+                    ),
+                )
+            )
 
+            threads[i].start()
+
+        for i in range(0, parallel):
+            threads[i].join()
```
fpacifici, line 182: I don't think we can do this. 
We cannot run parallel optimize for the entire duration of the job. As the process will end too close to peak load time.
We need to set a time boundary where we start running in parallel and where we stop running parallel and go back to serial. 

Which means this code needs to be arranged differently, having a number of threads,  then scheduling the next part on a thread as soon as we can and change the number of available threads depending what time it is.

PR #2566, opened 2022-03-29
feat(debt) Consumer cleanup step 1: demolish the stateful consumer.

fpacifici:
This consumer was built in order to ensure we would load a Change Data Capture snapshot from postgres consistently without possibility of double writes on Clickhouse. 
We never used it. It probably does not work and the way we store CDC data in Clickhouse does not really require avoiding double writes. 
So I think we can just remove it entirely and that reduces the reasons for having a consumer builder and makes the consumer building process simpler.
demolition


evanh (approved):
🗑️

PR #2567, opened 2022-03-30
config(migrations) - post migrations output to github PRs

onewland:
This does a diff with master and uses the existing Sentry commenting code to write out any new migration SQL to the current PR


snuba/migrations/groups.py
```diff
@@ -155,6 +155,7 @@ def get_migrations(self) -> Sequence[str]:
             "0002_outcomes_remove_size_and_bytes",
             "0003_outcomes_add_category_and_quantity",
             "0004_outcomes_matview_additions",
+            "0005_outcomes_copy",
```
onewland, line 158: This change will be removed before merge (needed for testing)

snuba/migrations/snuba_migrations/outcomes/0005_outcomes_copy.py
```diff
@@ -0,0 +1,128 @@
+from typing import Sequence
```
onewland, line 1: This file will be removed before merge (needed for testing)

lynnagara:
This is cool! Does it rerun and repost the DDL on every push? IIRC Sentry updates the original message- can we do the same?

onewland:
> This is cool! Does it rerun and repost the DDL on every push? IIRC Sentry updates the original message- can we do the same?

We're reusing their action and so it does reuse the existing message on every push

.github/workflows/ddl-changes.yml
```diff
@@ -0,0 +1,38 @@
...
+          python -m pip install --upgrade pip
+          make install-python-dependencies
+      - name: Run the migration script
+        run: |
+          SNUBA_SETTINGS=test_distributed python scripts/ddl-changes.py
+      - name: Generate SQL for migration
+        uses: getsentry/action-migrations@v1.0.8
+        env:
+          SENTRY_LOG_LEVEL: ERROR
+          PGPASSWORD: postgres
```
nikhars, line 33: Is this a security concern? Does Sentry open source also do this?

nikhars (commented):
Nice job. Left a followup question.

.github/workflows/ddl-changes.yml
onewland: This is copied directly from the way sentry does it, but I actually think I should get rid of SENTRY_LOG_LEVEL and PGPASSWORD as they don't apply to us

nikhars (approved):
Please remove the migrations from the PR which were added for testing.

PR #2571, opened 2022-04-01
fix(optimize): Make optimize jobs stateful

nikhars:
### Context
The optimize jobs are run as K8s cronjob. It can sometimes take close to 18 hours to run this job. K8s can move the job around to another pod at any moment of time. If it decides to do that, the optimize job has to start from the beginning again. Which means partitions which have already been optimized are optimized again. And partitions which haven't been optimized at all won't be optimized because of cut-off time.

### Solution
Introduce a mechanism of saving state of optimize jobs. This is done by saving the partitions which have already been optimized in redis. The optimized partitions are stored as redis sets. When the job starts it first gets the list of completed partitions from redis. Then it does a diff of partitions which need to be optimized with the partitions which have already been optimized. It then proceeds to perform the optimization on the diff'd set of partitions.

The PR also introduces a notion of threaded cutoff time. When an optimization is running in threaded mode we want to avoid running parallel optimizations when peak traffic starts. The threaded implementation uses the threaded cut off time to run the optimizations. Once all the threads work is done we check again to see if any partitions remain to be optimized using the saved state in redis. The optimization is then performed on remaining partitions using the original cutoff time.

### Tests
Added tests for redis states and running optimizations with the optimized partition tracker.


evanh (approved):
Nice, this looks good to me.

snuba/optimize_tracker.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class RedisOptimizedPartitionTracker(OptimizedPartitionTracker):
+    """
+    This class keeps track of partitions which have already been optimized by keeping state of
+    optimized partitions in redis.
+    """
+
+    def __init__(self, host: str, expire_time: datetime) -> None:
+        self.__bucket = f"{state.optimize_prefix}{host}"
```
volokluev, line 61: If you're going to have the prefix be configurable, it should be passed into the `__init__`, otherwise, you're making an implicit tie between this class and the `state` module, and it's not clear why

snuba/optimize_tracker.py
```diff
@@ -0,0 +1,81 @@
+from abc import ABC, abstractmethod
+from datetime import datetime
+from typing import Optional, Set
+
+from snuba import state
+from snuba.state import rds
+
+
+class OptimizedPartitionTracker(ABC):
```
volokluev, line 9: I don't see why any of this code needs to exist here. The InMemoryTracker is only used in a test and we have redis in our test environment. I would say let's not write code we don't use

snuba/optimize_tracker.py
volokluev: I would also say that the redis client should be passed into this as opposed to implicitly assumed to be the `snuba.state.rds` object.

volokluev: The `expire_time` should be part of the bucket as well. That way you're not relying on the expiration mechanism to separate the optimize jobs

snuba/optimize_tracker.py
```diff
@@ -0,0 +1,81 @@
...
+        partitions_set: Set[str] = set()
+        for partition in completed_partitions:
+            assert isinstance(partition, bytes)
+            partitions_set.add(partition.decode("utf-8"))
+
+        return partitions_set
+
+    def update_completed_partitions(self, part_name: str) -> None:
+        rds.sadd(self.__bucket, part_name.encode("utf-8"))
+        rds.expireat(self.__bucket, self.__key_expire_time)
```
volokluev, line 78: nit: you can do this in one redis call using `rds.pipeline()`

snuba/settings/__init__.py
```diff
@@ -198,6 +198,7 @@
 MAX_ROWS_TO_CHECK_FOR_SIMILARITY = 1000
 
 OPTIMIZE_JOB_CUTOFF_TIME = timedelta(hours=23)
+PARALLEL_OPTIMIZE_JOB_CUTOFF_TIME = timedelta(hours=11)
```
volokluev, line 202: Let's explain what these two constants mean and why these numbers were chosen

snuba/optimize_tracker.py
nikhars: > If you're going to have the prefix be configurable, it should be passed into the `__init__`, otherwise, you're making an implicit tie between this class and the `state` module, and it's not clear why

The prefix in now locally defined in the file. I had incorrectly assumed that all redis related definition should be in state.

> I would also say that the redis client should be passed into this as opposed to implicitly assumed to be the snuba.state.rds object.

Done

> The expire_time should be part of the bucket as well. That way you're not relying on the expiration mechanism to separate the optimize jobs

While it's not harmful to add the expire time to the bucket, I would not like to do so since it could imply that optimize jobs can run for more than a day when someone looks at a bucket name like `snuba-optimize:some-hostname:2022-04-01 23:00:00`.

snuba/optimize_tracker.py
nikhars: > I don't see why any of this code needs to exist here. The InMemoryTracker is only used in a test and we have redis in our test environment. I would say let's not write code we don't use

I don't like the idea of using infrastructure components unnecessarily for testing purposes. That's why I had introduced the InMemoryTracker. But I agree that it could be just in the test file. Removed it from the code for now.

snuba/optimize_tracker.py
volokluev: > While it's not harmful to add the expire time to the bucket, I would not like to do so since it could imply that optimize jobs can run for more than a day 

I understand that they are not *supposed to* run for more than a day but let's say that the mechanism we have to enforce fails for some reason. Would we rather have these keys collide or not collide? If we would rather have them collide that's fine by me

snuba/optimize_tracker.py
volokluev: I don't think the abstract class needs to exist here either. We do way too much Java-like interface design when it's not really necessary in python

snuba/optimize.py
```diff
@@ -226,6 +319,12 @@ def optimize_partitions(
         query = (query_template % args).strip()
         logger.info(f"Optimizing partition: {part.name}")
         start = time.time()
+
+        # Update tracker before running clickhouse.execute since its possible that the connection can get
+        # disconnected while the server is still executing the optimization. In that case we don't want to
+        # run optimization on the same partition twice.
+        if optimize_partition_tracker:
+            optimize_partition_tracker.update_completed_partitions(part.name)
```
fpacifici, line 327: I don't think this conceptually correct. 
In this implementation, every time we restart the job without the state being cleaned `get_partitions_to_optimize` is re-executed and the partitions that were already processed are excluded.
The problem is that `get_partitions_to_optimize` already pre-filters partitions that do not need to be optimized. Granted, most of the times we may need to optimize everything but can happen.
So, without keeping in redis the whole list of partitions you wanted to optimize on a given day, you may end up optimizing a partition after a restart (let's say at 5 am) that was not in the list for that day right after midnight because no merge had touched it that day. 
This will make you optimize more partitions during a day than we are doing today. 

I would store the state slightly differently: 
- at the beginning of the job, store (with the current date) the list of partitions to optimize.
- every time a partition is completed flag it as complete
- if you restart during the same day, just load the state and do not ask which partitions you need to Clickhouse and trust the state in redis.
At the start also clean up any possible remaining state from the previous date (but you can just set a TTL for the entire set to 24 hours).

This also ensures that the state starts fresh every day.

snuba/optimize_tracker.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    This class keeps track of partitions which have already been optimized by keeping state of
+    optimized partitions in redis.
+    """
+
+    def __init__(
+        self, redis_client: RedisClientType, host: str, expire_time: datetime
+    ) -> None:
+        self.__redis_client = redis_client
+        self.__bucket = f"{OPTIMIZE_PREFIX}:{host}"
```
fpacifici, line 47: I think you should also add the table. Today we only have one use case, but this is not granted. We may want (or more likely be forced to because I would definitively not want) to optimize multiple tables on the same host.

snuba/optimize.py
```diff
@@ -36,20 +37,43 @@ def run_optimize(
...
+        table=table,
+        parts=parts,
+        parallel=parallel,
+        cutoff_time=cutoff_time,
+        optimize_partition_tracker=optimize_partition_tracker,
+        clickhouse_host=clickhouse_host,
     )
+
+    if optimize_partition_tracker:
+        optimize_partition_tracker.remove_all_partitions()
```
fpacifici, line 76: How precise is the guarantee that Redis will get rid of a key at the TTL? If that is guaranteed all good. But if that means that Redis will keep the key "at least" up to the TTL, we may end up restarting the job the following day in case of exception or crash with the state still present.
A way to be 100% certain is to partition the state by day. See the comment below. if your state contains the date, you are sure never to read the state of the day before.

snuba/settings/__init__.py
```diff
@@ -198,8 +198,13 @@
 
 MAX_ROWS_TO_CHECK_FOR_SIMILARITY = 1000
 
+# Cutoff time from UTC 00:00:00 to stop running optimize jobs to avoid spilling over to the next day.
 OPTIMIZE_JOB_CUTOFF_TIME = timedelta(hours=23)
 
+# Cutoff time from UTC 00:00:00 to stop running optimize jobs in parallel to avoid running in parallel
+# when peak traffic starts.
+PARALLEL_OPTIMIZE_JOB_CUTOFF_TIME = timedelta(hours=11)
```
fpacifici, line 206: We cannot do this past 9 am UTC. That is the time we start the weekly report queries on monday morning.
Also We cannot do this from the start of the job as, when we start the job (which is around 1 am UTC) the TTL merge is still on going generally (it takes 2.5 hours now). Which means at the start of the job we still have 2 merges already on going. I would have a parameter to decide when to start the parallel section of this job and maybe start around 3 am UTC

fpacifici (changes requested):
Please see my comments on the state and the time frame for the threaded optimize

snuba/optimize.py
```diff
@@ -145,72 +169,141 @@ def _subdivide_parts(
...
+    optimize_partition_tracker: Optional[OptimizedPartitionTracker],
     clickhouse_host: Optional[str] = None,
 ) -> None:
-    if parallel <= 1:
-        optimize_partitions(
-            clickhouse, database, table, parts, ignore_cutoff, clickhouse_host
+    """
+    Run optimize job concurrently using multiple threads.
+
+    It subdivides the partitions into number of threads which are needed to be run. Each thread then runs the optimize
```
fpacifici, line 185: nit: PEP8 suggest 72 charactes limit per line on docstring to make them easier to read in code reviews 
https://peps.python.org/pep-0008/#maximum-line-length

snuba/optimize.py
```diff
@@ -145,72 +169,141 @@ def _subdivide_parts(
...
+    job on the list of partitions assigned to it.
+
+    The threads are provides a different cutoff_time than what is provided when calling this function. This is to avoid
```
fpacifici, line 188: nit: `The threads are provided`

snuba/optimize.py
```diff
@@ -145,72 +169,141 @@ def _subdivide_parts(
...
+    if parallel <= 1:
+        optimize_partitions(
+            clickhouse=clickhouse,
+            database=database,
+            table=table,
+            parts=parts,
+            cutoff_time=cutoff_time,
+            optimize_partition_tracker=optimize_partition_tracker,
+            clickhouse_host=clickhouse_host,
+        )
```
fpacifici, line 279: nit: I don;t think you are gaining much by having this two code paths handled separately. 
The job takes hours to complete, the threading overhead is none basically on that scale, why not just using the threaded version everywhere with 1 thread if you do not want parallelism?
So we will have only one code path to think about instead of two.

snuba/optimize_tracker.py
nikhars: Added the date to the time bucket to avoid chances of conflict

snuba/optimize_tracker.py
nikhars: Removed the abstract class

snuba/optimize.py
nikhars: > Moreover, in the current state, if the job restarts and there is a merge on going (merges are asynchronous, if the job crashes, the job does not stop) it will not care about on going merges and just schedule a new one.
This is a problem and we should avoid it. Imagine if K8S is having a bad day and keeps restarting this job over and over again. We will keep scheduling new merges. Clickhosue does not respect any quota on OPTIMIZE FINAL calls. So we may take down clickhouse.

>I think it is important to keep the state of the last merge we scheduled and query system.merges when we start. If we find that a merge we had started but not completed is still on going, just wait fora moment when clickhouse will be doing no merges.

This PR has already gotten pretty big than I would have liked it to be. I have incorporated all suggestions except the one mentioned above. I would like to tackle that in a separate PR to keep things sane. In the meantime, we could avoid doing parallel clickhouse optimizations even if this PR is merged.

snuba/optimize_tracker.py
nikhars: Done

snuba/optimize.py
nikhars: Date is part of the key now

snuba/settings/__init__.py
nikhars: Added buckets for optimization which define different parallel and cutoff times

snuba/optimize.py
```diff
@@ -36,20 +37,43 @@ def run_optimize(
     storage: ReadableTableStorage,
     database: str,
     before: Optional[datetime] = None,
-    ignore_cutoff: bool = False,
     parallel: int = 1,
+    cutoff_time: Optional[datetime] = None,
     clickhouse_host: Optional[str] = None,
+    optimize_partition_tracker: Optional[OptimizedPartitionTracker] = None,
```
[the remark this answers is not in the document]
nikhars: This was because of replacer being able to run optimize job. The call from replacer is separated out now

snuba/optimize.py
nikhars: Done

snuba/optimize.py
```diff
@@ -31,31 +34,144 @@ def _get_metrics_tags(table: str, clickhouse_host: Optional[str]) -> Mapping[str
     )
 
 
+@dataclass(frozen=True)
+class OptimizationBucket:
+    parallel: int
+    cutoff_time: datetime
+
+
+def _build_optimization_buckets(
```
fpacifici, line 43: Please have a unit test for this method. It is easy and stateless

fpacifici (commented):
Some comments. Will review again.

snuba/optimize_tracker.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        all_partitions = self.__redis_client.smembers(self.__all_bucket)
+        if not all_partitions:
+            return None
+
+        return self.__redis_set_of_bytes_to_set_of_strings(all_partitions)
+
+    def update_all_parts(self, part_names: Sequence[str]) -> None:
+        """
+        Update the list of all partitions which need to be optimized.
```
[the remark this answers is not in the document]
nikhars: Renamed all references of parts to partitions in the code

snuba/optimize.py
```diff
@@ -128,19 +253,29 @@ def get_partitions_to_optimize(
...
-    Subdivide a list of parts into number_of_subdivisions lists of parts so that optimize can be executed
-    on different parts list by different threads.
+    Subdivide a list of partitions into number_of_subdivisions lists of partitions
+    so that optimize can be executed on different partitions list by different
+    threads.
     """
 
-    sorted_parts = sorted(parts, key=lambda part: part.date, reverse=True)
-    output: MutableSequence[Sequence[util.Part]] = []
+    def sort_partitions(partition_name: str) -> str:
```
fpacifici, line 265: Why sorting at all ?

snuba/optimize.py
```diff
@@ -169,49 +314,49 @@ def optimize_partition_runner(
...
 
+        # Wait for all threads to finish. Any thread can raise
+        # JobTimeoutException. Its an indication that not all partitions
+        # could be optimized in the given amount of threaded_cutoff_time.
         for i in range(0, parallel):
-            threads[i].join()
+            try:
+                threads[i].join()
+            except JobTimeoutException:
+                pass
```
fpacifici, line 334: Doesn't this make the JobTimeoutException completely transparent? 
We should have the job crash loudly if the job goes past its overall cutoff time so we can alert on it.

snuba/optimize.py
```diff
@@ -31,31 +34,153 @@ def _get_metrics_tags(table: str, clickhouse_host: Optional[str]) -> Mapping[str
...
+        # We don't have any recorded state of partitions needing optimization
+        # for today. So we need to build it.
+        partitions = get_partitions_to_optimize(
+            clickhouse, storage, database, table, before
+        )
+        if len(partitions) == 0:
+            logger.info("No partitions need optimization")
+            return 0
+        partition_names = [partition.name for partition in partitions]
+        tracker.update_all_partitions(partition_names)
```
fpacifici, line 159: What happens if the job gets restarted past the daily job cutoff time? Like 23:11.
The TTL we set on Redis will be in the past. Will it be ok ?

snuba/optimize.py
```diff
@@ -128,19 +253,29 @@ def get_partitions_to_optimize(
     return parts
 
 
-def _subdivide_parts(
-    parts: Sequence[util.Part], number_of_subdivisions: int
-) -> Sequence[Sequence[util.Part]]:
+def _subdivide_partitions(
```
fpacifici, line 256: I am a bit concerned that the code in this module is becoming quite hard to follow. I think it is because there are two concerns that are mixed together while an abstraction could help.
Specifically we mix the scheduling aspect and the execution aspect.

What about having an `OptimizeScheduler` class that has a method  `get_next_schedule`. This method returns a sequence of seuqences of partitions. Each sequence is the schedule for a thread and a cutoff time. Once the cutoff time is passed you get the next schedule (and provide the set of successfully executed partitions).

fpacifici (approved):
Seems ok to me except for the job failing silently. Please fix that before merging.
Up to you if you want to take the `OptimizeScheduler` suggestion.

snuba/optimize_scheduler.py
```diff
@@ -0,0 +1,114 @@
...
+        Subdivide a list of partitions into number_of_subdivisions lists
+        of partitions so that optimizations can be executed in parallel.
+
+        We sort the partitions so that the more recent partitions are
+        optimized first since more recent partitions might have more
+        replacements.
+        """
+
+        def sort_partitions(partition_name: str) -> str:
+            date_regex = re.compile("\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])")
```
evanh, line 69: I know this code was copied from another place, but it's bad practice to compile regexes in a loop. I would move this at least out of the `sort_partitions` function and maybe even into the init of the class to avoid this happening more than once.

evanh (commented):
Sorry I just noticed one small thing.

snuba/optimize.py
nikhars: Sorting the partitions since I want to ensure that the most recent partitions are optimized before the older ones. Our product mostly serves queries which are more recent. If we ever hit the timeout condition this would ensure that the queries for recent data would not have issues.

snuba/optimize.py
nikhars: That was a good suggestion. I have incorporated the changes. Could you review the code again and see if the changes are inline with what you had in mind.

snuba/optimize_scheduler.py
nikhars: Good callout. Moved the regex compilation out.

PR #2572, opened 2022-04-01
feat(admin): Added SnQL to SQL conversion

rahul-kumar-saini:
### Overview
- Currently, the admin console only supports ClickHouse tracing data for ClickHouse SQL queries
- Added a SnQL to SQL converter
SnQL to SQL

### How it works
- The SnQL tracing console is given a SnQL Query along with a Dataset to query
- This is used for a "debug" dry-run which returns the generated SQL, along with the storage(s) the query would use
- This SQL and Storage can then be used to hit the existing ClickHouse tracing API which results in the actual tracing data


rahul-kumar-saini:
Most of the front end code is copied from the existing tracing stuff, but I refactored the directories and added a `/common` directory under `/tracing` which serves as some setup for a future refactor PR which would pull out many reused components out of both tracing tools into a single directory.

lynnagara:
The SnQL -> SQL conversion seems good to me.

One small concern about how the SnQL tracing works though. It's behaviour feels a bit at odds with the ClickHouse tracing tool since  it basically picks a random storage (well, the first one in the list)  to find the query node on which to run the query, whereas the ClickHouse tracing tool makes you explicitly pick a storage which is unambiguous about which query node the query is run on.

Today in prod this doesn't really make much difference because I think we have configured all of the storages of an entity to share query nodes. But Snuba does not really enforce this in any way, and it's possible for this not to be the case.

If you want to avoid this, some ideas to solve could be:
- Make the user select both dataset and storage in the interface (the storage list gets filtered to only the relevant ones when the pick a dataset)
- Just make the user use the SnQL -> SQL converter first and then paste that into the regular ClickHouse tracing tool

If you do decide to keep as is, I think it's worth adding a comment about it.

snuba/web/query.py
```diff
@@ -185,7 +186,13 @@ def _run_query_pipeline(
         metrics.increment("sample_without_turbo", tags={"referrer": request.referrer})
 
     if request.settings.get_dry_run():
-        query_runner = _dry_run_query_runner
+        storage_name = "<multiple>"
+        if isinstance(request.query, LogicalQuery):
+            entity = get_entity(request.query.get_from_clause().key)
+            storage = entity.get_all_storages()[0]
+            assert isinstance(storage, ReadableTableStorage)
+            storage_name = storage.get_storage_key().value
```
rahul-kumar-saini, line 194: @lynnagara 
Are you talking about line 192 here? From most entity definitions I've seen, we have the form:
```python3
super().init(storages=[some_selected_storage], ...)
```

So when we do `entity.get_all_storages()[0]`, it just returns that one right?

Eg: https://github.com/getsentry/snuba/blob/de14212c8f504144359fa63692a261c796d41067/snuba/datasets/entities/events.py#L198-L199

How different is this than how a regular Snuba query is processed, the entity returns the appropriate storage from its list of storages no?

The only deviations I've seen from this is in the metrics, outcomes, and sessions entities and even those just have two storages instead of one, with the first one being good enough to query from as the storage.

I could change it to be two drop downs, one for dataset and one for storage but I think the storage dropdown would contain exactly 1 or 2 storages to choose from (read/writable + maybe materialized).

lynnagara: AFAIK it's quite different to how a regular query is processed which goes through the StorageSelector. I don't think we should rely on the fact that today our storages happen to be just one or two within each entity, because we'd want it to still work with future datasets/entities where that might not be the case.

To be honest, I personally don't think the SnQL tracing tool adds all that much additional value once we have the SnQL -> SQL converter, since we can just grab the output of that and stick it into the existing tool. Especially if combining the two functionalities forces us to make these kinds of assumptions about storages within an entity.

lynnagara:
It's pretty easy to input malformed SnQL when writing it by hand. I think you should catch the exception so it doesn't get logged to Sentry and return it to the user.

snuba/web/query.py
rahul-kumar-saini: I'm all for deleting code, so I'll remove that section in favor of the SnQL to SQL translate -> SQL Tracing workflow.

snuba/admin/static/api_client.tsx
```diff
@@ -31,6 +36,8 @@ interface Client {
   getDescriptions: () => Promise<ConfigDescriptions>;
   getAuditlog: () => Promise<ConfigChange[]>;
   getClickhouseNodes: () => Promise<[ClickhouseNodeData]>;
+  getSnubaDatasetNames: () => Promise<SnubaDatasetName[]>;
+  executeSnQLQuery: (query: SnQLRequest) => Promise<SnQLResult>;
```
lynnagara, line 40: How about`convertSnQLQuery` maybe? No query is being executed

snuba/web/query.py
```diff
@@ -215,7 +223,7 @@ def _dry_run_query_runner(
     return QueryResult(
         {"data": [], "meta": []},
         {
-            "stats": {},
+            "stats": {"storage_names": storage_names},
```
lynnagara, line 226: What's the reason for this change?

snuba/web/query.py
```diff
@@ -287,10 +295,8 @@ def _format_storage_query_and_run(
     """
     Formats the Storage Query and pass it to the DB specific code for execution.
     """
-    from_clause = clickhouse_query.get_from_clause()
     visitor = TablesCollector()
-    visitor.visit(from_clause)
-    table_names = ",".join(sorted(visitor.get_tables()))
+    table_names = _get_clickhouse_tables(clickhouse_query, visitor)
```
lynnagara, line 299: How come we need to make these changes to the query pipeline?

snuba/web/query.py
```diff
@@ -185,7 +186,13 @@ def _run_query_pipeline(
         metrics.increment("sample_without_turbo", tags={"referrer": request.referrer})
 
     if request.settings.get_dry_run():
-        query_runner = _dry_run_query_runner
+        storage_names = []
```
lynnagara, line 189: Why do we need to collect storage names here? Isn't that the job of the storage selector later in the query pipeline?

snuba/admin/views.py
```diff
@@ -346,3 +349,21 @@ def clickhouse_nodes() -> Response:
...
+def snuba_datasets() -> Response:
+    return Response(
+        json.dumps(
+            get_enabled_dataset_names(), 200, {"Content-Type": "application/json"}
+        )
+    )
+
+
+@application.route("/snql_to_sql", methods=["POST"])
+def snql_to_sql() -> Response:
```
lynnagara, line 364: Can you add tests for these APIs?

snuba/web/query.py
rahul-kumar-saini: Oops this was part of a first attempt at getting more data into stats in dry run pipeline, forgot to undo.

snuba/web/query.py
rahul-kumar-saini: I was thinking about having a preloaded set of storages a user can choose from if they want to get tracing data. Assuming we're given a non-composite query, we could return a list of probable storages they would choose from? I don't know how useful it would be though so I can remove this now and re-add it if we want to pursue this route in another tech debt week.

snuba/web/query.py
rahul-kumar-saini: Same reason as above

evanh (approved):

snuba/admin/static/snql_to_sql/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      return {
+        ...prevQuery,
+        query,
+      };
+    });
+  }
+
+  function convertQuery() {
+    if (isExecuting) {
+      window.alert("A query is already running");
```
lynnagara, line 48: I'd return after this line, otherwise it'll rerun the query.

tests/admin/test_api.py
```diff
@@ -213,3 +214,48 @@ def test_query_trace_invalid_query(admin_api: FlaskClient[Any]) -> None:
...
+    AND project_id IN tuple(100)
+    AND started >= toDateTime('2022-01-01 00:00:00')
+    AND started < toDateTime('2022-02-01 00:00:00')
+    """
+    response = admin_api.post(
+        "/snql_to_sql", data=json.dumps({"dataset": "sessions", "query": snql_query})
+    )
+    assert response.status_code == 200
+    data = json.loads(response.data)
+    assert data["sql"] != ""
```
lynnagara, line 261: Probably good to put the whole query here (or at least the important parts of it), not just assert it's anything except an empty string.

PR #2574, opened 2022-04-02
feat(consumers): Support the --cooperative-rebalancing option

lynnagara:
The consumers, multistorage-consumers, subscriptions consumer and
subscriptions-executor now support passing a --cooperative-rebalancing
flag which, if passed, will use the "cooperative-sticky" partition
assignment strategy.

Note that cooperative and non-cooperative strategies cannot be
combined in the same group so this cannot be deployed with a rolling
strategy. When enabling cooperative rebalancing for the first time,
all old consumers in the group should be stopped before restarting.

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -88,6 +88,13 @@ from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
 @click.option(
     "--profile-path", type=click.Path(dir_okay=True, file_okay=False, exists=True)
 )
+# TODO: For testing alternate rebalancing strategies. To be eventually removed.
+@click.option(
+    "--cooperative-rebalancing",
+    is_flag=True,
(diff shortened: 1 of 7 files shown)


snuba/cli/consumer.py
```diff
@@ -107,6 +114,7 @@ def consumer(
     output_block_size: Optional[int],
     log_level: Optional[str] = None,
     profile_path: Optional[str] = None,
+    cooperative_rebalancing: bool = False,
```
fpacifici, line 117: Nobody does anything with this ?

fpacifici (approved):
Please see my comment on the standard consumer

snuba/cli/consumer.py
lynnagara: Oops, wired up properly now. Thanks.

PR #2581, opened 2022-04-05
feat(profiling): Whitelist a few dates and strings functions

phacops:
This will whitelist a few functions we'd like to use for profiling.


snuba/query/functions.py
```diff
@@ -149,6 +149,9 @@ def is_aggregation_function(func_name: str) -> bool:
     "toStartOfHour",
     "toStartOfYear",
     "toUnixTimestamp",
+    "date_add",
+    "date_sub",
+    "now",
```
fpacifici, line 154: could you please provide some context on why you need these and which kinds of queries you plan to use them for ?

phacops: We'd like to use the date functions to calculate the range we need in the query. Sometimes, we have to query via unique ID and Snuba forces us to pass a timestamp condition so we need to set an arbitrary date range. Instead of calculating before and setting it in the query, I'd rather set it in my query.

For `concat`, we have 2 fields which are stored separately (`version_name` and `version_code`). We'd want to consider them one and format them as we're going to use them directly in the query as opposed to doing that later.

fpacifici: > Sometimes, we have to query via unique ID and Snuba forces us to pass a timestamp condition so we need to set an arbitrary date range. Instead of calculating before and setting it in the query, I'd rather set it in my query.

Snuba requires you to provide a time range so it can narrow down your data by primary key.
I am not sure if you are talking about setting a time condition as `timestamp > date_add(columns, something)` if that is not the case, then the problem is different. If the condition you plan to add is like the one I provided instead you have this situation.
For the cases where you need to access an individual transaction we added a bloomfilter index on the transaction id column to narrow down by that one.
Now, there are two scenarios:
- you actually know a time range in which your transaction can be (however large) or you can find it out somehow. Then pass relevant conditions: `timestamp> X AND timestamp < Y`
- you do not know at all the time range the transactions can be. In this case you may are likely to be better off passing the maximum time range (3 months ago to now) than defining the time range as a function of a specific column.

if your query sets a time condition as `timestamp > literal` or `timestamp > a_function(literal)`, this is good. Clickhouse can use the primary key. 
But if you data contains a condition like `timestamp > a_function(another_column)`, then clickhouse cannot use the primary key and will have to calculate that function for every row in the table for the projects you are querying as the function execution would happen after the primary key filtering.

So if you plan to pass a column name to `date_add`  and use it in your condition, be aware that it will actually make the query worse than if you provide the maximum time range. At least in that case Clickhouse would not have to calculate the function for every row.  So you probably do not need to enable those two functions.

fpacifici: Also I would encourage you to add a prewhere processor to your storage:
```
PrewhereProcessor(["transaction_id"]),
```

Here: https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/profiles.py#L21

This will improve the performance of your queries that look for individual transactions ids as Clickhouse will process conditions on the `transaction_id` column before loading any other column and right after filtering by primary key/index

fpacifici: Could you please avoid using `now()` in your query? It makes it really hard to troubleshoot the query later on from the querylog to understand what Clickhouse was doing because not only we will have a hard time knowing what the exact timestamp was on the clickhouse server when the query was executed, but the value of `now()` will change during the execution of the query for long queries. 
Enabling it, will enable the function for all datasets, there are cases where it would generate wrong and non reproducible results. (like GROUP BY `toStartOfMinute(now() - something)`

Also, I don't think you need that function for `timestamp_sub(day, 30, now())`. You can express it as `toDateTime(your date) - number_of_seconds`.

phacops: My goal is to not specify a date, that's why I don't want to use `toDateTime` with my date. I very much want to have that part of the query "static". It fits my use case very well since I want to target the whole range of possible timestamps so I wouldn't make troubleshooting very difficult.

It seems like you want to prevent me to use it because it could cause issues with other's people queries which could then be difficult to troubleshoot. I would suggest to then educate people wanting to use that of the danger of using it as oppose to blocking this entirely.

fpacifici: > It seems like you want to prevent me to use it because it could cause issues with other's people queries which could then be difficult to troubleshoot. I would suggest to then educate people wanting to use that of the danger of using it as opposed to blocking this entirely.

This is not viable in a reasonable period of time, nor in a reliable way as we tried in the past to take this approach and reverted back. Snuba being a multi tenant platform that relies on multi tenant clusters that involve multiple use cases and is used by many teams we need to put safeguards to keep the usage sane.

So if you want to take the risk for your own dataset it is up to you, we cannot provide support though. In any case we are not going to enable this for the entire platform. 
While I would discourage you from enabling this specific function, if you want a specific exception for your  dataset there is a way to achieve that.
Implement `now` as a custom function and declare it in your entity. Do not call it now otherwise it will conflict with the Clickhouse now() function . You can do that by adding this one line `CustomFunction("now", [], simple_function("now()"))` to the query processors of your entity. That declares a custom function that you can call from the Snuba client and expand to whatever you want. You can use it to declare whatever function you want snuba to expand for you. An example is here: https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/transactions.py#L218. Then add it to the `CUSTOM_FUNCTION` list in `query/functions.py` 

@evanh  I remember we wanted to allow specific validators per dataset/entity to enable a function in a dataset only, did we ever make progress on that ? It seems to me we allowed validators per entity but we did not add a way to allow specific functions to be enabled in a dataset.

phacops:
Not worth it.

snuba/query/functions.py
evanh: There's a task in the backlog I believe to start validating functions on a per-entity basis.

PR #2582, opened 2022-04-05
fix(metrics) - update a test with a hard-coded timestamp beyond database TTL

onewland:
Fix of a test failure that hit a TTL condition today at 1:35 PM PT

### DIFF
--- a/tests/test_consumer.py
+++ b/tests/test_consumer.py
@@ -16,7 +16,6 @@ from arroyo.processing.strategies.streaming import KafkaConsumerStrategyFactory
 from arroyo.types import Position
 from arroyo.utils.clock import TestingClock
 
-from snuba import settings
 from snuba.clickhouse.errors import ClickhouseWriterError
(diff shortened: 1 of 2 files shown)


tests/test_consumer.py
```diff
@@ -346,35 +345,28 @@ def test_dead_letter_step() -> None:
 
 def test_metrics_writing_e2e() -> None:
```
onewland, line 346: I would actually be fine removing this test entirely, but it does mean we'd have a bit less coverage of the consumer flow

tests/test_consumer.py
```diff
@@ -346,35 +345,28 @@ def test_dead_letter_step() -> None:
...
     from snuba.datasets.storages.metrics import (
-        counters_storage,
```
onewland, line 349: This and sets_storage are no longer in use, and we're never going to set `settings.WRITE_METRICS_AGG_DIRECTLY = True`

volokluev (approved):

PR #2585, opened 2022-04-06
feat(dlq): Introduce DLQ to Metrics with Ignore policy.

rahul-kumar-saini:
### Overview
- Added a DLQ to for Metrics storages with an `ignore` policy - all invalid messages raised will be dropped and a datadog metric will be incremented

### Changes
- Updated arroyo to 0.0.15
- Added an optional DLQ Policy closure arg to stream loader so any storage can start adding DLQ's easily
- Updated redis count policy to reflect new arroyo changes, added some tests for the policy as well


snuba/datasets/storages/metrics.py
```diff
@@ -102,7 +105,9 @@
     ),
     query_processors=[ArrayJoinKeyValueOptimizer("tags"), TableRateLimit()],
     stream_loader=build_kafka_stream_loader_from_settings(
-        SetsAggregateProcessor(), default_topic=Topic.METRICS,
+        SetsAggregateProcessor(),
+        default_topic=Topic.METRICS,
+        dead_letter_queue_policy=IgnoreInvalidMessagePolicy(),
```
fpacifici, line 110: In order for the IgnoreInvalidMessagePolicy to be useful, when it drops a message and records the metric it needs to be able to provide context. Now we would not know which consumer received a bad message. 

Though there is a larger issue here: https://github.com/getsentry/arroyo/blob/d05860814366ff7d37ed90346385439de600c7bc/arroyo/processing/strategies/dead_letter_queue/policies/ignore.py#L10
Metrics will be initialized at the initialization of the class. This is a problem since now you are creating the policy at the initialization of the storage, which likely happens before this line 
https://github.com/getsentry/snuba/blob/master/snuba/cli/consumer.py#L125-L130
Which is when we set up the metrics. 
If your policy is initialized before that line (as it will) your policy will be initialized with a dummy metrics backend and never record any metric.

In practice the issue is that the data model should not reference arroyo components that rely on metrics initialization. @lynnagara  we may want to revisit the design decision of initializing metrics and have them as a singleton for arroyo. This can bite us back in these scenarios.

You have two ways to solve this:
- do not initialize the policy here. Provide instead a closure that initialize the policy. This closure will be called by KafkaConsumerStrategyFactory when the strategy is created.
- make the `IgnoreInvalidMessagePolicy` load the metrics each time it needs them, but this is going to be a problem as it creates an implicit coupling between the way arroyo works and the way snuba works. The two should exist independently from each other or, at best , Snuba should adapt to arroyo as Snuba is a client of the library, not vice versa.

rahul-kumar-saini: > we would not know which consumer received a bad message.

If the closure solution to the larger problem works, does this still apply? It seems the metrics will have the consumer group and storage key tags.

rahul-kumar-saini:
@getsentry/open-source

Apparently Arroyo is unlicensed 😅

chadwhitacre:
On it @rahul-kumar-saini, one sec ...

chadwhitacre:
Should be better now, @rahul-kumar-saini. Also for now the check is not required to merge while we ensure it is working properly. It will start to be required once we trust it, hopefully in a week or so. Cheers! 🙂 🍻

snuba/datasets/storages/metrics.py
fpacifici: How do you plan to support the policy that needs a Kafka producer to be provided to the constructor ?

fpacifici (commented):
Did you test this locally by starting the metrics consumer and sending an invalid message? Without the PR the consumer should crash. Without it should not crash.

snuba/datasets/storages/metrics.py
rahul-kumar-saini: Wouldn't this work?

```python3
def produce_policy_closure() -> DeadLetterQueuePolicy:
    return ProduceInvalidMessagePolicy(
        KafkaProducer(build_kafka_producer_configuration(dead_letter_topic)),
        dead_letter_topic,
    )
```

snuba/datasets/storages/metrics.py
```diff
@@ -102,7 +113,9 @@
     ),
     query_processors=[ArrayJoinKeyValueOptimizer("tags"), TableRateLimit()],
     stream_loader=build_kafka_stream_loader_from_settings(
-        SetsAggregateProcessor(), default_topic=Topic.METRICS,
+        SetsAggregateProcessor(),
+        default_topic=Topic.METRICS,
+        dead_letter_queue_policy_closure=ignore_policy_closure,
```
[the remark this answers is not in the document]
rahul-kumar-saini: Would these storages ever have write in logic or message processing in future? I could just remove the DLQ for these if not.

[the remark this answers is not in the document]
onewland: We have no short- or long-term plans to use them. I think it's likely we'd make these read-only storages by the end of 2022 as a tech debt task.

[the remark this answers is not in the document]
rahul-kumar-saini: It looks like each storage has validation though, and each one should be able to throw something to a DLQ so I'll just leave it in for now.

https://github.com/getsentry/snuba/blob/010c5dccfc776eb1756a6a53f4c6e6d4ca0b3c40/snuba/datasets/metrics_bucket_processor.py#L75-L108

rahul-kumar-saini:
@fpacifici 
Tested locally with a `raise InvalidMessages()` and the DLQ catches the exception, otherwise consumer dies ✅

PR #2587, opened 2022-04-06
chore(retention): Fix errors and transactions retention period

nikhars:
When we receive messages from upstream, we are not really sure what
values are being sent for retention days. That means it is possible
that we could have any number for retention days. That is bad since
our partitions are by retention days and timestamps.

The change buckets the retention days to either 30 or 90 by using
the ceiling of the retention days parameter passed to us.


fpacifici (approved):
Please apply this to all other consumers that process messages with a retention time.

lynnagara:
This was originally intentional on my part, since a number of on premise users complained about the retention they had configured in Sentry not being respected.

nikhars:
> This was originally intentional on my part, since a number of on premise users complained about the retention they had configured in Sentry not being respected.

Can you elaborate a bit on this. Was the concern around retention longer than 90 days? Or the concern was that they would configure retention days as 60 days but the data was there for 90 days?

lynnagara:
@nikhars From memory it came up more in relation to users wanting shorter retention than the 30 option as they found Sentry quite space hungry.

PR #2590, opened 2022-04-08
feat(subscriptions): Record more metrics from the subscriptions verifier

lynnagara:
Now the subscriptions verifier classifies the results it sees into one of
four categories:
- missing from original topic
- missing from new topic
- present in both and result is the same
- present in both and result is different

A metric with the count of each is recorded for each second.


evanh (approved):
This works as is, but I would make these four cases separate metrics instead of separate tags. This is a stylistic nit, but with this change the metric will be meaningless unless you split by outcome. I feel like a metric should mean something before splitting by tags.

lynnagara:
> This works as is, but I would make these four cases separate metrics instead of separate tags. This is a stylistic nit, but with this change the metric will be meaningless unless you split by outcome. I feel like a metric should mean something before splitting by tags.

Fair enough though I would not call the metric is entirely meaningless. Combined it's the sum of all unique subscriptions seen within that second. Then it can be broken down by outcome type.

PR #2579, opened 2022-04-04
cleanup(metrics) - remove unused metrics materialized views and tables

onewland:
Show tables before:

```
7b701b64404c :) show tables

SHOW TABLES

┌─name──────────────────────────────────────────┐
│ discover_local                                │
│ errors_local                                  │
│ groupassignee_local                           │
│ groupedmessage_local                          │
│ metrics_buckets_local                         │
│ metrics_counters_buckets_local                │
│ metrics_counters_consolidated_mv_local        │
│ metrics_counters_local                        │
│ metrics_counters_mv_10s_local                 │
│ metrics_counters_mv_3600s_local               │
│ metrics_counters_mv_86400s_local              │
│ metrics_counters_mv_local                     │
│ metrics_counters_polymorphic_mv_local         │
│ metrics_counters_polymorphic_mv_v2_local      │
│ metrics_counters_polymorphic_mv_v3_local      │
│ metrics_counters_polymorphic_mv_v4_local      │
│ metrics_counters_v2_local                     │
│ metrics_distributions_buckets_local           │
│ metrics_distributions_consolidated_mv_local   │
│ metrics_distributions_local                   │
│ metrics_distributions_mv_10s_local            │
│ metrics_distributions_mv_3600s_local          │
│ metrics_distributions_mv_86400s_local         │
│ metrics_distributions_mv_local                │
│ metrics_distributions_polymorphic_mv_local    │
│ metrics_distributions_polymorphic_mv_v2_local │
│ metrics_distributions_polymorphic_mv_v3_local │
│ metrics_distributions_polymorphic_mv_v4_local │
│ metrics_distributions_v2_local                │
│ metrics_raw_local                             │
│ metrics_raw_v2_local                          │
│ metrics_sets_consolidated_mv_local            │
│ metrics_sets_local                            │
│ metrics_sets_mv_10s_local                     │
│ metrics_sets_mv_3600s_local                   │
│ metrics_sets_mv_86400s_local                  │
│ metrics_sets_mv_local                         │
│ metrics_sets_polymorphic_mv_local             │
│ metrics_sets_polymorphic_mv_v2_local          │
│ metrics_sets_polymorphic_mv_v3_local          │
│ metrics_sets_polymorphic_mv_v4_local          │
│ metrics_sets_v2_local                         │
│ migrations_local                              │
│ outcomes_hourly_local                         │
│ outcomes_mv_hourly_local                      │
│ outcomes_raw_local                            │
│ sessions_hourly_local                         │
[...]

PR #2588, opened 2022-04-07
ref(devserver): update metrics-consumer command

MeredithAnya:
When trying to run sentry tests that depend on the snuba image, I'm getting:

```
23:24:21 metrics-consumer | Usage: snuba multistorage-consumer [OPTIONS]
23:24:21 metrics-consumer | Try 'snuba multistorage-consumer --help' for help.
23:24:21 metrics-consumer |

23:24:21 metrics-consumer | Error: Invalid value for '--storage': invalid choice: metrics_counters_buckets. (choose from groupedmessages, groupassignees, metrics_distributions, metrics_sets, metrics_counters, metrics_raw, errors, outcomes_raw, querylog, sessions_raw, transactions, spans, transactions_v2, errors_v2, profiles)
```

While I can fix this by just using local snuba instead we should probs update this for other people (who have will be using `SENTRY_USE_METRICS_DEV` in sentry)

PR #2591, opened 2022-04-08
fix(metrics): Remove bucket storages from devserver

jjbayer:
Follow-up to https://github.com/getsentry/snuba/pull/2579:

`snuba devserver` recently failed to start with the following error message:

```
Invalid value for '--storage': invalid choice: metrics_counters_buckets. (choose from groupedmessages, groupassignees, metrics_distributions, metrics_sets, metrics_counters, metrics_raw, errors, outcomes_raw, querylog, sessions_raw, transactions, spans, transactions_v2, errors_v2, profiles)
```

This PR fixes the error by removing remaining traces of `metrics_*buckets`.


onewland (approved):
Thanks for submitting this change -- I'm embarrassed I didn't catch this.

snuba/cli/devserver.py
```diff
@@ -144,9 +144,6 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                 [
                     "snuba",
                     "multistorage-consumer",
```
onewland, line 146: let's change this to be just `consumer` (there is only one storage) like in Meredith's PR in #2588

onewland: no problem, you had the ship it already

PR #2595, opened 2022-04-11
feat(subscriptions): Log non matching results to Sentry

lynnagara:


lynnagara:
@getsentry/open-source Do you know why this is failing the license validation check?

snuba/subscriptions/verifier.py
```diff
@@ -282,22 +283,34 @@ def _record(self, until: int) -> None:
...
 
             self.__metrics.increment(
                 METRIC_OUTCOME_NAME,
-                len(intersection_ids) - matching_results,
+                len(non_matching_results),
                 {"outcome": "different_result"},
             )
 
+            for id in non_matching_results:
+                logger.warning(
```
onewland, line 305: So should we have 100% matching because subscriptions only run based on matching commit offsets? Or should this perhaps be sampled because it's normal for e.g. backdated data to be inserted after tick creation but before subscription?

chadwhitacre:
> Do you know why this is failing the license validation check?

Sorry for the trouble, should be back on track now. 🤞

nikhars (approved):
LGTM

snuba/subscriptions/verifier.py
lynnagara: There's not a lot of non matching results normally, but I'll still make a change to limit it to maybe 100 or so per second just in case.

PR #2596, opened 2022-04-11
feat(subscriptions): Ignore stale subscriptions

lynnagara:
The subscriptions scheduler now supports passing --stale-threshold-seconds.
If a value is passed, subscriptions whose scheduled timestamp is more than
this number of seconds before the current time in the scheduler will be ignored.
This means that if the scheduler sees stale subscriptions they will be skipped
instead of scheduled for execution.

Under normal circumstances we should not really see stale subscriptions, however
there are a few situations where it happens.
- The main consumers were heavily backlogged for some reason
- Consumer or scheduler crashed for some time and we need to catch up
- In dev when devserver is stopped for an extended period of time

Since very delayed alerts are generally of little value to users, having the option
to skip them will be useful in some environments.

### DIFF
--- a/snuba/cli/subscriptions_scheduler.py
+++ b/snuba/cli/subscriptions_scheduler.py
@@ -48,6 +48,11 @@ logger = logging.getLogger(__name__)
 @click.option("--schedule-ttl", type=int, default=60 * 5)
 @click.option("--log-level", help="Logging level to use.")
 @click.option("--delay-seconds", type=int)
+@click.option(
+    "--stale-threshold-seconds",
+    type=int,
+    help="Skip scheduling if timestamp is beyond this threshold compared to the system time",
+)
 def subscriptions_scheduler(
     *,
     entity_name: str,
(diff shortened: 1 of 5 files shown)


onewland (approved):

snuba/cli/subscriptions_scheduler.py
```diff
@@ -48,6 +48,11 @@
 @click.option("--schedule-ttl", type=int, default=60 * 5)
 @click.option("--log-level", help="Logging level to use.")
 @click.option("--delay-seconds", type=int)
+@click.option(
+    "--stale-threshold-seconds",
+    type=int,
+    help="Skip scheduling if timestamp is beyond this threshold compared to the system time",
+)
```
nikhars, line 55: Should we add this to the subscriptions executor as well? It is possible that the scheduler is running ok but the executor was down for 5 minutes. Shouldn't the executor not run the queries in that case?

lynnagara: Yes, I will add the same to the executor too

PR #2597, opened 2022-04-11
fix(query): Avoid race in pipeline delegator

nikhars:
There's a data race which happens when the query is run via the
pipeline delegator. The same query object is modified by the entity
processors being run by different threads. To solve this race, we
create a copy of the query object and use it to avoid the race.


nikhars:
I will be disabling running dual subscription queries since Lyn is already going to verify the results with the new subscriptions executor/verifier. So the performance problem on subscriptions won't be seen.

fpacifici (approved):

lynnagara (approved):

PR #2602, opened 2022-04-12
feat: Add bucketed_started to sessions Entity

evanh:
The bucketed_started column is a virtual column, that ultimately gets
transformed into a function on the started column. However since it is a valid
column in the SnQL query, it should really be part of the Entity data model.

Also create a new test class that uses SnQL instead of legacy, that we can
start migrating tests into over time.


tests/test_sessions_api.py
```diff
@@ -283,6 +283,136 @@ def test_minute_granularity_range(self, get_project_id: Callable[[], int]) -> No
         }
 
 
+class TestSessionsApi(BaseSessionsMockTest, BaseApiTest):
```
volokluev, line 286: Are these tests related to the column change?

volokluev: how would it fail if the column change wasn't made

evanh: It won't on it's own. This is a small part of the larger initiative to start validating columns. I'm breaking that work down into more reviewable chunks. Ultimately I will turn on column validation for this entity, and at that point this test would start failing without the check.

volokluev: So when validation is turned on this test is expected to fail? On what line? 

Also, is the generating of data necessary for the validation test?

evanh: > So when validation is turned on this test is expected to fail? On what line?

This PR both introduces an error and fixes it. Turning the column validation to `WARN` would cause warning messages to get sent to Sentry, except that I added `bucketed_started` to the Entity data model, which should fix the warnings. So no, the test should pass and continue to pass. I added it because we weren't testing that code path at all.

> Also, is the generating of data necessary for the validation test?

I want to make sure that `bucketed_started` was returning bucketed data correctly.

volokluev (approved):

PR #2604, opened 2022-04-12
feat: Add time to outcomes_raw Entity data model

evanh:
This column is a valid column to query, so it should be part of the logical data
model, even if it does get mapped to a different column later on.


tests/test_outcomes_api.py
```diff
@@ -294,3 +294,125 @@ def test_category_quantity_sum_querying(
             },
         ]
         assert data["data"] == correct_data
+
+
+class TestOutcomesAPI(BaseApiTest):
```
fpacifici, line 299: I don't think we should test this change with a full integration test but with a specific unit test that stress validation.
- Failures on integration tests are generally tricky to troubleshoot. The test does not point you to the specific issue but it just tells you that something in the whole query processing + execution is wrong.
- You do not need to bear the full performance impact of running the query end to end just to test validation.
- You are not testing the invalid query case.

snuba/datasets/entities/outcomes_raw.py
```diff
@@ -20,13 +21,14 @@
 class OutcomesRawEntity(Entity):
     def __init__(self) -> None:
         storage = get_storage(StorageKey.OUTCOMES_RAW)
-
+        read_columns = storage.get_schema().get_columns()
+        time_columns = ColumnSet([("time", DateTime())])
```
rahul-kumar-saini, line 25: Couldn't we just extended the original columns in the storage schema to add a time column? (I'm not sure if that would have other implications)

evanh: The storage schema is supposed to reflect what is actually in the tables, while bucketed_time is a "magic" column that only exists on the Entity and not in the storage itself. That's why I added it only to the Entity.

nikhars (approved):
LGTM

PR #2605, opened 2022-04-13
feat(subscriptions): Ignore stale subscriptions in executor

lynnagara:
Same idea as https://github.com/getsentry/snuba/pull/2596 in the scheduler.
In case the executor was down or not making progress, it may be useful
to skip execution of stale subscriptions as these are likely not not
be useful anymore after a certain point.


snuba/subscriptions/executor_consumer.py
```diff
@@ -326,6 +334,14 @@ def submit(self, message: Message[KafkaPayload]) -> None:
             subscription_id_to_float(subscription_id) < executor_sample_rate
         )
 
+        # Don't execute stale subscriptions
+        if (
+            self.__stale_threshold_seconds is not None
+            and time.time() - datetime.timestamp(task.timestamp)
+            >= self.__stale_threshold_seconds
+        ):
+            should_execute = False
```
fpacifici, line 343: THis would not commit though while you churn the backlog. Is that correct and intentional?

fpacifici (commented):
I wonder if a better option would be to use the dead letter queue for this use case and send the stale subscriptions there so that we can separately decide to store them and maybe schedule offline slowly.

lynnagara:
> I wonder if a better option would be to use the dead letter queue for this use case and send the stale subscriptions there so that we can separately decide to store them and maybe schedule offline slowly.

I'm not sure this is a good use case for the dead letter queue. I can't imagine a scenario where we would want to execute or do anything with an earlier subscription after later ones have already run. I don't imagine we'll use this setting in production at least right now. The more immediate motivation for this is for dev and self hosted environments where a consumer is more likely to be stopped for a longer period of time and when it is restarted after, say, a few days, causes an insane number of subscriptions to be executed all at once.

snuba/subscriptions/executor_consumer.py
lynnagara: Well, it was somewhat intentional in that I didn't want to have to bring the commit function over to the ExecuteQuery just for this- committing usually happens in the following step. My thinking was that if no query is actually being executed, the backlog can be churned so fast by the executor that in practice not committing until we see a recent timestamp will never really be an issue.

fpacifici:
I can see this use case, whether or not we execute them afterwards: consume the dead letter queue and notify the the involved customers that alerts were dropped for a period of time. Or even for us to assess the impact of the incident. 
For dev I think we could do this, but in order to properly support this feature we should have something that preserves the skipped queries somewhere. Thus the dead letter queue. 
In principle a dead letter queue is not only for invalid messages. It is meant to put work aside for whatever reason. 

What do you think? I can accept this for the sake of dev but let's track the follow up somewhere like in a task and TODO

lynnagara:
@fpacifici I think what you suggest is reasonable but is also a bigger feature and we have yet to define what we actually want to achieve there. My goal here is just to solve some big immediate pain in non-prod environments.

lynnagara:
@fpacifici Any objections to merging this as-is? DLQ sounds fine but what you're suggesting cannot be a high priority right now
- the DLQ functionality doesn't exist yet and is still being built
- we're not running it in prod

It does however cause a lot of pain in dev

snuba/subscriptions/executor_consumer.py
nikhars: Now that you would be introducing the commit function via https://github.com/getsentry/snuba/pull/2656/files#diff-b81082d36647653087818cb560a918148759eb0e770ef1c04313589ea32167d7R196 to this processing step, you have an option to actually perform a commit in this step. Not saying that you need to since you mentioned that it's a temporary code in the other PR.

nikhars (approved):
Left a minor comment else changes look good to me.

PR #2606, opened 2022-04-13
feat: Bump arroyo to 0.0.16

lynnagara:
Includes changes to support incremental assignments in the stream processor.
This is required in order to switch to using the cooperative-sticky strategy
with Arroyo.


fpacifici (approved):

lynnagara:
The count policy changed as well, had to change CountInvalidMessagePolicy slightly.

PR #2608, opened 2022-04-13
fix(query): Avoid race in pipeline delegator

nikhars:
This fixes the race in pipeline delegator by making a copy of the query.
But there is an override mechanism via options wherein we can override
copying query for certain referrers. These would be the
subscription and subscriptions_executor referrers since they are
too sensitive to copying and they would be disallowed from running
parallel queries.

### DIFF
--- a/snuba/pipeline/pipeline_delegator.py
+++ b/snuba/pipeline/pipeline_delegator.py
@@ -1,6 +1,9 @@
+import copy
+from dataclasses import replace
 from functools import partial
 from typing import Callable, List, Mapping, Optional, Tuple
 
+from snuba import state
 from snuba.datasets.plans.query_plan import ClickhouseQueryPlan, QueryRunner
 from snuba.pipeline.query_pipeline import (
     QueryExecutionPipeline,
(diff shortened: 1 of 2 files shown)


snuba/pipeline/pipeline_delegator.py
```diff
@@ -26,6 +29,20 @@
 ]
 
 
+def _is_query_copying_disallowed(current_referrer: str) -> bool:
+    """
+    Check if the given referrer is disallowed from making a copy of the query.
```
MeredithAnya, line 34: maybe you could add why a given referrer would be disallowed from making a copy of the query? I'm not sure it's fully obvious to me why subscriptions and subscriptions executor are too sensitive for using deepcopy

onewland:
Too sensitive is referring to the increased query latency we saw on the original deploy of that change, right?

onewland:
Is there any way we could do gradual/run-time config roll-out so that it's very straightforward to roll back?

evanh (approved):
We need a better way of doing this. That's a bigger problem though.

PR #2609, opened 2022-04-14
feat(subscriptions): Improve the off by one calculation in the verifier

lynnagara:
Some subscription results are a list of dicts and not just an integer.
Update the off by one calculation to also take the other format into
account.


lynnagara:
This is probably not right

PR #2610, opened 2022-04-14
fix(subscriptions): Fix off by one calculation in verifier

lynnagara:
There were a number of bugs in https://github.com/getsentry/snuba/pull/2607.
It was always returning False for `is_off_by_one()`, and then logging the off by
one results to Sentry (always none due to this bug) instead of those that are
different in more significant ways.

### DIFF
--- a/snuba/subscriptions/verifier.py
+++ b/snuba/subscriptions/verifier.py
@@ -51,14 +51,19 @@ class SubscriptionResultData:
         data_a = a.pop("data")
         data_b = b.pop("data")
 
-        if (
-            a == b
-            and isinstance(data_a, int)
-            and isinstance(data_b, int)
-            and abs(data_a - data_b) == 1
-        ):
-            return True
-
+        # Data is a list of dicts
+        try:
+            if len(data_a) == len(data_b):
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/verifier.py
```diff
@@ -51,14 +51,19 @@ def is_off_by_one(self, other: SubscriptionResultData) -> bool:
...
-
+        # Data is a list of dicts
+        try:
+            if len(data_a) == len(data_b):
+                for (a, b) in zip(data_a, data_b):
+                    a_keys = a.keys()
+                    b_keys = b.keys()
+                    if a_keys == b_keys:
+                        for key in a_keys:
+                            if abs(a[key] - b[key]) > 1:
```
evanh, line 62: Just to be clear, if a single result is off by more than one, the entire result will be marked as False. Is that intended?

evanh (approved):

snuba/subscriptions/verifier.py
lynnagara: Yup. True -> Off by only one (or less). False -> Off by more than one.

PR #2612, opened 2022-04-14
fix: Enable pre-commit CI check

asottile-sentry:
commits are split into atomic tasks -- it's recommended to view the reformat commits with [?w=1](https://youtu.be/qYJWBJvHqj0)

___

previously the check in CI was always silently passing due to `master` not being
available with `actions/checkout@v2`:

```
fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,20 @@ jobs:
           python-version: 3.8
       - name: Setup pre-commit
         run: make setup-git
+      - uses: getsentry/paths-filter@v2
+        id: files
+        with:
+          # Enable listing of files matching each filter.
+          # Paths to files will be available in `${FILTER_NAME}_files` output variable.
+          # Paths will be escaped and space-delimited.
+          # Output is usable as command line argument list in linux shell
+          list-files: shell
+
+          # It doesn't make sense to lint deleted files.
+          # Therefore we specify we are only interested in added or modified files.
+          filters: |
+            all:
+              - added|modified: '**/*'
       - name: Run pre-commit checks
         # Run pre-commit to lint and format check files that were changed (but not deleted) compared to master.
         # XXX: there is a very small chance that it'll expand to exceed Linux's limits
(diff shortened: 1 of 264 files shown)


asottile-sentry (commented):
I read through the automated patches and commented on the surprises

evanh:
I'm a little confused, what is the actual change here? Even ignoring whitespace this PR is >2000 lines long. Are all the reformatting changes because of a new version of a linter? If so, can you configure black to not make all these changes?

asottile-sentry:
> I'm a little confused, what is the actual change here? Even ignoring whitespace this PR is >2000 lines long. Are all the reformatting changes because of a new version of a linter? If so, can you configure black to not make all these changes?

the linters are "configured" but aren't being enforced in CI -- this PR fixes them to be enforced in CI (the fourth commit) and resolves the drift from un-linted/un-formatted changes (second and third commits) which have been introduced due to the CI check erroneously passing.

evanh:
OK I understand what is being changed.

> which have been introduced due to the CI check erroneously passing.

The pre-commit linters were still happening locally though. My suggestion here would be to fix the CI check, but don't lint the entire code base. The files that are currently not linted correctly will get fixed if/when they get changed in subsequent PRs.

asottile-sentry:
> OK I understand what is being changed.
> 
> > which have been introduced due to the CI check erroneously passing.
> 
> The pre-commit linters were still happening locally though. My suggestion here would be to fix the CI check, but don't lint the entire code base. The files that are currently not linted correctly will get fixed if/when they get changed in subsequent PRs.

they were happening when people used them, but as far as I can tell they weren't running in the general case (`black` for instance in its current configuration cannot run successfully at all!  the version configured always crashes with the common `_unicodefun` error: https://github.com/psf/black/issues/2964)

generally I [recommend doing it all in one PR](https://github.com/readme/guides/code-review) otherwise other PRs will end up with unrelated linter/formatter noise which makes it more difficult to accurately review changes being made.  I specifically separated the noisy commits into separately reviewable patches for this reason

evanh (approved):
I meant to say, I think this would have been working when it was run as part of the pre-commit locally on `git commit`. However this is fine, I'll approve before you get more merge conflicts.

PR #2613, opened 2022-04-14
fix: Match only migration files in scripts/ddl-changes.py

asottile-sentry:
previously ddl-changes.py would match non-migration files and error.

I removed the fragile regex matching and moved the "added/deleted" and
path matching into the `git diff` command itself.

___

testing done: see https://github.com/getsentry/snuba/pull/2612 where I cherry-picked this commit on top to fix CI

### DIFF
--- a/scripts/ddl-changes.py
+++ b/scripts/ddl-changes.py
@@ -1,7 +1,6 @@
-import re
+import os.path
 import subprocess
 from shutil import ExecError
-from sys import stderr
 
 from snuba.migrations.groups import MigrationGroup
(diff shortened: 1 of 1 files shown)


asottile-sentry:
the diff is slightly easier to view using [?w=1](https://github.com/getsentry/snuba/pull/2613/files?w=1) -- more on that [here](https://youtu.be/qYJWBJvHqj0)

onewland (approved):
who wrote that messy code in the first place? 😄

PR #2614, opened 2022-04-14
dlq(metrics): Metrics produce policy + basic validation replaced with DLQ stuff

rahul-kumar-saini:
### Overview
- DLQ stuff is set up and hasn't broken anything yet
- Introduced a policy to produce invalid messages to dead letter topic
- Updated existing basic validation on metrics which would have restarted consumers to now throw messages to DLQ instead
- These should show up on the metrics dead letter topic and be visible in Datadog metrics


snuba/datasets/metrics_bucket_processor.py
```diff
@@ -16,8 +18,10 @@
     ProcessedMessage,
     _ensure_valid_date,
 )
+from snuba.utils.streams.topics import Topic
 
 DISABLED_MATERIALIZATION_VERSION = 1
+METRICS_TOPIC = Topic.METRICS.value
```
rahul-kumar-saini, line 24: Hard coding the topic here since all messages processed by these processors will come from the metrics topic. This topic is just telling the DLQ (and in turn whoever looks at the bad messages in the dead letter topic) where the message was originally produced to. No implications on any logic.

snuba/datasets/metrics_bucket_processor.py
```diff
@@ -101,10 +113,13 @@ def _should_process(self, message: Mapping[str, Any]) -> bool:
 
     def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
         values = message["value"]
-        for v in values:
-            assert isinstance(
-                v, (int, float)
-            ), "Illegal value in set. Int expected: {v}"
```
rahul-kumar-saini, line 107: Some of these comments mismatched what the check was, take this one for example. It's asserting and `int` or `float` but comment says "Int expected". I updated these wherever I found a discrepancy, should be reviewed.

snuba/datasets/metrics_bucket_processor.py
```diff
@@ -78,8 +82,13 @@ def _should_process(self, message: Mapping[str, Any]) -> bool:
 
     def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
         values = message["value"]
-        for v in values:
-            assert isinstance(v, int), "Illegal value in set. Int expected: {v}"
```
rahul-kumar-saini, line 82: These were probably meant to be f-strings and not regular strings so I updated them.

evanh (approved):
Looks good to me but would be great to get someone from storage to do a once over.

snuba/utils/streams/topics.py
```diff
@@ -28,6 +28,8 @@ class Topic(Enum):
     QUERYLOG = "snuba-queries"
     PROFILES = "processed-profiles"
     DEAD_LETTER_QUEUE_INSERTS = "snuba-dead-letter-inserts"
+    DEAD_LETTER_METRICS = "snuba-dead-letter-metrics"
+    DEAD_LETTER_SESSIONS = "snuba-dead-letter-sessions"
```
onewland, line 32: is the sessions DLQ addition intended?

onewland (approved):

snuba/utils/streams/topics.py
rahul-kumar-saini: Yeah, both of these topics exist now so I just added it here to use when we deploy a DLQ on sessions.

PR #2615, opened 2022-04-15
feat(dlq): Added CLI tool to display all messages in dead letter topic

rahul-kumar-saini:
### Overview
- Added a CLI command to list all messages in a dead letter topic from earliest to latest.
- Added a CLI command to show some info about a storage's dead letter queue.
- Added two new topics to list of topics
    - https://getsentry.atlassian.net/browse/OPS-1683 

Usage looks like:
```
$ snuba dlq-manager list --storage=<StorageSetKey>

Listing the first 10 messages from offset 0 in <storage> dead-letter topic:
(see --help for pagination)

--------------------------------------------------
{
  "offset": <offset in dead-letter topic>,
  "dlq-topic-timestamp": <timestamp at which dead-letter topic received the message>,
  "invalid_message": <message thrown to the DLQ>,
   ...
}
--------------------------------------------------
...
--------------------------------------------------
```
Or with pagination args:
```
$ snuba dlq-manager list --storage=<StorageSetKey> --offset=200 --limit=20

Listing the first 20 messages from offset 200 in <storage> dead-letter topic:
(see --help for pagination)
...
```

Info command:
```
$ snuba dlq-manager info --storage=<StorageSetKey>

Displaying info for <storage>:

Dead letter topic name: snuba-dead-letter-<storage>
Earliest offset: <earliest offset on dead-letter topic>
Latest offset: <latest offset on dead-letter topic>
```


snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,57 @@
...
+            bootstrap_servers=bootstrap_servers,
+            group_id="dlq-manager-consumer",
+            auto_offset_reset="earliest",
+        )
+    )
+    consumer.subscribe([Topic(StreamTopics.DEAD_LETTER_TOPIC.value)])
+    messages: MutableSequence[Message[KafkaPayload]] = []
+    while True:
+        try:
+            message = consumer.poll(10)
```
rahul-kumar-saini, line 49: For some reason, giving any timeout less than 10 seconds was flakey locally. It would just not show any messages half the time.

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,57 @@
...
+
+    consumer = KafkaConsumer(
+        build_kafka_consumer_configuration(
+            topic=StreamTopics.DEAD_LETTER_TOPIC,
+            bootstrap_servers=bootstrap_servers,
+            group_id="dlq-manager-consumer",
+            auto_offset_reset="earliest",
+        )
+    )
+    consumer.subscribe([Topic(StreamTopics.DEAD_LETTER_TOPIC.value)])
```
rahul-kumar-saini, line 45: This config works locally, I'm not sure if it's the correct way to do it so it works in prod.

snuba/utils/streams/topics.py
```diff
@@ -26,6 +26,7 @@ class Topic(Enum):
     QUERYLOG = "snuba-queries"
     PROFILES = "processed-profiles"
     DEAD_LETTER_QUEUE_INSERTS = "snuba-dead-letter-inserts"
+    DEAD_LETTER_TOPIC = "snuba-dead-letter-topic"
```
rahul-kumar-saini, line 29: I'm not sure if there will be a new dead letter topic or if `snuba-dead-letter-inserts` is the one we will be using. Whatever the topic name is will be updated here if it is going to be a new one.

snuba/cli/dlq_manager.py
rahul-kumar-saini: Should any of these configs be passed in with CLI flags?

snuba/utils/streams/topics.py
lynnagara: Are you saying everyone should use the same deal letter topic? How will we know which topic the messages originally came from in that case? I'm also not sure what `snuba-dead-letter-inserts` is on the line above.

rahul-kumar-saini: I believe the plan was always to have a single dead letter topic, with the original topic info encoded in the messages themselves. If you look at the example in the main comment of the PR you can see the structure I think we should use in this case. I think `snuba-dead-letter-inserts` is a topic that was created for @MeredithAnya's temporary DLQ.

rahul-kumar-saini: Update, each would use a different DLQ.

onewland:
We probably want to paginate here, right? Maybe arguments could be passed for minimum offset/max result count?

I can imagine a deploy issue producing 10s or 100s of thousands of DLQ messages in a very short period of time.

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,149 @@
...
+    else:
+        click.echo(f"\nNo messages found in {storage_set} dead-letter topic!")
+    click.echo(line_break)
+
+
+def _consume_dead_letters(
+    storage_set: str, offset: int, limit: int
+) -> Sequence[Message[KafkaPayload]]:
+    dead_letter_topic_arroyo = ArroyoTopic(
+        STORAGES_WITH_DLQ[StorageSetKey(storage_set)].value
```
evanh, line 75: I would probably handle the case that someone provided an invalid storage set more gracefully here.

onewland:
Looks good to me, but agree with Evan that it would be nice to handle invalid input more gracefully than a key error

snuba/cli/dlq_manager.py
rahul-kumar-saini: @onewland @evanh The click library handles this for me, the storage arg is a "choice" of items in that list defined in the type for the arg.

Eg
```bash
$ snuba dlq-manager list --storage abc
Usage: snuba dlq-manager list [OPTIONS]
Try 'snuba dlq-manager list --help' for help.

Error: Invalid value for '--storage': 'abc' is not one of 'metrics', 'sessions'.
```

nikhars:
Under what circumstances would we be using this tool? And is this tool built to help ops or SnS team? I am trying to figure out who would benefit from this tool. Why would someone use this over some open-source tool like kafkacat etc.

rahul-kumar-saini:
@nikhars This just makes it easier to query existing dead letter queue stuff, eventually we'd want to be able to replay certain messages which could be done by extending this tool. This could eventually be exposed to the admin tool as well so we have access to prod dead letter topics?

Someone could definitely do everything this CLI does right now by using kcat or something, but Kafka CLI tools are a little unintuitive, so I could see someone on call having an easier time debugging using this instead.

nikhars:
> @nikhars This just makes it easier to query existing dead letter queue stuff, eventually we'd want to be able to replay certain messages which could be done by extending this tool. This could eventually be exposed to the admin tool as well so we have access to prod dead letter topics?
> 
> Someone could definitely do everything this CLI does right now by using kcat or something, but Kafka CLI tools are a little unintuitive, so I could see someone on call having an easier time debugging using this instead.

Thanks for the response. The reason I brought it up is to see if we can make it a generic CLI to interact with Kafka. The change I would recommend is making it possible to read from any topic. Not just the dead letter topic. It should be a fairly straightforward change. Then when you expose it in snuba-admin, it's not just restricted to dead letter topics.

onewland:
> > @nikhars This just makes it easier to query existing dead letter queue stuff, eventually we'd want to be able to replay certain messages which could be done by extending this tool. This could eventually be exposed to the admin tool as well so we have access to prod dead letter topics?
> > Someone could definitely do everything this CLI does right now by using kcat or something, but Kafka CLI tools are a little unintuitive, so I could see someone on call having an easier time debugging using this instead.
> 
> Thanks for the response. The reason I brought it up is to see if we can make it a generic CLI to interact with Kafka. The change I would recommend is making it possible to read from any topic. Not just the dead letter topic. It should be a fairly straightforward change. Then when you expose it in snuba-admin, it's not just restricted to dead letter topics.

One reason I can think of that you may want a utility for "give me the DLQ messages for this main topic" is that it saves you the step of finding the DLQ topic name

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+@dlq_manager.command()
+@click.option(
+    "--storage",
+    "storage_set",
+    type=click.Choice(
+        [storage_set_key.value for storage_set_key in STORAGES_WITH_DLQ.keys()]
+    ),
+    help="The storage set to list dead letters for",
+    required=True,
```
lynnagara, line 30: A `storage` and a `storage set` are two quite distinct things in Snuba. Which one is this supposed to be?

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,151 @@
...
+)
+@click.option(
+    "--offset",
+    "offset",
+    type=int,
+    help="An offset to start reading from, 0 by default",
+)
```
lynnagara, line 37: why don't you pass default=0 here

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,151 @@
...
+    dead_letter_topic_snuba = STORAGES_WITH_DLQ[StorageSetKey(storage_set)]
+    click.echo(f"\nDisplaying info for {storage_set}:\n")
+    click.echo(f"Dead letter topic name: {dead_letter_topic_snuba.value}")
+    click.echo(f"Earliest offset: {earliest_offset}")
+    click.echo(f"Latest offset: {latest_offset}")
+
+
+def _get_offsets_info(storage_set: str) -> Tuple[Optional[int], Optional[int]]:
+    consumer = _build_consumer(storage_set)
+    message = consumer.poll(10)
```
lynnagara, line 128: Don't these functions take too long to run if the DLQ is empty?

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,151 @@
...
+    earliest_offset = message.offset if message else None
+    latest_offset = None
+    while message is not None:
```
lynnagara, line 131: Will this exit the while loop if there is nothing in the DLQ?

snuba/datasets/storages/factory.py
```diff
@@ -111,6 +113,11 @@
     **NON_WRITABLE_STORAGES,
 }
 
+STORAGES_WITH_DLQ: Mapping[StorageSetKey, Topic] = {
+    metrics_polymorphic_storage.get_storage_set_key(): Topic.DEAD_LETTER_METRICS,
+    sessions_raw_storage.get_storage_set_key(): Topic.DEAD_LETTER_SESSIONS,
+}
+
```
lynnagara, line 120: For our other Kafka topics, the topic is generally a property of the storage / stream loader definition. I don't think we should start a new convention here.

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        STORAGES_WITH_DLQ[StorageSetKey(storage_set)].value
+    )
+    consumer = _build_consumer(storage_set)
+    messages: MutableSequence[Message[KafkaPayload]] = []
+
+    if offset != 0:
+        try:
+            message = consumer.poll(10)
+            consumer.tell()
+            consumer.seek({Partition(dead_letter_topic_arroyo, 0): offset})
```
lynnagara, line 84: How come hard coding the partition index to 0? What if the dead letter topic has more than one partition?

lynnagara:
Can you clarify the motivation for this PR. Is it mostly for debugging in dev? Or for prod? If prod, how would it get used?

snuba/cli/dlq_manager.py
rahul-kumar-saini: The while loop doesn't start if there's nothing in the DLQ since `message` would be `None` to begin with

snuba/cli/dlq_manager.py
rahul-kumar-saini: I do mean storage set but I'm not entirely sure which it should be, DLQ's are defined per storage but I think storages under the same storage set share the same dead-letter topic.

snuba/cli/dlq_manager.py
rahul-kumar-saini: Yeah it takes too long to run, but when this timeout is less than 10 it's really flakey. I'm not sure why, it sometimes just doesn't get a message even though the topic has messages.

rahul-kumar-saini:
> Can you clarify the motivation for this PR. Is it mostly for debugging in dev? Or for prod? If prod, how would it get used?

@lynnagara 
I tried to clarify here:
https://github.com/getsentry/snuba/pull/2615#issuecomment-1124213973

But overall this tool is meant to make it easier to debug in prod. Someone with access to the prod Snuba CLI could quickly go in and grab invalid messages for us to debug on whatever storage starts to produce invalid messages.

Also see [SNS-961](https://getsentry.atlassian.net/browse/SNS-961), but this tool isn't doing everything we want yet.

nikhars:
> > Thanks for the response. The reason I brought it up is to see if we can make it a generic CLI to interact with Kafka. The change I would recommend is making it possible to read from any topic. Not just the dead letter topic. It should be a fairly straightforward change. Then when you expose it in snuba-admin, it's not just restricted to dead letter topics.
> 
> @nikhars I want to test that this tool actually works for now, I feel it might be a little flakey. Extending this or creating a separate CLI to just interact with Kafka and integrating it with the Snuba admin console could be done as a tech debt task. (I want to close out DLQ related tasks for now)

I am fine with merging this change to keep things moving. It's not a blocker for me. I will let Lyn give the final approval to merge since she had some comments.

snuba/cli/dlq_manager.py
lynnagara: Personally I think it should be defined at the storage level. Not all storages in a storage set might have a DLQ. That does not mean that storages can never share a DLQ, just that they are defined that the storage level and are not the same by default.

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,158 @@
...
+    type=click.Choice([storage_set for storage_set in STORAGE_SETS_WITH_DLQ]),
+    help="The storage set to list dead letters for",
+    required=True,
+)
+@click.option(
+    "--offset",
+    "offset",
+    type=int,
+    help="An offset on the dead-letter topic to start reading from, 0 by default",
+    default=0,
```
lynnagara, line 46: I'm not sure this is a sensible default we should pick. I think "earliest" would be better.

snuba/cli/dlq_manager.py
lynnagara: I see, I'd probably print a message for the user if it's waiting up to 10 seconds

snuba/datasets/storages/sessions.py
```diff
@@ -115,21 +116,23 @@ def process_query(self, query: Query, request_settings: RequestSettings) -> None
             )
 
 
+partially_built_stream_loader = partial(
+    build_kafka_stream_loader_from_settings,
+    processor=SessionsProcessor(),
+    default_topic=Topic.SESSIONS,
+    dead_letter_queue_topic=Topic.DEAD_LETTER_SESSIONS,
+)
```
lynnagara, line 124: Can you just put this inline below (even if there's a little duplication)? Sessions subscriptions is about to be killed and it'll make it easier to clean up.

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,158 @@
...
+    .get_stream_loader()
+    .get_dead_letter_queue_topic()
+    for storage in WRITABLE_STORAGES.values()
+}
+
+STORAGE_SETS_WITH_DLQ = {
+    storage.get_storage_set_key().value: topic
+    for storage, topic in __STORAGE_TO_DLQ_TOPIC.items()
+    if topic is not None
+}
```
lynnagara, line 22: I don't think you need this since DLQ topic is a property of the storage

snuba/cli/dlq_manager.py
rahul-kumar-saini: Technically, it is earliest by default since it only applies if the user passes anything other than 0. I'll just update the help message.

lynnagara: Should we also pass a partition index in? Otherwise how will you know which partition this offset refers to?

lynnagara:
Would it be possible to add some tests here?

snuba/cli/dlq_manager.py
rahul-kumar-saini: This makes more sense than just moving all partitions to the given offset, I'll update 😅

snuba/cli/dlq_manager.py
rahul-kumar-saini: If the DLQ is shared across the storages in a set, it would be weird to choose from a list of storages and not storage sets for the CLI. Choosing any storage under the same set would show the same results when querying for invalid messages.

Could we ever have different storages under the same storage set using different DLQ topics?

rahul-kumar-saini:
> Would it be possible to add some tests here?

I can try adding some simple tests once we're happy with the CLI

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,194 @@
...
+    else:
+        click.echo(f"\nNo messages found in {storage_set} dead-letter topic!")
+    click.echo(LINE_BREAK)
+
+
+def _consume_dead_letters(
+    storage_set: str, offset: int, partition_index: int, limit: int
+) -> Sequence[Message[KafkaPayload]]:
+    consumer = _build_consumer(storage_set)
+    message = consumer.poll(10)
```
lynnagara, line 94: I think we need to clarify the multiple partition scenario, there is no guarantee this message is from the partition index being passed in

rahul-kumar-saini: Well that's what the info command would help with. It would list first+last available offset, and then a list of partitions with whatever offset they contain up to.

And if a bad partition index is passed in, the error message would be displayed.

lynnagara: If you pass a partition and offset in though, you can get messages from another partition and offset entirely - is that expected? You can also get different results each time because there is no ordering between partitions.

snuba/cli/dlq_manager.py
lynnagara: If we look at this line...
```
STORAGE_SETS_WITH_DLQ = {
    storage.get_storage_set_key().value: topic
    for storage, topic in __STORAGE_TO_DLQ_TOPIC.items()
    if topic is not None
}
```
...we see that DLQ topic is defined at the storage level, and if there are 2 storages with different DLQ topics, then the first one will be clobbered by the second.

I think it's also possible multiple storage sets will end up sharing the same DLQ, in which case it might be equally weird to pick a storage set and see items from multiple storage sets in there.

If --storage or --storage-set is just a roundabout way of specifying a topic anyway, why don't you just pass the topic via the CLI instead?

lynnagara (changes requested):
I have a few outstanding concerns/questions
- The ticket you linked talks about building a standalone tool that can be used by both Snuba/Sentry not something in Snuba. Is there a reason we decided to build this in Snuba instead?
- We define DLQ by storage in Snuba, but the CLI takes a storage-set. It makes an assumption that all storages have the same topic and will not work properly if not.
- If we want this tool to work more generically (e.g. for Sentry), wouldn't it be better to just pass the topic instead of the storage or storage set and then try to use those to map back to the topic. The storage/storage set are very Snuba specific concepts.
- We pass a partition and offset to the tool but then it might consume from all other partitions as well if there are more than 1. This seems kind of unexpected. Wouldn't it be more expected to get results from only the partition being passed in starting at the provided offset?

snuba/cli/dlq_manager.py
```diff
@@ -0,0 +1,194 @@
...
+    type=click.Choice([storage_set for storage_set in STORAGE_SETS_WITH_DLQ]),
+    help="The storage set to list dead letters for",
+    required=True,
+)
+@click.option(
+    "--offset",
+    "offset",
+    type=int,
+    help="An offset on the dead-letter topic to start reading from, earliest by default",
+    default=0,
```
lynnagara, line 48: The description says the earliest is the default but then it's set to 0 specifically, which may not be the earliest one. What happens if 0 is out of range?

rahul-kumar-saini:
Closing as this is not necessary just yet, can be revisited later if needed.

PR #2616, opened 2022-04-15
feat(subscriptions): Temporarily run errors in global mode

lynnagara:
Temporarily run in global mode to verify the results are the same.
We will switch back later after the test.


volokluev:
whats global mode

lynnagara:
> whats global mode

The scheduler will wait for all partitions to reach a timestamp before scheduling rather than immediately schedule. It's for datasets not partitioned by project ID.

PR #2624, opened 2022-04-18
fix(subscriptions): Fix InvalidRangeError (again)

lynnagara:
https://github.com/getsentry/snuba/pull/2611 wasn't quite right
since the partitions we are keeping track of in self.__previous_messages
actually refer to the partitions on the original topic not the commit log
topic which is the one we are consuming from. We need to completely clear
self.__previous messages on revocation to avoid building invalid intervals
or skipping ticks during rebalance.


onewland (approved):
This looks fine. Is there a reason `__time_shift` shouldn't also be cleared?

lynnagara:
> This looks fine. Is there a reason `__time_shift` shouldn't also be cleared?

time_shift is a constant passed to the scheduler. It doesn't change with the message and partition data like the previous_messages does.

PR #2626, opened 2022-04-18
ref(metrics/sessions): enforce retention days

MeredithAnya:
**context**:
Sessions and metrics don't have retention enforced anywhere in snuba and rely on the the payload of the session/metric event. This PR makes sure that sessions and metrics events have a retention of either `30` or `90` days.

**changes**
I've spilt out the override functionality of the existing `enforce_retention` that is used for events/transactions. Overriding retention days is something that can be set on a project level basis for events/transactions.

Sessions/metrics may want to have override functionality later on but currently we don't and we'd want those overrides to be independent from other datasets (events/transactions).

### DIFF
--- a/snuba/datasets/events_format.py
+++ b/snuba/datasets/events_format.py
@@ -101,22 +101,21 @@ def extract_extra_contexts(
 
 
 def override_and_enforce_retention(
-    message: Mapping[str, Any], timestamp: Optional[datetime]
+    project_id: int, retention_days: Optional[int], timestamp: Optional[datetime]
 ) -> int:
     """
     Overriding retention with RETENTION_OVERRIDES is only
     used for events/transactions datasets.
     """
-    project_id = message["project_id"]
-    retention_days = settings.RETENTION_OVERRIDES.get(project_id)
-    if retention_days is None:
-        retention_days = int(
-            message.get("retention_days") or settings.DEFAULT_RETENTION_DAYS
-        )
+    if project_id:
+        retention_days = settings.RETENTION_OVERRIDES.get(project_id)
+
     return enforce_retention(retention_days, timestamp)
 
 
-def enforce_retention(retention_days: int, timestamp: Optional[datetime]) -> int:
+def enforce_retention(
+    retention_days: Optional[int], timestamp: Optional[datetime]
+) -> int:
     if retention_days is None:
         retention_days = settings.DEFAULT_RETENTION_DAYS
 
--- a/snuba/datasets/events_format.py
+++ b/snuba/datasets/events_format.py
@@ -100,13 +100,25 @@ def extract_extra_contexts(
     return (context_keys, context_values)
 
 
-def enforce_retention(message: Mapping[str, Any], timestamp: Optional[datetime]) -> int:
+def override_and_enforce_retention(
+    message: Mapping[str, Any], timestamp: Optional[datetime]
+) -> int:
+    """
+    Overriding retention with RETENTION_OVERRIDES is only
+    used for events/transactions datasets.
+    """
     project_id = message["project_id"]
     retention_days = settings.RETENTION_OVERRIDES.get(project_id)
     if retention_days is None:
         retention_days = int(
             message.get("retention_days") or settings.DEFAULT_RETENTION_DAYS
         )
+    return enforce_retention(retention_days, timestamp)
+
+
+def enforce_retention(retention_days: int, timestamp: Optional[datetime]) -> int:
+    if retention_days is None:
+        retention_days = settings.DEFAULT_RETENTION_DAYS
 
     if settings.ENFORCE_RETENTION:
         retention_days = (
--- a/snuba/datasets/events_format.py
+++ b/snuba/datasets/events_format.py
@@ -108,7 +108,7 @@ def override_and_enforce_retention(
     used for events/transactions datasets.
     """
     if project_id:
-        retention_days = settings.RETENTION_OVERRIDES.get(project_id)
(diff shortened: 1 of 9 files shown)


snuba/datasets/sessions_processor.py
```diff
@@ -27,6 +27,18 @@
...
+def enforce_retention_days(message: Mapping[str, Any]) -> int:
+    retention_days: int = message["retention_days"]
+
+    if settings.ENFORCE_RETENTION:
+        retention_days = (
+            settings.LOWER_RETENTION_DAYS
+            if retention_days <= settings.LOWER_RETENTION_DAYS
+            else settings.DEFAULT_RETENTION_DAYS
+        )
+    return retention_days
```
nikhars, line 39: Since this has become common code for most datasets, might be worth putting in a place where common code lives. Maybe the utils directory? You can just pass in the retention day number to the util function so that different datasets could have different keys and it would still work.

MeredithAnya: yeah I was actually going to just use the `enforce_retention` that is used in the events_formats

tests/datasets/test_metrics_processor.py
```diff
@@ -26,26 +26,32 @@
 
 MATERIALIZATION_VERSION = 4
 
+timestamp = int(datetime.now(timezone.utc).timestamp())
+# I don't know why this is off by 7 hours
+expected_timestamp = datetime.fromtimestamp(timestamp + 25200)
```
MeredithAnya, line 31: @onewland do you know why this is off by 7 hours?

onewland: That almost certainly seems like the difference between UTC/PT, but I don't know

evanh (approved):
This looks good to me.

tests/datasets/test_metrics_processor.py
evanh: Did something fail if you removed it?

snuba/datasets/metrics_aggregate_processor.py
```diff
@@ -94,7 +95,7 @@ def process_message(
                 "tags.key": _array_literal(keys),
                 "tags.value": _array_literal(values),
                 **self._process_values(message),
-                "retention_days": _literal(message["retention_days"]),
+                "retention_days": _literal(enforce_retention(message, timestamp)),
```
nikhars, line 98: Using the `enforce_retention` API from events_format, we introduce the following two constraints in sessions and metrics.
1. If a project has lower retention defined for either events or transactions, the lower retention would be applicable for sessions and metrics.
2. If the timestamp on the message is older than the retention days, we would crash the consumer since the `EventTooOld` exception would be raised.

I am not too familiar with the product to know if we want to add these additional constraints or not.

MeredithAnya: @nikhars those are good points. 

> 1. If a project has lower retention defined for either events or transactions, the lower retention would be applicable for sessions and metrics.


So can a project have its own `LOWER_RETENTION_DAYS` set? or do you mean the `RETENTION_OVERRIDES`? I think if we have `ENFORCE_RETENTION` set to true, regardless of project, the retention days will be either `30` or `90` right? Or are you specifically pointing out that we may want to have different retention days defined for different datasets? Like have events/transactions lower retention be 30 while metrics/sessions be 90 or something like that

> 2. If the timestamp on the message is older than the retention days, we would crash the consumer since the EventTooOld exception would be raised.

Yeah, it seems like from the metrics/sessions queries in sentry we hardcode in a time delta of 90days, so I was assuming the product isn't querying for that data anyway so we shouldn't be processing it. However, I see in the events_processor_base we handle that exception, so I should probs add handling it in the metrics and sessions processors as well

nikhars: > So can a project have its own LOWER_RETENTION_DAYS set? or do you mean the RETENTION_OVERRIDES? I think if we have ENFORCE_RETENTION set to true, regardless of project, the retention days will be either 30 or 90 right? Or are you specifically pointing out that we may want to have different retention days defined for different datasets? Like have events/transactions lower retention be 30 while metrics/sessions are 90 or something like that

Each project can set a lower retention period by using RETENTION_OVERRIDES. The RETENTION_OVERRIDES is currently set at 30 for certain projects. RETENTION_OVERRIDES were only being used for errors/transactions. So the question is if a project has lower retention for errors/transactions via RETENTION_OVERRIDES, should it also have lower retention on sessions/metrics?

fpacifici: > So the question is if a project has lower retention for errors/transactions via RETENTION_OVERRIDES, should it also have lower retention on sessions/metrics?

No. Retention have to be independent.

snuba/datasets/events_format.py
```diff
@@ -100,13 +100,25 @@ def extract_extra_contexts(
     return (context_keys, context_values)
 
 
-def enforce_retention(message: Mapping[str, Any], timestamp: Optional[datetime]) -> int:
+def override_and_enforce_retention(
+    message: Mapping[str, Any], timestamp: Optional[datetime]
```
nikhars, line 104: Nit: I would prefer if you passed in the parameter `project_id` and `retention_days` over passing the full message. Those fields could be called something else (like `projectid`) in some other dataset in the future. And let's assume we want to override and enforce this in the new dataset. Passing the message would fail in those cases.

nikhars (approved):
Left a minor comment. Else looks good

fpacifici (approved):

PR #2627, opened 2022-04-18
chore(tiger): Log exceptions from tiger with warn level

nikhars:
In order to avoid impacting our SLO's while still trying to run the
production load we would like to change the exceptions from the tiger
cluster to warn level.

### DIFF
--- a/snuba/utils/threaded_function_delegator.py
+++ b/snuba/utils/threaded_function_delegator.py
@@ -39,10 +39,12 @@ class ThreadedFunctionDelegator(Generic[TInput, TResult]):
         callback_func: Optional[
             Callable[[Optional[Result[TResult]], List[Result[TResult]]], None]
         ],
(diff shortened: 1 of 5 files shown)


snuba/utils/threaded_function_delegator.py
```diff
@@ -84,6 +86,9 @@ def execute_callback() -> None:
                     if self.__callback_func is not None:
                         self.__callback_func(primary_result, other_results)
                 except Exception as error:
-                    logger.exception(error)
+                    if self.__ignore_secondary_exceptions:
```
volokluev, line 89: how does this only ignore exceptions from tiger?

volokluev: oh nvm, I understand now

volokluev (approved):

PR #2631, opened 2022-04-19
feat(devserver): Support running the new subscriptions in dev

lynnagara:
Simplifies running the new subscriptions pipeline instead of
the old one in dev. This is not default so will not affect
most dev users.


fpacifici (approved):
I have an idea for dev and self hosted for cases where the executor is single node. Which is most scenarios outside of production. 
Why not building a combined scheduler + executor we use in single nodes setup. So we would run one container instead of two. What do you think ?

lynnagara:
> I have an idea for dev and self hosted for cases where the executor is single node. Which is most scenarios outside of production. Why not building a combined scheduler + executor we use in single nodes setup. So we would run one container instead of two. What do you think ?

We can consider this as a follow up, though right now it's useful for dev to be more similar to prod for debugging purposes.

PR #2632, opened 2022-04-19
fix: Use `python3` for `.envrc` virtualenv creation

asottile-sentry:
on the latest macos, `python` has been removed and is not
always a symlink to `python3`.

before this patch:

```console
$ direnv allow
direnv: loading ~/workspace/snuba/.envrc
./.envrc:5: python: command not found
./.envrc:6: .venv/bin/activate: No such file or directory
./.envrc:7: pip: command not found
pip install -e .
make: pip: No such file or directory
make: *** [install-python-dependencies] Error 1
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -2,7 +2,7 @@
 # shellcheck disable=SC1091
 
 if [ ! -d .venv ]; then
-    python -m venv .venv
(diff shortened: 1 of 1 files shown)


evanh (approved):
Thanks!

PR #2637, opened 2022-04-19
feat(metrics): Add entity and storage to allow for cross-org metric counter queries

wedamija:
To support converting the release monitor in sentry over to metrics we want to be able to make
queries without having to specify org and project ids in the query. This pr adds in
`OrgMetricsCountersEntity` and `org_counters_storage` to allow us to do this.

### DIFF
--- a/snuba/datasets/entities/metrics.py
+++ b/snuba/datasets/entities/metrics.py
@@ -1,6 +1,6 @@
 from abc import ABC
 from dataclasses import dataclass
-from typing import Optional, Sequence, Tuple, Union
+from typing import List, Optional, Sequence, Tuple, Union
 
 from snuba.clickhouse.columns import (
     AggregateFunction,
@@ -50,6 +50,7 @@ from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
 from snuba.query.validation.validators import (
     EntityRequiredColumnValidator,
     GranularityValidator,
+    QueryValidator,
 )
 from snuba.request.request_settings import RequestSettings
 
@@ -91,7 +92,7 @@ class MetricsEntity(Entity, ABC):
         if writable_storage:
             storages.append(writable_storage)
 
-        validators = []
+        validators: List[QueryValidator] = []
         if not allow_cross_org:
             validators.append(EntityRequiredColumnValidator({"org_id", "project_id"}))
         validators.append(GranularityValidator(minimum=10))
--- a/snuba/datasets/entities/metrics.py
+++ b/snuba/datasets/entities/metrics.py
@@ -82,6 +82,7 @@ class MetricsEntity(Entity, ABC):
         readable_storage_key: StorageKey,
         value_schema: Sequence[Column[SchemaModifiers]],
         mappers: TranslationMappers,
+        abstract_column_set: ColumnSet = None,
         validators: Optional[Sequence[QueryValidator]] = None,
     ) -> None:
         writable_storage = (
@@ -91,6 +92,18 @@ class MetricsEntity(Entity, ABC):
         storages = [readable_storage]
         if writable_storage:
             storages.append(writable_storage)
+
+        if abstract_column_set is None:
+            abstract_column_set = ColumnSet(
+                [
+                    Column("org_id", UInt(64)),
+                    Column("project_id", UInt(64)),
+                    Column("metric_id", UInt(64)),
+                    Column("timestamp", DateTime()),
+                    Column("tags", Nested([("key", UInt(64)), ("value", UInt(64))])),
+                    *value_schema,
+                ]
+            )
 
         if validators is None:
             validators = [
@@ -110,16 +123,7 @@ class MetricsEntity(Entity, ABC):
                     ).concat(mappers),
                 )
             ),
-            abstract_column_set=ColumnSet(
-                [
-                    Column("org_id", UInt(64)),
-                    Column("project_id", UInt(64)),
-                    Column("metric_id", UInt(64)),
-                    Column("timestamp", DateTime()),
-                    Column("tags", Nested([("key", UInt(64)), ("value", UInt(64))])),
-                    *value_schema,
-                ]
-            ),
+            abstract_column_set=abstract_column_set,
             join_relationships={},
             writable_storage=writable_storage,
             validators=validators,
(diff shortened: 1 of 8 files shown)


snuba/datasets/entities/metrics.py
```diff
@@ -77,16 +77,27 @@ def transform_expression(exp: Expression) -> Expression:
 class MetricsEntity(Entity, ABC):
     def __init__(
         self,
-        writable_storage_key: StorageKey,
+        writable_storage_key: Optional[StorageKey],
         readable_storage_key: StorageKey,
         value_schema: Sequence[Column[SchemaModifiers]],
         mappers: TranslationMappers,
+        allow_cross_org: bool = False,
```
wedamija, line 84: Not sure if we prefer to handle this via a param or possibly split this class up into a more generic base class.

Also could probably do with a better name here, but I couldn't think of one.

snuba/datasets/entities/metrics.py
```diff
@@ -77,16 +77,27 @@ def transform_expression(exp: Expression) -> Expression:
...
+        )
         readable_storage = get_storage(readable_storage_key)
+        storages = [readable_storage]
+        if writable_storage:
+            storages.append(writable_storage)
+
+        validators = []
+        if not allow_cross_org:
+            validators.append(EntityRequiredColumnValidator({"org_id", "project_id"}))
+        validators.append(GranularityValidator(minimum=10))
```
wedamija, line 97: Not sure if order matters here, otherwise I can just include `GranularityValidator` in the initial list

snuba/datasets/entities/metrics.py
```diff
@@ -161,6 +169,17 @@ def __init__(self) -> None:
         )
 
 
+class OrgMetricsCountersEntity(MetricsEntity):
+    def __init__(self) -> None:
+        super().__init__(
+            writable_storage_key=None,
+            readable_storage_key=StorageKey.METRICS_COUNTERS,
+            value_schema=[],
```
wedamija, line 177: I set `value_schema` and `mappers` to be empty since we don't need to do complex sums or aggregations here. This seems to work fine in the tests.

snuba/datasets/storages/metrics.py
```diff
@@ -143,6 +143,23 @@ def ignore_policy_closure() -> DeadLetterQueuePolicy:
...
+    storage_key=StorageKey.METRICS_COUNTERS,
+    storage_set_key=StorageSetKey.METRICS,
+    schema=TableSchema(
+        local_table_name="metrics_counters_v2_local",
+        dist_table_name="metrics_counters_v2_dist",
+        storage_set_key=StorageSetKey.METRICS,
+        columns=ColumnSet(
+            [
+                *aggregated_columns,
+                Column("value", AggregateFunction("sum", [Float(64)])),
```
wedamija, line 156: Can I drop this?

snuba/datasets/storages/metrics.py
```diff
@@ -143,6 +143,23 @@ def ignore_policy_closure() -> DeadLetterQueuePolicy:
...
+            ]
+        ),
+    ),
+    query_processors=[ArrayJoinKeyValueOptimizer("tags"), TableRateLimit()],
```
wedamija, line 160: Not sure if I need the `ArrayJoinKeyValueOptimizer` here, I assume it doesn't hurt.

snuba/datasets/entities/metrics.py
onewland: I wonder if it makes sense to just elevate `validators` to a param of this method given that we're already calling `super.__init__()` everywhere -- it could have a default of the current values. It looks like it's a parameter on the base `Entity` class so the pattern already exists

snuba/datasets/entities/metrics.py
onewland: I think I would keep those at parity. It's just nice to be able to write `sum(value)` in a SnQL query vs `sumMerge(value)` though both will work

snuba/datasets/storages/metrics.py
onewland: I would see what happens in tests if you do drop it -- I can't recall if we use the schema columns to do validation or create the return object

snuba/datasets/storages/metrics.py
onewland: I think it's probably safer to leave it, in case somebody wants to use tags later

onewland (commented):
Mostly looks good to me. I do think it could be cleaner if we have validators be a more generic parameter, though.

snuba/datasets/entities/metrics.py
wedamija: Agreed, this is much cleaner

snuba/datasets/storages/metrics.py
wedamija: Tests seem to run fine. I'm not sure whether I should or not, but was mostly trying to limit how much people can do with this entity/storage, since we don't want people making overly complex queries without being restricted to project/org

onewland: Sounds good, let's only grant access as we need it.

Following up on that, the mappers don't do anything for you if you don't want to surface the sum column.

snuba/datasets/entities/metrics.py
```diff
@@ -161,6 +171,22 @@ def __init__(self) -> None:
...
+            writable_storage_key=None,
+            readable_storage_key=StorageKey.METRICS_COUNTERS,
+            value_schema=[Column("value", AggregateFunction("sum", [Float(64)]))],
+            mappers=TranslationMappers(
+                functions=[
+                    FunctionNameMapper("sum", "sumMerge"),
+                    FunctionNameMapper("sumIf", "sumMergeIf"),
+                ],
+            ),
+            validators=[],
```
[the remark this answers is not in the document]
wedamija: Missed this when I refactored, thanks!

onewland (approved):

snuba/datasets/storages/metrics.py
```diff
@@ -143,6 +143,18 @@ def ignore_policy_closure() -> DeadLetterQueuePolicy:
     write_format=WriteFormat.VALUES,
 )
 
+org_counters_storage = ReadableTableStorage(
+    storage_key=StorageKey.METRICS_COUNTERS,
```
fpacifici, line 147: You cannot reuse the same storage key for two different storages. Please create a new one.

snuba/datasets/entities/metrics.py
```diff
@@ -161,6 +171,17 @@ def __init__(self) -> None:
         )
 
 
+class OrgMetricsCountersEntity(MetricsEntity):
+    def __init__(self) -> None:
+        super().__init__(
+            writable_storage_key=None,
+            readable_storage_key=StorageKey.METRICS_COUNTERS,
+            value_schema=[Column("value", AggregateFunction("sum", [Float(64)]))],
```
fpacifici, line 179: I don't think this is meant to expose all the fields.
See how this was done on sessions
https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/sessions.py#L305-L312

That only included org, project and time exactly to ensure this entity would not be used for anything other than returning a list of orgs and projects as it does not have the safeguards in place to prevent a query across organizations done by mistake.

I think we should do the same and only expose the same fields: org, project, timestamp (and granularity otherwise you cannot run the query).

We can consider tags if needed, not sure how the query is meant to look like, but please do not expose the aggregate values otherwise this could be used for scenarios other than your specific one.

fpacifici (changes requested):

fpacifici (approved):

PR #2641, opened 2022-04-20
Fix parse errors on tiger

volokluev:
### Context
Due to a backwards incomaptibility issue in clickhouse from 20.8 -> 21.8, aliased tuples within functions sometimes cause parsing errors in certain conditions. Issue: https://github.com/ClickHouse/ClickHouse/issues/35625

Introduce a query processor removes the aliases of`tuple` function calls whenever they are nested within other expressions. This change depends on alias expansion earlier in the pipeline working. I verified to see that it does in fact work as expected and the test coverage is adequate, but there is some risk being taken on here

### Notes:

This will likely not solve all the issues related to queries on tiger but I would like to roll this out to see if this fixes the majority of them. I think it should. Here are queries that I tried on a local 21.8 distributed clickhouse to reproduce the issue. Table name does not matter.

Query | Fails? | Comments
-- | -- | --
SELECT equals(tupleElement(tuple('a', 10) AS x, 1), 'a') FROM all_hits_2 LIMIT 1 | ✅ | Nikhar’s original
SELECT equals(tupleElement(tuple('a', 10), 1), 'a') FROM all_hits_2 LIMIT 1 | ❌ | No aliasing of the tuple
SELECT equals(tuple('a', 10) AS x, tuple('a', 1) AS y) FROM all_hits_2 LIMIT 1 | ❌ | tuple aliased inside the equals function but not within tupleElement
SELECT untuple(tuple(tupleElement(tuple('a', 10, 1) AS xyz, 1), 'poo') AS y) AS x FROM all_hits_2 LIMIT 1 | ❌ | aliasing the tuple element inside the tupleElement function seems to be fine as well. Maybe it’s just equals?
SELECT equals(tuple(tupleElement(tuple('a', 10, 1) AS xyz, 1), 'poo'), tuple('a', 'b')) AS x FROM all_hits_2 LIMIT 1 | ✅ | equals is probably the culprit
SELECT equals(tuple(tupleElement(tuple('a', 10, 1), 1), 'poo') AS xyz, tuple('a', 'b')) AS x FROM all_hits_2 LIMIT 1 | ❌ | aliasing the tuple outside of tupleElement within equals works

I was not able to get a full end to end test from snql -> bad clickhouse sql. It was taking me way too long and I ended up in the guts of the query pipeline again. A a pragmatic tradeoff, I would like to deploy this processor and see how it behaves in production

### Rollout
This QueryProcessor can be turned on/off in the runtime config

To turn on:
`SET tuple_unaliaser_rollout 1.0`

To turn on for ten percent of queries:
`SET tuple_unaliaser_rollout 0.1`

To turn off:
`SET tuple_unaliaser_rollout 0`
or
`DELETE tuple_unaliaser_rollout`

### DIFF
--- a/snuba/datasets/storages/errors_v2.py
+++ b/snuba/datasets/storages/errors_v2.py
@@ -34,6 +34,7 @@ from snuba.query.processors.mapping_promoter import MappingColumnPromoter
 from snuba.query.processors.prewhere import PrewhereProcessor
 from snuba.query.processors.slice_of_map_optimizer import SliceOfMapOptimizer
 from snuba.query.processors.table_rate_limit import TableRateLimit
+from snuba.query.processors.tuple_unaliaser import TupleUnaliaser
 from snuba.query.processors.type_converters.hexint_column_processor import (
     HexIntColumnProcessor,
 )
@@ -55,6 +56,7 @@ from snuba.utils.streams.topics import Topic
 # the new storage.
 query_processors = [
     UniqInSelectAndHavingProcessor(),
+    TupleUnaliaser(),
     PostReplacementConsistencyEnforcer(
         project_column="project_id", replacer_state_name=ReplacerState.ERRORS_V2,
     ),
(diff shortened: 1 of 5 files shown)


nikhars (approved):
LGTM

PR #2643, opened 2022-04-22
feat: Upgrade black and click

lynnagara:
The main motivation for this upgrade is to match the black version to the
one used in pre-commit. Previously they were different causing editor autoformatting
(e.g. the formatting applied on save in VSCode) to be different to autoformatting
applied in pre-commit.

----

### DIFF
--- a/snuba/cli/config.py
+++ b/snuba/cli/config.py
@@ -33,7 +33,7 @@ def config() -> None:
 
 
 @config.command("get-all")
-@click.option("--format", type=click.Choice(FORMATS.keys()), default="human")
+@click.option("--format", type=click.Choice([*FORMATS.keys()]), default="human")
 def get_all(*, format: str) -> None:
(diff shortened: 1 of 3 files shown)


snuba/cli/config.py
```diff
@@ -33,7 +33,7 @@ def config() -> None:
 
 
 @config.command("get-all")
-@click.option("--format", type=click.Choice(FORMATS.keys()), default="human")
+@click.option("--format", type=click.Choice([*FORMATS.keys()]), default="human")
```
nikhars, line 36: Is this something that is required by the new click version?

I see that click.Choice is used in other places which are not updated in this PR

lynnagara: Yeah it is. Choice now expects a Sequence rather than an Iterable.

nikhars (approved):

PR #2644, opened 2022-04-22
fix(subscriptions): Properly close producer

lynnagara:
This is an attempt to fix an error we are seeing in the executor logs
about the producer terminating with unflushed messages still in queue.

Now we wait for the producer.close() future to finish when closing the producer.

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -9,7 +9,7 @@ from typing import Callable, Deque, Mapping, MutableMapping, Optional, Sequence,
 import rapidjson
 from arroyo import Message, Partition, Topic
 from arroyo.backends.abstract import Producer
-from arroyo.backends.kafka import KafkaConsumer, KafkaPayload, KafkaProducer
+from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
 from arroyo.processing import StreamProcessor
 from arroyo.processing.strategies import MessageRejected, ProcessingStrategy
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -438,3 +438,11 @@ def join(self, timeout: Optional[float] = None) -> None:
...
                 {message.partition: Position(message.offset, message.timestamp)}
             )
+
+        # Flush producer
+        if isinstance(self.__producer, KafkaProducer):
+            remaining = timeout - (time.time() - start) if timeout is not None else None
+            # TODO: This is just for testing. If it works expose a flush method on KafkaProducer
+            producer = self.__producer._KafkaProducer__producer  # type: ignore
+            messages = producer.flush(*[remaining] if remaining is not None else [])
+            logger.debug(f"{messages} executor messages pending delivery")
```
fpacifici, line 448: I don't think you need to do this.
https://github.com/getsentry/arroyo/blob/main/arroyo/backends/kafka/consumer.py#L738-L740

The arroyo Producer has a close method that prevents it from accepting new messages and returns you a future that flushes the producer. Maybe we should ensure we log in arroyo whether the producer was flushed properly or not.

fpacifici (changes requested):
I think you are missing the call to `close` on the producer.

snuba/subscriptions/executor_consumer.py
lynnagara: Hmm, good point. Updated to wait for the future on close instead.

fpacifici (approved):

PR #2645, opened 2022-04-22
feat: Support separate transactions topic in Snuba

lynnagara:
Add the `transactions` and `snuba-transactions-commit-log` topics
to Snuba in order to support moving these to a separate topic entirely.
Currently these should still be mapped to the `events` and `snuba-commit-log`
topics in Snuba settings in every environment as they still share a topic
(this needs to be done before this PR is merged).
Once we are ready to start producing to the new topic, we can remove the
mapping in settings.


lynnagara:
I'm actually not sure of the status of the spans storage and whether that should be moved over as well.

evanh (approved):
When will the topics actually be switched?

lynnagara:
> When will the topics actually be switched?

As soon as possible. We will need the code change in the Snuba eventstream to switch between two topics ("events" and "transactions") before we can start rolling it out though

nikhars (approved):

PR #2647, opened 2022-04-25
fix(tests): Flakey DLQ policy test

rahul-kumar-saini:
### Overview
- Lyn encountered a `DID NOT RAISE` failure on a DLQ policy test
- This test ensures the policy ignores up to `n` invalid messages per second before raising an exception on the `n+1`st invalid message
- The flakeyness could be from the time bucket changing to the next second while the original `n` messages are being submitted, making the expected exception not raise since it's still under the limit
- Reduced `n` from 5 to 1 to reduce the window of possible time bucket change


lynnagara (approved):
Thank you!

volokluev (approved):

PR #2651, opened 2022-04-26
ref(tiger): Allow increasing cache wait time for tiger

nikhars:
We are noticing an increase in ExecutionTimeoutError on the tiger
cluster. In order to debug what's happening, I would like the ability
to increase the cache wait timeout period for tiger clusters. The max
query execution time for a clickhouse query would be still 30 seconds.
But I would increase the wait timeout to see if the result gets
populated with some additional wait time.

----

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -466,6 +466,18 @@ def execute_query_with_readthrough_caching(
         "cache_partition_loaded",
         tags={"partition_id": reader.cache_partition_id or "default"},
     )
+
+    cache_wait_timeout: int = int(query_settings.get("max_execution_time", 30))
+    # Increase the cache wait timeout for tiger clusters to debug
+    # ExecutionTimeoutErrors
+    if reader.cache_partition_id and reader.cache_partition_id in {
+        "tiger_errors",
+        "tiger_transactions",
+    }:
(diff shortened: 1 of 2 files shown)


snuba/web/db_query.py
```diff
@@ -466,6 +466,18 @@ def record_cache_hit_type(hit_type: int) -> None:
...
+    cache_wait_timeout: int = int(query_settings.get("max_execution_time", 30))
+    # Increase the cache wait timeout for tiger clusters to debug
+    # ExecutionTimeoutErrors
+    if reader.cache_partition_id and reader.cache_partition_id in {
+        "tiger_errors",
+        "tiger_transactions",
+    }:
+        tiger_wait_timeout_config = state.get_config("tiger-cache-wait-time")
+        if tiger_wait_timeout_config:
+            cache_wait_timeout = tiger_wait_timeout_config
```
volokluev, line 479: I'd recommend putting this into a helper function

volokluev (approved):

snuba/web/db_query.py
volokluev: Also something I like to do for code like this is put a note saying "if this code still exists by 2022-05-07, it should be removed, it was a hack to debug something"

PR #2652, opened 2022-04-26
fix(querylog): Flush querylog producer before terminating processes

lynnagara:
Another attempt to fix that producer termination with unflushed events error.
During the earlier test we discovered there were no unflushed events on the
subscription results producer so that one is unlikely to be the actual
culprit here. This time we are calling flush on the querylog producer.


snuba/state/__init__.py
```diff
@@ -325,6 +325,12 @@ def record_query(query_metadata: Mapping[str, Any]) -> None:
         logger.exception("Could not record query due to error: %r", ex)
 
 
+def flush_producer() -> None:
+    global kfk
+    if kfk is not None:
+        kfk.flush()
```
MeredithAnya, line 331: do we not want to give this a timeout?

lynnagara: What timeout do you suggest? In arroyo producer we don't so I think it's fine https://github.com/getsentry/arroyo/blob/4854c67b5ad19c304e3301eff98c45ccaed93a5e/arroyo/backends/kafka/consumer.py#L681

evanh (approved):
Looks good to me, pending Meredith's question.

lynnagara:
> Looks good to me, pending Meredith's question.

Going to merge this now. Confirmed this offline with Meredith.

PR #2654, opened 2022-04-26
fix(tracing): don't set the readonly mode on the client side

volokluev:
Currently the snuba-admin tool is broken for new hosts with the following error:

`An error occurred: DB::Exception: Cannot modify 'readonly' setting in readonly mode.`

From clickhouse:

```
SELECT *
FROM system.settings
WHERE name = 'readonly'
FORMAT Vertical

Query id: 4751d842-e00a-402c-9439-214afd1610d6

Row 1:
──────
name:        readonly
value:       2
changed:     1
IMPORTANT ------> description: 0 - everything is allowed. 1 - only read requests. 2 - only read requests, as well as changing settings, except for the 'readonly' setting.
min:         ᴺᵁᴸᴸ
max:         ᴺᵁᴸᴸ
readonly:    1
type:        UInt64
```
> 2 - only read requests, as well as changing settings, except for the 'readonly' setting.

Putting this setting into the client settings makes clickhouse interpret the query as changing the readonly setting (despite the fact that nothing is being changed. There is actually no need for us to pass this in the connection as the read only user can only issue read only queries

### DIFF
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -45,7 +45,7 @@ class ClickhouseClientSettings(Enum):
     )
     OPTIMIZE = ClickhouseClientSettingsType({}, 10000)
     QUERY = ClickhouseClientSettingsType({"readonly": 1}, None)
-    TRACING = ClickhouseClientSettingsType({"readonly": 2}, None)
+    TRACING = ClickhouseClientSettingsType({}, None)
     REPLACE = ClickhouseClientSettingsType(
         {
             # Replacing existing rows requires reconstructing the entire tuple for each


onewland:
I'm not 100% clear from your description -- are we breaking something by trying to set this value in this code or is this just a confusing no-op and we want to save other people from going down the path?

volokluev:
we are breaking something trying to set this value in code

onewland (approved):
I'm fine with the change though we should double-check that it doesn't break tracing on hosts where that currently works after deploy (e.g. I can do a trace on errors/transactions right now).

lynnagara (dismissed):
I don't think we should drop this. It adds an additional guarantee in code that no writing is being done by the code that uses these settings. Note that the readonly user only exists in SaaS because we have configured it there and not in any other environments.

Dropping it is not going to make tracing ever work if readonly was set to 1 so it seems like it would just be trading one error message for another.

nikhars:
> I think Volo might have a point here, though. We use the same read-only user for both read-only system table queries and tracing queries:
> 
> https://github.com/getsentry/snuba/blob/eac5b209d9ff1bdff6904dde4142fa85ac4afd73/snuba/admin/clickhouse/system_queries.py#L20-L28
> 
> https://github.com/getsentry/snuba/blob/eac5b209d9ff1bdff6904dde4142fa85ac4afd73/snuba/admin/clickhouse/tracing.py#L30-L32
> 
> https://github.com/getsentry/snuba/blob/eac5b209d9ff1bdff6904dde4142fa85ac4afd73/snuba/admin/clickhouse/common.py#L94-L96
> 
> https://github.com/getsentry/snuba/blob/eac5b209d9ff1bdff6904dde4142fa85ac4afd73/snuba/admin/clickhouse/common.py#L61-L69
> 
> so if you pass `{readonly: 1}` under one condition and `{readonly: 2}` under the other, one will invariably be wrong and cause that condition to fail. I think that having the client declare intent is a good idea but 20.7 must have had looser validation than 21.8 because I don't see how the same readonly client could send that value accurately for both use cases.
> 
> I think we need to either:
> 
> 1. split up tracing and query into two separate users, or
> 2. accept settings changes on `readonly` and pass it as 2 in the request for `QUERY` clickhouse settings.
> 
> Annoyingly long internal Slack thread where I was trying to deploy changes related to this: https://sentry.slack.com/archives/C039ZGT5K/p1651010922903109/. I could either get system table queries to work, or query tracing, but not both on 21.8.

Adding a new user should be relatively straightforward and keep things simpler in Snuba code. I would vote to go with that.

onewland:
Digging into the ops repo, I don't actually think there's a difference in 20.7 vs. 21.8 behavior, but a difference in how we provision hosts in chef vs. salt.

It looks like query nodes use attributes from different chef cookbooks than their storage nodes, but for salt-provisioned hosts they all come from the same template. So for a chef host:

`readonly=2` for user `readonly` on a query node:
https://github.com/getsentry/ops/blob/master/roles/snuba-query.json#L31
`readonly=1` for user `readonly` on a storage node:
https://github.com/getsentry/ops/blob/master/roles/snuba-errors.json#L20

but a salt host doesn't have distinct user definitions for query vs storage nodes.

I think I prefer the consistency of the salt way of doing things but it means defining separate `readonly` and `trace_r[ead]o[nly]` users in salt and also adding those definitions to chef. Do people like this idea? @volokluev @nikhars @lynnagara  @fpacifici 

If so I'll start doing it in chef

PR #2656, opened 2022-04-26
feat(subscriptions): Support transition between the old and new pipelines

lynnagara:
Supports transitioning between the old and new pipelines without skipping
(or duplicating) subscription queries.


snuba/subscriptions/executor_consumer.py
```diff
@@ -188,6 +190,8 @@ def __init__(
         max_concurrent_queries: int,
         metrics: MetricsBackend,
         next_step: ProcessingStrategy[SubscriptionTaskResult],
+        # TODO: To be removed once executor is fully rolled out
```
nikhars, line 193: Could you provide some additional context around why the commit would be removed when the executor is rolled out

snuba/subscriptions/utils.py
```diff
@@ -29,3 +34,78 @@ def time_shift(self, delta: timedelta) -> Tick:
...
+    - legacy (the default)
+    - new
+    - both
+    - transition_legacy
+    - transition_new
+
+    If transition_legacy or transition_new are being used a transition timestamp
+    (Unix time) must be provided. Any subscriptions scheduled up to the transition
+    time will use the prior pipeline and any subscriptions scheduled after the
+    transition time will use the pipeline being transitioned to.
```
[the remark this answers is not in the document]
lynnagara: My plan is to do a one shot upgrade for both of those other environments and not ever run the two subscriptions processes side by side.

snuba/subscriptions/executor_consumer.py
lynnagara: I added an additional comment. This comment will be removed pretty soon along with the argument anyway.

nikhars (approved):

PR #2658, opened 2022-04-27
chore(uwsgi): Upgrade the uwsgi version used

nikhars:
### Context
One of the potential causes of results not being available from the query cache for tiger queries could be that the uwsgi processes keep dying too often. We see that in the logs of snuba API attached in the PR.

### Solution
In order to find out what is causing the uwsgi processed to abort, we need to find out the reason. The latest uwsgi adds a reason why the process is aborted.

----
Screen Shot 2022-04-27 at 11 36 39 AM


nikhars:
@getsentry/open-source - This PR is updating the version of uwsgi we use in Snuba. It does not add a new library. But uwsgi is licensed as GPL with linking exception. Details can be found here https://github.com/unbit/uwsgi/blob/master/LICENSE. What should be the next steps?

lynnagara (approved):

chadwhitacre:
@nikhars Cleared per licensing exception and rerunning FOSSA. Should be set. 👍

PR #2660, opened 2022-04-29
fix(test): use datetime.utcfromtimestamp

MeredithAnya:
This test worked before in CI because in CI the local time is UTC.


onewland (approved):
looks good assuming the tests pass on CI

PR #2662, opened 2022-05-02
ref(freight): Remove metrics subscriptions executor from freight

MeredithAnya:
We are seeing an `OffsetOutOfRange` error crashing the metrics subscription executor so removing it from freight for now so that it doesn't block other snuba deployments


onewland (approved):
Fine with me though @lynnagara should definitely be aware

PR #2663, opened 2022-05-02
feat(admin) - split trace and system query users for admin

onewland:
Due to issues with CH 21.8 and strictness around the readonly setting, we need to split up readonly and trace users in ClickHouse -- this enables us to take advantage of changes in https://getsentry.atlassian.net/browse/OPS-1704

----


volokluev (dismissed):

snuba/clusters/cluster.py
```diff
@@ -44,8 +45,12 @@ class ClickhouseClientSettings(Enum):
         {"load_balancing": "in_order", "replication_alter_partitions_sync": 2}, 10000
     )
     OPTIMIZE = ClickhouseClientSettingsType({}, 10000)
-    QUERY = ClickhouseClientSettingsType({"readonly": 1}, None)
-    TRACING = ClickhouseClientSettingsType({"readonly": 2}, None)
+    QUERY = ClickhouseClientSettingsType(
+        {"readonly": 1}, None, settings.CLICKHOUSE_READONLY_USER
+    )
```
lynnagara, line 50: We use the QUERY settings for all of our queries in the Snuba API and subscriptions. We cannot use the readonly user (which has much lower limits for that)

lynnagara (changes requested):
Requesting changes so this doesn't accidentally get merged. We can't use the readonly user for prod/api usage

snuba/clusters/cluster.py
onewland: got it, I'll add another setting

onewland:
@lynnagara I think that should be safe

lynnagara:
This seems a bit inconsistent with the way the user is selected here https://github.com/getsentry/snuba/blob/master/snuba/admin/clickhouse/common.py#L64-L65. Should we try to pick one approach or the other for selecting the readonly user rather than combining the two?

onewland:
> This seems a bit inconsistent with the way the user is selected here https://github.com/getsentry/snuba/blob/master/snuba/admin/clickhouse/common.py#L64-L65. Should we try to pick one approach or the other for selecting the readonly user rather than combining the two?

yeah, i'm actually realizing that the way I just did it assumes the same password too, which it shouldn't. I'll come back with a new revision later today

lynnagara (approved):

PR #2664, opened 2022-05-02
ref(dev-tooling): Enable commit tracking

MeredithAnya:
**context:**
I want to be able to add suspect commits to sentry for the snuba project. We already have releases being created (although tbh im not sure exactly _where_ that is happening. It looks like we already have the logic in place to sync the commits in freight:
https://github.com/getsentry/freight/blob/25d9be3f64b2dd6ee10e5a2a96293d8a7f1fc7fd/freight/providers/pipeline.py#L140-L157

I don't know if there was a reason this was commented out - it seems like it was added like that so wondering if we can give it a try.


MeredithAnya:
adding @getsentry/ops here because I wanted to check to verify the freight code that I linked to actually will work

evanh (approved):
Looks good to me.

PR #2665, opened 2022-05-02
fix: Retry simultaneous query errors for non-robust queries

evanh:
Retry simultaneous query errors for non-robust queries. This logic is very
similar to what happens with robust queries, but the errors will only be retried
once.

Context:
We sometimes see issues where a single storage node gets imbalanced incorrectly and receives a lot more traffic than it should. The reason for the imbalance is unclear, but we can retry those queries so they get run on a different node.

### DIFF
--- a/snuba/clickhouse/native.py
+++ b/snuba/clickhouse/native.py
@@ -240,6 +240,23 @@ class ClickhousePool(object):
                             # balancer a chance to mark a bad host as down.
                             time.sleep(0.1)
                 except errors.Error as e:
+                    if e.code == errors.ErrorCodes.TOO_MANY_SIMULTANEOUS_QUERIES:
+                        attempts_remaining -= 1
+                        if attempts_remaining <= 0:
+                            raise ClickhouseError(e.message, code=e.code) from e
+
+                        attempts_remaining = min(
+                            attempts_remaining, 1
+                        )  # only retry once
+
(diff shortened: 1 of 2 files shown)


onewland:
Should we maybe put this behind a run-time config? I don't have a sense of how dangerous it is but it seems like something we might want to be able to turn off.

Code looks fine

onewland:
Should we maybe put this behind a run-time config? I don't have a sense of how dangerous it is but it seems like something we might want to be able to turn off.

Code looks fine

snuba/clickhouse/native.py
```diff
@@ -240,6 +240,23 @@ def execute(
                             # balancer a chance to mark a bad host as down.
                             time.sleep(0.1)
                 except errors.Error as e:
+                    if e.code == errors.ErrorCodes.TOO_MANY_SIMULTANEOUS_QUERIES:
```
onewland, line 243: How about we fetch the `simultaneous_queries_sleep_seconds` from below and as part of the conditional here make sure that it's not None? That gives us a simple kill-switch if we're retrying and causing more overload

If it's None we allow 0 retries -- if it's any value we continue into this clause.

evanh: Sure.

onewland (approved):

PR #2666, opened 2022-05-02
feat(subscriptions): Combined scheduler and executor

volokluev:
An alternative to the separate scheduler and executor that can both schedule and execute subscriptions for multiple entities in the same dataset in one process. Designed for use cases where the number of subscriptions to be run is not high and we want to reduce the number of containers that are run (e.g. dev and self-hosted).


nikhars (commented):
Can you refresh my mind on a few things?
1. When the scheduler and executor are separate, do both the consumers commit to the topic they consume from? the scheduler will commit on the commit log topic and the executor will commit to the topic where queries are being read from?
2. With the combined scheduler and executor, the scheduler passes the queries directly to the executor, so the scheduler does not need to commit. But executor process will call the commit on the topic from which scheduler is consuming?

lynnagara:
> Can you refresh my mind on a few things?
> 
> 1. When the scheduler and executor are separate, do both the consumers commit to the topic they consume from? the scheduler will commit on the commit log topic and the executor will commit to the topic where queries are being read from?
> 2. With the combined scheduler and executor, the scheduler passes the queries directly to the executor, so the scheduler does not need to commit. But executor process will call the commit on the topic from which scheduler is consuming?

Both points above are right. To confirm-
- Separate scheduler and executor consume from and commit offsets to the commit log topic and scheduled topic respectively.
- The main difference is the combined one completely skips the scheduled topic (no producing there, consuming from there or committing offsets there). So the commit step that happens at the very end after execution is actually committing to the commit log topic. (It is the same code that in the main executor is actually committing to the scheduled topic - only the topic we are committing offsets to has changed here)

nikhars (approved):

PR #2668, opened 2022-05-03
fix(docs): Update and clarify dev setup

mcannizz:
Fix two issues I got stuck on while setting up my environment:
* `make develop` failed because clang couldn't find libraries installed
with Homebrew.
* `make test` failed because Kafka wasn't running but the instructions
didn't describe how to run it.

### DIFF
--- a/docs/source/getstarted.rst
+++ b/docs/source/getstarted.rst
@@ -13,8 +13,14 @@ Snuba assumes:
 1. A Clickhouse server endpoint at ``CLICKHOUSE_HOST`` (default ``localhost``).
 2. A redis instance running at ``REDIS_HOST`` (default ``localhost``). On port
    `6379`
+3. A Kafka cluster running at ``localhost`` on port `9092`.
 
-A quick way to get these services running is to set up sentry, then use::
+A quick way to get these services running is to set up sentry, and add the following line
+in ``~/.sentry/sentry.conf.py``::
+
+    SENTRY_EVENTSTREAM = "sentry.eventstream.kafka.KafkaEventStream"
+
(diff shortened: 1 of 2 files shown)


lynnagara (approved):
Great, thanks!

evanh (approved):
Nice!

docs/source/getstarted.rst
```diff
@@ -13,8 +13,14 @@ Snuba assumes:
 1. A Clickhouse server endpoint at ``CLICKHOUSE_HOST`` (default ``localhost``).
 2. A redis instance running at ``REDIS_HOST`` (default ``localhost``). On port
    `6379`
+3. A Kafka cluster running at ``localhost`` on port `9092`.
 
-A quick way to get these services running is to set up sentry, then use::
+A quick way to get these services running is to set up sentry, and add the following line
+in ``~/.sentry/sentry.conf.py``::
+
+    SENTRY_EVENTSTREAM = "sentry.eventstream.kafka.KafkaEventStream"
```
onewland, line 21: It's a little nit-picky but you don't actually need to make this change in order to have sentry/snuba work together.

If you don't make this configuration change, events will be sent over the HTTP API through kind of a hacky backdoor, but that's what we use in the (sentry-side) integration tests so it's probably not going to go away and is maybe more reliable than Kafka in local dev environments

onewland: I think it's fine to merge this change -- most Snuba developers will want events to flow through Kafka -- but getting Kafka running and changing the sentry configuration to send events through Kafka vs. the HTTP endpoint are independent.

PR #2669, opened 2022-05-03
debug(replacer): add structured logs for errors_replacer

onewland:
This logs out JSON for the incoming message so that we can search for logs by attribute in GCP if we have a replacements issue we need to debug


snuba/datasets/errors_replacer.py
```diff
@@ -488,6 +489,9 @@ def __init__(
     def process_message(self, message: ReplacementMessage) -> Optional[Replacement]:
         type_ = message.action_type
 
+        attributes_json = json.dumps({"message_type": type_, **message.data})
```
MeredithAnya, line 492: do we care about all message types? It seems like down below we don't process all of them? 

```python
if type_ in (
            ReplacementType.START_DELETE_GROUPS,
            ReplacementType.START_MERGE,
            ReplacementType.START_UNMERGE,
            ReplacementType.START_UNMERGE_HIERARCHICAL,
            ReplacementType.START_DELETE_TAG,
        ):
            return None
```

MeredithAnya (approved):
my comment is just more of a question, not necessarily a suggestion that you'd need to change anything.

snuba/datasets/errors_replacer.py
onewland: I think it's useful to log receipt of them even if we don't act on them (we act on the `END_*` messages as you point out) but I'm not against an allowlist if that's what people prefer.

PR #2672, opened 2022-05-04
fix(snql): Return 400 error code on queries with type errors

mcannizz:
When queries fail with Clickhouse errors, return a 400 status code when
the Clickhouse error code indicates a type error in the query. This ensures
our SLOs are not impacted when users make mistakes in their queries.

### DIFF
--- a/tests/test_snql_api.py
+++ b/tests/test_snql_api.py
@@ -866,3 +866,47 @@ class TestSnQLApi(BaseApiTest):
         )
 
         assert response.status_code == 200
+
+    def test_clickhouse_type_mismatch_error(self) -> None:
+        """Test that snql queries that cause Clickhosue type errors have a 400 status code."""
+        response = self.post(
+            "/discover/snql",
+            data=json.dumps(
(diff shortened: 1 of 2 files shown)


tests/test_snql_api.py
```diff
@@ -866,3 +866,47 @@ def test_wrap_log_fn_with_ifnotfinite(self) -> None:
...
 
         assert response.status_code == 200
+
+    def test_clickhouse_type_mismatch_error(self) -> None:
+        """Test that snql queries that cause Clickhosue type errors have a 400 status code."""
+        response = self.post(
+            "/discover/snql",
+            data=json.dumps(
+                {
+                    "query": f"""MATCH (discover_events )
```
onewland, line 876: nit: weird space after `discover_events`

onewland (approved):

tests/test_snql_api.py
mcannizz: Good eye, fixed!

PR #2678, opened 2022-05-06
feat(subscriptions): Add logging for skipped subscriptions

lynnagara:
Log when subscriptions are skipped due to the `subscriptions_max_latency_sec`
override

### DIFF
--- a/snuba/subscriptions/worker.py
+++ b/snuba/subscriptions/worker.py
@@ -229,16 +229,17 @@ class SubscriptionWorker(
         max_validity_latency = state.get_config("subscriptions_max_latency_sec", 0)
         assert isinstance(max_validity_latency, int)
 
-        if max_validity_latency == 0:
-            tasks: Iterator[ScheduledSubscriptionTask] = self.__schedulers[
-                message.partition.index
-            ].find(tick)
-        else:
-            tasks = (
-                t
-                for t in self.__schedulers[message.partition.index].find(tick)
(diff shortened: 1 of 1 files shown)


snuba/subscriptions/worker.py
```diff
@@ -229,16 +229,17 @@ def process_message(
...
-                message.partition.index
-            ].find(tick)
-        else:
-            tasks = (
-                t
-                for t in self.__schedulers[message.partition.index].find(tick)
-                if datetime.timestamp(t.timestamp) > time.time() - max_validity_latency
+        if (
+            max_validity_latency != 0
+            and time.time() - datetime.timestamp(tick.timestamps.lower)
```
onewland, line 234: I don't doubt that this is the right change but it looks like you're changing the logic a bit here -- is that right and intended?

My read of the old logic:
> create a list of tasks and filter out the ones where the task timestamp is too old

My read of the new logic:
> if the incoming tick has a low timestamp older than a certain threshold, don't try to schedule any tasks related to that tick

Question about the new logic if I'm right: Should we be basing this on `upper`? Since it sounds like basing it on the `lower` could discard relevant ticks. If:

- max_valid_latency=10
- system time=160

and the tick you receive covers the range `[120,150]` could that tick represent timestamp 140 which is within your `max_valid_latency`? I think basing it on `lower` would discard that whole tick

You don't need to block merge on this question, I just want to check if I'm following along correctly.

onewland (approved):
You can ignore that first [deleted] comment, I answered my own question with more research

PR #2680, opened 2022-05-09
chore(tiger): Raise a different timeout exception from tiger clusters

nikhars:
### Context
When queries are run against the tiger cluster and the operation is performed via readthrough caching, we see a lot of ExecutionTimeoutErrors. This happens because uwsgi respawns the processes every 3 minutes. The respawn ensures that the requests sent to the processes have all returned their responses. So we don't see this issue with the primary queries. Unfortunately, there is no way to ask uwsgi to wait for queries being run in threads that need time to run. Those threads get killed mercilessly.

### Solution
Differentiate the exceptions being raised be queries from tiger cluster from regular production clusters. That way we can tune the alerts to ignore certain tiger exceptions.

----

### DIFF
--- a/snuba/state/cache/abstract.py
+++ b/snuba/state/cache/abstract.py
@@ -15,6 +15,10 @@ class ExecutionTimeoutError(ExecutionError):
     pass
 
 
+class TigerExecutionTimeoutError(ExecutionError):
+    pass
+
+
 class Cache(Generic[TValue], ABC):
     @abstractmethod
     def get(self, key: str) -> Optional[TValue]:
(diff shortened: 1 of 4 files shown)


volokluev:
This PR is a bit of a hack  because it makes the generic cache know what a `tiger` is. Could we make the ExecutionTimeoutError be the configurable thing and pass it in on cache instantiation?

nikhars:
> This PR is a bit of a hack because it makes the generic cache know what a `tiger` is. Could we make the ExecutionTimeoutError be the configurable thing and pass it in on cache instantiation?

Moved the type of Exception to be raised to be configured during instantiation.

evanh (approved):

PR #2681, opened 2022-05-09
feat(replays): initial replays clickhouse migration

JoshFerge:
## Summary
Creates the initial clickhouse migration for the replays dataset. Rows be uniquely added by their `replay_id` (many rows will have same replay_id) along with a monotonically increasing `sequence_id` that represents each additional piece of data for the replay.

This table will be used initially for simply searching and listing replays on the replays index page. basic aggregations will be done to gather the duration of the replay, and the list of trace_ids associated with the replay to start. 

further columns will likely be introduced as product requirements determine the types of searches we want to do.

See the spec here for more information. https://www.notion.so/sentry/Session-Replay-V1-alpha-Ingest-Backend-ae068d1e1d514221b6c3ea2233f360f4


snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -0,0 +1,96 @@
...
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+raw_columns: Sequence[Column[Modifiers]] = [
+    Column("replay_id", UUID()),
+    Column("sequence_id", UInt(16)),
```
lynnagara, line 20: Who generates the sequence_id? On the SDK? Sentry? Snuba? What's the max number allowed here?

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.REPLAYS,
+                table_name="replays_local",
+                columns=raw_columns,
+                engine=table_engines.ReplacingMergeTree(
+                    storage_set=StorageSetKey.REPLAYS,
+                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id), sequence_id)",
```
lynnagara, line 62: Just to confirm, items with the same replay_id can still span multiple days right?

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -0,0 +1,96 @@
...
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+raw_columns: Sequence[Column[Modifiers]] = [
+    Column("replay_id", UUID()),
+    Column("sequence_id", UInt(16)),
+    Column("trace_ids", Array(UUID())),
```
lynnagara, line 21: What are these trace_ids? Is it supposed to be a pointer to some other piece of data in one of our systems?

nikhars (commented):
Is there a document I can read on the use cases replays is trying to solve? More specifically what sort of queries would be run against this dataset.

snuba/migrations/snuba_migrations/replays/0001_replays.py
nikhars, line 20: Is there any sort of relation between `sequence_id` and `replay_id` field?

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -0,0 +1,96 @@
...
+    # user columns
+    Column("user", String()),
+    Column("user_hash", UInt(64)),
+    Column("user_id", String(Modifiers(nullable=True))),
+    Column("user_name", String(Modifiers(nullable=True))),
+    Column("user_email", String(Modifiers(nullable=True))),
+    # sdk info
+    Column("sdk_name", String()),
+    Column("sdk_version", String()),
+    Column("tags", Nested([("key", String()), ("value", String())])),
```
nikhars, line 43: For performance reasons, you might want to add bloom filter index on tags as we do on some of our other datasets

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.REPLAYS,
+                table_name="replays_dist",
+                columns=raw_columns,
+                engine=table_engines.Distributed(
+                    local_table_name="replays_local",
+                    sharding_key="project_id",
```
nikhars, line 86: What is the reason for sharding the data by `project_id` versus sharding randomly? One disadvantage I can see with sharding by `project_id is that if there is a big project which uses replays a lot, the shards could become imbalanced.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: the SDK will generate the sequence id. max number will be between ~100 and ~1000. (we will be capping replays max length time-wise and from there find a sane max seq_id).

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: yes. i'm glad you brought this up. the intention is that: replays which span across multiple days will only show up when the initial event is within the time range.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: https://github.com/getsentry/sentry-replay/issues/38#issuecomment-1110054329

replays can have N trace_ids, and each update may have N of them, and trace_id will be the link between them to start (we won't be doing any joins with them)

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: `sequence_id` will be a monotonically increasing counter, so for each `replay_id`, `sequence_id` is unique.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: good call 👍🏼  will look at adding those.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: good point, I think i just chose this arbitrarily. I'll shard by `replay_id` instead.

JoshFerge:
> Is there a document I can read on the use cases replays is trying to solve? More specifically what sort of queries would be run against this dataset.
I've added an addendum to the doc here:
https://www.notion.so/sentry/Addendum-Replay-Queries-fcfd8e68679e443e87649014cf10ae62

snuba/migrations/snuba_migrations/replays/0001_replays.py
fpacifici: Are you actually going to search for replays by tag key/value ?

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -0,0 +1,96 @@
...
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.REPLAYS,
+                table_name="replays_local",
+                columns=raw_columns,
+                engine=table_engines.ReplacingMergeTree(
```
fpacifici, line 60: Are you going to use the replacing feature for something (aside for removing duplicates, which is anyway a good idea) ?

snuba/migrations/snuba_migrations/replays/0001_replays.py
nikhars: Will `sequence_id` always start from a fixed number for each `replay_id`? And how many `sequence_id` are allowed for a single `replay_id`. Is there a maximum?

snuba/migrations/snuba_migrations/replays/0001_replays.py
fpacifici: @JoshFerge  Could you please provide some insights of the most common query pattern you expect? 
What are you going to filter by most times? 
What are you going to aggregate, if anything ?

The order by key has to be defined based on the expected query pattern. You cannot change it once done without rebuilding the table entirely and getting it wrong will make your query performance miserable. 
Also the expected query pattern impacts which (if any) data skipping indexes should be added. We cannot add indexes to all columns as the type of index depend on the query you want to make faster.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: on a per row basis it likely won't be more than 10

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: no, just removing duplicates.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: https://www.notion.so/sentry/Addendum-Replay-Queries-fcfd8e68679e443e87649014cf10ae62
see above ^^. We will be happy to rebuild the table entirely while we are testing for the next several months, so viewing all data as temporary and will make that clear to any customers testing. we will be building several use cases on top of this initial that may require us to re-build the tables at any rate.

fpacifici: Re: the document you linked:

"As a replays user, I want to see all replays where an error occurred"
The table designed here does not seem to have a reference to an issue or an error. Is that correct or a mistake ?

"As a performance user, I want to see if this trace has a replay associated with it"
If you want to search the replays table `WHERE has(trace_ids, 'asdasdasdasd')` please add a bloom filter index on that column (which may require you to create a materialized version with hashes of that column). Otherwise your search will be miserable. 
But it would be better to add a replay id on the transaction in some way so you do not have to scan the whole replay table to associate replays to traces.

JoshFerge: > The table designed here does not seem to have a reference to an issue or an error. Is that correct or a mistake ?

Not a mistake. For now, we will do very rudimentary query where we take the trace ids from a single page and look them up to determine if there is an error associated, or do a search on the errors table looking for the replays tag.

The issue is that since errors can be sampled / dropped, (and replays too in the future), tagging each other's events with the ids is problematic because it's not guaranteed that the tagged id will exist.

we'll likely need some separate table that's generated in event post_processing that can accurately associate ingested events with replays. this will come in a future iteration.

>But it would be better to add a replay id on the transaction in some way so you do not have to scan the whole replay table to associate replays to traces.

we'll also be adding replay_id on other events, so for example this search can use transactions tagged with a replay id. (there is still the sampling problem, but not going to worry about this in first iteration)

JoshFerge: and for now will not add bloom filter index, will add TODO. something I can follow up on.

JoshFerge: went ahead and added the index 👍🏼

JoshFerge:
> Please fix the type errors

fixed, accidentally included an errant file which caused the errors.

fpacifici (approved):

JoshFerge:
will merge monday. thanks for reviews all.

PR #2682, opened 2022-05-09
fix(querylog): compress, allow larger messages on the querylog producer

volokluev:
Sometimes the message is too large for the querylog topic. Add compression to the producer and allow larger messages. There is also an ops ticket to allow larger messages on the broker

### DIFF
--- a/snuba/state/__init__.py
+++ b/snuba/state/__init__.py
@@ -315,13 +315,20 @@ def record_query(query_metadata: Mapping[str, Any]) -> None:
                 # the querylog payloads can get really large so we allow larger messages
                 # (double the default)
                 # The performance is not business critical and therefore we accept the tradeoffs
-                # in more bandwidth for more observability
+                # in more bandwidth for more observability/debugability
                 # for this to be meaningful, the following setting has to be matched on the broker:
                 # max.message.bytes=2097176
                 build_kafka_producer_configuration(
                     topic=None,
                     override_params={
-                        "compression.type": "zstd",
+                        # at time of writing (2022-05-09) lz4 was chosen because it
+                        # compresses quickly. If more compression is needed at the cost of
+                        # performance, zstd can be used instead. Recording the query
+                        # is part of the API request, therefore speed is important
+                        # perf-testing: https://indico.fnal.gov/event/16264/contributions/36466/attachments/22610/28037/Zstd__LZ4.pdf
+                        # by default a topic is configured to use whatever compression method the producer used
+                        # https://docs.confluent.io/platform/current/installation/configuration/topic-configs.html#topicconfigs_compression.type
+                        "compression.type": "lz4",
                         "max.request.size": 2097176,
                     },
                 )
--- a/snuba/state/__init__.py
+++ b/snuba/state/__init__.py
@@ -24,9 +24,7 @@ from confluent_kafka import Producer
 from snuba import environment, settings
 from snuba.redis import redis_client
 from snuba.utils.metrics.wrapper import MetricsWrapper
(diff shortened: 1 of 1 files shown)


evanh (commented):
Does the consumer automatically handle the compressed messages as well? Or do we need to make changes to the consumers as well? Keep in mind that the super-big-consumers are not deployed with snuba.

snuba/state/__init__.py
```diff
@@ -313,7 +311,28 @@ def record_query(query_metadata: Mapping[str, Any]) -> None:
...
+                    override_params={
+                        # at time of writing (2022-05-09) lz4 was chosen because it
+                        # compresses quickly. If more compression is needed at the cost of
+                        # performance, zstd can be used instead. Recording the query
+                        # is part of the API request, therefore speed is important
+                        # perf-testing: https://indico.fnal.gov/event/16264/contributions/36466/attachments/22610/28037/Zstd__LZ4.pdf
+                        # by default a topic is configured to use whatever compression method the producer used
+                        # https://docs.confluent.io/platform/current/installation/configuration/topic-configs.html#topicconfigs_compression.type
+                        "compression.type": "lz4",
+                        "max.request.size": 2097176,
```
lynnagara, line 332: Where does `max.request.size` come from? I didn't see it in the list of librdkafka configurations

lynnagara: Our other producers use `message.max.bytes`

volokluev: https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#producerconfigs_max.request.size

volokluev:
> Does the consumer automatically handle the compressed messages as well?

The consumer should not need to care about this. They interact with the broker, not the producer

snuba/state/__init__.py
```diff
@@ -313,7 +311,28 @@ def record_query(query_metadata: Mapping[str, Any]) -> None:
...
+                    topic=None,
+                    override_params={
+                        # at time of writing (2022-05-09) lz4 was chosen because it
+                        # compresses quickly. If more compression is needed at the cost of
+                        # performance, zstd can be used instead. Recording the query
+                        # is part of the API request, therefore speed is important
+                        # perf-testing: https://indico.fnal.gov/event/16264/contributions/36466/attachments/22610/28037/Zstd__LZ4.pdf
+                        # by default a topic is configured to use whatever compression method the producer used
+                        # https://docs.confluent.io/platform/current/installation/configuration/topic-configs.html#topicconfigs_compression.type
+                        "compression.type": "lz4",
```
nikhars, line 331: Are there any implications of compressing the data in the middle of a stream? Let's say there are 1000 messages already on the topic which have not been read by the consumer of this topic. And then we enable this setting. So messages from 1001 to 2000 are compressed.

The consumer which reads from this topic would see messages from 0 to 1000 uncompressed and messages from 1001 to 2000 compressed. Will it gracefully handle this scenario?

nikhars: Just noticed that Evan asked the same question and you already responded to it. But I am not so sure about the answer. Are you saying that on a topic different messages can have different compressions and its the broker's job to specify the data to the consumer in an uncompressed form(or whatever setting if any the consumer has configured)?

volokluev: A consumer cannot even specify a compression type so yes  on a topic different messages can have different compressions and its the broker's job to specify the data to the consumer in an uncompressed form.

Only way it can break is if the broker specifies a compression type and the producer does not follow it.

https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html

nikhars (approved):

PR #2683, opened 2022-05-10
feat(replays): replays processor to ingest data into clickhouse

JoshFerge:
following up on #2681, we add a processor to insert replay data rows into clickhouse. 

common code has been copy/pasted from relevant consumers for things like tag/environment processing etc.

Will note that we'll be rapidly prototyping on top of this, so opting for quicker / more isolated changes (as opposed to e.g. refactoring to be more DRY for tag processing).


snuba/datasets/replays_processor.py
```diff
@@ -0,0 +1,170 @@
...
+        sdk = event_dict["data"].get("sdk", None) or {}
+        processed["sdk_name"] = _unicodify(sdk.get("name") or "")
+        processed["sdk_version"] = _unicodify(sdk.get("version") or "")
+
+        if processed["sdk_name"] == "":
+            metrics.increment("missing_sdk_name")
+        if processed["sdk_version"] == "":
+            metrics.increment("missing_sdk_version")
+
+    def process_message(
```
evanh, line 145: For now, can you put all the logic in this function into a try/except that logs a warning/sets a metric in Datadog? Currently if a bad message or something causes an exception it will crash the consumer.

Usually during the testing/experimentation phase we log errors on bad messages instead of crashing outright, and then once the data model etc. have been solidified we remove the error handling.

snuba/datasets/entities/replays.py
```diff
@@ -0,0 +1,51 @@
...
+from snuba.query.processors.object_id_rate_limiter import ProjectRateLimiterProcessor
+from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
+from snuba.query.validation.validators import (
+    ColumnValidationMode,
+    EntityRequiredColumnValidator,
+)
+from snuba.utils.schemas import Column
+
+replays_data_model = EntityColumnSet(
+    [Column("project_id", UInt(64)), Column("timestamp", DateTime())]
```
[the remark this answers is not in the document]
JoshFerge: yup, for now. api will come in a future PR.

snuba/datasets/replays_processor.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _structure_and_validate_message(
+        self, message: Mapping[Any, Any]
+    ) -> Optional[Tuple[EventDict, RetentionDays]]:
+
+        event = message
+        data = event["data"]
+        event_type = data.get("type")
+        if event_type != "replay_event":
+            return None
```
fpacifici, line 56: This seems odd. Why would you expect to have anything other than replays on the Kafka topic that you are consuming ?

snuba/datasets/replays_processor.py
fpacifici: Why not just applying the Dead letter queue? Now we have one. 
Just set up a dead letter queue policy when you define the storage (which is something you have to do anyway in order to consume)
https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/metrics.py#L96

snuba/datasets/replays.py
```diff
@@ -0,0 +1,7 @@
+from snuba.datasets.dataset import Dataset
+from snuba.datasets.entities import EntityKey
+
+
+class ReplaysDataset(Dataset):
+    def __init__(self) -> None:
+        super().__init__(default_entity=EntityKey.REPLAYS)
```
fpacifici, line 7: https://github.com/getsentry/snuba/blob/672406e88f6f1199cad7ad43b43d0fa9996078ce/snuba/datasets/dataset.py#L46-L54

Please override `is_experimental`. This will allow you to deploy a broken version of the dataset without impacting snuba production deployment

evanh (commented):
This looks good to me, but please mark the dataset as experimental (Filippo's comment)

snuba/datasets/replays_processor.py
JoshFerge: added a TODO for setting up a deadletter queue, need to provision the topic, etc.

JoshFerge:
ugh bad rebase, will re-open.

PR #2686, opened 2022-05-10
fix: Support array as well as tuple in condition

evanh:
The condition checker that gets used to find org/project IDs amongst other
things was always looking for IN conditions with a tuple. While that has been
working up until now, it should also support arrays as a way to pass IDs in.

### DIFF
--- a/snuba/query/conditions.py
+++ b/snuba/query/conditions.py
@@ -87,7 +87,7 @@ def __set_condition_pattern(
         (
             Param("lhs", lhs),
             Param(
-                "tuple",
+                "sequence",
                 Or(
                     [
                         FunctionCallPattern(String("tuple"), None),
--- a/snuba/clickhouse/query_dsl/accessors.py
+++ b/snuba/clickhouse/query_dsl/accessors.py
@@ -57,7 +57,7 @@ def get_object_ids_in_query_ast(
             Column(column_name=String(object_column))
         ).match(condition)
         if match is not None:
-            objects = match.expression("tuple")
+            objects = match.expression("sequence")
             assert isinstance(objects, FunctionCallExpr)
             return {
                 lit.value
--- a/snuba/query/processors/abstract_array_join_optimizer.py
+++ b/snuba/query/processors/abstract_array_join_optimizer.py
@@ -271,7 +271,7 @@ def string_literal_in_condition_extractor(
         if match is None:
             return set()
 
-        function = match.expression("tuple")
+        function = match.expression("sequence")
         assert isinstance(function, FunctionCallExpr)
 
         return {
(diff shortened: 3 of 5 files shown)


tests/query/processors/test_project_rate_limiter.py
```diff
@@ -42,7 +51,7 @@
             binary_condition(
                 ConditionFunctions.IN,
                 Column("_snuba_project_id", None, "project_id"),
-                FunctionCall(None, "array", (Literal(None, 4), Literal(None, 5))),
+                FunctionCall(None, "array", (Literal(None, 3), Literal(None, 5))),
```
volokluev, line 54: this change makes no difference right?

tests/query/processors/test_project_rate_limiter.py
```diff
@@ -54,7 +63,7 @@
 @pytest.mark.parametrize("unprocessed, project_id", tests)
 def test_project_rate_limit_processor(unprocessed: Expression, project_id: int) -> None:
     query = Query(
-        QueryEntity(EntityKey.EVENTS, ColumnSet([])),
+        QueryEntity(EntityKey.EVENTS, EntityColumnSet([])),
```
volokluev, line 66: does this change make any difference?

snuba/query/conditions.py
```diff
@@ -86,7 +86,15 @@ def __set_condition_pattern(
         String(operator),
         (
             Param("lhs", lhs),
-            Param("tuple", FunctionCallPattern(String("tuple"), None)),
+            Param(
```
volokluev, line 89: It seems like the only place this is tested is the rate limiter but it's in a place for generic use. Can we have a test in `tests/query/test_conditions.py` to test this as a unit?

snuba/query/conditions.py
```diff
@@ -86,7 +86,15 @@ def __set_condition_pattern(
...
+                "tuple",
```
volokluev, line 90: the name of this param is a little misleading. maybe let's call it `collection` or `sequence`?

volokluev (commented):
Looks good to me overall but I'd like to see a unit test of the main change.

tests/query/processors/test_project_rate_limiter.py
evanh: It fixes typing errors in the file.

tests/query/processors/test_project_rate_limiter.py
evanh: This change is necessary to have the tests pass. If there is a `=` condition and an `IN` condition, the code does an intersect to find the potential IDs. This worked before because the array wasn't being seen, so only the `=` condition was being considered. Since now both are being used, they have to overlap.

volokluev: cool!

volokluev (approved):
💯

PR #2691, opened 2022-05-11
Revert "ref(tiger): Allow increasing cache wait time for tiger"

nikhars:
We don't need this change since we found out the reason of query timeouts on tiger

Reverts getsentry/snuba#2651


onewland (approved):

lynnagara:
This seems to be stale, please reopen if still needed.

PR #2692, opened 2022-05-12
fix: Allow arbitrary AppIDs

evanh:
This was originally written with the intention of having a registry of AppIDs
that could be used to check and validate incoming values. It's not clear if that
is still a strategy we want to use in the future, so for now allow arbitrary
values so we can start collecting some basic data.


snuba/attribution/__init__.py
```diff
@@ -1,31 +1,24 @@
...
 metrics = MetricsWrapper(environment.metrics, "snuba.attribution")
 
 
-APPIDS: dict[str, AppID] = {}
-DEFAULT_APPID = AppID.from_dict(DEFAULT_APPIDS[0])
-INVALID_APPID = AppID.from_dict(DEFAULT_APPIDS[1])
-for obj in DEFAULT_APPIDS:
-    app_id = AppID.from_dict(obj)
-    APPIDS[app_id.key] = app_id
+DEFAULT_APPID = AppID("default", "evanh", datetime(2022, 3, 24))
```
nikhars, line 13: Nit: You can probably set the created by field to "snuba" and not have your name in the default cases

snuba/attribution/__init__.py
```diff
@@ -1,31 +1,24 @@
...
+INVALID_APPID = AppID("invalid", "evanh", datetime(2022, 3, 25))
+
 
+def get_app_id(app_id: str) -> AppID:
+    if not app_id:
```
nikhars, line 18:
```suggestion
return AppID(app_id) if app_id else INVALID_APPID
```

nikhars (approved):
Left minor comments. But I will approve.

snuba/attribution/__init__.py
evanh: But then people won't know who created them! ;)

PR #2697, opened 2022-05-13
feat(kafka): Extend KafkaMessageMetadata

rahul-kumar-saini:
### Overview
- Currently, `KafkaMessageMetadata` only contains a message timestamp, offset, and partition index
- It would be useful to also have the topic, and key+headers of the payload so eventually the `InvalidKafkaMessage` class in Arroyo introduced by DLQ changes could be used
- `MessageProcessor`s currently don't have enough data to use the appropriate invalid message class so only the `InvalidRawMessage` class (from Arroyo) is currently useful
- Eg. https://github.com/getsentry/snuba/pull/2614 Could have used `InvalidKafkaMessage` in order to produce more debug data to a dead-letter topic

### Changes
- Added extra fields to the metadata tuple
- Most of this PR is just updating tests to add the new fields to the tuple

### Notes
- Ideally, there should also be a way to figure out which consumer group a message was consumed/processed by, is there a way to do this?

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -607,7 +607,12 @@ def process_message(
     result = processor.process_message(
         rapidjson.loads(message.payload.value),
         KafkaMessageMetadata(
-            message.offset, message.partition.index, message.timestamp
+            offset=message.offset,
+            partition=message.partition.index,
+            timestamp=message.timestamp,
+            topic=message.partition.topic.name,
+            key=message.payload.key,
+            headers=message.payload.headers,
         ),
     )
 
@@ -651,7 +656,12 @@ def process_message_multistorage(
 ) -> MultistorageProcessedMessage:
     value = rapidjson.loads(message.payload.payload.value)
     metadata = KafkaMessageMetadata(
-        message.offset, message.partition.index, message.timestamp
+        offset=message.offset,
+        partition=message.partition.index,
+        timestamp=message.timestamp,
+        topic=message.partition.topic.name,
+        key=message.payload.payload.key,
+        headers=message.payload.payload.headers,
     )
 
     results: MutableSequence[
@@ -686,7 +696,12 @@ def process_message_multistorage_identical_storages(
     """
     value = rapidjson.loads(message.payload.payload.value)
     metadata = KafkaMessageMetadata(
-        message.offset, message.partition.index, message.timestamp
+        offset=message.offset,
+        partition=message.partition.index,
+        timestamp=message.timestamp,
+        topic=message.partition.topic.name,
+        key=message.payload.payload.key,
+        headers=message.payload.payload.headers,
     )
 
     intermediate_results: MutableMapping[
--- a/snuba/consumers/types.py
+++ b/snuba/consumers/types.py
@@ -1,8 +1,13 @@
 from datetime import datetime
-from typing import NamedTuple
+from typing import NamedTuple, Optional
+
+from arroyo.backends.kafka.consumer import Headers
 
 
 class KafkaMessageMetadata(NamedTuple):
     offset: int
     partition: int
     timestamp: datetime
+    topic: str
+    key: Optional[bytes]
+    headers: Headers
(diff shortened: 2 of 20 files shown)


lynnagara (changes requested):
Why do we need to put this in the metadata? The headers and key are considered part of the message itself

snuba/consumers/consumer.py
```diff
@@ -607,7 +607,12 @@ def process_message(
...
         KafkaMessageMetadata(
-            message.offset, message.partition.index, message.timestamp
+            offset=message.offset,
+            partition=message.partition.index,
+            timestamp=message.timestamp,
+            topic=message.partition.topic.name,
+            key=message.payload.key,
+            headers=message.payload.headers,
         ),
     )
```
rahul-kumar-saini, line 617: @lynnagara As far as I can tell, previously the processors only received the message payload's value (some bytes) and the offset, partition, and timestamp.

Key, Topic, and Headers are all missing here.

lynnagara: Why do we want to handle the invalid messages inside the MessageProcessor.process_message function though, which this is what is getting passed the KafkaMessageMetadata. Doesn't it mean every individual message processor needs to implement handling of the DLQ themselves? How about just raising the InvalidMessage from this function itself which already has access to all of this information from the original message?

rahul-kumar-saini: > Why do we want to handle the invalid messages inside the MessageProcessor.process_message function though

We don't want to handle the message in the processor, we just want this information _available_ to the processor in order to throw all of it to the DLQ.

> How about just raising the InvalidMessage from this function itself which already has access to all of this information from the original message?

I think the only information the message processors have right now is the message payload value (not including key+headers), the offset, partition, and timestamp. We want to add key, headers, and topic as well.

lynnagara: What I'm suggesting is having this function (or some other outer function) do:

```
try:
    result = processor.process_message(.....)
except Exception as exc:
    raise InvalidMessage(message.headers, message.key, etc....)
```

Otherwise all of the message processors (which are dataset and product specific) will need to implement code to raise this themselves in the right places. I think we could just catch any processing error and then reraise it as an InvalidMessage error here with all the right metadata attached rather than pass the metadata deep into every processing step and updating all of the processing steps to handle it.

rahul-kumar-saini: Oh hmm, I was assuming we _do_ want processing steps to be able to raise stuff to the DLQ where they want. What you're suggesting makes more sense, but it means that any exceptions we had previously would all be `InvalidMessages` exceptions now. Is that fine?

lynnagara: I think the error that gets logged to Sentry for debugging should always contain as much useful information as possible and be grouped properly whether or not there is a DLQ. So I think it's fine that it's an `InvalidMessage` error instead of what it is previously as long as it's properly grouped and contains all the useful information. I don't think we should raise the plural `InvalidMessages` (which I thought we were going to deprecate anwyay). I do think there should be one Sentry exception for every invalid message though, not one for any number of `InvalidMessages`.

We had a conversation earlier about getting rid of `InvalidMessages` and moving to `InvalidMessage` - where did this end up?

rahul-kumar-saini: > So I think it's fine that it's an `InvalidMessage` error instead of what it is previously as long as it's properly grouped and contains all the useful information

Alright cool, I do like that we can wrap the entirety of `process_message` with a try-catch, maybe we can update the DLQ Exception model to optionally contain some sort of exception that was caught.

> We had a conversation earlier about getting rid of `InvalidMessages` and moving to `InvalidMessage` - where did this end up?

Uh I think we ultimately decided on having 1 exception class `InvalidMessages` which contains 1 or more `InvalidMessage` objects.

`InvalidRawMessage` and `InvalidKafkaMessage` being the implementations of the abstract `InvalidMessage` class.

Relevant PR: https://github.com/getsentry/arroyo/pull/64/

rahul-kumar-saini:
Closing as this solution likely isn't necessary.

PR #2699, opened 2022-05-13
fix(tiger): don't call the explicit tuple function in clickhouse

volokluev:
After drilling down enough, it seems like this is actually the cause of the bugs. I thought I tried it a week ago when @evanh told me to but I guess I didn't try it correctly. Either way this should actually fix the problems. It looks like we already had to implement a similar fix for arrays so this should be no different.

### DIFF
--- a/snuba/clickhouse/formatter/expression.py
+++ b/snuba/clickhouse/formatter/expression.py
@@ -137,6 +137,11 @@ class ExpressionFormatterBase(ExpressionVisitor[str], ABC):
             # and work when they are passed as [1, 2, 3]
             return self._alias(f"[{self.__visit_params(exp.parameters)}]", exp.alias)
 
+        if exp.function_name == "tuple":
+            # Some distributed queries fail when tuples are passed as tuple(1,2,3)
+            # and work when they are passed as (1, 2, 3)
+            return self._alias(f"({self.__visit_params(exp.parameters)})", exp.alias)
+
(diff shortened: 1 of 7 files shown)


nikhars (approved):

nikhars:
Is there a way to enforce this only on tiger queries? I think the answer is No. But would it make sense to try and restrict this only to tiger queries?

volokluev:
Is it possible? Yes. Is it easily possible without dirty hacks? no

Does it make sense?

it makes sense-ish. This change would affect every query but there's no reason it shouldn't work I can hide it behind a runtime flag so we can turn it on slowly if that makes you feel more comfortable

volokluev:
I'm also not going to deploy this on a friday afternoon. I'll wait till monday

nikhars:
> Is it possible? Yes. Is it easily possible without dirty hacks? no
> 
> Does it make sense?
> 
> it makes sense-ish. This change would affect every query but there's no reason it shouldn't work I can hide it behind a runtime flag so we can turn it on slowly if that makes you feel more comfortable

The change seems relatively straightforward so I am fine to run this without the runtime flag.

nikhars:
> I'm also not going to deploy this on a friday afternoon. I'll wait till monday

I would probably try this out on a Friday than on Monday because of low traffic volumes. On Monday we run the weekly reports so we are running more queries. Just my opinion. You are free to decide when you want to try it out.

PR #2700, opened 2022-05-13
fix(subscriptions): Only flush the querylog producer when we are done producing

lynnagara:
Currently we are flushing the querylog producer before shutting the strategy,
however closing the strategy can result in some final query executions which
will produce more messages to the querylog. This changes moves it later
when we are finished producing to the querylog.

### Legal Boilerplate

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -495,3 +495,8 @@ class ProduceResult(ProcessingStrategy[SubscriptionTaskResult]):
             logger.debug("Flushing executor producer")
             messages = producer.flush(*[remaining] if remaining is not None else [])
             logger.debug(f"{messages} executor messages pending delivery")
+
+        # Flush querylog producer
+        logger.debug("Flushing querylog producer")
+        state.flush_producer()
+        logger.debug("Flushed querylog producer")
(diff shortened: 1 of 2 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -495,3 +495,8 @@ def join(self, timeout: Optional[float] = None) -> None:
             logger.debug("Flushing executor producer")
             messages = producer.flush(*[remaining] if remaining is not None else [])
             logger.debug(f"{messages} executor messages pending delivery")
+
+        # Flush querylog producer
+        logger.debug("Flushing querylog producer")
+        state.flush_producer()
```
MeredithAnya, line 501: would it be worth having metrics on how long it's taking to flush the querylog producer? cause now we will be flushing on rebalances as well right?

lynnagara: Yeah, you're right. I actually don't really want to flush it on rebalance. Let me see if I can move this somewhere else so that doesn't happen.

lynnagara: Moved it back out of the strategy, let me know what you think

MeredithAnya (approved):

PR #2702, opened 2022-05-16
fix(dlq): Skip bad message if DLQ inactive

rahul-kumar-saini:
### Overview
- Metrics DLQ is not yet active
- I added a sentry error in case we encounter a bad message so we can just skip but keep track
- I realize it wouldn't actually skip, it would just log an error and try processing it anyway
    - This is _bad_ and is currently in prod - thankfully no bad messages have been encountered yet.

### Changes
- Modified Metrics code to return a `None` from the message processor in case of bad message and DLQ is disabled
- I believe this effectively skips the message

### Notes
- The only reason this extra check exists is in case the DLQ does not work
- This additional logic will be removed in favor of just the DLQ logic once we verify it works as intended

### DIFF
--- a/snuba/datasets/metrics_bucket_processor.py
+++ b/snuba/datasets/metrics_bucket_processor.py
@@ -39,7 +39,9 @@ class MetricsBucketProcessor(MessageProcessor, ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+    def _process_values(
+        self, message: Mapping[str, Any]
+    ) -> Optional[Mapping[str, Any]]:
         raise NotImplementedError
 
     def process_message(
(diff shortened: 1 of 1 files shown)


MeredithAnya:
So is the whole `_raise_invalid_message` going to be replaced is that what you mean by "This additional logic will be removed in favor of just the DLQ logic once we verify it works as intended"? If so I would leave a #ToDo in the doc string or something

rahul-kumar-saini:
@MeredithAnya Just the `return None` everywhere and the if/else statement inside `_raise_invalid_message` will go away leaving just the `raise InvalidMessages`.

The logic going away is the one that is a backup for if raising that exception to hit the DLQ doesn't work for some reason

snuba/datasets/metrics_bucket_processor.py
```diff
@@ -76,14 +81,18 @@ def process_message(
         except EventTooOld:
             return None
 
+        processed_values = self._process_values(message)
```
MeredithAnya, line 84: I guess another option would be to have a try catch here if you wanted to do the logic of the runtime config here. 

```python
try:
    process_values = self.__process_values(message)
except InvalidMessages as e:
    if state.get_config("enable_metrics_dlq", True):
        raise
    else:
      logger.error() # whatever you wanna log
      return None
```

But not sure if that's easier

MeredithAnya (approved):

snuba/datasets/metrics_bucket_processor.py
rahul-kumar-saini: That is easier for `_process_values()` but I have the raise function being used a couple spots above too which would need a try/except for. I'm going to revert all this tomorrow after the prod test anyway so I don't think it's a big deal to clean this up for now.

PR #2705, opened 2022-05-16
ref: Remove subscriptions verifier

lynnagara:
This was used temporarily to diff results between the old and new subscriptions
pipeline. It's no longer needed.


volokluev (approved):
woooo

PR #2706, opened 2022-05-16
ref: Try removing the extra flush call to the producer

lynnagara:
https://github.com/getsentry/snuba/pull/2700 finally worked to
stop the producer termination errors we were seeing in production
when pods were being shut down. We're no longer seeing that error.
This extra call to flush the subscriptions executor producer is probably
unnecessary.


nikhars:
The change in https://github.com/getsentry/snuba/pull/2700, is regarding flushing the querylog producer. The code being removed here is removing the flush call from the producer of the ProduceResult step. I am a little confused as to why fixing the querylog flush would allow us to remove this code.

fpacifici (approved):
@nikhars  this was introduced before Lyn found out that the issue came from the querylog producer, we thought this producer was the one causing the issue. So no need to keep this hack around.

nikhars (approved):

PR #2708, opened 2022-05-16
feat(dlq): Metrics DLQ enabled

rahul-kumar-saini:
### Overview
- Removing backup logic and enabling DLQ on metrics validation


rahul-kumar-saini:
Enabled DLQ at general Snuba Consumer level instead.

PR #2711, opened 2022-05-17
fix(dlq): Revert Produce policy to Ignore on Metrics

rahul-kumar-saini:
### Overview
- There is a bug in produce policy - we never close the producer
- Reverting to Ignore policy while we figure out how to approach closing/rebalancing producer when metrics consumer is closed/rebalanced


fpacifici (approved):
Thanks. 

Please merge and deploy as soon as possible tomorrow morning

PR #2714, opened 2022-05-17
fix(tiger errors): Use v2 query processors

nikhars:
Fix a bug in the tiger errors ro storages to use the query processors
defined by the v2 entity.

### Legal Boilerplate


volokluev:
What's the bug?

volokluev (approved):

nikhars:
> What's the bug?

The errors_v2 processors differ from the errors processor for this specific processor https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/errors_v2.py#L62-L65. With the current implementation, errors_v2 is using a different cache for applying that processor. That's the bug.

volokluev:
I see. Good find!

PR #2720, opened 2022-05-18
fix(rate_limit): Allow passing a suffix to table rate limit

nikhars:
In order to allow having separate configurations for table rate limiters
that work on the same table but are residing in different clusters,
allow passing a suffix to the TableRateLimiter. The errors tiger
query processors will use a TableRateLimiter with a suffix.

### DIFF
--- a/snuba/datasets/storages/errors_v2.py
+++ b/snuba/datasets/storages/errors_v2.py
@@ -88,7 +88,7 @@ query_processors = [
         # merged together by the final.
         omit_if_final=["environment", "release", "group_id"],
     ),
-    TableRateLimit(),
+    TableRateLimit(suffix="errors_tiger"),
 ]
 
 schema = WritableTableSchema(
--- a/snuba/query/processors/table_rate_limit.py
+++ b/snuba/query/processors/table_rate_limit.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from snuba.clickhouse.processors import QueryProcessor
 from snuba.clickhouse.query import Query
 from snuba.request.request_settings import RequestSettings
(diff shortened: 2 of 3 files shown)


volokluev (approved):

fpacifici (commented):
Hold on, could you please give more context on that? 
Are you running into issues with the pipeline delegator (which is supposed to already partition the rate limiter so that the two pipelines have a different rate limtier space? Or are you trying to build a generic solution that is unrelated to the migraiton from errors to errors_V2 ?

If it is the first, the rate limiter scope implemented in https://github.com/getsentry/snuba/pull/2448 should be enough. If it does not work let's see why. 

If it is the second, all good. the same table name could end up being on two different clusters.

nikhars:
> Hold on, could you please give more context on that? Are you running into issues with the pipeline delegator (which is supposed to already partition the rate limiter so that the two pipelines have a different rate limtier space)? Or are you trying to build a generic solution that is unrelated to the migraiton from errors to errors_V2 ?
> 
> If it is the first, the rate limiter scope implemented in #2448 should be enough. If it does not work let's see why.
> 
> If it is the second, all good. the same table name could end up being on two different clusters.

I am not building a generic solution. So it's the first use case. I will look into #2448 and see if I can override settings for TableRateLimiter using that implementation.

nikhars:
> Hold on, could you please give more context on that? Are you running into issues with the pipeline delegator (which is supposed to already partition the rate limiter so that the two pipelines have a different rate limtier space)? Or are you trying to build a generic solution that is unrelated to the migraiton from errors to errors_V2 ?
> 
> If it is the first, the rate limiter scope implemented in #2448 should be enough. If it does not work let's see why.
> 
> If it is the second, all good. the same table name could end up being on two different clusters.

@fpacifici - The current implementation of the RateLimiterDelegate separates partition space so that the counting happens in different partition space https://github.com/getsentry/snuba/blob/master/snuba/pipeline/settings_delegator.py#L31-L37. The current problem I am fixing is quota management which needs to be applied to the partition space. The quotas are not separated by partition space. They are common as shown in the code above. So we have code like https://github.com/getsentry/snuba/blob/master/snuba/query/processors/table_rate_limit.py#L16-L21 where the same quota is applied across different partitions. The current change is the simplest way to get different quotas for different partitions. The quotas need to be different only for a few specific cases.

fpacifici:
Ok, good to me then

PR #2721, opened 2022-05-18
RFC: Simplify looking up things by name

volokluev:


snuba/datasets/entities/factory.py
```diff
@@ -11,69 +11,66 @@ class InvalidEntityError(SerializableException):
...
+    registered_entity = Entity.from_name(name.value)
+    if registered_entity is not None:
+        return cast(Entity, registered_entity())
+    import_entities()
+    registered_entity = Entity.from_name(name.value)
+    if registered_entity is not None:
+        return cast(Entity, registered_entity())
 
-    return entity
+    raise InvalidEntityError(f"entity {name!r} does not exist")
```
fpacifici, line 67: I think this is still continuing on the wrong data model we have now where entities can exist outside of datasets. 
Entities were meant to exist only within a dataset as a dataset is supposed to be a sealed partition of the snuba data model 
https://getsentry.github.io/snuba/architecture/datamodel.html#datasets

Which makes me think a `get_entity` method should only exist within a dataset and not by itself, which basically would change quite a bit the problem you are trying to solve. 

Now the constraint was not built from the beginning, so today it may be not trivial to add, though what about trying to fix the data model inconsistency first? Maybe the problem would not exist anymore.

lynnagara:
This hasn't moved in some time, please reopen if still needed.

PR #2726, opened 2022-05-19
ref: set an explicit uid/gid for the snuba user

asottile-sentry:
this will be used in conjunction with `runAsUser` / `runAsGroup` in the k8s metadata so `gosu` can be factored out (originally attempted in https://github.com/getsentry/snuba/pull/2696 but that had to be rolled back)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,8 +65,8 @@ RUN set -ex; \
 # so try not to do anything heavy beyond here.
 COPY . ./
 RUN set -ex; \
-    groupadd -r snuba; \
-    useradd -r -g snuba snuba; \
+    groupadd -r snuba --gid 1000; \
+    useradd -r -g snuba --uid 1000 snuba; \
     chown -R snuba:snuba ./; \
     pip install -e .; \
     snuba --help;


volokluev:
As far as I understand, this will not change anything about the current behavior but simply add another user and group.

What does

> be used in conjunction with runAsUser / runAsGroup in the k8s metadata

mean? How does that differ from the current situation with gosu?

asottile-sentry:
> As far as I understand, this will not change anything about the current behavior but simply add another user and group.
> 
> What does
> 
> > be used in conjunction with runAsUser / runAsGroup in the k8s metadata
> 
> mean? How does that differ from the current situation with gosu?

yeah this patch is a fancy noop -- I just want to make sure that the uid/gid is constant (it's already 1000/1000 from the current image)

I will follow up to patch `ops` so that instead of invoking `gosu snuba snuba` it will use `runAsUser: 1000` + `runAsGroup: 1000` (annoying k8s doesn't let you use usernames there) and just invoke `snuba`

volokluev (approved):
Thank you!

PR #2728, opened 2022-05-19
feat(subscriptions): Remove legacy transactions subscriptions from Freight

lynnagara:
We have moved the transactions subscriptions traffic to the new process now, this one will be completely removed soon.


fpacifici (commented):
What about waiting a few days before getting rid of this? At least till when we are sure we are not going to switch back.

lynnagara:
> What about waiting a few days before getting rid of this? At least till when we are sure we are not going to switch back.

Sure, i'll merge this only after the weekend

fpacifici (changes requested):
wait though. We found out just yesterday during the incident that we need manual intervention from ops to unblock the scheduler if it accumulates backlog. 

I think we should fix that issue before going all in on the new one.

That is something that can happen for many reasons that are not P0 incidents with sentry fully down. 
If ingestion stops for a little while we could accumulate enough commit log message at the restart to fill in the scheduler buffer. It is not so hard to reproduce, if the issue is that one.

lynnagara:
@fpacifici I can't imagine any scenario where we would go back to this. Even during the incident we did not and preferred to reset offsets instead.

fpacifici:
I am talking about the incident to mention the issue we found with the recovery, not about the failover process we applied.
Without the fix in place, quickly switching to the old system, is probably a better option than resetting offsets (which is what we had to do).
Let's remove this when the fixes are merged.

lynnagara:
@fpacifici All the planned fixes are already place. Is there something else you're looking to get done first? As there is no other work planned regarding recovering from long outages currently.

fpacifici (approved):
Discussed offline with @lynnagara .
Please ensure the failover procedure should the issue we had last week happen again and ensure the on-call knows about it. Then it is ok.

PR #2732, opened 2022-05-20
feat(profiling): Add new functions dataset

Zylphrex:
To leverage the aggregation capabilities of clickhouse more in aggregating
profiles, this new dataset will contain individual function frames from the
profile as rows. For now, to avoid having to join this functions table with the
profiles table, we will denormalize the attributes from profiles to enable
filtering.

Relevant docs: https://www.notion.so/sentry/Storing-Function-Frames-in-Clickhouse-04c02e550a654d17b6f588fcb19a4dcf

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -214,6 +214,21 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
             ),
         ]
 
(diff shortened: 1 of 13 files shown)


Zylphrex:
closing in favour of an aggregated merge tree approach

PR #2733, opened 2022-05-20
fix(rate-limits): Create copy of request settings

nikhars:
The current implementation of RateLimiterDelegate uses the same
RequestSettings passed to it in the constructor. That causes a bug
since the original rate_limit_params of HTTPRequestSettings are stored
in a list. The same list gets used by both the threads processing the
query since add_rate_limit_params from both execution pipelines
adds the rate limit to the same list.

The end result of this is that instead of processing 3 sets of rate
limits (project, table, global), secondary queries are actually getting 6
sets of rate limits (all of the above multiplied by 2). This leads
to overcounting since a query run in a single execution pipeline
gets counted in both execution pipelines.

To fix this we use a deep copy of the RequestSettings.

### Legal Boilerplate

### DIFF
--- a/snuba/pipeline/settings_delegator.py
+++ b/snuba/pipeline/settings_delegator.py
@@ -1,3 +1,4 @@
+import copy
 from typing import Optional, Sequence
 
 from snuba.request.request_settings import RequestSettings
@@ -25,7 +26,7 @@ class RateLimiterDelegate(RequestSettings):
 
     def __init__(self, prefix: str, delegate: RequestSettings):
         self.referrer = delegate.referrer
-        self.__delegate = delegate
+        self.__delegate = copy.deepcopy(delegate)
         self.__prefix = prefix
 
     def __append_prefix(self, rate_limiter: RateLimitParameters) -> RateLimitParameters:
--- a/snuba/pipeline/settings_delegator.py
+++ b/snuba/pipeline/settings_delegator.py
@@ -22,6 +22,12 @@ class RateLimiterDelegate(RequestSettings):
     Specifically the PipelineDelegator provides a RateLimiterDelegate to
     to each query pipeline so that the pipeline can use a separate
     rate limiter namespace without knowing about it.
+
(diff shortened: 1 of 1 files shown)


fpacifici (approved):
Good catch

snuba/pipeline/settings_delegator.py
```diff
@@ -25,7 +26,7 @@ class RateLimiterDelegate(RequestSettings):
 
     def __init__(self, prefix: str, delegate: RequestSettings):
         self.referrer = delegate.referrer
-        self.__delegate = delegate
+        self.__delegate = copy.deepcopy(delegate)
```
volokluev, line 29: great find on the race condition. I have a few questions:

1. Is this affecting customer facing queries right now?
2. This will have a performance impact as we will be deepcopying a lot of things (including the whole parsed query). Do we need to copy the whole request or just isolate the request settings?

volokluev: At the very least, put a comment here explaining why this deepcopy is done so nobody thinks "oh this seems like I could get a great performance impact if I removed it"

nikhars: > great find on the race condition. I have a few questions:
> 
> 1. Is this affecting customer facing queries right now?
> 2. This will have a performance impact as we will be deepcopying a lot of things (including the whole parsed query). Do we need to copy the whole request or just isolate the request settings?

1. Yes it is impacting customer facing queries. But the effects have been mitigated by increasing the quotas for now.
2. This does not copy the Request object(which has the parsed query). It only deepcopies the RequestSettings object which would have fields like https://github.com/getsentry/snuba/blob/master/snuba/request/request_settings.py#L95-L106

fpacifici: Context on the user impact. 
This was most likely the root cause of the increase in rate limiting of post process queries. The spiky load theory may have been wrong, so we should get back to Dan as he does not need to introduce anything new and maybe wants to move the queries back to consistent mode.

volokluev (approved):

PR #2734, opened 2022-05-20
feat(debug): Add table rate limit metrics

nikhars:
### Legal Boilerplate

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -43,6 +43,7 @@ from snuba.state.rate_limit import (
     GLOBAL_RATE_LIMIT_NAME,
     ORGANIZATION_RATE_LIMIT_NAME,
     PROJECT_RATE_LIMIT_NAME,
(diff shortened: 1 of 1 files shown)


snuba/web/db_query.py
```diff
@@ -341,6 +342,29 @@ def execute_query_with_rate_limits(
...
+            metrics.gauge(
+                name="table_per_second",
+                value=table_rate_limit_stats.rate,
+                tags={
+                    "cache_partition": reader.cache_partition_id
+                    if reader.cache_partition_id
+                    else "default"
+                },
+            )
+
```
volokluev, line 367: this is fine but can we put it into its own function so this one doesn't get too large?

volokluev (approved):

snuba/web/db_query.py
nikhars: Done.

volokluev (approved):

PR #2739, opened 2022-05-20
ref(pagination): stop changing the limit on queries in the ColumnSplitter

volokluev:
We have hit the unlikely case that two different projects assign the same UUID to an event. Us doing this "optimization" breaks pagination for that reason. In general, we should be respecting the limit that a request gives us rather than being clever about it

### DIFF
--- a/snuba/web/split.py
+++ b/snuba/web/split.py
@@ -331,10 +331,6 @@ class ColumnSplitQueryStrategy(QuerySplitStrategy):
             )
         )
         query.set_offset(0)
-        # TODO: This is technically wrong. Event ids are unique per project, not globally.
-        # So, if the minimal query only returned the same event_id from two projects, we
-        # would be underestimating the limit here.
-        query.set_limit(len(event_ids))
 
         project_ids = list(
             set([event[self.__project_column] for event in result.result["data"]])


snuba/web/split.py
```diff
@@ -331,10 +331,6 @@ def execute(
             )
         )
         query.set_offset(0)
-        # TODO: This is technically wrong. Event ids are unique per project, not globally.
-        # So, if the minimal query only returned the same event_id from two projects, we
-        # would be underestimating the limit here.
-        query.set_limit(len(event_ids))
```
lynnagara, line 337: Should this be `query.set_limit(result.result["data"])`? Basically the same number of rows that came back from the first minimal query. If we already know how many results we're supposed to get I think it's better to set the limit and stop scanning at that point.

volokluev: LIMIT just affects how much is returned, it does not change how much is scanned

lynnagara: > LIMIT just affects how much is returned, it does not change how much is scanned

Don't think I follow what you are saying. We already know at this point exactly how many rows there are supposed to be as we already ran the minimal query and got all the ids.

So if we know there is 1 match then why don't we just set the limit of 1 so ClickHouse can stop searching when it encountered that row. Otherwise it will needlessly scan the rest of the table even though we already know from the first query there is nothing else to be found.

volokluev: https://clickhouse.com/docs/en/sql-reference/statements/select/limit#limit-clause

> LIMIT m allows to select the first m rows from the result.

If I run this query in the tracing tool

```
SELECT (title AS _snuba_title), (transform((project_id AS _snuba_project_id), [5468329], ['their-story-frontend'], '') AS _snuba_project), (coalesce(IPv4NumToString(ip_address_v4), IPv6NumToString(ip_address_v6)) AS _snuba_ip_address), (timestamp AS _snuba_timestamp), (type AS _snuba_type), (replaceAll(toString(event_id), '-', '') AS _snuba_event_id), (transform(_snuba_project_id, [5468329], ['their-story-frontend'], '') AS `_snuba_project.name`) FROM errors_dist PREWHERE in(_snuba_project_id, [5468329]) WHERE equals(deleted, 0) AND equals(_snuba_ip_address, '67.240.253.31') AND greaterOrEquals(_snuba_timestamp, toDateTime('2022-05-19T13:00:43', 'Universal')) AND less(_snuba_timestamp, toDateTime('2022-05-20T13:00:43', 'Universal')) ORDER BY _snuba_timestamp ASC LIMIT 5 OFFSET 0
```

```
[Information] executeQuery: Read 206809 rows
```


And then I run the same query with LIMIT 51
```
[Information] executeQuery: Read 206703 rows
```

volokluev: https://clickhouse.com/docs/en/sql-reference/statements/select/limit#limit-clause

> LIMIT m allows to select the first m rows from the result.

If I run this query in the tracing tool

```
SELECT (title AS _snuba_title), (transform((project_id AS _snuba_project_id), [1234], ['something'], '') AS _snuba_project), (coalesce(IPv4NumToString(ip_address_v4), IPv6NumToString(ip_address_v6)) AS _snuba_ip_address), (timestamp AS _snuba_timestamp), (type AS _snuba_type), (replaceAll(toString(event_id), '-', '') AS _snuba_event_id), (transform(_snuba_project_id, [1234], ['somethingelse'], '') AS `_snuba_project.name`) FROM errors_dist PREWHERE in(_snuba_project_id, [1234]) WHERE equals(deleted, 0) AND equals(_snuba_ip_address, '11.111.111.31') AND greaterOrEquals(_snuba_timestamp, toDateTime('2022-05-19T13:00:43', 'Universal')) AND less(_snuba_timestamp, toDateTime('2022-05-20T13:00:43', 'Universal')) ORDER BY _snuba_timestamp ASC LIMIT 5 OFFSET 0
```

```
[Information] executeQuery: Read 206809 rows
```


And then I run the same query with LIMIT 51
```
[Information] executeQuery: Read 206703 rows
```

fpacifici: > LIMIT just affects how much is returned, it does not change how much is scanned

This is not accurate, it depends on the query itself. 
Specifically what prevented clickhouse from stopping in your query was the ORDER BY `ORDER BY _snuba_timestamp`. Clickhouse needed to scan the entire table because, after 51, results it did not know if there were more that would  go before the 51 it already found. 

If the query had no ORDER BY it would have stopped after finding 51 viable results.

volokluev: Understood, updated to what Lyn said and added a test. Thanks for the explanation

lynnagara (approved):

PR #2744, opened 2022-05-23
feat(subscriptions): Record information about size of each tick buffer

lynnagara:
We experienced an incident where some partitions heavily lagged
behind others. There was not a lot of visibility about the
contents of the TickBuffer and how long it was waiting for ticks
to be able to be submitted to the next step. This is important for
datasets that run in "global" mode, as subscriptions on those datasets
wait for all partitions to reach a timestamp before they can be executed.

### DIFF
--- a/snuba/subscriptions/scheduler_processing_strategy.py
+++ b/snuba/subscriptions/scheduler_processing_strategy.py
@@ -204,6 +204,7 @@ class TickBuffer(ProcessingStrategy[Tick]):
         partitions: int,
         max_ticks_buffered_per_partition: Optional[int],
         next_step: ProcessingStrategy[Tick],
+        metrics: MetricsBackend,
     ) -> None:
         if mode == SchedulingWatermarkMode.GLOBAL:
             assert max_ticks_buffered_per_partition is not None
@@ -212,12 +213,15 @@ class TickBuffer(ProcessingStrategy[Tick]):
         self.__partitions = partitions
         self.__max_ticks_buffered_per_partition = max_ticks_buffered_per_partition
         self.__next_step = next_step
+        self.__metrics = metrics
 
         self.__buffers: Mapping[int, Deque[Message[Tick]]] = {
             index: deque() for index in range(self.__partitions)
         }
 
         self.__closed = False
+        self.__record_frequency_seconds = 180  # 3 minutes
+        self.__last_recorded_time: float = 0
 
     def poll(self) -> None:
         self.__next_step.poll()
@@ -250,6 +254,17 @@ class TickBuffer(ProcessingStrategy[Tick]):
             )
             self.__next_step.submit(self.__buffers[tick_partition].popleft())
             return
+
+        # Periodically record the legnth of each buffer
+        now = time.time()
+        if now - self.__last_recorded_time > self.__record_frequency_seconds:
+            for partition_index in self.__buffers:
+                self.__metrics.gauge(
+                    "TickBuffer.queue_size",
+                    len(self.__buffers[partition_index]),
+                    tags={"partition": str(partition_index)},
+                )
+            self.__last_recorded_time = now
 
         # If there are any empty buffers, we can't submit anything yet.
         # Otherwise if all the buffers have ticks then we look for the partition/s
--- a/tests/subscriptions/test_scheduler_processing_strategy.py
+++ b/tests/subscriptions/test_scheduler_processing_strategy.py
@@ -37,10 +37,12 @@ from tests.backends.metrics import TestingMetricsBackend
 
 def test_tick_buffer_immediate() -> None:
     epoch = datetime(1970, 1, 1)
-
     next_step = mock.Mock()
+    metrics = TestingMetricsBackend()
 
-    strategy = TickBuffer(SchedulingWatermarkMode.PARTITION, 2, None, next_step)
+    strategy = TickBuffer(
+        SchedulingWatermarkMode.PARTITION, 2, None, next_step, metrics
+    )
 
     topic = Topic("messages")
     partition = Partition(topic, 0)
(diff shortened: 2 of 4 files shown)


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -212,12 +213,15 @@ def __init__(
...
         self.__max_ticks_buffered_per_partition = max_ticks_buffered_per_partition
         self.__next_step = next_step
+        self.__metrics = metrics
 
         self.__buffers: Mapping[int, Deque[Message[Tick]]] = {
             index: deque() for index in range(self.__partitions)
         }
 
         self.__closed = False
+        self.__record_frequency_seconds = 180  # 3 minutes
```
nikhars, line 223: Why not record every minute? In Datadog, you would see the metric value change every 3 minutes with the current implementation. So you would delay the time it takes to gauge if there's a problem

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -251,6 +255,17 @@ def submit(self, message: Message[Tick]) -> None:
             self.__next_step.submit(self.__buffers[tick_partition].popleft())
             return
 
+        # Periodically record the legnth of each buffer
+        now = time.time()
+        if now - self.__last_recorded_time > self.__record_frequency_seconds:
+            for partition_index in self.__buffers:
+                self.__metrics.gauge(
+                    "TickBuffer.queue_size",
```
nikhars, line 263: DD metrics are usually all lower case and use snake case.
```suggestion
                    "tick_buffer.queue_size",
```

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -251,6 +255,17 @@ def submit(self, message: Message[Tick]) -> None:
             self.__next_step.submit(self.__buffers[tick_partition].popleft())
             return
 
+        # Periodically record the legnth of each buffer
+        now = time.time()
+        if now - self.__last_recorded_time > self.__record_frequency_seconds:
+            for partition_index in self.__buffers:
+                self.__metrics.gauge(
```
nikhars, line 262: Is the metric already partitioned by dataset or entity over here? If not, then please add it

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -251,6 +255,17 @@ def submit(self, message: Message[Tick]) -> None:
...
+                    "TickBuffer.queue_size",
+                    len(self.__buffers[partition_index]),
+                    tags={"partition": str(partition_index)},
+                )
+            self.__last_recorded_time = now
+
```
volokluev, line 268: This should be in its own function

tests/subscriptions/test_scheduler_processing_strategy.py
```diff
@@ -410,14 +408,14 @@ def test_tick_buffer_with_commit_strategy_global() -> None:
...
     metrics_backend = TestingMetricsBackend()
-
     next_step = mock.Mock()
 
     strategy = TickBuffer(
         SchedulingWatermarkMode.GLOBAL,
         2,
         10,
         ProvideCommitStrategy(2, next_step, metrics_backend),
+        metrics_backend,
```
volokluev, line 418: from looking at this test I don't see that you've actually tested the new functionality you've introduced

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Already done at the top before the metrics object is passed in here

tests/subscriptions/test_scheduler_processing_strategy.py
lynnagara: I haven't, this just passes the now-required metrics argument in. I don't think we should need to have unit tests over logging to datadog.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Updated to 60 seconds

snuba/subscriptions/scheduler_processing_strategy.py
volokluev: I'll leave it up to you of course but as an external reader I don't find this more readable. Most importantly, being able to separate the observability logic from the application logic is important. Looking at the function, it's not obvious at a glance what is doing what. 

This function also does not fit into my monitor which makes its purpose and logic harder to follow.

snuba/subscriptions/scheduler_processing_strategy.py
volokluev: Can you show me the line? I don't see that anywhere

volokluev (approved):
Approved as long as:

* the metric is actually partitioned by dataset. I don't see where that is actually done
* You consider my comments about the test and function. I'll leave whether you actually change anything up to you

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: https://github.com/getsentry/snuba/blob/e1f0764c5b1f89552a2eda57d2d36bca7c8ae0e3/snuba/cli/subscriptions_scheduler.py#L111-L113

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Ok, I will move it out

tests/subscriptions/test_scheduler_processing_strategy.py
lynnagara: We almost never have tests for stuff like this, and I'm not sure we want to start now. Unfortunately it's always the case that logging and metrics would have an outsize impact on any high throughput consumer.

PR #2746, opened 2022-05-23
fix(subscriptions): Commit the right offset in the scheduler and executor

lynnagara:
The offset to be committed should always be the *next_offset* - the minimum
offset of the next message to be processed (which is one higher than the
offset of the last successfully processed message). Previously we
were committing the offset of the current message which is one offset behind
what it should be.

### DIFF
--- a/snuba/subscriptions/executor_consumer.py
+++ b/snuba/subscriptions/executor_consumer.py
@@ -337,7 +337,7 @@ class ExecuteQuery(ProcessingStrategy[KafkaPayload]):
 
             # Periodically commit offsets if we haven't started rollout yet
             self.__commit_data[message.partition] = Position(
-                message.offset, message.timestamp
+                message.next_offset, message.timestamp
             )
 
             now = time.time()
@@ -446,7 +446,7 @@ class ProduceResult(ProcessingStrategy[SubscriptionTaskResult]):
             self.__queue.popleft()
 
             self.__commit_data[message.partition] = Position(
-                message.offset, message.timestamp
+                message.next_offset, message.timestamp
             )
         self.__throttled_commit()
 
--- a/snuba/subscriptions/scheduler_processing_strategy.py
+++ b/snuba/subscriptions/scheduler_processing_strategy.py
@@ -133,7 +133,9 @@ class ProvideCommitStrategy(ProcessingStrategy[Tick]):
 
         # Keep track of the earliest partition based on its offset in the commit log topic.
         # This is used to determine the offset that we can safely commit.
-        earliest = message.offset
+        # The offset to be committed is the `next_offset`: 1 higher than the offset of
+        # the current message
+        earliest = message.next_offset
 
         for partition_message in self.__latest_messages_by_partition.values():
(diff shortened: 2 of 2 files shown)


snuba/subscriptions/executor_consumer.py
```diff
@@ -446,7 +446,7 @@ def poll(self) -> None:
             self.__queue.popleft()
 
             self.__commit_data[message.partition] = Position(
-                message.offset, message.timestamp
+                message.next_offset, message.timestamp
```
volokluev, line 449: Some questions:

1. What is the impact of this being incorrect right now?

2. This `next_offset` behavior appears in three separate places. Is there a way to make it so that it can't be misused? This seems like special behavior and its knowledge is duplicated. At the very least, I'd like there to be a comment in each one explaining at each place why this is happening because this is not intuitive

3. How and where is `next_offset` calculated? is it just `message.offset + 1`?

volokluev (commented):
The change is fine to me but there should be a comment explaining

snuba/subscriptions/executor_consumer.py
```diff
@@ -337,7 +337,7 @@ def submit(self, message: Message[KafkaPayload]) -> None:
 
             # Periodically commit offsets if we haven't started rollout yet
             self.__commit_data[message.partition] = Position(
-                message.offset, message.timestamp
+                message.next_offset, message.timestamp
```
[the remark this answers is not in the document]
lynnagara: Not really, but I can address that in a follow up. I think it's unrelated with the fact that we were always committing one offset behind.

snuba/subscriptions/executor_consumer.py
lynnagara: 1. We are always committing one offset behind what we should be. In practice it means when we rebalance / restart consumers we are likely to be double processing the last processed message. Subscriptions is an at least once system so it is considered harmless that we run it twice.
2. It appears in 3 different places because I was unaware that the correct offset to be committed is the next one and was committing the current one everywhere.
3.  Yes it is `message.offset + 1`. Somewhat related: https://github.com/getsentry/rust_arroyo/pull/18 and https://github.com/getsentry/arroyo/pull/73

snuba/subscriptions/executor_consumer.py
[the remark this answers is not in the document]
lynnagara: This is documented in https://github.com/getsentry/arroyo/blob/main/arroyo/backends/abstract.py#L14-L49

snuba/subscriptions/executor_consumer.py
lynnagara: This is documented in https://github.com/getsentry/arroyo/blob/main/arroyo/backends/abstract.py#L14-L49

volokluev: Ah I see, so `next_offset` what is always meant to be committed to kafka, not just for subscriptions. That was where I was confused.

volokluev (approved):

PR #2747, opened 2022-05-23
feat(datasets): allow marking datasets as experimental

volokluev:
Currently, if we deploy a misconfigured dataset, deploys will be blocked. This change allows marking a dataset as experimental, which will allow it to bypass the healthcheck and not block any deploys. 

# Rollout plan
* Merge this PR
* Open up the revert PR, wait until it passes CI
* deploy master, see that the deploy goes through
* Re-Deploy old commit if this is not sufficient to unblock deploys
* Monitor snuba API dashboard, sentry for errors
* Either close the revert PR or merge and deploy it


snuba/clusters/cluster.py
```diff
@@ -374,16 +375,49 @@ def __get_cluster_nodes(self, cluster_name: str) -> Sequence[ClickhouseNode]:
...
+        storage_set_key,
+        UndefinedClickhouseCluster(
+            host="undefined_clickhouse_cluster",
+            port=420,
+            user="filippo",
+            password="p455w0rd",
+            database="nonya",
+            http_port=69,
+            storage_sets={storage_set_key.value},
+            single_node=False,
```
fpacifici, line 420: ??

volokluev: Does the docstring not explain it?

fpacifici (commented):
Do we need bot the experimental dataset concept and the dev dataset concept?
It seems to me the experimental should make the dev entity/storage/etc. obsolete

volokluev:
> It seems to me the experimental should make the dev entity/storage/etc. obsolete

I would agree, I can add a task to remove that

snuba/utils/streams/topics.py
```diff
@@ -30,6 +30,7 @@ class Topic(Enum):
     DEAD_LETTER_QUEUE_INSERTS = "snuba-dead-letter-inserts"
     DEAD_LETTER_METRICS = "snuba-dead-letter-metrics"
     DEAD_LETTER_SESSIONS = "snuba-dead-letter-sessions"
+    EXPERIMENTAL = "does-not-exist"
```
lynnagara, line 33: This will be automatically created in self hosted and single tenant. Can we find a way to avoid that?

volokluev: as in an experimental kafka topic will automatically be created? Where is that done?

lynnagara:
Are some elements of this PR (e.g. the actual fake datasets and storages) but not others (the parts that allow experimental flags) supposed to be reverted and not others? It's unclear to me what's supposed to be permanent and what isn't.

snuba/clusters/storage_sets.py
```diff
@@ -31,6 +31,7 @@ class StorageSetKey(Enum):
     ERRORS_V2 = "errors_v2"
     ERRORS_V2_RO = "errors_v2_ro"
     PROFILES = "profiles"
+    EXPERIMENTAL = "experimental"
```
volokluev, line 34: will be removed

snuba/clusters/cluster.py
lynnagara: Why can't we just return None, instead of this weird object?

snuba/datasets/entities/experimental.py
```diff
@@ -0,0 +1,41 @@
+from abc import ABC
```
volokluev, line 1: this whole file will be removed

snuba/datasets/entities/__init__.py
```diff
@@ -24,3 +24,4 @@ class EntityKey(Enum):
     DISCOVER_TRANSACTIONS = "discover_transactions"
     DISCOVER_EVENTS = "discover_events"
     PROFILES = "profiles"
+    EXPERIMENTAL = "experimental"
```
volokluev, line 27: this will be removed

snuba/datasets/entities/factory.py
```diff
@@ -27,6 +27,7 @@ def get_entity(name: EntityKey) -> Entity:
         DiscoverTransactionsEntity,
     )
     from snuba.datasets.entities.events import EventsEntity
+    from snuba.datasets.entities.experimental import ExperimentalEntity
```
volokluev, line 30: will be removed

snuba/datasets/entities/factory.py
```diff
@@ -58,6 +59,7 @@ def get_entity(name: EntityKey) -> Entity:
         EntityKey.ORG_METRICS_COUNTERS: OrgMetricsCountersEntity,
         EntityKey.METRICS_DISTRIBUTIONS: MetricsDistributionsEntity,
         EntityKey.PROFILES: ProfilesEntity,
+        EntityKey.EXPERIMENTAL: ExperimentalEntity,
```
volokluev, line 62: will be removed

snuba/datasets/experimental.py
```diff
@@ -0,0 +1,11 @@
+from snuba.datasets.dataset import Dataset
```
volokluev, line 1: will be removed

snuba/datasets/factory.py
```diff
@@ -21,6 +21,7 @@
     "sessions",
     "transactions",
     "profiles",
+    "experimental",
```
volokluev, line 24: will be removed

snuba/datasets/factory.py
```diff
@@ -44,6 +45,7 @@ def get_dataset(name: str) -> Dataset:
     from snuba.datasets.cdc.groupedmessage import GroupedMessageDataset
     from snuba.datasets.discover import DiscoverDataset
     from snuba.datasets.events import EventsDataset
+    from snuba.datasets.experimental import ExperimentalDataset
```
volokluev, line 48: will be removed

snuba/datasets/factory.py
```diff
@@ -62,6 +64,7 @@ def get_dataset(name: str) -> Dataset:
         "sessions": SessionsDataset,
         "transactions": TransactionsDataset,
         "profiles": ProfilesDataset,
+        "experimental": ExperimentalDataset,
```
volokluev, line 67: will be removed

snuba/datasets/storages/__init__.py
```diff
@@ -32,6 +32,7 @@ class StorageKey(Enum):
     ERRORS_V2 = "errors_v2"
     PROFILES = "profiles"
     ERRORS_V2_RO = "errors_v2_ro"
+    EXPERIMENTAL = "experimental"
```
volokluev, line 35: will be removed

snuba/utils/streams/topics.py
lynnagara: No, everything on this list does. It doesn't know because experimental is not marked on this topic, that's only a property of the dataset. It's done by the bootstrap script.

snuba/datasets/storages/experimental.py
```diff
@@ -0,0 +1,44 @@
+from snuba.clickhouse.columns import UUID, ColumnSet, UInt
```
volokluev, line 1: will be removed

volokluev: (whole file)

snuba/datasets/storages/factory.py
```diff
@@ -9,6 +9,9 @@
 from snuba.datasets.storages.errors_ro import storage as errors_ro_storage
 from snuba.datasets.storages.errors_v2 import storage as errors_v2_storage
 from snuba.datasets.storages.errors_v2_ro import storage as errors_v2_ro_storage
+from snuba.datasets.storages.experimental import (
+    writable_storage as experimental_storage,
+)
```
volokluev, line 14: will be removed

snuba/datasets/storages/factory.py
```diff
@@ -74,6 +77,7 @@
             transactions_v2_storage,
             errors_v2_storage,
             profiles_writable_storage,
+            experimental_storage,
```
volokluev, line 80: will be removed

volokluev (commented):
Adding comments on the parts that will be removed after the test is complete

lynnagara:
@volokluev May I suggest splitting the PR into 2 instead of the inline comments about what will be removed? In addition to being easier to review, it'll allow you to just do a straight revert of a single commit afterwards.

snuba/clusters/cluster.py
fpacifici: My initial quick take was about the hardcoded values, which I would strongly discourage, as somebody getting that result would try to make sense of. 

In design term, agree with Lyn. If you want to avoid the exception, returning an invalid value seems not to solve the problem and does not seem to be easier to manage than None. At least with None, the callsite knows they have to do something special for that case. The type checker will tell them.

snuba/utils/streams/topics.py
lynnagara: One option for a quick test would be to just point your storage at some other real topic already present here. It should be harmless anyway since it doesn't produce or consume anything from there.

snuba/clusters/cluster.py
```diff
@@ -374,16 +375,49 @@ def __get_cluster_nodes(self, cluster_name: str) -> Sequence[ClickhouseNode]:
     if (s not in DEV_STORAGE_SETS or settings.ENABLE_DEV_FEATURES)
 }
 
-# Map all storages to clusters via storage sets
-_STORAGE_SET_CLUSTER_MAP = {
-    storage_set: cluster
-    for cluster in CLUSTERS
-    for storage_set in cluster.get_storage_set_keys()
-}
+_STORAGE_SET_CLUSTER_MAP: Optional[Dict[StorageSetKey, ClickhouseCluster]] = None
```
fpacifici, line 378: Any reason on why a MutableMapping would not achieve the same result ? At least you would keep the type more abstract.

volokluev:
good point, that will make things easier

fpacifici (commented):
A high level question. 
What should happen if you try to query an experimental dataset (which is basically the main way to use a dataset at runtime), but such dataset is not working properly and failing the health check ?

I think there are two types of failures to deal with:
- initialization (like the dataset is not properly defined/configured and basically just broken in a way that is not transient). In this way snuba should immediately reject the query, maybe the factories should even refuse to return the entity/dataset. No need to go through the pipeline and fail at potentially any point.
- transient issue (like I shut down the cluster), the data structures and config are valid, but we cannot effectively run a query and the health check does not pass (but could come back in a minute).  I feel in this case we should disable the dataset (and reject queries) when the health check does not pass and reenable it when it passes again. 

I am not sure which parts of the problem you are solving here and if some are meant to be addressed in a future PR.

PR #2760, opened 2022-05-24
chore(admin): Upgrade frontend dependencies

lynnagara:
The biggest change here is going from React 17 -> React 18


volokluev:
Does admin still run locally?

lynnagara:
> Does admin still run locally?

Yeah, did a quick navigate to each of the pages. Seems ok.

volokluev (approved):

PR #2761, opened 2022-05-24
ref(subscriptions): Enable executors to handle max_concurrent queries…

MeredithAnya:


lynnagara:
Is there an accompanying Arroyo PR for this?

snuba/subscriptions/executor_consumer.py
```diff
@@ -195,10 +199,26 @@ def __init__(
         # Commit is only passed here because we are temporarily
         # skipping executions during the transition phase.
         commit: Callable[[Mapping[Partition, Position]], None],
+        partitions: Mapping[Partition, int],
     ) -> None:
         self.__dataset = dataset
         self.__entity_names = set(entity_names)
+        # should be passed in via cli
+        self.__total_concurrent_queries = 64
```
MeredithAnya, line 207: https://github.com/getsentry/snuba/pull/2815

snuba/subscriptions/executor_consumer.py
```diff
@@ -200,11 +204,24 @@ def __init__(
         # Commit is only passed here because we are temporarily
         # skipping executions during the transition phase.
         commit: Callable[[Mapping[Partition, Position]], None],
+        partitions: Mapping[Partition, int],
     ) -> None:
         self.__dataset = dataset
         self.__entity_names = set(entity_names)
+
+        if state.get_config("use_calculated_max_concurrent_queries", False):
+            calculated_max_concurrent_queries = (
```
evanh, line 213: A comment explaining this equation (why 64?) would be good.

evanh (approved):

snuba/subscriptions/executor_consumer.py
fpacifici: If 64 is meant to represent the total number of partition for the topic, that will be a problem. 
Not all topics have the same number of partitions. The errors executor topic has 32 for example, but smaller ones may have less and the errors one is likely going to be increased. 
So that cannot be hardcoded.

The easy way is to add it in settings per executor. It is not great because we would have another manual operation to fix the config after adding partitions, which makes things less autonomous.

The right way is to use the Kafka admin api to get the current number of partitions for the consumer when you subscribe to a topic.

fpacifici (commented):
Please address the comment about the hardcoded `64` before merging.

MeredithAnya:
Updates: 
* made the flag `use_calculated_max_concurrent_queries` determine whether we actually use the calculated value (not whether we log it or not)
* validate the `partition_count`. However, this will still error if the `partition_count` is 0 regardless of whether we use the calculated value or not. I could make the default to log a message error instead of raise tho
* added tests (and kind of rearranged them to be easier to read imo)

snuba/subscriptions/executor_consumer.py
```diff
@@ -202,11 +204,24 @@ def __init__(
         # Commit is only passed here because we are temporarily
         # skipping executions during the transition phase.
         commit: Callable[[Mapping[Partition, Position]], None],
+        partitions: Mapping[Partition, int],
     ) -> None:
         self.__dataset = dataset
         self.__entity_names = set(entity_names)
+
+        if state.get_config("use_calculated_max_concurrent_queries", False):
```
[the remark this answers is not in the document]
MeredithAnya: moved this to actually gate whether we use the calculated value or not

snuba/subscriptions/executor_consumer.py
MeredithAnya: addressed in https://github.com/getsentry/snuba/pull/2888

snuba/subscriptions/executor_consumer.py
```diff
@@ -247,6 +277,34 @@ def __init__(
...
         )
 
+    @property
+    def max_concurrent_queries(self) -> int:
+        return self.__max_concurrent_queries
+
+    def __calculate_max_concurrent_queries(
+        self,
+        partitions: Mapping[Partition, int],
+        partition_count: int,
```
lynnagara, line 287: We are dealing with a lot of different partition counts here, `total_partition_count` would be clearer

lynnagara:
In https://github.com/getsentry/snuba/pull/2888#issuecomment-1170195125 you mentioned that this PR is no longer needed as it was superceded by the other one. What's changed since then?

snuba/subscriptions/executor_consumer.py
```diff
@@ -186,15 +193,18 @@ def create_with_partitions(
...
             self.__dataset,
             self.__entity_names,
             self.__max_concurrent_queries,
             self.__total_concurrent_queries,
+            self.__partition_count,
             self.__stale_threshold_seconds,
             self.__metrics,
             ProduceResult(self.__producer, self.__result_topic, commit),
             commit,
+            partitions,
```
lynnagara, line 207: partition_count and partitions are only being used to calculate the total_concurrent queries. Yet we are passing `total_concurrent_queries`, `partition_count` AND `partitions` through. Shouldn't we just be replacing the `total_concurrent_queries` and leaving this argument list unchanged?

snuba/subscriptions/executor_consumer.py
```diff
@@ -247,6 +277,34 @@ def __init__(
...
+        if partition_count <= 0:
+            raise InvalidPartitionCount(
+                f"Partition count {partition_count} must be a non-zero positive integer."
+            )
+        if partition_count < len(partitions):
+            raise InvalidPartitionCount(
+                f"Partition count {partition_count} cannot be lower than partitions received {len(partitions)}."
+            )
+
+        return int(total_concurrent_queries / (partition_count / len(partitions))) or 1
```
lynnagara, line 306: I think it is safer to round up here rather than down. Otherwise we would be potentially unnecessarily returning an error when ClickHouse would be fine with it. If we mistakenly let too many queries through, ClickHouse will error but then either way would have been an error anyway. You also don't need the `or 1` hack to avoid it being 0 if you round up.

snuba/subscriptions/executor_consumer.py
```diff
@@ -210,18 +220,38 @@ def __init__(
...
+
+        calculated_max_concurrent_queries = self.__calculate_max_concurrent_queries(
+            partitions, partition_count, total_concurrent_queries, use_calculated_value
+        )
+        # XXX(meredith): temporarily log both the calculated and passed in
+        # max concurrent queries
+        metrics.gauge(
+            "calculated_max_concurrent_queries", calculated_max_concurrent_queries
+        )
+        metrics.gauge("max_concurrent_queries", max_concurrent_queries)
```
lynnagara, line 248: Does this need to be sampled?

snuba/subscriptions/executor_consumer.py
```diff
@@ -210,18 +220,38 @@ def __init__(
...
         commit: Callable[[Mapping[Partition, Position]], None],
+        partitions: Mapping[Partition, int],
     ) -> None:
         self.__dataset = dataset
         self.__entity_names = set(entity_names)
-        self.__max_concurrent_queries = max_concurrent_queries
-        self.__total_concurrent_queries = total_concurrent_queries
+
+        use_calculated_value: bool = (
+            state.get_config("use_calculated_max_concurrent_queries", False) == True
```
lynnagara, line 237: How does actually work? AFAIK our config does not support boolean types

snuba/subscriptions/executor_consumer.py
```diff
@@ -247,6 +277,34 @@ def __init__(
...
+            return 1
+
+        if partition_count <= 0:
+            raise InvalidPartitionCount(
+                f"Partition count {partition_count} must be a non-zero positive integer."
+            )
+        if partition_count < len(partitions):
+            raise InvalidPartitionCount(
+                f"Partition count {partition_count} cannot be lower than partitions received {len(partitions)}."
+            )
```
lynnagara, line 304: You already checked for this on line 292

snuba/subscriptions/executor_consumer.py
```diff
@@ -247,6 +277,34 @@ def __init__(
...
+        if (
+            partition_count <= 0 or partition_count < len(partitions)
+        ) and not use_calculated_value:
+            logger.info(f"Partition count {partition_count} wouldve been invalid")
+            return 1
+
+        if partition_count <= 0:
+            raise InvalidPartitionCount(
+                f"Partition count {partition_count} must be a non-zero positive integer."
+            )
```
lynnagara, line 300: Also checked on line 292

snuba/subscriptions/executor_consumer.py
MeredithAnya: ah okays I can change this to use `1` or `0` then

snuba/subscriptions/executor_consumer.py
MeredithAnya: I didn't think it did because this is only on initialization, should it be sampled?

lynnagara: Ah so it is. No definitely not required then.

snuba/subscriptions/executor_consumer.py
MeredithAnya: these are not calculating `total_concurrent_queries` they are calculating `max_concurrent_queries`. We could pass in something like `calculated_max_concurrent_queries` or something like that instead if that's what you mean

snuba/subscriptions/executor_consumer.py
MeredithAnya: okay i can use `math.ciel` on this to round up. You are also saying then to forget this check:

 `partition_count < len(partitions)` 

right?

lynnagara (commented):
Oops, wrote these comments a couple of weeks ago and forgot to hit submit. Ignore if outdated.

tests/subscriptions/test_executor_consumer.py
```diff
@@ -490,3 +455,94 @@ def test_skip_stale_message() -> None:
...
+    "partitions, partition_count, total_concurrent_queries, expect_max_concurrent_queries",
+    max_concurrent_queries_tests,
+)
+def test_max_concurrent_queries(
+    partitions: Mapping[Partition, int],
+    partition_count: int,
+    total_concurrent_queries: int,
+    expect_max_concurrent_queries: int,
+) -> None:
+    state.set_config("use_calculated_max_concurrent_queries", True)
```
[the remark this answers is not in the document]
MeredithAnya: So after looking at the code

https://github.com/getsentry/snuba/blob/512e892741d8ea602d8042d6ea401a62eb22fce1/snuba/state/__init__.py#L133-L135

I learned today that `bool` is a subclass of `int` so `True` is an instance of `int`. So the assert line passes and we return `1`.

I think we use `state.set_config("config_str", True)` in a fair amount of tests so maybe we should be changing that and then can handle the boolean config correctly

snuba/subscriptions/executor_consumer.py
```diff
@@ -108,16 +109,16 @@ def get_topics_for_entity(
     )
 
     try:
-        partition_count = get_partition_count(scheduled_topic_spec.topic)
+        total_partition_count = get_partition_count(scheduled_topic_spec.topic)
     except Exception:
         logger.error("partition count unavailable..", exc_info=True)
-        partition_count = 0
+        total_partition_count = 0
```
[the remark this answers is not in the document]
MeredithAnya: changed to not catch the exception and let the consumer crash if we don't get the partition count (instead of providing a default in that case)

snuba/subscriptions/combined_scheduler_executor.py
```diff
@@ -180,6 +180,8 @@ def __init__(
         self.__executor_factory = SubscriptionExecutorProcessingFactory(
             max_concurrent_queries,
             total_concurrent_queries,
+            # total_partition_count should always be 1 when using the combined executor/scheduler
+            1,
```
lynnagara, line 184: I don't think this is a safe assumption. See https://github.com/getsentry/snuba/issues/2884

snuba/subscriptions/executor_consumer.py
```diff
@@ -46,6 +47,29 @@
 COMMIT_FREQUENCY_SEC = 1
 
 
+def calculate_max_concurrent_queries(
+    curr_partition_count: int,
```
lynnagara, line 51: suggestion `assigned_partition_count` - curr is a bit ambiguous about how it differs from total

snuba/subscriptions/executor_consumer.py
```diff
@@ -46,6 +47,29 @@
...
+    We use the total_partition_count and curr_partition_count to estimate
+    how many total replicas there are, which in turn lets us calc the
+    max_concurrent_queries for the given replica.
+
+    Round up since .5 of a replica doesnt really
+    make sense. total_partition_count could be 0 though if something
+    went wrong trying to fetch that from the kafka admin. In that case we
+    fall back to 1 replica (meaning the max concurrent queries == the total)
+    """
+    replicas = math.ceil(total_partition_count / curr_partition_count) or 1
```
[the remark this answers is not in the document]
lynnagara: Can we drop the `or 1` now? total_partition_count should never be 0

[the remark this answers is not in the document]
MeredithAnya: well it's rounding up to the nearest replica (if the `total_partition_count` was somehow lower than the `assigned_partition_count`). Which yeah ultimately would make the result lower. I was trying to enforce that the highest value `max_concurrent_queries` could be would be the same value as `total_concurrent_queries`.

 I think we kind of discussed earlier that it is a very unlikely case, maybe almost impossible, that the `total_partition_count` would be lower than `assigned_partition_count`, so I can probs jsut remove the `math.ceil` from line 69

lynnagara (approved):

PR #2766, opened 2022-05-25
feat(subscriptions): Record librdkafka buffer size in scheduler

lynnagara:
We record this metric for most of our other consumers, we should bring it
over to the scheduler as well as it was a blind spot for us during
the outage when the scheduler got stuck.


MeredithAnya (approved):
noice

PR #2771, opened 2022-05-25
feat(consumer): Snuba Consumer exceptions converted to `InvalidMessages` 

rahul-kumar-saini:
### Overview
- The DLQ prod test I'd like to do has certain risks which could result in a Metrics consumer going down
- Added some mitigations for failure modes to fool proof the test
- Tested these mitigations locally

### Note
- This should enable me to run the DLQ prod test


snuba/consumers/consumer.py
```diff
@@ -601,15 +605,69 @@ class MultistorageKafkaPayload(NamedTuple):
...
+    if (
+        f"{message.partition.topic.name}:{message.partition.index}:{message.offset}"
+        in messages_to_skip
+    ):
+        logger.warning(
+            f"A consumer for {message.partition.topic.name} skipped a message!",
+            exc_info=True,
+            extra=__message_to_dict(message),
+        )
+        return None
```
volokluev, line 638: you should write a test to make sure this is parsing the way you expect it to parse

rahul-kumar-saini: ✅

snuba/consumers/consumer.py
```diff
@@ -601,15 +605,75 @@ class MultistorageKafkaPayload(NamedTuple):
...
+    # Mitigation: JSON fails to deserialize
+    try:
+        deserialized_message = rapidjson.loads(message.payload.value)
+    except Exception as err:
+        logger.error(
+            err,
+            exc_info=True,
+            extra=__message_to_dict(message),
+        )
+        return None
```
fpacifici, line 655: We cannot do this. This logic applies to all consumer no matter whether the DLQ is configured. This means that the errors consumer would drop invalid json messages, while it is not supposed to have a DLQ.

snuba/consumers/consumer.py
```diff
@@ -601,15 +605,75 @@ class MultistorageKafkaPayload(NamedTuple):
...
+        result = processor.process_message(
+            deserialized_message,
+            KafkaMessageMetadata(
+                message.offset, message.partition.index, message.timestamp
+            ),
+        )
+    except Exception as err:
+        if (
+            message.partition.topic.name == StreamsTopic.METRICS.value
+            and not isinstance(err, InvalidMessages)
```
fpacifici, line 668: I don't think this really solves the problem. 
While you would not make the metrics consumer crash, you would not send errors to the DLQ that are not of the 2-3 specific types that the metrics consumer turn into InvalidMessage.

It seems to me that this could be much simpler by just catching the exception coming from process_message, turning it into an `InvalidMessage` and raising it again. Also remove the specific logic from the metrics processor.

fpacifici (changes requested):
I think there may be a misunderstanding here. 
I thought you were actually planning to complete the DLQ implementation for the metrics consumer (and snuba consumer in general) by ensuring that all errors raised by a snuba consumer during the  parsing and processing phase would be raised as InvalidMessage so that the consumer would send them to the DLQ if that is configured.

This seems to add a hack to allow the production test instead. I would discourage this approach as the test would not tell us that much, we would still have to do it again once the consumer is properly supporting the DLQ.
Also I suspect that ensuring all errors are turned in invalid messages would be simpler to implement than this version. It seems that you could just catch any error in `process_message`, turn it into `InvalidMessage` and raise again. If the DLQ is configured it will take care of it, if it is not, it will  crash the consumer as expected.

snuba/consumers/consumer.py
rahul-kumar-saini: Added a check for only metrics messages to be dropped.

rahul-kumar-saini:
> I thought you were actually planning to complete the DLQ implementation for the metrics consumer (and snuba consumer in general) by ensuring that all errors raised by a snuba consumer during the parsing and processing phase would be raised as InvalidMessage so that the consumer would send them to the DLQ if that is configured.

I have been under the assumption that the usage of the DLQ is up to whoever is implementing the consumer, ie the user decides where and where not to raise the DLQ specific exception. That is what I was trying to emulate with the modifications I made to the metrics consumer.

A DLQ for snuba consumer in general is a great idea, but wasn't scoped for this project as far as I know.

> This seems to add a hack to allow the production test instead. I would discourage this approach as the test would not tell us that much, we would still have to do it again once the consumer is properly supporting the DLQ.

Just allowing the prod test was the intention, I did not want to prescribe how the DLQ should be integrated into metrics. I still think it would be useful to run the test this way but sure, we can run it after making a "proper" version of the DLQ for metrics.

> It seems that you could just catch any error in `process_message`, turn it into InvalidMessage and raise again

This makes sense, but will require changing the `InvalidMessages` model to include the original exception caught as well if we want to see that info in the message produced to dead-letter topic. For now I can add the Exception args as the `reason` field for the DLQ exception.

rahul-kumar-saini:
> I thought you were actually planning to complete the DLQ implementation for the metrics consumer (and snuba consumer in general) by ensuring that all errors raised by a snuba consumer during the parsing and processing phase would be raised as InvalidMessage so that the consumer would send them to the DLQ if that is configured.

I have been under the assumption that the usage of the DLQ is up to whoever is implementing the consumer, ie the user decides where and where not to raise the DLQ specific exception. That is what I was trying to emulate with the modifications I made to the metrics consumer.

A DLQ for snuba consumer in general is a great idea, but wasn't scoped for this project as far as I know.

> This seems to add a hack to allow the production test instead. I would discourage this approach as the test would not tell us that much, we would still have to do it again once the consumer is properly supporting the DLQ.

Just allowing the prod test was the intention, I did not want to prescribe how the DLQ should be integrated into metrics. I still think it would be useful to run the test this way but sure, we can run it after making a "proper" version of the DLQ for metrics.

> It seems that you could just catch any error in `process_message`, turn it into InvalidMessage and raise again

This makes sense, but will require changing the `InvalidMessages` model to include the original exception caught as well if we want to see that info in the message produced to dead-letter topic. For now I can add the Exception args as the `reason` field for the DLQ exception.

fpacifici:
> This makes sense, but will require changing the InvalidMessages model to include the original exception caught as well if we want to see that info in the message produced to dead-letter topic. For now I can add the Exception args as the reason field for the DLQ exception.

You should not need to change anything in `InvalidMessages` just use python exception chaining.
https://docs.python.org/3/tutorial/errors.html#exception-chaining
```
catch Exception as e
   raise InvalidMessages from e
```

> I have been under the assumption that the usage of the DLQ is up to whoever is implementing the consumer, ie the user decides where and where not to raise the DLQ specific exception. That is what I was trying to emulate with the modifications I made to the metrics consumer.

That is true in general, my point is about snuba consumers, not all consumers. Snuba consumers can share the DLQ logic as they do the same thing: transform a Kafka message into a Clickhouse row. All errors in that process are equivalent. 

> A DLQ for snuba consumer in general is a great idea, but wasn't scoped for this project as far as I know.

I don't think it is very useful to discuss whether it was in plan or not as implementing the general DLQ for snuba consumer is simpler and should take shorter than adding this specific experiment logic.

rahul-kumar-saini:
@fpacifici 
(I guess I should've mentioned I've made the requested changes in the commits following my previous comment - code should be reviewed again)

> You should not need to change anything in `InvalidMessages` just use python exception chaining.
https://docs.python.org/3/tutorial/errors.html#exception-chaining

Yeah I've done this and it will show up nicely in Sentry, but I was talking about the message we produce to the DLQ topic. Should it not contain the "reason" info as well? Otherwise the messages present in the DLQ will not have much context.

snuba/consumers/consumer.py
```diff
@@ -601,15 +608,92 @@ class MultistorageKafkaPayload(NamedTuple):
...
+                        topic=message.partition.topic.name,
+                        consumer_group=consumer_group,
+                        partition=message.partition.index,
+                        offset=message.offset,
+                        headers=message.payload.headers,
+                        key=message.payload.key,
+                        reason=str(err.args),
+                    )
+                ]
+            ) from err
```
rahul-kumar-saini, line 696: This code block will remain in the error handling of `process_message`.

Adding `logger.error()` like I have in line 681 ensures that the exception is logged to Sentry (in case the DLQ handles it), but it doesn't really log the real exception which will be `InvalidMessages`.

I think ff there isn't a DLQ to handle it this would be double logged (once by logger.error() and once by raising the unhandled exception itself). Maybe the DLQ try/except itself should be logging to Sentry instead?

fpacifici:
> Should it not contain the "reason" info as well? Otherwise the messages present in the DLQ will not have much context.

If they do not contain a lot of details, it is not a big deal for as long as the message is there.

snuba/consumers/consumer.py
```diff
@@ -601,15 +608,92 @@ class MultistorageKafkaPayload(NamedTuple):
...
+    try:
+        deserialized_message = rapidjson.loads(message.payload.value)
+    except Exception as err:
+        if message.partition.topic.name == StreamsTopic.METRICS.value:
+            logger.error(
+                err,
+                exc_info=True,
+                extra=__message_to_dict(message),
+            )
+            return None
```
fpacifici, line 658: Why this ?
I think you can remove this part and just  raise the `InvalidMessages` error like for exceptions raised by the process_message.\

In the end you need to catch invalid json errors and producing such messages on the DLQ for all the storages that have a DLQ configured, not just for metrics.

snuba/consumers/consumer.py
```diff
@@ -601,15 +608,92 @@ class MultistorageKafkaPayload(NamedTuple):
...
+
+
+def skip_kafka_message(message: Message[KafkaPayload]) -> bool:
+    # TODO: Remove temporary mitigations for DLQ prod test
+
+    # expected format is "[topic:partition_index:offset,...]" eg [snuba-metrics:0:1,snuba-metrics:0:3]
+    messages_to_skip = (get_config("kafka_messages_to_skip") or "[]")[1:-1].split(",")
+    return (
+        f"{message.partition.topic.name}:{message.partition.index}:{message.offset}"
+        in messages_to_skip
```
fpacifici, line 630: I am not sure this will be very useful now that you are catching all exceptions. But if you want to preserve this feature, please identify the message with `partition:offset`, not with the full payload. It will be much easier to manage in terms of config.

snuba/consumers/consumer.py
```diff
@@ -601,15 +608,92 @@ class MultistorageKafkaPayload(NamedTuple):
...
+        if (
+            message.partition.topic.name == StreamsTopic.METRICS.value
+            and not get_config("enable_metrics_dlq", False)
+        ):
+            logger.error(
+                err,
+                exc_info=True,
+                extra={"cause": "Caught an exception on Metrics! (DLQ was disabled)"},
+            )
+            return None
```
fpacifici, line 679: Please remove this right away. It is not needed if you are catching all errors and sending them to the DLQ

fpacifici (commented):
I think this can be simplified further. Please see comments inline

snuba/consumers/consumer.py
rahul-kumar-saini: https://github.com/getsentry/snuba/pull/2771#discussion_r882201383

> We cannot do this. This logic applies to all consumer no matter whether the DLQ is configured. This means that the errors consumer would drop invalid json messages, while it is not supposed to have a DLQ.

Ok I read this as "only do this for Metrics", I've swapped it for a raise in general now.

snuba/consumers/consumer.py
```diff
@@ -601,15 +607,72 @@ class MultistorageKafkaPayload(NamedTuple):
...
+            deserialized_message,
+            KafkaMessageMetadata(
+                message.offset, message.partition.index, message.timestamp
+            ),
+        )
+    except Exception as err:
+        logger.error(err, exc_info=True)
+        raise InvalidMessages(
+            [__invalid_kafka_message(message, consumer_group, err)]
+        ) from err
```
fpacifici, line 675: You do not need two separate try catch blocks.
```
try:
        deserialized_message = rapidjson.loads(message.payload.value)
        result = processor.process_message(
            deserialized_message,
            KafkaMessageMetadata(
                message.offset, message.partition.index, message.timestamp
            ),
        )
except Exception as err:
        logger.error(err, exc_info=True)
        raise InvalidMessages(
            [__invalid_kafka_message(message, consumer_group, err)]
        ) from err
```

fpacifici (approved):

PR #2772, opened 2022-05-25
feat(datasets): Experimental datasets functionality

volokluev:
Splitting up #2747 into two separate PRs. This introduces the functionality to create experimental datasets but does not introduce them. This PR solves the following issues:

1. When an experimental dataset is improperly configured is defined in the snuba codebase, it does not fail the healthcheck (because the dataset is skipped from the healthcheck). This makes it not block deploys. 
2. When a cluster for a dataset is not defined in the settings for that particular environment, the queries will only fail when the cluster is accessed rather than at module import time


snuba/admin/clickhouse/nodes.py
```diff
@@ -29,7 +30,7 @@ def _get_local_nodes(storage_key: StorageKey) -> Sequence[Node]:
             {"host": node.host_name, "port": node.port}
             for node in storage.get_cluster().get_local_nodes()
         ]
-    except (AssertionError, KeyError):
+    except (AssertionError, KeyError, ClickhouseError):
```
lynnagara, line 33: Aren't you raising a `UndefinedClickhouseCluster` exception?

snuba/clusters/cluster.py
```diff
@@ -374,16 +376,37 @@ def __get_cluster_nodes(self, cluster_name: str) -> Sequence[ClickhouseNode]:
...
+def _get_storage_set_cluster_map() -> Dict[StorageSetKey, ClickhouseCluster]:
+    global _STORAGE_SET_CLUSTER_MAP
+    # Map all storages to clusters via storage sets
+    if _STORAGE_SET_CLUSTER_MAP is None:
+        _STORAGE_SET_CLUSTER_MAP = {
+            storage_set: cluster
+            for cluster in CLUSTERS
+            for storage_set in cluster.get_storage_set_keys()
+        }
+    return _STORAGE_SET_CLUSTER_MAP
```
lynnagara, line 391: Why is this change needed? Wouldn't it be better to just mark the invalid cluster so that all the other valid clusters are on the map rather than not populate the map at all.

lynnagara, line 391: Why is this change needed? Wouldn't it be better to just mark the invalid cluster so that all the other valid clusters are on the map rather than not populate the map at all.

snuba/datasets/profiles.py
```diff
@@ -5,3 +5,7 @@
 class ProfilesDataset(Dataset):
     def __init__(self) -> None:
         super().__init__(default_entity=EntityKey.PROFILES)
+
+    @classmethod
+    def is_experimental(cls) -> bool:
+        return True
```
lynnagara, line 11: The PR description states the experimental dataset functionality is not used yet. It seems to be applied here though.

snuba/web/views.py
```diff
@@ -108,12 +108,20 @@ def check_down_file_exists() -> bool:
             return False
 
 
-def check_clickhouse() -> bool:
+def check_clickhouse(filter_experimental: bool = True) -> bool:
```
lynnagara, line 111: Very much a nit but `ignore_experimental` seems more intuitive to me

snuba/clusters/cluster.py
volokluev: Why would the map not be populated at all?

snuba/admin/clickhouse/nodes.py
volokluev: Good catch

snuba/datasets/profiles.py
volokluev: will remove

snuba/web/views.py
volokluev: you're right it is

snuba/clusters/cluster.py
volokluev: This map does get populated except instead of at import time, it's populated at access time.

tests/conftest.py
```diff
@@ -70,12 +71,17 @@ def run_migrations() -> Iterator[None]:
...
+                nodes = [*cluster.get_local_nodes(), *cluster.get_distributed_nodes()]
+                for node in nodes:
+                    connection = cluster.get_node_connection(
+                        ClickhouseClientSettings.MIGRATE, node
+                    )
+                    connection.execute(
+                        f"TRUNCATE TABLE IF EXISTS {database}.{table_name}"
+                    )
+            except Exception as e:
+                logging.info(f"could not tear down all nodes for {storage_key} {e}")
```
lynnagara, line 84: What exception is actually expected here now? Is it just the connection error? If so I think we should just catch that specific one not any type of exception which could hide other issues.

Also, the try/except is around all of the nodes of the cluster not each individual one. Do you want to just give up on them all if 1 fails?

tests/web/test_check_clickhouse.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@mock.patch(
+    "snuba.web.views.get_enabled_dataset_names",
+    return_value=["events", "experimental"],
+)
+@mock.patch("snuba.web.views.get_dataset", side_effect=fake_get_dataset)
+def test_check_clickhouse(mock1: mock.MagicMock, mock2: mock.MagicMock) -> None:
+    assert check_clickhouse(filter_experimental=True)
+    assert not check_clickhouse(filter_experimental=False)
```
lynnagara, line 46: You could also add an assertion that the healthcheck does fail when the dataset is really misconfigured

volokluev: Wouldn't this already do that?
```
assert not check_clickhouse(filter_experimental=False)
```

the dataset is misconfigured and because we have not ignored it the healthcheck fails

volokluev: Or you want a non-experimental dataset to be misconfigured and for it to fail as well?

snuba/clusters/cluster.py
lynnagara: My apologies it is, I assumed you were handling some error and not populating it. So then, why the change at all? Isn't it more straightforward the original way?

tests/web/test_check_clickhouse.py
lynnagara: Yeah, basically the test against the function's original behavior (which we never had before)

tests/migrations/test_runner.py
```diff
@@ -184,26 +184,30 @@ def test_no_schema_differences() -> None:
     runner = Runner()
     runner.run_all(force=True)
 
-    for storage_key in STORAGES:
-        storage = get_storage(storage_key)
-        conn = storage.get_cluster().get_query_connection(
-            ClickhouseClientSettings.MIGRATE
-        )
+    for dataset_name in DATASET_NAMES:
```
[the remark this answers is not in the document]
volokluev: Now that you say this comment, I actually think this test shouldn't need to change. I originally put it in so you could define a completely broken dataset and have it work fine but realistically if you're adding a dataset that's even broken in the dev environment you should fix that. I'll revert my changes to this test

snuba/clusters/cluster.py
volokluev: I think this is an artifact of me trying to get the whole thing working end to end. You're probably right that this change is probably unnecessary and I can remove it if you feel strongly but here's why I think it should stay:

1. I try to minimize logic at import time unless it's absolutely necessary
2. _STORAGE_SET_CLUSTER_MAP is a global variable and now it's explicitly labeled as such
3. having this be a function makes it easier to mock for tests (In fact I've already used it in one of my commits addressing your comments)

But again if you say you'd still like it removed I can remove it.

tests/web/test_check_clickhouse.py
volokluev: Sure let's be diligent

volokluev: done

snuba/clusters/cluster.py
lynnagara: I'm not sure I agree with those points. For example, we do 1 extensively in the Snuba codebase (all storages, entities and datasets). Unclear to me why it's a problem here. Since it's not relevant to this change I think it's better to remove it.

volokluev: Well for point number 3 I already used it to help me write a test for the improperly caught exception you pointed out

https://github.com/getsentry/snuba/pull/2772/files#diff-3dcdd4ef5de62043035d4ab2e57d185826e6c7d147b21cb7a4824f7e956f2666R12

volokluev: I put the map creation logic back to the import level but I still kept its access behind a function because it makes testing easier

volokluev:
@lynnagara it was a flaky kafka test that had nothing to do with these changes

lynnagara (approved):

PR #2773, opened 2022-05-26
test: Fix flaky test

lynnagara:
This was flaky before since we were previously generating random subscription UUIDs and asserting they were executed within a 50 second window. Since the jitter spread things out over a minute this would fail roughly 10/60 of the time.

### DIFF
--- a/tests/subscriptions/test_scheduler_processing_strategy.py
+++ b/tests/subscriptions/test_scheduler_processing_strategy.py
@@ -1,3 +1,4 @@
+import time
 import uuid
 from collections import deque
 from concurrent.futures import Future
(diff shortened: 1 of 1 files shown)


tests/subscriptions/test_scheduler_processing_strategy.py
```diff
@@ -747,7 +747,7 @@ def test_produce_stale_message() -> None:
             Tick(
                 0,
                 offsets=Interval(3, 4),
-                timestamps=Interval(now - timedelta(seconds=50), now),
+                timestamps=Interval(now - timedelta(seconds=30), now),
```
volokluev, line 750: if this number is important, a comment saying why may be helpful to the next reader

volokluev (approved):

tests/subscriptions/test_scheduler_processing_strategy.py
lynnagara: I think i was wrong about this being the issue, pushed a better fix now

lynnagara:
Looks like this is still flaky. Going to be making more changes here soon.

lynnagara:
@volokluev Tests look like they are passing consistently now. Since I've changed this since you first approved, mind taking another look?

volokluev:
What's the learning here? Why did these changes make the test pass consistently?

lynnagara:
> What's the learning here? Why did these changes make the test pass consistently?

The original test was wrong. I reverted the bits that weren't needed to make the actual fix clearer. The gist of it is that we were generating a random UUID and then asserting that it got executed in a 50 second window. But the jitter distributes subscription ids throughout the minute. So 10/60 times that wouldn't actually be true. The fix is to make the tick in the test just match the 60 second window, then we know it must always be scheduled once.

volokluev:
I would say your comment there would be useful context in the test. Otherwise as an outside contributor it's not clear why one worked and one didn't

volokluev (approved):
Approved but please add a comment explaining what makes the test work as a kindness to the next person that looks at it

lynnagara:
> Approved but please add a comment explaining what makes the test work as a kindness to the next person that looks at it

I'll write it since you say it's useful. Personally I'm not sure since it's not specific to this particular test and is actually true in nearly all our other subscriptions tests as well. I just happened to only make the mistake of updating to 50 seconds in this test.

PR #2774, opened 2022-05-31
ref(datasets): Add experimental dataset

volokluev:
# THIS PR WILL BE REVERTED SHORTLY AFTER BEING DEPLOYED

Adding the experimental dataset which is broken. The functionality to support it was merged in #2772 . Merging and deploying this PR will make sure of the following:

1. Deploying this broken dataset will not block deploys
2. Queries that do not touch this dataset will not fail

### Rollout plan

1. Merge this PR
2. Open the revert for merging this PR. Wait for CI to pass
3. Deploy master
4. Observe that snuba still works. If it doesn't, redeploy previous commit
5. Merge the revert (this dataset is not necessary in the codebase)


evanh (commented):
A few questions:

1) Have you verified that we won't get any datadog/sentry alerts?
2) Have you tried to run Snuba locally and seen it work?
3) Can you clarify the expected behaviour once this is deployed?
4) Is there an associated ops PR for this? 
5) Do you plan to have an actual consumer for this dataset that also fails?

volokluev:
> Have you verified that we won't get any datadog/sentry alerts?

How can I verify that without deploying it?

> Have you tried to run Snuba locally and seen it work?

Yes

> Can you clarify the expected behaviour once this is deployed?

The expected behavior is that nothing happens. This should be a non-event

> Is there an associated ops PR for this?

No. The point is that even though this dataset's storage is misconfigured, it will not fail any queries that don't go to that dataset

> Do you plan to have an actual consumer for this dataset that also fails?

I haven't taken the consumer side into account here. This is only for the query side. Once that's working, we can re-introduce this dataset with a bad consumer

evanh (approved):
You could take a quick look through our monitors to see if something might trigger. But this is happening during business hours so not a big deal.

PR #2775, opened 2022-05-31
ref(attribution): Split RequestSettings into QuerySetings and attribution

volokluev:
This PR is step one of an effort to decouple attribution of queries from the settings passed to the query pipeline. 

The main changes are in `snuba/request/validation.py` and `snuba/request/schema.py` The rest of this PR is just renaming `RequestSettings` to `QuerySettings` 

### Could this be split up into multiple PRs?
Not really, I gave it a shot however as soon as `RequestSettings` no longer exists, you can't even run a test in this repo, therefore unfortunately I have to to this in one shot. I have left comments in this PR where there is actually new logic introduced. If you really want me to separate it into separate commits I don't want to go through with that unless someone really wants it.


snuba/query/query_settings.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import List
+
+from snuba.state.quota import ResourceQuota
+from snuba.state.rate_limit import RateLimitParameters
+
+
+@dataclass()
```
fpacifici, line 10: Design question. The original intent of the RequestSettings object (which seems to me is being replaced by QuerySettings) was to keep it immutable (and then, obviously, we violated our own architecture and added the mutable rate limiter list but regardless ...). The idea was that it would only contain settings provided by the user, thus not modified by the query processing pipeline.

Do you plan to make `QuerySettings` immutable or do you expect the query pipeline could make changes to it ?

tests/request/test_schema.py
```diff
@@ -0,0 +1,34 @@
+from snuba.query.query_settings import HTTPQuerySettings
```
volokluev, line 1: new test

volokluev (commented):
Some comments which provide guidance to reviewers

snuba/request/schema.py
```diff
@@ -6,13 +6,13 @@
 import jsonschema
 
 from snuba import environment
+from snuba.query.query_settings import (
```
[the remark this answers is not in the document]
volokluev: this file now parses the JSON request into three parts, query, query_settings, and attribution_info

PR #2781, opened 2022-06-03
fix(devserver): remove extra scheduler executor command

MeredithAnya:
originally changed in https://github.com/getsentry/snuba/pull/2767 but messed up the rebase 😢


onewland:
How did you verify the fix? Did you check that it's right with both `settings.SEPARATE_SCHEDULER_EXECUTOR_SUBSCRIPTIONS_DEV` being True and False?

onewland (approved):
seems fine pending an answer to my question

PR #2782, opened 2022-06-03
feat(metrics): create generic sets aggregate table + indices

onewland:
## create generic_metrics_sets in ClickHouse
`generic_metrics_sets` is the first aggregate table I'm creating for the metrics-enhanced performance project (internal doc: https://www.notion.so/sentry/Metrics-Enhanced-Performance-Q2-SNS-Plan-79247be0ab134aae90269de0523c35f5). The goal with this approach is to create end-to-end functionality for sets (starting from the indexer ingestion in sentry, to writing to ClickHouse, and returning results via the query API) before implementing the other metric types (counters, distributions) and their processors.

Changes in this PR:
- new storage set key (for sets type only)
- new storage set (for sets type only)
- new migration group (will be shared with other "generic" metrics clusters)
- a migration to create the table

Next steps:
- create a raw table and materialized view for writing to `generic_metrics_sets`.
- create a processor and consumer in snuba for writing to the raw input table for `generic_metrics_sets`

Differences from old sets aggregate table:
- we are going to allow two modes of tag storage: (indexed key) -> (indexed value) and (indexed key) -> (raw value (string)). This means separate columns and indices.
- granularity is no longer delineated in seconds but will essentially be a constant so we can use a smaller storage format (e.g. 0=10s, 1=60s, 2=3600s, 3=86400s vs. having to store 86400 as the day value)
- a new `timeseries_id` column that lets us control sharding from within the python codebase.

Testing:
- ran local/dist migrations locally

### DIFF
--- a/snuba/datasets/storages/generic_metrics.py
+++ b/snuba/datasets/storages/generic_metrics.py
@@ -29,8 +29,10 @@ aggregated_columns = [
     Column("granularity", UInt(32)),
     Column("timestamp", DateTime()),
     Column("retention_days", UInt(16)),
-    Column("tags", Nested([("key", UInt(64)), ("value", String())])),
-    Column("_tags_hash", Array(UInt(64), SchemaModifiers(readonly=True))),
+    Column("raw_tags", Nested([("key", UInt(64)), ("value", String())])),
+    Column("_raw_tags_hash", Array(UInt(64), SchemaModifiers(readonly=True))),
+    Column("indexed_tags", Nested([("key", UInt(64)), ("value", String())])),
+    Column("_indexed_tags_hash", Nested([("key", UInt(64)), ("value", UInt(64))])),
 ]
 
 sets_storage = ReadableTableStorage(
@@ -47,5 +49,9 @@ sets_storage = ReadableTableStorage(
             ]
         ),
     ),
-    query_processors=[ArrayJoinKeyValueOptimizer("tags"), TableRateLimit()],
+    query_processors=[
+        ArrayJoinKeyValueOptimizer("raw_tags"),
+        ArrayJoinKeyValueOptimizer("indexed_tags"),
+        TableRateLimit(),
+    ],
 )
--- a/snuba/datasets/storages/generic_metrics.py
+++ b/snuba/datasets/storages/generic_metrics.py
@@ -1,0 +1,51 @@
+"""
+The storages defined in this file are for the generic metrics system,
+initially built to handle metrics-enhanced performance.
+"""
+from snuba.clickhouse.columns import (
+    AggregateFunction,
+    Array,
+    Column,
+    ColumnSet,
+    DateTime,
+    Nested,
+    SchemaModifiers,
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.datasets.schemas.tables import TableSchema
+from snuba.datasets.storage import ReadableTableStorage
+from snuba.datasets.storages import StorageKey
+from snuba.query.processors.arrayjoin_keyvalue_optimizer import (
+    ArrayJoinKeyValueOptimizer,
+)
+from snuba.query.processors.table_rate_limit import TableRateLimit
+
+aggregated_columns = [
+    Column("org_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("metric_id", UInt(64)),
+    Column("granularity", UInt(32)),
+    Column("timestamp", DateTime()),
+    Column("retention_days", UInt(16)),
+    Column("tags", Nested([("key", UInt(64)), ("value", String())])),
+    Column("_tags_hash", Array(UInt(64), SchemaModifiers(readonly=True))),
+]
+
+sets_storage = ReadableTableStorage(
+    storage_key=StorageKey.GENERIC_METRICS_SETS,
+    storage_set_key=StorageSetKey.GENERIC_METRICS_SETS,
+    schema=TableSchema(
+        local_table_name="generic_metrics_sets_local",
+        dist_table_name="generic_metrics_sets_dist",
+        storage_set_key=StorageSetKey.GENERIC_METRICS_SETS,
+        columns=ColumnSet(
+            [
+                *aggregated_columns,
+                Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
+            ]
+        ),
+    ),
+    query_processors=[ArrayJoinKeyValueOptimizer("tags"), TableRateLimit()],
+)
--- a/snuba/datasets/storages/generic_metrics.py
+++ b/snuba/datasets/storages/generic_metrics.py
@@ -24,7 +24,7 @@ from snuba.query.processors.table_rate_limit import TableRateLimit
 
 aggregated_columns = [
     Column("org_id", UInt(64)),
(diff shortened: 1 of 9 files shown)


snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
```diff
@@ -0,0 +1,143 @@
...
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("granularity", UInt(8)),
+        Column("timestamp", DateTime()),
+        Column("retention_days", UInt(16)),
+        Column("indexed_tags", Nested([("key", UInt(64)), ("value", UInt(64))])),
+        Column("raw_tags", Nested([("key", UInt(64)), ("value", String())])),
+        Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
+        Column("timeseries_id", UInt(64)),
```
onewland, line 36: `timeseries_id` will be a hash generated by the consumer for sharding so we can isolate that logic from ClickHouse itself

snuba/datasets/storages/generic_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+    Column("org_id", UInt(64)),
+    Column("use_case_id", String()),
+    Column("project_id", UInt(64)),
+    Column("metric_id", UInt(64)),
+    Column("granularity", UInt(8)),
+    Column("timestamp", DateTime()),
+    Column("retention_days", UInt(16)),
+    Column("raw_tags", Nested([("key", UInt(64)), ("value", String())])),
+    Column("_raw_tags_hash", Array(UInt(64), SchemaModifiers(readonly=True))),
+    Column("indexed_tags", Nested([("key", UInt(64)), ("value", String())])),
```
fpacifici, line 35: You should not need to store the tag keys twice. Can't we have a tag keys column and two tag values (one for strings and one for ints) ? 
Also there is an issue here. The indexed_tags value should be a `UInt64`.

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
```diff
@@ -0,0 +1,143 @@
...
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    granularity = "2048"
+    local_table_name = "generic_metric_sets_local"
+    columns: Sequence[Column[Modifiers]] = [
+        Column("use_case_id", String(Modifiers(low_cardinality=True))),
```
fpacifici, line 26: Just a nit: Since this is not going to be in the primary key, I would put it after the primary key columns.

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
```diff
@@ -0,0 +1,143 @@
...
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("granularity", UInt(8)),
+        Column("timestamp", DateTime()),
+        Column("retention_days", UInt(16)),
+        Column("indexed_tags", Nested([("key", UInt(64)), ("value", UInt(64))])),
+        Column("raw_tags", Nested([("key", UInt(64)), ("value", String())])),
```
fpacifici, line 34: Same comment I had on the storage. We shouldn't need to store the tag keys column twice.

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
fpacifici: You should not need this in the aggregated merge tree. Only in the raw table.
Sharding happens at the raw table level. 
- you write on the raw table
- the distributed raw table partitions data across shards according to the sharding key
- each storage node writes data on the raw table locally
- the materialized view generates the aggregated data and stores them in the aggregatingMergeTree locally (from the local raw table). There is no sharding happening at this step.

(also I think we will be able to get away with 32 bit for the time being, since it is data retained for a short period of time it will not be a problem.)

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
```diff
@@ -0,0 +1,143 @@
...
+    ]
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.GENERIC_METRICS_SETS,
+                table_name=self.local_table_name,
+                engine=table_engines.AggregatingMergeTree(
+                    storage_set=StorageSetKey.GENERIC_METRICS_SETS,
+                    order_by="(org_id, project_id, metric_id, granularity, timestamp, indexed_tags.key, indexed_tags.value, raw_tags.key, retention_days)",
```
fpacifici, line 46: I think this is wrong
`indexed_tags.key, indexed_tags.value, raw_tags.key` 

It should be 
`indexed_tags.key, indexed_tags.value, raw_tags.value` 

Also please add here the use case id at the end. Otherwise you will not have partitioning. Two (unlikely) identical time series on two separate use case ids would be merged.

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
```diff
@@ -0,0 +1,143 @@
...
+        ]
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.GENERIC_METRICS_SETS,
+                table_name="generic_metric_sets_aggregated_dist",
+                engine=table_engines.Distributed(
+                    local_table_name=self.local_table_name,
+                    sharding_key="timeseries_id",
```
fpacifici, line 131: You do not need to add this sharding key to the aggregated distributed table. 
That goes on the raw table as materialization happens on local nodes. Not distributed ones.

fpacifici (changes requested):
Please see comments inline

nikhars (approved):
LGTM

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
```diff
@@ -0,0 +1,143 @@
...
+        Column("retention_days", UInt(16)),
+        Column(
+            "tags",
+            Nested(
+                [
+                    ("key", UInt(64)),
+                    ("indexed_value", UInt(64)),
+                    ("raw_value", String()),
+                ]
+            ),
```
fpacifici, line 40: You do not need to do anything right now. 
But watch out when you introduce the storage. I think some of the storage query processor to optimize tags may rely on the value column being called `value` and they will likely have to be updated.

fpacifici (approved):

snuba/migrations/snuba_migrations/generic_metrics/0001_sets_aggregate_table.py
onewland: yeah, I noticed that might be an issue on the ArrayJoinKeyValueOptimizer

PR #2784, opened 2022-06-03
feat(rate-limit): Add rate limit metrics

nikhars:
### Context
We would like to add alerts when the global or table rate limits are reached

### Solution
In order to get there, we need the metrics so that we can create alerts based on them. Added rate limited metrics for global and table rate limits. They can be differentiated by either them being per-second or concurrent one. If its a table rate limit we also add the table on which the limit was reached.


MeredithAnya:
I thought we did have a `rate-limited` status that we capture in metrics. Is this just to break down that further so we know which rate limit it was?

snuba/state/rate_limit.py
```diff
@@ -267,6 +267,28 @@ def get_global_rate_limit_params() -> RateLimitParameters:
...
+    Record rate limit metrics if needed.
+
+    We only record the metrics for global and table rate limits since
+    those indicate capacity of clickhouse clusters.
+
+    We get the scope and name from the message of RateLimitExceeded
+    exception since we want to know whether we exceeded the concurrent
+    or per second limit.
+    """
+    scope, name = exc.message.split()[:2]
```
evanh, line 283: Relying on the exception message doesn't seem a like a good idea. I would pass that information directly into the function.

evanh (commented):
Are these metrics going to be tied to Pagerduty? If so, I don't know that rate limiting warrants an incident.

nikhars:
> I thought we did have a `rate-limited` status that we capture in metrics (specific metric being `snuba.api.query.count`). Is this just to break down that further so we know which rate limit it was?

That is correct. The current one is based on the HTTP return status which does not provide insights into the type of rate limit. We have a few of them and we would like to know if global or table one caused it since it points to potential capacity problems

snuba/state/rate_limit.py
nikhars: I also didn't want to rely on it initially.

The other way I was doing it initially is by raising different exception types in the code here https://github.com/getsentry/snuba/blob/master/snuba/state/rate_limit.py#L199-L228. But the problem is that the code is agnostic to the type of rate-limiting it is currently running. So I wanted to keep it that way. That meant passing the exception to raise in the `rate_limit_params`. Then the exception raised would also need to be different based on whether it is per-second or concurrent. So it starts getting complicated.

So I went with this mechanism. I am open to other ideas of achieving this without doing a lot of refactoring.

nikhars:
> Are these metrics going to be tied to Pagerduty? If so, I don't know that rate limiting warrants an incident.

The global and table rate limit would be of interest to us. While setting up the alert we can decide if we want the alert to page us or notify us.

snuba/state/rate_limit.py
onewland: Would it be cleaner to make separate attributes for the scope and the name on `RateLimitExceeded`? Maybe they could be optional if not every caller uses them

nikhars: Made the changes as per the suggestion

onewland (approved):

PR #2789, opened 2022-06-06
Revert "feat(metrics): create generic sets aggregate table + indices"

onewland:
Reverts getsentry/snuba#2782

I didn't add the cluster to the ops definition in https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L11 yet so that would block deploys.


onewland:
I think I'm wrong and this change was fine

PR #2791, opened 2022-06-07
feat(subscriptions): Allow scheduling watermark mode to be overridden

lynnagara:
This is a temporary change that allows the scheduling watermark mode
to be overridden in the combined scheduler/executor. This is needed for
single tenant as we often have empty partitions there and empty partitions are
incompatible with "global" mode. Once transactions are no longer semantically
partitioned, this will not be necessary and this can be removed.

### DIFF
--- a/snuba/cli/subscriptions_scheduler_executor.py
+++ b/snuba/cli/subscriptions_scheduler_executor.py
@@ -72,9 +72,9 @@ from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
 )
 @click.option("--log-level", help="Logging level to use.")
 @click.option(
-    "--override-partition-mode",
+    "--scheduling-mode",
     type=click.Choice(["partition", "global"]),
-    help="Skip scheduling if timestamp is beyond this threshold compared to the system time",
+    help="Overrides the partition scheduling mode associated with the dataset.",
 )
 def subscriptions_scheduler_executor(
     *,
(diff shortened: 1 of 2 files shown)


snuba/cli/subscriptions_scheduler_executor.py
```diff
@@ -70,6 +71,11 @@
     help="Skip scheduling if timestamp is beyond this threshold compared to the system time",
 )
 @click.option("--log-level", help="Logging level to use.")
+@click.option(
+    "--override-partition-mode",
```
nikhars, line 75: Nit: Instead of calling this override-partition-mode, would it make more sense to call it `--scheduling-mode`. What this means is that if a parameter is provided from the CLI then use that instead of the default configuration. That would make it more explicit what scheduling mode the consumer is running on.

snuba/cli/subscriptions_scheduler_executor.py
```diff
@@ -70,6 +71,11 @@
...
+    type=click.Choice(["partition", "global"]),
+    help="Skip scheduling if timestamp is beyond this threshold compared to the system time",
```
nikhars, line 77: help string has a copy/paste error.

onewland:
Can you define what a scheduling watermark mode is?

snuba/cli/subscriptions_scheduler_executor.py
lynnagara: Sure, ok

snuba/cli/subscriptions_scheduler_executor.py
lynnagara: Thanks

lynnagara:
@nikhars Updated with your feedback

lynnagara:
> Can you define what a scheduling watermark mode is?

@onewland It defines whether scheduling happens either immediately or waits for all partitions. It's described in https://github.com/getsentry/snuba/blob/8118e109a435445009b79fdcf46502ced2ed4c81/snuba/subscriptions/scheduler_processing_strategy.py#L180-L201

PR #2797, opened 2022-06-08
fix(rate-limit): Don't override SerializableException constructor

nikhars:
Overriding SerializableException constructor is a bad idea since making
the original exception back from a serialized version does not work
correctly. Use extra_data instead.


snuba/utils/serializable_exception.py
```diff
@@ -86,6 +86,14 @@ def _get_registry() -> _ExceptionRegistry:
 
 
 class SerializableException(Exception):
+    """
```
volokluev, line 89: You could enforce this in the `__init_subclass__` function if you wanted

volokluev (approved):
LGTM, some comments that are purely educational

PR #2798, opened 2022-06-09
feat: Remove legacy events subscriptions consumer from Freight

lynnagara:
This has been replaced by the scheduler/executor and is no longer
serving any prod traffic


volokluev (approved):
hell yeah

PR #2801, opened 2022-06-09
fix: Update to latest version of Redis (4.3.3)

evanh:
Our current version of Redis is EOL, so update to the latest. Dependabot only
updated to 4.1.3 but 4.3.3 contains fixes for things that were broken in the
version Dependabot wanted. redis-py-cluster has been added to the redis library
so we don't need to import it separately any more.


volokluev (approved):
I'm guessing `register_scipt` works in the updated redis version?

PR #2802, opened 2022-06-13
fix: Update frontend packages

evanh:
This is a duplicate of a couple dependabot PRs. They were failing CI and it's
easier to debug CI problems on an internal PR vs. external.

Dependabot PRs:
https://github.com/getsentry/snuba/pull/2779
https://github.com/getsentry/snuba/pull/2790
https://github.com/getsentry/snuba/pull/2804

### DIFF
--- a/snuba/admin/package.json
+++ b/snuba/admin/package.json
@@ -15,6 +15,6 @@
   },
   "devDependencies": {
     "webpack": "^5.73.0",
-    "webpack-cli": "^4.9.2"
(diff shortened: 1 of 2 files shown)


lynnagara:
Is there a new js build to be checked in, or did bumping these not change that at all?

evanh:
> did bumping these not change the bundle at all?

No change to the bundle, I made sure.

lynnagara (approved):

PR #2806, opened 2022-06-13
feat(replays): replays processor to ingest data into clickhouse

JoshFerge:
(re-open of #2683)

following up on https://github.com/getsentry/snuba/pull/2681, we add a processor to insert replay data rows into clickhouse.

common code has been copy/pasted from relevant consumers for things like tag/environment processing etc.

Will note that we'll be rapidly prototyping on top of this, so opting for quicker / more isolated changes (as opposed to e.g. refactoring to be more DRY for tag processing).

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -295,6 +295,19 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                 ],
             ),
         ]
+    if settings.ENABLE_REPLAYS_CONSUMER:
+        daemons += [
+            (
+                "replays-consumer",
+                [
+                    "snuba",
+                    "consumer",
+                    "--auto-offset-reset=latest",
+                    "--log-level=debug",
+                    "--storage=replays",
+                ],
+            ),
+        ]
 
     manager = Manager()
     for name, cmd in daemons:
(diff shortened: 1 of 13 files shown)


JoshFerge:
updates after last feedback:
- added a TODO for setting up deadletter queue
- added is_experimental override
- removed unnecessary event_type check

rahul-kumar-saini:
Adding a DLQ is super easy:

1. Create a function that returns a [`DeadLetterQueuePolicy`](https://github.com/getsentry/arroyo/tree/main/arroyo/processing/strategies/dead_letter_queue/policies).
2. Add the function to the stream loader in the storage.

Example [here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/metrics.py#L96).

Any unhandled exceptions which occur during the processing of a message will be caught and routed to the policy to be handled. This policy can do whatever you want really, in the example above it just produces the bad messages to a dead-letter topic for manual inspection.

JoshFerge:
@rahul-kumar-saini awesome, thanks for the instructions! a DLQ isn't exactly necessary for our initial roll out here, and i've got half a dozen other pieces i'm trying to finish up, is it OK if i leave it as a TODO? if not let me know!

evanh (commented):
Nice, this is looking good! Note that to actually deploy this you'll also need to add an entry to the freight.yml file. Is there an associated ops PR for this?

JoshFerge:
@evanh i'll make that now.

nikhars (approved):
LGTM

JoshFerge:
@nikhars @evanh I assume this can be merged in without the ops changes, will do in a bit so unless i hear otherwise.

PR #2813, opened 2022-06-14
fix: Update various dependencies

evanh:
bump flake8 from 3.7.8 to 4.0.1
bump progressbar2 from 3.53.1 to 4.0.0
bump chardet from 3.0.4 to 4.0.0
bump pyparsing from 2.2.0 to 3.0.9

These are copied from dependabot PRs:
https://github.com/getsentry/snuba/pull/2751
https://github.com/getsentry/snuba/pull/2752
https://github.com/getsentry/snuba/pull/2812
https://github.com/getsentry/snuba/pull/2811

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ atomicwrites==1.1.5
 blinker==1.4
 black==22.3.0
 certifi==2018.4.16
-chardet==3.0.4
+chardet==4.0.0
 click==8.0.4
 clickhouse-driver==0.2.4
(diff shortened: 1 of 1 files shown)


volokluev (approved):
approved but I think flake8 should be a test dependency

evanh:
> approved but I think flake8 should be a test dependency

It's top level because it runs outside of tests (in the linting step). In the SDK I created a separate linting requirements, we could do that here too.

PR #2825, opened 2022-06-15
feat(profiling): Add aggregating merge tree table for functions

Zylphrex:


snuba/migrations/groups.py
```diff
@@ -17,6 +17,7 @@ class MigrationGroup(Enum):
     SESSIONS = "sessions"
     QUERYLOG = "querylog"
     PROFILES = "profiles"
+    FUNCTIONS = "functions"
```
fpacifici, line 20: Any reason for this feature to be separate from the profiles group?

snuba/migrations/snuba_migrations/functions/0001_functions.py
```diff
@@ -0,0 +1,189 @@
...
+    Column("transaction_name", String()),
+    Column("timestamp", DateTime()),
+    Column("depth", UInt(32)),
+    Column("parent_fingerprint", UInt(64)),
+    Column("fingerprint", UInt(64)),
+    Column("symbol", String()),
+    Column("image", String()),
+    Column("filename", String()),
+    Column("is_application", UInt(8)),
+    Column("tags", Nested([Column("key", String()), Column("value", String())])),
```
fpacifici, line 30: Why do we need custom arbitrary tags? These are not a good fit for rollup scenarios as a customer could be able to multiply the size of your table by just adding new tags.

snuba/migrations/snuba_migrations/functions/0001_functions.py
```diff
@@ -0,0 +1,189 @@
...
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+common_columns: List[Column[Modifiers]] = [
+    Column("org_id", UInt(64)),
+    Column("project_id", UInt(64)),
```
fpacifici, line 20: WHy do you need both organization and project? Are you expecting to query across projects in an organization? It seems it would be weird as all queries are filtered by transaciton_name.

In practice, unless you are filtering thousands of specific project ids, filtering by a group of projects is more efficient than doing it by organization as you do not need to load an additional UInt64 column (organization).
It also keeps your primary key more efficient as there would be one fewer field.

snuba/migrations/snuba_migrations/functions/0001_functions.py
```diff
@@ -0,0 +1,189 @@
...
+        return [
+            operations.CreateTable(
+                storage_set=self.storage_set,
+                table_name=self.local_raw_table,
+                columns=raw_columns,
+                engine=table_engines.MergeTree(
+                    storage_set=self.storage_set,
+                    order_by="(org_id, project_id, transaction_name, timestamp)",
+                    partition_by="(toStartOfInterval(timestamp, INTERVAL 3 day))",
+                    ttl="timestamp + toIntervalDay(7)",
```
fpacifici, line 80: Do you need 7 days of data? 
How backdated can profile be? Are you going to allow the ingestion of profiles that happened days ago ?

snuba/migrations/snuba_migrations/functions/0001_functions.py
```diff
@@ -0,0 +1,189 @@
...
+                ),
+            ),
+            operations.CreateTable(
+                storage_set=self.storage_set,
+                table_name=self.local_materialized_table,
+                columns=agg_columns,
+                engine=table_engines.AggregatingMergeTree(
+                    storage_set=self.storage_set,
+                    order_by="(org_id, project_id, transaction_name, timestamp, depth, parent_fingerprint, fingerprint, symbol, image, filename, is_application, tags.key, tags.value, retention_days)",
+                    primary_key="(org_id, project_id, transaction_name, timestamp, depth, parent_fingerprint, fingerprint)",
```
fpacifici, line 90: This matters when you filter your data. Do you expect to ever be filtering anything past timestamp? `depth, parent_fingerprint, fingerprint`? If not then keeping them here only makes the primary key less efficient as the primary index would not fit in memory.

snuba/migrations/snuba_migrations/functions/0001_functions.py
```diff
@@ -0,0 +1,189 @@
...
+    Column("project_id", UInt(64)),
+    Column("transaction_name", String()),
+    Column("timestamp", DateTime()),
+    Column("depth", UInt(32)),
+    Column("parent_fingerprint", UInt(64)),
+    Column("fingerprint", UInt(64)),
+    Column("symbol", String()),
+    Column("image", String()),
+    Column("filename", String()),
+    Column("is_application", UInt(8)),
```
fpacifici, line 29: How do you plan to use these additional fields.

snuba/migrations/snuba_migrations/functions/0001_functions.py
Zylphrex: This will be useful for constructing visualizations flamegraphs/calltrees. In order to construct them, we need to page through the result frames ordered by depth, parent_fingeprint and fingerprint to be able to consistently connect frames to their parents.

The condition will look something like the following

```sql
SELECT
  symbol,
  image,
  filename,
  fingerprint,
  parent_fingerprint,
  count(),
  sum(duration)
FROM functions_dist
WHERE
  start_timestamp >= ...
  AND start_timestamp < ...
  AND project_id = ''
  AND transaction_name = ''
  AND ...
  // we need to order the frames by depth/parent_fingerprint/fingerprint
  // so we can page through the results and merge them together
  AND (
    (depth > max depth from last batch)
    OR (
      depth = max depth from last batch
      AND parent_fingerprint >= max parent fingerprint from max depth from last batch
      AND fingerprint >= max fingerprint from max depth from last batch
    )
GROUP BY symbol, address, fingerprint, parent_fingerprint
ORDER BY depth ASC, parent_fingerprint ASC, fingerprint ASC
LIMIT 1000
```

Would it make sense to remove these fields from the primary keys in this case?

snuba/migrations/snuba_migrations/functions/0001_functions.py
Zylphrex: For ios/android, this will largely be a few device fields like
- device os name/version
- release
- android API level
These are mostly system generated, but release is user generated. In practice, this should be fairly low cardinality but we will verify this as we see more data.

snuba/migrations/snuba_migrations/functions/0001_functions.py
Zylphrex: Can you elaborate a little more on how this `materialization_version` is used? If I understand it correctly, if we want to populate a new materialization with new columns, we would have to do a migration to insert the column on the raw table as well. Is the use here just so that we can target only rows with the right `materialization_version` when the table is materialized?

nikhars: When you want to move to a new materialization view having a `materialization_version` would allow you to gracefully do it. Lets assume you have a `fn_raw_table` with `materialization_version` fields being written with value 1. It populates `fn_mv_1` based on `WHERE materialization_version = 1`. Now you have a use case where you add another column to `fn_raw_table`, say column `x`. You want to build a new materialization using column `x`. So you define `fn_mv_2` using `WHERE materialization_version=2`. Doing this allows you to alter the raw table and populate new materializations.

snuba/migrations/groups.py
fpacifici: Migration groups are not contextual to tables, they are logical grouping related to specific product. The reason they exist is to exclude/include a product from migrations entirely.
So it is fundamentally our decision. For your use case I would be ok one way or another.

But when you put two migrations in the same group you are saying that there is a sequential dependency between them. You won't be able to skip one while still running the rest of the group.

snuba/migrations/snuba_migrations/functions/0001_functions.py
fpacifici: In the end it will be your choice, though I would discourage pretty strongly tags in aggregating merge trees. 
That tends to generate wild variance in your storage size and query performance.
Let's say that you add a simple tag to all your messages somewhere early in the product without thinking about that too much -> you just doubled your storage size in one shot as, from that point on, you store multiple buckets when you were storing one.

If you know which tags you want, just add columns for them

snuba/settings/__init__.py
```diff
@@ -169,7 +170,7 @@
 COLUMN_SPLIT_MAX_RESULTS = 5000
 
 # Migrations in skipped groups will not be run
-SKIPPED_MIGRATION_GROUPS: Set[str] = {"querylog", "profiles", "replays"}
+SKIPPED_MIGRATION_GROUPS: Set[str] = {"querylog", "profiles", "functions", "replays"}
```
fpacifici, line 173: Now this setting is also overridden in the production config.
Please remember to send a PR to add your group there.

snuba/migrations/snuba_migrations/functions/0001_functions.py
```diff
@@ -0,0 +1,225 @@
...
+        return [
+            operations.CreateTable(
+                storage_set=self.storage_set,
+                table_name=self.local_raw_table,
+                columns=raw_columns,
+                engine=table_engines.MergeTree(
+                    storage_set=self.storage_set,
+                    order_by="(project_id, transaction_name, timestamp)",
+                    partition_by="(toStartOfInterval(timestamp, INTERVAL 12 HOUR))",
+                    ttl="timestamp + toIntervalDay(1)",
```
fpacifici, line 80: Are you allowing backdated profiles in the DB? 
Like would you accept a profile with a timestamp older than 24 hours ago? 
If yes here you may have a problem and you should verify that. If you insert a row with a timestamp older than 24 hours, does it get materialized or is it dropped immediately ?

fpacifici (approved):

snuba/migrations/groups.py
Zylphrex: In this case, I think it makes sense here to separate the migrations into two migration groups so that we can easily make changes to the datasets independently.

snuba/migrations/snuba_migrations/functions/0001_functions.py
Zylphrex: Understood, this is definitely something that we'll keep in mind. I think what we'll do here is start with these tag columns for the flexibility right now, and tread carefully. Once we have a clearer understanding of the right fields to keep for the various platforms, we'll work on adding columns for them.

snuba/migrations/snuba_migrations/functions/0001_functions.py
Zylphrex: For now, this should be impossible because profiles are dated with the timestamp they were received. But I have verified if a row is inserted with a timestamp older than 24 hours, it's not guaranteed to be dropped immediately. It looks like it'll automatically delete rows past its TTL but clickhouse doesn't guarantee when that'll happen, so we probably should not rely on this behaviour if we want to accept backdated profiles.

snuba/settings/__init__.py
Zylphrex: where can i find this production config?

snuba/migrations/snuba_migrations/functions/0001_functions.py
Zylphrex: This has been removed in favour of a few columns that we know we want.

snuba/settings/__init__.py
phacops: https://github.com/getsentry/ops/pull/4902

PR #2830, opened 2022-06-16
fix(admin): make get local table name resilient to bad clickhouse configs

volokluev:
Currently this breaks the admin page when an experimental cluster is not properly configured. Make it so that only the bad cluster config is affected

### DIFF
--- a/snuba/admin/clickhouse/nodes.py
+++ b/snuba/admin/clickhouse/nodes.py
@@ -18,9 +18,12 @@ Storage = TypedDict(
 
 
 def _get_local_table_name(storage_key: StorageKey) -> str:
-    schema = get_storage(storage_key).get_schema()
-    assert isinstance(schema, TableSchema)
-    return schema.get_table_name()
+    try:
+        schema = get_storage(storage_key).get_schema()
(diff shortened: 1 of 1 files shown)


snuba/admin/clickhouse/nodes.py
```diff
@@ -18,9 +18,12 @@
...
 def _get_local_table_name(storage_key: StorageKey) -> str:
-    schema = get_storage(storage_key).get_schema()
-    assert isinstance(schema, TableSchema)
-    return schema.get_table_name()
+    try:
+        schema = get_storage(storage_key).get_schema()
+        assert isinstance(schema, TableSchema)
+        return schema.get_table_name()
+    except UndefinedClickhouseCluster:
+        return "badcluster"
```
onewland, line 26: should we also filter out this value from what `get_storage_info` returns?

onewland (approved):

snuba/admin/clickhouse/nodes.py
volokluev: I'd rather we query it and just have it blow up

PR #2837, opened 2022-06-17
ref(subscriptions): Remove the ability to run legacy subscriptions

lynnagara:
This has been replaced by the scheduler, executor (or combined scheduler
and executor) command. It is no longer being used.


onewland (approved):
looks good pending tests passing

PR #2839, opened 2022-06-17
Add replays-consumer to Freight

rgibert:
Update Freight config to deploy to `replays-consumer`.


lynnagara (changes requested):
We need to hold off as there is currently a bug in the schema causing all messages to fail processing

fpacifici:
@lynnagara  is this still the case ?

fpacifici (approved):

rahul-kumar-saini:
https://github.com/getsentry/snuba/pull/2997

PR #2841, opened 2022-06-20
fix: Package updates

evanh:
Dependabot wanted werkzeug to go to 2.1.2, but doing that required updating
Flask, which required updating Jinja2 etc. so I used a slightly less version
and we can tackle updating Flask etc. separately.

This is aggregating a few dependabot PRs:
https://github.com/getsentry/snuba/pull/2833
https://github.com/getsentry/snuba/pull/2834
https://github.com/getsentry/snuba/pull/2835
https://github.com/getsentry/snuba/pull/2836

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,13 @@
 atomicwrites==1.4.0
 blinker==1.4
 black==22.3.0
-certifi==2018.4.16
+certifi==2022.6.15
 chardet==4.0.0
 click==8.0.4
 clickhouse-driver==0.2.4
 colorama==0.3.9
 confluent-kafka==1.7.0
-coverage==5.0
+coverage==6.4.1
 datadog==0.21.0
 deprecation==2.0.3
 docopt==0.6.2
@@ -31,7 +31,7 @@ parsimonious==0.8.1
 py==1.10.0
 pyparsing==3.0.9
 pytest==6.2.0
-pytest-cov==2.10.1
+pytest-cov==3.0.0
 pytest-watch==4.2.0
 pathtools==0.1.2
 pluggy==0.13.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -31,7 +31,7 @@ parsimonious==0.8.1
 py==1.10.0
 pyparsing==3.0.9
 pytest==6.2.0
-pytest-cov==2.10.1
+pytest-cov==3.0.0
```
volokluev, line 34: this should probably be in requirements-test.txt

requirements.txt
```diff
@@ -47,4 +47,4 @@ simplejson==3.15.0
 urllib3==1.26.9
 uWSGI==2.0.20
 wcwidth==0.1.7
-Werkzeug==0.16.1
+Werkzeug==2.0.3
```
volokluev, line 50: damn, that's an update alright

snuba/admin/views.py
```diff
@@ -79,6 +79,9 @@ def clickhouse_queries() -> Response:
 def clickhouse_system_query() -> Response:
     req = request.get_json()
     try:
+        if req is None:
+            req = {}
```
volokluev, line 83: what made this necessary?

volokluev: also this will fail on the next line

evanh: Added a comment, it's for a typing error. I thought it best to preserve the `KeyError` happening if `req` is `None` to avoid making logical changes.

requirements.txt
evanh: This might be used outside of tests too, I'll have to look into it. For these PRs, I am trying to avoid any change besides the version numbers.

volokluev (approved):

PR #2875, opened 2022-06-22
db(metrics): migrations to update generic sets schema

onewland:
Allow for custom granularities to be entered in the raw table and expanded in the materialized view

### DIFF
--- a/snuba/migrations/snuba_migrations/generic_metrics/0005_sets_replace_mv.py
+++ b/snuba/migrations/snuba_migrations/generic_metrics/0005_sets_replace_mv.py
@@ -1,0 +1,96 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import (
+    AggregateFunction,
+    Column,
+    DateTime,
+    Nested,
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    view_name = "generic_metric_sets_aggregation_mv"
+    dest_table_columns: Sequence[Column[Modifiers]] = [
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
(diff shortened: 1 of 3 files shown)


fpacifici (approved):

snuba/migrations/snuba_migrations/generic_metrics/0005_sets_replace_mv.py
```diff
@@ -0,0 +1,96 @@
...
+                    org_id,
+                    project_id,
+                    metric_id,
+                    arrayJoin(granularities) as granularity,
+                    tags.key,
+                    tags.indexed_value,
+                    tags.raw_value,
+                    toDateTime(multiIf(granularity=0,10,granularity=1,60,granularity=2,3600,granularity=3,86400,-1) *
+                      intDiv(toUnixTimestamp(timestamp),
+                             multiIf(granularity=0,10,granularity=1,60,granularity=2,3600,granularity=3,86400,-1))) as timestamp,
```
nikhars, line 63: Since we are allowing granularities to be passed in now, I feel we should get rid of the optimization where we are mapping values like 0 -> 10, 1 -> 60 etc. Just use the raw values of granularities

snuba/migrations/snuba_migrations/generic_metrics/0005_sets_replace_mv.py
```diff
@@ -0,0 +1,96 @@
...
+        Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
+        Column("use_case_id", String(Modifiers(low_cardinality=True))),
+    ]
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.DropTable(
+                storage_set=StorageSetKey.GENERIC_METRICS_SETS,
+                table_name=self.view_name,
+            ),
```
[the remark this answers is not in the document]
onewland: These tables are not in use in prod

snuba/migrations/snuba_migrations/generic_metrics/0005_sets_replace_mv.py
onewland: That's a fair point on the raw table, but it still save us index size on the aggregate table

PR #2877, opened 2022-06-22
feat(perf-on-metrics): processing infrastructure for generic metrics sets

onewland:


snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,116 @@
...
+            "granularities": [
+                GRANULARITY_ONE_MINUTE,
+                GRANULARITY_ONE_HOUR,
+                GRANULARITY_ONE_DAY,
+            ],
+        }
+        return InsertBatch([processed], None)
+
+
+class GenericSetsMetricsProcessor(GenericMetricsBucketProcessor):
```
evanh, line 111: This is just a design thing, but I bet you could make one class that just encodes the actual constant value. Then the type-specific classes could override just that value.

```
class GenericMetricsBucketProcessor(...):
    internal_type: str = METRICS_TYPE.DEFAULT
    
    def _should_process(self, message: Mapping[str, Any]) -> bool:
        return message["type"] is not None and message["type"] == self.internal_type
    ....

class GenericSetsMetricsProcessor(GenericMetricsBucketProcessor):
    internal_type: METRICS_TYPE.SET

```

onewland: yeah, this whole series of classes should definitely be split up in a different way. When I get to counters and distributions I think I might create some type aliases e.g.

```
messageFilter = Callable[Mapping[str,Any],bool]
valueGenerator = Callable[Mapping[str,Any],Mapping[str,Numeric]]
```

and then each of these subclasses can be replaced with an instance that sets the right values for those (and even those methods can probably be shared with some currying)

tests/datasets/test_metrics_processor.py
```diff
@@ -451,3 +468,46 @@ def test_metrics_polymorphic_processor(
...
+                "project_id": 2,
+                "metric_id": 1232341,
+                "timestamp": expected_timestamp,
+                "tags.key": [10, 20, 30],
+                "tags.indexed_value": [11, 22, 33],
+                "tags.raw_value": ["value-1", "value-2", "value-3"],
+                "metric_type": "set",
+                "set_values": [324234, 345345, 456456, 567567],
+                "materialization_version": 1,
+                "timeseries_id": ANY,
```
fpacifici, line 489: I think we should test the time series id calculation works. 

Having the field populated here ensures at least that it is not None, but this is secondary problem.
I think you should have a dedicated unit test for the time series id generator that validates that messages with different tags keys/vals/metric id  do not share the same time series id.

fpacifici (changes requested):
Please see my comment on time series id generation.

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,116 @@
...
+        _hash_timeseries_id should return a UInt32 whose distribution should shard
+        as evenly as possible while ensuring that an average query will not have to
+        cross shards to read a results (so for the same org, project, metric, and tags
+        ClickHouse should not have to aggregate results from multiple nodes).
+        """
+        use_case_id: str = message.get("use_case_id") or ""
+        org_id: str = message["org_id"]
+        project_id: str = message["project_id"]
+        metric_id: str = message["metric_id"]
+        tag_keys_comma_sep: str = ",".join(message["tags"].keys())
```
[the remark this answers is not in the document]
onewland: good call, I actually intended to do the tag key sorting.

one issue with using the indexed tag values is that they may not be present in the future, but supporting your approach is that we already can't handle it properly if we don't have indexed values anyway.

tests/datasets/test_metrics_processor.py
onewland: agreed, will add

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,116 @@
...
+        raise NotImplementedError
+
+    def _hash_timeseries_id(self, message: Mapping[str, Any]) -> int:
+        """
+        _hash_timeseries_id should return a UInt32 whose distribution should shard
+        as evenly as possible while ensuring that an average query will not have to
+        cross shards to read a results (so for the same org, project, metric, and tags
+        ClickHouse should not have to aggregate results from multiple nodes).
+        """
+        use_case_id: str = message.get("use_case_id") or ""
```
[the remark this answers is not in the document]
onewland: good call, I'll take it out. I sort of made this method as a skeleton definition while I was verifying that the rest of the plumbing worked, but as you point out in your comments the details need to be revisited

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,116 @@
...
+        org_id: str = message["org_id"]
+        project_id: str = message["project_id"]
+        metric_id: str = message["metric_id"]
+        tag_keys_comma_sep: str = ",".join(message["tags"].keys())
+
+        return zlib.adler32(
```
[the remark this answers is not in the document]
onewland: I think I just picked it because of this sentence in the [docs](https://docs.python.org/3/library/zlib.html):

> (An Adler-32 checksum is almost as reliable as a CRC32 but can be computed much more quickly.)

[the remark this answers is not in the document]
onewland: seems like fairly similar performance:

```
>>> byte_strings = [secrets.token_bytes(500) for i in range(1,1000000)]
>>> start = timeit.default_timer(); x = [zlib.adler32(s) for s in byte_strings]; end = timeit.default_timer()
>>> end-start
0.22351763100004973
>>> start = timeit.default_timer(); x = [zlib.crc32(s) for s in byte_strings]; end = timeit.default_timer()
>>> end-start
0.1954255269999976
```

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,116 @@
...
+
+        keys = []
+        indexed_values = []
+        tags = message["tags"]
+        assert isinstance(tags, Mapping), "Invalid tags type"
+
+        for key, value in sorted(tags.items()):
+            assert key.isdigit() and isinstance(value, int), "Tag key/value invalid"
+            keys.append(int(key))
+            indexed_values.append(value)
```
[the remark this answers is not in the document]
onewland: yeah, that's right

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,130 @@
...
+    def _process_values(self, message: Mapping[str, Any]) -> Mapping[str, Any]:
+        raise NotImplementedError
+
+    #
+    # This is mainly split out from _hash_timeseries_id for unit-testing purposes
+    #
+    def _timeseries_id_token(self, message: Mapping[str, Any]) -> bytearray:
+        field_separator: int = ord(";")
+        kv_separator: int = ord(",")
+        kv_assign: int = ord("=")
```
fpacifici, line 40: What about making these constants? The value does not change for very message

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,130 @@
...
+        org_id: int = message["org_id"]
+        project_id: int = message["project_id"]
+        metric_id: int = message["metric_id"]
+        sorted_tag_items: Sequence[Tuple[str, int]] = sorted(message["tags"].items())
```
fpacifici, line 44: This is sorted in another place as well. (line 90). 
Please sort them once and use the sorted version.

snuba/datasets/generic_metrics_processor.py
```diff
@@ -0,0 +1,130 @@
...
+
+        buffer = bytearray()
+        for field in [org_id, project_id, metric_id]:
+            buffer.extend(field.to_bytes(length=4, byteorder="little"))
+            buffer.append(field_separator)
```
fpacifici, line 49: Up to you, but this is an opaque identifier, The separators do nothing useful. You are not trying to use it to establish uniqueness, in such case you would not want ambiguity. 
As long as the same time series always yields the same id and they are reasonably evenly distributed you are fine.

fpacifici (approved):

PR #2878, opened 2022-06-22
feat(consumers): Experimental Consumers

rahul-kumar-saini:
### Overview
- We want the ability to mark a consumer as experimental so new consumers can crash on deploy without blocking the entire deploy
- By default, the k8s health check that blocks the deploy if a consumer crashes is just a simple check of whether or not the container is still running
- We need the consumer to be able to crash without the container itself restarting

### Changes
- Wrapped the entire `snuba consumer` CLI command with a try/except that just starts an infinite loop to keep the process from crashing
- Introduced setting which can contain consumer groups to be marked experimental - in turn marking the consumers experimental

### Tests
- Tested locally, marking a consumer group experimental and starting a broken consumer will log but not kill the process

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -1,5 +1,6 @@
 import logging
 import signal
+import time
 from typing import Any, Optional, Sequence
 
 import click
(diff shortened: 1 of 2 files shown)


snuba/cli/consumer.py
```diff
@@ -133,56 +134,77 @@ def consumer(
...
+        logger.info("Keeping process alive to avoid crash loop...")
+
+        def handler(signum: int, frame: Any) -> None:
+            exit()
+
+        signal.signal(signal.SIGINT, handler)
+        signal.signal(signal.SIGTERM, handler)
+
+        while True:
+            time.sleep(10000)
```
onewland, line 210: maybe we can emit a metric here for `broken-consumer-keepalive-loop`, tagged with the storage, so we have additional visibility on top of the Sentry app

snuba/cli/consumer.py
```diff
@@ -133,56 +134,77 @@ def consumer(
...
+        def handler(signum: int, frame: Any) -> None:
+            consumer.signal_shutdown()
+
+        signal.signal(signal.SIGINT, handler)
+        signal.signal(signal.SIGTERM, handler)
+
+        consumer.run()
+
+    except Exception:
+        if consumer_group not in settings.EXPERIMENTAL_CONSUMER_GROUPS:
```
onewland, line 192: Is there a reason to key this by `consumer_group` rather than `storage_key`?

Nice thing about using `storage_key` is that it's validated/constrained by the enum

snuba/cli/consumer.py
```diff
@@ -133,56 +134,77 @@ def consumer(
     profile_path: Optional[str] = None,
     cooperative_rebalancing: bool = False,
 ) -> None:
+    try:
```
onewland, line 137: could the try be around `consumer.run()` only?

onewland (commented):
Mostly good with this but I'd like answers to my questions, especially whether the `try` block could be much more contained

snuba/cli/consumer.py
rahul-kumar-saini: I think currently a consumer group is 1-1 with a topic (so the group of consumers on that topic) and storage keys might not be. Only the writable storages spawn consumers but there are plenty of read only storages in that Enum.

snuba/cli/consumer.py
rahul-kumar-saini: I had it this way originally, but creating the `consumer_builder` object itself could crash the startup of the consumer. We basically want the entire logic of running the `snuba consumer` CLI captured in case anything goes wrong.

snuba/cli/consumer.py
onewland: I was thinking once per loop iteration, so we can quickly see where we have uselessly spinning hosts

snuba/cli/consumer.py
onewland: makes sense

snuba/cli/consumer.py
rahul-kumar-saini: I think what I have now should do it, it will increment when a consumer crashes and decrement when the pod is exited with SIGINT/SIGTERM. On DataDog we can see if there's any non zero value, there is a uselessly spinning host.

snuba/cli/consumer.py
fpacifici: I think Oliver is right. What identifies a consumer processor in Snuba is the storage, not the consumer group. If you have two consumer on the same storage, they would share the processor so there should be no reason to allow some consumer for a storage to be experimental and other not to.
I would revert to using storages.

snuba/settings/__init__.py
```diff
@@ -208,6 +208,9 @@
 # Enable replays ingestion
 ENABLE_REPLAYS_CONSUMER = os.environ.get("ENABLE_REPLAYS_CONSUMER", False)
 
+# Used to tag consumers as experimental to avoid them crash looping in production
+EXPERIMENTAL_CONSUMER_GROUPS: Set[str] = set()
```
fpacifici, line 212: Why a dedicated setting? We already have experimental datasets, I would say by default, all storages and consumers of an experimental dataset should be experimental.

In the same way you should not have a consumer marked as experimental for a dataset that is marked as production. Can we use the same setting ?

snuba/cli/consumer.py
```diff
@@ -133,56 +134,83 @@ def consumer(
...
+        if consumer_group not in settings.EXPERIMENTAL_CONSUMER_GROUPS:
+            raise
+
+        tags = {"consumer_group": consumer_group, "storage": storage_key.value}
+        KEEPALIVE_METRIC = "broken_consumer_keepalive"
+
+        logger.exception(
+            "An experimental consumer crashed!",
+            extra=tags,
+        )
```
fpacifici, line 201: Please provide information about the Exception that caused the crash by passing `exc_info`

snuba/cli/consumer.py
```diff
@@ -133,56 +134,83 @@ def consumer(
...
+
+        logger.info(f"Experimental consumer for {consumer_group} has crashed!")
+        logger.info("Keeping process alive to avoid crash loop...")
```
fpacifici, line 204: Please do not log these. You just logged the exception at the line above that captures all the context. These do not add any context and would actually clutter your logs.

snuba/cli/consumer.py
```diff
@@ -133,56 +134,83 @@ def consumer(
...
+
+        def handler(signum: int, frame: Any) -> None:
+            metrics.increment(KEEPALIVE_METRIC, -1, tags)
+            exit()
+
+        signal.signal(signal.SIGINT, handler)
+        signal.signal(signal.SIGTERM, handler)
```
fpacifici, line 211: Why do you need this again ?

snuba/cli/consumer.py
fpacifici: `time.sleep(10000)` sleeps for 10,000 seconds. Are you sure you did not want to sleep for 10 seconds ?

fpacifici (changes requested):
I am not sure this approach will work well (the idea of preventing a consumer from restarting). 
Examples where this can have a problem:
- any transient kafka or clickhouse availability issue. These happen all the time, and we rely on the consumer to restart and re-establish the connection. With this approach, an experimental consumer would just hang in those cases instead of recovering. I don't think this would be an acceptable behavior even for experimental consumers. IT would introduce a ton of manual work that is not needed. 
- It would not cover crashes during the startup of the python interpreter (like settings validation). 

Can we let the consumer crash and have a quick and cheap process to exclude it from deployment manually without review when that happens ? Or excluding that only after it has been crashlooping for a while ?

fpacifici:
Regarding your alternatives 

> Create a new Step type in Freight to do everything KubernetesDeployment does except the k8s health check
To mark consumer experimental, modify freight.yaml
Unblocks deploy but doesn't stop the consumer from crash looping

I think I like this option. Letting the consumer crashloop without blocking the deployment seems a good approach. 
It would not impose a different process to troubleshoot a consumer nor any specific alert for those consumers than what people already have set it up.
Also it is resilient to a consumer crashing once or twice for specific transient errors.

evanh:
> Was there any downside ?

The downsides were:
1) It moves the "experimental" setting out of snuba, in the sense that the snuba application is not aware of what is experimental or not. So we would solve the very specific problem of allowing deploys, but we would have no room to improve/expand on this in the future.
2) It requires making changes to Freight itself, which is not trivial.
3) It adds another way datasets need to be configured. At least marking things in the settings file is an established convention.

rahul-kumar-saini:
Closing as this solution isn't the best

PR #2880, opened 2022-06-23
fix: Update Flask

evanh:
This required updating a few other packages as well in order to get
everything working.


onewland:
given the risk of a major version upgrade here, did you try running `snuba api` and issuing some basic requests against it?

evanh:
I tested locally and it works.

onewland (approved):
looks good to me

PR #2885, opened 2022-06-24
ref(subscriptions): Add partition_count for executors

MeredithAnya:
**context**
As per comment: https://github.com/getsentry/snuba/pull/2761#pullrequestreview-1017852897, we cannot hardcode `64` as the number of partitions across all consumers since that is not the case. 

The executors consume from the `scheduled-subscriptions-*` topics. So in order for this PR to pass the correct number of partitions, we'd need to add the following to the production settings `TOPIC_PARTITION_COUNTS`:

```
{
    "scheduled-subscriptions-events": 32,
    "scheduled-subscriptions-transactions": 32,
    "scheduled-subscriptions-metrics": 16,
}
```


MeredithAnya:
closing in favor of https://github.com/getsentry/snuba/pull/2888

PR #2888, opened 2022-06-28
ref(consumers): Add get_partition_count

MeredithAnya:
**context**
Okay so for the last piece of https://github.com/getsentry/snuba/pull/2761, we need to get the total partition count to use in our calculation for the max_concurrent_queries.

As stated in this comment https://github.com/getsentry/snuba/pull/2761#discussion_r905614595, there are a couple ways to go about this:

1. Keep track of the total number of partitions per topic in settings (implemented in https://github.com/getsentry/snuba/pull/2885)
2. Use the kafka admin api to get the number of partitions (as is implemented in this PR)

For this PR I was thinking that I would just log the metrics for `partition_count` to verify this works, but also to see what the impact of making these requests to the kafka admin is. I'm assuming it shouldn't be too much of an impact since this would only be when the consumers are starting up. If this ends up being a reasonable approach I'll modify https://github.com/getsentry/snuba/pull/2761 to use `get_partition_count` and pass through the result to the strategy

### DIFF
--- a/snuba/consumers/utils.py
+++ b/snuba/consumers/utils.py
@@ -8,8 +8,7 @@ from snuba.utils.streams.configuration_builder import get_default_kafka_configur
 from snuba.utils.streams.topics import Topic
 
 
-def get_partition_count_from_kafka_admin(topic: Topic, logger: Logger) -> int:
-
+def get_partition_count(topic: Topic, logger: Logger) -> int:
     override_params = {
         # Default is 60s, do we need that long?
         "socket.timeout.ms": 2000,
--- a/snuba/consumers/utils.py
+++ b/snuba/consumers/utils.py
@@ -1,0 +1,52 @@
+import time
+from logging import Logger
+
+from confluent_kafka import KafkaException
+from confluent_kafka.admin import AdminClient
+
+from snuba.utils.streams.configuration_builder import get_default_kafka_configuration
+from snuba.utils.streams.topics import Topic
+
+
+def get_partition_count_from_kafka_admin(topic: Topic, logger: Logger) -> int:
+
+    override_params = {
+        # Default is 60s, do we need that long?
+        "socket.timeout.ms": 2000,
+    }
+    attempts = 0
+    while True:
+        try:
+            logger.info("Attempting to connect to Kafka (attempt %d)...", attempts)
+            client = AdminClient(
+                get_default_kafka_configuration(
+                    topic=topic, override_params=override_params
+                )
+            )
+            cluster_metadata = client.list_topics(topic=topic.value, timeout=1.0)
+            break
+        except KafkaException as err:
+            logger.debug(
+                "Connection to Kafka failed (attempt %d)", attempts, exc_info=err
+            )
+            attempts += 1
+            # How many attempts is too many?
+            if attempts == 3:
+                raise
+            time.sleep(1)
+
+    try:
+        logger.info(f"Getting topic metadata for {topic.value}...")
+        topic_metadata = cluster_metadata.topics[topic.value]
+    except KeyError:
+        logger.info(f"No topic found for: {topic.value}")
+
+    if topic_metadata.error:
+        # (TODO) handle if there is a topic error, value is KafkaError object
+        # although not sure if this is necessary give the KafkaException
+        # handling above.
+        pass
+
+    total_partition_count = len(topic_metadata.partitions.keys())
+    logger.info(f"Total of {total_partition_count} partition(s) for {topic.value}.")
+    return total_partition_count
--- a/snuba/consumers/utils.py
+++ b/snuba/consumers/utils.py
@@ -15,20 +15,15 @@ class InvalidTopicName(Exception):
 
 
 def get_partition_count(topic: Topic) -> int:
-    override_params = {
-        # Default is 60s, do we need that long?
-        "socket.timeout.ms": 2000,
-    }
+    if not isinstance(topic, Topic):
+        raise InvalidTopicName("Must provide a valid Topic.")
+
     attempts = 0
     while True:
         try:
             logger.info("Attempting to connect to Kafka (attempt %d)...", attempts)
-            client = AdminClient(
-                get_default_kafka_configuration(
-                    topic=topic, override_params=override_params
-                )
-            )
-            cluster_metadata = client.list_topics(topic=topic.value, timeout=2.0)
+            client = AdminClient(get_default_kafka_configuration(topic=topic))
+            cluster_metadata = client.list_topics(timeout=2.0)
             break
         except KafkaException as err:
             logger.debug(
@@ -44,7 +39,7 @@ def get_partition_count(topic: Topic) -> int:
     topic_metadata = cluster_metadata.topics.get(topic.value)
 
     if not topic_metadata:
-        raise InvalidTopicName(f"Topic {topic.value} was not found")
+        raise InvalidTopicName(f"Topic {topic.value} was not found.")
 
     if topic_metadata.error is not None:
         raise KafkaException(topic_metadata.error)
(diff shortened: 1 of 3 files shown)


snuba/consumers/utils.py
```diff
@@ -0,0 +1,52 @@
...
+from logging import Logger
+
+from confluent_kafka import KafkaException
+from confluent_kafka.admin import AdminClient
+
+from snuba.utils.streams.configuration_builder import get_default_kafka_configuration
+from snuba.utils.streams.topics import Topic
+
+
+def get_partition_count_from_kafka_admin(topic: Topic, logger: Logger) -> int:
```
lynnagara, line 11:
```suggestion
def get_partition_count(topic: Topic) -> int:
```

I think `get_partition_count` is a better name. The fact that we are using an admin client is an implementation detail that I don't think the caller of this function should worry about.

Why do you need to pass the logger into this function?

MeredithAnya:
Things I'm unsure about:

* What is an appropriate `"socket.timeout.ms"`?
* How many attempts should we be allowing?
* It looks like the [TopicMetadata](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.admin.TopicMetadata) itself can have an error but I'm not quite sure when that would be that would not cause the earlier `KafkaException` - so just not sure if we should be re raising there as well?

snuba/consumers/utils.py
```diff
@@ -0,0 +1,52 @@
...
+            # How many attempts is too many?
+            if attempts == 3:
+                raise
+            time.sleep(1)
+
+    try:
+        logger.info(f"Getting topic metadata for {topic.value}...")
+        topic_metadata = cluster_metadata.topics[topic.value]
+    except KeyError:
+        logger.info(f"No topic found for: {topic.value}")
```
lynnagara, line 42: Are you sure you want to quietly swallow the exception here?

snuba/subscriptions/executor_consumer.py
```diff
@@ -106,6 +107,22 @@ def get_topics_for_entity(
         strict_offset_reset=strict_offset_reset,
     )
 
+    try:
+        partition_count = get_partition_count_from_kafka_admin(
+            scheduled_topic_spec.topic, logger
+        )
+    except Exception:
```
lynnagara, line 114: What exception is this meant to be catching? That function is never expected to raise an error

MeredithAnya: If we exceed the number of attempts we re raise the error

snuba/consumers/utils.py
MeredithAnya: cool yeah I can simplify the name. 

> Why do you need to pass the logger into this function?

I figured this would retain the logger instance associated with the specific consumer if this gets used by more than the executor consumers so we'd know if it was the executor, scheduler etc. Maybe that's not the best way to do that?

snuba/subscriptions/executor_consumer.py
lynnagara: Ah ok, didn't see that earlier

snuba/consumers/utils.py
MeredithAnya: Actually yeah that should probs raise an error, esp cause right now it will give an `UnboundLocalError: local variable 'topic_metadata' referenced before assignment`

snuba/consumers/utils.py
lynnagara: We don't typically use the logger name for that. Otherwise we would need pass the logger around to every function and that seems like a pain. Sentry always tags the server_name anyway. And if you were looking at the raw server logs well, I feel like you should probably know which server you are looking at anyway.

snuba/consumers/utils.py
```diff
@@ -0,0 +1,54 @@
...
+from snuba.utils.streams.topics import Topic
+
+logger = logging.getLogger("snuba.consumer.utils")
+
+
+class InvalidTopicName(Exception):
+    pass
+
+
+def get_partition_count(topic: Topic) -> int:
```
volokluev, line 17: can we add a test? Also, if this somehow fails, won't the consumer just not work?

MeredithAnya: > can we add a test?

yeah I can do that

> Also, if this somehow fails, won't the consumer just not work?

Right now I wrap this in a try except for any exception, and if it does fail we just log that to sentry. Since Im not sure what the right balance is for the timeout values, I figured this would be a safe way to test it out first.

snuba/consumers/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    override_params = {
+        # Default is 60s, do we need that long?
+        "socket.timeout.ms": 2000,
```
fpacifici, line 20: Why changing this ? If you cannot connect to the broker likely your consumer will not work anyway. I think the timeout that matters is the one you correctly set at line 31.

snuba/consumers/utils.py
```diff
@@ -0,0 +1,54 @@
...
+            logger.debug(
+                "Connection to Kafka failed (attempt %d)", attempts, exc_info=err
+            )
+            attempts += 1
+            # How many attempts is too many?
+            if attempts == 3:
+                raise
+            time.sleep(1)
+
+    logger.info(f"Checking topic metadata for {topic.value}...")
```
fpacifici, line 43: I think you may want to move this line above the loop, so that, if the consumer hangs, you know what it was trying to do.

snuba/consumers/utils.py
```diff
@@ -0,0 +1,48 @@
...
+logger = logging.getLogger("snuba.consumer.utils")
+
+
+class InvalidTopicName(Exception):
+    pass
+
+
+def get_partition_count(topic: Topic) -> int:
+    if not isinstance(topic, Topic):
+        raise InvalidTopicName("Must provide a valid Topic.")
```
lynnagara, line 19: Isn't this redundant since the type checker would prevent this from being anything other than a Topic?

lynnagara:
Does this PR supercede https://github.com/getsentry/snuba/pull/2885 or is that one still needed?

MeredithAnya:
> Does this PR supercede #2885 or is that one still needed?

yeah this one supercedes it, assuming this works fine when in prod

snuba/consumers/utils.py
MeredithAnya: removed it

snuba/consumers/utils.py
```diff
@@ -0,0 +1,45 @@
...
+                "Connection to Kafka failed (attempt %d)", attempts, exc_info=err
+            )
+            attempts += 1
+            # How many attempts is too many?
+            if attempts == 3:
+                raise
+            time.sleep(1)
+
+    if not topic_metadata:
+        raise TopicNotFound(f"Topic {topic.value} was not found.")
```
nikhars, line 38:
```suggestion
        raise TopicNotFound(f"Topic metadata for {topic.value} was not found.")
```

snuba/subscriptions/executor_consumer.py
```diff
@@ -106,6 +107,20 @@ def get_topics_for_entity(
         strict_offset_reset=strict_offset_reset,
     )
 
+    try:
+        partition_count = get_partition_count(scheduled_topic_spec.topic)
+    except Exception:
+        logger.error("partition count unavailable..", exc_info=True)
+        partition_count = 0
```
nikhars, line 114: Is partition_count 0 a good default to fall back on? Or should we have some other default? 

In failure scenarios where a consumer is not able to find out the partition count, is the expected behavior going to be that the consumer will not not run at all?

MeredithAnya: @nikhars I think for the purpose of this PR this is probably fine since we aren't using the `partition_count` yet, we are just logging it. But I think that if we can't get the partition_count that would be pretty problematic in that we couldn't accurately calculate the max_concurrent_queries.

PR #2894, opened 2022-06-29
feat(profiling): Add functions consumer and dataset

Zylphrex:
This adds the new functions dataset and introduces a consumer for it.

### DIFF
--- a/snuba/datasets/functions.py
+++ b/snuba/datasets/functions.py
@@ -1,0 +1,7 @@
+from snuba.datasets.dataset import Dataset
+from snuba.datasets.entities import EntityKey
+
+
+class FunctionsDataset(Dataset):
+    def __init__(self) -> None:
+        super().__init__(default_entity=EntityKey.FUNCTIONS)
--- a/snuba/datasets/functions_processor.py
+++ b/snuba/datasets/functions_processor.py
@@ -1,0 +1,48 @@
+import uuid
+from datetime import datetime
+from typing import Any, Mapping, Optional
+
+from snuba.consumers.types import KafkaMessageMetadata
+from snuba.processor import InsertBatch, MessageProcessor, ProcessedMessage
+
+
+class FunctionsMessageProcessor(MessageProcessor):
+    def process_message(
+        self, message: Mapping[str, Any], metadata: KafkaMessageMetadata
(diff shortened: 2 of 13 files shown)


snuba/datasets/functions.py
```diff
@@ -0,0 +1,7 @@
+from snuba.datasets.dataset import Dataset
+from snuba.datasets.entities import EntityKey
+
+
+class FunctionsDataset(Dataset):
```
evanh, line 5: Can you make this dataset experimental please?

snuba/datasets/functions_processor.py
```diff
@@ -0,0 +1,56 @@
...
+from snuba.processor import InsertBatch, MessageProcessor, ProcessedMessage
+
+
+class FunctionsMessageProcessor(MessageProcessor):
+    def process_message(
+        self, message: Mapping[str, Any], metadata: KafkaMessageMetadata
+    ) -> Optional[ProcessedMessage]:
+        functions = {}
+
+        for thread, root_frames in message["call_trees"].items():
```
evanh, line 15: Is there a cap on the number of `root_frames` in a `call_tree`? And a cap on the children frames? This function is going to get slow very quickly if there isn't some reasonable cap.

snuba/datasets/storages/functions.py
```diff
@@ -0,0 +1,93 @@
...
+                Column("profile_id", UUID()),
+                Column("materialization_version", UInt(8)),
+            ]
+        ),
+    ),
+    query_processors=[],
+    stream_loader=build_kafka_stream_loader_from_settings(
+        processor=FunctionsMessageProcessor(),
+        default_topic=Topic.PROFILES_FUNCTIONS,
+    ),
```
rahul-kumar-saini, line 63: Do you want to add a DLQ? (It would catch exceptions raised by the message processor and you can choose what to do with invalid messages so the consumer doesn't crash loop)

Here's an example stream loader configured with a DLQ:
https://github.com/getsentry/snuba/blob/c033f975aef88f69eaa1a405acf8285507b1e413/snuba/datasets/storages/generic_metrics.py#L166-L170

If you'd like invalid messages to be forwarded to a new dead-letter topic for this consumer, please leave a message in #discuss-eng-sns and we can set up/configure.

PR #2896, opened 2022-06-29
feat(metrics): expose generic_metrics dataset, generic_metrics_sets entity, add e2e test

onewland:
I wanted to make sure we didn't have regressions in terms of metrics_sets consumption, which required adding an e2e test, which itself required exposing the metrics_sets dataset. 

The `GENERIC_METRICS_SETS` entity is not ready for prod traffic, but I didn't want to blow up the scope of this PR. If people prefer, I can increase the scope of this PR

Next steps:
1. getting proper granularity processing and validation on queries
1. adding rate limiting for queries
1. proper array join optimizations (specifically because tags now have raw and indexed values)

I think this can be used by clients once next steps are complete

### DIFF
--- a/snuba/datasets/entities/generic_metrics.py
+++ b/snuba/datasets/entities/generic_metrics.py
@@ -1,0 +1,63 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import (
+    AggregateFunction,
+    Column,
+    ColumnSet,
+    DateTime,
+    Nested,
+    UInt,
+)
+from snuba.clickhouse.translators.snuba.mappers import (
+    FunctionNameMapper,
+    SubscriptableMapper,
+)
+from snuba.clickhouse.translators.snuba.mapping import TranslationMappers
+from snuba.datasets.entity import Entity
+from snuba.datasets.plans.single_storage import SingleStorageQueryPlanBuilder
+from snuba.datasets.storages.generic_metrics import sets_bucket_storage, sets_storage
+from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
+from snuba.query.processors import QueryProcessor
+from snuba.query.validation.validators import EntityRequiredColumnValidator
+
+
+class GenericMetricsSetsEntity(Entity):
+    READABLE_STORAGE = sets_storage
+    WRITABLE_STORAGE = sets_bucket_storage
+
+    def __init__(self) -> None:
+        super().__init__(
+            storages=[sets_bucket_storage, sets_storage],
+            query_pipeline_builder=SimplePipelineBuilder(
+                query_plan_builder=SingleStorageQueryPlanBuilder(
+                    self.READABLE_STORAGE,
+                    mappers=TranslationMappers(
+                        subscriptables=[
+                            SubscriptableMapper(None, "tags", None, "tags"),
+                        ],
+                        functions=[
+                            FunctionNameMapper("uniq", "uniqCombined64Merge"),
+                            FunctionNameMapper("uniqIf", "uniqCombined64MergeIf"),
+                        ],
+                    ),
+                )
+            ),
+            abstract_column_set=ColumnSet(
+                [
+                    Column("org_id", UInt(64)),
+                    Column("project_id", UInt(64)),
+                    Column("metric_id", UInt(64)),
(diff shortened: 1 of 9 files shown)


tests/test_generic_metrics_api.py
```diff
@@ -0,0 +1,139 @@
...
+class TestGenericMetricsApiSets(BaseApiTest):
+    placeholder_counter = 0
+
+    @pytest.fixture
+    def test_app(self) -> Any:
+        return self.app
+
+    @pytest.fixture
+    def test_entity(self) -> Union[str, Tuple[str, str]]:
+        return "metrics_counters"
```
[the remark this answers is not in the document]
evanh: Although I'm not sure what this fixture is doing anymore. I remember adding it when I was testing legacy/snql at the same time, but I don't know why I left it.

evanh (approved):

snuba/datasets/generic_metrics.py
```diff
@@ -0,0 +1,16 @@
...
+
+from snuba.datasets.dataset import Dataset
+from snuba.datasets.entities import EntityKey
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.entity import Entity
+
+DEFAULT_GRANULARITY = 60
+
+
+class GenericMetricsDataset(Dataset):
```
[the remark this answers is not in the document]
onewland: yeah, I was having trouble getting that working but I'll try again, I agree that makes sense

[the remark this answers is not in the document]
onewland: yeah, I don't think this can work with e2e testing without major changes. The application is initialized in `BaseApiTest`, and at that point `settings.ENABLE_DEV_FEATURES` is already False (so we get a test error for the `generic_metrics` dataset not being defined). I tried setting it to true before the `import application` line here: https://github.com/getsentry/snuba/blob/8a31d3066810d5ed55276dbfb16c008a1720f17f/tests/base.py#L6

but I think something else may have already imported the dataset factory elsewhere and `generic_metrics` is still not set in [DATASET_NAMES](https://github.com/getsentry/snuba/blob/8a31d3066810d5ed55276dbfb16c008a1720f17f/snuba/datasets/factory.py#L13)

snuba/datasets/entities/generic_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+from snuba.clickhouse.translators.snuba.mapping import TranslationMappers
+from snuba.datasets.entity import Entity
+from snuba.datasets.plans.single_storage import SingleStorageQueryPlanBuilder
+from snuba.datasets.storages.generic_metrics import sets_bucket_storage, sets_storage
+from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
+from snuba.query.processors import QueryProcessor
+from snuba.query.validation.validators import EntityRequiredColumnValidator
+
+
+class GenericMetricsSetsEntity(Entity):
```
fpacifici, line 24: How are you going to manage the granularity in this dataset ?
https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/metrics.py#L112

The product is used to provide the granularity in seconds. Now granularity are discrete values instead. Are we going to solve the discrepancy here or in the product ?

onewland: I was planning to hide this detail from product and have them send granularity in seconds (https://getsentry.atlassian.net/browse/SNS-1480)

fpacifici: I think we should think twice before doing that as it is exactly against the idea of removing the logical data model from Snuba. 
The transformation from seconds to the specific granularity enum value is arguably a product concern, that would fit in the snuba logical data model as an entity query processor today, but which we should move away from.

Any chance to move this into Sentry ?

fpacifici: If it is a big deal we can prioritize deploying this first.

onewland: yeah, I'm going to deploy this as-is and we can finalize the granularity approach in the next few days

evanh: Late to the party, but +1 on making this a product concern instead of a hidden Snuba feature.

PR #2897, opened 2022-06-29
feat(replays): Add dead letter queue policy to storage stream-loader

cmanallen:
This pull adds a dead-letter-queue policy to the replays storage stream-loader.  I followed the instructions Rahul outlined here: https://github.com/getsentry/snuba/pull/2806#issuecomment-1155794695.

I'm under the impression that this is all I need but I'm new to the Snuba library.  Let me know if I missed something.


snuba/datasets/storages/replays.py
```diff
@@ -58,7 +66,14 @@
     storage_set_key=StorageSetKey.REPLAYS,
 )
 
-# TODO: set up deadletter queue for bad messages.
+
+def produce_policy_creator() -> DeadLetterQueuePolicy:
+    """Produce all bad messages to dead-letter topic."""
+    return ProduceInvalidMessagePolicy(
+        KafkaProducer(build_kafka_producer_configuration(Topic.DEAD_LETTER_METRICS)),
+        KafkaTopic(Topic.DEAD_LETTER_METRICS.value),
```
fpacifici, line 74: You will need your own topic. That is the Topic we use for the metrics consumer.
Please add one in the topics file here https://github.com/getsentry/snuba/blob/master/snuba/utils/streams/topics.py#L28-L31

fpacifici (changes requested):
Please register the missing topics. 
Once that is done you will need to ask the ops team to create the topic and add it to the production config. In order to achieve that, please contact the search and storage team in slack discuss-eng-sns.

rahul-kumar-saini:
Echoing Filippo's comment, you should create a new dead-letter topic by filing an [ops request](https://getsentry.atlassian.net/secure/CreateIssue.jspa?issuetype=10301&pid=10303) for it. It should be created on the same cluster as the replay events topic. We have a simple convention going `snuba-dead-letter-XXX`.

It can then be added to the production config [here](https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L205-L239).
Probably just something like this:
```
  'snuba-dead-letter-replay-events': {
    'bootstrap.servers': KAFKA_SPECTO_BOOTSTRAP_SERVERS,
  },
```

snuba/datasets/storages/replays.py
cmanallen: 👍

snuba/utils/streams/topics.py
```diff
@@ -29,6 +29,7 @@ class Topic(Enum):
     DEAD_LETTER_METRICS = "snuba-dead-letter-metrics"
     DEAD_LETTER_SESSIONS = "snuba-dead-letter-sessions"
     DEAD_LETTER_GENERIC_METRICS = "snuba-dead-letter-generic-metrics"
+    DEAD_LETTER_REPLAYS = "snuba-dead-letter-replays"
```
fpacifici, line 32: You will also need to add the topic to the config validation rules.
https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py#L21-L46

cmanallen: @fpacifici Forgot to push after my last comment.  Should be good to go.

rahul-kumar-saini (approved):
LGTM pending https://github.com/getsentry/ops/pull/5018

cmanallen:
@fpacifici Is this ready?

fpacifici (approved):

PR #2901, opened 2022-06-30
fix: Fix unbound local error if blacklist is used

untitaker:
This doesn't look right. If the blacklist is used, this function will crash for those projects


untitaker:
nevermind can't read

PR #2902, opened 2022-06-30
feat: Upgrade mypy and split requirements files

evanh:
Split linting out of the main requirements, and upgrade mypy. That required some
typing fixes and changes to the mypy.ini


snuba/migrations/parse_schema.py
```diff
@@ -72,18 +72,21 @@ def merge_modifiers(
         return col_type.set_modifiers(existing_modifiers.merge(modifiers))
 
 
+COLUMN_MAPPERS: Mapping[str, type[ColumnType[MigrationModifiers]]] = {
```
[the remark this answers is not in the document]
evanh: Since the dictionary was being built every time, it wasn't possible to give it a type without creating a variable for it that could then have a type. If I was going to do that, I might as well move it out of the function so the dictionary isn't being rebuilt over and over.

Dockerfile
```diff
@@ -74,5 +74,5 @@ CMD [ "api" ]
 FROM application AS testing
 
 USER 0
-RUN pip install -r requirements-test.txt
+RUN pip install -r test-requirements.txt
```
lynnagara, line 77: Why changing the file name?

snuba/consumers/consumer.py
```diff
@@ -68,7 +68,11 @@ class BytesInsertBatch(NamedTuple):
...
 
-    def __reduce_ex__(
+    # TODO: Argument 1 of "__reduce_ex__" is incompatible with supertype "object"; supertype defines the argument type as "SupportsIndex"
+    # This is coming from the type hints builtins.pyi:
+    # if sys.version_info >= (3, 8):
+    #    def __reduce_ex__(self, __protocol: SupportsIndex) -> str | tuple[Any, ...]: ...
+    def __reduce_ex__(  # type: ignore
         self, protocol: int
     ) -> Tuple[Any, Tuple[Sequence[Any], Optional[datetime]]]:
         if protocol >= 5:
```
lynnagara, line 78: Still not ideal but how about:

```suggestion
    def __reduce_ex__(
        self, protocol: SupportsIndex
    ) -> Tuple[Any, Tuple[Sequence[Any], Optional[datetime]]]:
        if int(protocol) >= 5:
```

snuba/request/validation.py
```diff
@@ -85,10 +85,11 @@ def build_request(
                 }
                 query_settings["referrer"] = referrer
                 # TODO: referrer probably doesn't need to be passed in, it should be from the body
+                # this should fix the typing problem
```
lynnagara, line 88: What does this mean?

snuba/utils/metrics/wrapper.py
```diff
@@ -28,7 +28,9 @@ def __merge_tags(self, tags: Optional[Tags]) -> Optional[Tags]:
         elif tags is None:
             return self.__tags
         else:
-            return ChainMap(tags, self.__tags)
+            # TODO: ChainMap takes a MutableMapping, but both snuba and arroyo type tags as a Mapping.
+            # Fixing this requires updating both packages.
```
lynnagara, line 32: Is this actually an issue in snuba/arroyo or is it a mypy bug/limitation?

tests/test_api.py
```diff
@@ -150,7 +150,7 @@ def generate_fizzbuzz_events(self) -> None:
     def redis_db_size(self) -> int:
         # dbsize could be an integer for a single node cluster or a dictionary
         # with one key value pair per node for a multi node cluster
-        dbsize = redis_client.dbsize()
+        dbsize: int = redis_client.dbsize()
```
lynnagara, line 153: What's the `isinstance(dbsize, dict)` on the line below for? If this is an int why would that be needed?

snuba/consumers/consumer.py
evanh: Because then I get an error at `if int(protocol) >= 5`

Screen Shot 2022-07-13 at 3 50 05 PM

snuba/utils/metrics/wrapper.py
evanh: It's not a mypy bug, `ChainMap` is a core library that is typed to `MutableMapping`. I think the stubs got updated or something. So it is an "issue" with how the `Tags` are typed in snuba/arroyo, but it's just a minor typing issue.

snuba/consumers/consumer.py
lynnagara: I'm not sure I get why you would see this error if the LHS was being cast to an int.

snuba/utils/metrics/wrapper.py
lynnagara: Ok

tests/test_api.py
lynnagara: If the comment above was right then should the type be `int | dict`

Dockerfile
lynnagara: A bit of a nitpick but I prefer the requirements-* naming convention and we use it in most Sentry repos.

snuba/consumers/consumer.py
evanh: Me either, I couldn't figure it out.

Dockerfile
evanh: Renamed them.

snuba/consumers/consumer.py
lynnagara: This actually worked for me when I tried it locally. Mind double checking?

evanh:
@mwarkentin

PR #2903, opened 2022-06-30
feat(dependencies): Update Arroyo to 0.2.2

rahul-kumar-saini:
### Overview
- Upgrade Arroyo for DLQ bug fixes
    - DLQ would not catch errors on `join()`
    - DLQ exception repr was way too long, just displaying a list of messages in their entirety
- Also introduces `create_with_partitions` in processing strategies
    - https://github.com/getsentry/arroyo/pull/82


MeredithAnya (approved):
the `create_with_partitions` change looks fine to me

PR #2908, opened 2022-07-05
feat(mep): process generic-metrics distributions

onewland:
Verified locally with `snuba consumer --storage generic_metrics_distributions_raw --auto-offset-reset latest --consumer-group distributions-testing` and kafka-console-producer. Messages correctly make it to the aggregate table.

My plan is that when I add the entities for querying I'll add a full e2e test for distributions. If people prefer, I can roll that into this PR

### DIFF
--- a/.freight.yml
+++ b/.freight.yml
@@ -57,6 +57,8 @@ steps:
   - image: us.gcr.io/sentryio/snuba:{sha}
     name: metrics-consumer
   - image: us.gcr.io/sentryio/snuba:{sha}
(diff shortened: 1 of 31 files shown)


evanh (approved):
I'm OK with adding the e2e test in the next PR. Does this mean the table will start being populated once this PR is merged?

onewland:
> I'm OK with adding the e2e test in the next PR. Does this mean the table will start being populated once this PR is merged?

https://github.com/getsentry/ops/pull/4942 has to be merged to create the consumer instances

PR #2910, opened 2022-07-06
experiment(consumers): Experimental Consumers Freight Config

rahul-kumar-saini:
### Overview
- I'm working on extending Freight to support multiple config files per repo 
    - https://github.com/getsentry/freight/pull/303
- Each config file will require a separate Freight app to deploy
- I've added a new config file to act as a placeholder for "experimental" consumers
- Ideally, this config file is picked up by some `Snuba-experimental` freight app while the original freight config is run by the existing Snuba freight app
- I've just added the new generic metrics consumers into this config for now to test that this setup works

### DIFF
--- a/.freight.experimental.yml
+++ b/.freight.experimental.yml
@@ -12,11 +12,3 @@ steps:
         name: generic-metrics-sets-consumer
       - image: us.gcr.io/sentryio/snuba:{sha}
         name: generic-metrics-distributions-consumer
-  - kind: KubernetesCronJob
-    selector:
-      label_selector: service=snuba
-    containers:
-      - image: us.gcr.io/sentryio/snuba:{sha}
-        name: cleanup
-      - image: us.gcr.io/sentryio/snuba:{sha}
-        name: optimize


joshuarli (approved):

.freight.experimental.yml
```diff
@@ -0,0 +1,22 @@
...
+      - image: us.gcr.io/sentryio/snuba:{sha}
+        name: generic-metrics-distributions-consumer
+  - kind: KubernetesCronJob
+    selector:
+      label_selector: service=snuba
+    containers:
+      - image: us.gcr.io/sentryio/snuba:{sha}
+        name: cleanup
+      - image: us.gcr.io/sentryio/snuba:{sha}
+        name: optimize
```
fpacifici, line 22: Why are optimize and cleanup here ?

rahul-kumar-saini: I don't know what these do, what would happen if this step gets executed both in the regular consumers deployment and the experimental deployment?

I will remove it, just thought it was necessary when copying over the config.

evanh: They can be removed, they are for the cron jobs that automatically clean up Clickhouse partitions.

evanh (approved):

PR #2911, opened 2022-07-06
feat(mep): expose distributions entity, add e2e test

onewland:
1. creates the distributions entity for generic_metrics
2. some refactoring for shared logic of generic_metrics entities
3. e2e testing for the new entity

next steps:
1. getting rate limiters/granularity guard rails in place (should be mostly copy/paste from release health metrics)
2. fix the subscriptable mapper to work with tags (which has two value columns)

### DIFF
--- a/snuba/datasets/entities/generic_metrics.py
+++ b/snuba/datasets/entities/generic_metrics.py
@@ -82,6 +82,9 @@ class GenericMetricsEntity(Entity, ABC):
             required_time_column="timestamp",
         )
 
+    def get_query_processors(self) -> Sequence[QueryProcessor]:
+        return []
+
 
 class GenericMetricsSetsEntity(GenericMetricsEntity):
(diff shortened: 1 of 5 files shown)


snuba/datasets/entities/generic_metrics.py
```diff
@@ -13,51 +14,157 @@
     SubscriptableMapper,
 )
 from snuba.clickhouse.translators.snuba.mapping import TranslationMappers
+from snuba.datasets.entities.metrics import (
+    AggregateCurriedFunctionMapper,
+    AggregateFunctionMapper,
+)
```
onewland, line 20: not sure if we consider this bad practice -- these couldn't directly go into the standard mappers file because of mypy issues but I didn't spend a lot of time trying to figure out the root cause.

fpacifici (approved):

PR #2912, opened 2022-07-07
fix(profiling): Disable vertical merge algorithm

phacops:
Since our `profile` column is so big, we have merge issues. Switching horizontal merges helps with this issue.

### DIFF
--- a/snuba/migrations/snuba_migrations/profiles/0002_disable_vertical_merge_algorithm.py
+++ b/snuba/migrations/snuba_migrations/profiles/0002_disable_vertical_merge_algorithm.py
@@ -1,0 +1,31 @@
+from typing import Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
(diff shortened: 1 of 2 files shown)


snuba/migrations/snuba_migrations/profiles/0002_disable_vertical_merge_algorithm.py
```diff
@@ -0,0 +1,31 @@
...
+            operations.RunSql(
+                storage_set=StorageSetKey.PROFILES,
+                statement=(
+                    "ALTER TABLE profiles_local MODIFY SETTINGS enable_vertical_merge_algorithm=0"
+                ),
+            )
+        ]
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return []
```
fpacifici, line 25: Please add the backwards migration:
`ALTER TABLE RESET SETTING enable_vertical_merge_algorithm`

fpacifici (approved):
Could you please add the backwards migration before merging ?

fpacifici (commented):
`RESET SETTING` was introduced only in 21.8 so we cannot use it in a migration.
Please feel free to move ahead and merge this version.

PR #2913, opened 2022-07-07
feat(replays): add replays dataset to dataset factory

JoshFerge:
- adding here as I noticed its needed for the SnQL UI page, not sure there are any other ramifications I should know about.


evanh (approved):
> not sure there are any other ramifications I should know about.

Not that I can think of.

PR #2916, opened 2022-07-08
feat(replacements) Script to execute batches of replacements

fpacifici:
This script is for disaster recovery scenarios where we need to temporarily skip a number of replacements and we dump them into a file.

The script scans the file and re-executes them one by one using the same logic as the replacement consumer.
It provides a dry run mode, and an offset mode to skip part of the file. 
The format of the file must be equivalent to the one from Kafka.

### DIFF
--- a/snuba/cli/offline_replacer.py
+++ b/snuba/cli/offline_replacer.py
@@ -1,0 +1,118 @@
+import logging
+import time
+from datetime import datetime
+from typing import Optional
+
+import click
+import simplejson as json
+from arroyo import Message
+from arroyo.backends.kafka import KafkaPayload
+from arroyo.types import Partition, Topic
+
+from snuba import environment
+from snuba.datasets.storages import StorageKey
+from snuba.datasets.storages.factory import get_writable_storage
+from snuba.environment import setup_logging, setup_sentry
+from snuba.replacer import ReplacerWorker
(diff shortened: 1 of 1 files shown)


snuba/cli/offline_replacer.py
```diff
@@ -0,0 +1,119 @@
...
+    file_name: str,
+    project_id: int,
+    dry_run: bool,
+    start_from: int,
+    delay: int,
+    log_level: Optional[str] = None,
+) -> None:
+    """
+    Executes a list of replacements taken from a file instead of taking
+    them from Kafka.
```
nikhars, line 69: Could you add an example of what the contents of the file might look like?

snuba/cli/offline_replacer.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    logger = logging.getLogger(__name__)
+    setup_logging(log_level)
+    setup_sentry()
+
+    metrics = MetricsWrapper(environment.metrics, "errors.replacer")
```
nikhars, line 75: I am not sure whether metrics would be reported when the script is run. But if they are going to be reported then maybe have a different prefix for metrics reported from script vs what is run by the consumer

nikhars (approved):
Left some minor comments. Else looks ok to me.

PR #2917, opened 2022-07-08
feat(replacements) Reimplement bypass properly

fpacifici:
Redo of https://github.com/getsentry/snuba/pull/2916.

This allows to bypass multiple projects, add proper logging and bypassess all replacements for a given project so we can preserve in order delivery


evanh (approved):
So is this just throwing replacements away? How do we reprocess them in this case?

onewland (approved):

fpacifici:
> So is this just throwing replacements away? How do we reprocess them in this case?

WIth this.
https://github.com/getsentry/snuba/pull/2916

But this is still an intermediate solution. In the next step I will add a dead letter queue so we can have a second consumer attempt those messages directly again and they would not impact the main queue

PR #2918, opened 2022-07-08
fix: Flaky redis test

evanh:
Digging into this test, everything seemed to be working correctly when the test
failed. My current theory is that there is a small flake in the test that ends
up with the same result.

Either the last slow worker can timeout and find an error in redis, or it can
time out and find nothing in redis. In the first case, it would surface an error
that would be correctly caught by the test. In the second case, it was returning
an error that wasn't correctly handled by the test. Both cases are valid for the
test (slow worker times out waiting for a result from the original query) but
slightly different exceptions.

Since the way `pytest.raises` was not correct, fixing it to correctly watch for
a type seems like a valid fix to me.


lynnagara (changes requested):
If the test is sometimes not raising the correct error (a redis error instead of the ExecutionTimeoutError) I feel that we should actually fix the implementation to always return the right error instead of changing the test to support returning any kind of exception.

onewland:
One alternative to Lyn's suggestion, which sounds like it might not be feasible for this code path, would be having a set of expected exceptions and we pass as long as the raised exception is in that set (of 2)

evanh:
OK I'll try to make this cleaner.

> the test is sometimes not raising the correct error

To be clear: the test is raising a correct error, I think the way pytest was catching the exception changed. In all of my testing, everything was working as expected in the cache.

tests/state/test_cache.py
```diff
@@ -213,7 +213,7 @@ def worker(timeout: int) -> bytes:
     with pytest.raises(TimeoutError):
         waiter_fast.result()
 
-    with pytest.raises(exc):
+    with pytest.raises((ExecutionTimeoutError, TigerExecutionTimeoutError)):
```
nikhars, line 216: Since you are testing against one of several fixed exceptions then you can remove the parametrized exc being passed in since its not used anymore.

onewland (approved):
looks good but I agree with Nikhar's comment, let's remove the unused parameter

tests/state/test_cache.py
onewland: or move the tuple into the parameterization

evanh:
@lynnagara Is this OK with you?

PR #2928, opened 2022-07-11
feat(mep): add commit log topics config

onewland:


evanh (approved):

onewland:
blocked by https://github.com/getsentry/ops/pull/4967

mwarkentin:
Added the topics to validation in a separate PR so that we can deploy this, then do https://github.com/getsentry/ops/pull/4967, and then finally roll out the rest of this PR.

https://github.com/getsentry/snuba/pull/2930

snuba/datasets/storages/generic_metrics.py
```diff
@@ -119,6 +119,7 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
         processor=GenericSetsMetricsProcessor(),
         default_topic=Topic.GENERIC_METRICS,
         dead_letter_queue_policy_creator=produce_policy_creator,
+        commit_log_topic=Topic.GENERIC_METRICS_SETS_COMMIT_LOG,
```
lynnagara, line 122: You may want to move this line (and the distributions equivalent) into a separate PR. So we create the topics first and have them in topics.py but then start actually producing to them later.

onewland: removing that now, and then I'll merge and then we can do the ops PR

lynnagara (approved):

PR #2929, opened 2022-07-11
feat(mep): do kafka-header filtering to avoid JSON overhead for generic-metrics

onewland:
Right now we parse the JSON of the incoming message and compare it against the `metric_type` for the specific consumer, but we can do a direct string compare of the header value and avoid JSON parsing altogether for many incoming messages.

### DIFF
--- a/snuba/datasets/storages/generic_metrics.py
+++ b/snuba/datasets/storages/generic_metrics.py
@@ -29,6 +29,7 @@ from snuba.datasets.generic_metrics_processor import (
     GenericSetsMetricsProcessor,
 )
 from snuba.datasets.message_filters import KafkaHeaderFilter
+from snuba.datasets.metrics_messages import InputType
 from snuba.datasets.schemas.tables import TableSchema, WritableTableSchema
 from snuba.datasets.storage import ReadableTableStorage, WritableTableStorage
 from snuba.datasets.storages import StorageKey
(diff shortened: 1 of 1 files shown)


rahul-kumar-saini:
How could I tell what currently exists in a metrics message header?

onewland:
> How could I tell what currently exists in a metrics message header?

Without doing another deploy to log a metric or something, I don't think I have permission to anything that lets me answer that question in prod.

Here's where we set the type: https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/consumers/indexer/processing.py#L226

we validate that it's not None and also in `ACCEPTED_TYPES`:
https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/consumers/indexer/processing.py#L127-L133

`ACCEPTED_TYPES` contains "s", "c", and "d": https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/consumers/indexer/processing.py#L21

I can test locally if you're concerned, but this is also not released to customers so it can be rolled back fairly safely.

snuba/datasets/storages/generic_metrics.py
```diff
@@ -119,6 +120,7 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
         processor=GenericSetsMetricsProcessor(),
         default_topic=Topic.GENERIC_METRICS,
         dead_letter_queue_policy_creator=produce_policy_creator,
+        pre_filter=KafkaHeaderFilter("metric_type", "s"),
```
lynnagara, line 123: Should we use `InputType.SET.value`?

onewland: if you think it's clearer, I'm happy to make that change

lynnagara: Well it's up to you, I'm not going to insist on it or anything. But isn't that why the enum exists?

lynnagara (approved):

evanh (approved):

PR #2930, opened 2022-07-12
Add topics to validation

mwarkentin:
Add new topics to settings validation list.


onewland (approved):

lynnagara:
The topic doesn't exist anywhere in code yet. I think you can bring the changes to topic.py from here https://github.com/getsentry/snuba/pull/2928 over here too. Basically everything except passing it to the stream loaders so nothing gets produced there yet.

onewland:
This will be done in other PRs

PR #2937, opened 2022-07-13
feat(mep): support time bucketing in queries

onewland:
add support for bucketed time and handling the mapping of `granularity : enum value` in query processing

@wmak @ahmedetefy for visibility

### DIFF
--- a/snuba/datasets/entities/generic_metrics.py
+++ b/snuba/datasets/entities/generic_metrics.py
@@ -30,7 +30,11 @@ from snuba.datasets.storages.generic_metrics import (
 )
 from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
 from snuba.query.processors import QueryProcessor
-from snuba.query.processors.granularity_processor import MappedGranularityProcessor
+from snuba.query.processors.granularity_processor import (
+    DEFAULT_MAPPED_GRANULARITY_ENUM,
+    PERFORMANCE_GRANULARITIES,
+    MappedGranularityProcessor,
+)
 from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
 from snuba.query.validation.validators import (
     EntityRequiredColumnValidator,
@@ -101,7 +105,10 @@ class GenericMetricsEntity(Entity, ABC):
     def get_query_processors(self) -> Sequence[QueryProcessor]:
         return [
             TagsTypeTransformer(),
-            MappedGranularityProcessor(),
+            MappedGranularityProcessor(
+                accepted_granularities=PERFORMANCE_GRANULARITIES,
+                default_granularity_enum=DEFAULT_MAPPED_GRANULARITY_ENUM,
+            ),
             TimeSeriesProcessor({"bucketed_time": "timestamp"}, ("timestamp",)),
         ]
 
--- a/snuba/datasets/entities/generic_metrics.py
+++ b/snuba/datasets/entities/generic_metrics.py
@@ -31,6 +31,7 @@ from snuba.datasets.storages.generic_metrics import (
 from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
 from snuba.query.processors import QueryProcessor
 from snuba.query.processors.granularity_processor import MappedGranularityProcessor
+from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
 from snuba.query.validation.validators import (
     EntityRequiredColumnValidator,
     QueryValidator,
@@ -98,7 +99,11 @@ class GenericMetricsEntity(Entity, ABC):
         )
 
     def get_query_processors(self) -> Sequence[QueryProcessor]:
-        return [TagsTypeTransformer(), MappedGranularityProcessor()]
+        return [
+            TagsTypeTransformer(),
+            MappedGranularityProcessor(),
+            TimeSeriesProcessor({"bucketed_time": "timestamp"}, ("timestamp",)),
+        ]
 
 
 class GenericMetricsSetsEntity(GenericMetricsEntity):
--- a/snuba/datasets/entities/generic_metrics.py
+++ b/snuba/datasets/entities/generic_metrics.py
@@ -30,6 +30,7 @@ from snuba.datasets.storages.generic_metrics import (
 )
 from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
 from snuba.query.processors import QueryProcessor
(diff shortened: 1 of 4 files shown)


snuba/query/processors/granularity_processor.py
```diff
@@ -39,3 +41,53 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
...
+    Use the granularity set on the query to filter on the granularity column,
+    supporting generic-metrics style enum mapping
+    """
+
+    def __init__(
+        self,
+        accepted_granularities: Sequence[Tuple[int, int]],
+        default_granularity_enum: int,
+    ):
+        self._accepted_granularities = accepted_granularities
```
lynnagara, line 61: Maybe I'm being paranoid but I'd probably throw `sorted()` around this just to make sure it's ordered as the next step depends on it. I know it's sorted now but you never know when we might want to add a granularity later and it won't be in order anymore. Could also go ahead and reverse it here so you don't need to do that on every query to be processed.

snuba/query/processors/granularity_processor.py
```diff
@@ -39,3 +41,53 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
                 Literal(None, granularity),
             )
         )
+
+
+PERFORMANCE_GRANULARITIES = [(60, 1), (3600, 2), (86400, 3)]
```
lynnagara, line 46: You could make these NamedTuples

snuba/query/processors/granularity_processor.py
```diff
@@ -39,3 +41,53 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
...
+        """Find the best fitting granularity for this query"""
+        requested_granularity = query.get_granularity()
+
+        if requested_granularity is None:
+            return self._default_granularity_enum
+        elif requested_granularity > 0:
+            for (granularity, mapped_value) in reversed(self._accepted_granularities):
+                if (
+                    requested_granularity % granularity == 0
+                    or requested_granularity == mapped_value
```
lynnagara, line 77: Why is this second condition needed?

snuba/query/processors/granularity_processor.py
```diff
@@ -39,3 +41,53 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
...
+                ):
+                    return mapped_value
+
+        raise InvalidGranularityException(
+            f"Granularity must be multiple of one of {self._available_granularities_values}"
+        )
+
+    def process_query(self, query: Query, query_settings: QuerySettings) -> None:
+        granularity = self.__get_granularity(query)
+        query.add_condition_to_ast(
```
lynnagara, line 87: It is just a matter of adding the condition to the query? Is there no existing granularity condition to be removed/replaced?

tests/test_generic_metrics_api.py
```diff
@@ -180,7 +180,7 @@ def test_raw_tags(self) -> None:
                     AND tags_raw[{tag_key}] = '{value_as_string}'
                     AND timestamp >= toDateTime('{self.start_time}')
                     AND timestamp < toDateTime('{self.end_time}')
-                    GRANULARITY 1
+                    GRANULARITY 60
```
lynnagara, line 183: I don't think I follow why these queries changed? Are you saying they were wrong before?

snuba/query/processors/granularity_processor.py
```diff
@@ -39,3 +41,53 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
...
+
+class MappedGranularityProcessor(QueryProcessor):
+    """
+    Use the granularity set on the query to filter on the granularity column,
+    supporting generic-metrics style enum mapping
+    """
+
+    def __init__(
+        self,
+        accepted_granularities: Sequence[Tuple[int, int]],
```
evanh, line 58: Why do you pass these in? Everywhere you are using this class it is using the constants above.

snuba/datasets/entities/generic_metrics.py
```diff
@@ -97,7 +103,14 @@ def __init__(
...
 
     def get_query_processors(self) -> Sequence[QueryProcessor]:
-        return [TagsTypeTransformer()]
+        return [
+            TagsTypeTransformer(),
+            MappedGranularityProcessor(
+                accepted_granularities=PERFORMANCE_GRANULARITIES,
+                default_granularity_enum=DEFAULT_MAPPED_GRANULARITY_ENUM,
+            ),
+            TimeSeriesProcessor({"bucketed_time": "timestamp"}, ("timestamp",)),
```
evanh, line 112: Do you need this processor if you have the granularity processor above?

snuba/query/processors/granularity_processor.py
onewland: it's probably not, but we do want product to eventually own handling this mapping so I thought it could be a good way to transition

snuba/query/processors/granularity_processor.py
onewland: yes, that's correct

snuba/datasets/entities/generic_metrics.py
onewland: yes, because the product wants to round timestamps to odd intervals like 15m or 3h or whatever which isn't handled by the granularity processor alone

tests/test_generic_metrics_api.py
lynnagara: Which one are you saying is the correct value and which one is the wrong one? And if you're saying it will be wrong after this PR can you give some info about why and how will it be addressed in future?

snuba/query/processors/granularity_processor.py
onewland: it wasn't being set previously on generic_metrics unless you added a `where` clause with the granularity column

tests/test_generic_metrics_api.py
onewland: When I originally put up the PR, I was thinking that both providing raw and enum values were correct. I've reconsidered and I now think we should accept only the "raw" (in seconds) granularity value to reduce confusion.

Our ideal is to take this non-infrastructure concern and move it to the product side (e.g. sentry should maintain the list of enum values for granularities it cares about and use them directly in queries). They don't have the flexibility in their query layer at the moment, though, to have two different approaches for building requests.

How it will be addressed in the future is undetermined but they're aware that we don't like having this logic in our side.

lynnagara (approved):

PR #2939, opened 2022-07-15
dev(generic-metrics) run generic-metrics consumers in dev

onewland:
this enables developers to automatically run generic metrics consumers for testing metrics-enhanced performance locally

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -215,6 +215,30 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--consumer-group=snuba-metrics-consumers",
                 ],
             ),
+            (
+                "generic-metrics-distributions-consumer",
+                [
+                    "snuba",
+                    "consumer",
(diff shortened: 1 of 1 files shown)


snuba/cli/devserver.py
```diff
@@ -215,6 +215,30 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--consumer-group=snuba-metrics-consumers",
                 ],
             ),
+            (
```
nikhars, line 218: Do you want to use the same settings flag we use for release health metrics or use a different flag `settings.ENABLE_SENTRY_METRICS_DEV`? Will there be a use case where someone wants to run just one of release health metrics vs performance metrics?

To enable the indexer in Sentry for release health vs performance metrics, is there a single setting or different settings? Maybe it might be a good idea to keep it consistent.

onewland: I did in fact copy the single-setting behavior from Sentry (Markus added this recently to help local e2e testing): https://github.com/getsentry/sentry/pull/36682

PR #2940, opened 2022-07-15
feat: Add a str default to JSON encoding

evanh:
There are some types of data (IP addresses, UUIDs) that are converted to
classes in the result (ipaddress.IPv4Address, uuid.UUID). These classes are not
JSON serializable, so would throw an exception when encoded into the cache or
returned in the JSON response. Instead, automatically encode them as strings.

Context:
In previous datasets (errors, transactions) that had these types of columns
there are mappers and query processors that transform the query so that
Clickhouse returns a string instead of the class. With newer datasets like
replays they do not have these processors. Instead of forcing new datasets to
have these processors, or to wrap all calls with `toString`, we will
automatically JSON encode these results as strings.

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -64,7 +64,7 @@ metrics = MetricsWrapper(environment.metrics, "db_query")
 
 class ResultCacheCodec(ExceptionAwareCodec[bytes, Result]):
     def encode(self, value: Result) -> bytes:
-        return cast(str, rapidjson.dumps(value)).encode("utf-8")
+        return cast(str, rapidjson.dumps(value, default=str)).encode("utf-8")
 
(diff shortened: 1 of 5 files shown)


nikhars (approved):
Looks good to me.

If we wanted to do something similar for the existing datasets(errors/transactions) where we have the custom classes, would removing the custom classes/processors cause any backward incompatible change?

evanh:
> would removing the custom classes/processors cause any backward incompatible change?

@nikhars For some of them yes it would. For example, errors and transactions have a weird way of processing event IDs to (strip/add) dashes. That processor happens to also cast the result to a string. For the IP columns we actually coalesce the IPv4 and IPv6 values, so we could remove the stringify but would need to keep the automatic `coalesce` in order to preserve existing behaviour.

PR #2941, opened 2022-07-15
Add urls column to replays dataset

cmanallen:
-- start migrations

-- migration replays : 0002_add_url
Local operations:
ALTER TABLE replays_local ADD COLUMN IF NOT EXISTS url String AFTER title;

Dist operations:
ALTER TABLE replays_dist ADD COLUMN IF NOT EXISTS url String AFTER title;
-- end migration replays : 0002_add_url

### DIFF
--- a/snuba/datasets/storages/replays.py
+++ b/snuba/datasets/storages/replays.py
@@ -25,6 +25,7 @@ columns = ColumnSet(
             Array(UUID()),
         ),  # TODO: create bloom filter index / materialize column
         ("title", String(Modifiers(readonly=True))),
+        ("urls", Array(String())),
         ### common sentry event columns
         ("project_id", UInt(64)),
         # release/environment info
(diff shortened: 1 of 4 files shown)


snuba/datasets/storages/replays.py
```diff
@@ -25,6 +25,7 @@
             Array(UUID()),
         ),  # TODO: create bloom filter index / materialize column
         ("title", String(Modifiers(readonly=True))),
+        ("urls", Array(String())),
```
nikhars, line 28: Could you provide additional context on why this column needs to be an array?

cmanallen: @nikhars It doesn't necessarily have to be an array.  We could write all the urls as separate replay record.  Is there a benefit to choosing one over the other?

cmanallen: This became a moot point after a requirements change.

nikhars (approved):

evanh:
Please add the generated SQL of the migration to the PR description.

cmanallen:
@evanh How do I do this?  I don't see anything in the docs.

JoshFerge (approved):

nikhars:
> @evanh @nikhars Build failing with `ERROR: build step 4 "gcr.io/sentryio/docker-compose" failed: context deadline exceeded`. Thoughts?

Please re-run the failed step and see what happens. Might have been an infrastructure problem.

PR #2942, opened 2022-07-15
fix(replacements) Move tag condition out of the PREWHERE clause

fpacifici:
In tag deletion we cannot select the rows to replace by running the tag condition in the PREWHERE clause.
The query today is something like
```
SELECT ....
PREWHERE has(tags.key, 'asd')
FROM errors_dist FINAL
WHERE 
    project_id = .... AND
    ....
```

This won't work if we replace twice the same row  by removing the same tag multiple times before a merge (that would squash the duplicates) happens.
This can happen if you try to delete the same tag twice 

The issue is that the `FINAL` clause of the query would squash the duplicates on the fly and exclude the old row that has already been replaced.
Unfortunately the tag condition is in the PREWHERE that is processed before the FINAL clause. Thus, when running the SELECT, only the old rows makes it to the FINAL clause that has nothing to squash.

This moves the tag condition to the WHERE. It makes the query slower but at least correct.

### DIFF
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -1159,16 +1159,23 @@ def process_delete_tag(
     tag_column_name = tag_column_map["tags"].get(tag, tag)
     is_promoted = tag in promoted_tags["tags"]
 
-    where = """\
-        WHERE project_id = %(project_id)s
-        AND received <= CAST('%(timestamp)s' AS DateTime)
+    # We cannot put the tag condition (which is what we are mutating) in the
+    # prewhere clause. This is because the prewhere clause is processed before
+    # the FINAL clause.
+    # So if we are trying to mutate a row that was mutated before but not merged
+    # yet, the PREWHERE would return the old row that has already been
+    # replaced.
+    prewhere = " PREWHERE project_id = %(project_id)s"
+
+    where_base = """\
+        WHERE received <= CAST('%(timestamp)s' AS DateTime)
         AND NOT deleted
     """
 
     if is_promoted and use_promoted_prewhere:
-        prewhere = " PREWHERE %(tag_column)s IS NOT NULL "
+        where = where_base + " AND %(tag_column)s IS NOT NULL "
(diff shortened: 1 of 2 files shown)


snuba/datasets/errors_replacer.py
```diff
@@ -1159,16 +1159,23 @@ def process_delete_tag(
...
+    # yet, the PREWHERE would return the old row that has already been
+    # replaced.
+    prewhere = " PREWHERE project_id = %(project_id)s"
+
+    where_base = """\
+        WHERE received <= CAST('%(timestamp)s' AS DateTime)
         AND NOT deleted
     """
 
     if is_promoted and use_promoted_prewhere:
```
nikhars, line 1175: This change makes the parameter `use_promoted_prewhere` useless. The logic is simply whether the tags are promoted columns or not. Can you remove the parameter altogether?

snuba/datasets/errors_replacer.py
```diff
@@ -1159,16 +1159,23 @@ def process_delete_tag(
...
-    where = """\
-        WHERE project_id = %(project_id)s
-        AND received <= CAST('%(timestamp)s' AS DateTime)
+    # We cannot put the tag condition (which is what we are mutating) in the
+    # prewhere clause. This is because the prewhere clause is processed before
+    # the FINAL clause.
+    # So if we are trying to mutate a row that was mutated before but not merged
+    # yet, the PREWHERE would return the old row that has already been
+    # replaced.
+    prewhere = " PREWHERE project_id = %(project_id)s"
```
nikhars, line 1168: I am guessing this is just for optimizing the query and not necessarily a requirement.

nikhars (commented):
Good find

snuba/datasets/errors_replacer.py
fpacifici: Actually thinking over it, this is useless as the project_id is already in the primary key

nikhars (approved):

PR #2950, opened 2022-07-18
feat(mep): add generic_metrics_distributions, generic_metrics_sets to subscripton-scheduler

onewland:
This is needed to deploy subscriptions/alerting for MEP

### DIFF
--- a/snuba/cli/subscriptions_scheduler.py
+++ b/snuba/cli/subscriptions_scheduler.py
@@ -24,7 +24,16 @@ logger = logging.getLogger(__name__)
     "--entity",
     "entity_name",
     required=True,
-    type=click.Choice(["events", "transactions", "metrics_sets", "metrics_counters"]),
+    type=click.Choice(
+        [
+            "events",
+            "transactions",
+            "metrics_sets",
+            "metrics_counters",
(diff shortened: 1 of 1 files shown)


snuba/cli/subscriptions_scheduler.py
```diff
@@ -24,7 +24,16 @@
...
-    type=click.Choice(["events", "transactions", "metrics_sets", "metrics_counters"]),
+    type=click.Choice(
+        [
+            "events",
+            "transactions",
+            "metrics_sets",
+            "metrics_counters",
+            "generic_metrics_sets",
+            "generic_metrics_distributions",
+        ]
```
rahul-kumar-saini, line 35: I think we should get values from the `EntityKey` enum instead of hard coding:

https://github.com/getsentry/snuba/blob/26c314be6ca493987ed53df8af83c81395cf502b/snuba/datasets/entities/__init__.py#L4

onewland: yeah, this verification is annoying but I also don't think EntityKey should be the check because many/most entities don't support subscriptions

probably should be more dynamically based on (non-code) config

lynnagara (approved):

PR #2951, opened 2022-07-18
fix(tiger): Tiger clusters do not require different limits

nikhars:
Tiger error clusters don't require different table rate limits. Use the same rate limits as prod cluster.


evanh (approved):

nikhars:
> Why is this no longer needed? Didn't we separate them to avoid double counting?

The cache namespace is still different. This PR removes the separate quotas we had for the different clusters.

PR #2952, opened 2022-07-19
feat: Add app_id to snuba timer query metrics

enochtangg:
To help better monitor query performance of the Snuba API, we want to add
the app_id tag to timer query metrics in order to add an extra dimension
to filtering.

### DIFF
--- a/snuba/querylog/__init__.py
+++ b/snuba/querylog/__init__.py
@@ -20,6 +20,7 @@ def _record_timer_metrics(
 ) -> None:
     final = str(request.query.get_final())
     referrer = request.referrer or "none"
+    app_id = request.attribution_info.app_id.key or "none"
     timer.send_metrics_to(
         metrics,
         tags={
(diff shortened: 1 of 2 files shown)


enochtangg:
> Your test is not actually checking the correct thing. The test you changed mocks out the entire `record_query` function, which means that your change is not being hit.
> 
> You need to either add a new test, or modify a test that doesn't mock the `record_query` function. You can modify this test for example: https://github.com/getsentry/snuba/blob/master/tests/test_snql_api.py#L614
> 
> You need to add a new call to `get_recorded_metric_calls` in the test that checks the metrics being written here: https://github.com/getsentry/snuba/blob/master/snuba/querylog/__init__.py#L23
> 
> If you want the name of the metric, you can add a debugger/print for `self.__name` here: https://github.com/getsentry/snuba/blob/master/snuba/utils/metrics/timer.py#L86

Good catch, added a new test which calls `test_timing_metrics_tags()` for the timing metrics query and checks for the appropriate tags.

evanh (approved):
Nice, looks good!

PR #2953, opened 2022-07-19
fix(optimize): Sort partitions each time when getting schedule

nikhars:
One of the reasons the optimize jobs are timing out is that there can be overlaps between the TTL merge and optimize merge. When this happens, the other merge is blocked with a reason along the following lines
`Not executing log_entry <x> of type MERGE_PARTS because part <y> is not ready yet`

TTL merges work with the partition which is 13 weeks old. If we order the partitions to optimize in such a manner that the 13 week old partition is merged in the end, we decrease the likelihood of overlaps happening and optimize jobs timing out

### DIFF
--- a/snuba/optimize_scheduler.py
+++ b/snuba/optimize_scheduler.py
@@ -55,6 +55,17 @@ class OptimizeScheduler:
         )
 
     @staticmethod
+    def sort_partitions(partitions: Sequence[str]) -> Sequence[str]:
+        def sort_ordering_key(partition_name: str) -> str:
+            match = re.search(CLICKHOUSE_PARTITION_RE, partition_name)
+            if match is not None:
+                return match.group()
+
+            return partition_name
+
+        return sorted(partitions, key=sort_ordering_key, reverse=True)
+
+    @staticmethod
     def subdivide_partitions(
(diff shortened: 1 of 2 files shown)


evanh (approved):
Can you elaborate on why moving that partition to the end helps? It doesn't take less time to get optimized. Do the TTL and optimize jobs always start at the same time?

lynnagara:
What happens when there are retention periods of different lengths? Can you add a test with both 30 and 90 days partitions?

nikhars:
> Can you elaborate on why moving that partition to the end helps? It doesn't take less time to get optimized. Do the TTL and optimize jobs always start at the same time?

The TTL merges kick in at 17:00 PST. It starts working on partition (current_week - 13). Since its a horizontal merge, it takes longer. Sometimes we see this running for 6-8 hours. The optimize cron job kicks in at 18:00 PST and runs for approximately 15 hours. By having the TTL merge for partition (current_week - 13) be at the end, the overlap time would be reduced

nikhars:
> What happens when there are retention periods of different lengths? Can you add a test with both 30 and 90 days partitions?

The sorting based on date would still apply. I have added the mixed 90 and 30 day retention to one of the tests

PR #2955, opened 2022-07-20
fix(optimize): Remove exception from optimize_partitions

nikhars:
The `optimize_partitions` api is primarily called from `optimize_partition_runner` from the cron job to control the execution based on different schedules When a timeout is provided to the api, it means that the current schedule can run only till the provided timeout.

Because `optimize_partitions` is run in a threaded context, it is not easy to capture its exception it raises since by the time join is called on the thread, the threads stack trace is already gone.

What is important for us is that the optimize job raise an exception when OPTIMIZE_JOB_CUTOFF_TIME is reached. Which it does in OptimizeScheduler's get_next_schedule. The code is here https://github.com/getsentry/snuba/blob/master/snuba/optimize_scheduler.py#L96-L98. We don't care about transitions between different schedules.

### DIFF
--- a/tests/test_optimize.py
+++ b/tests/test_optimize.py
@@ -9,12 +9,7 @@ from snuba import optimize, settings
 from snuba.clusters.cluster import ClickhouseClientSettings
 from snuba.datasets.storages import StorageKey
 from snuba.datasets.storages.factory import get_writable_storage
-from snuba.optimize import (
-    JobTimeoutException,
-    _get_metrics_tags,
-    optimize_partition_runner,
-    optimize_partitions,
-)
+from snuba.optimize import _get_metrics_tags, optimize_partition_runner
 from snuba.optimize_scheduler import OptimizedSchedulerTimeout, OptimizeScheduler
 from snuba.optimize_tracker import OptimizedPartitionTracker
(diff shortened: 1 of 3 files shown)


tests/test_optimize.py
```diff
@@ -253,17 +248,6 @@ def test_optimize_partitions_raises_exception_with_cutoff_time() -> None:
     tracker.update_all_partitions([dummy_partition])
     scheduler = OptimizeScheduler(2)
 
-    with pytest.raises(JobTimeoutException):
-        optimize_partitions(
```
evanh, line 257: Why did you remove the call to `optimize_partitions` entirely? Shouldn't the test make the call with a different exception?

nikhars: There is more context in the description of the PR. In short, optimize_partitions does not raise exceptions anymore. So I removed it from this test since this tests purpose is for raised exceptions. The logic for catching exceptions is below that code https://github.com/getsentry/snuba/pull/2955/files#diff-2e9c0212505684117bae2d1a5003a9cae77e63141310ea78ba8fc860d38bc2dfR254-R263

evanh (approved):
Makes sense, thanks.

PR #2958, opened 2022-07-20
[wip] config(redis): redis isolation by use case

onewland:
Will probably do some cleanup before even a preliminary review but I'll use this PR to see what people think of this approach

I know that the separate `redis_multi` package is weird and would be addressed before merge. Basically I was just trying to avoid a circular import between `settings` importing `redis` to see the new setting types and vice-versa

### DIFF
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -38,7 +38,8 @@ from snuba.datasets.events_processor_base import (
 )
 from snuba.datasets.schemas.tables import WritableTableSchema
 from snuba.processor import InvalidMessageType, _hashify
-from snuba.redis import redis_client
+from snuba.redis import get_redis_client
+from snuba.redis_multi.configuration import ClusterFunction
 from snuba.replacers.replacer_processor import Replacement as ReplacementBase
 from snuba.replacers.replacer_processor import (
     ReplacementMessage,
@@ -57,6 +58,8 @@ In theory this will be needed only during the events to errors migration.
 
 logger = logging.getLogger(__name__)
 metrics = MetricsWrapper(environment.metrics, "errors.replacer")
+
+redis_client = get_redis_client(ClusterFunction.REPLACEMENT_STORAGE)
 
 
 @dataclass(frozen=True)
--- a/snuba/redis.py
+++ b/snuba/redis.py
@@ -111,7 +111,6 @@ def _build_cluster_client(
             ClusterNode(host=node.host, port=node.port)
             for node in connection_descriptor
         ],
-        db=connection_descriptor[0].db,
         password=connection_descriptor[0].password,
         max_connections_per_node=True,
     )
--- a/snuba/redis.py
+++ b/snuba/redis.py
@@ -2,12 +2,17 @@ from __future__ import absolute_import
 
 import time
 from functools import wraps
-from typing import Any, Callable, Union, cast
+from typing import Any, Callable, Mapping, MutableMapping, Sequence, Union, cast
 
 from redis.client import StrictRedis
 from redis.cluster import ClusterNode, NodesManager, RedisCluster, RedisClusterException
 from redis.exceptions import BusyLoadingError, ConnectionError
 from snuba import settings
+from snuba.redis_multi.configuration import (
+    ClusterFunction,
+    ConnectionDescriptor,
+    NodeDescriptor,
+)
 from snuba.utils.serializable_exception import SerializableException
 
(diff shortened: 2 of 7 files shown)


snuba/settings/__init__.py
```diff
@@ -78,16 +83,29 @@
 CLICKHOUSE_TRACE_USER = os.environ.get("CLICKHOUSE_TRACE_USER", "default")
 CLICKHOUSE_TRACE_PASSWORD = os.environ.get("CLICKHOUSE_TRACE_PASS", "")
 
-# Redis Options
 USE_REDIS_CLUSTER = os.environ.get("USE_REDIS_CLUSTER", "0") != "0"
```
onewland, line 86: I'm not a fan of this option but for now I kept it so that you can run local tests as well as tests in CI with SNUBA_SETTINGS=test, though I think the entirety of redis connection settings should live in the `REDIS_CLUSTER_MAP`

snuba/redis.py
```diff
@@ -72,29 +77,80 @@ def wrapper() -> RedisClientType:
...
-            db=settings.REDIS_DB,
-            socket_keepalive=True,
-        )
-
-
-redis_client: RedisClientType = _initialize_redis_cluster()
+def _build_cluster_client(
+    connection_descriptor: ConnectionDescriptor,
+) -> RedisClientType:
+    assert isinstance(connection_descriptor, Sequence)
```
evanh, line 106: If we want to be this strict about the typing, then `ConnectionDescriptor` should be a class that validates its own input.

snuba/redis.py
```diff
@@ -72,29 +77,80 @@ def wrapper() -> RedisClientType:
...
+    assert len(connection_descriptor) > 0
+    assert isinstance(connection_descriptor[0], NodeDescriptor)
```
evanh, line 108: What if there is more than one entry in the array?

snuba/settings/__init__.py
```diff
@@ -78,16 +83,29 @@
...
 REDIS_DB = int(os.environ.get("REDIS_DB", 1))
 REDIS_INIT_MAX_RETRIES = 3
 
+# Set self-hosted users up with the default setup where everything is
+# on one redis. Users who want to multiplex to multiple clusters should
+# fill in the values of REDIS_CLUSTER_MAP with lists of NodeDescriptor[s]
+REDIS_SINGLE_NODE_DESCRIPTOR = NodeDescriptor(
+    REDIS_HOST, REDIS_PORT, REDIS_DB, REDIS_PASSWORD
+)
+REDIS_CLUSTER_MAP: Mapping[ClusterFunction, ConnectionDescriptor] = {
```
evanh, line 102: I don't think you should put these enums and classes in here. I think the settings file should be simple JSON or something equivalent that the redis files pull in and build into the various connection classes. That solves the import cycle as well.

snuba/settings/__init__.py
evanh: How hard is it to have CI/local testing use the same Redis connection?

snuba/datasets/errors_replacer.py
```diff
@@ -58,6 +59,8 @@
 logger = logging.getLogger(__name__)
 metrics = MetricsWrapper(environment.metrics, "errors.replacer")
 
+redis_client = get_redis_client(ClusterFunction.REPLACEMENT_STORAGE)
```
onewland, line 62: This pattern should be copied everywhere that uses `redis_client` from `snuba.redis` and then that variable should be deleted

snuba/settings/__init__.py
onewland: I didn't really investigate but it means either using single-node configured Redis in CI or using cluster-configured Redis locally (the two connection types are not compatible)

snuba/redis.py
onewland: In that case, something will blow up. 

This is less about safety and more to satisfy mypy (L106 can check `isinstance(Sequence)` but not `Sequence[NodeDescriptor]`)

snuba/settings/__init__.py
onewland: OK that makes sense, maybe we could add this developer guideline and other settings stuff to https://getsentry.github.io/snuba/index.html

One thing I really wasn't sure about is how we might deprecate a setting (which IMO we want to be able to do)

Additionally things like "don't send datadog metrics in consumers without setting a sample rate" etc.

lynnagara:
Closing due to staleness, would be good to eventually solve this though.

PR #2959, opened 2022-07-21
fix(subscriptions): Fix bug where scheduler got stuck in global mode

lynnagara:
Remove the code that prevented the tick buffer from making progress
when the buffers for all partitions are full. This was supposed to
be an optimization to avoid checking the buffers for every tick
unnecessarily but actually prevented the TickBuffer from making progress
in some circumstances.

### DIFF
--- a/snuba/subscriptions/scheduler_processing_strategy.py
+++ b/snuba/subscriptions/scheduler_processing_strategy.py
@@ -270,25 +270,13 @@ class TickBuffer(ProcessingStrategy[Tick]):
         # Periodically record the legnth of each buffer
         self.__record_tick_buffer_length()
 
-        # If there are any empty buffers, we can't submit anything yet.
-        # Otherwise if all the buffers have ticks then we look for the partition/s
-        # with the earliest tick (i.e. the tick with the earliest upper timestamp
-        # interval value) and submit it to the next step.
-        if len(self.__buffers[tick_partition]) > 1:
-            return
-
         while all(len(buffer) > 0 for buffer in self.__buffers.values()):
             earliest_ts = message.payload.timestamps.upper
             assert message.payload.partition is not None
             earliest_ts_partitions = {message.payload.partition}
 
             for partition_index in self.__buffers:
-                if partition_index == message.payload.partition:
-                    continue
-
                 buffer = self.__buffers[partition_index]
-                if len(buffer) == 0:
-                    return
 
                 tick = buffer[0].payload
 


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -270,25 +270,13 @@ def submit(self, message: Message[Tick]) -> None:
         # Periodically record the legnth of each buffer
         self.__record_tick_buffer_length()
 
-        # If there are any empty buffers, we can't submit anything yet.
-        # Otherwise if all the buffers have ticks then we look for the partition/s
-        # with the earliest tick (i.e. the tick with the earliest upper timestamp
-        # interval value) and submit it to the next step.
-        if len(self.__buffers[tick_partition]) > 1:
```
MeredithAnya, line 277: I'm pretty confused on what this was supposed to do before. Was this to enforce there being exactly 1 tick per partition?

lynnagara: It was supposed to be an optimization to skip checking all the buffers when we know there was an empty one as we don't make progress if any are empty: we only submit the earliest tick to the next stage when we have a tick in the buffer of every partition. So if the current message is added to a non empty buffer then we could infer that one of the other buffers has to be empty and skip the loop below. But.. there is an edge case in which is this not actually true.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -270,25 +270,13 @@ def submit(self, message: Message[Tick]) -> None:
         # Periodically record the legnth of each buffer
         self.__record_tick_buffer_length()
```
fpacifici, line 271: What about producing a metric with the low watermark (or the high one, either would work) from every buffer as a counter or gauge? 
Then we can see where we get stuck and if there is a partition that is lagging particularly behind. Just as Kafka record a metric containing the broker offset.

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -270,25 +270,13 @@ def submit(self, message: Message[Tick]) -> None:
...
             assert message.payload.partition is not None
             earliest_ts_partitions = {message.payload.partition}
 
             for partition_index in self.__buffers:
-                if partition_index == message.payload.partition:
-                    continue
-
                 buffer = self.__buffers[partition_index]
-                if len(buffer) == 0:
-                    return
```
fpacifici, line 291: Could this actually ever happen ?
It seems you cannot get here with an empty buffer if you entered the while loop.

fpacifici (commented):
I would suggest we add a few more metrics to the `TickBuffer`.
Specifically:
one to count the calls to submit. This tells us that we are actually receiving ticks. We may not even need to sample this
the other is a counter that tells us when we enter the while loop. The goal is to know if we get stuck with a list of buffers that is never full.

snuba/subscriptions/scheduler_processing_strategy.py
fpacifici: > But.. there is an edge case in which is this not actually true.

Is this speculative or do we know what this edge case is ?

fpacifici (approved):
Please see the comments in line

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: It's not really speculative, if you closely inspect how ticks get removed from the buffer in the loop below you can see how it is possible. Basically it happens when all buffers have got ticks in them except for the one of the current message's partitions AND the message's timestamp is later than some other partition and that other partition's buffer has more than one tick in it already.

Since it's normal for partitions to progress at different speeds it's not even than unusual an edge case and I'm actually surprised we didn't hit the issue before.

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: No that's why I deleted it as the condition will never be met. I think it's a holdover when this was simply a while True loop.

lynnagara:
> I would suggest we add a few more metrics to the `TickBuffer`. Specifically: one to count the calls to submit. This tells us that we are actually receiving ticks. We may not even need to sample this the other is a counter that tells us when we enter the while loop. The goal is to know if we get stuck with a list of buffers that is never full.

I'd prefer to skip this for now. Though it may be interesting, I don't think it tells us anything that we were able to infer already from existing metrics (albeit perhaps a little more indirectly). 
For the issue we saw, we already know that submit was being called 0 times for a while. This is because the metrics relating to the next processing step didn't fire.
It is heavily sampled but we have a rough idea of when we would enter the while loop - this is because we periodically record the length of every buffer. So we know in the recent incident that we were never entering this loop.

PR #2962, opened 2022-07-21
feat(replays): change replays processor to consume directly from relay event

JoshFerge:
With some simplifications to replay's architecture, we can directly consume replay events that are emitted from relay. This PR changes the processor to consume this schema, and also removes some unnecessary logic around promoted tags.

Replays is not live yet, so there is no concern about worrying about the old schema.

Updated architecture page here: https://www.notion.so/sentry/Replay-Backend-Architecture-Update-7-19-7c4ecc8af3fb4027992f3ee75dae9529

### DIFF
--- a/snuba/datasets/replays_processor.py
+++ b/snuba/datasets/replays_processor.py
@@ -21,6 +21,7 @@ from snuba.processor import (
     _ensure_valid_ip,
     _unicodify,
 )
+from snuba.util import force_bytes
 from snuba.utils.metrics.wrapper import MetricsWrapper
 
 logger = logging.getLogger(__name__)
@@ -116,7 +117,14 @@ class ReplaysProcessor(MessageProcessor):
     def _process_event_hash(
         self, processed: MutableMapping[str, Any], replay_event: ReplayEventDict
     ) -> None:
-        processed["event_hash"] = md5(replay_event["segment_id"]).hexdigest()
+        # event_hash is used to uniquely identify a row within a replay
+        # for our segment updates we'll simply hash the segment_id
+        # for future additions, we'll use whatever unique identifier (e.g. event_id)
+        # as the input to the hash
+
+        segment_id_bytes = force_bytes(str((replay_event["segment_id"])))
+        segment_hash = md5(segment_id_bytes).hexdigest()
+        processed["event_hash"] = str(uuid.UUID(segment_hash))
 
     def process_message(
         self, message: Mapping[Any, Any], metadata: KafkaMessageMetadata
--- a/snuba/datasets/replays_processor.py
+++ b/snuba/datasets/replays_processor.py
@@ -1,6 +1,7 @@
 import logging
 import uuid
 from datetime import datetime
+from hashlib import md5
 from typing import Any, Mapping, MutableMapping, Optional, cast
 
 from snuba import environment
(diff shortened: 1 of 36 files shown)


snuba/datasets/replays_processor.py
```diff
@@ -27,146 +27,121 @@
...
+            if "url" in replay_event["request"]:
+                return cast(str, replay_event["request"]["url"])
+        return None
+
+    def _process_base_replay_event_values(
+        self, processed: MutableMapping[str, Any], replay_event: ReplayEventDict
+    ) -> None:
+        processed["replay_id"] = str(uuid.UUID(replay_event["replay_id"]))
+        processed["sequence_id"] = replay_event["sequence_id"]
+        processed["trace_ids"] = replay_event.get("trace_ids")
```
evanh, line 65: `trace_ids` is not a nullable column from what I can see in the schema. I'm not sure it will work if you try to write `None` to Clickhouse.

snuba/datasets/replays_processor.py
```diff
@@ -27,146 +27,121 @@
...
-    ) -> MutableMapping[str, Any]:
-
-        processed["replay_id"] = str(uuid.UUID(event_dict["replay_id"]))
-        processed["project_id"] = event_dict["project_id"]
-
-        processed["sequence_id"] = event_dict["sequence_id"]
-        processed["trace_ids"] = event_dict["trace_ids"]
-
+    def _add_user_field_tag(
+        self, processed: Any, replay_event: ReplayEventDict
```
evanh, line 45:
```suggestion
        self, processed: MutableMapping[str, Any], replay_event: ReplayEventDict
```

snuba/datasets/replays_processor.py
```diff
@@ -27,146 +27,121 @@
...
     def _process_tags(
-        self,
-        processed: MutableMapping[str, Any],
-        event_dict: EventDict,
+        self, processed: MutableMapping[str, Any], replay_event: ReplayEventDict
     ) -> None:
-
-        tags: Mapping[str, Any] = _as_dict_safe(event_dict["data"].get("tags", None))
+        tags: MutableMapping[str, Any] = _as_dict_safe(replay_event.get("tags", None))
+        processed["title"] = tags.get("transaction")
```
evanh, line 79: `title` is not nullable.

snuba/datasets/replays_processor.py
```diff
@@ -27,146 +27,121 @@
...
-        processed["project_id"] = event_dict["project_id"]
-
-        processed["sequence_id"] = event_dict["sequence_id"]
-        processed["trace_ids"] = event_dict["trace_ids"]
-
+    def _add_user_field_tag(
+        self, processed: Any, replay_event: ReplayEventDict
+    ) -> None:
+        if "user" not in replay_event:
+            return
```
evanh, line 48: `user` and `user_hash` are not nullable in your schema.

evanh (commented):
I think you need to double check this processor against your schema for which columns are nullable. There are more than just the ones I commented on.

JoshFerge:
@evanh i've adjusted the processor to handle the null cases you've mentioned (and other possible ones), and that the clickhouse writes work. i've adjusted the schema a bit, in #2972, which is required for this PR. I've seen that if a string is omitted but not nullable (e.g. title), the writes are still successful, and i believe an empty string will exist in that column.

snuba/datasets/replays_processor.py
JoshFerge: 👍🏼

snuba/datasets/replays_processor.py
JoshFerge: 👍🏼 let me know if it's preferred to write empty string, or we can also make it nullable instead. i saw this: https://stackoverflow.com/questions/63057886/clickhouse-string-field-disk-usage-null-vs-empty

snuba/datasets/replays_processor.py
JoshFerge: 👍🏼

evanh (approved):
Thanks for that, looks good.

PR #2963, opened 2022-07-22
Add kafka ssl authentication params to snuba

bumarcell:
### Legal Boilerplate

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -114,8 +114,12 @@ BULK_BINARY_LOAD_CHUNK = 2**22  # 4 MB
 # Processor/Writer Options
 
 BROKER_CONFIG: Mapping[str, Any] = {
-    # See snuba/utils/streams/backends/kafka.py for the supported options
+    # See https://github.com/getsentry/arroyo/blob/main/arroyo/backends/kafka/configuration.py#L16-L38 for the supported options
     "bootstrap.servers": os.environ.get("DEFAULT_BROKERS", "localhost:9092"),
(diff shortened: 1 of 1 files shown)


onewland:
Hi @bumarcell,

Thanks for your PR.

If you'd like to add these options to your local environment, you can do that without this change. You can create a new file in the settings/ directory that overrides the values in `settings/__init__.py` and run snuba with `SNUBA_SETTINGS=your_env_name`. The makefile does this for `make test` (`SNUBA_SETTINGS=test`).

Your comment fix is an improvement if you're interested in contributing just that.

bumarcell:
Hello @onewland 
Thanks for your response! 

So my goal is to deploy sentry (with snuba) in k8s through the helm chart (v15).
The workaround you suggested still needs me to clone snuba and build it locally, which what I already did up till this point.
Ideally I'd like to be able to use sentry and Snuba from your source (with updates possible) while being able to use external Kafka with ssl auth and without having to build Snuba myself every time I need to get an update :) 

P.S. Unfortunately at the moment when I try to use latest Snuba 22.7.0 the deployment fails at `snuba-migration-job` - apparently because Snuba tries to migrate some field called GENERIC_METRICS_.. which internal Clickhouse doesn't recognize and breaks the whole thing..

onewland:
re: the migration, that's a separate issue that I think is worth filing at https://github.com/getsentry/snuba/issues

We'll discuss internally about whether we think it's a good idea to add this environment-based control for Kafka connection settings.

onewland:
I think we're fine with this change in principle. Letting the CI run go and if that passes we need to make sure it doesn't conflict with our production settings.

onewland (approved):

PR #2965, opened 2022-07-22
feat(profiling): Add timeseries processor for profiles entity

Zylphrex:
We want to generate some timeseries on the profiles. The timeseries processor
will allow us to use the granularity clause in snql.


evanh (approved):
Just FYI, we are planning to deprecate/change some of these "magic" columns. They only exist because the expressions they generate were not possible with the old JSON Snuba API. There's no reason this logic couldn't know be done in SnQL.

PR #2971, opened 2022-07-25
feat(query): Mechanism to override query settings

nikhars:
Added a mechanism to override the query settings that are fetched from runtime config. This is to allow different query settings on the errors cluster.

The current mechanism provides a way of overriding using cache partitions which is necessary because prod and tiger cluster use the same entity.

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -553,7 +553,7 @@ def _get_cache_wait_timeout(
 
 
 def _get_query_settings_from_config(
-    cache_partition: Optional[str],
+    override_prefix: Optional[str],
 ) -> MutableMapping[str, Any]:
     """
     Helper function to get the query settings from the config.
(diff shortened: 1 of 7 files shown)


snuba/web/db_query.py
```diff
@@ -552,6 +552,37 @@ def _get_cache_wait_timeout(
...
+    cache_partition: Optional[str],
+) -> MutableMapping[str, Any]:
+    """
+    Helper function to get the query settings from the config.
+
+    #TODO: Make this configurable by entity/dataset. Since we want to use
+    #      different settings across different clusters belonging to the
+    #      same entity/dataset, using cache_partition right now. This is
+    #      not ideal but it works for now.
+    """
```
fpacifici, line 565: This seems quite dangerous. Since the cache partition key has nothing to do with the way you want to partition the query settings, we will forget about this when making changes to the cache partition config.
I think you can clean this up a bit by hiding the partitioning in the Cluster class or the Reader class (just as we did for cache partition).
There we can associate the partition for the query settings to the Cluster and not making it implicitly tied to the cache in code that has nothing to do with cache. 

You can start by having a `get_settings_prefix` or `get_settings_partition` in the Cluster (and then in the reader). You may want to populate that with the cache partition but at that point you can just populate it with something cleaner if you want.

nikhars: Made the changes according to your comments. Let me know if that's what you had in mind.

onewland (approved):

PR #2972, opened 2022-07-25
ref(replays): adjust original schema

JoshFerge:
- sequence_id 
    - remove from the sorting key, and replace with timestamp
    - make nullable (so that way we can receive out-of-bands updates)
    - rename to segment_id (sequence_id is kind of ambiguous)
- add default empty string for the user field
- add the materialized function for user_hash (missed this when copying the schema from txs, etc.)

We have received no production traffic so no problem with simply rerunning the initial migration, and Colton and I are the only developers who have run the migration locally and we can handle re-running.

### DIFF
--- a/snuba/datasets/replays_processor.py
+++ b/snuba/datasets/replays_processor.py
@@ -71,7 +71,7 @@ class ReplaysProcessor(MessageProcessor):
         processed["replay_id"] = str(uuid.UUID(event_dict["replay_id"]))
         processed["project_id"] = event_dict["project_id"]
 
-        processed["sequence_id"] = event_dict["sequence_id"]
+        processed["segment_id"] = event_dict["segment_id"]
         processed["trace_ids"] = event_dict["trace_ids"]
 
         processed["timestamp"] = self.__extract_timestamp(
--- a/snuba/migrations/snuba_migrations/replays/0001_replays.py
+++ b/snuba/migrations/snuba_migrations/replays/0001_replays.py
@@ -99,7 +99,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 columns=raw_columns,
                 engine=table_engines.Distributed(
                     local_table_name="replays_local",
-                    sharding_key="cityHash64(toString(replay_id))",
+                    sharding_key="cityHash64(replay_id)",
                 ),
             ),
         ]
--- a/snuba/migrations/snuba_migrations/replays/0001_replays.py
+++ b/snuba/migrations/snuba_migrations/replays/0001_replays.py
@@ -65,7 +65,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 columns=raw_columns,
                 engine=table_engines.ReplacingMergeTree(
                     storage_set=StorageSetKey.REPLAYS,
-                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id), timestamp)",
+                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id))",
                     partition_by="(retention_days, toMonday(timestamp))",
                     settings={"index_granularity": "8192"},
                     ttl="timestamp + toIntervalDay(retention_days)",
--- a/snuba/migrations/snuba_migrations/replays/0001_replays.py
+++ b/snuba/migrations/snuba_migrations/replays/0001_replays.py
@@ -18,7 +18,7 @@ from snuba.migrations.columns import MigrationModifiers as Modifiers
 raw_columns: Sequence[Column[Modifiers]] = [
     Column("replay_id", UUID()),  # the id of the encompassing replay
     Column(
-        "event_hash", UInt(64)
+        "event_hash", UUID()
     ),  # a hash that individually identifies a unique update within the replay
     Column("segment_id", UInt(16, Modifiers(nullable=True))),
     Column("trace_ids", Array(UUID())),
--- a/snuba/migrations/snuba_migrations/replays/0001_replays.py
+++ b/snuba/migrations/snuba_migrations/replays/0001_replays.py
@@ -17,7 +17,7 @@ from snuba.migrations.columns import MigrationModifiers as Modifiers
 
 raw_columns: Sequence[Column[Modifiers]] = [
     Column("replay_id", UUID()),
-    Column("sequence_id", UInt(16)),
+    Column("segment_id", UInt(16, Modifiers(nullable=True))),
     Column("trace_ids", Array(UUID())),
     Column(
         "_trace_ids_hashed",
@@ -38,8 +38,8 @@ raw_columns: Sequence[Column[Modifiers]] = [
     Column("ip_address_v4", IPv4(Modifiers(nullable=True))),
     Column("ip_address_v6", IPv6(Modifiers(nullable=True))),
     # user columns
-    Column("user", String()),
-    Column("user_hash", UInt(64)),
+    Column("user", String(Modifiers(default="''"))),
+    Column("user_hash", UInt(64, Modifiers(materialized="cityHash64(user)"))),
     Column("user_id", String(Modifiers(nullable=True))),
     Column("user_name", String(Modifiers(nullable=True))),
     Column("user_email", String(Modifiers(nullable=True))),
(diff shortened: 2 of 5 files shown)


snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -38,8 +38,8 @@
     Column("ip_address_v4", IPv4(Modifiers(nullable=True))),
     Column("ip_address_v6", IPv6(Modifiers(nullable=True))),
     # user columns
-    Column("user", String()),
-    Column("user_hash", UInt(64)),
+    Column("user", String(Modifiers(default="''"))),
+    Column("user_hash", UInt(64, Modifiers(materialized="cityHash64(user)"))),
```
JoshFerge, line 42: i'm actually not sure if we need `user_hash` -- what's the purpose on our other tables?

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -38,8 +38,8 @@
     Column("ip_address_v4", IPv4(Modifiers(nullable=True))),
     Column("ip_address_v6", IPv6(Modifiers(nullable=True))),
     # user columns
-    Column("user", String()),
-    Column("user_hash", UInt(64)),
+    Column("user", String(Modifiers(default="''"))),
```
JoshFerge, line 41: i'm adding this as i see it on other tables, but does a string actually need an empty string modifier? (or is it empty by default if not set).

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -65,7 +65,7 @@ def forwards_local(self) -> Sequence[operations.SqlOperation]:
                 columns=raw_columns,
                 engine=table_engines.ReplacingMergeTree(
                     storage_set=StorageSetKey.REPLAYS,
-                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id), sequence_id)",
+                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id), timestamp)",
```
MeredithAnya, line 68: why do we have both `toStartOfDay(timestamp)` and `timestamp` as part of the sorting key ?

JoshFerge: ahh i thought sorting keys have to be unique, but they dont. Good point.
>ClickHouse does not require a unique primary key. You can insert multiple rows with the same primary key.
https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/

snuba/migrations/snuba_migrations/replays/0001_replays.py
lynnagara: Yeah I can't explain why it's in some other places. Seems unnecessary to me.

snuba/migrations/snuba_migrations/replays/0001_replays.py
lynnagara: As far as I can tell, we never even query this column in other datasets. So I think we could omit it.

snuba/migrations/snuba_migrations/replays/0001_replays.py
```diff
@@ -65,7 +65,7 @@ def forwards_local(self) -> Sequence[operations.SqlOperation]:
                 columns=raw_columns,
                 engine=table_engines.ReplacingMergeTree(
                     storage_set=StorageSetKey.REPLAYS,
-                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id), sequence_id)",
+                    order_by="(project_id, toStartOfDay(timestamp), cityHash64(replay_id))",
```
lynnagara, line 68: This is the primary key. If we drop `sequence_id` doesn't that mean there will be many records with the same primary key? (which will get merged together since this is a replacing merge tree)

JoshFerge: ahh right. yes we don't want that -- so we could either timestamp in to the end of the order by, or make it a regular MergeTree. I think the only pro of a ReplacingMergeTree is we can overwrite later on if we want to remove a row, don't see any use for it otherwise.

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: (correction: since this a ReplacingMergeTree, the sorting key will be unique)

lynnagara:
We don't normally allow modifications to existing migrations. But this one is ok since replays is still in the default set of SKIPPED_MIGRATION_GROUPS and thus hasn't gone out to most users.

snuba/migrations/snuba_migrations/replays/0001_replays.py
lynnagara: No they definitely have to be unique if you want to preserve those rows. Replacing merge tree merges rows based on this sorting key. But I don't think `timestamp` is guaranteed to be unique - in our other datasets its rounded to the second. Why not segment ID?

JoshFerge: @lynnagara and I talked off GH -- to handle, we add a new column in addition to segment_id, `event_hash`. for segment updates, this will simply be a hash of the segment_id, and for non_segment updates, it can be a hash of a unique part of the update. (e.g. for event associations, it can be a hash of the event_id). 

This way, we can keep a ReplacingMergeTree (which is useful for deletions (via overwrite), and handling duplicate kafka messages).

JoshFerge:
think i may need to fix a couple of tests, but schema should be good for review 👍🏼

snuba/migrations/snuba_migrations/replays/0001_replays.py
nikhars: The `default` modifier is not needed. You can get rid of it. But if this column could be empty then I would recommend making it nullable.

snuba/migrations/snuba_migrations/replays/0001_replays.py
nikhars: If you don't have a specific use case to use the column right now then I would recommend not adding it. There would be an unnecessary materialization which would have to be computed during inserts.

snuba/migrations/snuba_migrations/replays/0001_replays.py
nikhars: Changing a Table engine is not possible once you have data in them. I would recommend that you spend some time thinking which table engine you would like to use. ReplacingMergeTree seems like a decent engine which would allow you to perform replacing rows if you ever needed that functionality.

JoshFerge: Yeah discussed offline, many good reasons to stick with the ReplacingMergeTree. see comment above https://github.com/getsentry/snuba/pull/2972#discussion_r929357751

snuba/migrations/snuba_migrations/replays/0001_replays.py
JoshFerge: does this stackoverflow post apply here: https://stackoverflow.com/questions/63057886/clickhouse-string-field-disk-usage-null-vs-empty? or is null preferred?

snuba/datasets/replays_processor.py
```diff
@@ -71,7 +71,7 @@ def _process_base_event_values(
         processed["replay_id"] = str(uuid.UUID(event_dict["replay_id"]))
         processed["project_id"] = event_dict["project_id"]
 
-        processed["sequence_id"] = event_dict["sequence_id"]
+        processed["segment_id"] = event_dict["segment_id"]
```
lynnagara, line 74: You will need to write the event_hash column. Are you planning to do this here or as a follow up?

JoshFerge: followup, doing it in #2962

lynnagara: Ok. Just keep in mind any replay data currently in this table will not be valid until then as rows will be incorrectly merged

JoshFerge: table is empty right now 👍🏼

PR #2973, opened 2022-07-25
feat(mep): add support for generic-metrics dataset subscription executors

onewland:
Tested locally

### DIFF
--- a/snuba/cli/subscriptions_executor.py
+++ b/snuba/cli/subscriptions_executor.py
@@ -33,12 +33,12 @@ from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
     multiple=True,
     type=click.Choice(
         [
-            "events",
-            "transactions",
-            "metrics_counters",
-            "metrics_sets",
-            "generic_metrics_sets",
-            "generic_metrics_distributions",
+            EntityKey.EVENTS.value,
+            EntityKey.TRANSACTIONS.value,
+            EntityKey.METRICS_COUNTERS.value,
+            EntityKey.METRICS_SETS.value,
+            EntityKey.GENERIC_METRICS_SETS.value,
+            EntityKey.GENERIC_METRICS_DISTRIBUTIONS.value,
         ]
     ),
     help="The entity to target.",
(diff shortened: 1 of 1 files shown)


snuba/cli/subscriptions_executor.py
```diff
@@ -23,15 +23,24 @@
...
-    type=click.Choice(["events", "transactions", "metrics_counters", "metrics_sets"]),
+    type=click.Choice(
+        [
+            "events",
+            "transactions",
+            "metrics_counters",
+            "metrics_sets",
+            "generic_metrics_sets",
+            "generic_metrics_distributions",
+        ]
```
rahul-kumar-saini, line 42: Following up on my comment from a previous PR: https://github.com/getsentry/snuba/pull/2950#discussion_r923849164

I didn't mean have the choice be all values from the `EntityKey` Enum, but to generate this list of choices using values from the Enum.

ie I'm suggesting we replace this list of strings with something like:
```python3
[
    EntityKey.EVENTS.value,
    ...,
    EntityKey.GENERIC_METRICS_DISTRIBUTIONS.value,
]
```
where the list _only_ contains values we want, not all possible entities. Although not the ideal non-code config solution, it's probably still better than hardcoding strings.

lynnagara (approved):

snuba/cli/subscriptions_executor.py
onewland: sure, I think that makes sense

nikhars (approved):

snuba/cli/subscriptions_executor.py
```diff
@@ -23,15 +23,24 @@
...
+    type=click.Choice(["events", "transactions", "metrics", "generic_metrics"]),
     help="The dataset to target.",
 )
 @click.option(
     "--entity",
     "entity_names",
     required=True,
     multiple=True,
-    type=click.Choice(["events", "transactions", "metrics_counters", "metrics_sets"]),
+    type=click.Choice(
```
fpacifici, line 34: Is there any reason for not picking up the full list of entities? 
If the entity does not support subscriptions the user gets an error immediately (nothing different than getting the error from click).
This way, instead, we always need to touch the CLI file when adding an entity, which we generally forget and likely cause another back and forth when trying to deploy in prod.

onewland: I'm going to merge this just to get alerts unblocked, but maybe I'll follow up with a PR that allows all entity names and then gives a more clearer message for the assertion of `assert result_topic_spec is not None` on line 117/126 which I assume is the place this would die

PR #2974, opened 2022-07-26
build: Clean up requirements.txt

lynnagara:
We have too many unused dependencies in our requirements.txt, remove
everything that is not actually required.

This removes the transitive dependencies from requirements.txt which
means we will no longer be pinning specific versions of some packages.


evanh (approved):
Wow lots of dead packages. Were any of these dependencies of other packages? For example I thought Flask was dependent on Jinja2?

lynnagara:
> Wow lots of dead packages. Were any of these dependencies of other packages? For example I thought Flask was dependent on Jinja2?

Yes a lot of them were. We were fairly inconsistent about this - sometimes we pin exact versions of a package's dependencies and sometimes we do not.

nikhars:
Were you able to run Snuba api and Snuba consumers locally with these changes?

lynnagara:
> Were you able to run Snuba api and Snuba consumers locally with these changes?

Thanks for reminding me to do this. Added back 2 packages that were needed. Now devserver works.

nikhars (approved):

evanh:
> Were you able to run Snuba api and Snuba consumers locally with these changes?

Also the snuba admin.

PR #2976, opened 2022-07-26
ref: freeze requirements using pip-tools

asottile-sentry:
I determined the minimal set of requirements using this script:

```python
import ast
import sys
from classify_imports import Classified, classify_base

class FindImports(ast.NodeVisitor):
    def __init__(self):
        self.mods = set()

    def visit_Import(self, node: ast.Import) -> None:
        for alias in node.names:
            base, _, _ = alias.name.partition('.')
            if classify_base(base) is Classified.THIRD_PARTY:
                self.mods.add(base)
        self.generic_visit(node)

    def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
        if not node.level and node.module is not None:
            base, _, _ = node.module.partition('.')
            if classify_base(base) is Classified.THIRD_PARTY:
                self.mods.add(base)
        self.generic_visit(node)


def main() -> int:
    visitor = FindImports()
    for filename in sys.argv[1:]:
        with open(filename, 'rb') as f:
            visitor.visit(ast.parse(f.read(), filename))

    for mod in sorted(visitor.mods):
        print(mod)

if __name__ == '__main__':
    raise SystemExit(main())
```

I think eventually dev-infra wants to provide some centralized freezing / verifying scripts similar to this.  the current version of this PR doesn't enforce `pip-tools` but probably should (and should probably freeze dev requirements as well though I didn't approach that)


lynnagara (commented):
This PR changes the role of requirements.txt and how we manage dependencies quite a bit. It's not clear to me how we should be ensuring requirements.txt is up to date or regenerating that file when we want to add/remove/update a dependency. Does it mean pip-tools is required in dev now? What is the process to re-freeze dependencies?

I am also wondering how this fits with our exiting Dependabot configuration. Manually updating every transitive dependency of all our libraries in requirements.txt via Dependabot doesn't feel sustainable.

asottile-sentry:
> This PR changes the role of requirements.txt and how we manage dependencies quite a bit. It's not clear to me how we should be ensuring requirements.txt is up to date or regenerating that file when we want to add/remove/update a dependency. Does it mean pip-tools is required in dev now? What is the process to re-freeze dependencies?
> 
> I am also wondering how this fits with our exiting Dependabot configuration. Manually updating every transitive dependency of all our libraries in requirements.txt via Dependabot doesn't feel sustainable.

yeah totally agree -- I don't think the tooling is quite there yet (it does require to run `pip-tools` locally, ideally that'd get run in CI as well).  as for the versioning, I tend to prefer lower-bounded versioning  (and only use `!=`, `==` when absolutely needed) -- `pip-tools` will prefer the already pinned versions when updating and it ~generally reduces conflicts when managing a lot of deps

we'll be providing better generic tooling for this in the future, it's not there yet though so I'd understand not moving forward with this right now.  that said, it is a little concerning that dependencies can slip out from under us

evanh:
> it is a little concerning that dependencies can slip out from under us

Can you clarify what this means? You mean we don't update often enough?

asottile-sentry:
> > it is a little concerning that dependencies can slip out from under us
> 
> Can you clarify what this means? You mean we don't update often enough?

for any unpinned transitive dependency snuba will start using it without a code change because pip defaults to latest.

requirements.txt
```diff
@@ -1,50 +1,58 @@
...
 itsdangerous==2.1.2
-Jinja2==3.1.2
+jinja2==3.1.2
 jsonschema==4.7.1
-lz4==4.0.1
-Markdown==3.3.7
-MarkupSafe==2.1.1
+markdown==3.3.7
+markupsafe==2.1.1
+milksnake==0.1.5
```
evanh, line 30: I noticed there's a bunch of packages like this that weren't in the requirements before. Is the plan to somehow filter these out?

evanh (commented):
~~Sorry I'm still confused, all our packages are pinned to specific versions.~~

This makes sense now.

requirements.txt
asottile-sentry: these are the transitive dependencies which on master are currently unpinned

evanh: Got it.

PR #2978, opened 2022-07-26
ref(replays): load the payload as a byte array, switch topic names

JoshFerge:
Final adjustment for our replays events consumer to ingest directly from relay: 

- relay sends the payload as a byte array within the json blob, so we need to convert the bytearray to `bytes`, then load that as json.
- changing the topic that we ingest from. `snuba-replay-events` replaced by `ingest-replay-events`. `snuba-replay-events` will not be used.


evanh (approved):

JoshFerge:
think this needs to go out together with https://github.com/getsentry/ops/pull/5042

MeredithAnya:
@JoshFerge just a question, is relay producing to the `ingest-replay-events` topic? and then snuba is just consuming straight from that topic without any intermediate steps?

JoshFerge:
@MeredithAnya yup

snuba/datasets/replays_processor.py
```diff
@@ -133,7 +134,7 @@ def process_message(
         self, message: Mapping[Any, Any], metadata: KafkaMessageMetadata
     ) -> Optional[ProcessedMessage]:
         try:
-            replay_event = message["payload"]
+            replay_event = json.loads(bytes(message["payload"]))
```
nikhars, line 137: I would suggest you use rapidjson instead of json since it offers better performance. Most of our consumers use rapidjson

JoshFerge: oh yeah, was gonna ask about that. yeah totally makes sense.

nikhars (approved):

PR #2979, opened 2022-07-26
feat(config): Non Experimental Consumers

rahul-kumar-saini:
### Overview
- As a result of our discussion [here](https://www.notion.so/sentry/Snuba-s-Freight-configs-c63670dad9954feea617c7c51ba86e97?d=c8dcd80ff5924f6182c2e43dc674fcf7#75971ae34f344f888799632f5bf82a07), I've replaced the `experimental` config with a new `non-experimental` config
- A consumer is marked experimental by adding it to `.freight.yml` and not adding it to `.freight.stable.yml`
- A consumer is promoted to production ready by adding it to `.freight.stable.yml`

I will look into adding some sort of GitHub action which will comment on PRs that promote/introduce consumers, something similar to how a comment is added to PRs which introduce migrations.

<strong>Here is a list of consumers I've "marked" experimental by excluding them from this list:</strong>
- profiles-consumer
- loadtest-outcomes-consumer
- loadtest-transactions-consumer
- loadtest-loadbalancer-outcomes-consumer
- loadtest-sessions-consumer
- loadtest-errors-consumer

*This PR should not be merged until a new Freight app is added supporting this config*


.freight.non-experimental.yml
```diff
@@ -0,0 +1,56 @@
...
+      - image: us.gcr.io/sentryio/snuba:{sha}
+        name: snuba-admin
+  - kind: KubernetesCronJob
+    selector:
+      label_selector: service=snuba
+    containers:
+      - image: us.gcr.io/sentryio/snuba:{sha}
+        name: cleanup
+      - image: us.gcr.io/sentryio/snuba:{sha}
+        name: optimize
```
rahul-kumar-saini, line 56: I can remove this but for the sake of keeping this file as the standard for a minimum production ready config I've left it in

onewland:
metrics-* and generic-metrics-* consumers are not experimental

.freight.non-experimental.yml
evanh: I think you have the right idea. This should be everything except experimental consumers.

MeredithAnya:
@rahul-kumar-saini  how tied are we to the naming of `non-experimental`? To me it makes more sense to name that one based on what it is, rather than what it isn't. Maybe something like `freight.stable.yml`? Not sure if anyone else has opinions on that, could just be personal preference

lynnagara:
Does it mean that to deploy all services we need to kick off two deployments (experimental and non-experimental)? I would prefer "all" and "experimental" as I think the Snuba team should always be deploying all by default.

rahul-kumar-saini:
> @MeredithAnya
To me it makes more sense to name that one based on what it is, rather than what it isn't. Maybe something like freight.stable.yml

✅ Done, it was just an arbitrary name I wrote.



> @lynnagara 
Does it mean that to deploy all services we need to kick off two deployments (experimental and non-experimental)? I would prefer "all" and "stable" as I think we should always be deploying all by default.

What you're suggesting is the way I've implemented it, the naming was just confusing.

We would deploy just `.freight.yml` via the `Snuba` app on Freight (which is the "all consumers" deployment). Then if an experimental consumer within that deployment is blocking us, we deploy `.freight.stable.yml` which would be a different Freight app.

Thoughts on what to name this freight app? [`snuba-stable`](https://freight.getsentry.net/snuba-stable)?

nikhars (approved):

PR #2982, opened 2022-07-27
feat(migrations): Add source column to transactions

andriisoldatenko:
Add new column `source` to transactions group.

PR based on [Snuba migrations field guide](https://github.com/getsentry/snuba/blob/master/MIGRATIONS.md).

Processor implementation lives in https://github.com/getsentry/snuba/pull/2985

### DIFF
--- a/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
@@ -23,7 +23,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 storage_set=StorageSetKey.TRANSACTIONS,
                 table_name=table_name,
                 column=Column(
-                    "transactions_source",
+                    "transaction_source",
                     String(Modifiers(nullable=True, low_cardinality=True)),
                 ),
                 after="title",
--- a/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
@@ -35,7 +35,7 @@ class Migration(migration.ClickhouseNodeMigration):
     ) -> Sequence[operations.SqlOperation]:
         return [
             operations.DropColumn(
-                StorageSetKey.TRANSACTIONS, table_name, "transactions_source"
+                StorageSetKey.TRANSACTIONS, table_name, "transaction_source"
             ),
         ]
 
--- a/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
@@ -8,7 +8,9 @@ from snuba.migrations.columns import MigrationModifiers as Modifiers
 
 class Migration(migration.ClickhouseNodeMigration):
     """
-    Add source column required for ...
+    The source column is required to query for transaction events
+    in the transaction summary when the transaction name has been
+    redacted in metrics.
     """
 
     blocking = False
--- a/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
@@ -21,12 +21,8 @@ class Migration(migration.ClickhouseNodeMigration):
                 storage_set=StorageSetKey.TRANSACTIONS,
                 table_name=table_name,
                 column=Column(
-                    "source",
-                    String(
-                        Modifiers(
-                            low_cardinality=True, materialized="transaction_source"
-                        )
-                    ),
+                    "transactions_source",
+                    String(Modifiers(nullable=True, low_cardinality=True)),
                 ),
                 after="title",
             ),
@@ -36,7 +32,9 @@ class Migration(migration.ClickhouseNodeMigration):
         self, table_name: str
     ) -> Sequence[operations.SqlOperation]:
         return [
-            operations.DropColumn(StorageSetKey.TRANSACTIONS, table_name, "source"),
+            operations.DropColumn(
+                StorageSetKey.TRANSACTIONS, table_name, "transactions_source"
+            ),
         ]
 
     def forwards_local(self) -> Sequence[operations.SqlOperation]:
--- a/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
@@ -23,7 +23,9 @@ class Migration(migration.ClickhouseNodeMigration):
                 column=Column(
                     "source",
                     String(
-                        Modifiers(low_cardinality=True, materialized="'transaction'")
+                        Modifiers(
+                            low_cardinality=True, materialized="transaction_source"
+                        )
                     ),
                 ),
                 after="title",
--- a/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
@@ -1,0 +1,50 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, String
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Add source column required for ...
+    """
+
+    blocking = False
+
+    def __forward_migrations(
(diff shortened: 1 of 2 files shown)


snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
```diff
@@ -0,0 +1,50 @@
...
+
+from snuba.clickhouse.columns import Column, String
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Add source column required for ...
```
jan-auer, line 11: The source column is required to query for transaction events in the transaction summary when the transaction name has been redacted in metrics.

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
```diff
@@ -0,0 +1,50 @@
...
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name=table_name,
+                column=Column(
+                    "source",
+                    String(
+                        Modifiers(low_cardinality=True, materialized="'transaction'")
```
jan-auer, line 26: What is the correct `materialized` value here?

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __forward_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name=table_name,
+                column=Column(
+                    "transactions_source",
```
andriisoldatenko, line 26:
```suggestion
                    "transaction_source",
```

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
```diff
@@ -0,0 +1,52 @@
...
+                after="title",
+            ),
+        ]
+
+    def __backwards_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.DropColumn(
+                StorageSetKey.TRANSACTIONS, table_name, "transactions_source"
```
andriisoldatenko, line 38:
```suggestion
                StorageSetKey.TRANSACTIONS, table_name, "transaction_source"
```

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
```diff
@@ -0,0 +1,52 @@
...
+    def __forward_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name=table_name,
+                column=Column(
+                    "transaction_source",
+                    String(Modifiers(nullable=True, low_cardinality=True)),
```
lynnagara, line 27: The linked processor PR is defaulting to an empty string. So I don't think this needs to be nullable. Using nullable takes more storage space and thus affects performance.

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
```diff
@@ -0,0 +1,52 @@
...
+from snuba.clickhouse.columns import Column, String
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    The source column is required to query for transaction events
+    in the transaction summary when the transaction name has been
```
evanh, line 12: How is this going to be used? In every query on the transactions summary page?

lynnagara (commented):
If you are planning to query this column directly please also add it to the transactions schema here https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/transactions_common.py

Also, what does transactions source represent? What kind of values are we actually expecting here? Is it something end users provide? (basically, how do we know it's low cardinality)

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
andriisoldatenko, line 27:
```suggestion
                    String(Modifiers(low_cardinality=True)),
```

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
andriisoldatenko: @jan-auer  can you please help here to answer?

andriisoldatenko:
> If you are planning to query this column directly please also add it to the transactions schema here https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/transactions_common.py
> 
> Also, what does transactions source represent? What kind of values are we actually expecting here? Is it something end users provide? (basically, how do we know it's low cardinality)

Example values:

image

https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
andriisoldatenko: fixed.

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
jan-auer: This will be used to render a special version of the transaction summary page where we don't query transaction events by transaction name, but rather by source. It is to collect all the transaction events where we dropped the transaction name during metric extraction.

The majority of transaction summary pages will still use the regular transaction names, so this is for the cases where we get unsanitized URLs from browsers, and such. 

cc @k-fish for details.

jan-auer:
> If you are planning to query this column directly please also add it to the transactions schema here https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/transactions_common.py

Since this was also mentioned by @evanh in https://github.com/getsentry/snuba/pull/2985#pullrequestreview-1052690244, should that go in this or processor PR?

evanh:
> should that go in this or processor PR?

I would put it in the processor PR. It's a little cleaner to semantically have one PR that is a migration, and another that enables access to the new column.

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
evanh: Let me give more context: I ask because if you are using this column to filter every/a lot of queries on the performance page, then it might be worth also adding an index of some kind. But this is also a "potential future problem" more than a "fix in this PR" problem.

andriisoldatenko: @evanh do you want me to add TODO or create a ticket to not forget about it?

lynnagara: Just my 2c - Based on the comment below this is an extremely low cardinality column with only 7 unique values. Low cardinality column type should already be fairly optimized and I don't know that an index here would really improve much (Clickhouse only has data skipping indexes). I wouldn't worry about it unless it's a problem.

evanh: I don't think there's any further action here. Let's see how it works and if there's a performance issue we can revisit it.

evanh (approved):
Looks good to me. @fpacifici or @nikhars can help you get this into prod.

andriisoldatenko:
```
snuba migrations reverse --group transactions --migration-id 0015_transactions_add_source_column --dry-run
```

**Local operations:**

```
ALTER TABLE transactions_local DROP COLUMN IF EXISTS transaction_source;


Dist operations:
```

snuba/migrations/snuba_migrations/transactions/0015_transactions_add_source_column.py
k-fish: For usage context;

Upon LA we'll be either querying this across all transactions on transaction summary, or potentially directing people to a discover page to list their 'high cardinality' transactions that need to be manually fixed. We aren't planning on querying this by default on any views, only for the person who is wondering what 'unparameterized' means on their metrics data and clicks something to look at it.

nikhars:
Migration has been applied to production

PR #2983, opened 2022-07-27
feat(profiling): add architecture column to profiles_local table

viglia:
This PR adds a new `architecture` column to the `profiles_local` table.

`architecture` is a string describing the architecture of the CPU that is used in the device that collected the specific profile.

### DIFF
--- a/snuba/migrations/snuba_migrations/profiles/0003_add_device_architecture.py
+++ b/snuba/migrations/snuba_migrations/profiles/0003_add_device_architecture.py
@@ -16,7 +16,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 table_name="profiles_local",
                 column=Column(
                     "architecture",
-                    String(Modifiers(low_cardinality=True, nullable=True)),
+                    String(Modifiers(low_cardinality=True, default="'unknown'")),
                 ),
                 after="device_os_version",
             )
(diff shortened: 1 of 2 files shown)


evanh (approved):

snuba/migrations/snuba_migrations/profiles/0003_add_device_architecture.py
```diff
@@ -0,0 +1,38 @@
...
+                table_name="profiles_local",
+                column_name="architecture",
+            )
+        ]
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
+
+    def backwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
```
nikhars, line 38: Is there a reason why the column isn't added to the distributed tables? In production, all queries go through the distributed tables. If you don't have the column defined in the distributed table you won't be able to access the column in the local tables

nikhars (changes requested):

snuba/migrations/snuba_migrations/profiles/0003_add_device_architecture.py
viglia: @nikhars I was looking at a previous migration ([002](https://github.com/getsentry/snuba/blob/master/snuba/migrations/snuba_migrations/profiles/0002_disable_vertical_merge_algorithm.py)) as a reference.

Thanks for pointing that out.

Zylphrex (approved):
LGTM

nikhars (approved):

viglia:
The migration has been applied successfully. I'm going to merge this.

PR #2985, opened 2022-07-27
feat(transactions): Start processing `transaction_source` in transactions_processor.py

andriisoldatenko:
Start processing new transaction annotation:

```
{ "transaction_info": {"source": "url"} }
```

https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

and storing  it using new column `transaction_source` from https://github.com/getsentry/snuba/pull/2982

**PLEASE DON'T MERGE** until https://github.com/getsentry/snuba/pull/2982

resolves [TET-270]

[TET-270]: https://getsentry.atlassian.net/browse/TET-270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/snuba/datasets/transactions_processor.py
+++ b/snuba/datasets/transactions_processor.py
@@ -110,6 +110,9 @@ class TransactionsMessageProcessor(MessageProcessor):
         processed["transaction_name"] = _unicodify(
             event_dict["data"].get("transaction") or ""
         )
+        processed["transaction_source"] = _unicodify(
+            event_dict["data"].get("transaction_info", {}).get("source") or ""
+        )
         processed["start_ts"], processed["start_ms"] = self.__extract_timestamp(
             event_dict["data"]["start_timestamp"],
         )
--- a/snuba/datasets/transactions_processor.py
+++ b/snuba/datasets/transactions_processor.py
@@ -111,7 +111,7 @@ class TransactionsMessageProcessor(MessageProcessor):
             event_dict["data"].get("transaction") or ""
         )
         processed["transaction_source"] = _unicodify(
-            event_dict["data"].get("transaction_info", {}).get("source") or ""
+            (event_dict["data"].get("transaction_info") or {}).get("source") or ""
         )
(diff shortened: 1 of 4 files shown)


snuba/datasets/transactions_processor.py
```diff
@@ -110,6 +110,9 @@ def _process_base_event_values(
         processed["transaction_name"] = _unicodify(
             event_dict["data"].get("transaction") or ""
         )
+        processed["transaction_source"] = _unicodify(
+            event_dict["data"].get("transaction_info", {}).get("source") or ""
```
jan-auer, line 114: Just to be extra careful, could we write it like this?

```suggestion
            (event_dict["data"].get("transaction_info") or {}).get("source") or ""
```

That would also cover the case where the dict contains an explicit `None`.

andriisoldatenko:
> Looks good. I may have forgotten to mention, in order to expose this column in the data model, you will need to add a column here: https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/transactions_common.py#L40

Done.

evanh (approved):

nikhars (approved):
The migration has been applied to production. Please merge and deploy during SnS working hours (PST).

andriisoldatenko:
@nikhars  are we good to merge it today?

nikhars:
> @nikhars are we good to merge it today?

Let me do a sanity check of the clusters to ensure the migration has happened on both the clusters. Want to be cautious since there are 2 clusters right now.

nikhars:
> > @nikhars are we good to merge it today?
> 
> Let me do a sanity check of the clusters to ensure the migration has happened on both the clusters. Want to be cautious since there are 2 clusters right now.

Verified. Feel free to merge now.

andriisoldatenko:
@nikhars  done!

PR #2986, opened 2022-07-27
ref(replays): add new topic to validation prior to consumer change

JoshFerge:
- adds a new topic to validation.py, so we can deploy snuba with the new topic configured on the ops side.

After that, we can deploy the consumer to listen to the topic. https://github.com/getsentry/snuba/pull/2978


mwarkentin:
@JoshFerge we've got this + another topic in https://github.com/getsentry/snuba/pull/2987

JoshFerge:
> @JoshFerge we've got this + another topic in #2987

ah okay cool!

PR #2988, opened 2022-07-27
feat: Sort raw sql fields and conditions [Experiment]

enochtangg:
Context:
To help improve the frequency of raw sql cache hits, this PR is
responsible for sorting column fields and expression conditions in raw
SQL queries. Note: this code should not interferer with the current 
operations of Snuba query caching. This is an experiment to investigate 
the consistency of the natural ordering of columns and conditions.

Test Plan:
* Provide functionality for sorting, and select a certain percentage
  (defined in runtime config) of raw queries to be part of the
  experiment.
* Exclude queries from an explicit set of referrers to avoid
  some computational overhead.
* Record metrics of cache hits of sorted and unsorted queries.
* Determine whether or not sorting should be implemented for all queries.

Sorting is *only* to expressions simple Columns in SELECT clause and conditionals (or nested) in the WHERE clause

### DIFF
--- a/snuba/clickhouse/formatter/query.py
+++ b/snuba/clickhouse/formatter/query.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from typing import Optional, Sequence, Type, Union
 
 from snuba.clickhouse.formatter.expression import (
@@ -28,7 +29,7 @@ from snuba.query.parsing import ParsingContext
 FormattableQuery = Union[Query, CompositeQuery[Table]]
 
 
-def format_query(query: FormattableQuery) -> FormattedQuery:
+def format_query(query: FormattableQuery, sort_fields: bool = False) -> FormattedQuery:
     """
     Formats a Clickhouse Query from the AST representation into an
     intermediate structure that can either be serialized into a string
@@ -37,11 +38,17 @@ def format_query(query: FormattableQuery) -> FormattedQuery:
     This is the entry point for any type of query, whether simple or
     composite.
     """
-    return FormattedQuery(_format_query_content(query, ClickhouseExpressionFormatter))
+    return FormattedQuery(
+        _format_query_content(query, ClickhouseExpressionFormatter, sort_fields)
+    )
 
 
-def format_query_anonymized(query: FormattableQuery) -> FormattedQuery:
-    return FormattedQuery(_format_query_content(query, ExpressionFormatterAnonymized))
+def format_query_anonymized(
+    query: FormattableQuery, sort_fields: bool = False
+) -> FormattedQuery:
+    return FormattedQuery(
+        _format_query_content(query, ExpressionFormatterAnonymized, sort_fields)
+    )
 
 
 class DataSourceFormatter(DataSourceVisitor[FormattedNode, Table]):
@@ -87,6 +94,7 @@ class DataSourceFormatter(DataSourceVisitor[FormattedNode, Table]):
 def _format_query_content(
     query: FormattableQuery,
     expression_formatter_type: Type[ExpressionFormatterBase],
+    sort_fields: bool = False,
 ) -> Sequence[FormattedNode]:
     """
     Produces the content of the formatted query.
@@ -101,7 +109,7 @@ def _format_query_content(
     return [
         v
         for v in [
-            _format_select(query, formatter),
+            _format_select(query, formatter, sort_fields),
             PaddingNode(
                 "FROM",
                 DataSourceFormatter(expression_formatter_type).visit(
@@ -109,12 +117,18 @@ def _format_query_content(
                 ),
             ),
             _format_arrayjoin(query, formatter),
-            _build_optional_string_node("PREWHERE", query.get_prewhere_ast(), formatter)
+            _build_optional_string_node(
+                "PREWHERE", query.get_prewhere_ast(), formatter, sort_fields
+            )
             if isinstance(query, Query)
             else None,
-            _build_optional_string_node("WHERE", query.get_condition(), formatter),
+            _build_optional_string_node(
+                "WHERE", query.get_condition(), formatter, sort_fields
+            ),
             _format_groupby(query, formatter),
-            _build_optional_string_node("HAVING", query.get_having(), formatter),
+            _build_optional_string_node(
+                "HAVING", query.get_having(), formatter, sort_fields
+            ),
             _format_orderby(query, formatter),
             _format_limitby(query, formatter),
             _format_limit(query, formatter),
@@ -124,11 +138,13 @@ def _format_query_content(
 
 
 def _format_select(
-    query: AbstractQuery, formatter: ExpressionVisitor[str]
+    query: AbstractQuery, formatter: ExpressionVisitor[str], sort_fields: bool = False
 ) -> StringNode:
     selected_cols = [
         e.expression.accept(formatter) for e in query.get_selected_columns()
     ]
+    if sort_fields:
+        selected_cols = sorted(selected_cols)
     return StringNode(f"SELECT {', '.join(selected_cols)}")
 
 
@@ -136,12 +152,18 @@ def _build_optional_string_node(
     name: str,
     expression: Optional[Expression],
     formatter: ExpressionVisitor[str],
+    sort_fields: bool = False,
 ) -> Optional[StringNode]:
-    return (
-        StringNode(f"{name} {expression.accept(formatter)}")
-        if expression is not None
-        else None
-    )
+    if expression is not None:
+        expression_string = expression.accept(formatter)
+        # Sorting will not work if sub-queries in expressions becomes supported in the future
+        if sort_fields:
+            expression_list = re.split("(?=AND)|(?=OR)", expression_string)
+            expression_list[1:] = sorted(expression_list[1:])
+            expression_string = "".join(expression_list)
+        return StringNode(f"{name} {expression_string}")
+    else:
+        return None
 
 
 def _format_groupby(
--- a/snuba/clickhouse/formatter/query.py
+++ b/snuba/clickhouse/formatter/query.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import re
 from typing import Optional, Sequence, Type, Union
 
 from snuba.clickhouse.formatter.expression import (
@@ -103,13 +102,13 @@ def _format_query_content(
     Should we have more differences going on we should break this
     method into smaller ones.
     """
-    parsing_context = ParsingContext()
+    parsing_context = ParsingContext(sort_fields=sort_fields)
     formatter = expression_formatter_type(parsing_context)
 
     return [
         v
         for v in [
-            _format_select(query, formatter, sort_fields),
+            _format_select(query, formatter),
             PaddingNode(
                 "FROM",
                 DataSourceFormatter(expression_formatter_type).visit(
@@ -117,18 +116,12 @@ def _format_query_content(
                 ),
             ),
             _format_arrayjoin(query, formatter),
-            _build_optional_string_node(
-                "PREWHERE", query.get_prewhere_ast(), formatter, sort_fields
-            )
+            _build_optional_string_node("PREWHERE", query.get_prewhere_ast(), formatter)
             if isinstance(query, Query)
             else None,
-            _build_optional_string_node(
-                "WHERE", query.get_condition(), formatter, sort_fields
-            ),
+            _build_optional_string_node("WHERE", query.get_condition(), formatter),
             _format_groupby(query, formatter),
-            _build_optional_string_node(
-                "HAVING", query.get_having(), formatter, sort_fields
-            ),
+            _build_optional_string_node("HAVING", query.get_having(), formatter),
             _format_orderby(query, formatter),
             _format_limitby(query, formatter),
             _format_limit(query, formatter),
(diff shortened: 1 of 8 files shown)


snuba/clickhouse/formatter/query.py
```diff
@@ -37,11 +38,17 @@ def format_query(query: FormattableQuery) -> FormattedQuery:
     This is the entry point for any type of query, whether simple or
     composite.
     """
-    return FormattedQuery(_format_query_content(query, ClickhouseExpressionFormatter))
+    return FormattedQuery(
```
evanh, line 41: Your current implementation doesn't handle complex expressions properly. Firstly, it doesn't sort all the parts of the query (group by, order by). Also because it is relying on string manipulation to sort expressions, it is going to improperly sort a lot of cases or just break the query. You are not handling nested conditions for example: 

```
"WHERE d = 1 AND c = 2 AND (b = 3 OR a = 4)" 
# becomes
["WHERE d = 1", "AND (b = 3 ", "AND c = 2 ", "OR a = 4)"]
```
This isn't just incorrectly sorted, it is also breaking the query by splitting the expression in the parentheses.

I don't think this is the best way of structuring this code. A cleaner way would be to extend `ClickhouseExpressionFormatter`, in a similar way that the  `ExpressionFormatterAnonymized` does. Then you could sort the expressions before passing them to the parent function to stringify. The `sort_fields` bool here would determine which formatter to use.

snuba/clickhouse/formatter/query.py
```diff
@@ -124,24 +138,32 @@ def _format_query_content(
...
-    return (
-        StringNode(f"{name} {expression.accept(formatter)}")
-        if expression is not None
-        else None
-    )
+    if expression is not None:
+        expression_string = expression.accept(formatter)
+        # Sorting will not work if sub-queries in expressions becomes supported in the future
+        if sort_fields:
+            expression_list = re.split("(?=AND)|(?=OR)", expression_string)
```
evanh, line 161: FYI, best practice is to compile regexes globally, then use the compiled version here.

snuba/web/db_query.py
```diff
@@ -388,6 +388,37 @@ def get_query_cache_key(formatted_query: FormattedQuery) -> str:
...
+        key_sorted = get_query_cache_key(formatted_query_sorted)
+        percentage = state.get_config(
+            "sort_sql_fields_and_conditions_rollout_percentage", 0
+        )
+        assert isinstance(percentage, int)
+        if percentage > hash_to_probability(key_sorted):
+            key_sorted_with_prefix = f"sorted:{key_sorted}"
+            result = cache_partition.get(key_sorted_with_prefix)
+            if result is not None:
+                write_query_cache_hit_metric("sorted", reader, key_sorted)
```
evanh, line 405: This shouldn't happen here. There should be one single call to the metrics, after both sorted and unsorted caches have been checked. We want to look at the intersection between the two results, and if you log them separately that won't work.

I would change this function to return the results of the cache check instead.

snuba/web/db_query.py
```diff
@@ -530,6 +567,19 @@ def record_cache_hit_type(hit_type: int) -> None:
     )
 
 
+def write_query_cache_hit_metric(
+    is_sorted: str, reader: Reader, key_sorted: str
+) -> None:
+    metrics.increment(
+        "snuba.api.query.cache_hit",
```
evanh, line 574: Please a different prefix from `snuba.api.query`. Otherwise you will collide with the other `.query` metrics which have specific meaning (tracking performance of different parts of our query pipeline).

snuba/web/db_query.py
```diff
@@ -530,6 +567,19 @@ def record_cache_hit_type(hit_type: int) -> None:
...
+        tags={
+            "partition_id": reader.cache_partition_id or "default",
```
evanh, line 576: Why do you need this?

snuba/web/db_query.py
```diff
@@ -530,6 +567,19 @@ def record_cache_hit_type(hit_type: int) -> None:
...
+            "query_info": is_sorted,
+            "key": key_sorted,
```
evanh, line 578: You cannot use `query_id`  as a key here. Datadog only allows ~300 unique tag values in a metric in total. Please remove this.

snuba/web/query.py
```diff
@@ -50,6 +50,16 @@
 metrics = MetricsWrapper(environment.metrics, "api")
 
 MAX_QUERY_SIZE_BYTES = 256 * 1024  # 256 KiB by default
+EXCLUDED_REFERRERS_FROM_FIELDS_SORTING = {
```
evanh, line 53: How did you determine this list?

snuba/web/query.py
```diff
@@ -300,6 +310,13 @@ def _format_storage_query_and_run(
         _apply_turbo_sampling_if_needed(clickhouse_query, query_settings)
 
         formatted_query = format_query(clickhouse_query)
+        formatted_query_sorted = None
+
+        # Build formatted query with sorted fields and expressions
+        if referrer not in EXCLUDED_REFERRERS_FROM_FIELDS_SORTING:
+            formatted_query_sorted = format_query(clickhouse_query, sort_fields=True)
+
+        # do modular hash algorithm
```
evanh, line 319: Why is this comment here?

evanh (changes requested):

snuba/web/db_query.py
enochtangg: Metric name updated to `sorted_and_unsorted_cache_hit`.

snuba/web/db_query.py
enochtangg: Updated as suggested!

snuba/clickhouse/formatter/query.py
enochtangg: Good point. I revised the sorting implementation to be conducted in the AST level. The query is now sorted based on an added attribute on the `ParsingContext` class. The actual sorting is now done at each recursive level when traversing the expressions in the FunctionCalls.

Additionally, since we are potentially sorting items of different types, the following ordering was created to ensure some level of consistency among classes: Literals < Columns < FunctionCalls (nested expressions). Note that we only sort string literals, else we default.

snuba/web/query.py
enochtangg: This list was chosen based on the highest average queries by referrers to our cache. Removing them (along with using a variable percentage of queries to roll-out this experiment, it will help ensure our cache does not become overloaded. Do you have suggestions/more context to what should be removed or added?

snuba/web/db_query.py
```diff
@@ -388,6 +388,39 @@ def get_query_cache_key(formatted_query: FormattedQuery) -> str:
...
+    formatted_query_sorted: Optional[FormattedQuery], reader: Reader
+) -> bool:
+    cache_partition = _get_cache_partition(reader)
+    if formatted_query_sorted:
+        key_sorted = get_query_cache_key(formatted_query_sorted)
+        percentage = state.get_config(
+            "sort_sql_fields_and_conditions_rollout_percentage", 0
+        )
+        assert isinstance(percentage, int)
+        if percentage > hash_to_probability(key_sorted):
```
evanh, line 401: Should this be `>` or `<`? If we have an experiment set to "10%", by convention that means 10% is _in_ the experiment, and 90% is the control.

snuba/web/db_query.py
```diff
@@ -497,6 +538,8 @@ def record_cache_hit_type(hit_type: int) -> None:
         if hit_type == RESULT_VALUE:
             stats["cache_hit"] = 1
             span_tag = "cache_hit"
+            if sorted_key_exists:
+                metrics.increment("sorted_and_unsorted_cache_hit")
```
evanh, line 542: You need to also track the other cases of this experiment. There should be 4:

**Sorted/Unsorted**
Miss/Miss
Miss/Hit
Hit/Miss <- the one we care about most
Hit/Hit

evanh (commented):
I mentioned this to you in person, but please add some unit tests to test your sorting logic on expressions. Some examples:

Regular tests: https://github.com/getsentry/snuba/blob/master/tests/query/test_expressions.py
Parameterized tests: https://github.com/getsentry/snuba/blob/master/tests/datasets/validation/test_no_time_condition_validator.py

snuba/web/db_query.py
enochtangg: Rewritten to `hash_to_probability(key_sorted) < percentage` for clarity. This means that if the query's random probability is *within* the experiment percentage range (e.g. 10%), we check the sorted cache.

snuba/query/expressions.py
```diff
@@ -417,6 +472,30 @@ def functional_eq(self, other: Expression) -> bool:
                 return False
         return True
 
+    def __gt__(self, other: Any) -> bool:
+        if (
```
evanh, line 476: This could also check the name of the function itself. `and` > `or`.

snuba/web/query.py
evanh: I would also add `group.unhandled-flag`. If we think of any others we can add them later.

snuba/query/expressions.py
```diff
@@ -85,6 +86,32 @@ def functional_eq(self, other: Expression) -> bool:
...
+        elif isinstance(self, Literal):
+            return relate[op]
+        elif isinstance(other, Literal):
+            return not relate[op]
+        elif isinstance(self, FunctionCall) and isinstance(other, FunctionCall):
+            if not self.parameters:
+                return relate[op]
+            elif not other.parameters:
+                return not relate[op]
+            elif isinstance(self.parameters[0], Column) and isinstance(
```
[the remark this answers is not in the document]
enochtangg: Moved empty parameter logic to FunctionCall.

snuba/query/expressions.py
```diff
@@ -85,6 +86,32 @@ def functional_eq(self, other: Expression) -> bool:
         """
         raise NotImplementedError
 
+    # Sort by Class first and then alphabetically
+    # Class order: Literals < Columns < FunctionCalls (nested)
+    def compare_expressions(
+        self, other: Expression, op: Callable[[Any, Any], Any]
+    ) -> bool:
+        relate = {operator.lt: True, operator.gt: False}
+
```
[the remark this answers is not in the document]
enochtangg: To your points:

1. When only one side is a Column, this would imply that the other side is a literal (at least in the test cases). The recursive `.compare_expressions()` is called and orders them accordingly. However, this implication might not always be true outside of test cases and if we decide to support more types in the future. I will add that in.

2. When comparing two FunctionCalls, we're actually comparing the first two objects in the first index of the tuple. For example:
* A = FunctionCall([Column('b') = 1, Column('d') = 1])
* B = FunctionCall([Column('a') = 1, Column('c') = 1])

In this case, B < A because Column('a') is < Column('b').
My initial thought was to keep this logic outside of FunctionCall and put it in the parent class for isolation. I agree that it makes more sense for the parent function to solely be used to compare two different types. However, this requires `FunctionCall` to know about its sibling classes `Column` and `Literal`. I've made this change to reflect this.

snuba/web/db_query.py
```diff
@@ -388,6 +388,48 @@ def get_query_cache_key(formatted_query: FormattedQuery) -> str:
...
+                    "profile": None,
+                    "trace_output": "",
+                }
+                cache_partition.set(key_sorted_with_prefix, empty_result)
+                return False
+    return False
+
+
+def hash_to_probability(hexadecimal: str) -> float:
+    MAX_HASH_PLUS_ONE = 16 ** len(hexadecimal)
```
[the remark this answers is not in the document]
enochtangg: To get the max float value of the MD5 hash, we convert the 32-character hexadecimal to decimal by raising 16 to power of 32 (length of MD5). Since the MD5 hash returns a uniformly distributed number from 0 to the max float value, we just convert the hexadecimal to a float and divide by the max value to get a random decimal from [0,1).

My initial thought was to calculate it each time in the case we switch hashing algorithms (e.g. SHA-256) in the future, but I guess that will be unlikely. Updated as global.

snuba/query/expressions.py
```diff
@@ -417,6 +467,54 @@ def functional_eq(self, other: Expression) -> bool:
...
+    def __gt__(self, other: Any) -> bool:
+        if isinstance(other, FunctionCall):
+            if (
+                self.parameters
+                and other.parameters
+                and isinstance(self.parameters[0], FunctionCall)
+                and isinstance(other.parameters[0], FunctionCall)
+            ):
+                return self.parameters[0] > (other.parameters[0])
+            elif self.parameters and not other.parameters:
```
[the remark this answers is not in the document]
enochtangg: Updated to follow this. Added operator overrides to Expression parents class.

snuba/query/expressions.py
```diff
@@ -417,6 +485,50 @@ def functional_eq(self, other: Expression) -> bool:
                 return False
         return True
 
+    def __gt__(self, other: Any) -> bool:
```
evanh, line 488: This is a lot of code that is duplicated twice. I think you can move all of this into one function, use the `operator` module to collapse this down.

An overly simplified example of what I mean:
```
def _gt_lt_helper(self, other, op, other_op):
    ...
    if op(self.function_name, other.function_name):
        return False
    elif other_op(self.function_name, other.function_name):
        return True
    ...

def __gt__(self, other):
    return self._gt_lt_helper(other, operator.gt, operator.lt)
...
```

snuba/web/db_query.py
```diff
@@ -59,6 +59,7 @@
 )
 from snuba.web import QueryException, QueryResult
 
+MAX_HASH_PLUS_ONE = 16**32
```
evanh, line 62: Please put a comment explaining what this constant is (max value of a md5).

evanh (commented):
Almost there!

snuba/web/db_query.py
```diff
@@ -459,15 +502,20 @@ def execute_query_with_caching(
         key = get_query_cache_key(formatted_query)
         clickhouse_query_settings["query_id"] = key
         if use_cache:
+            sorted_key_exists = check_sorted_sql_key_in_cache(
```
evanh, line 505: Can you wrap this in a try/except? I don't want any transient Redis issues here to cause query failures.

evanh (approved):
One last thing for safety, otherwise I think we are ready to test this!

snuba/web/db_query.py
enochtangg: Added try/except. Also added some logic for handling whether or not to write the metric depending on if the query was selected for the experiment

PR #2989, opened 2022-07-27
ref: re-organize ArrayJoinKeyValueOptimizer tests

onewland:
I need to make a change to ArrayJoinKeyValueOptimizer to support two value columns. To make that cleaner, I figured I would start by reorganizing this test file based on a "HACK" comment left by @evanh


evanh (approved):
Thanks!

PR #2990, opened 2022-07-27
fix(config): Add comment identifying experimental consumers

rahul-kumar-saini:
- There is no way to know whether a consumer is experimental without diffing the two config files.
- Added a comment identifying split between stable and experimental consumers in the config

### DIFF
--- a/.freight.yml
+++ b/.freight.yml
@@ -4,77 +4,78 @@ sentry:
   repository: getsentry/snuba
 
 steps:
-- kind: KubernetesDeployment
-  selector:
-    label_selector: service=snuba
-  containers:
-  - image: us.gcr.io/sentryio/snuba:{sha}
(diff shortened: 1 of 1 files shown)


rahul-kumar-saini:
Not sure why the diff isn't realizing there is just whitespace change but that's just my formatter, only real change is I moved the entry for the profiles consumer down to the bottom and added 1 line for the comment.

lynnagara:
Why changing the indentation?

lynnagara (approved):
Hard to read the diff but I'll take your word those are the only 2 things that changed 🙂

rahul-kumar-saini:
> Why changing the indentation?

Auto format on save - I assumed GitHub would show just whitespace added and not the entire line changed..

PR #2991, opened 2022-07-27
feat(profiling): Add thread id to functions table

Zylphrex:
The example profile id needs to be paired with a thread id. This introduces the
thread id column and rematerialize the view.


snuba/migrations/snuba_migrations/functions/0002_add_thread_id.py
```diff
@@ -0,0 +1,180 @@
...
+                table_name=self.local_raw_table,
+                column=Column("thread_id", UInt(64)),
+                after="profile_id",
+            ),
+            operations.DropTable(
+                storage_set=self.storage_set, table_name=self.local_materialized_table
+            ),
+            operations.DropTable(
+                storage_set=self.storage_set, table_name=self.local_view_table
+            ),
```
Zylphrex, line 80: Not sure if there's a better way to do this but dropping the existing data is not a problem right now.

snuba/migrations/snuba_migrations/functions/0002_add_thread_id.py
```diff
@@ -0,0 +1,180 @@
...
+                countState(arrayJoin(durations) AS duration) AS count,
+                quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)(duration) AS percentiles,
+                minState(duration) AS min,
+                maxState(duration) AS max,
+                avgState(duration) AS avg,
+                sumState(duration) AS sum,
+                argMaxState((profile_id, thread_id), duration) as worst,
+                groupUniqArrayState(5)((profile_id, thread_id)) as examples
+            FROM {self.local_raw_table}
+            WHERE materialization_version = 1
```
Zylphrex, line 162: Bumping the materialization version here as we only want the rows that have `thread_id` populated

snuba/migrations/snuba_migrations/functions/0002_add_thread_id.py
```diff
@@ -0,0 +1,180 @@
...
+                    ttl="timestamp + toIntervalDay(retention_days)",
+                ),
+            ),
+            operations.CreateMaterializedView(
+                storage_set=self.storage_set,
+                view_name=self.local_view_table,
+                destination_table_name=self.local_materialized_table,
+                columns=columns,
+                query=self.__MATVIEW_STATEMENT,
+            ),
```
Zylphrex, line 100: This is mostly identical to what it was previously. A few notable changes
- the worst + examples aggregations now return a tuple containing the profile id + thread id
- the primary key has been cut down to exclude some of the higher cardinality columns like fingerprint
- bumped the materialization version from 0 to 1 for the new data

PR #3001, opened 2022-07-29
feat(replays): add error_ids column

JoshFerge:
- add error ids UUID Array column for storing error ids associated with a replay.

### DIFF
--- a/snuba/datasets/replays_processor.py
+++ b/snuba/datasets/replays_processor.py
@@ -34,7 +34,8 @@ from sentry_sdk import capture_exception
 ReplayEventDict = Mapping[Any, Any]
 RetentionDays = int
 
-
+# A sane upper bound of 1000 was chosen for error_ids. If you know better, change it.
+ERROR_IDS_LIMIT = 1000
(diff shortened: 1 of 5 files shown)


evanh (commented):
Couple questions:

- Is there a limit to how many error IDs can be in the array?
- Why is the index being added?

cmanallen:
@evanh Hey, I'm taking over this PR.

> Is there a limit to how many error IDs can be in the array?

This field will likely be empty in most cases and then have a long tail of gradually decreasing probably for the number of errors found.

While I'm not an expert, I think its possible a large number could exist.  Someone could throw `n` malicious errors and whatever the SDK processes is what we receive.  Does that make sense?  We could cap this in the consumer if you foresee a problem.

> Why is the index being added?

When looking at an error you'll want to see if there was a replay associated with it.  Definitely a needle in a haystack.  The bloom filter will facilitate faster lookup.

evanh:
> We could cap this in the consumer if you foresee a problem.

I don't foresee a problem, but I would advise against having unbounded fields. You can make it quite high (1000?).

cmanallen (approved):

evanh (approved):

PR #3002, opened 2022-07-29
feat(admin): Allow describe table command

nikhars:
Once a migration has been run in production I would like to have an
easy way to find out if the schema has been applied correctly across
all nodes of the cluster. Being able to do so via the admin tool
allows me not to ssh into each individual machine.

The describe table is pretty non-intrusive and just shows the schema
of the table. Hence no custom validation is needed. More information
about the command can be found here https://clickhouse.com/docs/en/sql-reference/statements/describe-table

### Legal Boilerplate


lynnagara (approved):
Thanks, this is useful!

PR #3014, opened 2022-08-01
config(mep): add generic metrics schedulers/executors to freight

onewland:
These are deployed in prod but we want freight to deploy them automatically


nikhars (approved):

rahul-kumar-saini:
Should add these to .freight.stable.yml as well

onewland:
@rahul-kumar-saini will do, but let's watch/verify this deploys successfully first

PR #3016, opened 2022-08-01
fix(admin): Fix deletion of configs with `/` in them

nikhars:
The query_settings are configured with forward slashes in them. In
order for flask route's to recognize them correctly we need to specify
them explicitly as path which can have forward slashes.


onewland (approved):
You could add a slash-containing test case here if you wanted:

https://github.com/getsentry/snuba/blob/master/tests/admin/test_api.py#L100-L123

but you don't need to

PR #3018, opened 2022-08-02
config(ds): get rid of existing logic to move to dynamic sampling

onewland:
According to @untitaker `parent_sampled` is already a special case in the python SDK (see internal slack thread https://sentry.slack.com/archives/C02N6582ZFA/p1659476660850549?thread_ts=1659473664.627579&cid=C02N6582ZFA for more info) and traces with this context set will never be rejected, even with a sample rate of 0.

We'd like to start testing service-side sampling, and this would unblock us from that.

This should be deployed with an eye on our sentry project to make sure that our assumption is correct

### DIFF
--- a/snuba/environment.py
+++ b/snuba/environment.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 
 import logging
 import os
-from typing import Any, Mapping, Optional
+from typing import Optional
 
 import sentry_sdk
 from sentry_sdk.integrations.flask import FlaskIntegration
@@ -24,10 +24,6 @@ def setup_logging(level: Optional[str] = None) -> None:
     )
 
 
-def traces_sampler(sampling_context: Mapping[str, Any]) -> Any:
-    return sampling_context["parent_sampled"] or False
-
-
 def setup_sentry() -> None:
     sentry_sdk.init(
         dsn=settings.SENTRY_DSN,
@@ -38,7 +34,7 @@ def setup_sentry() -> None:
             RedisIntegration(),
         ],
         release=os.getenv("SNUBA_RELEASE"),
-        traces_sampler=traces_sampler,
+        traces_sample_rate=0,
     )
 
 
--- a/snuba/environment.py
+++ b/snuba/environment.py
@@ -34,7 +34,7 @@ def setup_sentry() -> None:
             RedisIntegration(),
         ],
         release=os.getenv("SNUBA_RELEASE"),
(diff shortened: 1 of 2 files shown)


snuba/environment.py
```diff
@@ -38,7 +34,7 @@ def setup_sentry() -> None:
             RedisIntegration(),
         ],
         release=os.getenv("SNUBA_RELEASE"),
-        traces_sampler=traces_sampler,
+        traces_sample_rate=0,
```
MeredithAnya, line 37: Okay so having the `traces_sample_rate=0` is the same as the `traces_sampler` return `False` correct?  

Do you know if `sampling_context["parent_sampled"]` was even doing anything to begin with? 

Do we know what

onewland: Your understanding is correct, according to Markus the default SDK/ingest behavior was to send if `sampling_context["parent_sampled"]` was truthy and default to not sending a trace otherwise.

untitaker: yup exactly

nikhars:
This might not be relevant but wanted to bring it up. Since Snuba is open source and run in both Self hosted and Single Tenant, are there any implications of making this change to those environments as well?

untitaker:
This particukar change shouldn't have any implications tbh. I think the
question is valid for when you have to turn up the sample rate (which the
dynamic sampling flow will ask you for), and I don't have an answer to that

onewland:
> This might not be relevant but wanted to bring it up. Since Snuba is open source and run in both Self hosted and Single Tenant, would there be any impact to those environments with this change?

I think this is not a huge issue because part of dynamic sampling is being able to change the client/server sample rates in the Sentry UI without updating the client config. So if self-hosted/ST users want a non-zero sample rate they don't have to adjust this file at all.

If you're concerned about making it more flexible maybe I could make that 0 come from settings.py?

MeredithAnya:
@onewland If the future is to always rely on dynamic sampling, it seems fine to me to leave `0` there, but if it's reasonable for folks to not want to use dynamic sampling, I think it makes sense to put it in as a setting.

untitaker:
Dynamic sampling flow will ask you to turn up the sample rate in the SDK. There will be onprem users that complain that they're now processing and storing more events because they won't have the same sample rates set up as you do. So I think it does make sense to move this configuration out of onprem, but I think this particular PR is fine for onprem users.

nikhars (approved):

PR #3020, opened 2022-08-03
fix(discover): add missing transaction_source to TRANSACTIONS_COLUMNS

andriisoldatenko:
After working on https://github.com/getsentry/sentry/pull/37295, 
@wmak suggested to add `transaction_source` to `TRANSACTIONS_COLUMNS` so 
we can query it from discover with this patch:
```diff
--- a/src/sentry/snuba/events.py
+++ b/src/sentry/snuba/events.py
@@ -466,8 +466,7 @@ class Columns(Enum):
         group_name=None,
         event_name=None,
         transaction_name="transaction_source",
-        # Only available in transactions, pretend like its a tag so we dont error and just null instead
-        discover_name="tags[transaction_source]",
+        discover_name="transaction_source",
         alias="transaction.source",
     )
```

### DIFF
--- a/snuba/datasets/entities/discover.py
+++ b/snuba/datasets/entities/discover.py
@@ -249,6 +249,7 @@ TRANSACTIONS_COLUMNS = ColumnSet(
         ("transaction_hash", UInt(64, Modifiers(nullable=True))),
         ("transaction_op", String(Modifiers(nullable=True))),
         ("transaction_status", UInt(8, Modifiers(nullable=True))),
+        ("transaction_source", String(Modifiers(nullable=True))),
         ("duration", UInt(32, Modifiers(nullable=True))),
         ("measurements", Nested([("key", String()), ("value", Float(64))])),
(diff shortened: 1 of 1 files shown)


lynnagara (changes requested):
I don't really get how this works without the column also being present in the errors table. Wouldn't it throw an error due to the column being missing if if you tried to query the column via Discover and it didn't exist?

andriisoldatenko:
> I don't really get how this works without the column also being present in the errors table. Wouldn't it throw an error due to the column being missing if if you tried to query the column via Discover and it didn't exist?

we use this test from sentry to check it:
```
def test_transaction_source(self):
        query = {
            "field": ["transaction"],
            "query": "transaction.source:task",
            "project": [self.project.id],
        }
        response = self.do_request(query)
        assert response.status_code == 200, response.content
```

@lynnagara  can you please suggest how to check your case?

lynnagara:
Oh ignore my comment, I see this is being mapped to None in errors

lynnagara (approved):

PR #3021, opened 2022-08-03
feat(replays): Add urls, user_agent, and replay_start_timestamp migration

cmanallen:
Companion of: https://github.com/getsentry/snuba/pull/3023

- Adds migration.

PR #3023, opened 2022-08-04
feat(replays): Add urls, user_agent, and replay_start_timestamp fields and processor

cmanallen:
Depends on: https://github.com/getsentry/snuba/pull/3021

- Add `urls`, `user_agent`, and `replay_start_timestamp` columns.
- Removes `url` column.
- Removes old url processor.


evanh (approved):
Pending the migration, this looks good to me.

PR #3024, opened 2022-08-04
feat(MDC): Generic Metrics storages built from config

rahul-kumar-saini:
### Overview
- We want to generate the Generic Metrics Dataset entirely from config files
- A big part of that are the storages - Sets (read/write) + Distributions (read/write)

### Changes
  - Added 4 config files (1 for each storage)
  - Updated the JSON Schema for storage configs to support `KafkaStreamLoader`
  - Added 2 generalized storage builder functions (one for readonly one for writable)
    - These functions create a `Storage` object given a `StorageKey` by reading the corresponding config file
  - Replaced all usages of the existing storage objects with the new objects built from config
  - Attempted to test the new storage objects by "deep" comparing the to the existing storage objects
    - The tests and existing storage definitions should be removed once we are confident in the new version
 
### Notes
- The storage building logic is non exhaustive and is minimally implemented just to support the generic metric storages
  - This logic along with the JSON schemas should be incrementally improved as we continue to change the storage definitions over to config
- Dynamically loading storage configs is not done in this PR (config file locations are currently hardcoded)


snuba/datasets/configurations/config_loader.py
```diff
@@ -0,0 +1,207 @@
...
+)
+from snuba.datasets.message_filters import KafkaHeaderSelectFilter
+from snuba.datasets.storage import WritableTableStorage
+from snuba.utils.streams.configuration_builder import build_kafka_producer_configuration
+from snuba.utils.streams.topics import Topic
+
+
+def dataclass_from_dict(cls: Type[Any], d: Any) -> Any:
+    # recursively type cast dict to classes
+    # HACK: super hacks
```
evanh, line 42: I know this is a hack so I am not criticizing, just leaving thoughts on how this should work.

This parsing function is assuming that the dataclasses can be arbitrarily combined, which is not true. The structure of our config files will be quite strict, meaning we don't need generic recursive parsers, we can parse each object knowing what it should be. For example we know that a `schema` has a `columns` key and the value is a sequence of `Column` objects. We can parse that directly. If it fails, then the config is bad anyways.

snuba/datasets/storages/generic_metrics.py
```diff
@@ -160,7 +171,7 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
     query_processors=shared_query_processors,
 )
 
-distributions_bucket_storage = WritableTableStorage(
+distributions_bucket_storage_old = WritableTableStorage(
```
evanh, line 174: I'd like to create a completely separate storage, like in a separate file separate. Basically we should avoid editing any existing files (except maybe the factory files).

snuba/datasets/storages/generic_metrics.py
```diff
@@ -181,3 +192,49 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
...
+file = open(
+    "./snuba/datasets/configuration/generic_metrics/storage_distributions_raw.yaml"
+)
+conf_yml = safe_load(file)
+assert isinstance(conf_yml, dict)
+conf = dataclass_from_dict(StorageConfig, conf_yml)
+assert isinstance(conf, StorageConfig)
+
+
+distributions_bucket_storage = WritableTableStorage(
```
fpacifici, line 206: I see that you create an intermediate dataclass representation for your config and then, you create the data model object from that.
What is the reason for these two to be separate ? IT seems the config object do not do anything and do not add any value.

If they provided at least config validation, they would be more useful.
On the other hand having a data model class and config class for each data model object seems very verbose (every time we want to add one). If rely on functions to build the data model from YAML directly you can skip a lot of these steps, but you would need to figure out how to do validation.

snuba/datasets/storages/generic_metrics.py
```diff
@@ -181,3 +192,49 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
...
 )
+
+
+file = open(
+    "./snuba/datasets/configuration/generic_metrics/storage_distributions_raw.yaml"
+)
+conf_yml = safe_load(file)
+assert isinstance(conf_yml, dict)
+conf = dataclass_from_dict(StorageConfig, conf_yml)
+assert isinstance(conf, StorageConfig)
```
fpacifici, line 203: What is the process to instantiate a storage from config ?
Do we expect the storages to be still created statically at module initialization ? And is the factory going to still have a role ?

snuba/datasets/storages/generic_metrics.py
rahul-kumar-saini: The only value the intermediate representation was adding was having everything be typed instead of doing something like `conf["schema"]["columns"]` but it got complicated and I just got kinda carried away. I'm going to use JSONSchema to validate and then just use the dictionaries untyped as is.

evanh: The other advantage of the intermediate interpretation is for lazy loading. We can load the config files on import into the intermediate classes, and only build the actual object when it's requested. That isn't something I'm sure we need though.

snuba/datasets/storages/generic_metrics.py
evanh: What do you mean by "process"?

I do expect to have Storages built on module initialization, since that is how they are currently initialized. We are trying to avoid changing any existing logic we don't have too. The factory will not have a role in configuration storages, but the `get_storage` function will still be used.

snuba/datasets/storages/generic_metrics.py
fpacifici: You can automatically create TypedDict from json schema. No need to create a class 
https://pypi.org/project/jsonschema-typed-v2/

We tried to use that to generate a TypedDict representation of the Events automatically from the jsonschema produced by relay 
https://github.com/getsentry/sentry-data-schemas/tree/main/py

snuba/datasets/configuration/template.yaml
```diff
@@ -0,0 +1,41 @@
+version: v1
```
rahul-kumar-saini, line 1: I've added this `template.yaml` file for reference since there isn't a good way to know what a storage config can contain (aside from looking at existing config files), can be removed if deemed unnecessary.

snuba/datasets/storages/factory.py
```diff
@@ -56,6 +48,20 @@
 from snuba.datasets.storages.transactions_ro import storage as transactions_ro_storage
 from snuba.datasets.storages.transactions_v2 import storage as transactions_v2_storage
 
+gen_metrics_dists_bucket_storage = build_writable_storage(
+    StorageKey.GENERIC_METRICS_DISTRIBUTIONS_RAW
+)
+gen_metrics_dists_aggregate_storage = build_readonly_storage(
+    StorageKey.GENERIC_METRICS_DISTRIBUTIONS
+)
+generic_metrics_sets_bucket_storage = build_writable_storage(
```
evanh: Weird that this one is `generic_` instead of `gen_`.

evanh (commented):
Please make the cutover to the configuration storages based on a setting so we can easily cut back if things go sideways.

Also can you please make note of anything that you _didn't_ get done in this PR? One thing I mentioned is that we ultimately want to load things out of the configuration directory automatically, without needing to put the names of specific files into the Snuba codebase.

snuba/datasets/storages/factory.py
rahul-kumar-saini: True

snuba/datasets/configuration/generic_metrics/storages/sets.yaml
```diff
@@ -0,0 +1,53 @@
+version: v1
+kind: readonly_storage
```
onewland, line 2: I think we should call this a readable storage (not `readonly`). For generic metrics, the writable/readable are separate but for many entities/datasets these things aren't separate

snuba/datasets/configuration/storage_builder.py
```diff
@@ -0,0 +1,146 @@
...
+}
+
+
+def build_readonly_storage(storage_key: StorageKey) -> ReadableTableStorage:
+    config = __load_storage_config(storage_key)
+    storage_kwargs = __build_readonly_storage_kwargs(config)
+    return ReadableTableStorage(**storage_kwargs)
+
+
+def build_writable_storage(storage_key: StorageKey) -> WritableTableStorage:
```
onewland, line 64: This doesn't seem like it should be based on `StorageKey` to me. Why not just take a `str` here? (same with `build_readonly_storage`). And `__load_storage_config` definitely seems like it should take a filename, not a storage key

snuba/datasets/storages/factory.py
```diff
@@ -55,6 +59,27 @@
...
+gen_metrics_dists_bucket_storage = gen_metrics_dists_bucket_storage_original
+gen_metrics_dists_aggregate_storage = gen_metrics_dists_aggregate_storage_original
+gen_metrics_sets_bucket_storage = gen_metrics_sets_bucket_storage_original
+gen_metrics_sets_aggregate_storage = gen_metrics_sets_aggregate_storage_original
+
+if get_config("use_generic_metrics_storages_from_configs", 0):
+    gen_metrics_dists_bucket_storage = build_writable_storage(
+        StorageKey.GENERIC_METRICS_DISTRIBUTIONS_RAW
+    )
+    gen_metrics_dists_aggregate_storage = build_readonly_storage(
```
onewland, line 73: given that we have the `kind` value in the config, why don't we use that to determine what kind of storage we're building, and then everything can be loaded from one generic entrypoint loader?

onewland:
Maybe it's too much to add to this PR but I'd like to see something more like:

```python
def load_dataset_object(str: path) -> Union[Entity, Storage, Dataset]:
   validate_against_all_schemas()
   if(yaml_object["kind"] == "readable_storage"):
      return __readable_storage_from_yaml(yaml_object)
   ...
```

and then `load_dataset_object` is the only thing that's exposed outside the module (in fact, I'd expect we eventually hide even that and do some kind of auto-import on startup).

evanh (commented):
Please add an E2E api test for this as well. Be careful, I don't know if overriding the setting for the test will be enough, you need the storages to actually be rebuilt.

snuba/datasets/storages/factory.py
rahul-kumar-saini: I had that before and there were messy typing issues when returning `ReadableTableStorage | WritableTableStorage` so I split it into 2 functions. I guess since these will be automatically loaded I do have to go back to doing it that way, I'll change it.

snuba/datasets/configuration/template.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+storage:
+  key: string
+  set_key: string
+
+schema:
+  columns: [{ name: string, type: string, args: object }]
+  local_table_name: string
+  dist_table_name: string
+  mandatory_conditions: [string] # | null
```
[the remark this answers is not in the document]
rahul-kumar-saini: It can be omitted entirely, JSON Schema treats null the same as the key not existing.

snuba/datasets/configuration/template.yaml
volokluev, line 1: read through this as if you didn't build it and answer all the questions you may have

snuba/datasets/configuration/storage_builder.py
```diff
@@ -0,0 +1,144 @@
...
+            dist_table_name=config[SCHEMA]["dist_table_name"],
+            storage_set_key=StorageSetKey(config[STORAGE][SET_KEY]),
+        ),
+        QUERY_PROCESSORS: get_query_processors(
+            config[QUERY_PROCESSORS] if QUERY_PROCESSORS in config else []
+        ),
+    }
+
+
+def __build_stream_loader(loader_config: dict[str, Any]) -> KafkaStreamLoader:
```
volokluev, line 95: There should be a test just for this function

snuba/datasets/configuration/utils.py
```diff
@@ -15,12 +19,54 @@
...
+        def produce_policy_creator() -> DeadLetterQueuePolicy:
+            return ProduceInvalidMessagePolicy(
+                KafkaProducer(build_kafka_producer_configuration(Topic(dlq_topic))),
+                KafkaTopic(dlq_topic),
+            )
+
+        return produce_policy_creator
+    return None
+
+
```
volokluev, line 53: Add a TODO to remove all these dictionaries

snuba/datasets/configuration/utils.py
```diff
@@ -15,12 +19,54 @@
...
+    Creates a DLQ Policy creator function.
+    """
+    if dlq_policy_config["type"] == "produce":
+        dlq_topic = dlq_policy_config["args"][0]
+
+        def produce_policy_creator() -> DeadLetterQueuePolicy:
+            return ProduceInvalidMessagePolicy(
+                KafkaProducer(build_kafka_producer_configuration(Topic(dlq_topic))),
+                KafkaTopic(dlq_topic),
+            )
```
volokluev, line 48: This is incomplete because it only works for produce policies, make a note of that. There should also be a test for this

snuba/datasets/configuration/storage_builder.py
```diff
@@ -0,0 +1,144 @@
...
+    StorageKey.GENERIC_METRICS_SETS: f"{CONFIG_FILES_PATH}/sets.yaml",
+}
+
+STORAGE_VALIDATION_SCHEMAS = {
+    "readable_storage": V1_READABLE_STORAGE_SCHEMA,
+    "writable_storage": V1_WRITABLE_STORAGE_SCHEMA,
+}
+
+
+def build_storage(
```
volokluev, line 58: let's add a test for:

* what happens if you omit a required field
*

snuba/datasets/configuration/storage_builder.py
```diff
@@ -0,0 +1,144 @@
...
+DLQ_POLICY = "dlq_policy"
+
+# TODO: The config files should be discovered automatically
+CONFIG_FILES_PATH = "./snuba/datasets/configuration/generic_metrics/storages"
+CONFIG_FILES = {
+    StorageKey.GENERIC_METRICS_DISTRIBUTIONS: f"{CONFIG_FILES_PATH}/distributions.yaml",
+    StorageKey.GENERIC_METRICS_DISTRIBUTIONS_RAW: f"{CONFIG_FILES_PATH}/distributions_bucket.yaml",
+    StorageKey.GENERIC_METRICS_SETS_RAW: f"{CONFIG_FILES_PATH}/sets_bucket.yaml",
+    StorageKey.GENERIC_METRICS_SETS: f"{CONFIG_FILES_PATH}/sets.yaml",
+}
```
volokluev, line 50: Move these to the storage factory as a temporary place for them. Make this file not depend on StorageKey

snuba/datasets/configuration/template.yaml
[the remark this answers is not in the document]
evanh: Got it, OK.

snuba/datasets/storages/factory.py
```diff
@@ -55,6 +56,25 @@
...
+    assert isinstance(dists_storage_new, WritableTableStorage)
+    gen_metrics_dists_bucket_storage = dists_storage_new
+    gen_metrics_dists_aggregate_storage = build_storage(
+        StorageKey.GENERIC_METRICS_DISTRIBUTIONS
+    )
+    sets_storage_new = build_storage(StorageKey.GENERIC_METRICS_SETS_RAW)
+    assert isinstance(sets_storage_new, WritableTableStorage)
+    gen_metrics_sets_bucket_storage = sets_storage_new
+    gen_metrics_sets_aggregate_storage = build_storage(StorageKey.GENERIC_METRICS_SETS)
+
```
volokluev, line 77: add a test to make sure this works how you think it does

volokluev (commented):
In person review comments

evanh (approved):
Looks good to me, pending the CI passes.

PR #3026, opened 2022-08-05
fix(inc): Introduce denylist that stops attempting consistent queries per project [INC-190]

untitaker:
Adding a project to this list will cause Snuba to ~give up on consistency~ give up on fine-grained consistency
in queries that include that project.

### DIFF
--- a/snuba/datasets/storages/processors/replaced_groups.py
+++ b/snuba/datasets/storages/processors/replaced_groups.py
@@ -56,7 +56,10 @@ class PostReplacementConsistencyEnforcer(QueryProcessor):
         for denied_project_id_string in (
             get_config("post_replacement_consistency_projects_denylist") or "[]"
         )[1:-1].split(","):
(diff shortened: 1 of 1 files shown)


snuba/datasets/storages/processors/replaced_groups.py
```diff
@@ -52,6 +53,13 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
             self._set_query_final(query, False)
             return
 
+        for denied_project_id_string in (
+            get_config("post_replacement_consistency_projects_denylist") or "[]"
+        )[1:-1].split(","):
+            if int(denied_project_id_string) in project_ids:
+                metrics.increment(name=CONSISTENCY_DENYLIST_METRIC)
+                return
```
fpacifici, line 61: Please turn the query final in these cases.

fpacifici (approved):

PR #3027, opened 2022-08-05
fix(replacements): Limit size of excluded groups set [INC-190]

untitaker:
This is an insufficient fix for correctness, it mainly ensures the redis
set does not grow unbounded.

The redis set should be trimmed to the most recent groups. I have **not** thought through the product implications of doing that (we should automatically fallback to FINAL), but it seems better to me than manually denylisting projects over the weekend.

### DIFF
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -215,7 +215,9 @@ def set_project_exclude_groups(
     p.execute()
 
 
-def truncate_group_id_replacement_set(p: StrictClusterPipeline, key: str, now: float) -> None:
+def truncate_group_id_replacement_set(
+    p: StrictClusterPipeline, key: str, now: float
+) -> None:
     # remove group id deletions that should have been merged by now
     p.zremrangebyscore(key, -1, now - settings.REPLACER_KEY_TTL)
 
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -213,7 +213,7 @@ def set_project_exclude_groups(
     p.execute()
 
 
-def truncate_group_id_replacement_set(p, key, now):
+def truncate_group_id_replacement_set(p: Any, key: str, now: float) -> None:
     # remove group id deletions that should have been merged by now
     p.zremrangebyscore(key, -1, now - settings.REPLACER_KEY_TTL)
     # remove group id deletions that exceed the maximum number of deletions
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -218,11 +218,17 @@ def set_project_exclude_groups(
 def truncate_group_id_replacement_set(p: Any, key: str, now: float) -> None:
     # remove group id deletions that should have been merged by now
     p.zremrangebyscore(key, -1, now - settings.REPLACER_KEY_TTL)
+
     # remove group id deletions that exceed the maximum number of deletions
     # snuba's query processor will put in a query.
-    # Add +1 such that the query processor will still recognize that we
-    # exceeded the limit and fall back to FINAL.
-    p.zremrangebyrank(key, 0, -(settings.REPLACER_MAX_GROUP_IDS_TO_EXCLUDE + 1))
+    #
+    # Add +2 because:
+    #
+    # - ranges are exclusive in redis (apparently)
+    # - such that the query processor will recognize that we exceeded the limit
+    #   and fall back to FINAL (because the set doesn't contain all group ids to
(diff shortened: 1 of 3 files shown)


tests/test_replacer.py
```diff
@@ -401,6 +401,29 @@ def test_query_time_flags_groups(self) -> None:
...
+
+        errors_replacer.set_project_exclude_groups(
+            project_id,
+            list(range(10)),
+            ReplacerState.ERRORS,
+            ReplacementType.EXCLUDE_GROUPS,
+        )
+
+        flags = ProjectsQueryFlags.load_from_redis([project_id], ReplacerState.ERRORS)
+        assert len(flags.group_ids_to_exclude) == 5
```
MeredithAnya, line 425: should we assert which group_ids these are?

snuba/datasets/errors_replacer.py
```diff
@@ -192,25 +192,37 @@ def set_project_exclude_groups(
...
 
 
+def truncate_group_id_replacement_set(p: Any, key: str, now: float) -> None:
+    # remove group id deletions that should have been merged by now
+    p.zremrangebyscore(key, -1, now - settings.REPLACER_KEY_TTL)
+    # remove group id deletions that exceed the maximum number of deletions
+    # snuba's query processor will put in a query.
+    # Add +1 such that the query processor will still recognize that we
+    # exceeded the limit and fall back to FINAL.
+    p.zremrangebyrank(key, 0, -(settings.REPLACER_MAX_GROUP_IDS_TO_EXCLUDE + 1))
```
rahul-kumar-saini, line 223: Is my understanding correct?

Previous logic:
Remove all stale replacements. Relevant replacements still remain in Redis - no limit on the total number.

New logic:
Remove all stale replacements. Relevant replacements still remain but are limited to a maximum for this project.

`REPLACER_MAX_GROUP_IDS_TO_EXCLUDE` is the maximum number of group ids we add to a *single* query to exclude before we just set the query to final.

So there isn't any need in Redis to store more than that max (+ 1) since logically it will still be FINAL and the additional storage does not help?

tests/test_replacer.py
untitaker: yup, thanks

snuba/datasets/errors_replacer.py
```diff
@@ -186,31 +187,44 @@ def set_project_exclude_groups(
...
     }
     p.zadd(type_key, replacement_type_data)
-    p.zremrangebyscore(type_key, -1, now - settings.REPLACER_KEY_TTL)
+    truncate_group_id_replacement_set(p, type_key, now)
     p.expire(type_key, int(settings.REPLACER_KEY_TTL))
 
     p.execute()
 
 
+def truncate_group_id_replacement_set(p: Any, key: str, now: float) -> None:
```
rahul-kumar-saini, line 218:
```suggestion
def truncate_group_id_replacement_set(p: StrictClusterPipeline, key: str, now: float) -> None:
```

snuba/datasets/errors_replacer.py
untitaker: https://github.com/getsentry/snuba/blob/0a2f2bb79d9f56174bb495498b8bb363512ec668/snuba/datasets/storages/processors/replaced_groups.py#L100

the idea is that the query processor sees that there are strictly more group ids in redis than the maximum, and falls back to final

however, I just realized that this won't work at all, because the length check there is against the number of groups that would be excluded (based on whether there are group ids in the query). so it might actually misbehave

also I just remembered that filippo pointed out that we can just set the final flag for this project, so I'll do that now

snuba/datasets/errors_replacer.py
untitaker: thanks :+1

tests/test_replacer.py
```diff
@@ -401,6 +403,38 @@ def test_query_time_flags_groups(self) -> None:
             {ReplacementType.EXCLUDE_GROUPS},
         )
 
+    @mock.patch.object(settings, "REPLACER_MAX_GROUP_IDS_TO_EXCLUDE", 5)
+    def test_query_time_flags_bounded_size(self) -> None:
+        redis_client.flushdb()
+        project_id = 4
+        now = time.time()
```
evanh, line 410: We have `freezegun` in the snuba repo you can use to set the time. Then you don't need to manually pass the time into the function (which I would like to avoid).

https://github.com/getsentry/snuba/blob/master/tests/test_optimize_scheduler.py#L249 <- Example

evanh (approved):
See comment about the test, otherwise looks good.

PR #3036, opened 2022-08-09
feat(mdc): Make EntityKey no longer an enum

volokluev:
### Context

Every entity has to be defined as an enum and then added to the factory. We want to stop having that be the case. This PR:

1. Makes the `EntityKey` class just a wrapper around a string
2. Introduces an `EntityKeys` class which allows backwards compatibilty with the `EntityKey.EVENTS` way of specifying `EntityKey`(s)
3. Renames `EntityKey.GROUPEDMESSAGES` to `EntityKeys.GROUPEDMESSAGE` to make the enum key and value consistent

### How to Review
The first commit contains all the non-rename changes

### DIFF
--- a/snuba/cli/entities.py
+++ b/snuba/cli/entities.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 import click
 
-from snuba.datasets.entities import EntityKey, EntityKeys
+from snuba.datasets.entities import EntityKey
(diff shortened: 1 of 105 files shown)


volokluev:
Because the cli needs to iterate over every entity, we're going to need a way to make that happen first

lynnagara:
Closing due to staleness, please reopen if needed

PR #3037, opened 2022-08-09
[prototype] [not for merge] loading entity definitions from YAML

onewland:
```
➜  snuba git:(configurable-entities) ✗ python snuba/datasets/entity_definitions.py \ 
snuba/datasets/configuration/generic_metrics/entity_generic_metrics_distributions_friendlier.yaml          

{'schema': 'v1', 'kind': 'entity', 'name': 'generic-metrics-distributions', 'spec': {'readable_storage': 'distributions_storage', 'writable_storage': 'distributions_bucket_storage', 'columns': [{'name': 'org_id', 'type': 'UInt(64)'}, {'name': 'project_id', 'type': 'UInt(64)'}, {'name': 'metric_id', 'type': 'UInt(64)'}, {'name': 'timestamp', 'type': 'DateTime'}, {'name': 'bucketed_time', 'type': 'DateTime'}, {'name': 'tags', 'type': 'Nested', 'subcolumns': [{'name': 'key', 'type': 'UInt(64)'}, {'name': 'value', 'type': 'UInt(64)'}]}, {'name': 'percentiles', 'type': 'AggregateFunction', 'function': 'quantiles(0.5, 0.75, 0.9, 0.95, 0.99)', 'datatype': 'Float(64)'}], 'query_processors': [{'transform_tag_types': None}, {'handle_mapped_granularities': {'accepted_granularities': [[60, 1], [3600, 2], [86400, 3]], 'default_granularity_enum': 1}}], 'validators': {'require_column': {'columns': ['org_id', 'project_id']}, 'org_id_rate_limit': {'org_column': 'org_id'}}}}
EntityDefinition(query_processors=[<snuba.datasets.entities.metrics.TagsTypeTransformer object at 0x10bb18160>, <snuba.query.processors.granularity_processor.MappedGranularityProcessor object at 0x10bb18040>], columns=ColumnSet([Column('org_id', UInt(64)[None]), Column('project_id', UInt(64)[None]), Column('metric_id', UInt(64)[None]), Column('timestamp', DateTime()[None]), Column('bucketed_time', DateTime()[None]), Column('tags', Nested([Column('key', UInt(64)[None]), Column('value', UInt(64)[None])])[None]), Column('percentiles', AggregateFunction('quantiles(0.5, 0.75, 0.9, 0.95, 0.99)', Float(64)[None])[None])]))
```


snuba/query/processors/granularity_processor.py
```diff
@@ -64,11 +65,28 @@ class MappedGranularityProcessor(QueryProcessor):
 
     def __init__(
         self,
-        accepted_granularities: Sequence[GranularityMapping],
+        accepted_granularities: Union[
+            Sequence[GranularityMapping], Sequence[Sequence[int]]
+        ],
         default_granularity_enum: int,
     ):
+        # handle YAML not knowing about the GranularityMapping NamedTuple
```
onewland, line 73: We may want to consider having all query processors/validators take basic python types (int, str, Sequence[str], Sequence[int], Mapping[str,str] etc.) to avoid this kind of stuff

onewland: A quick survey of the codebase says we're mostly not using custom types in QueryProcessor subclass initializers

snuba/query/processors/__init__.py
```diff
@@ -29,3 +30,23 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
         # existing one in place. We can move towards an immutable structure
         # after changing Request.
         raise NotImplementedError
+
+
+_SEEN_QUERY_PROCESSORS: MutableMapping[str, Type[QueryProcessor]] = {}
+
+# Decorator for registering your query processor in the static list
+def query_processor(
```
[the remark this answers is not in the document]
volokluev: and there you can also enforce a "config_name" property being defined in there as well

snuba/query/processors/granularity_processor.py
volokluev: I'm for this

snuba/datasets/configuration/generic_metrics/entity_generic_metrics_distributions_friendlier.yaml
```diff
@@ -0,0 +1,29 @@
...
+        accepted_granularities:
+          -
+            - 60
+            - 1
+          -
+            - 3600
+            - 2
+          -
+            - 86400
+            - 3
```
volokluev, line 21: this looks a little weird

snuba/query/processors/__init__.py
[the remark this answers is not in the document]
onewland: sure, that's a python thing I didn't know about but I'll read about it

snuba/datasets/configuration/generic_metrics/entity_generic_metrics_distributions_friendlier.yaml
onewland: yeah, it probably makes more sense as a mapping e.g.

```
accepted_granularities:
  60: 1
  3600: 2
  86400: 3
```

snuba/query/processors/granularity_processor.py
evanh: That makes sense to me too.

snuba/query/processors/__init__.py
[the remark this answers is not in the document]
onewland: I think `init_subclass` is the better approach and I'm moving to it

snuba/datasets/configuration/generic_metrics/entity_generic_metrics_distributions_friendlier.yaml
```diff
@@ -0,0 +1,60 @@
+schema: v1
+kind: entity
+name: generic-metrics-distributions
+spec:
+  readable_storage: distributions_storage
+  writable_storage: distributions_bucket_storage
+  columns:
```
fpacifici, line 7: I would suggest you look how the schema for the errors table would look like in this format. That schema is way larger, it may make you consider less verbose alternatives.

Also, a bit of a crazy idea as it would break yaml, but more compact. 
We have a schema parser in the snuba code, that is capable of parsing the output of a describe command in clickhouse 
https://github.com/getsentry/snuba/blob/master/snuba/migrations/parse_schema.py
We use it to compare the schema defined by a migration with the schema from Clickhouse.

So technically we could parse a schema written in the Clickhouse syntax, which is more compact, if you find a reasonable place to store it.

PR #3040, opened 2022-08-10
ref: remove future (unused python2 porting library)

asottile-sentry:


lynnagara:
Hmm, I think there was a reason why I couldn't safely remove this.

volokluev (approved):

asottile-sentry:
> Hmm, I think there was a reason why I couldn't safely remove this. IIRC it was needed to start the server successfully or something. Can you please ensure that the `snuba devserver` command still works (this is not tested in CI)

I don't have an x86_64 machine on hand at the moment and until I can get this to confluent-kafka==1.9.2 snuba doesn't build for me on arm :( -- do you have an intel machine that can run this?  I'd be very surprised if this change broke anything because `python-future` is very dead heh

lynnagara:
@asottile-sentry Just tested it, it seems to be working fine now without future

lynnagara (approved):

asottile-sentry:
thank you for testing it for me!  I really appreciate it -- I'll be able to test these myself soon :D

PR #3044, opened 2022-08-10
RFC: Registered Class

volokluev:
### Context
Snuba needs an easy way to map a string to a class in order to move configuration out of python into yaml and to have the config objects and python objects live side by side while we migrate everything over. This functionality needs to exist for:

- Datasets
- Entities
- Storages
- Query Processors
- Maybe others?

### What this PR does
Defines a metaclass that provides this functionality to any class defining it as a metaclass (see tests for examples). 

Upside: 
* don't need to implement a registry and `__init_subclass__` functionality for every class that needs to be looked up by name.
* Don't need the enum weirdness that we have now in order to map strings to classes

Downside: The `from_name` function is hard to make type safe. If you look in the `TypedFromName` class in the tests, you can see I need to override the method and cast it. I don't think there's another way to do it but maybe someone will show me something I'm missing


snuba/utils/registered_class.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self) -> None:
+        self.__mapping: Dict[str, RegisteredClass] = {}
+
+    def register_class(self, cls: "RegisteredClass") -> None:
+        key = cls.config_key()
+        existing_class = self.__mapping.get(key)
+        if not existing_class:
+            self.__mapping[key] = cls
+
+    def get_class_from_name(self, cls_name: str) -> Optional["RegisteredClass"]:
```
onewland, line 19: It could be handled by the caller but in general I think this should be a hard failure.

onewland (approved):
I think this approach makes sense, and yeah, I've already been having to sprinkle `# type: ignore`s around in various places to get around mypy in my PR, I think there's no way around it

onewland:
I do think others outside our sprint should review, though, as this is going to touch a lot of the codebase eventually.

requirements.txt
```diff
@@ -6,7 +6,6 @@ confluent-kafka==1.8.2
 datadog==0.21.0
 flake8==4.0.1
 Flask==2.1.2
-future==0.18.2
```
evanh, line 9: Why this change?

evanh (commented):
Can you give a concrete example of how this would work, maybe with a QueryProcessor?

snuba/utils/registered_class.py
volokluev: fair enough, I'll make that change

PR #3046, opened 2022-08-11
feat(MDC): Add Storage yaml parsing and schema

rahul-kumar-saini:
### Overview
- Added base JSON Schemas to parse Storage config yamls
- This includes definitions for Column parsing which can be extended upon for other configs

### DIFF
--- a/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
@@ -19,24 +19,25 @@ schema:
         name: tags,
         type: Nested,
         args:
-          [
-            { name: key, type: UInt, args: { size: 64 } },
-            { name: indexed_value, type: UInt, args: { size: 64 } },
-            { name: raw_value, type: String },
-          ],
+          {
+            subcolumns:
+              [
+                { name: key, type: UInt, args: { size: 64 } },
+                { name: indexed_value, type: UInt, args: { size: 64 } },
+                { name: raw_value, type: String },
+              ],
+          },
       },
 
       {
         name: _raw_tags_hash,
         type: Array,
-        args: { type: UInt, arg: 64 },
-        schema_modifiers: [readonly],
+        args: { type: UInt, arg: 64, schema_modifiers: [readonly] },
       },
       {
         name: _indexed_tags_hash,
         type: Array,
-        args: { type: UInt, arg: 64 },
-        schema_modifiers: [readonly],
+        args: { type: UInt, arg: 64, schema_modifiers: [readonly] },
       },
       { name: granularity, type: UInt, args: { size: 8 } },
 
@@ -77,8 +78,7 @@ schema:
       {
         name: histogram_buckets,
         type: AggregateFunction,
-        args:
-          { func: "histogram(250)", arg_types: [{ type: Float, arg: 64 }] },
+        args: { func: "histogram(250)", arg_types: [{ type: Float, arg: 64 }] },
       },
     ]
 
--- a/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
@@ -1,3 +1,7 @@
+version: v1
+kind: readonly_storage
+name: generic-metrics-distributions
+
 storage:
   key: generic_metrics_distributions
   set_key: generic_metrics_distributions
@@ -73,7 +77,8 @@ schema:
       {
         name: histogram_buckets,
         type: AggregateFunction,
-        args: { func: "histogram(250)", arg_types: [{ type: Float, arg: 64 }] },
+        args:
+          { func: "histogram(250)", arg_types: [{ type: Float, arg: 64 }] },
       },
     ]
 
--- a/snuba/datasets/configuration/generic_metrics/storages/distributions_bucket_storage.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/distributions_bucket_storage.yaml
@@ -1,0 +1,53 @@
+storage:
+  key: generic_metrics_distributions_raw
+  set_key: generic_metrics_distributions
+
+schema:
+  columns:
+    [
+      { name: org_id, type: UInt, args: { size: 64 } },
+      { name: use_case_id, type: String },
+      { name: project_id, type: UInt, args: { size: 64 } },
+      { name: metric_id, type: UInt, args: { size: 64 } },
+      { name: timestamp, type: DateTime },
+      { name: retention_days, type: UInt, args: { size: 16 } },
+      {
+        name: tags,
+        type: Nested,
+        args:
+          [
+            { name: key, type: UInt, args: { size: 64 } },
+            { name: indexed_value, type: UInt, args: { size: 64 } },
+            { name: raw_value, type: String },
+          ],
+      },
+
+      { name: granularities, type: Array, args: { type: UInt, arg: 8 } },
+      { name: count_value, type: Float, args: { size: 64 } },
+      { name: set_values, type: Array, args: { type: UInt, arg: 64 } },
+      {
+        name: distribution_values,
+        type: Array,
+        args: { type: Float, arg: 64 },
+      },
+      { name: timeseries_id, type: UInt, args: { size: 32 } },
+    ]
+  local_table_name: generic_metric_distributions_raw_local
+  dist_table_name: generic_metric_distributions_raw_local
+
+stream_loader:
+  processor: generic_distributions_metrics_processor
+  default_topic: snuba-generic-metrics
+  commit_log_topic: snuba-generic-metrics-distributions-commit-log
+  subscription_scheduled_topic: scheduled-subscriptions-generic-metrics-distributions
+  subscription_scheduler_mode: global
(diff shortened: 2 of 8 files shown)


snuba/datasets/configuration/generic_metrics/storages/distributions_bucket_storage.yaml
```diff
@@ -0,0 +1,53 @@
+storage:
```
onewland, line 1: I'd like some kind of header metadata in these YAML files similar to what I did in my PR that lets us do a quick sanity check of what we're trying to parse:

https://github.com/getsentry/snuba/pull/3037/files#diff-384b271ce3ab4bf13b8aa9156f53945508d69e6b59fd612a2e873a0675ecdcafR1-R4

I copied mine from Kubernetes and I don't really care if we follow that format precisely, but I think at the very least we should have schema version and type of file encoded at the top

snuba/datasets/configuration/generic_metrics/storages/distributions_storage.yaml
```diff
@@ -0,0 +1,80 @@
...
+      { name: use_case_id, type: String },
+      { name: project_id, type: UInt, args: { size: 64 } },
+      { name: metric_id, type: UInt, args: { size: 64 } },
+      { name: timestamp, type: DateTime },
+      { name: retention_days, type: UInt, args: { size: 16 } },
+      {
+        name: tags,
+        type: Nested,
+        args:
+          [
```
onewland, line 18: I think all argument lists that correspond to Python argument lists should be named for more straightforward forwards/backwards compatibility so this should be something like:

```yaml
args:
  subcolumns:
    (what you have now)
```

snuba/datasets/configuration/generic_metrics/storages/distributions_storage.yaml
```diff
@@ -0,0 +1,80 @@
...
+        name: _raw_tags_hash,
+        type: Array,
+        args: { type: UInt, arg: 64 },
+        schema_modifiers: [readonly],
+      },
+      {
+        name: _indexed_tags_hash,
+        type: Array,
+        args: { type: UInt, arg: 64 },
+        schema_modifiers: [readonly],
```
onewland, line 35: should schema_modifiers be part of args?

snuba/datasets/configuration/json_schema.py
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+
+Number_schema: Any = make_column_schema(
+    column_type={"enum": ["UInt", "Float"]},
+    args={"type": "object", "properties": {"size": {"type": "number"}}},
+)
+
+
+No_arg_schema = make_column_schema(
```
onewland, line 56: nit: casing here is weird, i'd be fine with `NO_ARG_SCHEMA` or `no_arg_schema` (or maybe `NoArgSchema`?)

tests/datasets/configuration/test_generic_metrics.py
```diff
@@ -1,6 +1,66 @@
...
+        "histogram_buckets",
+        AggregateFunction("histogram(250)", [Float(64)]),
+    ),
+]
+
+
+def test_distributions_storage() -> None:
+    file = open(f"{CONF_PATH}/distributions_storage.yaml")
+    config = safe_load(file)
+    validate(config, readable_storage_schema)
```
onewland, line 42: I think `parse YAML and validate config` should be encapsulated in a single `load_config` method/class somewhere and we should strongly discourage any sideload that doesn't also do enforcement

onewland (commented):
If the files are in a directory named `storages`, I think they don't need the suffix `_storage.yaml`

Some comments on the approach but overall I think what you're doing makes sense

snuba/datasets/configuration/json_schema.py
evanh: I agree, constants should be capital case.

snuba/datasets/configuration/json_schema.py
```diff
@@ -0,0 +1,141 @@
...
+                "type": TYPE_STRING,
+                "args": {"type": "array", "items": TYPE_STRING},
+            },
+        },
+    },
+}
+
+######
+# Column specific json schemas
+def make_column_schema(column_type: Any, args: Any) -> Any:
```
evanh, line 36: Why is this untyped?

tests/datasets/configuration/test_generic_metrics.py
```diff
@@ -1,6 +1,66 @@
...
 
-def test_file_loads() -> None:
-    file = open("./snuba/datasets/configuration/generic_metrics/dataset.yaml")
-    safe_load(file)
+CONF_PATH = "./snuba/datasets/configuration/generic_metrics/storages"
+
+DISTRIBUTIONS_STORAGE_COLUMN_SET = [
+    *common_columns,
+    *aggregate_common_columns,
+    Column(
```
evanh, line 23: Can we also import the aggregate functions instead of manually writing them out for the test?

tests/datasets/configuration/test_generic_metrics.py
```diff
@@ -1,6 +1,66 @@
...
+    ),
+]
+
+
+def test_distributions_storage() -> None:
+    file = open(f"{CONF_PATH}/distributions_storage.yaml")
+    config = safe_load(file)
+    validate(config, readable_storage_schema)
+    assert (
+        parse_columns(config["schema"]["columns"]) == DISTRIBUTIONS_STORAGE_COLUMN_SET
```
evanh, line 44: If you are going to start this PR with stream loaders etc., you need to validate the entire storage not just columns.

snuba/datasets/configuration/generic_metrics/storages/distributions_storage.yaml
rahul-kumar-saini: Good catch, this is what I intended to do

tests/datasets/configuration/test_generic_metrics.py
rahul-kumar-saini: Done, I've created a function that loads a config given a `StorageKey`.

tests/datasets/configuration/test_generic_metrics.py
rahul-kumar-saini: Eventually, would we get rid of our manual definitions of those columns since they'd be generated from config? In that case, we eventually remove the storage specific tests in favor of some generic testers for the JSON Schema?

snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
```diff
@@ -0,0 +1,85 @@
...
+        args: { func: "count", arg_types: [{ type: Float, arg: 64 }] },
+      },
+      {
+        name: histogram_buckets,
+        type: AggregateFunction,
+        args: { func: "histogram(250)", arg_types: [{ type: Float, arg: 64 }] },
+      },
+    ]
+
+query_processors: ["TableRateLimit", "TupleUnaliaser"]
```
fpacifici, line 85: Do we plan to have a list of query processors that are added by default? I think there are a few that do not need to be specified. 
Example `TableRateLimit`. A rate limiter per table should always be present. There should be no need to ever remove it.

tests/datasets/configuration/test_generic_metrics.py
rahul-kumar-saini: I've added a test for the stream loader, not sure if this is what you meant. Would we actually keep tests like these around? Seems kind of like rewriting the config but in test cases.

snuba/datasets/configuration/json_schema.py
```diff
@@ -0,0 +1,155 @@
...
+        "schema": SCHEMA_SCHEMA,
+        "query_processors": QUERY_PROCESSORS_SCHEMA,
+        "stream_loader": STREAM_LOADER_SCHEMA,
+    },
+}
+
+READABLE_STORAGE_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "version": TYPE_STRING,
```
onewland, line 148: I think we should verify that the version is 1 (maybe rename this and the other to `V1_{READABLE,WRITEABLE}_STORAGE_SCHEMA`)

onewland (approved):
Looks good to me but let's fast-fail if the version isn't v1

snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
rahul-kumar-saini: I've been 1-1 mapping existing logic to config so far, we can think about or introduce this always-add logic once we get the config stuff working (I'll create a tech debt task so it's not forgotten)

tests/datasets/configuration/test_generic_metrics.py
evanh: Sorry for the confusion, I think there are a few things going on here.

1) You need to write tests for all the functionality you're adding. You have tests for column parsing, but not for query processor parsing or DLQ parsing. 
2) If there is functionality in this PR that is not ready to be tested yet (like the stream loader config) then ideally you would remove it. I'm not really sure why that's in the PR, you aren't actually parsing it.

evanh (approved):
Looks good, first MDC commit!

PR #3048, opened 2022-08-11
SNS-1570-rely-on-the-entity-enum-in-the-subscriptions-scheduler-cli-command

dbanda:
Uses the entity enum for the list of options in entity enum name CLI choice option. Addressing this https://getsentry.atlassian.net/browse/SNS-1570

### DIFF
--- a/snuba/cli/subscriptions_scheduler.py
+++ b/snuba/cli/subscriptions_scheduler.py
@@ -24,16 +24,7 @@ logger = logging.getLogger(__name__)
     "--entity",
     "entity_name",
     required=True,
-    type=click.Choice(
-        [
-            "events",
-            "transactions",
-            "metrics_sets",
-            "metrics_counters",
-            "generic_metrics_sets",
-            "generic_metrics_distributions",
-        ]
-    ),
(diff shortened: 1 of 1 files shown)


fpacifici:
Thanks. this is useful. 
We do not have a very good test coverage for the CLI scripts (the logic is covered, the script itself is not).
Have you try to run the scheduler locally to ensure it works ?
The scheduler would start either by relying on `snuba devserver` or by manually running it. 
If you try to manually run it, the devserver script gives you an idea of the parameters to pass: https://github.com/getsentry/snuba/blob/master/snuba/cli/devserver.py#L113-L127

dbanda:
> Thanks. this is useful. We do not have a very good test coverage for the CLI scripts (the logic is covered, the script itself is not). Have you try to run the scheduler locally to ensure it works ? The scheduler would start either by relying on `snuba devserver` or by manually running it. If you try to manually run it, the devserver script gives you an idea of the parameters to pass: https://github.com/getsentry/snuba/blob/master/snuba/cli/devserver.py#L113-L127

I run the server a couple of time and everything works fine. I'll look into how I can improve the coverage within the tests.

fpacifici (approved):
good thanks.

PR #3049, opened 2022-08-11
SNS-1572-rely-on-the-entity-enum-in-the-subscriptions-executor-cli-command

dbanda:
Uses the entity enum in the list of options for the the entity name CLI choice option. Addresses this Jira https://getsentry.atlassian.net/browse/SNS-1572

### DIFF
--- a/snuba/cli/subscriptions_executor.py
+++ b/snuba/cli/subscriptions_executor.py
@@ -31,16 +31,7 @@ from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
     "entity_names",
     required=True,
     multiple=True,
-    type=click.Choice(
-        [
-            EntityKey.EVENTS.value,
-            EntityKey.TRANSACTIONS.value,
-            EntityKey.METRICS_COUNTERS.value,
-            EntityKey.METRICS_SETS.value,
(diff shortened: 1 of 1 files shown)


snuba/cli/subscriptions_executor.py
```diff
@@ -31,16 +31,7 @@
...
-        [
-            EntityKey.EVENTS.value,
-            EntityKey.TRANSACTIONS.value,
-            EntityKey.METRICS_COUNTERS.value,
-            EntityKey.METRICS_SETS.value,
-            EntityKey.GENERIC_METRICS_SETS.value,
-            EntityKey.GENERIC_METRICS_DISTRIBUTIONS.value,
-        ]
-    ),
+    type=click.Choice([entity_key.value for entity_key in EntityKey]),
```
lynnagara, line 34: Agree that enumerating all of the entity keys is verbose, but there was a reason for that. It's because not all entities actually support subscriptions. Now we are allowing entities to be passed that we cannot actually have subscriptions on. I think I'd prefer to leave it as is.

dbanda: oh got it. will abandon this PR. thanks

dbanda:
Abandoning because not all entities are supported.

PR #3050, opened 2022-08-11
config(datasets) - create a PluggableEntity class

onewland:
The goal of this new class is to provide a less flexible version of Entity that will not be subclassed but will be instantiated based on contents of YAML config files.

Once this and https://github.com/getsentry/snuba/pull/3046 are merged, we can make some changes to dynamically populate entities from YAML that fit certain characteristics (single storage plan, a limited set of query processors/mappers/etc.)

This should look somewhat similar to https://github.com/getsentry/snuba/pull/3037/files when done though that's obviously not complete

### DIFF
--- a/tests/datasets/entities/test_pluggable_entity.py
+++ b/tests/datasets/entities/test_pluggable_entity.py
@@ -1,0 +1,163 @@
+from datetime import datetime, timedelta
+from functools import partial
+from typing import MutableMapping
+
+import pytest
+import pytz
+
+from snuba.attribution import get_app_id
+from snuba.attribution.attribution_info import AttributionInfo
+from snuba.clickhouse.translators.snuba.mappers import (
+    FunctionNameMapper,
+    SubscriptableMapper,
+)
+from snuba.clickhouse.translators.snuba.mapping import TranslationMappers
+from snuba.datasets.entities.generic_metrics import GenericMetricsSetsEntity
+from snuba.datasets.entities.metrics import TagsTypeTransformer
+from snuba.datasets.factory import get_dataset
+from snuba.datasets.pluggable_entity import PluggableEntity
+from snuba.datasets.storages.generic_metrics import sets_storage
+from snuba.query import Query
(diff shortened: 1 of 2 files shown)


volokluev:
question about this: the `get_entity` function returns an `Entity` class. How would these live side by side in the same process?

onewland:
> question about this: the `get_entity` function returns an `Entity` class. How would these live side by side in the same process?

this returns an Entity instance, right?

```python
def get_entity(name: EntityKey) -> Entity:
    if name in ENTITY_IMPL:
        return ENTITY_IMPL[name]
```

volokluev:
> something else would load up a separate map with YAML-loaded entities

yes so is that signature now going to be:

```python
def get_entity(name: EntityKey | str) -> Entity | PluggableEntity:
    pass
```

onewland:
> > something else would load up a separate map with YAML-loaded entities
> 
> yes so is that signature now going to be:
> 
> ```python
> def get_entity(name: EntityKey | str) -> Entity | PluggableEntity:
>     pass
> ```

I was assuming we would keep that class return type as Entity (PluggableEntity is an Entity subclass)

volokluev:
oh woops I missed that subclass. All good

tests/datasets/entities/test_pluggable_entity.py
```diff
@@ -0,0 +1,166 @@
...
+        original_body=query_body,
+        query=query,
+        snql_anonymized=snql_anonymized,
+        query_settings=HTTPQuerySettings(referrer=""),
+        attribution_info=AttributionInfo(get_app_id("blah"), "blah", None, None, None),
+    )
+    return request
+
+
+def test_generic_metrics_sets_vs_pluggable_similar_behavior(
```
volokluev, line 124:
```suggestion
def test_generic_metrics_sets_vs_pluggable_similar_pipeline_behavior(
```

volokluev (approved):

PR #3051, opened 2022-08-11
ref(MDC): cleanup dataset factory, formalize factory pattern

volokluev:
YAML datasets and python datasets will need to coexist for some time in production. 
The factory functions are the gateway to accessing datasets.

In order to make the transition easier, a few things were refactored in the dataset factory:

1. One container for all global state in the factory (down from four)
2. Only one list of dataset names maintained
3. remove complicated mapping/caching logic
4. Remove unused `DISABLED_DATASETS` feature

This PR also introduces the ConfigComponentFactory class that I would like to use for all other ConfigComponents (Datasets, Entities, Storages, QueryProcessors) to make the loading and merging of ConfigComponents have a uniform, predictable interface.

### DIFF
--- a/snuba/datasets/factory.py
+++ b/snuba/datasets/factory.py
@@ -7,11 +7,11 @@ from snuba.utils.serializable_exception import SerializableException
 
 
 class _DatasetFactory(ConfigComponentFactory[Dataset, str]):
-    def __init__(self):
+    def __init__(self) -> None:
         self._dataset_map: MutableMapping[str, Dataset] = {}
         self._name_map: MutableMapping[Type[Dataset], str] = {}
 
-    def initialize(self):
+    def initialize(self) -> None:
         from snuba.datasets.cdc.groupassignee import GroupAssigneeDataset
         from snuba.datasets.cdc.groupedmessage import GroupedMessageDataset
         from snuba.datasets.discover import DiscoverDataset
@@ -75,7 +75,7 @@ class InvalidDatasetError(SerializableException):
 _DS_FACTORY = None
 
 
-def _ds_factory():
+def _ds_factory() -> _DatasetFactory:
     global _DS_FACTORY
     if _DS_FACTORY is None:
         _DS_FACTORY = _DatasetFactory()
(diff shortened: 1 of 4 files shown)


snuba/datasets/factory.py
```diff
@@ -1,93 +1,96 @@
...
+                "outcomes": OutcomesDataset(),
+                "outcomes_raw": OutcomesRawDataset(),
+                "sessions": SessionsDataset(),
+                "transactions": TransactionsDataset(),
+                "profiles": ProfilesDataset(),
+                "functions": FunctionsDataset(),
+                "generic_metrics": GenericMetricsDataset(),
+                "replays": ReplaysDataset(),
+            }
+        )
```
enochtangg, line 45: I'm excited to the dataset factory in a cleaner state!

One question, is there a reason why the datasets mappings were changed to an object of Dataset as opposed to the class itself and returning a fresh instance in `get()`? Would this become problematic if we ever decide to call `select_entity` on one dataset, and later in the execution, use `get()` for a new query but same dataset? Not sure if that case exists.

snuba/datasets/factory.py
```diff
@@ -1,93 +1,96 @@
...
+        _DS_FACTORY.initialize()
+    return _DS_FACTORY
+
+
 @with_span()
 def get_dataset(name: str) -> Dataset:
-    if name in DATASETS_IMPL:
-        return DATASETS_IMPL[name]
-
-    if name in settings.DISABLED_DATASETS:
```
evanh, line 40: Disabled datasets is not an unused feature, we are actively using it. Remember that any setting can be overriden by the prod settings, or could be used by self hosted instances.

evanh (changes requested):

snuba/datasets/factory.py
volokluev: where is it being used?

volokluev: either way I brought it back and put in a test for it so we know it's being used and it works

snuba/datasets/factory.py
volokluev: We discussed offline but TL;DR:

The old dataset code actually did use a map of strings to instances, just in a needlessly complicated way. That is now fixed

evanh (approved):
Thanks for bringing that back.

PR #3059, opened 2022-08-15
ref: Dependabot updates

evanh:
Collapse a number of dependabot changes into one PR.


volokluev:
The test failure seems like an actual failure, but I'm not sure which upgrade would have done this

evanh:
> The test failure seems like an actual failure, but I'm not sure which upgrade would have done this

I think it was a flake, the tests all passed for me when I re-ran them. I'm also not sure what could have caused the initial failure.

untitaker (approved):

PR #3060, opened 2022-08-15
ref(MDC): cleanup dataset factory, formalize factory pattern [updated]

enochtangg:
See the original reverted PR here: https://github.com/getsentry/snuba/commit/46566686b32be90f1d95ce1eed633e70cdbcf5ab

@volokluev initial hypothesis of multiple threads accessing incomplete states of the dataset factory map was tested and verified. This PR includes an additional change for moving the initialization of the dataset mapping to the constructor of _DatasetFactory. This ensures the dataset factory object either contains the full map or is None. Test was added accordingly

### DIFF
--- a/snuba/datasets/factory.py
+++ b/snuba/datasets/factory.py
@@ -1,93 +1,106 @@
-from typing import Callable, MutableMapping, Sequence, Set
+from typing import Generator, MutableMapping, Sequence, Type
 
 from snuba import settings
 from snuba.datasets.dataset import Dataset
 from snuba.util import with_span
+from snuba.utils.config_component_factory import ConfigComponentFactory
 from snuba.utils.serializable_exception import SerializableException
 
-DATASETS_IMPL: MutableMapping[str, Dataset] = {}
-DATASETS_NAME_LOOKUP: MutableMapping[Dataset, str] = {}
 
-DEV_DATASET_NAMES: Set[str] = set()
+class _DatasetFactory(ConfigComponentFactory[Dataset, str]):
+    def __init__(self) -> None:
+        self._dataset_map: MutableMapping[str, Dataset] = {}
+        self._name_map: MutableMapping[Type[Dataset], str] = {}
+        self.__initialize()
 
-DATASET_NAMES: Set[str] = {
-    "discover",
-    "events",
-    "groupassignee",
-    "groupedmessage",
-    "metrics",
-    "outcomes",
-    "outcomes_raw",
-    "sessions",
-    "transactions",
-    "profiles",
-    "functions",
-    "generic_metrics",
-    "replays",
-    *(DEV_DATASET_NAMES if settings.ENABLE_DEV_FEATURES else set()),
-}
+    def __initialize(self) -> None:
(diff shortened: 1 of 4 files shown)


volokluev (approved):

snuba/datasets/factory.py
```diff
@@ -1,93 +1,105 @@
...
 
 
+_DS_FACTORY = None
+
+
+def _ds_factory() -> _DatasetFactory:
+    global _DS_FACTORY
+    if _DS_FACTORY is None:
+        _DS_FACTORY = _DatasetFactory()
+    return _DS_FACTORY
```
volokluev, line 92: I would maybe put a note here that this is accessed by many threads at once but they're supposed to create the same object every time so it shouldn't be a problem if two threads recreate the same object

snuba/datasets/factory.py
```diff
@@ -1,93 +1,105 @@
...
-    *(DEV_DATASET_NAMES if settings.ENABLE_DEV_FEATURES else set()),
-}
+
+class _DatasetFactory(ConfigComponentFactory[Dataset, str]):
+    def __init__(self) -> None:
+        self._dataset_map: MutableMapping[str, Dataset] = {}
+        self._name_map: MutableMapping[Type[Dataset], str] = {}
+        self.initialize()
+
+    def initialize(self) -> None:
```
volokluev, line 16:
```suggestion
    def _initialize(self) -> None:
```
I would say this does not need to be a public method at this time, it can probably be removed from the ConfigComponentFactory as well or made private there.

volokluev (commented):
Two small comments

tests/datasets/test_dataset_factory.py
```diff
@@ -0,0 +1,85 @@
...
+        "functions",
+        "generic_metrics",
+        "replays",
+    ]:
+        factory_ds = get_dataset(ds_name)
+        assert isinstance(factory_ds, Dataset)
+        assert get_dataset_name(factory_ds) == ds_name
+
+
+def test_get_dataset_multithreaded_collision() -> None:
```
evanh, line 37: What happens if this test fails? Does the exception automatically get bubbled up?

enochtangg: Good point, the exception was not automatically bubbled up to the main thread, only existed in child thread's context, and only writing error to stdout. Added some error handling to fix this and fail the test.

snuba/utils/config_component_factory.py
```diff
@@ -0,0 +1,12 @@
+from typing import Generator, Generic, TypeVar
+
+T = TypeVar("T")
+KeyType = TypeVar("KeyType")
+
+
+class ConfigComponentFactory(Generic[T, KeyType]):
```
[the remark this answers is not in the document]
volokluev: It's a pattern we're going to replicate for entities, storages, and query processors.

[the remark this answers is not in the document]
volokluev: I figure since we're going to have this be repeated at least 3 times, it's worth abstracting

PR #3061, opened 2022-08-15
test(settings): Ensure Topics validation and Topics enum are in sync (SNS-1573)

ayirr7:
Add a test that allows to catch the error where a new Topic definition for Kafka is added but is not added to settings validation. Make sure the Topics enum and Topics validation are in sync. JIRA: https://getsentry.atlassian.net/browse/SNS-1573


tests/settings/test_settings.py
```diff
@@ -18,3 +18,28 @@ def test_invalid_storage() -> None:
...
+    topic_map = all_settings["KAFKA_TOPIC_MAP"]
+    default_map = deepcopy(topic_map)
+    # Remove default settings and build dummy Topic Map
+    topic_map.clear()
+    topic_map = {t.value: {} for t in Topic}
+
+    try:
+        validate_settings(all_settings)
+    except Exception as exc:
+        assert False, f"'validate_settings' raised an exception {exc}"
```
volokluev, line 43: Is this `assert False` necessary?

tests/settings/test_settings.py
```diff
@@ -18,3 +18,28 @@ def test_invalid_storage() -> None:
         assert False, f"'validate_settings' raised an exception {exc}"
     finally:
         cluster[0]["storage_sets"].remove("non_existing_storage")
+
+
+def test_topics_sync() -> None:
```
volokluev, line 23: You should probably explain what exactly this test is testing because I don't have a clear picture just glancing at the code

tests/settings/test_settings.py
```diff
@@ -18,3 +18,28 @@ def test_invalid_storage() -> None:
...
+    from copy import deepcopy
+
+    from snuba.settings.validation import validate_settings
+    from snuba.utils.streams.topics import Topic
+
+    all_settings = {
+        key: value
+        for key, value in settings.__dict__.items()
+        if not key.startswith("__") and not callable(key)
+    }
```
fpacifici, line 33: All tests on the settings module are likely to have this code. (See the test above). Please put it in a function so that it will be easier to focus on the code you are testing without being side tracked by the boilerplate.

tests/settings/test_settings.py
```diff
@@ -18,3 +18,28 @@ def test_invalid_storage() -> None:
...
+    topic_map = all_settings["KAFKA_TOPIC_MAP"]
+    default_map = deepcopy(topic_map)
```
fpacifici, line 35: I think you may not need this deepcopy as you are not updating the values of this mapping.

tests/settings/test_settings.py
fpacifici: Could you please provide a more specific error message (this is what people will see when the test fails). The message should tell us that the test failed because the list of Topics defined does not correspond to the list in the validator.

fpacifici (commented):
thanks

tests/settings/test_settings.py
onewland: I would just not bother catching the exception, unless it's more specific than `Exception`, since the error will make the test fail

tests/settings/test_settings.py
onewland: or maybe it could be titled something like `test_all_topics_in_settings_validator()`?

onewland:
Can you just describe how you tested this? (no need for a code change)

tests/settings/test_settings.py
```diff
@@ -18,3 +18,28 @@ def test_invalid_storage() -> None:
         assert False, f"'validate_settings' raised an exception {exc}"
     finally:
         cluster[0]["storage_sets"].remove("non_existing_storage")
+
+
+def test_topics_sync() -> None:
+    from copy import deepcopy
+
+    from snuba.settings.validation import validate_settings
+    from snuba.utils.streams.topics import Topic
```
evanh, line 27: Is there a reason these imports are inside the test itself?

tests/settings/test_settings.py
ayirr7: I realized that there was a mistake with this test that I wrote -- it does not accurately update KAFKA_TOPIC_MAP as I wanted. 

all_settings[KAFKA_TOPIC_MAP] should change so that it imports and includes all of the Topics enum values, and then gets validated by validate_settings(). After validate_settings() it should go back to its "default" state, which is why a deepcopy is needed if I understand correctly.

tests/settings/test_settings.py
ayirr7: I think it would be good to modify the error handling here to be specific to this test case, which is to check whether the two Topics lists correspond to one another. Changing it now.

tests/settings/test_settings.py
ayirr7: Good point, I'll add some comments and rename test.

tests/settings/test_settings.py
ayirr7: Also, I think the `assert False` is necessary to actually show that the test has failed when running pytest

tests/settings/test_settings.py
ayirr7: I was following the style of the above test. I think the idea is to place relevant imports where they belong, so the imports that are at the top of the file are relevant to _every_ test.

tests/settings/test_settings.py
volokluev: In that case, define a validation exception and raise that

ayirr7: Sounds good. In that case should the above previous test also be modified in that way?

tests/settings/test_settings.py
fpacifici: You should not need that. 
The settings module cannot import anything else as it must be possible to import it from anywhere. 
Though this should not be a problem in the test. 

I would try to move all the import at the beginning of the file. This is for both tests.

ayirr7:
Some description on the test:

So this basically temporarily overwrites the default `KAFKA_TOPIC_MAP` in https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L127, and includes all of the defined Kafka topics (https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py#L21). This test now uses `validate_settings` (https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py) to validate these new settings. `validate_settings` checks whether all of the defined Kafka topic names are also included in the topic name list in `validation.py`.

tests/settings/test_settings.py
onewland: I don't actually agree that `assert False` is necessary to show a failure in pytest:

```
➜  snuba git:(load-entity-from-yaml) ✗ cat tests/test_failure.py 
def test_failure():
	raise NotImplementedError() 

def test_passes():
	assert 1 + 1 == 2
➜  snuba git:(load-entity-from-yaml) ✗ pytest -vv tests/test_failure.py
=================================================================================================================================================================== test session starts ===================================================================================================================================================================
platform darwin -- Python 3.8.12, pytest-7.1.2, pluggy-0.13.0 -- /Users/oliver/workplace/snuba/.venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/oliver/workplace/snuba, configfile: setup.cfg
plugins: cov-3.0.0
collected 2 items                                                                                                                                                                                                                                                                                                                                         

tests/test_failure.py::test_failure FAILED                                                                                                                                                                                                                                                                                                          [ 50%]
tests/test_failure.py::test_passes PASSED                                                                                                                                                                                                                                                                                                           [100%]

======================================================================================================================================================================== FAILURES =========================================================================================================================================================================
______________________________________________________________________________________________________________________________________________________________________ test_failure _______________________________________________________________________________________________________________________________________________________________________
  File "/Users/oliver/workplace/snuba/tests/test_failure.py", line 2, in test_failure
    raise NotImplementedError()
NotImplementedError
================================================================================================================================================================= short test summary info =================================================================================================================================================================
FAILED tests/test_failure.py::test_failure - NotImplementedError
=============================================================================================================================================================== 1 failed, 1 passed in 7.07s ===============================================================================================================================================================
```

I think we should avoid catching an exception unless we know how to handle it/are doing something specific with it. I agree we can rewrite the above test to just not have a try/catch

ayirr7: The new test does not have an `assert False` (although the test above it, test_invalid_storage, does). Were you referring to removing the `assert` in that test?

I'm not entirely sure how catching the exception can be avoided in this case. If I understood the JIRA task correctly, we are supposed to temporarily modify KAFKA TOPIC MAP and run `validate_settings` on this new settings. `validate_settings` will throw an exception, see here: https://github.com/getsentry/snuba/blob/76b9d912e2ca4e1d5579a698d07450ba7c7ac43d/snuba/settings/validation.py#L62-L64. Since I think we are supposed to use this function, and this function throws an exception when the lists are not in sync (which is what we're checking for), I'm not sure how else we can make sure the test fails when it should. Unless we don't have to use `validate_settings`

tests/settings/test_settings.py
```diff
@@ -18,3 +32,25 @@ def test_invalid_storage() -> None:
...
+    default_map = deepcopy(all_settings["KAFKA_TOPIC_MAP"])
+    # Overwrite topic map temporarily to include all defined topic names
+    all_settings["KAFKA_TOPIC_MAP"] = {t.value: {} for t in Topic}
+
+    # Validate settings with the new topic map to check
+    # whether all defined topic names correspond to the
+    # topic names in the settings validator
+    try:
+        validate_settings(all_settings)
+    except Exception:
```
onewland, line 50: we want to catch your new exception type here, right?

tests/settings/test_settings.py
```diff
@@ -18,3 +32,25 @@ def test_invalid_storage() -> None:
         assert False, f"'validate_settings' raised an exception {exc}"
     finally:
         cluster[0]["storage_sets"].remove("non_existing_storage")
+
+
+def test_topics_sync_in_settings_validator() -> None:
+
```
onewland, line 38: nit: extra newline

tests/settings/test_settings.py
```diff
@@ -1,15 +1,29 @@
...
         key: value
         for key, value in settings.__dict__.items()
         if not key.startswith("__") and not callable(key)
     }
+
+    return all_settings
+
+
+def test_invalid_storage() -> None:
+
```
onewland, line 24: nit: extra newline

tests/settings/test_settings.py
```diff
@@ -1,15 +1,29 @@
...
+import pytest
+
 from snuba import settings
+from snuba.settings.validation import validate_settings
+from snuba.utils.streams.topics import Topic
 
 
-def test_invalid_storage() -> None:
-    from snuba.settings.validation import validate_settings
+def build_settings_dict() -> Dict[str, Any]:
```
onewland, line 12: nit: extra newline

onewland (approved):
Clean up whitespace and the exception type in the catch and this looks good to me

tests/settings/test_settings.py
ayirr7: oops! yes

PR #3063, opened 2022-08-16
feat(mdc): load entity definitions from yaml

onewland:
This PR adds support for loading entities from YAML. 

Changes in this PR:
- defines the `generic_metrics_sets` entity in YAML
- adds support in the entity factory for loading the definition from YAML, if a new setting is true (`settings.PREFER_PLUGGABLE_ENTITIES`, which will be off by default)
- puts new JSON schema validation around entities, along with some minor touch-ups to existing validations
- some changes to PluggableEntity that were necessary for `snuba api` to function
 
Next steps:
- mapping of support classes in a more coherent way (e.g. less centralized `query_processor[name] = class` but using something like https://github.com/getsentry/snuba/pull/3044 to have sub-classes add themselves to the "repository")
- generic_metrics_distributions definition

https://getsentry.atlassian.net/browse/SNS-1580


snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,136 @@
...
+
+_VALIDATOR_MAPPING: Mapping[str, Type[QueryValidator]] = {
+    "entity_required": EntityRequiredColumnValidator
+}
+
+logger = logging.getLogger("snuba.entity_builder")
+
+
+def _initialize_mappings() -> None:
+    # circular imports 🤪
```
onewland, line 49: Obviously the emoji comment shouldn't stay but I'm wondering what degree of bad we consider having to do these imports in-method. I looked at doing some rearranging but that would have really blown up the scope of this PR

snuba/query/processors/granularity_processor.py
```diff
@@ -48,11 +48,11 @@ class GranularityMapping(NamedTuple):
     enum_value: int
 
 
-PERFORMANCE_GRANULARITIES: Sequence[GranularityMapping] = [
-    GranularityMapping(60, 1),
-    GranularityMapping(3600, 2),
-    GranularityMapping(86400, 3),
-]
+PERFORMANCE_GRANULARITIES: Mapping[int, int] = {
```
evanh, line 51: Why did you make this change in this PR?

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,130 @@
...
```
evanh, line 48: Can you elaborate on this? Why is there a circular import here?

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,130 @@
...
+from snuba.datasets.pluggable_entity import PluggableEntity
+from snuba.datasets.storages import StorageKey
+from snuba.datasets.storages.factory import get_storage, get_writable_storage
+from snuba.query.processors import QueryProcessor
+from snuba.query.validation.validators import (
+    EntityRequiredColumnValidator,
+    QueryValidator,
+)
+
+__MAPPINGS_HAVE_BEEN_INITIALIZED = False
```
evanh, line 28: Why have a variable for this? Could you not check the length of mappings?

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    __MAPPINGS_HAVE_BEEN_INITIALIZED = True
+
+
+def _build_entity_validators(
+    config_validators: list[dict[str, Any]]
+) -> Sequence[QueryValidator]:
+    return [
+        _VALIDATOR_MAPPING[qv_config["validator"]](**qv_config["args"])  # type: ignore
```
evanh, line 81: I suspect this isn't working correctly, the `EntityRequiredColumnValidator` takes a `set` not an array.

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,130 @@
...
+        for qv_config in config_validators
+    ]
+
+
+def _build_entity_query_processors(
+    config_query_processors: list[dict[str, Any]],
+) -> Sequence[QueryProcessor]:
+    return [_QP_MAPPING[config_qp["processor"]](**(config_qp["args"] if config_qp.get("args") else {})) for config_qp in config_query_processors]  # type: ignore
```
evanh, line 89: One thing that should also be in next steps is better error handling I think. This will return a generic `KeyError` if someone misconfigures things.

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _build_entity_translation_mappers(
+    config_translation_mappers: dict[str, Any],
+) -> TranslationMappers:
+    function_mappers: list[FunctionCallMapper] = [
+        _FUNCTION_MAPPER_MAPPING[fm_config["mapper"]](**fm_config["args"])  # type: ignore
```
evanh, line 96: I understand we can't actually type these, but I would rather have loose typing than the ignore comment. We could achieve that (I think) by adding a `def __init__(self, *args: Any, **kwargs: Any)` to the `QueryProcessor` class.

evanh: This is actually already an issue, see my comment above about the `EntityRequiredColumnValidator`. Do we need to change processors to strictly validate their args? This feels like a larger discussion.

snuba/datasets/configuration/json_schema.py
```diff
@@ -119,22 +118,59 @@ def make_column_schema(
...
+    "required": ["validator"],
+}
+
+ENTITY_TRANSLATION_MAPPER_SUB_LIST = {
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "mapper": TYPE_STRING,
+            "args": {"type": "object"},
```
evanh, line 149: The args for `functions` and `subscriptables` are not arbitrary. I think you can add them to the schema as well and ensure they are correctly populated.

snuba/datasets/pluggable_entity.py
```diff
@@ -54,6 +55,8 @@ def get_all_join_relationships(self) -> Mapping[str, JoinRelationship]:
         return self.join_relationships
 
     def get_query_pipeline_builder(self) -> QueryPipelineBuilder[ClickhouseQueryPlan]:
+        from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
```
evanh, line 58: Is this also a circular import change?

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
```diff
@@ -0,0 +1,102 @@
+version: v1
+kind: entity
+name: generic_metrics_sets
```
enochtangg, line 3: Might be a dumb question, but why is this config name `sets.yaml` and why is the entity called `generic_metric_sets` as opposed to `generic_metrics`

snuba/cli/entities.py
```diff
@@ -60,3 +62,12 @@ def describe(entity_name: str) -> None:
         entity.describe().accept(CLIDescriber())
     except InvalidEntityError:
         click.echo(f"Entity {entity_name} does not exists or it is not registered.")
+
+
+@entities.command()
+@click.argument("entity_path", type=str)
+def load(entity_path: str) -> None:
```
enochtangg, line 69: What is this used for? Would it be used in the testing workflow when a developer adds a new entity?

enochtangg (commented):
Just had a few questions 👍

snuba/datasets/configuration/entity_builder.py
onewland: There was a circular import if these imports were at the top of the file.

snuba/datasets/pluggable_entity.py
onewland: yeah, it is

snuba/datasets/configuration/entity_builder.py
onewland: yeah, I agree completely

snuba/datasets/configuration/json_schema.py
onewland: Unfortunately they are a little bit arbitrary. For example, for `functions`, the expected argument type is `Sequence[FunctionCallMapper]`, which has multiple concrete subclasses like `DefaultFunctionMapper`, `FunctionNameMapper`, `AggregateCurriedFunctionMapper`.

snuba/cli/entities.py
onewland: exactly, just a quick way to validate your YAML if you haven't written tests yet

snuba/query/processors/granularity_processor.py
onewland: We need the constructors of QueryProcessors to conform with what yaml loading returns and this change was required to make the MappedGranularityProcessor constructor conform to a more primitive input time (`Mapping[int,int]`)

snuba/datasets/configuration/entity_builder.py
onewland: > This is actually already an issue, see my comment above about the `EntityRequiredColumnValidator`. Do we need to change processors to strictly validate their args? This feels like a larger discussion.

Maybe this could be incorporated into the registry/factory pattern where in addition to a friendly config name you also have to describe the method signature types? Maybe this could be done with introspection? Or it might be as simple as a practice of 
```
assert isinstance(arg1, type1) 
assert isinstance(arg2, type2) 
```

though that is pretty manual and error-prone

snuba/datasets/configuration/entity_builder.py
onewland: I was able to run queries so I think it's working correctly.

I just think we're not doing anything with that variable that requires set-specific behavior:

https://github.com/getsentry/snuba/blob/2c44f305625761fbde7acf84d7ff3d4da84459ef/snuba/query/validation/validators.py#L66-L74

but I agree that it's not ideal that we're able to pass an array here and not get any validation error on the constructor arguments

snuba/datasets/configuration/entity_builder.py
onewland: happy to change to a length check

snuba/datasets/configuration/entity_builder.py
onewland: I tried doing what you suggested with a method body of `raise NotImplementedError` but I it breaks the cases where there is no initializer defined on the subclass (see `TagsTypeTransformer`, `TagsExpanderTransformer`). Would you want to just have an __init__ with a pass as the method body?

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
onewland: The entity is the aspect of the product data model that you actually query. The dataset is `generic_metrics` which is really a container for entities and storages.

It's described in some detail on this page: https://getsentry.github.io/snuba/architecture/datamodel.html

If that's a typo and you're wondering why it's named `sets.yaml` vs. `generic_metrics_sets.yaml` it's just because generic_metrics is already in the path -- I'm OK with changing that though if we think it makes the code more discoverable

snuba/datasets/configuration/entity_builder.py
evanh: Per our live review: please leave a small comment explaining the tech debt problem.

snuba/datasets/configuration/entity_builder.py
evanh: After our discussion, let's leave this as is in this PR. There's a larger task here to figure out how to validate arguments in these situations that I will make a separate ticket for.

evanh (approved):
Minus a couple explanatory comments, this looks good.

PR #3064, opened 2022-08-17
feat(transactions): Add group_ids column to transactions

udameli:
To link transactions to their associated groups, this PR adds a new array column to transactions table. `group_ids` would contain at most 10 group ids.
I will follow up with changes to the transactions processor once the column is in prod.

Fixes WOR-2086

### DIFF
--- a/snuba/migrations/snuba_migrations/transactions/0016_transactions_add_group_ids_column.py
+++ b/snuba/migrations/snuba_migrations/transactions/0016_transactions_add_group_ids_column.py
@@ -1,40 +1,8 @@
-from typing import List, Sequence, Tuple
+from typing import Sequence
 
 from snuba.clickhouse.columns import Array, Column, UInt
 from snuba.clusters.storage_sets import StorageSetKey
 from snuba.migrations import migration, operations
-from snuba.migrations.columns import MigrationModifiers as Modifiers
-
-new_columns: Sequence[Tuple[Column[Modifiers], str]] = [
-    (Column("group_ids", Array(UInt(64))), "timestamp"),
-    (
-        Column(
-            "_group_ids_hashed",
-            Array(
-                UInt(64),
-                Modifiers(materialized="arrayMap(t -> cityHash64(t), group_ids)"),
-            ),
-        ),
-        "group_ids",
-    ),
-]
-
-new_indexes: List[operations.SqlOperation] = [
-    operations.AddIndex(
-        storage_set=StorageSetKey.TRANSACTIONS,
-        table_name="transactions_local",
-        index_name="bf_group_ids_hashed",
-        index_expression="_group_ids_hashed",
-        index_type="bloom_filter()",
-        granularity=1,
-    )
-]
-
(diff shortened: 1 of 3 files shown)


snuba/migrations/snuba_migrations/transactions/0016_transactions_add_group_ids_column.py
```diff
@@ -0,0 +1,86 @@
...
+new_columns: Sequence[Tuple[Column[Modifiers], str]] = [
+    (Column("group_ids", Array(UInt(64))), "timestamp"),
+    (
+        Column(
+            "_group_ids_hashed",
+            Array(
+                UInt(64),
+                Modifiers(materialized="arrayMap(t -> cityHash64(t), group_ids)"),
+            ),
+        ),
```
fpacifici, line 17: Why the hashed version ?
The group id is already a UInt64 so the hashed version is still the same type. THe bloom filter data structure already computes hashes (multiple of them) of the value.

snuba/migrations/snuba_migrations/transactions/0016_transactions_add_group_ids_column.py
```diff
@@ -0,0 +1,86 @@
...
+
+new_indexes: List[operations.SqlOperation] = [
+    operations.AddIndex(
+        storage_set=StorageSetKey.TRANSACTIONS,
+        table_name="transactions_local",
+        index_name="bf_group_ids_hashed",
+        index_expression="_group_ids_hashed",
+        index_type="bloom_filter()",
+        granularity=1,
+    )
```
fpacifici, line 30: I'd suggest we started without index.
If the assumption that the vast majority of transactions will not have issues, then we are likely not going to gain much by adding an index (it may actually be counter productive as we would have to load the index as well during queries).

Specifically if the array is empty most of the time, I suspect clickhouse will be as fast at excluding those blocks where the column is always empty than it would be through the index.

udameli: With the current detection logic we're finding that around 3-4 % of transaction events contain some kind of performance issue but with group creation rate limit and better thresholds, the actual rate will be even lower

snuba/migrations/snuba_migrations/transactions/0016_transactions_add_group_ids_column.py
udameli: addressed 👍

udameli: is there anything I need to do to explicitly enable bloom filter?

nikhars:
You need to add this migration to the group https://github.com/getsentry/snuba/blob/master/snuba/migrations/groups.py#L129

udameli:
> You need to add this migration to the group https://github.com/getsentry/snuba/blob/master/snuba/migrations/groups.py#L129

done 👍

nikhars (approved):

PR #3065, opened 2022-08-17
ref(MDC): Create Entity Factory Pattern

enochtangg:
Similar to dataset factory in https://github.com/getsentry/snuba/pull/3060, both YAML and Python defined entities will need to coexist. This entity factory pattern helps clean up the old method of initializing and accessing entities. 

1. One central container for all entities to live in (both yaml and python defined) 
2. Remove complicated mapping/caching logic
3. Remove unused `dev_entity_factories`

This PR is *not* responsible for the following:
1. Removing the EntityEnum (the entity factory still requires the full list of enums)
2. Dynamic importing of entities (each entity remains hard-coded in factory)

Additionally, the multithreaded tests for accessing the dataset factory was moved to the top of the test file to ensure test is ran with fresh objects imported from factory module.

### DIFF
--- a/snuba/datasets/entities/factory.py
+++ b/snuba/datasets/entities/factory.py
@@ -1,83 +1,118 @@
-from typing import Callable, MutableMapping
-
-from snuba import settings
+from typing import Generator, MutableMapping, Sequence, Type
 from snuba.datasets.entities import EntityKey
 from snuba.datasets.entity import Entity
 from snuba.datasets.table_storage import TableWriter
+from snuba.utils.config_component_factory import ConfigComponentFactory
 from snuba.utils.serializable_exception import SerializableException
+
+
+class _EntityFactory(ConfigComponentFactory[Entity, EntityKey]):
+    def __init__(self) -> None:
+        self._entity_map: MutableMapping[EntityKey, Entity] = {}
+        self._name_map: MutableMapping[Type[Entity], EntityKey] = {}
+        self.__initialize()
+
+    def __initialize(self) -> None:
+        from snuba.datasets.cdc.groupassignee_entity import GroupAssigneeEntity
+        from snuba.datasets.cdc.groupedmessage_entity import GroupedMessageEntity
+        from snuba.datasets.entities.discover import (
+            DiscoverEntity,
+            DiscoverEventsEntity,
+            DiscoverTransactionsEntity,
+        )
+        from snuba.datasets.entities.events import EventsEntity
+        from snuba.datasets.entities.functions import FunctionsEntity
+        from snuba.datasets.entities.generic_metrics import (
+            GenericMetricsDistributionsEntity,
+            GenericMetricsSetsEntity,
+        )
+        from snuba.datasets.entities.metrics import (
+            MetricsCountersEntity,
+            MetricsDistributionsEntity,
+            MetricsSetsEntity,
+            OrgMetricsCountersEntity,
+        )
+        from snuba.datasets.entities.outcomes import OutcomesEntity
+        from snuba.datasets.entities.outcomes_raw import OutcomesRawEntity
+        from snuba.datasets.entities.profiles import ProfilesEntity
+        from snuba.datasets.entities.replays import ReplaysEntity
+        from snuba.datasets.entities.sessions import OrgSessionsEntity, SessionsEntity
+        from snuba.datasets.entities.transactions import TransactionsEntity
+
+        self._entity_map.update(
+            {
+                EntityKey.DISCOVER: DiscoverEntity(),
+                EntityKey.EVENTS: EventsEntity(),
+                EntityKey.GROUPASSIGNEE: GroupAssigneeEntity(),
+                EntityKey.GROUPEDMESSAGES: GroupedMessageEntity(),
+                EntityKey.OUTCOMES: OutcomesEntity(),
+                EntityKey.OUTCOMES_RAW: OutcomesRawEntity(),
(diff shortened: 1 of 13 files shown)


snuba/datasets/entities/factory.py
```diff
@@ -1,89 +1,125 @@
...
+            raise InvalidEntityError(f"entity {entity} has no name") from error
+
+
 class InvalidEntityError(SerializableException):
     """Exception raised on invalid entity access."""
 
 
-ENTITY_IMPL: MutableMapping[EntityKey, Entity] = {}
-ENTITY_NAME_LOOKUP: MutableMapping[Entity, EntityKey] = {}
+_ENT_Factory = None
```
evanh, line 97:
```suggestion
_ENT_FACTORY = None
```

tests/datasets/test_dataset_factory.py
```diff
@@ -13,27 +13,7 @@
...
-        "functions",
-        "generic_metrics",
-        "replays",
-    ]:
-        factory_ds = get_dataset(ds_name)
-        assert isinstance(factory_ds, Dataset)
-        assert get_dataset_name(factory_ds) == ds_name
-
-
+# This test should be the first to ensure dataset factory module is has fresh set of objects
```
evanh, line 16:
```suggestion
# This test should be the first to ensure dataset factory module has a fresh set of objects
```

tests/datasets/test_entity_factory.py
```diff
@@ -0,0 +1,77 @@
...
+    EntityKey.ORG_METRICS_COUNTERS,
+    EntityKey.METRICS_DISTRIBUTIONS,
+    EntityKey.PROFILES,
+    EntityKey.FUNCTIONS,
+    EntityKey.REPLAYS,
+    EntityKey.GENERIC_METRICS_SETS,
+    EntityKey.GENERIC_METRICS_DISTRIBUTIONS,
+]
+
+# This test should be the first to ensure dataset factory module is has fresh set of objects
```
evanh, line 35: It occurs to me that you should be able to reset the value of `_ENT_FACTORY` in the test to ensure that it gets rebuilt.

```
from snuba.datasets.entities import factory # You can't import _ENT_FACTORY directly since that will copy it to the local scope
...
def test_...() -> None:
    factory._ENT_FACTORY = None
    # run the test knowing that the factory is now empty.
```

Otherwise this is a very brittle way of running these tests.

evanh: Obligatory SO post: https://stackoverflow.com/questions/3536620/how-to-change-a-module-variable-from-another-module

evanh: Oh actually we are doing something similar in an even easier way in other tests: https://github.com/getsentry/snuba/blob/master/tests/migrations/test_runner.py#L183

enochtangg: Seeing something quite strange, it looks like `importlib.reload(factory)` does indeed reset `factory._DS_FACTORY` however, the threads are somehow still able to pick previous "working" states of the factory. My hypothesis for why this is happening is because `reload` only recompiles the code in the module on disk and re-evaluates it with the existing `sys.modules.__dict__`. As a result, a new factory module object is not being created which causes it to still referencing some old symbol (even though doing an `id()` on the old and reloaded factory return the same).

snuba/datasets/entities/factory.py
```diff
@@ -1,89 +1,135 @@
...
 
 def enforce_table_writer(entity: Entity) -> TableWriter:
     writable_storage = entity.get_writable_storage()
 
     assert (
         writable_storage is not None
-    ), f"Entity {ENTITY_NAME_LOOKUP[entity]} does not have a writable storage."
+    ), f"Entity {_ent_factory().get_entity_name(entity)} does not have a writable storage."
     return writable_storage.get_table_writer()
+
```
enochtangg, line 126: Needed to add entity map override functionality to keep some test cases happy (ref. https://github.com/getsentry/snuba/blob/df9871f31d88da6e2d01ee7b0422280a63fa6b6e/tests/query/joins/test_equivalence_adder.py#L261)

tests/datasets/test_entity_factory.py
enochtangg: I did but I think it makes the module a bit confusing to read. Also, the import method from the SO post requires all sub-modules of `snuba.datasets` to be implicitly imported in the test just to access the factory sub-module.

Instead I've revised the test cases to force a failure if the factory object initialization is decoupled from the mapping initialization. This seems to have fixed the variable name spacing issue and removes the need to spin up more threads (which is be inconsistent).

tests/datasets/test_dataset_factory.py
```diff
@@ -81,7 +49,18 @@ def test_disabled(disable_datasets: Iterator[None]) -> None:
         get_dataset("events")
 
 
+def test_get_dataset_multithreaded_collision() -> None:
```
evanh, line 52: I don't think this is testing the same thing as the original test. The original test was ensuring that multiple threads initializing the factory at the same time wouldn't cause any issues.

enochtangg: Right, this does not explicitly test with multiple threads, but it ensures that the factory object initialization is not decoupled from the entity mapping creation. If atomicity is guaranteed between these operations, then multiple thread should have no issues accessing the factory in concurrently.

evanh: OK. Can you change the test name then please?

evanh (approved):

PR #3066, opened 2022-08-17
feat(transactions): Add group_ids to transactions processor

udameli:
This PRs adds an extra step to transaction processing for group_ids.


snuba/datasets/transactions_processor.py
```diff
@@ -135,6 +135,12 @@ def _process_base_event_values(
         processed["duration"] = max(int(duration_secs * 1000), 0)
 
         processed["platform"] = _unicodify(event_dict["platform"])
+
+        group_ids = event_dict["data"].get("group_ids") or []
```
udameli, line 139: I've noticed there's already `group_id` on transaction and it's not in `data`. Does it make sense to expect `group_ids` to be in `data`?

nikhars: The answer to this question depends on how data is written to the eventstream. I do not have the background on why certain fields were taken out from `data` and put into the outer message. But if on the eventstream you decide to write the `group_ids` only on the outer part of the message, you do not need to find it from `data`

tests/datasets/test_transaction_processor.py
```diff
@@ -16,6 +18,7 @@ class TransactionEvent:
     event_id: str
     trace_id: str
     span_id: str
+    group_ids: list[int]
```
nikhars, line 21: Instead of a list, you can use Sequence over here

tests/datasets/test_transaction_processor.py
```diff
@@ -49,11 +52,13 @@ def serialize(self) -> Tuple[int, str, Mapping[str, Any]]:
                 "event_id": self.event_id,
                 "message": "/organizations/:orgId/issues/",
                 "group_id": None,
+                "group_ids": self.group_ids,
```
nikhars, line 55: Ideally, it's best to have `group_ids` only in 1 part of the message, not both. The message sizes get bigger unnecessarily and there is really no value.

tests/datasets/test_transaction_processor.py
udameli: done!

udameli:
I'm making an assumption here that `group_ids` would in the outer part of the message similar to `group_id`. Note, `group_ids` and `group_id` will never be written at the same.

cc: @wedamija fyi this assumption affects how group ids are written in event stream

nikhars (approved):

snuba/datasets/transactions_processor.py
```diff
@@ -135,6 +135,12 @@ def _process_base_event_values(
         processed["duration"] = max(int(duration_secs * 1000), 0)
 
         processed["platform"] = _unicodify(event_dict["platform"])
+
+        group_ids = event_dict.get("group_ids") or []
+        if len(group_ids) > GROUP_IDS_LIMIT:
```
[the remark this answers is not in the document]
udameli: how would we know if the limit was exceeded if I truncate it here?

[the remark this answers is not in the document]
wedamija: Oh sorry, I missed the truncation a couple of lines down.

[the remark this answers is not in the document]
udameli: ah okay haha 👍

PR #3069, opened 2022-08-17
feat(MDC): Discover config files

rahul-kumar-saini:
### Overview
- Replaced hardcoded config file paths with auto discovery of config files
- These discovered config files dynamically populate the `CONFIG_BUILT_STORAGES` mapping

### DIFF
--- a/snuba/datasets/storages/factory.py
+++ b/snuba/datasets/storages/factory.py
@@ -1,4 +1,5 @@
 import logging
+from glob import glob
 from typing import Mapping
 
 from snuba import settings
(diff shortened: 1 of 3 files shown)


snuba/settings/__init__.py
```diff
@@ -239,9 +239,8 @@
 # Maximum jitter to add to the scheduling of threads of an optimize job
 OPTIMIZE_PARALLEL_MAX_JITTER_MINUTES = 30
 
-# File path for storage configs
-# TODO: This is hardcoded, the config files should be discovered automatically
-STORAGE_CONFIG_FILES_PATH = "./snuba/datasets/configuration/generic_metrics/storages"
+# File path glob for storage configs
+STORAGE_CONFIG_FILES_GLOB = "./snuba/datasets/configuration/**/storages/*.yaml"
```
onewland, line 243: I don't know the idiom in python but I think you want this to be something like:
 `f"{__application_root__}/snuba/datasets/configuration/**/storages/*.yaml"` to avoid that issue Volo ran into earlier of only being able to run commands from application root (also deploys might start things from a different directory)

snuba/datasets/storages/factory.py
```diff
@@ -63,19 +64,14 @@
...
-    StorageKey.GENERIC_METRICS_SETS_RAW: f"{settings.STORAGE_CONFIG_FILES_PATH}/sets_bucket.yaml",
-    StorageKey.GENERIC_METRICS_SETS: f"{settings.STORAGE_CONFIG_FILES_PATH}/sets.yaml",
-}
-
 CONFIG_BUILT_STORAGES = {
-    storage_key: build_storage(CONFIG_FILES[storage_key])
-    for storage_key in CONFIG_FILES
+    storage.get_storage_key(): storage
+    for storage in [
+        build_storage(config_file)
```
evanh, line 70: What happens here if a non-storage config file ends up in the folder?

evanh (approved):

onewland (approved):

snuba/datasets/storages/factory.py
rahul-kumar-saini: The glob function specifically looks for the pattern `/storages/*.yaml`. So if there is a yaml file under a folder named `storages` in any subfolder of `datasets/configuration/`, it _should_ be a storage. If it isn't, the validation or building the actual storage will raise exceptions well before anything is merged to master.

PR #3070, opened 2022-08-17
feat(MDC): Storage Factory Pattern

rahul-kumar-saini:
### Overview
- Refactored storage factory into standard dataset component factory pattern
- Builds off of #3069
- Inspired by #3065


snuba/datasets/cdc/groupassignee_entity.py
```diff
@@ -22,7 +23,7 @@ class GroupAssigneeEntity(Entity):
     """
 
     def __init__(self) -> None:
-        storage = get_cdc_storage(StorageKey.GROUPASSIGNEES)
+        assert isinstance(storage := get_storage(StorageKey.GROUPASSIGNEES), CdcStorage)
```
evanh, line 26: Why did you get rid of `get_cdc_storage`?

rahul-kumar-saini: I wanted to consolidate to a single `get_storage()` function but getting rid of `get_writable_storage()` would've been too much refactoring with `assert isinstance()` everywhere. `get_cdc_storage()` was only used in a few places so I replaced it with `get_storage()`. I can revert this back to having an individual `get_cdc_storage()` but my goal was to break it down to a single `get_storage()`.

evanh: I see what you're doing here. I think the problem is that this leaves us in a "halfway" state. Currently we have a different `get_*` function for each storage type (writable, readable, cdc, etc.). If we change that, but only for cdc storages, then it's a half solution. I'd rather leave this unchanged, and solve the problem fully in a different PR.

rahul-kumar-saini: Done

evanh (approved):

PR #3071, opened 2022-08-17
ref(MDC): Represent migrations in configuration

evanh:
This changes the migration groups to load automatically from configuration files
and populate the various global dictionaries we are currently using.

There is a new `ConfigurationLoader` which reads a migration group from a config
path. That path is detected automatically. The loaders are then added to the
global list of loaders and can be accessed in the same method as the hardcoded
loaders.

It is also showing one way to solve the "Enum" problem we have with different
components of Snuba.


lynnagara:
A couple of initial thoughts/questions:

1. What is the goal of representing individual migrations as configuration at this point? Is it due to safety concerns (running custom code)? Usability concerns with the current system? Even if we want to support this at a point, I believe we will always have a need for migrations as code, and am worried about taking this flexibility away from users. We needed code migrations a few times already such as when we rebuilt the errors table and I don't think this will go away.

8. The migration groups mechanism seems fine to me - what is the purpose of having "kind" and "version" here though? Version at least seems like an entirely new concept, why do we need it?

evanh:
The general consensus I'm gathering on this is: Don't make individual migrations part of the config, but making migration groups part of the configuration makes sense. I'm going to move forward with that.

> what is the purpose of having "kind" and "version" here though?

These are standard fields we are adding to all the config files, so they are self describing and we can change the schemas a little easier in the future.

lynnagara:
Separate to the representation, I think there will be a few other questions we will have to answer as part of this. How will the initial system migrations (which are mandatory) be fit into this system. How will dependencies between migration groups be resolved? We currently use the OPTIONAL_MIGRATION_GROUPS mechanism for experimental product features - what is the future of this?

snuba/datasets/configuration/generic_metrics/migrations/0001_sets_aggregate_table.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+  - add_column
+    storage_set: generic_metrics_sets
+    table_name: generic_metric_sets_local
+    column:
+      name: _raw_tags_hash
+      type: Array
+      args: { type: UInt, arg: 64 }
+      modifiers: { materialized: "arrayMap((k, v) -> cityHash64(concat(toString(k), '=', v)), tags.key, tags.raw_value)" }
+    after: null
```
fpacifici, line 66: @onewland, fyi, these columns should be created on the distributed table as well. In fact they are missing in production.

fpacifici:
> Don't make individual migrations part of the config, but making migration groups part of the configuration makes sense. I'm going to move forward with that.

Should we make migration groups auto-discoverable ? 
We did not do that at the beginning because there were corner cases that were not trivial to manage the sequence and potential merge conflicts in the same group. 
Though in a world where we want to reduce the friction and reduce the places the product teams need to touch it may be a direction to reconsider. 
We could do something like this:
- only mention a new group in settings when they are created, or even in the dataset configuration. 
- auto-discover the migration files in the group directory 
- expect the migration files to be in sequence (keep the number at the beginning). It is something relatively easy to automatically test.

snuba/migrations/groups.py
```diff
@@ -257,25 +254,68 @@ def get_migrations(self) -> Sequence[str]:
...
+    loader.migration_group_name: loader
+    for loader in [
+        ConfigurationLoader(config_file)
+        for config_file in glob(settings.MIGRATION_CONFIG_FILES_GLOB, recursive=True)
+    ]
+}
 
 
-_REGISTERED_GROUPS = {
+class _MigrationGroup(type):
```
volokluev, line 280: Why does this need to be a metaclass?

evanh: Because if I don't make it a metaclass, `__getattr__` will only work on instances of `MigrationGroup`, instead of on the class itself.

volokluev: Damn, we really getting funky now! Love it

snuba/datasets/configuration/migration_parser.py
```diff
@@ -0,0 +1,17 @@
...
+from jsonschema import validate
+from yaml import safe_load
+
+from snuba.datasets.configuration.migration_json_schema import MIGRATION_GROUP_SCHEMA
+
+
+def load_migration_group(path_to_file: str) -> dict[str, Any]:
+    yaml_file = open(path_to_file)
+    config = safe_load(yaml_file)
+    print("CONFIG", config)
```
volokluev, line 14: do we need this print in there?

snuba/migrations/groups.py
```diff
@@ -257,25 +255,68 @@ def get_migrations(self) -> Sequence[str]:
...
+
+        return MigrationGroup(attr.lower())
+
+    def __iter__(cls) -> Iterator[MigrationGroup]:
+        return iter(REGISTERED_GROUPS)
+
+
+class MigrationGroup(metaclass=_MigrationGroup):
+    """
+    This class is a replacement for the Enum class that used to define the different migration groups. In order to avoid having to go through the entire codebase to find instances of MigrationGroup and replace them, this will mimic the functionality of an Enum class while allowing arbitrary values to be populated.
```
volokluev, line 303: nit: can you break this docstring up over multiple lines?

tests/datasets/configuration/test_migration_groups.py
```diff
@@ -0,0 +1,29 @@
...
+        "0004_sets_raw_add_granularities",
+        "0005_sets_replace_mv",
+        "0006_sets_raw_add_granularities_dist_table",
+        "0007_distributions_aggregate_table",
+        "0008_distributions_raw_table",
+        "0009_distributions_mv",
+    ]
+
+    m = generic_metrics_loader.load_migration("0005_sets_replace_mv")
+    assert isinstance(m, Migration)
```
volokluev, line 29: since migrations are likely to be added, I would just iterate over them and `load_migration` in order to make sure that they are all indeed migrations. Maybe assert that the `len(migrations) > 5` or something.
 This way when another one is added, you won't need to change the test

volokluev (approved):
all my comments are not required for merging

snuba/migrations/groups.py
```diff
@@ -73,6 +50,27 @@ def load_migration(self, migration_id: str) -> Migration:
             raise MigrationDoesNotExist("Invalid migration ID")
 
 
+class ConfigurationLoader(DirectoryLoader):
```
lynnagara, line 53: I think this should inherit from the base `GroupLoader` not DirectoryLoader which supposed to be a completely separate implementation

snuba/migrations/groups.py
lynnagara: I don't think this enum like interface is the right one anymore since Migration groups can be defined on the fly. It made sense previously because all of the members were defined in code. The rest of the codebase never references MigrationGroup (only a handful of places within the migrations module) so there are not that many places to change.

lynnagara:
Please also update MIGRATIONS.md

snuba/migrations/groups.py
evanh: I did this because otherwise I'd have to copy the `load_migration` function over.

lynnagara:
Reposting an earlier comment since it was marked resolved without a response. I'm still interested in your thoughts on this. If you have other reasons for keeping this interface I'll be happy to hear them but the only reason stated is to avoid changing a few spots in the codebase and that just doesn't seem right to me.

I don't think this enum like interface is the right one anymore since Migration groups can be defined on the fly. It made sense previously because all of the members were defined in code. The rest of the codebase never references MigrationGroup (only a handful of places within the migrations module) so there are not that many places to change.

PR #3072, opened 2022-08-18
feat(admin):  allow show command to system queries SNS-1625

dbanda:
This will allow [`SHOW` commands ](https://clickhouse.com/docs/en/sql-reference/statements/show/) in system queries on snuba admin. Show queries do not reveal customer data and are safe, read-only. Addresses JIRA https://getsentry.atlassian.net/browse/SNS-1625. 

Reviewers, please confirm that the Regex is safe. Right now, it is `SHOW [\w\s\]+$` so theoretically an adversary could insert arbitrary subqueries with the `[\w\s\]` matching. We can probably come up with a more complex regex to skip matching any unsafe subqueries but I'm not sure if its worth the work given that the admin threat model is targeted towards accidental mistakes rather than adversarial attacks.

### DIFF
--- a/snuba/admin/clickhouse/system_queries.py
+++ b/snuba/admin/clickhouse/system_queries.py
@@ -62,12 +62,32 @@ DESCRIBE_QUERY_RE = re.compile(
     re.VERBOSE,
 )
 
+SHOW_QUERY_RE = re.compile(
+    r"""
+        ^ # Start
+        (SHOW|show)
+        \s
+        [\w\s]+
+        ;? # Optional semicolon
+        $ # End
+    """,
+    re.VERBOSE,
+)
+
 
 def is_query_select(sql_query: str) -> bool:
     """
     Simple validation to ensure query is a select command
     """
     match = SYSTEM_QUERY_RE.match(sql_query)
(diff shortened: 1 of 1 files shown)


snuba/admin/clickhouse/system_queries.py
```diff
@@ -71,6 +83,14 @@ def is_query_select(sql_query: str) -> bool:
     return True if match else False
 
 
+def is_query_show(sql_query: str) -> bool:
+    """
+    Simple validation to ensure query is a show command
+    """
+    match = SHOW_QUERY_RE.match(sql_query)
+    return True if match else False
```
volokluev, line 91: nit
```suggestion
    return bool(SHOW_QUERY_RE.match(sql_query))
```

volokluev (approved):

snuba/admin/clickhouse/system_queries.py
dbanda: Did it that way to match the style of the other "is_query" funcs. But your style is definitely more concise.

nikhars (approved):

PR #3074, opened 2022-08-18
fix(replays): process trace_ids / error_ids as uuids

JoshFerge:
- process the trace_ids / error_ids as uuids. This normalizes uuids without dashes / with dashes.
- also apply list element limits to trace_ids

### DIFF
--- a/snuba/datasets/replays_processor.py
+++ b/snuba/datasets/replays_processor.py
@@ -31,15 +31,12 @@ from snuba.utils.metrics.wrapper import MetricsWrapper
 logger = logging.getLogger(__name__)
 
 metrics = MetricsWrapper(environment.metrics, "replays.processor")
-from sentry_sdk import capture_exception
 
 ReplayEventDict = Mapping[Any, Any]
 RetentionDays = int
 
-# A sane upper bound of 1000 was chosen for error_ids. If you know better, change it.
-ERROR_IDS_LIMIT = 1000
-# A sane upper bound of 1000 was chosen for urls. If you know better, change it.
-URLS_LIMIT = 1000
+# Limit for error_ids / trace_ids / urls array elements
+LIST_ELEMENT_LIMIT = 1000
 
 USER_FIELDS_PRECEDENCE = ("user_id", "username", "email", "ip_address")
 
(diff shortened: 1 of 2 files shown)


snuba/datasets/replays_processor.py
```diff
@@ -225,5 +235,5 @@ def process_message(
             return InsertBatch([processed], None)
         except Exception as e:
             metrics.increment("consumer_error")
-            capture_exception(e)
+            logger.exception(e)
```
cmanallen, line 238: `logger.exception("Replay could not be consumed.")` or something similar.  Traceback is pulled implicitly.

snuba/datasets/replays_processor.py
```diff
@@ -81,12 +78,29 @@ def __extract_url(self, replay_event: ReplayEventDict) -> list[str]:
...
+        return [str(uuid.UUID(t)) for t in trace_ids[:LIST_ELEMENT_LIMIT]]
+
+    def __process_error_ids(self, error_ids: list[str] | None) -> list[str]:
+        if not error_ids:
+            return []
+        if len(error_ids) > LIST_ELEMENT_LIMIT:
+            metrics.increment("error_ids exceeded list limit")
+
+        return [str(uuid.UUID(e)) for e in error_ids[:LIST_ELEMENT_LIMIT]]
+
```
cmanallen, line 96: This section is interesting.  Clickhouse stores UUID's as 128-bit integers[1].  Normalizing the UUID should have no effect unless clickhouse preserves ingest style?  That seems unlikely though.  Are you seeing UUIDs ingested with/without dashes returning different response outputs?

1. https://github.com/ClickHouse/ClickHouse/blob/master/base/base/UUID.h#L8

cmanallen (changes requested):

snuba/datasets/replays_processor.py
JoshFerge: yes, uuids without dashes are failing for me locally. (also see we were already doing this for replay_id, and do this same processing on other uuids in other processors)

snuba/datasets/replays_processor.py
JoshFerge: right.

snuba/datasets/replays_processor.py
cmanallen: Makes sense to validate regardless.

cmanallen (approved):

snuba/datasets/replays_processor.py
cmanallen: I see this is unrelated to the query issue.  That's right.  I remember you said that now.

JoshFerge: 👍🏼

evanh (approved):

PR #3076, opened 2022-08-19
ref: switch from mywsgi/uwsgi to pyuwsgi

asottile-sentry:
pyuwsgi provides wheels making it much faster to install whereas uwsgi isn't wheelable

this also cuts about 10 minutes from the uncached arm64 image build (17 minutes -> 7 minutes)

### DIFF
--- a/snuba/cli/api.py
+++ b/snuba/cli/api.py
@@ -5,6 +5,7 @@ import click
 
 from snuba.attribution.log import flush_attribution_producer
 from snuba.environment import setup_logging
+from snuba.utils import uwsgi
 
 
 @click.command()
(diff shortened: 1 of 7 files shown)


snuba/utils/uwsgi.py
```diff
@@ -0,0 +1,75 @@
...
+        "protocol": protocol,
+        "single_interpreter": True,
+        "threads": 1,
+        "thunder_lock": True,
+        "vacuum": True,
+        "virtualenv": sys.prefix,
+        "wsgi_env_behavior": "holy",
+        f"{protocol}_socket": bind,
+        **kwargs,
+    }
```
asottile-sentry, line 69: these are copied from `mywsgi` which appear to be copied from `sentry`

asottile-sentry:
not planning to merge or roll this out until after `sentry` has and after hackweek -- unless there's interest in doing it sooner rather than later :shrug:

snuba/cli/api.py
```diff
@@ -46,13 +47,11 @@ def api(
         WSGIRequestHandler.protocol_version = "HTTP/1.1"
         application.run(host=host, port=port, threaded=True, debug=debug)
     else:
-        import mywsgi
-
         if log_level:
             os.environ["LOG_LEVEL"] = log_level
 
         with flush_attribution_producer():
```
asottile-sentry, line 53: unrelated to this patch but this call does nothing!  both `myuwsgi.run` and `snuba.utils.uwsgi.run` end with `os.execvp` which replaces the current process

evanh (approved):

PR #3077, opened 2022-08-19
feat(replacer): Create bucket timer class to track replacement processing

enochtangg:
Currently, replacement consumers are prone to be overloaded if a single consumer issues too many replacements. This PR is responsible for building observability into which projects are spending the majority of the processing time in a replacement batch.

To accomplish this, a `Bucket` is used to capture per minute slices and the processing time a project spends within the minute. A `Counter` class used to keep track of the past 10 minutes/buckets. When a replacer worker is started, a counter attribute is instantiate with the worker. Therefore, this counter persists through batches, and is only restarted when the service itself restarts. After a replacement is processed, the the counter inserts the processing time into bucket(s) and is stored within the `Counter`. Finally, the buckets are grouped by `project_id` and their totals (over the past WINDOW_SIZE=10 minutes) is compared to `WINDOW_SIZE * some percentage). If the total is greater (aka takes up the majority of the system's processing time), then write a metric to DataDog.

### DIFF
--- a/snuba/replacer.py
+++ b/snuba/replacer.py
@@ -28,7 +28,7 @@ from snuba.replacers.replacer_processor import (
     ReplacementMessageMetadata,
 )
 from snuba.state import get_config
-from snuba.utils.bucket_timer import Counter, compare_counters_and_write_metric
+from snuba.utils.bucket_timer import Counter, compare_counter_and_write_metric
 from snuba.utils.metrics import MetricsBackend
 from snuba.utils.rate_limiter import RateLimiter
 
@@ -271,7 +271,7 @@ class ReplacerWorker(AbstractBatchWorker[KafkaPayload, Replacement]):
         self.__storage = storage
 
         self.metrics = metrics
-        self.__global_counter = Counter()
+        self.__process_time_counter = Counter()
         processor = storage.get_table_writer().get_replacer_processor()
         assert (
             processor
@@ -390,7 +390,6 @@ class ReplacerWorker(AbstractBatchWorker[KafkaPayload, Replacement]):
             ClickhouseClientSettings.REPLACE
         )
 
-        project_counter = Counter()
         for replacement in batch:
             start_time = datetime.now()
 
@@ -422,9 +421,7 @@ class ReplacerWorker(AbstractBatchWorker[KafkaPayload, Replacement]):
             if callable(get_project_id):
                 project_id = get_project_id()
                 end_time = datetime.now()
-                self._update_and_compare_counters(
-                    project_counter, start_time, end_time, project_id
-                )
+                self._update_and_compare_counter(start_time, end_time, project_id)
 
         if need_optimize:
             from snuba.optimize import run_optimize
@@ -520,23 +517,17 @@ class ReplacerWorker(AbstractBatchWorker[KafkaPayload, Replacement]):
             ]
         )
 
-    def _update_and_compare_counters(
+    def _update_and_compare_counter(
         self,
-        project_counter: Counter,
         start_time: datetime,
         end_time: datetime,
         project_id: int,
     ) -> None:
-        project_counter.write_to_bucket(
+        self.__process_time_counter.write_to_bucket(
             project_id,
             start_time,
             end_time,
         )
-        self.__global_counter.write_to_bucket(
-            None,
-            start_time,
-            end_time,
-        )
-        compare_counters_and_write_metric(
-            self.__global_counter, project_counter, self.__consumer_group
+        compare_counter_and_write_metric(
+            self.__process_time_counter, self.__consumer_group
         )
--- a/snuba/replacer.py
+++ b/snuba/replacer.py
@@ -12,13 +12,14 @@ from arroyo import Message
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.processing.strategies.batching import AbstractBatchWorker
 
-from snuba import settings
+from snuba import environment, settings
 from snuba.clickhouse.native import ClickhousePool
 from snuba.clusters.cluster import (
     ClickhouseClientSettings,
     ClickhouseCluster,
     ClickhouseNode,
 )
+from snuba.datasets.errors_replacer import Replacement as ErrorReplacement
 from snuba.datasets.storage import WritableTableStorage
 from snuba.processor import InvalidMessageVersion
 from snuba.redis import redis_client
@@ -30,6 +31,7 @@ from snuba.replacers.replacer_processor import (
 from snuba.state import get_config
 from snuba.utils.bucket_timer import Counter
 from snuba.utils.metrics import MetricsBackend
+from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.rate_limiter import RateLimiter
 
 logger = logging.getLogger("snuba.replacer")
@@ -417,12 +419,12 @@ class ReplacerWorker(AbstractBatchWorker[KafkaPayload, Replacement]):
 
             self._check_timing_and_write_to_redis(replacement, start_time.timestamp())
 
-            get_project_id = getattr(replacement, "get_project_id", None)
-            if callable(get_project_id):
-                project_id = get_project_id()
+            if isinstance(replacement, ErrorReplacement):
+                project_id = replacement.get_project_id()
                 end_time = datetime.now()
-                self._update_and_compare_counter(start_time, end_time, project_id)
-
+                self._attempt_emitting_metric_for_projects_exceeding_limit(
+                    start_time, end_time, project_id
+                )
         if need_optimize:
             from snuba.optimize import run_optimize
 
@@ -517,15 +519,30 @@ class ReplacerWorker(AbstractBatchWorker[KafkaPayload, Replacement]):
             ]
         )
 
-    def _update_and_compare_counter(
+    def _attempt_emitting_metric_for_projects_exceeding_limit(
         self,
         start_time: datetime,
         end_time: datetime,
         project_id: int,
     ) -> None:
-        self.__processing_time_counter.write_to_bucket(
+        """
+        Attempts to emit a metric for projects which have a total processing time greater than some limit. This is accomplished
+        by first writing the time spent on processing the current replacement in Counter class. Then, a list of all projects
+        exceeding some limit is obtained and a metric is written to DataDog.
+        """
+        self.__processing_time_counter.record_time_spent(
             project_id,
             start_time,
             end_time,
         )
-        self.__processing_time_counter.compare_and_write_metric()
+        projects_exceeding_limit = (
+            self.__processing_time_counter.get_projects_exceeding_limit()
+        )
+        for project_id in projects_exceeding_limit:
+            metrics = MetricsWrapper(
+                environment.metrics, "replacer", tags={"group": self.__consumer_group}
+            )
+            metrics.increment(
+                "project_processing_time_exceeded_time_limit",
+                tags={"project_id": str(project_id)},
+            )
--- a/snuba/replacer.py
+++ b/snuba/replacer.py
@@ -28,7 +28,7 @@ from snuba.replacers.replacer_processor import (
     ReplacementMessageMetadata,
 )
 from snuba.state import get_config
-from snuba.utils.bucket_timer import Counter, compare_counter_and_write_metric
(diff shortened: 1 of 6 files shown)


snuba/replacer.py
```diff
@@ -414,7 +415,13 @@ def flush_batch(self, batch: Sequence[Replacement]) -> None:
 
             self.__replacer_processor.post_replacement(replacement, count)
 
-            self._check_timing_and_write_to_redis(replacement, start_time)
+            self._check_timing_and_write_to_redis(replacement, start_time.timestamp())
+
+            get_project_id = getattr(replacement, "get_project_id", None)
```
fpacifici, line 420: Please do not go around the type checker.
The `Replacement` object does not have a `get_project_id` because it is higher level than an `ErrorReplacement`. 
So if the project id is needed outside the replacer processor, you will have to change the abstraciton and move the `get_project_id` method to the parent class. 

The fact the method is in the subclass but you have access to the parent class implies you are only supposed to use the methods from the parent class otherwise. Having the method in the subclass means you could receive a different type of replacement here that does not have `get_project_id` or that `get_project_id` could be removed from the replacement subclass. This code is not supposed to depend on it. 
Going around that abstraction fundamentally breaks the design.

fpacifici (changes requested):
First one comment about the Replacement class. Will continue the review later.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        self.consumer_group: str = consumer_group
+        self.buckets: Deque[Bucket] = deque()
+
+    def get_time_sum(self, group: List[timedelta]) -> timedelta:
+        return sum([processing_time for processing_time in group], timedelta())
+
+    def floor_minute(self, time: datetime) -> datetime:
+        return time - timedelta(seconds=time.second, microseconds=time.microsecond)
+
+    def ceil_minute(self, time: datetime) -> datetime:
```
fpacifici, line 39: These methods seem to be used only within this class and do not seem to fit the public interface of this class. Please make them "private" `__get_time_sum` or, move them out of the class entirely if they do not need to access the state of the object.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        right = self.ceil_minute(start)
+        while right <= end:
+            self.create_and_add_bucket(project_id, start_minute, right - left)
+            left = start_minute = right
+            right += timedelta(minutes=1)
+        self.create_and_add_bucket(project_id, start_minute, end - left)
+
+    def print_buckets(self) -> None:
+        for bucket in self.buckets:
+            print(bucket.minute, bucket.project_id, bucket.processing_time)
```
fpacifici, line 85: This does not seem to fit within the scope of the class.
This is a class that manages data structures and it is self contained. Assuming there is a legitimate use case for printing the state on the console (I doubt) adding it into the class means mixing the presentation concern (print) with the internal functioning of the class. 
If you need to print the buckets, have a method that returns the list of buckets and print it outside.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def compare_and_write_metric(self) -> None:
```
fpacifici, line 87: This is another case where we are mixing the presentation with the  bucketing rate limiter logic. In order to keep the two responsibilities separate, please return the list of outstanding project and let the code that invoke this method decide what to do with that. 

Emitting the metrics inside ties the design of this class (which really is only about counting) with the purpose you are using this class for, which is instead relevant code in the replacer.
Example of how this approach breaks: Recording a metric in this case is only the initial step, we will want to replace this logic with something more drastic, which will demote the offending project ids to a separate queue.
When we will introduce that logic, the accounting logic will not change. The accounting policy stays the same, though we will have to touch this class because it combines the accounting function with the action to take. 
If you stick to one responsibility per class this will not happen, the counter will remain the same, which will be a sign of a better modularization.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        now = datetime.now()
+        self.trim_expired_buckets(now)
+        project_groups = defaultdict(list)
+        for bucket in self.buckets:
+            project_groups[bucket.project_id].append(bucket.processing_time)
+
+        # Compare the replacement total grouped by project_id with system time
+        for project_id, processing_times in project_groups.items():
+            project_processing_time = self.get_time_sum(processing_times)
+            if project_processing_time > WINDOW_SIZE * 0.5 and len(project_groups) > 1:
```
fpacifici, line 100: I think this 0.5 needs to be a runtime parameter. I bet we will be adjusting it quite a bit. 
https://github.com/getsentry/snuba/blob/master/snuba/state/__init__.py#L191

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+from datetime import datetime, timedelta
+from typing import Deque, List, Optional
+
+from snuba import environment
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+WINDOW_SIZE = timedelta(minutes=10)  # minutes
+
+
+class Bucket:
```
fpacifici, line 13: It seems this could be a dataclass (from dataclasses import dataclass) . So you will not have to write the constructor.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import annotations
+
+from collections import defaultdict, deque
+from datetime import datetime, timedelta
+from typing import Deque, List, Optional
+
+from snuba import environment
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+WINDOW_SIZE = timedelta(minutes=10)  # minutes
```
fpacifici, line 10: Please define the window as a setting (settings/__init__.py)

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        processing_time: timedelta,
+    ) -> None:
+        bucket = self.get_existing_bucket(project_id, start_minute)
+        if not bucket:
+            bucket = Bucket(project_id, start_minute, processing_time)
+            self.buckets.append(bucket)
+        else:
+            bucket.add_processing_time_seconds(processing_time)
+
+    def write_to_bucket(
```
fpacifici, line 71: nit: naming. The public interface of a class should be intuitive and give you an idea of what the class does for you (more interesting that how it does it). A good interface is one you do not know how to document because it is obvious. Think about a list: `append`, `get`. 

If you were describing what this class does you would probably explain it as:
- it accounts for the time spent on some activity for a project
- it returns the project ids that spent too much time.

So I would call this method something like `account`, `record_time_spent`. And I would call `compare_and_write_metric` something like `get_top_projects` or `get_greedy_projects`, or anything that give you an idea of a property of the projects you are returning. 

Calling this method `write_to_bucket` does not tell me what you are writing and why I would use it. On the other end it exposes an implementation detail (buckets) that the call site should not know about because it is irrelevant to the interface.
Calling the method below `compare_and_write_metric` instead does not tell me what it is comparing with what. 
Basically this naming would assume the call site knows the internals of the class, which is exactly what you want to hide by adding a class.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        return time - timedelta(seconds=time.second, microseconds=time.microsecond)
+
+    def ceil_minute(self, time: datetime) -> datetime:
+        if time.second == 0 and time.microsecond == 0:
+            return time
+        return self.floor_minute(time + timedelta(minutes=1))
+
+    def trim_expired_buckets(self, now: datetime) -> None:
+        current_minute = self.floor_minute(now)
+        window = current_minute - WINDOW_SIZE
```
fpacifici, line 46: nit: rename window with `window_start` or soemthing that tells you that it is the beginning of the window. The window is an interval while you represent a point in time with this variable.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        while self.buckets and self.buckets[0].minute < window:
+            self.buckets.popleft()
+
+    def get_existing_bucket(
+        self, project_id: Optional[int], minute: datetime
+    ) -> Optional[Bucket]:
+        for bucket in self.buckets:
+            if bucket.project_id == project_id and bucket.minute == minute:
+                return bucket
+        return None
```
fpacifici, line 56: You are using this method only once, it is only 3 statements long and it represent code at the same abstraction level compared to the code where it is called from.
I think you can just inline the method there.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+        project_id: Optional[int],
+        minute: datetime,
+        processing_time: timedelta = timedelta(seconds=0),
+    ) -> None:
+        self.project_id = project_id  # is None for global bucket
+        self.minute = minute
+        self.processing_time = processing_time
+
+    def add_processing_time_seconds(self, seconds: timedelta) -> None:
+        self.processing_time += seconds
```
fpacifici, line 25: Do you really need this method ? The method name itself is longer than increasing the value of the property.
`bucket.add_processing_time_seconds(seconds)` vs `bucket.processing_time += seconds`

Basically I think having this method is actually doing more harm than good by disrupting the reading process while the abstraction value is questionable.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def get_existing_bucket(
+        self, project_id: Optional[int], minute: datetime
+    ) -> Optional[Bucket]:
+        for bucket in self.buckets:
+            if bucket.project_id == project_id and bucket.minute == minute:
+                return bucket
+        return None
+
+    def create_and_add_bucket(
```
fpacifici, line 58: Please rename to something like `add_to_bucket`. 
`create_and_add` means you are always creating a bucket, which is not true as you are just incrementing the existing bucket if it is there.

fpacifici (commented):
I think the approach is ok. See my comments in line

enochtangg:
Thanks for the detailed review @fpacifici. Learned lots about structuring code in a modular format in Snuba for the future. I've revised my code to address your comments.

fpacifici (changes requested):
Thanks. this is way better. 

Please see my comment in line specifically the one about not using `project_id` as tags in our metrics.

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,95 @@
...
+    To accomplish this, the `record_time_spent()` function captures some processing time range and splits it by a per
+    minute resolution (Bucket). The buckets older than settings.COUNTER_WINDOW_SIZE are trimmed. Finally, the `get_bucket_totals_exceeding_limit()`
+    function returns all project ids who's total processing time has exceeded self.limit.
+    """
+
+    def __init__(self, consumer_group: str) -> None:
+        self.consumer_group: str = consumer_group
+        self.buckets: Deque[Bucket] = deque()
+
+        percentage = state.get_config("counter_window_size_percentage", 1.0)
```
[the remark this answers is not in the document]
enochtangg: Renamed to `project_quota_time_percentage`

snuba/utils/bucket_timer.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def __add_to_bucket(
+        self,
+        project_id: int,
+        start_minute: datetime,
+        processing_time: timedelta,
+    ) -> None:
+        curr_bucket = None
+        for bucket in self.buckets:
+            if bucket.project_id == project_id and bucket.minute == start_minute:
```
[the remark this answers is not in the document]
enochtangg: Makes sense, extending this idea further, I would propose that self.__buckets can even be a nested dictionary to leverage quicker access by whole minute AND project_id.

```
self.__buckets: MutableMapping[datetime, MutableMapping[int, timedelta]] = {}
```

If we let N be the total number of minutes saved and M be the total number of project_ids. Then...
For recording a bucket (creating new entry or adding processing time to existing bucket):
* Nested dictionary: O(1)
* Dictionary with deque: O(N) 

For trimming buckets:
* Nested dictionary: O(N)
* Dictionary with deque: O(N*M) <-- in most cases N will be smaller though because the probability of needing to trim the entire window is low

fpacifici (approved):

PR #3082, opened 2022-08-22
refs: Remove `EventsProcessor` and `EventsProcessorBase`

wedamija:
Was reading through this code to understand error processing, and found the two processors a bit
confusing. It seems like `EventsProcessor` is now completely unused, so I removed it.

It seems like `EventsProcessorBase` is also unnecessary now, since we only have `ErrorsProcessor`
subclassing it, so also removed that and consolidated all the logic into `ErrorsProcessor`.

It could make sense to move the various enums and type definitions out of the `events_processor_base`
file as well. Can do so in this pr or a follow-up if people agree.


wedamija:
This only moves code around - no logic changes should be present. Confirmed this with `git diff master --color-moved=dimmed_zebra`

PR #3083, opened 2022-08-26
upgrade mypy to 0.971

asottile-sentry:


snuba/request/validation.py
```diff
@@ -85,11 +85,8 @@ def build_request(
                 }
                 query_settings["referrer"] = referrer
                 # TODO: referrer probably doesn't need to be passed in, it should be from the body
-                settings_obj: Union[
-                    HTTPQuerySettings, SubscriptionQuerySettings
-                ] = settings_class(
-                    **query_settings,
-                )
+                settings_obj: Union[HTTPQuerySettings, SubscriptionQuerySettings]
+                settings_obj = settings_class(**query_settings)  # type: ignore
```
evanh, line 89: Why is this ignored now?

asottile-sentry: the parameters are either `str` or `bool` but don't accept `str | bool` -- previously mypy didn't check this call at all

the only way mypy would be able to figure out this call is if this was a `TypedDict` -- https://github.com/getsentry/snuba/blob/b85fb5b9f68d868f2d57a3927440602d9410eda6/snuba/query/query_settings.py#L67-L73

evanh: Got it. Can you leave a quick comment so other people don't get tripped up?

evanh (approved):

PR #3089, opened 2022-08-29
feat: Add initial documentation for Dataset configuration

evanh:
Add some quick placeholder files for Dataset configuration. Any new components
should have a new file created (`entity.rst`) and then add that file name to
`overview.rst` so it gets indexed in the ToC.

### DIFF
--- a/docs/source/configuration/dataset.rst
+++ b/docs/source/configuration/dataset.rst
@@ -1,0 +1,10 @@
+Dataset
+========
+
+A Dataset is a container of components necessary to ingest and read data from Clickhouse.
+
+Fields
+---------
+
+name : string
+    The name of the Dataset.
--- a/docs/source/configuration/intro.rst
+++ b/docs/source/configuration/intro.rst
@@ -1,0 +1,5 @@
+=====================
+Dataset Configuration
+=====================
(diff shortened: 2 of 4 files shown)


fpacifici:
What is your strategy for documenting the yaml config? Are we planning to automatically generate it or to manually write it?

Since we validate with jsonschema, there is a clean way to automatically generate a reference documentation and keep it close to the schema code:
https://github.com/getsentry/sentry-data-schemas/blob/main/relay/event.schema.json#L36

This is how we do for the event schema. We also generate (or at least used to) the official sentry documentation for the event payload from the jsonschema doc, which is produced by relay. Those description fields are in the Relay code together with the schema definition.

I think we should try to follow a similar approach: keep the reference docs together with the code, make it part of the jsonschema and generate the readable docs from there automatically.
Then manually write a description of the main concepts: why yaml? how to validate? etc.

evanh:
The plan is to have the schemas be automatically documented. So the schemas will be annotated with a description, and then docs can be generated from that and put into the Snuba documentation.

This is a placeholder for now so that other people can add documentation.

onewland (approved):

PR #3090, opened 2022-08-29
ref(subscriptions): Remove max_concurrent_queries cli arg

MeredithAnya:
**context:**
This is a follow up of the completion of https://github.com/getsentry/snuba/pull/2761. Now that the executors are using the calculated max concurrent queries in production, we can get rid of the cli arg. 

https://github.com/getsentry/ops/pull/5172 must be merged first so that we can safely remove the cli without there being a reference to it in the ops repo.


evanh (approved):
🔥

PR #3093, opened 2022-08-29
Revert "revert(mdc): revert broken entity loader change (#3088)"

onewland:
This reverts commit 8002e3fbb359edb125d0ca8c2d459fbe0f124764, restoring the config-based entity loading, with changes for Enoch's new entity factory

Tested `snuba api` locally with `settings.PREFER_PLUGGABLE_ENTITIES` set to both True and False


snuba/cli/entities.py
```diff
@@ -60,3 +62,12 @@ def describe(entity_name: str) -> None:
         entity.describe().accept(CLIDescriber())
     except InvalidEntityError:
         click.echo(f"Entity {entity_name} does not exists or it is not registered.")
+
+
+@entities.command()
```
volokluev, line 67: this is a script with a purpose, can you put its purpose in a comment in this file?

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,135 @@
...
+    "simple_func": FunctionNameMapper
+}
+
+_SUB_MAPPER_MAPPING: Mapping[str, Type[SubscriptableReferenceMapper]] = {
+    "subscriptable": SubscriptableMapper
+}
+
+_VALIDATOR_MAPPING: Mapping[str, Type[QueryValidator]] = {
+    "entity_required": EntityRequiredColumnValidator
+}
```
volokluev, line 40: Are these all incomplete? If so, that should be specified with a `#TODO` otherwise it's confusing. I know for a fact there are other validators

snuba/datasets/configuration/entity_builder.py
```diff
@@ -0,0 +1,135 @@
...
+
+logger = logging.getLogger("snuba.entity_builder")
+
+
+def _initialize_mappings() -> None:
```
volokluev, line 45: this pattern can cause threading issues. instead, have the function return the constructed object.

```python3
_QP_MAPPING = None

def _qp_mapping() -> dict[str, Type[QueryProcessor]]:
    global _QP_MAPPING
    if not _QP_MAPPING:
        # initialize here
    return _QP_MAPPING
```

This way if there's a race condition, worst case is you'll have two threads construct and return the same object

snuba/datasets/entities/factory.py
```diff
@@ -65,7 +71,13 @@ def __initialize(self) -> None:
...
         )
 
-        # TODO: load the yaml entities here
+        if settings.PREFER_PLUGGABLE_ENTITIES:
+            self._entity_map.update(
+                {
+                    key: build_entity_from_config(path)
+                    for (key, path) in entity_to_config_path_mapping.items()
+                }
+            )
```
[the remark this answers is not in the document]
volokluev: This will require a redeploy to change back and forth won't it? Isn't the point to be able to toggle it on and off? 

Once this global variable gets initialized, it won't be changed for the runtime of the program (except by some threads). If you want to have this be changeable at runtime, that logic should go into the `_ent_factory` method

snuba/datasets/entities/generic_metrics.py
```diff
@@ -110,7 +110,7 @@ def get_query_processors(self) -> Sequence[QueryProcessor]:
             TagsTypeTransformer(),
             MappedGranularityProcessor(
                 accepted_granularities=PERFORMANCE_GRANULARITIES,
-                default_granularity_enum=DEFAULT_MAPPED_GRANULARITY_ENUM,
+                default_granularity=DEFAULT_MAPPED_GRANULARITY_ENUM,
```
volokluev, line 113: any reason for this change?

snuba/query/processors/granularity_processor.py
```diff
@@ -48,11 +48,11 @@ class GranularityMapping(NamedTuple):
...
-PERFORMANCE_GRANULARITIES: Sequence[GranularityMapping] = [
-    GranularityMapping(60, 1),
-    GranularityMapping(3600, 2),
-    GranularityMapping(86400, 3),
-]
+PERFORMANCE_GRANULARITIES: Mapping[int, int] = {
+    60: 1,
+    3600: 2,
+    86400: 3,
+}
```
volokluev, line 55: this `int, int` mapping makes the object hierarchy simpler but as a reader of the code, it's not clear to me what `PERFORMANCE_GRANULARITIES` are

snuba/datasets/configuration/entity_builder.py
onewland: We aren't trying to complete the mapping -- we want to use the registered class pattern to obsolete this

snuba/datasets/entities/factory.py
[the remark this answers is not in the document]
onewland: I don't have any intention of changing back and forth. My thought is just that self-hosted users can't get burned if we only do this in prod (and they may never have use for pluggable entities anyway)

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
```diff
@@ -0,0 +1,102 @@
...
+    -
+      mapper: simple_func
+      args:
+        from_name: uniqIf
+        to_name: uniqCombined64MergeIf
+  subscriptables:
+    -
+      mapper: subscriptable
+      args:
+        from_column_table:
```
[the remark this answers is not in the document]
onewland: that's how it's used in the Pythonic construction:

https://github.com/getsentry/snuba/blob/7abbe01bc220e157738e875a9ee9c1610d18ea6a/snuba/datasets/entities/generic_metrics.py#L83-L89

I think it's not a bad idea as all arguments are required and we're not allowing positional arguments here

snuba/datasets/configuration/entity_builder.py
onewland: I actually don't see the functional difference between my approach and yours, assuming that dict.update is atomic.

In my approach:
```
t=0, Thread T1 sees _QP_MAPPING with length 0
t=1, Thread T2 sees _QP_MAPPING with length 0
t=2, Thread T2 finishes importing
t=3, Thread T1 finishes importing
t=4, Thread T2 calls _QP_MAPPING.update(constants)
t=5, Thread T1 calls _QP_MAPPING.update(constants)
```
extra update, no data inconsistency

Your approach:
```
t=0, Thread T1 sees _QP_MAPPING != None
t=1, Thread T2 sees _QP_MAPPING != None
t=2, Thread T2 finishes importing
t=3, Thread T1 finishes importing
t=4, Thread T2 sets _QP_MAPPING = constants
t=5, Thread T1 sets _QP_MAPPING = constants
t=6 or 7, Threads T1 and T2 should return _QP_MAPPING with the same contents
```

I think the issue with my code before is that within that `# initialize here` section of yours I was re-checking the original condition and the value changed and I mishandled it.

Also, we hope to remove this code (or at least move out of this file into the metaclass/QueryProcessor definition) when we go with the registered class pattern

snuba/query/processors/granularity_processor.py
onewland: That's fair but when the move to YAML is complete this constant will be deleted.

snuba/datasets/entities/generic_metrics.py
onewland: I thought it made the YAML definition look cleaner

snuba/cli/entities.py
onewland: added

snuba/datasets/configuration/entity_builder.py
volokluev: I don't think that `_QP_MAPPING.update` is atomic (I could be wrong). So you could see `_QP_MAPPING` be partially updated but not fully.

volokluev: There are other benefits to this approach as well. In a test, you can mock the return value of the function to return what you want in that instance of the test. It's also a good practice in general to hide accesses of a global variable behind a function.

snuba/datasets/configuration/entity_builder.py
volokluev: > we want to use the registered class pattern to obsolete this

Oh I understand that,I just want there to be a #TODO here to make sure I haven't forgotten something

snuba/query/processors/granularity_processor.py
volokluev: > Would a doc comment on the constructor help?

Yeah that would be good, just so as the reader I can figure out what that dict is for. The constructor is not going away either way

snuba/datasets/configuration/entity_builder.py
untitaker: i have absolutely no stake in this but this particular thread caught my eye. `dict.update` is holding the GIL continuously (= "is atomic") for as long as the dictionary keys are primitive types and their `__hash__` is written in C. This is the case for `str`, so this case is fine.

_If_ you happen to update with a dictionary key whose `__hash__` is written in _Python_, you will release the GIL and could probably maybe observe inconsistent state. I think in this case it would mean that one thread will observe its dictionary being mutated, updated with values that are equal with the previous ones, but have a different `id()`.

A variable assignment is guaranteed to be atomic no matter what the variable type is.

volokluev (approved):

snuba/datasets/configuration/entity_builder.py
onewland: added

snuba/query/processors/granularity_processor.py
onewland: added

PR #3096, opened 2022-08-30
ref(MDC): StorageKey enum to class

rahul-kumar-saini:
StorageKey enum replaced with a class.

Based on a working solution for MigrationGroup here: 
#3071 

Relevant file changes:
- `storage_builder.py`
- `__init__.py`
- `storage_key.py`
- `test_storages.py`

The rest of the files changed are simple import updates across the repo.


volokluev:
tests?

snuba/datasets/storages/__init__.py
```diff
@@ -1,44 +1,73 @@
-from enum import Enum
 from functools import lru_cache
+from glob import glob
+from typing import Any
 
+from yaml import safe_load
```
onewland, line 5: I don't think we should directly load YAML outside of the `build_storage` code path

rahul-kumar-saini: I agree, but the `storage_builder` file imports the Enum from this file so it creates a circular import. Either all of the storage building code is moved into this init file or the `StorageKey` definition is moved into that file. I didn't think either way was very clean so this is what I was left with.

I'm open to any other suggestions you may have.

rahul-kumar-saini: Alright I've moved the loading out into `loader.py` since `built_storage` is in a file that imports the `StorageKey` enum which would cause a circular import

snuba/datasets/configuration/loader.py
```diff
@@ -19,3 +22,16 @@ def load_configuration_data(
...
+
+
+def load_config_built_storage_keys() -> dict[str, str]:
+    """
+    Loads all storage keys from configuration files
+    """
+    return {
+        key.upper(): key
+        for key in [
+            safe_load(open(config_file))["storage"]["key"]
```
onewland, line 34: I'm still not a fan of this, I think there should be one single entrypoint for any loading of any storage configuration file.

evanh: The key dictionary could be moved into the same place as the storage factory. Then it could be populated at the same time the storages are loaded from the config.

onewland: Yeah, that sounds like a good approach to me.

rahul-kumar-saini: I don't think we can move this dictionary out into the Storage factory file and then import it in the init file where the enum exists. This is the first thing I tried and it caused a circular import since the storage factory file imports the enum itself.

onewland (approved):
Looks good to me, a couple thoughts:
1. probably one other person should approve because this touches so many files now
2. make sure you can run all the `snuba/cli` commands, especially devserver and consumers, since the tests often don't check that things will get initialized correctly in that flow

snuba/datasets/storages/storage_key.py
```diff
@@ -0,0 +1,81 @@
...
+    "REPLAYS": "replays",
+    "ERRORS_V2_RO": "errors_v2_ro",
+}
+
+REGISTERED_STORAGE_KEYS: dict[str, str] = {}
+
+
+class _StorageKey(type):
+    def __getattr__(cls, attr: str) -> "StorageKey":
+        if attr not in HARDCODED_STORAGE_KEYS and attr not in REGISTERED_STORAGE_KEYS:
```
evanh, line 42: The next step of this needs to be locking down the dependency injection of things. It feels like we have all these disparate modules with a mix of things built lazily and on import that all happen to be initialized in the right order. Feels like it's pretty brittle right now.

evanh (approved):

PR #3098, opened 2022-08-30
ref(MDC): consolidate query processors into one folder

volokluev:
In order to be able to auto-import the query processors and have them be accessible by config, they can't be scattered all over the codebase. Find the rogue processors and put them under

`snuba/query/processors/(physical|logical)`

#### This PR  contains no logical changes, only file renames and import renames


nikhars (approved):

volokluev:
@evanh 

> I think a more subtle approach is needed here. The storage processors are separate because they operate on a different class from the "query" processors. One operates on the logical query and one on the physical query.

Wow that's very confusing. Especially given they're the same class and take what looks like the same arguments. What if the folder structure was 

`snuba.query.processors.logical`
`snuba.query.processors.physical`

onewland:
> @evanh
> 
> > I think a more subtle approach is needed here. The storage processors are separate because they operate on a different class from the "query" processors. One operates on the logical query and one on the physical query.
> 
> Wow that's very confusing. Especially given they're the same class and take what looks like the same arguments. What if the folder structure was
> 
> `snuba.query.processors.logical` `snuba.query.processors.physical`

I think that distinction makes sense

evanh:
> What if the folder structure was...

Yep that's fine with me. Maybe we should make the base `QueryProcessor` class names different as well? `LogicalQueryProcessor` and `ClickhouseQueryProcessor`?

rahul-kumar-saini (approved):

PR #3099, opened 2022-08-30
feat(mdc): Make translators, validators, and mappers registered classes

volokluev:
This PR removes 3 of the hacky maps from the `entity_builder` and replaces them with the `RegisteredClass` metaclass which allows any imported class to be looked up by name

This PR does not do auto-importing, but these classes are imported as part of runtime. I can either do auto-importing in this PR or a different one


snuba/datasets/configuration/entity_builder.py
```diff
@@ -30,10 +26,7 @@
 from snuba.query.processors.quota_processor import ResourceQuotaProcessor
 from snuba.query.processors.tags_type_transformer import TagsTypeTransformer
 from snuba.query.processors.timeseries_processor import TimeSeriesProcessor
-from snuba.query.validation.validators import (
-    EntityRequiredColumnValidator,
-    QueryValidator,
```
onewland, line 35: If you're not auto-importing, how are these getting registered?

I'm wondering if this is only succeeding because other entities are loading the validators. What happens if you run `snuba entities load snuba/datasets/configuration/generic_metrics/entities/sets.yaml`?

onewland (dismissed):
I'm directionally good with this if the test failures are flukes -- did you want to do the safe-ish factory pattern later? e.g. the `QueryValidatorSubclass.construct_from_name(name, kwargs)` that validates kwargs before calling the existing `__init__` on QueryValidatorSubclass?

snuba/datasets/configuration/entity_builder.py
onewland: Also, does `pytest tests/datasets/configuration/test_entity_loader.py` on its own work? That loads the python-created entity *after* using the entity builder, but if you use the entire suite of tests then the other classes might get loaded and make these imports unnecessary.

onewland: never mind, I cloned your branch locally and I now see that the validators/mappers we're using are already part of the base import path, which is probably why you punted on QueryProcessors.

evanh (approved):
I'm OK with this approach, pending tests passing.

snuba/utils/registered_class.py
```diff
@@ -0,0 +1,70 @@
...
+    """Keep a mapping of classes to their names"""
+
+    def __init__(self) -> None:
+        self.__mapping: Dict[str, RegisteredClass] = {}
+
+    def register_class(self, cls: "RegisteredClass") -> None:
+        key = cls.config_key()
+        existing_class = self.__mapping.get(key)
+        if not existing_class:
+            self.__mapping[key] = cls
```
onewland, line 23: I actually think we should fail hard if a mapping already exists, I could see this being very confusing

onewland (approved):
looks good pending mypy fixes

PR #3100, opened 2022-08-31
ref: use internal pypi for prebuilt packages

asottile-sentry:
this changes from using public pypi to using our "internal" (but publicly accessible) pypi

the main advantage is on the 4 supported platforms (linux, macos) x (x86_64, arm64) all python dependencies are prebuilt and will not need to be built from source

pypi packages are sourced from https://github.com/getsentry/pypi

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,10 @@
+--index-url https://pypi.devinfra.sentry.io/simple
+
 black==22.6.0
 blinker==1.5
 click==8.1.3
 clickhouse-driver==0.2.4
-confluent-kafka==1.8.2
+confluent-kafka==1.9.2
 datadog==0.21.0
 flake8==5.0.4
 Flask==2.2.2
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,8 +1,10 @@
+--index-url https://pypi.devinfra.sentry.io/simple
+
 black==22.6.0
 blinker==1.5
 click==8.1.3
 clickhouse-driver==0.2.4
-confluent-kafka==1.8.2
```
asottile-sentry, line 5: I had to additionally update confluent-kafka and arroyo because I didn't build confluent-kafka<1.9.2 for pypi.  previous versions have various trouble building on the 4 supported platform/arch 2-ples

evanh (approved):
Please ping the sns on-call when you deploy this, just so we can watch the kafka changes.

asottile-sentry:
> Please ping the sns on-call when you deploy this, just so we can watch the kafka changes.

sounds good -- I plan to do this after the holidays !

PR #3102, opened 2022-08-31
feat(mdc) add configuration for generic metrics distributions

onewland:
add the generic_metrics_distributions entity.

Tested locally by:
- running generic_metrics unit tests with `settings.PREFER_PLUGGABLE_ENTITIES` set to true
- used the `snuba entities load` command with a breakpoint and inspected the output to make sure the mappers were there


onewland:
mypy failing is from the parent (non-master) branch

evanh (approved):

PR #3104, opened 2022-08-31
feat(discover): add group_ids to discover

udameli:
This PR is meant to enable querying transactions by `group_ids` in Discover, more in https://github.com/getsentry/sentry/pull/38024

The goal is to use discover endpoints to get affected transaction events for performance issues


evanh (commented):
I would add a test in `test_snql_api.py` that inserts some data, queries for the column and validates the result.

snuba/datasets/entities/discover.py
```diff
@@ -284,6 +284,7 @@ def attempt_map(
                 ]
             ),
         ),
+        ("group_ids", Array(UInt(64, Modifiers(nullable=True)))),
```
nikhars, line 287: This might not be a concern today, but something to think about for the long term. If we are planning to make issues as a platform, I would imagine that there would be multiple datasets the discover queries would like to be able to query in the future (like profiles, replays). In that case, would it be better if the column names are all prefixed with the dataset name? Something like `transactions_group_ids`?

wedamija: I don't think we would want to differentiate there. We'd want to treat `group_ids` the same across all the datasets.

But tbh I have no idea what these additional issues will look like and whether we will be adding them to discover. Probably too early to plan for that.

evanh (approved):
Looks good, thanks.

PR #3106, opened 2022-08-31
Create a Snuba dataset task

ayirr7:


snuba/migrations/snuba_migrations/audit_log/0001_auditlog.py
```diff
@@ -0,0 +1,60 @@
...
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.AUDIT_LOG,
+                table_name="audit_log_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    storage_set=StorageSetKey.AUDIT_LOG,
+                    order_by="timestamp",
+                    partition_by="user",
```
lynnagara, line 29: Any particular reason for choosing user as the partitioning key?

snuba/datasets/entities/audit_log.py
```diff
@@ -0,0 +1,45 @@
...
+        Column("event_id", UUID()),
+        Column("timestamp", DateTime()),
+        Column("event_type", String()),
+        Column("user", String()),
+        Column("details", String()),
+    ]
+)
+
+
+class AuditLogEntity(Entity, ABC):
```
lynnagara, line 24: I don't think this one needs to be an abstract base class, unless you planning to subclass it?

lynnagara:
If you rebase/merge master you might notice that the `snuba_migrations` folder moved. Sorry about that - we are in the process of trying to split out the migrations framework part into a separate module that will eventually move out of Snuba.

ayirr7:
Onboarding task completed

PR #3107, opened 2022-09-01
feat(mdc): Load entity subscriptions from config

enochtangg:
This PR is responsible for adding support for loading entity subscriptions from YAML.

Changes:
1. Defines `generic_metrics_sets_subscription` and `generic_metrics_distributions_subscription` in YAML
2. Added JSON schema for validating entity subscription YAML configs
3. Created `PluggableEntitySubscription`, loader, and mappings required to convert config to code
4. Added `_EntitySubscriptionFactory` which houses both Python and config defined entity subscriptions
5. Test cases for running entity subscription from config

Note:
One of the main differences between defining entities from config vs. entity subscription from config is that entity subscription classes require a parent class (`BaseEventsSubscription`, `SessionsSubscription`) in python. As discussed, we will initially only support the `SessionsSubscription` base class for now. Eventually down the road, we can add functionality for other classes.


evanh:
> entitiy_subscriptions directory was created in snuba/subscriptions/ which contains all the configuration, loaders, yamls, and etc. needed to run entity subscriptions from config.

This isn't correct. All the parsing files should live in `datasets/configuration/` and the YAML files should be in a sub folder of `datasets/configuration/generic_metrics`.

snuba/subscriptions/entity_subscriptions/factory.py
```diff
@@ -0,0 +1,118 @@
...
+            EventsSubscription,
+            GenericMetricsDistributionsSubscription,
+            GenericMetricsSetsSubscription,
+            MetricsCountersSubscription,
+            MetricsSetsSubscription,
+            TransactionsSubscription,
+        )
+
+        entity_subscription_to_config_path_mapping: Mapping[EntityKey, str] = {
+            EntityKey.GENERIC_METRICS_SETS: "snuba/subscriptions/entity_subscriptions/generic_metrics/sets.yaml",
```
evanh, line 37: This dictionary should be built automatically, see how the storages are loaded automatically: https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/factory.py#L97

snuba/subscriptions/entity_subscriptions/generic_metrics/distributions.yaml
```diff
@@ -0,0 +1,7 @@
+version: v1
+kind: entity_subscription
+name: generic_metrics_distributions_subscription
+
+parent_subscription_entity_class: sessions_subscription
```
onewland, line 5: I'm pretty against specifying parent classes here -- I'd much rather duplicate information than have this thing support inheritance

onewland:
Is there a reason that information about how subscriptions work can't be stored directly on the entity itself?

It seems at least at the YAML level that should be a 1:1 relationship (but I could be wrong about that)

snuba/subscriptions/entity_subscriptions/configuration/entity_subscription_builder.py
```diff
@@ -0,0 +1,54 @@
...
+    ]
+    PluggableEntitySubscription.disallowed_aggregations = config_data[
+        "disallowed_aggregations"
+    ]
+
+    # Plug parent class dynamically from config
+    NewPluggableEntitySubscription = type(
+        "PluggableEntitySubscription",
+        (
+            _PARENT_SUBSCRIPTION_CLASS_MAPPING[
```
evanh, line 45: I think you can simplify all this by assuming `PluggableEntitySubscription` always extends `SessionsSubscription` for now.

We can revisit this once we migrate datasets that need the other class as well. I would leave a comment somewhere though that documents that decision.

snuba/subscriptions/entity_subscriptions/configuration/entity_subscription_builder.py
```diff
@@ -0,0 +1,54 @@
...
+    PluggableEntitySubscription.name = config_data["name"]
+    PluggableEntitySubscription.MAX_ALLOWED_AGGREGATIONS = config_data[
+        "max_allowed_aggregations"
+    ]
+    PluggableEntitySubscription.disallowed_aggregations = config_data[
+        "disallowed_aggregations"
+    ]
+
+    # Plug parent class dynamically from config
+    NewPluggableEntitySubscription = type(
```
evanh, line 42: I am just re-realizing now how terrible EntitySubscriptions are structured. Needing to embed this information in the actual class instead of instances was a foolish choice.

snuba/subscriptions/entity_subscriptions/generic_metrics/distributions.yaml
enochtangg: Changed PluggableEntitySubscription to only extend from SessionSubscription as mentioned in comment below. We can revisit this once we migrate datasets that require a different base class.

snuba/subscriptions/entity_subscriptions/factory.py
enochtangg: This might require some class (e.g. https://github.com/getsentry/snuba/pull/3109/) or relationship between (EntitySubscriptions and Entities) to map automatically imported file paths and EntityKeys. Perhaps I could create a ticket for this to discuss storing EntitySubscriptions onto Entities directly?

snuba/utils/loader.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+
+from typing import Any, Mapping
+
+from jsonschema import validate
+from yaml import safe_load
+
+
+def load_configuration_data(
```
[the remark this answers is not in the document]
enochtangg: Resolved.

tests/subscriptions/entity_subscriptions/test_pluggable_entity_subscription.py
```diff
@@ -0,0 +1,113 @@
...
+
+    sets_identifier = SubscriptionCreator(dataset, entity_key).create(
+        sets_subscription, timer
+    )
+    print(f"Created {sets_identifier}")
+
+    pluggable_identifier = SubscriptionCreator(dataset, entity_key).create(
+        pluggable_subscription, timer
+    )
+    print(f"Created {pluggable_identifier}")
```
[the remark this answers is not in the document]
enochtangg: Resolved.

snuba/subscriptions/entity_subscriptions/configuration/entity_subscription_builder.py
enochtangg: Resolved.

snuba/subscriptions/entity_subscriptions/configuration/entity_subscription_builder.py
enochtangg: See reply to other comment above.

snuba/subscriptions/entity_subscriptions/factory.py
evanh: Sure, that's fine.

snuba/subscriptions/entity_subscriptions/configuration/entity_subscription_builder.py
evanh: Yes please do.

evanh (approved):
This is an good migration of a bad feature. Let's make sure we have a ticket to decide how to kill this feature in the future.

PR #3108, opened 2022-09-01
feat(MDC): Generic Metrics Dataset loaded from Config

rahul-kumar-saini:
### Overview
- Generic Metrics Dataset is loaded from a config file
- Added minimal `PluggableDataset` which can be extended as we config more complicated datasets
- Added validation for Dataset config
- New dataset configs will be automatically picked up

### Notes
- Any dataset config file should be named `dataset.yaml` and placed as follows to be automatically picked up
    - `snuba/datasets/configuration/<dataset_name>/dataset.yaml`

### DIFF
--- a/snuba/datasets/configuration/json_schema.py
+++ b/snuba/datasets/configuration/json_schema.py
@@ -121,8 +121,6 @@ STORAGE_SCHEMA = {
 
 STORAGE_QUERY_PROCESSORS_SCHEMA = TYPE_STRING_ARRAY
 
-KIND_SCHEMA = {"enum": ["writable_storage", "readable_storage", "entity"]}
-
 ENTITY_QUERY_PROCESSOR = {
     "type": "object",
     "properties": {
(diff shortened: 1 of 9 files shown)


snuba/datasets/configuration/json_schema.py
```diff
@@ -213,3 +214,17 @@ def make_column_schema(
         "required_time_column",
     ],
 }
+
+V1_DATASET_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "version": {"const": "v1"},
+        "kind": {"const": "dataset"},
```
evanh, line 222: Shouldn't this go in `KIND_SCHEMA`?

evanh (approved):

snuba/datasets/configuration/json_schema.py
rahul-kumar-saini: Removed `KIND_SCHEMA` entirely since we don't want a dataset to be able to have `entity` as a kind etc. Each kind should be a const as it is now.

PR #3109, opened 2022-09-01
ref(MDC): EntityKey enum to class

rahul-kumar-saini:
### Overview
- Replace `EntityKey` enum with a simple class which will be slowly phased out as more entities are made into config

#### Most of this PR consists of updated import statements:
`from snuba.datasets.entities import EntityKey` -> `from snuba.datasets.entities.entity_key import EntityKey`

#### Relevant file changes to review are the following:
- [ ] `snuba/datasets/configuration/entity_builder.py`
    - Entity Keys are registered as they are loaded from config
- [ ] `snuba/datasets/entities/__init__.py`
    - `EntityKey` enum removed
- [ ] `snuba/datasets/entities/entity_key.py`
    - New `EntityKey` class created
- [ ] `snuba/datasets/entities/factory.py`
    - Updated factory to load entities automatically instead of from hardcoded file paths
- [ ] `snuba/datasets/pluggable_entity.py`
    - Replaced attribute `name` with `entity_key`
- [ ] `snuba/datasets/storages/storage_key.py`
    - Added iterator in case StorageKeys need iterating
- [ ] `snuba/settings/__init__.py`
    - Added file path glob for auto loading entity configs
- [ ] `snuba/subscriptions/entity_subscription.py`
    - Call to init entity factory
- [ ] `tests/datasets/entities/test_entity_key.py`


snuba/subscriptions/entity_subscription.py
```diff
@@ -14,6 +14,8 @@
     SubscriptionAllowedClausesValidator,
 )
 
+initialize_entity_factory()
```
rahul-kumar-saini, line 17: Hack?: `entity_subscription.py` stuff requires all entities to be loaded so I'm not sure where else to initialize. Tests fail if entity factory isn't initialized otherwise.

snuba/datasets/entities/factory.py
```diff
@@ -19,6 +20,17 @@ def __init__(self) -> None:
...
+
+        self._config_built_entities = {
+            entity.entity_key: entity
+            for entity in [
+                build_entity_from_config(config_file)
+                for config_file in glob(
+                    settings.ENTITY_CONFIG_FILES_GLOB, recursive=True
+                )
+            ]
+        }
```
onewland, line 32: thanks for handling this piece

snuba/subscriptions/entity_subscription.py
onewland: I'd like to see this file mostly go away -- these just seem like properties that should live on Entity to me

tests/datasets/entities/test_entity_key.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def test_entity_key() -> None:
+    reset_entity_factory()
+    with pytest.raises(AttributeError):
+        EntityKey.NON_EXISTENT_ENTITY
+    assert REGISTERED_ENTITY_KEYS == {
+        "GENERIC_METRICS_DISTRIBUTIONS": "generic_metrics_distributions",
+        "GENERIC_METRICS_SETS": "generic_metrics_sets",
+    }
```
onewland, line 14: I don't love that this is going to require changing for every entity added to `snuba/datasets/configuration/`. Not sure of a better way to test this behavior but I think that we should not assume the entity list will be so static

onewland (commented):
Looks pretty good to me, did we test running subscription schedulers + executors + other CLI things that depend on entities? Sometimes that can catch errors

snuba/subscriptions/entity_subscription.py
volokluev: Yes this is a hack. The thing that it supports is also a hack. Put the two hacks near each other and encase it in a comment block that explains the hack and file a ticket to remove it

https://github.com/getsentry/snuba/blob/53fadadab46b75aa0bf456c5e1d256eeb962fca3/snuba/subscriptions/entity_subscription.py#L146-L157

tests/datasets/entities/test_entity_key.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from snuba.datasets.entities.entity_key import REGISTERED_ENTITY_KEYS, EntityKey
+from snuba.datasets.entities.factory import reset_entity_factory
+
+
+def test_entity_key() -> None:
+    reset_entity_factory()
```
volokluev, line 8: Does this function only exist for testing? I can't imagine any other reason why it would need to.

If you need a blank entity factory for a test, just mock the `_ent_factory` function to return what you want

https://github.com/getsentry/snuba/blob/53fadadab46b75aa0bf456c5e1d256eeb962fca3/snuba/datasets/entities/factory.py#L120

volokluev: this reset method shouldn't exist because it invites mutability into global state, which we should not have.

volokluev (commented):
some questions

tests/datasets/entities/test_entity_key.py
rahul-kumar-saini: I'll remove it, this was introduced as part of #3065 and I just used it here

enochtangg: This was introduced purely for testing the multithreaded race condition. As opposed to spawning new threads and accessing the entity factory in a test case, the test was revised to ensure the entity factory initialization and the entity mapping attribute creation are not decoupled.

volokluev: (off-topic for this PR)

It's best to keep testing functions within tests especially for things like this which work with global state. Having a function that alters global state, should not be used in normal operation of the system, in the public api of a module introduces more surface area for issues.

This is a usecase for the python mocking library which can inject these kinds of corner conditions into the runtime of the test but without changing the way the system operates

snuba/settings/__init__.py
```diff
@@ -245,6 +245,7 @@
 
 # File path glob for configs
 STORAGE_CONFIG_FILES_GLOB = f"{CONFIG_FILES_PATH}/**/storages/*.yaml"
+ENTITY_CONFIG_FILES_GLOB = f"{CONFIG_FILES_PATH}/**/entities/*.yaml"
```
evanh, line 248: We're going to need a better way of doing this I think. I feel like this works now because there is only one folder in the config file path.

tests/datasets/entities/test_entity_key.py
evanh: We will definitely need to figure this out, because it also implies that if a product person wants to add a new entity they will also have to edit this test. The point of the configuration is that changing them doesn't require Python changes.

evanh (approved):

PR #3121, opened 2022-09-06
fix(grouping): Group clickhouse errors by referrer

untitaker:
Discussed with evan, we want to split up ClickhouseError into smaller
groups that can be ignored individually.

Alternatively we can just start tagging all of those properties, and
use server-side fingerprinting rules (in project settings) to convert
those tags into fingerprints. The advantage of that would be that you
can iterate on grouping much more quickly, but the downside is that
changes to those rules don't live in code and don't undergo a review
process.


snuba/web/db_query.py
```diff
@@ -717,7 +717,12 @@ def raw_query(
             with configure_scope() as scope:
                 if isinstance(cause, ClickhouseError):
                     error_code = cause.code
-                    scope.fingerprint = ["{{default}}", str(cause.code)]
+                    scope.fingerprint = [
+                        "{{default}}",
+                        str(cause.code),
+                        query_metadata.dataset,
+                        query_metadata.request.referrer,
```
fpacifici, line 724: Why do we want to break down by referrer ? 
This will create a very large amount of groups every time we have a system issue. I would advise against.

untitaker: @fpacifici i discussed this with @evanh, but my understanding is that the current clickhouse error is hard to monitor because it conflates all the datasets

"referrer" is imo a good approximation of "where did this issue come from", a piece of information typically provided by a stacktrace

there are not that many _actually observed_ `(dataset, referrer)` combinations: https://sentry.io/organizations/sentry/issues/3111024867/events/9498896d977f4c2abc70a0edc9c705da/tags/transaction/?project=300688

evanh: I think this addresses the more common case, which is intermittent Clickhouse errors. It is true that a system wide problem would create many issues, but that is a rarer case. Also this applies only to errors  with a Clickhouse code, which is a smaller slice of potential problems.

evanh (approved):

snuba/web/db_query.py
untitaker: what is the expected cardinality of referrer? if it is less than the number of possible stacktraces, our system really should be able to handle it...

untitaker: >Also this applies only to errors with a Clickhouse code, which is a smaller slice of potential problems.

I don't see any branching specific to existence of code, I think all clickhouse errors have one, no?

fpacifici: I don't think this is a rare case.

The most common error we generate is the project rate limitier error. The number of issues will multiply. by the number of referrer (hundreds). 
I would still strongly suggest using a tag for this and not a new issue.

fpacifici: > I think this addresses the more common case, which is intermittent Clickhouse errors. It is true that a system wide problem would create many issues, but that is a rarer case.

Not following. While the larger issues are less common, having them properly grouped allows a proper investigation  when they happen. If we generate hundreds of issues because a clickhouse node decided it was receiving too many concurrent queries or stopped responding on the network, even counting the occurrences will be very hard. 

Again, this seems the right problem to be solved with a tag, which provides the same querying ability without breaking grouping.

evanh: The problem we are trying to solve is the day to day work of triaging and fixing bugs in the Sentry dashboard. A tag will not help that problem. This change will make 99% of the investigations in Sentry easier. I don't know why we would optimize for the 1%, especially because I am skeptical this change will actually _hinder_ an investigation into a catastrophic incident.

evanh: > I don't see any branching specific to existence of code

I was referring to the implicit branching. We have many errors that happen before sending the query to Clickhouse, so there is no error code.

untitaker: is it possible to determine the clickhouse error codes that indicate systematic failures and not problems that pertain to a specific query?

evanh: We can (and have) isolated some of the error codes in this way, but the problem is that some error codes apply to both cases. For example `TIMEOUT_EXCEEDED = 159` can mean a single query is taking too long, but if it is happening across many queries at the same time it can mean Clickhouse itself is in a bad state.

volokluev:
@untitaker do you want to go ahead and merge this?

snuba/web/db_query.py
fpacifici: > Maybe a better compromise would be to pick the few error codes that are signals of general Clickhouse degradation (timeouts etc.) and not split those.

As discussed offline this still has to be done.

fpacifici (changes requested):
As per discussion offline we still need to compile the list of errors to break down/not break down

evanh:
This fell of my radar, sorry about that. This is the list of error codes that I think are not referrer specific: https://gist.github.com/evanh/6654674490f652b2c8b1de5ff944e21c

This is the result of me looking through the error codes found here: https://github.com/mymarilyn/clickhouse-driver/blob/master/clickhouse_driver/errors.py

untitaker:
@fpacifici @evanh this PR now groups clickhouse errors by referrer, except those defined in a certain list. ptal

fpacifici (approved):

PR #3124, opened 2022-09-07
feat(transactions): Add app_start_type migration

philipphofmann:
Adds the column app_start_type to transactions. For more context,
see https://github.com/getsentry/sentry/issues/36927.

### DIFF
--- a/snuba/datasets/storages/transactions_common.py
+++ b/snuba/datasets/storages/transactions_common.py
@@ -101,7 +101,7 @@ columns = ColumnSet(
         ("transaction_source", String()),
         ("timestamp", DateTime(Modifiers(readonly=True))),
         ("group_ids", Array(UInt(64))),
-        ("app_start_type", String(Modifiers(readonly=True))),
+        ("app_start_type", String()),
     ]
 )
 
--- a/snuba/datasets/storages/transactions_common.py
+++ b/snuba/datasets/storages/transactions_common.py
@@ -101,6 +101,7 @@ columns = ColumnSet(
         ("transaction_source", String()),
         ("timestamp", DateTime(Modifiers(readonly=True))),
         ("group_ids", Array(UInt(64))),
+        ("app_start_type", String(Modifiers(readonly=True))),
     ]
 )
 
--- a/snuba/snuba_migrations/transactions/0017_transactions_add_app_start_type_column.py
+++ b/snuba/snuba_migrations/transactions/0017_transactions_add_app_start_type_column.py
@@ -1,0 +1,51 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, String
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    The app start type column is required to query for different
+    app start types.
+    """
+
+    blocking = False
+
+    def __forward_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name=table_name,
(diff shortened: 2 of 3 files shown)


philipphofmann:
@evanh, are there any further actions required to make that searchable in Discover?

snuba/datasets/transactions_processor.py
```diff
@@ -248,6 +248,8 @@ def _process_contexts_and_user(
             if context in contexts:
                 del contexts[context]
 
+        processed["app_start_type"] = contexts["app"]["start_type"]
```
evanh, line 251: Is this field guaranteed to always exist?

snuba/datasets/entities/discover.py
```diff
@@ -284,6 +284,7 @@ def attempt_map(
                 ]
             ),
         ),
+        ("app_start_type", String(Modifiers(nullable=True))),
```
evanh, line 287: I would add/modify an E2E test in `test_snql_api.py` or `test_snql_sdk_api.py` that ensures this column can be read correctly.

evanh (commented):
Please separate the migration change from the other changes. Currently migrations in Snuba are run manually, so if this is merged in the current state inserts will start failing because the column doesn't exist on the table. The steps for adding a new column are:

1) Open a migration PR and a second PR for the reading/writing changes.
2) Merge the migration PR
3) Someone from SnS will manually run the migration
4) Merge the second PR.

> are any further actions required to make that searchable in Discover?

The column also needs to be exposed in Sentry, via https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/events.py

snuba/datasets/transactions_processor.py
philipphofmann: No, definitely not.

philipphofmann:
Thanks for the feedback, @evanh. I put everything together in one PR to be able to get feedback at one glance. I'm going to split the PR now.

snuba/datasets/transactions_processor.py
philipphofmann: Ah, I just learned the difference between `dict[key]` and `dict.get(key)`. I'm new to Python.

snuba/datasets/storages/transactions_common.py
```diff
@@ -101,6 +101,7 @@
         ("transaction_source", String()),
         ("timestamp", DateTime(Modifiers(readonly=True))),
         ("group_ids", Array(UInt(64))),
+        ("app_start_type", String(Modifiers(readonly=True))),
```
nikhars, line 104: Why do you add `readonly=True` modifier to this column? Is this column going to be materialized? If not, I would recommend that you remove `readonly=True` modifier

snuba/snuba_migrations/transactions/0017_transactions_add_app_start_type_column.py
```diff
@@ -0,0 +1,51 @@
...
+    def __forward_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name=table_name,
+                column=Column(
+                    "app_start_type",
+                    String(Modifiers(low_cardinality=True)),
```
nikhars, line 26: Is app_start_type field user generated or machine generated? If it is machine generated and would have mostly a few values then having it as low_cardinality is ok. If it is user generated, what is the likelihood of having high cardinality of this value?

philipphofmann: The SDKs create these values, and it's not exposed to the user. Currently, we have four different values. The likelihood of this having a high cardinality is very low. Maybe it increases to 10 or 20 at some point, but not more than 100 for sure.

snuba/datasets/storages/transactions_common.py
philipphofmann: What exactly do you mean by `materialized`? I am OK with removing it.

PR #3126, opened 2022-09-07
fix(settings): Remove requirement of CDC and events to be on same cluster

nikhars:
In order to swap the mapping of dataset to clusters for events and errors_v2, we would need to get rid of the validation since CDC datasets are not defined in the tiger clusters.


lynnagara:
See the failing tests. I think an alternate approach could be to leave JOINABLE_STORAGE_SETS unchanged and just drop the code that enforces that they ever need to be on the same cluster.

lynnagara (approved):

PR #3128, opened 2022-09-09
feat(MDC): Validate configs in CI

rahul-kumar-saini:
### Overview
- Adds an optional step in CI that validates all `.yaml` files located under the config files directory
- See step "Dataset Config Validation" in some CI runs on this PR to see how it shows up


evanh (commented):
Nice, this looks good! The only thing I would change is to move the validation script into the `snuba` app, so we can do `snuba validate-configs`.

evanh (approved):

enochtangg (commented):
The `print` statements can probably be converted to `click.echo` to follow convention. 

Something I noticed when running `snuba validate-configs`, the `snuba` CLI-command by default imports all modules within `/cli`, which includes the `subscriptions-executor command`. This command runs `get_enabled_dataset_names()` in a command-line argument (https://github.com/getsentry/snuba/blob/2d5817d632f4327302a94619360e87ba6cb92dd2/snuba/cli/subscriptions_executor.py#L27) which runs `load_configuration_data()`. 

Therefore, when a user has an invalid config, and they run `snuba validate-configs`, the validation occurs in the loading of the `subscriptions-executor` module as opposed to `snuba validate-configs` itself.

PR #3135, opened 2022-09-13
feat(slicing): add partition mapping

onewland:
Add a mapping of logical to physical partitions, to be used by storage slicing


docs/source/architecture/partitioning.rst
```diff
@@ -0,0 +1,33 @@
...
+===========================
+Adding a physical partition
+===========================
+
+Add the logical:physical mapping
+--------------------------------
+To add a physical partition to the logical:physical mapping, or reshard, increment the
+value of ``settings.LOCAL_PHYSICAL_SLICES`` and change
+the mapping of relevant partitions in ``settings.DATASET_PARTITION_MAPPING``.
+Every logical partition **must** be assigned to a physical partition and the
```
evanh, line 27: Is the reverse true? Must every physical partition be assigned a logical partition?

onewland: No, I think it's reasonable for a physical partition to be unused.

snuba/settings/__init__.py
```diff
@@ -28,6 +28,16 @@
 # Clickhouse Options
 CLICKHOUSE_MAX_POOL_SIZE = 25
 
+SENTRY_LOGICAL_PARTITIONS = 256
```
nikhars, line 31: Is there any consequence of this value being 256 for ST or Self hosted? I feel like slicing is something we would need only for SaaS, so it might be better to keep this as 1 for defaults and override in our prod settings file.

snuba/settings/__init__.py
```diff
@@ -28,6 +28,16 @@
...
+LOCAL_PHYSICAL_SLICES = 1
```
nikhars, line 32: Can you add comments around the usage of these settings?

docs/source/architecture/partitioning.rst
```diff
@@ -0,0 +1,33 @@
+===========================================
```
nikhars, line 1: Thank you for writing the document

snuba/settings/__init__.py
onewland: The *logical* partitions should probably be 256 (or 512 or whatever we chose) everywhere IMO. They exist to make migration between different numbers of *physical* partitions easier. I set `LOCAL_PHYSICAL_SLICES` to 1 for the reason you state.

Arguably this should be a constant rather than a setting, it should essentially never change. But I wasn't sure where that would belong

snuba/settings/__init__.py
onewland: as I said above, we really shouldn't change `SENTRY_LOGICAL_PARTITIONS` anywhere. I can add a doc for `LOCAL_PHYSICAL_SLICES`, though

ethanhs:
Looks like you need to merge the default branch to fix CI, c.f. https://github.com/getsentry/snuba/pull/3134

snuba/settings/__init__.py
lynnagara: @onewland Agreed, please don't put this in the settings file. It's very important that 256 is constant and never overridden. Settings exists for things that need to be configured by users.

snuba/settings/__init__.py
```diff
@@ -28,6 +28,18 @@
...
+SENTRY_LOGICAL_PARTITIONS = 256
+# The number of physical slices that we will need to map resources to
+# for dataset partitioning
+LOCAL_PHYSICAL_SLICES = 1
+# Mapping of logical (key) to physical (value) partitions for datasets
+# that are partitioned across Sentry
+DATASET_PARTITION_MAPPING: Mapping[str, int] = {
+    str(x): 0 for x in range(0, SENTRY_LOGICAL_PARTITIONS)
+}
+# Storage names to apply dataset partitioning to
```
lynnagara, line 40: Why do datasets need to be partitioned as well? Shouldn't partitioning be defined at the storage layer (i.e. closer to the physical layer)

tests/settings/test_settings.py
```diff
@@ -51,3 +51,26 @@ def test_topics_sync_in_settings_validator() -> None:
...
+
+def test_validation_catches_unmapped_logical_parts() -> None:
+    all_settings = build_settings_dict()
+
+    part_mapping = all_settings["DATASET_PARTITION_MAPPING"]
+    del part_mapping["2"]
+
+    with pytest.raises(AssertionError):
+        validate_settings(all_settings)
+    part_mapping["2"] = 0
```
lynnagara, line 76: Do these tests actually validate anything that isn't covered in validation.py? I think it's more important to have a comprehensive validation there. In that was the case I think we wouldn't need this test on top of that as everything would immediately fail on import if you provided an invalid config.

snuba/settings/__init__.py
onewland: This is just imprecise wording, I can remove the "dataset" wording

snuba/settings/__init__.py
onewland: moved that out of settings, let me know what you think

snuba/datasets/sharding.py
```diff
@@ -0,0 +1,6 @@
+"""
+The number of logical partitions used to distinguish between where records
+should be stored. These do not require individual physical partitions but allow
+for resharding with less code changes per physical change.
+"""
+SENTRY_LOGICAL_PARTITIONS = 256
```
lynnagara, line 6: Thanks for moving this. Just a nit, we use the wort partitioning for this variable, sharding for the module and slicing in settings and elsewhere. I think it would be best to standardise on one of these words in our codebase (and ideally outside of it too).

docs/source/architecture/partitioning.rst
```diff
@@ -0,0 +1,33 @@
...
+Snuba Data Partitioning (under development)
+===========================================
+
+*This feature is under active development and is subject to change*
+
+To support a higher volume of data, we are building out support for
+datasets and storages that span multiple ClickHouse clusters with the same
```
lynnagara, line 8: Just ClickHouse clusters? How about mentioning Kafka clusters and potentially Redises (tbc) and Postgreses (well, that one would be part of Sentry but still...). I was under the impression the main purpose of this common partitioning scheme we are doing is to facilitate a single approach over multiple systems and not just ClickHouse though.

onewland: that's a good call, I can update the description

snuba/datasets/sharding.py
onewland: I'm going to change to partition everywhere

onewland:
I believe I've addressed all comments here if folks want to take another look

snuba/settings/validation.py
```diff
@@ -83,3 +85,13 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
                 # We allow definition of storage_sets in configuration files
                 # that are not defined in StorageSetKey.
                 pass
+
+    for logical_part in range(0, SENTRY_LOGICAL_PARTITIONS):
+        physical_part = locals["LOGICAL_PARTITION_MAPPING"].get(str(logical_part))
+        slice_count = locals["LOCAL_PHYSICAL_PARTITIONS"]
```
lynnagara, line 91:
```suggestion
        partition_count = locals["LOCAL_PHYSICAL_PARTITIONS"]
```

snuba/settings/validation.py
```diff
@@ -83,3 +85,13 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
...
+        assert (
+            physical_part is not None
+        ), f"missing physical partition for logical partition {logical_part}"
+        assert (
+            physical_part < locals["LOCAL_PHYSICAL_PARTITIONS"]
+        ), f"physical partition for logical partition {logical_part} is {physical_part}, but only {slice_count} slices exist"
```
lynnagara, line 97:
```suggestion
        ), f"physical partition for logical partition {logical_part} is {physical_part}, but only {partition_count} partitions exist"
```

docs/source/architecture/partitioning.rst
```diff
@@ -0,0 +1,35 @@
...
+data from consumers to different ClickHouse clusters by overriding the
+StorageSet key that exists in configuration.
+
+===========================
+Adding a physical partition
+===========================
+
+Add the logical:physical mapping
+--------------------------------
+To add a physical partition to the logical:physical mapping, or reshard, increment the
```
lynnagara, line 26:
```suggestion
To add a physical partition to the logical:physical mapping, or repartition, increment the
```

snuba/datasets/partitioning.py
```diff
@@ -0,0 +1,6 @@
+"""
+The number of logical partitions used to distinguish between where records
+should be stored. These do not require individual physical partitions but allow
+for resharding with less code changes per physical change.
```
lynnagara, line 4:
```suggestion
for repartitioning with less code changes per physical change.
```

lynnagara (approved):

PR #3137, opened 2022-09-13
fix(MDC): Validate required fields and that no additional fields are added

rahul-kumar-saini:
- Ensure no unexpected keys are in configs
- Ensure all required fields are marked required
- Make config validation CI step required


snuba/datasets/configuration/json_schema.py
```diff
@@ -247,6 +300,7 @@ def make_column_schema(
         "kind",
         "name",
     ],
+    "additionalProperties": False,
```
evanh, line 303: Can you add this to the migration group schema as well?

evanh (commented):
You should add a test for this as well.

evanh (approved):

PR #3139, opened 2022-09-13
feat(slicing): Remove StorageSetKey enum 

ayirr7:
- Removed the StorageSetKey enum, created a class in `snuba/clusters/storage_set_key.py`
- Closely followed the (very similar) StorageKey and EntityKey enum to class examples
- Modified all relevant import statements yet 
- Files with meaningful changes (not just imports): `snuba/clusters/storage_set_key.py`, `snuba/clusters/storage_sets.py`, `snuba/datasets/configuration/storage_builder.py`


snuba/cli/migrations.py
```diff
@@ -4,7 +4,7 @@
 import click
 
 from snuba.clusters.cluster import CLUSTERS, ClickhouseNodeType
-from snuba.clusters.storage_sets import StorageSetKey
+from snuba.clusters.storage_set_key import StorageSetKey
```
lynnagara, line 7: What's the reason for moving it to a separate file? The two modules still seems pretty closely related to me that it makes sense to look at together, and half of this diff is updating the imports.

snuba/clusters/storage_set_key.py
```diff
@@ -0,0 +1,80 @@
...
+    "ERRORS_V2_RO": "errors_v2_ro",
+    "PROFILES": "profiles",
+    "FUNCTIONS": "functions",
+    "REPLAYS": "replays",
+    "GENERIC_METRICS_SETS": "generic_metrics_sets",
+    "GENERIC_METRICS_DISTRIBUTIONS": "generic_metrics_distributions",
+}
+
+
+REGISTERED_STORAGE_SET_KEYS: Dict[str, str] = {}
```
lynnagara, line 28: Can we mark this as private? (even if it's just using a leading underscore to convey intent). I am worried about someone accidentally importing this and it being an empty dictionary.

snuba/cli/migrations.py
onewland: I think we had an import problem with EntityKey and the entity factory that forced us to separate the module there. Riya was sort of copying the pattern that Rahul used in https://github.com/getsentry/snuba/pull/3109/files

Given that storage sets really just map to configuration I wonder if we could leave this one in the old module, and just add the registry/type change

lynnagara: Previously we had storages depend on storage sets but never the other way. Storages are very high level abstractions and storage sets are supposed to be low level ones. I am a bit concerned about us introducing dependencies that go both ways now.

lynnagara: I think this may be very problematic for things like migrations that are not supposed to be part of Snuba long term. They do share storage sets (with the assumption that these are low level classes that don't depend on anything else). Now it seems like we may be introducing more dependencies.

lynnagara: I am not sure if this circular dependency actually exists here or we are doing the pattern for another reason?

snuba/datasets/configuration/storage_builder.py
```diff
@@ -59,14 +59,14 @@ def build_storage(
...
-        "storage_set_key": StorageSetKey(config[STORAGE][SET_KEY]),
+        "storage_set_key": register_storage_set_key(config[STORAGE][SET_KEY]),
         SCHEMA: (
             WritableTableSchema if config[KIND] == WRITABLE_STORAGE else TableSchema
         )(
             columns=ColumnSet(parse_columns(config[SCHEMA]["columns"])),
             local_table_name=config[SCHEMA]["local_table_name"],
             dist_table_name=config[SCHEMA]["dist_table_name"],
-            storage_set_key=StorageSetKey(config[STORAGE][SET_KEY]),
+            storage_set_key=register_storage_set_key(config[STORAGE][SET_KEY]),
```
rahul-kumar-saini, line 69: You need only register a key once, this second call to `register_storage_set_key(...)` can be replaced with `StorageSetKey(...)` since the mapping should have the key registered and available to use now

snuba/cli/migrations.py
onewland: > Previously we had storages depend on storage sets but never the other way. Storages are very high level abstractions and storage sets are supposed to be low level ones. I am a bit concerned about us introducing dependencies that go both ways now.

Are we introducing a new direction of dependency here? My goal with this work was to make it so that storage set keys can be dynamic (not hardcoded). Then we can dynamically compute depending on environment and other factors which storage set to query from a given storage (we currently instantiate storages with a hardcoded storage set key value). But I don't think we're introducing a storage set -> storage dependency here.

onewland: Is your concern the `register_storage_set_key()` being called in [snuba/datasets/configuration/storage_builder.py](https://github.com/getsentry/snuba/pull/3139/files/0ce9dbf95722ea97da621b703048a5d881f46e91#diff-00e1a5d3eb99b9b762cd6e92f2d9f91271d0e1f44ea4be71ad2b1c4dffdb23d9)?

lynnagara: If not then great! I thought that's the reason you were saying it got moved into a separate module to avoid the circular dependency.

snuba/clusters/storage_set_key.py
ayirr7: good point, I'll change that

snuba/cli/migrations.py
ayirr7: I moved everything back into one module. So `storage_sets.py` now includes the StorageSetKey class definition

onewland: > If not then great! I thought that's the reason you were saying it got moved into a separate module to avoid the circular dependency.

yeah, it's a little confusing, but I think what was an issue for EntityKey/EntityFactory was not an issue with StorageSetKey and I probably sent Riya down the wrong path a bit.

snuba/clusters/storage_sets.py
```diff
@@ -1,8 +1,54 @@
-from enum import Enum
-from typing import FrozenSet
+from __future__ import annotations
 
+from typing import Any, Dict, FrozenSet, Iterator
 
-class StorageSetKey(Enum):
+HARDCODED_STORAGE_SET_KEYS = {
```
onewland, line 5: when you make `REGISTERED_STORAGE_SET_KEYS` private, let's also do this one

onewland (approved):
looks good to me (pending the changes Lyn and I mentioned about visibility)

snuba/datasets/configuration/storage_builder.py
ayirr7: oops, good catch

lynnagara (approved):

snuba/cli/migrations.py
ayirr7: All good! Fixed now

PR #3141, opened 2022-09-14
feat(mdc): Add pre-commit check for config validation and additionalProperties to schema

enochtangg:
* Added component config validation pre-commit hook. The python file `snuba/validate_configs.py` was used over the `check-jsonchema` CLI because that only supports `.json` files as an input argument. Our schemas are `.py`. Also note, print statements are not forwarded to stdout in pre-commit hook unless the test fails. 
* `additionalProperties` were added to each full schema to ensure that no arbitrary keys are defined in config. Schema will not validate if a user inputs a key that is not present in `properties`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,10 @@ jobs:
           key: cache-epoch-1|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
       - name: Setup pre-commit
         run: make setup-git
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          make install-python-dependencies
       - uses: getsentry/paths-filter@v2
         id: files
         with:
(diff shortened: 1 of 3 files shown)


enochtangg:
> I think you'll need to add a step in the `pre-commit` CI job that installs python dependencies so you can have `jsonschema` available when validating.
> 
> Eg:
> 
> https://github.com/getsentry/snuba/blob/8232af134571238542de8a874bd4d8767bd40ab1/.github/workflows/ci.yml#L90-L93

This worked! 
Why it worked: The `pre-commit hooks` allows a `language: python` flag which manages a new environment for the test to run in. This is typically used in conjunction with `additional_dependencies` to install external packages that might be needed. This however does not include the repo's modules itself. To solve this, a virtual environment is installed onto the runner itself as part of a CI-step.

enochtangg:
Update: Removed `additionalProperties: False` in json_schema.py because that is done in https://github.com/getsentry/snuba/pull/3137

rahul-kumar-saini (approved):

PR #3142, opened 2022-09-14
feat(mdc): Remove default entity from base Dataset class

enochtangg:
Now that all queries sent to Snuba specify an Entity, we want to remove the legacy `default_entity` attribute in a Dataset.

* Removed `default_entity` from base Dataset class and its get method.
* Enforced jsonschema to require configs to specify entities `all` sub-property.
* Added `all_entities` attribute to dataset base class and pass entity keys to child initialization
* Removed all `get_default_entity` occurrences in tests and added support where needed
* Creation of new testing eventstream and insert endpoints to remove `default_entity` and dataset parameter altogether. Plan to re-route to new endpoints are specified here: https://github.com/getsentry/sentry/pull/38877

### DIFF
--- a/bin/mocks/mock-subscriptions
+++ b/bin/mocks/mock-subscriptions
@@ -1,9 +1,9 @@
 #!/usr/bin/env python
 import argparse
 import random
-from datetime import timedelta
 
-from snuba.datasets.entities.factory import get_entity_name
+from snuba.datasets.entities.entity_key import EntityKey
+from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.entity_subscriptions.factory import get_entity_subscription
 from snuba.datasets.factory import get_dataset
 from snuba.redis import redis_client
@@ -31,12 +31,21 @@ parser.add_argument(
     choices=["events"],
     default="events",
 )
+parser.add_argument(
+    "--entity",
+    type=str,
+    help="Entity for subscriptions",
+    dest="entity",
+    choices=["events"],
+    default="events",
+)
 parsed = parser.parse_args()
 
 dataset_name = parsed.dataset
 dataset = get_dataset(dataset_name)
-entity = dataset.get_default_entity()
-entity_key = get_entity_name(entity)
+entity_str = parsed.entity
+entity_key = EntityKey(entity_str)
+entity = get_entity(entity_key)
 storage = entity.get_writable_storage()
 assert storage is not None
 stream_loader = storage.get_table_writer().get_stream_loader()
@@ -52,8 +61,8 @@ for _ in range(parsed.number):
     subscription_data = SubscriptionData(
         query=f"MATCH ({entity_key.value}) SELECT count() AS count WHERE project_id = {project_id}",
         project_id=project_id,
-        time_window=timedelta(minutes=1),
-        resolution=timedelta(minutes=1),
+        time_window_sec=60,
+        resolution_sec=60,
         entity_subscription=entity_subscription,
     )
 
--- a/snuba/datasets/cdc/groupassignee.py
+++ b/snuba/datasets/cdc/groupassignee.py
@@ -1,9 +1,6 @@
-from typing import Sequence
-
 from snuba.datasets.dataset import Dataset
 from snuba.datasets.entities.entity_key import EntityKey
 from snuba.datasets.entities.factory import get_entity
-from snuba.datasets.entity import Entity
 
 
 class GroupAssigneeDataset(Dataset):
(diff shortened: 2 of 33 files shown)


bin/mocks/mock-subscriptions
```diff
@@ -52,8 +61,8 @@ for _ in range(parsed.number):
     subscription_data = SubscriptionData(
         query=f"MATCH ({entity_key.value}) SELECT count() AS count WHERE project_id = {project_id}",
         project_id=project_id,
-        time_window=timedelta(minutes=1),
-        resolution=timedelta(minutes=1),
+        time_window_sec=60,
+        resolution_sec=60,
```
lynnagara, line 65: Thanks for fixing this script!

snuba/datasets/cdc/groupassignee.py
```diff
@@ -11,5 +15,7 @@ class GroupAssigneeDataset(Dataset):
     there is no issue in fixing the name.
     """
 
-    def __init__(self) -> None:
-        super().__init__(default_entity=EntityKey.GROUPASSIGNEE)
+    def get_all_entities(self) -> Sequence[Entity]:
```
lynnagara, line 18: No need to redefined the `get_all_entities()` function inside every single dataset. I think you should pass the entity keys to the constructor of the base Dataset class, then add the `get_all_entities()` function there to return them.

enochtangg: When you say pass entity keys to constructor of base, I'm interpreting that as defining a new attribute say `__all_entities` in our base Dataset, and initializing the keys in every dataset using `super().__init__()` which passes to the base constructor. Is that correct?

lynnagara: Yes, basically exactly the same as how default entity is passed through today, but replace that with a list of all the entities instead of just one.

enochtangg: Makes sense, this would be congruent with what we do with `PluggableDataset` (https://github.com/getsentry/snuba/blob/master/snuba/datasets/pluggable_dataset.py) and how we store Dataset related properties as attributes on the class itself .

snuba/datasets/cdc/groupassignee.py
```diff
@@ -12,4 +13,8 @@ class GroupAssigneeDataset(Dataset):
     """
 
     def __init__(self) -> None:
-        super().__init__(default_entity=EntityKey.GROUPASSIGNEE)
+        super().__init__(
+            all_entities=[
+                get_entity(EntityKey.GROUPASSIGNEE),
```
lynnagara, line 18: nit: No need to do the get entity call in every dataset, let's just pass the entity keys and do that in base as well?

snuba/web/views.py
```diff
@@ -604,16 +604,14 @@ def _write_to_entity(*, entity: EntityType) -> RespTuple:
...
-        return _write_to_entity(entity=dataset.get_default_entity())
-
     @application.route("/tests/entities/<entity:entity>/insert", methods=["POST"])
     def write_to_entity(*, entity: EntityType) -> RespTuple:
         return _write_to_entity(entity=entity)
 
-    @application.route("/tests/<dataset:dataset>/eventstream", methods=["POST"])
-    def eventstream(*, dataset: Dataset) -> RespTuple:
+    @application.route(
+        "/tests/<dataset:dataset>/<entity:entity>/eventstream", methods=["POST"]
```
lynnagara, line 612: What's the current status of this in Sentry? Have you updated this endpoint to pass entity everywhere?

enochtangg: From what I understand, this is purely a test endpoint used in both Snuba and Sentry. The Snuba eventstream tests have been updated accordingly in this PR to support this change. On the other hand, the Sentry `SnubaEventStream` class (which is only used in tests) have been updated here: https://github.com/getsentry/sentry/pull/38877

enochtangg: Updated: Both the old and new endpoints are kept, after this PR is merged, the https://github.com/getsentry/sentry/pull/38877 can be merged, finally the old eventstream endpoint can be safely removed.

lynnagara: Does dataset actually need to be in here? I believe that inserting generally happens on a storage/entity and dataset is largely only for the query interface. I think that these eventstream endpoints should match the insert endpoints here (i.e. if you look at the write_to_entity function, there is no dataset passed, only an entity.

enochtangg: Right, both new test `insert` and `eventstream` endpoints have been updated to only specify entity

snuba/web/views.py
```diff
@@ -606,14 +606,79 @@ def _write_to_entity(*, entity: EntityType) -> RespTuple:
...
+        else:
+            from snuba.replacer import ReplacerWorker
+
+            worker = ReplacerWorker(storage, "consumer_group", metrics=metrics)
+            processed = worker.process_message(message)
+            if processed is not None:
+                batch = [processed]
+                worker.flush_batch(batch)
+
+        return ("ok", 200, {"Content-Type": "text/plain"})
```
lynnagara, line 678:
```suggestion
        # This endpoint is only used by sentry, this will be deprecated after https://github.com/getsentry/sentry/pull/38877 is merged
       return self.eventstream(entity=dataset.get_all_entities()[0])
```

lynnagara (approved):

PR #3144, opened 2022-09-15
feat(multistorage-consumer): Add --raw-events-topic and --bootstrap-server options

lynnagara:
This brings the `multistorage-consumer` options closer to the `consumer` options. These two comamnds should eventually be combined anyway, so this is a step in that direction.

The primary motivation for making this change now though is that we are rolling out errors/transactions split together with the transactions/transactions_v2 double writes. This requires us to pass non default topic and Kafka cluster options to one of the multistorage consumers.

### DIFF
--- a/snuba/cli/multistorage_consumer.py
+++ b/snuba/cli/multistorage_consumer.py
@@ -43,13 +43,19 @@ logger = logging.getLogger(__name__)
     multiple=True,
     required=True,
 )
+@click.option("--raw-events-topic", help="Topic to consume raw events from.")
+@click.option(
+    "--commit-log-topic",
+    help="Topic for committed offsets to be written to, triggering post-processing task(s)",
+)
 @click.option(
     "--consumer-group",
     default="snuba-consumers",
 )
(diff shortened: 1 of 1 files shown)


snuba/cli/multistorage_consumer.py
```diff
@@ -113,8 +119,10 @@
 )
 def multistorage_consumer(
     storage_names: Sequence[str],
-    consumer_group: str,
+    raw_events_topic: Optional[str],
     commit_log_topic: str,
+    consumer_group: str,
+    bootstrap_server: Sequence[str],
```
nikhars, line 125: Shouldn't this be optional? Most of the times we wouldn't want consumers to provide the bootstrap_server.

lynnagara: Since it's a sequence, if it's not provided it's just an empty list and the default value gets applied in that case. This is the same way it's done in the other consumer CLI.

nikhars (approved):

PR #3146, opened 2022-09-15
feat(gen-metrics): Accept strings as tag values

nikhars:
When the metrics indexer starts writing tag values as strings, the payload of the message would change. We need to start accepting strings for tag values during migration to the new system. Once we migrate, we can change the protocol to only accept strings.

### Legal Boilerplate

### DIFF
--- a/snuba/datasets/generic_metrics_processor.py
+++ b/snuba/datasets/generic_metrics_processor.py
@@ -107,31 +107,25 @@ class GenericMetricsBucketProcessor(MessageProcessor, ABC):
         indexed_values: MutableSequence[int] = []
         raw_values: MutableSequence[str] = []
         tags = message["tags"]
+        version = message.get("version", 1)
         assert isinstance(tags, Mapping), "Invalid tags type"
 
         sorted_tag_items = sorted(tags.items())
         for key, value in sorted_tag_items:
-            assert key.isdigit() and (
-                isinstance(value, int) or isinstance(value, str)
-            ), "Tag key/value invalid"
-
+            assert key.isdigit(), "Tag key invalid"
             keys.append(int(key))
-            if isinstance(value, int):
+
+            if version == 1:
+                assert isinstance(value, int), "Tag value invalid"
                 indexed_values.append(value)
-                raw_values.append("")
-            elif isinstance(value, str):
+            elif version == 2:
+                assert isinstance(value, str), "Tag value invalid"
                 indexed_values.append(0)
                 raw_values.append(value)
 
-        raw_values_index = self._get_raw_values_index(message)
-        processed_raw_values: MutableSequence[str] = []
-        for offset, indexed_value in enumerate(indexed_values):
-            if indexed_value != 0:
-                processed_raw_values.append(
-                    raw_values_index.get(str(indexed_value), "")
-                )
-            else:
-                processed_raw_values.append(raw_values[offset])
+        if version == 1:
+            raw_values_index = self._get_raw_values_index(message)
+            raw_values = [raw_values_index.get(str(v), "") for v in indexed_values]
 
         processed = {
             "use_case_id": message["use_case_id"],
@@ -140,7 +134,7 @@ class GenericMetricsBucketProcessor(MessageProcessor, ABC):
             "metric_id": message["metric_id"],
             "timestamp": timestamp,
             "tags.key": keys,
-            "tags.raw_value": processed_raw_values,
+            "tags.raw_value": raw_values,
             "tags.indexed_value": indexed_values,
             **self._process_values(message),
             "materialization_version": 1,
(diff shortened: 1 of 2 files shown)


snuba/datasets/generic_metrics_processor.py
```diff
@@ -86,24 +98,40 @@ def process_message(
...
-        try:
-            retention_days = enforce_retention(message["retention_days"], timestamp)
-        except EventTooOld:
-            return None
+            keys.append(int(key))
+            if isinstance(value, int):
+                indexed_values.append(value)
+                raw_values.append("")
+            elif isinstance(value, str):
+                indexed_values.append(0)
```
fpacifici, line 123: Technically 0 is a valid value. It would be assigned to the first value we indexed. Though probably not a big deal as, if we are here, we are not trusting indexed tag values anymore.

snuba/datasets/generic_metrics_processor.py
```diff
@@ -86,24 +98,40 @@ def process_message(
...
+                raw_values.append(value)
 
         raw_values_index = self._get_raw_values_index(message)
-        raw_values = [raw_values_index.get(str(v), "") for v in indexed_values]
+        processed_raw_values: MutableSequence[str] = []
+        for offset, indexed_value in enumerate(indexed_values):
+            if indexed_value != 0:
```
fpacifici, line 129: I think it would be better to trust whether the raw value is populated or not as 0 is a valid value. 
Though empty string is valid as well. 

I think a more robust strategy would be to decide whether we are in indexed mode or not depending on the type of the tag values column and stick to that as both 0 and "" are valid values.

Even better, I would not even implicitly trust the type. 
Just add a `mode` flag in the message from the indexer that tells us whether we are in indexed mode or not. Then trust that. Messing up that way would be much harder.

fpacifici (commented):
I think I would make the `mode` explicit instead of inferring it from the tag value type which is a bit fragile. Please see more details in my comments inline about the reasoning.

snuba/datasets/generic_metrics_processor.py
nikhars: Using the `version` field now to determine whether to interpret tag values are strings or ints

snuba/datasets/generic_metrics_processor.py
untitaker: i think there's two reasons this is ok

* https://github.com/getsentry/sentry/blob/master/src/sentry/migrations/0284_metrics_indexer_alter_seq.py

* 0 is already used as a sentinel value returned by clickhouse for the group in `group by tags[foo]` where `foo` does not exist as a tag

snuba/datasets/generic_metrics_processor.py
```diff
@@ -86,24 +98,32 @@ def process_message(
...
 
-        try:
-            retention_days = enforce_retention(message["retention_days"], timestamp)
-        except EventTooOld:
-            return None
-
-        raw_values_index = self._get_raw_values_index(message)
-        raw_values = [raw_values_index.get(str(v), "") for v in indexed_values]
+            if version == 1:
+                assert isinstance(value, int), "Tag value invalid"
```
untitaker, line 120: not really something to be fixed in this pr, but I wonder if it would be feasible to type out both messages like this:

```
class MessageV1(TypedDict):
    tags: Mapping[str, int]
    version: Literal[1]

class MessageV2(TypedDict):
    tags: Mapping[str, str]
    version: Literal[2]

Message = Union[MessageV1, MessageV2]
```

snuba/datasets/generic_metrics_processor.py
fpacifici: One counter argument to that approach. 
Defensive design would prompt us not to rely on that sequence migration in postgres or on the fact we use the value on clickhouse as a sentinel to consider this ok. 
Those two are implementation details are not reflected in the design:
- The mutations is totally specific to postgres, it is not written anywhere that the sequence has to start from 1. So relying on that would be pretty dangerous as we create an implicit assumption that would be easy to forget. 
- Using 0 as a sentinel value is basically a bug, and it relies itself on a very weird and coutner intuitive clickhouse behavior that is not documented (returning 0 or empty string when looking for an entry that does not exists). Clickhouse can change the way it works.

Relying on the fact that the tag values column is meaningless, thus not queried, when not in indexed mode is more solid because it relies on a property of the data model design: either we query indexed values or we query non indexed values, we do not support hybrid queries. If we wanted to support hybrid queries (that use indexed values where the values are indexed and use string where values are not indexed) we would likely add an `index_state` column in the clickhosue table and rely on that one to decide if the row has indexed tags or not. 
That way we would be relying on something that is specifically designed to tell us whether the row is indexed. Thus we would be relying on something more robust than a property that happens to be true for indexed rows but does not really have to. This would appear evident to somebody new to the code base when the tribal knowledge is lost.

fpacifici (approved):

PR #3148, opened 2022-09-15
[wip] support second slice in migrations

onewland:
this is far from ready, has lots of bugs, but it does use the slicing to create a table in a new DB

### DIFF
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -21,12 +21,13 @@ def migrations() -> None:
 
 
 @migrations.command()
-def list() -> None:
+@click.option("--partition-id")
+def list(partition_id: Optional[int]) -> None:
     """
     Lists migrations and their statuses
     """
     check_clickhouse_connections()
-    runner = Runner()
+    runner = Runner(partition_id)
     for group, group_migrations in runner.show_all():
         click.echo(group)
         for migration_id, status, blocking in group_migrations:
@@ -52,13 +53,16 @@ def list() -> None:
 @click.option(
     "--log-level", help="Logging level to use.", type=click.Choice(LOG_LEVELS)
 )
-def migrate(force: bool, log_level: Optional[str] = None) -> None:
+@click.option("--partition-id")
+def migrate(
+    force: bool, partition_id: Optional[int] = None, log_level: Optional[str] = None
+) -> None:
     """
     Runs all migrations. Blocking migrations will not be run unless --force is passed.
     """
     setup_logging(log_level)
     check_clickhouse_connections()
-    runner = Runner()
+    runner = Runner(partition_id)
 
     try:
         runner.run_all(force=force)
@@ -77,12 +81,14 @@ def migrate(force: bool, log_level: Optional[str] = None) -> None:
 @click.option(
     "--log-level", help="Logging level to use.", type=click.Choice(LOG_LEVELS)
 )
+@click.option("--partition-id")
 def run(
     group: str,
     migration_id: str,
     force: bool,
     fake: bool,
     dry_run: bool,
+    partition_id: Optional[int] = None,
     log_level: Optional[str] = None,
 ) -> None:
     """
@@ -96,7 +102,7 @@ def run(
     if not dry_run:
         check_clickhouse_connections()
 
-    runner = Runner()
+    runner = Runner(partition_id)
     migration_group = MigrationGroup(group)
     migration_key = MigrationKey(migration_group, migration_id)
 
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -66,8 +66,13 @@ class CodeMigration(Migration, ABC):
                 print(f"Non SQL operation - {desc}")
             return
 
-        migration_id, logger, update_status = context
-        logger.info(f"Running migration: {migration_id}")
+        migration_id, logger, update_status, partition_id = context
+
+        if partition_id is not None:
+            logger.info(
+                f"Ignoring code migration {migration_id}, received partition ID {partition_id}"
+            )
+            return
         update_status(Status.IN_PROGRESS)
 
         for op in self.forwards_global():
@@ -83,7 +88,13 @@ class CodeMigration(Migration, ABC):
                 print(f"Non SQL operation - {desc}")
             return
 
-        migration_id, logger, update_status = context
+        migration_id, logger, update_status, partition_id = context
+        if partition_id is not None:
+            logger.info(
+                f"Ignoring code migration {migration_id}, received partition ID {partition_id}"
+            )
+            return
+
         logger.info(f"Reversing migration: {migration_id}")
         update_status(Status.IN_PROGRESS)
         for op in self.backwards_global():
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -134,7 +134,7 @@ class ClickhouseNodeMigration(Migration, ABC):
             self.__dry_run(self.forwards_local(), self.forwards_dist())
             return
 
-        migration_id, logger, update_status = context
+        migration_id, logger, update_status, partition_id = context
         logger.info(f"Running migration: {migration_id}")
 
         # The table does not exist before the first migration is run
@@ -142,9 +142,9 @@ class ClickhouseNodeMigration(Migration, ABC):
         if not self.is_first_migration():
             update_status(Status.IN_PROGRESS)
         for op in self.forwards_local():
-            op.execute(local=True)
+            op.execute(local=True, partition_id=partition_id)
         for op in self.forwards_dist():
-            op.execute(local=False)
+            op.execute(local=False, partition_id=partition_id)
         logger.info(f"Finished: {migration_id}")
         update_status(Status.COMPLETED)
 
@@ -153,13 +153,13 @@ class ClickhouseNodeMigration(Migration, ABC):
             self.__dry_run(self.backwards_local(), self.backwards_dist())
             return
 
-        migration_id, logger, update_status = context
+        migration_id, logger, update_status, partition_id = context
         logger.info(f"Reversing migration: {migration_id}")
         update_status(Status.IN_PROGRESS)
         for op in self.backwards_dist():
-            op.execute(local=False)
+            op.execute(local=False, partition_id=partition_id)
         for op in self.backwards_local():
-            op.execute(local=True)
+            op.execute(local=True, partition_id=partition_id)
         logger.info(f"Finished reversing: {migration_id}")
 
         # The migrations table will be destroyed if the first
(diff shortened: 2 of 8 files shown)


snuba/settings/__init__.py
```diff
@@ -71,6 +71,19 @@
...
+        "user": os.environ.get("CLICKHOUSE_USER", "default"),
+        "password": os.environ.get("CLICKHOUSE_PASSWORD", ""),
+        "database": os.environ.get("CLICKHOUSE_DATABASE", "slice_1_default"),
+        "http_port": int(os.environ.get("CLICKHOUSE_HTTP_PORT", 8123)),
+        "storage_sets": {
+            "#slice(migrations,1)",
+            "#slice(generic_metrics_sets,1)",
+        },
+        "single_node": True,
+    },
```
fpacifici, line 86: What is the rationale for adding a db for another slice in the local config? 
Is it only for your tests or do you expect to merge this part ?

snuba/migrations/operations.py
```diff
@@ -17,8 +18,21 @@ def __init__(self, storage_set: StorageSetKey):
     def storage_set(self) -> StorageSetKey:
         return self._storage_set
 
-    def execute(self, local: bool) -> None:
-        cluster = get_cluster(self._storage_set)
+    def execute(self, local: bool, partition_id: Optional[int]) -> None:
```
fpacifici, line 21: I think we should ensure to be very verbose in the log reminding to the user what partition they are running migrations on. 
Example:
- when listing the migrations, please ensure the output mentions the partition
- when doing a dry-run, same
- when executing a migraiton, please mention in the output which partition the migration was run.

snuba/migrations/runner.py
```diff
@@ -60,8 +60,13 @@ class MigrationDetails(NamedTuple):
 
 
 class Runner:
-    def __init__(self) -> None:
-        migrations_cluster = get_cluster(StorageSetKey.MIGRATIONS)
+    def __init__(self, partition_id: Optional[int]) -> None:
```
fpacifici, line 63: What is the plan to initialize the migrations table? 
Now the migrations table is created by this migration https://github.com/getsentry/snuba/blob/master/snuba/migrations/system_migrations/0001_migrations.py.

We will need a way to innitialize that table on a new slice when the new slice shows up in the configuration.

snuba/cli/migrations.py
```diff
@@ -21,12 +21,13 @@ def migrations() -> None:
 
 
 @migrations.command()
-def list() -> None:
+@click.option("--partition-id")
+def list(partition_id: Optional[int]) -> None:
```
fpacifici, line 25: Should we require the partition when running a migration that impacts a storage that is partitioned ?

snuba/migrations/migration.py
```diff
@@ -134,17 +145,17 @@ def forwards(self, context: Context, dry_run: bool = False) -> None:
             self.__dry_run(self.forwards_local(), self.forwards_dist())
             return
 
-        migration_id, logger, update_status = context
+        migration_id, logger, update_status, partition_id = context
         logger.info(f"Running migration: {migration_id}")
```
fpacifici, line 149: Please add the partition id to the message

snuba/migrations/operations.py
```diff
@@ -17,8 +18,21 @@ def __init__(self, storage_set: StorageSetKey):
...
+        if (
+            partition_id is not None
+            and self._storage_set.value not in settings.PARTITIONED_STORAGE_SETS
```
fpacifici, line 24: Are we going to have a `migrate_all` command that executes all migrations on all slices ? 
Normally the process for self hosted installs is to run bootstrap (or really any other command), that executes all the pending migrations. 

Since slices will be something people can set up, they should be taken into account automatically when moving to a new version.

snuba/settings/__init__.py
```diff
@@ -38,7 +38,7 @@
     str(x): 0 for x in range(0, SENTRY_LOGICAL_PARTITIONS)
 }
 # Storage names to apply dataset partitioning to
-PARTITIONED_STORAGES: Set[str] = set()
+PARTITIONED_STORAGE_SETS: Set[str] = {"migrations", "generic_metrics_sets"}
```
fpacifici, line 41: Is this config specific to the migration framework or is it used by snuba in other places. 
I think we should try to keep them separated as we want to see the migrations tool as a stand alone entity that we could break out of snuba. 
Right now it only depends on Storage set keys and storage keys. Those are not going to be enums anymore soon so no big deal, though we should try to avoid the logical dependencies as well.

fpacifici (commented):
Some early comments

How do you envision the process to run migrations on a multi slice system will work ?
Specifically these cases:
- I am installing a new version of sentry so I have a few migrations to run. My config contains multiple slices, how does it work? Is the `migrate` command going to run migrations on all slices ?
- I am up to date with the migration and I am adding a new slice. How do I bring that sice up to date ? Is there going to be a single command that runs all the pending migrations on partitioned storages ?
- I have a new migration to run on a partitioned storage. Can I run it slice by slice?

snuba/settings/__init__.py
onewland: I just did this so that I could run `snuba migrate` locally. I think before merge I would add a second slice to `settings_test` so we could automate that migrations run correctly on a second slice, but this is going to be removed.

snuba/migrations/runner.py
onewland: in my local environment, this just sort of worked (e.g. when I created the new database (`slice_1_default`) and ran the initial migrations against it it built a new copy of the migrations table.

onewland:
> Some early comments
> 
> How do you envision the process to run migrations on a multi slice system will work ? Specifically these cases:
> 
> * I am installing a new version of sentry so I have a few migrations to run. My config contains multiple slices, how does it work? Is the `migrate` command going to run migrations on all slices ?

I hadn't thought about this use case, but I think it would be fairly straightforward to use to iterate from 0 to `settings.LOCAL_PHYSICAL_PARTITIONS-1` and run the equivalent of the migrate command on each physical partition ID

> * I am up to date with the migration and I am adding a new slice. How do I bring that sice up to date ? Is there going to be a single command that runs all the pending migrations on partitioned storages ?

I was able to run `snuba migrate --parition-id=1` locally and it was successful (it ignored storage sets that aren't partitioned, and ran all migrations for generic_metrics_distributions).

> * I have a new migration to run on a partitioned storage. Can I run it slice by slice?

Yes, in fact right now the alternative isn't supported (running a migration on all slices).

snuba/settings/__init__.py
onewland: I was assuming `StorageSetKey.at_partition` would be used in the query path to switch between storage sets depending on logical partition (though logical : physical mapping isn't yet being used):

https://github.com/getsentry/snuba/pull/3148/files#diff-045703230a5e4a12e1bde538fccabfab59d530d2bf982f8b21c94ab0d8010ddbR79-R87

nikhars (commented):
I might be digressing a bit, but I don’t think migrations should know anything about partition/slice. That means migration code does not support partition at all. We should have something similar to a proxy(via config overrides) which can route the migration to the appropriate slice. There is a drawback that migrations can then only address one slice at a time. But that is a good thing in my opinion. Once a migration has been run there should be a verification step which we don’t have automated right now. So it has to be manual for now.

fpacifici:
> That means migration code does not support partition at all. We should have something similar to a proxy(via config overrides) which can route the migration to the appropriate slice.

What Oliver is doing is to fully partition the migration state. We will have fully separate migration tables per slice. 
Then the question is only how we target one table vs another. 

Now the question is how do you target one table vs another. CLI argument vs config vs others (like environment variables). 
I think there may be merits in both solutions so I will just leave here the list of scenarios we need to address:
- CLI in our production environment. Here we can apply either env vars or settings override or pass CLI arguments 
- Snuba admin UI. Here applying settings override will be trickier but possible. Envvars are likely going to be impossible unless we spin up multiple admins (which I think we should not do)
- Snuba bootstrap. That expects all migrations to run so there has to be a way to run migrations for all slices in one command, whenter the mgiration framework knows about them or whether it is the bootstrap tool that knows about slices and issues migrations on all slices.

snuba/settings/__init__.py
fpacifici: I am not sure migrations can be partitioned at storage set level (allowing a migrations group to contain both partitioned and not partitioned migrations).
The migration tool enforces that, for a migration table, migrations for the same group are executed in sequence (you cannot skip them). 
Though if a migration group contains migrations for multiple storage sets (which I think is possible) you have a problem:
- let's assume group `A` contains migrations from storage sets `B` and `C`.  Now `B` is partitioned on 5 slices while `C` has no slices. 
- You will not have problem in running migrations on the table for slice 0. All Storage sets are present there. 
- The table for slice 1 will become inconsistent as it will be missing the migrations from storage set `C`. The tool will fail there. 

So I think we need to revisit this.
Either allow groups to skip migrations, or disallow different storage sets in the same group.
Or make slicing not tied to the storage set but to something else.

I think this may become controversial.

onewland: I think this whole system gets incredibly complicated if we allow for different levels of physical slicing (at a settings/conceptual level) within the same deployment. So if we have n physical partitions (n << 256) I think those partitions need to apply across all CH storages *even if* some storages will internally map multiple physical partitions to one physical CH cluster.

Example:
There are 4 physical partitions in a snuba deploy.
`generic_metrics_distributions`, which contains the most metrics data, maps 
```
{
    physical_part_1: cluster_1,
    physical_part_2: cluster_2,
    physical_part_3: cluster_3,
    physical_part_4: cluster_4
}
```

whereas `generic_metrics_sets` which doesn't require nearly as much space will map
```
{
    physical_part_1: cluster_1,
    physical_part_2: cluster_1,
    physical_part_3: cluster_1,
    physical_part_4: cluster_1
}
```

It's sort of like a second level of logical mapping, but it means we can reuse the migrations table for all the physical partitions in the second group but run them independently for the first.

(I'm not taking into consideration Kafka here)

fpacifici: I am not sure we are addressing the same problem. 
My concern is that having a migrations table per slice breaks the migration tools data model. 
- The migration tool data model require sequentiality within a migration group. You cannot skip migrations.
- There is nothing forcing a group to contain migrations that only reference one storage key/set 
- Thus, the same group can contain at the same time migrations for a sliced storage and migrations for a non sliced storage.
- We would not like to change any of the constraint above

This implies, that running slice 2 on a separate migration table may require you to skip migrations in a group as the group may contain migrations that should not be applied to slice 2 as they are for a non sliced storage.
Skipping migrations in a group breaks the sequentiality point above.

onewland:
we're not going forward with this approach

PR #3149, opened 2022-09-15
feat(slicing): StorageSetKey override config for slicing

ayirr7:
- removes `LOCAL_PHYSICAL_PARTITIONS` which assumes a constant number of slices across storages, and an associated test
- adds config for storage set key overrides based on slice ID

- addresses SNS-1688, SNS-1707, and partially addresses SNS-1708
(https://getsentry.atlassian.net/browse/SNS-1688, https://getsentry.atlassian.net/browse/SNS-1708 and https://getsentry.atlassian.net/browse/SNS-1707)

### DIFF
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -404,7 +404,7 @@ class UndefinedClickhouseCluster(SerializableException):
 
 
 def get_cluster(
-    storage_set_key: StorageSetKey, slice_id: Optional[int]
+    storage_set_key: StorageSetKey, slice_id: Optional[int] = None
 ) -> ClickhouseCluster:
     """Return a clickhouse cluster for a storage set key.
 
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -421,7 +421,8 @@ def get_cluster(
         ]
         storage_set_key = StorageSetKey(sliced_key_name)
 
-    # we assume that below map will be updated accordingly with the new clusters
+    # we assume that below map will be updated accordingly with
+    # the new clusters and sliced StorageSetKeys
     res = _get_storage_set_cluster_map().get(storage_set_key, None)
     if res is None:
         raise UndefinedClickhouseCluster(
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -403,7 +403,9 @@ class UndefinedClickhouseCluster(SerializableException):
     pass
 
 
-def get_cluster(storage_set_key: StorageSetKey) -> ClickhouseCluster:
+def get_cluster(
+    storage_set_key: StorageSetKey, slice_id: Optional[int]
+) -> ClickhouseCluster:
     """Return a clickhouse cluster for a storage set key.
 
     If the storage set key is not defined
@@ -412,6 +414,14 @@ def get_cluster(storage_set_key: StorageSetKey) -> ClickhouseCluster:
     assert (
         storage_set_key not in DEV_STORAGE_SETS or settings.ENABLE_DEV_FEATURES
     ), f"Storage set {storage_set_key} is disabled"
+
+    if slice_id:
+        sliced_key_name = settings.OVERRIDE_STORAGE_SET_KEYS[storage_set_key.value][
+            slice_id
+        ]
+        storage_set_key = StorageSetKey(sliced_key_name)
+
+    # we assume that below map will be updated accordingly with the new clusters
     res = _get_storage_set_cluster_map().get(storage_set_key, None)
     if res is None:
         raise UndefinedClickhouseCluster(
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -29,6 +29,7 @@ DISABLED_DATASETS: Set[str] = set()
 # Clickhouse Options
 CLICKHOUSE_MAX_POOL_SIZE = 25
 
+# Mapping of StorageSetKeys to a Mapping of slice id to sliced StorageSetKeys
 OVERRIDE_STORAGE_SET_KEYS: Mapping[str, Mapping[int, str]] = {}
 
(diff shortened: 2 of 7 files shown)


snuba/settings/__init__.py
```diff
@@ -29,9 +29,9 @@
 # Clickhouse Options
 CLICKHOUSE_MAX_POOL_SIZE = 25
 
-# The number of physical partitions that we will need to map resources to
-# for storage partitioning
-LOCAL_PHYSICAL_PARTITIONS = 1
```
nikhars, line 34: Even though this is not needed across all of storage sets, there would be a requirement to map logical partitions to a specific physical partitions based on storage set.

snuba/clusters/cluster.py
```diff
@@ -403,7 +403,9 @@ class UndefinedClickhouseCluster(SerializableException):
     pass
 
 
-def get_cluster(storage_set_key: StorageSetKey) -> ClickhouseCluster:
+def get_cluster(
```
nikhars, line 406: Can you write a test for the changes made here?

tests/clusters/test_cluster.py
```diff
@@ -196,3 +218,20 @@ def test_cache_connections() -> None:
...
+@patch("snuba.settings.OVERRIDE_STORAGE_SET_KEYS", OVERRIDE_STORAGE_SET_KEY_CONFIG)
+@patch("snuba.settings.CLUSTERS", TEST_SLICE_CONFIG)
+@patch(
+    "snuba.clusters.storage_sets._REGISTERED_STORAGE_SET_KEYS",
+    {"GENERIC_METRICS_DISTRIBUTIONS_1": "generic_metrics_distributions_1"},
+)
+def test_override_storage_set_keys() -> None:
+    importlib.reload(cluster)
+
+    res_cluster = cluster.get_cluster(StorageSetKey.GENERIC_METRICS_DISTRIBUTIONS, 1)
```
nikhars, line 232: Can you also add a test which asserts that without the optional slice, we map to the original cluster?

tests/clusters/test_cluster.py
```diff
@@ -80,6 +80,28 @@
...
+    }
+]
+
+
+OVERRIDE_STORAGE_SET_KEY_CONFIG = {
+    "generic_metrics_distributions": {
+        0: "generic_metrics_distributions",
+        1: "generic_metrics_distributions_1",
+    }
+}
```
lynnagara, line 102: I don't really understand how this pieces fit together. How is 0 and 1 defined? I would prefer if we do not attempt to override storage sets at all and just map storage set/partition pairs which are unique to a cluster. I feel it is unnecessary and adds additional complexity.

onewland: I'm not sure I understand this comment. 0 and 1 are physical partitions and so she is using a storage set key+partition pair to do this override.

So are you saying that CLUSTERS should contain partition information? My thought is that should not be necessary for unsliced storages which are the vast majority, and will be completely unused by single-tenant/self-hosted users

onewland: Would you have the same opposition if this was written out (and documented) as:

```python
STORAGE_SET_KEY_PARTITION_MAP = {
     "storage_set_key_no_partitions": {
           part_number_0: "storage_set_key_partition_0",
           # ...
           part_number_n: "storage_set_key_partition_n"
     }
}
```

lynnagara: Yes, to summarize my point is that this is currently mapping:

storage set / partition pair -> another thing that we are also kind of calling a storage set but is not really -> cluster

Why not just map:
storage set / partition pair (which is already unique) -> cluster

My suggestion is to remove the part in the middle and go straight from (storage set / partition) to cluster. The extra complexity is both in Snuba code as well as partially exposed to users via these settings which is why I am opposed to it.

snuba/settings/__init__.py
ayirr7: I'll add the LOCAL_PHYSICAL_PARTITIONS back.

tests/clusters/test_cluster.py
onewland: I don't agree with "another thing that we are also kind of calling a storage set but is not really". My interpretation is that `storage_set_key_partition_n` is a storage set as it defines colocated resources to clusters (as defined by settings.CLUSTERS) for slice n. It is perhaps not referred to directly in the storage definition, but that's fine with me.

If we changed CLUSTERS to contain partition information, then I think we're adding extra complexity to cluster definitions and parsing from which almost every dataset and almost every user derives no value. Therefore I still think we'd probably want to create an override structure with your approach.

If we create that cluster override structure, then we end up with a mapping like:

```python
PARTITIONED_CLUSTERS = [
   {
        "host": "host_slice",
        "port": 9000,
        "user": "default",
        "password": "",
        "database": "default",
        "http_port": 8123,
        "storage_sets": {("generic_metrics_distributions",0)},
        "single_node": True,
   },
   {
        "host": "host_slice",
        "port": 9001,
        "user": "default",
        "password": "",
        "database": "default",
        "http_port": 8124,
        "storage_sets": {("generic_metrics_distributions",1)},
        "single_node": True,
   },
]
```

which in my opinion is the same complexity to support in Snuba as what Riya did. I'm going to write up a bit more in notion but unless you think that this is functionally limiting I don't think it should block this approach

lynnagara: This example is more straightforward to me that the one in the code with the storage set overrides since there is no need to write `storage_set_key_partition_n` in two different places which was the additional complexity I was referring to (to clarify I think this is more mental complexity of having another thing here more than overall code complexity which is likely on par either way). You could leave CLUSTERS unchanged and only put the overrides in a structure like this.

onewland:
This is close-able, right?

ayirr7:
Yes, this can be closed. I will do it.

PR #3158, opened 2022-09-19
feat(optimize): track merge state

dbanda:
Addresses [SNS-1483](https://getsentry.atlassian.net/browse/SNS-1483). 

Adds functionality to query `system.merges` at the start of an optimization and delay running `OPTIMIZE` on a partition if there are already ongoing merges on that partition.

It works by calling `optimize.is_busy_merging()` to check if it should execute a merge or sleep then retry later. 

`is_busy_merging` queries `system.merges` table for merges executing and returns true and the estimated sleep time if clickhouse is busy with merges in progress for the table. Clickhouse is considered busy if:
1. there are more than `settings.OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS` merges in progress with an elapsed time > `settings.OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME`. These are merges that we consider long running and should probably wait on before attempting again.
2. or there is a merge where `total_size_bytes_compressed` is greater than `settings.OPTIMIZE_MERGE_SIZE_CUTOFF`. These are merges on large parts and we should wait for them.

In both cases, the sleep time is `setting.OPTIMIZE_BASE_SLEEP_TIME + estimated_sleep_time` where `estimated_sleep_time` is the max of `elapsed/progress` of the ongoing merges found in the query.

### DIFF
--- a/snuba/cleanup.py
+++ b/snuba/cleanup.py
@@ -31,7 +31,7 @@ def get_active_partitions(
 
     response = clickhouse.execute(
         """
-        SELECT DISTINCT partition
+        SELECT DISTINCT partition, partition_id
         FROM system.parts
         WHERE database = %(database)s
         AND table = %(table)s
--- a/snuba/cleanup.py
+++ b/snuba/cleanup.py
@@ -31,7 +31,7 @@ def get_active_partitions(
 
     response = clickhouse.execute(
         """
-        SELECT DISTINCT partition, partition_id
+        SELECT DISTINCT partition
         FROM system.parts
         WHERE database = %(database)s
         AND table = %(table)s
--- a/snuba/optimize.py
+++ b/snuba/optimize.py
@@ -204,7 +204,7 @@ def get_current_merging_partitions_info(
     table: str,
 ) -> Sequence[util.MergeInfo]:
     """
-    Returns a dictionary of partitions that are currently part of large merged. Merges
+    Returns a dictionary of partitions that are currently part of large merge. Merges
     are considered large if they are longer than OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME
     or if they are larger than OPTIMIZE_MERGE_MIN_PARTS_CUTTOFF_TIME.
     """
--- a/snuba/optimize.py
+++ b/snuba/optimize.py
@@ -1,11 +1,12 @@
+import asyncio
 import logging
 import threading
 import time
 from datetime import datetime, timedelta
-from typing import Mapping, MutableSequence, Optional, Sequence
+from typing import Any, Dict, Mapping, MutableSequence, Optional, Sequence
 
 from snuba import environment, util
-from snuba.clickhouse.native import ClickhousePool
+from snuba.clickhouse.native import ClickhousePool, ClickhouseResult
 from snuba.datasets.schemas.tables import TableSchema
 from snuba.datasets.storage import ReadableTableStorage
 from snuba.optimize_scheduler import OptimizeScheduler
@@ -40,15 +41,23 @@ def run_optimize(
     assert isinstance(schema, TableSchema)
     table = schema.get_local_table_name()
     database = storage.get_cluster().get_database()
+    part_format = schema.get_part_format()
+    assert part_format is not None
 
     partitions = get_partitions_to_optimize(
         clickhouse, storage, database, table, before
     )
+
+    merging_partitions_info = get_current_merging_partitions_info(
+        clickhouse, database, table, partitions
+    )
+
     partition_names = [partition.name for partition in partitions]
 
     optimize_partitions(
         clickhouse=clickhouse,
         database=database,
+        current_merge_info=merging_partitions_info,
         table=table,
         partitions=partition_names,
     )
@@ -78,6 +87,8 @@ def run_optimize_cron_job(
     assert isinstance(schema, TableSchema)
     table = schema.get_local_table_name()
     database = storage.get_cluster().get_database()
+    part_format = schema.get_part_format()
+    assert part_format
     optimize_scheduler = OptimizeScheduler(parallel=parallel)
 
     try:
@@ -88,6 +99,10 @@ def run_optimize_cron_job(
         partitions = get_partitions_to_optimize(
             clickhouse, storage, database, table, before
         )
+        merging_partitions_info = get_current_merging_partitions_info(
+            clickhouse, database, table, partitions
+        )
+
         if len(partitions) == 0:
             logger.info("No partitions need optimization")
             return 0
@@ -106,6 +121,7 @@ def run_optimize_cron_job(
         database=database,
         table=table,
         partitions=list(partitions_to_optimize),
+        current_merge_info=merging_partitions_info,
         tracker=tracker,
         scheduler=optimize_scheduler,
         clickhouse_host=clickhouse_host,
@@ -161,14 +177,15 @@ def get_partitions_to_optimize(
         """
         SELECT
             partition,
+            partition_id,
             count() AS c
         FROM system.parts
         WHERE active
         AND database = %(database)s
         AND table = %(table)s
-        GROUP BY partition
+        GROUP BY partition, partition_id
         HAVING c > 1
-        ORDER BY c DESC, partition
+        ORDER BY c DESC, partition, partition_id
         """,
         {"database": database, "table": table},
     )
@@ -190,11 +207,63 @@ def get_partitions_to_optimize(
     return parts
 
 
+def get_current_merging_partitions_info(
+    clickhouse: ClickhousePool,
+    database: str,
+    table: str,
+    partitions: Sequence[util.Part],
+) -> Dict[str, Any]:
+    """
+    Returns a dictionary of partitions that are currently being merged.
+    """
+
+    merging_parts = clickhouse.execute(
+        """
+        SELECT
+            result_part_name,
+            max(elapsed),
+            min(progress),      -- get the slowest merge combination possible
+            count() AS c
+        FROM system.merges
+        WHERE database = %(database)s
+        AND table = %(table)s
+        GROUP BY result_part_name
+        ORDER BY c DESC, result_part_name
+        """,
+        {"database": database, "table": table},
+    )
+
+    return __build_merging_partitions_info(merging_parts, partitions)
+
+
+def __build_merging_partitions_info(
+    merge_resuts: ClickhouseResult, partitions: Sequence[util.Part]
+) -> Dict[str, Any]:
+
+    merge_partitions_ids = {}
+    for part, elapsed, progress, count in merge_resuts.results:
+        partition_id = part.split("_")[0]
+        merge_partitions_ids[partition_id] = {
+            "partition_id": partition_id,
+            "elapsed": elapsed,
+            "progress": progress,
+            "estimated_time": elapsed / (progress + 0.0001),
+        }
+
+    merging_partitions_info = {
+        part.name: merge_partitions_ids[part.partition_id]
+        for part in partitions
+        if part.partition_id in merge_partitions_ids
+    }
+    return merging_partitions_info
+
+
 def optimize_partition_runner(
     clickhouse: ClickhousePool,
     database: str,
     table: str,
     partitions: Sequence[str],
+    current_merge_info: Dict[str, Any],
     scheduler: OptimizeScheduler,
     tracker: OptimizedPartitionTracker,
     clickhouse_host: str,
@@ -229,6 +298,7 @@ def optimize_partition_runner(
                         database,
                         table,
                         schedule.partitions[i],
+                        current_merge_info,
                         schedule.cutoff_time,
                         tracker,
                         clickhouse_host,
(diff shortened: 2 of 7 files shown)


snuba/optimize.py
```diff
@@ -291,15 +359,20 @@ def optimize_partitions(
...
         # executing the optimization. In that case we don't want to
         # run optimization on the same partition twice.
         if tracker:
-            tracker.update_completed_partitions(partition)
+            tracker.update_scheduled_partitions(partition)
+
+        # if theres a merge in progress for this partition, wait for it to finish
+        if partition in current_merge_info:
+            est_sleep_time = current_merge_info[partition]["estimated_time"]
+            time.sleep(OPTIMIZE_BASE_SLEEP_TIME + est_sleep_time)
```
evanh, line 373: Is there a way to do this without pausing the entire process? As in, move on to another partition while this one finishes?

evanh (commented):
This looks good to me but I'll leave final review to someone with more context.

snuba/optimize.py
dbanda: One thing I tried is wrapping this in an async function but then I spoke to @nikhars about that and he said it's actually preferred to pause the entire process because we want to be running only one at a time. He prob can provide extra context here too.

fpacifici, line 373: I think this algorithm to make the decision is quite weak. 
- It relies on the estimate time to be at least reasonably accurate. I don;'t think we ever measured this accuracy, but if there is an increase in load or a slow down in the disk latency, definitively the estimated time will increase. 
- we run two optimize merges at a time in general. they do not necessarily start at the same time, though one will add load to the system while the other is running thus impacting the estimate.
- The query returns only one of the large merges while we may have mutations going on. 

I think you can make it safer:
- never start anything if there are more than 3 merges on going. 
- ignore all the merges that are expected to take short in total (like 10 minutes or so). 
- If there are long merges on going (>10 minutes), get the expected time, wait till then and repeat the checks.
- make the constants above (3 concurrent merges and min 10 minutes configurable via settings)

So you are sure you are not going to rely on an unreliable estimate and you take into account mutations that may have started in the meantime

fpacifici (changes requested):
please see the comments inline

snuba/optimize.py
nikhars: Agreed with what Filippo has mentioned above. I have commented here https://github.com/getsentry/snuba/pull/3158/files#diff-85a4b5d71565524e6a2d3f9540361a4b0093941afb3ae0a8f94bc0532ff87d49R206 what the api should do. Ideally that api should be used here directly so that it can be used in a loop to keep checking that the count has gone to 0. You don't need to know the exact partition which is having the long merge. You want to know the count of them.

snuba/optimize_tracker.py
```diff
@@ -108,7 +108,7 @@ def get_partitions_to_optimize(self) -> Set[str]:
 
         """
         all_partitions = self.get_all_partitions()
-        completed_partitions = self.get_completed_partitions()
+        completed_partitions = self.get_scheduled_partitions()
```
nikhars, line 111:
```suggestion
        scheduled_partitions = self.get_scheduled_partitions()
```

snuba/cleanup.py
```diff
@@ -31,7 +31,7 @@ def get_active_partitions(
 
     response = clickhouse.execute(
         """
-        SELECT DISTINCT partition
+        SELECT DISTINCT partition, partition_id
```
nikhars, line 34: Could you elaborate a bit on why you need the `partition_id` in the query?

snuba/optimize.py
```diff
@@ -189,6 +197,48 @@ def get_partitions_to_optimize(
     return parts
 
 
+def get_current_merging_partitions_info(
+    clickhouse: ClickhousePool,
+    database: str,
+    table: str,
+) -> Sequence[util.MergeInfo]:
+    """
+    Returns a dictionary of partitions that are currently part of large merged. Merges
```
nikhars, line 207:
```suggestion
    Returns a dictionary of partitions that are currently part of large merge. Merges
```

snuba/optimize.py
```diff
@@ -291,18 +341,58 @@ def optimize_partitions(
...
+    merge_info = get_current_merging_partitions_info(clickhouse, database, table)
+
+    if len(merge_info) > OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS:
+        estimated_sleep_time = max(
+            merge_info, key=lambda x: x.estimated_time
+        ).estimated_time
+        logger.info(
+            f"too many concurrent merges {len(merge_info)}, sleeping for {estimated_sleep_time}s"
+        )
+        return True, estimated_sleep_time
```
nikhars, line 390: We should not have this check. It is possible for clickhouse to run many small merges concurrently. Those have no impact on the system. We should pause only if big merges are ongoing.

snuba/optimize.py
dbanda: Updated the algorithm. We sleep if clickhouse is busy, where Clickhouse is considered busy if:

1. there are more than settings.OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS merges in progress with an elapsed time > settings.OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME. These are merges that we consider long running and should probably wait on before attempting again.
2. Or there is a merge where total_size_bytes_compressed is greater than settings.OPTIMIZE_MERGE_SIZE_CUTOFF. These are merges on large parts and we should wait for them.

snuba/optimize.py
dbanda: I updated `get_current_merging_partitions_info` to return only long running merges as Filippo suggested then sleep if we have too many of those going on. Should I only check size of merges?

snuba/cleanup.py
dbanda: was a typo. removed.

snuba/optimize.py
nikhars: Ok. I think we can improve readability of the code. `get_current_merging_partitions_info` should be called something like `get_long_merges_info` or something. The current name suggests that it gets all of the current merges.

snuba/optimize.py
```diff
@@ -292,18 +342,58 @@ def optimize_partitions(
...
+        2. or there is a merge of size greater than OPTIMIZE_MERGE_SIZE_CUTOFF
+    """
+    merge_info = get_current_merging_partitions_info(clickhouse, database, table)
+
+    if len(merge_info) > OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS:
+        estimated_sleep_time = max(
+            merge_info, key=lambda x: x.estimated_time
+        ).estimated_time
+        logger.info(
+            f"too many concurrent merges {len(merge_info)}, sleeping for {estimated_sleep_time}s"
```
nikhars, line 388:
```suggestion
            f"too many concurrent long merges {len(merge_info)}, sleeping for {estimated_sleep_time}s"
```

snuba/optimize.py
```diff
@@ -292,18 +342,58 @@ def optimize_partitions(
...
         )
+
+
+def is_busy_merging(
+    clickhouse: ClickhousePool, database: str, table: str
+) -> Tuple[bool, float]:
+    """
+    Returns true and the estimated sleep time if clickhouse is busy with merges in progress
+    for the table. Clickhouse is considered busy if
+        1. there are more than OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS merges in progress
```
nikhars, line 378:
```suggestion
        1. there are more than OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS merges in progress with an elapsed time greater than OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME
```

snuba/settings/__init__.py
```diff
@@ -284,6 +284,14 @@ class RedisClusters(TypedDict):
...
 OPTIMIZE_JOB_CUTOFF_TIME = timedelta(hours=23)
 OPTIMIZE_QUERY_TIMEOUT = 4 * 60 * 60  # 4 hours
+# sleep time to wait for a merge to complete
+OPTIMIZE_BASE_SLEEP_TIME = 300  # 5 mins
+# merges longer than this will be considered long running
+OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME = 10  # 10 mins
+# merges larger than this will be considered long running and will be waited on
+OPTIMIZE_MERGE_SIZE_CUTOFF = 50_000_000_000  # 50GB
+# max number of running merges to wait on
+OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS = 3
```
nikhars, line 294: Please rename this variable to reflect jobs running greater than elapsed time cutoff

Also, the value of this constant should be 1. It is possible that you might have 2 merges of 100GB partitions and they satisfy the criteria of being greater than elapsed time. We should pause in those scenarios

snuba/optimize.py
```diff
@@ -189,11 +203,63 @@ def get_partitions_to_optimize(
...
+
+    merge_partitions_ids = {}
+    for part, elapsed, progress, count in merge_resuts.results:
+        partition_id = part.split("_")[0]
+        merge_partitions_ids[partition_id] = {
+            "partition_id": partition_id,
+            "elapsed": elapsed,
+            "progress": progress,
+            "estimated_time": elapsed / (progress + 0.0001),
+        }
```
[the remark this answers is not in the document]
dbanda: moved to a dataclass.

snuba/settings/__init__.py
dbanda: The current algo sleeps if there's any merge greater than 50GB . So in that case, we would sleep on a single 100GB merge. The `OPTIMIZE_MERGE_MAX_CONCURRENT_JOBS` check is only applied to long running merges of smaller size. So say if there were 3 running merges with elapsed time > 10 min of size 5 GB each, it would sleep.

snuba/util.py
```diff
@@ -275,3 +276,16 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
         return cast(F, wrapper)
 
     return decorator
+
+
+@dataclass
+class MergeInfo:
+    partition_id: str
```
nikhars, line 283: I think you can remove `partition_id` from this dataclass since you don't use this anywhere.

nikhars (approved):
Left a minor comment. Else looks good

snuba/optimize.py
```diff
@@ -189,11 +203,63 @@ def get_partitions_to_optimize(
...
+    partitions: Sequence[util.Part],
+) -> Dict[str, Any]:
+    """
+    Returns a dictionary of partitions that are currently being merged.
+    """
+
+    merging_parts = clickhouse.execute(
+        """
+        SELECT
+            result_part_name,
```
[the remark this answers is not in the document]
dbanda: Updated the query to instead get all ongoing merges then filter by size and running time.

snuba/util.py
dbanda: removed

snuba/util.py
```diff
@@ -275,3 +276,16 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+class MergeInfo:
+    result_part_name: str
+    elapsed: float
+    progress: float
+    size: int
+
+    @property
+    # estimated time remaining in seconds
+    def estimated_time(self) -> float:
+        return self.elapsed / (self.progress + 0.0001)
```
fpacifici, line 291: I would suggest not adding this to the util module. 
Besides the fact that there should be no util module as it become the kitchen sink, I think this belongs to where we keep the optimize code. Now, unfortunately, that is a module in the root directory.  I think you should create a new directory for the optimize work.  
Put it inside this https://github.com/getsentry/snuba/tree/master/snuba/clickhouse

snuba/optimize.py
```diff
@@ -292,18 +341,58 @@ def optimize_partitions(
...
 
+        # if theres a merge in progress for this partition, wait for it to finish
+        while True:
+            busy_merging, estimated_sleep_time = is_busy_merging(
+                clickhouse, database, table
+            )
+            if not busy_merging:
+                break
+            else:
+                time.sleep(OPTIMIZE_BASE_SLEEP_TIME + estimated_sleep_time)
```
fpacifici, line 360: I think you can be more aggressive here. If Clickhosue tells it will take 4 hours. You can still check once in a while. The elapsed time speeds up a lot if the load diminishes. Just check every 5 minutes and call it a day.

snuba/optimize.py
```diff
@@ -292,18 +341,58 @@ def optimize_partitions(
...
         )
+
+
+def is_busy_merging(
+    clickhouse: ClickhousePool, database: str, table: str
+) -> Tuple[bool, float]:
+    """
+    Returns true and the estimated sleep time if clickhouse is busy with merges in progress
+    for the table. Clickhouse is considered busy if
+        1. there are more than OPTIMIZE_MERGE_MAX_LONG_CONCURRENT_JOBS merges in progress with an elapsed time greater than OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME
```
fpacifici, line 377: Mind cutting this line shorter. It is really hard to read. According to PEP-8 docstrings should be 80 characters wide so they remain more readable on code reviews.

fpacifici (approved):
Please address the comments in the review before merging.

fpacifici:
@dbanda I noticed now this was set to auto-merge. Please issue a new PR to fix the comment in the review.

PR #3159, opened 2022-09-19
feat(partitioning): Method to map logical partition between nodes

nikhars:
Added a mechanism to map logical partitions to nodes/slices. The mechanism uses consistent hashing to have minimal data movement when new nodes/slices are added.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,3 +30,4 @@ urllib3==1.26.11
 pyuwsgi==2.0.20
 Werkzeug==2.2.2
 PyYAML==6.0
+uhashring~=2.1
--- a/snuba/datasets/partitioning.py
+++ b/snuba/datasets/partitioning.py
@@ -1,6 +1,37 @@
+from typing import TypeVar
+
+from uhashring import HashRing
+
 """
 The number of logical partitions used to distinguish between where records
 should be stored. These do not require individual physical partitions but allow
 for repartitioning with less code changes per physical change.
 """
 SENTRY_LOGICAL_PARTITIONS = 256
+
+LOGICAL_PARTITION = TypeVar("LOGICAL_PARTITION", bound=int)
+
+# NODE can represent anything which can be uniquely identified in a hash ring.
+# In the context of Snuba, this is a string representing the storage set which
+# should be used for querying data.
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -30,3 +30,4 @@ urllib3==1.26.11
 pyuwsgi==2.0.20
 Werkzeug==2.2.2
 PyYAML==6.0
+uhashring~=2.1
```
volokluev, line 33: is this available in the getsentry pypi?

https://github.com/getsentry/pypi

volokluev (commented):
It's hard for me to review this PR without understanding how it fits into the big picture. If we check this in, it will be dead code.

Can we instead define the interface to select partitions first, implement it (for `physical_partitions=1`), and then add the consistent hashing logic?

Also, is there anything written down about how data will be moved between nodes (facilitated by this consistent hashing)?

snuba/datasets/partitioning.py
```diff
@@ -1,6 +1,37 @@
...
+    logical_partition: LOGICAL_PARTITION, nodes: HashRing
+) -> NODE:
+    """
+    Given a hash ring containing all the nodes on which the partition could
+    exist, returns the node that the logical partition is assigned to.
+
+    We use consistent hashing to ensure minimal data movement of logical
+    partitions when nodes are added or removed from the cluster.
+    """
+    return nodes.get_node(str(logical_partition))
```
lynnagara, line 37: I thought the the mapping of logical to physical partitions was supposed to be done via this mechanism https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L37-L39 introduced in https://github.com/getsentry/snuba/pull/3135  + was supposed to be configurable via settings?

fpacifici: Indeed, mapping between orgs and logical shards is via hashing. Logical shards are statically mapped to physical resources so we can move them at will.

nikhars: Ok. Going to close this PR. I had seen that mapping in the settings before. But I wasn't sure how it was going to evolve when new slices are added. Definitely should not be mod based in the future.

PR #3163, opened 2022-09-21
feat: Update the default log level to warn

lynnagara:
INFO is too verbose for most scenarios, let's default to WARN

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -6,7 +6,7 @@ from typing import Any, Mapping, MutableMapping, Optional, Sequence, Set
 from snuba.datasets.partitioning import SENTRY_LOGICAL_PARTITIONS
 from snuba.settings.validation import validate_settings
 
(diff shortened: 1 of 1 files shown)


snuba/settings/__init__.py
```diff
@@ -6,7 +6,7 @@
 from snuba.datasets.partitioning import SENTRY_LOGICAL_PARTITIONS
 from snuba.settings.validation import validate_settings
 
-LOG_LEVEL = os.environ.get("LOG_LEVEL", "INFO")
+LOG_LEVEL = os.environ.get("LOG_LEVEL", "WARN")
```
fpacifici, line 9: Hold on. This way we will loose logs we actually rely on today (unless we start those pods explicit as info):
- Batches commit in the consumers 
- Partitions assigned/revoked 

I don't think we should loose those

lynnagara: Ah, fair enough. I assumed this was going to be a more straightforward task

PR #3166, opened 2022-09-21
(fix): Use venv created by pre-commit for validate config

enochtangg:
The `validate-configs-syntax` pre-commit hook should be using the venv created by pre-commit by default when specifying `language: python`. Calling `python3 ./snuba/validate_configs.py` does not import the current directory by default. This can be achieved by running the library module as a script. Thanks @asottile-sentry for the reference.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,8 +41,8 @@ repos:
     hooks:
       - id: validate-configs-syntax
         name: validate-configs-syntax
-        entry: python3 ./snuba/validate_configs.py
-        language: system
-        pass_filenames: false
+        entry: python3 -m snuba.validate_configs
+        language: python
+        additional_dependencies: [ 'jsonschema', 'pyyaml' ]
 default_language_version:
   python: python3.8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,7 @@ repos:
         entry: python3 -m snuba.validate_configs
         language: python
         additional_dependencies: [ 'jsonschema', 'pyyaml' ]
-        files: 'snuba/datasets/configuration/*'
+        pass_filenames: false
+        files: 'snuba/datasets/configuration/.*'
 default_language_version:
   python: python3.8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,5 +44,6 @@ repos:
         entry: python3 -m snuba.validate_configs
         language: python
         additional_dependencies: [ 'jsonschema', 'pyyaml' ]
+        files: 'snuba/datasets/configuration/*'
 default_language_version:
   python: python3.8


.pre-commit-config.yaml
```diff
@@ -41,8 +41,8 @@ repos:
     hooks:
       - id: validate-configs-syntax
         name: validate-configs-syntax
-        entry: python3 ./snuba/validate_configs.py
-        language: system
-        pass_filenames: false
+        entry: python3 -m snuba.validate_configs
+        language: python
+        additional_dependencies: [ 'jsonschema', 'pyyaml' ]
```
asottile-sentry, line 46: this should also set a `files` filter such that it's running on specific filenames -- and the validator itself should take specific filenames and only validate those files

enochtangg: Currently, the config files are recursed through a file path specified in a settings variable (https://github.com/getsentry/snuba/blob/5786877b22c8ce24b9cbc75142ace77fdf6a5410/snuba/validate_configs.py#L14).  Also, this validator is used in our CI via `make` command (https://github.com/getsentry/snuba/blob/347aab5c3f117cd3ab4114f347386037941f7290/.github/workflows/ci.yml#L100). Passing config filenames will require a different validation function, and the additional maintenance of where configs live. Would it be possible to leave this?

asottile-sentry: as it is right now it's going to always validate every config every time even when it's not changed (which on my machine adds about half a second when committing (even when committing unrelated things!)) -- maybe that overhead is ok but it's right on the borderline where I'd consider turning it off

evanh: That's a good idea, our other pre-commit hooks are only triggered when the relevant files are changed.

enochtangg: Ah I misunderstood the purpose of the `files` flag. Updated the pre-commit hook to only execute if files within `snuba/datasets/configuration` has been modified (includes configs, loaders, jsonschemas, etc.)

.pre-commit-config.yaml
```diff
@@ -41,8 +41,9 @@ repos:
...
+        files: 'snuba/datasets/configuration/*'
```
asottile-sentry, line 47: hmmmm this is supposed to trigger a warning that apparently isn't happening -- but anyway here's the warning message it's supposed to print:

```
[WARNING] The 'files' field in hook 'validate-configs-syntax' is a regex, not a glob -- matching '/*' probably isn't what you want here
```

.pre-commit-config.yaml
asottile-sentry: a further improvement would be to make the hook only check the files it is passed rather than all of them whenever triggered

as it is right now you probably need `pass_filenames: false` otherwise it's going to try and run it in parallel (`pass_filenames` is usually a hack, the right thing to do is to make it respond to just the filenames passed to it)

.pre-commit-config.yaml
enochtangg: How did you generate that warning? When switching to glob, I get the following error:
```An error has occurred: InvalidConfigError:
==> File .pre-commit-config.yaml
==> At Config()
==> At key: repos
==> At Repository(repo='local')
==> At key: hooks
==> At Hook(id='validate-configs-syntax')
==> At key: files
=====> 'snuba/datasets/configuration/**/*{.py,.yaml}' is not a valid python regex
```

Also tried with `snuba/datasets/configuration/**/*`

asottile-sentry: I copied it from the source (it's supposed to show up but current doesn't do to an oversight pre-commit/pre-commit#2521) -- the point is it's *not* a glob, it's a regex -- but you've written it like a glob

for example your pattern currently matches `snuba/datasets/configuration` and `snuba/datasets/configuration//////////////////////////////////////////////////////`

asottile-sentry (approved):
good enough improvement for now :+1:

volokluev (approved):

PR #2633, opened 2022-04-19
feat(profiling): Add index on profile_id

phacops:

PR #3169, opened 2022-09-21
feat: Run ClickHouse 21.8 in CI

lynnagara:
This aligns to the ClickHouse version we are using for the majority of tables in production.

It is needed for features like https://github.com/getsentry/snuba/pull/2633.


lynnagara:
Not yet

PR #3170, opened 2022-09-21
feat(slicing): Remove hardcoded generic metrics StorageSetKeys

ayirr7:
This PR:

- Removes the generics metrics StorageSetKeys from a hardcoded list that was present for the StorageSetKey Enum


evanh (approved):
This makes sense to me. I do think that we could probably get rid of the `StorageSet` class, which would remove the need for `register_storage_set_key` but that can be done separately.

PR #3171, opened 2022-09-21
ref(clickhouse-clusters): Add cluster_nodes to ClickhouseCluster

MeredithAnya:
I've been playing around with using the `clickhouse-setup` repo, (specifically this setup https://github.com/getsentry/clickhouse-setup/pull/4) to test out the migrations tool. This works well except for the fact that because the clickhouse nodes are running in containers, the migrations tool doesn't have access to the internal ports. And those are needed to in order to run the migrations on the nodes. 

When you connect to a clickhouse node through `docker exec -it` we can see that the `system.clusters` table doesn't give us those internal ports:

```sql
clickhouse-01 :) select * from system.clusters

SELECT *
FROM system.clusters

┌─cluster─────────────────────┬─shard_num─┬─shard_weight─┬─replica_num─┬─host_name────────────────┬─host_address─┬─port─┬─is_local─┬─user────┬─default_database─┬─errors_count─┬─estimated_recovery_time─┐
│ cluster_example_one         │         1 │            1 │           1 │ clickhouse-01            │ 172.18.0.5   │ 9000 │        1 │ default │                  │            0 │                       0 │
│ cluster_example_one         │         1 │            1 │           2 │ clickhouse-01-replica    │ 172.18.0.6   │ 9000 │        0 │ default │                  │            0 │                       0 │
│ cluster_example_one         │         2 │            1 │           1 │ clickhouse-02            │ 172.18.0.7   │ 9000 │        0 │ default │                  │            0 │                       0 │
│ cluster_migrations          │         1 │            1 │           1 │ clickhouse-migrations-01 │ 172.18.0.3   │ 9000 │        0 │ default │                  │            0 │                       0 │
│ test_shard_localhost        │         1 │            1 │           1 │ localhost                │ 127.0.0.1    │ 9000 │        1 │ default │                  │            0 │                       0 │
│ test_shard_localhost_secure │         1 │            1 │           1 │ localhost                │ 127.0.0.1    │ 9440 │        0 │ default │                  │            0 │                       0 │
└─────────────────────────────┴───────────┴──────────────┴─────────────┴──────────────────────────┴──────────────┴──────┴──────────┴─────────┴──────────────────┴──────────────┴─────────────────────────┘

6 rows in set. Elapsed: 0.005 sec.
```

This is why I've added an option `cluster_nodes` param to be able to specify what those ports are so that instead of querying the system tables, the nodes are predefined. This is not something that should be used in production, but would make it easier to test out with the `clickhouse-setup` repostitory. I've added a [not so finished `README.md`](https://github.com/getsentry/clickhouse-setup/blob/0779b1e17ef72d1312d5d4a624a48af5285244d2/examples/new_replica/README.md) to show how this flow would work.

### DIFF
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -208,6 +208,9 @@ class ClickhouseCluster(Cluster[ClickhouseWriterOptions]):
         distributed_cluster_name: Optional[str] = None,
         cache_partition_id: Optional[str] = None,
         query_settings_prefix: Optional[str] = None,
+        cluster_nodes: Optional[
+            Mapping[str, Sequence[Tuple[str, int, Optional[int], Optional[int]]]]
+        ] = None,
     ):
         super().__init__(storage_sets)
         self.__host = host
@@ -224,6 +227,7 @@ class ClickhouseCluster(Cluster[ClickhouseWriterOptions]):
         self.__connection_cache = connection_cache
         self.__cache_partition_id = cache_partition_id
         self.__query_settings_prefix = query_settings_prefix
+        self.__cluster_nodes = cluster_nodes
 
     def __str__(self) -> str:
         return str(self.__query_node)
@@ -334,6 +338,11 @@ class ClickhouseCluster(Cluster[ClickhouseWriterOptions]):
         )
 
     def __get_cluster_nodes(self, cluster_name: str) -> Sequence[ClickhouseNode]:
+        if self.__cluster_nodes and self.__cluster_nodes.get(cluster_name):
+            return [
+                ClickhouseNode(*node) for node in self.__cluster_nodes[cluster_name]
+            ]
+
         return [
             ClickhouseNode(*host)
             for host in self.get_query_connection(ClickhouseClientSettings.QUERY)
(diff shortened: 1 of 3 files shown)


nikhars (approved):
I am fine with these changes. Please address the comment about documentation

fpacifici (commented):
I am assuming this is meant to allow us to make the migration tool access each node when we run it in a docker compose file. 
Can't we solve the problem by using a real google cloud cluster if this is just for tests or by running the snuba migration tool from a container inside the docker compose ?
They feel simpler option that would not break Snuba abstractions

fpacifici:
Actually no, ignore the gcloud part of my comment. You would have to configure firewalls and networking to allow you to run migrations from your laptop against the gcloud cluster. 
Though the idea of running the script/admin in a docker container inside the docker compose bundle remains. 

Anyway, if we merge this can we consider it temporary for this project with the goal of reverting it asap ?

MeredithAnya:
NOTE: I haven't merged this in because I think I might have found another way to test this functionality without introducing this PR, BUT since I don't have the finalized I have been using this branch for some local testing

MeredithAnya:
No longer need now that https://github.com/getsentry/clickhouse-setup/blob/master/examples/new_replica/README.md exists

PR #3172, opened 2022-09-21
feat(slices): Add a mechanism to map queries to correct clusters

nikhars:
### Legal Boilerplate


lynnagara:
This is adding quite a bit of code generic metrics entity definition for the slice. Are you sure about this approach? It means product engineers have to now care about slicing considerations. I think it would be simpler if that decision was made entirely by operators via settings overrides, and entity definitions do not concern themselves with this.

nikhars:
> This is adding quite a bit of code generic metrics entity definition for the slice. Are you sure about this approach? It means product engineers have to now care about slicing considerations. I think it would be simpler if that decision was made entirely by operators via settings overrides, and entity definitions do not have any knowledge or concern about this.

Let me break down the problem from an API perspective. An API request comes in for a dataset(which is unpartitioned). We need to make it to a storage key (which is unpartitioned). Storage keys have a 1-1 mapping with a storage set key in existing code. Slicing will introduce a 1-n mapping from a storage key to storage set. We need to find a solution on how that mapping happens. The mapping from a single storage key to a storage set key (based on org_id) is problem I am trying to solve. 

A few additional details on the thought process behind the solution being proposed.
- The logic of mapping from a storage key to storage set key does not have to happen for all storages. Only generic metrics needs to do that. Hence the introduction of `PartitionedStorageQueryPlanBuilder`.
- Now we need a mechanism to selecting the storage_set key within `PartitionedStorageQueryPlanBuilder`. That is why the ABC `StoragePartitionSelector` whose responsibility it is to provide a single storage set key and cluster which can be used down the line.

onewland:
> This is adding quite a bit of code generic metrics entity definition for the slice. Are you sure about this approach? It means product engineers have to now care about slicing considerations. I think it would be simpler if that decision was made entirely by operators via settings overrides, and entity definitions do not have any knowledge or concern about this.

Also note that it's part of my responsibility to move this to the generic_metrics configuration (along with support for this new query plan builder to YAML-based datasets)

snuba/datasets/storage.py
```diff
@@ -157,6 +158,45 @@ def get_mandatory_condition_checkers(self) -> Sequence[ConditionChecker]:
         return self.__mandatory_condition_checkers
 
 
+class PartitionedReadableTableStorage(ReadableStorage):
```
onewland, line 161: I'm not sure I want us to introduce this class, I see that it's for safety so that `get_cluster()` can't be used improperly, but IMO we should just add the `partition_id: Optional[int] = None` parameter to the parent class and ignore it almost all the time

snuba/datasets/entities/generic_metrics.py
```diff
@@ -205,3 +223,32 @@ def __init__(self) -> None:
                 ],
             ),
         )
+
+
+class GenericMetricsStoragePartitionSelector(StoragePartitionSelector):
```
onewland, line 228: I think this could be a little more generalized, and not specific to generic_metrics (which will be useful for me when I move this to config). 

Let's move the name "org_id" to be a constructor argument (`partition_key_column_name`) which we'll store on the object and then we can call this a `ColumnBasedStoragePartitionSelector` and put it somewhere less "generic metrics" specific. It could maybe even be in `snuba/datasets/plans/partitioned_storage.py`

snuba/datasets/plans/partitioned_storage.py
```diff
@@ -0,0 +1,144 @@
...
+
+class StoragePartitionSelector(ABC):
+    """
+    The component provided by a dataset and used at the beginning of the
+    execution of a query to pick the storage set a query should be executed
+    onto.
+    """
+
+    @abstractmethod
+    def select_storage(
```
onewland, line 43: following the above, maybe this should be `select_cluster`?

snuba/datasets/plans/partitioned_storage.py
```diff
@@ -0,0 +1,144 @@
...
+    MandatoryConditionEnforcer,
+)
+from snuba.query.processors.physical.mandatory_condition_applier import (
+    MandatoryConditionApplier,
+)
+from snuba.query.query_settings import QuerySettings
+from snuba.util import with_span
+
+
+class StoragePartitionSelector(ABC):
```
onewland, line 35: Should this maybe be called StorageClusterSelector?

snuba/datasets/partitioning.py
```diff
@@ -3,4 +3,21 @@
...
+    fixed, an org id will always be mapped to the same logical partition.
+    """
+    return org_id % SENTRY_LOGICAL_PARTITIONS
+
+
+def map_logical_partition_to_physical_partition(logical_partition: int) -> int:
+    """
+    Maps a logical partition to a physical partition.
+    """
+    return LOGICAL_PARTITION_MAPPING[str(logical_partition)]
```
onewland, line 23: I think we will change the type here to int so you don't have to do this `str` conversion, but that can be outside the scope of this PR

onewland (approved):
Looks like a pretty clean approach to me, thanks for following all the twists/turns of the design discussion

snuba/datasets/plans/partitioned_storage.py
```diff
@@ -0,0 +1,153 @@
...
+        self.partition_key_column_name = partition_key_column_name
+
+    def select_cluster(
+        self, query: LogicalQuery, query_settings: QuerySettings
+    ) -> ClickhouseCluster:
+        """
+        Selects the cluster to use for a query if the storage set is partitioned.
+        If the storage set is not partitioned, it returns the default cluster.
+        """
+        if not is_storage_set_partitioned(self.storage_set):
```
nikhars, line 71: @onewland - A storage may or may not be partitioned based on the environment we are running in (SaaS, ST, Self hosted). I have used https://github.com/getsentry/snuba/pull/3172/files#diff-4242bc3893c530314f2c6131401b0da4629e90cd6a24783a09273befb003c7daR28-R34 to make that decision. But I am not sure if that is the correct way to find out.

snuba/datasets/plans/partitioned_storage.py
```diff
@@ -0,0 +1,153 @@
...
+
+        org_ids = get_object_ids_in_query_ast(query, self.partition_key_column_name)
+        assert org_ids is not None
+        assert len(org_ids) == 1
+        org_id = org_ids.pop()
+
+        physical_partition = map_logical_partition_to_physical_partition(
+            map_org_id_to_logical_partition(org_id)
+        )
+        cluster = get_cluster(self.storage_set, physical_partition)
```
nikhars, line 82: @onewland - When a query needs to span across clickhouse clusters, we need a way to find that out. Now if we are going to build a different storage set for the cross clickhouse cluster, things might start to get messy over here. Query comes for a dataset A. Dataset A is bound to storage set A. Processing decides it has to go across clusters. Change Storage set to B. That means modify ColumnBasedStoragePartitionSelector to take 2 storage sets. Which would mean when defining the dataset entities there would most likely be circluar imports. 

My suggestion is to have a config which can say whether to go across clusters or not but use the same storage set.

nikhars: Thinking about this a bit more, it might be ok. But this is something to be aware of

onewland: I was sort of assuming that this call to `cluster = get_cluster(storage_set, slice_id)` would be changed when we introduced the run-time flag to

```
if override_cluster_exists(storage_set, logical_partition):
   cluster = get_cluster(get_override_cluster_storage_set_key(storage_set, slice_id))
else:
   cluster = get_cluster(storage_set, slice_id)
```

does that seem like a bad idea to you for some reason?

snuba/datasets/plans/partitioned_storage.py
onewland: I think that's right, thanks.

snuba/datasets/plans/partitioned_storage.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Storage partition selector for the generic metrics storage. This is needed
+    because the generic metrics storage can be partitioned and we would need to
+    know which partition to use for a specific query.
+    """
+
+    def __init__(
+        self,
+        storage: StorageKey,
+        storage_set: StorageSetKey,
```
onewland, line 62: Both these values are derivable from the ReadableTableStorage which I'd be happy to take as a parameter if others prefer that

PR #3173, opened 2022-09-21
feat(profiling): Handle sample format

phacops:
We have a new format for profiles and we need to map fields slightly differently than the usual format we handle. We detect this new format by looking if a field called `version` is set in the message.

### DIFF
--- a/snuba/datasets/profiles_processor.py
+++ b/snuba/datasets/profiles_processor.py
@@ -4,6 +4,7 @@ from uuid import UUID
 
 from snuba import environment
 from snuba.consumers.types import KafkaMessageMetadata
+from snuba.datasets.event_format import EventTooOld, enforce_retention
 from snuba.processor import InsertBatch, MessageProcessor, ProcessedMessage
 from snuba.utils.metrics.wrapper import MetricsWrapper
 
@@ -15,10 +16,22 @@ class ProfilesMessageProcessor(MessageProcessor):
         self, message: Mapping[str, Any], metadata: KafkaMessageMetadata
     ) -> Optional[ProcessedMessage]:
         try:
+            received = datetime.fromutctimestamp(message["received"])
+            retention_days = enforce_retention(
+                message["project_id"], message["retention_days"], received
+            )
+
             if "version" in message:
-                processed = _normalize_sample_format(message, metadata)
+                processed = _normalize_sample_format(
+                    message, metadata, retention_days, received
+                )
             else:
-                processed = _normalize_legacy_format(message, metadata)
+                processed = _normalize_legacy_format(
+                    message, metadata, retention_days, received
+                )
+        except EventTooOld:
+            metrics.increment("event_too_old")
+            return None
         except IndexError:
             metrics.increment("invalid_transaction")
             return None
@@ -32,7 +45,10 @@ class ProfilesMessageProcessor(MessageProcessor):
 
 
 def _normalize_legacy_format(
-    message: Mapping[str, Any], metadata: KafkaMessageMetadata
+    message: Mapping[str, Any],
+    metadata: KafkaMessageMetadata,
+    retention_days: int,
+    received: datetime,
 ) -> Mapping[str, Any]:
     return {
         "android_api_level": message.get("android_api_level"),
@@ -54,7 +70,7 @@ def _normalize_legacy_format(
         "profile_id": str(UUID(message["profile_id"])),
         "project_id": message["project_id"],
         "received": datetime.utcfromtimestamp(message["received"]),
-        "retention_days": message["retention_days"],
+        "retention_days": retention_days,
         "trace_id": str(UUID(message["trace_id"])),
         "transaction_id": str(UUID(message["transaction_id"])),
         "transaction_name": message["transaction_name"],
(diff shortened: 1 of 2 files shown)


evanh (commented):
I'm not sure why there was no test added for this.

Also, I would recommend wrapping this new code in a try-except and log an error, just while you are testing it out. Otherwise your processor will start crash looping.

phacops:
The `try`/`except` is already there and we're comfortable with the level of logging.

tests/datasets/test_profiles_processor.py
```diff
@@ -41,13 +92,57 @@ def serialize(self) -> Mapping[str, Any]:
...
         result = asdict(self)
+        result["profile"] = ""
         result["received"] = datetime.utcfromtimestamp(self.received)
         result["offset"] = meta.offset
         result["partition"] = meta.partition
         return result
 
 
 class TestProfilesProcessor:
+    def test_sample_profile_message(self) -> None:
```
volokluev, line 103: this is only testing messages with a version field. What if there is no version field?

phacops: A message with no version field would be the current message we process and it's tested here: https://github.com/getsentry/snuba/pull/3173/files#diff-4ced175fc87fe811e95d6e65609911fb917d0c2389691f91190968e9f531076fR181.

snuba/datasets/profiles_processor.py
```diff
@@ -9,44 +9,80 @@
...
-                "version_code": message["version_code"],
-                "retention_days": retention_days,
                 "offset": metadata.offset,
+                "organization_id": message["organization_id"],
                 "partition": metadata.partition,
+                "platform": message["platform"],
+                "profile": "",  # deprecated
+                "project_id": message["project_id"],
+                "received": datetime.utcfromtimestamp(message["received"]),
+                "retention_days": message["retention_days"],
```
MeredithAnya, line 79: is there a reason that the retention days aren't enforced to be `30` or `90` anymore?

snuba/datasets/profiles_processor.py
```diff
@@ -9,44 +9,80 @@
...
+                )
+                additional_fields["version_name"] = message["release"]
+
+                for field in ["name", "version", "build_number"]:
+                    additional_fields[f"device_os_{field}"] = message["os"].get(field)
+
+                transaction = message["transactions"][0]
+                additional_fields.update(
+                    {
+                        "transaction_id": transaction["id"],
```
Zylphrex, line 39: Do we need to format this with
```suggestion
                        "transaction_id": str(UUID(transaction["id"])),
```

snuba/datasets/profiles_processor.py
```diff
@@ -9,44 +9,80 @@
...
+                        "transaction_name": transaction["name"],
+                        "trace_id": transaction["trace_id"],
```
Zylphrex, line 41: Same formatting here
```suggestion
                        "trace_id": str(UUID(transaction["trace_id"])),
```

snuba/datasets/profiles_processor.py
```diff
@@ -9,44 +9,80 @@
...
-                "transaction_id": str(UUID(message["transaction_id"])),
-                "profile_id": str(UUID(message["profile_id"])),
-                "received": datetime.utcfromtimestamp(message["received"]),
-                "profile": message["profile"],
                 "android_api_level": message.get("android_api_level"),
-                "device_classification": message["device_classification"],
-                "device_locale": message["device_locale"],
-                "device_manufacturer": message["device_manufacturer"],
-                "device_model": message["device_model"],
                 "device_os_build_number": message.get("device_os_build_number"),
```
Zylphrex, line 70: I think you'll want to move `device_os_build_number` into the else block since the sampled format overwrites this value.

snuba/datasets/profiles_processor.py
phacops: We pull the value from https://github.com/getsentry/sentry/blob/master/src/sentry/profiles/task.py#L148 now and it didn't seem useful to have some guard here to switch between 30 or 90, purely because we might end up with a different value at some point (like 60). Before, we were kind of setting it ourselves and it was definitely a good idea to guard it.

snuba/datasets/profiles_processor.py
phacops: I mean sure, it'll be the 3rd time in the pipeline we format it but you know.

volokluev (approved):

snuba/datasets/profiles_processor.py
```diff
@@ -9,48 +9,92 @@
...
-
 
 class ProfilesMessageProcessor(MessageProcessor):
     def process_message(
         self, message: Mapping[str, Any], metadata: KafkaMessageMetadata
     ) -> Optional[ProcessedMessage]:
         try:
-            retention_days = message["retention_days"]
-            if retention_days not in RETENTION_DAYS_ALLOWED:
-                retention_days = 30
```
nikhars, line 22: Please do not remove this validation. If you allow users to drive the retention, the work of TTL merges on Clickhouse would increase since each partition could have parts which might require TTL merge. That increases load on you cluster unnecessarily. I would advise against doing so.

nikhars (changes requested):

snuba/datasets/profiles_processor.py
phacops: Sure, I'll add it back. It's still not users driving the retention but Sentry via the quota configured by their plan. This logic is just too simple.

I'll use whatever you use in transactions instead.

nikhars (approved):

PR #3174, opened 2022-09-21
ref(migrations): Add copy_tables command

MeredithAnya:
**context:**
We have some ClickHouse [migrations tooling](https://github.com/getsentry/snuba/blob/master/MIGRATIONS.md#running-migrations) that currently used to `run` and `reverse` migrations in addition to `list`-ing the current status of migrations. This is great for existing datasets where we already have the clusters set up how we want. However, if we want to add a replica or a shard to an existing cluster we cannot easy do so.

**`add_node` today:**
There is some prior art which is the current [add_node](https://github.com/getsentry/snuba/blob/master/MIGRATIONS.md#adding-a-new-node) command. This basically runs all the migrations in order for a specific storage set on a new node. The end result is hopefully that the node is now in the same state as all the other nodes in the clusters. This works only in the case where the initial `CREATE TABLE` statement doesn't differ from what the current schema is. 

**why it doesn't quite work:**
ClickHouse requires the table schema to be same in order to replicate any data. If we have a table that has gone through many schema iterations, the moment we try to run the `0001_add_table...` migration, the table on the new node is in a different state than the other nodes in the cluster. When we try to grab parts from ZooKeeper to replicate data we will get some sort of error. 

For example, in the case below, the first migration for the `generic_metrics` actually has multiple SQL operations, first `CREATE TABLE`, then `ADD COLUMN` (and then some other add columns/add indexes). But right after we run the first `CREATE TABLE`, we get an error about the `_indexed_tags_hash` missing, which makes sense because it hasn't been added yet (that was the next SQL statement). 
```bash
15:41 $ snuba migrations add-node --type=local --storage-set=generic_metrics_sets --port=9051
migration <snuba.snuba_migrations.generic_metrics.0001_sets_aggregate_table.Migration object at 0x10f8d8790>
  File "/Users/meredith/snuba/snuba/clickhouse/native.py", line 192, in execute
    result_data = query_execute()
  File "/Users/meredith/snuba/.venv/lib/python3.8/site-packages/clickhouse_driver/client.py", line 304, in execute
    rv = self.process_ordinary_query(
  File "/Users/meredith/snuba/.venv/lib/python3.8/site-packages/clickhouse_driver/client.py", line 491, in process_ordinary_query
    return self.receive_result(with_column_types=with_column_types,
  File "/Users/meredith/snuba/.venv/lib/python3.8/site-packages/clickhouse_driver/client.py", line 151, in receive_result
    return result.get_result()
  File "/Users/meredith/snuba/.venv/lib/python3.8/site-packages/clickhouse_driver/result.py", line 50, in get_result
    for packet in self.packet_generator:
  File "/Users/meredith/snuba/.venv/lib/python3.8/site-packages/clickhouse_driver/client.py", line 167, in packet_generator
    packet = self.receive_packet()
  File "/Users/meredith/snuba/.venv/lib/python3.8/site-packages/clickhouse_driver/client.py", line 184, in receive_packet
    raise packet.exception
clickhouse_driver.errors.ServerException: Code: 47.
DB::Exception: Missing columns: '_indexed_tags_hash' while processing query: '_indexed_tags_hash', required columns: '_indexed_tags_hash', source columns: 'value' 'tags.raw_value' 'use_case_id' 'tags.key' 'tags.indexed_value' 'org_id' 'project_id' 'timestamp' 'retention_days' 'granularity' 'metric_id'. Stack trace:
```

**solution**
We don't really care about running the migrations in order, since as far as the cluster in concerned, the migrations are up to date (its just the new node that's not). Instead we can copy the current table statement, using `SHOW CREATE TABLE`, and the use that to create the tables on the new node. This way the schema should be identical to all the other nodes in the cluster, preventing any errors of missing columns, missing indexes etc. 

In order to do this we need:
* access a current node in the cluster to run the `SHOW CREATE TABLE` command
* the table names that we actually want to copy over 
* access to the new node to run the collected table statements.

In this PR I've added the option to either pass in the list of table names, or for the command to do that automatically by running the `SHOW TABLE` command on the node we are copying the table statements from.

### DIFF
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -436,7 +436,7 @@ class Runner:
                     python_op.execute_new_node(storage_sets)
 
     @classmethod
-    def create_tables(
+    def copy_tables(
         self,
         node_type: ClickhouseNodeType,
         host_name: str,
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -434,3 +434,69 @@ class Runner:
             elif isinstance(migration, CodeMigration):
                 for python_op in migration.forwards_global():
                     python_op.execute_new_node(storage_sets)
+
+    @classmethod
+    def create_tables(
+        self,
+        node_type: ClickhouseNodeType,
+        host_name: str,
+        port: int,
+        user: str,
+        password: str,
+        database: str,
+        new_port: int,
+        table_names: Optional[str],
+        dry_run: bool = False,
+    ) -> None:
+        """
+        Create tables on a new node in a cluster by using the existing create table
+        statement from an existing node in a cluster. This can be done for a new shard
+        or new replica.
+        """
+        client_settings = ClickhouseClientSettings.QUERY.value
+        clickhouse = ClickhousePool(
+            host_name,
+            port,
+            user,
+            password,
+            database,
+            client_settings=client_settings.settings,
+            send_receive_timeout=client_settings.timeout,
+        )
+
+        # TODO: assert that the new node is not in the system.clusters definitions yet
+
+        if table_names:
+            ch_table_names = table_names.split(",")
+        else:
+            ch_table_names = [
+                result[0] for result in clickhouse.execute("SHOW TABLES").results
+            ]
+
+        show_table_statments = {}
+
+        for name in ch_table_names:
+            ((curr_create_table_statement,),) = clickhouse.execute(
+                f"SHOW CREATE TABLE {database}.{name}"
+            ).results
+            show_table_statments[name] = curr_create_table_statement
+
+        new_client_settings = ClickhouseClientSettings.MIGRATE.value
+        new_clickhouse_node = ClickhousePool(
+            host_name,
+            new_port,
+            user,
+            password,
+            database,
+            client_settings=new_client_settings.settings,
+            send_receive_timeout=new_client_settings.timeout,
+        )
+
+        # TODO: print out the system.macros for the new node
+
+        for table_name, statement in show_table_statments.items():
+            print(f"creating {table_name}...")
+            if dry_run:
+                print(f"create table statement: \n {statement}")
+                return
+            new_clickhouse_node.execute(statement)


snuba/migrations/runner.py
```diff
@@ -418,3 +418,69 @@ def add_node(
             elif isinstance(migration, CodeMigration):
                 for python_op in migration.forwards_global():
                     python_op.execute_new_node(storage_sets)
+
+    @classmethod
+    def create_tables(
```
lynnagara, line 423: Can we call this copy_tables? To ensure there is no confusion about when to use and to keep it more distinct from the existing create table operation.

snuba/migrations/runner.py
```diff
@@ -434,3 +434,69 @@ def add_node(
...
+    @classmethod
+    def copy_tables(
+        self,
+        node_type: ClickhouseNodeType,
+        host_name: str,
+        port: int,
+        user: str,
+        password: str,
+        database: str,
+        new_port: int,
```
fpacifici, line 447: What is this for ?

snuba/migrations/runner.py
```diff
@@ -434,3 +434,69 @@ def add_node(
...
+            host_name,
+            port,
+            user,
+            password,
+            database,
+            client_settings=client_settings.settings,
+            send_receive_timeout=client_settings.timeout,
+        )
+
+        # TODO: assert that the new node is not in the system.clusters definitions yet
```
fpacifici, line 467: Did we discuss this before? Not sure why we want to enforce this. 
Did you discuss this process with @nikhars  and @mwarkentin ? 
We want to make the addition of a new node quick, that means having as few manual steps as possible. adding the tables before adding the node to the cluster implies at least three steps: provision node, add tables, then add the node to the cluster. 

I think we should ensure we are on the same page on the quickest process to add a node.

snuba/migrations/runner.py
```diff
@@ -434,3 +434,69 @@ def add_node(
...
+
+        if table_names:
+            ch_table_names = table_names.split(",")
+        else:
+            ch_table_names = [
+                result[0] for result in clickhouse.execute("SHOW TABLES").results
+            ]
```
fpacifici, line 474: I think this is going to be a bit fragile. If we trust the content of the tables on another node we should at least validate they are the same you have in the storage set you are trying to deploy on this node. 
More in general, how do you expect this script to be used? Is the user going to tell "please duplicate this node" or is the user going to tell "Please create all the tables you should have on this node according to the storage sets on this cluster" ?

snuba/migrations/runner.py
```diff
@@ -434,3 +434,69 @@ def add_node(
...
+        for name in ch_table_names:
+            ((curr_create_table_statement,),) = clickhouse.execute(
+                f"SHOW CREATE TABLE {database}.{name}"
+            ).results
+            show_table_statments[name] = curr_create_table_statement
+
+        new_client_settings = ClickhouseClientSettings.MIGRATE.value
+        new_clickhouse_node = ClickhousePool(
+            host_name,
+            new_port,
```
fpacifici, line 487: Why are we assuming the host name is going to be the same? In most cases this is not going to be true.

MeredithAnya: happened to be true for testing, but yes I should update to have a `new_host_name` param as well

MeredithAnya:
Closing in favor of https://github.com/getsentry/snuba/pull/3241, we may need to come back and add some cli command utility but we can do that in a new PR when the time comes

PR #3179, opened 2022-09-22
feat(slicing): Map StorageSetKey, partition pairs to Clusters

ayirr7:
This PR:

- Attempts to map (storage set, partition) pairs to separate clusters, in settings. This approach came up from discussion in another PR (https://github.com/getsentry/snuba/pull/3149) which attempted to partition via "override" StorageSetKeys
- Also integrates the changes from https://github.com/getsentry/snuba/pull/3170 (remove hardcoded generics metrics storage set keys). Once PR 3170 is merged, the same/copied changes will be removed from this PR.

### DIFF
--- a/docs/source/architecture/partitioning.rst
+++ b/docs/source/architecture/partitioning.rst
@@ -30,6 +30,14 @@ Every logical partition **must** be assigned to a physical partition and the
 valid values of physical partitions are in the range
 of ``[0,settings.LOCAL_PHYSICAL_PARTITIONS)``.
 
+Defining partitioned clusters
+--------------------------------
+To add a cluster with an associated (storage set key, partition) pair, add cluster definitions
+to ``settings.PARTITIONED_CLUSTERS`` in the desired environment's settings. Follow the same structure as
+regular cluster definitions in ``settings.CLUSTERS``. In the ``storage_sets`` field, partitioned storage
+sets should be added in the form of ``(StorageSetKey, partition_id)`` where partition_id is in
+the range ``[0,settings.LOCAL_PHYSICAL_PARTITIONS)``.
+
 
 TODO: adding storages, migrating subscriptions, etc.
 ----------------------------------------------------
--- a/docs/source/architecture/partitioning.rst
+++ b/docs/source/architecture/partitioning.rst
@@ -34,7 +34,7 @@ Defining partitioned clusters
 --------------------------------
 To add a cluster with an associated (storage set key, partition) pair, add cluster definitions
 to ``settings.PARTITIONED_CLUSTERS`` in the desired environment's settings. Follow the same structure as
-regular cluster definitions in ``settings.CLUSTERS``. In the ``storage_sets`` field, partitioned storage
+regular cluster definitions in ``settings.CLUSTERS``. In the ``storage_set_partitions`` field, partitioned storage
 sets should be added in the form of ``(StorageSetKey, partition_id)`` where partition_id is in
 the range ``[0,settings.LOCAL_PHYSICAL_PARTITIONS)``.
 
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -447,12 +447,16 @@ def get_cluster(
     if partition_id:
         part_storage_set_cluster_map = _get_partitioned_storage_set_cluster_map()
         res = part_storage_set_cluster_map.get((storage_set_key, partition_id), None)
+        if res is None:
+            raise UndefinedClickhouseCluster(
+                f"{(storage_set_key, partition_id)} is not defined in the PARTITIONED_CLUSTERS setting for this environment"
+            )
 
     else:
         storage_set_cluster_map = _get_storage_set_cluster_map()
         res = storage_set_cluster_map.get(storage_set_key, None)
         if res is None:
             raise UndefinedClickhouseCluster(
-                f"{storage_set_key} is not a defined in the CLUSTERS setting for this environment"
+                f"{storage_set_key} is not defined in the CLUSTERS setting for this environment"
             )
     return res
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -403,7 +403,9 @@ def _build_partitioned_cluster(cluster: Mapping[str, Any]) -> ClickhouseCluster:
         password=cluster.get("password", ""),
         database=cluster.get("database", "default"),
         http_port=cluster["http_port"],
-        storage_sets={storage_tuple[0] for storage_tuple in cluster["storage_sets"]},
+        storage_sets={
+            storage_tuple[0] for storage_tuple in cluster["storage_set_partitions"]
+        },
         single_node=cluster["single_node"],
         cluster_name=cluster["cluster_name"] if "cluster_name" in cluster else None,
         distributed_cluster_name=cluster["distributed_cluster_name"]
@@ -424,7 +426,7 @@ def _get_partitioned_storage_set_cluster_map() -> Dict[
 ]:
     if len(_PARTITIONED_STORAGE_SET_CLUSTER_MAP) == 0:
         for cluster in settings.PARTITIONED_CLUSTERS:
-            for storage_set_tuple in cluster["storage_sets"]:
+            for storage_set_tuple in cluster["storage_set_partitions"]:
                 _PARTITIONED_STORAGE_SET_CLUSTER_MAP[
                     (StorageSetKey(storage_set_tuple[0]), storage_set_tuple[1])
                 ] = _build_partitioned_cluster(cluster)
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -443,8 +443,15 @@ def get_cluster(
 ) -> ClickhouseCluster:
     """Return a clickhouse cluster for a storage set key.
 
-    If the storage set key is not defined
-    in the CLUSTERS config, it will raise an UndefinedClickhouseCluster Exception.
+    If passing in a partitioned storage set, a slice_id must be specified.
+    This ID will be used to return the matching cluster in PARTITIONED_CLUSTERS.
+    If passing in an unpartitioned storage set, a slice_id should not be
+    specified. The StorageSetKey will be used to return the matching
+    cluster in CLUSTERS.
+
+    If the storage set key is not defined either in CLUSTERS or in
+    PARTITIONED_CLUSTERS, then an UndefinedClickhouseCluster Exception
+    will be raised.
     """
     assert (
         storage_set_key not in DEV_STORAGE_SETS or settings.ENABLE_DEV_FEATURES
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -105,18 +105,7 @@ class Cluster(ABC, Generic[TWriterOptions]):
         self.__storage_sets = storage_sets
 
     def get_storage_set_keys(self) -> Set[StorageSetKey]:
-        all_storage_sets = set(key.value for key in StorageSetKey)
-
-        storage_set_keys = set()
-
-        for storage_set in self.__storage_sets:
-            # We ignore invalid storage set keys since new storage sets will
-            # need to be registered to configuration before they can be used
-            # in Snuba.
-            if storage_set in all_storage_sets:
-                storage_set_keys.add(StorageSetKey(storage_set))
-
-        return storage_set_keys
+        return {StorageSetKey(storage_set) for storage_set in self.__storage_sets}
 
     @abstractmethod
     def get_reader(self) -> Reader:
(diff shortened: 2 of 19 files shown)


snuba/clusters/cluster.py
```diff
@@ -399,11 +395,46 @@ def _get_storage_set_cluster_map() -> Dict[StorageSetKey, ClickhouseCluster]:
     return _STORAGE_SET_CLUSTER_MAP
 
 
+def _build_partitioned_cluster(cluster: Mapping[str, Any]) -> ClickhouseCluster:
+
```
onewland, line 399: nit: unnecessary newline

tests/clusters/test_cluster.py
```diff
@@ -196,3 +219,15 @@ def test_cache_connections() -> None:
     assert cluster_1.get_query_connection(
         cluster.ClickhouseClientSettings.QUERY
     ) != cluster_3.get_query_connection(cluster.ClickhouseClientSettings.QUERY)
+
+
+@patch("snuba.settings.PARTITIONED_CLUSTERS", PARTITIONED_CLUSTERS_CONFIG)
+def test_partitioned_cluster() -> None:
```
onewland, line 225: can we also test that `get_cluster` on partition 0 returns the other cluster values? (it can be in the same method)

snuba/clusters/cluster.py
```diff
@@ -399,11 +395,46 @@ def _get_storage_set_cluster_map() -> Dict[StorageSetKey, ClickhouseCluster]:
...
+        else None,
+        cache_partition_id=cluster.get("cache_partition_id"),
+        query_settings_prefix=cluster.get("query_settings_prefix"),
+    )
+
+
+def _get_partitioned_storage_set_cluster_map() -> Dict[
+    Tuple[StorageSetKey, int], ClickhouseCluster
+]:
+    _PARTITIONED_STORAGE_SET_CLUSTER_MAP = {}
```
onewland, line 421: We should probably only build this map once (e.g `_PARTITIONED_STORAGE_SET_CLUSTER_MAP` should be a global and we only assign it if it's equal to empty map at the start of the function call).

onewland (approved):
This is fine with me if it works for Lyn. If so, let's close out that other PR

onewland:
> This PR:
> 
> * Also integrates the changes from [feat(slicing): Remove hardcoded generic metrics StorageSetKeys #3170](https://github.com/getsentry/snuba/pull/3170) (remove hardcoded generics metrics storage set keys). Once PR 3170 is merged, the same/copied changes will be removed from this PR.

Note that you can change the base branch of PRs so people can focus on just the changeset you care about

onewland:
Also, assuming we get approval from Lyn, let's see what we should update in https://github.com/getsentry/snuba/blob/master/docs/source/architecture/partitioning.rst

snuba/datasets/storage.py
```diff
@@ -39,8 +39,8 @@ def __init__(self, storage_set_key: StorageSetKey, schema: Schema):
     def get_storage_set_key(self) -> StorageSetKey:
         return self.__storage_set_key
 
-    def get_cluster(self) -> ClickhouseCluster:
-        return get_cluster(self.__storage_set_key)
+    def get_cluster(self, partition_id: Optional[int] = None) -> ClickhouseCluster:
```
nikhars, line 42: While theoretically speaking this works, from a logical perspective it is not useful since we are not partitioning storages. What we are partitioning is storage sets in Snuba. So a storage would never have any idea about partitioning. So I would suggest we remove it from here.

snuba/clusters/cluster.py
```diff
@@ -408,9 +443,20 @@ def get_cluster(storage_set_key: StorageSetKey) -> ClickhouseCluster:
     assert (
         storage_set_key not in DEV_STORAGE_SETS or settings.ENABLE_DEV_FEATURES
     ), f"Storage set {storage_set_key} is disabled"
-    res = _get_storage_set_cluster_map().get(storage_set_key, None)
-    if res is None:
-        raise UndefinedClickhouseCluster(
-            f"{storage_set_key} is not a defined in the CLUSTERS setting for this environment"
-        )
+
+    if partition_id:
```
nikhars, line 447: If a storage set is partitioned, it would provide a different storage set key with a suffix like `_1` without passing the partition_id. Will that behavior still work with this change?

onewland: We're not using storage set key alone to route with this approach.

A single storage set can map to multiple clusters based on the (storage set key, partition ID) pair. If we go with this approach, you'll want your mapping system to pass partition_id here, there won't be a re-routing to another storage set key.

snuba/datasets/storage.py
ayirr7: Sure, ok that makes sense

tests/clusters/test_cluster.py
ayirr7: Done

snuba/clusters/cluster.py
```diff
@@ -395,11 +395,50 @@ def _get_storage_set_cluster_map() -> Dict[StorageSetKey, ClickhouseCluster]:
...
+    return _PARTITIONED_STORAGE_SET_CLUSTER_MAP
+
+
 class UndefinedClickhouseCluster(SerializableException):
     pass
 
 
-def get_cluster(storage_set_key: StorageSetKey) -> ClickhouseCluster:
+def get_cluster(
+    storage_set_key: StorageSetKey, partition_id: Optional[int] = None
```
lynnagara, line 440: How is this function supposed to be used? Please add a docstring as it's imported in quite a few places. Is None is supposed to be passed only for unpartitioned, and an int for partitioned? Since they are handled completely different you may want to consider splitting it into 2 functions instead.

Another idea I'll just throw out there would be to make partition_id required, and treat everything as partitioned. An unpartitioned storage set is nothing more than a storage set with 1 partition as far as most of Snuba is concerned.

lynnagara: We'll probably also need a `get_clusters(storage_set_key)` to retrieve them all in this module. There's a bunch of pieces of code (e.g. migrations, admin, etc) that will likely need to fetch all the clusters of a storage set and iterate through them.

snuba/settings/__init__.py
```diff
@@ -40,6 +40,8 @@
 # Storage names to apply dataset partitioning to
 PARTITIONED_STORAGES: Set[str] = set()
 
+PARTITIONED_CLUSTERS: Sequence[Mapping[str, Any]] = []
```
lynnagara, line 43: Please provide an example of how to use in comments above this line.

tests/clusters/test_cluster.py
```diff
@@ -79,6 +79,29 @@
...
 
+PARTITIONED_CLUSTERS_CONFIG = [
+    {
+        "host": "host_slice",
+        "port": 9000,
+        "user": "default",
+        "password": "",
+        "database": "default",
+        "http_port": 8123,
+        "storage_sets": {("generic_metrics_distributions", 0)},
```
lynnagara, line 90: nit: I'd call this `storage_set_partitions` or something

snuba/settings/__init__.py
lynnagara: Also, can you please clarify (in this file ideally) how CLUSTERS and PARTITIONED_CLUSTERS interacts

lynnagara: It's a nitpick i know but can you put PARTITIONED_CLUSTERS below CLUSTERS in this file

lynnagara:
This approach is generally fine to me. The biggest open question I think is how this would fit with the query path ideas laid out here https://github.com/getsentry/snuba/pull/3172

tests/clusters/test_cluster.py
onewland: +1 to this suggestion

snuba/clusters/cluster.py
ayirr7: I like the idea of making partition id required. I'll add it and see what the others also think

onewland: > Another idea I'll just throw out there would be to make partition_id required, and treat everything as partitioned. An unpartitioned storage set is nothing more than a storage set with 1 partition as far as most of Snuba is concerned. Signature of this function would then have `partition_id: int = 0`

I've thought about this and don't like the idea. Based on the way I think we should handle topologies, you'd need to map every physical partition to a storage (https://www.notion.so/sentry/Generic-Metrics-Slicing-9c206e02f92a4abca5948d2e21c6c00d#a1c689d0674c4671ba646ad98d1cf44d).

If you don't like that idea, then we can discuss in that doc or tomorrow's design meeting, but I'd prefer to leave the unpartitioned clusters alone for the time being

lynnagara: Oh no, just to clarify I am not suggesting to change PARTITIONED_CLUSTERS or CLUSTERS anything about that interface. Only to hide the difference between PARTITIONED_CLUSTERS and CLUSTERS inside this function and make them the same to the outside world. Basically everything from CLUSTERS gets treated as partition 0 in here.

ayirr7: Hm, I'm not sure I quite follow. I think if we have it so that partition id is not optional (i.e. for the default case we pass in 0 always), the function logic would have to involve deciding whether to look into CLUSTERS or PARTITIONED_CLUSTERS. Because the function doesn't know if the storage set key we passed in is one that is not partitioned (resides in CLUSTERS) or partitioned, with multiple partition ids (resides in PARTITIONED_CLUSTERS). I think that would interfere with the clear distinction between CLUSTERS and PARTITIONED_CLUSTERS, which is defined in a comment now in settings.

lynnagara: CLUSTERS vs PARTITIONED_CLUSTERS only exists for one reason - to keep settings backwards compatible and CLUSTERS unchanged for those who don't need to care about extra slices. There is no real reason for all the different parts of Snuba to keep this distinction, in fact it adds complexity everywhere.

In this current implementation our whole code base has to know about the difference between partitioned and unpartitioned clusters. What I'm proposing is we have this function decide which one to look in. The benefit of that is that no one else has to. Right now every caller of this module needs to know the difference between partitioned and unpartitioned clusters. Why not move that job into this module and make them exactly the same to all the other parts of Snuba.

onewland: so @lynnagara is your thought that we do something like

```
_SEEN_PARTITIONED_STORAGE_SET_KEYS = extract_storage_set_keys(_PARTITIONED_STORAGE_SET_CLUSTER_MAP)

def get_cluster(storage_set_key, partition_id = 0):
   if storage_set_key in _SEEN_PARTITIONED_STORAGE_SET_KEYS:
     get from partitioned_clusters
   else:
     assert partition_id == 0
     get from unpartitioned_clusters
```

I'm fine with that if you feel strongly but I aesthetically/subjectively prefer Riya's current approach -- I don't see how it pollutes the codebase more than yours given both approaches rely on the optionality of the argument in both to avoid adding `partition_id=n` everywhere.

docs/source/architecture/partitioning.rst
```diff
@@ -30,6 +30,14 @@ Every logical partition **must** be assigned to a physical partition and the
 valid values of physical partitions are in the range
 of ``[0,settings.LOCAL_PHYSICAL_PARTITIONS)``.
 
+Defining partitioned clusters
+--------------------------------
+To add a cluster with an associated (storage set key, partition) pair, add cluster definitions
+to ``settings.PARTITIONED_CLUSTERS`` in the desired environment's settings. Follow the same structure as
+regular cluster definitions in ``settings.CLUSTERS``. In the ``storage_sets`` field, partitioned storage
```
onewland, line 37:
```suggestion
regular cluster definitions in ``settings.CLUSTERS``. In the ``storage_set_partitions`` field, partitioned storage
```

snuba/clusters/cluster.py
onewland: OK, I'd like to continue as-is, but I agree that we are invalidating the documentation (currently) on lines 444-448 of this file and it should be corrected to clearly explain the new behavior.

nikhars: Is this a blocker or can we follow up on this separately. If it is not a blocker, can we open a follow up task and get the current implementation merged? I would like to use the functionality being introduced in this PR for https://github.com/getsentry/snuba/pull/3172

ayirr7: Right, I'll update the function docstring (apologies for not getting that done earlier)

lynnagara: Not a blocker, but I think it would be a lot better if this function took a logical partition instead of a physical partition and did the mapping in one place. I don't think there's a reason we need to call `map_logical_partition_to_physical_partition()` everywhere that then calls this function instead of just doing it just once here.

ayirr7: I see, is this referring to how this code will integrate with #3172? Do you think that change (logical to physical partition mapping inside get_cluster()) should be done before this PR gets merged?

lynnagara: No, go ahead and merge, we can figure out those details later.

lynnagara (approved):

lynnagara (approved):

snuba/clusters/cluster.py
ayirr7: I think I understand the idea about how every call to get_cluster() now would need to "think" about whether or not it is dealing with a partitioned or unpartitioned cluster (and whether or not to pass a slice id or not). Should changing this be a follow up task? Or should it be done before merging?

docs/source/architecture/partitioning.rst
ayirr7: oops, thanks for pointing that out

PR #3180, opened 2022-09-22
feat(replays): Add is_archived column and migration

cmanallen:
- Adds new is_archived column.

### DIFF
--- a/snuba/snuba_migrations/replays/0006_add_is_archived_column.py
+++ b/snuba/snuba_migrations/replays/0006_add_is_archived_column.py
@@ -1,12 +1,20 @@
 from typing import List, Sequence, Tuple
 
-from snuba.clickhouse.columns import Column, UInt
+from snuba.clickhouse.columns import Column, String, UInt
 from snuba.clusters.storage_sets import StorageSetKey
 from snuba.migrations import migration, operations
 from snuba.migrations.columns import MigrationModifiers as Modifiers
 
 new_columns: Sequence[Tuple[Column[Modifiers], str]] = [
-    (Column("is_archived", UInt(8, Modifiers(nullable=True))), "urls"),
+    (Column("is_archived", UInt(8, Modifiers(nullable=True))), "urls")
+]
+
+alter_forward: Sequence[Column[Modifiers]] = [
+    Column("platform", String(Modifiers(nullable=True)))
+]
+
+alter_backward: Sequence[Column[Modifiers]] = [
+    Column("platform", String(Modifiers(nullable=False, default="''")))
 ]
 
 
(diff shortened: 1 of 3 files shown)


snuba/snuba_migrations/replays/0006_add_is_archived_column.py
```diff
@@ -0,0 +1,112 @@
...
+new_columns: Sequence[Tuple[Column[Modifiers], str]] = [
+    (Column("is_archived", UInt(8, Modifiers(nullable=True))), "urls")
+]
+
+alter_forward: Sequence[Column[Modifiers]] = [
+    Column("platform", String(Modifiers(nullable=True)))
+]
+
+alter_backward: Sequence[Column[Modifiers]] = [
+    Column("platform", String(Modifiers(nullable=False, default="''")))
```
evanh, line 17: Is the new column supposed to have a default? Have you confirmed that running the `MODIFY` with just the nullable removes the default = ''?

evanh (commented):
Adding the column looks fine to me, I am curious about the change to platform though.

Why is platform now nullable? I am not sure of the impact of running that `MODIFY`, Clickhouse might lock up while that runs.

onewland:
Is it expected that the state of `is_archived` will change for a record over time? If yes, have you discussed the implications of mutability in a table with somebody in SNS?

Also do you need to be able to distinguish between empty string and null for platform? Because that's all that the nullable modifier buys you.

cmanallen:
@onewland 

> Is it expected that the state of `is_archived` will change for a record over time? If yes, have you discussed the implications of mutability in a table with somebody in SNS?

Replays are not mutable.  A Replay is an aggregation of many rows.  Replays emulate mutability in that new rows influence the aggregated output.

> Also do you need to be able to distinguish between empty string and null for `platform`? Because that's all that the nullable modifier in the migration buys you. See https://clickhouse.com/docs/en/sql-reference/data-types/nullable/#storage-features

Nullability provides benefits to aggregate functions like `groupUniqArray` which will filter null values automatically.  Otherwise we have to `map` in the `select` or in the application server.

The Replays schema has 19 nullable fields.

snuba/snuba_migrations/replays/0006_add_is_archived_column.py
cmanallen: @evanh If this is going to cause operational issues we can drop the nullability requirements for platform.

evanh: It's not an operational concern, it's just that I'm not sure if the `default = ''` is preserved in your migration. If it is, you'll have a nullable column that is always setting a default of `''`. The default should appear when you do `DESCRIBE TABLE`. To test, I would see that it is there before the migration, run the migration, and then be sure it's gone.

cmanallen: I'm going to drop the nullability requirements for platform.  If the need arises in the future, we can address it then.

Re: migration misunderstandings.  I'm looking at the SQL output and this ALTER statement makes sense to me.  Where am I going wrong?  If I'm altering a nullable column to not-null column it needs a default on the column, right?

> ALTER TABLE replays_local MODIFY COLUMN platform String DEFAULT '';

evanh: Ah I see, yes OK this is fine. Thank you.

evanh (approved):

PR #3183, opened 2022-09-23
ref(migrations): Allow running migrations per group

MeredithAnya:
**context:**
Right now if you run ClickHouse migrations from scratch it is going to run all the migrations for all the groups (that are active). This can make sense for local development on snuba datasets, but for some of the development on the migrations tooling, often it's enough to test with just one group.

Additionally, long term goals for migrations are for folks to own their datasets, and if you were to bring up a new cluster, you are not going to need all the groups, just the groups for that cluster. You could _technically_ get away with that by running each migration individually for each group using `snuba migrations run --migration-id=<migration_id>` but that isn't super user friendly.

This PR adds the ability to run migrations per group at a time. It also makes sure to run the `system` migrations (if they haven't been run yet) to make sure we have the `migrations_local` table created before we try to add data to it. 

**_future work_**
There are some groups that may need to be run together but in order e.g. `discover` migration group relies on other group migrations being run. We can add some sort of `depends_on` for migration groups in the future so that a user can't mistaken run group migrations out of order, but this is not covered in this PR specifically

### DIFF
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -136,14 +136,25 @@ class Runner:
 
         return migrations
 
-    def run_all(self, *, force: bool = False) -> None:
+    def run_all(self, *, force: bool = False, group: Optional[str] = None) -> None:
         """
-        Run all pending migrations. Throws an error if any migration is in progress.
+        If group is specified, runs all pending migrations for that specific group. Makes
+        sure to run any pending system migrations first so that the migrations table is
+        created before running migrations for other groups. Throw an error if any migration
+        is in progress.
+
+        If no group is specified, run all pending migrations. Throws an error if any
+        migration is in progress.
 
         Requires force to run blocking migrations.
         """
 
-        pending_migrations = self._get_pending_migrations()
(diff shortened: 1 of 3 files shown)


dbanda (approved):

snuba/migrations/runner.py
```diff
@@ -136,14 +136,25 @@ def get_status(migration_key: MigrationKey) -> Status:
...
+    def run_all(self, *, force: bool = False, group: Optional[str] = None) -> None:
         """
-        Run all pending migrations. Throws an error if any migration is in progress.
+        If group is specified, runs all pending migrations for that specific group. Makes
+        sure to run any pending system migrations first so that the migrations table is
+        created before running migrations for other groups. Throw an error if any migration
+        is in progress.
+
+        If no group is specified, run all pending migrations. Throws an error if any
+        migration is in progress.
```
lynnagara, line 147: nit: this line is written twice, once is enough: `Throws an error if any migration is in progress`

lynnagara (approved):

PR #3184, opened 2022-09-23
feat(test): add tests for dist migrations clusters

dbanda:
Add a docker compose setup for testing migrations on multi-node clusters. By running docker compose with the `multi_node` profile, the multinode setup will create 5 extra clickhouse nodes, running 3 clickhouse clusters that loosely approximate the cluster layout in prod.

The tests live in a folder `test_distributed_migrations` that is separate from the other single node unit tests. This is done so that the multi-node tests get their own set-up and teardown functions. In CI these test are run with `SNUBA_SETTINGS=test_distributed_migrations`.


snuba/clusters/cluster.py
```diff
@@ -47,7 +47,13 @@ class ClickhouseClientSettings(Enum):
     CLEANUP = ClickhouseClientSettingsType({}, None)
     INSERT = ClickhouseClientSettingsType({}, None)
     MIGRATE = ClickhouseClientSettingsType(
-        {"load_balancing": "in_order", "replication_alter_partitions_sync": 2}, 10000
+        {
+            "load_balancing": "in_order",
+            "replication_alter_partitions_sync": 2,
+            "mutations_sync": 2,
+            "database_atomic_wait_for_drop_and_detach_synchronously": 1,
```
MeredithAnya, line 54: Do we know that these settings work for  clickhouse version `20.7` ? If not, we maybe have to have a different setting enum

snuba/migrations/operations.py
```diff
@@ -111,7 +111,7 @@ def __init__(self, storage_set: StorageSetKey, table_name: str) -> None:
         self.table_name = table_name
 
     def format_sql(self) -> str:
-        return f"DROP TABLE IF EXISTS {self.table_name};"
+        return f"DROP TABLE IF EXISTS {self.table_name} SYNC;"
```
MeredithAnya, line 114: same here about supported by older clickhouse versions. If there are a bunch of places like this, we may want to have like a setting that enables the new features or something

snuba/settings/settings_test_distributed_migrations.py
```diff
@@ -0,0 +1,64 @@
...
+        "host": os.environ.get("CLICKHOUSE_HOST", "clickhouse-01"),
+        "port": int(os.environ.get("CLICKHOUSE_PORT", 9000)),
+        "user": os.environ.get("CLICKHOUSE_USER", "default"),
+        "password": os.environ.get("CLICKHOUSE_PASSWORD", ""),
+        "database": os.environ.get("CLICKHOUSE_DATABASE", "snuba_test"),
+        "http_port": int(os.environ.get("CLICKHOUSE_HTTP_PORT", 8229)),
+        "storage_sets": {
+            "migrations",
+        },
+        "single_node": False,
```
MeredithAnya, line 27: so I actually think we could make the migrations cluster a single node, sine that's what it is production as of right now

Makefile
```diff
@@ -11,7 +11,12 @@ setup-git:
 	pre-commit install --install-hooks
 
 test:
-	SNUBA_SETTINGS=test pytest -vv -v -m "not ci_only"
+	SNUBA_SETTINGS=test pytest -vv tests -v -m "not ci_only"
+
+test-distributed-migrations:
+	docker-compose -f docker-compose.gcb.clusters.yml down
+	SNUBA_IMAGE=snuba-test SNUBA_SETTINGS=test docker-compose -f docker-compose.gcb.clusters.yml run --rm snuba-test
```
MeredithAnya, line 18: im assuming `SNUBA_SETTINGS` here should also be `test_distributed_migrations` yeah?

snuba/settings/settings_test_distributed_migrations.py
dbanda: Ok I'll try that out. My understanding was setting single_node would disable the dist tables.

snuba/migrations/operations.py
dbanda: removed

snuba/clusters/cluster.py
dbanda: In my testing everything seemed fine on 20.7

Makefile
dbanda: fixed

MeredithAnya (approved):

PR #3185, opened 2022-09-23
Auto-generate documentation for jsonschema

enochtangg:
This PR is responsible for automatically generating jsonschema documentation for snuba components. This includes the following:
* Script for converting jsonschema to markdown
* Added Sphinx support for markdown files
* Added CI step to automatically generate config snuba docs
* Updated descriptions for every key in jsonschema. Note: jsonschema was reformatted to be a bit more explicit with what is included in the dictionary. References to `TYPE_STRING` were removed for clarity because `description` key needs to live in the same level.

Example screenshots of what it looks like:
image
image


requirements.txt
```diff
@@ -9,7 +9,7 @@ datadog==0.21.0
 flake8==5.0.4
 Flask==2.2.2
 honcho==1.1.0
-jsonschema==4.16.0
+jsonschema==4.9.1
```
evanh, line 12: Why this change?

evanh (approved):
This looks pretty nice to me. One small question.

requirements.txt
enochtangg: Good catch, this was from testing out different jsonschema to markdown or restructuredtext libs which required a different version of jsonschema. I've changed this back.

snuba/datasets/configuration/generate_config_docs.py
```diff
@@ -0,0 +1,31 @@
...
+                show_examples="all",
+            )
+            md_lines = parser.parse_schema(schema)
+            reformatted_md_lines = []
+            for line in md_lines:
+                # jsonschema2md adds some extra lines of its own. This removes them.
+                if "**Items**" not in line:
+                    line = line.replace("`", "")
+                    line = line.replace(" Cannot contain additional properties.", "")
+                    reformatted_md_lines.append(line)
```
onewland, line 26: Can you share what the bad `jsonschema2md` output looks like? Just having trouble picturing it.

snuba/datasets/configuration/generate_config_docs.py
```diff
@@ -0,0 +1,31 @@
...
+
+import jsonschema2md
+
+from snuba import settings
+from snuba.datasets.configuration.json_schema import V1_ALL_SCHEMAS
+
+CONFIGURATION_DOCS_PATH = f"{settings.ROOT_REPO_PATH}/docs/source/configuration"
+
+
+if __name__ == "__main__":
```
onewland, line 11: If our JSON schema was not embedded in python but present in files that can parse raw JSON schema text, would we be able to use an existing utility rather than rolling our own here?

Not trying to stop you from merging this but I want to continue to reiterate that I think there are benefits to not having the JSON schema be encoded in Python dicts (e.g. you could have [first-class text editor support](https://code.visualstudio.com/docs/languages/json#_mapping-to-a-schema-in-the-workspace) for authoring new datasets, or validate with other [language bindings](https://json-schema.org/implementations.html#validators) if we want to support Rust in the future)

snuba/datasets/configuration/generate_config_docs.py
enochtangg: Yeah, so there are two things which `jsonschema2md` adds. Firstly, whenever there is a key with `TYPE_STRING_ARRAY`, there will be an extra `Items` bullet point without any type. See `columns` and `query_processors` keys. The description on the the upper level key is already sufficient. Secondly, if any `properties` contain the `additionalProperties: False` key/value, then the "Cannot contain any additional properties" string is added. This gets very repetitive and causes some confusion.  

image

snuba/datasets/configuration/generate_config_docs.py
enochtangg: I agree. The work behind implementing jsonschema validation for pre-commit further supports this idea. Currently, our pre-commit uses a custom script (https://github.com/getsentry/snuba/blob/master/snuba/validate_configs.py) to validate our configs. This was required because the built-in hook`check-jsonschema` (https://github.com/python-jsonschema/check-jsonschema) only accepts .json or .yaml files as an input parameter.

PR #3187, opened 2022-09-26
feat(MDC): Auto import logical processors

volokluev:
Currently query processors work in the entity builder because of a hack that manually imports them and adds them to a dictionary. This PR makes the LogicalQueryProcessor a registered class and auto imports the files. In order to make this work, the `config_key` property had to be defined on all the processors which is most of the lines in this PR.

There shouldn't be much to review here

### DIFF
--- a/snuba/datasets/configuration/entity_builder.py
+++ b/snuba/datasets/configuration/entity_builder.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
-import logging
-from typing import Any, Sequence, Type
+from typing import Any, Sequence
 
 import snuba.clickhouse.translators.snuba.function_call_mappers  # noqa
 from snuba.clickhouse.translators.snuba.allowed import (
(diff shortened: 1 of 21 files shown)


snuba/datasets/configuration/generic_metrics/entities/distributions.yaml
```diff
@@ -78,7 +78,7 @@ query_processors:
         bucketed_time: timestamp
       time_parse_columns: [timestamp]
   -
-    processor: referrer_rate_limit
+    processor: referrer_rate_limiter
```
volokluev, line 81: this name was made consistent with other rate limiters

snuba/datasets/configuration/entity_builder.py
```diff
@@ -18,35 +17,8 @@
...
-    "transform_tag_types": TagsTypeTransformer,
-    "handle_mapped_granularities": MappedGranularityProcessor,
-    "translate_time_series": TimeSeriesProcessor,
-    "referrer_rate_limit": ReferrerRateLimiterProcessor,
-    "org_rate_limiter": OrganizationRateLimiterProcessor,
-    "project_referrer_rate_limiter": ProjectReferrerRateLimiter,
-    "project_rate_limiter": ProjectRateLimiterProcessor,
-    "resource_quota_limiter": ResourceQuotaProcessor,
-}
-
```
onewland, line 47: 🎉

snuba/query/processors/logical/custom_function.py
```diff
@@ -39,6 +42,13 @@ def partial_function(body: str, constants: Sequence[Tuple[str, Any]]) -> Express
 
 class CustomFunction(LogicalQueryProcessor):
     """
+    WARNING
```
evanh, line 45: Can you add a tech debt task in JIRA to track this?

snuba/query/processors/physical/null_column_caster.py
```diff
@@ -1,13 +1,15 @@
...
 
 from snuba.clickhouse.columns import FlattenedColumn, SchemaModifiers
 from snuba.clickhouse.query import Query
-from snuba.datasets.storage import ReadableTableStorage
 from snuba.query.expressions import Column, Expression, FunctionCall, Literal
 from snuba.query.functions import AGGREGATION_FUNCTIONS
 from snuba.query.processors.physical import ClickhouseQueryProcessor
 from snuba.query.query_settings import QuerySettings
 
+if TYPE_CHECKING:
```
evanh, line 10: What happened here?

snuba/query/processors/logical/custom_function.py
volokluev: sure

snuba/query/processors/physical/null_column_caster.py
volokluev: This is done to prevent a circular import. ReadableTableStorabe is only necessary for type checking but importing it during runtime causes an import cycle. I don't remember the cycle path exactly but this is a documented way to solve import cycle issues in mypy

https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking

evanh: Sorry I should have been more clear, what was the circular import, and was that import caused by our code being poorly laid out?

volokluev: `snuba.datasets.storage` import imports ClickhouseQueryProcessor (for type checking)
this file imports `ReadableTableStorage` for type checking

evanh: Got it, OK thanks.

PR #3189, opened 2022-09-26
feat: Remove global rate limiter

evanh:
The global rate limiter is a very hot Redis key that can't be sharded. As we move 
to a multi cluster environment that is a non-starter. Also, each table has its own
rate limiter so the global one is less useful.


nikhars (approved):
LGTM. Before merging, could you verify that there is a `TableRateLimit` being applied on all storages?

evanh:
Checked through all the storages, and they are all configured to have table rate limits. Not all of them have values other than the defaults though.

I also checked through the rate limiting history, and saw that we rejected 4 queries in the past month with the global rate limiter.

volokluev (approved):

PR #3190, opened 2022-09-26
fix(errors): Allow input format mismatch

nikhars:
If there is a mismatch between the input format and clickhouse format, by default clickhouse throws an error. That will lead to ingestion backlogs in case of format mismatches. Allow the error storage to skip unknown fields during insertion to make it more robust. This is already happening when writing to transactions storage https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/transactions.py#L43-L46

### DIFF
--- a/snuba/datasets/storages/errors.py
+++ b/snuba/datasets/storages/errors.py
@@ -55,4 +55,7 @@ storage = WritableTableStorage(
         promoted_tags={"tags": list(promoted_tag_columns.keys()), "contexts": []},
         state_name=ReplacerState.ERRORS,
     ),
+    writer_options={
+        "input_format_skip_unknown_fields": 1,
+    },
 )


onewland (approved):

fpacifici:
How would we find out that we forgot to add a column to a distributed table if it does not raise errors anymore?
I think erroring on format mismatch is a good thing, not a lack of robustness. 

It was added to transactions only to support a migration. And it was forgotten afterwards, it should have been removed.
The migration called for adding a column to a nested field, which requires data to be in the input before we add the column. In general this scenario is an error.
I think we should not allow this behavior by default.

evanh:
I agree with Filippo on this: the fact that transactions skips unknown fields is a bug, not a feature.

nikhars:
My rationale was the following. When introducing a new migration, in case you forget to add a column to the distributed table, it should be a bug. Not something that impacts ingestion and causes an incident. Most of the times when new columns are being introduced, new features are being introduced and it it ok to think of failures in those scenarios as failures of that specific feature. Not something which breaks the whole application

PR #3192, opened 2022-09-27
ref(logging): make logging structured for snuba admin

onewland:
Add the structured_log package and move all of snuba admin to use it. I think eventually we'd like to move everything to it but this lets us go forward a little more cautiously.

Depends on https://github.com/getsentry/pypi/pull/88

### Advantages of structlog:
- lets us query logs via structured attributes (see https://cloud.google.com/logging/docs/structured-logging if you're not familiar)
  - this is especially helpful for access logs which I want to add this week)
- makes google cloud logs understand the correct log level (everything isn't error)
- convenient binding lets us attach context to logs at various levels before writing (see https://www.structlog.org/en/stable/contextvars.html)

I've verified that this all works on the Google Cloud SNS project, though it's possible prod will be different. 

### Changes:
- added structlog and sentry-structlog packages
   - sentry-structlog is third-party and not as widely used as structlog. It's basically a copy of python sentry-sdk logic for forwarding `sys.exc_info()` information to Sentry when `logger.error` is called. I wasn't sure how much to worry about a 3rd party dependency, but it seemed to serve our needs here
- start using loggers from structlog and not logging in the admin code
- change the "friendly" logging in RuntimeConfigClient to be structured

### Sample output from my local environment 
(filtered out logs that aren't using structured_log, but I can provide an example with them mixed too):
```
{"module": "snuba.admin.views", "user": "unknown", "event": "authorize.finished", "severity": "info", "timestamp": "2022-09-28T19:43:54.765098Z"}
{"module": "snuba.admin.views", "user": "unknown", "event": "authorize.finished", "severity": "info", "timestamp": "2022-09-28T19:43:54.779433Z"}
{"module": "snuba.admin.views", "user": "unknown", "event": "authorize.finished", "severity": "info", "timestamp": "2022-09-28T19:43:54.796352Z"}
{"module": "snuba.admin.views", "user": "unknown", "event": "authorize.finished", "severity": "info", "timestamp": "2022-09-28T19:43:54.811762Z"}
{"module": "snuba.admin.views", "user": "unknown", "event": "authorize.finished", "severity": "info", "timestamp": "2022-09-28T19:44:01.610722Z"}
{"module": "snuba.admin.notifications.base", "user": null, "action": "updated", "option": "events_read_only_table", "old": 6, "new": 5, "timestamp": "2022-09-28T19:58:12.303542Z", "event": "value_updated", "severity": "info"}
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,6 +26,7 @@ sentry-arroyo==1.0.4
 sentry-relay==0.8.13
 sentry-sdk==1.9.8
 simplejson==3.17.6
+structlog==22.1.0
 urllib3==1.26.11
 pyuwsgi==2.0.20
 Werkzeug==2.2.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,6 +27,7 @@ sentry-relay==0.8.13
 sentry-sdk==1.9.8
 simplejson==3.17.6
 structlog==22.1.0
+structlog-sentry==2.0.0
 urllib3==1.26.11
 pyuwsgi==2.0.20
 Werkzeug==2.2.2
--- a/snuba/admin/notifications/base.py
+++ b/snuba/admin/notifications/base.py
@@ -49,7 +49,7 @@ class NotificationBase(ABC):
 
 class RuntimeConfigLogClient(NotificationBase):
     def __init__(self) -> None:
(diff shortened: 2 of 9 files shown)


nikhars (approved):

volokluev (approved):
nice

PR #3196, opened 2022-09-28
ci(tests): Add tests against clickhouse 21.8

nikhars:


MeredithAnya:
@nikhars I know this is just a draft but do you think we should be separating out the multi node distributed testing like this as well? Basically so that we can test multiple version of clickhouse for those tests cc @dbanda

dbanda:
> @nikhars I know this is just a draft but do you think we should be separating out the multi node distributed testing like this as well? Basically so that we can test multiple version of clickhouse for those tests cc @dbanda

@MeredithAnya I refactored my PR to use the the same docker compose. so in theory we should just be able to reuse the same version variable to test on the multi node case. hopefully 🤞

nikhars:
> @nikhars I know this is just a draft but do you think we should be separating out the multi node distributed testing like this as well? Basically so that we can test multiple version of clickhouse for those tests cc @dbanda

I don't have the full context. But ideally testing against multiple versions should be a testing matrix. One thing I stumbled upon while doing this PR is that having parallel tests run in multiple distributed clusters throws errors. So I am not tackling that use case for now.

nikhars:
Continuing this work via https://github.com/getsentry/snuba/pull/3292. So closing this.

PR #3198, opened 2022-09-28
ref(tests): Move scheduler inside freezetime

nikhars:
Since it is the optimize scheduler's cutoff time which determines whether to raise an exception or not, move the creation of the scheduler inside frozen time to have a consistent behavior during tests.

The setting `settings.OPTIMIZE_JOB_CUTOFF_TIME` also defaults to 1 day by default to avoid tests running into timeout exception. For this test, we should change it to 23 hours.


nikhars:
Test still fails. Need to investigate more

fpacifici (approved):

PR #3199, opened 2022-09-29
fix(logging): remove auth for health check

onewland:
Now that we log the result of all auth calls (currently a no-op) we're getting spammed about once/second with what I assume must be health check calls (given the low traffic to snuba admin).

Verified locally that this bypasses the method for `/health` but keeps it intact for others.

Additionally, add some logging context to benefit us in the future

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -50,9 +50,10 @@ def handle_invalid_json(exception: UnauthorizedException) -> Response:
 @application.before_request
 def authorize() -> None:
     logger.debug("authorize.entered")
-    user = authorize_request()
-    logger.info("authorize.finished", user=user)
-    g.user = user
+    if request.endpoint != "health":
+        user = authorize_request()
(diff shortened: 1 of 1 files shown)


snuba/admin/views.py
```diff
@@ -47,12 +48,19 @@ def handle_invalid_json(exception: UnauthorizedException) -> Response:
     )
 
 
+@application.before_request
+def set_logging_context() -> None:
+    clear_contextvars()
```
evanh, line 53: Just read up on this package, this is very cool. Something we might want to leverage for the api as well.

evanh (approved):

snuba/admin/views.py
onewland: yeah, i think once you start "thinking" in terms of structured logging it's an extremely useful tool. IMO it should be the basis for all logs, but admin is a safe place to start

PR #3200, opened 2022-09-29
feat(logs): Snuba initialization monitoring/logging

rahul-kumar-saini:
### Overview
- Snuba CLI init takes longer and longer as Snuba grows
    - This is possibly causing health check timeouts on Snuba deploy (See #inc-215)
- Added logging and a Sentry transaction for Snuba CLI initialization (which is effectively Snuba's init)
- All CLI imports are instrumented with additional instrumentation for Dataset as Config initialization to monitor performance impact over time of adding/converting more datasets as config
    - Dataset config instrumentation is tagged with dataset component names

Transaction example:

Screen Shot 2022-09-29 at 12 38 58 PM

Logging example:

Screen Shot 2022-09-29 at 11 41 40 AM


snuba/datasets/factory.py
```diff
@@ -111,7 +113,8 @@ def _ds_factory() -> _DatasetFactory:
     # This function can be acessed by many threads at once. It is okay if more than one thread recreates the same object.
     global _DS_FACTORY
     if _DS_FACTORY is None:
-        _DS_FACTORY = _DatasetFactory()
+        with sentry_sdk.start_span(op="function", description="Load Dataset Factory"):
```
volokluev, line 116: this should probably be in the `__init__` function. Where we call this may change and we want the profiling to not dissapear

snuba/datasets/storages/factory.py
```diff
@@ -197,7 +199,8 @@ def get_config_built_storages(self) -> dict[StorageKey, Storage]:
 def _storage_factory() -> _StorageFactory:
     global _STORAGE_FACTORY
     if _STORAGE_FACTORY is None:
-        _STORAGE_FACTORY = _StorageFactory()
+        with sentry_sdk.start_span(op="function", description="Load Storage Factory"):
+            _STORAGE_FACTORY = _StorageFactory()
```
volokluev, line 203: ditto

volokluev (approved):
Good overall, just some suggestions

mcannizz:
Nice! Would it make sense to also expose init time as a metric in Datadog? I ask because we do expect init time to increase over time, so maybe it would be handy to have in our dashboards or alerts.

snuba/cli/__init__.py
```diff
@@ -17,8 +29,13 @@ def main() -> None:
     `OoO'  o   O `OoO'o `OoO' `OoO'o"""
 
 
-for loader, module_name, is_pkg in walk_packages(__path__, __name__ + "."):
-    module = __import__(module_name, globals(), locals(), ["__name__"])
-    cmd = getattr(module, module_name.rsplit(".", 1)[-1])
-    if isinstance(cmd, click.Command):
-        main.add_command(cmd)
+with sentry_sdk.start_transaction(op="snuba_init", name="Snuba CLI Initialization"):
+    for loader, module_name, is_pkg in walk_packages(__path__, __name__ + "."):
```
fpacifici, line 33: This is likely to initialize a transaction on consumers as well.
We cannot afford to do that without doing a throughout test before and without testing locally what gets produced there. We know snuba produces billions of locally dropped transactions which likely come from consumers.
So please check that consumer does not produce transactions

fpacifici (changes requested):
Please verify the behavior of tracing on consumers

MeredithAnya:
From the screenshot it kinds of seems like we are missing more details about what is taking `Load Dataset Factory` so long, but not sure if Im not reading things right

snuba/cli/__init__.py
rahul-kumar-saini: Tested behavior of consumers locally - all consumer startups create a transaction which is finished and sent to Sentry before message processing begins. There should only be 1 transaction per consumer start (or restart). The added spans may make it into any new transactions created around Dataset component factory initialization but even this initialization should only happen once per consumer startup.

rahul-kumar-saini:
> Nice! Would it make sense to also expose init time as a metric in Datadog? I ask because we do expect init time to increase over time, so maybe it would be handy to have in our dashboards or alerts.

@mcannizz I've added a Datadog metric so we can easily see it in a dashboard but I think we can probably just monitor/alert on the performance data in Sentry

snuba/cli/__init__.py
```diff
@@ -1,6 +1,21 @@
...
 
 import click
+import sentry_sdk
+
+from snuba.environment import metrics as environment_metrics
+from snuba.environment import setup_logging, setup_sentry
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+setup_sentry()
+setup_logging("INFO")
```
fpacifici, line 13: Since each script is supposed to honor the `--log-level` parameter. Have you verified that you can reinitialize the logger with a different level if the script then tries to initialize it with a different level ?

snuba/cli/__init__.py
```diff
@@ -17,8 +32,15 @@ def main() -> None:
...
 
 
-for loader, module_name, is_pkg in walk_packages(__path__, __name__ + "."):
-    module = __import__(module_name, globals(), locals(), ["__name__"])
-    cmd = getattr(module, module_name.rsplit(".", 1)[-1])
-    if isinstance(cmd, click.Command):
-        main.add_command(cmd)
+with sentry_sdk.start_transaction(op="snuba_init", name="Snuba CLI Initialization"):
+    for loader, module_name, is_pkg in walk_packages(__path__, __name__ + "."):
+        logger.info(f"Loading module {module_name}")
```
fpacifici, line 37: It seems this will be really really verbose. Are you sure about logging this in `INFO`? It will show up in google cloud log.I feel it is not a very useful piece of info. I would demote this to DEBUG

fpacifici (approved):

snuba/cli/__init__.py
rahul-kumar-saini: Good catch, I've set `setup_logging` to force reset logging when the function is run to override the saved log level.

PR #3204, opened 2022-09-29
feat(admin): use JWT to validate user ID and e-mail

onewland:
Use the actual JWT from google to validate user ID and e-mail

Next steps would be to extract roles and potentially permissions from IAM based on the user ID (https://cloud.google.com/resource-manager/reference/rest/v1/projects/getIamPolicy?apix_params=%7B%22resource%22%3A%22search-and-storage%22%2C%22body_resource%22%3A%7B%7D%7D)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ datadog==0.21.0
 flake8==5.0.4
 Flask==2.2.2
 honcho==1.1.0
+python-jose[cryptography]==3.3.0
 jsonschema==4.16.0
 Markdown==3.4.1
 packaging==21.3
--- a/snuba/admin/jwt.py
+++ b/snuba/admin/jwt.py
@@ -1,5 +1,6 @@
 from typing import Any, Optional
 
+import requests
 import structlog
 
 from snuba.admin.user import AdminUser
@@ -16,7 +17,6 @@ def _certs() -> Any:
     validating Google-signed JWTs. Since these change rarely, the result
     is cached on first request for faster subsequent responses.
     """
-    import requests
 
     global CERTS
     if CERTS is None:
@@ -30,8 +30,6 @@ def _get_metadata(item_name: str) -> str:
     See https://cloud.google.com/compute/docs/storing-retrieving-metadata for
     possible item_name values.
     """
-    import requests
-
     endpoint = "http://metadata.google.internal"
     path = "/computeMetadata/v1/project/"
     path += item_name
--- a/snuba/admin/jwt.py
+++ b/snuba/admin/jwt.py
@@ -1,24 +1,14 @@
-from dataclasses import dataclass
 from typing import Any, Optional
 
 import structlog
+
+from snuba.admin.user import AdminUser
 
 logger = structlog.get_logger().bind(module=__name__)
 
 
 CERTS: Optional[Any] = None
 AUDIENCE: Optional[str] = None
-
-
-@dataclass
-class AdminUser:
-    """
-    Basic encapsulation of a user of the admin panel. In the future,
-    should be extended to contain permissions among other things
-    """
-
-    email: str
-    id: str
 
 
 def _certs() -> Any:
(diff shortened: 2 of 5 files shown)


snuba/admin/jwt.py
```diff
@@ -0,0 +1,78 @@
+from dataclasses import dataclass
```
onewland, line 1: This is mostly a copy of this GCP example: https://cloud.google.com/python/docs/getting-started/authenticate-users#create_the_source_code

but it seemed reasonable to me

snuba/admin/jwt.py
```diff
@@ -0,0 +1,68 @@
...
+def _get_metadata(item_name: str) -> str:
+    """Returns a string with the project metadata value for the item_name.
+    See https://cloud.google.com/compute/docs/storing-retrieving-metadata for
+    possible item_name values.
+    """
+    import requests
+
+    endpoint = "http://metadata.google.internal"
+    path = "/computeMetadata/v1/project/"
+    path += item_name
```
MeredithAnya, line 37: I think i might still be confused by `item_name` is that like the project `attributes` https://cloud.google.com/compute/docs/metadata/default-metadata-values#project-attributes-metadata ?

[edit]: I see it's project id or the numeric project id

requirements.txt
```diff
@@ -9,6 +9,7 @@ datadog==0.21.0
 flake8==5.0.4
 Flask==2.2.2
 honcho==1.1.0
+python-jose[cryptography]==3.3.0
```
evanh, line 12: Just for my own understanding, you had to add these to the pypi repo as well right?

snuba/admin/jwt.py
```diff
@@ -0,0 +1,68 @@
...
+CERTS: Optional[Any] = None
+AUDIENCE: Optional[str] = None
+
+
+def _certs() -> Any:
+    """Returns a dictionary of current Google public key certificates for
+    validating Google-signed JWTs. Since these change rarely, the result
+    is cached on first request for faster subsequent responses.
+    """
+    import requests
```
evanh, line 19: I think you could probably import this outside the function.

nikhars, line 19: Any specific reason you are importing modules locally within different function?

nikhars (approved):
Left some minor comment. Else looks good.

snuba/admin/jwt.py
```diff
@@ -0,0 +1,68 @@
...
+    """Returns a string with the project metadata value for the item_name.
+    See https://cloud.google.com/compute/docs/storing-retrieving-metadata for
+    possible item_name values.
+    """
+    import requests
+
+    endpoint = "http://metadata.google.internal"
+    path = "/computeMetadata/v1/project/"
+    path += item_name
+    response = requests.get(
```
evanh, line 38: I am not good at auth: don't we need to pass in some headers or something so Google knows which project we are trying to access?

requirements.txt
onewland: yes

snuba/admin/jwt.py
onewland: My understanding is that this API returns the project ID based on the GKE cluster that's initiating the request.

From https://cloud.google.com/python/docs/getting-started/authenticate-users#create_the_source_code:

> The App Engine metadata service (and similar metadata services for other Google Cloud computing services) looks like a web site and is queried by standard web queries. However, it isn't actually an external site, but an internal feature that returns requested information about the running app, so it is safe to use http instead of https requests. It's used to get the current Google Cloud identifiers needed to define the JWT assertion's intended audience.

snuba/admin/jwt.py
onewland: I was just copying the GCP example -- happy to move out that import, they probably do that in case you copy/paste into a much bigger and higher traffic application:

https://cloud.google.com/python/docs/getting-started/authenticate-users#create_the_source_code

snuba/admin/jwt.py
evanh: Ah I see. Man Google is a smart company.

evanh (approved):

PR #3205, opened 2022-09-30
ref(upgrade): Revert pipeline builders for errors and transactions

nikhars:
For performing dual reads during upgrades, we had modified the pipeline builder of errors and transaction entities to be pipeline delegator to support running parallel queries. Now that the upgrades are done, we can move back to simple pipeline builder for both of those entities.


snuba/datasets/entities/events.py
```diff
@@ -116,82 +110,23 @@ def select_storage(
...
-
 class BaseEventsEntity(Entity, ABC):
     """
     Represents the collection of classic sentry "error" type events
     and the particular quirks of storing and querying them.
     """
 
     def __init__(self, custom_mappers: Optional[TranslationMappers] = None) -> None:
-        v1_pipeline_builder = SimplePipelineBuilder(
+        events_storage = get_writable_storage(StorageKey.ERRORS)
```
MeredithAnya, line 120: Maybe this is a dumb question, but is the new storage `ERRORS` now? or would this be `ERRORS_V2`?

nikhars: The new storage is `ERRORS`. We cannot remove `ERRORS` as Single Tenant and Self hosted still use them. So in production, we swapped the mapping so that `ERRORS` points to the new Tiger cluster.

MeredithAnya (approved):

PR #3214, opened 2022-10-03
feat(MDC): Allow init arguments for storage query processors

volokluev:
This was neglected for speed's sake but cannot continue this way.

### DIFF
--- a/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml
@@ -84,4 +84,6 @@ schema:
   local_table_name: generic_metric_distributions_aggregated_local
   dist_table_name: generic_metric_distributions_aggregated_dist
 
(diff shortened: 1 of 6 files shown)


tests/datasets/storages/test_storages.py
```diff
@@ -1,23 +1,12 @@
...
     with pytest.raises(AttributeError):
         StorageKey.NON_EXISTENT_STORAGE
 
-    assert REGISTERED_STORAGE_KEYS == {
-        "GENERIC_METRICS_DISTRIBUTIONS": "generic_metrics_distributions",
-        "GENERIC_METRICS_DISTRIBUTIONS_RAW": "generic_metrics_distributions_raw",
-        "GENERIC_METRICS_SETS": "generic_metrics_sets",
-        "GENERIC_METRICS_SETS_RAW": "generic_metrics_sets_raw",
-    }
-
```
volokluev, line 20: I removed this assert because it's a bad test. Any newly registered storage will have to update this test.

PR #3218, opened 2022-10-04
feat(mdc) Make MessageProcessor a RegisteredClass for dataset configuration

evanh:
This makes non-replacement message processors into RegisteredClasses so that
they can be detected automatically by the configuration.

There are a few major changes here:

- Move all the non-replacement message processors into their own folder
- Create an intermediate parent class `DatasetMessageProcessor` that
becomes the parent of all those message processors
- The `DatasetMessageProcessor` becomes the registered class

This meant also moving some code that is shared between event and replacement
processors into a new place.


snuba/datasets/configuration/storage_builder.py
```diff
@@ -76,7 +76,7 @@ def __build_readable_storage_kwargs(config: dict[str, Any]) -> dict[str, Any]:
 
 
 def build_stream_loader(loader_config: dict[str, Any]) -> KafkaStreamLoader:
-    processor = CONF_TO_PROCESSOR[loader_config["processor"]]()
+    processor = DatasetMessageProcessor.get_from_name(loader_config["processor"])()
```
evanh, line 79: I didn't add init params to this yet. The ErrorsProcessor will eventually require this though.

snuba/datasets/processors/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    def get_from_name(cls, name: str) -> Type["DatasetMessageProcessor"]:
+        return cast(Type["DatasetMessageProcessor"], cls.class_from_name(name))
+
+    @classmethod
+    def from_kwargs(cls, **kwargs: str) -> DatasetMessageProcessor:
+        return cls(**kwargs)
+
+    @classmethod
+    def config_key(cls) -> str:
+        return camel_to_snake(cls.__name__)
```
evanh, line 42: I am not sure how I feel about this. If we specified the class name exactly in the configuration, we could get rid of a lot of logic. It does ruin the aesthetics of the configuration though.

snuba/processor.py
```diff
@@ -70,6 +72,52 @@ class InvalidMessageVersion(SerializableException):
     pass
 
 
+class ReplacementType(str, Enum):
```
evanh, line 75: These were moved from `events_processor_base`.

volokluev: These seem a little too specific to be in such a high level file. Wouldn't ReplacementType(s) live with the replacement processor?

volokluev (approved):
LGTM just two things that shouldn't require re-review

snuba/processor.py
evanh: Potentially, but right now this type is imported in `snuba/datasets` as well as `snuba/replacers` and I think organizing the replacement code is out of scope for this PR.

snuba/datasets/configuration/storage_builder.py
volokluev: can we add a ticket to support init args for `DatasetMessageProcessor(s)`? Also, you may want to call `from_kwargs` for completeness here

evanh: Done and done.

PR #3219, opened 2022-10-04
feat: Ability to use separate Redis clusters for rate limiting, replacer state and caching

untitaker:
## Motivation

During incidents documented in SNS-1526 we saw a bug in Snuba replacements impacting query performance because a redis instance was shared. We want to avoid problems like that going forward, so this PR introduces logic to allow for uses of multiple redises.

## prior art

https://github.com/getsentry/snuba/pull/2958 does the same thing, but this one addresses some of the review comments, e.g. wrt json serializability of the config. It is also quite a bit simpler because it does not attempt to deduplicate connections at all: if two usecases are configured to point to the same (non-default) cluster, there will be two clients regardless.

## TODO

The usecases and their designation I still need to re-check before this leaves draft status, but the basics in `redis.py`  I think are ready.


snuba/redis.py
```diff
@@ -67,35 +65,77 @@ def wrapper() -> RedisClientType:
...
+
+class RedisClients(TypedDict):
+    cache: RedisClientType
+    rate_limiter: RedisClientType
+    subscription_store: RedisClientType
+    replacements_store: RedisClientType
+    misc: RedisClientType
+
+
+redis_clients: RedisClients = {
```
onewland, line 129: I think we should probably the actual mapping behind a `get_redis_client(name: str)` or something

onewland (approved):
approach looks generally fine to me (though I do think that mapping should be hidden from importers)

snuba/redis.py
untitaker: done

snuba/settings/__init__.py
```diff
@@ -118,6 +134,25 @@
...
+    replacements_store: RedisClusterConfig | None
+    misc: RedisClusterConfig | None
+
+
+REDIS_CLUSTERS: RedisClusters = {
+    "cache": None,
+    "rate_limiter": None,
+    "subscription_store": None,
+    "replacements_store": None,
+    "misc": None,
```
fpacifici, line 153: Why not removing this `misc` ? 
We can always map two logical clusters to the same physical cluster. 
Not having a `misc` will discourage people from using it for all new use cases and will remove the cognitive load of making a decision whether we need a new logical cluster for a new use case or not. 
Then a lot of small use cases will be mapped to the same physical cluster in the settings.

snuba/settings/__init__.py
```diff
@@ -118,6 +134,25 @@
...
+class RedisClusters(TypedDict):
+    cache: RedisClusterConfig | None
+    rate_limiter: RedisClusterConfig | None
+    subscription_store: RedisClusterConfig | None
+    replacements_store: RedisClusterConfig | None
+    misc: RedisClusterConfig | None
+
+
+REDIS_CLUSTERS: RedisClusters = {
+    "cache": None,
```
evanh, line 149: I think we should have these keys be in an Enum, to avoid raw strings being passed into `get_redis_client`.

evanh (commented):
This strategy looks solid to me.

You've probably thought of this, but I thought I should write it down: this can't be merged until these settings have been added to the prod settings file.

snuba/settings/__init__.py
untitaker: are you talking about the interface to `get_redis_client` here, or would you prefer to configure things with an enum as key? the strings are still typed using `Literal`.

evanh:
> are you sure?

You're right, I missed the check on config being None. My mistake.

snuba/settings/__init__.py
untitaker: i split up the usecase now. i was a bit hesitant because i didn't pay close attention there as to whether all those usecases in `snuba.state` can actually be split safely. ptal

evanh (approved):

untitaker:
@fpacifici @evanh this PR now groups clickhouse errors by referrer, except those defined in a certain list. ptal

PR #3220, opened 2022-10-04
feat(MDC) auto-import storage pre-filters from config

enochtangg:
This PR is responsible for auto-importing storage pre-filters defined in config. This is accomplished by adding a `config_key` to all pre-filter classes and registering them to `RegisteredClasses`.

Changes include:
* Registered `StreamMessageFilter` to `RegisteredClasses` and added support methods
* Added `config_key` to all children pre-filter classes
* Remove hardcoded mappings and updated builder to use registry
* Consolidated cdc pre-filters with all other pre-filters
* Rename `KafkaHeaderFilterWithBypass` to `KafkaHeaderWithBypassFilter` to ensure consistent naming convention
* Rename `config_key`: `kafka_header_select_filter` to `kafka_header_select` for consistency


enochtangg:
Currently pending: Use a single parent `config_key` method which leverages a function to convert class name to snake case
https://github.com/getsentry/snuba/pull/3218#discussion_r987149879

snuba/datasets/message_filters.py
```diff
@@ -57,6 +84,10 @@ class KafkaHeaderSelectFilter(StreamMessageFilter[KafkaPayload]):
     header_key: str
     header_value: str
 
+    @classmethod
+    def config_key(cls) -> str:
+        return "KafkaHeaderSelectFilter"
```
volokluev, line 89: we can just set the `config_key` in the superclass to `cls.__name__`

snuba/datasets/message_filters.py
```diff
@@ -1,25 +1,48 @@
...
 
-class StreamMessageFilter(ABC, Generic[TPayload]):
+class StreamMessageFilter(ABC, Generic[TPayload], metaclass=RegisteredClass):
     """
     A filter over messages coming from a stream. Can be used to pre filter
     messages during consumption but potentially for other use cases as well.
     """
 
+    def __init__(self) -> None:
+        pass
```
MeredithAnya, line 26: why do we need this `__init__` if it doesn't do anything?

snuba/datasets/cdc/message_filters.py
```diff
@@ -1,32 +0,0 @@
...
-from arroyo.backends.kafka import KafkaPayload
-
-from snuba.datasets.message_filters import StreamMessageFilter
-
-KAFKA_ONLY_PARTITION = (
-    0  # CDC only works with single partition topics. So partition must be 0
-)
-
-
-class CdcTableNameMessageFilter(StreamMessageFilter[KafkaPayload]):
```
volokluev, line 11: so this file is just dead code?

snuba/datasets/message_filters.py
enochtangg: That was my initial thought as well, but wouldn't this decrease the grep-ability?

snuba/datasets/cdc/message_filters.py
enochtangg: The code in this file was consolidated to `snuba/datasets/message_filters.py` with the rest of the filters.

snuba/datasets/message_filters.py
volokluev: no because you can grep the class name and get the class

snuba/datasets/configuration/json_schema.py
```diff
@@ -13,7 +13,7 @@
 FUNCTION_CALL_SCHEMA = {
     "type": "object",
     "properties": {
-        "type": {"type": "string", "description": "Name of class key"},
+        "type": {"type": "string", "description": "Name of class config key"},
```
evanh, line 16: I don't think these changes are necessary anymore.

snuba/datasets/message_filters.py
```diff
@@ -69,7 +89,7 @@ def should_drop(self, message: Message[KafkaPayload]) -> bool:
 
 
 @dataclass
-class KafkaHeaderFilterWithBypass(KafkaHeaderFilter):
+class KafkaHeaderWithBypassFilter(KafkaHeaderFilter):
```
evanh, line 92: Why this rename?

enochtangg: All filter class names seem are suffixed with `Filter`. Renaming this class helps with consistency. This rename also does not change the understanding of what the pre-filter does.

snuba/datasets/configuration/json_schema.py
enochtangg: Updated, removed config_key

snuba/datasets/configuration/storage_builder.py
```diff
@@ -81,13 +81,15 @@ def build_stream_loader(loader_config: dict[str, Any]) -> KafkaStreamLoader:
...
-    pre_filter = (
-        CONF_TO_PREFILTER[loader_config[PRE_FILTER]["type"]](
+    pre_filter = None
+    if PRE_FILTER in loader_config and loader_config[PRE_FILTER] is not None:
+        pre_filter = StreamMessageFilter.get_from_name(
+            loader_config[PRE_FILTER]["type"]
+        ).from_kwargs(
             **loader_config[PRE_FILTER]["args"]
+            if loader_config[PRE_FILTER].get("args")
+            else {}
```
volokluev, line 91: This is a little cleaner

```suggestion
            loader_config[PRE_FILTER].get("args", {})
```

snuba/datasets/configuration/storage_builder.py
```diff
@@ -81,13 +81,15 @@ def build_stream_loader(loader_config: dict[str, Any]) -> KafkaStreamLoader:
     ).from_kwargs()
     default_topic = Topic(loader_config["default_topic"])
     # optionals
-    pre_filter = (
-        CONF_TO_PREFILTER[loader_config[PRE_FILTER]["type"]](
+    pre_filter = None
+    if PRE_FILTER in loader_config and loader_config[PRE_FILTER] is not None:
```
volokluev, line 85: looks like you've added extra logic. do you have a test for that?

snuba/datasets/message_filters.py
evanh: Sounds good.

snuba/datasets/configuration/storage_builder.py
enochtangg: This if statement logic is the same, but isn't written inline. Writing it this way save us a few extra lines of code.

PR #3221, opened 2022-10-04
feat(admin): Add Clickhouse Migrations page

MeredithAnya:
**context:**
We are working to get the Clickhouse Migrations tooling into snuba admin so that folks can do the following:
* run a migration in a given group
* reverse (or revert) a migration in a given group
* show the migration status for a group
where a group is defined in the [migrations group section](https://github.com/getsentry/snuba/blob/master/MIGRATIONS.md#migration-groups) in the migrations documentation.

This PR is meant to just have a very bare bones starting point for the UI portion of adding this tooling. The API endpoints that will be used will be done in subsequent PRs which is why for now, I've just created some dummy migration data in the `index.jsx` file. 

So far this is what we working with: 

> Screen Shot 2022-10-04 at 12 02 21 PM


evanh (approved):
You didn't want a dummy submit button? I am excited for this tool!

dbanda (approved):

PR #3222, opened 2022-10-04
feat(slices): Add Sliced Kafka Topics

ayirr7:
This PR:

- Adds the `SLICED_KAFKA_TOPIC_MAP` mapping, which can be thought of as analogous to the `SLICED_CLUSTERS` mapping. If a Topic is associated with multiple slices, it will be defined here. Here, keys are (logical topic name, slice id) tuples, and they are mapped to custom physical topic names. 
- Adds some logic to `validate_settings()` that allows us to make sure that any (logical topic name, slice id) tuple added to `SLICED_KAFKA_TOPIC_MAP` is also added to `SLICED_KAFKA_BROKER_CONFIG`
- Includes a test for above settings validation
- Defines `SLICED_KAFKA_BROKER_CONFIG` as a mapping of (logical topic name, slice id) tuples to broker configs
- Adds docs

Update: Per the discussion below and in channel, this PR was waiting for https://github.com/getsentry/snuba/pull/3226. It is now completed.


snuba/settings/validation.py
```diff
@@ -95,3 +95,14 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
...
             physical_part < locals["LOCAL_SLICES"]
         ), f"physical slice for logical partition {logical_part} is {physical_part}, but only {partition_count} physical slices exist"
+
+    # build a set of sliced Kafka logical topics
+    sliced_topics = set()
+    for topic_tuple in locals["SLICED_KAFKA_TOPICS"]:
+        sliced_topics.add(topic_tuple[0])
+
+    for logical_topic_name in sliced_topics:
+        for slice_num in range(0, locals["LOCAL_SLICES"]):
```
fpacifici, line 105: I may have missed something, but isn't the slices number a property of a dataset or storage? Why do we have a global `LOCAL_SLICES` setting ?

onewland: I'm thinking through what the level of effort would be to change this to be per-storage. For future extensibility it may be worth more complexity

onewland: FWIW I do think `LOCAL_SLICES` belongs in settings so that it is deployment-specific and not polluting the dataset configuration

snuba/settings/__init__.py
```diff
@@ -160,7 +165,7 @@
     "snuba-transactions-commit-log": "snuba-commit-log",
 }
 
-# Mapping of default Kafka topic name to broker config
+# Mapping of physical Kafka topic name to broker config
```
lynnagara, line 168: Why change this? The `default` was very important -> it is supposed to be the name hardcoded in the Snuba codebase/dataset config and NOT the physical name that the user might override

snuba/settings/__init__.py
```diff
@@ -144,6 +144,11 @@
 
 # Processor/Writer Options
 
+# Mapping of (logical topic names, slice id) pairs to physical topic names
+# This is for Kafka topics that are associated
+# with multiple slices
+SLICED_KAFKA_TOPICS: Mapping[Tuple[str, int], str] = {}
```
lynnagara, line 150: IS this supposed to be equivalent to `KAFKA_TOPIC_MAP` but for slices? If so please place it directly below the other variable and describe the interaction between the two settings. I also kind of wonder if we really need to support this right now. Why not just have every slice use the same topic names?

onewland: Separate physical topic names were raised by @fpacifici as a hard requirement, no compromise possible. Happy to consider the possibility of same topic names with different broker connections if my understanding was wrong

snuba/settings/validation.py
ayirr7: Updated PR (https://github.com/getsentry/snuba/pull/3226) addresses the removal of `LOCAL_SLICES` and the change to make slice number a property of storage

snuba/settings/__init__.py
lynnagara: That's fine, just a though. Please make the variable names mirror the original ones so the relationship is clearer though.

nikhars, line 150: Could you add an example usage of this setting in the partitioning docs?

snuba/settings/validation.py
```diff
@@ -95,3 +95,14 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
         assert (
             physical_part < locals["LOCAL_SLICES"]
         ), f"physical slice for logical partition {logical_part} is {physical_part}, but only {partition_count} physical slices exist"
+
+    # build a set of sliced Kafka logical topics
+    sliced_topics = set()
```
nikhars, line 100: I feel like validations for slicing should start moving to their own function now. There's a lot of validations happening just for slicing and it becomes hard to reason about them in non-sliced cases

snuba/settings/__init__.py
onewland: I think because self-hosted and single-tenant have using a single Kafka broker as a requirement (though to your point, it doesn't seem highly likely to me that they'll need slicing and not be able to bring in a second broker)

snuba/settings/__init__.py
ayirr7: This make sense. I added a new `SLICED_KAFKA_BROKER_CONFIG` that will deal with sliced topics specifically, and have physical/custom names as keys

snuba/settings/validation.py
ayirr7: Good point, done

ayirr7:
One thing I'm thinking about adding to validation is a check for whether there is a key for each (logical partition, slice) pair in `SLICED_KAFKA_TOPIC_MAP`.  Since slice number now varies by storage, this will require finding the corresponding storage(s) for each Kafka topic. Any thoughts on this?

onewland:
> One thing I'm thinking about adding to validation is a check for whether there is a physical topic name for each possible (logical partition, slice) pair in `SLICED_KAFKA_TOPIC_MAP`. Since slice number now varies by storage, this will require finding the corresponding storage(s) for each Kafka topic. Any thoughts on this?

I think this will have to be done on storage YAML load/initialization. If you see a simple way to do it in `storage_builder.py` then go for it, but if not I think it's probably OK to just fail hard on initialization for now (startup will fail if the sliced topic can't be found)

snuba/settings/__init__.py
```diff
@@ -278,6 +255,46 @@
...
+# to slice id
+LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {}
+
+# The slice configs below are the "SLICED" versions to
+# the equivalent default settings above. For example,
+# "SLICED_KAFKA_TOPIC_MAP" is the "SLICED" version of
+# "KAFKA_TOPIC_MAP". These should be filled out
+# for any corresponding sliced storages defined above,
+# with the applicable number of slices in mind.
+
```
onewland, line 274: nit: extra newline

snuba/settings/validation.py
```diff
@@ -90,6 +90,9 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
                 # that are not defined in StorageSetKey.
                 pass
 
+
+def validate_slicing_settings(locals: Mapping[str, Any]) -> None:
+
```
onewland, line 95: nit: extra newline

onewland (approved):
Looks good to me but I would ask Lyn to approve this since she was the most concerned with the approach here

snuba/settings/__init__.py
```diff
@@ -278,6 +255,46 @@
...
+# which are not sliced, i.e. are associated with
+# only the default slice_id (0). CLUSTERS is defined in
+# the default way, without adding slice id in
+# the storage_sets field. We define sliced clusters,
+# i.e. clusters that reside on multiple slices
+# in SLICED_CLUSTERS. We define all associated
+# (storage set, slice id) pairs in SLICED_CLUSTERS
+# in the storage_sets field. Other fields are defined
+# in the same way as they are in CLUSTERS.
+SLICED_CLUSTERS: Sequence[Mapping[str, Any]] = []
```
lynnagara, line 287: Don't these all need the sliced topic in the key like `SLICED_KAFKA_TOPIC_MAP` has? e.g. `SLICED_CLUSTERS: Sequence[Mapping[Tuple[str, int], Any]] = []`.

snuba/settings/__init__.py
lynnagara: Agree with this. Examples are much easier to parse than the long descriptions. The generic metrics use case would be a very good example to write in the comment since that is the first one we are gonna slice.

snuba/settings/__init__.py
```diff
@@ -278,6 +255,46 @@
...
+
+# Mapping of (logical topic names, slice id) pairs to custom physical topic names
+# This is only for sliced Kafka topics
+SLICED_KAFKA_TOPIC_MAP: Mapping[Tuple[str, int], str] = {}
+
+# Mapping of physical Kafka topic names to broker config
+# This is only for sliced Kafka topics
+SLICED_KAFKA_BROKER_CONFIG: Mapping[str, Mapping[str, Any]] = {}
```
lynnagara, line 295: Don't these all need the sliced topic in the key like `SLICED_KAFKA_TOPIC_MAP` has? e.g. `SLICED_KAFKA_BROKER_CONFIG: Mapping[Tuple[str, int], Mapping[str, Any]] = {}`. Otherwise how do you know which slice the config is for?

ayirr7: So the way it currently works is `SLICED_KAFKA_TOPIC_MAP` maps something like ("events", 1) to "events-1". And `SLICED_KAFKA_BROKER_CONFIG` maps "events-1" to [broker config].

snuba/settings/__init__.py
ayirr7: Yep, I'm on it now! I was waiting for discussions and to see if any other config would be changed before I write up in docs and add examples etc

ayirr7:
> If you see a simple way to do it in `storage_builder.py` then go for it, but if not I think it's probably OK to just fail hard on initialization for now (startup will fail if the sliced topic can't be found)

Sounds good. I think I'll keep validation in this PR the way it is, but I'll also look into adding any extra validation in storage_builder next.

snuba/settings/__init__.py
lynnagara: I don't think this is the right approach. Reasons:
1. It is very fragile. It assumes that the user has assigned unique topic names on all slices and breaks if they reused the same default name (which imo is the normal thing to do). Tuple of default topic and slice ID is guaranteed to be unique so is a much better choice to me
2. It basically now works the opposite of KAFKA_BROKER_CONFIG where the map is based on the default topic name. This is unintuitive as one would expect `SLICED_KAFKA_BROKER_CONFIG` to operate more like `KAFKA_BROKER_CONFIG`

lynnagara (changes requested):

snuba/settings/__init__.py
ayirr7: If I'm understanding correctly: no `SLICED_KAFKA_TOPIC_MAP`, and changing the keys of `SLICED_KAFKA_BROKER_CONFIG`. So for example:

`SLICED_KAFKA_BROKER_CONFIG` = {("snuba-generic-metrics", 1), BROKER_CONFIG}

How does this look to you?

lynnagara: Yes, this is what I was thinking. This is exactly how KAFKA_TOPIC_MAP works. The only extra bit is the 1.

lynnagara: No, that's not exactly what I was suggesting. You still probably want to keep the ability to remap all the topic names per slice so you want to keep the `SLICED_KAFKA_TOPIC_MAP` for that. But yes, change the `SLICED_KAFKA_BROKER_CONFIG` to that.

ayirr7: Ok, understood. Making that change now!

lynnagara (approved):
Looks good to me

snuba/settings/__init__.py
fpacifici: >  As for no compromise possible, curious why.

Compromise is always possible. Though here we would be compromising on the ability to consolidate Kafka clusters.
If we have two topics in the fleet with the same name, they will have to be on two separate brokers forever. 
For all practical purpose, Kafka is one of the most scalable components we have, likely we could have shared the broker between two slices. Sharing the name would prevent us from doing so.

So, taking a step back, allowing name conflicts between topics means the application imposing an operational constraint for no real reason. Keeping the names separate allows us to be flexible when we deploy in multiple environments as we can consolidate or split topic as we want.

snuba/settings/__init__.py
```diff
@@ -278,6 +255,45 @@
...
 
+# Slicing Configuration
+
+# Mapping of storage key to slice count
+# This is only for sliced storages
+SLICED_STORAGES: Mapping[str, int] = {}
+
+# Mapping storage key to a mapping of logical partition
+# to slice id
+LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {}
```
fpacifici, line 266: I think a direct mapping of logical partitions to slice id will be very impractical. 
There are hundreds of logical partitions (if we are talking about the logical shards from the original design). You probably want to map ranges of logical partitions to slice ids.:

```
Mapping[str, Mapping[Tuple[int, int], int]]
```

Or better:
```
PartitionRange = Tuple[int,int]
Mapping[str, Mapping[PartitionRange, int]]
```

snuba/settings/__init__.py
```diff
@@ -278,6 +255,45 @@
...
+# or SLICED_CLUSTERS. CLUSTERS will define clusters
+# which are not sliced, i.e. are associated with
+# only the default slice_id (0). CLUSTERS is defined in
+# the default way, without adding slice id in
+# the storage_sets field. We define sliced clusters,
+# i.e. clusters that reside on multiple slices
+# in SLICED_CLUSTERS. We define all associated
+# (storage set, slice id) pairs in SLICED_CLUSTERS
+# in the storage_sets field. Other fields are defined
+# in the same way as they are in CLUSTERS.
```
fpacifici, line 285: I am not sure how this is going to work. 
For a sliced cluster, if the slice id is an integer, why is the key of the mapping at line 286 a string ?

ayirr7: This typing follows from that of `CLUSTERS` above. The keys of the mappings are strings because these are the names of the fields. We kept it this way so that `SLICED_CLUSTERS` and `CLUSTERS` will have similar structure. There are examples in https://github.com/getsentry/snuba/blob/master/tests/clusters/test_cluster.py, but I can also try to revise this comment so that the definition/structure of `SLICED_CLUSTERS` is clearer.

fpacifici: Thanks for the clarification, but I think the question remains. 
How, in this data structure, is a cluster associated to a slice number? Is the slice number a field in the mapping ? Or is the slice number implied by the order in the sequence? If it is the second how would we manage multiple sliced storages as different storages can have different number of slices ?

ayirr7: I'm still thinking about your questions, but perhaps an example will make things clearer:

```
SLICED_CLUSTERS_CONFIG = [
    {
        "host": "host_slice",
        "port": 9000,
        "user": "default",
        "password": "",
        "database": "default",
        "http_port": 8123,
        "storage_set_slices": {("generic_metrics_distributions", 0)},
        "single_node": True,
    },
    {
        "host": "host_slice",
        "port": 9001,
        "user": "default",
        "password": "",
        "database": "slice_1_default",
        "http_port": 8124,
        "storage_set_slices": {("generic_metrics_distributions", 1)},
        "single_node": True,
    },
]
```
Here, the `storage_set_slices` field will contain tuples of the form (storage set key, slice id).

fpacifici: Ah ok, so the schema changes as the storage_set list contains the slice for each storage set. Thanks. That answers my question.

snuba/settings/__init__.py
onewland: I see the merit of the approach you're proposing but I want to push back on doing this right now.

1. We already have code that depends on this `LOGICAL_PARTITION_MAPPING` (it may look new in this PR but that's because Riya moved it to consolidate with other slicing settings) and
2. We don't really have slicing in place yet, and in my mind the risk for this project is not "can we represent configuration in a friendly way?" but "how can we get slicing working at all?". Making this change now takes us further from addressing the risk, not closer.
3. This configuration is functionally equivalent to what you're proposing

I just don't see a huge difference in operational burden between:

```python
LOGICAL_PARTITION_MAPPING = { 
  "generic-metrics-distributions": {
     **{ x: 0 for x in range(0,128) },
     **{ x: 1 for x in range(128,256) }
   }
}
```

and

```python
LOGICAL_PARTITION_MAPPING = { "generic-metrics-distributions": {(0, 128): 0, (128,256): 1}}
```

but to make your change now requires taking focus off functional slicing changes to address this (IMO) aesthetic concern. I'm happy to ticket doing this later though

onewland: Yeah, I've added a ticket to the epic (SNS-1766) to revisit this once we are more functionality-complete

PR #3226, opened 2022-10-05
feat(slices): Make slice count vary by storage

ayirr7:
This PR:

- Removes `LOCAL_SLICES` from settings
- Modifies the definitions of `LOGICAL_PARTITION_MAPPING` and `SLICED_STORAGE_SETS` accordingly
- Updates validation and tests
- Renames `SLICED_STORAGES` --> `SLICED_STORAGE_SETS` (point to potentially discuss)
- Updates docs

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -31,16 +31,37 @@ DISABLED_DATASETS: Set[str] = set()
 # Clickhouse Options
 CLICKHOUSE_MAX_POOL_SIZE = 25
 
-# The number of physical partitions that we will need to map resources to
-# for storage partitioning
-LOCAL_SLICES = 1
-# Mapping of logical (key) to physical (value) partitions for storages
-# that are partitioned in Snuba resources
-LOGICAL_PARTITION_MAPPING: Mapping[str, int] = {
-    str(x): 0 for x in range(0, SENTRY_LOGICAL_PARTITIONS)
+# Mapping storage set key to a mapping of logical partition
+# to slice id
+LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {
+    storage_set: {x: 0 for x in range(0, SENTRY_LOGICAL_PARTITIONS)}
+    for storage_set in {
+        "cdc",
+        "discover",
+        "events",
+        "events_ro",
+        "metrics",
+        "migrations",
+        "outcomes",
+        "querylog",
+        "sessions",
+        "transactions",
+        "transactions_ro",
+        "transactions_v2",
+        "errors_v2",
+        "errors_v2_ro",
+        "profiles",
+        "functions",
+        "replays",
+        "generic_metrics_sets",
+        "generic_metrics_distributions",
+    }
 }
-# Storage names to apply dataset partitioning to
-SLICED_STORAGES: Set[str] = set()
+
+# Mapping of storage set key to slice count
+# Only includes storage sets that are
+# assocated with multiple slices
+SLICED_STORAGE_SETS: Mapping[str, int] = {}
 
 CLUSTERS: Sequence[Mapping[str, Any]] = [
     {
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -5,7 +5,6 @@ from datetime import timedelta
 from pathlib import Path
 from typing import Any, Mapping, MutableMapping, Sequence, Set
 
-from snuba.datasets.partitioning import SENTRY_LOGICAL_PARTITIONS
 from snuba.settings.validation import validate_settings
 
 LOG_LEVEL = os.environ.get("LOG_LEVEL", "INFO")
@@ -31,37 +30,14 @@ DISABLED_DATASETS: Set[str] = set()
 # Clickhouse Options
 CLICKHOUSE_MAX_POOL_SIZE = 25
 
-# Mapping storage set key to a mapping of logical partition
+# Mapping storage key to a mapping of logical partition
 # to slice id
-LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {
-    storage_set: {x: 0 for x in range(0, SENTRY_LOGICAL_PARTITIONS)}
-    for storage_set in {
-        "cdc",
-        "discover",
-        "events",
-        "events_ro",
-        "metrics",
-        "migrations",
-        "outcomes",
-        "querylog",
-        "sessions",
-        "transactions",
-        "transactions_ro",
-        "transactions_v2",
-        "errors_v2",
-        "errors_v2_ro",
-        "profiles",
-        "functions",
-        "replays",
-        "generic_metrics_sets",
-        "generic_metrics_distributions",
-    }
-}
+LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {}
 
-# Mapping of storage set key to slice count
-# Only includes storage sets that are
+# Mapping of storage key to slice count
+# Only includes storages that are
 # assocated with multiple slices
-SLICED_STORAGE_SETS: Mapping[str, int] = {}
+SLICED_STORAGES: Mapping[str, int] = {}
 
 CLUSTERS: Sequence[Mapping[str, Any]] = [
     {
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -90,14 +90,13 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
                 # that are not defined in StorageSetKey.
                 pass
 
-    for storage in locals["LOGICAL_PARTITION_MAPPING"]:
-        storage_mapping = locals["LOGICAL_PARTITION_MAPPING"][storage]
+    for storage in locals["SLICED_STORAGES"]:
+        assert (
+            storage in locals["LOGICAL_PARTITION_MAPPING"]
+        ), "sliced mapping must be defined for sliced storage {storage}"
 
-        # check if this is a sliced storage
-        if storage in locals["SLICED_STORAGES"]:
-            defined_slice_count = locals["SLICED_STORAGES"][storage]
-        else:
-            defined_slice_count = 1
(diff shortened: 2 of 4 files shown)


snuba/settings/__init__.py
```diff
@@ -31,16 +31,37 @@
...
+        "transactions_ro",
+        "transactions_v2",
+        "errors_v2",
+        "errors_v2_ro",
+        "profiles",
+        "functions",
+        "replays",
+        "generic_metrics_sets",
+        "generic_metrics_distributions",
+    }
```
onewland, line 58: Is there a reason why we're doing this mapping by default, rather than only specifying it for sliced storages?

ayirr7: Right, this was something I wanted to clarify as well. Before, it was a general mapping of logical partitions to slices (and this would apply by default to all storage sets). My understanding was that same default meaning needs to be preserved with this new mapping, so this is why I included all of the storage sets. Let me know if I misunderstood something, though.

onewland:
Given that:

1. our approach uses the same storage set key regardless of which ClickHouse cluster is being accessed for a record (vs. before when we were thinking of dispatching by storage set key)
2. other aspects besides storage set need to be sliced upstream and downstream (specifically Kafka topics for ingestion and subcsriptions)

I'm wondering if the StorageKey should be the basis for slicing rather than the StorageSetKey. These generally have a 1:1 relationship but storage set is more specific to ClickHouse alone

snuba/settings/__init__.py
onewland: I think we can preserve this functionality without building the mapping explicitly. e.g. not all storage sets will be present in `LOGICAL_PARTITION_MAPPING`, code that relies on `LOGICAL_PARTITION_MAPPING` can assume any non-present keys only have a single slice

ayirr7:
In `get_cluster()` and related work, each sliced cluster is associated directly with "sliced" storage sets, not storages. It's analogous to `CLUSTERS` where each cluster has a `storage_set` field. Thus, a StorageSetKey and slice id are currently being used to identify and get a specific cluster. If we were to make StorageKey the basis of slicing, I think these would be identified by storages instead of by storage sets. I think it should be doable, but I'm wondering if there are any other impacts we should consider.

ayirr7:
> I'm wondering if the StorageKey should be the basis for slicing rather than the StorageSetKey. These often have a 1:1 relationship but storage set is more specific to ClickHouse alone and slicing is broader than that. Does this conflict with what we did in `get_cluster` etc.?

In get_cluster() and related work, each sliced cluster in SLICED_CLUSTERS is associated directly with sliced storage sets, not storages. It's analogous to CLUSTERS where each cluster has a storage_sets field. Thus, a StorageSetKey and slice id are currently being used to identify and get a specific cluster. If we were to make StorageKey the basis of slicing, clusters would be identified by storages instead of by storage sets. I think it should be doable, but I'm wondering if there are any other impacts we should consider. One thing I notice is that CLUSTERS and SLICED_CLUSTERS will no longer be same/analogous in the way they are identified.

onewland:
> In `get_cluster()` and related work, each sliced cluster in `SLICED_CLUSTERS` is associated directly with sliced storage sets, not storages. It's analogous to `CLUSTERS` where each cluster has a `storage_set` field. Thus, a StorageSetKey and slice id are currently being used to identify and get a specific cluster. If we were to make StorageKey the basis of slicing, clusters would be identified by storages instead of by storage sets.

Is there a problem in doing validation of the number of slices in SLICED_CLUSTERS which are addressed by (storage set key, slice ID) and not (storage key, slice ID)?

Because I think we can still slice on storage key upstream and whatever fetches the clickhouse cluster (query plan builder, Clickhouse writier for consumers) can know that it needs to pass a slice ID for the storage set key that it obtains from the storage.

snuba/settings/__init__.py
ayirr7: Ok, that makes sense. I can make it such that `LOGICAL_PARTITION_MAPPING` is empty in default settings.

ayirr7:
> Because I think we can still slice on storage key at a higher level and whatever fetches the clickhouse cluster (query plan builder for the API, Clickhouse writer for consumers) can know that it needs to pass a slice ID for the storage set key that it obtains from the storage.

Ok, this makes sense to me. Are you suggesting we keep the way `SLICED_CLUSTERS` works right now (being identified by storage_sets)? Also, are you suggesting any changes to `SLICED_STORAGE_SETS`?

onewland:
> > Because I think we can still slice on storage key at a higher level and whatever fetches the clickhouse cluster (query plan builder for the API, Clickhouse writer for consumers) can know that it needs to pass a slice ID for the storage set key that it obtains from the storage.
> 
> Ok, this makes sense to me. Are you suggesting we keep the way `SLICED_CLUSTERS` works right now (being identified by storage_sets)? Also, are you suggesting any changes to `SLICED_STORAGE_SETS`?

We'll see if Nikhar agrees with my comment in the other thread, maybe we want to get rid of this variable entirely, but if it remains I think it should be called `SLICED_STORAGES` and the types (str -> int) can remain the same

snuba/settings/__init__.py
```diff
@@ -31,16 +31,37 @@
...
 }
-# Storage names to apply dataset partitioning to
-SLICED_STORAGES: Set[str] = set()
+
+# Mapping of storage set key to slice count
+# Only includes storage sets that are
+# assocated with multiple slices
+SLICED_STORAGE_SETS: Mapping[str, int] = {}
```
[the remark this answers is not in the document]
nikhars: We can add a constraint which is checked during validations that slice number definitions should be a sequence for a given dataset. That way the problem you mentioned should not exist. 

The current problem is that you need to put a constraint both ways. That number of entries for a storage set in `LOGICAL_PARTITION_MAPPING` matches the number of slices defined in `SLICED_STORAGE_SETS`.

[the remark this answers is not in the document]
onewland: I'm fine with this though I think it gets a little more gnarly when we add `SLICED_KAFKA_TOPICS`. Then we're going to need to make sure the number of slice IDs represented in both are equal for a given storage (or storage set) key.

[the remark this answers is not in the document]
nikhars: I see your point. You need slices for different infrastructure resources. So having it at a common place makes validations simpler. Lets keep it this way then.

ayirr7:
> I think it should be called `SLICED_STORAGES` and the types (str -> int) can remain the same

Just so I'm clear, this means that the keys of `SLICED_STORAGES` and `LOGICAL_PARTITION_MAPPING` are now going to be StorageKeys, correct?

onewland:
> > I think it should be called `SLICED_STORAGES` and the types (str -> int) can remain the same
> 
> Just so I'm clear, this means that the keys of `SLICED_STORAGES` and `LOGICAL_PARTITION_MAPPING` are now going to be StorageKeys, correct?

that's right (though in practice these values are the same strings for our first storages, "generic-metrics-sets" and "generic-metrics-distributions")

snuba/settings/validation.py
```diff
@@ -86,12 +90,22 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
...
-            physical_part < locals["LOCAL_SLICES"]
-        ), f"physical slice for logical partition {logical_part} is {physical_part}, but only {partition_count} physical slices exist"
+    for storage in locals["LOGICAL_PARTITION_MAPPING"]:
+        storage_mapping = locals["LOGICAL_PARTITION_MAPPING"][storage]
+
+        # check if this is a sliced storage
+        if storage in locals["SLICED_STORAGES"]:
+            defined_slice_count = locals["SLICED_STORAGES"][storage]
+        else:
+            defined_slice_count = 1
```
onewland, line 100: I think it makes more sense to invert this a little and make the outer loop

```python
for storage in locals["SLICED_STORAGES"]:
   assert storage in locals["LOGICAL_PARTITION_MAPPING"], "sliced mapping must be defined for sliced storage {storage}"
   # validate all logical partitions are assigned to a reasonable slice
```

does that make sense or is there a reason it's structured this way?

onewland (commented):
Mostly good with this pending that question

snuba/settings/validation.py
ayirr7: Oh, right. So this was written originally under the assumption that all storages will be defined in `LOGICAL_PARTITION_MAPPING`, whether sliced or not. However, if I understand correctly, it is now such that both `LOGICAL_PARTITION_MAPPING` and `SLICED_STORAGES` are only defined for sliced storages. In this case, I'm not sure we can have validation for the non-sliced storages. They won't be in `LOGICAL_PARTITION_MAPPING`, so we can't check if they have valid logical:physical mappings.

onewland: yeah, I wouldn't bother worrying about the case where `LOGICAL_PARTITION_MAPPING` doesn't contain a storage key (assuming `SLICED_STORAGES` also does not contain that storage)

ayirr7: Ok, this makes sense. The inversion above also makes sense since `SLICED_STORAGES` is where we note down which storages are sliced. I'll make that change.

snuba/settings/validation.py
```diff
@@ -86,12 +90,21 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
...
+        defined_slice_count = locals["SLICED_STORAGES"][storage]
+
+        for logical_part in range(0, SENTRY_LOGICAL_PARTITIONS):
+            slice_id = storage_mapping.get(logical_part)
+
+            assert (
+                slice_id is not None
+            ), f"missing physical slice for storage {storage}'s logical partition {logical_part}"
+
+            assert slice_id < defined_slice_count, slice_count_validation_msg.format(
```
onewland, line 108: nit: let's change this condition `slice_id >= 0 and slice_id  < defined_slice_count`

onewland (approved):
Looks good to me. Thanks for hanging in through all the feedback

A small note: on merge, let's change the commit to say "vary by storage" rather than "vary by storage set"

PR #3227, opened 2022-10-05
ref(admin): add migration_groups API

MeredithAnya:
**context:**
A bit of context is in https://github.com/getsentry/snuba/pull/3221, but basically we want to add migrations tooling to the snuba admin tool. As seen in the previously mentioned PR, in order to know which migrations to run, a user must first pick a group. This endpoint is to populate that first dropdown of all the possible groups that one could run (or reverse) a migration for. 

**example payload:**

```json
{
    "group": "generic_metrics",
    "migration_ids": [
        "0001_sets_aggregate_table",
        "0002_sets_raw_table",
        "0003_sets_mv",
        "0004_sets_raw_add_granularities",
        "0005_sets_replace_mv",
        "0006_sets_raw_add_granularities_dist_table",
        "0007_distributions_aggregate_table",
        "0008_distributions_raw_table",
        "0009_distributions_mv"
     ]
}
```

Users can pick any migration to run (or reverse) since they could technically run each migration one by one if setting up a new cluster. The logic of whether a migration is appropriate to run will be handled when some tries to do a dry run or actually run the migration itself, and that endpoint and error handling will be done in a future PR. 

**ADMIN_ALLOWED_MIGRATION_GROUPS**:
There are certain groups that may not be safe to manage through the snuba admin tool (or at least not yet without some more fine grain permissions). So for now I've added a hardcoded list of migration groups that are allowed, and we can add to this once we audit the other groups. 

I'm also filtering by `active` groups, which is still TBD on if there is a case of for allowing folks to be able to run migrations on their own datasets that have infrastructure set up, but just aren't live for customers in any way.

This PR does not address filtering `migration_ids` by those that are considered dangerous, that will be addressed later once we've implemented a mechanism for determining said "dangerous" status.

### DIFF
--- a/snuba/admin/clickhouse/migration_groups.py
+++ b/snuba/admin/clickhouse/migration_groups.py
@@ -1,8 +1,0 @@
-from typing import Sequence, TypedDict
-
-from snuba.migrations.groups import MigrationGroup
-
-
-class MigrationGroupData(TypedDict):
-    group: MigrationGroup
-    migration_ids: Sequence[str]
--- a/snuba/admin/clickhouse/migration_groups.py
+++ b/snuba/admin/clickhouse/migration_groups.py
@@ -1,0 +1,8 @@
+from typing import Sequence, TypedDict
+
+from snuba.migrations.groups import MigrationGroup
+
+
+class MigrationGroupData(TypedDict):
+    group: MigrationGroup
+    migration_ids: Sequence[str]
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -1,16 +1,15 @@
 from __future__ import annotations
 
-from typing import Any, List, Optional, Sequence, Tuple, cast
+from typing import Any, List, Mapping, Optional, Sequence, Tuple, cast
 
 import simplejson as json
 import structlog
 from flask import Flask, Response, g, jsonify, make_response, request
 from structlog.contextvars import bind_contextvars, clear_contextvars
 
-from snuba import state
+from snuba import settings, state
 from snuba.admin.auth import UnauthorizedException, authorize_request
 from snuba.admin.clickhouse.common import InvalidCustomQuery
-from snuba.admin.clickhouse.migration_groups import MigrationGroupData
 from snuba.admin.clickhouse.nodes import get_storage_info
 from snuba.admin.clickhouse.predefined_system_queries import SystemQuery
 from snuba.admin.clickhouse.system_queries import run_system_query_on_host_with_sql
@@ -27,7 +26,7 @@ from snuba.datasets.factory import (
     get_dataset,
     get_enabled_dataset_names,
 )
-from snuba.migrations.groups import get_group_loader
+from snuba.migrations.groups import MigrationGroup, get_group_loader
 from snuba.migrations.runner import get_active_migration_groups
 from snuba.query.exceptions import InvalidQueryException
 from snuba.utils.metrics.timer import Timer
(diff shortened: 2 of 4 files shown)


volokluev:
test?

MeredithAnya:
> test?

Yeah I still need to add that

snuba/admin/clickhouse/migration_groups.py
```diff
@@ -0,0 +1,8 @@
+from typing import Sequence, TypedDict
+
+from snuba.migrations.groups import MigrationGroup
+
+
+class MigrationGroupData(TypedDict):
```
evanh, line 6: I would make this a dataclass/namedtuple instead of a TypedDict. I have a bit of a preference that way since TypedDict isn't quite as easy to type (you [can't do](https://mypy.readthedocs.io/en/stable/more_types.html#unions-of-typeddicts) `isinstance` checks).

It should JSON encode in exactly the same way.

lynnagara: I would not even add this type here. It's only used briefly in an intermediate step of processing the API response.

dbanda (approved):

evanh (approved):
Nice.

snuba/settings/__init__.py
```diff
@@ -23,6 +23,16 @@
...
+# Migrations Groups that are allowed to be managed
+# in the snuba admin tool.
+ADMIN_ALLOWED_MIGRATION_GROUPS = {
+    "system",
+    "generic_metrics",
+    "profiles",
+    "functions",
+    "replays",
+}
+
```
fpacifici, line 35: Any reason for not using the existing SKIPPED_MIGRATIONS_GROUPS
https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L206
Which is the flag that makes the migration tool ignore migration groups?

Or is this meant to serve a different purpose? 
Specifically how are we dealing with these use cases:
- a group that has not been vetted yet. That should be disabled to the tool entirely whether it is CLI or through admin.
- a group whose cluster is not ready. The dataset was merged by the cluster is not provisioned yet. Should these migrations be visible in the tool? Visible and disabled ?

snuba/admin/views.py
```diff
@@ -73,6 +75,16 @@ def health() -> Response:
     return Response("OK", 200)
 
 
+@application.route("/migrations/groups")
+def migrations_groups() -> Response:
+    res: List[Mapping[str, MigrationGroup | Sequence[str]]] = []
+    for migration_group in get_active_migration_groups():
+        if migration_group in settings.ADMIN_ALLOWED_MIGRATION_GROUPS:
+            group_migrations = get_group_loader(migration_group).get_migrations()
+            res.append({"group": migration_group, "migration_ids": group_migrations})
```
fpacifici, line 84: I think you will want to return the state of each migration otherwise how would you show to the client which migrations they can run and which ones have already been applied ?

snuba/settings/__init__.py
MeredithAnya: > Any reason for not using the existing `SKIPPED_MIGRATIONS_GROUPS`

I wanted to have a list that you have to explicitly define which groups are allowed, vs implicitly assuming all that are not in the set are allowed. I think by using `get_active_migration_groups` I'm actually using the `SKIPPED_MIGRATIONS_GROUPS` as well, but I wanted to limit more than just that. 

> Or is this meant to serve a different purpose?

`ADMIN_ALLOWED_MIGRATION_GROUPS` is meant to serve as a safeguard for the admin tooling specifically right now for datasets who we haven't determined are safe to run. (thinking of errors in particular). I think we will have to have a more robust system for permissions going forward.

> a group that has not been vetted yet. That should be disabled to the tool entirely whether it is CLI or through admin.

This is kind of what `ADMIN_ALLOWED_MIGRATION_GROUPS` is used for except only specifically for admin. I think if migrations are not to be run at all, via admin or cli, then they should go in `SKIPPED_MIGRATION_GROUPS`.

> a group whose cluster is not ready. The dataset was merged by the cluster is not provisioned yet. Should these migrations be visible in the tool? Visible and disabled ?

I think this is taken care of by `SKIPPED_MIGRATIONS_GROUPS`

snuba/admin/views.py
MeredithAnya: I thought about that, I think that this will depend on the cluster they want to run the migration for. For now I'm going to leave this since there are checks in the runner that will prevent users from running migrations that have already been run and such, but yeah I think it would be ideal to show only the migrations that someone would be able to run successfully

PR #3229, opened 2022-10-06
feat(slices): Add a mechanism to map queries to correct clusters

onewland:
For sliced storages, the API will have to dispatch queries to the correct ClickHouse clusters. This PR creates:
1. a new partitioning module for mapping org_id to slice_id given a storage key
2. a new query plan builder that uses the partitioning module to select a cluster for an incoming query request

This is additional work on top of https://github.com/getsentry/snuba/pull/3172 but re-opened so I could be the PR owner


snuba/datasets/plans/partitioned_storage.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Storage partition selector for the generic metrics storage. This is needed
+    because the generic metrics storage can be partitioned and we would need to
+    know which partition to use for a specific query.
+    """
+
+    def __init__(
+        self,
+        storage: StorageKey,
+        storage_set: StorageSetKey,
```
onewland, line 62: these are both derivable from the `ReadableTableStorage` object itself but it seemed a little cleaner to pass the keys around -- happy to change that though

snuba/datasets/partitioning.py
```diff
@@ -3,4 +3,32 @@
...
+    return org_id % SENTRY_LOGICAL_PARTITIONS
+
+
+def map_logical_partition_to_slice(storage: StorageKey, logical_partition: int) -> int:
+    """
+    Maps a logical partition to a slice.
+    """
+    from snuba.settings import LOGICAL_PARTITION_MAPPING
+
+    return LOGICAL_PARTITION_MAPPING[storage.value][logical_partition]
```
nikhars, line 25: This could potentially throw an exception in cases where someone calls this for a storage which is not partitioned. We should wrap this in a try/except and raise an exception

nikhars (approved):
Left some minor comment. You can remove some of the logic which is based off Riya's PR and then should be fine to merge.

snuba/datasets/partitioning.py
ayirr7: If I understand correctly, this will only be called in contexts where the storage is sliced?

onewland: I think I'm going to add assertions, this would be a misconfiguration and we shouldn't try to recover from that as an exception.

PR #3231, opened 2022-10-06
 ref(admin): add migration groups list endpoint

dbanda:
More context in #3227 

Add an endpoint `migrations/{group}/list` to show the migration for a group. Returns error 400 : invalid group  if the `group` is not found

e.g

```
curl http: //localhost:1219/migrations/replays/list 

[{
      "blocking": false,
      "migration_id": "0001_replays",
      "status": "completed"
}, .... {
      "blocking": false,
      "migration_id": "0006_add_is_archived_column",
      "status": "completed"
}]
```

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -28,7 +28,7 @@ from snuba.datasets.factory import (
     get_enabled_dataset_names,
 )
 from snuba.migrations.groups import MigrationGroup, get_group_loader
-from snuba.migrations.runner import get_active_migration_groups
+from snuba.migrations.runner import Runner, get_active_migration_groups
 from snuba.query.exceptions import InvalidQueryException
 from snuba.utils.metrics.timer import Timer
 from snuba.web.views import dataset_query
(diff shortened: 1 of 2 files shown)


evanh (commented):
I think you should also restrict this endpoint using the setting that Meredith added in https://github.com/getsentry/snuba/pull/3227.

To that end, I think actually the best practice here would be to add a middleware that checks the group being referenced is allowed, and add it to every endpoint you create.

tests/admin/test_api.py
```diff
@@ -15,6 +17,43 @@ def admin_api() -> FlaskClient:
...
+        },
+        {
+            "blocking": False,
+            "migration_id": "0003_sessions_matview",
+            "status": "completed",
+        },
+    ]
+
+    def sort_by_migration_id(migration: Any) -> Any:
+        return migration["migration_id"]
```
lynnagara, line 49: How come sorting by migration ID? Although by convention we use incrementing numbers for migration IDs, this is just a convention and the ordering is defined explicitly by the migration group.

snuba/admin/views.py
```diff
@@ -73,6 +75,27 @@ def health() -> Response:
     return Response("OK", 200)
 
 
+@application.route("/migrations/<group>/list")
```
lynnagara, line 78: Just a thought but would it be easier to return the list of all migrations by group? (basically equivalent to what the `snuba migrations list` command outputs)? There's really hardly any data, and the UI could easily filter them. The groups and ordering of those groups would also be present there so you wouldn't need to fire so many API calls.

tests/admin/test_api.py
dbanda: This was just to compare the lists in the assert.

snuba/admin/views.py
dbanda: that would be easier, but the plan is to have some access control so users can only list the allowed groups.

MeredithAnya (approved):
LGTM but I did add a separate testing file for `clickhouse_migrations` just so the `test_api` doesn't get to long if you wanted to add to https://github.com/getsentry/snuba/pull/3227/files

snuba/admin/views.py
```diff
@@ -73,6 +75,27 @@ def health() -> Response:
...
+def migrations_groups(group: str) -> Response:
```
nikhars, line 79: Is this supposed to be run on a per host basis? Like the system queries one where you select the host on which you want to run the command?

MeredithAnya: This should connect to the migrations cluster for which there is only one node as far as I know

PR #3232, opened 2022-10-07
feat(admin): Add extremely rudimentary kafka topic debugging

lynnagara:
This is to help debug an incident where a topic did not appear to exist in the metadata from the admin client.
Let's add some visibility in Snuba admin to try and figure out where it went.


snuba/admin/static/kafka/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+function TopicData(props: { api: Client }) {
+  const [data, setData] = useState<KafkaTopicData[] | null>(null);
+
+  useEffect(() => {
+    props.api.getKafkaData().then((res) => {
+      setData(res);
+    });
+  }, []);
+
+  return <div>{JSON.stringify(data)}</div>;
```
evanh, line 15: You could do `JSON.stringify(data, null, 4)` and it would at least auto indent the JSON blob. Otherwise this might be pretty hard to read, unless I'm missing some formatting somewhere.

evanh (approved):

snuba/admin/static/kafka/index.tsx
lynnagara: You're not, this is quick and dirty at like 10pm last night to help with an INC. But I assume the spaces would not have HTML line breaks in them so wouldn't work that well in browser anyway?

lynnagara: Ok it definitely doesn't work. But for other reasons. Unfortunately i had to cast whole parts of the response to a string on the server as it wasn't json serializable

evanh: Did a quick test, and yeah it doesn't really change anything in the browser. Up to you if you want to try and deal with that or not.

PR #3236, opened 2022-10-11
ref(MDC): Topics Enum to Class

enochtangg:
Similar to what we've done with `StorageKey` and `EntityKey`, we also want to use to metaclass design pattern to remove the Topic enum. However, the only difference between `Topic` and `StorageKey` and `EntityKey` is that the `Topic` key/value pairs are not uppercased/lowercased mirrors of each other. For example: `"QUERYLOG": "snuba-queries"`. As a result, when topics are registered through definition in config, they are assigned a different key name than the existing ones in code. Since topic keys are **only** used to reference topics in code and do not reflect real infrastructure (done with value pair), this was fixed by refactoring topic keys to mirror their respective value pairs.

* Deleted Topic enum and added Topic class and metaclass
* Refactored the topic keys to uppercased and underscored strings of value pair
* Updated all old references of topic keys

Note: In a subsequent PR, the generic metric topics can be removed from `_HARDCODED_TOPICS`

Edit:
* Move topics in `validation.py` to settings `ALLOWED_TOPICS`

### DIFF
--- a/snuba/datasets/storages/validation.py
+++ b/snuba/datasets/storages/validation.py
@@ -1,0 +1,28 @@
+from typing import Set
+
+from snuba import settings
+from snuba.utils.streams.topics import Topic
+
+
+class InvalidTopicError(ValueError):
+    pass
+
+
+def validate_storages() -> None:
+    validate_topics_with_settings()
+
+
+def validate_topics_with_settings() -> None:
+    """
+    This functoion validates topics specified in settings are valid
+    topics that have been loaded into the Topics registry (via storage builder).
+    """
+    topic_names: Set[str] = set([t.value for t in Topic])
+
+    for key in settings.KAFKA_TOPIC_MAP.keys():
+        if key not in topic_names:
+            raise InvalidTopicError(f"Invalid topic value: {key}")
+
+    for key in settings.KAFKA_BROKER_CONFIG.keys():
+        if key not in topic_names:
+            raise ValueError(f"Invalid topic value {key}")
--- a/snuba/attribution/log.py
+++ b/snuba/attribution/log.py
@@ -106,7 +106,7 @@ def record_attribution(attr_data: AttributionData) -> None:
         producer.poll(0)  # trigger queued delivery callbacks
         producer.produce(
             settings.KAFKA_TOPIC_MAP.get(
-                Topic.ATTRIBUTION.value, Topic.ATTRIBUTION.value
+                Topic.SNUBA_ATTRIBUTION.value, Topic.SNUBA_ATTRIBUTION.value
             ),
             data_str.encode("utf-8"),
             on_delivery=_record_attribution_delivery_callback,
--- a/snuba/datasets/configuration/storage_builder.py
+++ b/snuba/datasets/configuration/storage_builder.py
@@ -26,7 +26,7 @@ from snuba.datasets.table_storage import (
     build_kafka_stream_loader_from_settings,
 )
 from snuba.subscriptions.utils import SchedulingWatermarkMode
-from snuba.utils.streams.topics import Topic
+from snuba.utils.streams.topics import Topic, register_topic
 
 KIND = "kind"
 WRITABLE_STORAGE = "writable_storage"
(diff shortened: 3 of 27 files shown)


volokluev (approved):

evanh (approved):
Nicely done.

fpacifici (commented):
How is the topic name validation going to work when overriding topics or assigning them to a specific broker ? (settings `KAFKA_TOPIC_MAP` and `KAFKA_BROKER_CONFIG`). 
https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py#L31-L77

Today that one forces you to list all topics again because the settings validation code cannot import the `Topic` enum. With this change topics are not in the enum anymore, still the problem remains, the validation, as it is implemented today, prevents you from removing the topic names from python code. 
I think you will have to think how to validate `KAFKA_TOPIC_MAP` and `KAFKA_BROKER_CONFIG` at startup considering topics may be only stated in config.

There is a test that ensures topics from the enum and from the validator are in sync 
https://github.com/getsentry/snuba/blob/94c83c37d92d0e50ce768a65d72d55f03eb896e8/tests/settings/test_settings.py#L42-L60

evanh:
> How is the topic name validation going to work when overriding topics or assigning them to a specific broker ?

I think this list of "allowed" topics will have to move, either into a settings file or a configuration file. Really this validation is somewhat dubious to begin with, since it's not actually connected to Kafka at all. Really what should happen is the values in the settings/configuration should be validated in the actual environment to be sure they actually exist.

enochtangg:
The fact that I didn't know that another hardcoded list of topics existed in `validation.py` is insightful. I'd assume when other developers define new datasets/topics in config, they might also overlook this file. What do people think about moving this product specific code (topics set in `validation.py`) into our settings and calling it something like `ALLOWED_TOPICS`? This would provide better visibility and the ability to validate topics without having to the load storage factory.

Currently, this is what we do with `CLUSTERS` as all `storage_sets` are defined in settings.
As we eventually migrate all datasets to config,  HARDCODED_TOPICS will be obsolete anyways.

Also from a separate discussion with @evan.hicks, in a different task, we should look into replacing this validation system with something that checks against the real world.

fpacifici:
> Really what should happen is the values in the settings/configuration should be validated in the actual environment to be sure they actually exist.

That is only 50% of it. 
There are two ways to mess up the configuration:
- try to override a topic that does not exist in the Topic enum. Whether the topic is present in prod does not matter as the the topic you try to override is just the logical name. The worst part of this failure is that it would be silent, overriding would just not happen, and when you noticed it it would be too late as data would be in the wrong topic.
- try to provide an override topic that does not exist in the environment you are running Kafka on the broker. This is the case you mentioned.

fpacifici:
> What do people think about moving this product specific code (topics set in validation.py) into our settings and calling it something like ALLOWED_TOPICS? This would provide better visibility and the ability to validate topics without having to the load storage factory.

I think this kind of misses the point of what this validation is for. 
This is not meant to ensure that a topic you declare in your dataset is present in production. That would be a separate use case.
This validation is meant to ensure that if you declare a topic called `my_topic` in your storage (now config), you cannot make a mistake when you try to map that logical topic to a specific broker via settings. An example would writing this in the `KAFKA_BROKER_CONFIG` setting.
```
KAFKA_BROKER_CONFIG = {
    "my-topic": ..... # typo in the topic name.
}
```


The validation relied on all topics being hardcoded in the enum. Since now topics are created dynamically via config, why not moving this validation from the settings validation step to the dataset config validation? 
You cannot load storages from settings as it would cause circular dependencies. So why not making this check part of the storage validation? Once you loaded all data model and you have all topics created, ensure the settings are valid and do not reference a topic that does not exist in the `Topic` class?

evanh:
> Once you loaded all data model and you have all topics created, ensure the settings are valid and do not reference a topic that does not exist in the Topic class?

Is there a use case for having a topic in the settings that is not defined in configuration?

enochtangg:
Moved topics settings validation into a new validation step (`snuba/datasets/storages/validation.py`) after initializing storage factory. The unit test has been altered to ensure that topics specified in `KAFKA_TOPIC_MAP` and `KAFKA_BROKER_CONFIG` are present in the Topics registry (after loading from configs). Removed hardcoded list of topics in `validation.py`

snuba/datasets/storages/validation.py
```diff
@@ -0,0 +1,28 @@
+from typing import Set
```
evanh, line 1: I know we were planning to have a dedicated place for configuration validation to happen. If that exists, I think this should go there.

fpacifici:
> Is there a use case for having a topic in the settings that is not defined in configuration?

One at least. The querylog topic is hardcoded. Not sure about others. 
I am assuming you expect all datasets to be in configuration, otherwise now most topics are hardcoded as the datasets are defined in code.

fpacifici (commented):
Apart for the two hardcoded topics: querylog and attribution I think this will work. If you hardcode them in the Topic class it should be fine.

PR #3241, opened 2022-10-12
ref(migrations): add copy-tables script

MeredithAnya:
**context:**
Similar to https://github.com/getsentry/snuba/pull/3174 except that this script is independent of Snuba dependencies. The reason for adding this would be so that this script could be used by ops easily without having to run the snuba cli commands.

```shell
(.venv) ✔ ~/snuba [meredith/SNS-1753|⚑ 121]
17:20 $ python scripts/copy-tables.py --table=generic_metric_sets_aggregation_mv,generic_metric_distributions_aggregation_mv --source-host=localhost --target-host=localhost

Materialized View Check:
...checking table statement: generic_metric_sets_aggregation_mv
...looking for local tables: generic_metric_sets_local, generic_metric_sets_raw_local
...local tables found, mv check complete !


Materialized View Check:
...checking table statement: generic_metric_distributions_aggregation_mv
...looking for local tables: generic_metric_distributions_aggregated_local, generic_metric_distributions_raw_local
...local tables found, mv check complete !

creating generic_metric_sets_aggregation_mv...

create table statement:
 CREATE MATERIALIZED VIEW default.generic_metric_sets_aggregation_mv TO default.generic_metric_sets_local (`org_id` UInt64, `project_id` UInt64, `metric_id` UInt64, `granularity` UInt8, `timestamp` DateTime CODEC(DoubleDelta), `retention_days` UInt16, `tags.key` Array(UInt64), `tags.indexed_value` Array(UInt64), `tags.raw_value` Array(String), `value` AggregateFunction(uniqCombined64, UInt64), `use_case_id` LowCardinality(String)) AS SELECT use_case_id, org_id, project_id, metric_id, arrayJoin(granularities) AS granularity, tags.key, tags.indexed_value, tags.raw_value, toDateTime(multiIf(granularity = 0, 10, granularity = 1, 60, granularity = 2, 3600, granularity = 3, 86400, -1) * intDiv(toUnixTimestamp(timestamp), multiIf(granularity = 0, 10, granularity = 1, 60, granularity = 2, 3600, granularity = 3, 86400, -1))) AS timestamp, retention_days, uniqCombined64State(arrayJoin(set_values)) AS value FROM default.generic_metric_sets_raw_local WHERE (materialization_version = 1) AND (metric_type = 'set') GROUP BY use_case_id, org_id, project_id, metric_id, tags.key, tags.indexed_value, tags.raw_value, timestamp, granularity, retention_days

creating generic_metric_distributions_aggregation_mv...

create table statement:
 CREATE MATERIALIZED VIEW default.generic_metric_distributions_aggregation_mv TO default.generic_metric_distributions_aggregated_local (`org_id` UInt64, `project_id` UInt64, `metric_id` UInt64, `granularity` UInt8, `timestamp` DateTime CODEC(DoubleDelta), `retention_days` UInt16, `tags.key` Array(UInt64), `tags.indexed_value` Array(UInt64), `tags.raw_value` Array(String), `value` AggregateFunction(uniqCombined64, UInt64), `use_case_id` LowCardinality(String)) AS SELECT use_case_id, org_id, project_id, metric_id, arrayJoin(granularities) AS granularity, tags.key, tags.indexed_value, tags.raw_value, toDateTime(multiIf(granularity = 0, 10, granularity = 1, 60, granularity = 2, 3600, granularity = 3, 86400, -1) * intDiv(toUnixTimestamp(timestamp), multiIf(granularity = 0, 10, granularity = 1, 60, granularity = 2, 3600, granularity = 3, 86400, -1))) AS timestamp, retention_days, quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)(arrayJoin(distribution_values) AS values_rows) AS percentiles, minState(values_rows) AS min, maxState(values_rows) AS max, avgState(values_rows) AS avg, sumState(values_rows) AS sum, countState(values_rows) AS count, histogramState(250)(values_rows) AS histogram_buckets FROM default.generic_metric_distributions_raw_local WHERE (materialization_version = 1) AND (metric_type = 'distribution') GROUP BY use_case_id, org_id, project_id, metric_id, tags.key, tags.indexed_value, tags.raw_value, timestamp, granularity, retention_days
```

### DIFF
--- a/scripts/copy-tables.py
+++ b/scripts/copy-tables.py
@@ -1,5 +1,7 @@
 import argparse
 import re
+from collections import OrderedDict
+from typing import Optional, Sequence
 
 from clickhouse_driver import Client
 
@@ -11,7 +13,9 @@ SETTINGS = {
 }
 
 
-def _get_client(host, port, user, password, database) -> Client:
+def _get_client(
+    host: str, port: int, user: str, password: str, database: str
+) -> Client:
     return Client(
         host=host,
         port=port,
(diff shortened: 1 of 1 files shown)


evanh (approved):
LGTM. How have you tested it?

lynnagara (changes requested):
Why does it need to be independent of Snuba dependencies?

I think this logic should be part of the migration system. If we want to add another entrypoint here for internal reasons, that is fine by me but I would prefer the bulk of the logic also to be available via the migration system.

scripts/copy-tables.py
```diff
@@ -0,0 +1,113 @@
...
+
+        built_replica_path = f"{create_table_path}/replicas/{replica}"
+
+        assert (
+            built_replica_path == replica_path
+        ), f"{built_replica_path} should match zk path: {replica_path}"
+        print(f"Zookeeper replica paths verified for table: {table} ! :)")
+
+
+def copy_tables(source_client, new_client, tables, database, dry_run) -> None:
```
nikhars, line 58: Please add docstrings to this api.

scripts/copy-tables.py
```diff
@@ -0,0 +1,113 @@
...
+import re
+
+from clickhouse_driver import Client
+
+SETTINGS = {
+    "load_balancing": "in_order",
+    "replication_alter_partitions_sync": 2,
+    "mutations_sync": 2,
+    "database_atomic_wait_for_drop_and_detach_synchronously": 1,
+}
```
nikhars, line 11: Where did you get these settings from? Could you elaborate why each of them are needed?

scripts/copy-tables.py
```diff
@@ -0,0 +1,113 @@
...
+            f"SHOW CREATE TABLE {database}.{name}"
+        )
+        show_table_statments[name] = curr_create_table_statement
+
+    for table_name, statement in show_table_statments.items():
+        print(f"creating {table_name}...")
+        if dry_run:
+            print(f"create table statement: \n {statement}")
+            return
+        new_client.execute(statement)
```
nikhars, line 78: Please put this in an else block to make it explicit that the code is only run when dry_run is set to False.

scripts/copy-tables.py
```diff
@@ -0,0 +1,113 @@
...
+
+    source_client = _get_client(
+        host=args.source_host,
+        port=args.source_port,
+        user=args.user,
+        password=args.password,
+        database=args.database,
+    )
+
+    new_client = _get_client(
```
nikhars, line 103: Rename this to destination_client. I would recommend renaming everything `new` in the code to `destination` to make it clearer

scripts/copy-tables.py
```diff
@@ -0,0 +1,113 @@
...
+        print(f"creating {table_name}...")
+        if dry_run:
+            print(f"create table statement: \n {statement}")
+            return
+        new_client.execute(statement)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Copy ClickHouse Tables")
+    parser.add_argument("--source-host", default="localhost", help="clickhouse host")
```
nikhars, line 83: Remove this default. Make it a required argument

scripts/copy-tables.py
```diff
@@ -0,0 +1,113 @@
...
+    parser.add_argument("--source-port", help="source port")
+    parser.add_argument("--new-host", default="localhost")
```
nikhars, line 85: Same. Remove the default

nikhars:
> Why does it need to be independent of Snuba dependencies?
> 
> I think this logic should be part of the migration system. If we want to add another entrypoint here for internal reasons, that is fine by me but I would prefer the bulk of the logic also to be available via the migration system.

It's a conscious decision. The script is supposed to be pluggable with automation tools like salt to avoid manual steps. That means able to run from places where snuba necessarily does not exist.

lynnagara:
Right but this implementation still depends on Python libraries and a bunch of Python libraries. If you need to install stuff anyway, why not just install Snuba. If you wanted to make it easily runnable from anywhere then why not a standalone binary. This feels halfway between both options.

nikhars:
> Right but this implementation still depends on Python and a bunch of Python libraries. If you need to install stuff anyway, why not just install Snuba. If you wanted to make it easily runnable from anywhere then why not a standalone binary. This feels halfway between both options.

It is not easy to run Snuba from production machines. I will let @fpacifici chime in on the problems he faces while trying to run a snuba migration from one of our prod instances. Python dependency is fine. The idea is that this file could ultimately move into the ops repository without any sort of dependencies. Once it is in the ops repository, with only python dependencies, you can plug it into the Salt workflow more easily. @mwarkentin can provide more insights into that workflow

scripts/copy-tables.py
mwarkentin: We discussed also flipping it so dry-run is the default and you need to pass `--execute` (similar to `kafkactl`) to actually perform the copy.

mwarkentin:
This script is a helpful replacement for our current process of running `SHOW CREATE TABLE` on one replica and pasting the output into a new replica, so I agree that it should be simple to install wherever we need to run it from.

I think a standalone binary would be even better but I know that can be a bit painful to do in Python (is this a good case for a [Go script](https://github.com/ClickHouse/clickhouse-go)?)

Installing 1 dependency isn't too much to deal with, but might make sense to stick this into its own `scripts/copy-tables/*` directory and add a `requirements.txt` to simplify that further? We can do this when we move to `ops` as well if that's what ends up happening.

If there are cases where this might be used by both teams, sticking this into a `clickhouse-copy-tables` repo / binary might make sense.

scripts/copy-tables.py
MeredithAnya: I used `target` instead of destination, but I think that's better than new

scripts/copy-tables.py
MeredithAnya: made the default to run without `execute` but also flipped the check to explicitly do 

```python
if execute:
    target_client.execute(statement)

```

scripts/copy-tables.py
MeredithAnya: i copied this over from `ClickhouseClientSettings.MIGRATE` but just doubled checked all of them and none actually apply to this script so removing

scripts/copy-tables.py
```diff
@@ -0,0 +1,207 @@
...
+            print("\nMaterialized View Check:")
+            verify_local_tables_exist_from_mv(
+                target_client, curr_create_table_statement, name
+            )
+
+        if engine.startswith("Replicated"):
+            print("\nReplicated Table Check:")
+            verify_zk_replica_path(source_client, curr_create_table_statement, name)
+
+        show_table_statments[name] = curr_create_table_statement
```
nikhars, line 143: Is there some sort of ordering guarantees in the way create table commands are run if no tables are provided in the api call? Lets take an example of the materialized view. The sequencing is both (raw_table, aggregated_table) and then materialized_view. Does running this loop order that guarantee?

If not, you could build 2 dictionaries. One with the base tables and another with the materialized_views. And then execute all base table ones before the materialized views one.

scripts/copy-tables.py
```diff
@@ -0,0 +1,207 @@
...
+            f"creating {table_name}... on replica: {target_replica}, shard: {target_shard}"
+        )
+        if execute:
+            target_client.execute(statement)
+            print(f"created {table_name} !")
+            return
+        print(f"\ncreate table statement: \n {statement}\n")
+
+
+if __name__ == "__main__":
```
nikhars, line 159: Is there a way to test this functionality during CI? I would recommend spending sometime figuring out how this can be tested. An example could be bring up a cluster with 2 nodes and create some tables (raw and materialized) on one of the nodes. Then run the binary to copy those tables on the second node and verify that the tables exist.

dbanda: We could do something similar to test_distributed_migrations/test_runner_add_node.py where we add clickhouse--04 to the cluster. Also would be good to have unit tests for verify_zk_path functions to check the regex matching.

mwarkentin:
@MeredithAnya does this feel about ready to be merged? We've used it a few times successfully already.

MeredithAnya:
> @MeredithAnya does this feel about ready to be merged? We've used it a few times successfully already.

yeah I'm going to merge it, I'm going to follow up with testing in a separate PR

PR #3244, opened 2022-10-12
cleanup(transactions): remove transactions_ro from the codebase

volokluev:
We have learned that this is completely unnecessary so we're removing it. Everything will now go to `transactions` storage as god intended


nikhars:
Before removing it from code, I would suggest you remove it from [the ops prod_settings](https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L106). I am not sure if snuba deployments would have a problem if the code is gone but the settings are still there in prod.

lynnagara (approved):
I think the other way might actually fail (required storage set not being assigned) but please double check locally before merging

snuba/datasets/entities/transactions.py
```diff
@@ -94,43 +90,16 @@
...
-            query_plan_builder=SelectedStorageQueryPlanBuilder(
-                selector=TransactionsQueryStorageSelector(
-                    mappers=transaction_translator
-                    if custom_mappers is None
-                    else transaction_translator.concat(custom_mappers)
-                )
-            ),
+            query_plan_builder=SingleStorageQueryPlanBuilder(
+                storage=get_storage(StorageKey.TRANSACTIONS),
+                mappers=transaction_translator,
```
lynnagara, line 101: Actually you need that custom_mappers line? How do you know it's safe to remove?

lynnagara (changes requested):

volokluev:
> Before removing it from code, I would suggest you remove it from [the ops prod_settings](https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L106). I am not sure if snuba deployments would have a problem if the code is gone but the settings are still there in prod.

The problem is actually the other way around. If it's defined in code but not defined in prod settings, everything comes crashing down

snuba/datasets/entities/transactions.py
volokluev: you are right

tests/clusters/test_cluster.py
```diff
@@ -23,7 +23,7 @@
 
 ALL_STORAGE_SETS = {
     "outcomes",
-    "transactions_ro",
+    "transaction_ro",
```
lynnagara, line 26: Why not just delete this line?

lynnagara (approved):

tests/clusters/test_cluster.py
volokluev: I will, I was just making sure that having non-existent storage set keys in cluster settings wasn't going to break anything

PR #3224, opened 2022-10-04
cleanup(mdc): remove flag for enabling/disabling datasets from config

onewland:
This was just a temporary flag while we dialed this up, it's been on in prod for a while, and we're building on the new YAML-based entities.

PR #3245, opened 2022-10-12
fix(MDC): Incorrect column name in Gen Metrics

rahul-kumar-saini:
An incorrectly named column in Generic Metrics config is causing some queries to fail.

This PR fixes the issue ([SNS-1760](https://getsentry.atlassian.net/browse/SNS-1760))


onewland:
just added this to my PR: https://github.com/getsentry/snuba/pull/3224

onewland (approved):

volokluev:
Is there any way we could have caught this before going out into production

onewland:
> Is there any way we could have caught this before going out into production

I think it was a mistake to leave the feature flag in so long, had we removed it that would have broken sentry on snuba tests (as it did for my PR, which is how I discovered it)

PR #3247, opened 2022-10-13
feat(admin): run and reverse migrations

dbanda:
Adds endpoints to run and reverse migration via snuba admin. 

endpoints:
`/migrations/<group>/<run,reverse>/<migration_id>`

Clients can pass in the force, and fake params via the query url e.g `/migrations/system/run/0001_migrations?force=true`

If a migration is not in `settings.ADMIN_ALLOWED_MIGRATION_GROUPS` returns error 400, otherwise runs and returns `"OK", 200` if successful.

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -6,6 +6,7 @@ import simplejson as json
 import structlog
 from flask import Flask, Response, g, jsonify, make_response, request
 from structlog.contextvars import bind_contextvars, clear_contextvars
+from werkzeug.routing import BaseConverter
 
 from snuba import settings, state
 from snuba.admin.auth import UnauthorizedException, authorize_request
@@ -28,12 +29,10 @@ from snuba.datasets.factory import (
     get_enabled_dataset_names,
 )
 from snuba.migrations.groups import MigrationGroup, get_group_loader
-from snuba.migrations.runner import Runner, MigrationKey, get_active_migration_groups
+from snuba.migrations.runner import MigrationKey, Runner, get_active_migration_groups
 from snuba.query.exceptions import InvalidQueryException
 from snuba.utils.metrics.timer import Timer
 from snuba.web.views import dataset_query
-from werkzeug.routing import BaseConverter
-
 
 logger = structlog.get_logger().bind(module=__name__)
 
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -28,6 +28,7 @@ from snuba.datasets.factory import (
     get_dataset,
     get_enabled_dataset_names,
 )
+from snuba.migrations.errors import MigrationError
 from snuba.migrations.groups import MigrationGroup, get_group_loader
 from snuba.migrations.runner import MigrationKey, Runner, get_active_migration_groups
 from snuba.query.exceptions import InvalidQueryException
@@ -145,8 +146,17 @@ def run_or_reverse_migration(group: str, action: str, migration_id: str) -> Resp
             runner.reverse_migration(
                 migration_key, force=force, fake=fake, dry_run=dry_run
             )
-    except KeyError:
+    except KeyError as err:
+        logger.error(err, exc_info=True)
         return make_response(jsonify({"error": "Group not found"}), 400)
+    except MigrationError as err:
+        logger.error(err, exc_info=True)
+        return make_response(jsonify({"error": "migration error: " + err.message}), 400)
+    except ClickhouseError as err:
+        logger.error(err, exc_info=True)
+        return make_response(
+            jsonify({"error": "clickhouse error: " + err.message}), 400
+        )
 
     return Response("OK", 200)
 
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -136,14 +136,17 @@ def run_or_reverse_migration(group: str, action: str, migration_id: str) -> Resp
 
     force = request.args.get("force", False, type=str_to_bool)
     fake = request.args.get("fake", False, type=str_to_bool)
+    dry_run = request.args.get("dry_run", False, type=str_to_bool)
 
     try:
         if action == "run":
-            runner.run_migration(migration_key, force=force, fake=fake)
+            runner.run_migration(migration_key, force=force, fake=fake, dry_run=dry_run)
         else:
-            runner.reverse_migration(migration_key, force=force, fake=fake)
-    except Exception as e:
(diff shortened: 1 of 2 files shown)


snuba/admin/views.py
```diff
@@ -111,6 +112,42 @@ def migrations_groups_list(group: str) -> Response:
...
+    force = request.args.get("force", False, type=str_to_bool)
+    fake = request.args.get("fake", False, type=str_to_bool)
+
+    try:
+        if action == "run":
+            runner.run_migration(migration_key, force=force, fake=fake)
+        else:
+            runner.reverse_migration(migration_key, force=force, fake=fake)
+    except Exception as e:
+        return make_response(jsonify({"error": repr(e)}), 500)
```
evanh, line 146: You should also do `logger.error(...` here so that the error gets sent to Sentry.

snuba/admin/views.py
```diff
@@ -111,6 +112,42 @@ def migrations_groups_list(group: str) -> Response:
...
+    migration_key = MigrationKey(migration_group, migration_id)
+
+    def str_to_bool(s: str) -> bool:
+        return s.strip().lower() in ("yes", "true", "t", "1")
+
+    force = request.args.get("force", False, type=str_to_bool)
+    fake = request.args.get("fake", False, type=str_to_bool)
+
+    try:
+        if action == "run":
```
evanh, line 141: What happens if someone passes in something other than `run` or `reverse` as an action?

snuba/admin/views.py
```diff
@@ -111,6 +112,42 @@ def migrations_groups_list(group: str) -> Response:
...
+
+    runner = Runner()
+    migration_group = MigrationGroup(group)
+    migration_key = MigrationKey(migration_group, migration_id)
+
+    def str_to_bool(s: str) -> bool:
+        return s.strip().lower() in ("yes", "true", "t", "1")
+
+    force = request.args.get("force", False, type=str_to_bool)
+    fake = request.args.get("fake", False, type=str_to_bool)
```
evanh, line 138: Why not include dry run here as well?

There's also a part of me that thinks this tool should default to `dry_run=True`.

snuba/admin/views.py
evanh: It's best practice to only expose an error message in the API, and not the underlying Python logic.

tests/admin/clickhouse_migrations/test_api.py
```diff
@@ -91,3 +92,56 @@ def sort_by_migration_id(migration: Any) -> Any:
...
+        "snuba.settings.ADMIN_ALLOWED_MIGRATION_GROUPS", {"system", "invalid_group"}
+    ):
+        # invalid action
+        response = admin_api.post("/migrations/system/invalid_action/0001_migrations/")
+        assert response.status_code == 404
+
+        # invalid migration group
+        response = admin_api.post(f"/migrations/invalid_group/{action}/0001_migrations")
+        assert response.status_code == 500
+        assert json.loads(response.data) == {"error": "KeyError('invalid_group')"}
```
evanh, line 112: I mentioned this above but I think this is illustrative: the API should return a simple string message, not a stringified Python object. So `{"error": "group not found"}`

snuba/admin/views.py
dbanda: fixed

snuba/admin/views.py
dbanda: good point. added dry_run

tests/admin/clickhouse_migrations/test_api.py
dbanda: fixed this

snuba/admin/views.py
```diff
@@ -111,6 +113,54 @@ def migrations_groups_list(group: str) -> Response:
...
+    "/migrations/<group>/<migration_action:action>/<migration_id>",
+    methods=["POST"],
+)
+def run_or_reverse_migration(group: str, action: str, migration_id: str) -> Response:
+    if group not in settings.ADMIN_ALLOWED_MIGRATION_GROUPS:
+        return make_response(jsonify({"error": "Group not allowed"}), 400)
+
+    runner = Runner()
+    migration_group = MigrationGroup(group)
+    migration_key = MigrationKey(migration_group, migration_id)
```
fpacifici, line 133: How are we going to filter out and reject dangerous migrations ?

snuba/admin/views.py
```diff
@@ -111,6 +113,54 @@ def migrations_groups_list(group: str) -> Response:
...
+
+    def str_to_bool(s: str) -> bool:
+        return s.strip().lower() in ("yes", "true", "t", "1")
+
+    force = request.args.get("force", False, type=str_to_bool)
+    fake = request.args.get("fake", False, type=str_to_bool)
+    dry_run = request.args.get("dry_run", False, type=str_to_bool)
```
fpacifici, line 140: How does the user flow look like when doing a migration dry run? It does not seem you return anything to the user in any case, so I wonder what conclusion is the user going to draw from the result of a dry run.

snuba/admin/views.py
dbanda: The plan is to add some middleware that will check for safety. That will be in a separate PR.  This one just simply checks if the group is allowed.

snuba/admin/views.py
dbanda: Fair point. Right now dry run is a no-op. I'll fix by using `contextlib` to capture stdout the return to user. Hopefully that doesn't messup the logging in other parts.

dbanda: fixed

snuba/admin/views.py
```diff
@@ -111,6 +115,65 @@ def migrations_groups_list(group: str) -> Response:
...
+    def do_action() -> None:
+        if action == "run":
+            runner.run_migration(migration_key, force=force, fake=fake, dry_run=dry_run)
+        else:
+            runner.reverse_migration(
+                migration_key, force=force, fake=fake, dry_run=dry_run
+            )
+
+    try:
+        if dry_run:
```
evanh, line 153: Why not capture the output for every action? I think having more feedback will be better for the tool in all cases, but especially for errors.

MeredithAnya: agreed ^ probs good to always capture the stdout

dbanda: fixed it

MeredithAnya (approved):

PR #3250, opened 2022-10-13
feat: Remove default mapping of "transaction" topic to "events"

lynnagara:
Should be (roughly) synchronized with https://github.com/getsentry/sentry/pull/39988 which will send events to the new topic.

This would only affect self hosted and dev users who are using the Kafka backend. All Sentry prod environments have already been migrated via other methods to ensure zero downtime during the transition.


evanh (approved):
How will this affect ST customers?

lynnagara:
> How will this affect ST customers?

All ST customers were migrated a few weeks ago and are already running with these same overrides in their settings.

PR #3253, opened 2022-10-13
fix(TDC) Move TypeConverters into same folder as QueryProcessor

evanh:
The TypeConverter classes were in a separate folder from the rest of the query
processors, since they inherited from BaseTypeConverter, which is a subclass of
QueryProcessor.

This broke the registered class paradigm, since all the children of a registered
class need to be in the same folder. This solution seemed better than creating a
new registered class for type converters, since there is currently no way to
distinguish a type converter from a regular processor in the configuration.


volokluev (approved):

enochtangg (approved):

evanh:
I didn't save the original error, but I do know that when I added `reset_dataset_factory` to the test, it started passing.

That made me think, we should probably do that for every test just as a best practice. That way we can be assured that one set of configuration tests aren't messing up the global variable for a different test. That's why I created the base class.

PR #3254, opened 2022-10-13
ref(profiling): Drop profile column

phacops:
This was used to store profile data and has now been on GCS for a while. We don't need this column anymore.


nikhars:
Is this column being written in code anywhere right now? Or has populating this column in code already been removed?

phacops:
This column has not been written to in the last few months, there still was some code reading it but it's also being removed https://github.com/getsentry/vroom/pull/83.

nikhars (approved):
When applying the migration to your cluster please ensure that you remove the column from the distributed tables and then remove the column from the local tables.

PR #3256, opened 2022-10-13
feat(TDC): Make custom function processors accessible from config

volokluev:
The CustomFunctions processor takes python objects as arguments. Rather than trying to make those objects yaml-able, simply make its users be exposed query processors and make _CustomFunctions a private utility class.


snuba/query/processors/logical/custom_function.py
```diff
@@ -133,3 +143,37 @@ def apply_function(expression: Expression) -> Expression:
...
+    def __init__(self) -> None:
+        self.__processor = _CustomFunction(
+            "apdex",
+            [("column", ColType({UInt})), ("satisfied", LiteralType({int}))],
+            simple_function(
+                "divide(plus(countIf(lessOrEquals(column, satisfied)), divide(countIf(and(greater(column, satisfied), lessOrEquals(column, multiply(satisfied, 4)))), 2)), count())"
+            ),
+        )
+
+    def process_query(self, query: Query, settings: QuerySettings) -> None:
```
evanh, line 158: I think `process_query` can still be defined in the parent class.

evanh (approved):
One comment, otherwise looks good.

snuba/query/processors/logical/custom_function.py
volokluev: There's no parent class here. It's a wrapper class around the _CustomFunction processor

PR #3257, opened 2022-10-13
feat: Always pass a commit policy to Arroyo stream processor

lynnagara:
Soon this will become a mandatory argument

### DIFF
--- a/snuba/subscriptions/combined_scheduler_executor.py
+++ b/snuba/subscriptions/combined_scheduler_executor.py
@@ -5,6 +5,7 @@ from typing import Mapping, NamedTuple, Optional, Sequence, cast
 from arroyo import Message, Partition, Topic
 from arroyo.backends.abstract import Producer
 from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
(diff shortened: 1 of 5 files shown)


snuba/subscriptions/combined_scheduler_executor.py
```diff
@@ -109,9 +110,7 @@ def get_topic_configuration_for_entity(
     )
 
     return StreamProcessor(
-        tick_consumer,
-        Topic(commit_log_topic.topic_name),
-        factory,
+        tick_consumer, Topic(commit_log_topic.topic_name), factory, ONCE_PER_SECOND
```
evanh, line 113: Out of curiosity, why is this one `ONCE_PER_SECOND`? Was that defined somewhere else already? Based on the other changes I assumed `IMMEDIATE` was the default.

lynnagara: You're right, this is actually changing from IMMEDIATE to ONCE_PER_SECOND. This is because committing every offset to Kafka is a bad idea and we should actually throttle here. It's the only functional change in this PR.

evanh (approved):

PR #3258, opened 2022-10-13
ref(migrations): Add MigrationPolicy

MeredithAnya:
**context**
As we build out snuba admin tooling for clickhouse migrations, we want to make sure we have some way of deciding which migrations are actually "safe" to run in the admin tool, which may be different for various folks and migration groups. 

**MigrationPolicy**
To start out, each migration group can have a policy. The default policy (if we want a default can be the `NoMigrationsPolicy`.

```python
# these are just examples, not necessarily the
# levels we would give these groups
{
    "generic_metrics": NonBlockingMigrationsPolicy,
    "replays": AllMigrationsPolicy,
}
```
so that when we look up a migration group we know what policy to apply. Down the road we could expand this to have different access groups (google groups for example) which would have their own definitions for the groups

```python
# these are just examples, not necessarily the
# levels we would give these groups
{
    "generic_metrics": {
         "default": NoMigrationsPolicy,
         "team-sns": NonBlockingMigrationsPolicy,
    }
}
```

### DIFF
--- a/snuba/migrations/policies.py
+++ b/snuba/migrations/policies.py
@@ -7,12 +7,12 @@ from snuba.migrations.status import Status
 
 class MigrationPolicy(ABC):
     """
-    A MigrationPolicy implements an `is_allowed` method
-    that determines whether or not a migration can be applied
-    from snuba admin.
+    A MigrationPolicy implements `can_run` and
+    `can_reverse` methods that determines whether or not
+    a migration can be run or reversed from snuba admin.
 
     A policy can be used by a migration group to determine the
-    level of access a group has to run migrations.
+    level of access a group has to run/reverse migrations.
     """
 
     @abstractmethod
@@ -38,17 +38,12 @@ class ReadOnlyPolicy(MigrationPolicy):
 
 class WriteSafeAndPendingPolicy(MigrationPolicy):
     """
-    Safe migrations can be run in addition to dangerous migrations that
-    are in a pending state. All backwards migrations are considered
-    dangerous in this policy.
+    Migrations can be run that are non-blocking, as determined
+    by the `blocking` attribute set on a migration.
 
-    This policy defers to the `blocking` attribute for a migration to
-    determine if the migration is safe or not.
+    Reversing a migration is considered blocking by this policy,
+    and only migrations in a pending state can be reversed.
     """
-
-    def _get_status(self, migration_key: MigrationKey) -> Status:
-        status, _ = Runner().get_status(migration_key)
-        return status
 
     def can_run(self, migration_key: MigrationKey) -> bool:
         migration = get_group_loader(migration_key.group).load_migration(
@@ -57,7 +52,8 @@ class WriteSafeAndPendingPolicy(MigrationPolicy):
         return False if migration.blocking else True
 
     def can_reverse(self, migration_key: MigrationKey) -> bool:
-        if self._get_status(migration_key) == Status.IN_PROGRESS:
+        status, _ = Runner().get_status(migration_key)
+        if status == Status.IN_PROGRESS:
             return True
         return False
 
--- a/snuba/migrations/policies.py
+++ b/snuba/migrations/policies.py
@@ -1,13 +1,8 @@
 from abc import ABC, abstractmethod
-from enum import Enum
 
-from snuba.migrations.migration import Migration
+from snuba.migrations.groups import get_group_loader
+from snuba.migrations.runner import MigrationKey, Runner
 from snuba.migrations.status import Status
-
-
-class MigrationAction(Enum):
-    FORWARDS = "forwards"
-    BACKWARDS = "backwards"
 
 
 class MigrationPolicy(ABC):
@@ -21,7 +16,11 @@ class MigrationPolicy(ABC):
     """
 
     @abstractmethod
-    def allows(migration: Migration, action: MigrationAction) -> bool:
+    def can_run(self, migration_key: MigrationKey) -> bool:
+        raise NotImplementedError
+
+    @abstractmethod
+    def can_reverse(self, migration_key: MigrationKey) -> bool:
         raise NotImplementedError
 
 
@@ -30,7 +29,10 @@ class ReadOnlyPolicy(MigrationPolicy):
     No migration is allowed to be run or reversed.
     """
 
-    def allows(migration: Migration, action: MigrationAction) -> bool:
+    def can_run(self, migration_key: MigrationKey) -> bool:
+        return False
+
+    def can_reverse(self, migration_key: MigrationKey) -> bool:
         return False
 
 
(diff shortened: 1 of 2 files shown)


snuba/migrations/policies.py
```diff
@@ -0,0 +1,63 @@
...
+    that determines whether or not a migration can be applied
+    from snuba admin.
+
+    A policy can be used by a migration group to determine the
+    level of access a group has to run migrations.
+    """
+
+    @abstractmethod
+    def allows(migration: Migration, action: MigrationAction) -> bool:
+        raise NotImplementedError
```
lynnagara, line 25: How about
```suggestion
    def can_run(migration: Migration) -> bool:
        raise NotImplementedError
       
     def can_reverse(migration: Migration) -> bool:
        raise NotImplementedError
```
Since there's only two options the enum doesn't feel all that necessary. `allows` is a bit vague as it doesn't specify what you're being allowed to do

evanh: Is there  case we would want someone to have access to only one of these actions? Why split the access between run/reverse?

evanh (commented):
How do these policies work with the dry run attribute? It seems like users should be able to do dry run on migrations even if they are dangerous.

MeredithAnya:
> How do these policies work with the dry run attribute? It seems like users should be able to do dry run on migrations even if they are dangerous.

What would be the benefit of running a dry run for a migration that they couldn't run within the tooling? I think that maybe what we would do if we wanted that to be the case would be to skip the policy checking and just let them run it dry run, probs doesn't make sense to add it into the policies if the result would always be `True`.

snuba/migrations/policies.py
MeredithAnya: @lynnagara yeah could make sense to do. Also I think it might make sense to pass the `migration_key` instead of migration since getting the migration itself is not aware of it's own status.

evanh:
> What would be the benefit of running a dry run for a migration that they couldn't run within the tooling?

Testing and verification spring to mind. I don't feel strongly about having dry-run have the same restrictions though.

snuba/migrations/policies.py
```diff
@@ -0,0 +1,75 @@
+from abc import ABC, abstractmethod
+
+from snuba.migrations.groups import get_group_loader
+from snuba.migrations.runner import MigrationKey, Runner
+from snuba.migrations.status import Status
+
+
+class MigrationPolicy(ABC):
+    """
+    A MigrationPolicy implements an `is_allowed` method
```
lynnagara, line 10: Can you update the docstring?

snuba/migrations/policies.py
```diff
@@ -0,0 +1,75 @@
...
+    def can_run(self, migration_key: MigrationKey) -> bool:
+        migration = get_group_loader(migration_key.group).load_migration(
+            migration_key.migration_id
+        )
+        return False if migration.blocking else True
+
+    def can_reverse(self, migration_key: MigrationKey) -> bool:
+        if self._get_status(migration_key) == Status.IN_PROGRESS:
+            return True
+        return False
```
lynnagara, line 62: nit: I don't think the get_status method adds value
```suggestion
        return Runner().get_status(migration_key) == Status.IN_PROGRESS:
```

snuba/migrations/policies.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def can_reverse(self, migration_key: MigrationKey) -> bool:
+        return False
+
+
+class WriteSafeAndPendingPolicy(MigrationPolicy):
+    """
+    Safe migrations can be run in addition to dangerous migrations that
+    are in a pending state. All backwards migrations are considered
```
lynnagara, line 42: This docstring was a bit unclear to me:
- The migrations framework uses the word `blocking`, here you also introduce the word `safe` and `dangerous` but it's not really clear what these mean. I would prefer to standardise all terminology on blocking.
- I think it would be clearer to say that any migration in a pending state can be reversed. Your comment implies that both directions are ok for pending migrations when it is not.

snuba/migrations/policies.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    All migrations are allowed - both CodeMigrations and
+    ClickhouseNodeMigration (SQL) migrations
+    """
+
+    def can_run(self, migration_key: MigrationKey) -> bool:
+        return True
+
+    def can_reverse(self, migration_key: MigrationKey) -> bool:
+        return True
```
lynnagara, line 75: Reversing completed migrations should be very rare and not normal usage.

The CLI does allow this but requires an extra confirmation via passing a `--force` flag. I think it could be a good idea for the UI to do something similar, e.g. by requiring an additional confirmation step. Should we consider supporting a `force` arg in the policy as well for this scenario?

snuba/migrations/policies.py
MeredithAnya: I can clear up this doc string

snuba/migrations/policies.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+
+    def can_run(self, migration_key: MigrationKey) -> bool:
+        return False
+
+    def can_reverse(self, migration_key: MigrationKey) -> bool:
+        return False
+
+
+class WriteSafeAndPendingPolicy(MigrationPolicy):
```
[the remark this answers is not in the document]
MeredithAnya: but this isn't just running non-blocking migrations, it's also reversing migrations, which would then be considered blocking. I think because blocking is used to describe a whole migration vs the migration and the action it's confusing.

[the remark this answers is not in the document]
lynnagara: Yes, i suspect there is a better option than this. But `WriteSafeAndPendingPolicy` is confusing for me and it also doesn't capture all of these caveats that you mentioned.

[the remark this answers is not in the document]
MeredithAnya: The only thing I can think of would be like `RunNonBlockingReverseBlockingPolicy` which is kind of long (although `WriteSafeAndPendingPolicy` is already long)

snuba/migrations/policies.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Migrations can be run that are non-blocking, as determined
+    by the `blocking` attribute set on a migration.
+
+    Reversing a migration is considered blocking by this policy,
+    and only migrations in a pending state can be reversed.
+    """
+
+    def can_run(self, migration_key: MigrationKey) -> bool:
+        migration = get_group_loader(migration_key.group).load_migration(
```
lynnagara, line 49: What's the interaction of migration policies and groups? Since we are locking down access by group doesn't the policy need to check for access to the group before loading the migration?

MeredithAnya: Do you think the policy should keep the state of the group then? And have a separate policy for every group? even if it's the same policy type

I have the interaction described at the top but you could do something like look up the mapping and then create the policy with the group `policy = WriteAll("sessions")` or whatever in the endpoint where you are actually checking the policy

snuba/migrations/policies.py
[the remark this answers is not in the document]
lynnagara: Imo `RunNonBlocking` is both shorter and better captures the intent of the policy which is only to run the non blocking migrations. The ability to reverse the pending migration is only really in service of being able to revert what you just did via the tool in case it failed halfway. IMO this policy should actually only allow pending, non-blocking migrations to be reversed as that's all the tool allowed in the first place.

snuba/migrations/policies.py
MeredithAnya: update: I added this to the docstring but the assumption is that the access to the group has been verified before checking the policy. So the policy doesn't have the responsibility of verifying the access to the group

lynnagara (approved):
great, thanks!

PR #3259, opened 2022-10-14
feat(slices): Make consumers "slice-aware"

ayirr7:
This PR:

- Passes in slice id as a CLI argument to `snuba consumer`
- When relevant, converts the default logical topic to the physical topic (specific to the slice) using sliced topic configurations in settings. This mapping is performed for the input consumption topic and the commit log topic. 
- Mainly modifies the `ConsumerBuilder` logic to take into account any slice id that is passed in
- Builds upon the changes in the "Topic enum to class" [PR](https://github.com/getsentry/snuba/pull/3268), and uses `register_topic()` to register slice-specific physical topics that are defined in `SLICED_KAFKA_TOPIC_MAP` in settings

Next steps:
- Add some sort of verification that allows us to check whether the slice id being passed in is valid for the specific storage 
- Further testing


onewland:
Your change is currently not selecting the correct ClickHouse cluster for the given slice, so that needs to be done before a merge is possible (you can't rely on the changes we made to the query path).

Given that the `TableWriter` gives you access to the Kafka topic (indirectly via the stream loader) and selects the cluster, I wonder if you could pass through the slice ID to that object and hide more slice awareness from the consumer builder

https://github.com/getsentry/snuba/blob/efd9d420374f4fce054d4ccada84cd4e94c32087/snuba/datasets/table_storage.py#L243-L250

snuba/consumers/consumer_builder.py
```diff
@@ -93,12 +96,20 @@ def __init__(
             .topic
         )
 
+        if slice_id is not None:
+            # there should be some sort of assertion to check
+            # whether the slice_id passed is valid for this storage
+            physical_topic = SLICED_KAFKA_TOPIC_MAP[(topic.value, slice_id)]
+            # this will require registration of the physical topic
+            topic = register_topic(physical_topic)
```
onewland, line 104: I wonder if `register_topic` itself could validate whether the Kafka topic configuration exists and is present on the configured broker

snuba/consumers/consumer_builder.py
```diff
@@ -93,12 +96,20 @@ def __init__(
             .topic
         )
 
+        if slice_id is not None:
+            # there should be some sort of assertion to check
+            # whether the slice_id passed is valid for this storage
+            physical_topic = SLICED_KAFKA_TOPIC_MAP[(topic.value, slice_id)]
```
volokluev, line 102: Can you put this behind a function? this makes an assumption that the way logical topics and physical topics are mapped and is harder to test in case it changes

volokluev: in fact you can save a lot of if statements in this function if said function can take a `None` slice_id and just return the `topic.value`

snuba/consumers/consumer_builder.py
```diff
@@ -172,15 +196,29 @@ def __init__(
...
             .get_stream_loader()
             .get_default_topic_spec()
-            .topic,
+            .topic
+        )
+
+        # get the physical topic in case of slicing
+        if slice_id is not None:
+            physical_topic = SLICED_KAFKA_TOPIC_MAP[(topic.value, slice_id)]
+            topic = register_topic(physical_topic)
```
volokluev, line 215: You should only have to compute the `physical_topic` once throughout the consumer builder

snuba/datasets/storages/functions.py
```diff
@@ -59,7 +59,7 @@
     query_processors=[],
     stream_loader=build_kafka_stream_loader_from_settings(
         processor=FunctionsMessageProcessor(),
-        default_topic=Topic.PROFILES_FUNCTIONS,
+        default_topic=Topic.PROFILES_CALL_TREE,
```
volokluev, line 62: it seems like you have some o @enochtangg 's changes here. I think you need to merge master?

ayirr7: Fixed, I merged changes from his branch into mine.

ayirr7:
> Given that the `TableWriter` gives you access to the Kafka topic (indirectly via the stream loader) AND selects the cluster, I wonder if you could pass through the slice ID to `TableWriter`s creation or to `get_table_writer()` and hide more slice awareness from the consumer builder

In order to facilitate getting the cluster using slice id, I modified `get_table_writer()` and the associated superclass to take in slice id. This populates a TableWriter object's slice id attribute, which I added. Now, we can get a slice- and cluster-specific BatchWriter from the TableWriter object. 

An alternative to this would have been to modify the WritableTableStorage class to take in slice id upon initialization. However, it seemed to me that this would be a more invasive/refactoring-heavy change.

snuba/datasets/table_storage.py
```diff
@@ -225,6 +232,7 @@ def get_batch_writer(
         options: ClickhouseWriterOptions = None,
         table_name: Optional[str] = None,
         chunk_size: int = settings.CLICKHOUSE_HTTP_CHUNK_SIZE,
+        slice_id: Optional[int] = None,
```
fpacifici, line 235: If the table writer already contains the slice_id as part of its state `self.__slice_id` why would you need to pass the slice id when instantiating the batch writer ?

```
get_cluster(self.__storage_set, self.get_slice_id()).get_batch_writer(
```

On the other hand if the same table writer is supposed to be able to return a batch writer for multiple slices I don;t think you should have a `get_slice_id` method nor `add_slice_id`.

The big problem you have with this approach (`add_slice_id` method) is that you are making talbe writers (thus storages) mutable. This can have quite substantial negative ramifications. If the storage can mutate after initialization, it means we will have to be careful, any time we reference it ,to manage properly both scenarios: before mutations and after. 
We also need to be very careful to the storage lifecycle in order not to use the storage before the slice has been added.

Immutable data models simplify things a lot specifically when these objects are conceptually singletons. Making a singleton mutable tends to introduce a lot of hard to manage side effects.

onewland: I completely agree with Filippo's comments, especially because in this scenario we do not want `slice_id` to change within the lifetime of a process.

We should either figure out how to wire in the slice_id from the constructor (even if that means a bigger change) or require slice_id to be passed for every call of both `get_cluster` and `get_topic` (your original approach).

If you want to go with the constructor approach and you think it would help minimize complexity, I think it would be reasonable to split the work into:
- [refactor] PR 1: a mostly no-op change to take slice_id in the `TableWriter` constructor and store it, and accept that argument via the CLI
- [feature] PR 2: a functional change to actually use the new slice_id

ayirr7: In order to pass in slice_id into the `TableWriter` constructor, we will also have to change the `WritableTableStorage` constructor to take in slice_id. This will mean changing the way storages will be initialized, and the way they will be built from YAML config, etc. We will then need a new alternative to the current `get_writable_storage()` in the storage factory, perhaps something like `get_sliced_writable_storage()` and changes to `_StorageFactory` as well to include sliced storages. I just wanted to make sure that these changes are expected and seem reasonable/okay for me to implement? Or perhaps is there an alternative, simpler approach that I'm missing?

fpacifici: The alternative is that the storage is initialized in an agnostic way to the slice id and then the slice id is passed when you retrieve any resource that is slice specific, here there would be many possible ways depending which level should be slice specific: `get_table_writer` or within the `TableWriter` `get_batch_writer`, etc.

If instead you want to make the storage initialization slice aware, that can make all further usages simpler as no slice id parameter is needed anymore.

Which one is better depends on where and when you have the `slice_id` available, specifically you will need to look into both the query pipeline and the consumer since both rely on the storage to know where to read from and where to write to.

snuba/consumers/consumer_builder.py
ayirr7: Are you suggesting making the physical topic (and therefore, also the logical topic) an instance attribute?

snuba/datasets/storage.py
```diff
@@ -167,9 +168,12 @@ def __init__(
             writer_options=writer_options,
             write_format=write_format,
         )
+
         self.__ignore_write_errors = ignore_write_errors
 
     def get_table_writer(self) -> TableWriter:
+        # if slice_id is not None:
+        #     self.__table_writer.add_slice_id(slice_id)
```
onewland, line 176: this comment should be removed

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,33 @@ class MockParameters:
     std_deviation: int
 
 
+def get_sliced_snuba_topic(
+    topic: SnubaTopic, slice_id: Optional[int] = None
+) -> SnubaTopic:
+
```
onewland, line 71: nit: extra newline

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,33 @@ class MockParameters:
...
+        # this will require registration of the physical topic
+        topic = register_topic(physical_topic)
+
+    return topic
+
+
+def get_sliced_arroyo_topic(
+    passed_topic: str, slice_id: Optional[int] = None
+) -> ArroyoTopic:
+
```
onewland, line 85: nit: extra newline

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,33 @@ class MockParameters:
...
+def get_sliced_snuba_topic(
+    topic: SnubaTopic, slice_id: Optional[int] = None
+) -> SnubaTopic:
+
+    if slice_id is not None:
+        # there should be some sort of assertion to check
+        # whether the slice_id passed is valid for this storage
+        physical_topic = SLICED_KAFKA_TOPIC_MAP[(topic.value, slice_id)]
+        # this will require registration of the physical topic
+        topic = register_topic(physical_topic)
```
onewland, line 77: let's not reassign `topic`, we could just `return register_topic(physical_topic)` here since we're not doing anything else with it

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,33 @@ class MockParameters:
     std_deviation: int
 
 
+def get_sliced_snuba_topic(
```
onewland, line 68: Since this and the method below work on sliced and non-sliced topics, should they just be called `get_snuba_topic/get_arroyo_topic`? The majority of calls will be for non-sliced topics

snuba/consumers/consumer_builder.py
ayirr7: I'm not sure if I can modify `register_topic` for this, but I think I can add the validation in the consumer builder itself as soon as it has the storage and slice_id.

onewland (commented):
pending my comments, I think I'm fine with this approach.

Have you tried testing locally by creating a sliced cluster/topic, manually starting a consumer, and seeing if you can write to slice 1?

snuba/consumers/consumer_builder.py
ayirr7: It's true the methods work also for non-sliced topics. I think it could be helpful to call them something like `get_physical_snuba_topic`/`get_physical_arroyo_topic` though?

snuba/consumers/consumer_builder.py
volokluev: I would say yes, unless I'm confused about how the consumer builder works and it builds many consumers per one instance of it

snuba/consumers/consumer_builder.py
onewland: sounds good

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,39 @@ class MockParameters:
...
 
 
+def validate_passed_slice(
+    storage_key: StorageKey, slice_id: Optional[int] = None
+) -> None:
+    # make sure this is a storage that is sliced
+    # and that the slice id passed in is valid
+    if slice_id is not None:
+        assert storage_key.value in SLICED_STORAGES
+        assert slice_id < SLICED_STORAGES[storage_key.value]
```
lynnagara, line 75: I think all of this validation could also be part of the centralised KafkaTopicSpec implementation rather than requiring each piece of code to do it on their own

ayirr7:
> Have you tried testing locally by creating a sliced cluster/topic, manually starting a consumer, and seeing if you can write to slice 1?

A few updates to this PR: 

Tested consumer behavior using the `snuba-generic-metrics` topic on slice_id = 1 (physical topic `snuba-generic-metrics-1`). I was able to successfully produce to and consume from `snuba-generic-metrics-1`, with the stroage being `generic_metrics_distributions_raw`. I've written up the configuration changes and commands I used for testing on a Notion doc, which I am happy to share. 

Removed the logic for mapping logical Snuba topics to physical/sliced topics. Snuba topics are only used for retrieving Kafka broker config in the Consumer builder, and broker config can be found using simply a (logical topic name, slice id) pair. 

Kept the logic for mapping logical Arroyo topics to physical/sliced topics, and passed this sliced topic into StreamProcessor logic so that the consumer connects to this correct topic (this is `snuba-generic-metrics-1` for our example above). 

Finally, made a modification which may be a small typo in the Topic enum to class PR (since this PR builds on that PR).

snuba/consumers/consumer_builder.py
```diff
@@ -217,10 +260,15 @@ def __build_consumer(
                 commit_retry_policy=self.__commit_retry_policy,
             )
 
-        return StreamProcessor(consumer, self.raw_topic, strategy_factory, IMMEDIATE)
+        # Can we assume that the raw topic would be mapped
+        # to a physical topic in the sliced context?
```
onewland, line 264: do you know the answer to this question? seems like it should be "yes", and if so we can remove the comment

snuba/datasets/storage.py
```diff
@@ -159,6 +159,7 @@ def __init__(
             mandatory_condition_checkers,
         )
         assert isinstance(schema, WritableTableSchema)
+
```
onewland, line 162: nit: did `flake8` force these changes? I'd rather not commit only whitespace changes in a file

snuba/consumers/consumer_builder.py
onewland: This seems like it's no longer about Kafka topic validation -- I'm guessing Lyn's comment no longer applies?

onewland (commented):
Looks like we're going in the right direction to me. Thanks for doing more in-depth testing! It might be annoying but it's definitely worse to find errors post-deploy

snuba/consumers/consumer_builder.py
ayirr7: In the new approach, `physical_topic` no longer exists in the constructor. So, I think this suggestion would no longer be possible/needed?

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,39 @@ class MockParameters:
...
+def get_sliced_arroyo_topic(
+    passed_topic: str, slice_id: Optional[int] = None
+) -> ArroyoTopic:
+    if slice_id is not None:
+        physical_topic = SLICED_KAFKA_TOPIC_MAP[(passed_topic, slice_id)]
+        set_topic = ArroyoTopic(physical_topic)
+    else:
+        set_topic = ArroyoTopic(passed_topic)
+
+    return set_topic
```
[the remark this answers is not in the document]
ayirr7: `KafkaTopicSpec` currently only works with Snuba topics. This PR has been updated to only include `get_physical_arroyo_topic()` (the Snuba equivalent was removed). Are you suggesting to make `KafkaTopicSpec` compatible with both topic types? i.e.something like:

```class KafkaTopicSpec:
    def __init__(self, snuba_topic: SnubaTopic, arroyo_topic: ArroyoTopic) -> None:
        self.__snuba_topic = snuba_topic
        self.__arroyo_topic = arroyo_topic
```

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +63,28 @@ class MockParameters:
     std_deviation: int
 
 
+def validate_passed_slice(
+    storage_key: StorageKey, slice_id: Optional[int] = None
+) -> None:
+    # make sure this is a storage that is sliced
```
MeredithAnya, line 69: should this be a docstring instead of a comment? also it was not obvious to me why 

```
assert slice_id < SLICED_STORAGES[storage_key.value]
```

was used, maybe you could say something like "make sure the slice id passed in is less than the number of total slices for the given storage" 
maybe its obvious for others tho

snuba/consumers/consumer_builder.py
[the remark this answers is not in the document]
lynnagara: It's not so much about SnubaTopic vs ArroyoTopic, more about logical vs physical topic and how it gets resolved it a common way.`KafkaTopicSpec` is supposed to be the singular way of resolving a Snuba logical topic name to a physical topic, via `KafkaTopicSpec.topic_name`. This is useful so that all code in Snuba can just reference the Snuba topic and get resolved in this exact same way and we don't need separate resolving functions like this in many parts of our codebase. This seems like it solves a similar problem as you are trying to do here so I would prefer we find a way to adapt that to make it work for sliced topics as well.

snuba/consumers/consumer_builder.py
lynnagara: Yes. Although I would still encourage you to think about whether you think this code should be reusable and part of the common slice validation logic rather than specific to the consumer builder.

snuba/consumers/consumer_builder.py
ayirr7: That's a good point, I can write a docstring that clarifies why these validations are in place.

snuba/utils/streams/configuration_builder.py
```diff
@@ -32,14 +41,15 @@ def get_default_kafka_configuration(
...
+    slice_id: Optional[int] = None,
     auto_offset_reset: Optional[str] = None,
     queued_max_messages_kbytes: Optional[int] = None,
     queued_min_messages: Optional[int] = None,
     bootstrap_servers: Optional[Sequence[str]] = None,
     override_params: Optional[Mapping[str, Any]] = None,
     strict_offset_reset: Optional[bool] = None,
 ) -> KafkaBrokerConfig:
-    default_topic_config = _get_default_topic_configuration(topic)
+    default_topic_config = _get_default_topic_configuration(topic, slice_id)
```
nikhars, line 52: You can simplify this a bit. This logic can be moved out to the caller and you only pass in the topic config to this api. So you don't need to pass in `topic` and `slice_id` here. This just reduces the amount of parameters which are being passed around.

snuba/datasets/table_storage.py
```diff
@@ -57,6 +58,14 @@ def topic_creation_config(self) -> Mapping[str, str]:
     def __eq__(self, other: object) -> bool:
         return isinstance(other, KafkaTopicSpec) and self.topic == other.topic
 
+    def get_physical_topic_name(self, slice_id: Optional[int] = None) -> str:
```
nikhars, line 61: Can you add a test case for this?

snuba/utils/streams/configuration_builder.py
```diff
@@ -10,19 +10,28 @@
 from snuba.utils.streams.types import KafkaBrokerConfig
 
 
-def _get_default_topic_configuration(topic: Optional[Topic]) -> Mapping[str, Any]:
+def _get_default_topic_configuration(
```
nikhars, line 13: Please add a test case for this.

snuba/consumers/consumer_builder.py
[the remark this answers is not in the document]
ayirr7: I updated the PR to use `KafkaTopicSpec` for resolving the logical to sliced physical topic name

snuba/consumers/consumer_builder.py
```diff
@@ -62,6 +65,19 @@ class MockParameters:
...
+    """
+    Verifies that the given storage can be sliced
+    and that the slice_id passed in is within the range
+    of the total number of slices for the given storage
+    """
+    if slice_id is not None:
+        assert storage_key.value in SLICED_STORAGES
+        assert slice_id < SLICED_STORAGES[storage_key.value]
+
+
```
lynnagara, line 80: Please do this inside the shared `KafkaTopicSpec` implementation as well. There is no reason to implement it multiple times inside every entrypoint.

snuba/consumers/consumer_builder.py
```diff
@@ -109,29 +129,42 @@ def __init__(
 
         stream_loader = self.storage.get_table_writer().get_stream_loader()
 
-        self.raw_topic: Topic
+        self.raw_topic: ArroyoTopic
```
lynnagara, line 132: Personal preference, I'd rather you just left all these names.

snuba/consumers/consumer_builder.py
```diff
@@ -109,29 +129,42 @@ def __init__(
...
         if kafka_params.raw_topic is not None:
-            self.raw_topic = Topic(kafka_params.raw_topic)
+            # we do not support topic override in a sliced context
+            assert slice_id is None
```
lynnagara, line 135: Why not?

snuba/utils/streams/configuration_builder.py
```diff
@@ -10,19 +10,28 @@
...
+    topic: Optional[Topic], slice_id: Optional[int] = None
+) -> Mapping[str, Any]:
+
     if topic is not None:
-        return settings.KAFKA_BROKER_CONFIG.get(topic.value, settings.BROKER_CONFIG)
+        if slice_id is not None:
+            return settings.SLICED_KAFKA_BROKER_CONFIG.get(
+                (topic.value, slice_id), settings.BROKER_CONFIG
```
lynnagara, line 20: Please make slice=0 behave the same as slice=None. The subtle differences here make it too easy to pass the wrong values.

lynnagara (changes requested):
Thanks, I think this is looking better. My main comments are around consistency. We need to make the slices behave the same as slice 0, currently there are too many subtle differences.

- Passing slice=0 should work, and it should behave the same as slice=None, or not passing the slice ID. Currently it behaves in pretty unexpected ways and you will probably get something broken if you pass slice=0.
- We need to support the topic overrides and the same features in general for slices. How come you can't override a topic if you are using a slice? I think that code for various slices should behave the same, these small differences are very unexpected
- Validation should also be done in the shared KafkaTopicSpec so it will be automatically run every time we fetch the slice data not implemented and called independently in every separate entrypoint, of which there will be many.

snuba/consumers/consumer_builder.py
ayirr7: I'm not sure how this would be moved into `KafkaTopicSpec`. This validation has no relationship with any Topic/Topics as a concept. This is simply checking whether a slice_id (an int) and a corresponding storage_key are "compatible' (using those two assert statements). There is no reference to Topic maps or creation of sliced topics here. In order to put this into `KafkaTopicSpec`, we would need to pass storage_key into some function of `KafkaTopicSpec` -- does that make sense to you?

snuba/consumers/consumer_builder.py
ayirr7: The issue with leaving the names is the potential confusion between (Snuba) `Topic` and an (Arroyo) `Topic`. If both are `Topic` (how it was originally), it can become confusing to work with the code in future iterations.

snuba/consumers/consumer_builder.py
ayirr7: This was part of a discussion I had with Oliver, and he'd said that we do not plan to support this as of now.

snuba/utils/streams/configuration_builder.py
ayirr7: This does indeed work the same way between slice=0, slice=1, etc. for sliced topics. If you look into `settings/__init__.py`, you will see under the Slicing Configuration section, all of these configs/maps are for sliced topics/storages, i.e. those that are associated with multiple slices. So, in this code snippet, as long as the topic being passed in is one that is indeed defined to be "sliced", `SLICED_KAFKA_BROKER_CONFIG` will contain information for it's slice 0, slice 1, and so on. I hope that makes sense

lynnagara: I'm not sure I really understand the slice configuration. It seems ambiguous in many cases and so there are scenarios where it doesn't work the same way. For example, what if the user puts the slice 0 configuration in KAFKA_BROKER_CONFIG then they add the slice 1 configuration in SLICED_KAFKA_BROKER_CONFIG. If that is not valid, it should be validated and prevented but there is nothing preventing the user from doing that now.

Now, I would expected passing slice=0 to fall back to the KAFKA_BROKER_CONFIG but it does not. Instead it goes back to BROKER_CONFIG. This is pretty unexpected to me.

snuba/consumers/consumer_builder.py
lynnagara: Since we already support topic override on unsliced storages, the effort to do so on a sliced storage which uses the same code seems pretty small. Almost the same as throwing this AssertionError here.

I don't think it's a good idea to not support this as there is no particular reason for it.

In general I think we should not be treating sliced storages as a special case unless there is a very strong reason for omitting a feature.

snuba/consumers/consumer_builder.py
lynnagara: Ok you are right. KafkaTopicSpec may not be the right place since it's not storage aware. But how do we prevent this code from being redefined in every endpoint? And ideally, how do we also ensure this function gets automatically called on any attempt to fetch a sliced storage instead of relying on the user to remember to do so?

snuba/consumers/consumer_builder.py
ayirr7: Sure, this makes sense to me. I can have the topic override exist for sliced settings as well.

snuba/consumers/consumer_builder.py
ayirr7: Good point. I will think about how I can make this part of an automatic validation and update.

snuba/utils/streams/configuration_builder.py
nikhars: To add some context, an example of how the slicing convention would work is here https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L413-L537.

To the point of slice=0 behaving same as slice=None, it makes things implicit. I prefer explicit over implicit. So the slicing configuration for different objects which would get sliced would have their own configuration. What slice 0 has in the sliced configs is same as what the original config has. You can see that in the settings file I linked above.

It also keeps thing simpler to reason about. If slicing is enabled on an object(broker config, topic, clusters), look at the sliced version of the config. If not, look at the regular config.

snuba/consumers/consumer_builder.py
nikhars: Is topic override for sliced environments a hard requirement? I know it is good to have. But we are building iteratively and feel like this can be supported at a later point in time.

snuba/utils/streams/configuration_builder.py
lynnagara: If that's how settings is supposed to be defined, this is ok. But this was not clear to me. If sliced settings cannot be put into the KAFKA_BROKER_CONFIG then this needs to be clear from:
- the comments in the settings file
- properly validated in https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py so it will be impossible to provide such settings

Otherwise, I would assume it might work in the other way where slice 0 can go in KAFKA_BROKER_CONFIG and other slices can go in SLICED_KAFKA_BROKER_CONFIG or that passing slice=0 is fine if there's only 1 slice / unsliced.

snuba/consumers/consumer_builder.py
lynnagara: Isn't the effort to support a sliced storage literally just removing the assertion on this one line? Not sure why we would want to not do that now and add it later.

snuba/consumers/consumer_builder.py
lynnagara: Let me ask from a different angle? What would happen if we passed a storage that was not actually sliced in here with a slice ID, and vice versa? Would it fail anyway and at which later point?

snuba/consumers/consumer_builder.py
nikhars: Under what circumstances do we use the topic overrides? From my understanding, it is for overriding topics which are hard-coded in the code based on different environments. With slicing, there is an explicit way of overriding topics per slice which can be seen here https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L469-L496. This keeps the method of overrides consistent across different model objects(kafka topics, kafka brokers, clusters, etc.) which can be sliced. So the precedence order in a sliced environment is always prefer slice configuration if it exists.

Now individual model objects can provide their own mechanism of overrides. In a non-sliced environment that works fine. But if we start allowing overrides for different model objects and the precedence order of the different model objects varies by type of model, it makes things complicated.

lynnagara: What you described about the topic overrides for slicing is not any different for non sliced topics. There also exists an explicit way to override the topics in settings. It's essentially exactly the same - KAFKA_TOPIC_MAP and SLICED_KAFKA_TOPIC_MAP. Slicing and non slicing is exactly the same in this regard, the value just gets picked up from a different variable in settings.

There are scenarios where we want to pass via the CLI. Doing the errors/transactions topic split was a recent example where it was needed. The precedence (CLI always takes precedence!) should be exactly the same as for non sliced topics and it's like this on all our consumers, replacers, etc. I don't think there is anything unique or extra complicated when there is a slice involved.

If you look a few lines below this one you will see that this PR already supports overrides for the replacement and commit log topics via CLI for both sliced and non sliced consumer. So I don't really get why we woudn't also want the same for the raw topic.

nikhars: You bring up a good point. The CLI can override topic names today. The question I would like to answer is should we allow CLI overrides when slice_id is passed? There can be conflicts between the topic-name passed in via CLI and the configuration in SLICED_KAFKA_TOPIC_MAP.

Lets assume you have settings like these
SLICED_KAFKA_TOPIC_MAP -> (orig-topic-name, 1) = new-topic-name-1
The consumer code is run with 
`consumer --slice-id=1 --topic=cli-topic-name ...`
When the consumer tries to resolve the topic name, it tries to lookup using `cli-topic-name` instead of `orig-topic-name`. The lookup would fail. The question is how do we avoid these human errors. The problem of trying to keep 2 different sets of configuration in sync is a manual operation.

A simpler way to avoid this problem might be to disallow topic overrides when slice-id is passed. The current usage of topic overrides can be achieved by modifying the configuration of `SLICED_KAFKA_TOPIC_MAP` itself. I am open to other suggestions as well.

snuba/consumers/consumer_builder.py
```diff
@@ -217,10 +259,14 @@ def __build_consumer(
                 commit_retry_policy=self.__commit_retry_policy,
             )
 
-        return StreamProcessor(consumer, self.raw_topic, strategy_factory, IMMEDIATE)
+        raw_topic_spec = KafkaTopicSpec(SnubaTopic(self.raw_topic.name))
```
lynnagara, line 262: This is not the correct usage of KafkaTopicSpec. You should never be constructing it yourself but just fetching it from the stream loader. ie. `stream_loader.get_default_topic_spec()` then calling `get_physical_topic_name` on that to resolve it.

snuba/consumers/consumer_builder.py
lynnagara: I disagree that just disallowing topic overrides is a simple way of solving the problem. I think that it makes it more complex because what it does is make sliced logic and unsliced logic different in this code path, when this code path should not be concerned with slicing at all and should behave exactly the same way in both scenarios.

About your comment about lookups failing. I believe it's because of a separate bug in how resolving is done in this PR. Left a comment on it here https://github.com/getsentry/snuba/pull/3259#discussion_r1014501873. If this is fixed we can just remove this one line and everything behaves the same in both sliced and unsliced variants.

snuba/consumers/consumer_builder.py
```diff
@@ -217,10 +259,14 @@ def __build_consumer(
...
+        physical_topic = ArroyoTopic(raw_topic_spec.get_physical_topic_name(slice_id))
+
+        return StreamProcessor(consumer, physical_topic, strategy_factory, IMMEDIATE)
```
lynnagara, line 265: I think this is going to break all CLI topic overrides since you are attempting to remap it to something else in here.

snuba/consumers/consumer_builder.py
ayirr7: Yes, it would fail when trying to retrieve the cluster. So, `build_batch_writer` in __build_streaming_strategy_factory calls `get_batch_writer` which calls `get_cluster` for some storage set key and slice id. If the storage and slice id pair are incompatible, this will fail and return an `UndefinedClickhouseCluster` (as you can see in the `get_cluster` function).

snuba/consumers/consumer_builder.py
```diff
@@ -109,29 +129,42 @@ def __init__(
...
-        self.commit_log_topic: Optional[Topic]
+        self.commit_log_topic: Optional[ArroyoTopic]
+
         if kafka_params.commit_log_topic is not None:
-            self.commit_log_topic = Topic(kafka_params.commit_log_topic)
+            kafka_param_topic_spec = KafkaTopicSpec(
+                SnubaTopic(kafka_params.commit_log_topic)
+            )
+            self.commit_log_topic = ArroyoTopic(
+                kafka_param_topic_spec.get_physical_topic_name(slice_id)
```
lynnagara, line 159: You should not be calling get_physical_topic_name here since we were already given the actual topic name that needs no further resolving. We only want it below in the else statement.

snuba/consumers/consumer_builder.py
lynnagara: Please ensure that all the way of mapping of commit log topic, replacement topic and raw topic is being handled the same. We are solving the same thing for all 3.

snuba/consumers/consumer_builder.py
ayirr7: Ah ok. So if I follow correctly, what you mean is that I can support topic override as it is in un-sliced contexts already, and then get the physical topic name in this way?

snuba/consumers/consumer_builder.py
ayirr7: I think I follow Lyn's suggestion, and indeed using the default topic spec can allow us to resolve consistently and accurately to the physical topic name. It should be a small change, so if we can easily retain the original behavior of the ConsumerBuilder, I think it makes sense to do so.

snuba/utils/streams/configuration_builder.py
ayirr7: I see where the confusion is occurring. I can make the comments clearer (although it does say, within the slicing section in settings/init, that each of those configs are for sliced topics/storages/etc only). I'll add some validation as well so that it is impossible to incorrectly configure.

snuba/consumers/consumer_builder.py
lynnagara: Yes. The KafkaTopicSpec is now taking care of all the sliced vs unsliced differences. So the consumer builder doesn't have to care about it anymore - as long as it passes the slice_id along to get_physical_topic_name(). Since that's taken care of in KafkaTopicSpec, this file should hardly change at all now. All you need to do is make sure that anywhere that was previously calling `KafkaTopicSpec.topic_name` now calls `KafkaTopicSpec.get_physical_topic_name(slice_id)` so that function can figure everything out taking the slice or lack of slice into account. There are 3 places that need to be replaced - default/raw topic, commit log topic and replacement topic - all these should be done the same way.

snuba/consumers/consumer_builder.py
lynnagara: I think the issue is the other topic (SnubaTopic) is being imported here but that should never be the case. It should always be accessed via the stream loader and KafkaTopicSpec mechanism to ensure that topic resolving is done correctly.

snuba/consumers/consumer_builder.py
ayirr7: Thanks, this makes sense and is a lot cleaner and more consistent!

snuba/consumers/consumer_builder.py
ayirr7: This makes sense. I no longer have those imports. Shall I change `ArroyoTopic` back to `Topic`?

snuba/utils/streams/configuration_builder.py
lynnagara: That behaviour seems good to me.

I think maybe the settings for KAFKA_BROKER_CONFIG just needs a comment saying that broker config for sliced storages can not go in there.

I think we should error in 2 cases:
- if we see the topic for a sliced storage in in KAFKA_BROKER_CONFIG
- an unsliced topic present in SLICED_KAFKA_BROKER_CONFIG.

Since this is sort of tangential to this particular PR though, feel free to tackle it separately.

snuba/consumers/consumer_builder.py
lynnagara: That seems good to me. It feels a little bit more consistent with what the other files do.

snuba/consumers/consumer_builder.py
```diff
@@ -61,6 +62,19 @@ class MockParameters:
     std_deviation: int
 
 
+def validate_passed_slice(
+    storage_key: StorageKey, slice_id: Optional[int] = None
```
lynnagara, line 66: nit: I think it's better to omit the None default here. It feels safer to force the caller to pass the slice ID. Otherwise we might risk applying the wrong default. There's a few other functions in this diff that do a similar thing too.

snuba/consumers/consumer_builder.py
```diff
@@ -77,6 +91,7 @@ def __init__(
         max_batch_time_ms: int,
         metrics: MetricsBackend,
         parallel_collect: bool,
+        slice_id: Optional[int] = None,
```
lynnagara, line 94: Let's remove the default here too, we want the real value

snuba/datasets/storage.py
```diff
@@ -167,6 +167,7 @@ def __init__(
             writer_options=writer_options,
             write_format=write_format,
         )
+
```
lynnagara, line 170: nit: No need to change this file

tests/datasets/test_table_storage.py
```diff
@@ -0,0 +1,10 @@
+# from snuba.datasets.table_storage import KafkaTopicSpec
+
+# # define some SLICED_KAFKA_TOPIC_MAP
+
+# # patch the settings with above map
+# def test_get_physical_topic_name() -> None:
```
lynnagara, line 6: I think this test is a good idea

snuba/utils/streams/configuration_builder.py
ayirr7: That sounds good. I'll make this in a separate PR.

snuba/consumers/consumer_builder.py
ayirr7: Update: we've decided to move the validation step to `partitioning.py`, but this and underlying discussions (slicing by storage vs storage set) will be revisited.

snuba/consumers/consumer_builder.py
```diff
@@ -93,12 +95,15 @@ def __init__(
             .topic
         )
 
+        validate_passed_slice(storage_key, slice_id)
```
lynnagara, line 98: I think we should also move this call to the centralized code so it happens automatically on an attempt to load the storage and does not require the user to explicitly call it.

ayirr7: Where are you referring to by "centralized code"?

snuba/consumers/consumer_builder.py
ayirr7: I'm not sure if we would want to remove the default. A handful of different sliced functions (like `get_cluster`) rely on this behavior (only pass in a slice id if the storage set is sliced). This is also analogous to the way our settings work: only define in `SLICED_` settings if this is a sliced cluster/storage set/etc. If something is unsliced, define it in the regular settings.

lynnagara: The consumer builder is only called in one place and we always pass a value from there. So why introduce the chance of calling it wrong / applying an incorrect default?

snuba/consumers/consumer_builder.py
ayirr7: The code which you linked is currently only used for the QueryPlanBuilder, and doesn't actually directly load the storages. The Consumer Builder only uses `get_writable_storage` to get and load the storage from config, but we decided not to make that path slice-aware.

snuba/consumers/consumer_builder.py
ayirr7: ConsumerBuilder calls `get_cluster()` and the `slice_id` that is passed into the ConsumerBuilder in order to initialize it is the same as that which goes into the `get_cluster()` call. Are you suggesting that we change the behavior of `get_cluster` right now?

lynnagara: I'm saying there was a reason why `get_cluster()` had to be that way with the default (which is that it's used in many non slice aware places right now). But I don't think that reason exists here.

snuba/consumers/consumer_builder.py
lynnagara: Personally I think that there may not be a very obvious place. I think the reason for this is might be twofold:
1. storage objects are unsliced
2. slicing is not really even done at the storage level, it's done at the storage set level

So I think this is fine if you want to leave this here now. But I'm not sure if this setting that we are validating against is actually the right one given that slices are generally defined against storage sets and not the storages themselves.

snuba/consumers/consumer_builder.py
ayirr7: Oh, ok I think I'm starting to follow. Are you suggesting we keep it so that we explicitly pass in `None` for slice_id into ConsumerBuilder when necessary? And therefore, also into the Consumer CLI (via `--slice-id=None`)?

lynnagara: The consumer CLI doesn't need to change anymore. Not passing any `--slice-id` ***is*** passing none. I think slice_id just should be passed in most of the places in this PR. There will be no real impact since we are already always passing the slice ID into most of the places anyway. But I think it helps prevent against incorrect usage in the future.

ayirr7: Right, ok this makes sense.

nikhars (approved):
LGTM

snuba/consumers/consumer_builder.py
ayirr7: I think what I'll do is leave this here for now, but make a note to revisit the location for this when we do the rename from `SLICED_STORAGES` to `SLICED_STORAGE_SETS`?

snuba/utils/streams/configuration_builder.py
ayirr7: If I understand what you are suggesting, I can move the logic inside `_get_default_topic_configuration` into `get_default_kafka_configuration` itself. Only thing is this would remove our ability to separately test the mapping functionality in `_get_default_topic_configuration`.

snuba/consumers/consumer_builder.py
lynnagara: Bumping this. Can you please revert the change?

lynnagara (approved):

PR #3260, opened 2022-10-14
fix(gen-metrics): distributed table name wrong in config

onewland:
Stuff broke when we tried switching to the [distributed] query node which correctly does not have local tables

https://sentry.io/organizations/sentry/issues/3671781868/?project=300688&query=is%3Aunresolved
https://sentry.io/organizations/sentry/issues/3671799059/?project=300688&query=is%3Aunresolved

🤦


mwarkentin (approved):
LGTM

PR #3261, opened 2022-10-14
feat(MDC): Add Translation Mapper Support for Transaction Dataset.

enochtangg:
Before migrating the transaction dataset to config, there are a few translation mappers which are currently not supported in config (namely the ones which inherit from `ColumnMapper`). 

Adding support for `column_mappers` is not as simple as the `function_mappers` and `subscriptable_mappers`. This is because there is a `ColumnToFunction` mapper which contains an attribute called `to_function_params: Tuple[Expression, ...]`. As a result, we also need to register `Expressions` classes to be looked up by name. Additionally, `Expression` include `FunctionCall` which is a class that can have another nested `Expression`. This PR is responsible for registering the necessary classes and adding support to `entity_builder.py` to be able to define all column mappers.

More specifically:
* Register column mappers and expressions to `metaclass=RegisteredClassed`
* Added column mapping support and recursive expression parsing to entity_builder
* Added tests for processing column mappers in config

### DIFF
--- a/snuba/datasets/configuration/entity_builder.py
+++ b/snuba/datasets/configuration/entity_builder.py
@@ -17,7 +17,6 @@ from snuba.datasets.entities.entity_key import register_entity_key
 from snuba.datasets.pluggable_entity import PluggableEntity
 from snuba.datasets.storages.factory import get_storage, get_writable_storage
 from snuba.datasets.storages.storage_key import StorageKey
-from snuba.query.expressions import Expression
 from snuba.query.processors.logical import LogicalQueryProcessor
 from snuba.query.validation.validators import QueryValidator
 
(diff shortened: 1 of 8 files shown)


snuba/datasets/configuration/entity_builder.py
```diff
@@ -40,9 +42,44 @@ def _build_entity_query_processors(
...
+            Expression.get_from_name(expression_name)(**expression_config["args"])
+        )
+    return expressions
+
+
+def _pre_process_column_mapper_config(sub_config: dict[str, Any]) -> dict[str, Any]:
+    # TODO: Will need to add support for ColumnToCurriedFunction when migrating sessions dataset to config
+    if sub_config["mapper"] == "ColumnToFunction":
+        sub_config["args"]["to_function_params"] = tuple(
+            _parse_expressions(sub_config["args"]["to_function_params"])
```
evanh, line 65: There is actually already a parser that will convert a string into a correct expression. https://github.com/getsentry/snuba/blob/master/snuba/query/parser/expressions.py#L160

It's used in the CustomFunction processors (amongst other places).

Maybe a simpler solution here would be to represent the function as a string in the config, and then use the existing parser to convert it into an `Expression`. For example, `expression: "coalesce(IPv4NumToString(ip_address_v4), IPv6NumToString(ip_address_v6))"` .

Then `Expression` doesn't need to be a registered class and you don't need to reimplement an expression parser. This would also carry over to `ColumnToCurriedFunction` when we get there.

tests/datasets/configuration/test_entity_loader.py
```diff
@@ -54,3 +62,47 @@ def test_bad_configuration_broken_validator() -> None:
...
+
+    # Check that ColumnToFunction mapper was successfully loaded from config
+    column_to_function = get_object_in_list_by_class(column_mappers, ColumnToFunction)
+    assert isinstance(column_to_function, ColumnToFunction)
+
+    # Check that nested expressions were loaded correctly in ColumnToFunction
+    assert len(column_to_function.to_function_params) == 2
+    function_call = get_object_in_list_by_class(
+        column_to_function.to_function_params, FunctionCall
+    )
```
onewland, line 88: seems like you forgot to delete this

snuba/datasets/configuration/entity_builder.py
volokluev: ☝️ yes let's do that. `Expression` classes are for the computer, not for humans. Humans should be able to type in actual expressions

volokluev (approved):
well done. I would update the PR description before merging though a what happened in this PR does not match the description anymore

tests/datasets/configuration/entity_with_ip_column_mappers.yaml
```diff
@@ -0,0 +1,71 @@
...
+  -
+    processor: TagsTypeTransformer
+  -
+    processor: ReferrerRateLimiterProcessor
+translation_mappers:
+  columns:
+    -
+      mapper: ColumnToIPAddress
+      args:
+        from_table_name:
```
evanh, line 26: Is it necessary to have these empty values? Could we maybe set a reasonable default so they don't need to be defined in the YAML?

evanh (approved):
Looks good to me, one small question.

tests/datasets/configuration/entity_with_ip_column_mappers.yaml
enochtangg: Right now, these empty values are also required by other mappers (subscriptables, functions, curried_functions). I agree, we should set some reasonable default either as a config override or within the class constructor itself. Since this PR is blocking some other work, I will create a ticket for this issue.

PR #3262, opened 2022-10-14
fix(TDC) Have configuration tests automatically reset factories

evanh:
When migrating the transactions storage, there was a test failure that running
`reset_dataset_factory` fixed. Instead of manually putting that into every test,
instead have a base class that does it automatically.


tests/datasets/configuration/test_storage_loader.py
```diff
@@ -143,13 +140,13 @@ def test_processor_with_constructor():
...
-        assert qp._MappingOptimizer__hash_map_name == "hashmap"
-        assert qp._MappingOptimizer__killswitch == "kill"
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            filename = os.path.join(tmpdirname, "file.yaml")
+            with open(filename, "w") as f:
+                f.write(yml_text)
+            storage = build_storage(filename)
+            assert len(storage.get_query_processors()) == 1
+            qp = storage.get_query_processors()[0]
+            assert getattr(qp, "_MappingOptimizer__column_name") == "a"
```
evanh, line 150: Using `getattr` was to fix a typing error.

volokluev:
what was the failing test and how did it fail?

volokluev:
I'm not opposed to this change, but I want to better understand what a `ConfigurationTest` is and when we would want to use it

evanh:
I didn't save the original error, but I do know that when I added `reset_dataset_factory` to the test, it started passing.

That made me think, we should probably do that for every test just as a best practice. That way we can be assured that one set of configuration tests aren't messing up the global variable for a different test. That's why I created the base class.

rahul-kumar-saini (approved):

PR #3263, opened 2022-10-14
feat: Drop filtering by offset on errors dataset

lynnagara:
None of the other datasets apply this condition (they cannot since they are non semantically partitioned).

We sometimes flip large projects into to random partitioning mode to avoid backlogs and incidents. Doing so causes the alert to be completely wrong unless we remove this condition as the offset across different partitions can no longer be compared.


MeredithAnya:
I'm a little confused, so if we take this out does this mean that the alert could be wrong in semantic partitioning mode then? are we saying this check isn't useful at all anymore then?

lynnagara:
Technically it means that a backdated event that falls in the evaluation window that happens to have a later offset will be counted where today it will be excluded. But that is how alerts on all of our other datasets work, and I think we should just align on it rather than preserve this for errors.

fpacifici (approved):

PR #3266, opened 2022-10-17
test(migrations): Add example migrations for testing

MeredithAnya:
An attempt to create some dummy migrations that could be used by https://github.com/getsentry/snuba/pull/3258/ to test out the migration policies. 

My thought was to create a group loader that could override the `get_group_loader` in tests for the migration policies that need to load the migration.

### DIFF
--- a/tests/migrations/examples/0001_sql_migration.py
+++ b/tests/migrations/examples/0001_sql_migration.py
@@ -1,0 +1,19 @@
+from typing import Sequence
+
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return []
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return []
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
+
+    def backwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return []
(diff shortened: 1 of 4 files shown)


tests/migrations/examples/0001_sql_migration.py
```diff
@@ -0,0 +1,19 @@
+from typing import Sequence
+
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_local(self) -> Sequence[operations.SqlOperation]:
+        return []
```
MeredithAnya, line 10: This will work for any tests that don't need to check the operations. The operations depend on specifying a `StorageSetKey` so I wasn't sure how we envision a migrations module that is independent of snuba  `StorageSetKey`s

lynnagara: We should have at least one operation in here to make this migration more realistic. Eventually storage set keys will probably need to get registered in the migration framework (and not just shared from Snuba), but for now I think it's ok to hardcode an operation using some existing Snuba storage set keys here. We can swap it out later.

onewland:
Is this still something we want to try to merge? (Just found this looking for open non-dependabot PRs that need SNS approvals to merge)

MeredithAnya:
> Is this still something we want to try to merge? (Just found this looking for open non-dependabot PRs that need SNS approvals to merge)

ehh I don't think it's super important, I'll close it out. I think it can be brought back if need be in the future

PR #3268, opened 2022-10-17
ref(MDC): Topics Enum to Class (with fix)

enochtangg:
Retrying this PR: https://github.com/getsentry/snuba/pull/3236

This PR also contains the fix which caused consumers to go into CrashLoopBackOff when deployed. This PR contains a change to remove the list of all topics in the settings `validation.py`. Instead of doing validation there, it was proposed that we do validation after storages were loaded. This allowed us to remove `validation.py` and validate against the Topics class (after all topics were loaded from configuration). However, it was overlooked that there were 4 topics in the `validation.py` that were not present in the Topics class. 

Namely:

```
snuba-sessions-commit-log
scheduled-subscriptions-sessions
sessions-subscription-results
snuba-replay-events
```

This issue went unnoticed in unit testing because we override session topics in `prod_settings`. This PR also includes the fix which adds the 4 topics into `topics.py`.

### DIFF
--- a/snuba/datasets/storages/generic_metrics.py
+++ b/snuba/datasets/storages/generic_metrics.py
@@ -48,9 +48,7 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
     """
     return ProduceInvalidMessagePolicy(
         KafkaProducer(
-            build_kafka_producer_configuration(
-                Topic.DEAD_LETTER_GSNUBA_DEAD_LETTER_GENERIC_METRICSENERIC_METRICS
-            )
+            build_kafka_producer_configuration(Topic.SNUBA_DEAD_LETTER_GENERIC_METRICS)
         ),
         KafkaTopic(Topic.SNUBA_DEAD_LETTER_GENERIC_METRICS.value),
     )
--- a/snuba/datasets/storages/generic_metrics.py
+++ b/snuba/datasets/storages/generic_metrics.py
@@ -48,9 +48,11 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
     """
     return ProduceInvalidMessagePolicy(
         KafkaProducer(
-            build_kafka_producer_configuration(Topic.DEAD_LETTER_GENERIC_METRICS)
+            build_kafka_producer_configuration(
+                Topic.DEAD_LETTER_GSNUBA_DEAD_LETTER_GENERIC_METRICSENERIC_METRICS
+            )
         ),
-        KafkaTopic(Topic.DEAD_LETTER_GENERIC_METRICS.value),
+        KafkaTopic(Topic.SNUBA_DEAD_LETTER_GENERIC_METRICS.value),
     )
 
 
@@ -137,12 +139,12 @@ sets_bucket_storage = WritableTableStorage(
     query_processors=[],
     stream_loader=build_kafka_stream_loader_from_settings(
         processor=GenericSetsMetricsProcessor(),
-        default_topic=Topic.GENERIC_METRICS,
+        default_topic=Topic.SNUBA_GENERIC_METRICS,
         dead_letter_queue_policy_creator=produce_policy_creator,
-        commit_log_topic=Topic.GENERIC_METRICS_SETS_COMMIT_LOG,
-        subscription_scheduled_topic=Topic.SUBSCRIPTION_SCHEDULED_GENERIC_METRICS_SETS,
+        commit_log_topic=Topic.SNUBA_GENERIC_METRICS_SETS_COMMIT_LOG,
+        subscription_scheduled_topic=Topic.SCHEDULED_SUBSCRIPTIONS_GENERIC_METRICS_SETS,
         subscription_scheduler_mode=SchedulingWatermarkMode.GLOBAL,
-        subscription_result_topic=Topic.SUBSCRIPTION_RESULTS_GENERIC_METRICS_SETS,
+        subscription_result_topic=Topic.GENERIC_METRICS_SETS_SUBSCRIPTION_RESULTS,
         pre_filter=KafkaHeaderSelectFilter("metric_type", InputType.SET.value),
     ),
 )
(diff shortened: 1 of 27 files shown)


snuba/utils/streams/topics.py
```diff
@@ -1,62 +1,90 @@
...
         Topic.EVENTS: {"message.timestamp.type": "LogAppendTime"},
         Topic.TRANSACTIONS: {"message.timestamp.type": "LogAppendTime"},
-        Topic.METRICS: {"message.timestamp.type": "LogAppendTime"},
-        Topic.PROFILES: {"message.timestamp.type": "LogAppendTime"},
-        Topic.REPLAYEVENTS: {"message.timestamp.type": "LogAppendTime"},
-        Topic.GENERIC_METRICS: {"message.timestamp.type": "LogAppendTime"},
+        Topic.SNUBA_METRICS: {"message.timestamp.type": "LogAppendTime"},
+        Topic.PROCESSED_PROFILES: {"message.timestamp.type": "LogAppendTime"},
+        Topic.INGEST_REPLAY_EVENTS: {"message.timestamp.type": "LogAppendTime"},
+        Topic.SNUBA_GENERIC_METRICS: {"message.timestamp.type": "LogAppendTime"},
```
fpacifici, line 88: Dp you plan to move this somehow to config ? 
This setting is required for any storage we want to run alerts on. So that would mean we would still have to make changes to Snuba code to create a new dataset that has subscriptions enabled. 

I think you have two options. 
- move these creation settings to config.
- make this setting applied automatically when creating a topic that has subscriptions associated.

Though I think this opens another problem: should the dataset definition carry enough information to create all topics the dataset needs or should the config be split between the topic name (defined in the dataset) and the environment specific settings to create a topic (like partitions number and retention)?
I think the environment parameters should not be together with the dataset config. The dataset is the schema, it is part of the product.

@lynnagara  what do you think ?

evanh: Can you clarify you suggestions?

1) Have the Kafka settings in the configuration.
```
topic:
  - name: snuba-metrics
  - settings:
    - message.timestamp.type: LogAppendTime
```

2) Make the setting applied automatically:
```
topic:
  - name: snuba-metrics
  - subscriptions: yes
```
and then
```
if config["subscriptions"] == "yes":
    kafka_settings["message.timestamp.type"] = "LogAppendTime"
```

fpacifici: Context:
The way snuba manages Kafka topic relies a lot on them being in an enum. This was all meant and designed long before we decided to move dataset in config. So at this point they suffer of all the same limitations and constraints `StorageSetKey` have plus others. 
Like `StorageSetKey`, topics are a logical view on an environment specific resource. As such, part of them are part of the application (the logical name) and part of that needs to be overridden per environment (name and assignment to broker).
Contrarily to `StorageSetKey`, Topics are a step ahead as they carry the configuration that should be applied when the topic is created. We do not have anything that automatically configure a Clickhouse cluster when it is created. 
The topic creation configuration contains some fields that are actually environment specific (like partitions and retention) and other that the application depends on (like `LogAppendTime`. Subscriptions cannot work without). 

So it seems that treating all creation settings as environment settings would be wrong and treating all of them as application specific would be wrong.

To solve this problem without requiring anything to be hardcoded when a new dataset is added, I think we should allow the following:
- a basic sets of settings that are applied based on the fact that the storage has subscriptions. So something like your option 2. The dataset config says whether the storage has subscriptions and the topic is created with that flag on. This does not solve it for existing topics when we activate subscriptions after the fact but I believe this is not solve today either. 
- an additional set of settings to override parameters for the specific environment. This is not needed in this PR though as it is not supported today either. 

Or, in alternative, just always create topics with `LogAppendTime` @lynnagara  what do you think about always creating topics that way ?

lynnagara: I think it's very important for topics to be "registered" (in some form or another) in a single place. I think Evan's option 1 is the best of the proposed ideas here and by far the closest to how it already works today.

Why I think the other options should not be pursued:
- Firstly the `subscriptions: yes` option. Topic configuration should not know about high level Snuba concepts. One day this may actually move outside Snuba as this is how different services talk to each other. The only reason we put it inside Snuba is because nothing like that exists today.
- About putting `LogAppendTime` everywhere. Well, this solves our immediate problem but closes the door to all other custom configuration. This is inflexible and seems like it would be a pain on the day we need it.

lynnagara: Now, if we want to do things like Validate LogAppendTime is there when a topic is chosen for subscriptions, we can still do that inside the storage building logic. But it should not be the job of the topics definition to validate Snuba specific concepts.

enochtangg: It sounds like there are two different perspectives on how to solve this problem. Whether or not we decide on a solution here, would it make sense for this to be moved to some documentation and shared for visibility? For the scope of this PR, the **function** of the Topic class is the same as the Enum. There is another [PR](https://github.com/getsentry/snuba/pull/3259) which depends on some of the `register_topic()` functionality implemented here.

Since as mentioned, logical Topics still carry configuration that is applied in code. What I'm proposing moving forward:
1. I'll write up documentation to capture this discussion and share it for visibility
2. We can more formally decide on which option to pursue
3. This will be a separate task that gets carried out during tech debt week

Does that sound sensible? Or does it make more sense for this PR to contain these changes?  @evanh @lynnagara @fpacifici

evanh: Yes, let's do that @enochtangg . Thank you.

I think there are two underlying questions that this is touching on: How much of the `Topic` settings is logical and belongs in configuration, and how much is environment specific? Once we figure that out, where do the environment specific settings go?

snuba/datasets/storages/generic_metrics.py
```diff
@@ -48,9 +48,11 @@ def produce_policy_creator() -> DeadLetterQueuePolicy:
     """
     return ProduceInvalidMessagePolicy(
         KafkaProducer(
-            build_kafka_producer_configuration(Topic.DEAD_LETTER_GENERIC_METRICS)
+            build_kafka_producer_configuration(
+                Topic.DEAD_LETTER_GSNUBA_DEAD_LETTER_GENERIC_METRICSENERIC_METRICS
```
ayirr7, line 52: Apologies if this has been addressed already, but is the Topic here on line 52 a valid topic? While working on testing sliced consumers, this threw an Attribute error as I think it is not in `_HARDCODED_TOPICS` or `_REGISTERED_TOPICS`.

enochtangg: Good catch, that's definitely a copy paste mistake. Updated!

snuba/utils/streams/topics.py
```diff
@@ -1,62 +1,90 @@
...
+
+    def __eq__(self, other: Any) -> bool:
+        return isinstance(other, Topic) and other.value == self.value
+
+    def __repr__(self) -> str:
+        return f"Topic.{self.value.upper()}"
+
+
+def register_topic(key: str) -> Topic:
+    _REGISTERED_TOPICS[key.upper().replace("-", "_")] = key.lower()
```
dbanda, line 77: nit: assuming this input is coming from the cli,  you may also want to call `.strip()` to remove whitespace

PR #3270, opened 2022-10-18
feat(observability): Add measurements to readthrough cache

volokluev:
We are having issues with occasionally spiking query times in snuba. We have evidence to believe that this is because of the readthrough cache but we don't know why exactly because its inner workings are not visible at runtime. This PR introduces metrics about the cache's operation such that they can be observed when another spike happens

### DIFF
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -233,7 +233,7 @@ class RedisCache(Cache[TValue]):
                 effective_timeout,
             )
             queue_len = self.__client.llen(build_notify_queue_key(task_ident))
-            metrics.gauge("notify_queye_len", queue_len, tags=metric_tags)
+            metrics.gauge("notify_queue_len", queue_len, tags=metric_tags)
             notification_received = (
                 self.__client.blpop(
                     build_notify_queue_key(task_ident), effective_timeout
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -147,7 +147,6 @@ class RedisCache(Cache[TValue]):
         if result[0] == RESULT_VALUE:
             # If we got a cache hit, this is easy -- we just return it.
             logger.debug("Immediately returning result from cache hit.")
-            metrics.increment("cache_hit", tags=metric_tags)
             return self.__codec.decode(result[1])
         elif result[0] == RESULT_EXECUTE:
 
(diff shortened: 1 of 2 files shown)


snuba/state/cache/redis/backend.py
```diff
@@ -136,13 +139,15 @@ def build_notify_queue_key(task_ident: str) -> str:
...
             timer.mark("cache_get")
+        metric_tags = timer.tags if timer is not None else {}
 
         # This updates the stats object and querylog
         record_cache_hit_type(result[0])
 
         if result[0] == RESULT_VALUE:
             # If we got a cache hit, this is easy -- we just return it.
             logger.debug("Immediately returning result from cache hit.")
+            metrics.increment("cache_hit", tags=metric_tags)
```
onewland, line 150: is this going to cause an issue with sending too many messages to Datadog e.g. do we need some kind of sampling?

onewland (approved):
fine with this, but if we're emitting a metric on the happy path and that includes subscriptions + API queries, we should be confident that it won't cause performance issues

volokluev:
yeah maybe I'll just remove the happy path

snuba/state/cache/redis/backend.py
evanh: We already track cache hits/misses etc., through the timer. Is there a reason this needs to be separate?

snuba/state/cache/redis/backend.py
```diff
@@ -224,6 +232,8 @@ def build_notify_queue_key(task_ident: str) -> str:
                 task_ident,
                 effective_timeout,
             )
+            queue_len = self.__client.llen(build_notify_queue_key(task_ident))
+            metrics.gauge("notify_queue_len", queue_len, tags=metric_tags)
```
evanh, line 236: Are you sure you want a gauge here? Every hit to this metric will overwrite the previous value. Unless the `metrics_tags` provide enough granularity for what you want to track.

volokluev: hmm I guess what I wanted is a distribution of the queue lengths. Is that not possible with datadog?

snuba/state/cache/redis/backend.py
volokluev: I'll delete this one

snuba/state/cache/redis/backend.py
volokluev: Ah I need to use the histogram type

https://docs.datadoghq.com/metrics/types/?tab=histogram#metric-types

Our metrics abstraction doesn't support this yet so maybe I'll take it out for now and put it back in later

evanh (approved):

PR #3271, opened 2022-10-18
feat(TDC): Transactions Entity Config

rahul-kumar-saini:
### Overview
- Transactions entity config'd
- Extended `test_entity_loader.py` with Transactions entity
- Modified json schema as needed

### DIFF
--- a/snuba/datasets/configuration/utils.py
+++ b/snuba/datasets/configuration/utils.py
@@ -22,7 +22,7 @@ from snuba.clickhouse.columns import (
 from snuba.datasets.plans.splitters import QuerySplitStrategy
 from snuba.query.processors.condition_checkers import ConditionChecker
 from snuba.query.processors.physical import ClickhouseQueryProcessor
-from snuba.utils.schemas import UUID, AggregateFunction
+from snuba.utils.schemas import UUID, AggregateFunction, IPv4, IPv6
 from snuba.utils.streams.configuration_builder import build_kafka_producer_configuration
 from snuba.utils.streams.topics import Topic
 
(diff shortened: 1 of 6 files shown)


snuba/datasets/configuration/utils.py
```diff
@@ -95,45 +95,48 @@ def get_mandatory_condition_checkers(
...
+NUMBER_COLUMN_TYPES = {"UInt": UInt, "Float": Float}
+
+SIMPLE_COLUMN_TYPES = {
+    **NUMBER_COLUMN_TYPES,
+    "String": String,
+    "DateTime": DateTime,
+    "UUID": UUID,
+    "IPv4": IPv4,
+    "IPv6": IPv6,
+}
```
volokluev, line 107: what if we just made Column(s) a RegisteredClass?

evanh: That's an interesting idea, but I think maybe should be a separate task/PR.

tests/datasets/configuration/test_entity_loader.py
```diff
@@ -12,44 +12,65 @@
 )
 from snuba.datasets.configuration.entity_builder import build_entity_from_config
 from snuba.datasets.entities.entity_key import EntityKey
-from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.entity import Entity
+from snuba.datasets.factory import reset_dataset_factory
 from snuba.datasets.pluggable_entity import PluggableEntity
 from snuba.query.expressions import Column, FunctionCall, Literal
 
+reset_dataset_factory()
```
evanh, line 20: This will conflict with the testing changes I rolled out. If you move your tests inside the `ConfigurationTest` class you shouldn't need this call.

evanh (commented):
This is looking pretty good to me.

Sorry for the merge conflicts on the tests.

snuba/datasets/configuration/utils.py
volokluev: for sure

PR #3273, opened 2022-10-19
feat(rate_limit): Shard sets of table rate limiter [sns-1576]

untitaker:
Redis sets are not sharded across multiple nodes. The table rate limiter
manages a set that grows proportional to the amount of concurrent
queries allowed + whatever is configured for history purposes.

It's therefore possible for a clickhouse cluster of sufficient size to
allow for much more per-second and concurrent queries than what a single
redis node can handle.

In order to bypass this problem, we shard the redis set by query ID.

We did something similar for the cardinality limiter in Sentry
("sentry.ratelimits.cardinality"), where there's the concept of
"num_shards" and "num_physical_shards".


snuba/state/rate_limit.py
```diff
@@ -119,24 +130,43 @@ def rate_limit(
...
     if bypass_rate_limit == 1:
         yield None
         return
 
+    now = time.time()
+
+    query_id_uuid = uuid.uuid4()
+    query_id = str(query_id_uuid)
+
+    bucket_shard = int(query_id_uuid) % rate_limit_shard_factor
```
evanh, line 158: How does this work when we increase the factor? Won't we allocate queries to the wrong shards?

untitaker: @evanh I don't think so, because there's no consistency required across calls to this function. query_id itself is rolled at the start of this function, and as long as we keep using the same shard for the lifetime of the context manager, we should be good.

If we _decrease_ the number of shards, it may mean that we temporarily undercount rate limits as we stop zcounting sets that still have relevant data in them. This is a temporary effect that should fix itself within a couple minutes.

If we increase number øf shards, I think there should not be a problem because it just means we will read additional redis keys, but keep readinig the old ones.

I suggest we'll set this option once (or tweak it as little as possible) and just accept the dataloss.

If that's not acceptable we can also have two settings: how many shards to write, and how many shards to read. This would allow us to decrease shard count without dataloss.

evanh (approved):

PR #3274, opened 2022-10-19
feat(TDC): Transactions storage as configuration

evanh:
Duplicate the transactions writable storage in configuration.

This was tested locally by also removing the Python storage completely, to
ensure backwards compatibility.

Changes made to support the storage:
- Add UUID, IPv4 and IPv6 type
- Create a function `registered_class_schema` to reduce duplicated schemas
- Add `writer_options` to storage schema
- Remove `importlib` calls in a test that are not needed and were breaking the configuration tests

### DIFF
--- a/snuba/datasets/configuration/json_schema.py
+++ b/snuba/datasets/configuration/json_schema.py
@@ -204,10 +204,10 @@ SCHEMA_SCHEMA = {
             "type": "string",
             "description": "The distributed table name in distributed ClickHouse",
         },
-        "part_format": {
+        "partition_format": {
             "type": "array",
             "items": {"type": "string"},
-            "description": "The format of the parts in Clickhouse. Used in the cleanup job.",
+            "description": "The format of the partitions in Clickhouse. Used in the cleanup job.",
         },
     },
     "additionalProperties": False,
--- a/snuba/datasets/configuration/json_schema.py
+++ b/snuba/datasets/configuration/json_schema.py
@@ -331,6 +331,7 @@ ENTITY_TRANSLATION_MAPPERS = {
         "functions": ENTITY_TRANSLATION_MAPPER_SUB_LIST,
         "curried_functions": ENTITY_TRANSLATION_MAPPER_SUB_LIST,
         "subscriptables": ENTITY_TRANSLATION_MAPPER_SUB_LIST,
+        "columns": ENTITY_TRANSLATION_MAPPER_SUB_LIST,
     },
     "additionalProperties": False,
 }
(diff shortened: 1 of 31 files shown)


snuba/datasets/configuration/json_schema.py
```diff
@@ -224,24 +224,52 @@ def make_column_schema(
...
 }
 
-STORAGE_QUERY_PROCESSOR = {
-    "type": "object",
-    "properties": {
-        "processor": {
-            "type": "string",
-            "description": "Name of ClickhouseQueryProcessor class config key. Responsible for the transformation applied to a query.",
+
+def registered_class_schema(
```
volokluev, line 228: nice

volokluev (approved):
Looks good to me. Rahul had a test in his to verify that the config object was the same as the code one. Is that possible here?

snuba/datasets/configuration/json_schema.py
```diff
@@ -204,6 +204,11 @@ def make_column_schema(
             "type": "string",
             "description": "The distributed table name in distributed ClickHouse",
         },
+        "part_format": {
```
nikhars, line 207: This is actually partition format, not part format. I know the code is wrong today in referring to these as part. But it should be partitions. Since you are creating new dataset configuration, may I suggest we fix the name now?
```suggestion
        "partition_format": {
```

snuba/datasets/configuration/json_schema.py
```diff
@@ -204,6 +204,11 @@ def make_column_schema(
...
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "The format of the parts in Clickhouse. Used in the cleanup job.",
```
nikhars, line 210:
```suggestion
            "description": "The format of the partitions in Clickhouse. Used in the cleanup job.",
```

snuba/datasets/configuration/transactions/storages/transactions.yaml
```diff
@@ -0,0 +1,216 @@
...
+      { name: timestamp, type: DateTime, args: { schema_modifiers: [readonly] } },
+      { name: group_ids, type: Array, args: { type: UInt, arg: 64 } },
+      { name: app_start_type, type: String },
+    ]
+  local_table_name: transactions_local
+  dist_table_name: transactions_dist
+  part_format: [retention_days, date]
+
+query_processors:
+  -
```
nikhars, line 121: This might be a stupid question but is there a specific reason why the processors are always defined from the next line of hyphen? It kind of looks unintuitive when reading the file

evanh: Not a stupid question. I think we did that initially to try and line up the `processor` and `args` more but I'm also not a fan. I changed it.

PR #3276, opened 2022-10-20
Add migration policy middleware SNS-1776

dbanda:
Makes the migration policy checking middleware in snuba admin. The middleware gets the migration group policy from the settings map then runs `can_run` / `can_reverese` to if the migration key is allowed to run.

### DIFF
--- a/snuba/admin/migrations_policies.py
+++ b/snuba/admin/migrations_policies.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+
+from functools import wraps
+from typing import Any, Callable, Dict
+
+from flask import Response, jsonify, make_response
+from werkzeug.routing import BaseConverter
+
+from snuba import settings
+from snuba.migrations.policies import (
+    AllMigrationsPolicy,
+    MigrationPolicy,
+    NoMigrationsPolicy,
+    NonBlockingMigrationsPolicy,
+)
+from snuba.migrations.runner import MigrationKey
+
+policy_map = {
+    "AllMigrationsPolicy": AllMigrationsPolicy(),
+    "NoMigrationsPolicy": NoMigrationsPolicy(),
+    "NonBlockingMigrationsPolicy": NonBlockingMigrationsPolicy(),
+}
+
+
+class MigrationActionConverter(BaseConverter):
+    regex = r"(?:run|reverse)"
+
+
+def check_migration_perms(f: Callable[..., Response]) -> Callable[..., Response]:
+    @wraps(f)
+    def check_group_perms(*args: Any, **kwargs: Any) -> Response:
+        ADMIN_ALLOWED_MIGRATION_GROUPS: Dict[str, MigrationPolicy] = {
+            k: policy_map[v] for k, v in settings.ADMIN_ALLOWED_MIGRATION_GROUPS.items()
+        }
+        group = kwargs["group"]
+        if group not in ADMIN_ALLOWED_MIGRATION_GROUPS:
+
+            return make_response(jsonify({"error": "Group not allowed"}), 400)
+
+        if "action" in kwargs:
+            action = kwargs["action"]
+            migration_id = kwargs["migration_id"]
+            migration_key = MigrationKey(group, migration_id)
+            policy = ADMIN_ALLOWED_MIGRATION_GROUPS[group]
+            if action == "run":
+                if not policy.can_run(migration_key):
+                    return make_response(
+                        jsonify({"error": "Group not allowed run policy"}), 400
+                    )
+            elif action == "reverse":
+                if not policy.can_reverse(migration_key):
+                    return make_response(
+                        jsonify({"error": "Group not allowed reverse policy"}), 400
+                    )
+        return f(*args, **kwargs)
+
+    return check_group_perms
--- a/snuba/admin/migrations_policies.py
+++ b/snuba/admin/migrations_policies.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from functools import wraps
 from typing import Any, Callable, Dict
 
-from flask import Response, jsonify, make_response
+from flask import Response, jsonify, make_response, request
 from werkzeug.routing import BaseConverter
 
 from snuba import settings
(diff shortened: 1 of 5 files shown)


snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,57 @@
...
+
+class MigrationActionConverter(BaseConverter):
+    regex = r"(?:run|reverse)"
+
+
+def check_migration_perms(f: Callable[..., Response]) -> Callable[..., Response]:
+    @wraps(f)
+    def check_group_perms(*args: Any, **kwargs: Any) -> Response:
+        ADMIN_ALLOWED_MIGRATION_GROUPS: Dict[str, MigrationPolicy] = {
+            k: policy_map[v] for k, v in settings.ADMIN_ALLOWED_MIGRATION_GROUPS.items()
```
MeredithAnya, line 33: should we handle the `KeyError` here if someone sets a policy in settings that's not in the `policy_map`?

snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,57 @@
...
+    "NoMigrationsPolicy": NoMigrationsPolicy(),
+    "NonBlockingMigrationsPolicy": NonBlockingMigrationsPolicy(),
+}
+
+
+class MigrationActionConverter(BaseConverter):
+    regex = r"(?:run|reverse)"
+
+
+def check_migration_perms(f: Callable[..., Response]) -> Callable[..., Response]:
```
MeredithAnya, line 29: maybe we can add a docstring here ? maybe just explaining that this is to be used for all migrations endpoints and checks both group access, but then also if a migration is provided, whether or not there is access to run/reverse

snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,57 @@
...
+    def check_group_perms(*args: Any, **kwargs: Any) -> Response:
+        ADMIN_ALLOWED_MIGRATION_GROUPS: Dict[str, MigrationPolicy] = {
+            k: policy_map[v] for k, v in settings.ADMIN_ALLOWED_MIGRATION_GROUPS.items()
+        }
+        group = kwargs["group"]
+        if group not in ADMIN_ALLOWED_MIGRATION_GROUPS:
+
+            return make_response(jsonify({"error": "Group not allowed"}), 400)
+
+        if "action" in kwargs:
```
MeredithAnya, line 40: I think we may want to have a check for the dry_run scenario and by pass checking the policies.

snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,57 @@
...
+def check_migration_perms(f: Callable[..., Response]) -> Callable[..., Response]:
+    @wraps(f)
+    def check_group_perms(*args: Any, **kwargs: Any) -> Response:
+        ADMIN_ALLOWED_MIGRATION_GROUPS: Dict[str, MigrationPolicy] = {
+            k: policy_map[v] for k, v in settings.ADMIN_ALLOWED_MIGRATION_GROUPS.items()
+        }
+        group = kwargs["group"]
+        if group not in ADMIN_ALLOWED_MIGRATION_GROUPS:
+
+            return make_response(jsonify({"error": "Group not allowed"}), 400)
```
evanh, line 38: I think permission errors should be 403, not 400.

tests/admin/clickhouse_migrations/test_api.py
```diff
@@ -100,14 +110,24 @@ def test_run_reverse_migrations(admin_api: FlaskClient, action: str) -> None:
...
+            "system": "AllMigrationsPolicy",
+            "invalid_but_allowed_group": "AllMigrationsPolicy",
+            "generic_metrics": "NoMigrationsPolicy",
+            "events": "NonBlockingMigrationsPolicy",
+        },
     ):
         # invalid action
-        response = admin_api.post("/migrations/system/invalid_action/0001_migrations/")
+        response = admin_api.post(
+            "/migrations/system/invalid_but_allowed_group/0001_migrations/"
```
lynnagara, line 123: To be honest, I'm not a big fan of this URL scheme. I don't understand what `system` means here, and would personally prefer something that aligns more closely with our CLI commands, for example:

`POST /migrations/<action> -d {"group": "some_group", "migration_id": "0001...", }` where action is one of run, reverse, list, etc

snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,57 @@
...
+
+policy_map = {
+    "AllMigrationsPolicy": AllMigrationsPolicy(),
+    "NoMigrationsPolicy": NoMigrationsPolicy(),
+    "NonBlockingMigrationsPolicy": NonBlockingMigrationsPolicy(),
+}
+
+
+class MigrationActionConverter(BaseConverter):
+    regex = r"(?:run|reverse)"
```
lynnagara, line 26: why do we need this regex? can't we just exact match on `run` and `reverse` which are really the only two actions we want to support?

tests/admin/clickhouse_migrations/test_api.py
MeredithAnya: I think it makes sense to have the group in the top level since we also have a list for the groups

what if we had

`/migrations_groups/<group>/list`
`/migrations_groups/<group>/<run | reverse>/` -> and this can have the migration id, fake, force and dry run in the data?

lynnagara: Wouldn't there be cases though where passing a group might not be necessary? e.g. we might want the ability to list everything at once or get all the groups. How would those fit in?

IMO it's cleaner to put all migration system functionality under `/migrations` rather than potentially split the interface in two,

snuba/admin/migrations_policies.py
dbanda: This allows us to reuse the same function for both. Otherwise we would have to have two separate functions for run/reverse the have those call an auxiliary function. I'm fine either way.

snuba/admin/migrations_policies.py
dbanda: the keys are set in `settings.ADMIN_ALLOWED_MIGRATION_GROUPS` in the settings file. `policy_map` is a static field that maps the policy name to class. It's not to be changed by the user. moved it inside to the function to make that clear

snuba/admin/migrations_policies.py
dbanda: Added

snuba/admin/migrations_policies.py
dbanda: added

snuba/admin/migrations_policies.py
dbanda: fixed

tests/admin/clickhouse_migrations/test_api.py
dbanda: also fixed a typo on this line. it should be `"/migrations/system/invalid_action/0001_migrations/"`

snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,68 @@
...
+    that the migration has the appropriate run/reverse policy
+    """
+
+    @wraps(f)
+    def check_group_perms(*args: Any, **kwargs: Any) -> Response:
+        policy_map = {
+            "AllMigrationsPolicy": AllMigrationsPolicy(),
+            "NoMigrationsPolicy": NoMigrationsPolicy(),
+            "NonBlockingMigrationsPolicy": NonBlockingMigrationsPolicy(),
+        }
```
fpacifici, line 36: We now have a concept of registered classes to resolve a class by name. 
https://github.com/getsentry/snuba/blob/master/snuba/utils/registered_class.py#L46

You may want to remove this boilerplate. 
These are only three and quite static so no big deal, up to you.

snuba/admin/migrations_policies.py
```diff
@@ -0,0 +1,68 @@
...
+            return make_response(jsonify({"error": "Group not allowed"}), 403)
+
+        if "action" in kwargs:
+            action = kwargs["action"]
+            migration_id = kwargs["migration_id"]
+            migration_key = MigrationKey(group, migration_id)
+            policy = ADMIN_ALLOWED_MIGRATION_GROUPS[group]
+
+            def str_to_bool(s: str) -> bool:
+                return s.strip().lower() in ("yes", "true", "t", "1")
```
fpacifici, line 51: Let's avoid this and require a single valid value for true and a single valid value for false.

snuba/admin/migrations_policies.py
dbanda: fixed

snuba/admin/migrations_policies.py
dbanda: fixed

MeredithAnya (approved):

PR #3277, opened 2022-10-20
feat(TDC): Create transactions dataset config file

enochtangg:
Changes made:
* Transactions dataset converted to .yaml

### DIFF
--- a/snuba/datasets/configuration/transactions/dataset.yaml
+++ b/snuba/datasets/configuration/transactions/dataset.yaml
@@ -1,0 +1,8 @@
+version: v1
+kind: dataset
+name: transactions
+
+is_experimental: false
+
+entities:
+  all: [transactions]


lynnagara (dismissed):

lynnagara:
The comment says `Transactions dataset converted to .yaml` but this seems like it's only adding the yaml version not converting anything. When is the old version going to be deleted?

lynnagara:
@enochtangg If both exist, can you please confirm which one is the one that actually applies in prod and which is the one that is ignored?

evanh:
> which one is the one that actually applies in prod

We are defaulting to using configuration versions of Datasets in production.

> or do we want to keep it for some reason?

We've been preserving these initial Python versions so we can more easily roll back if an issue happens.

lynnagara:
Ok. Are those __init__ files needed in yaml land?

rahul-kumar-saini:
https://github.com/getsentry/snuba/blob/master/tests/datasets/configuration/test_dataset_loader.py
This can probably be extended to cover the additional dataset like the [entity loader test](https://github.com/getsentry/snuba/blob/master/tests/datasets/configuration/test_entity_loader.py#L41) was extended to cover the new transactions entity.

enochtangg:
Attempted to do some Git magic and ended up deleting this by accident, created a new PR here: https://github.com/getsentry/snuba/pull/3280

PR #3278, opened 2022-10-20
feat(TDC): Add support for transactions entity subscription to config

enochtangg:
This PR is responsible for adding support for the `TransactionsSubscription` to move into config. For some context, the original `PluggableEntitySubscription` only inherited from `SessionsSubscription`. This is because we did not want config to support inheritance. However, this becomes an issue we begin to migrate the Transactions entity subscription. Since the TransactionsSubscription inherits from something different we need to add support for this. There were a few options that were considered:

* Create a separate pluggable entity which supports a different base class
* Dynamic inheritance with using `type()` in `__new__()` of `PluggableEntitySubscription`
* Encode subscription into the Entity itself.

For the purpose of this task, creating a separate pluggable entity subscription class seemed like a quick/easiest approach (future work would be to eliminate entity subscriptions altogether and embed into entity?).

As a result, this PR contains the following changes:
* Added additional pluggable entity subscription class which inherits the necessary classes that `TransactionsSubscription` requires.
* Added field in config file called `type: [simple | complex]` which allows user to configure which base class to use.
* Created two new top level classes: `SimpleEntitySubscription` and `ComplexEntitySubscription`
* Refactored all entity subscriptions to inherit from these two classes (previously a bunch of them were inheriting from the SessionsSubscription).

### DIFF
--- a/snuba/datasets/configuration/entity_subscription_builder.py
+++ b/snuba/datasets/configuration/entity_subscription_builder.py
@@ -1,12 +1,13 @@
 import logging
-from typing import Type
+from typing import Type, Union
 
 import sentry_sdk
 
 from snuba.datasets.configuration.json_schema import V1_ENTITY_SUBSCIPTION_SCHEMA
 from snuba.datasets.configuration.loader import load_configuration_data
 from snuba.datasets.entity_subscriptions.pluggable_entity_subscription import (
-    PluggableEntitySubscription,
+    PluggableEntitySubscriptionComplex,
+    PluggableEntitySubscriptionSimple,
 )
 
 logger = logging.getLogger("snuba.entity_subscriptions_builder")
@@ -14,18 +15,29 @@ logger = logging.getLogger("snuba.entity_subscriptions_builder")
 
 def build_entity_subscription_from_config(
     file_path: str,
-) -> Type[PluggableEntitySubscription]:
+) -> Type[Union[PluggableEntitySubscriptionComplex, PluggableEntitySubscriptionSimple]]:
     config = load_configuration_data(
         file_path, {"entity_subscription": V1_ENTITY_SUBSCIPTION_SCHEMA}
     )
+
     with sentry_sdk.start_span(
         op="build", description=f"Entity Subscription: {config['name']}"
     ):
+        # Default to simple entity subscription
+        PluggableEntitySubscription: Type[
+            Union[PluggableEntitySubscriptionComplex, PluggableEntitySubscriptionSimple]
+        ] = PluggableEntitySubscriptionSimple
+
+        if "type" in config and config["type"] == "complex":
+            PluggableEntitySubscription = PluggableEntitySubscriptionComplex
+
         PluggableEntitySubscription.name = config["name"]
-        PluggableEntitySubscription.MAX_ALLOWED_AGGREGATIONS = config[
-            "max_allowed_aggregations"
-        ]
-        PluggableEntitySubscription.disallowed_aggregations = config[
-            "disallowed_aggregations"
-        ]
-        return PluggableEntitySubscription
+        if "max_allowed_aggregations" in config:
+            PluggableEntitySubscription.MAX_ALLOWED_AGGREGATIONS = config[
+                "max_allowed_aggregations"
+            ]
+        if "disallowed_aggregations" in config:
+            PluggableEntitySubscription.disallowed_aggregations = config[
+                "disallowed_aggregations"
+            ]
+    return PluggableEntitySubscription
--- a/snuba/datasets/entity_subscriptions/entity_subscription.py
+++ b/snuba/datasets/entity_subscriptions/entity_subscription.py
@@ -62,9 +62,17 @@ class EntitySubscriptionValidation:
             NoTimeBasedConditionValidator(entity.required_time_column).validate(query)
 
 
-class SessionsSubscription(EntitySubscriptionValidation, EntitySubscription):
-    MAX_ALLOWED_AGGREGATIONS: int = 2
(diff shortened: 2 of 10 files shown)


lynnagara (changes requested):
I'm not sure that classifying all entity subscriptions as `complex` of `simple` is the right way to go, nor does it seem like an intuitive user experience for end users to be picking one of these options. If the simple one is so much easier why can't we just treat it as the basic case of the "complex" pluggable entity subscription, which seems like it should be able to handle all cases.

snuba/datasets/configuration/entity_subscription_builder.py
```diff
@@ -1,31 +1,43 @@
...
+
     with sentry_sdk.start_span(
         op="build", description=f"Entity Subscription: {config['name']}"
     ):
+        # Default to simple entity subscription
+        PluggableEntitySubscription: Type[
+            Union[PluggableEntitySubscriptionComplex, PluggableEntitySubscriptionSimple]
+        ] = PluggableEntitySubscriptionSimple
+
+        if "type" in config and config["type"] == "complex":
```
evanh, line 31: I don't think we need this check or the `type` field. If the configuration has one of the `aggregations` fields below, we know it's complex. Otherwise it's simple.

snuba/datasets/entity_subscriptions/entity_subscription.py
```diff
@@ -62,9 +62,17 @@ def validate_query(self, query: Union[CompositeQuery[Entity], Query]) -> None:
             NoTimeBasedConditionValidator(entity.required_time_column).validate(query)
 
 
-class SessionsSubscription(EntitySubscriptionValidation, EntitySubscription):
-    MAX_ALLOWED_AGGREGATIONS: int = 2
+class SimpleEntitySubscription(EntitySubscriptionValidation, EntitySubscription):
```
evanh, line 65: This will be controversial, but I don't think you need to change this file at all. The new `Pluggable` class you added can inherit from the `TransactionsSubscription` class and the other one can inherit from `SessionsSubscription` like it does now.

snuba/datasets/entity_subscriptions/pluggable_entity_subscription.py
```diff
@@ -1,15 +1,30 @@
...
 
-from snuba.datasets.entity_subscriptions.entity_subscription import SessionsSubscription
+from snuba.datasets.entity_subscriptions.entity_subscription import (
+    ComplexEntitySubscription,
+    SimpleEntitySubscription,
+)
 
 
-class PluggableEntitySubscription(SessionsSubscription):
+class PluggableEntitySubscriptionComplex(ComplexEntitySubscription):
```
evanh, line 9: Nit: I don't think `Simple` and `Complex` are the right terms. I would leave the `complex` one just as `PluggableEntitySubscription` and add `BasicPluggableEntitySubscription`.

enochtangg:
> I'm not sure that classifying all entity subscriptions as `complex` of `simple` is the right way to go, nor does it seem like an intuitive user experience for end users to be picking one of these options. If the simple one is so much easier why can't we just treat it as the basic case of the "complex" pluggable entity subscription, which seems like it should be able to handle all cases.

Makes sense, the purpose of that was to allow certain entity subscriptions to extend different methods. The main distinguishing feature between `simple` and `complex` is that `ComplexEntitySubscription` has an `org_id` and a few specific methods that uses that `org_id`.

I'm not sure how we can couple the two together given their properties? Would renaming the field flag to `has_org_id: [true | false]` and renaming `PluggableEntitySubscriptionComplex` -> `PluggableEntitySubscriptionWithOrgId` and `PluggableEntitySubscriptionSimple` -> `PluggableEntitySubscription` make more sense?

evanh (commented):
I'm not sure that classifying all entity subscriptions as complex of simple is the right way to go, nor does it seem like an intuitive user experience for end users to be picking one of these options. If the simple one is so much easier why can't we just treat it as the basic case of the "complex" pluggable entity subscription, which seems like it should be able to handle all cases.

> nor does it seem like an intuitive user experience for end users to be picking one of these options.

You're right. We have a tech debt task to make all of this simpler that I'm hoping will get done in this next tech debt week.

> If the simple one is so much easier why can't we just treat it as the basic case of the "complex" pluggable entity subscription

Short answer: this is the fastest way to get TDC done and anything else is tech debt.

Long answer: Right now `EntitySubscription` is an abstract class, which means a concrete class is needed even for the base case. That's what the "simple" class is. The "complex" class doesn't technically extend the "simple" class, it extends the abstract base class. We could remove the abstract base class, but now we're going down the road of refactoring `EntitySubscription` and if we're going to do that I want to remove these classes completely.

snuba/datasets/entity_subscriptions/entity_subscription.py
enochtangg: Right, that was my initial thought, but the `TransactionsSubscription` is itself the transactions subscriptions entity and the `SessionsSubscription` is itself the sessions subscriptions entity. Which means when we eventually remove these components because they are all in config, then things will break. That's my rational for making two parent classes which all python defined subscriptions entity inherits from. That way, when we eventually delete the child python entity subscriptions, the config ones wont break.

lynnagara:
Some ideas discussed offline with Enoch

- We can add a `has_org_id` property and remove the simple vs complex thing
- Pluggable entity does not seem to add value and just causes indirection, just use the underlying entities directly? That way users can easily grep for them in code
- May be better to keep the underlying EntitySubscription class names (similar to column mapper approach) which would mean any type of entity subscription with custom conditions can be added in the future. This approach is highly coupled to the ones we happen to have in the system right now and is not very generic

PR #3279, opened 2022-10-20
fix(logs): Snuba init transaction was not sampled

rahul-kumar-saini:
### Overview
- Snuba's sampling rate is 0 and we don't ever start a transaction anywhere
- I introduced a transaction in #3200 which is populating the Datadog metric just fine but Snuba isn't sending the transaction so we basically still have no visibility
- This PR swaps out our hardcoded sampling rate with a function that sets sampling rate to 1 if the transaction op is `snuba_init` and the default 0 otherwise

### Questions
- This works fine locally - could this break anything in prod? (There are no instances where we start a fresh transaction anywhere in Snuba so this is the first time we're adding one)
- Is it possible that some global transaction (coming from wherever the current Snuba transactions are started) could have the op name "snuba_init" and hence be sampled 100%?
    - If so, would having the check be more thorough be sufficient?
    - Eg `if op == "snuba_init" and name == "Snuba CLI Initialization"` instead of just `if op == "snuba_init"`

### DIFF
--- a/snuba/environment.py
+++ b/snuba/environment.py
@@ -1,8 +1,8 @@
-from __future__ import absolute_import, annotations
+from __future__ import absolute_import
 
 import logging
 import os
-from typing import Any, Optional
+from typing import Optional
 
 import sentry_sdk
 import structlog
@@ -70,12 +70,6 @@ def setup_logging(level: Optional[str] = None) -> None:
     )
 
 
-def _traces_sampler(sampling_context: dict[str, Any]) -> int:
-    if sampling_context["transaction_context"]["op"] == "snuba_init":
-        return 1
-    return settings.SENTRY_TRACE_SAMPLE_RATE
-
-
 def setup_sentry() -> None:
     sentry_sdk.init(
         dsn=settings.SENTRY_DSN,
@@ -86,7 +80,7 @@ def setup_sentry() -> None:
             RedisIntegration(),
         ],
         release=os.getenv("SNUBA_RELEASE"),
-        traces_sampler=_traces_sampler,
+        traces_sample_rate=settings.SENTRY_TRACE_SAMPLE_RATE,
     )
 
 
--- a/snuba/environment.py
+++ b/snuba/environment.py
@@ -1,8 +1,8 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, annotations
 
 import logging
 import os
-from typing import Optional
+from typing import Any, Optional
 
 import sentry_sdk
 import structlog
@@ -70,6 +70,12 @@ def setup_logging(level: Optional[str] = None) -> None:
     )
 
 
+def _traces_sampler(sampling_context: dict[str, Any]) -> int:
+    if sampling_context["transaction_context"]["op"] == "snuba_init":
+        return 1
+    return settings.SENTRY_TRACE_SAMPLE_RATE
+
+
 def setup_sentry() -> None:
     sentry_sdk.init(
         dsn=settings.SENTRY_DSN,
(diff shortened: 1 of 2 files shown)


onewland:
I'm good with this in principle -- given that we have this outstanding issue where we believe Snuba is erroneously sending a bunch of transactions from the consumer (https://getsentry.atlassian.net/browse/SNS-1637) I wonder if we could ask ops deploy this to just a low-volume consumer first and make sure that we don't see a spike in transactions in Sentry or S4S for Snuba.

snuba/environment.py
```diff
@@ -70,6 +70,12 @@ def setup_logging(level: Optional[str] = None) -> None:
     )
 
 
+def _traces_sampler(sampling_context: dict[str, Any]) -> int:
+    if sampling_context["transaction_context"]["op"] == "snuba_init":
+        return 1
+    return settings.SENTRY_TRACE_SAMPLE_RATE
+
```
fpacifici, line 77: I don't think this works as you would want. 
https://github.com/getsentry/sentry-python/blob/c05bcf598c5455a6f35eabd18c840c4544c9392c/sentry_sdk/tracing.py#L723-L739

If you do not define a `trace_sampler`, snuba honors the decision made by sentry, thus making the trace consistent. Which is what we want. 
But if you define a `traces_sampler`, your sampler needs to explicitly honor the parent decision, you cannot just return a sampling rate.


Also the call `start_transaction` is supposed to have a way to force a the sampling 
```
start_transaction(name: "my transaction", sampled: True)
```
so you should not need this,

fpacifici (changes requested):

snuba/environment.py
MeredithAnya: as per the conversation that happened in https://github.com/getsentry/snuba/pull/3018, my understanding is that we could just be increasing `SENTRY_TRACE_SAMPLE_RATE` in our prod settings, which would still honor the sentry decisions it would just mean new traces in snuba would get that new rate.

fpacifici: You do not want to sample 100% of the transactions started in snuba that are not of the specific type Rahul is interested into. That would also sample the massive amount of transactions that seem to originate from our consumers that we never managed to investigate. (Looking at the outcomes table there are several orders of magnitude more transactions dropped by the sdk than those that reach sentry. These are likely to originate in our consumers).

MeredithAnya: I thought there were no other `start_transaction` beside rahul's. But anyway according to our docs the `start_transaction` does have a way to enable sampling and it seems like it overrides whatever sample rate is else where so filippo's suggestion still stands

rahul-kumar-saini: I see, I'll try setting `sampled: True` and see if it'll force Sentry to sample the transaction. As far as Snuba transactions go, this is the only transaction explicitly started within Snuba but I didn't know whether the sampling rate in settings only mattered for transactions started within Snuba or all transactions Snuba's logic is a part of.

evanh: There are some docs on the `traces_sampler` function [here](https://docs.sentry.io/platforms/python/guides/falcon/configuration/options/#traces-sampler). Filippo's link actually explains it better.

The sampling happens to every transaction started with our DSN. The transactions are still started in Snuba, even though they are implicitly created by our integrations. The web API creates transactions through the Flask integration. Because we don't have `traces_sampler` set, we are effectively sampling based on whether the parent transaction (from Sentry in our case) is sampling or not. So if Sentry samples, so does Snuba, and vice versa.

evanh (approved):

fpacifici (approved):

PR #3280, opened 2022-10-21
feat(TDC): Create transactions dataset config file

enochtangg:
This PR is the same as https://github.com/getsentry/snuba/pull/3277

### DIFF
--- a/snuba/datasets/configuration/transactions/dataset.yaml
+++ b/snuba/datasets/configuration/transactions/dataset.yaml
@@ -5,4 +5,4 @@ name: transactions
 is_experimental: false
 
 entities:
-  all: [transactions]
+  - transactions
--- a/snuba/datasets/configuration/transactions/dataset.yaml
+++ b/snuba/datasets/configuration/transactions/dataset.yaml
@@ -1,0 +1,8 @@
+version: v1
+kind: dataset
+name: transactions
+
+is_experimental: false
+
+entities:
+  all: [transactions]
--- a/snuba/datasets/configuration/dataset_builder.py
+++ b/snuba/datasets/configuration/dataset_builder.py
@@ -15,6 +15,6 @@ def build_dataset_from_config(config_file_path: str) -> PluggableDataset:
     with sentry_sdk.start_span(op="build", description=f"Dataset: {config['name']}"):
         return PluggableDataset(
             name=config["name"],
-            all_entities=[EntityKey(key) for key in config["entities"]["all"]],
+            all_entities=[EntityKey(key) for key in config["entities"]],
             is_experimental=bool(config["is_experimental"]),
         )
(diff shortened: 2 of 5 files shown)


enochtangg:
Added test to dataset factory for Transactions dataset loaded from config

evanh (approved):

snuba/datasets/configuration/transactions/dataset.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
+kind: dataset
+name: transactions
+
+is_experimental: false
```
nikhars, line 5: Is this field mandatory for all datasets? Can we make it optional such that only new datasets need to set it up as experimental.

snuba/datasets/configuration/transactions/dataset.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+entities:
+  all: [transactions]
```
nikhars, line 8: Could you elaborate what the `all` parameter does here and why it is needed?

snuba/datasets/configuration/transactions/dataset.yaml
enochtangg: The field itself is not required by json_schema: https://github.com/getsentry/snuba/blob/3d12cf9af8a2f4acd28c5dcb825d1ece2a0070b8/snuba/datasets/configuration/json_schema.py#L434-L439

However, in `dataset_builder`, it looks like it is required or else it raises KeyError: https://github.com/getsentry/snuba/blob/3d12cf9af8a2f4acd28c5dcb825d1ece2a0070b8/snuba/datasets/configuration/dataset_builder.py#L19

I've applied the fix to make it required which was the original intention I believe. 

To your question: Since the `Dataset` class requires `is_experimental`, I think we could make it "optional" by setting some default if not specified. However, I will differ to others with more context to decide how important it is to be explicit with this flag.

snuba/datasets/configuration/transactions/dataset.yaml
enochtangg: From my understanding, there is no functional reason why this is here other than to map closer to the Python definition of Dataset? I think we can remove this.

snuba/datasets/configuration/transactions/dataset.yaml
evanh: My opinion is that explicit is better than implicit. A Dataset needs to know whether it is experimental or not, so we should explicitly codify that.

snuba/datasets/configuration/transactions/dataset.yaml
evanh: I think you can list the entities themselves, without the `all` keyword:

```
entities:
- transactions
- ...
```

PR #3281, opened 2022-10-21
feat(redis): Add second shard to rate limiter and cache

untitaker:
We want to split out those usecases to a new redis cluster, but instead
of doing a hard cutover to a new redis instance we want to do it slowly.

We don't care much about dataloss (particularly for those usecases), our
primary concern is whether the new Redis instance can take the load of
those usecases. We have little insight into what causes the most load on
the current redis cluster, and at the same time hope to get away with a
single memorystore instance per-usecase as memorystore does not support
sharding yet.

In order to facilitate a "slow" cutover we introduce two new rollout
rate settings. Increasing a rollout rate will make the corresponding
usecase gradually start a different redis cluster (represented as a new
"v2 usecase"), and once we hit 1.0 on both of them we can essentially
revert this PR and point the "v1" usecase directly to the new cluster
(in one deploy)

There is a chance this doesn't go well and a single memorystore instance
cannot take the load of either the rate limit or caching usecase, in
which case we have to consider building sharding into Snuba.

Longer-term I'd ideally want us to move to Envoy's Redis proxy or a
similar abstraction layer between the redis client and the actual
instances. That would give us dual-writes and key-based sharding OOTB
without having to build it into the app itself.

### DIFF
--- a/snuba/state/rate_limit.py
+++ b/snuba/state/rate_limit.py
@@ -5,12 +5,14 @@ import uuid
 from collections import ChainMap, namedtuple
 from contextlib import AbstractContextManager, ExitStack, contextmanager
 from dataclasses import dataclass
+from hashlib import md5
 from types import TracebackType
 from typing import ChainMap as TypingChainMap
 from typing import Iterator, MutableMapping, Optional, Sequence, Type
 
 from snuba import environment, state
 from snuba.redis import RedisClientKey, get_redis_client
+from snuba.util import force_bytes
 from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.serializable_exception import SerializableException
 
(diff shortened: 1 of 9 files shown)


snuba/state/rate_limit.py
```diff
@@ -124,16 +129,27 @@ def rate_limit(
...
+        ]
     )
     assert isinstance(rate_history_s, (int, float))
+    assert isinstance(v2_cluster_rollout_rate, float)
 
     if bypass_rate_limit == 1:
         yield None
         return
 
+    should_use_v2 = (
```
evanh, line 147: I think this sharding choice logic should be inside the RedisCache, instead of duplicated in the code every place we want a v2 shard. Maybe we need a wrapper class around multiple `RedisCache`?

untitaker: i agree with this but i had hoped this code will be removed after we have moved the usecases over (so, within one week)

evanh (approved):

snuba/state/rate_limit.py
untitaker: "shard" is a misnomer i guess, we shard the keyspace across old/new cluster and make one shard gradually bigger in order to migrate traffic/load

PR #3290, opened 2022-10-24
feat(TDC): Migrate transactions migrate group to configuration

evanh:
Move the transaction migration group to configuration. Since the transactions
migrations need to be executed before the discover migrations, also add the
concept of an "execute_before" field, which can be used to organize the
migrations.

### DIFF
--- a/snuba/datasets/configuration/json_schema.py
+++ b/snuba/datasets/configuration/json_schema.py
@@ -515,6 +515,10 @@ V1_MIGRATION_GROUP_SCHEMA = {
             "type": "boolean",
             "description": "Flag to determine if migration group is optional",
         },
+        "execute_before": {
+            "type": "string",
+            "description": "The migration group this should be run directly before.",
+        },
         "migrations": {
             "type": "array",
             "items": {"type": "string", "description": "Names of migrations"},
--- a/snuba/datasets/configuration/transactions/migrations/transactions_migration_group.yaml
+++ b/snuba/datasets/configuration/transactions/migrations/transactions_migration_group.yaml
@@ -1,0 +1,24 @@
+version: v1
+kind: migration_group
+name: transactions
+optional: false
+execute_before: discover
+
+migrations:
+  - 0001_transactions
+  - 0002_transactions_onpremise_fix_orderby_and_partitionby
+  - 0003_transactions_onpremise_fix_columns
+  - 0004_transactions_add_tags_hash_map
+  - 0005_transactions_add_measurements
+  - 0006_transactions_add_http_fields
+  - 0007_transactions_add_discover_cols
+  - 0008_transactions_add_timestamp_index
+  - 0009_transactions_fix_title_and_message
+  - 0010_transactions_nullable_trace_id
+  - 0011_transactions_add_span_op_breakdowns
+  - 0012_transactions_add_spans
+  - 0013_transactions_reduce_spans_exclusive_time
+  - 0014_transactions_remove_flattened_columns
+  - 0015_transactions_add_source_column
+  - 0016_transactions_add_group_ids_column
+  - 0017_transactions_add_app_start_type_column
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -65,6 +65,7 @@ class ConfigurationLoader(DirectoryLoader):
         ]
         self.optional = self.migration_group.get("optional", False)
         migration_path = migration_path or "snuba.snuba_migrations"
+        self.execute_before = self.migration_group.get("execute_before")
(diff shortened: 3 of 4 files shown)


volokluev:
I think the migrations folks should really take a look at this @MeredithAnya @dbanda

snuba/datasets/configuration/json_schema.py
```diff
@@ -515,6 +515,10 @@ def registered_class_schema(
             "type": "boolean",
             "description": "Flag to determine if migration group is optional",
         },
+        "execute_before": {
+            "type": "string",
+            "description": "The migration group this should be run directly before.",
+        },
```
fpacifici, line 521: I am not following the reasoning for this. 
Migrations group are designed to be independent from each other so that you do not need to run all of them in your environment. 
Which is what we do as we have disabled groups. 
This implies that there cannot be a dependency relationship between groups.

lynnagara:
A couple of comments:
1. Migrations are not a property of datasets and cannot be nested under the datasets directory. Can you move them to a different module? Datasets can specify the migration groups they depend on but we cannot have any relationship going the other way (that is, migrations must know nothing about datasets).

2. Dependencies between migration groups are complicated and it might be better to not support this. There are many edge cases to consider - e.g. if one depends on the other what order do the migrations get run in. For adding and removing columns for example we might need to flip the order just like for local and distributed migrations. So instead how about we just have the dataset enumerate the list of migration groups it depends on. For most this will be only 1 group, but for Discover dataset this will be `errors, transactions, discover`. What do you think?

snuba/datasets/configuration/json_schema.py
evanh: Well there is a dependency between groups, since the discover group can't be run before the transactions/errors group. Right now that dependency is encoded implicitly in the order of `REGISTERED_GROUPS` but that is no longer reliable with the way configurations are loaded.

evanh:
> Migrations are not a property of datasets and cannot be nested under the datasets directory.

We already have other migration groups in the `datasets` folder, this is the same process we went through with generic metrics (see [here](https://github.com/getsentry/snuba/pull/3071). The group is not aware of the Dataset in any way, except that it happens to be in a folder called `transactions`.

> Dependencies between migration groups are complicated and it might be better to not support this. ... For most this will be only 1 group, but for Discover dataset this will be errors, transactions, discover. What do you think?

That makes some sense to me, but doesn't solve the immediate problem of how do I ensure that the discover migration group runs after the transactions migration group.

evanh:
@lynnagara @fpacifici Moving the discussion on this [here](https://www.notion.so/sentry/Migration-Groups-in-Configuration-c02d42a6436646f390fc2e5c502d6a61). I'll summarize any decisions in this PR as well.

snuba/datasets/configuration/json_schema.py
fpacifici: > Well there is a dependency between groups, since the discover group can't be run before the transactions/errors group. Right now that dependency is encoded implicitly in the order of REGISTERED_GROUPS but that is no longer reliable with the way configurations are loaded.

That's because the data model for discover and transaction migrations is broken. It is not true that discover depends on transactions. If we drop a column the dependency is inverted. In that case discover *must* be executed first.
So we cannot introduce that relationship in the framework knowing that it generates an impossible scenario that could not work.

lynnagara:
> We already have other migration groups in the `datasets` folder, this is the same process we went through with generic metrics (see [here](https://github.com/getsentry/snuba/pull/3071). The group is not aware of the Dataset in any way, except that it happens to be in a folder called `transactions`.

This isn't right and we need to revisit it. Sorry I didn't catch it on the earlier PR. I don't think it makes sense to be in a folder called `transactions` either since technically the tables created by migration groups can be shared between datasets. This is enforcing a 1:1 relationship here that shouldn't exist.

> > Dependencies between migration groups are complicated and it might be better to not support this. ... For most this will be only 1 group, but for Discover dataset this will be errors, transactions, discover. What do you think?
> 
> That makes some sense to me, but doesn't solve the immediate problem of how do I ensure that the discover migration group runs after the transactions migration group.

The migration groups were previously enumerated in one place and we ran them in that order. The easiest thing to do would be to just keep it that way. Is this something you are planning to fully remove? I'm not sure I fully understand how the ordering of groups is determined now. Can you clarify?

dbanda:
My 2c is that ordering with `execute_before` would be tricky with many edge cases. We found when doing the migration there can be many different ordering constraints depending on the SQL expression, and if it is being applied to distributed or local tables.

evanh:
This architecture needs more discussion, so I'm going to close this PR for now.

PR #3291, opened 2022-10-25
feat(querylog): Expose ClickHouse Querylog to Snuba Admin API

rahul-kumar-saini:
## Overview
This is the first of a few PRs which will enable ClickHouse Querylog table to be queryable from Snuba Admin console. This PR sets up the backend endpoints which will be requested by the front end.

The audit log solution is fairly simple - any query that is attempted to be executed against the querylog table (aside from describing the table) is logged to GCP via a new logger using structlog. Each log contains the `user`, `query`, `status` (failed/succeeded), and start/end timestamps. (See sample log below)

### Blast Radius
- Snuba Admin API - added 2 new endpoints which are locked behind a runtime config.

### Before State
- No access to ClickHouse Querylog table except via SSH
    - Only accessible to those with SSH access to production ClickHouse

### After State
- A new admin API endpoint introduced to view the Schema of the Querylog table
    - Executes and returns the result of `DESCRIBE TABLE <querylog_table_name>`
- A new admin API endpoint introduced to run SELECT queries on the Querylog table
    - Validates query for read only command and valid querylog table being queried
    - Logs to GCP when a query is executed
- A new runtime config `enable_clickhouse_querylog_api` introduced to enable/disable the above APIs from executing queries. API routes will still exist and respond with 400 unless config is enabled.

### Sample GCP log for Audit logging
```
{"module": "snuba.admin.notifications.querylog.audit_log", "user": "rahul.saini@sentry.io", "query": "SELECT request_body, request_id FROM querylog_local WHERE projects=[1]", "status": "success", "event": "run_query", "start_timestamp": "2022-10-25T23:38:08.499432Z", "end_timestamp": "2022-10-25T23:38:08.499432Z", "severity": "info", "user_ip": "127.0.0.1", "endpoint": "clickhouse_querylog_query"}
```

## Testing Notes
Local Querylog Setup:
- Settings: flipped `RECORD_QUERIES` to `True`, removed `querylog` from `SKIPPED_MIGRATIONS`
- Ran `snuba migrations migrate --force` to create `querylog_local` table
- Ran `snuba consumer --storage=querylog --auto-offset-reset=latest --log-level=info` to start Querylog consumer
Dev testing:
- Ran `snuba admin` to start up admin API
- Ran `snuba devserver` to start up consumers
- Executed some queries via `http://localhost:1218/discover/snql`
- cURL'd some querylog queries to verify everything works with runtime config enabled/disabled

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,6 +29,7 @@ sentry-sdk==1.9.8
 simplejson==3.17.6
 structlog==22.1.0
 structlog-sentry==2.0.0
+sql-metadata==2.6.0
 urllib3==1.26.12
 pyuwsgi==2.0.20
 Werkzeug==2.2.2
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import MutableMapping
 
-from sql_metadata import Parser, QueryType
+from sql_metadata import Parser, QueryType  # type: ignore
 
 from snuba import settings
 from snuba.clickhouse.native import ClickhousePool
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -137,9 +137,7 @@ def validate_ro_query(sql_query: str, allowed_tables: list[str] | None = None) -
     if parsed.query_type != QueryType.SELECT:
         raise InvalidCustomQuery("Only SELECT queries are allowed")
 
-    if allowed_tables:
-        for table in parsed.tables:
-            if table not in allowed_tables:
-                raise InvalidCustomQuery(
-                    f"Invalid FROM clause, only the following tables are allowed: {allowed_tables}"
-                )
+    if allowed_tables and not set(parsed.tables).issubset(set(allowed_tables)):
+        raise InvalidCustomQuery(
+            f"Invalid FROM clause, only the following tables are allowed: {allowed_tables}"
+        )
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import MutableMapping
 
-# import sqlparse
 from sql_metadata import Parser, QueryType
 
 from snuba import settings
@@ -126,11 +125,14 @@ def validate_ro_query(sql_query: str, allowed_tables: list[str] | None = None) -
 
     Raises InvalidCustomQuery if query is invalid or not allowed.
     """
+    lowered = sql_query.lower()
+    disallowed_keywords = ["insert", ";"]
 
-    parsed = Parser(sql_query.lower())
+    for kw in disallowed_keywords:
+        if kw in lowered:
+            raise InvalidCustomQuery(f"{kw} is not allowed in the query")
 
-    if ";" in parsed.query:
-        raise InvalidCustomQuery("';' is not allowed in the query")
+    parsed = Parser(lowered)
 
     if parsed.query_type != QueryType.SELECT:
         raise InvalidCustomQuery("Only SELECT queries are allowed")
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -2,6 +2,9 @@ from __future__ import annotations
 
 from typing import MutableMapping
 
+# import sqlparse
+from sql_metadata import Parser, QueryType
+
 from snuba import settings
 from snuba.clickhouse.native import ClickhousePool
 from snuba.clusters.cluster import ClickhouseClientSettings, ClickhouseCluster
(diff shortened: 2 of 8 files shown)


snuba/admin/clickhouse/common.py
```diff
@@ -110,3 +112,33 @@ def get_ro_query_node_connection(
 
     CLUSTER_CONNECTIONS[storage_name] = connection
     return connection
+
+
+def validate_ro_query(sql_query: str, allowed_tables: list[str] | None = None) -> None:
```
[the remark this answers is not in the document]
volokluev: I agree, let's file a separate ticket for that

snuba/admin/clickhouse/querylog.py
```diff
@@ -0,0 +1,54 @@
...
+def run_querylog_query(query: str, user: str) -> ClickhouseResult:
+    """
+    Validates, audit logs, and executes given query against Querylog
+    table in ClickHouse.
+    """
+    schema = get_storage(StorageKey.QUERYLOG).get_schema()
+    assert isinstance(schema, TableSchema)
+    validate_ro_query(sql_query=query, allowed_tables=[schema.get_table_name()])
+    if get_config(ENABLE_QUERYLOG_API_CONFIG, 0):
+        audit_log_query(query, user)
```
volokluev, line 28: Can we instead make a decorator like 

```python3
@auditlog_func
def run_querylog_query(...):
    pass
```

This decorator should log every call to said function and whether it errored. That way we can be sure nobody forgets to put it in the right place if the function changes

snuba/admin/views.py
```diff
@@ -278,6 +279,93 @@ def clickhouse_trace_query() -> Response:
...
+@application.route("/clickhouse_querylog_query", methods=["POST"])
+def clickhouse_querylog_query() -> Response:
+    user = request.headers.get(USER_HEADER_KEY)
+    if not user:
+        return Response(
+            json.dumps({"error": "Unauthorized"}),
+            401,
+            {"Content-Type": "application/json"},
+        )
+    req = json.loads(request.data)
```
volokluev, line 291: Can you file a tech debt ticket to make this automatic? this is currently very easy to misuse/forget

snuba/admin/clickhouse/querylog.py
volokluev: You can do it in this PR or file a ticket to do it later

volokluev (approved):
LGTM overall. Some stuff that should be cleaned up but I wouldn't block the PR over them

volokluev:
Actually echoing what @evanh said. There should be a test to make sure that the audit log was created. If you make it into a decorator, you can test the decorator in isolation as well

snuba/admin/views.py
```diff
@@ -278,6 +279,93 @@ def clickhouse_trace_query() -> Response:
...
+        )
+    try:
+        result = run_querylog_query(raw_sql, user)
+        rows = []
+        rows, columns = cast(List[List[str]], result.results), result.meta
+
+        if columns:
+            res = {}
+            res["column_names"] = [name for name, _ in columns]
+            res["rows"] = [[str(col) for col in row] for row in rows]
```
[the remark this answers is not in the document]
rahul-kumar-saini: Just followed an example endpoint which returns ClickHouse data:
https://github.com/getsentry/snuba/blob/85c38174832ecad29103b1daac697cd1a78d119f/snuba/admin/views.py#L205-L215

I assumed there is a reason for this, I don't have enough local data to cover some edge case this may be handling but I'll try testing without the cast.

snuba/admin/views.py
rahul-kumar-saini: I think it may not be necessary, there is an auth function that runs before every request so I don't think I even needed to do this.

I just added an `if not user` check so mypy won't complain about it possible being `None`, but I could've probably just `assert user is not None` since it won't ever be None in prod.

https://github.com/getsentry/snuba/blob/85c38174832ecad29103b1daac697cd1a78d119f/snuba/admin/views.py#L64-L70

snuba/admin/clickhouse/common.py
```diff
@@ -110,3 +112,33 @@ def get_ro_query_node_connection(
...
+
+    sql_query_split = sql_query.lower().split()
+    lowered = " ".join(sql_query_split)
+    if not lowered.startswith("select"):
+        raise InvalidCustomQuery("Only SELECT queries are allowed")
+
+    if allowed_tables:
+        if (
+            sql_query_split.count("from") != 1
+            or sql_query_split[sql_query_split.index("from") + 1] not in allowed_tables
```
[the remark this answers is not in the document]
rahul-kumar-saini: You're right, SQL parser time.

snuba/admin/views.py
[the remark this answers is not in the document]
evanh: I don't think we want this with the querylog. The system queries have a limit to what they can return. The Querylog queries could be returning numbers, lists, booleans or whatever. Preserving that will make it easier to format in the front end.

requirements.txt
```diff
@@ -29,6 +29,7 @@ sentry-sdk==1.9.8
 simplejson==3.17.6
 structlog==22.1.0
 structlog-sentry==2.0.0
+sql-metadata==2.6.0
```
evanh, line 32: Was this added to the https://github.com/getsentry/pypi/ repo as well?

snuba/admin/clickhouse/common.py
```diff
@@ -110,3 +114,30 @@ def get_ro_query_node_connection(
...
+    for kw in disallowed_keywords:
+        if kw in lowered:
+            raise InvalidCustomQuery(f"{kw} is not allowed in the query")
+
+    parsed = Parser(lowered)
+
+    if parsed.query_type != QueryType.SELECT:
+        raise InvalidCustomQuery("Only SELECT queries are allowed")
+
+    if allowed_tables and not set(parsed.tables).issubset(set(allowed_tables)):
```
evanh, line 140: You could make `allowed_tables` a set, so you don't have to cast it. Unless there's some other reason for it to be a list.

evanh (approved):
One comment about the response values, otherwise looks good.

requirements.txt
rahul-kumar-saini: Yep

PR #3292, opened 2022-10-25
ci(clickhouse): Enable running tests against clickhouse version 21.8

nikhars:
We have multiple versions of clickhouse running across different environments. Some clusters on SaaS run on version 21.8 while self hosted and single tenant run anywhere from 20.3 to 20.7. The goal of this PR is to ensure that the code we introduce passes on the minimum and maximum versions of clickhouse we support i.e. 20.3 and 21.8. This is done by parametrizing the clickhouse version in the docker compose used by gcb.

Only the tests with `SNUBA_SETTINGS=test` are run on multiple clickhouse versions. The others are not. We can add additional tests to run in the future if needed.

The way current github workflow is setup, `Tests and code coverage (test)` and `Tests and code coverage (test_distributed)` are required checks. This PR modifies those jobs so they have a different name. Once this PR is approved, I will disable the required checks from github so that this PR can be merged. After this PR is merged, I will make the required checks to be `Tests and code coverage (test, yandex/clickhouse-server:20.3.9.70)`, `Tests and code coverage (test, altinity/clickhouse-server:21.8.13.1.altinitystable)` and `Tests and code coverage (test_distributed, yandex/clickhouse-server:20.3.9.70)`.


tests/test_cleanup.py
```diff
@@ -55,6 +55,7 @@ class TestCleanup:
         test_data,
     )
     @patch("snuba.cleanup.current_time")
+    @pytest.mark.skip(reason="This test does not work with clickhouse version 21.8")
```
nikhars, line 58: Skipping this because ingesting an event past the TTL causes the partition to be inactive at ingest time on clickhouse version 21.8. On the older clickhouse version, it would still be an active partition.

tests/test_discover_api.py
```diff
@@ -191,17 +191,8 @@ def test_handles_columns_from_other_dataset(self) -> None:
...
 
         assert response.status_code == 200
-        assert data["data"] == [
-            {
-                "type": "transaction",
-                "count": 0,
-                "uniq_group_id": 0,
-                "uniq_ex_stacks": None,
-            }
-        ]
```
nikhars, line 204: clickhouse version 21.8 returns an empty response for this query.

tests/test_cleanup.py
evanh: Is this a temporary change? We shouldn't be skipping tests unless we plan to make them work at a later date.

.github/workflows/ci.yml
```diff
@@ -226,8 +226,14 @@ jobs:
...
       matrix:
-        snuba_settings:
-          ["test", "test_distributed", "test_distributed_migrations"]
+        snuba_settings: ["test"]
+        clickhouse_image: ['yandex/clickhouse-server:20.3.9.70',
+                           'altinity/clickhouse-server:21.8.13.1.altinitystable']
+        include:
+          - snuba_settings: "test_distributed"
+            clickhouse_image: 'yandex/clickhouse-server:20.3.9.70'
+          - snuba_settings: "test_distributed_migrations"
```
evanh, line 235: Why are these tests not done on 21.8?

nikhars: Running the distributed tests in parallel raises problems which I would like to punt fixing in the future. Here was a run where distributed tests were run in parallel https://github.com/getsentry/snuba/actions/runs/3147126533/jobs/5116339287

I would like to have something working and then iterating from there.

tests/test_cleanup.py
nikhars: Yes, it is temporary. The way the test is written right now, I can't make it work easily for both clickhouse versions. So I will look into how to enable running it only on one of them as a later step.

I agree that we shouldn't be skipping tests. If there were a bunch of them, I would start looking into them. But since i am disabling only 1 for now, I believe it is ok.

dbanda (approved):
LGTM

.github/workflows/ci.yml
evanh: OK makes sense.

evanh (approved):

PR #3294, opened 2022-10-25
ref(upgrades): Remore transactions_v2 storage

nikhars:
We have already upgraded our clickhouse fleet to the newer tiger clusters. The way they are mapped in production is `transactions` storage set key points to the tiger cluster. We have also stopped performing reads and writes on the old clusters. The storage set key `transactions_v2` is mapped to the older cluster. So we have the flexibility to remove the code referencing `transactions_v2`. Once this PR is merged, we can remove the mapping for `transactions_v2` from the `prod_settings.py` file.


lynnagara (approved):
🎉

PR #3298, opened 2022-10-25
ref(upgrades): Remove errors_v2 and errors_v2_ro storages

nikhars:
Similar to https://github.com/getsentry/snuba/pull/3294, but this one is for errors_v2 and errors_v2_ro storages.


snuba/cli/offline_replacer.py
```diff
@@ -21,7 +21,7 @@
 @click.option(
     "--storage",
     "storage_name",
-    type=click.Choice(["errors", "errors_v2"]),
+    type=click.Choice(["errors"]),
```
fpacifici, line 24: Should we just remove the parameter altogether ?

fpacifici (commented):
Please fix the type error

snuba/cli/offline_replacer.py
lynnagara: I think explicit is good. Offline replacer should work for others too, just a coincidence there is only one today.

nikhars: I would like to keep it for now. There is no harm in having it.

PR #3299, opened 2022-10-25
ci(reverts): Allow fast reverts

nikhars:
Add the workflow which allows fast reverts. This workflow has been adopted from the [sentry](https://github.com/getsentry/sentry/blob/master/.github/workflows/fast-revert.yml) repository. Note that this hasn't been tested in any way. Once this gets merged, I will make a small PR which I will try to revert using the new workflow.


mattgauntseo-sentry (approved):

evanh (approved):

.github/workflows/fast-revert.yml
```diff
@@ -0,0 +1,40 @@
...
+        co_authored_by: ${{ github.event.inputs.co_authored_by || format('{0} <{1}+{0}@users.noreply.github.com>', github.event.sender.login, github.event.sender.id) }}
+        committer_name: getsentry-bot
+        committer_email: bot@sentry.io
+        token: ${{ secrets.BUMP_SENTRY_TOKEN }}
+    - name: comment on failure
+      run: |
+        curl \
+            --silent \
+            -X POST \
+            -H 'Authorization: token ${{ secrets.BUMP_SENTRY_TOKEN }}' \
```
asottile-sentry, line 37: you will need this secret configured

nikhars: Thanks for pointing it out. How do I go about configuring the secret? Do we use a shared secret across all sentry repositories? Or should I create one just for the snuba repository using https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md

asottile-sentry: hmmm this is a good question -- `sentry` and `getsentry` use a PAT for a bot -- I wonder if the sentry internal app (globally available) has enough permissions to write a revert

if you want to try that out you could use:

```yaml
      - uses: getsentry/action-github-app-token@38a3ce582e170ddfe8789f509597c6944f2292a9  # v1
        id: token
        with:
          app_id: ${{ secrets.SENTRY_INTERNAL_APP_ID }}
          private_key: ${{ secrets.SENTRY_INTERNAL_APP_PRIVATE_KEY }}
```

and then use `${{ steps.token.outputs.token }}` in place of `${{ secrets.BUMP_SENTRY_TOKEN }}`

nikhars: Did what you suggested. Mind reviewing it again?

nikhars: You are right. Removed all usages of `BUMP_SENTRY_TOKEN`

asottile-sentry (approved):
hope it works!

PR #3301, opened 2022-10-26
ref(consumer): Merge Snuba Consumer and Multistorage Consumer, step 1

ayirr7:
Currently, the (single-storage) Consumer and Multistorage Consumer live in separate files, but share a lot of common logic and there's duplication within/across these files. The goal of this PR is to start the process of merging the two into a single consumer CLI entrypoint (`snuba/cli/consumer.py`) and Consumer Builder (`snuba/consumers/consumer_builder.py`). More context on this change can be found here: https://getsentry.atlassian.net/browse/SNS-1810. 

This PR: 
- Serves as the first incremental step
- Allows the Snuba consumer CLI to accept multiple storages 
- Introduces some logic from `multistorage_consumer.py` in constructor of `ConsumerBuilder` by resolving multiple topics for multiple storages
- `__build_consumer` and `__build_streaming_strategy_factory` do not have any multistorage logic yet
- Note that the consumer is still meant only for single-storage use cases (we are incrementally adding/transferring functionality until it works for both single and multistorage)

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -43,11 +43,12 @@ logger = logging.getLogger(__name__)
 )
 @click.option(
     "--storage",
-    "storage_name",
+    "storage_names",
     type=click.Choice(
         [storage_key.value for storage_key in get_writable_storage_keys()]
     ),
-    help="The storage to target",
+    help="The storage(s) to target",
+    multiple=True,
     required=True,
 )
 @click.option(
@@ -120,7 +121,7 @@ def consumer(
     commit_log_topic: Optional[str],
     consumer_group: str,
     bootstrap_server: Sequence[str],
-    storage_name: str,
+    storage_names: str,
     max_batch_size: int,
     max_batch_time_ms: int,
     auto_offset_reset: str,
@@ -139,21 +140,33 @@ def consumer(
     setup_logging(log_level)
     setup_sentry()
     logger.info("Consumer Starting")
-    storage_key = StorageKey(storage_name)
+    storage_keys = [
+        getattr(StorageKey, storage_name.upper()) for storage_name in storage_names
+    ]
 
-    metrics = MetricsWrapper(
-        environment.metrics,
-        "consumer",
-        tags={"group": consumer_group, "storage": storage_key.value},
-    )
-    configure_metrics(StreamMetricsAdapter(metrics))
+    if len(storage_keys) == 1:
+        metrics = MetricsWrapper(
+            environment.metrics,
+            "consumer",
+            tags={"group": consumer_group, "storage": storage_keys[0].value},
+        )
+        configure_metrics(StreamMetricsAdapter(metrics))
+    else:
+        metrics = MetricsWrapper(
+            environment.metrics,
+            "consumer",
+            tags={
+                "group": consumer_group,
+                "storage": "_".join([storage_keys[0].value, "m"]),
+            },
+        )
 
     def stats_callback(stats_json: str) -> None:
         stats = rapidjson.loads(stats_json)
         metrics.gauge("librdkafka.total_queue_size", stats.get("replyq", 0))
 
     consumer_builder = ConsumerBuilder(
-        storage_key=storage_key,
+        storage_keys=storage_keys,
         kafka_params=KafkaParameters(
             raw_topic=raw_events_topic,
             replacements_topic=replacements_topic,
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -116,6 +116,10 @@ class ConsumerBuilder:
                 },
             )
 
+            # XXX: This can result in a producer being built in cases where it's
+            # not actually required.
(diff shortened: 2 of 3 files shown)


snuba/cli/consumer.py
```diff
@@ -139,21 +140,33 @@ def consumer(
...
+            tags={"group": consumer_group, "storage": storage_keys[0].value},
+        )
+        configure_metrics(StreamMetricsAdapter(metrics))
+    else:
+        metrics = MetricsWrapper(
+            environment.metrics,
+            "consumer",
+            tags={
+                "group": consumer_group,
+                "storage": "_".join([storage_keys[0].value, "m"]),
```
lynnagara, line 160: What is this "m" about?

snuba/cli/consumer.py
```diff
@@ -139,21 +140,33 @@ def consumer(
...
+        getattr(StorageKey, storage_name.upper()) for storage_name in storage_names
+    ]
 
-    metrics = MetricsWrapper(
-        environment.metrics,
-        "consumer",
-        tags={"group": consumer_group, "storage": storage_key.value},
-    )
-    configure_metrics(StreamMetricsAdapter(metrics))
+    if len(storage_keys) == 1:
```
lynnagara, line 147: I don't think you need this if/else statement. Why don't you join all the storages together?

snuba/consumers/consumer_builder.py
```diff
@@ -83,62 +83,98 @@ def __init__(
...
             .get_default_topic_spec()
             .topic
-        )
+            for storage in self.storages.values()
+        }
+        topic = topics.pop()
+        if topics:
+            raise ValueError("only one topic is supported")
+
+        if len(self.storages) == 1:
```
lynnagara, line 102: I think the goal would be to ideally share all of this implementation, not move a ton of if/else statements into this function

snuba/consumers/consumer_builder.py
```diff
@@ -83,62 +83,98 @@ def __init__(
...
 
-        stream_loader = self.storage.get_table_writer().get_stream_loader()
+            # XXX: This can result in a producer being built in cases where it's
+            # not actually required.
+            self.producer = Producer(self.producer_broker_config)
 
-        self.raw_topic: Topic
+        storage_keys = [*self.storages.keys()]
+
+        self.raw_topic: ArroyoTopic
```
lynnagara, line 125: Why calling this ArroyoTopic?

snuba/consumers/consumer_builder.py
```diff
@@ -221,7 +261,11 @@ def __build_consumer(
     def __build_streaming_strategy_factory(
         self,
     ) -> ProcessingStrategyFactory[KafkaPayload]:
-        table_writer = self.storage.get_table_writer()
+        # this assumes single storage for now
```
lynnagara, line 264: This function is quite different currently between the two implementations. What do you think about leaving it unchanged for now and building an alternativve `__build_streaming_multistorage_strategy_factory` alongside it?

snuba/cli/consumer.py
ayirr7: I'm not completely sure, I got it from: https://github.com/getsentry/snuba/blob/master/snuba/cli/multistorage_consumer.py#L235-L242.

snuba/cli/consumer.py
ayirr7: That could be possible. The reason why I didn't join them all was because of the code snippet from multistorage_consumer.py (https://github.com/getsentry/snuba/blob/master/snuba/cli/multistorage_consumer.py#L235-L242) that shows how the multistorage case would create a MetricsWrapper

snuba/consumers/consumer_builder.py
ayirr7: There's a mix of Topic object types in this file. Some are Snuba topics, some are Arroyo topics, and some functions can only take in Snuba Topics and some Arroyo Topics. I figured that it would be helpful to distinguish when we're initializing an ArroyoTopic so that we don't mistake the two/pass one in when we meant the other.

snuba/consumers/consumer_builder.py
ayirr7: I think that's a good idea. It would probably be one of my next steps.

snuba/cli/consumer.py
fpacifici: We have alerts and consumer dashboards that are based on the storage tag. That's probably a bad idea, using the consumer group could be better.
That breaks with the multistorage consumer as there are multiple storages. That `m` was to tell apart the multistorage ones in our metrics.

fpacifici (commented):
Have you considered going the other way and fold the single storage consumer into the multistorage consumer instead of adapting the consumer builder?
A multi-storage consumer is also a single storage consumer (if you can write in two storages you can also write in one). Maybe that one would be quicker to adapt. 

There are, though, some features that were built in the consumer builder that were not ported  (read: I did not port) to the multi storage consumer. But it should not be much.

PR #3155, opened 2022-09-19
fix migrations for clickhouse 21.9+, 0014_backfill_errors, error: snu…

Slach:
fix migrations for clickhouse 21.9+, `0014_backfill_errors`, error: snuba.clickhouse.errors.ClickhouseError: DB::Exception: Illegal column Nullable(FixedString(32)) of first argument of function toString, 
look https://github.com/ClickHouse/ClickHouse/issues/41490 for details

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #3302, opened 2022-10-26
Update 0014_backfill_errors.py

turbotankist:
the same as in PR
https://github.com/getsentry/snuba/pull/3155
second script require NotNull too
I've run this migration on clickhouse 22.6 and it works

### Legal Boilerplate

### DIFF
--- a/snuba/snuba_migrations/events/0014_backfill_errors.py
+++ b/snuba/snuba_migrations/events/0014_backfill_errors.py
@@ -98,7 +98,7 @@ COLUMNS = [
     ("group_id", "group_id"),
     (
         "primary_hash",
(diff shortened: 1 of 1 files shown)


evanh:
What's the motivation for this change? The PR you referenced was later reverted.

turbotankist:
> What's the motivation for this change? The PR you referenced was later reverted.

The migration doesn't work with clickhouse version 22+
I don't know if can this work with old clickhouse versions but migrations definitely won't work without it with a new version.
PR was reverted but the changes is in [master](https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/events/0014_backfill_errors.py#L64) now

lynnagara (approved):
This is ok to me. Since `type != 'transaction'` is in the prewhere clause and transactions were the only event types without primary_hash, this should be safe.

PR #3305, opened 2022-10-26
fix(consumer): Remove shared producers in consumer

ayirr7:
This PR:

- Removes the shared default producer code in the consumer
- Instead, creates a replacements-specific producer and a commit log-specific producer in the respective locations 
- Having dedicated producers is also more similar to how the Multistorage Consumer creates producers (in `snuba/cli/multistorage_consumer.py`). 

This change is not associated with any particular JIRA ticket, but was something I noticed while working on https://github.com/getsentry/snuba/pull/3301.

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -188,3 +188,5 @@ def consumer(
     signal.signal(signal.SIGTERM, handler)
 
     consumer.run()
+    consumer_builder.commit_log_producer.close()
+    consumer_builder.replacements_producer.close()
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -188,5 +188,4 @@ def consumer(
     signal.signal(signal.SIGTERM, handler)
 
     consumer.run()
-    consumer_builder.commit_log_producer.close()
-    consumer_builder.replacements_producer.close()
+    consumer_builder.close_producers()
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -187,4 +187,5 @@ def consumer(
     signal.signal(signal.SIGINT, handler)
     signal.signal(signal.SIGTERM, handler)
 
-    consumer_builder.close(consumer)
+    consumer.run()
+    consumer_builder.flush()
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -187,5 +187,4 @@ def consumer(
     signal.signal(signal.SIGINT, handler)
     signal.signal(signal.SIGTERM, handler)
 
-    consumer.run()
-    consumer_builder.close_producers()
+    consumer_builder.close(consumer)
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -86,6 +86,9 @@ class ConsumerBuilder:
         self.storage = get_writable_storage(storage_key)
         self.bootstrap_servers = kafka_params.bootstrap_servers
         self.consumer_group = kafka_params.group_id
+        self.replacements_producer = None
+        self.commit_log_producer = None
+
         topic = (
             self.storage.get_table_writer()
             .get_stream_loader()
@@ -238,11 +241,7 @@ class ConsumerBuilder:
             collector=build_batch_writer(
                 table_writer,
                 metrics=self.metrics,
-                replacements_producer=(
-                    self.replacements_producer
-                    if self.replacements_topic is not None
-                    else None
-                ),
+                replacements_producer=self.replacements_producer,
                 replacements_topic=self.replacements_topic,
             )
             if self.__mock_parameters is None
@@ -270,6 +269,13 @@ class ConsumerBuilder:
             )
 
         return strategy_factory
+
+    def close_producers(self) -> None:
+        if self.commit_log_producer is not None:
+            self.commit_log_producer.close()
+
+        if self.replacements_producer is not None:
+            self.replacements_producer.close()
 
     def build_base_consumer(self) -> StreamProcessor[KafkaPayload]:
         """
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -274,14 +274,18 @@ class ConsumerBuilder:
         return strategy_factory
 
     def close(self, consumer: StreamProcessor[KafkaPayload]) -> None:
-        if not self.commit_log_producer and not self.replacements_producer:
+        if self.commit_log_producer is None and self.replacements_producer is None:
             consumer.run()
 
-        elif self.commit_log_producer and not self.replacements_producer:
+        elif (
+            self.commit_log_producer is not None and self.replacements_producer is None
+        ):
             with closing(self.commit_log_producer):
                 consumer.run()
 
-        elif self.replacements_producer and not self.commit_log_producer:
+        elif (
+            self.replacements_producer is not None and self.commit_log_producer is None
+        ):
             with closing(self.replacements_producer):
                 consumer.run()
 
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -134,12 +134,7 @@ class ConsumerBuilder:
             if commit_log_topic_spec is not None:
                 self.commit_log_topic = Topic(commit_log_topic_spec.topic_name)
                 self.commit_log_producer = Producer(
-                    build_kafka_producer_configuration(commit_log_topic_spec.topic),
-                    bootstrap_servers=kafka_params.bootstrap_servers,
-                    override_params={
-                        "partitioner": "consistent",
-                        "message.max.bytes": 50000000,  # 50MB, default is 1MB)
-                    },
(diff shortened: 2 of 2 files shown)


snuba/consumers/consumer_builder.py
```diff
@@ -97,14 +97,6 @@ def __init__(
             topic, bootstrap_servers=kafka_params.bootstrap_servers
         )
         logger.info(f"librdkafka log level: {self.broker_config.get('log_level', 6)}")
-        self.producer_broker_config = build_kafka_producer_configuration(
-            topic,
-            bootstrap_servers=kafka_params.bootstrap_servers,
-            override_params={
-                "partitioner": "consistent",
-                "message.max.bytes": 50000000,  # 50MB, default is 1MB
-            },
```
lynnagara, line 106: How come you dropped these parameters? Do we not need them?

lynnagara (commented):
This is good, thanks. Can you make sure to close both producers?

snuba/consumers/consumer_builder.py
ayirr7: Right, good point. I wasn't too sure whether we wanted to keep them. Do you have any insight?

ayirr7: Added them back to replacements producer (multistorage consumer passes them in for replacements).

snuba/consumers/consumer_builder.py
```diff
@@ -131,15 +133,14 @@ def __init__(
             commit_log_topic_spec = stream_loader.get_commit_log_topic_spec()
             if commit_log_topic_spec is not None:
                 self.commit_log_topic = Topic(commit_log_topic_spec.topic_name)
+                self.commit_log_producer = Producer(
+                    build_kafka_producer_configuration(commit_log_topic_spec.topic)
```
lynnagara, line 137: How come you didn't pass the bootstrap servers here?

snuba/consumers/consumer_builder.py
```diff
@@ -271,4 +274,7 @@ def build_base_consumer(self) -> StreamProcessor[KafkaPayload]:
         """
         Builds the consumer.
         """
-        return self.__build_consumer(self.__build_streaming_strategy_factory())
+        processor = self.__build_consumer(self.__build_streaming_strategy_factory())
+        self.commit_log_producer.close()
+        self.replacements_producer.close()
```
lynnagara, line 279: You can't close the producers immediately after building the consumer. Otherwise they won't be able to do any producing anymore. Only close them when the consumer is shutting down.

snuba/consumers/consumer_builder.py
```diff
@@ -237,7 +238,9 @@ def __build_streaming_strategy_factory(
                 table_writer,
                 metrics=self.metrics,
                 replacements_producer=(
-                    self.producer if self.replacements_topic is not None else None
+                    self.replacements_producer
+                    if self.replacements_topic is not None
+                    else None
```
lynnagara, line 243: It should be either a producer instance or None
```suggestion
                    self.replacements_producer
```

ayirr7: I think that's what this does: if self.replacements_topic is not None, self.replacements_producer is a valid instance and that gets passed to the field. If it is None, then None is passed to the field.

snuba/consumers/consumer_builder.py
ayirr7: Oops, good catch

snuba/consumers/consumer_builder.py
lynnagara: Functionally yes it is the same, but I think it's much more straightforward and less error prone to always define replacements_producer and drop this conditional here.

ayirr7: Ah ok, that makes sense.

snuba/cli/consumer.py
```diff
@@ -188,3 +188,5 @@ def handler(signum: int, frame: Any) -> None:
     signal.signal(signal.SIGTERM, handler)
 
     consumer.run()
+    consumer_builder.commit_log_producer.close()
+    consumer_builder.replacements_producer.close()
```
lynnagara, line 192: These are internal details of the consumer builder. Why don't you add a public close method there that in turn calls close on the producers?

snuba/consumers/consumer_builder.py
```diff
@@ -267,6 +270,13 @@ def __build_streaming_strategy_factory(
 
         return strategy_factory
 
+    def close_producers(self) -> None:
```
lynnagara, line 273: Let's just call this `close()`? I think the fact that it is closing producers is an internal detail of the consumer builder.

lynnagara:
Were you able to run this locally and verify that things worked and producers got closed correctly, etc?

snuba/consumers/consumer_builder.py
```diff
@@ -86,6 +86,9 @@ def __init__(
         self.storage = get_writable_storage(storage_key)
         self.bootstrap_servers = kafka_params.bootstrap_servers
         self.consumer_group = kafka_params.group_id
+        self.replacements_producer = None
+        self.commit_log_producer = None
```
lynnagara, line 90: A bit of a nitpick here but I would probably set this in the else statement where `replacements_topic` and `commit_log_topic` get set to None just to mirror how those variables get set

ayirr7: Sure, that makes sense to me

snuba/consumers/consumer_builder.py
```diff
@@ -270,12 +273,22 @@ def __build_streaming_strategy_factory(
 
         return strategy_factory
 
-    def close_producers(self) -> None:
-        if self.commit_log_producer is not None:
-            self.commit_log_producer.close()
+    def close(self, consumer: StreamProcessor[KafkaPayload]) -> None:
+        if not self.commit_log_producer and not self.replacements_producer:
+            consumer.run()
```
lynnagara, line 278: Wait... why did this move in here? Now your close function is not just closing, it's actually responsible for running the consumer too. That doesn't seem right.

ayirr7: This will change now since we realized that this type of Producer can't be closed. I had it like this originally because the usages of `closing()` I found in the codebase all did it like something similar to this:

```if dead_letter_producer:
        with closing(dead_letter_producer):
            processor.run()
    else:
        processor.run()```

fpacifici: I think @ayirr7  is doing this because, after creating the consumer she does not have access to the producers anymore from the code that calls the consumer builder and the consumer builder is gone. 

This highlights a larger problem we have with resources life cycle in this case and in Arroyo in general. 
- Arroyo is not strongly opinionated. It allows you to pass resources in (like producers) to factories or consumers and retain the ownership but it also allows you to give the ownership entirely to the consumer. The processing factory creates and destroys what it needs. The second approach is troublesome because it makes unit tests much harder so should be avoided. 
- The consumer builder is not built with resource management in mind. Cannot remember if this was intentional or a mistake, but definitively it does not work with Arroyo resources management. 

It seems that the consumer builder would work in a world where it creates the resources (the producer) pass the ownership to the consumer and let the consumer figure out when to close them. This is what we generally avoid in Arroyo.

I think the options you are are these (they are all bad):
- Make the consumer builder return a new class that retains ownership of the consumer and of the producers. This class has a run method and a close method. When you call close it closes the producers. So you preserve the lifecycle of the builder, which is not supposed to survive after it has built the consumer (which is cumbersome and full of boiler plate)
- You make the consumer builder return both producers and consumer (which is gross). Please let's avoid.
- We get rid of the consumer builder entirely. But I feel that is slightly beyond the scope of this PR.
- You instantiate the producer in the consumer builder, pass them to the consumer and the batch writer and rely on them to close the producer. This is confusing as it splits the life cycle so hard to verify, though possible with the `KafkaConsumerWithCommitLog`. Not sure about the batch writer.

I would avoid making the consumer instantiate its own producers as it makes it untestable.

lynnagara: @fpacifici I think you're referring to a somewhat related but different problem. To clarify I think this was fine before the latest change in https://github.com/getsentry/snuba/pull/3305/commits/392db09886d3bf768fe4f52be65f01f9a0a6b6e5 which started passing a consumer in here. I think we should just revert that last part and call it a day.

ayirr7:
Updates: We realized that the Confluent Kafka Producer can't be closed, so we removed that functionality. We are using `flush()` on the Producer instances now.

lynnagara (approved):
Looks good, thanks!

snuba/consumers/consumer_builder.py
fpacifici: Could we at least call that method `run_and_close` ? A `close` method that also runs all the business logic would be counter intuitive.

fpacifici:
You do not need to `flush` the producer used by the `KafkaConsumerWithCommitLog`. `KafkaConsumerWithCommitLog` does it for you when it is closed.

snuba/consumers/consumer_builder.py
lynnagara: Running went back to it's original place now.

PR #3307, opened 2022-10-27
Revert "ci(clickhouse): Enable running tests against clickhouse version 21.8"

evanh:
Reverts getsentry/snuba#3292

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,14 +226,8 @@ jobs:
     timeout-minutes: 25
     strategy:
       matrix:
-        snuba_settings: ["test"]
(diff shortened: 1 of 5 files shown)


MeredithAnya (approved):

MeredithAnya:
actually @evanh I think the changes in https://github.com/getsentry/snuba/pull/3304 also would need to reverted with this change or at least the `CLICKHOUSE_IMAGE` changes

evanh:
> actually @evanh I think the changes in #3304 also would need to reverted with this change or at least the `CLICKHOUSE_IMAGE` changes

Hmm, OK I'll revert that one too.

PR #3309, opened 2022-10-27
evanh/fix/revert ci changes

evanh:
One or both of these changes are blocking master from deploying. Revert both to
unblock deploys and then we can figure out what can be safely added back.

- Revert "use clickhouse image var in makefile and fix migration test cluster bug"
- Revert "ci(clickhouse): Enable running tests against clickhouse version 21.8 (#3292)"

### DIFF
--- a/snuba/settings/settings_test_distributed_migrations.py
+++ b/snuba/settings/settings_test_distributed_migrations.py
@@ -26,7 +26,7 @@ CLUSTERS: Sequence[Mapping[str, Any]] = [
         "storage_sets": {
             "migrations",
         },
-        "single_node": False,
+        "single_node": True,
         "cluster_name": "migrations_cluster",
         "distributed_cluster_name": "query_cluster",
     },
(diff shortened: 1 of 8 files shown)


snuba/settings/settings_test_distributed_migrations.py
```diff
@@ -26,7 +26,7 @@
         "storage_sets": {
             "migrations",
         },
-        "single_node": False,
+        "single_node": True,
```
MeredithAnya, line 29: I don't think you need to revert this part

MeredithAnya (approved):
Up to you if you want to take out the part that's unrelated to the CI changes

snuba/settings/settings_test_distributed_migrations.py
evanh: We can add it back in later, but for now we need to get master unblocked.

lynnagara (approved):

PR #3310, opened 2022-10-27
ref(entity subscriptions): Move entity subscriptions to entity

enochtangg:
## Context
This change was initially conceived when building the entity subscription loader for config. References to this can be found here: 
* https://github.com/getsentry/snuba/pull/3107#issuecomment-1234734703
* https://github.com/getsentry/snuba/pull/3107#discussion_r961068006

Currently, `EntitySubscriptions` are embedded in a class. There is an EntitySubscription factory (separate from the Entity factory) which loads the mappings between an `EntityKey` and `EntitySubscriptions` class. Since these mappings are 1:1, it is possible `EntitySubscriptions` to be directly associated with the `Entity` itself which removes the need for the factory.

## Overview
This PR is responsible for moving the EntitySubscriptions into the Entity class. To accomplish this, there are a few components were changed:

### Refactored EntitySubscription base class

As a proposed in [https://github.com/getsentry/snuba/pull/3278](https://github.com/getsentry/snuba/pull/3278), a single `BaseEntitySubscription` was created to replace all entity subscriptions. In the current state, `SessionsSubscription` is inherited by a subset of `EntitySubscription` child classes. These classes contain an attribute which does not exist in `EventSubscription` and `TransactionSubscription`([code snippet](https://github.com/getsentry/snuba/blob/11cd169ba0a5e1a808f975b63d156f5a2643854c/snuba/datasets/entity_subscriptions/entity_subscription.py#L92-L129)).

This PR removes all entity subscription classes and replaces them with a single `EntitySubscription` class which supports all interfaces. To accomplish this, methods were refactored and class attributes where changed to instance attributes to preserve the unique nature of each individual entity subscription.

### A entity_subscription attribute added to Entity

This `entity_subscription: Optional[EntitySubscription]` attribute was added to the entity class. All python defined entities were updated to include this object. On the other hand, the entities loaded from config were updated to specify subscriptions with the following JSONschema:
```
version: v1
kind: entity
name: generic_metrics_sets

...
required_time_column: timestamp
partition_key_column_name: org_id
...
entity_subscriptions:
  max_allowed_aggregations: 3
  disallowed_aggregations: ["having", "orderby"]
```

The entity_loader was updated to parse and create the entity_subscription object and set it in the PluggableEntity constructor. When the subscription field is not specified in YAML, the entity_subscription is set to None.

### Removing organization_id from EntitySubscription
The snuba data model upper layer (dataset, entity, storage, query processo) is designed to be static and immutable. However, there is an `organization: int` attribute in EntitySubscriptions that violates this. This PR removes this query specific attribute from `EntitySubscriptions` and puts it on `SubscriptionData`. That way, entity subscriptions are only instantiated once when the entity is created and remain unchanged regardless of the query.

### Testing

To test this, all test which used the old entity subscriptions factory were updated to reference the entity subscription from entity object itself which is attained by the `Entity` factory. Additionally, the `SubscriptionData` tests were updated and checked to make sure that the subscriptions itself remained unchanged. Finally, all entity subscriptions factory related code was removed.


volokluev:
@lynnagara would definitely want to look at this

snuba/datasets/configuration/json_schema.py
```diff
@@ -444,6 +444,23 @@ def registered_class_schema(
...
+            "description": "Specifies whether entity has subscriptions enabled",
+            "properties": {
+                "max_allowed_aggregations": {
+                    "type": "integer",
+                    "description": "The max number of allowed of aggregations in subscription",
+                },
+                "disallowed_aggregations": {
+                    "type": "array",
+                    "items": TYPE_STRING,
+                    "description": "The disallowed aggregation clauses in subscription query",
```
volokluev, line 458:
```suggestion
                    "description": "The disallowed aggregation clauses in subscription query. e.g. [\"having\", \"orderby\"]",
```

snuba/datasets/entity.py
```diff
@@ -156,3 +175,106 @@ def describe(self) -> Description:
...
+                    Column(None, None, "org_id"),
+                    Literal(None, self.organization),
+                ),
+            ]
+        return []
+
+    def to_dict(self) -> Mapping[str, Any]:
+        if self.organization:
+            return {"organization": self.organization}
+        return {}
```
volokluev: Why does this need to be 3 classes when the inheritance tree is just a straight line?

volokluev: Also `Base*` by convention implies a base class that is to be extended. Is this one of those cases or is it just an `EntitySubscription`?

snuba/datasets/entity.py
```diff
@@ -156,3 +175,106 @@ def describe(self) -> Description:
...
+            super().__init__(max_allowed_aggregations, disallowed_aggregations)
+        self.set_org(data_dict)
+
+    def set_org(self, data_dict: Mapping[str, Any] = {}) -> BaseEntitySubscription:
+        if "organization" in data_dict:
+            try:
+                self.organization: int = data_dict["organization"]
+            except KeyError:
+                raise InvalidQueryException(
+                    "organization param is required for any query over sessions entity"
```
volokluev, line 264: This is kind of weird. why are we talking about a specific product entity in the error message? I think this whole class could use a docstring which describes:

1. the function it accomplishes and how it fits into the larger system
2. usage examples not tied to any specific entity

enochtangg: Right, this was copy pasta from the original EntitySubscription (which all extended from SessionsSubscription). Will update this. Sounds good, will add a descriptive docstring with examples.

snuba/datasets/entity.py
enochtangg: Updated back to the original structure. The idea behind the base class is so that we can extend the functionality of subscriptions the same way we extend things like translation mappers. Whether or not that should extend from `EntitySubscription` or the newly created `BaseEntitySubscription` is still to be determined.

snuba/datasets/entity.py
```diff
@@ -156,3 +174,129 @@ def describe(self) -> Description:
...
+        entity = get_entity(from_clause.key)
+
+        SubscriptionAllowedClausesValidator(
+            self.max_allowed_aggregations, self.disallowed_aggregations
+        ).validate(query)
+        if entity.required_time_column:
+            NoTimeBasedConditionValidator(entity.required_time_column).validate(query)
+
+
+class BaseEntitySubscription(EntitySubscriptionValidation, EntitySubscription):
```
evanh, line 255: I think you should rename this class to `EntitySubscription` and collapse the other two classes into it. We don't need to extend this, and having the validation be separate isn't useful.

I also think all this code shouldn't be in the entity file. I am actually OK with it being in the `entity_subscription` folder, unless that causes import cycles.

snuba/datasets/entity.py
```diff
@@ -156,3 +174,129 @@ def describe(self) -> Description:
...
+        max_allowed_aggregations: Optional[int] = None,
+        disallowed_aggregations: Optional[Sequence[str]] = None,
+        data_dict: Mapping[str, Any] = {},
+    ) -> None:
+        super().__init__()
+        if max_allowed_aggregations and disallowed_aggregations:
+            super().__init__(max_allowed_aggregations, disallowed_aggregations)
+        self.set_org(data_dict)
+
+    def set_org(self, data_dict: Mapping[str, Any] = {}) -> BaseEntitySubscription:
```
evanh, line 278: If this is going to `set_org` it should take an organization ID as a parameter. Parsing the organization out of the data can happen where this function is called.

snuba/subscriptions/data.py
```diff
@@ -172,7 +167,13 @@ def build_request(
...
         cls, data: Mapping[str, Any], entity_key: EntityKey
     ) -> SubscriptionData:
-        entity_subscription = get_entity_subscription(entity_key)(data_dict=data)
+
+        entity_subscription = get_entity(entity_key).get_entity_subscription()
+        if not entity_subscription:
+            raise InvalidSubscriptionError(
+                f"entity subscription for {entity_key} does not exist"
+            )
+        entity_subscription = entity_subscription.set_org(data_dict=data)
```
evanh, line 176: The org should be parsed out of the data here, instead of inside the `set_org` function.

snuba/datasets/entity.py
```diff
@@ -156,3 +174,129 @@ def describe(self) -> Description:
...
+        Setting self.organization is optional because not every entity requires this field.
+        Organization is set after Entity and EntitySubscription objects are created using set_org().
+        Child classes can extend from this to provide additional entity specific functionality.
+    """
+
+    def __init__(
+        self,
+        max_allowed_aggregations: Optional[int] = None,
+        disallowed_aggregations: Optional[Sequence[str]] = None,
+        data_dict: Mapping[str, Any] = {},
```
lynnagara, line 271: Same here re: mutable default args

lynnagara (commented):
In general I like the direction this is going into. My main feedback is that EntitySubscription should be defined once and not mutated later via the `set_org` mechanism. A few other comments inline.

snuba/datasets/entity.py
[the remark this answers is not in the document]
enochtangg: The existing way EntitySubscriptions work is through an EntitySubscription factory which maps an `EntityKey` to a Type[EntitySubscription]`. Therefore, the object is not created until the `SubscriptionData` is created later on and the system knows which query it wants to run. Since we are consolidating all EntitySubscription into a single class, we can't use class attributes and need to move to instance attributes because they are supposedly "different" for each entity. So with this change, when the entity is first created, it creates the `EntitySubscription` object with the appropriate parameters (max_allowed_aggregations, disallowed_aggregations). However, it isn't until later when the subscription is created that the system knows the org_id.

snuba/datasets/entity.py
```diff
@@ -156,3 +174,129 @@ def describe(self) -> Description:
...
+    def set_org(self, data_dict: Mapping[str, Any] = {}) -> EntitySubscription:
+        raise NotImplementedError
+
+
+class EntitySubscriptionValidation:
+    def __init__(
+        self,
+        max_allowed_aggregations: int = 1,
+        disallowed_aggregations: Sequence[str] = ["groupby", "having", "orderby"],
+        organization: Optional[int] = None,
```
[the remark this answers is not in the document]
enochtangg: Would it make more sense for `org_id` to be part of `SubscriptionData` instead? https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/data.py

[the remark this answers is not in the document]
lynnagara: I think so. See how `offset` is passed to the `get_entity_subscription_conditions_for_snql` method. Well, we are never using that to build the snql query anymore (it used to be until recently) but that was essentially solving the same problem of needing query parameters to build the snql query. You may want to consider a similar approach.

snuba/datasets/entity.py
enochtangg: Moved files back to original directories as suggested.

snuba/datasets/entity.py
enochtangg: Collapsed classes and renamed.

snuba/datasets/entity.py
```diff
@@ -156,3 +174,129 @@ def describe(self) -> Description:
...
+    @abstractmethod
+    def set_org(self, data_dict: Mapping[str, Any] = {}) -> EntitySubscription:
+        raise NotImplementedError
+
+
+class EntitySubscriptionValidation:
+    def __init__(
+        self,
+        max_allowed_aggregations: int = 1,
+        disallowed_aggregations: Sequence[str] = ["groupby", "having", "orderby"],
```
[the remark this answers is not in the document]
enochtangg: Fixed

snuba/datasets/configuration/transactions/entities/transactions.yaml
```diff
@@ -347,3 +347,6 @@ validators:
   ]
 
 required_time_column: finish_ts
+subscriptions:
```
lynnagara, line 350: How is `entity_subscription=None` supposed to be represented here? Does the absence of `subscriptions` mean none?

Also, please align all the names in the yaml and python versions. In python you called it `entity_subscription`. Here the key is `subscriptions`. It is much clearer for the user/reader if all of these names which essentially represent the same thing line up

lynnagara (changes requested):
I had a closer look and I think the changes to EntitySubscription in particular are too risky. What this change is doing is instantiating a single EntitySubscription object and reusing it over and over again for all subscriptions instead of building it from scratch for every query. The function `set_org` has been added so for every query that comes in we can just update the new org on the same object that was used to run the query for other orgs. This feels like a bit of a hack to me. Individual fields (like organization) are not really supposed to be part of the entity subscription interface. Aside from design aspects though the main concern I have here is that we might have multiple copies of the subscription data at the same time and now we are constantly mutating those objects to replace the org, setting all the orgs to the wrong one. The risk of building a subscription query and returning data for the wrong org because they are all sharing a single object which contains org/query level data is too high.

snuba/datasets/configuration/transactions/entities/transactions.yaml
enochtangg: Yes, the absence of `subscription` means there is no entity_subscription set which means this entity does not use subscriptions. This idea of "define what you need" is true for the way we define datasets as config right now. For example, a user can leave out `writeable_storage` which implies that `writeable_storage=None`

Sure, I can update the name to `entity_subscription` align with python definition.

fpacifici:
> we might have multiple copies of the subscription data at the same time and now we are constantly mutating those objects to replace the org

Good catch. A small correction: we are guaranteed to have multiple queries being processed concurrently.
This is the code that starts the processing the queries. It schedules on a thread pool, so they are executed concurrently.
https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/executor_consumer.py#L348-L353

fpacifici:
Adding some additional context to my previous comment around the mutability of the data model objects as, thinking over that, this is not the first attempt I saw recently at making some elements mutable. 

Simply put: the Snuba dataset data model is designed to be immutable and the correctness of the query processing pipelines actively depends on this invariant. 

More in details:
This paragraph explains that the data model is divided in two layers https://github.com/getsentry/snuba/blob/master/snuba/datasets/dataset.py#L36-L40

The upper layer depends on the lower layer. The other way around is not allowed and would introduce circular dependencies. The upper layer defines the dataset schema. 
The upper layer includes: dataset, entity, storage, query processor, etc. Here immutability is a requirement. Beyond immutability, all these are designed to be static semantically so that they could be replaced by configuration easily. This last property implies that whether or not you have multiple instances of the same class does not matter as they would be identical (if you instantiate twice the same entity, they are identical).

The lower layer is query specific and can be referenced by a query object. This includes DataSource, SubscriptionEntity, etc. This part is supposed is supposed to still be immutable but it does not have to be static. You may instantiate different instances of the same SubscriptionEntities with query specific properties. So here the requirements are a bit less strict, but, if this class has query specific fields then it cannot be shared. Keeping them immutable is a good safeguard.

Regarding entity subscription. While this was designed to be part of the upper data model, thus static, it seems this was violated as it seems the instance carries the organization id. Which means this has to be instantiated brand new for every query or everything is broken.
https://github.com/getsentry/snuba/blob/master/snuba/datasets/entity_subscriptions/entity_subscription.py#L71
That makes it a lower level concern and it is critical a new instance with new init parameters is instantiated at every query.

So in general two rules of thumb:
- immutable, static -> always good.
- non static (query specific fields) -> has to be instantiated new with every query and must be in the lower level. It cannot depend on the upper level data model. If it does you get circular dependencies.
- mutable -> please stay away, either you are really precise at guaranteeing a new instance is created for each queries or you will have subtle failures as concurrent query processing through threads is quite common.

enochtangg:
@lynnagara  @fpacifici  Knowing this requirement and the nature of concurrent queries, it sounds like this PR is a good place to also fix this design violation (will update PR description). I think `organization_id` should be moved out of `EntitySubscription` and placed on the `SubscriptionData` instead (context: `SubscriptionData` is what calls `EntitySubscription` methods). 

Moving forward, how does making `EntitySubscription` instantiated once when the `Entity` is created (immutable, static) and moving query specific params outside of `EntitySubscription` and only passed as args into its methods?

The `EntitySubscription` object itself can be used over and over (like writable storage object) as long as there is no query specific fields part of its interface

evanh (commented):
I feel like the ideal solution is:

The `EntitySubscription` class, if it even exists, is simply a container for subscription configuration. In other words, we could delete the class and have `entity.use_organization_in_subscriptions`, `entity.max_allowed_aggregations`...

Then in the subscriptions code, the subscription would fetch the relevant entity, and use the settings to build the appropriate conditions. All the logic about building the conditions can be stand alone functions or part of the subscriptions code, instead of being part of the EntitySubscription class.

snuba/subscriptions/data.py
```diff
@@ -172,14 +171,20 @@ def build_request(
     def from_dict(
         cls, data: Mapping[str, Any], entity_key: EntityKey
     ) -> SubscriptionData:
-        entity_subscription = get_entity_subscription(entity_key)(data_dict=data)
+
+        entity_subscription = get_entity(entity_key).get_entity_subscription()
+        if not entity_subscription:
```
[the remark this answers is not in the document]
enochtangg: All datasets with subscriptions enabled has an entity subscription set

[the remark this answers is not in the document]
evanh: OK cool, just wanted to be sure.

snuba/datasets/entity_subscriptions/entity_subscription.py
```diff
@@ -18,112 +17,49 @@ class InvalidSubscriptionError(Exception):
...
+        disallowed_aggregations: Sequence[str] = ["groupby", "having", "orderby"],
+    ) -> None:
+        self.max_allowed_aggregations = max_allowed_aggregations
+        self.disallowed_aggregations = disallowed_aggregations
 
-class EntitySubscriptionValidation:
-    MAX_ALLOWED_AGGREGATIONS: int = 1
-    disallowed_aggregations: Sequence[str] = ["groupby", "having", "orderby"]
+    def validate_query(self, query: Union[CompositeQuery[EntityDS], Query]) -> None:
+        # Import get_entity() when used, not at import time to avoid circular imports
```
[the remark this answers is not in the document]
enochtangg: This import cycle is not caused by type checking, but from the nature of how entity factory is set up. Entity factory requires this `entity_subscription.py` module which has a specific function that looks up an entity. However, this is done in validation so it's guaranteed that entities are already created beforehand.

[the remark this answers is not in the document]
evanh: This feels like our code is structured incorrectly. Do you think it would be difficult to move the validation functions out of the `EntitySubscription` class, and make them functions in the subscriptions module?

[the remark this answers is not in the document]
enochtangg: I don't think it would be difficult to do that given how this validate query method is only called once in the subscriptions module. Is this something we want to commit to? Personally, I don't think there is a significant difference given how entity subscription is just a container for these methods used by the subscriptions module. However, if in the future we want to support entity specific validation on a query, then this might useful to keep?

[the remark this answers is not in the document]
lynnagara: Agree this is a sign that the code is not structured correctly. If EntitySubscription is a property of an entity, it doesn't make sense for it to also require the entity. Even though you found a way around the circular import issue, this is a huge hack.

lynnagara:
This code feels like it's doing a lot:
- refactoring entity subscription class and changing it's purpose
- moving files around
- updating all the python and yaml entities to move the entity subscription definitions in there
- merging all the entity subscription classes into one
- adding and changing tests

I think this makes it harder to review thoroughly and not miss something. I think you should split out independent PRs for the separate tasks, especially the parts that affect the query pipeline e.g. some of the refactoring of EntitySubscription since it largely doesn't depend on the rest of the changes.

PR #3311, opened 2022-10-27
Explicitly initialize snuba

volokluev:
## Blast Radius

This affects every snuba deployment as it fundamentally changes how the snuba cli commands are invoked.

## Before State
Every CLI command would be imported no matter what cli command was called. This would make initialization time go up with each command that was added. All datasets and their sub-components would be loaded lazily,

## After State 
The CLI still discovers the files in this directory as before but does not actually run their code until they are invoked. Thus `snuba api` would only ever import the `snuba/cli/api.py` module.

A new module is added `snuba/core/initialize.py` which explicitly initializes the storages, entities and datasets. Any other initialization steps that apply to all commands should go in there from now on. 

## Testing Notes
I ran the `snuba <cmd>` for every subcommand in the CLI directory and made sure it still works. Also made sure to test the startup speed:

master:
```
❯ time snuba api --help
2022-10-25 09:24:06,784 Initializing Snuba...
2022-10-25 09:24:06,786 Loading module snuba.cli.admin
2022-10-25 09:24:06,789 Loading module snuba.cli.api
2022-10-25 09:24:06,886 Loading module snuba.cli.bootstrap
2022-10-25 09:24:07,346 Loading module snuba.cli.bulk_load
2022-10-25 09:24:07,961 Loading module snuba.cli.cleanup
2022-10-25 09:24:07,962 Loading module snuba.cli.config
2022-10-25 09:24:07,964 Loading module snuba.cli.consumer
2022-10-25 09:24:07,980 Loading module snuba.cli.devserver
2022-10-25 09:24:07,981 Loading module snuba.cli.entities
2022-10-25 09:24:08,335 Loading module snuba.cli.migrations
2022-10-25 09:24:08,337 Loading module snuba.cli.multistorage_consumer
2022-10-25 09:24:08,339 Loading module snuba.cli.offline_replacer
2022-10-25 09:24:08,346 Loading module snuba.cli.optimize
2022-10-25 09:24:08,352 Loading module snuba.cli.replacer
2022-10-25 09:24:08,354 Loading module snuba.cli.subscriptions_executor
2022-10-25 09:24:09,569 Loading module snuba.cli.subscriptions_scheduler
2022-10-25 09:24:09,576 Loading module snuba.cli.subscriptions_scheduler_executor
2022-10-25 09:24:09,579 Loading module snuba.cli.test_consumer
2022-10-25 09:24:09,581 Snuba initialization took 2.796583355s
Usage: snuba api [OPTIONS]

Options:
  --bind TEXT          Address to listen on.
  --debug
  --log-level TEXT     Logging level to use.
  --processes INTEGER
  --threads INTEGER
  --help               Show this message and exit.
snuba api --help  2.60s user 0.42s system 85% cpu 3.560 total
```

current branch:
```
❯ time snuba api --help
2022-10-27 12:55:17,492 Initializing Snuba CLI...
2022-10-27 12:55:17,493 Snuba initialization took 0.0006562770000000384s
2022-10-27 12:55:17,494 Loading command api
2022-10-27 12:55:17,523 Initializing snuba
Usage: snuba api [OPTIONS]

Options:
  --bind TEXT          Address to listen on.
  --debug
  --log-level TEXT     Logging level to use.
  --processes INTEGER
  --threads INTEGER
  --help               Show this message and exit.
snuba api --help  1.41s user 0.19s system 97% cpu 1.647 total
```

^ we save about a second on initialization because we don't need to import every other module

### New tests added

A new test was added specifically for initialization in the `test_initialization` directory. It was put in a separate directory to avoice conflicting with conftest and other tests in the `tests` directory as testing initialization requires a fresh state.

### DIFF
--- a/snuba/cli/__init__.py
+++ b/snuba/cli/__init__.py
@@ -5,6 +5,7 @@ import os
 import time
 
 import click
+import sentry_sdk
 import structlog
 
 from snuba.environment import metrics as environment_metrics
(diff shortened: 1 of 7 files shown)


snuba/core/initialize.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+import logging
+
+logger = logging.getLogger("snuba_core_init")
+
+
+def _load_datasets() -> None:
+    from snuba.datasets.factory import get_enabled_dataset_names
+
+    get_enabled_dataset_names()
```
evanh, line 17: Is there a reason you're using this function instead of the `reset_dataset_factory` function?

snuba/cli/__init__.py
```diff
@@ -1,51 +1,68 @@
...
+                            O
+                            o
+        .oOo  'OoOo. O   o  OoOo. .oOoO'
+        `Ooo.  o   O o   O  O   o O   o
+            O  O   o O   o  o   O o   O
+        `OoO'  o   O `OoO'o `OoO' `OoO'o"""
+
+    init_time = time.perf_counter() - start
+    metrics.gauge("snuba_init", init_time)
+    logger.info(f"Snuba initialization took {init_time}s")
```
evanh, line 66: Is this still true? I don't think  `init_time` is accurate anymore.

snuba/core/initialize.py
volokluev: My understanding was that we would only use the `reset_dataset_factory` function for testing.

snuba/cli/__init__.py
volokluev: fixed

snuba/cli/__init__.py
```diff
@@ -1,51 +1,67 @@
...
-metrics = MetricsWrapper(environment_metrics, "cli")
+with sentry_sdk.start_transaction(
+    op="snuba_init", name="Snuba CLI Initialization", sampled=True
+):
 
+    setup_logging("DEBUG")
+    logger = logging.getLogger("snuba_init")
 
-@click.group()
-@click.version_option()
```
MeredithAnya, line 23: im not too too familiar with click but do we not need these anymore?

volokluev: We don't need `group` because we have switched main to a `MultiCommand` but I can put the version option back in

snuba/core/initialize.py
```diff
@@ -0,0 +1,42 @@
...
+
+    get_all_entity_names()
+
+
+def initialize() -> None:
+    logger.info("Initializing snuba")
+
+    # The order of the functions matters. At time of writing (2022-10-27), snuba datasets are not guaranteed to load
+    # entities and storages. This is because datasets defined in code reference entities via EntityKey(s) and are
+    # thus loaded lazily
```
lynnagara, line 37: I would remove this. Or consider rewording. This comment doesn't help me understand what's actually happening here.

Specifically from reading this function it looks like the entire storage, entity and dataset objects are being loaded. But then when looking at the implementations of each of the functions it looks like only the "names" or "keys' are being loaded, e.g. get_all_entity_names, get_all_storage_keys, etc, which implies that we might just be loading a small list of names or something. Can you clarify what's happening exactly?

volokluev: You're right. The functions I was using to initialize the factories are confusing. I used their proper init methods now. Let me know if it still doesn't make sense

volokluev:
CI finally works. I have released my mortal coil

snuba/core/initialize.py
volokluev: nvm I'm using it now

lynnagara:
@volokluev What do you think about supporting an option to initialize the minimum required set of datasets/storages/entities upon startup instead of all of them every time?

Apart from the API which serves a bunch of datasets, most of our processes are only concerned with one of those entities. For example, consumers only need to care about the storage they are writing to, and shouldn't need to care about the higher level concepts or unrelated storages that they are not targeting.

How much longer will the initialization take?

Separately to load time, as we march closer to a world where the number of datasets is getting large and the ownership of that datasets is distributed (and potentially eventually outside of Snuba), it may help to avoid a bug in one dataset definition having the possibility of affecting others if we are never loading them at all.

snuba/core/initialize.py
evanh: I'm not sure the reason provided here makes sense. Fundamentally the order matters because entities reference storages, so the storages need to be built first, and datasets reference entities etc.

In other words, if we changed the code so that datasets didn't load entities lazily, we would still probably want to maintain the bottom up ordering of initialization. We also want to be more transparent about what is being initialized when, so just loading the dataset and having everything else be initialized implicitly is counter to the benefits of this change.

volokluev:
@lynnagara 

> What do you think about supporting an option to initialize the minimum required set of datasets/storages/entities upon startup instead of all of them every time?

I think that's a good idea and I'm down to file a task for it but I don't want to do it in this PR. This PR is already an improvement on load times for all applications and there is now an initialization step to snuba. Doing what you described would require more careful consideration which I don't have time for right now

snuba/core/initialize.py
volokluev: Updated the comment. Let me know if that makes sense

evanh (approved):

PR #3323, opened 2022-10-31
RFC: InitRecorder

volokluev:
## Why
This is possibly a necessary step to converting the datasets which are in code to configuration automatically

## Blast Radius
This metaclass would be on every Column, QueryProcessor, QuerySplitter, etc. It would be a no-op in any actual production environment and will be removed after the python -> yaml conversion is complete. 

## The problem

Many classes that make up a storage or entity are instantiated with arguments which are then used to drive its behavior. Those arguments are not guaranteed to be accessible after instantiation ([example](https://github.com/getsentry/snuba/blob/master/snuba/query/processors/physical/array_has_optimizer.py#L29)). For these classes to be declared in configuration, we need to know what arguments they were declared with in such a way that we can reliably access them.

## The (proposed) solution

Introduce a temporary metaclass into the classes that are part of the data model declaration. which will allow the inspection of the arguments they were called with.

### DIFF
--- a/snuba/settings/settings_test.py
+++ b/snuba/settings/settings_test.py
@@ -22,6 +22,11 @@ ENABLE_DEV_FEATURES = True
 # to explore the unrefined Expression structure
 PRETTY_FORMAT_EXPRESSIONS = True
 
+# For converting code storages/entities/datasets to yaml we need to record what
+# arguments certain classes were instantiated with. This has no purpose in a production
+# environment. Thus it is only enabled here
+ENABLE_INIT_RECORDER = True
+
 # override replacer threshold to write to redis every time a replacement message is consumed
 REPLACER_PROCESSING_TIMEOUT_THRESHOLD = 0  # ms
 
--- a/tests/datasets/storages/test_py_2_yaml.py
+++ b/tests/datasets/storages/test_py_2_yaml.py
@@ -1,0 +1,58 @@
+from abc import ABCMeta
+from copy import deepcopy
+from typing import Any, Dict, Tuple, Type
+
+from snuba import settings
+
+
+class InitRecorder(ABCMeta):
+    """Records the arguments passed to the __init__ function of classes
+    in an init_kwargs member. Only useful if said objects need to be converted
+    to declarative configuration
+
+    Handles positional arguments by inspecting the signature and turning them into kwargs
+    for ease of use
+    """
+
+    def __new__(cls, name: str, bases: Tuple[Type[Any]], dct: Dict[str, Any]) -> Any:
+        res = super().__new__(cls, name, bases, dct)
+        # this is not useful in production in any way, hence this is a noop unless explicitly enabled
+        if settings.ENABLE_INIT_RECORDER:
+            from inspect import signature
+
+            orig_init = getattr(res, "__init__")
+
+            def __init__(self, *args, **kwargs):
+                self.init_kwargs = deepcopy(kwargs)
+                if args:
+                    init_param_names = [p for p in signature(orig_init).parameters]
+                    i = 0
+                    while i < len(args):
+                        self.init_kwargs[init_param_names[i + 1]] = args[i]
+                        i += 1
+                return orig_init(self, *args, **kwargs)
+
+            setattr(res, "__init__", __init__)
+        return res
+
+
+def test_record_init_args():
+    class A(metaclass=InitRecorder):
+        def __init__(self, a, b):
+            self.a = a
+            self.b = b
+
+    instance = A(a="a", b="b")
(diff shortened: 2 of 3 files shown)


enochtangg:
Nice stuff! Had a two questions:
* Are there any classes which define attributes that are not part of its initialization? If so, do we care about them?
* Are there any methods which alter the attributes set by original arguments passed into the `__init__()` ?

snuba/settings/settings_test.py
```diff
@@ -22,6 +22,11 @@
 # to explore the unrefined Expression structure
 PRETTY_FORMAT_EXPRESSIONS = True
 
+# For converting code storages/entities/datasets to yaml we need to record what
+# arguments certain classes were instantiated with. This has no purpose in a production
+# environment. Thus it is only enabled here
+ENABLE_INIT_RECORDER = True
```
evanh, line 28: Does this even need to be enabled here? It's possible to override settings on the CLI (`ENABLE_INIT_RECORDER=1 snuba translate-dataset ...`).

evanh (commented):
Another way to do this would be to parse and inspect the code itself. That could be done entirely in a script, and wouldn't require changing any of the existing snuba code.

rahul-kumar-saini:
> Another way to do this would be to parse and inspect the code itself. That could be done entirely in a script, and wouldn't require changing any of the existing snuba code.

If you mean writing a parser to read the python files themselves it may be hard for classes that reference and import shared constants, eg the same list of columns

tests/datasets/storages/test_py_2_yaml.py
```diff
@@ -0,0 +1,58 @@
+from abc import ABCMeta
+from copy import deepcopy
+from typing import Any, Dict, Tuple, Type
+
+from snuba import settings
+
+
+class InitRecorder(ABCMeta):
```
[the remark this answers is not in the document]
volokluev: yes exactly

snuba/settings/settings_test.py
volokluev: TIL, I'll remove it from here then!

volokluev:
Another way to do this would be to parse and inspect the code itself. That could be done entirely in a script, and wouldn't require changing any of the existing snuba code.

@evanh Rahul is exactly on point here. The [errors storage](https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/errors.py) for example has a lot of shared constants being imported. And following those references seems like a lot of work that won't pay off for anything else

volokluev:
> Are there any classes which define attributes that are not part of its initialization? If so, do we care about them?

I believe all the classes we care about here are defined through initialization. It may come up that there's an edge case somewhere but I'll handle that when it's a problem.

> Are there any methods which alter the attributes set by original arguments passed into the __init__() ?

Can you clarify what you mean with an example? The init method will transform the arguments in some way but I don't think we care about that for configuration

evanh:
> following those references seems like a lot of work that won't pay off for anything else

I don't think this idea should be dismissed out of hand. All of this is work that won't pay off for anything else. This class is explicitly going to be deleted after we do the migration.

The upside of the code parsing is that it means we won't introduce potentially breaking changes in Snuba just to make it easier to run one-off scripts.

volokluev:
@evanh

> I don't think this idea should be dismissed out of hand

I don't want to dismiss it out of hand either. I just see it as an order of magnitude harder. If we don't plan to convert errors using the script (which has a lot of reference chasing). This does still complicate discover which defines its [query processors in a different way](https://github.com/getsentry/snuba/blob/master/snuba/datasets/entities/discover.py#L426-L435). Taking all that into account, we *could* do it with parsing the file. Even then however, we would have to parse it using the python ast and deal with a lot more complexity (as we are parsing the whole python file). With this method, we would always be dealing with the storage/entity object itself which takes care of that complexity for us.

I understand that this code has more risk (as it *could* be run at runtime). But it wouldn't be run in any environment except our local machines which de-risks it a lot.

Am I maybe thinking about the parsing approach incorrectly or overestimating its complexity?

evanh:
> Am I maybe thinking about the parsing approach incorrectly or overestimating its complexity?

I don't actually know the complexity, I've never had to work with the Python AST. I think you're right, it would certainly be more complex to parse the files. I wanted to balance that against the risk.

All that being said, I had an idea: what if we didn't actually check in the script changes to master? We could leave all those changes in a separate (permanent) branch. Since this is going to be run locally anyways, it doesn't actually need to be in master. That would de-risk the whole thing, and we could change what we needed in Snuba.

volokluev:
> All that being said, I had an idea: what if we didn't actually check in the script changes to master?

I think that's a good idea. Let's do that. Thanks Evan

PR #3324, opened 2022-10-31
ref (migrations): allow specifying order in migrations  SNS-1816

dbanda:
Allow migrations to specify if they want to execute the local ops before distributed ops for run and reverse. This is the first PR in a change in migrations to better handle the order of operations to avoid getting into an inconsistent state.

**Context**
There are certain SQL ops that must be applied to the distributed tables before local tables and vice versa. Right now, our migrations always run local first and this leads to errors.  By introducing the local_first flag, creaters of migrations can specify the correct order.  The follow up to this PR will include a best effort checker that will run test to check that the order specified is consistent with the SQL ops in the migration and will not cause errors.

**Blast Radius**
Migrations Base Class

**Before**
All migrations run local ops before distributed ops

**After** 
Run local ops first by default, but allow users to specify distributed ops first in their migrations.

### DIFF
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -113,6 +113,9 @@ class ClickhouseNodeMigration(Migration, ABC):
     completely unrelated, they are probably better as separate migrations.
     """
 
+    forwards_local_first: bool = True
+    backwards_local_first: bool = True
+
     @abstractmethod
     def forwards_local(self) -> Sequence[SqlOperation]:
         raise NotImplementedError
@@ -141,10 +144,19 @@ class ClickhouseNodeMigration(Migration, ABC):
         # so do not update status yet
         if not self.is_first_migration():
             update_status(Status.IN_PROGRESS)
-        for op in self.forwards_local():
-            op.execute(local=True)
-        for op in self.forwards_dist():
-            op.execute(local=False)
+
+        local_ops = list(self.forwards_local())
+        dist_ops = list(self.forwards_dist())
+        ops = (
+            local_ops + dist_ops if self.forwards_local_first else dist_ops + local_ops
+        )
+
+        for op in ops:
+            if op in local_ops:
+                op.execute(local=True)
+            if op in dist_ops:
+                op.execute(local=False)
+
         logger.info(f"Finished: {migration_id}")
         update_status(Status.COMPLETED)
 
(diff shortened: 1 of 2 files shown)


evanh (commented):
Would it be possible, instead of a flag, to automatically detect the order of operations of a migration?

If we know the set of SQL ops that require a certain order, why add the extra overhead to users of having to figure that out themselves?

dbanda:
> Would it be possible, instead of a flag, to automatically detect the order of operations of a migration?
> 
> If we know the set of SQL ops that require a certain order, why add the extra overhead to users of having to figure that out themselves?


We had some discussion about that and found its hard to do because the dependency graph could get complex with cycles and cant be resolved to a single order.

snuba/migrations/migration.py
```diff
@@ -113,6 +113,9 @@ class ClickhouseNodeMigration(Migration, ABC):
     completely unrelated, they are probably better as separate migrations.
     """
 
+    forwards_local_first: bool = True
+    backwards_local_first: bool = True
```
lynnagara, line 117: Why do we need to define two variables? Shouldn't backwards always the reverse of forwards? At least I can't think of a  valid scenario where that wouldn't be the case.

dbanda: It should be that way, but in our code there is nothing that enforces this. Backwards is another sequence or arbitrary SQL ops.

MeredithAnya: I think the default should be that `backwards_local_first` is `False` if we have the default be that `forwards_local_first` is `True`.

MeredithAnya:
Can we add a test ?

snuba/migrations/migration.py
dbanda: agreed. changed the defaults.

dbanda:
> Can we add a test ?

Added

tests/migrations/test_operations.py
```diff
@@ -182,3 +187,57 @@ def test_modify_ttl() -> None:
...
+
+            return add_op
+
+        op.execute.side_effect = effect(op)
+
+    test_migration = TestMigration()
+    test_migration.forwards(context)
+    assert order == [create_local_op, create_dist_op]
+    order.clear()
+    test_migration.backwards(context, False)
```
MeredithAnya, line 234: why does the `backwards` method need the extra arg?

dbanda: not sure why, but on our migrations `dry_run` isn't optional for backwards. surprised me too.

MeredithAnya: hmm my guess is that's an oversight, but if it isn't thats not super clear haha. okay then we can leave the test as is and find out if we should be passing a default value of `False` for backwards as well

MeredithAnya (approved):

tests/migrations/test_operations.py
```diff
@@ -182,3 +187,57 @@ def test_modify_ttl() -> None:
...
+            return [drop_dist_op]
+
+    ops = [create_local_op, create_dist_op, drop_local_op, drop_dist_op]
+    order = []
+    for op in ops:
+
+        def effect(op: Mock) -> Callable[[bool], None]:
+            def add_op(local: bool) -> None:
+                order.append(op)
+
```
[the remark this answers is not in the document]
dbanda: The linter forced them on me :(

PR #3326, opened 2022-11-01
Explicitly initialize snuba take 2

volokluev:
This is mostly the same as #3311 with the only difference being the name transformation

if you have a click command like this:

```python3
@click.command
def my_command():
    pass
```

it will be accessible in the cli as `my-command`. #3311 naively expected it to be `my_command`. 

This PR fixes that and adds a comment explaining why it's important

### Testing

Today I remembered about the `snuba devserver` command which starts up everything with the correct names. If I had done that initially I wouldn't have had to revert.

With this change, `snuba devserver` starts up properly (before it did not)

### DIFF
--- a/snuba/cli/__init__.py
+++ b/snuba/cli/__init__.py
@@ -34,15 +34,26 @@ with sentry_sdk.start_transaction(
             rv = []
             for filename in os.listdir(plugin_folder):
                 if filename.endswith(".py") and filename != "__init__.py":
-                    rv.append(filename[:-3])
+                    # IMPORTANT!!!!
+                    # Click replaces underscores with dashes for command names
+                    # by default. To mimic this behavior we have to do that here
+                    # since all of our infrastructure depends on this being the
+                    # case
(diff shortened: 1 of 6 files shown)


snuba/cli/__init__.py
```diff
@@ -1,51 +1,79 @@
...
+            return rv
+
+        def get_command(self, ctx: Any, name: str) -> click.Command:
+            logger.info(f"Loading command {name}")
+            # IMPORTANT!!!!
+            # Click replaces underscores with dashes for command names
+            # by default. To mimic this behavior we have to do that here
+            # since all of our infrastructure depends on this being the
+            # case
+            actual_command_name = name.replace("-", "_")
```
volokluev, line 53: bulk of the change

evanh (approved):
I am OK with this change, but it feels a little brittle. If someone checks in a file with an actual dash in it, this will break again.

volokluev:
>  If someone checks in a file with an actual dash in it, this will break again.

True but presumably they would run the command at least once before checking it in

PR #3330, opened 2022-11-02
ref(MDC): Move organization_id to SubscriptionData

enochtangg:
This PR is the first of 3 smaller PRs from the original https://github.com/getsentry/snuba/pull/3310#issuecomment-1299000241. As the original PR was evolved, the new updates caused many file changes across the repository which made reviewing difficult.

## Overview
As part of the entity subscription refactor, it was discovered that we currently store query specific data (`organization: int`) in `EntitySubscriptions`. As mentioned previously https://github.com/getsentry/snuba/pull/3310#issuecomment-1296303561, this is not ideal as the Snuba dataset data model is designed to be immutable and static. Because of this, we are currently instantiating a new EntitySubscription for every subscription query. To eventually move away from this design, this PR moves organization_id from `EntitySubscription` into `SubscriptionData` (where the query is stored). By doing this, we can move one step closer to making `EntitySubscription` defined on the `Entity` and make it immutable + static.


snuba/datasets/entity_subscriptions/entity_subscription.py
```diff
@@ -65,33 +65,26 @@ def validate_query(self, query: Union[CompositeQuery[Entity], Query]) -> None:
...
                 ConditionFunctions.EQ,
                 Column(None, None, "org_id"),
-                Literal(None, self.organization),
+                Literal(None, organization),
             ),
         ]
 
     def to_dict(self) -> Mapping[str, Any]:
-        return {"organization": self.organization}
+        return self.__dict__
```
evanh, line 82: I don't think this is right. This will start returning a bunch of values that weren't being returned before. Really this should return at more an empty dictionary, if organization is being encoded elsewhere.

That being said, it doesn't make a lot of sense to have a function that just returns an empty dictionary. If it doesn't overcomplicate the change, I would remove `to_dict` from the entity subscription completely.

tests/subscriptions/entity_subscriptions/test_entity_subscriptions.py
```diff
@@ -124,35 +130,38 @@
...
     exception: Optional[Type[Exception]],
 ) -> None:
+    es = entity_subscription()
     if exception is not None:
         with pytest.raises(exception):
-            entity_subscription(**creation_dict)
-    else:
-        es = entity_subscription(**creation_dict)
-        assert es.to_dict() == creation_dict["data_dict"]
+            es.get_entity_subscription_conditions_for_snql(None, organization)
```
evanh, line 148: I don't think you changed this test correctly. You shouldn't remove the `else` clause, that will skip a bunch of the tests. It should look like:

```
if exception is not None:
    with pytest.raises(exception):
        something()
else:
    something()
```

Also, you should assert something on this call in the successful case. As written this doesn't actually verify the conditions are correct in any way.

tests/subscriptions/test_codecs.py
```diff
@@ -70,247 +72,248 @@ def build_snql_subscription_data(
...
-        assert isinstance(subscription.entity_subscription, subscription_cls)
-        assert getattr(subscription.entity_subscription, "organization") == organization
-    else:
-        assert isinstance(subscription.entity_subscription, EventsSubscription)
-        with pytest.raises(AttributeError):
-            getattr(subscription.entity_subscription, "organization")
-
-
-@pytest.mark.parametrize("builder, organization, entity_key", SNQL_CASES)
-def test_basic(
```
evanh, line 90: I don't want to read through all this: you didn't actually remove any tests right? Just moved them around?

snuba/subscriptions/data.py
```diff
@@ -172,14 +173,15 @@ def build_request(
...
-        entity_subscription = get_entity_subscription(entity_key)(data_dict=data)
+        entity_subscription = get_entity_subscription(entity_key)()
 
         return SubscriptionData(
             project_id=data["project_id"],
             # Always cast to int in case any old subscriptions are stored as float
             time_window_sec=int(data["time_window"]),
             resolution_sec=int(data["resolution"]),
             query=data["query"],
+            organization=data["organization"] if "organization" in data else None,
```
evanh, line 184:
```suggestion
            organization=data.get("organization"),
```

evanh (commented):
Couple comments on the test refactors.

snuba/datasets/entity_subscriptions/entity_subscription.py
enochtangg: Right, organization is already being encoded on the `SubscriptionData`. Was thinking since `entity_subscription` is an attribute of the subscription data, it might make sense to encode that data too. But since this changes the way we do thinks right now, I will remove.

tests/subscriptions/test_codecs.py
enochtangg: Yup, wrapped everything inside `ConfigurationTest` to reset data factory

snuba/subscriptions/data.py
```diff
@@ -79,6 +79,7 @@ class SubscriptionData:
     time_window_sec: int
     entity_subscription: EntitySubscription
     query: str
+    organization: Optional[int] = None
```
lynnagara, line 82:
```suggestion
    org_id: Optional[int]
```
- org_id is the actual name of the column, let's align to this
- remove the default value, i think it should be explicitly passed
- i would probably probably put this directly below project_id

bin/mocks/mock-subscriptions
```diff
@@ -39,6 +39,13 @@ parser.add_argument(
     choices=["events"],
     default="events",
 )
+parser.add_argument(
+    "--organization",
```
lynnagara, line 43:
```suggestion
    "--organization-id",
```
So this is consistent with the project-id arg above

snuba/subscriptions/data.py
```diff
@@ -172,14 +173,15 @@ def build_request(
...
-        entity_subscription = get_entity_subscription(entity_key)(data_dict=data)
+        entity_subscription = get_entity_subscription(entity_key)()
 
         return SubscriptionData(
             project_id=data["project_id"],
             # Always cast to int in case any old subscriptions are stored as float
             time_window_sec=int(data["time_window"]),
             resolution_sec=int(data["resolution"]),
             query=data["query"],
+            organization=data.get("organization", None),
```
lynnagara, line 184: Ideally this would be like:
```suggestion
            organization=data["organization"],
```

Curious why you decided to provide a fallback here? Was it to deal with deployment issues where some older data from before this change gets deployed won't have the "organization" key yet?

lynnagara (commented):
EntitySubscription was originally designed in this way in order that entities can extend the subscription data with custom logic and validations that is highly specific to the entity. In order to support these customised validations, an new EntitySubscription instance was created each time we create SubscriptionData which has access to the query's data. It is essentially an extension for SubscriptionData..

These changes here will make entity subscription close to obsolete, since all the logic of entity subscriptions moves into SubscriptionData and SubscriptionData must know about all the different subscription variants for all the entities.

Given that we have so few different subscription types we right now, I think this is fine and SubscriptionData will not be overly complex. 

But I think it imposes a lot of constraints on data models now that didn't previously exist. For example if someone has a column called "organization_id" instead of "org_id" this won't work anymore. If someone wants to have a special validation on "my_product_id" then they would have to go and modify the base class in Snuba and change the data format for all subscriptions instead of building an EntitySubscription that only affects them. So previously we could've just created a custom EntitySubscription for that dataset and not changed anyone else's data but that would no longer be possible. If we one day get to a point where we have a lot of datasets and a lot more subscription formats and variation in data models I think this may need to be revisited and we may actually want to bring something like entity subscription back again.

In some ways I think this might actually go against the goals of product autonomy even though there may be slightly less boilerplate for users to define in a dataset. This is not a blocker for me to move forward with this approach, but I wanted to point this out.

snuba/subscriptions/data.py
enochtangg: Sure, it's also important to note that `organization` is the default key in the data payload. For processes (e.g. tests) that both encode and decode its own payload, we would need to support parsing for both `organization` and `org_id` then.

snuba/subscriptions/data.py
enochtangg: Yes, it is for lagging subscriptions that might not have `organization` key.

snuba/subscriptions/data.py
evanh: Yeah I think this is important to note. There is no change to how the data is being stored (or there shouldn't be). Before this change, the org ID was stored in the `organization` key in the dictionary.

I would advise against changing anything in how the data is actually stored. That's a risk we don't need to take.

snuba/subscriptions/data.py
evanh: I don't know that `organization` is passed for all subscriptions. It's only required by sessions, so I don't think we can assume there will be an organization for transactions subscriptions for example.

snuba/subscriptions/data.py
```diff
@@ -75,6 +75,7 @@ class SubscriptionData:
     """
 
     project_id: int
+    organization: Optional[int]
```
fpacifici, line 78: This approach would have an issue. Not all datasets know about organizations. Errors and transactions do not have an organization column. 
We did not put the organization id here back when we introduced sessions subscriptions exactly because it was not a common field to all datasets types we wanted to run subscriptions onto. 
Thus the SubscriptionEntity. 

This seems to go against a design where subscriptions are customizable per dataset (they are exactly because not everybody has the same fields and validators).

fpacifici: If your goal is to make `SubscriptionEntity` immutable and static, which can be a good idea, why not just putting the logic into`EntitySubscription` ? 
What we needed `EntitySubscription` for was to process the subscription query before executing in a way that was specific to the `Entity`. There is nothing `EntitySubscription` needs that is not already in the query: like the condition on the organization. 
This is how the EntitySubscription is used 
https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/data.py#L145-L169
The entity subscription logic could be executed after the parsing and could extract the query condition from the query itself.

evanh: I don't think we are breaking that design necessarily. The organization data was already being stored in redis alongside all these other fields. Instead of reading that field out and storing it in a different class from the other fields, we are simply putting into this class.

The customization is still possible, because the entity can configure whether or not to use these fields.

fpacifici: Wouldn't that mean that, if tomorrow profiles decides to add a new field to the subscription data class, we would add a profile specific field in the generic `SubscriptionData` ? 

Though, taking a step back, I am not 100% sure this whole problem needs to exist. 
@lynnagara  do you remember why we needed to have organization id as a separate field in the SubscriptionData representation in redis? That is constant for a subscription id, it is not like the time window that changes at every iteration. Is there any reason the `organization_id` condition cannot just be in the SnQL query string ?

@evanh , @enochtangg you should also ensure any design change is in line with what the Slicing project is planning to change around subscriptions: https://www.notion.so/sentry/Generic-Metrics-Slicing-9c206e02f92a4abca5948d2e21c6c00d#cdb80a402b894440adf05cf1805363ec

lynnagara: > Though, taking a step back, I am not 100% sure this whole problem needs to exist. @lynnagara do you remember why we needed to have organization id as a separate field in the SubscriptionData representation in redis? That is constant for a subscription id, it is not like the time window that changes at every iteration. Is there any reason the `organization_id` condition cannot just be in the SnQL query string ?
I believe it was done for the same reason that project_id is not in the SnQL query string. We need to be extra careful to ensure there is only one.

fpacifici: > I believe it was done for the same reason that project_id is not in the SnQL query string. We need to be extra careful to ensure there is only one.

Project id, was not there because of historical reasons:
- it was before SnQL. project id was not in the query at that time.
- project id was supposed to be known by the pipeline for semantic partitioning, and before the AST there was no way. 

So in both cases, this should not have applied to organization id when this was introduced and if these are the reasons  it should not be needed anymore, unless it is needed somehow in the results.
I would consider getting rid of it and fixing the content of the storage.

evanh: > it should not be needed anymore, unless it is needed somehow in the results

We would need to run a migration to update all the subscriptions in Redis to make sure all the ones with an `organization` tag also have an organization condition. This feels like a lot of unnecessary work. Even if we don't remove it, organization is a concept that is useful across many datasets.

PR #3331, opened 2022-11-02
ref(migrations): Add `CLICKHOUSE_MIGRATIONS_USER`

MeredithAnya:
_description incoming_


snuba/clusters/cluster.py
```diff
@@ -55,6 +55,16 @@ class ClickhouseClientSettings(Enum):
         },
         10000,
     )
+    MIGRATE_DDL = ClickhouseClientSettingsType(
+        {
+            "readonly": 2,
```
MeredithAnya, line 60: This shouldn't affect the ability to run DDL queries but when testing this out locally I get: 

```
snuba.clickhouse.errors.ClickhouseError: DB::Exception: default: Cannot execute query in readonly mode.
```

so not sure if that's a 20.3.9.70 thing or what

lynnagara: Did you get that error on all operations or just the non DDL ones? Note thatthe migration system needs write access to the migrations table only but not the rest.

dbanda: > This shouldn't affect the ability to run DDL queries but when testing this out locally I get:
> 
> ```
> snuba.clickhouse.errors.ClickhouseError: DB::Exception: default: Cannot execute query in readonly mode.
> ```
> 
> so not sure if that's a 20.3.9.70 thing or what

Maybe this is thrown by the test setup fixture which runs before all tests and needs write access.

MeredithAnya:
We can just use the default user given that there is no way to combine `allow_ddl` and `readonly` user with levels greater than 0 see https://github.com/ClickHouse/ClickHouse/issues/43068

PR #3334, opened 2022-11-02
feat(api): Snuba healthcheck observability

rahul-kumar-saini:
### Overview
- Closes [SNS-1740](https://getsentry.atlassian.net/browse/SNS-1740)
- This PR introduces a new `healthcheck_failed` metric for non 200 responses in the Snuba API `/health` route
    - The metric also records a couple relevant booleans as tags

As far as Clickhouse goes, I believe we are already logging the errors on health check:

https://github.com/getsentry/snuba/blob/ed5cbef9086204cbd5dc105a4a69091f96080ba6/snuba/web/views.py#L155-L164

We can check Sentry/GCP logs for when we see a failed health check where `clickhouse_ok` is `False`

### DIFF
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -468,7 +468,11 @@ def get_cluster(
         res = part_storage_set_cluster_map.get((storage_set_key, slice_id), None)
         if res is None:
             raise UndefinedClickhouseCluster(
-                f"{(storage_set_key, slice_id)} is not defined in the SLICED_CLUSTERS setting for this environment"
+                f"{(storage_set_key, slice_id)} is not defined in the SLICED_CLUSTERS setting for this environment",
+                extra_data={
+                    "storage_set_key_not_defined": storage_set_key,
+                    "slice_id": slice_id,
+                },
             )
 
     else:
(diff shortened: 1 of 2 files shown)


volokluev:
Can we instrument the `check_clickhouse` better? Send to datadog which clusters are failing the healthcheck?

rahul-kumar-saini:
> Can we instrument the `check_clickhouse` better? Send to datadog which clusters are failing the healthcheck?

I looked into that but I think if some cluster fails to connect it would just throw an exception, I could try to catch a specific exception and see if I can parse out the cluster from it

evanh (approved):
I'm going to approve this, but I kind of agree with @volokluev  It would be nice to get the table/cluster back from `check_clickhouse` so we can add that to the metric. Would potentially save time digging through logs.

volokluev:
@rahul-kumar-saini the main point of the ticket was to get that data. You can make another PR for it if you like but I wouldn't close the ticket until we have that

rahul-kumar-saini:
@volokluev I've got a few tags set based on type of error or if a table name isn't found on a cluster now. If it's a Clickhouse error it will at least report the causing exceptions name.

I'm not sure we can extract any further data unless there are specific cases we want to address, the logs for INC-215 aren't in GCP anymore but I don't think they would've helped anyway.

evanh (commented):
Remove the `print` and this looks good to me.

evanh (approved):

PR #3337, opened 2022-11-02
Add yaml profiles storage

volokluev:
this yaml file was autogenerated and checked using the script in #3336


rahul-kumar-saini:
#3359

PR #3338, opened 2022-11-03
feat(slices): Make Subscription Scheduler filter by slice ID

ayirr7:
For general context:
https://getsentry.atlassian.net/browse/SNS-1759 

Up-to-date Approach for scheduling Subscriptions in a Sliced context:

- Getting the partition key from the corresponding `EntitySubscription`
- Mapping the partition key's value (some integer) to slice id (using partitioning functions)
- Check if this slice id is equal to the slice id that is passed into the `SubscriptionScheduler` instance
- If equal, keep the `Subscription`. Otherwise, filter out.
- Ensure that tasks are being scheduled on a sliced (physical) subscriptions scheduler topic, if a slice_id is passed in

Updates: There was discussion around changing this approach, however, we are now deciding to move forward with this approach for now and will most likely move around/refactor once changes to `EntitySubscription` are merged.


evanh:
@enochtangg You should look at this, we'll need to factor this into the `EntitySubscription` changes.

snuba/subscriptions/scheduler.py
```diff
@@ -344,7 +352,33 @@ def __get_subscriptions(self) -> List[Subscription]:
...
+                # get the EntitySubscription from the Subscription
+                sub_data = subscription.data
+                entity_sub = sub_data.entity_subscription
+
+                # partition key is only defined for SessionsSubscriptions
+                if isinstance(entity_sub, SessionsSubscription):
+                    partition_key_value = entity_sub.get_partitioning_key()
+
+                    # map the partition key's value to the slice ID
+                    logical_part = map_org_id_to_logical_partition(partition_key_value)
```
dbanda, line 368: Seems like here you are assuming the partition key for `entity_sub` is always org_id. Is that always the case? Seems like this is tightly coupled to the implementation of `entity_sub` and  would break if we change the partitioning key. SessionSubscription has an organization attribute, maybe we can use that to make it clear we are getting the org_id? Or maybe `map_org_id_to_logical_partition` can be refactored to be more general i.e `map_patition_key_to_logical_partition`

ayirr7: That's a good point, and it makes sense. I think it would be good to change it to `map_partition_key_to_logical_partition` as you are suggesting.

snuba/subscriptions/scheduler.py
```diff
@@ -321,6 +329,36 @@ def __reset_builder(self) -> None:
             # We are transitioning between jittered and immediate mode. We must use the delegate builder.
             self.__builder = self.__delegate_builder
 
+    def __get_filtered_subscriptions(self) -> List[Subscription]:
```
nikhars, line 332: I would recommend that you pass the original list of subscriptions to this method rather than relying on `self.__subscriptions` and rename the method to `__filter_subscriptions`. Looking at the usage of `self.__subscriptions` I don't think there is value in having the private instance variable. You could probably get rid of it completely.

snuba/subscriptions/scheduler.py
```diff
@@ -321,6 +329,36 @@ def __reset_builder(self) -> None:
...
+        if self.__slice_id is not None:
```
nikhars, line 333: It would be better if we move the slice_id checks in __get_subscriptions and call __get_filtered_subscriptions only in the case of slice_id being enabled.

snuba/datasets/entity_subscriptions/entity_subscription.py
```diff
@@ -88,6 +92,9 @@ def get_entity_subscription_conditions_for_snql(
     def to_dict(self) -> Mapping[str, Any]:
         return {"organization": self.organization}
 
+    def get_partitioning_key(self) -> int:
```
nikhars, line 95: Let's implement this method for `GenericMetricsSetsSubscription` and `GenericMetricsDistributionsSubscription` only for now since they are the only entities which need slicing.

nikhars (commented):
Please add slice_id tag to the MetricsWrapper when slice_id is passed via the CLI

snuba/subscriptions/scheduler.py
ayirr7: Sure, this makes sense. Updated.

snuba/datasets/entity_subscriptions/entity_subscription.py
ayirr7: Ok, this makes sense

snuba/subscriptions/scheduler.py
```diff
@@ -321,21 +327,50 @@ def __reset_builder(self) -> None:
...
+    def __filter_subscriptions(
+        self, subscriptions: List[Subscription]
+    ) -> List[Subscription]:
+
+        filtered_subscriptions: List[Subscription] = []
+        for subscription in subscriptions:
+            # get the metadata and org_id from the Subscription
+            sub_data = subscription.data
+            sub_metadata = sub_data.metadata
+            org_id = sub_metadata["organization"]
```
nikhars, line 339: Is there a way to make access to the metadata fields be specific to an Entity? Not all subscriptions would have an `organization` field. So if tomorrow this method gets called for some other Entity, this would cause an Exception

ayirr7: Good point. I can just move the `if` statement up, which checks for generics metrics EntityKeys

ayirr7: Good point. Would it make sense to move the `if` statement that checks for generic metrics EntityKeys to `__get_subscriptions` to the filter step? This way, if the corresponding Entity for a subscription is not within generic metrics, we don't do the filter step.

ayirr7: Good point. I think can add a condition for checking the EntityKey that corresponds with a Subscription. Where we add  this condition depends:

- if `slice_id` is not None, but the entity of the subscription is not within generic metrics, should we just return the regular list of subscriptions? 

Or, should we just restrict the filter step to Subscriptions that satisfy these conditions: (1) `slice_id` is not None (2) the EntityKey for the Subscription is within generic metrics

nikhars: Lets allow filtering to be called when slice_id is enabled. Within the filter method you can restrict it to specific EntityKey.

ayirr7: Ok, that makes sense to me. Just to be clear, does this mean that if we have a `slice_id` enabled, but a non-generic metrics entity, we return all of the subscriptions (no filtering needed) or something else?

nikhars: Yes. The filter can just return the original subscriptions list for non-generic metrics entity

ayirr7: Done!

onewland (commented):
This isn't currently reading from the correct commit topic.

Either:
https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/scheduler_consumer.py#L260-L266

or 

https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/scheduler_consumer.py#L235-L237

needs to take slice_id into account

snuba/subscriptions/scheduler.py
```diff
@@ -321,6 +328,36 @@ def __reset_builder(self) -> None:
             # We are transitioning between jittered and immediate mode. We must use the delegate builder.
             self.__builder = self.__delegate_builder
 
+    def __filter_subscriptions(self) -> List[Subscription]:
+
+        # only generic metrics storage sets are currently sliced
+        if (
+            self.__entity_key == EntityKey.GENERIC_METRICS_SETS
+            or self.__entity_key == EntityKey.GENERIC_METRICS_DISTRIBUTIONS
```
onewland, line 336: can we check if the storage set for the given entity key is sliced using the global logic rather than hardcoding these here?

the entity key is constant after initialization so we should only have to test for that once

snuba/subscriptions/scheduler.py
```diff
@@ -321,6 +328,36 @@ def __reset_builder(self) -> None:
...
+        ):
+
```
onewland, line 338: nit: extra line

snuba/subscriptions/scheduler.py
```diff
@@ -321,6 +328,36 @@ def __reset_builder(self) -> None:
...
+            filtered_subscriptions: List[Subscription] = []
+            for subscription in self.__subscriptions:
+                # get the metadata and org_id from the Subscription
+                sub_data = subscription.data
+                sub_metadata = sub_data.metadata
+                org_id = sub_metadata["organization"]
```
onewland, line 344: should we do something if `organization` is None? maybe we can emit a metric and skip the current subscription?

onewland (commented):
Mostly good with this approach (and tested it locally) but let's make sure there's a unit test on the filtering before we commit to master

snuba/subscriptions/scheduler.py
ayirr7: makes sense, changed

snuba/subscriptions/scheduler.py
ayirr7: good point, done

snuba/subscriptions/scheduler.py
```diff
@@ -278,6 +283,43 @@ def add_tag(tags: Tags, builder_type: str) -> Tags:
         ]
 
 
+def filter_subscriptions(
+    subscriptions: List[Subscription],
+    entity_key: EntityKey,
+    metrics: MetricsBackend,
+    slice_id: Optional[int] = None,
+) -> List[Subscription]:
```
nikhars, line 291:
```suggestion
) -> MutableSequence[Subscription]:
```

snuba/subscriptions/scheduler.py
```diff
@@ -278,6 +283,43 @@ def add_tag(tags: Tags, builder_type: str) -> Tags:
...
+                if org_id is not None:
+                    # map the org_id to the slice ID
+                    logical_part = map_org_id_to_logical_partition(org_id)
+                    part_slice_id = map_logical_partition_to_slice(
+                        storage.get_storage_set_key(), logical_part
+                    )
+                    if part_slice_id == slice_id:
+                        filtered_subscriptions.append(subscription)
+                else:
+                    metrics.increment("queries_with_orgID=None")
```
nikhars, line 316: I don't know if datadog supports using `=` in tag names. But in general, you should avoid using it. You can rename this to something like `subscription_with_empty_org_id`

onewland: agree with this

snuba/subscriptions/scheduler.py
onewland: FWIW, I think this type is copied from self.__subscriptions so it should probably be updated in both places or neither

tests/subscriptions/test_filter_subscriptions.py
```diff
@@ -0,0 +1,47 @@
...
+subs = expected_subs + extra_subs
+
+
+@patch("snuba.settings.SLICED_STORAGE_SETS", {"events": 3})
+@patch("snuba.settings.LOGICAL_PARTITION_MAPPING", {"events": {0: 0, 1: 1, 2: 2}})
+def test_filter_subscriptions() -> None:
+    importlib.reload(scheduler)
+
+    filtered_subs = filter_subscriptions(
+        subs, EntityKey.EVENTS, DummyMetricsBackend(strict=True), 2
```
onewland, line 45: A bit of a nit but can we used named rather than positional arguments here? It's just not super obvious that 2 is the org_id

I think it's generally good practice to use named arguments if the count is greater than 3

tests/subscriptions/test_filter_subscriptions.py
```diff
@@ -0,0 +1,47 @@
...
+            entity=get_entity(EntityKey.EVENTS),
+            metadata={"organization": org_id},
+        ),
+    )
+
+
+# create a list of subscriptions
+expected_subs = [build_subscription(timedelta(minutes=1), 2) for count in range(20)]
+extra_subs = [build_subscription(timedelta(minutes=3), 1) for count in range(10)]
+subs = expected_subs + extra_subs
```
onewland, line 36: Let's put these in the test method, or create fixture methods (https://docs.pytest.org/en/6.2.x/fixture.html#) to avoid future coupling

snuba/datasets/slicing.py
```diff
@@ -10,12 +10,12 @@
 SENTRY_LOGICAL_PARTITIONS = 256
 
 
-def map_org_id_to_logical_partition(org_id: int) -> int:
+def map_org_id_to_logical_partition(partition_key: int) -> int:
     """
-    Maps an org_id to a logical partition. Since SENTRY_LOGICAL_PARTITIONS is
+    Maps a partition key to a logical partition. Since SENTRY_LOGICAL_PARTITIONS is
     fixed, an org id will always be mapped to the same logical partition.
```
onewland, line 16: this doc comment and the method title still reference org_id, IMO we might just want to remove the rename, but I don't feel strongly

onewland (approved):
Overall looks good except for the little things pointed out by Nikhar and me

snuba/subscriptions/scheduler.py
ayirr7: I can change it back!

tests/subscriptions/test_filter_subscriptions.py
ayirr7: makes sense! fixed

snuba/datasets/slicing.py
ayirr7: this is true. i can revert it and we can revisit later if needed

nikhars (approved):

PR #3339, opened 2022-11-03
Pause optimize on large merges. attempt 2

dbanda:
This is a retry of #3158 which got reverted

## Context

We want to make our Optimize cron job wait for large merges to complete before starting. This avoids spinning up too many optimize jobs especially in cases where K8s reschedules the cron jobs. The main change in this PR is a check for ongoing large merges at the start of the cron job. 

In optimize.py the function `get_current_large_merges` queries Clickhouse to find large merge jobs. The function `is_busy_merging` repeatedly calls `get_current_large_merges` to find if there are existed merges that are large and returns true if there are large merges that pass our threshold for waiting. At the start of the optimize cron job, we repeatedly poll `is_busy_merging`, sleeping each time it returns true and progress only once it returns false.

This PR also refactors the optimize code to move it to `snuba/clickhouse/optimize` to avoid cluttering the root with optimize-specific utility functions.

## Blast Radius
This affects the optimize cron job. No impact on calling optimize via other methods.

## Before State
Optimize cron job would immediately spin up threads calling OPTIMIZE query when started

## After State
Optimize cron job checks if there are large merges going on at start that meet a threshold. If they exist, the job repeatedly sleeps until those other merges are finished.

## Testing Notes
It's difficult to create actual large merges, instead we mocked the some of the clickhouse query responses to `system.merges` with predefined partitions of large size; then checked that sleep was called when responses indicated large partitions.


dbanda:
For some reason, Github does not detect that `test_optimize_tracker.py` was moved and thinks it a new file. The only changes there are `test_run_optimize_with_ongoing_merges()` and `test_merge_info()`

nikhars (approved):

snuba/clickhouse/optimize/optimize.py
```diff
@@ -80,6 +89,11 @@ def run_optimize_cron_job(
     database = storage.get_cluster().get_database()
     optimize_scheduler = OptimizeScheduler(parallel=parallel)
 
+    # if theres a merge in progress wait for it to finish
+    while is_busy_merging(clickhouse, database, table):
+        logger.info(f"busy merging, sleeping for {OPTIMIZE_BASE_SLEEP_TIME}s")
+        time.sleep(OPTIMIZE_BASE_SLEEP_TIME)
```
nikhars, line 95: One optimization we could do is potentially add a higher bound on how much time to sleep. Usually merges take 1.5 hours to complete. So you could add an upper bound to protect from code logic bug. The upper bound could be 2 hours. So even if the signal says that there is an ongoing merge, you could proceed after 2 hours have passed from the check starting.

dbanda: That makes sense. Added the check to set upper bound to 2 hours.

PR #3342, opened 2022-11-04
fix(slices): Rename partition to slice 

ayirr7:
`is_storage_partitioned` to `is_storage_sliced`

Calling the storage "partitioned" in the context of slicing can be confusing, so changing it to "sliced" since this will refer to storages that are associated with a storage set key that is sliced. Storage sets are still the level at which slicing is happening, but we want to be consistent with slicing config names in settings like `SLICED_STORAGES`


nikhars (commented):
Let's rename the file snuba/datasets/partitioning.py to snuba/datasets/slicing.py. 

Let's also update the docs and rename partitioning to slicing. And change the name of the file to slicing.rst. https://github.com/getsentry/snuba/blob/master/docs/source/architecture/partitioning.rst to

snuba/datasets/pluggable_entity.py
```diff
@@ -69,7 +69,7 @@ def get_all_join_relationships(self) -> Mapping[str, JoinRelationship]:
     def get_query_pipeline_builder(self) -> QueryPipelineBuilder[ClickhouseQueryPlan]:
         from snuba.pipeline.simple_pipeline import SimplePipelineBuilder
 
-        if is_storage_partitioned(self.readable_storage.get_storage_key()):
+        if is_storage_sliced(self.readable_storage.get_storage_key()):
             assert (
                 self.partition_key_column_name is not None
```
lynnagara, line 74: `slice_key_column_name`?

lynnagara (commented):
Thanks for this! 😍 for naming consistency everywhere

nikhars (approved):

nikhars:
Please fix the docs build failure before merging.

PR #3343, opened 2022-11-04
feat: Make multistorage consumer dlq respect the policy defined

lynnagara:
Currently the multistorage consumer completely ignores the policy defined on the storage itself and just applies it's own policy which produces the message to a topic. This is wrong and also different to how the main consumer works where the policy defined on the storage is the one that is used. It was also applied on a different part of the pipeline (only wraps the collector step). This also means the collector implementation was inconsistent in the two consumer variants.

When this hack was first introduced, it's likely there were practical reasons we needed this at Sentry, but since this particular code path (multistorage consumer with dlq) is never used today, those obviously don't exist anymore.

This is a step on the path to folding the multistorage consumer into the main consumer and making them behave in the same way.

### DIFF
--- a/snuba/cli/multistorage_consumer.py
+++ b/snuba/cli/multistorage_consumer.py
@@ -1,12 +1,11 @@
 import logging
 import signal
-from contextlib import closing
 from typing import Any, Optional, Sequence
 
 import click
 import rapidjson
 from arroyo import Topic, configure_metrics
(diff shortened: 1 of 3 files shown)


evanh (approved):
How did you confirm these aren't used anymore?

lynnagara:
> How did you confirm these aren't used anymore?

It was fairly straightforward since we only define one multistorage consumer these days (for groupedmessages + groupassignees together). There is no dlq on those.

PR #3345, opened 2022-11-04
fix: Fix the multistorage replacements producer Kafka configuration

lynnagara:
Previously this used the producer configuration from the default topic. If this happened to not be the same cluster as the replacements topic this wouldn't work. Now, it doesn't matter which cluster the replacement topic is on, the right producer configuration will be built.


untitaker (approved):

untitaker (commented):
without fully understanding what this does I approved it since I think it's just about bringing L796 in line with L803

lynnagara:
> without fully understanding what this does I approved it since I think it's just about bringing L796 in line with L803

It gets used to build the configuration for the producer based on mappings of topic to brokers that users can provide in settings. Most of the time it doesn't have an impact. But if we had the 2 topics on different clusters then it would be broken.

PR #3347, opened 2022-11-04
Explicitly initialize snuba take 3

volokluev:
Compiling changes in #3311 and #3326 and moving the initialization up  before subcommand compilation

The deploy of #3326 caused a crashloop on `snuba-generic-metrics-distributions-subscriptions-scheduler-production`and `snuba-generic-metrics-distributions-subscriptions-scheduler-production` (and only those two)

This happened because:

1. generic_metrics entities are defined in config
2. the subscriptions scheduler cli has the entity argument defined like so:

```
@click.option(
    "--entity",
    "entity_name",
    required=True,
    type=click.Choice([entity_key.value for entity_key in EntityKey]),
    help="The entity to target",
)
```

This makes the cli depend on the entity key being fully populated *at import time*. This also makes it so that the code as is only work *by accident*. 

## The fix

We move up initialization *before* we compile the subcommand code so that all the entity factories are initialized at import time

## The test

This issue was reproducible locally by trying to run those consumer groups `snuba devserver` did not call those commands either. They have been added.

Undoing the fix will make this command fail:

```
SEPARATE_SCHEDULER_EXECUTOR_SUBSCRIPTIONS_DEV=1 ENABLE_METRICS_SUBSCRIPTIONS=1 ENABLE_SENTRY_METRICS_DEV=1 snuba devserver
```

### DIFF
--- a/snuba/cli/__init__.py
+++ b/snuba/cli/__init__.py
@@ -53,11 +53,15 @@ with sentry_sdk.start_transaction(
             actual_command_name = name.replace("-", "_")
             with sentry_sdk.start_span(op="import", description=name):
                 ns: dict[str, click.Command] = {}
+                # NOTE: WE initialize snuba before we compile the command code
+                # That way if any command code references any snuba construct that needs
+                # to be initialized (e.g. a factory) at import time, it is already initialized
+                # into the runtime
+                initialize.initialize()
                 fn = os.path.join(plugin_folder, actual_command_name + ".py")
                 with open(fn) as f:
                     code = compile(f.read(), fn, "exec")
                     eval(code, ns, ns)
-                initialize.initialize()
                 init_time = time.perf_counter() - start
                 metrics.gauge("snuba_init", init_time)
                 logger.info(f"Snuba initialization took {init_time}s")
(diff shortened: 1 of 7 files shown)


snuba/cli/__init__.py
```diff
@@ -1,51 +1,83 @@
...
+            logger.info(f"Loading command {name}")
+            # IMPORTANT!!!!
+            # Click replaces underscores with dashes for command names
+            # by default. To mimic this behavior we have to do that here
+            # since all of our infrastructure depends on this being the
+            # case
+            actual_command_name = name.replace("-", "_")
+            with sentry_sdk.start_span(op="import", description=name):
+                ns: dict[str, click.Command] = {}
+                # NOTE: WE initialize snuba before we compile the command code
```
volokluev, line 56: This is the main notable change

snuba/cli/devserver.py
```diff
@@ -270,6 +270,34 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                             "--schedule-ttl=10",
                         ],
                     ),
+                    (
```
volokluev, line 273: these invocations added to the devserver

evanh (approved):

PR #3350, opened 2022-11-07
feat(replays): Add additional parsing safety to string column types

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/192

This will be the first in a series of refactors to the replays processor.  The goal is to address the issues raised by INC-250 and to add preventative safety measures.


evanh (approved):
LGTM

PR #3352, opened 2022-11-07
feat(replays): Add additional parsing safety to integer and datetime column types

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/193

- Strictly collapses integers or errs.
- Generates timezone aware datetimes from integer timestamps.
- Defaults to timezone aware utc now.
- Adds missing key safety for default-able fields.

### DIFF
--- a/snuba/datasets/processors/replays_processor.py
+++ b/snuba/datasets/processors/replays_processor.py
@@ -86,7 +86,7 @@ class ReplaysProcessor(DatasetMessageProcessor):
     ) -> None:
         processed["replay_id"] = str(uuid.UUID(replay_event["replay_id"]))
         processed["segment_id"] = maybe(
-            coerce_segment_id, replay_event.get("segment_id")
+            _coerce_segment_id, replay_event.get("segment_id")
         )
         processed["trace_ids"] = self.__process_trace_ids(replay_event.get("trace_ids"))
 
@@ -251,11 +251,6 @@ def maybe(into: Callable[[T], U], value: T | None) -> U | None:
     return None if value is None else into(value)
 
 
-def coerce_segment_id(value: Any) -> int:
-    """Return a 16-bit integer or err."""
-    return _collapse_or_err(_collapse_uint16, _intify(value))
-
-
 def stringify(value: Any) -> str:
     """Return a string or err.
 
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/replays_processor.py
```diff
@@ -273,3 +283,35 @@ def stringify(value: Any) -> str:
...
+            f'Expected value of type "int"; received "{type(value)}" with a value of "{value}".'
+        )
+
+
+def _collapse_or_err(callable: Callable[[int], int | None], value: int) -> int:
+    """Return the integer or error if it overflows."""
+    if callable(value) is None:
+        # This exception can only be triggered through abuse.  We choose not to suppress these
+        # exceptions in favor of identifying the origin.
+        raise ValueError(f'Integer "{value}" overflowed.')
```
MeredithAnya, line 305: Won't this cause the whole replay processing pipeline to stop if one org is causing the abuse?

cmanallen: This particular error *should* trigger the replays DLQ (let me know if this assumption is incorrect).  An exception raised after the processing step (e.g. the write step) caused INC-250.  I don't believe this change introduces any pipeline risk.

snuba/datasets/processors/replays_processor.py
```diff
@@ -274,6 +271,51 @@ def _encode_utf8(value: str) -> str:
...
+def datetimeify(value: Any) -> datetime:
+    """Return a datetime instance or err.
+
+    Datetimes for the replays schema standardize on 32 bit dates.
+    """
+    return _timestamp_to_datetime(_collapse_or_err(_collapse_uint32, _intify(value)))
+
+
+def _intify(value: Any) -> int:
+    """Return an integer or err."""
```
evanh, line 283: Doesn't the `int` cast already do that? It will throw a `ValueError` if the thing it is casting is not int-able.

snuba/datasets/processors/replays_processor.py
```diff
@@ -274,6 +271,51 @@ def _encode_utf8(value: str) -> str:
...
+    if isinstance(value, int):
+        return value
+    elif isinstance(value, (str, float)):
+        return int(value)
```
evanh, line 287: This will return a `ValueError` if the string is not int-able.

snuba/datasets/processors/replays_processor.py
cmanallen: Ah, duh..  replaced with `int`.

evanh (approved):
Looks good, nice to see the additional tests!

PR #3353, opened 2022-11-07
feat(migrations): add migration order validator SNS-1831

dbanda:
Adds a script to validate the order of migration and introduce a test in CI that checks existing migrations

## Context
#3324 added the option to set flags to specify order, this PR introduces a checker to validate the specified order does not introduce errors. The checker uses the following spec:

1. For two `AddColumn` OPs with the same target table and target column, the local OP must be applied first
2. For two `CreateTable` OPs with the same target table name, the local OP must be applied first
3. For two `DropColumn` OPs with the same target table and target column, the dist OP must be applied first

The target table for distributed ops is then extracted either by looking at the engine attribute in the case of `CreateTable`, or by parsing the the local table name from the table engine in Clickhouse via querying `SELECT engine_full FROM system.tables WHERE name = table_name` for `AddColumn`,`DropColumn` ops. For distributed tables, the target is the third argument in the `Distributed()` engine_full value.

## Before
We wouldn't be able to detect if the wrong order was specified in migrations

## After
A checker script is run that can catch some errors when the wrong migration operation order is specified

##  Blast Radius
The validator is an independent script in the migrations package. It doesn't affect the migrations themselves, but this PR add a test to check the existing migrations. 

To make it easier to implement, we make public some attributes of `SqlOperations` such as the table name and column.

For migrations `0004_drop_profile_column.py`,  `0016_drop_legacy_events` and `0014_transactions_remove_flattened_columns.py` we corrected the order flags as the checker detected errors.

## Testing
Unit tests check the utility methods for detecting conflicting ops that target the same underlying table and column 

A parameterized test creates a series of Migrations with create, add column and drop column ops and checks that if the order is violated the validator throws the correct exception and message

A unit test creates empty tables in Clickhouse and checks that the local table name can be parsed properly from the engine name

### DIFF
--- a/snuba/migrations/validator.py
+++ b/snuba/migrations/validator.py
@@ -1,0 +1,159 @@
+import re
+from typing import Union
+
+from snuba.clickhouse.native import ClickhousePool
+from snuba.clusters.cluster import ClickhouseClientSettings, get_cluster
+from snuba.migrations.migration import ClickhouseNodeMigration
+from snuba.migrations.operations import AddColumn, CreateTable, DropColumn, SqlOperation
+
+ENGINE_REGEX = r"^Distributed(\(.+\))$"
+
+
+class InvalidMigrationOrderError(Exception):
+    """
+    Raised when a migration ops are not in the correct order.
+    """
+
+    pass
+
+
+class NoDistClusterNodes(Exception):
+    """
+    The migration has no distributed cluster nodes, so we can't validate the order.
+    """
+
+    pass
+
+
+def validate_migration_order(migration: ClickhouseNodeMigration) -> None:
+    """
+    Validates that the migration order is correct. Chesk that the order of
+    AddColumn, CreateTable and DropColumn operations are correct with reagards to being
+    applied on local and distributed tables.
+    """
+    local_ops = migration.forwards_local()
+    dist_ops = migration.forwards_dist()
+
+    if migration.forwards_local_first:
+        for dist_op in dist_ops:
+            if isinstance(dist_op, DropColumn):
+                if any(
+                    conflicts_drop_column_op(local_op, dist_op)
+                    for local_op in local_ops
+                    if isinstance(local_op, DropColumn)
+                ):
+                    raise InvalidMigrationOrderError(
+                        f"DropColumn {dist_op.table_name}.{dist_op.column_name} operation must applied on dist table before local"
+                    )
+
+    else:
+        for local_op in local_ops:
+            if isinstance(local_op, AddColumn):
+                if any(
+                    conflicts_add_column_op(local_op, dist_op)
+                    for dist_op in dist_ops
+                    if isinstance(dist_op, AddColumn)
+                ):
+                    raise InvalidMigrationOrderError(
+                        f"AddColumn {local_op.table_name}.{local_op.column.name} operation must applied on local table before dist"
+                    )
+
+            if isinstance(local_op, CreateTable):
+                if any(
+                    conflicts_create_table_op(local_op, dist_op)
+                    for dist_op in dist_ops
+                    if isinstance(dist_op, CreateTable)
+                ):
+                    raise InvalidMigrationOrderError(
+                        f"CreateTable {local_op.table_name} operation must applied on local table before dist"
+                    )
+
+
+def conflicts_create_table_op(
+    local_create: CreateTable, dist_create: CreateTable
+) -> bool:
+    """
+    Returns True if create table operation and local create table operation
+    target same underlying local table.
+    """
+    if local_create._storage_set != dist_create._storage_set:
+        return False
+    try:
+        if local_create.table_name == _get_local_table_name(dist_create):
+            return True
+    except NoDistClusterNodes:
+        # If there are no distributed nodes, we can't validate the order.
+        pass
+
+    return False
+
+
+def conflicts_add_column_op(local_add: AddColumn, dist_add: AddColumn) -> bool:
+    """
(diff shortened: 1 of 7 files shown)


snuba/migrations/validator.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+
+    def validate_order(
+        local_ops: Sequence[SqlOperation],
+        dist_ops: Sequence[SqlOperation],
+        local_first: bool,
+    ) -> None:
+        if local_first:
+            for dist_op in dist_ops:
+                if isinstance(dist_op, DropColumn):
```
MeredithAnya, line 50: I think we can make some of this logic a bit more DRY by pulling out the logic for validating local and dist ops

```python
def validate_dist_op(dist_op, local_ops, operation) -> None:
    if not isinstance(dist_op, operation):
        return
    for local_op in local_ops:
        if not isinstance(local_op, operation):
            return
        if conflicts_ops(operation, local_op, dist_op):
            raise InvalidMigrationOrderError(
                f"{operation} {dist_op} \
                operation must be applied on dist table before local"
            )


def validate_local_op(local_op, dist_ops, operation) -> None:
    if not isinstance(local_op, operation):
        return
    for dist_op in dist_ops:
        if not isinstance(dist_op, operation):
            return
        if conflicts_ops(operation, local_op, dist_op):
            raise InvalidMigrationOrderError(
                f"{operation} {dist_op} \
                operation must be applied on local table before dist"
            )
```

You'd have to have some way to repr the operation as a string depending on the op but
this would all you to write something like 

```python
def validate_order(
        local_ops: Sequence[SqlOperation],
        dist_ops: Sequence[SqlOperation],
        local_first: bool,
    ) -> None:
        if local_first:
            for dist_op in dist_ops:
                validate_dist_op(dist_op, local_ops, DropColumn)

        else:
            for local_op in local_ops:
                validate_local_op(local_op, dist_ops, AddColumn)
                validate_local_op(local_op, dist_ops, CreateTable)
```

You'd need to unify the logic for the conflicts function, but you could have mapping from operation to callable or possibly there is a way to make the conflicts_ops encapsulate the logic you have for all three methods right now. They look pretty similar except for create table one

MeredithAnya:
Also thanks for the helpful PR description!! 🎉

snuba/migrations/validator.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Cant parser the distributed engine string value
+    """
+
+    pass
+
+
+def validate_migration_order(migration: ClickhouseNodeMigration) -> None:
+    """
+    Validates that the migration order is correct. Chesk that the order of
```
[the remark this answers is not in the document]
dbanda: reminds me to set up my vscode spell check plugin

snuba/migrations/validator.py
dbanda: Agreed. I have refactored it to be more DRY now. The only issue is the type checker complains so I had to have a few more conditionals to check op type when inside the same function.

snuba/migrations/validator.py
```diff
@@ -0,0 +1,199 @@
...
+    def conflicts_ops(local_op: SqlOperation, dist_op: SqlOperation) -> bool:
+        if isinstance(local_op, CreateTable) and isinstance(dist_op, CreateTable):
+            return conflicts_create_table_op(local_op, dist_op)
+        elif isinstance(local_op, AddColumn) and isinstance(dist_op, AddColumn):
+            return conflicts_add_column_op(local_op, dist_op)
+        elif isinstance(local_op, DropColumn) and isinstance(dist_op, DropColumn):
+            return conflicts_drop_column_op(local_op, dist_op)
+        return False
+
+    def validate_add_col_or_create_table(
```
MeredithAnya, line 52: is there a reason to keep these functions under the `validate_migration_order`? versus pulling them out to be at the top level? That would make it easier to test these functions in unit tests.

dbanda: Two reasons:
1. These functions were very specific to this method and I did want them exposed as part of the module API. It made sense to keep them private.
2. Test should test functionality rather than implementation. That way when we make changes to the implementation (which I anticipate we will do once we identify more conflicting ops), we can use the same tests

snuba/migrations/validator.py
```diff
@@ -0,0 +1,199 @@
...
+
+class DistributedEngineParseError(Exception):
+    """
+    Cant parser the distributed engine string value
+    """
+
+    pass
+
+
+def validate_migration_order(migration: ClickhouseNodeMigration) -> None:
```
fpacifici, line 36: How do you plan to run this validation? During unit tests or when you try to execute the migration in the distributed cluster? The second would be too late. At that point the migration will already be merged and we will not be able to revert it.

dbanda: In unit tests. I have a test here https://github.com/getsentry/snuba/blob/8f986081747206bbac5d12655a0271917c0ef5b0/tests/migrations/test_validator.py#L25-L47 that validates all migrations.

MeredithAnya (approved):

Veteran-ChengQin:
test

Veteran-ChengQin:
@Veteran-ChengQin

Veteran-ChengQin:
@CodiumAI-Agent

Veteran-ChengQin:
@Composio

Veteran-ChengQin:
@Composio

Veteran-ChengQin:
@Veteran-ChengQin

Veteran-ChengQin:
@Veteran-ChengQin hello

PR #3354, opened 2022-11-07
feat(replays): Add additional safety to tags processor

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/194

- Add upper limit to the number of tags accepted per request.
- Enforces non-null string keys.
- Enforces nullable string values.
- Preserves duplicate tag keys (supported by the replays product).
- Skips (instead of erring on) malformed tag tuples.

### DIFF
--- a/snuba/datasets/processors/replays_processor.py
+++ b/snuba/datasets/processors/replays_processor.py
@@ -293,14 +293,15 @@ def process_tags_object(value: Any) -> Tag:
         return Tag.empty_set()
 
     # Excess tags are trimmed.
-    tags = capped_list(normalize_tags(value))
+    tags = capped_list("tags", normalize_tags(value))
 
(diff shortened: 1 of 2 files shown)


evanh (approved):
> Preserves duplicate tag keys (supported by the replays product).

While you can certainly store duplicate tag keys, I don't know that this will actually work in practice. Queries against the tags in Snuba are transformed into something equivalent to  `tags.values[indexOf(tags.keys, <value>)]`. If you are storing duplicate tag keys, you will only ever get the first value in the array when you query the tags.

Otherwise this looks fine to me.

cmanallen:
@evanh Replays uses its own tags aggregation query so this shouldn't be a problem.  We display tags like `{"key": [value1, value2]}`

PR #3358, opened 2022-11-08
ref(admin): update migrations/groups/ endpoint

MeredithAnya:
**context:**
The goal of this PR is to power the new UI added in https://github.com/getsentry/snuba/pull/3341. Instead of just returning the `migration_ids` as a list, we need extra information so I'm changing the API to return a list of objects with the following attributes:

* `migration_id`: the migration id e.b. `0001_migration`
* `status`: one of the following - `not_started`, `in_progess`, `completed`
* `blocking`: whether the migration is blocking or not
* `can_run`: can the user run the migration in question
* `run_reason`: if the user cant run the migration, a string message of why not, otherwise empty string
* `can_reverse`: can the user reverse the migration in question
* `reverse_reason`: if the user cant reverse the migration, a string message of why not, otherwise empty string

**Checkers**
The checkers help determine whether a migration can be run or reversed based on criteria the specific checker has. Checkers are independent of each other but for the specific checkers I've implemented we want the `StatusChecker` to come before the `PolicyChecker`. Really the only thing that matters about the ordering here is the failure reason the user is going to get back.

### DIFF
--- a/snuba/admin/clickhouse/migration_checks.py
+++ b/snuba/admin/clickhouse/migration_checks.py
@@ -29,7 +29,8 @@ class ResultReason:
     reason: Reason
 
 
-class MigrationData(TypedDict):
+@dataclass
+class MigrationData:
     migration_id: str
     status: str
     blocking: bool
@@ -179,22 +180,6 @@ def do_checks(
     return run_result, reverse_result
 
 
-def format_migration_data(
-    details: MigrationDetails,
-    run_result: ResultReason,
-    reverse_result: ResultReason,
-) -> MigrationData:
-    return {
-        "migration_id": details.migration_id,
-        "status": details.status.value,
-        "blocking": details.blocking,
-        "can_run": run_result.allowed,
-        "run_reason": run_result.reason.value,
-        "can_reverse": reverse_result.allowed,
-        "reverse_reason": reverse_result.reason.value,
-    }
-
-
 def checks_for_group(
     migration_group: MigrationGroup, migrations: Sequence[MigrationDetails]
 ) -> Sequence[MigrationData]:
@@ -207,20 +192,16 @@ def checks_for_group(
 
     for details in migrations:
         run_result, reverse_result = do_checks(checkers, details.migration_id)
-        migration_ids.append(format_migration_data(details, run_result, reverse_result))
+        migration_ids.append(
+            MigrationData(
+                migration_id=details.migration_id,
+                status=details.status.value,
+                blocking=details.blocking,
+                can_run=run_result.allowed,
+                can_reverse=reverse_result.allowed,
+                run_reason=run_result.reason.value,
+                reverse_reason=reverse_result.reason.value,
+            )
+        )
 
     return migration_ids
-
-
-def get_migration_ids_data(
-    groups: Sequence[str],
-) -> Sequence[MigrationGroupPayload]:
-    result = []
-    for group, migrations in runner.show_all(groups):
-        migration_ids = checks_for_group(group, migrations)
-        payload: MigrationGroupPayload = {
-            "group": group.value,
-            "migration_ids": migration_ids,
-        }
-        result.append(payload)
-    return result
--- a/snuba/admin/clickhouse/migration_checks.py
+++ b/snuba/admin/clickhouse/migration_checks.py
@@ -5,7 +5,6 @@ from typing import List, Optional, Sequence, Tuple, Union
 
 from snuba.admin.migrations_policies import get_migration_group_polices
 from snuba.migrations.groups import MigrationGroup, get_group_loader
-from snuba.migrations.policies import MigrationPolicy
 from snuba.migrations.runner import MigrationDetails, MigrationKey, Runner
 from snuba.migrations.status import Status
 
@@ -70,11 +69,11 @@ class Checker(ABC):
     """
 
     @abstractmethod
-    def can_run(self, migration_id: str) -> RunResult:
+    def can_run(self, migration_key: MigrationKey) -> RunResult:
         raise NotImplementedError
 
     @abstractmethod
-    def can_reverse(self, migration_id: str) -> ReverseResult:
+    def can_reverse(self, migration_key: MigrationKey) -> ReverseResult:
         raise NotImplementedError
 
 
@@ -106,15 +105,19 @@ class StatusChecker(Checker):
                 "migration_id": migration_id,
                 "status": status,
             }
+        assert self.__all_migration_ids == list(migration_statuses.keys())
         self.__migration_statuses = migration_statuses
 
-    def can_run(self, migration_id: str) -> RunResult:
+    def can_run(self, migration_key: MigrationKey) -> RunResult:
         """
         Covers the following cases:
         * no running a migration that is already pending or completed
         * no running a migration that has a preceeding migration that is not completed
         """
+        group, migration_id = migration_key
+        assert group == self.__group
         all_migration_ids = self.__all_migration_ids
+
         if self.__migration_statuses[migration_id]["status"] != Status.NOT_STARTED:
             return RunResult(False, RunReason.ALREADY_RUN)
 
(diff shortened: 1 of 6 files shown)


snuba/admin/views.py
```diff
@@ -81,20 +83,35 @@ def health() -> Response:
...
+    res: List[Mapping[str, str | Sequence[Mapping[str, str | bool]]]] = []
+    allowed_groups = [
+        group.value
+        for group in get_active_migration_groups()
+        if group.value in settings.ADMIN_ALLOWED_MIGRATION_GROUPS
+    ]
+    if not allowed_groups:
+        return make_response(jsonify(res), 200)
+
+    for group, migrations in runner.show_all(allowed_groups):
```
dbanda, line 95: Seems like a lot of the correctness of this algo depends on the show_all returning things in order. Do we know if that is guaranteed? Maybe we can add a line in docstring showing if that's the case.

snuba/admin/views.py
```diff
@@ -81,20 +83,35 @@ def health() -> Response:
...
+        migration_ids: Sequence[Mapping[str, str | bool]] = [
+            {
+                "migration_id": m.migration_id,
+                "status": m.status,
+                "blocking": m.blocking,
+                "can_run": m.can_run,
+                "can_reverse": m.can_reverse,
```
fpacifici, line 102: Just to confirm, giving details about the reason a migration can/cannot be run to the UI is good (so it can be displayed to the user). Though are you also doing enforcing the check in the API method that actually executes/reverses the migration? That is mandatory to ensure nobody can go around the UI running migrations they do not have permissions to run.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+
+from snuba.admin.migrations_policies import get_migration_group_polices
+from snuba.migrations.groups import MigrationGroup
+from snuba.migrations.policies import MigrationPolicy
+from snuba.migrations.runner import MigrationDetails, MigrationKey
+from snuba.migrations.status import Status
+
+
+class Reason(Enum):
+    NO_REASON_NEEDED = ""
```
fpacifici, line 14: `Reason.no_reason` seems confusing. There is a clear way to identify a case where there should be no reason, which is making the reason optional.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+    ALREADY_RUN = "already run"
+    NOT_RUN_YET = "can't reverse if not already run"
+    NEEDS_SUBSEQUENT_MIGRATIONS = "subsequent migrations must be reversed first"
+    NEEDS_EARLIER_MIGRATIONS = "earlier migrations must run first"
```
fpacifici, line 18: This class is mixing together reasons for running and reversing migrations. It feels like the data model is not representing your states in the right way. You could have a lot of inconsistent states. Like `can_run` returning an instance of `ResultReason.NOT_RUN_YET`.

Since you have two methods: `can_run` and `can_reverse`, You can have two different result types. The fields in the enum has only one value that applies to both.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+
+class Checker(ABC):
+    """
+    A checker is used to encapsulate logic about whether
+    a migration can be run or reversed based on some criteria
+    or state the checker has.
+
+    If a migration can be run/reversed, the ResultReason that
+    is returned should have allowed = True and
+    reason = Reason.NO_REASON_NEEDED.
```
fpacifici, line 48: Why both? `allowed` should be enough. In order to enforce that, please add a post_init method to the `ResultReason` class so that you cannot provide a reason if `allowed == True` and you must provide a reason if `allowed==False`

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+
+    for checker in checkers:
+        reverse_result = checker.can_reverse(migration_id)
+        if not reverse_result.allowed:
+            break
+
+    return run_result, reverse_result
+
+
+def migration_checks_for_group(
```
fpacifici, line 183: nit: if you want to make it clear what a method does, having a method name that does not contain a verb tends to be a bad sign (unless it returns a property). You probably want to rephrase it in a way that states what the method does (like `run_migrations....`.
Also this does not really run the checks for a group as it takes a list of migrations. Nothing says that it will run the checks for the entire group. It would run the checks on. the list of migrations you provide.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+    migration_group: MigrationGroup, migrations: Sequence[MigrationDetails]
+) -> Sequence[MigrationData]:
+    """
+    Responsible for initializing the checkers to be used to
+    validate whether migrations can be run/reversed for a specific
+    migration group.
```
fpacifici, line 189: This is not correct. It also runs the checks.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+    for checker in checkers:
+        reverse_result = checker.can_reverse(migration_id)
+        if not reverse_result.allowed:
+            break
+
+    return run_result, reverse_result
+
+
+def migration_checks_for_group(
+    migration_group: MigrationGroup, migrations: Sequence[MigrationDetails]
```
fpacifici, line 184: Is there a way to get the group from the list of migrations instead of passing both?
When your module provide an interface, you should make it as hard as possible to misuse it. By providing two disconnected structures: a group and a list of migrations you open the door to two potential bugs due to how the method is used:
- provide migrations that do not correspond to the passed group
- provide a migration list that only partially matches the provided group.

Making the method interface restrictive enough that cannot be misused (or at least is capable to spot misuse and reject them) makes your code more robust. Fewer scenarios to test and fewer ways to introduce bugs later on.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+        key = self._get_migration_key(migration_id)
+        if self.__policy.can_reverse(key):
+            return ResultReason(True, Reason.NO_REASON_NEEDED)
+        else:
+            return ResultReason(False, Reason.REVERSE_POLICY)
+
+
+def do_checks(
+    checkers: Sequence[Checker], migration_id: str
+) -> Tuple[ResultReason, ResultReason]:
```
fpacifici, line 158: Please make this a named tuple. 
Imagine you were reading the code that calls `do_checks`, that could be somewhere else in the code. How do you know which one of the field is the result for `run` and which one is the result for `reverse` ? 
Some body will eventually mess up the two while this can be prevented by having a named tuple that clearly states which one is which.

snuba/admin/clickhouse/migration_checks.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class PolicyChecker(Checker):
+    """
+    The PolicyChecker validates whether you can run or
+    reverse a migration based on the policy for the
+    migration's group.
+
+    Policies are defined in the ADMIN_ALLOWED_MIGRATION_GROUPS
+    setting.
```
fpacifici, line 129: Does this class need to exist at all ?
It feels we are maintaining state (the policy and the migration group) for no reason, and by doing so you introduce several possible ways of misusing this class (passing a migration_id that does not correspond to the group you initiualized the `PolicyChecker` with.

What about a two simple methods `can_run`, `can_reverse`. They take in the `MigrationKey`, from there they get the group, from there they load the policies and from there they apply them. 
Messing up becomes much harder.

snuba/admin/views.py
MeredithAnya: so looked into this, and when doing the status checks in the runner we look at the `get_migrations()` from the loader, which makes sense since that's where the order is explicitly defined. I've updated the logic to use to loader so we can make sure the ordering is consistent

snuba/admin/clickhouse/migration_checks.py
MeredithAnya: updated to separate out result types and reason types

snuba/admin/clickhouse/migration_checks.py
MeredithAnya: this is updated now that the result types are different

snuba/admin/clickhouse/migration_checks.py
MeredithAnya: I've updated this method to just take the `allowed_groups` instead so that it can do the logic of getting the statuses using the runner and run the checks for all the groups.

snuba/admin/views.py
MeredithAnya: @fpacifici the `@check_migration_perms` on the run/reverse endpoints checks for the policies and the runner already has checks in place that are the same as the `StatusChecker`

dbanda (approved):

PR #3360, opened 2022-11-08
MIGRATION: add bytes_scanned to querylog

volokluev:
The attribution log contains attribution for internal customers but not for sentry customers (no project_id in attribution log). We want to be able to analyze bytes scanned by sentry customers as well. Hence we need to add this column

### DIFF
--- a/snuba/snuba_migrations/querylog/0004_add_bytes_scanned.py
+++ b/snuba/snuba_migrations/querylog/0004_add_bytes_scanned.py
@@ -1,0 +1,55 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Array, Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Adds fields for query profile.
+    """
+
+    blocking = True
+
+    def __forward_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.QUERYLOG,
+                table_name=table_name,
+                column=Column(
+                    "clickhouse_queries.bytes_scanned",
+                    Array(
+                        UInt(64),
+                        Modifiers(
+                            default="arrayResize([0], length(clickhouse_queries.sql))"
+                        ),
+                    ),
+                ),
+                after="clickhouse_queries.array_join_columns",
+            ),
+        ]
+
+    def __backwards_migrations(
+        self, table_name: str
+    ) -> Sequence[operations.SqlOperation]:
+        return [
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/querylog/0004_add_bytes_scanned.py
```diff
@@ -0,0 +1,55 @@
...
+                storage_set=StorageSetKey.QUERYLOG,
+                table_name=table_name,
+                column=Column(
+                    "clickhouse_queries.bytes_scanned",
+                    Array(
+                        UInt(64),
+                        Modifiers(
+                            default="arrayResize([0], length(clickhouse_queries.sql))"
+                        ),
+                    ),
```
fpacifici, line 30: This will not be this simple.
If I applied this migration the way it is, the consumer would break.
The consumer would try write rows in the `clickhouse_queries` nested column that would not be of the same size as the `bytes_scanned` array would not be passed.
As such Clickhouse will consider this array as passed empty.

The migration has to happen in the consumer first passing the column first. You will have to set the `input_format_skip_unknown_fields` Clickhouse settings in your write so Clickhouse will ignore the unknown column. Then we can apply the migration.

This was the example https://github.com/getsentry/snuba/pull/2232/files

fpacifici (changes requested):

snuba/snuba_migrations/querylog/0004_add_bytes_scanned.py
```diff
@@ -0,0 +1,55 @@
...
+            operations.AddColumn(
+                storage_set=StorageSetKey.QUERYLOG,
+                table_name=table_name,
+                column=Column(
+                    "clickhouse_queries.bytes_scanned",
+                    Array(
+                        UInt(64),
+                        Modifiers(
+                            default="arrayResize([0], length(clickhouse_queries.sql))"
+                        ),
```
fpacifici, line 29: You should be able to remove this as you are now writing the actual empty array.

fpacifici (approved):

snuba/snuba_migrations/querylog/0004_add_bytes_scanned.py
volokluev: But not all rows have this column written. Doesn't it need to exist until that's no longer the case?

fpacifici: If not all rows have this column written, please hold off, that would break the consumer. 

In a nested column all rows MUST have all columns written. When you write a nested column you write an array per column. The arrays MUST have the same length or Clickhouse will reject the write and the consumer will get stuck.

fpacifici: No, because the column does not exist in the DB. 
The critical part is that you deploy the change that writes the column for all rows before we apply the migration. 

The default here proved to be useless right after we added it back in 2020. 
It was a desperate move to make a migration work without having to stop the consumer. We did not know about `input_format_skip_unknown_fields` back then.

PR #3367, opened 2022-11-09
feat(DC): More inner types for Array

rahul-kumar-saini:
### Overview
- Currently, the JSON Schema and Parser assume `Array` Column types will only have a `Float` or `UInt` inner type
- This was sufficient for many datasets but as we work on more configs, we need to add support for all types of Arrays
- This PR modifies the Array schema to now accept recursive inner types
    - The validation limits this to up to 1 layer of recursion, eg `Array(Array(String()))`
    - This is good enough as I haven't seen any instances of `Array(Array(Array(...)))`
    - The parser technically has no limit, it supports as much recursion as needed but the validation will block this

### Code level changes
- Updated JSON Schema
- Updated parser to support additional recursive inner types for Array
- Updated all existing configs with Array's to use the new format
- Modified `parse_columns()` function to split out responsibility into sub functions which deal solely with `ColumnType` while the main function takes care of building the `Column` object itself

### Blast radius
- Dataset Configs

### Notes
- As far as I know, the documentation surrounding JSON Schema _should_ auto update but I will do it manually if I'm wrong

### DIFF
--- a/snuba/datasets/configuration/utils.py
+++ b/snuba/datasets/configuration/utils.py
@@ -95,7 +95,7 @@ def get_mandatory_condition_checkers(
     ]
 
 
-NUMBER_COLUMN_TYPES: dict[str, Type[ColumnType[SchemaModifiers]]] = {
+NUMBER_COLUMN_TYPES: dict[str, Any] = {
     "UInt": UInt,
     "Float": Float,
 }
@@ -119,9 +119,9 @@ def __parse_simple(
 def __parse_number(
     col: dict[str, Any], modifiers: SchemaModifiers | None
 ) -> ColumnType[SchemaModifiers]:
-    col_type = NUMBER_COLUMN_TYPES[col["type"]]
+    col_type = NUMBER_COLUMN_TYPES[col["type"]](col["args"]["size"], modifiers)
     assert isinstance(col_type, UInt) or isinstance(col_type, Float)
-    return col_type(col["args"]["size"], modifiers)
+    return col_type
 
 
 def __parse_column_type(col: dict[str, Any]) -> ColumnType[SchemaModifiers]:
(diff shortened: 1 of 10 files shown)


snuba/datasets/configuration/utils.py
```diff
@@ -109,60 +112,50 @@ def get_mandatory_condition_checkers(
...
 def __parse_number(
     col: dict[str, Any], modifiers: SchemaModifiers | None
-) -> Column[SchemaModifiers]:
-    return Column(
-        col["name"], NUMBER_COLUMN_TYPES[col["type"]](col["args"]["size"], modifiers)
-    )
+) -> ColumnType[SchemaModifiers]:
+    col_type = NUMBER_COLUMN_TYPES[col["type"]]
+    # UInt and Float ColumnType accepts a size parameter
+    return col_type(col["args"]["size"], modifiers)  # type: ignore
```
evanh, line 124: Why the type ignore here?

evanh (commented):
Mostly looks good, but this should have a test.

snuba/datasets/configuration/utils.py
rahul-kumar-saini: Base `ColumnType` does not accept a `size` parameter like `Float` and `UInt` do so mypy complains that `ColumnType` was given too many parameters since `NUMBER_COLUMN_TYPES` is a `dict[str, ColumnType]`. I could add some assertion that `col_type` is a `Float` or `UInt` instead of the type ignore if that's a better way to do this.

evanh: Yeah I would rather do that.

evanh (approved):
Looks good, besides small typing fix.

PR #3370, opened 2022-11-10
fix(init): uWSGI Snuba API processes not initializing Snuba

rahul-kumar-saini:
### Overview
- When uWSGI starts a new Snuba API process, it simply starts the Flask app
- This is a problem as the first query which hits this new process will force Snuba to load all datasets/entities/storages from config before the query can be executed
- This is causing timeouts as queries which normally take <1s will randomly take >5s to execute

### Before State
- First query to any Snuba API process loads all datasets

### After State
- Datasets are loaded before the Snuba API is declared "ready" by uWSGI

### Blast Radius
- Snuba API
- Initialization is more explicit once again

### Testing Notes
- Tested locally to verify the old version forces datasets to load on first query and now it does not 👨‍🔬

### DIFF
--- a/snuba/core/initialize.py
+++ b/snuba/core/initialize.py
@@ -29,8 +29,8 @@ def _load_entities() -> None:
     initialize_entity_factory()
 
 
-def initialize() -> None:
-    logger.info("Initializing snuba")
+def initialize_snuba() -> None:
+    logger.info("Initializing Snuba")
 
     # The order of the functions matters The reference direction is
     #
--- a/snuba/web/wsgi.py
+++ b/snuba/web/wsgi.py
@@ -1,6 +1,8 @@
+from snuba.core.initialize import initialize_snuba
 from snuba.environment import setup_logging, setup_sentry
 
 setup_logging()
 setup_sentry()
+initialize_snuba()
 
 from snuba.web.views import application  # noqa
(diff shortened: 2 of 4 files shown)


snuba/core/initialize.py
```diff
@@ -29,8 +29,8 @@ def _load_entities() -> None:
     initialize_entity_factory()
 
 
-def initialize() -> None:
-    logger.info("Initializing snuba")
+def initialize_snuba() -> None:
```
rahul-kumar-saini, line 32: Simple rename to make it easier to understand the usage of this function outside of this module.

snuba/web/wsgi.py
```diff
@@ -1,6 +1,8 @@
+from snuba.core.initialize import initialize_snuba
 from snuba.environment import setup_logging, setup_sentry
 
 setup_logging()
 setup_sentry()
+initialize_snuba()
```
rahul-kumar-saini, line 6: This is the core change, the `snuba/web/wsgi.py` file is loaded by uWSGI to start the API process.

volokluev, line 6: not related to this ticket but I'm starting to think all of these should be in the `initialize` function

volokluev (approved):

PR #3371, opened 2022-11-10
feat(admin): add migrations dry run to UI sns-1867

dbanda:
Connects the forwards and backwards buttons of the UI to call the dry run endpoints. Removes the dummy code and actually makes dry run calls to the backend.


evanh (approved):
Was there any UI changes? If so, it would be nice to include screenshots.

snuba/admin/static/clickhouse_migrations/index.tsx
```diff
@@ -61,6 +60,28 @@ function ClickhouseMigrations(props: { api: Client }) {
...
+      console.assert(req.dry_run, "dry_run must be set")
+      props.api
+      .runMigration(req as RunMigrationRequest)
+      .then((res) => {
+        console.log(res)
+        setSQLText(() => res.stdout);
+      })
+      .catch((err) => {
+        console.log(err)
+        setSQLText(() => JSON.stringify(err));
```
MeredithAnya, line 80: we should be able to do `err.error` should we not?

dbanda: This is a more generic error type and can include both client and server side response errors. For server side we can do `err.error`, client side we cannot.

dbanda:
> Was there any UI changes? If so, it would be nice to include screenshots.

No real changes changes to the UI. Instead of dummy sql, we now show proper real dry-run SQL.

PR #3372, opened 2022-11-11
ref: Remove unused v1 query cache

untitaker:
This does not seem to be used anywhere, the new readthrough cache is
enabled by default (and in prod) since 2020

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -13,7 +13,7 @@ from clickhouse_driver import errors
 from sentry_sdk import Hub
 from sentry_sdk.api import configure_scope
 
-from snuba import environment, settings, state
+from snuba import environment, state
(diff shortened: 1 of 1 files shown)


evanh (approved):
I'm happy to see this code deleted, but I assumed it had been left in for this long since potentially self hosted single tenant might be using it.

I don't think the two solutions require different infrastructure though so we should be fine.

untitaker:
i haven't forgotten about this PR, just can't figure out why the tests fail atm. they pass locally

evanh (approved):
Is it possible for a self hosted customer to have this old version caching enabled, and then when this change is merged they stop caching all together? Or will they automatically fail over to readthrough caching?

volokluev:
@untitaker I actually also stumbled on this code and tried to remove it. I figured out the test failure as well and will have a PR up by EOD today

PR #3373, opened 2022-11-11
fix(query cache): Retry query with randomized ID if cache fails

untitaker:
Figured I'd open this without tests just to get some feedback on the
approach.

It seems that ClickHouse rejects the second of two queries that have the
same query ID. The query ID is directly computed from the formatted
query contents itself, so two identical queries hitting clickhouse
yields errors.

This is typically not a problem as we use a Redis cache to deduplicate
queries. However, that Redis cache can fail and lose data, such as
during a failover scenario or when I am specifically trying to migrate
the cache from one Redis cluster to another without doing data
migration.

I think something like this is the sanest attempt to avoid rejecting
valid queries without significantly affecting what query IDs mean (in
querylog for example) most of the time. A significantly simpler approach
would be to randomize the query ID that is being passed to clickhouse
while still using the formatted query for the cache key. I don't see a
strong reason for not doing that right now but it might affect auditing
when trying to correlate cache keys in redis with the query IDs in
querylog.

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import uuid
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial, reduce
 from hashlib import md5
@@ -511,7 +512,7 @@ def execute_query_with_caching(
 
 
 @with_span(op="db")
-def execute_query_with_readthrough_caching(
+def execute_query_with_query_id(
     clickhouse_query: Union[Query, CompositeQuery[Table]],
     query_settings: QuerySettings,
     formatted_query: FormattedQuery,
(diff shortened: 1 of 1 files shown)


evanh:
The risk I see with this approach is if it starts happening a lot, we could be hugely increasing the load on Clickhouse. It feels like in the case of missing data, we should be able to recover the cache. If a query gets the duplicate query ID error, that means there is another query that will return the results, even if the cache key is missing. We could recreate the cache key and wait for those results I think.

- Query A comes in, sets the cache key X and starts executing on Clickhouse
- Cache key X gets deleted
- Query B comes in, sees there is no cache key. It will set cache key X, and then start executing
- Query B will get a duplicate query ID error.
- At this point, it seems like we could tell query B to just wait for results on cache key X
- Query A finishes, writes results to cache key X
- Query B reads results from cache key X

untitaker:
One concern I see is that if Redis never persists writes successfully, this could lead to hanging queries. I am not sure if that's any worse than current behavior though.

evanh:
> One concern I see is that if Redis never persists writes successfully, this could lead to hanging queries.

This feels like a decision we need to make about what user experience we want. 

1. We can send the queries to Clickhouse. That will cause extra load on Clickhouse and/or Clickhouse will start rate limiting queries. However it's possible some queries will continue to work.
2. The queries fail/timeout. That will mitigate the load on Clickhouse but ingest etc. will probably be broken.

volokluev:
I think this is a fine approach as long as we:

1. Have a monitor on it to know when it is happening a lot
2. Have a way to kill this in runtime config. I imagine it's not necessary most of the time but can be useful when we're doing a redis migration

untitaker:
@volokluev I think that's a good idea. honestly I didn't want to spend much time on this because I'm trying to unblock redis cluster migration, didn't get anywhere yesterday while trying to understand our cache layer

I now log an error (translating into a sentry error when this happens), and have the entire behavior hidden behind a setting defaulting to "off"

fpacifici:
>  A significantly simpler approach
would be to randomize the query ID that is being passed to clickhouse
while still using the formatted query for the cache key. I don't see a
strong reason for not doing that right now but it might affect auditing
when trying to correlate cache keys in redis with the query IDs in
querylog.

We did rely on grouping by query id to audit and troubleshooting the Snuba cache/dedupe during incidents in the past. So randomization by default is undesirable.

fpacifici:
> It feels like in the case of missing data, we should be able to recover the cache. If a query gets the duplicate query ID error, that means there is another query that will return the results, even if the cache key is missing. We could recreate the cache key and wait for those results I think.

@evanh  Not necessarily. Redis replication is asynchronous. It is possible that the primary query could create the queue in redis for the duplicates, then redis does a hard failover without that queue ever being replicated. A second query could receive the rejection from Clickhouse. Now the duplicates queue does not exist in Redis on the new primary so there is no way to wait for the results. 

Worse it is possible that the primary thread may never be able to write the results in cache if it still tries to write on the redis primary.

fpacifici (commented):
The approach seems sane to me. I think adding a killswitch as Volo suggest is a good idea

evanh (approved):
Yeah this looks good to me too. Thanks!

PR #3376, opened 2022-11-11
feat(api-abuse): Add ability to throttle threads for an entire referrer

volokluev:
In case a bad feature ships. We want to throttle an entire referrer, not just a specific project's threads

### DIFF
--- a/snuba/query/processors/logical/quota_processor.py
+++ b/snuba/query/processors/logical/quota_processor.py
@@ -41,7 +41,7 @@ class ResourceQuotaProcessor(LogicalQueryProcessor):
             return
 
         # TODO: Like for the rate limiter Add logic for multiple IDs
-        project_id = str(project_ids.pop())
+        project_id = str(project_ids.pop()) if project_ids else "NO_PROJECT_ID"
         project_referrer_thread_quota = get_config(
             f"{REFERRER_PROJECT_CONFIG}_{query_settings.referrer}_{project_id}"
         )
--- a/snuba/query/processors/logical/quota_processor.py
+++ b/snuba/query/processors/logical/quota_processor.py
@@ -7,11 +7,22 @@ from snuba.state.quota import ResourceQuota
 
 ENABLED_CONFIG = "resource_quota_processor_enabled"
 REFERRER_PROJECT_CONFIG = "referrer_project_thread_quota"
+REFERRER_CONFIG = "referrer_thread_quota"
 
 
 class ResourceQuotaProcessor(LogicalQueryProcessor):
     """
-    Applies a referrer/project thread quota to the query.
+    Applies a referrer/project thread quota to the query. Can throttle a referrer
+    or a (referrer, project) pair. The more specific restriction takes precedence
+
+    Example:
+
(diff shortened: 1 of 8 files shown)


nikhars (commented):
When I look at how resource quota is used, this is the code where it is being used https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L343-L348. There is and condition which checks against `project_rate_limit_stats` and the concurrent parameter being greater than 1.

My question is if we apply only the referrer based rate limit settings, will the code where it is being used still be triggered? Maybe worth adding a test on that method to find out.

snuba/query/processors/logical/quota_processor.py
```diff
@@ -7,11 +7,22 @@
...
+    Applies a referrer/project thread quota to the query. Can throttle a referrer
+    or a (referrer, project) pair. The more specific restriction takes precedence
+
+    Example:
+
+        SET referrer_thread_quota_MYREFERRER = 20
+        # all requests with referrer = MYREFERRER are now capped at 20 threads
+
+        SET referrer_project_thread_quota_MYREFERRER_1337 = 2
+        # all requests with referrer = MYREFERRER, project_id = 1337 are now
```
evanh, line 24: I think there should be a test to confirm this behaviour.

evanh (commented):
Looks good to me, some extra tests required though.

snuba/query/processors/logical/quota_processor.py
```diff
@@ -7,11 +7,22 @@
...
 ENABLED_CONFIG = "resource_quota_processor_enabled"
 REFERRER_PROJECT_CONFIG = "referrer_project_thread_quota"
+REFERRER_CONFIG = "referrer_thread_quota"
 
 
 class ResourceQuotaProcessor(LogicalQueryProcessor):
     """
-    Applies a referrer/project thread quota to the query.
+    Applies a referrer/project thread quota to the query. Can throttle a referrer
+    or a (referrer, project) pair. The more specific restriction takes precedence
```
onewland, line 16: I wonder if precedence should be the minimum number of max threads. If `REFERRER_{X}` is capped at 2 threads globally, it doesn't seem like `REFERRER_{X}_PROJECT_{Y}` should be allowed to have 10.

Though in practice I'm guessing people won't tend to mess this up

nikhars (approved):
LGTM

snuba/web/db_query.py
```diff
@@ -317,6 +318,25 @@ def _record_rate_limit_metrics(
         )
 
 
+def _apply_thread_quota_to_clickhouse_query_settings(
+    query_settings: QuerySettings,
+    clickhouse_query_settings: MutableMapping[str, Any],
+    project_rate_limit_stats: Optional[RateLimitStats],
+) -> None:
```
onewland, line 325: Not a big deal because this method is private, but I think it would be nice (as a convention) to return a new `QuerySettings` with `max_threads` set and not modify the input.

onewland (approved):
Changes look good to me, but I have a low level of familiarity with our rate limiting code

PR #3383, opened 2022-11-14
feat(DC): Replays Storage as Config

rahul-kumar-saini:
### Overview
- Adding Replays storage as a config
- Storage config was generated and tested using scripts from #3336 

### Changes
- Any call to `get_storage(StorageKey.REPLAYS)` will return the replays storage built from config instead of the hardcoded version

### Testing Notes
- How config was generated:
    - `python3 scripts/pystorage_2_yaml.py replays replays.yaml`
    - Removed quotes around `columns` in schema, format yaml file with Prettier
- How config was tested:
    - `python3 scripts/check_yaml_against_code.py replays replays.yaml`
    - Updated `tests/datasets/configuration/test_storage_loader.py`

### DIFF
--- a/snuba/datasets/configuration/replays/storages/replays.yaml
+++ b/snuba/datasets/configuration/replays/storages/replays.yaml
@@ -1,0 +1,120 @@
+version: v1
+kind: writable_storage
+name: replays
+storage:
+  key: replays
+  set_key: replays
+schema:
+  columns:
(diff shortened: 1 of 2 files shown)


volokluev (approved):

volokluev:
Let's deploy these one at a time to start

PR #3384, opened 2022-11-15
ref(entity subscription): Modularize Entity Subscriptions

enochtangg:
## Overview
Currently, Entity Subscriptions contain configuration that is defined separately from the Entity. As a result, there is a separate factory that manages Entity Subscriptions. We would like to remove this factory and move this configuration to be defined in the Entity. 

However, Entity Subscriptions is not only composed of configuration. For each entity, there may be specific methods and unique parameters in its constructor (e.g. `org_id`). More importantly, we want to provide code isolation among entity subscriptions. For example, subscriptions that have no organization (`transactions`, `errors`) should not have an empty organization key in Redis. 

Therefore, we need a design which allows us to move what is defined in config to the entity config while preserving those unique parameters for each subscription.

## Requirements and Constraints
* Must be able to define all entity subscription related configuration in `Entity` configuration
* Must preserve functionality of existing entity subscription methods
* Entity subscription data must be isolated from each other in code
* Must not contain new data migrations (Redis should not be altered)

## Before State
* Entity subscription and Entity components were decoupled
* Entity subscription factory loads mapping of `EntityKey` and `Type[EntitySubscription]`.
* `EntitySubscription` objects are instantiated for every new subscription due to query specific params which were set in `EntitySubscription` (e.g. `organization_id`)
* A different `EntitySubscription` class for each entity
* Unique methods were defined in all entity subscription class to accomplish one-off tasks. This meant that all entity subscription classes shared some similar interface. In an ideal scenario, different entity subscriptions should be allowed to specific their own isolated methods and attributes without need to know other the data of other entity subscriptions.

## After State
* Remove entity subscription factory.
* Entity Subscriptions is split into `subscription_processors` and `subscription_validators`.
* The `EntitySubscription` now adopts a design pattern similar to `TranslationMapper`.  We can now specify a list of processors and/or validators (searched up by class name in config) which individually does some work on the subscription query. This modularizes entity subscriptions and enables us to defined configuration and extra keys in an isolated manner among entities.
* Subscription processors and validators are loaded from config only once. We can do this because we have removed query specific attributes on the `EntitySubscription` and set that in the `SubscriptionData` instead. (we call this `metadata` which would contain keys like `organization` or any other new keys product owners would want in the future)

```
class Entity(Describable, ABC):
    def __init__(
        self,
        ...
        subscription_processors: Optional[Sequence[EntitySubscriptionProcessor]],
        subscription_validators: Optional[Sequence[EntitySubscriptionValidator]],
```

## Blast Radius
* All entity configurations now can optionally require the `subscription_processors` and `subscription_validators` keys.
* The existing functionality of entity subscriptions are unchanged, but the way we define them have. 
* Additionally, the way subscriptions are encoded and stored in Redis are unchanged. 
* Although a metadata attribute is added to SubscriptionData, it does not get encoded. 

## Testing Notes
* New unit tests added under `tests/subscriptions/entity_subscriptions/`


ayirr7:
This is more of a question, but am I correct in my understanding that: for Subscriptions that have corresponding `EntityKeys` that fall under (metrics_sets, metrics_counters, metrics_distributions), there will be an additional field called "organization" in the `SubscriptionData`'s `metadata`?

enochtangg:
> This is more of a general question, but am I correct in my understanding of this approach that: for Subscriptions that have corresponding `EntityKeys` that fall under (metrics_sets, metrics_counters, metrics_distributions), there will be an additional field called "organization" in the `SubscriptionData`'s `metadata`?

That is correct. This is similar to the way it currently exists except right now, organization is placed as a top level key in `SubscriptionData`. With this change, any extra keys will be stored inside `metadata`

> Are there any other `EntityKeys` that would?
Yes, there are the ones you mentioned plus sessions and generic_metrics_sets. Currently, they are defined [here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/entity_subscriptions/entity_subscription.py)

lynnagara:
I personally find a PR of this size and scope very difficult to review thoroughly. Please consider breaking it up into smaller chunks. As an example of how this could be split, you may want to do all the refactoring for the class separately to translating everything to yaml as well.

snuba/datasets/configuration/json_schema.py
```diff
@@ -444,6 +482,15 @@ def registered_class_schema(
             "type": ["string", "null"],
             "description": "The column name, if this entity is partitioned, to select slice",
         },
+        "entity_subscription": {
+            "type": "object",
+            "description": "Specifies the processors and validators associated with subscriptions",
+            "properties": {
+                "processors": ENTITY_SUBSCRIPTION_PROCESSORS,
+                "validators": ENTITY_SUBSCRIPTION_VALIDATORS,
```
lynnagara, line 490: This pattern of separate processors and validators seems like a pretty significant departure from what was discussed offline (having linked processors/validators so that a single class would both validate the presence of the org ID then include it into the query). Reasons for the change?

enochtangg: I think there might have be some slight confusion with the selection of splitting or combining `processors` and `validators` offline. The problem with combining `processors` and `validators` into a single "processor" class is because currently, the `EntitySubscription` validation and query processing are two completely separate things. We currently do not validate org ID. The only validation we do right now is on the aggregation clauses of the query. On the contrary, we validate the aggregation but this does not mean we do some processing on the aggregation portion of the query. As a result, in the current state of the world, these two are separate components which live separately.

enochtangg:
> As an example of how this could be split, you may want to do all the refactoring for the class separately to translating everything to yaml as well.

The two are quite dependent because we already run dataset as configuration in production today.

> Please consider breaking it up into smaller chunks.

This refactor is quite difficult to split up without having to write a good amount of placeholder code. Doing that takes up a lot of time due to updating the substantial amount of subscriptions unit tests and updating each entity configuration. Also, I personally don’t see much value in writing placeholder code just so that it can be removed in a few days especially when the primary stakeholders are aware of the implemented solution from the proposal.

Notes: 
* The slicing work could use some of the new SubscriptionData `metadata` stuff implemented here which makes this time sensitive. 
* I’m not strictly against this being split, but I want to be clear that it is a lot of work which I’m not sure is necessary given the involvement in the proposal.

snuba/datasets/configuration/json_schema.py
enochtangg: I think there might be some slight confusing because I thought the option of splitting `processors` and `validators` was selected offline. The problem with combining processors and validators into a single processor class is because in the current state we run `EntitySubscriptions`, validation and query processing are two completely separate things. We currently do not validate org ID. We validate that the aggregation clauses of the query follow some specification in the query. On the contrary, we validate the aggregation but this does not mean we do some processing on the aggregation portion of the query. As a result, in the current state of the world, these two are separate components which live separately.

lynnagara:
@enochtangg Ultimately it is your call how to stage changes - it's just a suggestion to make it more reviewable as I tend to miss things in big PRs.

- I do think there are other ways this change can be pretty easily split. If you want to do Python + yaml entities in one go, that's fine. Another way would be to introduce all your new processor and validator classes exactly as they are here in one PR and but just call them from the existing EntitySubscription classes. Then in a subsequent change you can deprecate the EntitySubscription class and modify the callsites to reference those processors and validators directly
- About the `metadata` changes that make this PR time sensitive - similar question. Why is it bundled together here? This seems totally separate and makes this PR bigger and more complex

tests/subscriptions/test_subscription.py
```diff
@@ -296,6 +386,12 @@ def test(self) -> None:
             time_window_sec=10 * 60,
             resolution_sec=60,
             entity_subscription=create_entity_subscription(),
+            metadata={
+                "project_id": 1,
+                "time_window": 10 * 60,
+                "resolution": 60,
+                "query": "MATCH (events) SELECT count() AS count",
```
lynnagara, line 393: Why are these existing fields duplicated here? This isn't additional metadata - it's already in the SubscriptionData

enochtangg: Thanks, updated to remove duplicated fields in metadata. The data received from payload strips away `SubscriptionData` fields and leaves the extra keys. https://github.com/getsentry/snuba/pull/3384/files#diff-8e3672e3fdccd3ce17c850cd47dfe0b41f53feabd7dcf550c30077fe3f7e31fbR180-R183

snuba/datasets/entity_subscriptions/validators.py
```diff
@@ -0,0 +1,58 @@
...
+        self, max_allowed_aggregations: int, disallowed_aggregations: Sequence[str]
+    ):
+        self.max_allowed_aggregations = max_allowed_aggregations
+        self.disallowed_aggregations = disallowed_aggregations
+
+    def to_dict(self) -> Mapping[str, Any]:
+        return {}
+
+    def validate(self, query: Union[CompositeQuery[Entity], Query]) -> None:
+        from snuba.datasets.entities.factory import get_entity
```
evanh, line 47: I assume this is avoiding some kind of circular import? What was the actual problem? Generally when this happens one of two things is wrong:

1) A circular import is being caused solely by typing (the easy case). We can use the `TYPE_CHECKING` constant in the imports to avoid that case.
2) Our code is organized poorly (the hard case). We need to move code around to fix this.

snuba/subscriptions/data.py
```diff
@@ -181,16 +189,23 @@ def from_dict(
...
             "resolution": self.resolution_sec,
             "query": self.query,
-            **self.entity_subscription.to_dict(),
         }
+        if self.entity_subscription.processors:
+            for processor in self.entity_subscription.processors:
+                subscription_data_dict.update(processor.to_dict(self.metadata))
+        if self.entity_subscription.validators:
+            for validator in self.entity_subscription.validators:
+                subscription_data_dict.update(validator.to_dict())
```
evanh, line 207: I think the validator should also be passed the metadata. I know we don't use it currently but that will probably change.

tests/subscriptions/entity_subscriptions/test_entity_subscriptions.py
```diff
@@ -1,158 +1,296 @@
...
-        True,
-        id="Metrics sets subscription of type SNQL",
+        ),
+        {},
+        InvalidQueryException,
+        id="Metrics sets subscription",
     ),
 ]
 
+# TESTS_CONDITIONS_SNQL_METHOD = [
```
evanh, line 195: Why are these commented out?

tests/subscriptions/entity_subscriptions/test_entity_subscriptions.py
```diff
@@ -1,158 +1,296 @@
...
-    has_offset: bool,
+
+@pytest.mark.parametrize("entity_key, query, metadata, exception", TESTS)
+def test_entity_subscription_validators(
+    entity_key: EntityKey,
+    query: Union[CompositeQuery[QueryEntity], Query],
+    metadata: Mapping[str, Any],
+    exception: Optional[Type[Exception]],
 ) -> None:
-    offset = 5 if has_offset else None
```
evanh, line 154: This was testing the offset conditions being correctly added. Are there still tests for that somewhere?

tests/subscriptions/entity_subscriptions/test_entity_subscriptions_data.py
```diff
@@ -0,0 +1,86 @@
...
+        entity_subscription=entity_subscription,
+        query=query,
+        metadata=metadata,
+    )
+
+
+@pytest.fixture
+def entity_subscription() -> EntitySubscription:
+    entity_subscription = entity.get_entity_subscription()
+    assert isinstance(entity_subscription, EntitySubscription)
```
evanh, line 52: Why is this a fixture? `entity` is already a global, why not this too?

tests/subscriptions/test_subscription.py
```diff
@@ -126,6 +128,17 @@ def test_invalid_time_window(self) -> None:
                     time_window_sec=0,
                     resolution_sec=60,
                     entity_subscription=create_entity_subscription(),
+                    metadata={
+                        "project_id": 123,
```
evanh, line 132: Why is all this in the metadata?

evanh:
This generally looks good to me, some comments/questions though.

Also @lynnagara is correct, this PR is too big. If your PR is more than 500 lines of additions, then it's almost certainly too big. Next time, if you really want all the changes in one PR, you can break up your changes in the commits. For example, have one commit for the code changes, and a separate commit for the test changes. That can make it easier to review.

snuba/subscriptions/data.py
```diff
@@ -161,18 +166,21 @@ def build_request(
...
     def from_dict(
         cls, data: Mapping[str, Any], entity_key: EntityKey
     ) -> SubscriptionData:
-        entity_subscription = get_entity_subscription(entity_key)(data_dict=data)
+        entity_subscription = get_entity(entity_key).get_entity_subscription()
+        assert entity_subscription is not None
+
+        metadata = {}
+        for key in data.keys():
+            if key not in {"project_id", "time_window", "resolution", "query"}:
```
nikhars, line 182: Can you define these hardcoded keys as a constant with an explanation of why they are not part of metadata?

.vscode/settings.json
```diff
@@ -61,5 +61,8 @@
   // tests is often in your way and misclicked
   "python.testing.pytestEnabled": true,
   "python.testing.unittestEnabled": true,
-  "python.testing.pytestArgs": ["tests"]
+  "python.testing.pytestArgs": [
+    "tests"
+  ],
+  "esbonio.sphinx.confDir": ""
```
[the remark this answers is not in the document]
enochtangg: Must have been my formatter. Updated.

tests/subscriptions/entity_subscriptions/test_entity_subscriptions.py
enochtangg: These were stale which were already captured by the new validator and processor tests. Removed.

tests/subscriptions/entity_subscriptions/test_entity_subscriptions.py
enochtangg: The previous tests didn't actually check anything because offset was not used. So all tests with or without offset would just resolve which is why I removed it. Added it back in here in case we use/test offset in the future.

tests/subscriptions/entity_subscriptions/test_entity_subscriptions_data.py
enochtangg: Updated to global

snuba/datasets/entity_subscriptions/validators.py
```diff
@@ -0,0 +1,55 @@
...
+class AggregationValidator(EntitySubscriptionValidator):
+    def __init__(
+        self,
+        max_allowed_aggregations: int,
+        disallowed_aggregations: Sequence[str],
+        required_time_column: Optional[str] = None,
+    ):
+        self.max_allowed_aggregations = max_allowed_aggregations
+        self.disallowed_aggregations = disallowed_aggregations
+        self.required_time_column = required_time_column
```
enochtangg, line 45: > I assume this is avoiding some kind of circular import?

@evanh As discussed, added `required_time_column` as a param to AggregationValidator. This is used to fix the cyclic dependency.
Before: AggregationValidator uses entity factory to get required_time_column in entity. Causes cycle.
Now: `required_time_column` is specified in entity subscription config and used directly

snuba/datasets/entity_subscriptions/validators.py
enochtangg: Replied below.

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
```diff
@@ -101,3 +101,16 @@ validators:
...
+      validator: AggregationValidator
+      args:
+        max_allowed_aggregations: 3
+        disallowed_aggregations: [having, orderby]
+  processors:
+    -
+      processor: AddColumnCondition
+      args:
+        extra_condition_data_key: organization
+        extra_condition_column: org_id
```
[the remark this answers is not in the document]
enochtangg: There is some implicit semantic validation done when the system tries to build the entity. For entity subscription processors and validators, this happens [here](https://github.com/getsentry/snuba/pull/3384/files#diff-530c557e763d4aa8baeb0c82eb931da0d8e7206d886d3b266357330ed9506145R100). The builder looks up the class name by a base class config. As a result, there won't be a misconfiguration when compiling it. However, right now, we have no cues to inform users about these types of misconfiguration before runtime. The `make validate-configs` command only checks that the schema is congruent to json_schema definition (no semantic validation).

snuba/datasets/configuration/json_schema.py
lynnagara: If they are completely separate components why do they overlap in the `to_dict()` method. The fact that the validator can provide keys to be written to redis via `to_dict()` feels like an anti-pattern. It seems like it might be doing a lot more than validating.

Also, although we are not currently validating the org ID, shouldn't we be? That feels like it might be an oversight

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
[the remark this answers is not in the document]
lynnagara: If entity validations are processors are separate as you say (which I am not sure about b/c of https://github.com/getsentry/snuba/pull/3384#discussion_r1025863707), I don't really think you need the entity subscription parent at all.

Why not just have:
```
subscription_validators:
  - validator
subscription_processors:
  - processor

The entity subscription parent and the entity prefix serves no value anymore since we are already within the context of an entity

File: snuba/datasets/entity_subscriptions/processors.py [hunk 13]
```diff
@@ -0,0 +1,72 @@
...
+        if self.extra_condition_data_key not in metadata:
+            raise InvalidQueryException
+        condition_to_add: Sequence[Expression] = [
+            binary_condition(
+                ConditionFunctions.EQ,
+                Column(None, None, self.extra_condition_column),
+                Literal(None, metadata[self.extra_condition_data_key]),
+            ),
+        ]
+        new_condition = combine_and_conditions(condition_to_add)
```
lynnagara, line 66: Why bother making `condition_to_add` a list and then joining it with combine_and_conditions. There is only 1 item in there anyway so it's always a no-op

snuba/datasets/entity_subscriptions/processors.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def to_dict(self, metadata: Mapping[str, Any]) -> Mapping[str, Any]:
+        if self.extra_condition_data_key not in metadata:
+            raise InvalidQueryException
+        return {self.extra_condition_data_key: metadata[self.extra_condition_data_key]}
+
+    def process(
+        self,
+        query: Union[CompositeQuery[Entity], Query],
+        metadata: Mapping[str, Any],
```
lynnagara, line 54: I don't really get why metadata has to get passed everywhere now. Surely `offset` below would be a part of metadata?

snuba/datasets/entity_subscriptions/processors.py
```diff
@@ -0,0 +1,72 @@
...
+        offset: Optional[int] = None,
+    ) -> None:
+        raise NotImplementedError
+
+
+class AddColumnCondition(EntitySubscriptionProcessor):
+    def __init__(self, extra_condition_data_key: str, extra_condition_column: str):
+        self.extra_condition_data_key = extra_condition_data_key
```
lynnagara, line 43: What does data_key mean? Would `value` be a better name here?

snuba/subscriptions/data.py
```diff
@@ -161,18 +169,21 @@ def build_request(
...
         cls, data: Mapping[str, Any], entity_key: EntityKey
     ) -> SubscriptionData:
-        entity_subscription = get_entity_subscription(entity_key)(data_dict=data)
+        entity_subscription = get_entity(entity_key).get_entity_subscription()
+        assert entity_subscription is not None
+
+        metadata = {}
+        for key in data.keys():
+            if key not in SUBSCRIPTION_DATA_PAYLOAD_KEYS:
+                metadata[key] = data[key]
```
lynnagara, line 186: This seems too complicated and now you are keeping these keys in two places.

Instead of doing `data["project_id"]` etc below why don't you just `data.pop("project_id")`. Everything remaining in `data` after you've popped the parts needed for SubscriptionData will be your metadata.

snuba/datasets/entity_subscriptions/processors.py
lynnagara: Nvm i see it's the name of the key in redis

snuba/subscriptions/data.py
```diff
@@ -181,16 +192,23 @@ def from_dict(
...
             "resolution": self.resolution_sec,
             "query": self.query,
-            **self.entity_subscription.to_dict(),
         }
+        if self.entity_subscription.processors:
+            for processor in self.entity_subscription.processors:
+                subscription_data_dict.update(processor.to_dict(self.metadata))
+        if self.entity_subscription.validators:
+            for validator in self.entity_subscription.validators:
+                subscription_data_dict.update(validator.to_dict(self.metadata))
```
lynnagara, line 210: It's not a validators job to be return data to write to redis

evanh: Why not? What if we want to validate based on metadata in the subscription?

lynnagara: I think a validator needs to do only that - validate. That is provide a validation function that can return true if valid and false if not. So it's fine if it wants to validate subscription metadata but it is not ok that it is the one deciding what metadata to actually write there. A validator should never be mutate data in our systems. If it's doing more than validating, maybe it should not be called something else.

This is the reason why I suggest you do not split processors and validators as I think they might be inherently linked.

snuba/datasets/entity_subscriptions/processors.py
enochtangg: This makes sense; however, I'm not too familiar with why this was the case. Looking at how we currently do this, `tick_upper_offset` is a key in top-level key accessed in the [codec](https://github.com/getsentry/snuba/blob/master/snuba/subscriptions/codecs.py#L105). `SubscriptionData` on the other hand is obtained by `scheduled_subscription_dict["task"]["data"]` where `scheduled_subscription_dict` is the decoded raw payload as a dictionary. Moving this key (top-level --> inner-level) into the new metadata attribute we have defined will inherently change the shape of how subscriptions look like today.

snuba/subscriptions/data.py
evanh: @enochtangg If we removed the `to_dict` function from the validator and only had the processor write metadata, would that break any existing subscriptions?

enochtangg: That makes sense. We can remove the `to_dict` abstract method from the validator base. Right now it only returns an empty dictionary.

evanh: OK let's do that then. That keeps the logic in the processor alone.

snuba/datasets/entity_subscriptions/processors.py
```diff
@@ -0,0 +1,68 @@
...
+    ) -> None:
+        if self.extra_condition_data_key not in metadata:
+            raise InvalidQueryException
+        condition_to_add: Expression = binary_condition(
+            ConditionFunctions.EQ,
+            Column(None, None, self.extra_condition_column),
+            Literal(None, metadata[self.extra_condition_data_key]),
+        )
+        condition = query.get_condition()
+        if condition:
```
evanh, line 64: This statement is broken: if `condition` is falsey (`if condition: # False`) it will cause an error to be thrown because `new_condition` is not defined.

evanh (commented):
One thing to fix for me. I think this is pretty much ready.

snuba/datasets/entity_subscriptions/processors.py
enochtangg: Updated!

evanh: I don't think this is right either. If the query has no condition, you want to set it to be `condition_to_add`.

```
if condition:
    ...
    query.set_ast_condition(new_condition)
else:
    query.set_ast_condition(condition_to_add)
```

evanh (commented):
This looks good to me but before I approve, I'd like to make sure all the outstanding questions have been answered. Can you please go back through the comments on the PR and make sure you've answered any open questions? I can see a few questions from Lyn that don't have responses.

snuba/datasets/configuration/json_schema.py
enochtangg: Forgot to comment here directly, but the PR has been updated and `to_dict` abstract method has been removed from the `EntitySubscriptionValidator` class.

I think that's a good point and we should be validating the org ID. Since the current state of entity subscriptions does not do org_id validation either and adding a new validator seems like more code that can complicate this PR, I can file a ticket for this feature?

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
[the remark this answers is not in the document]
enochtangg: This makes sense to me. Before making any changes, I'd like to point out a few implications with this change:
* `SubscriptionData` class has an attribute `entity_subscription: EntitySubscription` as of today. This change would require us to change that interface.
* New use cases might be added to entity_subscriptions as we move towards providing service owners with more autonomy. This would require new attributes to be added to this class.

snuba/datasets/configuration/json_schema.py
lynnagara: Follow up seems fine to me. I was not really suggesting we should be adding a new validator in this PR though. It's about whether the validator and processor should be a shared implementation as in this case of org_id they will be inherently linked. In this PR, the validator is completely separate so user need to remember to include it rather than it being an extension of the processor.

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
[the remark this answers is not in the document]
lynnagara: What do you imagine these new use cases for entity subscriptions be? A processor already allows a user quite a lot of autonomy to transform a query in any way.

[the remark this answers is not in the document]
enochtangg: I don't think I know of any real use-cases today and I'm unsure about the future direction of subscriptions. Maybe adding rate limiting functionality to subscriptions.. idk? If most speculative use cases can fit within processors and validators, then we don't need to design for features that don't exist.

snuba/datasets/configuration/json_schema.py
enochtangg: That makes sense, I agree that in the case of `org_id`, the validator and processor is inherently linked and shared implementation would be a good safe-guard. Something I just thought of, for any processor a user defines, shouldn't the associated validator be automatically applied as opposed needing them to specific it explicitly? Could be out of scope.

Could I suggest a more incremental approach to this? The reason why validators and processors are implemented separately is because it matches what we have today. I'd be happy to create a ticket for this design change and carry it forward. Making this change requires some refactoring of the `build_request()` method in [SubscriptionData](https://github.com/getsentry/snuba/blob/33bb2c04aa08c123b15a4b5831cce0ca3e898b1b/snuba/subscriptions/data.py#L145).

snuba/datasets/configuration/generic_metrics/entities/sets.yaml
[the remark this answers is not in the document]
enochtangg: Updated to split subscription_validators and subscription_processors independently in Entity class

PR #3389, opened 2022-11-16
test(consumer): Add preliminary unit tests to ConsumerBuilder

ayirr7:
While working on a separate project (https://getsentry.atlassian.net/browse/SNS-1810), we realized that making changes to the `ConsumerBuilder` class is difficult due to the lack of tests. This PR attempts to add a few preliminary unit tests that may be helpful for validating the initialization of a `ConsumerBuilder`.

In this PR:

- Check whether initializations of `ConsumerBuilder` with various optional/non-optional arguments are valid. To start out, we look at the case (1) where only the required arguments are passed in, and the other case (2) where all required and optional (but those with no defaults) arguments are passed in. 

- Ensure that, in both of these cases, `ConsumerBuilder` has its attributes properly defined. For some of these attributes we can directly check the values since they should be fixed or fairly consistent across different initializations/settings. For some, we only check whether or not they are of the right Type. Depending on whether the attribute is optional or not, we may also check the bare minimum, i.e. if the attribute is defined at all. We generally try to choose the strongest possible test for each attribute. 

This is an initial attempt, so I appreciate any thoughts on the nature of these tests or suggestions for other tests that could be added!

[SNS-1810]: https://getsentry.atlassian.net/browse/SNS-1810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lynnagara:
I think we should focus on testing the main method of the ConsumerBuilder and specifically ensuring that `build_base_consumer()` builds a consumer with the right properties. Currently this is not being tested yet (I assume it was supposed to be a follow up)

I feel that some of the tests in this PR are too focussed on testing what is essentially internal attributes of the consumer builder. Well, technically they are not marked internal but all these properties like `self.storage` or self.bootrap_servers` etc are never actually accessed from outside. My 2c is that these parts are subject to change and having all these tests on internal properties makes refactoring more difficult later.

ayirr7:
> I think we should focus on testing the main method of the ConsumerBuilder and specifically ensuring that `build_base_consumer()` builds a consumer with the right properties. Currently this is not being tested yet (I assume it was supposed to be a follow up)

That's correct, my thought process was to do it incrementally.
 
> I feel that some of the tests in this PR are too focussed on testing what is essentially internal attributes of the consumer builder. Well, technically they are not marked internal but they probably should be as all these properties like `self.storage` or `self.bootrap_servers` etc are never actually accessed from outside. My 2c is that these parts are subject to change and having all these tests on internal properties makes refactoring more difficult later.

I see your point, I was also running into some confusion about how much of the internal attributes I should be testing, and in what way. I guess these tests were primarily motivated by the mistake that was made in our code a couple of weeks back for project week, where the `__init__` method wasn't covering/properly initializing some attributes in certain use cases (e.g. when an override Kafka topic is provided). 

I suppose that I can try and change the focus of these tests to testing the output of `build_base_consumer()` and ensuring that it has the right properties, since that should cover some of these attributes anyways

tests/consumers/test_consumer_builder.py
```diff
@@ -0,0 +1,254 @@
...
+    assert consumer_builder.max_batch_time_ms == 4
+    assert consumer_builder.auto_offset_reset == "earliest"
+    assert consumer_builder.queued_max_messages_kbytes == 1
+    assert consumer_builder.queued_min_messages == 2
+
+    assert consumer_builder_with_opt.max_batch_size == 3
+    assert consumer_builder_with_opt.max_batch_time_ms == 4
+    assert consumer_builder_with_opt.auto_offset_reset == "earliest"
+    assert consumer_builder_with_opt.queued_max_messages_kbytes == 1
+    assert consumer_builder_with_opt.queued_min_messages == 2
```
lynnagara, line 121: Since the assertions appear to be the same for the 2 consumer builders, could you use parametrize (or an equivalent for loop)? https://docs.pytest.org/en/6.2.x/parametrize.html

tests/consumers/test_consumer_builder.py
```diff
@@ -0,0 +1,254 @@
...
+    # In the case that optional Kafka topic overrides
+    # are passed in, stronger checks are performed
+    # in a separate test
+
+    try:
+        consumer_builder.bootstrap_servers
+    except NameError:
+        assert (
+            False
+        ), "ConsumerBuilder's bootstrap_servers must either be defined as some value or None"
```
lynnagara, line 137: How about removing the try/except/assert False. Makes it easier to keep these up to date if the attributes change.
```suggestion
    consumer_builder.bootstrap_servers
```

lynnagara (commented):
Took another look over this PR and I think these tests are pretty valuable still, how about we merge this first and then add additional testing for the main consumer/strategy as a follow up?

Would you mind reducing the boilerplate a bit by using parametrize where possible and simplifying some of the assertions to omit the try/except if not needed?

tests/consumers/test_consumer_builder.py
ayirr7: This works!

tests/consumers/test_consumer_builder.py
ayirr7: Done!

lynnagara (approved):
🎉

PR #3399, opened 2022-11-16
ref(migrations): make migration use single ordered sequence SNS-1830

dbanda:
Refactors migrations to use a singe sequence of operations, rather having the user specify operations in `_dist()` and `_local()`.

## Context
To avoid ordering errors in migrations we refactor them to be a single sequence. The user no longer has to specify the order flags for the whole migration. Instead, each SQL operation takes a target flag specifying if it should run on local or dist. To make a node migration, the user has to only implement `forwards_ops()` and `backwards_ops()`. This removes the limitation of having the user dump all their ops in either `_dist()` or `_local()` methods and executing one before the other. Now the user has complete control of when their operations can execute.

To avoid errors introduced by giving the user more power, we use a validator that runs unit tests against migrations to ensure that there are no ops that will cause local/dist ordering issues.

This refactoring maintains backwards compatibility with migrations that use the old classes, avoiding a hard fork of the repo. Old migrations using the legacy classes are wrapped with their own `backwards_ops` and `forwards_ops` methods which are executed by the runner.

## Before
Users would have to implement the `_dist()` and `_local()` methods each for each backwards and forwards, and specify which one is executed first.

## After
Users would only need to implement `forwards_ops` and `backwards_ops`, specifying for each op if it should run local or dist.

## Blast radius
Refactors the node migration class in `migration.py`. The old class is renamed to `Legacy` (this rename affects all existing migrations). Both the old and new form of the class are compatible with the runner. The runner is simplified to iterate through a single sequence of ops for backwards/forwards. SQL operations in `operation.py` are refactor to take in a target enum flag specifying if they are to run local or dist. 

## Testing
Since the new and old classes are functionally compatible, we repurpose a lot of the tests targeting the old classes to run on the new classes.

### DIFF
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -182,6 +182,9 @@ class ClickhouseNodeMigration(Migration, ABC):
 
 
 class ClickhouseNodeMigrationLegacy(ClickhouseNodeMigration, ABC):
+    """
+    This is now deprecated. Use ClickhouseNodeMigration instead.
+    """
 
     forwards_local_first: bool = True
     backwards_local_first: bool = False
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -166,19 +166,19 @@ class ClickhouseNodeMigration(Migration, ABC):
         self,
         ops: Sequence[SqlOperation],
     ) -> None:
-        def print_dist_op(op: SqlOperation, is_single_node: bool) -> None:
+        def print_dist_op(op: SqlOperation) -> None:
+            cluster = get_cluster(op._storage_set)
+            is_single_node = cluster.is_single_node()
             if not is_single_node:
                 print(f"Distributed op: {op.format_sql()}")
             else:
                 print("Skipped dist operation - single node cluster")
 
         for op in ops:
-            cluster = get_cluster(op._storage_set)
-            is_single_node = cluster.is_single_node()
             if op.target == OperationTarget.LOCAL:
                 print(f"Local op: {op.format_sql()}")
             elif op.target == OperationTarget.DISTRIBUTED:
-                print_dist_op(op, is_single_node)
+                print_dist_op(op)
 
 
 class ClickhouseNodeMigrationLegacy(ClickhouseNodeMigration, ABC):
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -114,61 +114,13 @@ class ClickhouseNodeMigrationLegacy(Migration, ABC):
     completely unrelated, they are probably better as separate migrations.
     """
 
-    forwards_local_first: bool = True
-    backwards_local_first: bool = False
-
-    def forwards_local(self) -> Sequence[SqlOperation]:
-        raise NotImplementedError
-
-    def backwards_local(self) -> Sequence[SqlOperation]:
-        raise NotImplementedError
-
-    def forwards_dist(self) -> Sequence[SqlOperation]:
-        raise NotImplementedError
-
-    def backwards_dist(self) -> Sequence[SqlOperation]:
-        raise NotImplementedError
-
-    def _set_targets(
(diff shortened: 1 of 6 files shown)


MeredithAnya:
It's a little cumbersome to review such a large PR, is it possible to split the name change out into it's own PR? basically just the changing of the name to Legacy.

snuba/migrations/operations.py
```diff
@@ -9,26 +10,48 @@
...
+    DISTRIBUTED = "distributed"
+
+
 class SqlOperation(ABC):
     def __init__(
-        self, storage_set: StorageSetKey, settings: Optional[Mapping[str, Any]] = None
+        self,
+        storage_set: StorageSetKey,
+        settings: Optional[Mapping[str, Any]] = None,
+        target: OperationTarget = OperationTarget.LOCAL,
```
lynnagara, line 28: I think this is a bit dangerous, the user should explicitly provide the target. Can we do away with the default?

lynnagara: I think all the child classes of SqlOperation need take the target argument in the constructor now. I think we should try and keep this immutable.

dbanda: The trouble with having no default is we would have to add the parameter in all 110 of the migrations files for the existing migrations. What I am doing instead is setting a default the applying a wrapper to correct the target for the legacy migrations.

dbanda:
> It's a little cumbersome to review such a large PR, is it possible to split the name change out into it's own PR? basically just the changing of the name to Legacy.

Agreed. Let me see how I can refactor it. The changes in `snub_migrations` are all part of the rename change. All the relevant changes are in `migrations.py`, `operations.py`, `runner.py` and `validator.py`

snuba/migrations/operations.py
lynnagara: The fact that not all child classes take this argument means it would still be wrong for any new migrations though.

I know it's a bit of grunt work but one option would be to have a separate PR that gets merged first that only adds the OperationTarget enum and makes it mandatory in any SqlOperation. I know this is annoying but some reasons for proposing this:
- I am not a fan of making things that should be immutable mutable and would like to find a way to avoid it
- I think having local as a default is a bit of a footgun and would actually prefer making the default None and then adding some validation later to ensure it is set even if we absolutely need to keep it mutable
- If not now I think we would just need to do this work later to make it a required argument as it affects the interface of the SqlOperation class going forward

dbanda:
@MeredithAnya I created #3417 to do the rename. Hopefully, that makes this one more managable.

snuba/migrations/operations.py
dbanda: Having target as a required is super tricky cause I can't see a way around manually adding it to every migration. Having default None is more doable and would throw errors on new migrations if it's not explicitly set. I think thats the better way to go for now.

dbanda: For typing purposes, setting None as default can be done by adding an UNSET enum to the OperationTarget. That also makes it more clear that it needs to be set.

snuba/migrations/migration.py
```diff
@@ -197,37 +164,75 @@ def backwards(self, context: Context, dry_run: bool) -> None:
...
+        def print_dist_op(op: SqlOperation, is_single_node: bool) -> None:
+            if not is_single_node:
+                print(f"Distributed op: {op.format_sql()}")
+            else:
+                print("Skipped dist operation - single node cluster")
 
-            for op in local_operations:
-                print(op.format_sql())
+        for op in ops:
+            cluster = get_cluster(op._storage_set)
```
MeredithAnya, line 176: why not move this line and the line below into the `print_dist_op` since it seems like you only need to get the cluster when we have a dist op?

dbanda: moved

MeredithAnya (approved):
I just have some smol comments, don't think they are necessarily blocking though

PR #3404, opened 2022-11-17
fix(clickhouse): Annotate INSERT error with the relevant row [SNS-1850]

untitaker:
## Background

When a snuba consumer crashes, the relevant clickhouse error does not show which data caused the error, and it's hard to figure out which of the 10k rows in the INSERT made the entire insert fail. This was somewhat inconvenient in an incident where the querylog consumer crashlooped and we had a hard time figuring out the shape of the malformed data in kafka.

## Solution

Set the row's payload as Sentry context, so that when the snuba consumer
crashes on INSERT, the relevant sentry error event contains the faulty
payload. Sentry contexts have the most generous size limits and are
least likely to be truncated by Relay. At the same time Relay should be
smart enough to discard this data before it attempts to trim down more
important info, like the error message or stacktrace. We'll see.

I didn't find a good place to grab this data. It seems that we stream
the rows from processing strategies without ever holding all of them in
memory. As such it's very likely that this patch significantly increases
memory consumption, because we need to buffer up all encoded rows in
case we need to add them to an error.

Also the rows are encoded bytestrings, so they will look a bit ugly in
Sentry but that's also the cheapest representation to keep in memory
until the end of the batch (one PyObject per message)


snuba/clickhouse/http.py
```diff
@@ -199,7 +200,7 @@ def append(self, value: bytes) -> None:
         assert not self.__closed
 
         self.__queue.put(value)
-        self.__rows += 1
+        self.__rows.append(value)
```
evanh, line 203: How big (in length and in bytes) is this expected to get?

untitaker: I don't know to be honest. We have many different datasets with different column counts and batch sizes. I might want to deploy a PR first that estimates this via DD metric but perhaps somebody can make an assessment straight away on whether this approach is feasible or whether I need to get more clever.

onewland: This DD notebook might be a rough estimate of row sizes/dataset based on total bytes inserted vs. total row count inserted:

https://app.datadoghq.com/notebook/3950488/oliver-nov-17-2022-13-23

nikhars: Why not as an initial step, you start emitting metrics for `self.__size` to get a sense of how big the total length of value is? Events and Transactions payload can be really big. And the processing strategy which writes to clickhouse happens in parallel to batches being filled with more messages from kafka concurrently. So in theory the memory usage might be close to doubling with this.

snuba/clickhouse/http.py
```diff
@@ -219,6 +220,14 @@ def join(self, timeout: Optional[float] = None) -> None:
                 code = int(details["code"])
                 message = details["message"]
                 row = int(details["row"]) if details["row"] is not None else None
```
lynnagara, line 222: Could we start by simply including this `details["row"]` value that comes back from ClickHouse to the Sentry error instead of storing a copy of all the raw rows.

lynnagara: edit: nvm, that is probably nothing useful. But what about attaching some of the other data that comes back from the ClickHouse error?

untitaker: @lynnagara I think the f/u from the incident was specifically about missing insight into the data that is being inserted, i don't think clickhouse returns that as part of the error. Is there something in particular that you would also like to have?

snuba/clickhouse/http.py
untitaker: This is a good idea nikhar, done in https://github.com/getsentry/snuba/pull/3431

snuba/clickhouse/http.py
lynnagara: Ah, ignore my comment then. I assumed we just wanted parts of the clickhouse error that was missing and might help us better diagnose. But that might not help us understand if bad data got there to begin with.

untitaker:
errors/transactions appear to have max batch size of 60MB, but I thought of something else.

I decided to move the max buffer size behind a runtime config flag that is disabled by default. This means that during incident resolution, one has to increase the value to increase the likelihood that the clickhouse error has the row annotated. That chance is 0% by default, but one can increase it gradually while the consumer crashloops, paying with increased memory consumption.

There is now also a Sentry tag to search for error events that have the erroring row as metadata.

snuba/clickhouse/http.py
```diff
@@ -233,7 +256,7 @@ def join(self, timeout: Optional[float] = None) -> None:
                 try:
                     if row is not None:
                         sentry_sdk.set_context(
-                            "snuba_errored_row", {"data": self.__rows[row]}
+                            "snuba_errored_row", {"data": self.__debug_buffer[row]}
```
evanh, line 259: Since the debug buffer doesn't necessarily contain all the rows, couldn't this throw an IndexError?

untitaker: yeah but the indexerror is caught down

evanh (approved):
Seems fine.

I would document how/when this should be used in a runbook in Notion as well. We have a section for Runbooks in our team Notion page.

untitaker:
https://www.notion.so/sentry/WIP-Debugging-crashloops-in-consumers-5442684f820b490cb8e4b236d82570c6

PR #3405, opened 2022-11-17
feat(slicing): Mega cluster using runtime overrides

nikhars:
### Context
With slicing, logical partitions could be one of 2 states
1. All data resides on a single sliced clickhouse cluster.
2. Partial data resides in some other sliced clickhouse cluster. This would happen when the mapping of a logical partition to a slice is modified. When the mapping is changed, **writes** of new data will happen on the new slice on which the logical partition is mapped. But we do not perform data migration of the logical partition from older slice to newer slice. So **reads** need to query across both the old slice and the new slice.

The overall system would look something like this
Screen Shot 2022-11-17 at 9 45 33 AM

### Implementation
A new runtime config is introduced to identify which logical partitions of a storage_set have should use the mega cluster. If a  query belongs to a logical partition which is set in the runtime config, it uses cluster information from settings.CLUSTER.

### Tests
1. Added a test to validate that when runtime override is set, the database selected is from settings.CLUSTER even when slicing is enabled.
2. Added tests to check behavior of different values of the runtime config.

### DIFF
--- a/snuba/datasets/plans/sliced_storage.py
+++ b/snuba/datasets/plans/sliced_storage.py
@@ -3,6 +3,7 @@ from typing import Optional, Sequence
 
 import sentry_sdk
 
+from snuba import state
 from snuba.clickhouse.query_dsl.accessors import get_object_ids_in_query_ast
 from snuba.clickhouse.translators.snuba.mapping import TranslationMappers
 from snuba.clusters.cluster import ClickhouseCluster, get_cluster
@@ -34,6 +35,8 @@ from snuba.query.processors.physical.mandatory_condition_applier import (
 from snuba.query.query_settings import QuerySettings
 from snuba.util import with_span
 
+MEGA_CLUSTER_RUNTIME_CONFIG_PREFIX = "slicing_mega_cluster_partitions"
+
 
 class StorageClusterSelector(ABC):
     """
@@ -49,11 +52,47 @@ class StorageClusterSelector(ABC):
         raise NotImplementedError
 
 
+def _should_use_mega_cluster(
+    storage_set: StorageSetKey, logical_partition: int
+) -> bool:
+    """
+    Helper method to find out whether a logical partition of a sliced storage
+    set needs to send queries to the mega cluster.
+
+    Mega cluster needs to be used when a logical partition's data might be
+    across multiple slices. This happens when a logical partition is mapped
+    to a new slice. In such cases, the old data resides in some different
+    slice than what the new mapping says.
+    """
+    key = f"{MEGA_CLUSTER_RUNTIME_CONFIG_PREFIX}_{storage_set.value}"
+    state.get_config(key, None)
+    slicing_read_override_config = state.get_config(key, None)
+
+    if slicing_read_override_config is None:
+        return False
+
+    slicing_read_override_config = slicing_read_override_config[1:-1]
+    if slicing_read_override_config:
+        logical_partition_overrides = [
+            int(p.strip()) for p in slicing_read_override_config.split(",")
+        ]
+        if logical_partition in logical_partition_overrides:
+            return True
+
+    return False
+
+
 class ColumnBasedStorageSliceSelector(StorageClusterSelector):
     """
-    Storage slice selector for the generic metrics storage. This is needed
-    because the generic metrics storage can be sliced and we would need to
-    know which slice to use for a specific query.
+    Storage slice selector based on a specific column of the query. This is
+    needed in order to select the cluster to use for a query. The cluster
+    selection depends on the value of the partition_key_column_name.
+
+    In most cases, the partition_key_column_name would get mapped to a logical
+    partition and then to a slice. But when a logical partition is being
+    transitioned to another slice, the old data resides in a different slice. In
+    those cases, we need to use the mega cluster to query both the old data and
+    the new data.
     """
 
     def __init__(
@@ -81,12 +120,13 @@ class ColumnBasedStorageSliceSelector(StorageClusterSelector):
         assert len(org_ids) == 1
         org_id = org_ids.pop()
 
-        slice_id = map_logical_partition_to_slice(
-            self.storage, map_org_id_to_logical_partition(org_id)
-        )
-        cluster = get_cluster(self.storage_set, slice_id)
-
-        return cluster
+        logical_partition = map_org_id_to_logical_partition(org_id)
+        if _should_use_mega_cluster(self.storage_set, logical_partition):
+            return get_cluster(self.storage_set)
+        else:
+            slice_id = map_logical_partition_to_slice(self.storage, logical_partition)
+            cluster = get_cluster(self.storage_set, slice_id)
+            return cluster
 
 
 class SlicedStorageQueryPlanBuilder(ClickhouseQueryPlanBuilder):
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -319,24 +319,27 @@ SLICED_STORAGES: Mapping[str, int] = {}
 # to slice id
 LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {}
 
-# The slice configs below are the "SLICED" versions to
-# the equivalent default settings above. For example,
-# "SLICED_KAFKA_TOPIC_MAP" is the "SLICED" version of
-# "KAFKA_TOPIC_MAP". These should be filled out
-# for any corresponding sliced storages defined above,
-# with the applicable number of slices in mind.
+# The slice configs below are the "SLICED" versions to the equivalent default
+# settings above. For example, "SLICED_KAFKA_TOPIC_MAP" is the "SLICED"
+# version of "KAFKA_TOPIC_MAP". These should be filled out for any
+# corresponding sliced storages defined above, with the applicable number of
+# slices in mind.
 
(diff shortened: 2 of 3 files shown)


snuba/datasets/plans/sliced_storage.py
```diff
@@ -81,12 +120,13 @@ def select_cluster(
...
 
-        slice_id = map_logical_partition_to_slice(
-            self.storage, map_org_id_to_logical_partition(org_id)
-        )
-        cluster = get_cluster(self.storage_set, slice_id)
-
-        return cluster
+        logical_partition = map_org_id_to_logical_partition(org_id)
+        if _should_use_mega_cluster(self.storage_set, logical_partition):
+            return get_cluster(self.storage_set)
```
onewland, line 125: I was sort of assuming that the storage set key for the "mega-cluster" would be distinct from the sliced cluster, though it's not a requirement. Presumably the run-time config used to define `_should_use_mega_cluster` could store that storage set key in the value.

Did you consider this and dismiss it or just assume that we would want the storage set key to remain consistent whether using the mega-cluster or the normal sliced cluster?

onewland (approved):
I'm fine with this change though I think we should consider making the storage set key variant vs. making it the same between "mega-cluster" and "slice cluster".

Thanks for the nice diagram in the PR description!

evanh (commented):
LGTM

snuba/datasets/plans/sliced_storage.py
ayirr7: I was also wondering about this. Currently, `settings.CLUSTERS` is only for non-sliced storage set keys, `settings.SLICED_CLUSTERS` only for sliced storage set keys. We can't have a storage set key be present in both.

nikhars: I considered using a separate storage_set but decided against it and using a simpler approach. My thought process was the following.

The way to access a clickhouse cluster is via `storage_set`. Example [here](https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L61-L77). Lets assume mega cluster access should happen via a separate `storage_set`. In that case we would want a generic_metrics [entity](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/entities/distributions.yaml#L3) to map to a mega cluster `storage_set`. Today it links with a `storage_set` from [here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/entities/distributions.yaml#L69) and then [here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml#L7)

To enable that lets see what would need to happen. `storage_set` is defined in the yaml files like [this](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml#L7). That code is used as part of building the [storage schema](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/storage_builder.py#L87) and [readable storage](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/storage_builder.py#L95). `ReadableTableStorage` is linked with an `Entity`. An `Entity` can only have one `ReadableTableStorage` as shown [here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/pluggable_entity.py#L44). We don't want to specify multiple `ReadableTableStorage` for a single `Entity`. That means defining a new `storage_set` would imply defining a new `Entity` as well which we do not want in this case. And it would also imply keeping schema `columns` same across different datasets and migrations consistent across different datasets. Which is another big headache. Also, the new storage_key is probably needed only when adding new slices. It has no purpose after all data has been migrated.

So I would like to simplify the problem. And the simplification is that somehow an API call needs to be routed to the mega-cluster. mega-cluster is just another host/port definition which needs to be looked up. If we want the mega-cluster definition to not be in settings.CLUSTER, a workaround I can think of is defining a reserved slice number (say 10000) which would provide access to the mega-cluster via settings.SLICED_CLUSTERS.

snuba/datasets/plans/sliced_storage.py
```diff
@@ -49,11 +52,47 @@ def select_cluster(
...
+    """
+    Helper method to find out whether a logical partition of a sliced storage
+    set needs to send queries to the mega cluster.
+
+    Mega cluster needs to be used when a logical partition's data might be
+    across multiple slices. This happens when a logical partition is mapped
+    to a new slice. In such cases, the old data resides in some different
+    slice than what the new mapping says.
+    """
+    key = f"{MEGA_CLUSTER_RUNTIME_CONFIG_PREFIX}_{storage_set.value}"
```
fpacifici, line 67: nit: runtime config keys are lower case

snuba/datasets/plans/sliced_storage.py
onewland: That argument makes sense to me. 

@ayirr7, do you see an inherent conflict between Nikhar's approach and the work you've been doing? You say 

> We can't have a storage set key be present in both.

So is this going to fail to start `snuba api` or `snuba consumer --storage=<sliced storage>` during validation or is this just violating an assumption?

ayirr7: That also makes sense to me. We can keep it in `settings.CLUSTERS` and I don't think any validation would break because of this. I don't think it makes sense to define any mega cluster in `settings.SLICED_CLUSTERS` because the mega cluster can't be sliced. I think it'd be helpful to add this in docs or comments in the settings file (the mega cluster for a sliced storage set would be defined in `settings.CLUSTERS` and not in `SLICED_CLUSTERS`).

lynnagara (commented):
I think this design introduces some complexity with how settings.CLUSTERS and settings.SLICED_CLUSTERS works today.

The settings file contains this comment:
```
# Storage set keys should be defined either in CLUSTERS
# or SLICED_CLUSTERS.
```
...but now that comment is wrong

Also, I am wondering why there is only one mega cluster? Shouldn't a mega cluster simply cover a subset of slices for a storage

How about introducing a new setting like:
```
settings.MEGA_CLUSTERS = {
    StorageSetKey.ERRORS: {
         slices: [1, 2],
         cluster: ClusterInfo
    }
}
```
I think there are some benefits of this way:
- Cluster configuration is all in settings instead of half split across runtime config and settings
- It is explicit about which slices the mega cluster covers which seems more aligned with how the mega cluster is supposed to work.
- It allows for some slices to not use the mega cluster unnecessarily

onewland:
> I think this design introduces some complexity with how settings.CLUSTERS and settings.SLICED_CLUSTERS works today.
> 
> The settings file contains this comment:
> 
> ```
> # Storage set keys should be defined either in CLUSTERS
> # or SLICED_CLUSTERS.
> ```
> 
> ...but now that comment is wrong
> 

Agree we should address this comment

> Also, I am wondering why there is only one mega cluster? Shouldn't a mega cluster simply cover a subset of slices for a storage
> 
> How about introducing a new setting like:
> 
> ```
> settings.MEGA_CLUSTERS = {
>     StorageSetKey.ERRORS: {
>          slices: [1, 2],
>          cluster: ClusterInfo
>     }
> }
> ```
> 

To be clear, the run-time config lets you override on a per-logical partition basis so that the mega-cluster isn't used for logical partitions that aren't rebalanced. See how Nikhar is using the config values in `_should_use_mega_cluster`.

> I think there are some benefits of this way:
> 
>     * Cluster configuration is all in settings instead of half split across runtime config and settings
> 
>     * It is explicit about which slices the mega cluster covers which seems more aligned with how the mega cluster is supposed to work.
> 
>     * It allows for some slices to not use the mega cluster unnecessarily

I actually prefer having the mega-cluster usage being driven by run-time config. To me it makes it more clear that this should be a temporary band-aid while we transition to a new logical->physical partition assignment scheme rather than a durable setting (though, to your point, it needs to be in place for 90 days before the setting can be remove)

snuba/datasets/plans/sliced_storage.py
nikhars: The actual runtime prefix is lower case. The constant is defined in upper case.

nikhars:
> I think this design introduces some complexity with how settings.CLUSTERS and settings.SLICED_CLUSTERS works today.
> 
> The settings file contains this comment:
> 
> ```
> # Storage set keys should be defined either in CLUSTERS
> # or SLICED_CLUSTERS.
> ```
> 
> ...but now that comment is wrong
I will fix the comment since it is now incorrect.

> Also, I am wondering why there is only one mega cluster? Shouldn't a mega cluster simply cover a subset of slices for a storage
> 
> How about introducing a new setting like:
> 
> ```
> settings.MEGA_CLUSTERS = {
>     StorageSetKey.ERRORS: {
>          slices: [1, 2],
>          cluster: ClusterInfo
>     }
> }
> ```
Mega-cluster is needed for any logical partition which gets moved. We don't really care about what the underlying slice is from where it is moving. If you look at the diagram in the PR description, the query node will have its cluster defined as hosts from all slices.

> I think there are some benefits of this way:
> 
> * Cluster configuration is all in settings instead of half split across runtime config and settings
> * It is explicit about which slices the mega cluster covers which seems more aligned with how the mega cluster is supposed to work.
> * It allows for some slices to not use the mega cluster unnecessarily
I prefer runtime config for now. It allows us to switch the logical partition rather than waiting for ops to apply and then take effect. When we do want to use it and something goes wrong, switching things is under our control.

PR #3406, opened 2022-11-17
ref(admin): Add AuditLog base

MeredithAnya:
Kind of another try at consolidating https://github.com/getsentry/snuba/pull/3393 into a more standard way of doing audit logging/notifications for the admin tooling 

**Goal**
Provide a simple way to audit admin usage and make it easy to search in one place (GCP for now). In addition to creating an audit log entry, we can also optionally send a notification to slack. 

**Changes from previous implementation**
* always log the record to GCP (the runtime config was defaulting to logging if slack wasn't configured but otherwise wasn't logging) I know we keep the runtime config changes in redis as well, but I think it makes sense for the admin audit logs to be in one place to easily look up
* structlog `logger` no longer binds the `module`. The only thing helpful was that you knew what resource it was (e.g. `querylog`) but the action of the auditlog should tell you that information
* `SlackClient` doesn't get a `channel_id` passed in, we always use the same id for admin

**What affected?**
* querylog audit logs will have the event be `ran.query` instead of `run_query`, `start_timestamp` is now `timestamp`, and the `module` attribute is removed
* runtime config audit logs will start being logged to GCP

cc @dbanda


snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,80 @@
...
+
+class QuerylogAuditLog(AuditLog):
+    def __init__(self) -> None:
+        self.logger = structlog.get_logger().bind(module=__name__)
+
+    def _record(
+        self, user: str, timestamp: str, action: str, data: Mapping[str, Any]
+    ) -> None:
+        query = data["query"]
+        status = data["status"]
```
nikhars, line 30: Would there be cases where some keys in `data` can be missing. You can probably make access to `data` keys safe at all the places you are using it.

snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,80 @@
...
+        start_timestamp = timestamp
+        self.logger.info(
+            event=action,
+            user=user,
+            query=query,
+            status=status.value,
+            start_timestamp=start_timestamp,
+            end_timestamp=datetime.now().strftime(DATETIME_FORMAT),
```
nikhars, line 38: Wouldn't the start_timestamp and end_timestamp be the same always in this case? start_timestamp would always be https://github.com/getsentry/snuba/pull/3406/files#diff-fc0f55816e7b2ce9c6703ffa2134728694149604e31e3f4a76ba78494666f90cR21-R22

snuba/admin/audit_log/base.py
```diff
@@ -0,0 +1,36 @@
+from abc import ABC, abstractmethod
+from datetime import datetime
+from typing import Any, Mapping, Optional
+
+
+class AuditLog(ABC):
```
fpacifici, line 6: If I understand this design correctly, I think this class has two independent responsibilities that are going to generate a lot of duplication.
On one side it is feature specific: it runs the application logic that formats the audit log record or notification payload.
On the other side it contains the channel specific logic that allows you to write the record somewhere.

The consequence of storing both in the same class is that if you want to allow the multiple audit log type to use the same channel (like writing on redis) you need to duplicate the channel specific logic.

Why not making the `AuditLog` class represent a channel instead (if you can define a common interface). At that point you will have one implementation for slack, one for the GCP logger, one for redis, etc. Each use case can use channels the way it wants without the need for an abstract class everyone needs to implement.
A sign that this class in itself is not very useful are:
- I cannot imagine a callsite referencing the abstract class `AuditLog`. Whenever you use the audit log you only instantiate the subclasses and reference the subclasses. So this class is not useful for polymorphism.
- It does almost anything that is shared by subclasses. Which means you are not using to share code. Also there are better ways to share code via composition if polymorphism is not your goal.

If you then find value in adding common logic between all the audit log implementations you can add an `audit` function that implements this logic and forwards the payload to the specific channel. An example could be making each feature load the right channel according to configuration.

nikhars (approved):
I have left minor comments but overall think the approach should be fine

snuba/admin/audit_log/querylog.py
rahul-kumar-saini: `start_timestamp` is coming from this line, which is executed before the query is run:
https://github.com/getsentry/snuba/pull/3406/files#diff-ba008dc0ca3311bd1841a5ccf1403e9c3bf554f270a1f15b735d4aa8f041a344R67

snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,80 @@
...
+    ) -> None:
+        query = data["query"]
+        status = data["status"]
+        start_timestamp = timestamp
+        self.logger.info(
+            event=action,
+            user=user,
+            query=query,
+            status=status.value,
+            start_timestamp=start_timestamp,
```
rahul-kumar-saini, line 37:
```suggestion
            start_timestamp=timestamp,
```

snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,80 @@
...
+class QuerylogAuditLog(AuditLog):
+    def __init__(self) -> None:
+        self.logger = structlog.get_logger().bind(module=__name__)
+
+    def _record(
+        self, user: str, timestamp: str, action: str, data: Mapping[str, Any]
+    ) -> None:
+        query = data["query"]
+        status = data["status"]
+        start_timestamp = timestamp
```
rahul-kumar-saini, line 31:
```suggestion
```

snuba/admin/audit_log/base.py
MeredithAnya: okay what about something like: https://github.com/getsentry/snuba/pull/3406/commits/163749b38804677d198fd41238f1d7944878f1a4
which just lets people override the `_format_data` if they want to. In the future I could see refactoring this to have the audit log take maybe a notifier and a recorder or something. In my mind you'd always want to record the audit log, but it would optional if you wanted to get a notification, which is why I wasn't treating slack and GCP as the same (like as a channel), although you could argue they do just send the data somewhere. Not sure if im making sense here.

fpacifici: > lets people override the _format_data if they want to. In the future I could see refactoring this to have the audit log take maybe a notifier and a recorder or something.

I think the original concern still applies. Why do you need a parent class for all subscriptions? There generally are two reasons for it: polymorphism and sharing code. 
Sharing code is generally better achieved by composition (put the common logic in its own class/function and use it from the code that wants to send the notification). You leave more flexibility to the callsite instead of adding a constraint that is not providing value.

Polymorphism does not seem to be your goal here. It seems to me you are never going to have a function or class that depends on `AurditLog`. They will always depend on the concrete implementation.

Also how would you implement two types of logs (logger and redis) with thelast version you submitted ?

snuba/admin/audit_log/base.py
```diff
@@ -0,0 +1,54 @@
...
+        time = datetime.now()
+        return time.strftime("%B %d, %Y %H:%M:%S %p")
+
+    def audit(
+        self,
+        user: str,
+        action: str,
+        data: Mapping[str, Any],
+        notify: bool = False,
+        timestamp: Optional[str] = None,
```
lynnagara, line 48: Why does timestamp need to be sometimes provided and sometimes not? Can we just do it one way? Also, curious why it's a str? This means that the caller gets to decide the formatting and so the format won't be standardised which I don't think is ideal.

snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,66 @@
...
+            "status": status.value if status else "unknown",
+            "start_timestamp": timestamp,
+            "end_timestamp": datetime.now().strftime(DATETIME_FORMAT),
+        }
+
+
+__querylog_audit_log_notification_client = QuerylogAuditLog(__name__)
+
+
+def audit_log(
```
lynnagara, line 36: What is the purpose of providing the alternative decorator interfaces here in addition to using the class directly? I don't think we have to provide multiple ways of doing the same thing here since we only really need to call this in one  place. This function is pretty heavily specific to the ClickHouseResult, and the other one is very specific to the runtime config scenario that it's hard to imagine it needs to be all that reusable.

snuba/admin/notifications/slack/client.py
```diff
@@ -49,3 +57,6 @@ def post_message(
 
         if not is_ok:
             logger.error(f"Slack error: {str(error_option)}")
+
+
+slack_client = SlackClient()
```
lynnagara, line 62: I think you can remove this. Just instantiate it in the scenarios where you need it, rather than on load of this module. I see that you are calling `client = SlackClient()` in some other places anyway and not even using this instance.

snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class QuerylogAuditLog(AuditLog):
+    def _format_data(
+        self, user: str, timestamp: str, action: str, data: Mapping[str, Any]
+    ) -> Mapping[str, Any]:
+        status = data.get("status")
+        return {
+            "query": data.get("query"),
+            "status": status.value if status else "unknown",
```
lynnagara, line 27: Shouldn't a status always be there?

snuba/admin/audit_log/querylog.py
```diff
@@ -0,0 +1,66 @@
...
+    FAILED = "failed"
+
+
+class QuerylogAuditLog(AuditLog):
+    def _format_data(
+        self, user: str, timestamp: str, action: str, data: Mapping[str, Any]
+    ) -> Mapping[str, Any]:
+        status = data.get("status")
+        return {
+            "query": data.get("query"),
```
lynnagara, line 26: Why not `data["query"]`? Under what scenarios would a query not be present?

lynnagara (commented):
Filippo mentioned this already but I agree that you should not subclass the audit log in this way.

The two variants are not really separate implementations at all, and the base class is highly specific - sharing the same logging and notification backends, including the exact same Slack channel / credentials. This means you also have two copies of an almost identical audit log decorator which I think should be avoided.

The only reason for subclassing here seems to be a kind of circular one - to allow you to pass a raw data object that then gets formatted via a custom `_format_data` method into some other data object that can get logged. I think there no real benefit for the audit log class to be responsible for doing this highly custom translation to begin with - it'd be more straightforward to pass the already formatted data into here and do away with subclasses + custom formatter method entirely.

The idea about having an audit log abstraction where you can plug in different notification backends and logging backends/databases may be interesting but I don't think it's mandatory since we have only got the one structlog and slack implementation right now.

snuba/admin/audit_log/base.py
MeredithAnya: I can just always have the timestamp be created by the audit log

snuba/admin/notifications/slack/client.py
MeredithAnya: I can change the audit log to instantiate it, but im not sure why it would be a problem to have one instance of the slack client if we wanted to use it more places that the audit log

snuba/admin/audit_log/querylog.py
MeredithAnya: this doesn't apply anymore

dbanda (approved):

PR #3407, opened 2022-11-17
feat(replays): Add tags transformer to replays query processor

cmanallen:
related: https://github.com/getsentry/replay-backend/issues/181

Enables selecting and filtering by `tags[key]`.


onewland (dismissed):
change seems fine, though I'm not sure what's going on with the github actions

evanh (approved):

evanh (changes requested):
This processor needs to be added to your entity, not the storage. That should fix CI.

nikhars (approved):

evanh:
I think I need more context on what you are trying to enable in this PR.

Moving the entity to YAML shouldn't cause any changes in how your Entity can be used/accessed. It seems like you are trying to add the `TagsTypeTransformer` processor, but that processor is pretty specific to metrics. It deals with expressions like `tags[123]`, by converting the `123` to an integer (since metrics stores tag values as ints). I'm not sure that's what you want here.

The PR you linked in your description talks about having problems with `SAMPLE`. Can you give more context on what other issues you are trying to fix?

cmanallen:
@evanh We're trying to integrate with `sentry.tagstore`.  We kept hitting roadblocks so we're just going to do a more custom implementation.  I did not know the tags transformer was specific to metrics.  We can ignore this pull.

PR #3408, opened 2022-11-17
exp(DC): Multi processed Storage config loading

rahul-kumar-saini:
### Overview
- Config loads and builds are taking [increasingly longer](https://sentry.io/organizations/sentry/performance/summary/?project=300688&transaction=Snuba+CLI+Initialization) as we introduce more and more configs
- So far there haven't been any problems but I can imagine potentially hitting some sort of timeouts especially with `snuba optimize`
    - The inits for this command take super long for some unknown reason - visible in p100

### Local experiment
- Attempting to multiprocess the initialization of storage configs to see if it has a positive impact on load times in production
- Locally it increases the load times on average *except* for the case where I hardcode each load to take at least 1s, in which case the load time significantly decreases
- Overall locally what I'm seeing is the p95/p99 cut down ~90% but the p50 go up ~20%

### Prod experiment
- Multiprocessed storage loading logic is placed behind a runtime config to start/stop the experiment
- Hoping to see similar reduction in p95+ for the initialization

### DIFF
--- a/snuba/datasets/storages/factory.py
+++ b/snuba/datasets/storages/factory.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from glob import glob
+from multiprocessing import Pool
 from typing import Generator
 
 import sentry_sdk
@@ -10,7 +11,7 @@ from snuba import settings
 from snuba.datasets.cdc import CdcStorage
 from snuba.datasets.configuration.storage_builder import build_storage_from_config
 from snuba.datasets.storage import ReadableTableStorage, Storage, WritableTableStorage
-from snuba.datasets.storages.storage_key import StorageKey
+from snuba.datasets.storages.storage_key import StorageKey, register_storage_key
 from snuba.state import get_config
 from snuba.utils.config_component_factory import ConfigComponentFactory
 
@@ -33,15 +34,27 @@ class _StorageFactory(ConfigComponentFactory[Storage, StorageKey]):
             self.__initialize()
 
     def __initialize(self) -> None:
-        self._config_built_storages = {
-            storage.get_storage_key(): storage
-            for storage in [
-                build_storage_from_config(config_file)
-                for config_file in glob(
-                    settings.STORAGE_CONFIG_FILES_GLOB, recursive=True
-                )
-            ]
-        }
+        if get_config("load_configs_multi_processed", 0):
+            with Pool() as p:
+                self._config_built_storages = {
+                    storage.get_storage_key(): storage
+                    for storage in p.map(
+                        build_storage_from_config,
+                        glob(settings.STORAGE_CONFIG_FILES_GLOB, recursive=True),
+                    )
+                }
+            for key in self._config_built_storages:
+                register_storage_key(key.value)
+        else:
+            self._config_built_storages = {
+                storage.get_storage_key(): storage
+                for storage in [
+                    build_storage_from_config(config_file)
+                    for config_file in glob(
+                        settings.STORAGE_CONFIG_FILES_GLOB, recursive=True
+                    )
+                ]
+            }
 
         # TODO: Remove these as they are converted to configs
         from snuba.datasets.storages.discover import storage as discover_storage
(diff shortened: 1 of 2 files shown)


snuba/datasets/storages/factory.py
```diff
@@ -33,15 +34,27 @@ def __init__(self) -> None:
...
+            with Pool() as p:
+                self._config_built_storages = {
+                    storage.get_storage_key(): storage
+                    for storage in p.map(
+                        build_storage_from_config,
+                        glob(settings.STORAGE_CONFIG_FILES_GLOB, recursive=True),
+                    )
+                }
+            for key in self._config_built_storages:
+                register_storage_key(key.value)
```
rahul-kumar-saini, line 47: Need to reregister the keys since multiprocessing only returns the built storage object and creates a copy of the `StorageKey` enum within the process which the key gets registered to, leaving the StorageKey enum in the main process unaltered.

evanh:
I'm down to try this. How do you want to measure the impact?

snuba/datasets/storages/factory.py
```diff
@@ -33,15 +34,27 @@ def __init__(self) -> None:
...
-            storage.get_storage_key(): storage
-            for storage in [
-                build_storage_from_config(config_file)
-                for config_file in glob(
-                    settings.STORAGE_CONFIG_FILES_GLOB, recursive=True
-                )
-            ]
-        }
+        if get_config("load_configs_multi_processed", 0):
+            with Pool() as p:
```
rahul-kumar-saini, line 38: By default, `Pool()` uses all available cores, should I specify an actual number of max processes (cores)?

rahul-kumar-saini:
@evanh 
> I'm down to try this. How do you want to measure the impact?

Just monitor the Snuba CLI init transaction, it should just measure this perfectly fine. It works locally, we just lose the individual spans for the configs and get the main "load factory" span but for an experiment I think that's ok.

Plan is:

- Enable multi processed config loading
- If nothing breaks, let it run overnight so we catch a `snuba optimize` aka Clickhouse merge which usually causes the p99 to spike to 12s+
- Watch Snuba CLI Initialization transactions to see if it helped

rahul-kumar-saini:
### Local results:
- Added a `sleep(1)` in build_storage_from_config() to emulate slow processing

Current version:
Screen Shot 2022-11-17 at 2 06 24 PM

Multi processed version:
image

evanh:
Seems like this speeds things up (👍 ) but also messes up the transaction? (👎 ) Shouldn't there be multiple "missing instrumentation" blocks in the second picture?

rahul-kumar-saini:
> Seems like this speeds things up (👍 ) but also messes up the transaction? (👎 ) Shouldn't there be multiple "missing instrumentation" blocks in the second picture?

I think each multi-processed call to `build_storage..` starts its own span which is lost when the process ends. The only thing that persists is what the function actually returns.

I need to figure out a way to propagate in the main transaction into the process or get the span out of the subprocess to attach it to the main transaction. I'm hoping this is a problem that has been solved already, I just wanna try running this experiment to validate the hypothesis. If it works I'd look into fixing the transaction before enabling it permanently.

I think the only time we get "missing instrumentation" is when Sentry expects a span but doesn't receive one? I'm not sure.

evanh (approved):
This is still worth testing, we can figure out the span problem if this turns out to be a useful solution.

fpacifici:
Screen Shot 2022-11-21 at 5 28 13 PM

Hi, I quickly ran `cProfile` on the snuba initialization. this is what came out. 
So most time is being spent in jsonschema validation/ I think you should look deeper into why. 

It is quite easy to reproduce:
```
import cProfile

            with cProfile.Profile() as pr:
                initialize.initialize_snuba()
            pr.dump_stats("stats.prof")
```
In snuba/cli/__init__.py


I can provide you the stats file but you can generate it with the code above I used `snakeviz` for visualization and `cProfile` to collect the stats 
https://docs.python.org/3/library/profile.html

I suggest you try to dig down the root cause of this behavior before just trying multi processing which would require us to allocate more CPU for the scripts in k8s (otherwise your processes would be throttled)

rahul-kumar-saini:
Closing as this is not necessary, the fast json schema changes were very effective. (#3434)

PR #3409, opened 2022-11-17
Add datadog events

dbanda:
Allows the datadog backend to send events. This is useful for setups where we may want to trigger alerts in datadog based on events.

### DIFF
--- a/snuba/utils/metrics/backends/datadog.py
+++ b/snuba/utils/metrics/backends/datadog.py
@@ -75,3 +75,19 @@ class DatadogMetricsBackend(MetricsBackend):
             tags=self.__normalize_tags(tags),
             sample_rate=self.__sample_rates.get(name, 1.0),
         )
+
+    def events(
+        self,
+        title: str,
+        text: str,
+        alert_type: str,
+        priority: str,
+        tags: Optional[Tags] = None,
+    ) -> None:
+        self.__client.event(
+            title=title,
+            text=text,
+            alert_type=alert_type,
+            tags=self.__normalize_tags(tags),
+            priority=priority,
+        )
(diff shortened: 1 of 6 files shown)


snuba/utils/metrics/backends/datadog.py
```diff
@@ -75,3 +75,19 @@ def timing(
...
+        priority: str,
+        tags: Optional[Tags] = None,
+    ) -> None:
+        self.__client.event(
+            title=title,
+            text=text,
+            alert_type=alert_type,
+            tags=self.__normalize_tags(tags),
+            priority=priority,
+        )
```
nikhars, line 93: The answer might be yes, but can you check that the datadog client version we use does actually support sending events?

snuba/utils/metrics/backends/testing.py
```diff
@@ -93,3 +93,20 @@ def timing(
...
+
+    def events(
+        self,
+        title: str,
+        text: str,
+        alert_type: str,
+        priority: str,
+        tags: Optional[Tags] = None,
+    ) -> None:
+        record_metric_call("events", title, 1, tags)
```
nikhars, line 105: You should probably define something like `record_event_call` and make it generic for events. Fitting events into metric does not seem right

snuba/utils/metrics/backends/datadog.py
dbanda: yes. confirmed

snuba/utils/metrics/backends/testing.py
```diff
@@ -13,12 +13,19 @@ class RecordedMetricCall:
     tags: Tags
 
 
+@dataclass(frozen=True)
+class RecordedEventCall:
+    value: str
```
nikhars, line 18: Why have a single value? Can't it actually contain the different fields like `title`, `text`, `alert_type` etc?

dbanda: I had it stored as a json string. Since this is mostly dummy data for testing I didn't bother too much with keeping its structure.

nikhars (approved):

PR #3411, opened 2022-11-18
feat(slicing): Introduce slicing into Subscriptions Executor

ayirr7:
This PR contains the changes necessary for slicing, at the level of subscription executors. Namely, we want to:

- Pass in `slice_id` optionally to the executor CLI
- Ensure that, when a `slice_id` is passed, the executor consumes from the "sliced" (physical) scheduled topic. These topics should be defined in slicing settings (`settings.SLICED_KAFKA_TOPIC_MAP`). Note that the result topic is not sliced, i.e. we use the common default across slices. 
- If no `slice_id` is passed, we should observe default behavior and the default scheduled and result topics should be used.

This work will synchronize with the slicing changes added to the subscriptions scheduler, once those are ready

### DIFF
--- a/snuba/cli/subscriptions_executor.py
+++ b/snuba/cli/subscriptions_executor.py
@@ -52,7 +52,7 @@ from snuba.utils.streams.topics import Topic
 )
 @click.option(
     "--slice-id",
-    help="The slice to write on/read data from",
+    help="The slice to load scheduled queries from",
 )
 @click.option(
     "--total-concurrent-queries",
--- a/snuba/cli/subscriptions_executor.py
+++ b/snuba/cli/subscriptions_executor.py
@@ -17,7 +17,6 @@ from snuba.subscriptions.executor_consumer import build_executor_consumer
 from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.streams.configuration_builder import build_kafka_producer_configuration
 from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
-from snuba.utils.streams.topics import Topic
 
 
 @click.command()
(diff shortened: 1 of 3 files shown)


snuba/cli/subscriptions_executor.py
```diff
@@ -49,6 +50,10 @@
     default="snuba-subscription-executor",
     help="Consumer group used for consuming the scheduled subscription topic/s.",
 )
+@click.option(
+    "--slice-id",
+    help="The slice to write on/read data from",
```
onewland, line 55: Let's call it "the slice to load scheduled queries from", because the executor is actually capable of executing queries for any slice (and during periods of transition it will execute queries across slices)

onewland (approved):
functionality seems fine to me

snuba/cli/subscriptions_executor.py
ayirr7: right, good point

onewland (commented):
Before merge, can you add the "slice_id" tag to the MetricsWrapper created by the CLI function?

Should be very small, like https://github.com/getsentry/snuba/pull/3374

snuba/cli/subscriptions_executor.py
```diff
@@ -116,10 +122,11 @@ def subscriptions_executor(
     stream_loader = storage.get_table_writer().get_stream_loader()
     result_topic_spec = stream_loader.get_subscription_result_topic_spec()
     assert result_topic_spec is not None
+    physical_topic_name = result_topic_spec.get_physical_topic_name(slice_id)
```
fpacifici, line 125: Just to ensure we are on the same page. 
Will it be possible to have a different scheduled topic per slice but share the result topic? 
the fact the snuba part of the pipeline (ingestion and subscription) is sliced does not mean the alert feature in the product is (which consumes the result topic)

nikhars: That should be the default behavior right now. The slice boundary ends at writing the result to the subscription result topic. All results have to be written to an unsliced topic name. 

@ayirr7 - What this means is that when you are resolving topic names based on slices, you should only do that for source topic names. Not for the result topic names.

ayirr7: Oh, right, good point. I didn't quite realize the slicing boundary ends for the results topics (they aren't sliced). Thanks for the clarifications.

nikhars (approved):

PR #3415, opened 2022-11-18
ref(querylog) Add pre-defined queries to Querylog Admin

evanh:
Similar to System Queries, add a selection of pre-defined queries in the
querylog tab with descriptions of what they mean and how to use them.

Add an endpoint that powers the selection, and tests for that endpoint as well
as the system queries equivalent. In order to avoid code duplication, also
change the system queries registry to use the RegisteredClass concept for
registering all the predefined queries. This could then be re-used for the
querylog queries.

querylog-tool


lynnagara (approved):
This is cool, thanks

snuba/admin/clickhouse/predefined_querylog_queries.py
```diff
@@ -0,0 +1,114 @@
...
+
+class QueryByID(QuerylogQuery):
+    """Find a query by its ID"""
+
+    sql = """
+    SELECT <fields>
+    FROM querylog_local
+    WHERE has(clickhouse_queries.query_id, '<UUID>')
+    AND timestamp > (now() - 60)
+    AND timestamp < toDateTime('2022-11-18T19:20:00')
```
nikhars, line 21: Any particular reason why you are using a specific end timestamp instead of now()

nikhars (approved):
Left a minor comment, else looks good. Thanks for adding this capability.

snuba/admin/clickhouse/predefined_querylog_queries.py
evanh: Just to show how to do that. Sometimes you want to isolate a specific time window.

PR #3427, opened 2022-11-21
feat(DC): All Metrics Storages as Configs

rahul-kumar-saini:
### Overview
Replaces:
- #3387 
- #3388 
- #3390
- #3391
- #3392


evanh (approved):
Why did you consolidate all those PRs?

rahul-kumar-saini:
> Why did you consolidate all those PRs?

Merging and deploying them one by one would take a long time, and I believe once #3434 is merged we don't need to worry about how much extra load putting all of these in at once will put on the init

lynnagara:
Was this written by hand or generated with a script? If script, can you please clarify in the description which one?

rahul-kumar-saini:
> Were these all written by hand or generated with a script? If script, can you please clarify in the description / commit message which one?

@lynnagara All were via script, updated the description!

PR #3432, opened 2022-11-22
fix(api-abuse): Add duration_ms to individual queries in querylog

enochtangg:
## Overview
The snuba querylog has a `duration_ms` which is the runtime of the entire snuba query. However, the runtime of the individual clickhouse query itself (time elapsed for Clickhouse to run a query on server) is not captured by querylog. This PR is responsible for adding this data to querylog.

## Before State
Querylog always writes `0` in `clickhouse_queries.duration_ms`

## After State
Querylog writes the elapsed time to run the clickhouse query in `clickhouse_queries.duration_ms`

## Testing
Tested locally to ensure messages produced to querylog topics are correct and contain accurate fields.

### DIFF
--- a/snuba/datasets/processors/querylog_processor.py
+++ b/snuba/datasets/processors/querylog_processor.py
@@ -59,8 +59,6 @@ class QuerylogProcessor(DatasetMessageProcessor):
             sql.append(query["sql"])
             status.append(query["status"])
             trace_id.append(str(uuid.UUID(query["trace_id"])))
-            # TODO: Calculate subquery duration, for now just insert 0s
-            duration_ms.append(0)
             stats.append(self.__to_json_string(query["stats"]))
             final.append(int(query["stats"].get("final") or 0))
             cache_hit.append(int(query["stats"].get("cache_hit") or 0))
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/querylog_processor.py
```diff
@@ -93,6 +91,7 @@ def __extract_query_list(
             groupby_columns.append(profile["groupby_cols"])
             array_join_columns.append(profile["array_join_cols"])
             bytes_scanned_columns.append(result_profile.get("bytes", 0))
+            duration_ms.append(result_profile.get("elapsed", 0))
```
nikhars, line 94: The result returned from clickhouse is in seconds while the column being written to is in milliseconds. https://clickhouse-driver.readthedocs.io/en/latest/features.html?highlight=elapsed#query-execution-statistics. You should convert it to milliseconds before writing to querylog

nikhars (changes requested):

snuba/datasets/processors/querylog_processor.py
enochtangg: Good catch. Does it make more sense to preserve the original nature of `ClickhouseProfile` and do the conversion in the querylog processor or before the message is produced to topic?

nikhars: What you did seems fine to me.

nikhars (approved):

PR #3434, opened 2022-11-22
feat(DC): Fast JSON Schema

rahul-kumar-saini:
### Overview
- The true bottleneck for Snuba init times is the JSON Schema validation
- Apparently a very common problem with plenty of solutions:
     - https://www.peterbe.com/plog/jsonschema-validate-10x-faster-in-python
- Locally I'm seeing a solid 50% reduction in load times by compiling validators once and reusing them across the configs
- Each new config should add very little to the init time as well since the bulk of the time will be spent compiling the validators in the beginning

### Previously, each time we validate a config the schema was compiled and validated against:
Screen Shot 2022-11-22 at 11 38 57 AM

### Now we compile the validators beforehand and just run them against each config:
Screen Shot 2022-11-22 at 12 01 33 PM

### Notes
- I removed the `build` span from everything since building the actual object takes <1ms

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,3 +1,4 @@
 jsonschema2md==0.4.0
+fastjsonschema==2.16.2
 myst-parser==0.18.0
 sphinx==5.1.1
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,4 +1,5 @@
 jsonschema2md==0.4.0
 fastjsonschema==2.16.2
+sentry-sdk==1.10.1
 myst-parser==0.18.0
 sphinx==5.1.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,7 @@ repos:
         name: validate-configs-syntax
         entry: python3 -m snuba.validate_configs
         language: python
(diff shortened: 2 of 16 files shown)


enochtangg:
LGTM!

docs-requirements.txt
```diff
@@ -1,3 +1,5 @@
 jsonschema2md==0.4.0
+fastjsonschema==2.16.2
```
evanh, line 2: Why do you need these in docs?

evanh (commented):
Why did you remove all the spans around building the Python objects (i.e. `description=f"Storage: {config['name']}")`?

rahul-kumar-saini:
> Why did you remove all the spans around building the Python objects (i.e. `description=f"Storage: {config['name']}")`?

I removed the `build` span from everything since building the actual object takes <1ms

I did add back the info of what storage is being loaded+validated in the description:

image

docs-requirements.txt
rahul-kumar-saini: Docs use the `json_schema.py` file to generate the docs, I moved the compilation of the schemas into validators into that file. All of that files dependencies need to be installed to run the docs command, I could move it out to another new file if this isn't a good solution.

lynnagara (approved):
Seems good to me

PR #3435, opened 2022-11-22
fix(admin): actually use slack channel id...

MeredithAnya:
I removed the passing of the slack channel in https://github.com/getsentry/snuba/pull/3406 or at least I THOUGHT I did, I think this change got lost in one of my rebases... FIXES SNUBA-30C


untitaker (approved):
🚢 🇮🇹

PR #3442, opened 2022-11-22
admin: run/reverse real migrations

dbanda:
Run real migrations and refresh the status table in the UI. Previously we were only running dry runs.

### DIFF
--- a/snuba/admin/static/clickhouse_migrations/index.tsx
+++ b/snuba/admin/static/clickhouse_migrations/index.tsx
@@ -42,16 +42,16 @@ function ClickhouseMigrations(props: { api: Client }) {
         `Migration ${migrationId} is blocking, are you sure you want to execute?`
       );
     }
-    console.log("executing !", action);
     executeRealRun(action);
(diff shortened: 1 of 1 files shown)


evanh (approved):
Excited for this!

snuba/admin/static/clickhouse_migrations/index.tsx
```diff
@@ -41,17 +43,16 @@ function ClickhouseMigrations(props: { api: Client }) {
       );
     }
     console.log("executing !", action);
+    executeRealRun(action);
```
MeredithAnya, line 46: we need to add the `force` flag or else this will not run for real, the dry run ignores the need for the `force` flag but we'll need it for real for the following situations

* a migration is blocking (already has a confirm dialog)
* trying to reverse a migration that has already be run (needs the confirm dialog)

dbanda: fixed. added the force arg

PR #3444, opened 2022-11-22
OPS-2635: setting up storage sets, migrations for snuba-access-logs

mwarkentin:
### OPS-2635

~~Depends on #3451~~

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -233,7 +233,12 @@ COLUMN_SPLIT_MAX_LIMIT = 1000
 COLUMN_SPLIT_MAX_RESULTS = 5000
 
 # Migrations in skipped groups will not be run
-SKIPPED_MIGRATION_GROUPS: Set[str] = {"querylog", "profiles", "functions", "access_logs", }
+SKIPPED_MIGRATION_GROUPS: Set[str] = {
+    "querylog",
+    "profiles",
+    "functions",
+    "access_logs",
+}
 
 MAX_RESOLUTION_FOR_JITTER = 60
 
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -74,6 +74,7 @@ CLUSTERS: Sequence[Mapping[str, Any]] = [
             "replays",
             "generic_metrics_sets",
             "generic_metrics_distributions",
+            "access_logs",
         },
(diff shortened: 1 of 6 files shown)


evanh:
I ran this locally (and fixed a few some errors). The migration runs correctly as far as I can tell.

I think this is ready for regular review.

mwarkentin:
@evanh looks like I might need to define some clusters settings?

https://github.com/getsentry/snuba/actions/runs/3549121456/jobs/5961125396

evanh:
New bug I'm trying to figure out.

`DB::Exception: Engine Distributed doesn't support TTL clause. Currently only the following engines have support for the feature: [MergeTree, ReplicatedVersionedCollapsingMergeTree, ReplacingMergeTree, ReplicatedSummingMergeTree, ReplicatedAggregatingMergeTree, ReplicatedCollapsingMergeTree, ReplicatedGraphiteMergeTree, ReplicatedMergeTree, ReplicatedReplacingMergeTree, VersionedCollapsingMergeTree, SummingMergeTree, GraphiteMergeTree, CollapsingMergeTree, AggregatingMergeTree]. Stack trace:
`

mwarkentin:
I think that makes sense, the distributed table wouldn't have data to expire. Guess we need to modify or duplicate the column definition to handle both engine types?

snuba/settings/__init__.py
```diff
@@ -227,9 +228,13 @@ class RedisClusters(TypedDict):
 COLUMN_SPLIT_MAX_LIMIT = 1000
 COLUMN_SPLIT_MAX_RESULTS = 5000
 
-# The migration groups that can be skipped are listed in OPTIONAL_GROUPS.
-# Migrations for skipped groups will not be run.
-SKIPPED_MIGRATION_GROUPS: Set[str] = {"querylog", "profiles", "functions"}
+# Migrations in skipped groups will not be run
+SKIPPED_MIGRATION_GROUPS: Set[str] = {
```
lynnagara, line 232: I think access_logs needs to be on the OPTIONAL_GROUPS list for skipping to be enabled on that group

evanh:
`DB::Exception: Sharding expression has type FixedString(32), but should be one of integer type. Stack trace:`

@mwarkentin I think you need a different sharding key.

PR #3445, opened 2022-11-22
ref(migrations): add migration target enum

dbanda:
A precursor to #3399 . This adds the target enum to specify which nodes to run an op on.

### Before:
The runner would pass in the local flag to let an op know which nodes to run on

### After:
Ops take in the target enum as a parameter. When executing ops already know what nodes to target. By default, the target enum is set to "unset" and will throw an error if the user does not explicitly set it.

### DIFF
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -3,7 +3,7 @@ from typing import Sequence
 
 from snuba.clusters.cluster import get_cluster
 from snuba.migrations.context import Context
-from snuba.migrations.operations import RunPython, SqlOperation
+from snuba.migrations.operations import OperationTarget, RunPython, SqlOperation
 from snuba.migrations.status import Status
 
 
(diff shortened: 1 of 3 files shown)


lynnagara:
The `RunSql` operation is missing the target arg.

dbanda:
> The `RunSql` operation is missing the target arg.

Added

snuba/migrations/migration.py
```diff
@@ -157,9 +157,11 @@ def forwards(self, context: Context, dry_run: bool = False) -> None:
 
         for op in ops:
             if op in local_ops:
-                op.execute(local=True)
+                op.target = OperationTarget.LOCAL
```
MeredithAnya, line 160: This is temporary right now right? until we add this to the migrations themselves?

dbanda: yes

snuba/migrations/operations.py
```diff
@@ -9,26 +10,51 @@
...
 
+class OperationTarget(Enum):
+    """
+    Represents the target nodes of an operation.
+    Either local, distributed.
+    """
+
+    LOCAL = "local"
+    DISTRIBUTED = "distributed"
+    UNSET = "unset"  # target is not set. will throw an error if executed
```
lynnagara, line 21: Mind commenting this as temporary / to be deprecated in code here?

lynnagara (approved):

PR #3449, opened 2022-11-23
admin: add migration audit log

dbanda:
Add audit log to the migration endpoints. It's logs requests to run/reverse migrations and sends a slack msg notification.


snuba/admin/notifications/slack/utils.py
```diff
@@ -38,6 +44,18 @@ def build_runtime_config_text(data: Any, action: AuditLogAction) -> Optional[str
...
+def build_migration_run_text(
+    data: Any, user: str, action: AuditLogAction
+) -> Optional[str]:
+    if action == AuditLogAction.RAN_MIGRATION:
+        action_text = f":runner: ran migration {data['migration']}"
+    elif action == AuditLogAction.REVERSED_MIGRATION:
+        action_text = f":uno-reverse: reversed migration {data['migration']}"
+    else:
+        return None
+    return f":warning: *Migration:* user '{user}' {action_text}  , force={data['force']}, fake={data['fake']}"
```
MeredithAnya, line 56: do you mind just adding like a couple new lines after the *Migration:* just so its a tad easier to read? 

This is what we working with: 
Screen Shot 2022-11-23 at 10 29 41 AM

otherwise I think it looks good!

MeredithAnya: also actually we do already put the user at the footer if you wanted to leave that out in the text above

locally it shows `by <>` but it will be `by accounts.google.com:meredith@sentry.io`

snuba/admin/notifications/slack/utils.py
```diff
@@ -38,6 +44,18 @@ def build_runtime_config_text(data: Any, action: AuditLogAction) -> Optional[str
         return None
 
 
+def build_migration_run_text(
+    data: Any, user: str, action: AuditLogAction
+) -> Optional[str]:
+    if action == AuditLogAction.RAN_MIGRATION:
+        action_text = f":runner: ran migration {data['migration']}"
+    elif action == AuditLogAction.REVERSED_MIGRATION:
+        action_text = f":uno-reverse: reversed migration {data['migration']}"
```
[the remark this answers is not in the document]
MeredithAnya: includes the group - the data["migration"] is actually the migration key which is the combo of the group and migration id

MeredithAnya (approved):
just had a couple nits about the display that are kind of just personal preference but otherwise ready to 🚢

snuba/admin/notifications/slack/utils.py
dbanda: Made some changes. It should look better now.

PR #3452, opened 2022-11-23
fix(migrations) Allow ttl_day on ModifyColumn operations

evanh:
There is a use case where someone wants to be able to give a daily TTL to a
specific column. The operation currently only allows monthly TTL.

Before:

Columns could only have monthly TTL if they needed a new  TTL from the table.

After:

The ModifyColumn operation takes an optional `ttl_day` parameter that works the
same way as `ttl_month` but on a daily interval instead of monthly.

### DIFF
--- a/snuba/migrations/operations.py
+++ b/snuba/migrations/operations.py
@@ -237,11 +237,14 @@ class ModifyColumn(SqlOperation):
         table_name: str,
         column: Column[MigrationModifiers],
         ttl_month: Optional[Tuple[str, int]] = None,
+        ttl_day: Optional[Tuple[str, int]] = None,
     ):
         super().__init__(storage_set)
         self.__table_name = table_name
         self.__column = column
+        assert not (ttl_month and ttl_day), "Only daily or monthly TTL allowed"
         self.__ttl_month = ttl_month
+        self.__ttl_day = ttl_day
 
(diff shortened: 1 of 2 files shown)


dbanda:
Maybe we should make this more general. The code will get messy if we have to have `ttl_day`, `ttl_month`, `ttl_year` ... Given that there are so many ways to express TTL, I think it would be better if we have a generic TTL expression. It could make use of the `tointerval` fuction which is used by clickhouse anyway. 
e.g
```[python]
def get_ttl(unit: str =  "Day" | "Month" | "Year" ... ,  val : int )
    return  f'tointerval{unit}({val})` 
```

I've learned the hard way that it is difficult to make changes to migrations once they get in and are used by the scripts.

MeredithAnya:
Did we not want to add the TTL as a it's own modifier?

lynnagara (changes requested):
Is this an urgent addition we need to make? If so, we can consider merging this as a hotfix but I think we need to build a system that works for all kinds of TTL expression rather than add to the existing (bad) convention. There are many more types of valid TTL expressions that still wouldn't work and I don't think we can keep increasing the number of arguments to this operation to cover all of them.

evanh:
Closing in favour of https://github.com/getsentry/snuba/pull/3451

PR #3454, opened 2022-11-23
fix(init): Snuba init timing metric

rahul-kumar-saini:
- Should use `timing` instead of `gauge` for a timing metric

### DIFF
--- a/snuba/cli/__init__.py
+++ b/snuba/cli/__init__.py
@@ -61,7 +61,7 @@ class SnubaCLI(click.MultiCommand):
                 code = compile(f.read(), fn, "exec")
                 eval(code, ns, ns)
             init_time = time.perf_counter() - start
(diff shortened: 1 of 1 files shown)


snuba/cli/__init__.py
```diff
@@ -61,7 +61,7 @@ def get_command(self, ctx: Any, name: str) -> click.Command:
                 code = compile(f.read(), fn, "exec")
                 eval(code, ns, ns)
             init_time = time.perf_counter() - start
-            metrics.gauge("snuba_init", init_time)
+            metrics.timing("snuba_init", init_time)
```
evanh, line 64: You might want to have a different name. I'm not sure how DD handles a metric with the same name but two different types.

evanh (approved):

PR #3457, opened 2022-11-29
feat(replays): Add replay_type column

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/210

@evanh Are enums unsupported?  Yaml validation throws.

### DIFF
--- a/snuba/datasets/configuration/replays/storages/replays.yaml
+++ b/snuba/datasets/configuration/replays/storages/replays.yaml
@@ -8,11 +8,6 @@ schema:
   columns:
     [
       { name: replay_id, type: UUID },
-      {
-        name: replay_type,
-        type: String,
-        args: { schema_modifiers: [nullable] },
-      },
       { name: event_hash, type: UUID },
       {
         name: segment_id,
--- a/snuba/datasets/configuration/replays/storages/replays.yaml
+++ b/snuba/datasets/configuration/replays/storages/replays.yaml
@@ -8,6 +8,11 @@ schema:
   columns:
     [
       { name: replay_id, type: UUID },
+      {
+        name: replay_type,
+        type: String,
+        args: { schema_modifiers: [nullable] },
+      },
       { name: event_hash, type: UUID },
       {
         name: segment_id,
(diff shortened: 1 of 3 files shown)


evanh:
Enums aren't supported properly in the configuration, I will add a task to get that fixed.

That being said, I would think carefully about whether you really want an Enum or not. Specifically, it's not trivial to add/remove values from the Enum once the column is created. If you wanted to support more values later on, you would need to do an `ALTER COLUMN`, which can be problematic to run in production, or add a new column with the supported types.

I would recommend leaving this column as a low cardinality string, that way you can easily update the allowed values in your processor.

snuba/datasets/configuration/replays/storages/replays.yaml
```diff
@@ -8,6 +8,11 @@ schema:
   columns:
     [
       { name: replay_id, type: UUID },
+      {
```
evanh, line 11: The storage change should be in the other PR. If you deploy this, your consumer might break until the migration is run.

snuba/snuba_migrations/replays/0007_add_replay_type_column.py
```diff
@@ -0,0 +1,74 @@
...
+        Column(
+            "replay_type",
+            Enum([("session", 1), ("error", 2)], Modifiers(nullable=True)),
+        ),
+        "replay_id",
+    )
+]
+
+
+class Migration(migration.ClickhouseNodeMigrationLegacy):
```
MeredithAnya, line 19: We are moving away from the `ClickhouseNodeMigrationLegacy` migrations, I think there is one last PR to be merged to be able to start using the new `ClickhouseNodeMigration`, I think this would be a good opportunity to use the new implementation provided that we get that PR merged cc @dbanda

dbanda: The PR is merged. To use `ClickhouseNodeMigration`, implement `forwards_ops` and `backwards_ops`. The `_local` and `_dist` functions are deprecated. Your ops now go in a single sequence.

dbanda: You use the `target=` parameter to your ops to specify if they should be local or distributed.

cmanallen: Looks great!  Love the new system.  Sounds like a good project for my next migration.  I have the generated SQL I need though.

evanh (approved):
If you don't want to use the new migration system for this, then this is good to go.

PR #3458, opened 2022-11-29
feat(replays): Add replay_type column processing

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/210


cmanallen:
Added replay_type to storage dataset per review feedback.

PR #3460, opened 2022-11-30
feat(data delete): allow arbitrary data deletions [WIP]

onewland:
This is a work-in-progress but I wanted to see what people think about the high-level algorithm in the `data_delete` method in `snuba/data_delete.py`. This current script will run a query against a set of parameters, waiting each time for the number of mutations in the system to be 0.

TODO:
- make the query template a parameter of the CLI
- make the input query parameters read from a file specified in the CLI
- time out with a failure if the number of mutations never reaches 0
- some kind of verification step with a second query template that makes sure the partitions are empty for all the input query parameters
- ???

### DIFF
--- a/snuba/cli/data_delete.py
+++ b/snuba/cli/data_delete.py
@@ -2,22 +2,48 @@ from typing import Optional
 
 import click
 
-from snuba.clusters.cluster import ClickhouseClientSettings
 from snuba.datasets.storages.factory import get_writable_storage
 from snuba.datasets.storages.storage_key import StorageKey
 from snuba.environment import setup_logging, setup_sentry
 
+TRANSACTIONS_DELETE_QUERY = """
+ALTER TABLE transactions_local
+DELETE WHERE
+    project_id = %(project_id)d
+    AND toMonday(finish_ts) = '%(to_monday_ts_lower)s'
+"""
+
+# FIXME: remove once I document how to get partition timestamps in an
+# easily repeatable way
+
+# fetch timestamps with:
+# SELECT partition, count(*)
+# from system.parts
+# where table='transactions_local'
+#    and active=1
+# group by partition
+# order by partition
+TEST_TIMESTAMPS = [
+    "2022-08-29",
+    "2022-09-05",
+    "2022-09-12",
+    "2022-09-19",
+    "2022-09-26",
+    "2022-10-03",
+    "2022-10-10",
+    "2022-10-17",
+    "2022-10-24",
+    "2022-10-31",
+    "2022-11-07",
+    "2022-11-14",
+    "2022-11-21",
+    "2022-11-28",
+]
+
+QUERY_PARAMS = [{"project_id": 2, "to_monday_ts_lower": ts} for ts in TEST_TIMESTAMPS]
+
 
 @click.command()
-@click.option(
-    "--clickhouse-host",
-    help="Clickhouse server to write to.",
-)
-@click.option(
-    "--clickhouse-port",
-    type=int,
-    help="Clickhouse native port to write to.",
-)
 @click.option(
     "--dry-run",
     type=bool,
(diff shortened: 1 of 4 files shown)


snuba/data_delete.py
```diff
@@ -0,0 +1,108 @@
...
+    connection: ClickhousePool,
+    query: str,
+    params: Mapping[str, Any],
+    dry_run: bool = False,
+) -> ClickhouseResult:
+    logger.info("data_delete.pre_delete", interpolated_query={query % params})
+    if dry_run:
+        return ClickhouseResult()
+    else:
+        return connection.execute(query, params, retryable=False)
```
fpacifici, line 60: Are you planning to execute the mutation per part or per partition ? And who is going to restrict the query to a specific partition.

snuba/cli/data_delete.py
```diff
@@ -0,0 +1,109 @@
...
+
+TRANSACTIONS_DELETE_QUERY = """
+ALTER TABLE transactions_local
+DELETE WHERE
+    project_id = %(project_id)d
+    AND toMonday(finish_ts) = '%(to_monday_ts_lower)s'
+"""
+
+# FIXME: remove once I document how to get partition timestamps in an
+# easily repeatable way
```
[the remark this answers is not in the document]
onewland: we're on the same page, this comment is a no longer relevant note-to-self.

snuba/data_delete.py
```diff
@@ -0,0 +1,108 @@
...
+        bind_contextvars(shard_id=shard)
+        logger.info("data_delete.shard.init")
+        replica_node = get_working_replica_node(
+            shard_replicas, ch_user, ch_pass, clickhouse_db
+        )
+        for query_parameter_row in query_parameters:
+            bind_contextvars(query_parameters=query_parameter_row)
+            assert_no_mutations_on_replica(replica_node)
+            execute_delete_command_on_replica(
+                replica_node, query_template, query_parameter_row, dry_run
```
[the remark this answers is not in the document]
onewland: I think the `connection.execute` call on line 60 will interpolate the params

snuba/data_delete.py
onewland: I thought that this had to be controlled by creating a safe input template and parameters. Are there actually query/DDL execution settings in ClickHouse that will let you manually limit the scope of a single query to a part or partition? 

Alternatively, were you hoping to programmatically verify that a delete will not span partitions by parsing the delete query conditions and comparing it to the table schema? That seems complicated to me, but I'd be fine with investigating.

nikhars: Ideally, this would be run per partition, not parts. Partitions are easier to keep track of. Can you check if our clickhouse version supports this https://clickhouse.com/docs/en/sql-reference/statements/alter/partition/#delete-in-partition. If it does, then you can get all the active partitions and apply the command in sequence to each one of them. You probably won't even need this condition then https://github.com/getsentry/snuba/pull/3460/files#diff-e6a3c68f04d4785448cb8190f9633b0e7109857bdbcaccc8898b96603bd1f89dR14.

snuba/data_delete.py
```diff
@@ -0,0 +1,108 @@
...
+
+    for (shard, shard_replicas) in cluster_nodes.items():
+        bind_contextvars(shard_id=shard)
+        logger.info("data_delete.shard.init")
+        replica_node = get_working_replica_node(
+            shard_replicas, ch_user, ch_pass, clickhouse_db
+        )
+        for query_parameter_row in query_parameters:
+            bind_contextvars(query_parameters=query_parameter_row)
+            assert_no_mutations_on_replica(replica_node)
```
nikhars, line 97: Instead of asserting here, why not keep looping until existing mutations are done like you do here https://github.com/getsentry/snuba/pull/3460/files#diff-ff88dadbd99312eff92430874f230151041fc27daaa5b81edb4d4a735aa832faR102-R104

onewland: I think because if you haven't reached a state at this point where `count(mutations) = 0`, it's an indication that somebody else is running the script (and you'd want to figure that out before continuing).

I think even these checks are not extremely safe (to prevent two scripts from running simultaneously) given that it's not possible to do an atomic `alter table delete IFF check for no mutations succeeds`

snuba/data_delete.py
onewland: I'll take a look at whether our ClickHouse versions support `DELETE IN PARTITION`

PR #3466, opened 2022-12-05
feat(api-abuse): Add rate limit page in Snuba admin

enochtangg:
### Overview
The management of rate limit configs is difficult in Snuba admin. More specifically, viewing applied rate limit configs can get mentally exhausting as rate limit configs are coupled with all runtime configs which might require some line-by-line parsing by on-call. This PR is a first iteration of what a Rate Limit page in Snuba admin might look like. This new page copies the Runtime Config page, removes all non-rate limit configs, and adds some filtering capabilities (by dimension, parameter, target).

rate_limit_faster


snuba/admin/static/ratelimit/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+  });
+
+  // All supported rate limiter dimensions and parameters in Snuba
+  const availableDimensions = [
+    "project_referrer",
+    "project",
+    "referrer",
+    "organization",
+  ];
+  const availableParameters = ["per_second_limit", "concurrent_limit"];
```
rahul-kumar-saini, line 89: What happens if a key contains some of these keywords but isn't a rate limit config? I can see `project`, `referrer` etc come up in unrelated configs.

snuba/admin/static/ratelimit/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+      }
+    });
+    setFilterTerms(filterTerm);
+
+    let searchData = data
+      .filter((obj) => {
+        if (filterTerm.dimension === "project") {
+          return (
+            obj.key.includes(filterTerm.dimension) &&
+            !obj.key.includes("project_referrer")
```
rahul-kumar-saini, line 177: `"project_referrer"`, `"organization"` etc should probably be constants

rahul-kumar-saini (commented):
I feel like most of the rate limit component is the same as the runtime config component so you could refactor out a lot of this code and reuse it in runtime config + ratelimit config

snuba/admin/static/ratelimit/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+    api.getConfigs().then((res) => {
+      let data = filterOnlyRateLimitConfigs(res);
+      populateFilterLists(data);
+      setSearchFilteredData(data);
+      setData(data);
+    });
+  }
+
+  function filterOnlyRateLimitConfigs(configs: Array<Config>) {
+    // Rate limit run time configs must include per_second_limit or concurrent_limit as substring
```
evanh, line 110: You should also include `thread_quota` here.

snuba/admin/static/ratelimit/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+  function fetchData() {
+    api.getConfigs().then((res) => {
+      let data = filterOnlyRateLimitConfigs(res);
+      populateFilterLists(data);
+      setSearchFilteredData(data);
+      setData(data);
+    });
+  }
+
+  function filterOnlyRateLimitConfigs(configs: Array<Config>) {
```
evanh, line 109: I thought we had talked about moving this functionality into the Python backend and either pass a parameter to the config endpoint to only return rate limiting or have a separate endpoint altogether?

lynnagara (changes requested):
Not a fan of this. It duplicates way too much code and even if it did not - makes it seem to the user like rate limit is something special other than config when it is not. Why not just add a simple filter to the page if you don't want to see the rest in this view?

evanh:
> makes it seem to the user like rate limit is something special other than config when it is not.

I don't think this is the right way to think about it. Rate limiting is a distinct use case of configuration, just like feature rollout is another. Separating those two things visually makes perfect sense: why would I want to sort through a bunch of rollout keys in order to check what rate limiting keys I have set, and vice versa?

Configuration in redis is a tool used by many different featrures of Snuba. Forcing every key of every feature to all be in the same place just because they happen to use the same tool is not a good way to manage our system.

evanh:
@lynnagara Would it make more sense if we namespaced the rate limit keys in Redis so the use case is clearly separated?

snuba/admin/static/ratelimit/index.tsx
enochtangg: yeah this is a good point, one thing change we've discussed offline is writing new endpoints which filter rate limit configs in the backend and serve that as a new API route to admin?

snuba/admin/static/ratelimit/index.tsx
enochtangg: We have, this was just the original PR. Probably should have been a draft. Those backend changes are still being made.

lynnagara:
@evanh The part that I'm opposed to is pretending that things are two different systems when they are not. If you think that the rate limit system should be a different one then let's make it a different one rather than put a bunch of filtering on strings (which is incredibly brittle) in front of runtime config.

evanh:
I agree with you about the front end filtering, I wanted to move the logic to the backend. What I'm trying to understand is, what would we have to change to make this "two different systems"? Is your concern just about doing filtering on the front end? What if the filtering was on the backend endpoint?

lynnagara:
If it was filtering alone on the backend endpoint, I think that would not be enough. A lot of the same issues would still remain since they are still the same redis database and namespace. For example: you enter a runtime config and if happens to have a specific format it ends up on the other rate limit page instead. This should never be possible in a separate system.

evanh:
OK, well that will require some deep thought on how to handle the data migration.

@lynnagara Do you think some version of this is worth launching now to help with managing the system? Or is it not worth the tech debt?

PR #3468, opened 2022-12-05
feat(search-issues): create new dataset to support issue platform search

barkbarkimashark:
Introduces a new dataset powering issue platform search for Profiling issues.

Will add more fleshed out tests for validating kafka message processing, wanted to land this PR first to unblock my team from using the new dataset.


snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,82 @@
...
+from snuba.consumers.types import KafkaMessageMetadata
+from snuba.datasets.events_format import EventTooOld, enforce_retention
+from snuba.datasets.processors import DatasetMessageProcessor
+from snuba.processor import InsertBatch, ProcessedMessage
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+metrics = MetricsWrapper(environment.metrics, "search_issues.processor")
+
+
+class SearchIssueMessage(TypedDict, total=False):
```
wedamija, line 14: Where does this get created? Currently, most event data is in the `data` key of the  message passed to kafka. I assumed we'd serialize the occurrence data and pass it via an `occurrence` key or something along those lines.

wedamija: https://github.com/getsentry/sentry/blob/f2c1f3648cb2000f12a4025b2e509ff16fbc8280/src/sentry/eventstream/snuba.py#L166-L197

Here's what the message looks like. Just not sure how that gets converted to this flattened message

barkbarkimashark: Yeah, we could go with that as well. I just have this class here as starting point to pull out data from the message. We can certainly update.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,96 @@
...
+                ),
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=StorageSetKey.SEARCH_ISSUES,
+                table_name="search_issues_dist",
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name="search_issues_local",
+                    sharding_key="cityHash64(project_id)",
```
barkbarkimashark, line 79: Not sure what the best sharding key is for this.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,96 @@
...
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.SEARCH_ISSUES,
+                table_name="search_issues_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(organization_id, project_id, toStartOfDay(detection_timestamp))",
+                    partition_by="(retention_days, toMonday(detection_timestamp))",
+                    # TODO: need to figure out what sample is here
+                    # sample_by="cityHash64(occurrence_id)",
```
barkbarkimashark, line 66: What's a good sample key for this dataset? What do we need to consider here?

snuba/datasets/processors/search_issues_processor.py
wedamija: Is the class here basically meant to be a reflection of what is sent via the Kafka message? Or is there a step where we translate it to this format?

barkbarkimashark: Yeah just a reference class so we know what fields are being sent. I made some changes to better reflect the messaging format we're more likely to be sending (based off the eventstream code).

snuba/snuba_migrations/search_issues/0002_search_issues_add_tags_hash_map.py
```diff
@@ -0,0 +1,49 @@
...
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Copy of 0004_transactions_add_tags_hash_map.py
+
+    """
```
barkbarkimashark, line 15: Since `_tags_hash_map` is materialized from `tags.key` and `tags.value` I think this needs to be a separate migration after 0001 runs. Lemme know if there's a way to combine these two migrations into one.

tests/datasets/configuration/test_storage_loader.py
```diff
@@ -129,7 +129,7 @@ def test_config_file_discovery(self) -> None:
             storage.get_storage_key() in CONFIG_BUILT_STORAGES
             for storage in self.python_storages
         )
-        assert len(CONFIG_BUILT_STORAGES) == len(self.python_storages)
+        assert len(CONFIG_BUILT_STORAGES) >= len(self.python_storages)
```
barkbarkimashark, line 132: This seems like a potential hack around this test. Since we're using config built storage for this dataset, we have can't import an instance of `Storage` to include in the list of storages.

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -0,0 +1,67 @@
...
+    ]
+  local_table_name: search_issues_local
+  dist_table_name: search_issues_dist
+query_processors:
+#  - processor: UUIDColumnProcessor
+#    args:
+#      columns: !!set
+#        trace_id: null
+#        transaction_id: null
+#        profile_id: null
```
volokluev, line 54: why is this commented out?

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,105 @@
...
+    event_id: Optional[str]
+    trace_id: Optional[str]
+    platform: str
+    environment: Optional[str]
+    release: Optional[str]
+    dist: Optional[str]
+    receive_timestamp: float
+    client_timestamp: float
+    tags: Mapping[str, Any]
+    user: Mapping[str, Any]  # user, user_hash, user_id, user_name, user_email
```
volokluev, line 29: why not use TypedDict here?

volokluev (commented):
Where can I find documentation about:

1. What queries you're expecting to be able to serve using this dataset?
2. How the topic you're going to consume from will be populated?

This will help answer some of the questions you're asking

tests/datasets/configuration/test_storage_loader.py
volokluev: yeah this is a bad test, we'll fix it

snuba/snuba_migrations/search_issues/0001_search_issues.py
nikhars: Historically, we have sharded the data so that it gets spread evenly across all shards of the cluster. I don't think there is a reason to deviate from that at the moment. So you could use `cityHash64(occurence_id)` here

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -0,0 +1,67 @@
...
+name: search_issues
+
+storage:
+  key: search_issues
+  set_key: search_issues
+
+schema:
+  columns:
+    [
+      { name: organization_id, type: UInt, args: { size: 64 } },
```
fpacifici, line 12: What is the rationale for adding the organization id to the table when we already have project id? 
Today all errors and transactions queries are based on project id alone and things are ok. 

Assuming org id will be in the order by key, if you add the organization id, you will always have to add a condition based on the organization to all your query, which in turn, will imply clickhouse will always load that column when processing your query, which has a performance impact on short and fast queries. 
This impact is generally useless as you already have the project id that serves the same purpose. 

Are there queries where you want to scan all project in an org without filtering them? Those will not be very efficient if we rely on organization id alone. This is because project id will be in the order by key, but it will not be in the query, thus no further field from the order by key will be used to narrow down the scope (including timestamp).

So I would suggest having only project id unless you plan to add organization without adding it to the order by key.

snuba/datasets/configuration/issues/entities/search_issues.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - processor: ProjectReferrerRateLimiter
+    args:
+      project_column: project_id
+  - processor: ProjectRateLimiterProcessor
+    args:
+      project_column: project_id
+  - processor: ResourceQuotaProcessor
+    args:
+      project_field: project_id
+  - processor: TimeSeriesProcessor
```
evanh, line 35: I would suggest that you don't need this processor. This allows some syntactic sugar, converting a "magic" column into a bucketed time (for example, magic `time` column becomes `toStartOfHour(timestamp)`). This logic isn't really useful for new Datasets since you can do that with SnQL now, and it will just cause headaches when trying to debug queries.

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,105 @@
...
+                    "project_id": event["project_id"],
+                    "organization_id": event["organization_id"],
+                    "group_id": group_id,
+                    "detection_timestamp": detection_timestamp,
+                    "retention_days": retention_days,
+                }
+                for group_id in group_ids or ()
+            ]
+        except IndexError:
+            metrics.increment("invalid_message")
```
evanh, line 94: I would recommend adding some more context to these errors. Either log something out so you can look in Sentry/the logs for what the error is, and/or add some low cardinality tags to the metrics call (for example, the key that is missing).

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,105 @@
...
+            return None
+        except ValueError:
+            metrics.increment("invalid_uuid")
+            return None
+        except KeyError:
+            metrics.increment("missing_field")
+            return None
+        except Exception:
+            metrics.increment("process_message_error")
+            raise
```
evanh, line 104: While you are initially testing this dataset, I would have nothing actually raise here. If this raises, the entire consumer will shut down and stop processing (since it will be stuck on the bad message).

snuba/snuba_migrations/search_issues/0002_search_issues_add_tags_hash_map.py
evanh: You could put the `AddColumn` operation after the `CreateTable` operation in your original migration. But leaving this separate seems a little cleaner, especially since `CreateTable` operations are pretty complicated.

snuba/snuba_migrations/search_issues/0001_search_issues.py
barkbarkimashark: For distributed aggregations, does each shard have to have all the rows in the group by clause to work effectively or is there some mechanism that would aggregate across shards?

Ex. 
search_issues_dist Shard 1:
------------------------------
occurrence_1, group_1
occurrence_2, group_2
occurrence_3, group_3
occurrence_4, group_4

search_issues_dist Shard 2:
------------------------------
occurrence_5, group_2
occurrence_6, group_2
occurrence_7, group_5
occurrence_8, group_6

query:
```
select group_id, count() as count 
from search_issues_dist
where group_id=group2
group by group_id
```

I think we mostly do aggregations by group_id for this dataset - would sharding on group_id make sense here?

barkbarkimashark: Would it make sense to shard on `project_id`, that way itll support aggregations on project and group between shards - assuming there is a limitation on aggregations across shards.

snuba/snuba_migrations/search_issues/0002_search_issues_add_tags_hash_map.py
barkbarkimashark: Ahh, right. Didn't realize I can have multiple operations in the same migration - that makes sense. I think I'll keep it as two though since it is cleaner.

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -0,0 +1,67 @@
...
+      { name: occurrence_type_id, type: UInt, args: { size: 8 } },
+      { name: detection_timestamp, type: DateTime },
+
+      { name: event_id, type: UUID, args: { schema_modifiers: [nullable] } },
+      { name: trace_id, type: UUID, args: { schema_modifiers: [nullable] } },
+      { name: platform, type: String, args: { schema_modifiers: [low_cardinality] } },
+      { name: environment, type: String, args: { schema_modifiers: [low_cardinality, nullable] } },
+      { name: release, type: String, args: { schema_modifiers: [low_cardinality, nullable] } },
+      { name: dist, type: String, args: { schema_modifiers: [low_cardinality, nullable] } },
+      { name: receive_timestamp, type: DateTime },
```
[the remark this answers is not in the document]
barkbarkimashark: That's the time when the event was ingested that resulted in the problem being detected.

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -0,0 +1,67 @@
...
+      { name: client_timestamp, type: DateTime },
+      { name: tags.key, type: Array, args: { inner_type: { type: String } } },
+      { name: tags.value, type: Array, args: { inner_type: { type: String } } },
+      { name: _tags_hash_map, type: Array, args: { inner_type: { type: UInt, args: { size: 64 } } } },
+      { name: user, type: String },
+      { name: user_id, type: UInt, args: { size: 64 } },
```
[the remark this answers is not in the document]
barkbarkimashark: Right. Will update.

snuba/datasets/configuration/issues/storages/search_issues.yaml
barkbarkimashark: Thanks for the context. I think dropping `organization_id` from the order by makes sense. I don't think we have any use-cases to scan all projects within an org. Would we be able to efficiently support something like that later on if we use `project_id` as the first order by column?

snuba/snuba_migrations/search_issues/0001_search_issues.py
barkbarkimashark: Hmm, when I tried running the migration locally with the `sample_by="cityHash64(occurrence_id)"` it errors out with ` Sampling expression must be present in the primary key.`. Does that mean the sampled column must be in the order by expression? What if we `sample_by="cityHash64(project_id)"? Would that be bad design?

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -0,0 +1,67 @@
...
+      { name: detection_timestamp, type: DateTime },
+
+      { name: event_id, type: UUID, args: { schema_modifiers: [nullable] } },
+      { name: trace_id, type: UUID, args: { schema_modifiers: [nullable] } },
+      { name: platform, type: String, args: { schema_modifiers: [low_cardinality] } },
+      { name: environment, type: String, args: { schema_modifiers: [low_cardinality, nullable] } },
+      { name: release, type: String, args: { schema_modifiers: [low_cardinality, nullable] } },
+      { name: dist, type: String, args: { schema_modifiers: [low_cardinality, nullable] } },
+      { name: receive_timestamp, type: DateTime },
+      { name: client_timestamp, type: DateTime },
```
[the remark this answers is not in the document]
barkbarkimashark: After talking with @wedamija  about `client_timestamp` I think we want to keep it non-null for now since we'll always have an event here to populate the timestamp for when it was captured, regardless of whether it was captured by the SDK or some other means.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,96 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.SEARCH_ISSUES,
+                table_name="search_issues_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(organization_id, project_id, toStartOfDay(detection_timestamp))",
```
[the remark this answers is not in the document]
fpacifici: While the organization_id issue was resolved, the uniqueness one was not.

snuba/datasets/processors/search_issues_processor.py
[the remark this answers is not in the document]
barkbarkimashark: Is this enough to configure a dead letter queue going:
https://github.com/getsentry/snuba/pull/3468/files#diff-a30dc1e9e23d00e1016c7b0293329c4c8e4dbc95f9b8d5067fd73646ce4af1e9R71-R73

Presumably, any uncaught exceptions raised here would get caught upstream and some handler would write to the dead letter queue? 

Should I just raise the IndexError, ValueError, etc after recording the metrics and/or logging?

barkbarkimashark:
Are there any considerations we should take with this new dataset to get replacements working? The volume of data this dataset should see is considerably lower than the current errors table.

snuba/snuba_migrations/search_issues/0001_search_issues.py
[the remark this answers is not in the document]
barkbarkimashark: What about `project_id, toStartOfDay(detection_timestamp)), occurrence_id`? If we primarily make queries with the `occurrence_id` missing, would that still be performant?

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+    Column("environment", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("release", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("dist", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("receive_timestamp", DateTime()),
+    Column("client_timestamp", DateTime()),
+    Column("tags", Nested([("key", String()), ("value", String())])),
+    Column("user", String(Modifiers(nullable=True))),
+    Column(
+        "user_hash", UInt(64, Modifiers(nullable=True, materialized="cityHash64(user)"))
+    ),
```
fpacifici, line 40: How are you planning to use this ?

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+    Nested,
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+columns: List[Column[Modifiers]] = [
```
fpacifici, line 19: You are missing the partition and offset columns to track down a row to the Kafka message. Please add them.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.SEARCH_ISSUES,
+                table_name="search_issues_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, toStartOfDay(detection_timestamp))",
```
fpacifici, line 65: What is the query pattern you expect ? 
The `order by` should reflect the query pattern coming from the client as in Clickhouse there is limited indexes support (you cannot create a BTree index on any column), data needs to be sorted in the optimal way for your queries. Then you can improve some query pattern that do not fit the sorting of the table through data skipping indexes but there as well, you need to be deliberate on which queries you are optimizing for.

An example I notice here is that there is nothing to make search by `group id` efficient. A query containing `group_id=x` would scan all the rows for the project involved in the time range provided. This may not be what you want. We have a minmax index on the group id column but this may or may not be the right solution here.

Also, I bet that you will never query across different issue types, which means that you will always have a condition like `issue type = x` in your query. This begs having the issue type in your sorting key.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+                    partition_by="(retention_days, toMonday(detection_timestamp))",
+                    sample_by="project_id",
```
fpacifici, line 67: I don't think this is a good sampling key. 
This means that, if you pass a `SAMPLE 0.5` to scan only 50% of the data the DB will scan 100% of data for 50% of the projects. Since all queries have a project id condition, it means that either sampling will be ignored or your results will be empty. 

https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/#sample-by

In a table like this, when you sample, you tend to want a random sample of your data, so you likely want to sample by occurrence_id. or event_id if you always want all occurrences for a specific event id to either be in the sample or not.
To summarize, the sample key is generally supposed to be a high cardinality column (with a lot of different values) to get a statistically relevant sample.

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,106 @@
...
+    def process_message(
+        self, message: Tuple[int, str, SearchIssueEvent], metadata: KafkaMessageMetadata
+    ) -> Optional[ProcessedMessage]:
+        if not (isinstance(message, (list, tuple)) and len(message) >= 2):
+            return None
+
+        # TODO: validate version
+        version = message[0]
+        if not version:
+            return None
```
fpacifici, line 60: Do you actually plan to support multiple versions ?

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,106 @@
...
+
+        type_, event = message[1:3]
+        if type_ != "insert":
+            return None
```
fpacifici, line 64: Having a type field in the processor was only specific to errors as they have replacers. Do you expect to implement replacement to mutate the table ?

snuba/datasets/processors/search_issues_processor.py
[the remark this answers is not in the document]
fpacifici: Correct. You let the exception bubble up and it will be caught by the DLQ and the appropriate policy will be applied according to what you configured.

fpacifici: Please let's not silently drop messages, The storage has a Dead Letter Queue configured, raising exceptions is ok.

snuba/utils/streams/topics.py
```diff
@@ -41,13 +41,15 @@ class Topic(Enum):
     GENERIC_METRICS_DISTRIBUTIONS_COMMIT_LOG = (
         "snuba-generic-metrics-distributions-commit-log"
     )
+    SEARCH_ISSUES = "search-issues"
```
wedamija, line 44: Was thinking we'd call this topic `eventstream-issue-platform`, although not too fussed either way. Just want to match this up to work in eventstream.

snuba/datasets/configuration/issues/storages/search_issues.yaml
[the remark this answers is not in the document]
rjo100: will be more useful for edge detection I believe

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("group_id", UInt(64)),
+    Column("search_title", String()),
+    Column("fingerprint", Array(String())),
+    Column("occurrence_id", UUID()),
+    Column("occurrence_type_id", UInt(8)),
```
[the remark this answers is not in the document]
rjo100: https://github.com/getsentry/sentry/blob/1ff09a4a380b651acc4872f26597271d2ea66d89/src/sentry/types/issues.py

issue type is currently an enum

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("group_id", UInt(64)),
+    Column("search_title", String()),
+    Column("fingerprint", Array(String())),
```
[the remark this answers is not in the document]
barkbarkimashark: As far as I know I don't believe we have any queries that utilize fingerprint. We might have plans in the future though. If performance is a concern could we always materialize the array into a denormalized column?

snuba/snuba_migrations/search_issues/0001_search_issues.py
barkbarkimashark: Tbh, this column schema was copied over from the errors/transaction table. I couldn't tell if existeing search queries on error issues or performance issues actually utilized `user_hash` column though.

snuba/snuba_migrations/search_issues/0001_search_issues.py
barkbarkimashark: ```If a sampling expression is used, the primary key must contain it. The result of a sampling expression must be an unsigned integer. Example: SAMPLE BY intHash32(UserID) ORDER BY (CounterID, EventDate, intHash32(UserID)).```

If we sample by `occurrence_id` it looks like we need to add it to the primary key. Are there concerns about adding `occurrence_id` to the last part of the primary key when most of our queries won't filter on `occurrence_id`?

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Yeah, it seems like a good idea if the payload evolves and we need to support multiple versions.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: For the short term (month or so) we don't, but I think we will have to later down the line.

snuba/snuba_migrations/search_issues/0001_search_issues.py
barkbarkimashark: What's the purpose of tracking down a row to a kafka message? Debugging?

snuba/snuba_migrations/search_issues/0001_search_issues.py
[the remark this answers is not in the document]
evanh: That begs a further question: how many fingerprints will there be in a given row? Having arbitrary sized arrays can lead to issues down the road.

snuba/utils/streams/topics.py
barkbarkimashark: Yup, renamed.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,97 @@
...
+    Column("occurrence_id", UUID()),
+    Column("occurrence_type_id", UInt(8)),
+    Column("detection_timestamp", DateTime()),
+    Column("event_id", UUID(Modifiers(nullable=True))),
+    Column("trace_id", UUID(Modifiers(nullable=True))),
+    Column("platform", String(Modifiers(low_cardinality=True))),
+    Column("environment", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("release", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("dist", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("receive_timestamp", DateTime()),
```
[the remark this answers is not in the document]
barkbarkimashark: The `detection_timestamp` would be the time when a vertical team processed an event and detected an issue, whereas `receive_timestamp` would be the event was ingested. Practically, the time difference between the two should be minimal since detection would likely happen close to when the event was ingested.

snuba/snuba_migrations/search_issues/0001_search_issues.py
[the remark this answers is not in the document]
barkbarkimashark: We won't enforce a limit but I don't see any reason a vertical team would be sending us a huge amount of fingerprints per issue. I think we can safely assume the size of the fingerprints array would be < 10 (or somewhere in that ballpark).

snuba/cli/consumer.py
```diff
@@ -46,6 +46,7 @@
     "storage_name",
     type=click.Choice(
         [storage_key.value for storage_key in get_writable_storage_keys()]
+        + ["search_issues"]
```
barkbarkimashark, line 49: Temporary workaround to enable running the consumer. Will remove this when the choices for storage options load from configs properly.

snuba/datasets/configuration/issues/entities/search_issues.yaml
```diff
@@ -0,0 +1,101 @@
+version: v1
+kind: entity
+name: search_issues
+
+schema:
+  [
+    { name: organization_id, type: UInt, args: { size: 64 } },
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: group_id, type: UInt, args: { size: 64 } },
```
lynnagara, line 9: No primary hash? Just group_id and fingerprints?

wedamija: What do we actually use `primary_hash` for in snuba? I don't think it's really used much in any search functionality.

If we want to include it, then we can extract it out of `fingerprints[0]` or make sure we pass it via eventstream.

barkbarkimashark: 👍 will add it now.

lynnagara:
If I'm not mistaken the correct spelling of the word occurrence is with a double `r`. This PR uses an incorrect spelling with single `r` throughout - could you please update?

snuba/cli/consumer.py
fpacifici: cc @evanh

snuba/utils/streams/topics.py
```diff
@@ -41,13 +41,15 @@ class Topic(Enum):
     GENERIC_METRICS_DISTRIBUTIONS_COMMIT_LOG = (
         "snuba-generic-metrics-distributions-commit-log"
     )
+    SEARCH_ISSUES = "eventstream-issue-platform"
```
lynnagara, line 44: Should this just be `issue-platform`? Afaik, all events go through the eventstream so I'm not sure that part is particularly meaningful here.

snuba/datasets/configuration/issues/entities/search_issues.yaml
lynnagara: Can you clarify the relationship between the occurrence_id and the primary hash / group id?

wedamija: An event can have multiple occurrences in it. An occurrence currently just has one fingerprint, although we left this as a list for use in the future. Each occurrence represents an issue happening one time.

So in terms of how they relate:
 - An occurrence will have a primary hash that is just the single fingerprint. These will never change
 - An occurrence is related to a group id in the same way an error event is related to a group. So in the future, we'd support merging/deletions and use replacements.

snuba/datasets/configuration/issues/entities/search_issues.yaml
```diff
@@ -0,0 +1,101 @@
...
+    { name: search_title, type: String },
+    { name: fingerprint, type: Array, args: { inner_type: { type: String } } },
+    { name: occurrence_id, type: UUID },
+    { name: occurrence_type_id, type: UInt, args: { size: 8 } },
```
lynnagara, line 13: What is the ocurrence_type_id? In other documents related to this schema that I've seen, this was listed as a UInt64. Why did it change to UInt8?

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,100 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.SEARCH_ISSUES,
+                table_name="search_issues_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, toStartOfDay(detection_timestamp), cityHash64(occurrence_id))",
```
lynnagara, line 68: Just curious about the choice of putting the detection timestamp in the primary key (instead of received like our other tables usually do). Do you anticipate wanting to query by this "detected_time" field?

snuba/cli/consumer.py
lynnagara: Reading though https://github.com/getsentry/snuba/blob/master/snuba/datasets/storages/factory.py#L132, it seems like this should not be needed as long as you turn on ENABLE_DEV_FEATURES and add it to the dev storages list

tests/helpers.py
```diff
@@ -43,7 +43,8 @@ def write_unprocessed_events(
 
 
 def write_raw_unprocessed_events(
-    storage: WritableStorage, events: Sequence[Union[InsertEvent, Mapping[str, Any]]]
+    storage: WritableStorage,
+    events: Sequence[Union[InsertEvent, Mapping[str, Any], Any]],
```
lynnagara, line 47: Any chance we can do better than `Any` for the message type?

snuba/snuba_migrations/search_issues/0001_search_issues.py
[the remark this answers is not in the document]
evanh: Is there a limit being enforced somewhere? There should be a hard cap on this somewhere. If one doesn't already exist I would suggest adding something to your processor to cap the array. It can be a high limit, just something to prevent some crazy edge case from writing a huge array.

snuba/cli/consumer.py
evanh: This PR should fix this bug: https://github.com/getsentry/snuba/pull/3502

snuba/datasets/configuration/issues/entities/search_issues.yaml
wedamija: It's the type of the occurrence, it won't have many unique values. The values represent the specific issue type, so `n+1 database query`, `file io on main thread`, etc. I think it was a mistake that we listed it as a UInt64 in the spec doc.

snuba/snuba_migrations/search_issues/0001_search_issues.py
wedamija: I think that it probably makes sense for this to be received. Users care most about when the thing happened, not when we actually detected it.

snuba/snuba_migrations/search_issues/0001_search_issues.py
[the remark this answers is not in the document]
wedamija: I think it probably makes sense to limit this here just to be thorough and in our ingestion consumer during our validation step. Let's just put a limit of 100?

snuba/datasets/configuration/issues/entities/search_issues.yaml
lynnagara: Ok

barkbarkimashark: Yeah, that was me not updating the doc. It should be UInt8.

snuba/snuba_migrations/search_issues/0001_search_issues.py
barkbarkimashark: Yeah, that's fair. Will use received_timestamp instead.

snuba/utils/streams/topics.py
barkbarkimashark: @wedamija what do you think about `ISSUE_PLATFORM = "issue-platform"` or `ISSUE_PLATFORM = "issue-platform-occurrence"`?

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,102 @@
...
```
nikhars, line 70: Since you have added `primary_hash` column, you would also need to add it to the order_by key in order for replacements to work.

snuba/snuba_migrations/search_issues/0002_search_issues_add_tags_hash_map.py
```diff
@@ -0,0 +1,49 @@
...
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.datasets.storages.tags_hash_map import TAGS_HASH_MAP_COLUMN
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Copy of 0004_transactions_add_tags_hash_map.py
```
nikhars, line 13: Nit: Comment is wrong

snuba/cli/consumer.py
barkbarkimashark: Looking at the factory and how its initialized, I can't tell how to register a storage as a `_dev_writable_storages`. 

Could we keep this workaround for this PR since it looks like that PR is still awaiting a merge to master.

We're looking to get this PR merged to start testing the whole flow within sentry.

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -0,0 +1,179 @@
...
+    def process_message(
+        self, message: Tuple[int, str, SearchIssueEvent], metadata: KafkaMessageMetadata
+    ) -> Optional[ProcessedMessage]:
+        if not (isinstance(message, (list, tuple)) and len(message) >= 2):
+            raise InvalidMessageFormat(
+                f"Expected message format (<version:int>, <operation:str>, <event>>)), got {message} instead"
+            )
+
+        version = message[0]
+        if not version or version != 2:
```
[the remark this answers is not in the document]
barkbarkimashark: Yeah, its more of an artifact of how we've structured inserting to the eventstream. Basically we'll be piggybacking the eventstream impl which is also sending over errors and transactions. We might look to split up the versioning if we start introducing breaking protocol changes, but as it currently stands the message format should work for all events.

snuba/snuba_migrations/search_issues/0001_search_issues.py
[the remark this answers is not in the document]
barkbarkimashark: Sounds good. Will introduce a hard limit and drop any fingerprints greater than 100.

snuba/utils/streams/topics.py
wedamija: let's call it `issue-occurrences`? Or maybe `generic-events` to match what we're calling the events.

snuba/snuba_migrations/search_issues/0001_search_issues.py
```diff
@@ -0,0 +1,102 @@
...
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("group_id", UInt(64)),
+    Column("search_title", String()),
+    Column("primary_hash", UUID()),
+    Column("fingerprint", Array(String())),
+    Column("occurrence_id", UUID()),
+    Column("occurrence_type_id", UInt(8)),
+    Column("detection_timestamp", DateTime()),
+    Column("timestamp", DateTime(Modifiers(materialized="receive_timestamp"))),
```
lynnagara, line 29: Curious why you needed this materialization? Isn't this kind of a waste to keep 2 copies of the same thing?

barkbarkimashark: This was more of an assumption on my part on how snuba_sdk builds and enforces queries with the time column.
https://github.com/getsentry/snuba-sdk/blob/8b8fd04a56f8afb3ca9909ad26b937b595360e2d/snuba_sdk/entity.py#L65-L66

I was thinking of adding a default value of `timestamp` there so any new datasets spun up would have a reasonable default for a time column. 

Unfortunately, our search code is fairly old and there's a lot of .. accumulated logic there that we need to keep for the short term.

snuba/utils/streams/topics.py
barkbarkimashark: Ok. To be consistent, I'll go with `generic-events`.

snuba/snuba_migrations/search_issues/0001_search_issues.py
wedamija: I think `timestamp` should just be the value of `client_timestamp` tbh. I'm pretty sure that's what it is in the errors table. I think I was also confused before, and the `order_by` should be basically the same as errors:

`order_by="(project_id, toStartOfDay(timestamp), primary_hash, cityHash64(event_id))`

lynnagara: Snuba (and snuba sdk) does enforce a time column, but the time column can be anything you want, it doesn't have to be called `timestamp`. If you absolutely must have the other column, how about making it an expression that just points to `receive_timestamp` instead of a materialization that keeps another copy of the data?

nikhars: The comment about the order_by change is a blocker if you want to support replacements in the future. Once an order by field is setup on a table, you cannot change it.

wedamija: We're ok with rebuilding the table from scratch short term, but we may as well fix this now since we're aware of it.

tests/helpers.py
barkbarkimashark: Ended up using `write_unprocessed_events` instead and reverted the typing here.

snuba/datasets/configuration/issues/storages/search_issues.yaml
barkbarkimashark: It looks like eventstream send the `organization__id` as part of the payload. I don't think we have any plans to make queries aggregated by the `organization_id` but filling in the column value doesn't look like a problem. I think we can keep it for now and drop it later instead of removing it now and potentially adding it back in the future.

lynnagara:
Are there any more changes planned here or are you looking for a re-review from sns now?

barkbarkimashark:
> Are there any more changes planned here or are you looking for a re-review from sns now?

I think its in a decent state right now for a re-review. I think there are straggling minor stuff about some columns but we can always make modifications and re-build the table if we need to. Please take a look again and see if its in a good enough state to merge. Thanks!

lynnagara (approved):

nikhars (approved):

PR #3469, opened 2022-12-05
fix(api abuse): Track the max threads per referrer

evanh:
This is a convenience so we can get some insight into how the resource quota
processor is working. It should be removed when the feature becomes widespread.


volokluev (approved):

evanh:
This was created to initially debug a supposed bug in the max thread usage. It turns out that was a PEBKAC problem, so this is no longer necessary.

PR #3473, opened 2022-12-06
ref(admin) Add a CSV copy button for querylog results

evanh:
Sometimes I find it easier to paste the results into a spreadsheet or another
tool. This saves me having to manually format the JSON output into a CSV.

Before state:

There was just a "Copy to clipboard" button that only returned the JSON formatted results.

After state:

There are two buttons, one which returns the JSON formatted results and one which returns the CSV formatted results.

copy-to-csv


volokluev (approved):
hell yeah

PR #3475, opened 2022-12-07
ref(slicing): Make Storage Set the level of Slicing

ayirr7:
As per design, slicing occurs at the level of storage set. However, this wasn't reflected entirely in slicing-related code changes, in slicing config, and in further usages. This PR attempts to do small refactors and renamings to make it clear that slicing indeed occurs by storage sets, not storages. Docs are and will continue to update accordingly to reflect this.


onewland (commented):
I think this looks fine. Did you test that the consumer starts properly with this change?

ayirr7:
> Did you test that the consumer starts properly with this change?

It appears to work as expected, yes

onewland (approved):

PR #3476, opened 2022-12-07
Add admin UI build to docker image SNS-1479

dbanda:
Currently the bundle must be manually checked into source control.  This will fix it to be part of the image build process.


evanh (approved):

Dockerfile
```diff
@@ -53,7 +53,25 @@ ENV SNUBA_RELEASE=$SOURCE_COMMIT \
...
+    curl -fsSL https://deb.nodesource.com/setup_${NODE_VERSION}.x | bash - &&\
+    apt-get update && \
+    apt-get install -y yarn nodejs --no-install-recommends && \
+    cd snuba/admin && \
+    yarn install && \
+    yarn run build && \
+    yarn cache clean && \
+    rm -rf node_modules && \
+    apt-get purge -y --auto-remove yarn curl nodejs gnupg && \
+    rm -rf /var/lib/apt/lists/*
```
onewland, line 74: Based on the line 38 comment, could we move the install of the tools/libraries to above where we copy the source over, and at this point only run the necessary yarn commands?

I could see this making CI runs a lot slower

dbanda: It has to execute after the copy otherwise there are no files to build. The Layer cache gets invalidated after the COPY every time because any change to the files in the snuba repo will will make docker redo the COPY. Definitely makes the CI slower but I'm not sure if there's a way to work around it.

onewland: I agree that 

```
    cd snuba/admin && \
    yarn install && \
    yarn run build && \
    yarn cache clean && \
    rm -rf node_modules && \
    apt-get purge -y --auto-remove yarn curl nodejs gnupg && \
    rm -rf /var/lib/apt/lists/*
```

has to run after the copy but I think all the other commands above those can run earlier and it's the fetches from apt that I worry about being slow

onewland: Also if you look at lines 9-36 there's some more complex logic (which I'm not saying is necessary) for copying over python's `requirements.txt` first and then doing the install of those dependencies before copying application logic over.

That's probably the ideal way to handle the JS dependencies but I think it might be overkill for how long `yarn install` takes.

dbanda: I could put the fetches from apt above copy but that would make the image bigger since the layer would be split. 

Instead, I moved the layer above the copy and did a copy of `snuba/admin` first to do the build. That should be fast and keep the size small, however it introduced the risk of unintentionally overwriting the bundle with the second copy, so I added bundle to the .dockerignore.

onewland:
@lynnagara you originally added the .bundle.js IIRC. are you good with this change?

I'm fine with it if you are.

lynnagara (approved):
Awesome, thanks!

PR #3478, opened 2022-12-07
feat(DAC): make storage file for discover

volokluev:
converted and verified using scripts in  #3336


volokluev:
closed due to #3480

PR #3481, opened 2022-12-07
feat(DC): Outcomes Storages as a Config

rahul-kumar-saini:
### Overview
- Adding Outcomes storages as a config
- Storage config were generated and tested using scripts from #3336 

### Changes
- Any call to `get_storage()` with an outcomes storage key will return the storage built from config instead of the hardcoded version

### Testing Notes
- How config was generated (eg outcomes_raw):
    - `python3 scripts/pystorage_2_yaml.py outcomes_raw raw.yaml`
    - Removed quotes around `columns` in schema, format yaml file with Prettier
- How config was tested:
    - `python3 scripts/check_yaml_against_code.py outcomes_raw raw.yaml`
    - Updated `tests/datasets/configuration/test_storage_loader.py`

### DIFF
--- a/tests/datasets/configuration/test_storage_loader.py
+++ b/tests/datasets/configuration/test_storage_loader.py
@@ -43,6 +43,8 @@ from snuba.datasets.storages.metrics import (
 from snuba.datasets.storages.metrics import org_counters_storage
 from snuba.datasets.storages.metrics import polymorphic_bucket as metrics_raw
 from snuba.datasets.storages.metrics import sets_storage as metrics_sets
+from snuba.datasets.storages.outcomes import materialized_storage as outcomes_hourly
+from snuba.datasets.storages.outcomes import raw_storage as outcomes_raw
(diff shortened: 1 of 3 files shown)


tests/datasets/configuration/test_storage_loader.py
```diff
@@ -134,6 +136,8 @@ class TestStorageConfiguration(ConfigurationTest):
         metrics_distributions_storage,
         metrics_raw,
         org_counters_storage,
+        outcomes_hourly,
+        outcomes_raw,
```
volokluev, line 140: while we're here can we fix this test? If the storage never existed in code (which new ones won't) it won't work

https://github.com/getsentry/snuba/pull/3468#discussion_r1041551487

rahul-kumar-saini: I've removed that check entirely since it assumes the only config storages are ones that have a python version.

evanh (approved):

PR #3483, opened 2022-12-08
feat(CDC): Dynamic Row Processors

rahul-kumar-saini:
### Overview
- Currently CDC Storages hardcode row processors as lambda functions
- It would be nice to have these defined in one place and referred to by some key to make it possible to convert CDC storages to config

### Changes
- `CdcRowProcessor` is now a new registered class replacing `RowProcessor`


volokluev:
Can we just continue the RegisteredClass pattern and make these into classes we can reference from config? I don't see much of a point in adding this special case when there's an established pattern for this

evanh (approved):
Thanks for that, this looks good.

PR #3488, opened 2022-12-08
Add test migration

dbanda:
Add a simple test migration. This migration creates an empty table in the querylog storage set and adds a column to it. We can use this to test migrations tool on a real cluster.

### DIFF
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -20,6 +20,7 @@ class MigrationGroup(Enum):
     FUNCTIONS = "functions"
     REPLAYS = "replays"
     GENERIC_METRICS = "generic_metrics"
+    TEST_MIGRATION = "test_migration"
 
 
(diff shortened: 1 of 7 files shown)


MeredithAnya:
> This is going to create this table for ST and open source customers as well. I think you should disable it by default and then you can override that in our prod settings.

yeah we can just add it to `SKIPPED_MIGRATION_GROUPS`

dbanda:
> > This is going to create this table for ST and open source customers as well. I think you should disable it by default and then you can override that in our prod settings.
> 
> yeah we can just add it to `SKIPPED_MIGRATION_GROUPS`
> 
> in prod it won't be included in `SKIPPED_MIGRATION_GROUPS` so we should still see this in admin

added

evanh (approved):

MeredithAnya (approved):

PR #3489, opened 2022-12-08
ref(abuse) Add more predefined querylog queries

evanh:
Add two new queries to help narrow down with referrers might be causing the
abuse. To use these queries you should already know which cluster is
experiencing the abuse, and at what time the abuse started.

The idea behind the queries is: take half an hour before the abuse started, and
half an hour after. Compare the referrers between those two time periods, and
surface the ones that had the highest percentage change from before to after.

The purpose of these is not to automatically find the smoking gun, but to at
least remove some of the noise and show some actionable signal that can be
followed up on to confirm if one of the referrers actually is the source of
the problem.

I tested these queries on some known cases of abuse, and they did surface the
referrers that turned out to be the problem.

### DIFF
--- a/snuba/admin/clickhouse/predefined_querylog_queries.py
+++ b/snuba/admin/clickhouse/predefined_querylog_queries.py
@@ -112,3 +112,57 @@ class QueryDurationForReferrerByProject(QuerylogQuery):
     )
     ORDER BY projects ASC, time ASC
     """
+
+
+class BeforeAfterBytesScannedComparison(QuerylogQuery):
+    """Given a certain time that abuse started on a certain cluster, specify a time range before and after the abuse (recommend 30 minutes). This will show the referrers with the largest change in bytes_scanned. In the example, the abuse started on Dec 7. at 21:54:50."""
+
+    sql = """
+    SELECT referrer, before_scanned, after_scanned, (after_scanned - ifNull(before_scanned, 0)) as diff, (diff / if(equals(after_scanned, 0), 1, after_scanned))  * 100 as pct_diff
+    FROM
+    (
+        SELECT referrer, sum(arrayReduce('sum', clickhouse_queries.bytes_scanned)) as after_scanned
+        FROM querylog_local
+        WHERE timestamp >= toDateTime('2022-12-07T21:55:00')
+        AND timestamp <= toDateTime('2022-12-07T22:20:00')
+        AND dataset IN ('events', 'discover')
+        GROUP BY referrer
+    ) `after` LEFT OUTER JOIN
+    (
+        SELECT referrer, sum(arrayReduce('sum', clickhouse_queries.bytes_scanned)) as before_scanned
+        FROM querylog_local
+        WHERE timestamp >= toDateTime('2022-12-07T21:24:00')
+        AND timestamp <= toDateTime('2022-12-07T21:54:00')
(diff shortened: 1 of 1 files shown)


snuba/admin/clickhouse/predefined_querylog_queries.py
```diff
@@ -112,3 +112,57 @@ class QueryDurationForReferrerByProject(QuerylogQuery):
...
+    """Given a certain time that abuse started on a certain cluster, specify a time range before and after the abuse (recommend 30 minutes). This will show the referrers with the largest change in bytes_scanned. In the example, the abuse started on Dec 7. at 21:54:50."""
+
+    sql = """
+    SELECT referrer, before_scanned, after_scanned, (after_scanned - ifNull(before_scanned, 0)) as diff, (diff / if(equals(after_scanned, 0), 1, after_scanned))  * 100 as pct_diff
+    FROM
+    (
+        SELECT referrer, sum(arrayReduce('sum', clickhouse_queries.bytes_scanned)) as after_scanned
+        FROM querylog_local
+        WHERE timestamp >= toDateTime('2022-12-07T21:55:00')
+        AND timestamp <= toDateTime('2022-12-07T22:20:00')
```
volokluev, line 127:
```suggestion
        WHERE timestamp >= toDateTime(<time_str_after_abuse>)
        AND timestamp <= toDateTime(<time_str_after_abuse + 30 mins)
```

volokluev: Making these explicit descriptive parameters may be easier for the user

snuba/admin/clickhouse/predefined_querylog_queries.py
```diff
@@ -112,3 +112,57 @@ class QueryDurationForReferrerByProject(QuerylogQuery):
...
+        AND dataset IN ('events', 'discover')
+        GROUP BY referrer
+    ) `after` LEFT OUTER JOIN
+    (
+        SELECT referrer, sum(arrayReduce('sum', clickhouse_queries.bytes_scanned)) as before_scanned
+        FROM querylog_local
+        WHERE timestamp >= toDateTime('2022-12-07T21:24:00')
+        AND timestamp <= toDateTime('2022-12-07T21:54:00')
```
volokluev, line 135:
```suggestion
        WHERE timestamp >= toDateTime(<time_str_before_abuse>)
        AND timestamp <= toDateTime(<time_str_before_abuse + 30mins>)
```

snuba/admin/clickhouse/predefined_querylog_queries.py
volokluev: I filed a [ticket](https://getsentry.atlassian.net/browse/SNS-1931) to make this a feature in the tool, but I do think this is still an improvement

volokluev (approved):
cool!

snuba/admin/clickhouse/predefined_querylog_queries.py
evanh: I deliberately left these like this, since I find it easier when I have an actual example to edit. Otherwise I have to remind myself of the exact date format Clickhouse expects.

I would like to have our tools ultimately work using parameters though, in a similar way to how saved queries work on Redash. Users fill out a form with the values they want and we template them into the query.

PR #3490, opened 2022-12-08
test(slicing): Add more validation for slicing

ayirr7:
This PR:

- Adds validation for a few other cases with sliced/regular Kafka broker configs and sliced Clusters settings
- Also adds the corresponding tests to catch any bad configuration in settings

### DIFF
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -118,12 +118,13 @@ def validate_slicing_settings(locals: Mapping[str, Any]) -> None:
             topic_tuple in locals["SLICED_KAFKA_BROKER_CONFIG"]
         ), f"missing broker config definition for sliced Kafka topic {topic_tuple[0]} on slice {topic_tuple[1]}"
 
-    _STORAGE_SET_CLUSTER_MAP: Dict[str, Mapping[str, Any]] = {}
+    _STORAGE_SET_CLUSTER_MAP = {
+        storage_set: cluster
+        for cluster in locals["CLUSTERS"]
+        for storage_set in cluster["storage_sets"]
+    }
 
     _SLICED_STORAGE_SET_CLUSTER_MAP: Dict[Tuple[str, int], Mapping[str, Any]] = {}
-    for cluster in locals["CLUSTERS"]:
-        for storage_set in cluster["storage_sets"]:
-            _STORAGE_SET_CLUSTER_MAP[storage_set] = cluster
 
     for sliced_cluster in locals["SLICED_CLUSTERS"]:
         for storage_set_tuple in sliced_cluster["storage_set_slices"]:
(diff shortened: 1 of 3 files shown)


ayirr7:
Update: Added validation to ensure that all Storage Set Keys have the same `single_node` value across their associated clusters.

nikhars (approved):
Thanks for testing against prod configuration

snuba/settings/validation.py
```diff
@@ -117,3 +117,48 @@ def validate_slicing_settings(locals: Mapping[str, Any]) -> None:
         assert (
             topic_tuple in locals["SLICED_KAFKA_BROKER_CONFIG"]
         ), f"missing broker config definition for sliced Kafka topic {topic_tuple[0]} on slice {topic_tuple[1]}"
+
+    _STORAGE_SET_CLUSTER_MAP: Dict[str, Mapping[str, Any]] = {}
+
+    _SLICED_STORAGE_SET_CLUSTER_MAP: Dict[Tuple[str, int], Mapping[str, Any]] = {}
+    for cluster in locals["CLUSTERS"]:
+        for storage_set in cluster["storage_sets"]:
+            _STORAGE_SET_CLUSTER_MAP[storage_set] = cluster
```
onewland, line 126: this is clear so it's fine but i think it could be done as a list comprehension and then you wouldn't need a mutable variable type:

```python
_STORAGE_SET_CLUSTER_MAP = { 
  storage_set: cluster 
  for storage_set in cluster["storage_sets"] 
  for cluster in locals["CLUSTERS"] 
}
```

onewland (approved):
looks good

snuba/settings/validation.py
```diff
@@ -117,3 +117,49 @@ def validate_slicing_settings(locals: Mapping[str, Any]) -> None:
...
             topic_tuple in locals["SLICED_KAFKA_BROKER_CONFIG"]
         ), f"missing broker config definition for sliced Kafka topic {topic_tuple[0]} on slice {topic_tuple[1]}"
+
+    _STORAGE_SET_CLUSTER_MAP = {
+        storage_set: cluster
+        for cluster in locals["CLUSTERS"]
+        for storage_set in cluster["storage_sets"]
+    }
+
+    _SLICED_STORAGE_SET_CLUSTER_MAP: Dict[Tuple[str, int], Mapping[str, Any]] = {}
```
fpacifici, line 127: nit

```
Dict[Tuple[str, int], Mapping[str, Any]]
```

Is a quite hard type to read and wrap one's mind around. Adding aliases to give some context would be helpful. Also please `MutableMapping` instead of Dict as it is more generic.

ayirr7: Ok, modified!

PR #3492, opened 2022-12-09
ref(admin): minor migrations audit log fixes

MeredithAnya:
Found some minor things while I was testing this out for a demo:

* we should be returning early if we get an exception from Slack, otherwise we hit a `KeyError`
* ~we should be run/reversing migrations before we do the audit log, because otherwise we will create an audit log record even if we failed (and the audit log doesn't show that something failed)~ addressing in https://github.com/getsentry/snuba/pull/3493
* personal preference but I just updated the emojis to be 👟 and 🔙 , the :uno-reverse: is a custom slack emoji so just wanted something more generic and I got annoyed that the running person 🏃 looked like they were running backwards and not forwards lol

### DIFF
--- a/snuba/admin/notifications/slack/client.py
+++ b/snuba/admin/notifications/slack/client.py
@@ -37,6 +37,7 @@ class SlackClient(object):
             )
         except Exception as exc:
             logger.error(exc, exc_info=True)
+            return
 
             # Slack error handling
             # Copied from https://github.com/getsentry/sentry/blob/601f829c9246ae73c8169510140fd7f47fc6dfc3/src/sentry/integrations/slack/client.py#L36-L53
--- a/snuba/admin/notifications/slack/utils.py
+++ b/snuba/admin/notifications/slack/utils.py
@@ -45,9 +45,9 @@ def build_runtime_config_text(data: Any, action: AuditLogAction) -> Optional[str
 
 
 def build_migration_run_text(data: Any, action: AuditLogAction) -> Optional[str]:
-    if action == AuditLogAction.RAN_MIGRATION:
+    if action == AuditLogAction.RAN_MIGRATION_COMPLETED:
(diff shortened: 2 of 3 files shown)


MeredithAnya:
@dbanda did you purposefully add the audit log to be before the running/reversing so that we would create a record even if we failed? I think if that was your intention, maybe we could figure out how we want to denote that in the records

dbanda:
> @dbanda did you purposefully add the audit log to be before the running/reversing so that we would create a record even if we failed? I think if that was your intention, maybe we could figure out how we want to denote that in the records

Yes. My view is that it's better to log a false positive than to have the migration run and miss logging if something happens just before the log.  Worst case we get a notification that something was ran then discover that it failed. We can do a before and after annotation, but that seemed a little overkill for me.

MeredithAnya:
> > @dbanda did you purposefully add the audit log to be before the running/reversing so that we would create a record even if we failed? I think if that was your intention, maybe we could figure out how we want to denote that in the records
> 
> Yes. My view is that it's better to log a false positive than to have the migration run and miss logging if something happens just before the log. Worst case we get a notification that something was ran then discover that it failed. We can do a before and after annotation, but that seemed a little overkill for me.



> > @dbanda did you purposefully add the audit log to be before the running/reversing so that we would create a record even if we failed? I think if that was your intention, maybe we could figure out how we want to denote that in the records
> 
> Yes. My view is that it's better to log a false positive than to have the migration run and miss logging if something happens just before the log. Worst case we get a notification that something was ran then discover that it failed. We can do a before and after annotation, but that seemed a little overkill for me.

So I added a before and after: https://github.com/getsentry/snuba/pull/3493 I actually think this may not be overkill, it would make it easy to search to logs for failed runs, and it would also let us see how long it took for a migration to run which might be interesting to know.

dbanda (approved):

PR #3496, opened 2022-12-09
ref(replacements): Refactor Schema out of ReplacerProcessor init

rahul-kumar-saini:
### Overview
- ReplacerProcessor currently requires a schema as an init param
- This makes the config representation of the errors storage very messy since it will first define the schema as the storage param and then once again for the processor

### Refactor
- Moved schema initialization into a new method which basically does what was done on init before, but runs later (once all storages are built)
- The parent storage key is passed to the processor instead of the schema
- This method would be called on the first process_message() or get_schema() call of the processor to initialize the schema

### DIFF
--- a/snuba/datasets/errors_replacer.py
+++ b/snuba/datasets/errors_replacer.py
@@ -508,6 +508,11 @@ class ErrorsReplacer(ReplacerProcessor[Replacement]):
         self.__storage_key_str = storage_key_str
 
     def __initialize_schema(self) -> None:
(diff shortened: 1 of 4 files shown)


evanh (commented):
I think it might be better to deal with this similar to how the `NullColumnCaster` was changed: instead of a schema as an argument, the name of the storage/entity gets passed in. Then when `process_message` is called for the first time, it automatically calls `initialize_schema`, fetches the component and builds things.

evanh (approved):

PR #3497, opened 2022-12-09
ref(abuse) Add the parent_api back to the metrics

evanh:
The parent_api metric can be useful to see which feature has spiked in abuse,
since one feature can call many referrers and make it harder to see changes in
traffic usage.

This metric was removed in the past because we found cases of the endpoint not
being correct. Referrers were attached to endpoints that didn't actually call
the referrer. However the Sentry SDK has improved quite a bit since then and
even if the endpoint is not correct, it can still be useful signal of a problem.

### DIFF
--- a/snuba/query/query_settings.py
+++ b/snuba/query/query_settings.py
@@ -29,6 +29,14 @@ class QuerySettings(ABC):
         pass
 
     @abstractmethod
+    def get_parent_api(self) -> str:
+        """
+        This is an identifier used to denote the higher level request. That could be an API endpoint,
+        a celery task or something else entirely.
+        """
+        pass
+
+    @abstractmethod
     def get_dry_run(self) -> bool:
         pass
 
@@ -71,6 +79,7 @@ class HTTPQuerySettings(QuerySettings):
         # TODO: is this flag still relevant?
         legacy: bool = False,
         referrer: str = "unknown",
+        parent_api: str = "<unknown>",
     ) -> None:
         super().__init__()
         self.__turbo = turbo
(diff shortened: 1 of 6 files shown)


tests/subscriptions/entity_subscriptions/test_entity_subscriptions_data.py
```diff
@@ -52,7 +52,7 @@ def test_entity_subscriptions_data() -> None:
         RedisSubscriptionDataStore(redis_client, entity_key, PartitionId(i))
         for i in range(topic_spec.partitions_number)
     ]
-    print(stores[0].all())
```
evanh, line 55: This is unrelated, I just happened to notice this extra print.

snuba/query/query_settings.py
```diff
@@ -91,6 +101,9 @@ def get_consistent(self) -> bool:
     def get_debug(self) -> bool:
         return self.__debug
 
+    def get_parent_api(self) -> str:
```
volokluev, line 104: why does this need to be in the query settings?

volokluev (commented):
So once this merges, will it be automatically populated?The sentry side has been sending it all this time and we have just been ignoring it?

snuba/query/query_settings.py
evanh: I needed it in order to get some tests passing, but I think I was in a "Friday afternoon" haze and I did something wrong. I'm changing it.

snuba/request/schema.py
```diff
@@ -175,7 +175,7 @@ def generate_template(self) -> Any:
             # TODO: move this to attribution
             "legacy": {"type": "boolean", "default": False},
             "referrer": {"type": "string", "default": "<unknown>"},
-            "parent_api": {"type": "string", "default": "<unknown>"},
+            # "parent_api": {"type": "string", "default": "<unknown>"},
```
volokluev, line 178: should this be commented out?

evanh:
> So once this merges, will it be automatically populated?The sentry side has been sending it all this time and we have just been ignoring it?

No, there is an SDK change that requires a new version, which requires a pypi change, and then Sentry needs to be updated.

lynnagara (approved):

PR #3499, opened 2022-12-11
fix coverage paths

dbanda:
Some of the tests were overwriting each others coverage files. This fix renames the coverage paths to give each tests its own coverage file.

### DIFF
--- a/docker-compose.gcb.yml
+++ b/docker-compose.gcb.yml
@@ -19,9 +19,9 @@ services:
       - '--cov'
       - '.'
       - '--cov-report'
-      - 'xml:/.artifacts/coverage.xml'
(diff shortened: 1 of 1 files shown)


lynnagara (approved):
Thanks!

lynnagara:
Is this why we have reports like the one posted above in this PR? (project coverage changed by -69.46%)

dbanda:
> Is this why we have reports like the one posted above in this PR? (project coverage changed by -69.46%)

That's one of the reasons. But the bigger issue is that sometimes codecov uploads fail. They are pretty unreliable. But by default codecov will not return a fail code if that happens. you have to specify that behavior with the `-Z` flag.

PR #3502, opened 2022-12-12
bug(DC): Add config built storage keys to all_storages dictionary in factory

enochtangg:
This PR fixes a consumer [CLI](https://github.com/getsentry/snuba/blob/ff935eeb47b56fe54b0fda294b9055ab0acfcf48/snuba/cli/consumer.py#L44-L52) bug which was not returning the writable storages defined in config. When defining a new dataset (storage, processors, etc.), a user would eventually try to run their consumer locally through `snuba consumer`. However, since config define storages are not part of `_StorageFactory._writable_storages`, they would not be able to select their storage. The following code adds combines config built storages with `all_storages`.


snuba/datasets/storages/factory.py
```diff
@@ -130,6 +137,7 @@ def __initialize(self) -> None:
                 ]
             },
             **(self._dev_writable_storages if settings.ENABLE_DEV_FEATURES else {}),
+            **self._config_built_writable_storages,
```
volokluev, line 140: instead of creating this big dictionary, why not just remove it and have `get_writable_storage_keys` check if the storage is writeable?

rahul-kumar-saini: You'd have to modify `get_writable_storage_keys` to now build a list from a subset of `_all_storages`, and it seems `_all_storages` isn't really _all_ storages since it doesn't contain the config built storages. So one way or another, you have to add these storages to a dict.

Maybe the readable storages should be added as well and then the `get_writable_storage_keys` function could be modified to filter from `_all_storages`

volokluev: > Maybe the readable storages should be added as well and then the get_writable_storage_keys function could be modified to filter from _all_storages

Yeah that's what I mean, we should load all the storages into one place and these convenience functions should just be filters

enochtangg: Not sure if I understand the suggestion correctly, but ideally, all processing is done only once during initialization as opposed to every time the method is called.

volokluev: >  but ideally, all processing is done only once during initialization as opposed to every time the method is called.

Why is that ideal?

rahul-kumar-saini: If by processing you mean having to filter the dictionary every time the getter is called, we are already doing a bunch of processing to convert the dict keys to a list before returning. You could prebuild a big all_storages dict and then build small lists of storage keys on init and then just return them in the getter?

enochtangg: I think if there we were filtering on numerous parameters, it would be nonsensical to maintain numerous dictionaries which are subsets one set. However, if what we really need is two sets (assuming all storages have been migrated to config), one for all storages and one for only writable storages, then I think two dictionaries is reasonable to maintain. It doesn't cost a lot of memory, but saves us from filtering a dictionary every time the factory method is called.

volokluev: The only reason these `get_<something>_storage_keys` function exist is to provide command line options. They're only called once (if at all). They iterate through a dictionary that's at most 30 elements long, it's a few microseconds of impact.

Meanwhile, look at how many dictionaries are in this init method:

https://github.com/getsentry/snuba/blob/251cad3ab8b1ea648e6599da6eb443859f13ac93/snuba/datasets/storages/factory.py#L24-L34

Those are all extra pieces of state that live on the factory class (and thus globally). Does that state change? Are they serving a useful purpose?

enochtangg: Sure, I've added the readable storages to `self._all_storages` and update the methods to filter from that main dictionary instead. Let me know your thoughts on the approach. 

FWIW, I'm not familiar with the context behind `_dev_writable_storages`, `_dev_cdc_storages`, `_non_writable_storages`, and `_dev_non_writable_storages`, but it seems like they are all empty and can be removed which would make that list much less cluttered.

volokluev: Approach is good, let's clean up the __init__ and remove the unnecessary state. Otherwise great job!

snuba/datasets/storages/factory.py
```diff
@@ -23,7 +23,6 @@ class _StorageFactory(ConfigComponentFactory[Storage, StorageKey]):
     def __init__(self) -> None:
         with sentry_sdk.start_span(op="initialize", description="Storage Factory"):
             self._config_built_storages: dict[StorageKey, Storage] = {}
-            self._writable_storages: dict[StorageKey, Storage] = {}
```
volokluev, line 26: do you still need these?

```
 self._config_built_storages: dict[StorageKey, Storage] = {}
            self._dev_writable_storages: dict[StorageKey, Storage] = {}
            self._cdc_storages: dict[StorageKey, Storage] = {}
            self._dev_cdc_storages: dict[StorageKey, Storage] = {}
            self._non_writable_storages: dict[StorageKey, Storage] = {}
            self._dev_non_writable_storages: dict[StorageKey, Storage] = {}
```

enochtangg: Yes, to `_config_built_storages` because errors storage has not been move to config yet. I don't know have context behind what all the other ones do but they are empty so I think they can be removed? What do you think?

volokluev: Yeah I think they can go, they're not used anywhere

snuba/datasets/storages/factory.py
```diff
@@ -168,10 +151,18 @@ def get(self, storage_key: StorageKey) -> Storage:
...
+            for storage_key, storage in self._all_storages.items()
+            if type(storage) is WritableTableStorage
+        ]
 
     def get_cdc_storage_keys(self) -> list[StorageKey]:
-        return list(self._cdc_storages.keys())
+        return [
+            storage_key
+            for storage_key, storage in self._all_storages.items()
+            if type(storage) is CdcStorage
```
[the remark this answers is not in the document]
enochtangg: Since our `WritableTableStorage` extends from `ReadableTableStorage`, if we filtered for readable storages using `isinstance(), it would be incorrect because all writable storages are also instances of readable storages. Doing `type(object) is class` does an exact match. 

`isinstance()` still works for `CdcStorage` and `WritableTableStorage` so I can update that.

snuba/datasets/storages/factory.py
```diff
@@ -168,10 +151,18 @@ def get(self, storage_key: StorageKey) -> Storage:
         return self._all_storages[storage_key]
 
     def get_writable_storage_keys(self) -> list[StorageKey]:
-        return list(self._writable_storages.keys())
+        return [
```
volokluev, line 154: do you have a test to see if you're fixing the bug this PR was supposed to fix?

```
This PR fixes a consumer [CLI](https://github.com/getsentry/snuba/blob/ff935eeb47b56fe54b0fda294b9055ab0acfcf48/snuba/cli/consumer.py#L44-L52) bug which was not returning the writable storages defined in config.
```

enochtangg: Added file and a test for ensuring config-built writable storage key exists in get_writable_storage_keys()

snuba/datasets/storages/factory.py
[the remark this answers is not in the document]
lynnagara: Ok. For non writeable, i would probably still use `not isinstance(WritableTableStorage)` rather than type. That way in case we end up subclassing these anywhere, it still works.

snuba/datasets/storages/factory.py
volokluev: `self._cdc_storages` doesn't need to exist outside of the `__initialize` function scope.

volokluev (approved):
There's one more instance variable which can be removed. But other than that look good

PR #3503, opened 2022-12-12
fix(admin): bind log to configure it before using

dbanda:
The logger in `AuditLog` wasn't configured properly and was lazy so it was ignored in k8s. Adding bind to force it to configure properly and add the processors.

### DIFF
--- a/snuba/admin/audit_log/base.py
+++ b/snuba/admin/audit_log/base.py
@@ -10,6 +10,8 @@ from snuba.admin.notifications.slack.utils import build_blocks
 
 class AuditLog:
     def __init__(self) -> None:
+        # because we don't previously configure structlog, we need to bind the logger otherwise
+        # we will have a lazy bound logger without processors
         self.logger = structlog.get_logger().bind(
             module=self.__class__.__module__, context_class=self.__class__.__qualname__
         )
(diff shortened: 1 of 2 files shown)


snuba/admin/audit_log/base.py
```diff
@@ -10,7 +10,9 @@
 
 class AuditLog:
     def __init__(self) -> None:
-        self.logger = structlog.get_logger()
+        self.logger = structlog.get_logger().bind(
```
volokluev, line 13: would be good to explain why the bind is important in a comment

volokluev (approved):

snuba/admin/audit_log/base.py
```diff
@@ -10,7 +10,9 @@
...
+            module=self.__class__.__module__, context_class=self.__class__.__qualname__
```
MeredithAnya, line 14: what does the `context_class` give us? is the module not enough?

MeredithAnya:
Also, from reading the docs, it looks like we can just call `bind` without any parameters ( in the "important" note in https://www.structlog.org/en/stable/configuration.html#configuration). Is the problem that we didn't call `bind`? or that we didn't call `bind` with the parameters you added?

snuba/admin/audit_log/base.py
dbanda: module is enough. anything to trigger the binding would work

snuba/admin/audit_log/base.py
dbanda: added

PR #3509, opened 2022-12-13
ref(DC): Consolidate Query Plan Builders

enochtangg:
### Overview
Currently, there are classes which inherit from `ClickhouseQueryPlanBuilder`. They all accomplish the same goal, but with slight modifications to support storage selection and cluster selection. In order to support the current initiative to move all datasets to config files, these query plan builders need to be consolidated into one class.

### What has changed (Before and After)?
Most file changes in this PR are from the entity files. The main changes reside in the following files:
* `snuba/datasets/plans/storage_plan_builder.py`: contains the consolidated `StorageQueryPlanBuilder` class which supports single storage, selected storage, and cluster selection. This class takes a list of `StorageAndMappers` and uses a selector (if provided) choose the storage and respective translation mappers to use. 
* Renamed `sliced_storage.py` to `snuba/datasets/plans/cluster_selecetor.py` which contains `StorageClusterSelector`

### What is not included in this PR
The following changes will be added into a subsequent PR to avoid complexity:
* Updating entity definition from `self.storages: Sequence[Storages]` -> `self.storages: Sequence[StorageAndMappers]`
* Adding/Supporting custom storage selectors in config.

### Blast Radius
* All entities were updated to use consolidated query plan builder.

### Testing
* All existing `ClickhouseQueryPlanBuilder` tests were updated to use consolidated class and are passing.


snuba/datasets/storage.py
```diff
@@ -176,9 +176,22 @@ def get_is_write_error_ignorable(self) -> bool:
...
 
 
+class StorageAndMappersNotFound(Exception):
+    pass
+
+
 class StorageAndMappers(NamedTuple):
     storage: ReadableStorage
     mappers: TranslationMappers
+    is_writable: bool
```
evanh, line 186: I think it would be better if `is_writable` wasn't passed in here. This could be calculated based on the storage being passed in. I'm not sure what NamedTuple supports, but some options

- Use the `@property` decorator on a function
- Add a function `is_writable()` that checks if the storage is writable

snuba/datasets/plans/storage_builder.py
```diff
@@ -92,112 +98,79 @@ def get_query_data_source(
...
         self, query: LogicalQuery, settings: QuerySettings
     ) -> Sequence[ClickhouseQueryPlan]:
-        with sentry_sdk.start_span(
-            op="build_plan.selected_storage", description="select_storage"
-        ):
-            storage, mappers = self.__selector.select_storage(query, settings)
+        if len(self.__storages) < 1:
+            raise QueryStorageSelectorError("No storages specified to select from.")
+
+        if not self.__selector:
```
evanh, line 126: This should only be used by entities that are defined in YAML correct? So this check would be removed once we support storage selection in config?

snuba/datasets/storage.py
enochtangg: I think that's a better approach too, but there's actually an underlying issue that is blocking the determination of a storage with only the object. The problem is that there a difference between actual storages which inherit from `ReadTableStorage` or `WriteTableStorage` and storages which config defines as "readable" or "writable". For example, see the metrics [sets and counters storages](https://github.com/getsentry/snuba/blob/53d6f3ec05c3e7d306c7697bf6b046c41a4df236/snuba/datasets/storages/metrics.py#L114-L158) which are both `WritableTableStorage`. However, if you look at the [entity](https://github.com/getsentry/snuba/blob/53d6f3ec05c3e7d306c7697bf6b046c41a4df236/snuba/datasets/entities/metrics.py#L127-L164) you'll see that they are "readable" storages on the entity. This is problematic because we use config defined definition to default the storage selector.

snuba/datasets/plans/storage_builder.py
enochtangg: Right, although jsonschema syntax validation should complain before this gets raised.

snuba/datasets/storage.py
volokluev: So if I can rephrase what you're saying, `StorageAndMappers` outline the relationship of an entity to a storage, and just because the storage is `Writeable` doesn't mean the entity sees it that way. Is that right?

snuba/datasets/plans/storage_builder.py
```diff
@@ -92,112 +98,78 @@ def get_query_data_source(
...
-                ),
-            )
-        ]
-
-
-class SelectedStorageQueryPlanBuilder(ClickhouseQueryPlanBuilder):
+class StorageQueryPlanBuilder(ClickhouseQueryPlanBuilder):
     """
-    A query plan builder that selects one of multiple storages in the dataset.
+    Builds the Clickhouse Query Execution Plan for a dataset which supports single storages,
```
volokluev, line 103: are there tests for all these branches of functionality? I understand that the existing tests pass but those depend on existing entities and storages right? Have we tested this component just in isolation?

snuba/datasets/storage.py
enochtangg: > just because the storage is Writeable doesn't mean the entity sees it that way
👍 

> StorageAndMappers outline the relationship of an entity to a storage
Sort of. StorageAndMappers outline the relationship of a storage and translation mappers. And we eventually would like an entity to specify a `List[StorageAndMappers]` to therefore outline the relationship between entity and storage.

snuba/datasets/plans/storage_builder.py
```diff
@@ -92,112 +98,78 @@ def get_query_data_source(
...
         self, query: LogicalQuery, settings: QuerySettings
     ) -> Sequence[ClickhouseQueryPlan]:
-        with sentry_sdk.start_span(
-            op="build_plan.selected_storage", description="select_storage"
-        ):
-            storage, mappers = self.__selector.select_storage(query, settings)
+        if len(self.__storages) < 1:
+            raise QueryStorageSelectorError("No storages specified to select from.")
+
+        if not self.__selector:
```
volokluev, line 126: This function should be split up into:

1. Selecting the StorageAndMappers
2. Selecting the cluster 

Currently it's easy to get lost in what's doing what and it's creating a lot of state in the scope of the function which is easy to misuse

snuba/datasets/pluggable_entity.py
```diff
@@ -73,25 +73,20 @@ def get_all_join_relationships(self) -> Mapping[str, JoinRelationship]:
...
-                        storage_set=self.readable_storage.get_storage_set_key(),
-                        partition_key_column_name=self.partition_key_column_name,
-                    ),
+        storages: List[StorageAndMappers] = [
+            StorageAndMappers(self.readable_storage, self.translation_mappers, False)
+        ]
+        if self.writeable_storage:
+            storages.append(
+                StorageAndMappers(
+                    self.writeable_storage, self.translation_mappers, True
```
volokluev, line 82: So are we making an assumption here that the translation mappers are the same no matter the storage?

volokluev: Is that an accurate assumption to make?

snuba/datasets/plans/storage_builder.py
enochtangg: There are individual tests for storage selection and cluster selection. Both of which have been updated to use the consolidated plan. However, there is no test for just the component in isolation. I can make one for that!

snuba/datasets/pluggable_entity.py
enochtangg: Correct, this is the assumption at the moment which what we have today. This PR description says that updating the pluggable entity definition to support multiple translation mappers via `List[StorageAndMappers]` will be included in a subsequent PR because those changes will be too large for this one.

snuba/datasets/plans/storage_builder.py
```diff
@@ -92,112 +98,78 @@ def get_query_data_source(
...
+            # Default to the first and only storage and mapper
+            if len(self.__storages) == 1:
+                storage, mappers, _ = self.__storages[0]
+            else:
+                # If there are both readable and writable storages, select the readable one.
+                # Multiple writable storages are not supported.
+                readable_storages = [
+                    storage for storage in self.__storages if not storage.is_writable
```
evanh, line 134: Related to https://github.com/getsentry/snuba/pull/3509#discussion_r1047529358

I think the problem is that this function is trying to be too generic and handle too many cases. There are currently three use cases for storage selection:
1) Single storage: The entity provides a single storage and mappers, and those are always used.
2) Selected storage: The entity provides a class which selects a storage/mappers. <- the problem case
3) Sliced storage: The entity provides a single storage and mappers, and some extra fields that allow choosing a cluster. Note that this is the same as the Single Storage, but changes how the cluster is fetched.

This logic only needs to allow and handle those three cases explicitly. So `StorageQueryPlanBuilder` would be initialized the same way you have now.

Pick a storage:
- Does the builder specify a selector? Pass the list of storages to the selector and get back a single `StorageAndMappers`. <- the major change.
- Otherwise, use the provided storage/mapper.

Fetch the cluster:
- Does the builder have sliced storage fields? Use those to fetch the cluster.
- Otherwise, get the cluster from the storage selected above.

The major change here is to have the storage selectors themselves deal with choosing the storage.

lynnagara:
Can you comment on why this change is needed specifically?

snuba/datasets/plans/storage_builder.py
enochtangg: The main function can definitely be separated to draw a clearer distinction as Volo mentioned. This PR attempts to cover those three main cases explicitly, but also adds some functionality in preparation for defining `StorageAndMappers` in the entity https://github.com/getsentry/snuba/pull/3509#discussion_r1047804306.

For that functionality to work, we wouldn't be able to "Otherwise, use the provided storage/mapper"  which is why the logic for defaulting to a readable storage is there.

enochtangg:
> Can you comment on why this change is needed specifically? I'm not sure I understand what it solves

This is one of the sub-task for getting storage selection working in config. Right now `SelectedStorageQueryPlanBuilder` (or any `StorageQueryPlanBuilder`) is not reference-able from configuration. As a result, we default to single storage. However for entities like `Sessions`, it has multiple storages, translation mappers, and one selector. To make this all work, we'd like a single point of entry for the query plan builder which can support all 3 cases list above. The proposal/architecture meeting is described [here](https://www.notion.so/sentry/Selecting-a-storage-in-dataset-config-208b7921094b472a954be881f659278b).

snuba/datasets/plans/storage_builder.py
enochtangg: Updated!

evanh (approved):
Couple small things, but mostly looking really good. Nicely done!

.vscode/settings.json
```diff
@@ -61,5 +61,8 @@
   // tests is often in your way and misclicked
   "python.testing.pytestEnabled": true,
   "python.testing.unittestEnabled": true,
-  "python.testing.pytestArgs": ["tests"]
+  "python.testing.pytestArgs": [
```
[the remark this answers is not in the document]
evanh: This file shouldn't be changing at all.

PR #3511, opened 2022-12-13
feat: KafkaConsumerWithCommitLog as a strategy (take 2)

lynnagara:
This is another attempt at replacing the KafkaConsumerWithCommitLog with a processing strategy instead of custom consumer code.

The original implementation (https://github.com/getsentry/snuba/pull/3506) had a bug because the CollectStep never called next_step.poll() where the producing was done.

A fix for this underlying issue in Arroyo is in https://github.com/getsentry/arroyo/pull/163.

This PR avoids this issue by producing in the submit method now.

The other change made in this version of the code is to completely remove produce throttling. Since this strategy is only ever called after a CollectStep which operates once per batch it is unnecessary to introduce an additional level of throttling here.


lynnagara:
On second thoughts, going to rethink this a bit

PR #3513, opened 2022-12-13
migrations UI fixes

dbanda:
UI fixes to the migrations tool

1. Makes it clearer what the dry run buttons do
2. Only presents the EXECUTE buttons run/reverse immediately after user has just run the dry run. Reduces risk of user mistakenly clicking forwards run when they are looking at a dry run from reverse.
3. Show a success/fail alert.

https://user-images.githubusercontent.com/1960911/207752692-28f0cfc0-49c8-480f-89aa-61f6d5c0dc35.mov

### DIFF
--- a/snuba/admin/static/clickhouse_migrations/index.tsx
+++ b/snuba/admin/static/clickhouse_migrations/index.tsx
@@ -100,7 +100,7 @@ function ClickhouseMigrations(props: { api: Client }) {
     executeRealRun(action, force);
   }
 
-  function executeRun(action: Action, dry_run: boolean, force: boolean, cb: (stdout: string) => void ) {
+  function executeRun(action: Action, dry_run: boolean, force: boolean, cb?: (stdout: string) => void ) {
     let req = {
       action: action,
       migration_id: migrationId,
@@ -132,7 +132,7 @@ function ClickhouseMigrations(props: { api: Client }) {
   function executeDryRun(action: Action) {
     console.log("executing dry run !", migrationId, action);
     setHeader(()=> dry_run_header)
-    executeRun(action, true, false, null)
+    executeRun(action, true, false)
     setShowAction(()=> true)
 
   }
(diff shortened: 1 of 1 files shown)


snuba/admin/static/clickhouse_migrations/index.tsx
```diff
@@ -17,6 +17,15 @@ function ClickhouseMigrations(props: { api: Client }) {
   const [migrationId, setMigrationId] = useState<string | null>(null);
   const [SQLText, setSQLText] = useState<string | null>(null);
 
+  const [header, setHeader] = useState<string | null>(null);
+  const [show_action, setShowAction] = useState<boolean | null>(false);
+
+  const dry_run_header = "Raw SQL for running a migration (forwards) or reversing \n (backwards). Good to do before executing a migration for real."
```
MeredithAnya, line 23: I think this text should always be present whenever the buttons are present, so I would move this text to display above the buttons. You could change it a little to be like:
`Before executing a migration, do a dry run first. This will generate the raw SQL for running a migration (forwards) or reversing (backwards) a migration so that you can verify it's contents.`

Something that basically tells them why they are doing this first.

But I do think it makes sense to have a header over the output textbox to indicate whether it's the raw sql from the dry run or if its the output from actually executing. so maybe you could just have that header be `Dry run SQL output`

snuba/admin/static/clickhouse_migrations/index.tsx
```diff
@@ -17,6 +17,15 @@ function ClickhouseMigrations(props: { api: Client }) {
...
+  const real_run_header = "Run log output"
+
+  const [forwards_dry_run, setFowardsDryRun] = useState<string | null>(null);
+  const [backwards_dry_run, setBackwardsDryRun] = useState<string | null>(null);
```
MeredithAnya, line 27: also might be nice to show the buttons as selected since you have this state now.

snuba/admin/static/clickhouse_migrations/index.tsx
dbanda: makes sense. updated

snuba/admin/static/clickhouse_migrations/index.tsx
dbanda: I fixed it to give the buttons a red text when selected.

MeredithAnya (approved):

PR #3514, opened 2022-12-13
fix(admin): build bundle outside of arm64 image 

dbanda:
Building the admin UI within the Arm64 docker image is very slow because GH use emulation. Changes to snuba/admin would invalidate the cache and trigger this slow building. This speeds up the CI by building the bundle.js outside of the docker image and copying the file into the container.


dbanda:
> Do you have to remove the `yarn` commands from the Dockerfile as well? Or are those just skipped if the `bundle.js` is already built?

They are skipped in this CI job because we use the target application argument.

evanh (approved):

PR #3516, opened 2022-12-14
fix(discover): DefaultNoneColumnMapper Group By 

rahul-kumar-saini:
### Overview
- If you select AND group by a column which will be set to `identity(NULL)`, we surface a ClickHouse error:
    - `DB::Exception: Size of permutation doesn't match size of column.`
- The query being sent to us isn’t technically wrong, so this shouldn’t cause a 500 error
- It is however, impossible to serve

### Changes
- Introduced a new QueryValidator which just applies the `DefaultNoneColumnMapper` to all columns in a group by condition to catch this case

### Notes
- This ONLY applies to Discover since that is the only entity which should ever use the mapper above

Closes [SNS-1942](https://getsentry.atlassian.net/browse/SNS-1942)

### DIFF
--- a/snuba/datasets/entities/discover.py
+++ b/snuba/datasets/entities/discover.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import Sequence
 
 from snuba.clickhouse.columns import (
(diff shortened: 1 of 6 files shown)


snuba/datasets/entities/discover.py
```diff
@@ -267,7 +276,10 @@ def __init__(self) -> None:
             ),
             join_relationships={},
             writable_storage=None,
-            validators=[EntityRequiredColumnValidator({"project_id"})],
+            validators=[
+                EntityRequiredColumnValidator({"project_id"}),
+                DefaultNoneColumnValidator(default_none_column_mappers(mappers)),
+            ],
```
volokluev, line 282: how is this going to translate to config?

rahul-kumar-saini:
Stale

PR #3517, opened 2022-12-14
feat(devserver): Refactor devserver to use CLI args instead of settings

lynnagara:
Our settings is getting very polluted with random settings for the devserver. None of this is relevant except for in the devserver anyway. This moves them to CLI args.

The following settings have been removed:
- ENABLE_SENTRY_METRICS_DEV
- ENABLE_METRICS_SUBSCRIPTIONS
- SEPARATE_SCHEDULER_EXECUTOR_SUBSCRIPTIONS
- ENABLE_PROFILES_CONSUMER
- ENABLE_REPLAYS_CONSUMER

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -1,12 +1,24 @@
 import click
 
-from snuba import settings
-
 
 @click.command()
 @click.option("--bootstrap/--no-bootstrap", default=True)
 @click.option("--workers/--no-workers", default=True)
-def devserver(*, bootstrap: bool, workers: bool) -> None:
+@click.option("--metrics/--no-metrics", default=False)
+@click.option("--profiles/--no-profiles", default=False)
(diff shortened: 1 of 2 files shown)


evanh (commented):
Is the devserver used in our or Sentry's CI anywhere?

lynnagara:
> Is the devserver used in our or Sentry's CI anywhere?

Oh it is used by sentry's devserver. Looks like we'll actually have to do this in 2 steps then - first introduce the new args but still support the old ones, then deprecate them as a follow up

untitaker:
i'm all for this change

> Oh it is used by sentry's devserver. Looks like we'll actually have to do this in 2 steps then - first introduce the new args but still support the old ones, then deprecate them as a follow up

I am not sure you need to be this fancy. I feel like it should not be a big deal to just send out a PSA that everybody has to run `devservices up`

PR #3521, opened 2022-12-14
feat: Delete offline replacer

lynnagara:
I don't think this has ever been used. Even in a disaster scenario, it seeems improbable to me that we would ever want to perform replacements from a file instead of a Kafka topic. Let's remove it.


onewland:
I think Filippo has in fact used this during an incident to "correct the record" after removing one org ID's replacements from the topic

lynnagara:
Ok. Let's wait for @fpacifici to confirm if we should keep this around

fpacifici:
> it seeems improbable to me that we would ever want to perform replacements from a file instead of a Kafka topic. Let's remove it.

Now I remember the context. The rationale and scenario to use this is that we would have a single project clogging the replacement topic. In such case we disable replacements for that project and drop all messages for such project. 
Later we recover those dropped messages and execute the script. We cannot execute the script from the kafka topic as it moved on.

We have some tickets to build a secondary queue to demote projects, but till we do not have that, we need this script to recover.

lynnagara:
Ok. I'll leave this alone until we implement the replacement then

PR #3522, opened 2022-12-15
feat(admin): implement basic admin user roles

MeredithAnya:
**context:**
The access to the snuba admin tool right now for SaaS is determined via [google IAP](https://cloud.google.com/iap) which is great for all or nothing access but doesn't really allow for more granular permissions. The reason that we want more granular permissions is so that we can open up parts admin tooling to more folks (outside the SNS team) without giving access to everything

**Roles**
I've added the concept of roles to the admin user. A role has two properties:
* `name` - the name of the role
* `actions` - the actions a user is allowed to take on specified resources, for migrations this would be the actions they could take on a migration(s) within a migration group

**Actions**
Actions correspond to specific resources, instead of generic actions that apply to all resources. Custom actions can be implemented for each `category` (e.g. "migrations" resource has the `MigrationsAction`) and those can be further split up in specific actions within a category

[EDIT]: Instead of scopes, the actions within a role map to a policy. e.g. `ExecuteNonBlockingAction` would correspond to the `NonBlockingMigrationsPolicy`.
~Scopes will be used to map to migration polices - this PR https://github.com/getsentry/snuba/pull/3543/files builds off of this one to do just that.~

This PR adds the roles to the user but doesn't do any checking of the roles. This will come later when implementing this for migrations.


snuba/admin/auth_scopes.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Set
+
+
+class AdminRole(Enum):
+    ADMIN = "admin"
+    MEMBER = "member"
+    MEMBER_READ = "member_read"
+
+
+ADMIN_CATEGORY_RESOURCES = {
```
evanh, line 12: Would it be possible/worth it to generate this list automatically from the migration groups in code?

lynnagara:
Ltgm though I think @dbanda may want to take a look as I saw some offline conversations about this.

Where will the user -> group -> role -> category -> resource mappings be kept? Can you give a rough idea of how you imagine this configuration would look/be managed?

snuba/admin/auth_scopes.py
fpacifici: +1, we should not have to add a new scope in the admin tool code each time we add a new migration group. That is something product teams do.

snuba/admin/auth_scopes.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@dataclass(frozen=True)
+class AuthScope:
+    category: str
+    resource: str
+    role: AdminRole
+
+    def to_str(self) -> str:
+        return f"{self.category}.{self.resource}.{self.role.value}"
```
fpacifici, line 32: I am a bit confused about this model. 
As I mentioned on the notion doc, most role based access system have at least three entities: 
- resources
- actions
- principals / roles 

There generally is a relationship of ownership between a role, and action and a resource so that a role is granted the permission to execute the action on a resource.
Then you assign roles to a person.

Here I think the resource is composed by the pair `category + resource`. Is this correct ?
I cannot see the action in this model. 
Also does the scope represent the relationship between the three ? 

Since these concepts in role based access control are generally familiar to developers I would suggest we stick to them without introducing new ones. That makes it much easier for somebody who does not know the snuba admin to pick them up without having to understand the custom model.

fpacifici (commented):
If somebody uses the migrations tool without going through the admin, should we have an equivalent permissions model ?

snuba/admin/auth_scopes.py
```diff
@@ -0,0 +1,43 @@
...
+
+def scopes() -> Set[AuthScope]:
+    scopes = set()
+    for item in ADMIN_CATEGORY_RESOURCES:
+        category, resource = item.split(".")
+        scopes.update([AuthScope(category, resource, role) for role in AdminRole])
+    return scopes
+
+
+ADMIN_SCOPES = scopes()
```
dbanda, line 43: This looks like it will generate all permutations of scopes for each resource and admin role. Is there a reason why this is done? It looks like the only place `ADMIN_SCOPES` is used is in `_set_scopes()` which will assign all these scopes to the user. Was that intentional?

dbanda (approved):

snuba/admin/auth_roles.py
```diff
@@ -0,0 +1,43 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Set
+
+from snuba import settings
+
+
+class Action(Enum):
+    EXECUTE = "execute"
+    LIMITED_EXECUTE = "limited_execute"
```
fpacifici, line 10: Have you considered calling the actions in the same way as the policy ?

fpacifici (commented):
I like this model more than the previous one. 
I would consider renaming the action so that they are more explicit about the policy they would adopt.

snuba/admin/auth_roles.py
```diff
@@ -0,0 +1,101 @@
...
+
+class ExecuteNonBlockingAction(MigrationAction):
+    pass
+
+
+class ExecuteNoneAction(MigrationAction):
+    pass
+
+
+# todo, shoudln't need .keys() once ADMIN_ALLOWED_MIGRATION_GROUPS is a set not dict
```
dbanda, line 75:
```suggestion
# todo, `shouldn't` need .keys() once ADMIN_ALLOWED_MIGRATION_GROUPS is a set not dict
```

dbanda (commented):
LGTM

PR #3525, opened 2022-12-15
feat(admin): Show dist nodes hosts

dbanda:
The dist nodes do not show up as hosts for the system queries tab. This usually isn't an issue locally because the dist hosts are the same a local hosts, but for distributed setups where the two are different, it means we cant run system queries on the dist hosts.

### Before:
The system queries tab would only show local hosts

### After:
The system queries tab shows dist hosts if they are not already included in the local hosts

### Blast Radius
Frontend JSX and `/clickhouse_nodes` admin endpoint.

### DIFF
--- a/snuba/admin/clickhouse/nodes.py
+++ b/snuba/admin/clickhouse/nodes.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Sequence, TypedDict
+from typing import Optional, Sequence, TypedDict
 
 from snuba import settings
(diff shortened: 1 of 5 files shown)


MeredithAnya:
So I think we can try this out, but Im not sure it will work given that we don't actually set the `distributed_cluster_name` in prod: https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L28

MeredithAnya (approved):
Again im interested to see if this works, if it doesn't it looks like we should be in the same spot we were before

PR #3526, opened 2022-12-15
feat(migration): conditionally auto-run migration based on env var

barkbarkimashark:
This PR is mostly needed as a stop-gap since we don't have the production infra setup yet for the table yet, but we still want to be able to reference the dataset in sentry CI and dev.

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -237,10 +237,11 @@ SKIPPED_MIGRATION_GROUPS: Set[str] = {
     "profiles",
     "functions",
     "test_migration",
+    "search_issues",
 }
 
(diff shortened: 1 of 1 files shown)


MeredithAnya:
So I thought at least for CI it would use the https://github.com/getsentry/snuba/blob/ef6260d7bbcdadc6c57aa54b00b08188d674cc94/snuba/settings/settings_test.py#L16 Does it not?

lynnagara (approved):

barkbarkimashark:
> So I thought at least for CI it would use the
> 
> https://github.com/getsentry/snuba/blob/ef6260d7bbcdadc6c57aa54b00b08188d674cc94/snuba/settings/settings_test.py#L16
> Does it not?

Does it? That would be great if it did. How could I confirm this?

PR #3527, opened 2022-12-15
fix(search_issues): parse out client_timestamp and map timestamp -> client_timestamp column

barkbarkimashark:
Populate `client_timestamp` column with the value from `event.datetime`. Map timestamp -> client_timestamp.

### DIFF
--- a/snuba/datasets/processors/search_issues_processor.py
+++ b/snuba/datasets/processors/search_issues_processor.py
@@ -53,7 +53,7 @@ class IssueEventData(TypedDict, total=False):
     request: Mapping[str, Any]  # http_method, http_referer
 
 
-class SearchIssueEvent(TypedDict):
+class SearchIssueEvent(TypedDict, total=False):
     # meta
     retention_days: int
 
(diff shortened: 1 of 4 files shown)


snuba/datasets/processors/search_issues_processor.py
```diff
@@ -110,6 +112,12 @@ def process_insert_v1(
         retention_days = enforce_retention(
             event.get("retention_days", 90), detection_timestamp
         )
+        client_timestamp = _ensure_valid_date(
+            datetime.strptime(event["datetime"], settings.PAYLOAD_DATETIME_FORMAT)
```
nikhars, line 116: Is it possible for the `datetime` field on the incoming event to be empty? If yes, then you should handle that case.

barkbarkimashark: Good point. I think we'll always have that present, but if its missing its probably better to fail the event processing instead of defaulting the value to received_timestamp. Re-worked the logic to reflect those changes.

wedamija (approved):

lynnagara (approved):

PR #3531, opened 2022-12-16
feat: Refactor the Kafka consumer with commit log (again)

lynnagara:
Same as https://github.com/getsentry/snuba/pull/3523 but with the `producer.poll(0.0)` call between each message submitted.

This was being called in the original KafkaConsumerWithCommitLog here (https://github.com/getsentry/snuba/blob/9b91174bd9a2d55761c510e934ceeabb096a528a/snuba/utils/streams/kafka_consumer_with_commit_log.py#L32) but was dropped in the previous version of this PR which converted this logic to the strategy.

Poll is required when using the Confluent producer directly in order to poll the producer for events and fire delivery callbacks.

We should merge https://github.com/getsentry/snuba/pull/3530 before this. On the off chance that another out of order message makes its way to one of the commit log topics, we don't want to cause another incident.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -312,7 +312,7 @@ class ProcessedMessageBatchWriter(
                     on_delivery=self.__commit_message_delivery_callback,
                 )
                 self.__commit_log_config.producer.poll(0.0)
-            self.__offsets_to_produce.clear()
+        self.__offsets_to_produce.clear()
 
     def terminate(self) -> None:
         self.__closed = True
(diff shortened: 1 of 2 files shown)


evanh (approved):
For my own curiousity, are there tests already in place for all these changes?

lynnagara:
This consumer is largely untested 😢. It's on our radar to add tests (started here https://github.com/getsentry/snuba/pull/3389) and ideally also deprecate it entirely in favour of using the multistorage implementation everywhere. But that is not done yet.

In the absence of those, I just ran the consumer end to end locally though to verify it works.

PR #3539, opened 2022-12-16
fix(init): don't return in the with block

volokluev:
I have a suspicion that the sentry sdk and click are interacting in weird ways. Try this to see if the transaction is labeled properly


lynnagara (approved):

untitaker:
@volokluev do you need help with this? It wouldn't make sense to me that the `return` causes the with-stmt to not properly close the transaction

PR #3540, opened 2022-12-16
fix(DC): Bring EntityContainsColumnsValidator back to config entities

volokluev:
Due to a bug in our testing, we have not been including the EntityContainsColumnsValidator in the entities built with config. Having a place where the built in validators are defined makes it easier to auto-convert entities as we can skip the default ones in the config. 

### Blast Radius

This will affect only config entities (metrics) and this is re-introducing a columns validator which is turned off. This is essentially a no-op.


tests/datasets/configuration/test_entity_loader.py
```diff
@@ -61,24 +61,30 @@ def _config_matches_python_definition(
         self, config_path: str, entity: Type[Entity], entity_key: EntityKey
     ) -> None:
         config_entity = build_entity_from_config(config_path)
+        print(config_entity.get_validators())
```
rahul-kumar-saini, line 64: ❗

snuba/datasets/pluggable_entity.py
```diff
@@ -58,6 +62,14 @@ class PluggableEntity(Entity):
     subscription_processors: Optional[Sequence[EntitySubscriptionProcessor]] = None
     subscription_validators: Optional[Sequence[EntitySubscriptionValidator]] = None
 
+    def _get_builtin_validators(self) -> Sequence[QueryValidator]:
+        return [
+            EntityContainsColumnsValidator(
+                EntityColumnSet(self.columns), ColumnValidationMode.DO_NOTHING
+            )
+        ]
+        return []
```
rahul-kumar-saini, line 71: ❓

rahul-kumar-saini (approved):

PR #3543, opened 2022-12-19
ref(migrations): use scopes with policies

MeredithAnya:


snuba/admin/migrations_policies.py
```diff
@@ -23,6 +30,29 @@ def get_migration_group_polices() -> Dict[str, MigrationPolicy]:
...
+    scopes: Set[AuthScope],
+) -> Dict[str, MigrationPolicy]:
+    migration_scopes = [scope for scope in scopes if scope.category == "migrations"]
+    group_policies = {}
+    for scope in migration_scopes:
+        if scope.resource == "all":
+            policy = MigrationPolicy.class_from_name(
+                ROLES_TO_POLICIES[scope.role.value]
+            )()
+            for group in settings.ADMIN_ALLOWED_MIGRATION_GROUPS.keys():
```
MeredithAnya, line 43: `settings.ADMIN_ALLOWED_MIGRATION_GROUPS` will just be a set of groups instead of a mapping to a policy but was just trying this out locally first.

PR #3544, opened 2022-12-19
feat(DC): CDC Storage support

rahul-kumar-saini:
### Overview
- Previously only Readable and Writable storages were supported
- This doesn't work for the CDC storages
- Introduced `cdc_storage` config type and refactored storage building to support the new type


snuba/datasets/configuration/storage_builder.py
```diff
@@ -53,27 +57,65 @@ def build_storage_from_config(
...
+        ),
+        QUERY_SPLITTERS: get_query_splitters(
+            config[QUERY_SPLITTERS] if QUERY_SPLITTERS in config else []
+        ),
+        MANDATORY_CONDITION_CHECKERS: get_mandatory_condition_checkers(
+            config[MANDATORY_CONDITION_CHECKERS]
+            if MANDATORY_CONDITION_CHECKERS in config
+            else []
+        )
+        # TODO: Rest of readable storage optional args
```
volokluev, line 89: what's left?

tests/datasets/configuration/test_storage_loader.py
```diff
@@ -112,6 +113,15 @@ def _deep_compare_storages(old: Storage, new: Storage) -> None:
             new.get_table_writer().get_stream_loader(),
         )
 
+    if isinstance(old, CdcStorage) or isinstance(new, CdcStorage):
```
volokluev, line 116: Is this code actually running on anything? We don't have any CDC storages in config to compare yet right?

volokluev (changes requested):
This needs tests that define a CDC storage (independent of the ones that exist right now) and makes sure it's parsed and populated correctly

tests/datasets/configuration/test_storage_loader.py
rahul-kumar-saini: This is for #3549, thought I'd add the support in this PR so the other PR is cleaner

snuba/datasets/configuration/storage_builder.py
rahul-kumar-saini: Nothing, good catch

evanh (approved):

PR #3545, opened 2022-12-19
feat(DC): Add Storage selector support for Entity configuration

enochtangg:
### Overview
This PR is responsible for adding capabilities for defining storage selectors in entity configuration. Now that [query plan builders have been consolidated](https://github.com/getsentry/snuba/pull/3509), we can now support custom or default storage selectors defined on an entity. For context, right now only Errors and Sessions have custom storage selectors based on a query. All other entities simply default to single storage regardless of the query.

### What has changed (before and after)?
* All pluggable entities now explicitly require a storage selector even if there is only one storage specified.
* `DefaultQueryStorageSelector` has been created and added to all entities (python defined ones too). This is used to make storage selection explicit.
* A new `snuba/datasets/storages/selectors/` directory was created to house all storage selectors. The custom sessions and errors selectors were also moved here.

### Blast Radius
This change is being applied to all entities. Entities must now specify a storage selector (default is provided).

### Testing
* Existing query plan builder tests were updated to use default storage selectors. 
* Entity tests are passing as expected with the selector. 
* A new isolated test for default storage selector was added.


snuba/datasets/storages/selectors/selector.py
```diff
@@ -0,0 +1,77 @@
...
+    ) -> StorageAndMappers:
+        raise NotImplementedError
+
+    def get_storage_mapping_pair(
+        self,
+        storage: ReadableStorage,
+        storage_and_mappers_list: Sequence[StorageAndMappers],
+    ) -> StorageAndMappers:
+        for sm_tuple in storage_and_mappers_list:
+            if storage == sm_tuple.storage:
```
evanh, line 49: This check feels a bit brittle. The problem is that we don't have a proper equality check for storages, so what you are actually checking here is that the instance of the storages is the same (Python compares the memory address of the variables and sees if they are the same).

In theory that should be fine, since we only instantiate each storage once, but it would be better to compare the storage key I think. Storages are aware of their own storage key. `storage.storage_key == ...`

snuba/datasets/storages/selectors/selector.py
```diff
@@ -0,0 +1,77 @@
...
+        query: Query,
+        query_settings: QuerySettings,
+        storage_and_mappers_list: List[StorageAndMappers],
+    ) -> StorageAndMappers:
+        raise NotImplementedError
+
+    def get_storage_mapping_pair(
+        self,
+        storage: ReadableStorage,
+        storage_and_mappers_list: Sequence[StorageAndMappers],
```
evanh, line 46:
```suggestion
        storage_and_mappers: Sequence[StorageAndMappers],
```

snuba/datasets/storages/selectors/selector.py
```diff
@@ -0,0 +1,77 @@
...
+
+class DefaultQueryStorageSelector(QueryStorageSelector):
+    """
+    A default query storage selector which chooses the only storage specified in config.
+    Entities which define multiple storages should not use this selector and should use
+    custom ones.
+    """
+
+    def __init__(self, storage: str) -> None:
+        self.storage = get_storage(StorageKey(storage))
```
evanh, line 64: I think you could probably just store the `StorageKey` here. Then in the `select_storage` function you could compare the storage key directly: `if sm.storage.storage_key == self.storage_key`.

tests/datasets/storages/selectors/test_storage_selector.py
```diff
@@ -0,0 +1,59 @@
...
+from snuba.datasets.storages.selectors.selector import (
+    DefaultQueryStorageSelector,
+    QueryStorageSelector,
+)
+from snuba.datasets.storages.storage_key import StorageKey
+from snuba.query.logical import Query
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.query.snql.parser import parse_snql_query
+
+TEST_CASES = [
```
evanh, line 19: I would add a couple test cases for the errors and sessions ones as well.

volokluev:
I have some questions about the design choices made here:


> All pluggable entities now explicitly require a storage selector even if there is only one storage specified.

Why not have the default be implicit and have the entity fail to instantiate if the deafult is not sufficient? You can accomplish this with the `__post_init__` method of the dataclass. Seems like one fewer config option for users to think about.


> A new snuba/datasets/storages/selectors/ directory was created to house all storage selectors. The custom sessions and errors selectors were also moved here.

Aren't storage selectors part of entities? Wouldn't it make more sense to put it under there?

snuba/datasets/storages/selectors/selector.py
```diff
@@ -0,0 +1,77 @@
...
+        query: Query,
+        query_settings: QuerySettings,
+        storage_and_mappers_list: List[StorageAndMappers],
+    ) -> StorageAndMappers:
+        for sm in storage_and_mappers_list:
+            if sm.storage == self.storage:
+                return sm
+        raise QueryStorageSelectorError(
+            "Unable to select storage. Storage not defined."
+        )
```
volokluev, line 77: In this function, the storage_and_mappers_list is only supposed to have one element. I don't think you need to be doing any comparisons here just

```python
assert len(storage_and_mappers_list) == 1
return storage_and_mappers_list[0]
```

snuba/datasets/storages/selectors/errors.py
```diff
@@ -0,0 +1,34 @@
...
+from snuba.datasets.storages.selectors.selector import QueryStorageSelector
+from snuba.datasets.storages.storage_key import StorageKey
+from snuba.query.logical import Query
+from snuba.query.query_settings import QuerySettings
+
+
+class ErrorsQueryStorageSelector(QueryStorageSelector):
+    def __init__(self) -> None:
+        self.__errors_table = get_writable_storage(StorageKey.ERRORS)
+        self.__errors_ro_table = get_storage(StorageKey.ERRORS_RO)
```
volokluev, line 15: This pattern is duplicating knowledge that is already in the configuration and then using `get_storage_mapping_pair` to do this reverse lookup, it's not very intuitive. I think there are more egonomic ways to do this. 

For example, we can provide the guarantee that `storage_and_mappers_list` is always going to be in the order that it is defined in configuration, that way you can do something like:


```python3
   def select_storage(
        self,
        query: Query,
        query_settings: QuerySettings,
        storage_and_mappers_list: List[StorageAndMappers],
    ) -> StorageAndMappers:
        use_readonly_storage = (
            state.get_config("enable_events_readonly_table", False)
            and not query_settings.get_consistent()
        )
        errors_ro, errors = storage_and_mappers_list
        if use_readonly_storage:
            return errors_ro
        return errors
```

That way you don't need to be worrying about implementing a good reverse mapping function and the coherence of the configuration is preserved

snuba/datasets/storages/selectors/selector.py
volokluev: You don't need the `__init__` function either in that case.

snuba/datasets/storages/selectors/selector.py
enochtangg: Makes sense, updated to compare using storage key instead.

snuba/datasets/storages/selectors/selector.py
enochtangg: Updated.

enochtangg:
> Why not have the default be implicit and have the entity fail to instantiate if the deafult is not sufficient? You can accomplish this with the `__post_init__` method of the dataclass. Seems like one fewer config option for users to think about.

I'm okay with making this default storage selector implicit too. My initial thought was since storages defined in the entity will eventually be formatted as an array (in a subsequent PR), implicit selection from an array might not be too obvious. I'm fine with either though.

> Aren't storage selectors part of entities? Wouldn't it make more sense to put it under there?
You're right, I will move this directory over.

snuba/datasets/storages/selectors/selector.py
enochtangg: Updated!

snuba/datasets/storages/selectors/errors.py
enochtangg: Right, this is duplicating knowledge that is already in configuration, but it's also the way we do things today (selectors define storages in its constructor). However, I'm also not a fan of the reverse lookup. 

Guaranteeing order is an interesting approach (e.g. sorting storages by some condition). Is what you're suggesting, read-only storages should always be defined first in the array, followed by writable storages? What if there are multiple readable storages? How should they be ordered?

To summarize the problem, it is the job of `select_storage()` to determine which storage (defined on entity) to use based on input parameters. If the selection choice is always binary (readable or writable), it could probably just check `isinstance()` to determine which storage in the array (len=2) to choose. However, this might not always be the case with multiple readable storages.

enochtangg: Updated this to remove the reverse lookup thing and to check instance type instead. This solution would not work for multiple readable storages though. But perhaps we don't need to design for that yet.

snuba/datasets/entities/sessions.py
```diff
@@ -310,7 +265,8 @@ def __init__(self) -> None:
             storages=[storage],
             query_pipeline_builder=SimplePipelineBuilder(
                 query_plan_builder=StorageQueryPlanBuilder(
-                    storages=[StorageAndMappers(storage, TranslationMappers())]
+                    storages=[StorageAndMappers(storage, TranslationMappers())],
+                    selector=DefaultQueryStorageSelector(),
```
evanh, line 269: How is it possible this is using the `DefaultQueryStorageSelector`? Shouldn't it be using the sessions selector?

evanh (approved):

snuba/datasets/entities/sessions.py
enochtangg: `ORG_SESSIONS` only uses the org materialized sessions storage so the default is possible. `SESSIONS` on the other hand requires the storage selector.

evanh: Ah got it. OK cool.

PR #3547, opened 2022-12-19
ref(multistorage-consumer): Stop using ConsumerStrategyFactory

lynnagara:
The KafkaConusmerStrategyFactory is supposed to get deprecated. This change refactors the multistorage consumer to stop using it there. It also removes the need for the generic `ConsumerStrategyFactory` which previously needed to be generic so it could take a MultistorageKafkaPayload instead of a KafkaPayload in the multistorage scenario.

Soon we will be adding a new processing step for payload decoding and schema validation - this will happen between the DLQ and the rest of the message processing steps.

This refactor will make it easier to introduce the decoder/validator as it needs to happen in different places in the main consumer vs the multistorage consumer.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -27,13 +27,22 @@ import rapidjson
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.backends.kafka.commit import CommitCodec
 from arroyo.commit import Commit as CommitLogCommit
-from arroyo.processing.strategies import FilterStep
+from arroyo.processing.strategies import (
+    CollectStep,
+    CommitOffsets,
+    FilterStep,
+    ParallelCollectStep,
+    ParallelTransformStep,
+)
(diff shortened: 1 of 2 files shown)


snuba/consumers/consumer.py
```diff
@@ -795,31 +814,62 @@ def __init__(
...
+                self.__collector,
+                CommitOffsets(commit),
+                self.__max_batch_size,
+                self.__max_batch_time,
+            )
+        )
+
+        transform_function = self.__process_message_fn
+
+        inner_strategy: ProcessingStep[MultistorageKafkaPayload]
```
fpacifici, line 846: Do you actually need to declare this here ?

fpacifici (approved):

snuba/consumers/consumer.py
lynnagara: Yeah, i think it was needed. Since it can be either TransformStep or ParallelTransformStep, it doesn't infer the right type.

PR #3550, opened 2022-12-21
ref: Remove "kafka_messages_to_skip" config

lynnagara:
Removes the ability to skip Kafka messages via config. This functionality was never used, and shouldn't ever be (invalid messages should be handled via dlqs) not this mechanism.

The motivations to remove this are to:
- remove unnecessary complexity that is never used
- align the consumer and multistorage consumer implementations which will be merged together soon
- make it easier to provide a shared decoder/schema validator implementation for both single storage and multistorage consumer

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -57,7 +57,6 @@ from snuba.processor import (
     MessageProcessor,
     ReplacementBatch,
 )
-from snuba.state import get_config
 from snuba.utils.metrics import MetricsBackend
 from snuba.utils.metrics.wrapper import MetricsWrapper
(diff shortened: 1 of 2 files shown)


evanh (commented):
Is this removing something that self hosted users might be using? If so, then I think this requires a major version change.

lynnagara:
> Is this removing something that self hosted users might be using? If so, then I think this requires a major version change.

No, definitely not. This is mechanism where you can manually enter a runtime config containing the offset of a particular message and it gets skipped. Self hosted users wouldn't be doing that and aren't even running the Snuba admin container.

evanh (approved):
Cool, thanks!

PR #1406, opened 2020-10-09
Add kafka tls authentication to settings

alexef:
Fixes: #974. Related to: https://github.com/getsentry/sentry/pull/21275

Add support for additional configuration (like sasl mechanisms and more) to the Snuba settings.

```python
BBROKER_CONFIG: Mapping[str, Any] = {
    "bootstrap.servers": os.environ.get("DEFAULT_BROKERS", "localhost:9092"),
    "sasl.mechanism": os.environ.get("SASL_MECHANISM"),
    "sasl.username": os.environ.get("SASL_PLAIN_USERNAME"),
    "sasl.password": os.environ.get("SASL_PLAIN_PASSWORD"),
    "security.protocol": os.environ.get("SECURITY_PROTOCOL"),
}
STORAGE_BROKER_CONFIG: Mapping[str, Mapping[str, Any]] = {}
```

Note that even there is a single mapping of configuration, one can still use multiple brokers for bootstrapping communication with kafka.

This is in line with how [Sentry itself](https://github.com/getsentry/sentry/blob/4389bf5a4a7421d5ac3322c7f27c695d8fcef1cb/src/sentry/conf/server.py#L1853-L1859) configures the connections to kafka.

PR #3551, opened 2022-12-21
Allow configuring kafka sasl parameters

cragonnyunt:
This PR allows snuba to connect to kafka brokers protected by sasl plaing username/password. The setting was initially introduced in #1406 but the author decided to remove in the end. 

Benefits for docker and kubernetes users - when I use the chart https://github.com/sentry-kubernetes/charts for deployment, the snuba scripts can't connect to external kafka brokers as sasl username/password can't be set via env. There is no need to turn off sasl protection of kafka, if it can be set via environment variables. 

### Legal Boilerplate


lynnagara (approved):

cragonnyunt:
After investigating the pipeline failure, I found the default value for `sasl.mechanism` is not empty text, it is set to `GSSAPI`. 

Ref - https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md

cragonnyunt:
It looks like transient failure in CI pipeline, please make it re-run again?

PR #3553, opened 2022-12-21
feat: Consumer schema validation

lynnagara:
The consumer entrypoint now has a `--validate-schema` option which (if passed) will perform schema validation on every message where a schema is published for that topic.

The `get_schema()` function is supposed to return either the json schema or none if a schema cannot be found for that topic. Currently this function is just a placeholder that never returns a schemas. Eventually it will be updated to return the schema published from `https://github.com/getsentry/sentry-kafka-topics` package.

The insert API used in test mode will always perform schema validation.


evanh (approved):

lynnagara:
Note: tests are currently failing because it depends on a new version of Arroyo being published that includes this change: https://github.com/getsentry/arroyo/pull/166

PR #3555, opened 2022-12-22
feat(consumer): Remove parallel collect option

lynnagara:
It is always better to use parallel collect rather than collect in Snuba consumers. It was already the default. Let's simplify and just remove the option (that was never used anyway).

This change depends on https://github.com/getsentry/snuba/pull/3547


evanh (approved):

snuba/consumers/consumer.py
```diff
@@ -824,21 +819,12 @@ def create_with_partitions(
...
-                CommitOffsets(commit),
-                self.__max_batch_size,
-                self.__max_batch_time,
-            )
+        collect = ParallelCollectStep(
+            self.__collector,
+            CommitOffsets(commit),
+            self.__max_batch_size,
+            self.__max_batch_time,
+            10.0,
```
onewland, line 827: nit: the meaning of 10.0 is is not super clear in this call. could we do one of:
1. use named arguments rather than positional here
2. make a named constant of DEFAULT_PARALLEL_COLLECT_TIMEOUT_MS = 10.0

in general I think named over positional is better when the argument count is > 3. It would be easy to mix up `self.__max_batch_time` and `DEFAULT_PARALLEL_COLLECT_TIMEOUT_MS`, right (same type)?

lynnagara: Updated to a named argument. The 10.0 value for the timeout actually happens to be the default in Arroyo so another option could've been to omit it. But it's probably not a bad idea to be explicit about it here.

untitaker (approved):

PR #3561, opened 2022-12-30
feat(oncall): add the request_id to the query breadcrumbs

volokluev:
Searching for queries by id in the querylog requires the request_id because that is the `orderBy` key for the table. Without it, the queries timeout. This assures that any query that makes it to sentry has a request_id attached to it


lynnagara (approved):

.github/workflows/ci.yml
```diff
@@ -361,6 +361,7 @@ jobs:
             -e REDIS_HOST=sentry_redis \
             -e REDIS_PORT=6379 \
             -e REDIS_DB=1 \
+            -e ENABLE_AUTORUN_MIGRATION_SEARCH_ISSUES=1 \
```
evanh, line 364: Why this change?

volokluev: It was breaking CI to not have it. It's in master now

PR #3569, opened 2023-01-05
feat: Deprecate settings.TOPIC_PARTITION_COUNTS

lynnagara:
The goal of this change is to reduce the amount of unnecessary configuration in Snuba. Requiring operators of Snuba to set the right TOPIC_PARTITION_COUNTS values is error prone and causes additional work and complexity during incidents. Now the KafkaTopicSpec.partitions_number fetches the real partitions number from the broker instead of the hardcoded setting.

The KafkaTopicSpec class now only returns actual partition counts instead of counts from configuration. The implication of this is that it can no longer be used to bootstrap topics that don't exist yet (e.g. in dev). Topics bootstrapped via the `snuba bootstrap` command are now always created with 1 partition.


snuba/utils/manage_topics.py
```diff
@@ -33,3 +39,37 @@ def create_topics(client: AdminClient, topics: Sequence[Topic]) -> None:
         except KafkaException as err:
             if err.args[0].code() != KafkaError.TOPIC_ALREADY_EXISTS:
                 logger.error("Failed to create topic %s", topic, exc_info=err)
+
+
+def delete_topics(client: AdminClient, topics: Sequence[Topic]) -> None:
```
evanh, line 44: Are these functions purely for use in tests?

evanh (commented):
Conceptually this looks good to me.

snuba/datasets/table_storage.py
```diff
@@ -57,20 +59,9 @@ def get_physical_topic_name(self, slice_id: Optional[int] = None) -> str:
 
         return physical_topic
 
-    @property
+    @cached_property
```
fpacifici, line 62: I don't think this will work, or at least the result will be very tricky to manage. 
@cached_property caches the response for the duration of the instance, which, for data model object means as long as the pod is not restarted. 

This would make a repartitioning of a topic (adding partitions/removing partitions) more complex as there would be a period of time when the number of partitions in the topic does not reflect what the pod thinks. 

We generally fetch the number of partitions when starting a node that depends on the number of partitions like the scheduler. The repartitioning always triggers a rebalancing. Why not removing this entirely from the `KafkaTopicSpec` and use the api directly upon rebalancing where needed ?

snuba/datasets/table_storage.py
```diff
@@ -57,20 +59,9 @@ def get_physical_topic_name(self, slice_id: Optional[int] = None) -> str:
...
-        # and KAFKA_BROKER_CONFIG
-        return settings.TOPIC_PARTITION_COUNTS.get(self.topic_name, 1)
-
-    @property
-    def replication_factor(self) -> int:
-        return 1
-
-    @property
-    def topic_creation_config(self) -> Mapping[str, str]:
-        return get_topic_creation_config(self.__topic)
```
fpacifici, line 73: What happened to these two ?

fpacifici (commented):
I would suggest a different approach that would allow us to do repartitioning without downtime and without having to restart snuba: stop making the partition count a property of the topic spec in the data model and use directly the Kafka client api when needed. Which is during a rebalancing.

PR #3571, opened 2023-01-05
feat(issue-platform): process extra fields from event

barkbarkimashark:
This PR extracts additional field values from the event payload so it can be used for querying.

### DIFF
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -65,6 +65,14 @@ schema:
 readable_storage: search_issues
 writable_storage: search_issues
 query_processors:
+  - processor: TimeSeriesProcessor
+    args:
+      time_group_columns:
+        time: client_timestamp
+      time_parse_columns:
+        - detection_timestamp
+        - client_timestamp
+        - receive_timestamp
   - processor: OrganizationRateLimiterProcessor
     args:
       org_column: organization_id
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -65,14 +65,6 @@ schema:
 readable_storage: search_issues
 writable_storage: search_issues
 query_processors:
-  - processor: TimeSeriesProcessor
-    args:
-      time_group_columns:
-        time: client_timestamp
-      time_parse_columns:
-        - detection_timestamp
-        - client_timestamp
-        - receive_timestamp
   - processor: OrganizationRateLimiterProcessor
     args:
       org_column: organization_id
--- a/snuba/datasets/processors/search_issues_processor.py
+++ b/snuba/datasets/processors/search_issues_processor.py
@@ -70,6 +70,7 @@ class IssueEventData(TypedDict, total=False):
     dist: Optional[str]  # tags[sentry:dist] -> dist
     # (tags[sentry:user] or user[id]) -> user
 
+    # TODO: when we process contexts, make sure to also conditionally extract out trace_id
     # contexts aliases
     # contexts.trace.trace_id -> trace_id
 
@@ -151,7 +152,7 @@ class SearchIssuesMessageProcessor(DatasetMessageProcessor):
     def _process_request_data(
         self, event_data: IssueEventData, processed: MutableMapping[str, Any]
     ) -> None:
-        request = event_data.get("request") or {}
+        request = event_data.get("request", {})
         http_data: MutableMapping[str, Any] = {}
         extract_http(http_data, request)
         processed["http_method"] = http_data["http_method"]
--- a/snuba/datasets/processors/search_issues_processor.py
+++ b/snuba/datasets/processors/search_issues_processor.py
@@ -131,11 +131,13 @@ class SearchIssuesMessageProcessor(DatasetMessageProcessor):
     def _process_tags(
         self, event_data: IssueEventData, processed: MutableMapping[str, Any]
     ) -> None:
-        tags_maybe = event_data.get("tags", None)
-        if not tags_maybe:
+        existing_tags = event_data.get("tags", None)
+        tags: Mapping[str, Any] = _as_dict_safe(cast(Dict[str, Any], existing_tags))
+        if not existing_tags:
             processed["tags.key"], processed["tags.value"] = [], []
-        tags: Mapping[str, Any] = _as_dict_safe(cast(Dict[str, Any], tags_maybe))
-        processed["tags.key"], processed["tags.value"] = extract_extra_tags(tags)
+        else:
+            processed["tags.key"], processed["tags.value"] = extract_extra_tags(tags)
+
         promoted_tags = {col: tags[col] for col in self.PROMOTED_TAGS if col in tags}
         processed["release"] = promoted_tags.get(
             "sentry:release",
--- a/snuba/datasets/processors/search_issues_processor.py
+++ b/snuba/datasets/processors/search_issues_processor.py
@@ -38,7 +38,6 @@ class IssueOccurrenceData(TypedDict, total=False):
 
 class IssueEventData(TypedDict, total=False):
     # general data from event.data map
-    trace_id: Optional[str]
     received: float
     client_timestamp: float
     tags: Mapping[str, Any]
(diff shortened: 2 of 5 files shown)


snuba/datasets/processors/search_issues_processor.py
```diff
@@ -99,6 +127,43 @@ def _process_user(
...
+        )
+        processed["user"] = promoted_tags.get("sentry:user")
+        processed["dist"] = _unicodify(
+            promoted_tags.get("sentry:dist", event_data.get("dist")),
+        )
+
+    def _process_request_data(
+        self, event_data: IssueEventData, processed: MutableMapping[str, Any]
+    ) -> None:
+        request = event_data.get("request") or {}
```
ceorourke, line 154: I think you can put the empty dict as the default arg in `.get` like `event_data.get("request", {})`

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -61,10 +82,10 @@ class SearchIssueEvent(TypedDict, total=False):
     organization_id: int
     project_id: int
     event_id: str
-    group_id: int  # backwards compatibility
-    group_ids: Sequence[int]
+    group_id: int
```
ceorourke, line 85: Why is this changing to singular `group_id` now? I thought we'd always have multiple ids for non-error issue stuff.

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -38,20 +56,23 @@ class IssueOccurrenceData(TypedDict, total=False):
...
     request: Mapping[str, Any]  # http_method, http_referer
 
+    # tag aliases
+    environment: Optional[str]  # tags[environment] -> environment
+    release: Optional[str]  # tags[sentry:release] -> release
+    dist: Optional[str]  # tags[sentry:dist] -> dist
+    # (tags[sentry:user] or user[id]) -> user
+
+    # contexts aliases
+    # contexts.trace.trace_id -> trace_id
```
ceorourke, line 74: Why is this and line 71 commented out? Something for later on or left in?

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -77,6 +98,13 @@ def ensure_uuid(value: str) -> str:
 class SearchIssuesMessageProcessor(DatasetMessageProcessor):
     FINGERPRINTS_HARD_LIMIT_SIZE = 100
 
+    PROMOTED_TAGS = {
```
ceorourke, line 101: What does promoted mean in this context?

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -99,6 +127,43 @@ def _process_user(
 
         return
 
+    def _process_tags(
+        self, event_data: IssueEventData, processed: MutableMapping[str, Any]
+    ) -> None:
+        tags_maybe = event_data.get("tags", None)
+        if not tags_maybe:
+            processed["tags.key"], processed["tags.value"] = [], []
```
evanh, line 135: I think you should `return` here.

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -99,6 +127,43 @@ def _process_user(
...
+    def _process_request_data(
+        self, event_data: IssueEventData, processed: MutableMapping[str, Any]
+    ) -> None:
+        request = event_data.get("request") or {}
+        http_data: MutableMapping[str, Any] = {}
+        extract_http(http_data, request)
+        processed["http_method"] = http_data["http_method"]
+        processed["http_referer"] = http_data["http_referer"]
+
+    def _process_sdk_data(
```
evanh, line 160: Do you need this data?

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Good suggestion! Will do.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Yeah, I had the same discussion with Dan. An occurrence event will be sent for each group(s) that resulted in processing that event. So if multiple Issues are detected for a single issue, then multiple events will be sent to the eventstream.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Yeah, this is reserved for later since right now we're not even sending the context data through. Will add a TODO comment to be safe.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: These are basically 'special' tag values we lift directly from tags and place them in top-level columns. I basically took this pattern from how transactions was written. I presume this was done for performance - so instead of indexing into the tags array, we do some special processing when ingesting events and normalize it into its own specific column.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: I actually had it return early before, but then the extra params won't get processed:
```
processed["release"] = promoted_tags.get(
            "sentry:release",
            event_data.get("release"),
        )
```

So events like:
```
{
        "project_id": 1,
        "organization_id": 2,
        "group_id": 3,
        "event_id": str(uuid.uuid4()),
        "retention_days": 90,
        "primary_hash": str(uuid.uuid4()),
        "datetime": datetime.utcnow().isoformat() + "Z",
        "platform": "other",
        "data": {
            "tags": {}
            "release": "release@1.2.3"
        }
        ...
    }
```
would get skipped entirely.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Not totally sure, but I think they're needed to support searching like this?
Screen Shot 2023-01-05 at 5 04 57 PM

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -99,6 +127,43 @@ def _process_user(
...
         return
 
+    def _process_tags(
+        self, event_data: IssueEventData, processed: MutableMapping[str, Any]
+    ) -> None:
+        tags_maybe = event_data.get("tags", None)
+        if not tags_maybe:
+            processed["tags.key"], processed["tags.value"] = [], []
+        tags: Mapping[str, Any] = _as_dict_safe(cast(Dict[str, Any], tags_maybe))
+        processed["tags.key"], processed["tags.value"] = extract_extra_tags(tags)
```
rjo100, line 137: does this overwrite line 135 if tags are `None`?

snuba/datasets/processors/search_issues_processor.py
```diff
@@ -99,6 +127,43 @@ def _process_user(
 
         return
 
+    def _process_tags(
+        self, event_data: IssueEventData, processed: MutableMapping[str, Any]
+    ) -> None:
+        tags_maybe = event_data.get("tags", None)
```
rjo100, line 133: maybe make this a more descriptive name like `event_tags`

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Yeah we do:
https://github.com/getsentry/snuba/blob/8fad778062c6e3db4be1bf36bc21ef08e61b7185/snuba/datasets/configuration/issues/entities/search_issues.yaml#L41-L42

snuba/datasets/configuration/issues/entities/search_issues.yaml
```diff
@@ -65,6 +65,14 @@ schema:
 readable_storage: search_issues
 writable_storage: search_issues
 query_processors:
+  - processor: TimeSeriesProcessor
```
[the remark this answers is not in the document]
barkbarkimashark: Fair enough. I recall you mentioning before about `TimeSeriesProcessor`. One of the legacy queries I was trying to get working looks to be grouping on `time`. Let me take a crack at porting that bit over. Appreciate the clarification.

snuba/datasets/processors/search_issues_processor.py
evanh: OK cool, just making sure.

snuba/datasets/processors/search_issues_processor.py
barkbarkimashark: Ahh, right. I rearranged `process_tags` to be a bit more readable and to avoid doing dupe work.

evanh (approved):
This looks fine to me.

PR #3574, opened 2023-01-05
docs(schema): why do we promote tags?

onewland:
wanted to document the rationale behind why you see "promotion" throughout Snuba as it's come up a couple times for me recently

https://github.com/getsentry/snuba/pull/3571/files#r1062820314

### DIFF
--- a/docs/source/clickhouse/schema_design.rst
+++ b/docs/source/clickhouse/schema_design.rst
@@ -1,0 +1,27 @@
+ClickHouse Schema Design Best Practices
+=======================================
+
+Columns based on dictionary data
+--------------------------------
+
+ClickHouse is a columnar datastore, and it loads columns on-demand based on the columns referenced
+in the query (both the columns ``SELECT``ed and those part of the ``WHERE`` clause). Commonly,
+a data schema contains a flexible key:value pair mapping (for example: tags) and stores that
+data in a column that contains two ``Nested`` arrays where the first array contains the keys
+of the dictionary and the second array contains the values. This works well when
+your dataset design gives you the ability to filter based on other attributes to a small
+number of rows and the flexibility of completely arbitrary keys and values are a real requirement.
+Often, however, a ClickHouse user is interested in rows that contain a specific tag key or a
+specific ``key=value`` pair and in this case one is often best-served by creating a top-level
+column and duplicating (we often call it "promoting" in code) the data.
+
+This is more efficient on two dimensions:
+
+1. A dictionary style-column can be arbitrarily sized and, especially if you know you will often only
+need a single value from the dictionary, it may be that just returning
+this column in your results blows up the size of your **query results** result in-memory. It also can
+exhaust the cache and in that manner slow down other unrelated queries.
+
+2. Queries that filter based on an array column have to load the entire column into memory
+in order to do filtering. This makes it more difficult and expensive for the query planner to build
+indexing on the presence of a key.
--- a/docs/source/clickhouse/schema_design.rst
+++ b/docs/source/clickhouse/schema_design.rst
@@ -4,9 +4,13 @@ ClickHouse Schema Design Best Practices
 Columns based on dictionary data
 --------------------------------
 
-ClickHouse is a columnar datastore, and it loads columns on-demand based on the columns referenced
-in the query (both the columns ``SELECT``ed and those part of the ``WHERE`` clause). Commonly,
-a data schema contains a flexible key:value pair mapping (for example: tags) and stores that
+ClickHouse is a columnar datastore, and at run-time it loads columns on-demand
+based on the columns referenced in the query (both the columns ``SELECT`` ed
+and those part of the ``WHERE`` clause). The ability to store different columns independently
+and not load them for every row for every query is part of the performance advantage that
+ClickHouse provides over a traditional RDBMBS (like PostgreSQL).
+
+Commonly, a data schema contains a flexible key:value pair mapping (for example: tags) and stores that
 data in a column that contains two ``Nested`` arrays where the first array contains the keys
 of the dictionary and the second array contains the values. This works well when
 your dataset design gives you the ability to filter based on other attributes to a small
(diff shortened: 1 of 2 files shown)


docs/source/clickhouse/schema_design.rst
```diff
@@ -0,0 +1,35 @@
...
+Columns based on dictionary data
+--------------------------------
+
+ClickHouse is a columnar datastore, and at run-time it loads columns on-demand
+based on the columns referenced in the query (both the columns ``SELECT`` ed
+and those part of the ``WHERE`` clause). The ability to store different columns independently
+and not load them for every row for every query is part of the performance advantage that
+ClickHouse provides over a traditional RDBMBS (like PostgreSQL).
+
+Commonly, a data schema contains a flexible key:value pair mapping (canonically: ``tags``) and stores that
```
nikhars, line 13: contexts are also used for flexible key:value pair mapping. Might want to add them here as well.

docs/source/clickhouse/schema_design.rst
```diff
@@ -0,0 +1,35 @@
...
+data in a column that contains two ``Nested`` arrays where the first array contains the keys
+of the dictionary and the second array contains the values. This works well when
+your dataset design gives you the ability to filter based on other attributes to a small
+number of rows and the flexibility of completely arbitrary keys and values are a real requirement.
+Often, however, a ClickHouse user is interested in rows that contain a specific tag key or a
+specific ``key=value`` pair and in this case one is often best-served by creating a top-level
+column and duplicating (we often call it "promoting" in code) the data.
```
nikhars, line 20: Storing duplicates has storage cost associated with it. Ideally, when a specific array element is being promoted to its own column, you would stop storing it in the array. But the problem is that the promotion does not help with data which is already persisted on Clickhouse. So the promotion would start happening from the day promotion code is merged. Then to use only the new columns in a consistent manner you would have to wait for 90 days data to be written. Then you can stop writing the duplicate data.

docs/source/clickhouse/schema_design.rst
```diff
@@ -0,0 +1,35 @@
...
+
+This duplication is more performance-efficient on two dimensions:
```
nikhars, line 22: I think we can simplify the answer to the first bullet point you provided. How does the following sound?

With array accesses Clickhouse needs to find out the index of the key whose value needs to be looked up. That is an O(N) lookup. The same key maybe located at different indexes for each row. Once the index is known. Looking up its value is O(1) since you perform direct index access. If the query has to lookup M rows, this becomes an O(M * N) in terms of time complexity.

With direct column access, there is no index lookup for a key. So the overall complexity to lookup N rows is O(N).

The second point about adding indexes is valid.

docs/source/clickhouse/schema_design.rst
onewland: I'll change the wording around duplication, that makes sense

onewland:
addressed all comments, I think

nikhars (approved):

docs/source/clickhouse/schema_design.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Commonly, a data schema contains a flexible key:value pair mapping
+(canonically: ``tags`` or ``contexts``) and stores that
+data in a column that contains two ``Nested`` arrays where the first array contains the keys
+of the dictionary and the second array contains the values. This works well when
+your dataset design gives you the ability to filter based on other attributes to a small
+number of rows and the flexibility of completely arbitrary keys and values are a real requirement.
+Often, however, a ClickHouse user is interested in rows that contain a specific tag key or a
+specific ``key=value`` pair. In this case one is often best-served by creating a top-level
+column and "promoting" the data to it [#dupe]_.
```
fpacifici, line 21: I think this is misleading as a reason. 
Reading this I would conclude that all scenarios where I query something like ``key=value`` should be promoted. That is the majority of the scenarios where we use tags.
Also the ``tag=value`` case does not really need promotion most of the times as it is covered by the bloomfilter index that makes is potentially cheaper than a dedicated column. 

The dedicated column helps us mostly for arrayJoin, when the tag is present on most rows and where the tag column tends to be really large.

- the arrayJoin case is because the bloomfilter index that filters out data does not help there. We have to scan all tags for all rows 
- the case where the tag key is present on most or all rows and the specific tag value is very very common, because the bloomfilter would filter out very little 


So I would rephrase this paragraph and fix the reason why we promote columns (if there is still a reason at all to be fair)

docs/source/clickhouse/schema_design.rst
```diff
@@ -0,0 +1,46 @@
...
+
+This promotion is preferable on a couple dimensions:
+
+1. Even though it looks like a dictionary, our tags and contexts "maps" are actually arrays of keys and values.
```
fpacifici, line 25: This is not 100% correct as it is missing the bloomfilter piece. 
tag keys/values have a bloomfilter index associated. 
It is true that, in the worst case scenario, the complexity of scanning the tag column to find a row with the ``tag=value`` condition is `O(m * n)` but it is also true that such scenario is supposed not to happen in most scenarios as the bloomfilter drops the majority of blocks that do not contain ``tag=value`` before scanning. 

The actual reason to take into account is the selectivity of your tag. That is what makes us decide whether to promote it or not.
Since the bloomfilter index just drops blocks before scanning, we evaluate if it would drop enough blocks.
- When checking if a tag key exists, the bloomfitler would not drop anything if the tag key exists (with different values) on all rows. Think about something like a `request_id` tag. It will be there always.
- When checking ``tag=value``, a value that is low selectivity (present in most rows) would make the bloomfilter discard almost no blocks. Think about an `http_response` tag, the value `200` will be there almost always. The bloomfilter is useless. 

We promote the tag when the tag is there very often, the bloomfilter is not capable of dropping anything and the column is queried very often.

fpacifici (changes requested):

onewland:
follow-up PR here: https://github.com/getsentry/snuba/pull/3576

PR #3575, opened 2023-01-06
feat(issue-platform): increase occurrence_type_id column size from UInt8 to UInt16

barkbarkimashark:
Turns out UInt8 isn't large enough for our purposes. This PR adds a migration to increase the column size to UInt16.

### DIFF
--- a/snuba/snuba_migrations/search_issues/0003_search_issues_modify_occurrence_type_id_size.py
+++ b/snuba/snuba_migrations/search_issues/0003_search_issues_modify_occurrence_type_id_size.py
@@ -7,11 +7,6 @@ from snuba.migrations.operations import OperationTarget
 
 
 class Migration(migration.ClickhouseNodeMigration):
-    """
-    Heavily influenced by 0004_transactions_add_tags_hash_map.py
-
-    """
-
     blocking = True
 
     def forwards_ops(self) -> Sequence[operations.SqlOperation]:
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/search_issues/0003_search_issues_modify_occurrence_type_id_size.py
```diff
@@ -0,0 +1,38 @@
...
+            operations.ModifyColumn(
+                storage_set=StorageSetKey.SEARCH_ISSUES,
+                table_name=params[0],
+                column=Column("occurrence_type_id", UInt(8)),
+                target=params[1],
+            )
+            for params in [
+                ("search_issues_dist", OperationTarget.DISTRIBUTED),
+                ("search_issues_local", OperationTarget.LOCAL),
+            ]
```
fpacifici, line 37: @nikhars  do you know if Clickhosue allows us to downscale a column in place (assuming there is no data larger than 8 bits yet)

nikhars: Yes downscaling is allowed. I tried with a simple MergeTree engine on my local environment. Even if you have an existing value larger than the new downscaled variant, Clickhouse would let the DDL change happen. During selects the larger values got casted to lower values.

dbanda (approved):

PR #3576, opened 2023-01-06
docs(schema): update tag promotion doc to talk about bloom filters

onewland:
@fpacifici added some comments on the original PR, so I geared this more into a discussion of why tag indexes can fail on certain kinds of queries

### DIFF
--- a/docs/source/clickhouse/schema_design.rst
+++ b/docs/source/clickhouse/schema_design.rst
@@ -31,8 +31,9 @@ This works well when your dataset and query design gives you the ability to
 filter for exact matches and a large number of rows will NOT be an exact match.
 Often, however, a ClickHouse query filters for rows that contain a substring match or regular
 expression match for a tag value of a given key. This makes bloom filter indexes
-not usable for the query and requires moving (or promoting) those relevant values for a given tag
-key to a new separate column which can support these queries more efficiently [#dupe]_.
+not usable for the query and, depending on the other selectivity attributes of your query,
+can necessitate moving (or promoting) those relevant values for a given tag key to a new separate
+column [#dupe]_.
 
 .. _selectivity:
 
--- a/docs/source/clickhouse/schema_design.rst
+++ b/docs/source/clickhouse/schema_design.rst
@@ -56,7 +56,7 @@ Bloom filter indexing on dictionary-like columns
 To facilitate faster searching on dictionary columns, we tend to create bloom filter indices
 on a hashes of both the unique ``key`` values of each row as well as hashes of all the ``key=value``
 pairs of each row. The `bloom filter <https://en.wikipedia.org/wiki/Bloom_filter>`_  registers these
-in a stochastic data structure designed to quickly determine which elements do NOT exist in a set,
+in a stochastic data structure designed to quickly determine which elements do NOT exist in a set.
 So that it can model the entire unbounded keyspace in a fixed amount of memory, a bloom filter
 is designed to have false positives. This means that there is actually a performance **penalty**
 if the value is often present in the underlying set: First, the value must be tested
--- a/docs/source/clickhouse/schema_design.rst
+++ b/docs/source/clickhouse/schema_design.rst
@@ -2,6 +2,14 @@
 ClickHouse Schema Design Best Practices
 =======================================
 
+.. tip::
+    This is a work-in-progress document for collecting ClickHouse schema and querying
+    best-practices based on experiences running ClickHouse at scale at Sentry.
+    It is subject to change and if something doesn't seem right please
+    submit a PR to Snuba.
+
+.. contents:: :local:
+
 
 Columns based on dictionary data (tag promotion)
 ------------------------------------------------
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

nikhars (approved):

docs/source/clickhouse/schema_design.rst
```diff
@@ -16,31 +18,37 @@ data in a column that contains two ``Nested`` arrays where the first array conta
...
-   With direct column access, there is no index lookup for a key. The overall complexity to lookup M rows
-   for "promoted columns" is O(M).
-2. Queries that filter based on an array column may have to load the entire column into memory
-   in order to do filtering. Adding a new index on top of a top-level column is a more
-   straightforward operation that guarantees more consistent performance outcomes vs. iterating
-   over an array
+Often, however, a ClickHouse user is predominantly interested in rows that contain a specific tag key or a
+specific ``key=value`` pair that occurs in a very large percentage of the table data (for example,
+``http_status_code=200`` or ``hasKey(http_status_code)``). This breaks any efficiency
+of the bloom-filter style index on dictionary columns (see :ref:`bloom`).
```
fpacifici, line 24: I am not sure I follow this paragraph. Does it mean that those conditions are not efficient with a bloom filter ?

It would be the other way around.
By default a nested column does not have a bloom filter index associated. 
That way, in order to find rows that contain a specific key-value pair require scanning the entire column and scanning all the tags for each row.

When we add the bloom filter index, we can make this search efficient as the bloom filter allows us to drop blocks that are very unlikely to contain `key=value` or `key` before scanning.

So the bloom filter index helps in these cases and only in these cases.

The bloom filter index is useless, though, in any other use of the nested column like:
- regex expressions
- key LIKE %something%
- arrayJoin(tags.value)

docs/source/clickhouse/schema_design.rst
```diff
@@ -16,31 +18,37 @@ data in a column that contains two ``Nested`` arrays where the first array conta
...
+In this case one is often best-served by creating a top-level column and "promoting"
+the data to it [#dupe]_.
+
+It is very efficient for queries to filter by a specific matching value that exists in
+a high number of rows for a given tag key being non-empty or key:value pair on a column
```
fpacifici, line 29: Are you talking about using the bloom filter or the promoted column ? 
The problem is that neither is efficient. If the value has low selectivity (present in few rows) the search can be made efficient.

- low selectivity, key=value condition -> do not promote the column, use the bloom filter on the nested column
- low selectivity, key LIKE %something% -> bloom filter does not help, if it is a common condition, promoted column is a bit better. It gets better as the nested column increases in size
- high selectivity -> both are bad. you need to scan all the blocks. The promoted column reduces the amount of data to scan making things a bit better

docs/source/clickhouse/schema_design.rst
```diff
@@ -16,31 +18,37 @@ data in a column that contains two ``Nested`` arrays where the first array conta
...
+
+.. _bloom:
+
+Bloom filter indexing on dictionary-like columns
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To facilitate faster searching on dictionary columns, we tend to create bloom filter indices
+on a hash of both the unique ``key`` values of each row as well as the ``key=value`` pairs
+of each row. The `bloom filter <https://en.wikipedia.org/wiki/Bloom_filter>`_ index that stores these
+is a stochastic data structure designed to quickly determine which elements do NOT exist in a set,
+but provides little to no optimization if a value is often present in the underlying set.
```
fpacifici, line 42: In this case it actually tend to be counter productive as you have to load the index first., you do not drop anything, and then you have to scan the whole column. Without the index you only scan the column.

docs/source/clickhouse/schema_design.rst
onewland: Those conditions wouldn't be made efficient with a bloom filter index because (presumably) most of the rows in the hypothetical table have a `http_status_code` key and have `http_status_code=200` and therefore your BF will almost always be falling back to a table scan.

I do think that a mention of selectivity in queries like your below comment will maybe help make this clearer

docs/source/clickhouse/schema_design.rst
onewland: good point, changing this wording

onewland:
@fpacifici let me know if my latest changes resolve your concerns

docs/source/clickhouse/schema_design.rst
```diff
@@ -11,36 +21,64 @@ and not load them for every row for every query is part of the performance advan
 ClickHouse provides over a traditional RDBMS (like PostgreSQL).
 
 Commonly, a data schema contains a flexible key:value pair mapping
-(canonically: ``tags`` or ``contexts``) and stores that
+(canonically at Sentry: ``tags`` or ``contexts``) and stores that
 data in a column that contains two ``Nested`` arrays where the first array contains the keys
-of the dictionary and the second array contains the values. This works well when
-your dataset design gives you the ability to filter based on other attributes to a small
+of the dictionary and the second array contains the values, indexed by a bloom filter. This works well when
+your dataset and query design gives you the ability to filter for exact matches to a small
```
fpacifici, line 27: This is not accurate. 

> stores that data in a column that contains two ``Nested`` arrays where the first array contains the keys of the dictionary and the second array contains the values

The column itself is `Nested`. The content is two standard arrays, one for keys and one for values.

> indexed by a bloom filter

We do not index the Nested column with a bloom filter. You can refer to the errors table schema for the details.
We index the tag keys array and a second column which is the tags hashmap. The tags hashmap is a simple array of integer where each entry is the hash of the `key=value` string.

docs/source/clickhouse/schema_design.rst
```diff
@@ -11,36 +21,64 @@ and not load them for every row for every query is part of the performance advan
...
 
-1. Even though it looks like a dictionary, our tags and contexts "maps" are actually arrays of keys and values.
-   With array accesses Clickhouse needs to locate the index of the key whose value needs to be looked up.
-   That is an O(N) lookup. The same key may be located at different indexes for each row.
-   Once the index is known, looking up its value is O(1) since you perform direct index access.
-   If the query has to lookup M rows, this becomes an O(M * N) in terms of time complexity.
+Queries are much more efficient when they have the attribute of being low-selectivity on
+the table indexes -- meaning the query conditions on indexed columns filter the dataset
+to a very small proportion of the overall number of rows. Therefore it's ideal to design
+schemas and queries that cause searches on a very low number of rows. High selectivity
```
fpacifici, line 49: I think this may be a little misleading. 
The point of a real time analytics database is to be able to produce aggregations over very large amount of rows very fast. This is the primary use case and Clickhouse does great at that.
What is written here make it seems that such usage is an anti-pattern. If we had a use case where most queries were very low selectivity Clickhouse would not be the right storage. A key value pair would be the one. 

The point of introducing these optimization is to make low selectivity queries still similarily efficient as they would be on a database with b-tree indexes like postgres by reducing the amount of data to scan.

fpacifici (commented):
Please see the comments inline, otherwise it is good.

docs/source/clickhouse/schema_design.rst
evanh: We also only do that on the tag values, we don't have the same indexes for contexts. So I don't know if this indexing pattern is "common".

docs/source/clickhouse/schema_design.rst
evanh: I agree, this is missing some nuance. It depends on the type of conditions used. A `count()` over 90 days with a simple `project_id` filter can execute much faster than a regex search on tags over 7 days, even though the second is much lower selectivity.

docs/source/clickhouse/schema_design.rst
onewland: > We also only do that on the tag values, we don't have the same indexes for contexts. So I don't know if this indexing pattern is "common".

There's nothing special about contexts that makes them less index-able, right? I can make a note that we don't currently have bloom filter indexes on contexts but that doesn't seem inherent to the text.

onewland: Also, I think I meant to imply that it's common across datasets, more than specific columns

onewland: ok, adding a note about that

docs/source/clickhouse/schema_design.rst
onewland: Rather than try to address the nuance Evan is suggesting, I think I'm just going to drop that wording and say "lower selectivity = better".

We note in the earlier and later sections that bloom filters can't do anything with fuzzy (regex/LIKE) filtering.

docs/source/clickhouse/schema_design.rst
```diff
@@ -11,36 +21,64 @@ and not load them for every row for every query is part of the performance advan
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To facilitate faster searching on dictionary columns, we tend to create bloom filter indices
+on a hash of both the unique ``key`` values of each row as well as the ``key=value`` pairs
+of each row. The `bloom filter <https://en.wikipedia.org/wiki/Bloom_filter>`_ index that stores these
+is a stochastic data structure designed to quickly determine which elements do NOT exist in a set,
+but there is actually a performance **penalty** if the value is often present in the underlying set.
+First, the value must be tested against the bloom filter (which will always return "maybe
+present"), and after that operation occurs a full scan of the column must be performed.
+
+When searching a dictionary column for a value likely to filter many rows (for example,
```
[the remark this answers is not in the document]
onewland: agree, dropping this paragraph and doing a slight rewording on the one above to account for the index hash descriptions Filippo mentioned above

onewland:
@fpacifici @evanh did some more cleanup here

docs/source/clickhouse/schema_design.rst
```diff
@@ -11,36 +21,54 @@ and not load them for every row for every query is part of the performance advan
...
+data in a ``Nested`` column that contains two arrays where the first array contains the keys
+of the dictionary and the second array contains the values. To make queries faster,
+a column like this can be indexed with bloom filters as described in :ref:`bloom`. In general
+we construct this index across datasets for ``tags`` but not for other columns.
+
+This works well when your dataset and query design gives you the ability to
+filter for exact matches and a large number of rows will NOT be an exact match.
+Often, however, a ClickHouse query filters for rows that contain a substring match or regular
+expression match for a tag value of a given key. This makes bloom filter indexes
+not usable for the query and requires moving (or promoting) those relevant values for a given tag
```
fpacifici, line 34: It does not "require" moving to a promoted column. 
This makes it seems like if we always promote a column in this case, which is not true. Most of the times we do not. 
Promotion makes things better but it is reserved for special cases.

fpacifici (approved):

docs/source/clickhouse/schema_design.rst
onewland: Tell me if my read on this is correct:

If your query pattern is low-selectivity *before* the inexact compare, it's fine to leave the tag un-promoted, because that will just be a filter in memory on a relatively small number of rows.

If your query pattern requires the inexact compare in order to be low-selectivity, then the promoted column becomes more necessary because ClickHouse will never finish iterating over all the arrays of a quasi table scan.

PR #3580, opened 2023-01-09
fix: Change DefaultNoneColumnMapper to use a normal set

evanh:
Dataset configuration doesn't support complex objects for registered classes
like this. In order for the discover entity to be migration to YAML, there needs
to be a different way to initialize this class that can be encoded in YAML.

This mapper is checking if a column name exists in the ColumnSet. That check
compares against the flattened column name stored in the ColumnSet. This can be
achieved by comparing to a normal set instead of a ColumnSet.

### Blast Radius

This should only concern anyone who works with the Discover entity. 
Also, nothing should change when this is merged.

### DIFF
--- a/snuba/clickhouse/translators/snuba/allowed.py
+++ b/snuba/clickhouse/translators/snuba/allowed.py
@@ -140,23 +140,24 @@ class ArgumentMapper(SnubaClickhouseMapper[Argument, Argument]):
     pass
 
 
-@dataclass(frozen=True)
+@dataclass
 class DefaultNoneColumnMapper(ColumnMapper):
     """
-    This maps a list of column names to None (NULL in SQL) as it is done
-    in the discover column_expr method today. It should not be used for
-    any other reason or use case, thus it should not be moved out of
-    the discover dataset file.
+    This takes a list of flattened column names and maps them to None (NULL in SQL) as it is done in the discover column_expr method today.
+    It should not be used for any other reason or use case, thus it should not be moved out of the discover dataset file.
     """
 
-    columns: set[str]
+    column_names: list[str]
+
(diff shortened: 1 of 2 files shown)


volokluev:
I'm not sure this will take us in the direction we want to go in. The problem is more that the DefaultNoneColumnMapper needs to take the columns of the two tables that it's creating a merge table of. Is the plan to just re-enumerate all those columns in the discover entity yaml?

evanh:
> Is the plan to just re-enumerate all those columns in the discover entity yaml?

That's what I was thinking. The list of transaction/event columns is going to need to be enumerated somewhere, since in theory the `entities/discover.py` file is going away.

That list of event/transaction specific columns is referenced in the class itself, but just as a convenience (instead of spelling out all the columns it does `<list of common columns> + event columns + transactions columns`.

I think the Discover entity would list out all the columns (common, event, transactions) as its schema, and then in the mappers define which ones are event specific vs. transaction specific. That will result in a lot of lines of configuration, but more closely maps to how these values are actually used.

snuba/clickhouse/translators/snuba/allowed.py
```diff
@@ -148,7 +149,7 @@ class DefaultNoneColumnMapper(ColumnMapper):
     the discover dataset file.
     """
 
-    columns: ColumnSet
+    columns: set[str]
```
volokluev, line 152: 1. Can we make this a list and then check uniqueness in the `__post_init__` the yaml syntax for sets is pretty ugly
2. Can we fix the docstring and explain that the columns list is a list of strings mapping to the column names?

volokluev (approved):

PR #3582, opened 2023-01-10
metrics: add filter to match SaaS

dbanda:


snuba/snuba_migrations/metrics/0035_metrics_add_prod_filter.py
```diff
@@ -0,0 +1,94 @@
...
+        return [
+            counters_op,
+            operations.DropTable(
+                storage_set=StorageSetKey.METRICS,
+                table_name=get_polymorphic_mv_variant_name(
+                    "distributions", self.mv_version
+                ),
+                target=operations.OperationTarget.LOCAL,
+            ),
+            distributions_op,
```
fpacifici, line 84: The materialized view version column is used differently. 
Clickhouse cannot replace a materialized view in an atomic manner (though newer version allow to mutate the statement, we should try that out). 
So we were used to create a first a new materialized view with a new version, then switch the version in the consumer so that we would start ingesting through the new matview, then we deleted the old matview.
This means the creation of the new matview must happen before you delete the old one and in two different migrations as the change to the consumer has to happen in between.

Details
https://www.notion.so/sentry/Migrations-vs-Materialized-Views-896bce040cb74a8fa9cca2e7c9a859ab

Probably a better option in new clickhosue versions:
https://clickhouse.com/docs/en/sql-reference/statements/alter/view/#alter-live-view-statement

fpacifici (changes requested):
I think we can ignore this discrepancy. 
We used this filter when we migrated to a new materialized view early on in the project to make the migration simpler and quicker. 
That date is far beyond the retention so it is basically irrelevant today. 
I think you can just ignore it.

PR #3587, opened 2023-01-11
feat(replays): Add migration for session and error sample rates

cmanallen:


snuba/snuba_migrations/replays/0008_add_sample_rate.py
```diff
@@ -0,0 +1,63 @@
+from typing import Iterator, Sequence, Tuple
+
+from snuba.clickhouse.columns import Column, Float
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+
+columns: Sequence[Tuple[Column[Modifiers], str]] = [
+    (
+        Column("error_sample_rate", Float(64, Modifiers(nullable=True))),
```
lynnagara, line 10: Just curious if it needs to be nullable / needs 64 bit precision?

lynnagara (approved):

snuba/snuba_migrations/replays/0008_add_sample_rate.py
cmanallen: @lynnagara Yeah they only want to send it on the initial segment.  Null values make sense to me but we could enforce non-null and default to `0.0`.  I can be persuaded to adopt that approach.

Re: precision.  Float32 is probably enough.  Do I need to do anything in my processor to cut the precision or will clickhouse do that for me?

lynnagara: @cmanallen No persuading - null is fine if that makes more sense for your use case

PR #3590, opened 2023-01-12
ref(migrations): revive clickhouse migrations user

MeredithAnya:
Originally we wanted to see if we could make a new clickhouse user that had different permissions than the default user, which resulted in https://github.com/getsentry/snuba/pull/3331, but we didn't end up needing it.

However, now there is work being done to stop using the default clickhouse user for everything. This means we will have to define a clickhouse user for migrations in the configuration (in the ops repo). The corresponding snuba change is to allow actually using a settings based user/password for the migrations, which is what this PR does


snuba/migrations/operations.py
```diff
@@ -50,11 +51,9 @@ def execute(self) -> None:
         else:
             raise ValueError(f"Target not set for {self}")
 
+        database = cluster.get_database()
         for node in nodes:
-            connection = cluster.get_node_connection(
-                ClickhouseClientSettings.MIGRATE, node
-            )
-
+            connection = get_ddl_node_connection(node, database)
```
evanh, line 56: Do these connections get cleaned up at all?

evanh (commented):
Seems fine, but there is some failing CI.

PR #3591, opened 2023-01-12
ref(DC): Reformat how entities specify storages in configuration

enochtangg:
### Summary
Currently, we are not able to move the `sessions` entity into configuration. This is because sessions contains different translation mappers for each of its selectable storages. This PR is responsible for reformatting how storages and translation mappers are defined in configuration. Entities now specify a `Sequence[EntityStorageConnection]`. In addition, a `SimpleQueryStorageSelector` was added to support this new format of storages in configuration.

### Before State
Entity configuration:
```
readable_storage: generic_metrics_distributions
writable_storage: generic_metrics_distributions_raw
translation_mappers: [x,y,z]
```
* Storages were "implicitly" selected in code. For example, when there are two or more storages defined in an entity, only one single storage (usually the materialized_storage) is passed into the query plan builder.

### After State
* Entity configuration:
```
storages:
  - storage: generic_metrics_distributions
  translation_mappers: [x,y,z]
  - storage: generic_metrics_distributions_raw
  is_writable: true
  translation_mappers: [x,y,z]
```
* All storages defined in configuration are passed into the query plan builder. It is the job of the query storage selector to determine which storage to use. As a result, all storages need to be selected explicitly in configuration. The addition of  `SimpleQueryStorageSelector` helps with that. 

### Blast Radius
The configuration of all entities have been changed to use this format.

### Testing
* Entity definitions and pluggable entity tests have been updated to use new format
* A new `SimpleQueryStorageSelector` added.


snuba/datasets/entity.py
```diff
@@ -28,20 +28,20 @@ class Entity(Describable, ABC):
     def __init__(
         self,
         *,
-        storages: Sequence[Storage],
+        storages: Sequence[StorageAndMappers],
```
volokluev, line 31: I think this is not modelling the reality quite accurately. If we think about entities and storages as nodes, each entry in this list is supposed to be an edge

```
Entity -- (mappers, is_writeable) --> Storage
```

But the way this is represented in code is not quite that. We have the `writeable_storage` field which is also connected in the same way but is given special treatment while still existing in the `storages` array. 

I think `StorageAndMappers` should become something more general like `EntityStorageConnection` which contains the mappers, is_writeable, and in the future whatever else we decide to add to that relationship.

The yaml is already specifying things that way, why not make the code do so as well? That way you won't need the special case that you have in the entity builder

https://github.com/getsentry/snuba/pull/3591/files#diff-530c557e763d4aa8baeb0c82eb931da0d8e7206d886d3b266357330ed9506145R145

volokluev: This way you can also remove that reverse lookup I was talking about in the last PR:

https://github.com/getsentry/snuba/pull/3545#discussion_r1054774522

volokluev (commented):
At the time I was writing the initial design for this, I did not fully understand the relationship between entities and storages. I think the configuration format you've laid out makes sense but I think the code organization should reflect it as well. Let me know what you think

snuba/datasets/entity.py
enochtangg: I think the improvement to make configuration and code more consistent with each other makes sense. 

So if I understand correctly, is what are suggesting to remove `writable_storage` as an attribute on `Entity` and change `entity.get_writable_storage()` to get a storage from `entity.storages: Sequence[EntityStorageConnection]`?

volokluev: exactly

snuba/datasets/entities/storage_selectors/selector.py
```diff
@@ -35,33 +35,30 @@ def select_storage(
...
-    ) -> Optional[StorageAndMappers]:
-        return next(
-            (
-                storage
-                for storage in storage_and_mappers
-                if isinstance(storage.storage, WritableTableStorage)
-            ),
-            None,
-        )
+    def get_readable_storage_connection(
```
volokluev, line 42: this is making an assumption (which is not stated or enforced anywhere) that there is only one readable storage and this selector will select the first one by default. If that is an expectation we have then it should be surfaced. The selector functions returning `None` will surface the problem in an obfuscated way. 

If there is no storage which can be selected, this should raise
If the storage to be selected is ambiguous, this should raise

snuba/datasets/pluggable_entity.py
```diff
@@ -88,13 +91,18 @@ def get_query_pipeline_builder(self) -> QueryPipelineBuilder[ClickhouseQueryPlan
...
-        return [storage_and_mapper.storage for storage_and_mapper in self.storages]
+        return [storage_connection.storage for storage_connection in self.storages]
 
-    def get_all_storages_and_mappers(self) -> Sequence[StorageAndMappers]:
+    def get_all_storage_connections(self) -> Sequence[EntityStorageConnection]:
         return self.storages
 
     def get_writable_storage(self) -> Optional[WritableTableStorage]:
-        return self.writable_storage
+        for storage_connection in self.storages:
```
volokluev, line 100: could this use the storage_selector method `get_writable_storage_connection`? Seems to be the same thing

volokluev (commented):
Overall looks good. But we should be clear on the readable storage policy. Are we only allowing one? In which case we should make that clear

snuba/datasets/entities/storage_selectors/selector.py
enochtangg: Right, that assumption is not true. The methods `get_readable_storage_connection()` and `get_writable_storage_connection()` shouldn't be on the `QueryStorageSelector` base class. Currently today, it is completely valid to specify multiple readable and writable storages (although we don't have any entities that do that). As long as the storage selector is contains the logic to choose from them all. 

I have removed those methods, and added the logic to `storage_selectors/errors.py` and  `storage_selectors/sessions.py` individually. Also `EntityStorageConnectionNotFound` is raised. Does that make sense?

snuba/datasets/pluggable_entity.py
enochtangg: See reply above

evanh (commented):
I think this looks good to me. I'll let Volo give the final approval.

snuba/datasets/configuration/entity_builder.py
```diff
@@ -152,6 +152,35 @@ def _build_storage_connections(
...
+            rhs_entity=rhs_key,
+            columns=columns,
+            join_type=join_type,
+            equivalences=equivalences,
+        )
+        relationships[key] = join
+
+    return relationships
+
+
```
volokluev, line 183: Did you change this in your PR or is this merge weirdness?

volokluev (approved):

snuba/datasets/configuration/entity_builder.py
evanh: Definitely merge weirdness, that was from a PR I merged.

enochtangg: Yup, merge weirdness. Updated.

PR #3594, opened 2023-01-13
fix(config) Support join relationships in configuration

evanh:
Add support to the schemas to allow entities to specify join relationships.
Also add a test for the join relationship parser.

Since CDC is a deprecated feature, there was a discussion about whether we should support joins in config, since CDC is the only user of them. But I don't think we want to eliminate support for joins entirely, so I decided to add support for the relationships. Perhaps in the future someone will want joins and we will be ready.

### DIFF
--- a/snuba/datasets/configuration/entity_builder.py
+++ b/snuba/datasets/configuration/entity_builder.py
@@ -20,8 +20,13 @@ from snuba.datasets.entity_subscriptions.validators import EntitySubscriptionVal
 from snuba.datasets.pluggable_entity import PluggableEntity
 from snuba.datasets.storages.factory import get_storage, get_writable_storage
 from snuba.datasets.storages.storage_key import StorageKey
+from snuba.query.data_source.join import ColumnEquivalence, JoinRelationship, JoinType
 from snuba.query.processors.logical import LogicalQueryProcessor
 from snuba.query.validation.validators import QueryValidator
+
+
+class InvalidEntityConfigException(Exception):
+    pass
 
 
 def _build_entity_validators(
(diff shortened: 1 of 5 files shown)


snuba/datasets/configuration/entity_builder.py
```diff
@@ -136,8 +170,9 @@ def build_entity_from_config(file_path: str) -> PluggableEntity:
         required_time_column=config["required_time_column"],
         validators=_build_entity_validators(config["validators"]),
         translation_mappers=_build_entity_translation_mappers(
-            config["translation_mappers"]
+            config.get("translation_mappers", {})
```
evanh, line 173: `translation_mappers` is not a required field in the schema, and in fact some entities (CDC) don't specify them. So I changed this to handle entities that don't specify any mappers.

snuba/datasets/configuration/json_schema.py
```diff
@@ -442,6 +442,50 @@ def registered_class_array_schema(
     },
 }
 
+ENTITY_JOIN_RELATIONSHIPS = {
+    "type": "object",
+    "patternProperties": {
```
evanh, line 447: This allows arbitrary keys with specific values.

snuba/datasets/configuration/json_schema.py
```diff
@@ -442,6 +442,50 @@ def registered_class_array_schema(
...
+            "properties": {
+                "rhs_entity": {
+                    "type": "string",
+                    "description": "The entity key of the rhs entity to join with",
+                },
+                "columns": {
+                    "type": "array",
+                    "items": {
+                        "type": "array",
+                        "prefixItems": [
```
evanh, line 460: `prefixItems` lets you do tuple style validation.

tests/datasets/configuration/entity_join_relationships.yaml
```diff
@@ -0,0 +1,38 @@
+version: v1
```
evanh, line 1: This is actually very close to what the groupassignee schema would actually be.

enochtangg (approved):

PR #3596, opened 2023-01-13
feat(DC): Discover Dataset Config

rahul-kumar-saini:
### Overview
- Discover dataset has a unique entity selection process
- This blocks conversion to config
- This logic shouldn't be generalized and should remain specific to discover, so I've moved it into the parser and freed up the dataset to be converted

### Notes
- Certain metrics recorded may end up having their name changed in datadog since the metrics object is instantiated by doing `metrics = MetricsWrapper(environment.metrics, "api.discover")`


evanh (approved):

volokluev (changes requested):
Don't merge this yet, I'm not convinced that this is a good idea. I need more time to read it

volokluev (approved):
I see this is not making this actively worse but the fact that the parser knows about the discover dataset is very bad. Please file a tech debt ticket to get all dataset specific stuff out of the parser file

PR #3597, opened 2023-01-13
feat(issue-platform): add mappers so discover queries can work with search_issues dataset

barkbarkimashark:
Basically mirrors the DiscoverEntity column mappers from here:
https://github.com/getsentry/snuba/blob/fa65895b0d6d62d5cb7003455ff994864d0339da/snuba/datasets/entities/discover.py#L195

### DIFF
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -91,6 +91,63 @@ translation_mappers:
         from_col_name: "timestamp"
         to_table_name:
         to_col_name: "client_timestamp"
+    - mapper: ColumnToColumn
+      args:
+        from_table_name:
+        from_col_name: "message"
+        to_table_name:
+        to_col_name: "search_title"
+    - mapper: ColumnToColumn
+      args:
+        from_table_name:
+        from_col_name: "title"
+        to_table_name:
+        to_col_name: "search_title"
+    - mapper: ColumnToIPAddress
+      args:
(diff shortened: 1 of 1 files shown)


snuba/datasets/configuration/issues/entities/search_issues.yaml
```diff
@@ -91,6 +91,63 @@ translation_mappers:
...
+      args:
+        from_table_name:
+        from_col_name: "title"
+        to_table_name:
+        to_col_name: "search_title"
+    - mapper: ColumnToIPAddress
+      args:
+        from_table_name:
+        from_col_name: "ip_address"
+        to_function_name: "coalesce"
```
wedamija, line 110: This needs to be coalescing the ipv4 and ipv6 cols right? Not sure if that's possible in this yaml format

wedamija: Nm, just realised `ColumnToIPAddress` is a magic translator

wedamija (approved):

snuba/datasets/configuration/issues/entities/search_issues.yaml
barkbarkimashark: I think its baked into the ColumnToIPAddress mapper:
https://github.com/getsentry/snuba/blob/306097c69963868c9381f806f77afee9104a7b3c/snuba/clickhouse/translators/snuba/mappers.py#L120-L134

evanh (approved):

PR #3601, opened 2023-01-16
test(consumer): Add processing strategy tests

ayirr7:
Another step towards having more complete testing for `ConsumerBuilder`. In this new test, we get the streaming strategy factory from the `ConsumerBuilder`, and then use the resulting processing strategy to submit an `InsertEvent` message. We check for the expected effects from this submit.


tests/consumers/test_consumer_builder.py
```diff
@@ -22,7 +32,7 @@
     kafka_params=KafkaParameters(
         raw_topic=None,
         replacements_topic=None,
-        bootstrap_servers=None,
+        bootstrap_servers=os.environ.get("DEFAULT_BROKERS", "localhost:9092"),
```
lynnagara, line 35: Can we leave this as None? I think the test is failing currently because this is expecting an array and is getting a string. `None` means use the default which is probably ok for this test.

lynnagara (commented):
CI is failing currently due to the bootstrap_servers configuration being passed. Otherwise test looks good to me

tests/consumers/test_consumer_builder.py
ayirr7: Thanks for catching this!

lynnagara (approved):
:shipit:

PR #3602, opened 2023-01-16
feat: Deprecate ParallelCollect strategy in favour of Reduce/RunTaskInThreads

lynnagara:
The parallel collect step is too complex, and offers little more than what can be achieved with Reduce chained to RunTaskInThreads.

As a follow up, Collect/ParallelCollect will be completely removed from Arroyo. The collect strategy is too complicated and we have introduced more intuitive ways to do batching now (via the Batch and Reduce steps).


evanh (approved):
This does seem way cleaner.

PR #3605, opened 2023-01-17
fix(migrations): make outcomes schema match SaaS SNS-1992

dbanda:
This migrations attempts to make the outcomes tables match what's in sass. 
1. It fixes some of the past migrations to display columns in a consistent order with SaaS. Functionally, this make no difference but allows the diff tool to more easily highlight changes.
2. Add back size, bytes received cols to match SaaS schema
3. Add indexes and codecs to match SaaS schema

Also fixes a bug where the validator was overly aggressive when checking for order errors.

### DIFF
--- a/snuba/migrations/validator.py
+++ b/snuba/migrations/validator.py
@@ -68,7 +68,9 @@ def _validate_add_col_or_create_table(
             )
 
 
-def _validate_drop(dist_op: SqlOperation, local_ops: Sequence[SqlOperation]) -> None:
+def _validate_drop_col(
+    dist_op: SqlOperation, local_ops: Sequence[SqlOperation]
+) -> None:
     if isinstance(dist_op, (DropColumn)):
         if any(_conflicts_ops(local_op, dist_op) for local_op in local_ops):
             raise InvalidMigrationOrderError(
@@ -87,7 +89,7 @@ def _validate_order_old(
     """
     if local_first:
         for dist_op in dist_ops:
-            _validate_drop(dist_op, local_ops)
+            _validate_drop_col(dist_op, local_ops)
     else:
         for local_op in local_ops:
             _validate_add_col_or_create_table(local_op, dist_ops)
@@ -112,7 +114,7 @@ def _validate_order_new(
                     _validate_add_col_or_create_table(op, dist_ops_before)
             elif isinstance(op, DropColumn):
                 if op.target == OperationTarget.DISTRIBUTED:
-                    _validate_drop(op, local_ops_before)
+                    _validate_drop_col(op, local_ops_before)
 
 
 def validate_migration_order(migration: ClickhouseNodeMigration) -> None:
--- a/snuba/migrations/validator.py
+++ b/snuba/migrations/validator.py
@@ -111,7 +111,8 @@ def _validate_order_new(
                 if op.target == OperationTarget.LOCAL:
                     _validate_add_col_or_create_table(op, dist_ops_before)
             elif isinstance(op, DropColumn):
-                _validate_drop(op, local_ops_before)
+                if op.target == OperationTarget.DISTRIBUTED:
+                    _validate_drop(op, local_ops_before)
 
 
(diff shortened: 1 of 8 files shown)


snuba/migrations/validator.py
```diff
@@ -111,7 +111,8 @@ def _validate_order_new(
                 if op.target == OperationTarget.LOCAL:
                     _validate_add_col_or_create_table(op, dist_ops_before)
             elif isinstance(op, DropColumn):
-                _validate_drop(op, local_ops_before)
+                if op.target == OperationTarget.DISTRIBUTED:
+                    _validate_drop(op, local_ops_before)
```
nikhars, line 115: Let's rename the method to make it clear that it is validating drop column ordering. For a moment I thought it was validating drop tables.
```suggestion
                    _validate_drop_col(op, local_ops_before)
```

snuba/snuba_migrations/outcomes/0006_outcomes_add_size_col.py
```diff
@@ -0,0 +1,69 @@
...
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Adds back (again) the size, bytes_received columns to match schema in SaaS
+    """
+
+    blocking = True
```
nikhars, line 13: Shouldn't this be a non-blocking migration since it is just adding columns? Same comment about the blocking = True in the other migration

snuba/snuba_migrations/outcomes/0007_outcomes_add_indexes_and_codecs.py
```diff
@@ -0,0 +1,76 @@
...
+                    UUID(
+                        Modifiers(
+                            nullable=True,
+                            codecs=["LZ4HC(0)"],
+                            ttl="timestamp + toIntervalDay(30)",
+                        )
+                    ),
+                ),
+                target=operations.OperationTarget.LOCAL,
+            ),
```
nikhars, line 33: I would recommend you move out the modify column to its own migration. The migration is applying TTL to a column (which is a heavy operation) and also adding indexes (which is a lighter operation). Even though the end state would be the same whether you write 1 migration or 2, you separate out DDL changes by the type of operation being performed.

snuba/migrations/validator.py
dbanda: fixed

snuba/snuba_migrations/outcomes/0006_outcomes_add_size_col.py
dbanda: fixed

snuba/snuba_migrations/outcomes/0007_outcomes_add_indexes_and_codecs.py
dbanda: moved to 0007_outcomes_add_event_id_ttl_codec

snuba/snuba_migrations/outcomes/0006_outcomes_add_size_col.py
```diff
@@ -0,0 +1,69 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Adds back (again) the size, bytes_received columns to match schema in SaaS
```
MeredithAnya, line 10: Are `size` and `bytes_received` actually used in prod? Did we decide that it was not worth it to remove these columns from saas outcomes if they are in fact unused?

dbanda: I haven't got a confirmation on that. The plan is to get both to a common state, then afterwards we can decide what to drop.

MeredithAnya (approved):

PR #3606, opened 2023-01-17
ref(migrations): Simplfy policies and status checks

MeredithAnya:
**context:**
Policies are no longer going to be defined in settings, but rather by the roles added https://github.com/getsentry/snuba/pull/3522. Actually using the roles to get the policies is done in https://github.com/getsentry/snuba/pull/3618 but I wanted to simplify the code a bit in order to make that change more manageable. 

**changes:**
* removal of the `PolicyChecker` class - it made more sense to get the group policies in the API endpoint since we'll need the user to do so. 
* contain all the checking logic within `run_migration_checks_and_policies` - I think ultimately this makes it simpler to read and understand

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -16,7 +16,7 @@ from snuba.admin.audit_log.action import AuditLogAction
 from snuba.admin.audit_log.base import AuditLog
 from snuba.admin.auth import USER_HEADER_KEY, UnauthorizedException, authorize_request
 from snuba.admin.clickhouse.common import InvalidCustomQuery
-from snuba.admin.clickhouse.migration_checks import run_migration_checks_for_groups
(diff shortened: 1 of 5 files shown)


snuba/admin/views.py
```diff
@@ -87,18 +90,17 @@ def health() -> Response:
 
 @application.route("/migrations/groups")
 def migrations_groups() -> Response:
+    group_policies = get_migration_group_polices()
+    allowed_groups = group_policies.keys()
+
     res: List[Mapping[str, str | Sequence[Mapping[str, str | bool]]]] = []
-    allowed_groups: Sequence[MigrationGroup] = [
-        group
-        for group in get_active_migration_groups()
```
MeredithAnya, line 93: actually we are going to need the `get_active_migration_groups()` because that's what the runner uses so need to add this back in

dbanda (approved):

PR #3607, opened 2023-01-17
feat(profiling): Add profile id to transactions table

Zylphrex:
This adds a new profile id column to the transactions table as per https://github.com/getsentry/rfcs/blob/main/text/0047-introduce-profile-context.md

### DIFF
--- a/snuba/snuba_migrations/transactions/0018_transactions_add_profile_id.py
+++ b/snuba/snuba_migrations/transactions/0018_transactions_add_profile_id.py
@@ -1,0 +1,39 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import UUID, Column
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/transactions/0018_transactions_add_profile_id.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.TRANSACTIONS,
+                table_name=table_name,
+                column=Column("profile_id", UUID()),
```
evanh, line 17: Shouldn't this be nullable?

Zylphrex: good catch, I added it in b1b4c72

evanh (approved):

PR #3609, opened 2023-01-17
feat(replays): Add logging for exceptions which violate integer sizing

cmanallen:
closes: https://github.com/getsentry/replay-backend/issues/254

### DIFF
--- a/snuba/datasets/processors/replays_processor.py
+++ b/snuba/datasets/processors/replays_processor.py
@@ -201,11 +201,8 @@ class ReplaysProcessor(DatasetMessageProcessor):
             self._process_event_hash(processed, replay_event)
             self._process_contexts(processed, replay_event)
             return InsertBatch([processed], None)
-        except LoggedException as exc:
-            logger.error(
-                "A replay schema constraint was not observed.",
-                exc_info=exc,
-            )
+        except LoggedException:
+            logger.error("A replay schema constraint was not observed.", exc_info=True)
             return None
         except Exception:
             metrics.increment("consumer_error")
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/replays_processor.py
```diff
@@ -196,6 +200,9 @@ def process_message(
             self._process_event_hash(processed, replay_event)
             self._process_contexts(processed, replay_event)
             return InsertBatch([processed], None)
+        except LoggedException:
+            logger.error("A replay schema constraint was not observed.", exc_info=True)
```
nikhars, line 204: Do you mind changing this to logger.warning instead? Sentry alerts for SnS counts number of errors and I believe these would still trigger our alerts.

cmanallen: @nikhars 👍

fpacifici:
I would advise against this approach (unless the goal is to mute the alert associated with this issue in the short term). 

This consumer has a Dead Letter Queue associated. This means it should be enough for the consumer to simply raise an exception for whatever cannot be processed (as invalid) and the DLQ will take care of that and produce consistent metrics about that. Making a choice in the processor between what should be dropped immediately and what should be taken care by the DLQ means adding complexity in the consumer as there are cases where the message is dropped immediately. 
Is there any specific reason you do not want these messages to end in the DLQ?

If the reason for this is silencing the alerts, then the issue is in the snuba consumer infra. Not in the replays processor as the snuba consumer infra logs the exception even if it rethrows it as an InvalidMessage exception 
https://github.com/getsentry/snuba/blob/master/snuba/consumers/consumer.py#L574

PR #3611, opened 2023-01-17
migrations: make querylog match SaaS

dbanda:
This adds migrations to make the `querylog_local` tables match the schema in SaaS. 

There are two migrations steps added:
1. Adds codecs to columns and updates the table settings to match SaaS. To accomplish this, we also add a SqlOperation to modify table settings.
2. Adds extra columns to the sorting key to match SaaS. This has to be a code migration because we cant use Sql `ALTER` statements to modify the sorting key. Instead, we have to create a new table and copy over the data then drop the old table. Inside this migration, we also add the codec to the primary key as this has to be done in code too.


snuba/migrations/operations.py
```diff
@@ -315,6 +315,48 @@ def optional_ttl_clause(self) -> str:
...
+    def __init__(
+        self,
+        storage_set: StorageSetKey,
+        table_name: str,
+        settings: Sequence[str],
+        target: OperationTarget = OperationTarget.UNSET,
+    ):
+        super().__init__(storage_set, target=target)
+        self.__table_name = table_name
+        self.__settings = settings
```
dbanda, line 353: Tests for these are in another PR

snuba/snuba_migrations/querylog/0005_add_codec_update_settings.py
```diff
@@ -0,0 +1,112 @@
...
+            ),
+            operations.ModifyTableSettings(
+                StorageSetKey.QUERYLOG,
+                table_name,
+                {"min_bytes_for_wide_part": "10000000", "ttl_only_drop_parts": 1},
+                target=operations.OperationTarget.LOCAL,
+            ),
+        ]
+
+    def backwards_ops(self) -> Sequence[operations.SqlOperation]:
```
MeredithAnya, line 68: From what I recall, to reset it to the default you'd need `CODEC(Default)` as was commented in https://github.com/ClickHouse/ClickHouse/issues/6227 which I also don't think was possible for older versions. Will these backwards ops actually do anything?

dbanda: that's correct. but `CODEC(Default)` doesn't work on clickhouse 20 😢

snuba/snuba_migrations/querylog/0005_add_codec_update_settings.py
```diff
@@ -0,0 +1,112 @@
...
+        return [
```
nikhars, line 69: Not that it matters in this specific case, but usually you want to write the sequence of backward ops in the reverse order of forward ops. The reason being that if you are adding a column and then lets assume adding an index on it in the forward ops, if you don't do the reverse order, i am not sure whether clickhouse would let you remove a column on which an index is built.

Do you know if our migration tests apply both forward and reverse to catch these sort of things?

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
```diff
@@ -0,0 +1,156 @@
...
+    return
+
+
+def forwards(logger: logging.Logger) -> None:
+    """
+    This migration is a bit complicated because we need to change the sorting key.
+    We can't do it inplace so we need to create a new table with the modified keys,
+    copy the data over, then drop the old table.
+    """
+    cluster = get_cluster(StorageSetKey.TRANSACTIONS)
```
nikhars, line 97: Why is the cluster fetched with `StorageSetKey.TRANSACTIONS`? Is it a typo?

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
```diff
@@ -0,0 +1,156 @@
...
+        update_querylog_table(connection, database)
+
+
+def backwards(logger: logging.Logger) -> None:
+    """
+    This method cleans up the temporary tables used by the forwards methods and
+    returns us to the original state if the forwards method has failed somewhere
+    in the middle. Otherwise it's a no-op.
+    """
+    cluster = get_cluster(StorageSetKey.TRANSACTIONS)
```
nikhars, line 111: Same comment as above

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
```diff
@@ -0,0 +1,156 @@
...
+
+    clickhouse.execute(f"RENAME TABLE {TABLE_NAME} TO {TABLE_NAME_OLD};")
+
+    clickhouse.execute(f"RENAME TABLE {TABLE_NAME_NEW} TO {TABLE_NAME};")
+
+    # Ensure each table has the same number of rows before deleting the old one
+    assert (
+        clickhouse.execute(f"SELECT COUNT() FROM {TABLE_NAME};").results
+        == clickhouse.execute(f"SELECT COUNT() FROM {TABLE_NAME_OLD};").results
+    )
```
nikhars, line 85: Is it possible for newer writes to happen while this loop is running https://github.com/getsentry/snuba/pull/3611/files#diff-8832387c64f114e4caa833a673d0d8b6cfb00cb5c12f77a40df3cf00f741ba08R58-R75? If yes, this assert can fail.

dbanda: Yes that is correct. Since this is a data migration you would have take the table inserts offline first. I'm not sure if there's an easy way around that.

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
dbanda: fixed

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
dbanda: fixed

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
nikhars: Got it. I would suggest you remove this assert to remove the ambiguity.

snuba/snuba_migrations/querylog/0006_sorting_key_change.py
```diff
@@ -0,0 +1,156 @@
...
+    batch_size = 100000
+    batch_count = math.ceil(row_count / batch_size)
+
+    orderby = "toStartOfDay(finish_ts), project_id, event_id"
+
+    for i in range(batch_count):
+        skip = batch_size * i
+        clickhouse.execute(
+            f"""
+            INSERT INTO {TABLE_NAME_NEW}
```
nikhars, line 69: Is there a specific reason you did not use https://github.com/getsentry/snuba/blob/master/snuba/migrations/operations.py#L416?

If you cannot re-use it as is, I would recommend building a different api to achieve what you are doing here. This use case of inserting into another using select from existing can re-surface for different reasons in the future. We should try to make an api out of it. That api could perform the task using batches or partitions etc.

dbanda: It was missing some parameters for doing batching. I fixed it up now.

nikhars (approved):

PR #3615, opened 2023-01-19
fix(outcomes): Add TTL to outcomes dataset

nikhars:
Since there is no TTL on the outcomes dataset, the storage size can grow unbounded. This PR adds a TTL of 30 days for the raw tables and 90 days for the aggregate tables.

### DIFF
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -167,6 +167,7 @@ class OutcomesLoader(DirectoryLoader):
             "0002_outcomes_remove_size_and_bytes",
             "0003_outcomes_add_category_and_quantity",
             "0004_outcomes_matview_additions",
+            "0005_outcomes_ttl",
         ]
 
 
--- a/snuba/snuba_migrations/outcomes/0005_outcomes_ttl.py
+++ b/snuba/snuba_migrations/outcomes/0005_outcomes_ttl.py
@@ -1,0 +1,34 @@
+from typing import Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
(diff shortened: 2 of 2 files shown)


volokluev (approved):

dbanda:
Is this change also being added to SaaS? I ask because I have a PR #3605 that does something similar by adding a TTL to a columns. Dont know if that will be repeating work.

nikhars:
> Is this change also being added to SaaS? I ask because I have a PR https://github.com/getsentry/snuba/pull/3605 that does something similar by adding a TTL to a columns. Dont know if that will be repeating work.

Yes, I will apply the TTL settings to SaaS. In your PR, the TTL change I see is
`Local op: ALTER TABLE outcomes_raw_local MODIFY COLUMN event_id Nullable(UUID) CODEC (LZ4HC(0)) TTL timestamp + toIntervalDay(30);`
Is that TTL change not on the SaaS environment yet? I thought the reconciliation was getting code to match SaaS. Could you provide some additional context.

dbanda:
> > Is this change also being added to SaaS? I ask because I have a PR #3605 that does something similar by adding a TTL to a columns. Dont know if that will be repeating work.
> 
> Yes, I will apply the TTL settings to SaaS. In your PR, the TTL change I see is `Local op: ALTER TABLE outcomes_raw_local MODIFY COLUMN event_id Nullable(UUID) CODEC (LZ4HC(0)) TTL timestamp + toIntervalDay(30);` Is that TTL change not on the SaaS environment yet? I thought the reconciliation was getting code to match SaaS. Could you provide some additional context.

yes the change is in SaaS. but to get them to match I will also have to add it to what we have here.

PR #3618, opened 2023-01-19
ref(migrations): use roles to get group policies

MeredithAnya:
Can be used after https://github.com/getsentry/snuba/pull/3606 is merged (still will have to rebase and update some code/tests)

### DIFF
--- a/snuba/migrations/policies.py
+++ b/snuba/migrations/policies.py
@@ -83,3 +83,24 @@ class AllMigrationsPolicy(MigrationPolicy):
 
     def can_reverse(self, migration_key: MigrationKey) -> bool:
         return True
+
+
+MIGRATION_WEIGHTS = {
+    "AllMigrationsPolicy": 3,
+    "NonBlockingMigrationsPolicy": 2,
+    "NoMigrationsPolicy": 1,
+}
+
+import functools
+
+
+def compare_policies(x: str, y: str) -> int:
+    return MIGRATION_WEIGHTS[y] - MIGRATION_WEIGHTS[x]
+
+
+from typing import Sequence
+
+
+def max_policy(policies: Sequence[str]) -> str:
+    sorted_policies = sorted(policies, key=functools.cmp_to_key(compare_policies))
+    return sorted_policies[0]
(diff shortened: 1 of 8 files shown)


snuba/migrations/policies.py
```diff
@@ -83,3 +83,24 @@ def can_run(self, migration_key: MigrationKey) -> bool:
 
     def can_reverse(self, migration_key: MigrationKey) -> bool:
         return True
+
+
+MIGRATION_WEIGHTS = {
```
MeredithAnya, line 88: Is there a better way to do this? Just wanted an easy way to compare policies

dbanda: This comparison strategy seems a little bit fragile. If we add more policies it seems like it may become unwieldy keeping a total order of policies. Also, I believe you would have to ensure the comparator is transitive for python to work properly and this can be hard to figure out for some permissions. 

A better way to do this would be dump all the policies into a list for each user, then the checker enumerates through and allows if it finds one that has the permissions. Its a little slower, but given that the size of these lists should be very small, the perf impact is negligible.

snuba/migrations/policies.py
```diff
@@ -83,3 +83,24 @@ def can_run(self, migration_key: MigrationKey) -> bool:
...
+    "AllMigrationsPolicy": 3,
+    "NonBlockingMigrationsPolicy": 2,
+    "NoMigrationsPolicy": 1,
+}
+
+import functools
```
evanh, line 94: Why are these imports done here instead of at the top of the file?

snuba/migrations/policies.py
evanh: I agree with Dalitso, this seems really complicated.

```
# Sorted by importance
MIGRATION_WEIGHTS = ["All", "Non", "No"]

def max_policy(policies: Sequence[str]) -> str:
    for p in MIGRATION_WEIGHTS:
        if p in policies:
            return p
```

MeredithAnya: update this re dalitso's suggestion, and now just check all the policies for a group simplifying this so that there is no need for weights

dbanda (approved):

PR #3622, opened 2023-01-20
ref: Remove runtime config override for stale-threshold-sec

lynnagara:
--stale-threshold-sec should be passed via the CLI, remove the runtime override for this value


fpacifici (approved):
Do we have it set to something meaningful in all cases ?

lynnagara:
> Do we have it set to something meaningful in all cases ?

There is a linked ops PR to pass all the values currently set via runtime config via CLI instead. Will merge this only once that is applied.

PR #3625, opened 2023-01-20
feat(api-abuse): Add triggered rate limiter name to Querylog stats

enochtangg:
## Overview
Currently, the snuba querylog contains `status` which specifies whether a query was rate-limited or not. However, it does not specify which rate limiter was applied. This PR is responsible for adding that into `stats`.

## After State
Querylog contains a new key `triggered_rate_limiter` in `stats` when a rate limiter is applied to a query.

## Testing
Tested locally to ensure messages produced to querylog topics are correct and contain accurate fields.

### DIFF
--- a/snuba/state/rate_limit.py
+++ b/snuba/state/rate_limit.py
@@ -107,6 +107,7 @@ def _get_bucket_key(prefix: str, bucket: str, shard_id: int) -> str:
 def rate_limit(
     rate_limit_params: RateLimitParameters,
 ) -> Iterator[Optional[RateLimitStats]]:
+    print("entered rate limit context manager")
     """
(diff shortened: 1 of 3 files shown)


snuba/state/rate_limit.py
```diff
@@ -107,6 +107,7 @@ def _get_bucket_key(prefix: str, bucket: str, shard_id: int) -> str:
 def rate_limit(
     rate_limit_params: RateLimitParameters,
 ) -> Iterator[Optional[RateLimitStats]]:
+    print("entered rate limit context manager")
```
evanh, line 110: 💀

enochtangg: oops, updated.

evanh (approved):

PR #3626, opened 2023-01-20
fix(consumer): Separate out the shared producers in consumer

ayirr7:
Bringing back https://github.com/getsentry/snuba/pull/3305

This time there are tests for `Consumer_Builder.` The test itself was also slightly modified to match this refactor and ensure valid topic names are being used.


lynnagara:
Is this implementation different from https://github.com/getsentry/snuba/pull/3305? What was the original reason for reverting that again?

snuba/consumers/consumer_builder.py
```diff
@@ -121,11 +113,23 @@ def __init__(
...
+            self.replacements_producer = Producer(
+                build_kafka_producer_configuration(
+                    SnubaTopic(self.replacements_topic.name),
+                    bootstrap_servers=kafka_params.bootstrap_servers,
+                    override_params={
+                        "partitioner": "consistent",
+                        "message.max.bytes": 50000000,  # 50MB, default is 1MB)
+                    },
+                )
+            )
```
lynnagara, line 128: nit: I think it's a bit clearer if you define `self.replacements_producer = None` in an else block starting on line 129. So that it is being defined in the same place in both cases. Same for the commit log producer.

lynnagara (approved):

ayirr7:
> Is this implementation different from #3305? What was the original reason for reverting that again?

Same conceptual implementation, with a fix. The issue with that PR was that the replacements and commit log producers were not defined for all scenarios. They weren't defined in the case that user passes in an override for replacements topic, or commit log topic. This PR fixes that by ensuring they are defined (the tests help verify this).

PR #3627, opened 2023-01-20
fixcodecov): add settings for codecov

volokluev:
Tested for validity with

```
cat codecov.yml | curl --data-binary @- https://codecov.io/validate
```


evanh (approved):
Something has to be better than nothing, right?

volokluev:
at least we'll know what we need to change next time

PR #3628, opened 2023-01-20
feat(api-abuse): Add timing metrics for table_concurrent

enochtangg:
## Overview
Currently, the `table_concurrent` metric is a gauge in Datadog. As a result, the agent only submits the last reported number of the metric in a given time interval. Since we want a more accurate representation of the maximum concurrent queries that hit our tables, we should change this to a `metric.timing`.

## After State
A new `table_concurrent_` metric is added to get a more accurate maximum overt time.

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -313,6 +313,16 @@ def _record_rate_limit_metrics(
                 else "default",
             },
         )
+        metrics.timing(
+            name="table_concurrent_",
+            value=table_rate_limit_stats.concurrent,
+            tags={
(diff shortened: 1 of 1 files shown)


snuba/web/db_query.py
```diff
@@ -313,6 +313,16 @@ def _record_rate_limit_metrics(
                 else "default",
             },
         )
+        metrics.timing(
+            name="table_concurrent_",
```
volokluev, line 317: name it something like `table_concurrent_hist`

volokluev (approved):

PR #3629, opened 2023-01-20
fix(codecov) Change codecov notifier to send after at least 2 builds

enochtangg:
Since the codecov.yml was added, Codecov does not notify (comment report on PR) anymore. This PR reduces `after_n_builds` to `2` so it will post after only 2 of our test CI steps.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,7 +6,7 @@ codecov:
   # until all of these are done, we don't want a faulty
   # codecov report
   notify:
-      after_n_builds: 3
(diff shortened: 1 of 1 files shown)


volokluev:
> This PR reduces after_n_builds to 2 so it will post after only 2 of our test CI steps.

It reduces it to post after 2 CI steps which upload to codecov, that's different from 2 CI steps period

enochtangg:
didn't work

PR #3630, opened 2023-01-20
ref(DC): Add FixedString field to configuration

volokluev:
The discover Entity contains `FixedString` fields which do not exist in the configuration grammar right now

* update the grammar
* update the parser

**Blast Radius**
No existing entities should be affected

**Feedback Desired**
Is there a better place to put the test that I don't know about?

### DIFF
--- a/tests/datasets/configuration/test_entity_loader.py
+++ b/tests/datasets/configuration/test_entity_loader.py
@@ -26,6 +26,11 @@ def get_object_in_list_by_class(object_list: Any, object_class: Any) -> Any:
 
 
 class TestEntityConfiguration(ConfigurationTest):
+    def test_entity_loader_fixed_string(self) -> None:
+        build_entity_from_config(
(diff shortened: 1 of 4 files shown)


tests/datasets/configuration/test_entity_loader.py
```diff
@@ -26,6 +26,15 @@ def get_object_in_list_by_class(object_list: Any, object_class: Any) -> Any:
 
 
 class TestEntityConfiguration(ConfigurationTest):
+    def test_entity_loader_fixed_string(self) -> None:
+        entity = build_entity_from_config(
+            "tests/datasets/configuration/entity_with_fixed_string.yaml"
+        )
+        columns = list(entity.get_data_model())
+        assert len(columns) == 3
+        assert columns[0].type.length == 420  # type: ignore
```
evanh, line 35: 😄

evanh (approved):

PR #3633, opened 2023-01-20
feat: Tag invalid messages for DLQ with `invalid_message`: `true`

lynnagara:
This allows us to better route Sentry errors to relevant teams based on tags.

The storage name is also tagged on the Snuba consumer so we can filter by consumers for monitoring.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -24,6 +24,7 @@ from typing import (
 )
 
 import rapidjson
+import sentry_sdk
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.backends.kafka.commit import CommitCodec
 from arroyo.commit import Commit as CommitLogCommit
@@ -566,6 +567,7 @@ def process_message(
             ),
         )
     except Exception as err:
+        sentry_sdk.set_tag("invalid_message", "true")
         logger.error(err, exc_info=True)
         raise InvalidMessages(
             [__invalid_kafka_message(message.value, consumer_group, err)]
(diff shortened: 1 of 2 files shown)


snuba/consumers/consumer.py
```diff
@@ -566,7 +567,8 @@ def process_message(
             ),
         )
     except Exception as err:
-        logger.error(err, exc_info=True)
+        sentry_sdk.set_tag("invalid_message", "true")
```
fpacifici, line 570: Isn't this going to be set to true for any further exception this process sends to sentry? 
In that case it would be wrong.

lynnagara: Oops, thanks

snuba/consumers/consumer.py
```diff
@@ -566,7 +567,9 @@ def process_message(
             ),
         )
     except Exception as err:
-        logger.error(err, exc_info=True)
+        with sentry_sdk.push_scope() as scope:
```
fpacifici, line 570: This has another issue though. 
the tag will be set to the warning message, not to the error message that is raised as InvalidMessage

lynnagara: Indented the raise InvalidMessage so it is in this block too.

fpacifici (approved):

snuba/consumers/consumer.py
```diff
@@ -566,7 +567,9 @@ def process_message(
...
+            scope.set_tag("invalid_message", "true")
+            logger.warning(err, exc_info=True)
```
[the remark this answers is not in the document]
lynnagara: Verified this locally. We get both the original warning and the InvalidMessage exception if there is no DLQ. Both have the stack trace.

PR #3635, opened 2023-01-20
build: Arroyo 2.5.0

lynnagara:


requirements.txt
```diff
@@ -24,7 +24,7 @@ python-dateutil==2.8.2
 python-rapidjson==1.8
 pytz==2022.2.1
 redis==4.3.4
-sentry-arroyo==2.4.0
+sentry-arroyo[avro,json,msgpack]==2.5.0
```
lynnagara, line 27: Fix in https://github.com/getsentry/arroyo/pull/176 so we only need json in future.

evanh (approved):

PR #3636, opened 2023-01-20
refDC): Extract session translation mappers

volokluev:
The sessions entity has a lot of complicated translation mappers which heavily make use of ColumnToFunction. We don't support ColumnToFunction because it would make a bad user experience and properly expressing it in config takes more thought than we care to give it right now. Therefore we are extracting all these complex mappers into their own file

**Blast Radius**

Only the session entity is affected which is deprecated. It is still used by single tenant and self hosted. There should be no code change here, this is refactoring


snuba/datasets/entities/sessions.py
```diff
@@ -2,9 +2,22 @@
 
 from snuba import environment
 from snuba.clickhouse.columns import ColumnSet, DateTime, UInt
-from snuba.clickhouse.translators.snuba.mappers import (
-    ColumnToCurriedFunction,
-    ColumnToFunction,
+
+# -------------
```
evanh, line 6: Why this comment?

evanh (approved):

PR #3639, opened 2023-01-23
ref(admin): Add in fetching GCP roles

MeredithAnya:
**context:**:
A little bit ago, I added roles in the snuba admin tooling: https://github.com/getsentry/snuba/pull/3522. Right now, set all those roles on a user:https://github.com/getsentry/snuba/blob/154cd213f3b9fdff41b2f7b98e311b4bd6e46b18/snuba/admin/auth.py#L39
We want to be able to assign different roles to different users, and we've decided we can use GCP roles to do that.

My hope is to create roles in GCP with identical names as the ones listed in `DEFAULT_ROLES` (these are not the end all be all for which roles, but just wanted to have a couple to start). And then by grabbing the roles from a user that is authenticated via google's auth, we can assign the admin roles appropriately. 

e.g user `meredith@sentry.io` has a custom GCP role `roles/MigrationsLimitedExecutor`, from that I should be able to assign the admin role with the name `MigrationsLimitedExecutor` on the `AdminUser` for `meredith@sentry.io`

Since the cloud resource manager API would be fetching all the role bindings for an entire project, we could just store that in memory (just the clickhouse migration ones) instead of fetching every time we check a user. We could decide when to refresh so that bindings are up to date.

### DIFF
--- a/snuba/admin/gcp_roles.py
+++ b/snuba/admin/gcp_roles.py
@@ -6,8 +6,11 @@ import googleapiclient.discovery
 # not sure if both of these are needed...
 from google.oauth2 import service_account
 
-GCP_PROJECT_ID = 123455789  # todo move to a setting
+from snuba.settings import ADMIN_GCP_PROJECT_ID
+
 GCP_USER_ROLES: MutableMapping[str, List[str]] = defaultdict(list)
+
+ROLE_PREFIX = "roles/clickhouseMigrations."
 
 
 def set_google_roles() -> None:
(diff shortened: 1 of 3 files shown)


snuba/admin/gcp_roles.py
```diff
@@ -0,0 +1,43 @@
...
+
+def set_google_roles() -> None:
+    """
+    Using the GCP cloud resource manager API, request the list
+    of bindings (roles to users) and use those mappings to populate
+    GCP_USER_ROLES for the project specified by the GCP_PROJECT_ID.
+
+    Every call to this function will refresh the GCP_USER_ROLES.
+    """
+    # is this needed? can we get creds a diff way?
```
dbanda, line 21: I dont think so. we will needs a service account one way or the other.

MeredithAnya:
outdated

PR #3640, opened 2023-01-23
feat(profiling): Process profile context on transactions

Zylphrex:
Process the profile context on transactions and extract the profile id.

### DIFF
--- a/snuba/datasets/storages/transactions_common.py
+++ b/snuba/datasets/storages/transactions_common.py
@@ -102,6 +102,7 @@ columns = ColumnSet(
         ("timestamp", DateTime(Modifiers(readonly=True))),
         ("group_ids", Array(UInt(64))),
         ("app_start_type", String()),
+        ("profile_id", UUID(Modifiers(nullable=True))),
     ]
 )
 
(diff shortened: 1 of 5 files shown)


nikhars (approved):

phacops (approved):

snuba/datasets/storages/transactions_common.py
```diff
@@ -102,6 +102,7 @@
         ("timestamp", DateTime(Modifiers(readonly=True))),
         ("group_ids", Array(UInt(64))),
         ("app_start_type", String()),
+        ("profile_id", UUID(Modifiers(nullable=True))),
```
Zylphrex, line 105: @nikhars is this correct? I was under the impression the yaml schema was all that was needed. Now we have to specify it in yaml as well as in python? Am I missing it anywhere else?

evanh: The YAML schema is all that is needed for the change to function. I suspect this change is to ensure we keep the schemas in lockstep until we can remove the Python ones.

evanh (approved):

PR #3644, opened 2023-01-24
feat(metrics): Add timeseries_id column to Metrics dataset

ayirr7:
As part of our release health cluster improvements, we are adding the `timeseries_id` column to the Metrics dataset (https://getsentry.atlassian.net/browse/SNS-1972). By doing so, we will be able to create a sharding key based on timeseries, shard data by timeseries, and scale release health cluster accordingly.


snuba/snuba_migrations/metrics/0035_metrics_sets_timeseries_id.py
```diff
@@ -0,0 +1,48 @@
...
+        return self.__forward_migrations("metrics_sets_v2_local")
+
+    def backwards_local(self) -> Sequence[operations.SqlOperation]:
+        return self.__backward_migrations("metrics_sets_v2_local")
+
+    def forwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return self.__forward_migrations("metrics_sets_v2_dist")
+
+    def backwards_dist(self) -> Sequence[operations.SqlOperation]:
+        return self.__backward_migrations("metrics_sets_v2_dist")
```
fpacifici, line 48: I don't think this is the right table. 
(same for the tables below)

When you write on metrics, you write on a raw table and then a number of materialized views fill in the aggregate tables: `metrics_sets`, `metrics_distributions`, `metrics_counters`. 
Sharding happens on the raw table, not on the aggregate tables as the aggregate tables are colocated with the raw table. 

So this field should be added to the raw polymorphic table instead than the aggregate ones. 

This is the right table 
https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/metrics/0020_polymorphic_buckets_table.py

fpacifici (changes requested):
I think this field is being added to the wrong tables. Please see comment inline

snuba/snuba_migrations/metrics/0035_metrics_sets_timeseries_id.py
ayirr7: Thank you for pointing this out, I'll look into it and make the change

fpacifici (approved):

snuba/snuba_migrations/metrics/0035_metrics_raw_timeseries_id.py
```diff
@@ -0,0 +1,48 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigrationLegacy):
```
nikhars, line 8: Lets not use the legacy migration. It is meant to be [deprecated](https://github.com/getsentry/snuba/blob/master/snuba/migrations/migration.py#L186-L189). Use the newer class instead https://github.com/getsentry/snuba/blob/master/snuba/migrations/migration.py#L97.

onewland: I wonder if the legacy migration could log a warning or something (don't think it should block migrating).

Though I guess being named `Legacy` is already nudging you to not use it

ayirr7: Thanks for pointing this out, I hadn't noticed this.

PR #3645, opened 2023-01-24
feat(DC): create events entity, expand testing, fix transactions yaml

volokluev:
Events entity generated with the [entity conversion script](https://github.com/getsentry/snuba/tree/python_2_yaml_entity_converter)

* Added test functionality to compare translation mappers
* Added test to verify join relationships
* added test to verify subscription validators
* The transactions entity was actually out of sync so I fixed it

### DIFF
--- a/snuba/datasets/configuration/events/entities/events.yaml
+++ b/snuba/datasets/configuration/events/entities/events.yaml
@@ -411,8 +411,7 @@ subscription_validators:
         - groupby
         - having
         - orderby
-      required_time_column:
-        - timestampp
+      required_time_column: timestamp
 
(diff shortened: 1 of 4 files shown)


enochtangg (approved):
Just curious, how did you find out that transactions entity had an extra translation mapper? Was it by inspection or something more automatic?

volokluev:
> how did you find out that transactions entity had an extra translation mapper?

The test I wrote surfaced the issue that the mappers did not match. I had to use my eyeballs to figure out which one was the duplicate one though

evanh (approved):

PR #3646, opened 2023-01-24
ref(config): Create Metrics Org Counters Entity yaml file

enochtangg:
This PR is responsible for creating the metrics org_counters entity yaml file using the [python_2_yaml conversion script](https://github.com/getsentry/snuba/blob/python_2_yaml_entity_converter/scripts/pyentity_2_yaml.py).


evanh (approved):
Looks good, but please test again once https://github.com/getsentry/snuba/pull/3645 is merged.

PR #3647, opened 2023-01-25
remove yarn ppa and instead install yarn off npm

dbanda:
There a bug where our yarn key has expired and is causing self-hosted tests to fail because the docker cache is not invalidated to refresh. This will make it so that we install yarn directly via npm and don't rely on caching the yarn key to fetch from ppa.


asottile-sentry:
here's a slightly less involved patch -- just prevents doing a stale, cached fetch later -- https://github.com/getsentry/snuba/pull/3648

volokluev:
@dbanda is this PR still necessary?

PR #3648, opened 2023-01-25
ref: remove yarn apt list after installation

asottile-sentry:


hubertdeng123 (approved):
Seems like this key expiring is breaking self-hosted. Users have reported a dirty fix that is similar to the changes in this PR
https://github.com/getsentry/self-hosted/issues/1932

dbanda (approved):

PR #3650, opened 2023-01-25
ref(DC): Remove Generic Metrics Python

rahul-kumar-saini:
### Overview
- Removes the Python version of Generic Metrics Dataset, Entity, and Storages


onewland (approved):
🔥

PR #3653, opened 2023-01-25
cleanup(retention): remove project ID-based retention overrides

onewland:
retention_days are set on every dataset upstream by using org-level info, this was a hack and nobody could remember why it existed.


nikhars (approved):
Please cleanup the RETENTION_OVERRIDE settings from the prod config as well.

PR #3656, opened 2023-01-26
ref(migrations): allow reverting completed non-blocking migrations

MeredithAnya:
**context**
Currently our non blocking migration policy allows users to revert migrations that are pending. However, there may be a case where a migration ran successfully and completed, but there may be a valid reason to revert. 

This PR adds an arbitrary time limit of 20 minutes to revert once completed. I figured 20 minutes would be enough time to assess whether the impact of a migration would necessitate a revert.

### DIFF
--- a/snuba/migrations/policies.py
+++ b/snuba/migrations/policies.py
@@ -75,7 +75,7 @@ class NonBlockingMigrationsPolicy(MigrationPolicy):
 
         if status == Status.COMPLETED and timestamp:
             oldest_allowed_timestamp = datetime.now() + timedelta(
-                minutes=-MAX_REVERT_TIME_WINDOW_HRS
+                hours=-MAX_REVERT_TIME_WINDOW_HRS
             )
             if timestamp >= oldest_allowed_timestamp:
                 return False if migration.blocking else True
--- a/snuba/migrations/policies.py
+++ b/snuba/migrations/policies.py
@@ -1,9 +1,12 @@
 from abc import ABC, abstractmethod
+from datetime import datetime, timedelta
 
 from snuba.migrations.groups import get_group_loader
 from snuba.migrations.runner import MigrationKey, Runner
 from snuba.migrations.status import Status
 from snuba.utils.registered_class import RegisteredClass
+
+MAX_REVERT_TIME_WINDOW_MINS = 20
 
 
 class MigrationPolicy(ABC, metaclass=RegisteredClass):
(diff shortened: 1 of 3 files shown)


evanh (commented):
Why is there a time limit at all?

MeredithAnya:
> Why is there a time limit at all?

In general I don't think we want to be allowing migrations to be reversed, since that being the case means no further migrations can be run. However, if someone needs to revert right after running a migration, because of an incident or something, it would be not ideal if they were blocked by the current policy. That would require manual intervention by the few with prod access. 

This is just trying to add a bit more flexibility to revert when needed while still having some safeguards

evanh:
OK, I see where you're coming from.

In that case 20 minutes is way too short. I would put it at 24 or even 72 hours (because of weekends).

onewland:
Is the migration lifecycle documented anywhere? I think the PR logic here is probably correct but I don't actually know what `pending` means or what other states a migration could have

MeredithAnya:
> Is the migration lifecycle documented anywhere? I think the PR logic here is probably correct but I don't actually know what `pending` means or what other states a migration could have

I'm not sure if/where this is documented but when we try to look up the status of a migration, if there is no entry in the table the status is not_started, when a migration is run through the tool, the status is updated to pending before we actually running the migration, if it was successful, the status is updated to completed, if something went wrong trying to run the migration, the status remains pending

snuba/migrations/policies.py
```diff
@@ -1,10 +1,13 @@
 from abc import ABC, abstractmethod
+from datetime import datetime, timedelta
 
 from snuba.migrations.groups import get_group_loader
 from snuba.migrations.runner import MigrationKey, Runner
 from snuba.migrations.status import Status
 from snuba.utils.registered_class import RegisteredClass
 
+MAX_REVERT_TIME_WINDOW_HRS = 24
```
nikhars, line 9: Lets define this in settings so that we only need setting changes to happen if we need to change it

tests/migrations/test_policies.py
```diff
@@ -103,3 +105,21 @@ def test_pending_migration_reverse(self, mock_get_status: Mock) -> None:
...
         assert NonBlockingMigrationsPolicy().can_reverse(migration_key) == False
+
+    @patch(
+        "snuba.migrations.runner.Runner.get_status",
+        return_value=(
+            Status.COMPLETED,
+            datetime.now() + timedelta(hours=-(MAX_REVERT_TIME_WINDOW_HRS - 5)),
+        ),
+    )
+    def test_completed_migration_reverse(self, mock_get_status: Mock) -> None:
```
nikhars, line 116: Would be a good idea to write some docstring to tell what the test is doing.

tests/migrations/test_policies.py
```diff
@@ -103,3 +105,21 @@ def test_pending_migration_reverse(self, mock_get_status: Mock) -> None:
...
+        migration_key = MigrationKey(
+            MigrationGroup("events"), "0016_drop_legacy_events"
+        )
+        assert NonBlockingMigrationsPolicy().can_reverse(migration_key) == True
+
+        # if forward migration was blocking, reverse not allowed even when
+        # within the max time limit to
```
nikhars, line 123:
```suggestion
        # within the max time limit
```

tests/migrations/test_policies.py
MeredithAnya: added for both this test and the pending migration test

nikhars (approved):

PR #3659, opened 2023-01-27
fix(search-issues): Remove hardcoded key in consumer cli

enochtangg:
This was fixed in https://github.com/getsentry/snuba/pull/3502. We do not need this hardcoded key anymore.

### Testing
Ran `snuba consumer --storage search_issues --auto-offset-reset earliest` locally.


evanh (commented):
Have you verified that this will work with what we have in the prod settings?

evanh:
> could you please explain?

Honestly I'm not sure what I meant thinking over this. My mistake.

evanh (approved):

PR #3661, opened 2023-01-27
feat(DC): discover_events yaml

volokluev:
discover_events entity generated with the [entity conversion script](https://github.com/getsentry/snuba/tree/python_2_yaml_entity_converter)

Join relationships and subscription validators had to be updated manually


snuba/datasets/configuration/discover/entities/discover.yaml
```diff
@@ -363,8 +363,8 @@ storages:
         - mapper: DefaultNoneSubscriptMapper
           args:
             subscript_names:
-              - measurements
               - span_op_breakdowns
+              - measurements
```
evanh, line 367: Why this change?

evanh (approved):
One quick question.

snuba/datasets/configuration/discover/entities/discover.yaml
volokluev: It's an artifact of me changing the order in the discover.py file first and then seeing that it caused a difference in this file. More operations than necessary but no impact either way

PR #3663, opened 2023-01-27
test(migrations): check legacy use SNS-1998

dbanda:
We would like people to use the new migrations style. The old migrations class `ClickhouseNodeMigrationLegacy` was left for backwards compatibility with the old migrations. Going forwards we would like a new migrations to use the the new `ClickhouseNodeMigration` class. This adds a test in CI that checks that new migrations use the new class.

For current migration groups, new migrations are defined as migrations past the cutoff indexes.

### DIFF
--- a/tests/migrations/test_legacy_use.py
+++ b/tests/migrations/test_legacy_use.py
@@ -1,0 +1,45 @@
+from typing import List, Mapping, Tuple
+
+import pytest
+
+from snuba.migrations import migration
+from snuba.migrations.groups import MigrationGroup, get_group_loader
+
+all_migrations: List[Tuple[str, MigrationGroup, migration.ClickhouseNodeMigration]] = []
+for group in MigrationGroup:
+    group_loader = get_group_loader(group)
+    for migration_id in group_loader.get_migrations():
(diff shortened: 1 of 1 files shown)


tests/migrations/test_legacy_use.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_legacy_use() -> None:
+    """
+    check that we don't use legacy types in new migrations that are past the
+    cutoff marks
+    """
+    for migration_id, group, snuba_migration in all_migrations:
+        if isinstance(snuba_migration, migration.ClickhouseNodeMigrationLegacy):
+            mark = int(migration_id.split("_")[0])
+            if mark > int(legacy_cutoff[group]):
```
MeredithAnya, line 42: If someone made their first migration for a new migration group maybe we can just check to see if the group is even in the legacy_cutoff mapping, otherwise it looks like this will just key error (although i mean still wont pass)

MeredithAnya (approved):

tests/migrations/test_legacy_use.py
dbanda: thanks. fixed it.

PR #3671, opened 2023-01-30
fix(profiling): Profile id only on transaction dataset

Zylphrex:
When querying the profile_id, make sure it only uses the transactions dataset in this case.

Fixes SNUBA-31S


evanh (approved):
Thanks for fixing this.

PR #3676, opened 2023-01-30
fix(admin): remove the execute role for the default user

dbanda:
Removes the execute-non-blocking action from the default user. This  will make all migrations in the snuba admin tool readonly except from the TestMigration by default.


MeredithAnya (approved):
im assuming this will affect some tests, but I think it's fine to remove this one. We can more specific roles later

PR #3677, opened 2023-01-30
docs(aggregation): basic documentation for how we use aggregated tables in Snuba

onewland:
- with a special focus on materialization_version


docs/source/clickhouse/schema_design.rst
```diff
@@ -72,3 +73,78 @@ Those styles of queries require a separate column to search.
...
+
+1. Create a new materialized view ``click_events_aggregation_mv_v1`` via the migration system. This new materialized
+   view will use the WHERE clause or some kind of filtering to avoid setting quantilesState(0)
+   in the write for the ``click_duration`` column. This new materialized will only operate on
+   inputs in ``click_events_raw_local`` where ``materialization_version = 1``
+2. Test that this fixes the issue in your local environment by changing your consumer to use
+   ``materialization_version = 1``. It can make sense to control this via the settings file in
+   (in ``snuba/settings/__init.py__``)
+3. Run the migration in all relevant environments.
+4. Change the main consumer code, or the setting mentioned above in a specific environment, to
```
nikhars, line 146: Not a blocker for this PR, but something to consider. Would it make sense to control this via settings only going forward? Code changes and migrations would not always sync across different environments (SaaS, ST, Self-hosted). To solve this problem, a consumer needs to support all previous versions of writing to the raw table using the appropriate `materialization_version`. If this makes sense then we should recommend that you should make the change using settings.

onewland: I think deprecation is complicated, but we actually do want to completely deprecate the old paths in the case of `metrics` and `generic_metrics` for example. I think it will be to our benefit if we can ensure that this happens in all environments where maintainers are keeping them up to date.

That said, I'm happy to change the recommendation here to be clearer and make it always settings-based because it's easy enough to switch code paths to something hard-coded

nikhars (approved):

PR #3678, opened 2023-01-31
chore(oncall): add timing to healthcheck

volokluev:
We have some pods failing the readinessprobe. My hypothesis is that it is happening because the healthcheck is taking too long. This gives us more information on its operation

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -423,7 +423,7 @@ def health() -> Response:
     if status != 200:
         metrics.increment("healthcheck_failed", tags=metric_tags)
     metrics.timing(
-        "healthcheck_timing", time.time() - start, tags={"thorough": str(thorough)}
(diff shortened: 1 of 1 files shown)


snuba/web/views.py
```diff
@@ -422,7 +422,9 @@ def health() -> Response:
 
     if status != 200:
         metrics.increment("healthcheck_failed", tags=metric_tags)
-
+    metrics.timing(
+        "healthcheck_timing", time.time() - start, tags={"thorough": str(thorough)}
```
onewland, line 426:
```suggestion
        "healthcheck.latency", time.time() - start, tags={"thorough": str(thorough)}
```

I think matches the naming convention of other `.timing` usages

onewland (approved):
looks good, consider the rename

PR #3682, opened 2023-01-31
Revert "fix(profiling): Profile id only on transaction dataset"

volokluev:
This PR did not have the most recent changes from master at the time that it merged which caused a CI breakage. Reverting to fix master CI

Reverts getsentry/snuba#3671


enochtangg (approved):

volokluev:
This has been rolled forward

PR #3684, opened 2023-01-31
feat(metrics): Populate timeseries_id column in metrics_raw table

ayirr7:
https://getsentry.atlassian.net/browse/SNS-1973

This PR follows the same method that the `GenericMetricsBucketProcessor` uses in order to fill the timeseries_id column with data from the message.

### DIFF
--- a/snuba/datasets/processors/metrics_bucket_processor.py
+++ b/snuba/datasets/processors/metrics_bucket_processor.py
@@ -1,7 +1,8 @@
 import logging
+import zlib
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import Any, Mapping, Optional
+from typing import Any, Iterable, Mapping, Optional, Tuple, Union
 
 from snuba import settings
 from snuba.consumers.types import KafkaMessageMetadata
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/metrics_bucket_processor.py
```diff
@@ -35,6 +36,39 @@ def _should_process(self, message: Mapping[str, Any]) -> bool:
...
+
+        buffer = bytearray()
+        for field in [org_id, project_id, metric_id]:
+            buffer.extend(field.to_bytes(length=8, byteorder="little"))
+        for (key, value) in sorted_tag_items:
+            buffer.extend(bytes(key, "utf-8"))
+            if isinstance(value, int):
+                buffer.extend(value.to_bytes(length=8, byteorder="little"))
+            elif isinstance(value, str):
+                buffer.extend(bytes(value, "utf-8"))
```
ayirr7, line 56: Not entirely sure if we need to be checking for both cases, if I understand correctly the metrics indexer takes has the mapping from strings to integer indices

nikhars: There is no harm in adding code to handle the case of tag value being a string. Your understanding is correct that the tag values would be ints right now on the release health metrics pipeline.

nikhars (commented):
Please write some tests for the code you are introducing in the PR.

ayirr7:
> Please write some tests for the code you are introducing in the PR.

Which code are you referring to exactly? The code added (`_timeseries_id_token`) is tested in test_generic_metrics_processor.py, and the MetricsBucketProcessor, with the addition of this new column, is being tested in test_metrics_processor.py.

snuba/datasets/processors/metrics_bucket_processor.py
fpacifici: This code already exists in dynamic sampling. 
Cna we use the same hash and support both the cases where tags are int and the cases where the tags are strings ?
https://github.com/getsentry/snuba/blob/ba2daa237a95da5b4700df1451908859c259b326/snuba/datasets/processors/generic_metrics_processor.py#L49

ayirr7: This line: https://github.com/getsentry/snuba/blob/master/snuba/datasets/processors/metrics_bucket_processor.py#L54 suggests that values must be ints, are you suggesting to change this?

ayirr7: Are you suggesting to pull out the hash and `_timeseries_id_token` functions outside of the Generic Metrics Processor? I thought it made sense to make it part of each class in case they ever had different implementations. I can pull these functions out, if you don't think that would be a concern.

nikhars: If we had a common implementation, there are more chances of bugs creeping in when a particular dataset would like to change the implementation of timeseries for its own purpose. So duplication should be fine for this.

nikhars (approved):

PR #3685, opened 2023-01-31
ref(config) Remove Discover Python components

evanh:
This also required restructuring some values that existed in the Python
component files. The definition of event / transaction columns for the purposes
of selecting an entity moved to the special discover file, and some tests were
updated to have the values stored directly.

Now that the Discover selection is all in one place, it should be fairly
straightforward to port that function directly to Sentry.


tests/datasets/test_transaction_translations.py
```diff
@@ -14,6 +21,24 @@
...
+    functions=[DefaultNoneFunctionMapper(["isHandled", "notHandled"])],
+)
+
+null_function_translation_mappers = TranslationMappers(
+    curried_functions=[DefaultIfNullCurriedFunctionMapper()],
+    functions=[DefaultIfNullFunctionMapper()],
+)
+
+
+reset_dataset_factory()
```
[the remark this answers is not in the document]
evanh: Must have been leftover from something I was testing. I removed it.

snuba/query/snql/discover_entity_selection.py
```diff
@@ -32,9 +41,81 @@
 
 """
 
-EVENTS = EntityKey.DISCOVER_EVENTS
-TRANSACTIONS = EntityKey.DISCOVER_TRANSACTIONS
-EVENTS_AND_TRANSACTIONS = EntityKey.DISCOVER
+
+# NOTE: These will need to be updated if/when events and transaction columns change.
```
volokluev, line 45: can't we just get this from the entities?

evanh: Yes but I think that should be in a separate PR, which I'm working on.

volokluev (approved):

PR #3687, opened 2023-01-31
chore(replays): Add test coverage for possible ip-address variants

cmanallen:
We've observed that a large number of replays are not recording their ip-addresses in clickhouse (stored as null).  I've added test coverage to assert how the replay processor responds to various ip-address values.


cmanallen:
I'm loathe to close additions of test coverage but since we found the source of this issue it feels like this pull is not so relevant anymore.

PR #3688, opened 2023-02-01
ref(config) Remove events/errors Python components

evanh:
Remove the events/errors Python components. Also remove the EVENTS StorageKey, which wasn't used anywhere except in tests. Move some objects that were only used in tests into the relevant test.

UPDATE:
This is now the last of these PRs to be merged! This means there are a few extra changes:
- Remove the tests that were comparing Python components to YAML ones
- Change the EntityKey and StorageKey to no longer have a HARDCODED component
- Update the factories to only use the configuration components. This also inadvertently fixed a bug with how DISABLED_ENTITIES was working.


evanh:
The migration change above is to fix a Python migration so it initializes the factories before running.

snuba/datasets/entities/entity_key.py
```diff
@@ -2,26 +2,18 @@
...
 
-HARDCODED_ENTITY_KEYS = {
-    "EVENTS": "events",
-    "GROUPS": "groups",
-}
-
 REGISTERED_ENTITY_KEYS: dict[str, str] = {}
 
 
 class _EntityKey(type):
```
evanh, line 8: This class stays for now, since most of our code uses `EntityKey.SOMETHING`.

snuba/datasets/storages/storage_key.py
```diff
@@ -2,29 +2,18 @@
...
-    "EVENTS_RO": "events_ro",
-    "ERRORS": "errors",
-    "ERRORS_RO": "errors_ro",
-    "SPANS": "spans",
-}
-
 REGISTERED_STORAGE_KEYS: dict[str, str] = {}
 
 
 class _StorageKey(type):
```
evanh, line 8: This class stays since our code uses `StorageKey.SOMETHING`.

snuba/snuba_migrations/events/0014_backfill_errors.py
```diff
@@ -128,6 +129,7 @@ def get_monday(day: date) -> date:
 
 
 def backfill_errors(logger: logging.Logger) -> None:
+    initialize_snuba()  # ensure configuration is loaded
```
enochtangg, line 132: What is this change?

evanh: This migration fails if I didn't add this, because it references some of the components. Normally migrations don't need the components but this one does since it is a data migration.

enochtangg: Ah okay, I also see you wrote that above. Thanks

enochtangg (approved):

PR #3690, opened 2023-02-02
fix(search-issues): rebuild search_issues dataset with a version column to support replacements in the future

barkbarkimashark:
We wanna rebuild the search_issues table to include a `deleted` column required for replacements. 

This table is only present in local and dev environments so its ok to drop the table and rebuild.

I tested this locally and didn't encounter any issues.

### DIFF
--- a/snuba/snuba_migrations/search_issues/0004_rebuild_search_issues_with_version.py
+++ b/snuba/snuba_migrations/search_issues/0004_rebuild_search_issues_with_version.py
@@ -113,7 +113,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 table_name="search_issues_dist_new",
                 columns=columns,
                 engine=table_engines.Distributed(
-                    local_table_name="errors_local",
+                    local_table_name="search_issues_dist_new",
                     sharding_key="cityHash64(occurrence_id)",
                 ),
                 target=OperationTarget.DISTRIBUTED,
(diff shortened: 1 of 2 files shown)


MeredithAnya:
@barkbarkimashark I think Im a little confused why we don't want to just add the version column in a new migration, but maybe I'm missing something?

barkbarkimashark:
> @barkbarkimashark I think Im a little confused why we don't want to just add the version column in a new migration vs rebuilding the table, but maybe I'm missing something?

I think its mainly this line change here:
https://github.com/getsentry/snuba/pull/3690/files#diff-1349769bd353bf3a225887b8021c31b9070fb962bf9d32b9f24476c471421d29R80

@nikhars mentioned we can't change the engine params when its already been applied so we have to rebuild the table from scratch.

snuba/snuba_migrations/search_issues/0004_rebuild_search_issues_with_version.py
```diff
@@ -0,0 +1,155 @@
...
+    Column("user_name", String(Modifiers(nullable=True))),
+    Column("user_email", String(Modifiers(nullable=True))),
+    Column("ip_address_v4", IPv4(Modifiers(nullable=True))),
+    Column("ip_address_v6", IPv6(Modifiers(nullable=True))),
+    Column("sdk_name", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("sdk_version", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("contexts", Nested([("key", String()), ("value", String())])),
+    Column("http_method", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("http_referer", String(Modifiers(nullable=True))),
+    Column("version", UInt(8)),
```
nikhars, line 53: You should rename this column to `deleted` in order to be consistent with other tables and to avoid making additional changes in the replacer code in the future.

nikhars (approved):

PR #3692, opened 2023-02-02
fix(admin): read role policies from file

dbanda:
### context
In order to make the migrations tool viable , we have to introduce a permissions system to assign policies to users based on their role. Ideally, we would use the roles assigned in GCP and fetch them from there (see #3639 ). However, to do so programmatically, we need to assign our program a service account to do the fetching. We have opened a ticket with ops to do so. 

This is a temporary fix to read to roles  from file in the meantime. The file uses the same format as the standard IAM policy we would be able to fetch from GCP.  Roles are hardcoded in a list and the mapping from users to roles is kept in the policy json file. 

### before
All users get default roles

### after
Users get additional roles from the policy file

### testing
include a test in test_api.py that writes policy to a temp file and tests loading from it.


snuba/admin/iam_policy/iam_policy.json
```diff
@@ -0,0 +1,19 @@
+{
```
MeredithAnya, line 1: this should probably live in the ops repo right?

snuba/admin/auth.py
```diff
@@ -33,10 +37,41 @@ def authorize_request() -> AdminUser:
...
+    # TODO this will be handled by gcp identity once we have the service account
+    # https://cloud.google.com/identity/docs/how-to/query-memberships#searching_for_all_memberships_in_a_group
+    return False
+
+
+def get_iam_roles_from_file(user: AdminUser) -> Sequence[str]:
+    iam_roles = []
+    try:
+        with open(settings.ADMIN_IAM_POLICY_FILE, "r") as policy_file:
+            policy = json.load(policy_file)
```
MeredithAnya, line 50: should we handle invalid json errors as well?

snuba/admin/auth.py
```diff
@@ -33,10 +37,41 @@ def authorize_request() -> AdminUser:
...
+        )
+
+    return iam_roles
+
+
 def _set_roles(user: AdminUser) -> AdminUser:
     # todo: depending on provider convert user email
     # to subset of DEFAULT_ROLES based on IAM roles
-    user.roles = DEFAULT_ROLES
+    iam_roles = get_iam_roles_from_file(user)
```
MeredithAnya, line 73: or maybe you could wrap this in a try/except, and have the except set the roles to `DEFAULT_ROLES` and log the failure reason (invalid json, missing file, etc)

snuba/admin/iam_policy/iam_policy.json
dbanda: makes sense. I'll move it to ops.

snuba/admin/auth.py
dbanda: Not sure about this. My feeling is that if someone put a policy file and it's not proper json its probably better to fail cause that means something is seriously wrong. I wouldn't expect in normal situations for someone to put invalid json.

MeredithAnya (approved):

PR #3694, opened 2023-02-02
feat(outcomes): Add additional discard reasons

cmanallen:
https://github.com/getsentry/sentry-javascript/issues/6774

Adds two new discard reasons to support replays javascript sdk use-cases.  More information can be found in the linked issue.

- Are there other areas I need to change?
- Test coverage?

### DIFF
--- a/snuba/datasets/processors/outcomes_processor.py
+++ b/snuba/datasets/processors/outcomes_processor.py
@@ -30,6 +30,8 @@ CLIENT_DISCARD_REASONS = frozenset(
         "before_send",
         "event_processor",
         "sample_rate",
+        "send_error",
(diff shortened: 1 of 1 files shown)


nikhars (approved):

evanh:
>    Are there other areas I need to change?

Not in Snuba. I can't speak to ingest etc.

>    Test coverage?

There are tests in this file: https://github.com/getsentry/snuba/blob/master/tests/test_outcomes_api.py
They aren't specific to the processor though. I think this change is innocuous enough to not require a new test though.

cmanallen:
@evanh Sounds good.  Just wanted to be sure this was a low cardinality field and not an enum which would have required a migration.

PR #3695, opened 2023-02-02
feat(oncall): add envoy specific healthcheck endpoint

volokluev:
We are currently conflating two different healthchecks as one. There's the healthcheck that envoy uses to not send traffic to the pod while it's draining, and the healthcheck that k8s uses to know whether to restart the pod. Currently those are the same thing.

The hypothesis is that if we have them as two separate endpoints, we will stop getting erroneous restarts by k8s of the snuba API

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -390,6 +390,21 @@ def config_changes() -> RespTuple:
     )
 
 
+@application.route("/health_envoy")
+def health_envoy() -> Response:
+    down_file_exists = check_down_file_exists()
+
+    body: Mapping[str, Union[str, bool]]
+    if not down_file_exists:
+        status = 200
+    else:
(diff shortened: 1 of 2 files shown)


nikhars:
> The hypothesis is that if we have them as two separate endpoints, we will stop getting erroneous restarts by k8s of the snuba API
With your change, is the end state going to be that envoy uses /health_envoy and k8s will use existing /health endpoint. If yes, can you elaborate how erroneous restarts by k8s would be resolved since that will continue to use the existing endpoint.

volokluev:
There is another PR [here](https://github.com/getsentry/snuba/pull/3702) to fix the k8s healthceck. We just cannot merge that until we switch envoy to use the new endpoint

evanh (approved):

PR #3697, opened 2023-02-03
feat(self-hosted): Allow running sessions subscriptions in self hosted

### LINKED ISSUE #3696: About the sessions-subscription-results subscription issue
### Environment

- sentry | snuba version ：23.1.1

https://github.com/getsentry/snuba/pull/2737 ，@lynnagara Hello, I have a question about this pr, hope to get your answer, thank you very much

- After removing subscriptions-scheduler-executor-session support in snuba, how to write data to the topic of sessions-subscription-results? Because I see that the crash rate warning code in sentry is still there and has not changed, for example

- https://github.com/getsentry/sentry/pull/28526

https://github.com/getsentry/sentry/blob/8e00dcdf463d916b9ca79ddbe13e99f161d58db1/src/sentry/snuba/query_subscription_consumer.py#L61-L61

My original question is as follows, I have enabled the organizations:incidents function in sentry and subscribed to sessions-results through the following script
```bash
sentry
run
query-subscription-consumer
--topic=sessions-subscription-results
```
Because there is no data in the sessions-subscription-results topic, the crash rate alarm cannot work
image

lynnagara:
Closes #3696

### DIFF
--- a/snuba/cli/subscriptions_scheduler_executor.py
+++ b/snuba/cli/subscriptions_scheduler_executor.py
@@ -24,7 +24,7 @@ from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
     "--dataset",
     "dataset_name",
     required=True,
-    type=click.Choice(["events", "transactions", "metrics"]),
+    type=click.Choice(["events", "transactions", "metrics", "sessions"]),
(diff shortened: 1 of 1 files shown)


volokluev (approved):
sigh

PR #3700, opened 2023-02-03
fix(consumer): Separate out producers in consumer_builder

ayirr7:
https://github.com/getsentry/snuba/pull/3626 (third time's the charm?)

This time, the Producers are configured using the logical/default topic name, which is derived from the topic spec, regardless of whether there is a topic override being passed in.

### DIFF
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -114,8 +114,7 @@ class ConsumerBuilder:
             else:
                 self.replacements_topic = None
 
-        if self.replacements_topic is not None:
-            assert replacement_topic_spec is not None
+        if replacement_topic_spec is not None:
             self.replacements_producer = Producer(
                 build_kafka_producer_configuration(
                     replacement_topic_spec.topic,
@@ -141,8 +140,7 @@ class ConsumerBuilder:
             else:
                 self.commit_log_topic = None
 
-        if self.commit_log_topic is not None:
-            assert commit_log_topic_spec is not None
+        if commit_log_topic_spec is not None:
             self.commit_log_producer = Producer(
                 build_kafka_producer_configuration(
                     commit_log_topic_spec.topic,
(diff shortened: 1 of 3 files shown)


snuba/consumers/consumer_builder.py
```diff
@@ -112,36 +103,56 @@ def __init__(
...
             else:
                 self.commit_log_topic = None
 
-        self.stats_callback = stats_callback
-
-        # XXX: This can result in a producer being built in cases where it's
-        # not actually required.
-        self.producer = Producer(self.producer_broker_config)
+        if self.commit_log_topic is not None:
+            assert commit_log_topic_spec is not None
```
lynnagara, line 145:
```suggestion
        if commit_log_topic_spec is not None:
```

snuba/consumers/consumer_builder.py
```diff
@@ -261,7 +272,9 @@ def build_streaming_strategy_factory(
                 table_writer,
                 metrics=self.metrics,
                 replacements_producer=(
-                    self.producer if self.replacements_topic is not None else None
+                    self.replacements_producer
+                    if self.replacements_topic is not None
+                    else None
```
lynnagara, line 277: This could just be `self.replacements_producer`. It's always defined and should already be set to None in cases where there's no replacement topic.

snuba/consumers/consumer_builder.py
```diff
@@ -112,36 +103,56 @@ def __init__(
...
-            replacement_topic_spec = stream_loader.get_replacement_topic_spec()
             if replacement_topic_spec is not None:
                 self.replacements_topic = Topic(
                     replacement_topic_spec.get_physical_topic_name(slice_id)
                 )
             else:
                 self.replacements_topic = None
 
+        if self.replacements_topic is not None:
+            assert replacement_topic_spec is not None
```
lynnagara, line 118:
```suggestion
        if replacement_topic_spec is not None:
```

snuba/consumers/consumer_builder.py
ayirr7: Yep, this is true

lynnagara (approved):

PR #3701, opened 2023-02-03
docs(clickhouse): describe basic Sentry CH usage

onewland:
- describe local vs dist tables
- describe storage vs. query nodes


docs/source/clickhouse/topology.rst
```diff
@@ -0,0 +1,45 @@
...
+-----------------------------------
+
+Astute snuba users might notice that migrations contain references to tables with names
+suffixed with ``_local`` and table names suffixed with ``_dist``. This is used to
+distinguish between distributed tables (generally using the ClickHouse table engine
+`Distributed <https://clickhouse.com/docs/en/engines/table-engines/special/distributed/>`_)
+and local tables (generally using one of the
+`MergeTree-derived <https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/>`_
+table engines). Distributed tables exist to aggregate from the shards of local tables, and following
+the sequence above, distributed tables tend to be created on query nodes and local tables
```
evanh, line 36: > the sequence above

Nit: This wording seems a little weird. What sequence above?

docs/source/clickhouse/topology.rst
```diff
@@ -0,0 +1,45 @@
+==================================
+ClickHouse Topology Best Practices
+==================================
+
+Sentry has a few conventions for using ClickHouse that are not inherent to
+the database but useful to consider if you intend to create new tables or
+datasets.
+
+Storage nodes vs. Query nodes
```
evanh, line 9: I had thought there was a general idea to move away from separate query nodes and put the distributed table directly on the storage nodes, like metrics. Or was that specific to metrics?

onewland: That was specific to release-health metrics (dataset `metrics`) and we're going to go back and add query nodes to those because it's actually been problematic for ops already to not have that layer of abstraction

docs/source/clickhouse/topology.rst
onewland: yeah, I'm referring to the two-point listing above. agree that it's weird, can you think of a better word than "sequence"? maybe section?

dbanda (approved):
Nice doc. Well written and clear 💯

docs/source/clickhouse/topology.rst
evanh: I would just remove that line. 

> Distributed tables exist to aggregate from the shards of local tables, so they tend to be created on query nodes instead of storage nodes.

I just realized, why do you say local tables `tend` to be on storage nodes? When would a local table be on a query node?

onewland: never, I can reword

docs/source/clickhouse/schema_design.rst
```diff
@@ -73,3 +74,78 @@ Those styles of queries require a separate column to search.
...
+* ``widgets_aggregation_mv`` for the materialized view (this only exists on storage nodes)
+* (``widgets_aggregated_local``, ``widgets_aggregated_dist``) for the roll-up/aggregated (local, distributed) tables
+
+Materialized views are immutable so it's normal to have multiple versions of
+``widgets_aggregation_mv`` when behavior is updated, with suffixes like
+``widgets_aggregation_mv_v1``, ``widgets_aggregation_mv_v2``, etc. Migration
+between materialized view versions are described in the next section but in general
+old materialized views should be discarded once they are no longer used.
+
+Schema migrations using materialization_version
```
fpacifici, line 107: Please deprecate and delete the internal documentation when this is merged so we do not have two independent places that will evolve on their own.

fpacifici:
> Content of the document looks good to me.
> 
> One question I do have is regarding privacy. Are we allowed to publish publicly what our internal layout looks like? There is nothing in here which is intellectual property. You might want to find out whether it is ok or not to publish this in external docs.

@nikhars  this level of detail is fine. It does not contain any sensitive information.

PR #3702, opened 2023-02-03
feat(oncall): don't fail the k8s healthcheck if the down file exists

volokluev:
Depends on this being merged:

https://github.com/getsentry/ops/pull/6017


snuba/web/views.py
```diff
@@ -390,6 +390,30 @@ def config_changes() -> RespTuple:
...
+    """K8s can decide to shut down the pod, at which point it will write the down file.
+    This down file signals that we have to drain the pod, and not accept any new traffic.
+    In SaaS, envoy is the thing that will decided to route traffic to this node or not. It
+    uses this endpoint to make that decision.
+
+    This differs from the generic health endpoint because the pod can still be healthy
+    but just not accepting traffic. That way k8s will not restart it until it is drained
+    """
+
+    down_file_exists = check_down_file_exists()
```
evanh, line 404: Shouldn't this also check clickhouse connections?

volokluev: No it should not. Envoy healthchecks are shallow currently:

https://github.com/getsentry/ops/pull/6017/files#diff-2670249352f9304aeb27eb921ee45113bdb174e2150f969baac421d747bee03eL49

evanh (approved):
Pending CI fixes this looks good to me.

PR #3705, opened 2023-02-06
ref(logging) Don't log full exception trace for rate limit errors

evanh:
Currently in order to send rate limit errors to Sentry, the LoggingIntegration
is used, which requires logging the exception.

This generally works fine, since we want the exception to be in the logs as well
as in Sentry.

However with rate limit exceptions, they tend to have many layers of tracebacks,
which results in multiple lines being logged for a single error, and an error
that having the traceback isn't useful in logs.

Add a new utlity that sends a warning to Sentry. This is a manual workaround.
Use that workaround to send the warning event to Sentry, but just log the
message without the full stack trace.

### Legal Boilerplate

### DIFF
--- a/snuba/util.py
+++ b/snuba/util.py
@@ -2,6 +2,7 @@ import inspect
 import logging
 import numbers
 import re
+import sys
 from datetime import date, datetime, timedelta
 from enum import Enum
 from functools import partial, wraps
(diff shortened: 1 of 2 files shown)


volokluev (approved):

evanh:
Out of date

PR #3707, opened 2023-02-06
ref(consumer): Merging consumers in consumer_builder

ayirr7:


snuba/cli/consumer.py
```diff
@@ -135,13 +138,23 @@ def consumer(
...
-
-    metrics_tags = {
-        "group": consumer_group,
-        "storage": storage_key.value,
+    storages = {
+        key: get_writable_storage(key)
+        for key in (StorageKey(name) for name in storage_names)
     }
+    storage_keys = [*storages.keys()]
+    # sentry_sdk.set_tag("storage", storage_name)
```
lynnagara, line 146: Why removing this line?

snuba/cli/consumer.py
```diff
@@ -135,13 +138,23 @@ def consumer(
...
+
+    if len(storages) == 1:
```
lynnagara, line 148: Just to confirm, it's not possible to pass 0 storages right?

snuba/consumers/consumer_builder.py
```diff
@@ -75,31 +122,38 @@ def __init__(
...
-            topic, slice_id, bootstrap_servers=kafka_params.bootstrap_servers
+            SnubaTopic(self.__topic),
+            slice_id,
+            bootstrap_servers=kafka_params.bootstrap_servers,
         )
         logger.info(f"librdkafka log level: {self.broker_config.get('log_level', 6)}")
 
-        stream_loader = self.storage.get_table_writer().get_stream_loader()
+        # Get a random storage from the storage dict
+        rand_storage = [*self.storages.values()][0]
```
lynnagara, line 146: It doesn't seem very random... isn't it just the first one?

snuba/consumers/consumer_builder.py
```diff
@@ -237,11 +283,52 @@ def log_general_error(e: KafkaError) -> None:
 
         return StreamProcessor(consumer, self.raw_topic, strategy_factory, IMMEDIATE)
 
+    # We currently don't having slicing implemented for multistorage case
```
lynnagara, line 286: I suggest you immediately throw an error if the user tries to pass a slice with multiple storages, so it is very clear that combination is not support, rather than just relying on someone reading comment to figure out it doesn't work

snuba/consumers/consumer_builder.py
```diff
@@ -237,11 +283,52 @@ def log_general_error(e: KafkaError) -> None:
...
+            processes=self.processes,
+            input_block_size=self.input_block_size,
+            output_block_size=self.output_block_size,
+            metrics=self.metrics,
+            dead_letter_policy_creator=dead_letter_policy_creator,
+            commit_log_config=commit_log_config,
+        )
+
+        return strategy_factory
+
```
lynnagara, line 322: Where did this code come from? Can we remove the other implementation rather than keeping 2 around? Alternatively, you can call the existing implementation from here and bring it over as a follow up. I think you should avoid ever having 2 copies of the same code.

tests/consumers/test_consumer_builder.py
```diff
@@ -95,21 +101,23 @@ def test_consumer_builder_non_optional_attributes(con_build) -> None:  # type: i
     # Depending on the attribute, we can verify this
     # to different degrees
 
-    assert con_build.storage == get_writable_storage(test_storage_key)
+    storages = list(con_build.storages.values())
+    if len(storages) == 1:
```
lynnagara, line 105: Can you add tests for the case where `len(storages) > 1`?

snuba/consumers/consumer_builder.py
```diff
@@ -54,6 +57,50 @@ class ProcessingParameters:
     output_block_size: Optional[int]
 
 
+def verify_single_topic(storages: Sequence[WritableTableStorage]) -> str:
```
lynnagara, line 60: Not sure I understand why this function returns a string. It verifies 3 topics - the input one, the commit log one, and the replacement one. Why does it need to return one of them?

This is similar to the code that already exists in the multistorage implementation. I would suggest you avoid making a second copy of that code by either:
1) calling this function from that implementation as well, or:
2) calling that existing implementation from here then later moving it over

I think 2 could be a good idea, so you can make smaller incremental changes.

lynnagara (commented):
When making incremental, risky refactoring like this I think it would be better to avoid making duplicate copies of functionality. Right now, there are some pieces of code copied over from the multistorage consumer. My preference would be to either immediately remove the other implementation or start by just calling the existing implementation from here and moving it to the other module as a follow up.

ayirr7:
> When making incremental, risky refactoring like this I think it would be better to avoid making duplicate copies of functionality. Right now, there are some pieces of code copied over from the multistorage consumer. My preference would be to either immediately remove the other implementation or start by just calling the existing implementation from here and moving it to the other module as a follow up.

Oh sorry, I forgot to mention, my plan is to remove the multistorage consumer file once this is finalized. Do you have a preference between 1) removing the multistorage_consumer.py file altogether or 2) calling functions from that file in here?

snuba/consumers/consumer_builder.py
ayirr7: Makes sense. I'll call the similar code that lives in multistorage_consumer here.

snuba/consumers/consumer_builder.py
ayirr7: Oops, I forgot to update the comment and names from earlier

snuba/consumers/consumer_builder.py
ayirr7: The reason why it's returning the string (which is the raw logical topic) is because I thought it'd be easy/cleaner to just return that from this function, which already computes `default_topics`. Otherwise I would just have to compute `default_topics` again and get the topic again outside of this `verify_single_topic` function. I can do that too, this was just for cleaniness/not repeating computations.

snuba/cli/consumer.py
ayirr7: Fixed

snuba/consumers/consumer_builder.py
ayirr7: In the new refactor, `build_multistorage_streaming_strategy_factory` is in multistorage_consumer.py, with defined args, so someone can't pass in slice_id.

snuba/cli/consumer.py
ayirr7: It's not possible because the CLI arg is non-optional.

tests/consumers/test_consumer_builder.py
ayirr7: Fixed

ayirr7:
Update: The main new changes here involve refactoring some of the code I had copied over from multistorage consumer and writing them as separate functions in the multistorage consumer. Now, the consumer_builder calls these functions directly. As one can see, with these new functions, there are ways to reduce the number of lines in both consumer_builder and multistorage_consumer (specifically in the parts of code where we resolve topics) but I wanted to put out the initial idea for the refactor / potentially get feedback for it first before I made the code cleaner and less repetitive.

snuba/cli/consumer.py
```diff
@@ -135,13 +139,23 @@ def consumer(
...
+        metrics_tags = {
+            "group": consumer_group,
+            "storage": storage_keys[0].value,
+        }
+    else:
+        metrics_tags = {
+            "group": consumer_group,
+            "storage": "_".join([storage_keys[0].value, "m"]),
```
fpacifici, line 157: I think this tag will end up not being very useful as you cannot search in Datadog by a substring of a tag (only prefix are allowed).
I think we should have a ticket to stop filtering our dashboards and alerts by storage and do it by consumer group as a consumer group is what identifies a consumer instead.
But then we could get into issues if we allowed multiple topics to be consumed by the same consumer on the same consumer group.

What do you think ?

lynnagara: We talked about updating this to `",".join(storage_names)` or something. It is true that we normally assign different names to each group, though that is nothing more than a convention we use at Sentry and nothing else actually stops the same name from being used in multiple places. So I'd personally just leave the tag in there as it seems harmless.

ayirr7:
[Just an update] This PR is not going to be merged, as we discussed the idea of breaking it down into multiple mini-PRs so that we can test each incremental change more easily. We'll keep it around for some reference until the scope of these changes have been merged.

PR #3708, opened 2023-02-06
feat: add org generic counters [TET-695]

andriisoldatenko:
This PR adds the ability to query cross org query for the generic metric counter.


snuba/datasets/configuration/generic_metrics/entities/org_counters.yaml
```diff
@@ -0,0 +1,57 @@
+version: v1
+kind: entity
+name: org_metrics_counters
```
evanh, line 3: I would make this `generic_org_metrics_counters`, otherwise the name will collide with the existing entity.

snuba/datasets/configuration/generic_metrics/entities/org_counters.yaml
```diff
@@ -0,0 +1,57 @@
...
+    { name: org_id, type: UInt, args: { size: 64 } },
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: metric_id, type: UInt, args: { size: 64 } },
+    { name: timestamp, type: DateTime },
+    { name: bucketed_time, type: DateTime },
+  ]
+required_time_column: timestamp
+
+storages:
+  - storage: org_metrics_counters
```
evanh, line 15: I mentioned this in Slack, but this storage is referencing the release health metrics counters. You will need a new storage in `generic_metrics/storages/` that will be very similar to [this one](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/metrics/storages/org_counters.yaml), but with a new name. Then you can reference that storage here.

snuba/datasets/configuration/generic_metrics/storages/org_counters.yaml
```diff
@@ -0,0 +1,19 @@
+version: v1
+kind: readable_storage
+name: generic_org_metrics_counters
+storage:
+  key: generic_org_metrics_counters
+  set_key: generic_org_metrics_counters
```
evanh, line 6: This should be `generic_metrics_counters`. The `set_key` is different from the `key`, and specifies which cluster the storage is referencing. You don't need a new cluster.

andriisoldatenko: fixed.

andriisoldatenko:
> This all looks good to me, but I think you should add a test here so you can do an E2E check of the new entity: https://github.com/getsentry/snuba/blob/master/tests/test_generic_metrics_api.py

do you want me to add `TestGenericMetricsApiCounter` and `TestOrgMetricsApiCounter` ?

evanh (approved):
Looks good!

untitaker (approved):

snuba/datasets/configuration/generic_metrics/storages/org_counters.yaml
```diff
@@ -0,0 +1,19 @@
...
+  key: generic_org_metrics_counters
+  set_key: generic_metrics_counters
+schema:
+  columns:
+    [
+      { name: org_id, type: UInt, args: { size: 64 } },
+      { name: project_id, type: UInt, args: { size: 64 } },
+      { name: metric_id, type: UInt, args: { size: 64 } },
+      { name: granularity, type: UInt, args: { size: 8 } },
+      { name: timestamp, type: DateTime },
```
onewland, line 14: is this missing the actual count column?

https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/storages/counters.yaml#L48-L56

or do you just not need it?

onewland (approved):

snuba/datasets/configuration/generic_metrics/storages/org_counters.yaml
andriisoldatenko: I think we need it here, @evanh  @untitaker wdyt? 
I'm not very familiar with how this will work under the hood.

andriisoldatenko: only i can say it works with and without this block. i just did quick test.

andriisoldatenko: I see, yeah I prefer to implement as minimally as we need. So later we can extend.

PR #3709, opened 2023-02-06
feat(search-issues): process context data when processing payload

barkbarkimashark:
Adds support for processing `context` arbitrary data bag from the incoming event for search_issues dataset.

### DIFF
--- a/snuba/datasets/processors/search_issues_processor.py
+++ b/snuba/datasets/processors/search_issues_processor.py
@@ -170,7 +170,7 @@ class SearchIssuesMessageProcessor(DatasetMessageProcessor):
 
     def _process_contexts(
         self, event_data: IssueEventData, processed: MutableMapping[str, Any]
-    ):
+    ) -> None:
         contexts = event_data.get("contexts", {})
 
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/search_issues_processor.py
```diff
@@ -167,6 +168,15 @@ def _process_sdk_data(
         processed["sdk_name"] = _unicodify(sdk.get("name"))
         processed["sdk_version"] = _unicodify(sdk.get("version"))
 
+    def _process_contexts(
```
barkbarkimashark, line 171: We're thinking about having the 'custom' context processing done upstream within sentry app code to avoid doing business logicy stuff in this processor. 
Some examples:
https://github.com/getsentry/snuba/blob/8cf41e25b0fc2f97a84ddca14ad7e569d0d6314d/snuba/datasets/processors/transactions_processor.py#L256-L260
https://github.com/getsentry/snuba/blob/0c8a30814f441c491d9db76aab59b54fdc372b6a/snuba/datasets/processors/errors_processor.py#L170-L173

So, ideally, the incoming kafka message context should be written to the row with little to no custom processing.

evanh (approved):

PR #3714, opened 2023-02-08
ref(consumer): Build strategy factory function in multistorage consumer

ayirr7:
We are breaking down https://github.com/getsentry/snuba/pull/3707 into smaller, incremental PRs. This way, we can understand and finalize each incremental change faster and with more confidence. 

This PR pulls out the existing code for building the multistorage streaming strategy factory into its function. Ultimately, this would be called from the consumer_builder.

### DIFF
--- a/snuba/cli/multistorage_consumer.py
+++ b/snuba/cli/multistorage_consumer.py
@@ -1,13 +1,14 @@
 import logging
 import signal
-from typing import Any, Optional, Sequence
+from typing import Any, Dict, Optional, Sequence
 
 import click
 import rapidjson
 from arroyo import Topic, configure_metrics
-from arroyo.backends.kafka import KafkaConsumer
+from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
 from arroyo.commit import IMMEDIATE
 from arroyo.processing import StreamProcessor
+from arroyo.processing.strategies import ProcessingStrategyFactory
 from confluent_kafka import Producer as ConfluentKafkaProducer
 
 from snuba import environment, settings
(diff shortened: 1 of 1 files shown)


snuba/cli/multistorage_consumer.py
```diff
@@ -259,6 +262,44 @@ def stats_callback(stats_json: str) -> None:
...
+    def handler(signum: int, frame: Any) -> None:
+        processor.signal_shutdown()
+
+    signal.signal(signal.SIGINT, handler)
+    signal.signal(signal.SIGTERM, handler)
+    processor.run()
+
+
+def build_multistorage_streaming_strategy_factory(
+    storages: Dict[Any, WritableTableStorage],
```
lynnagara, line 293: This type could probably be more specific than Any. Though would it make sense to pass a sequence of storage keys here instead of a dict? Since we already get a sequence of those from both the consumer and multistorage CLI, it could make sense to pass it straight into this function.

ayirr7: Hm, so I changed the type of the arg that gets passed into the build function. Were you suggesting to remove the `storages` dictionary altogether? 

I agree about validating the topics beforehand so we don't build the factory unless necessary. I think that may need to live in a different function though as it might be a fairly long function with different topic validations and/or resolutions.

lynnagara: It could still be pulled into a separate function. As long as that separate function is called from within `build_multistorage_streaming_strategy_factory` and not outside of it.

ayirr7: Yep, that makes sense, that would be my next PR.

lynnagara (approved):

PR #3717, opened 2023-02-08
feat(rust-consumer): Move code from rust arroyo to a separate crate

lynnagara:
We don't want the arroyo code to be in the root of the Snuba project, move it to a separate crate


untitaker (commented):
for more proper isolation, do a separate crate too

lynnagara:
@untitaker Update rust_arroyo to a crate. Lmk if this setup looks right.

rust_snuba/rust_arroyo/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
```
fpacifici, line 1: Do you plan to actually publish a crate from rust_arroyo and import that eventually or do you plan to vendor rust_arroyo into snuba permanently ?

The first is better.

lynnagara: I think the first is better in the long term and the second is better in the short term. I would hold off publishing a crate until we have a much better idea of what that library would look like. Right now we are unsure and prioritising for faster iteration.

dbanda (approved):

PR #3718, opened 2023-02-09
feat(admin): allow querylog to be queried by more threads

volokluev:
There are many useful querylog queries which we cannot do because we only have one thread to query the querylog with. Up the max_threads to 4 but make it runtime configurable.

NOTE:

I'm not sure how to test this to know that it works as expected (except for deploying it). It seems like you can pass anything you want into the `settings` dict, and have max_threads be any value and the query will still succeed. Would love some guidance on this if anyone has ideas.

### DIFF
--- a/snuba/admin/clickhouse/querylog.py
+++ b/snuba/admin/clickhouse/querylog.py
@@ -10,6 +10,8 @@ from snuba.datasets.schemas.tables import TableSchema
 from snuba.datasets.storages.factory import get_storage
 from snuba.datasets.storages.storage_key import StorageKey
 
+_MAX_CH_THREADS = 4
+
 
 @audit_log
 def run_querylog_query(query: str, user: str) -> ClickhouseResult:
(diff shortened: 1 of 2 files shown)


snuba/admin/clickhouse/querylog.py
```diff
@@ -39,5 +40,9 @@ def __run_querylog_query(query: str) -> ClickhouseResult:
     connection = get_ro_query_node_connection(
         StorageKey.QUERYLOG.value, ClickhouseClientSettings.QUERY
     )
-    query_result = connection.execute(query=query, with_column_types=True)
+    query_result = connection.execute(
+        query=query,
+        with_column_types=True,
+        settings={"max_threads": state.get_config("admin.querylog_threads", 4)},
```
fpacifici, line 46: please cap this value. 
If anybody can set whatever they want, they can set a value that is large enough to use enough CPU to trip the alert that would page the ops on-call.

volokluev: good call, done

snuba/admin/clickhouse/querylog.py
```diff
@@ -31,6 +34,18 @@ def describe_querylog_schema() -> ClickhouseResult:
...
 
+def _get_clickhouse_threads() -> int:
+    config_threads = state.get_config("admin.querylog_threads", _MAX_CH_THREADS)
+    try:
+        return min(
+            int(config_threads) if config_threads is not None else _MAX_CH_THREADS,
+            _MAX_CH_THREADS,
+        )
+    except ValueError:
+        # in case the config is set incorrectly
```
evanh, line 45: I don't think we should swallow this error. Otherwise someone will think they successfully changed the threads.

volokluev: done

evanh (approved):

PR #3721, opened 2023-02-09
feat(replays): add replay_id to transactions table

JoshFerge:
As per https://github.com/getsentry/rfcs/blob/main/text/0048-move-replayid-out-of-tags.md, this pull request adds a `replay_id` column to the transactions table.


evanh (approved):

JoshFerge:
will this migration get ran as part of CI? or do i need to make a ticket? @evanh

evanh:
> will this migration get ran as part of CI? or do i need to make a ticket?

You will need to make a ticket. Migrations aren't in CI yet.

PR #3722, opened 2023-02-10
feat(replays): add replay_id to errors table

JoshFerge:
As per https://github.com/getsentry/rfcs/blob/main/text/0048-move-replayid-out-of-tags.md, this pull request adds a replay_id column to the errors table. 

I'm not sure how `ERRORS_RO` is used, and if its needed in this migration, but not doing anything with it now.

### DIFF
--- a/snuba/snuba_migrations/events/0019_add_replay_id_column.py
+++ b/snuba/snuba_migrations/events/0019_add_replay_id_column.py
@@ -21,7 +21,7 @@ class Migration(migration.ClickhouseNodeMigration):
             )
             for table_name, target in [
                 ("errors_local", OperationTarget.LOCAL),
-                ("transactions_dist", OperationTarget.DISTRIBUTED),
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/events/0019_add_replay_id_column.py
```diff
@@ -0,0 +1,40 @@
...
+    def backwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.DropColumn(
+                storage_set=StorageSetKey.EVENTS,
+                table_name=table_name,
+                column_name="replay_id",
+                target=target,
+            )
+            for table_name, target in [
+                ("errors_local", OperationTarget.LOCAL),
```
evanh, line 37: I believe when dropping a column the dist table has to be before the local table.

cmanallen (approved):

evanh (approved):

PR #3724, opened 2023-02-10
migrations: add custom search issues role

dbanda:
Adds custom role to allow running the search issue migrations by default. This is part of the effort to run search issue with the migrations tool.  Follows up #3720


evanh (approved):

dbanda:
> @dbanda wanna add that role to some SNS folks in the bindings file? so we could be able to run the migrations?

Its non blocking for default members but I added it to the binding file too.

PR #3727, opened 2023-02-10
fix: Reject improperly typed tag conditions as invalid queries

evanh:
Validate that simple tag conditions (conditions with a tags column on the left
side and a literal or array of literals on the right side) are comparing against
strings and nothing else.

This is a common error that comes from Sentry, and other current checks don't
catch it so it gets sent to Clickhouse and reported to Sentry as an error.


snuba/query/conditions.py
```diff
@@ -313,40 +315,58 @@ def is_condition(exp: Expression) -> bool:
 
 
 def build_match(
```
volokluev, line 317: It's not clear to me what this function does, a high level description would be useful to orient me as a reviewer. Also this seems to be a core function and its functionality has changed but I don't see any tests of the function specifically in this PR.

snuba/query/validation/validators.py
```diff
@@ -255,3 +258,58 @@ def validate(self, query: Query, alias: Optional[str] = None) -> None:
...
+class TagConditionValidator(QueryValidator):
+    """
+    Sometimes a query will have a condition that compares a tag value to a
+    non-string literal. This will always fail (tags are always strings) so
+    catch this specific case and reject it.
+    """
+
+    def __init__(self) -> None:
+        self.condition_matcher = build_match(
+            subscriptable="tags",
```
volokluev, line 272: I think the `subscriptable` parameter should be an argument in the `__init__`

snuba/query/validation/validators.py
```diff
@@ -255,3 +258,58 @@ def validate(self, query: Query, alias: Optional[str] = None) -> None:
             raise InvalidQueryException(
                 f"granularity must be multiple of {self.minimum}"
             )
+
+
+class TagConditionValidator(QueryValidator):
```
volokluev, line 263: This should have a test of its own rather than relying on the api integration test

volokluev (commented):
Love the change and that we'll finally reject these queries. Just think the test coverage should be improved and while we're here let's add some context around `build_match` currently I don't have enough knowledge to review it quickly

snuba/query/validation/validators.py
evanh: I thought about that but this is a pretty specific use case so I decided to keep it limited to tags for now.

snuba/query/conditions.py
```diff
@@ -313,40 +315,67 @@ def is_condition(exp: Expression) -> bool:
...
+    ops: Optional[Sequence[str]] = None,
+    array_ops: Optional[Sequence[str]] = None,
+    param_type: Optional[Any] = None,
     alias: Optional[str] = None,
     key: Optional[str] = None,
 ) -> Or[Expression]:
-    # The IN condition has to be checked separately since each parameter
-    # has to be checked individually.
+    """
+    There is a common use case of matching a specific condition in our code base.
```
volokluev, line 327: This is better but not quite what I want as a reader.

What would make this easier are examples of usage and what patterns it would match (at least 2). 

Also, all the parameters are strings or `Any` and it's not clear what they mean unless I read the code in depth. A description of the parameters would be useful here too

evanh: This is a pretty simple function that does one simple thing. I'm not sure what else I could add that makes it clearer. If you want examples of usage, look at examples of where we use it in the codebase. "Reading the code in depth" is less than 50 lines of code. That doesn't seem like huge burden.

volokluev (approved):

PR #3728, opened 2023-02-10
ref(consumer): Separate out a get_logical_topics function in multistorage consumer

ayirr7:
This is the second mini step. There is scattered topic validation throughout the multistorage consumer which checks for and ensures that there is only one topic or one dead letter policy in a multistorage context. 

This PR pulls out those validations into a single function and reduces a bit of repetition. With this PR we can also predict some of the small refactors we will need to get multistorage consumer to look a bit more like ConsumerBuilder. In the next step(s), we can move some pieces of code around to get a `get_consumer` and a `build_streaming_strategy_factory` function like ConsumerBuilder has.

### DIFF
--- a/snuba/cli/multistorage_consumer.py
+++ b/snuba/cli/multistorage_consumer.py
@@ -1,6 +1,6 @@
 import logging
 import signal
-from typing import Any, Optional, Sequence
+from typing import Any, Callable, Optional, Sequence, Tuple
 
 import click
 import rapidjson
@@ -9,6 +9,9 @@ from arroyo.backends.kafka import KafkaConsumer, KafkaPayload
 from arroyo.commit import IMMEDIATE
 from arroyo.processing import StreamProcessor
 from arroyo.processing.strategies import ProcessingStrategyFactory
+from arroyo.processing.strategies.dead_letter_queue.policies.abstract import (
+    DeadLetterQueuePolicy,
+)
 from confluent_kafka import Producer as ConfluentKafkaProducer
 
 from snuba import environment, settings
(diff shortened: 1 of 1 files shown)


snuba/cli/multistorage_consumer.py
```diff
@@ -289,30 +257,91 @@ def handler(signum: int, frame: Any) -> None:
...
-def build_multistorage_streaming_strategy_factory(
+LogicalTopicTuple = Tuple[
+    Topic,
+    Optional[Topic],
+    Optional[Topic],
+    Optional[Callable[[], DeadLetterQueuePolicy]],
+]
+
+
+def get_logical_topics(
```
lynnagara, line 268: How about calling this `get_consumer_config()`. Not everything here is a topic, and it seems likely we will return more config from here in the future.

snuba/cli/multistorage_consumer.py
```diff
@@ -289,30 +257,91 @@ def handler(signum: int, frame: Any) -> None:
     processor.run()
 
 
-def build_multistorage_streaming_strategy_factory(
+LogicalTopicTuple = Tuple[
+    Topic,
+    Optional[Topic],
+    Optional[Topic],
+    Optional[Callable[[], DeadLetterQueuePolicy]],
+]
```
lynnagara, line 265: Can you make this a NamedTuple or dataclass? How about calling this `ConsumerConfig` or something along those lines?

snuba/cli/multistorage_consumer.py
ayirr7: right, makes sense

snuba/cli/multistorage_consumer.py
```diff
@@ -289,30 +255,93 @@ def handler(signum: int, frame: Any) -> None:
     processor.run()
 
 
-def build_multistorage_streaming_strategy_factory(
+@dataclass
```
lynnagara, line 258: We generally freeze these if they are not supposed to be mutated
```suggestion
@dataclass(frozen=True)
```

lynnagara (approved):

PR #3731, opened 2023-02-10
fix(admin): allow querying dist nodes too

dbanda:
Fix a bug where we didn't recognize dist nodes as valid.

### DIFF
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -25,7 +25,11 @@ class InvalidStorageError(SerializableException):
 
 
 def is_valid_node(host: str, port: int, cluster: ClickhouseCluster) -> bool:
-    nodes = [*cluster.get_local_nodes(), cluster.get_query_node()]
+    nodes = [
+        *cluster.get_local_nodes(),
+        cluster.get_query_node(),
+        *cluster.get_distributed_nodes(),
+    ]
 
     return any(node.host_name == host and node.port == port for node in nodes)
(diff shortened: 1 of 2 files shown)


snuba/admin/clickhouse/common.py
```diff
@@ -25,7 +25,11 @@ class InvalidStorageError(SerializableException):
 
 
 def is_valid_node(host: str, port: int, cluster: ClickhouseCluster) -> bool:
-    nodes = [*cluster.get_local_nodes(), cluster.get_query_node()]
+    nodes = [
+        *cluster.get_local_nodes(),
+        cluster.get_query_node(),
+        *cluster.get_distributed_nodes(),
```
MeredithAnya, line 31: if there is no `__distributed_cluster_name` is this going to error for some clusters?

dbanda: Good point. It supposed to be that if its None then single_node must be True. But I think they are not doing this in Ops repo. Maybe we should change this assert to warn instead and return an empty list.

MeredithAnya: >  Maybe we should change this assert to warn instead and return an empty list.

yeah I think this makes sense to me

dbanda: Ok. changed it log and return an empty list. Also I set the log level to error rather than warn to bother people to fix their configuration.

MeredithAnya (approved):

PR #3737, opened 2023-02-13
Add missing generic metric consumers to freight

enochtangg:


onewland (approved):
Definitely deploy as soon as tests pass on this one, you don't want to surprise the next person if the freight config is wrong here somehow

PR #3742, opened 2023-02-14
ref(schemas): Add basic generic-metrics message schema

untitaker:
I used https://app.quicktype.io/ to infer a schema from some example
messages, and edited the schema manually to be actually correct.

I verified its contents with some sample messages in local development.
It would be better if we had a way to run schema validation against all
existing test payloads we have checked in. But typically those test
payloads exist as part of tests that directly invoke the
dataset-specific processor.

Also remove a CLI flag and replace it with a runtime config sample rate
setting. We are unsure about the performance impact and stability of schema
validation, and this will allow us to sample a few messages to get metrics from
prod + turn down to zero quickly if the consumer starts rejecting messages
incorrectly.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -1,5 +1,6 @@
 import itertools
 import logging
+import random
 import time
 from collections import defaultdict
 from datetime import datetime
@@ -551,13 +552,15 @@ def process_message(
     processor: MessageProcessor,
     consumer_group: str,
     snuba_logical_topic: SnubaTopic,
-    validate: bool,
+    validate: float,
     message: Message[KafkaPayload],
 ) -> Union[None, BytesInsertBatch, ReplacementBatch]:
     assert isinstance(message.value, BrokerValue)
     try:
         codec = get_json_codec(snuba_logical_topic)
-        decoded = codec.decode(message.payload.value, validate=validate)
+        decoded = codec.decode(
+            message.payload.value, validate=random.random() < validate
+        )
         result = processor.process_message(
             decoded,
             KafkaMessageMetadata(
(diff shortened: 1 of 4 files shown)


snuba/consumers/consumer_builder.py
```diff
@@ -264,7 +262,7 @@ def build_streaming_strategy_factory(
                 processor,
                 self.consumer_group,
                 logical_topic,
-                self.__validate_schema,
+                float(get_config("validate_schema", 0) or 0.0),
```
lynnagara, line 265: Let's support controlling the schema validation rate for each consumer independently?
```suggestion
                float(get_config(f"validate_schema_{storage_key.value}", 0) or 0.0),
```

snuba/consumers/consumer.py
```diff
@@ -551,13 +552,15 @@ def process_message(
     processor: MessageProcessor,
     consumer_group: str,
     snuba_logical_topic: SnubaTopic,
-    validate: bool,
+    validate: float,
```
lynnagara, line 555: nit: `validate_sample_rate` or something?

snuba/consumers/schemas.py
```diff
@@ -4,6 +4,62 @@
 
 from snuba.utils.streams.topics import Topic
 
+_HARDCODED_SCHEMAS: Mapping[Topic, Mapping[str, Any]] = {
+    Topic.GENERIC_METRICS: {
+        "$schema": "http://json-schema.org/draft-06/schema#",
```
lynnagara, line 9: This seems like quite an old version, should we try to use `2020-12` everywhere?

lynnagara (approved):

PR #3747, opened 2023-02-15
feat: Add new statuses for SLO

evanh:
This adds a new type of status that is more granular than the existing status.
The new status is added in parallel to the old one, so that the migration can
happen seemlessly without losing data. Once the old status isn't being used
anymore it can be removed and this code simplified.

The logic for determing the status has been consolidated to where the statuses
are defined to make it easier to keep track of the logic behind the different
values.

Once this is merged the corresponding SLOs can be set up in Datadog and checked.


snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
...
+    CLICKHOUSE_TIMEOUT = "clickhouse-timeout"
+    # Network isssues, SOCKET_TIMEOUT NETWORK_ERROR
+    NETWORK_TIMEOUT = "network-timeout"
+    # Query used too much memory, MEMORY_LIMIT_EXCEEDED
+    MEMORY_EXCEEDED = "memory-exceeded"
+    # Any other error
+    ERROR = "error"
+
+
+class SLO(Enum):
```
evanh, line 70: I don't love these names but I don't know what to call them. Open to suggestions.

snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
...
+    FOR = "for"  # These are not counted against our SLO
+    AGAINST = "against"  # These are counted against our SLO
+
+
+@dataclass(frozen=True)
+class NewStatus:
```
evanh, line 76: This can be renamed to `Status` once the old status has been deprecated.

tests/test_api_status.py
```diff
@@ -0,0 +1,113 @@
...
+                "memory-exceeded",
+                "for",
+            ),
+            (TimeoutError("test"), "cache-set-timeout", "against"),
+            (ExecutionTimeoutError("test"), "cache-wait-timeout", "against"),
+        ]
+
+        for exception, status, slo in tests:
+            execute_mock.side_effect = exception
+            self.post()
```
evanh, line 107: It might seem counter-intuitive to test something so far down the query pipeline with an API call like this, but it was just very cumbersome to try and manually build all the objects that `db_query` uses.

snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
...
+    status: SLOStatus
+    slo: SLO
```
lynnagara, line 78: Isn't the SLO value (for/against) largely implied by the SLO status? Like invalid request should always not count, etc?

snuba/querylog/__init__.py
```diff
@@ -30,6 +30,8 @@ def _record_timer_metrics(
         metrics,
         tags={
             "status": query_metadata.status.value,
+            "new_status": query_metadata.new_status.value,
```
lynnagara, line 33: Just a personal preference but I like to use the naming conventions status/legacy_status rather than new_status/status. It makes it easier to deprecate later as it's clearer what should be removed

snuba/querylog/query_metadata.py
lynnagara: Rather than having both, it it possible to add the new members to QueryStatus to bring it in line with this without having the 2 side by side?

snuba/querylog/query_metadata.py
evanh: Yes, this is a convenient way to store that knowledge without calculating/looking it up every time.

snuba/querylog/__init__.py
evanh: That's fair but I am sort of locked into the `status` name because that's what is used in Datadog. If I change the name then the existing SLO in Datadog will have zero data to start with.

snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
     TIMEOUT = "timeout"
 
 
+QUERY_ERROR_MAPPINGS = {
```
nikhars, line 25: Nit: Makes it easier to understand the purpose of these mappings
```suggestion
CLICKHOUSE_ERROR_TO_SNUBA_ERROR_MAPPINGS = {
```

snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
...
+    ErrorCodes.TOO_SLOW: QueryStatus.TIMEOUT,
+    ErrorCodes.TIMEOUT_EXCEEDED: QueryStatus.TIMEOUT,
+    ErrorCodes.SOCKET_TIMEOUT: QueryStatus.TIMEOUT,
+    ErrorCodes.NETWORK_ERROR: QueryStatus.TIMEOUT,
+}
+
+
+def get_query_status_from_error_codes(code: ErrorCodes) -> QueryStatus | None:
+    return QUERY_ERROR_MAPPINGS.get(code)
```
nikhars, line 34: Wouldn't this cause an exception if you pass in a Clickhouse error code which is not mapped. Should there be a default?

snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
...
+
+
+class SLOStatus(Enum):
```
nikhars, line 37: I agree the names could be a bit better. Personally, I would refrain from using terms like `SLO` in code altogether. `SLO` is a service concern rather than a software concern. The service concern would be derived from a software concern.
```suggestion
class RequestStatus(Enum):
```
or
```suggestion
class APIStatus(Enum):
```

snuba/querylog/query_metadata.py
```diff
@@ -15,6 +22,107 @@ class QueryStatus(Enum):
...
+    NETWORK_TIMEOUT = "network-timeout"
+    # Query used too much memory, MEMORY_LIMIT_EXCEEDED
+    MEMORY_EXCEEDED = "memory-exceeded"
+    # Any other error
+    ERROR = "error"
+
+
+class SLO(Enum):
+    FOR = "for"  # These are not counted against our SLO
+    AGAINST = "against"  # These are counted against our SLO
```
nikhars, line 72: Should we model these the way HTTP error codes are laid out? Error codes from ranges 400-499 are client side errors. Ranges 500-599 are server side errors.

In a similar vein, should we assign fixed range numbers for all product concerns and another fixed range numbers for all snuba concerns?

I believe that would make the overall system simpler and categorizing new errors in the future easier as well.

snuba/querylog/query_metadata.py
volokluev: ergonomics wise I would make `slo` a property

```python3
class NewSatus:
    status: SLOStatus
    
    @property
    def slo(self) -> SLO:
         # do you dictionary lookups here

**Review by volokluev — COMMENTED** [review]
This looks good to me but I agree with Nikhar about the naming

File: snuba/querylog/query_metadata.py (hunk 7 above)
Comment (evanh) [reply snuba/querylog/query_metadata.py:34]: No, `.get` will return None by default.

File: snuba/querylog/query_metadata.py (hunk 9 above)
Comment (evanh) [reply snuba/querylog/query_metadata.py:72]: I don't know how I feel about this idea. It does make it easier to calculate whether a status is for/against the SLO, but I'm not sure it makes it easier to maintain. If a status changes how it impacts the SLO, does that mean we change its number? Does that mean our data is inconsistent?

While this would get rid of some of the mappings, it would also mean keeping a mapping of every error to an arbitrary number.

File: snuba/querylog/__init__.py [hunk 10]
```diff
@@ -115,32 +117,43 @@ def _add_tags(
...
     # table would be useful.
     if settings.RECORD_QUERIES:
         timer.send_metrics_to(
             metrics,
-            tags={"status": status.value, "referrer": referrer or "none"},
+            tags={
+                "status": status.value,
+                "referrer": referrer or "none",
+                "new_status": new_status.status.value,
+                "slo": new_status.slo.value,
```
fpacifici, line 156: This is likely going to be a problem with our metrics cardinality, thus cost.
These tags are attached to all the marks in a query (every phase of the query has a metrics and each one of them will have all these tags). 
Each mark is then multiplied by Veneur. This is likely to multiply the snuba metrics 4 times and Snuba is the main offender today. 
I would suggest against applying the new status to all the marks of the query. This will probably require touching the Timer class.

snuba/querylog/query_metadata.py
fpacifici: `SUCCESS` / `FAIL` instead of FOR and AGAINST as they clearly mean which one is good and which one is bad ?

snuba/querylog/__init__.py
fpacifici: There is no `query_status` tag in datadog. That would be better than `new_status`

snuba/querylog/__init__.py
evanh: I don't think this gets sent for every mark. There is a separate `mark_tags` that gets sent for the marks, and this is separate from that. https://github.com/getsentry/snuba/blob/master/snuba/utils/metrics/timer.py#L92-L94

So I don't think I need to make a change there.

snuba/querylog/__init__.py
evanh: I'll change it to `request_status`.

snuba/querylog/query_metadata.py
evanh: I think I got caught up in the semantics. A request that doesn't count against our SLO might actually be more bad than one that does. For example, a completely invalid query vs. a table rate limited query.

nikhars (approved):
I am ok with the current changes

PR #3750, opened 2023-02-16
feat: Expose Snuba Admin via devserver

untitaker:
Will also open a PR for exposing this port by default in sentry
devservices


snuba/cli/devserver.py
```diff
@@ -35,6 +35,7 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
         os.execvp(daemons[0][1][0], daemons[0][1])
 
     daemons += [
+        (("admin", ["snuba", "admin"])),
```
dbanda, line 38: I know ST does not have the admin tool. This might fail for them.

PR #3757, opened 2023-02-17
feat(search-issues): alias isHandled and notHandled functions to None

barkbarkimashark:
Handle queries to this dataset that use `isHandled` and `notHandled` function names. For our use case, we don't have stacktraces to parse out so just return nothing.

### DIFF
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -142,6 +142,12 @@ storages:
             from_column_name: contexts
             to_nested_col_table:
             to_nested_col_name: contexts
+      functions:
+        - mapper: DefaultNoneFunctionMapper
+          args:
+            function_names:
+              - isHandled
+              - notHandled
 
 storage_selector:
   selector: DefaultQueryStorageSelector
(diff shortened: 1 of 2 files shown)


evanh (commented):
Why are those functions being called on your dataset in the first place? Why wouldn't you make a change in Sentry to stop querying with that function?

barkbarkimashark:
> Why are those functions being called on your dataset in the first place? Why wouldn't you make a change in Sentry to stop querying with that function?

Yeah, this solution isn't ideal but we have a lot of common search code we're leveraging and re-writing everything would be prohibitively expensive. We'll look to do some refactoring in the future.

evanh (changes requested):
If you want to add tech debt to be removed later, you could add a check to Sentry that does the same thing as this. That would also make it easier to refactor later since all the code will be one place.

I didn't mean to reject this change outright and I can't fix that without approving. I do think though that if there is going to be tech debt around how queries are constructed, it would be better to keep that tech debt  in one place.

fire

barkbarkimashark:
> If you want to add tech debt to be removed later, you could add a check to Sentry that does the same thing as this. That would also make it easier to refactor later since all the code will be one place.
> 
> I didn't mean to reject this change outright and I can't fix that without approving. I do think though that if there is going to be tech debt around how queries are constructed, it would be better to keep that tech debt in one place.
> 
> fire [ fire ](https://user-images.githubusercontent.com/2727124/220457746-a965f27f-a068-453f-b2a5-9c7d9352aade.gif) [ ](https://user-images.githubusercontent.com/2727124/220457746-a965f27f-a068-453f-b2a5-9c7d9352aade.gif)

No, that's fair. After thinking about it some more I think we can add some pre-validation step when constructing the query for this dataset that shouldn't require a huge change to the search code.

PR #3759, opened 2023-02-17
feat(replays): Default event_hash when segment_id is null

cmanallen:
The primary-key of the replays table includes segment_id as a component.  When segment_id is null we need to provide a default event_hash value to prevent row merges.  Previously `None` was coerced to a UUID which was deterministic.

### DIFF
--- a/snuba/datasets/processors/replays_processor.py
+++ b/snuba/datasets/processors/replays_processor.py
@@ -161,7 +161,11 @@ class ReplaysProcessor(DatasetMessageProcessor):
     def _process_event_hash(
         self, processed: MutableMapping[str, Any], replay_event: ReplayEventDict
     ) -> None:
-        processed["event_hash"] = segment_id_to_event_hash(replay_event["segment_id"])
+        event_hash = replay_event.get("event_hash")
+        if event_hash is None:
+            event_hash = segment_id_to_event_hash(replay_event["segment_id"])
+
+        processed["event_hash"] = event_hash
 
     def process_message(
         self, message: Mapping[Any, Any], metadata: KafkaMessageMetadata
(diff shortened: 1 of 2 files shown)


lynnagara:
Would it make sense to do this in the replays consumer upstream of Snuba that produces this topic? That way it is the same in the replays data you are putting into file storage?

cmanallen:
> That way it is the same in the replays data you are putting into file storage?

Our null segments are a special case.  They're not associated with a file.  They come from user requests or internal updates.  Our delete behavior uses this.  We issue a null segment and set is-archived to true to "delete" the row.

snuba/datasets/processors/replays_processor.py
```diff
@@ -211,6 +204,19 @@ def process_message(
 U = TypeVar("U")
 
 
+def segment_id_to_event_hash(segment_id: int | None) -> str:
+    if segment_id is None:
+        # Rows with null segment_id fields are considered "out of band" meaning they do not
+        # originate from the SDK and do not relate to a specific segment.
+        #
+        # For example: archive requests.
+        return str(uuid.uuid4().hex)
```
dbanda, line 213: It seems a little unusual to assign a random id to the message here. If somehow a message gets processed twice, the same message would have different ids. You lose out on any benefit of clickhouse deduplication. Is this acceptable? I'm wondering if it is better to do the id assignment from the source so once a message is created it has a unique id for life.

cmanallen: @dbanda Updated to allow for setting event_hash explicitly.  Missing event hashes with null segment-id will still generate a random uuid.

evanh (approved):

PR #3765, opened 2023-02-21
fix(oncall): error on non-existent columns

volokluev:
Queries get sent to us with non-existent columns in them. No need to let those queries get all the way to clickhouse. Turn on the `EntityContainsColumnsValidator` to `ERROR` for all entities by default.

As part of this I moved the test for this validator to its own file and decoupled it from the outcomes entity (it now uses its a test configuration entity)

### DIFF
--- a/snuba/datasets/configuration/outcomes/entities/outcomes.yaml
+++ b/snuba/datasets/configuration/outcomes/entities/outcomes.yaml
@@ -42,4 +42,3 @@ validators:
   args:
     required_filter_columns:
     - org_id
-validate_data_model: warn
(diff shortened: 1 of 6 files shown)


evanh (commented):
I believe outcomes and maybe a couple other entities have the value explicitly set to WARN, can we change those to ERROR as well?

enochtangg (commented):
Does this solution also work for columns in the `WHERE` clause? Additionally, how about columns wrapped inside a function (e.g. `isNull(asdf)` = 1)

volokluev:
@enochtangg it does work, I added a test for it

PR #3766, opened 2023-02-21
fix: Fix a few bugs with message schema validation

untitaker:
* The defined schema was not for the generic-metrics topic, but for the
  metrics topic. An important difference is that the tag values are
  strings!

* The config was not reloaded dynamically, only on consumer start

* Validation errors crashed the consumers, but in this stage we don't
  want that


lynnagara:
Looks like the test failures are real

lynnagara (approved):

PR #3774, opened 2023-02-22
fix(oncall): remove useless validation

volokluev:
We do not have the resources (or need) to police every kafka topic that gets created and used with snuba. Remove this validation because it is not useful and causes crashloops whenever a new topic is added and we forgot to add it to this list


fpacifici (changes requested):
Please do not merge this. 
This validation was introduced after incidents caused by mistyped topics in the prod settings. 

There was a ticket to add a unit test that would check that the enum and this list matched. I am confident it was assigned at a point, but probably not compelted.

PR #3776, opened 2023-02-23
Add `loadbalancer-outcomes` to the list of valid topics

oioki:
I'd like to temporarily override the cluster for `loadbalancer-outcomes` topic via `KAFKA_BROKER_CONFIG`. However, according to [this logic](https://github.com/getsentry/snuba/blob/master/snuba/settings/validation.py#L80-L82) it would fail currently, because this topic is missing from `topic_names`.

### DIFF
--- a/snuba/settings/validation.py
+++ b/snuba/settings/validation.py
@@ -39,6 +39,7 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
         "cdc",
         "snuba-metrics",
         "outcomes",
+        "loadbalancer-outcomes",
         "ingest-sessions",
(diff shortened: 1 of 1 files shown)


lynnagara:
This shouldn't be needed. Overrides as specified in `KAFKA_BROKER_CONFIG` as `{"outcomes": "loadbalancer-outcomes"}` where the key is the logical/default name and the value is whatever you want to override it to. Only the keys are validated, override names shouldn't need to be added here.

oioki:
We'll go with `--bootstrap-server` option instead.

PR #3777, opened 2023-02-23
ref: Add more topic schemas to Snuba

untitaker:
The kafka schema registry isn't yet releaseable so in order to test out
performance, add a few more schemas here instead.

Also tag schema validation related metrics by topic.


lynnagara:
Looks like the tests are complaining about the uint64 type in the events schema

lynnagara (approved):

PR #3778, opened 2023-02-23
ref(consumer): Aligning the two consumers' CLI

ayirr7:


snuba/cli/consumer.py
```diff
@@ -42,15 +52,6 @@
     multiple=True,
     help="Kafka bootstrap server to use.",
 )
-@click.option(
-    "--storage",
```
lynnagara, line 46: This is going to break all current Snuba consumers in dev as well as prod since they are all passing `--storage` right now. Can we put a migration plan in place?

snuba/cli/consumer.py
```diff
@@ -133,6 +134,8 @@ def consumer(
     setup_logging(log_level)
     setup_sentry()
     logger.info("Consumer Starting")
+
+    storage_name = storage_names[0]
```
lynnagara, line 138: If you are (temporarily) just picking the first from the list and discarding the rest, please raise an exception if any more than 1 is passed. It may be very unexpected behavior.

snuba/cli/consumer.py
ayirr7: Oh oops...I can do that. Maybe we can sync offline about this as well

untitaker: just in case it's useful, a click option can have multiple aliases: `@click.option("--storage", "--storages")`

lynnagara: Looking at this again, I think it should probably stay singular anyway.

- It is also `--storage` on the multistorage consumer as well so making it singular would align to that
- `snuba consumer --storage errors --storage errors2` reads a lot better than `snuba consmer --storages errors --storages errors2`

lynnagara (approved):

PR #3780, opened 2023-02-23
fix(oncall): validate metrics counters tag condition

volokluev:
Had a bunch of queries come in to sentry that triggered an alert because non-string keys were being used on tags


evanh (commented):
This won't work like this, the TagConditionValidator assumes you want to compare strings to strings, not ints to ints. It will have to be modified/extended to handle that case.

volokluev:
thanks for letting me know

PR #3782, opened 2023-02-23
fix(ast): Apply visitor to LIMIT BY clause in transform

john-z-yang:
### Overview
- The AST for SnQL does not apply the visitor to the `LIMIT BY` clause in a `transform` call. Causing post-parse optimization steps like `expand_aliases` to not apply to the LIMIT BY clause.

### Changes
- Apply the visitor to the `LIMIT BY` clause
- Introduce new tests to verify `LIMIT BY` is visited by the visitor

Closes [SNS-1891](https://getsentry.atlassian.net/browse/SNS-1891)

[SNS-1891]: https://getsentry.atlassian.net/browse/SNS-1891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snuba/query/__init__.py
```diff
@@ -377,6 +377,11 @@ def transform(self, visitor: ExpressionVisitor[Expression]) -> None:
                 self.__order_by,
             )
         )
+        if self.__limitby is not None:
+            self.__limitby = LimitBy(
+                self.__limitby.limit,
+                [expression.accept(visitor) for expression in self.__limitby.columns],
+            )
```
lynnagara, line 384: Was there a reason why you couldn't call transform on the whole limit by clause?

i.e. `self.__limitby = self.__limit_by.transform(func) if self.__limitby else None`

john-z-yang: My understanding is that `Limitby` is a dataclass where only the [columns are expressions](https://github.com/getsentry/snuba/blob/master/snuba/query/__init__.py#L35), and the visitor applies transformation to only expressions.
```
@dataclass(frozen=True)
class LimitBy:
    limit: int
    columns: Sequence[Expression]
```

lynnagara (approved):
LGTM. Just keep a close eye on the query/failure rate when this gets deployed.

tests/query/test_query.py
```diff
@@ -77,3 +80,37 @@ def test_query_experiments() -> None:
 
     query.set_experiments({"optimization3": 0.5})
     assert query.get_experiments() == {"optimization3": 0.5}
+
+
+@patch("snuba.query.replace")
+def test_query_transformation(mock_replace: MagicMock) -> None:
```
volokluev, line 86: Why are there so many mocks in this test? It seems like you are testing the implementation rather than the desired behavior. All the changes in this PR are local to this repo. You should be able to test your change without any mocks

volokluev (changes requested):

volokluev (approved):
John and I talked about the test. It's okay, just needs to communicate its intent a little bit in the comments, otherwise good job

PR #3783, opened 2023-02-24
feat(CapMan): `NO_LIMIT` keyword to skip limiting

rahul-kumar-saini:
All this does is skip counting towards the metric for now, maybe this can become a keyword used to ignore or at least request skipping allocation management for certain requests


evanh (commented):
Why are you adding this? We generally don't allow any workarounds for rate limiting/caching etc.

Why not add a `subscription` referrer and track subscriptions that way?

volokluev:
I agree with evan, this is not how we want to do this

PR #3795, opened 2023-02-27
ref(CapMan): remove execute_query_strategy variable

volokluev:
in order to make room for capacity management, we have to understand how db_query.py works. As part of that process, I realized that we had dead code in this file `execute_query_with_caching` has not been used since 2020 and we're not going to use it again.

@untitaker actually tried to remove this in #3372 but tests were failing. 

### Significant changes

added a short circuit option to the readthrough cache. This was done for the following reasons:

1. The test that was failing was failing because it was hitting the readthrough cache, which never hit the `execute_query` function, thus never setting the `consistent` flag in the results. 
2. We had a single tenant incident where we wanted to disable the reathrough cache and were not able to so I figured it was fine to add it. This way if we lose redis for a consistent amount of time we can just pass the queries right through.

### DIFF
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -81,6 +81,11 @@ class RedisCache(Cache[TValue]):
         timeout: int,
         timer: Optional[Timer] = None,
     ) -> TValue:
+        # in case something is wrong with redis, we want to be able to
+        # disable the read_through_cache but still serve traffic.
+        if get_config("read_through_cache.short_circuit", 0):
+            return function()
+
         # This method is designed with the following goals in mind:
         # 1. The value generation function is only executed when no value
         # already exists for the key.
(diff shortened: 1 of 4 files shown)


untitaker (approved):
i guess that's a way to get rid of the test failure 👍

evanh (commented):
I have the same question for you as for Markus: is there a way for a self-hosted/single tenant customer to have things break when this change is made? What if a customer was using the old cache before?

volokluev:
>  is there a way for a self-hosted/single tenant customer to have things break when this change is made? What if a customer was using the old cache before?

I think they will be okay. If a query comes in that was in the old cache but was switched to the new cache, it just won't be cached and will go straight to clickhouse. In fact we did try and switch back to the old cache and back to the readthrough cache during an incident and there were no serious issues to speak of.

PR #3796, opened 2023-02-28
rust: add simple transform and produce strategy example

dbanda:
Adds the produce strategy and a simple transform -> produce example. 

The produce strategy is an arroyo strategy that produces messages on a given topic.

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
+++ b/rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
@@ -1,4 +1,7 @@
 
+use futures::Future;
+use futures::executor::block_on;
+use futures::stream::FuturesOrdered;
 use log::warn;
 use crate::backends::Producer;
 use crate::backends::kafka::producer::KafkaProducer;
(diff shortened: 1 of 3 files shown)


rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub producer: KafkaProducer,
+    pub next_step: Box<dyn ProcessingStrategy<TPayload>>,
+    queue: VecDeque<Message<TPayload>>,
+    topic: TopicOrPartition,
+    closed: bool,
+    max_queue_size: usize,
+}
+
+// TODO: make the queue store futures
+// TODO: make this generic
```
lynnagara, line 23: Does it actually need to be generic? i.e. Any reason to take anything other than a KafkaPayload here?

rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
```diff
@@ -0,0 +1,160 @@
...
+        return None
+    }
+
+    fn submit(&mut self, message: Message<KafkaPayload>) -> Result<(), MessageRejected> {
+        if self.queue.len() >= self.max_queue_size {
+            return Err(MessageRejected)
+        }
+        print!("Producing message: {:?}", &message.payload.payload);
+        self.producer.produce(&self.topic, &message.payload);
+        self.producer.flush();
```
lynnagara, line 55: `submit` is supposed to be a non blocking method, here we're waiting indefinitely for the message to get flushed

rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
```diff
@@ -0,0 +1,160 @@
...
+        self.queue.push_back(message);
+        return Ok(());
+    }
+
+    fn close(&mut self) {
+        self.closed = true;
```
lynnagara, line 61: `self.closed` isn't being used anywhere

lynnagara (commented):
Shall we decide how we want to deal with the asynchronous produce step?

A couple of option I can see:

1 - Have this strategy create the producer inside here instead of getting it passed in from outside by the user of this strategy. This would allow us to pass it the delivery callbacks we need so that we can get notified when messages have been successfully produced.

2 - Change producer.produce() to return a future, and avoid those callbacks.

What do you think?

untitaker (approved):
just gonna approve so we can merge it for now, i havent' read it in detail

dbanda:
> Shall we decide how we want to deal with the asynchronous produce step?
> 
> A couple of option I can see:
> 
> 1 - Have this strategy create the producer inside here instead of getting it passed in from outside by the user of this strategy. This would allow us to pass it the delivery callbacks we need so that we can get notified when messages have been successfully produced.
> 
> 2 - Change producer.produce() to return a future, and avoid those callbacks.
> 
> What do you think?

My preference would be for 2. Passing callbacks in rust seems a little messy, but I not sure what the best practices are for using futures cause that depends on `tokio` probably.

rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
dbanda: removed flush and made it place a future on the queue.

rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
dbanda: added a `panic` check to submit if closed is true

rust_snuba/rust_arroyo/src/processing/strategies/produce.rs
dbanda: Ur right. removed

PR #3797, opened 2023-02-28
migrations : add test to check that reverse has no side effects

dbanda:
In order for migrations to run in an automated pipeline, the reverse function must be idempotent. This is because Snuba doesn’t not keep per node state; it is either all or nothing. In the case that a failure occurs, a pipeline may call reverse several times on the same node until it is sure that all nodes have reversed the migration. Running reverse must be idempotent and must not have unintended side effects.

This test runs reverse on migrations multiple times and checks the clickhouse create table query statements to check that DDL are not changed on multiple runs.

### DIFF
--- a/tests/migrations/test_runner.py
+++ b/tests/migrations/test_runner.py
@@ -209,6 +209,56 @@ def test_reverse_all() -> None:
     ), "All tables should be deleted"
 
 
+def test_reverse_idempotency_all() -> None:
+    # This test is to ensure that reversing a migration twice does not cause any
+    # issues or unintended side effects. This is important because we may need to reverse
+    # a migration multiple times in the event of a rollback.
+    runner = Runner()
+    all_migrations = runner._get_pending_migrations()
+    runner.run_all(force=True)
+    connection = get_cluster(StorageSetKey.MIGRATIONS).get_query_connection(
+        ClickhouseClientSettings.MIGRATE
+    )
+    for migration in reversed(all_migrations):
+        runner.reverse_migration(migration, force=True)
+        if migration.group != MigrationGroup.SYSTEM:
+
+            def reverse_twice() -> None:
+                # reverse again to ensure idempotency
+                runner.run_migration(migration, fake=True)
+                runner.reverse_migration(migration, force=True)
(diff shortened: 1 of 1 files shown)


tests/migrations/test_runner.py
```diff
@@ -209,6 +209,56 @@ def test_reverse_all() -> None:
...
+            def reverse_twice() -> None:
+                # reverse again to ensure idempotency
+                runner.run_migration(migration, fake=True)
+                runner.reverse_migration(migration, force=True)
+                runner.run_migration(migration, fake=True)
+                runner.reverse_migration(migration, force=True)
+
+            group = migration.group.value
+
+            # for some reason this has to be here. it looks like
```
evanh, line 235: It would  be good to understand what's happening here.

tests/migrations/test_runner.py
```diff
@@ -209,6 +209,56 @@ def test_reverse_all() -> None:
...
+    runner.run_all(force=True)
+    connection = get_cluster(StorageSetKey.MIGRATIONS).get_query_connection(
+        ClickhouseClientSettings.MIGRATE
+    )
+    for migration in reversed(all_migrations):
+        runner.reverse_migration(migration, force=True)
+        if migration.group != MigrationGroup.SYSTEM:
+
+            def reverse_twice() -> None:
+                # reverse again to ensure idempotency
```
evanh, line 227: I thought the test was to run the reverse twice, e.g. `run, reverse, reverse`, instead of `run, reverse, run, reverse`?

tests/migrations/test_runner.py
lynnagara: Guessing it's because some other tests are doing stuff like this: https://github.com/getsentry/snuba/blob/3cff084ed27b6cb013437f8e9cb1add1f778d847/tests/clusters/test_cluster.py#L107

tests/migrations/test_runner.py
dbanda: The run has `fake=True`. So it doesn't do anything but mark the migrations table so that reverse is not rejected immediately.

lynnagara (approved):

PR #3801, opened 2023-02-28
ref(ratelimit): Create separate namespace if config redis for rate limit keys

evanh:
Motivation:

Currently all of the configuration for rate limiting is done alongside other
runtime configuration. This was fine when not much rate limiting was done, but
now rate limit keys are the dominant keys in the configuration tool.

Ultimately there should be a separate tool for configuring rate limiting. That
simplifies the user interface and allows product owners access to rate limits
without having to grant access to configuration in general.

Blast Radius:

This change adds a new config key hash to redis to store the rate limiting keys.
This still uses the configuration mechanism under the hood, largely to leverage
the audit log that state changes allow.

This is step one of the migration process, doing dual read/writes on the keys
when they are accessed. This will be followed up with a migration to move all
the old keys over automatically.

Notes:

The new rate limit keys will not show up in the config tool, so follow up
changes will be added to expose them.

### DIFF
--- a/snuba/state/__init__.py
+++ b/snuba/state/__init__.py
@@ -7,10 +7,10 @@ from dataclasses import dataclass
 from functools import partial
 from typing import (
     Any,
-    Callable,
     Iterable,
     Mapping,
     Optional,
+    Protocol,
     Sequence,
     SupportsFloat,
     Tuple,
@@ -42,6 +42,7 @@ config_history_hash = "snuba-config-history"
 config_changes_list = "snuba-config-changes"
 config_changes_list_limit = 25
 queries_list = "snuba-queries"
+rate_limit_config_hash = "snuba-ratelimit-config:"
 
 # Rate Limiting and Deduplication
 
@@ -106,22 +107,31 @@ def get_rates(bucket: str, rollup: int = 60) -> Sequence[Any]:
 # Runtime Configuration
 
 
+class ConfigKeyCallable(Protocol):  # Necessary for typing the memoize
+    def __call__(self, config_key: str = config_hash) -> Any:
+        pass
+
+
 class memoize:
     """
-    Simple expiring memoizer for functions with no args.
+    Simple expiring memoizer for state functions that takes a config key.
     """
 
     def __init__(self, timeout: int = 1) -> None:
         self.timeout = timeout
-        self.saved = None
-        self.at = 0.0
+        self.saved: dict[str, Any] = {}
+        self.at: dict[str, float] = {}
 
-    def __call__(self, func: Callable[[], Any]) -> Callable[[], Any]:
-        def wrapper() -> Any:
+    def __call__(self, func: ConfigKeyCallable) -> ConfigKeyCallable:
+        def wrapper(config_key: str = config_hash) -> Any:
             now = time.time()
-            if now > self.at + self.timeout or self.saved is None:
-                self.saved, self.at = func(), now
-            return self.saved
+            at = self.at.get(config_key, 0.0)
+            if now > at + self.timeout or config_key not in self.saved:
+                self.saved[config_key], self.at[config_key] = (
+                    func(config_key=config_key),
+                    now,
+                )
+            return self.saved[config_key]
 
         return wrapper
 
@@ -147,12 +157,12 @@ def set_config(
     value: Optional[Any],
     user: Optional[str] = None,
     force: bool = False,
+    config_key: str = config_hash,
 ) -> None:
     value = get_typed_value(value)
     enc_value = "{}".format(value).encode("utf-8") if value is not None else None
-
     try:
-        enc_original_value = rds.hget(config_hash, key)
+        enc_original_value = rds.hget(config_key, key)
         if enc_original_value is not None and value is not None:
             original_value = get_typed_value(enc_original_value.decode("utf-8"))
             if value == original_value and type(value) == type(original_value):
@@ -164,10 +174,10 @@ def set_config(
         change_record = (time.time(), user, enc_original_value, enc_value)
         p = rds.pipeline()
         if value is None:
-            p.hdel(config_hash, key)
+            p.hdel(config_key, key)
             p.hdel(config_history_hash, key)
         else:
-            p.hset(config_hash, key, enc_value)
+            p.hset(config_key, key, enc_value)
             p.hset(config_history_hash, key, json.dumps(change_record))
         p.lpush(config_changes_list, json.dumps((key, change_record)))
         p.ltrim(config_changes_list, 0, config_changes_list_limit)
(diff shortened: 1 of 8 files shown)


snuba/state/rate_limit.py
```diff
@@ -27,6 +31,57 @@
 rds = get_redis_client(RedisClientKey.RATE_LIMITER)
 
 
+def get_rate_limit_config(
+    per_second: tuple[str, float | None], concurrent: tuple[str, int | None]
+) -> tuple[Any, Any]:
```
volokluev, line 36: is the typing on this right? Doesn't seem like it should be `Any`

snuba/state/rate_limit.py
```diff
@@ -27,6 +31,57 @@
...
+
+def set_rate_limit_config(bucket: str, value: float | int | None) -> None:
+    """
+    This function to encapsulate how rate limit keys are set in Redis, since
+    that is conceptually a different process from fetching normal config keys.
+    """
+    set_config(
+        bucket,
+        value,
+        config_key=state.rate_limit_config_hash,
```
volokluev, line 81: if you're setting with `config_key=state.rate_limit_config_hash` but your `get_rate_limit_config` defaults to old values, are you not going to override whatever you set after you deploy this PR? Maybe it would make sense to write to both?

tests/query/processors/test_org_rate_limiter.py
```diff
@@ -102,3 +102,33 @@ def test_org_rate_limit_processor_overridden(
     assert rate_limiter.bucket == str(org_id)
     assert rate_limiter.per_second_limit == 5
     assert rate_limiter.concurrent_limit == 10
+
+
+def test_namespaced_rate_limit() -> None:
```
volokluev, line 107: I don't really understand why this test passes. You're writing a config without a namespace, reading it with a namespace and somehow it works but I'm not sure why

tests/state/test_rate_limit.py
```diff
@@ -151,8 +155,10 @@ def test_rate_limit_exceptions(self) -> None:
         bucket = "{}{}".format(state.ratelimit_prefix, params.bucket)
 
         def count() -> int:
-            return get_redis_client(RedisClientKey.RATE_LIMITER).zcount(
-                bucket, "-inf", "+inf"
+            return int(
```
volokluev, line 158: just a mypy error I presume?

tests/state/test_rate_limit.py
```diff
@@ -207,3 +215,59 @@ def test_rate_limit_failures(vals: Tuple[int, int, int], rate_limit_shards) -> N
...
+        set_rate_limit_config(ct_key, ct_in_new)
+    if ps_in_old is not None:
+        state.set_config(ps_key, ps_in_old)
+    if ct_in_old is not None:
+        state.set_config(ct_key, ct_in_old)
+
+    ps_found, ct_found = get_rate_limit_config((ps_key, 76), (ct_key, 77))
+
+    expected_ps_saved = ps_in_old
+    expected_ps_found = ps_in_old or 76
```
volokluev, line 259: nit: defining default variables makes this more readable, e.g. 

```python3
DEFAULT_PS_LIMIT = 76
DEFAULT_C_LIMIT = 77
```

snuba/state/rate_limit.py
volokluev: This function signature seems to bake in the assumption that every rate limit is both concurrent and per-second. Maybe that's fine but just something I wanted to throw out there as unnecessarily limiting

snuba/state/rate_limit.py
```diff
@@ -27,6 +31,57 @@
...
+        old_concurrent if old_concurrent is not None else concurrent_default
+    )
+
+    return (found_per_second, found_concurrent)
+
+
+def set_rate_limit_config(bucket: str, value: float | int | None) -> None:
+    """
+    This function to encapsulate how rate limit keys are set in Redis, since
+    that is conceptually a different process from fetching normal config keys.
```
volokluev, line 76:
```suggestion
    that is conceptually a different process from writing normal config keys.
```

volokluev (commented):
nice job, got some questions

snuba/state/rate_limit.py
evanh: I'm not sure I follow which case you're referring to here?

tests/query/processors/test_org_rate_limiter.py
evanh: Because the test should copy the value to the new namespace.

1) Write a value to the old namespace (no config key)
2) Run the test, which should copy the value to the new namespace
3) Check the value was copied (rate limit config key)

tests/state/test_rate_limit.py
evanh: Yep.

snuba/state/__init__.py
```diff
@@ -42,6 +42,7 @@
 config_changes_list = "snuba-config-changes"
 config_changes_list_limit = 25
 queries_list = "snuba-queries"
+rate_limit_config_hash = "snuba-ratelimit-config:"
```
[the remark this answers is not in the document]
evanh: That's what I thought but this appears to be the convention (see `config_hash`). I'll change it.

snuba/state/rate_limit.py
evanh: I decided to keep this opinionated for now, since it lets me keep the code cleaner. If we start to encounter edge cases where this needs to be less restrictive we can change it then.

enochtangg (commented):
Looks good! One thing I'd want to double check (although unlikely), does Redis set a default TTL for the new keys we are writing.

tests/query/processors/test_org_rate_limiter.py
```diff
@@ -102,3 +102,33 @@ def test_org_rate_limit_processor_overridden(
...
+        ),
+    )
+    org_id = 1
+    settings = HTTPQuerySettings()
+    ps_key = f"org_per_second_limit_{org_id}"
+    ct_key = f"org_concurrent_limit_{org_id}"
+    state.set_config(ps_key, 5)
+    state.set_config(ct_key, 10)
+
+    OrganizationRateLimiterProcessor("org_id").process_query(query, settings)
```
volokluev, line 124: I think this comment would help make clear what the test is testing. It took me a few re-reads to understand it otherwise

```suggestion
    # By running the rate limiter processor, the keys should be ported over to the 
    # rate_limit_config_hash namespace
    OrganizationRateLimiterProcessor("org_id").process_query(query, settings)
```

snuba/state/rate_limit.py
volokluev: Here's a test that fails which I think should pass:

```python3
def test_overwrite_new_with_old():
    ps_key = "project_per_second_limit_36"
    ct_key = "project_concurrent_limit_36"

    set_rate_limit_config(ps_key, 123)
    set_rate_limit_config(ct_key, 321)

    ps_found, ct_found = get_rate_limit_config((ps_key, 76), (ct_key, 77))
    assert ps_found == 123
    assert ct_found == 321
```


After re-reading this PR, I understand what is happening but it is pretty confusing. As I read this module, I see two public functions:

```
def get_rate_limit_config
def set_rate_limit_config
```

As a reader, I assume that rate limits are retrieved by `get_rate_limit_config` and set by `set_rate_limit_config`.

That is not actually true with the code in this PR. `set_rate_limit_config` is an internal function called in this module which sets the configuration with a namespace, however the way rate limits are _actually_ set is by calling `state.set_config` without a `config_key`, thus putting it by default into the old default namespace. I understand this is maybe an interim state and maybe we could document it, but I think it would just be easier if this module's getters and setters worked symmetrically.

evanh: Oh that's a good point. I'll change that.

evanh:
> does Redis set a default TTL for the new keys we are writing

It does not, although good question.

volokluev (approved):

PR #3802, opened 2023-02-28
feat(rust-consumer): Add yaml settings file

lynnagara:
Adds a yaml file with the minimal settings needed to run the Rust consumer (the Kafka broker config and the ClickHouse cluster mapping).

These values are essentially duplicated from the Python settings file. If Rust and Python code were being run side by side, these would need to be kept in sync. Though since only consumers would ever be in Rust, not every setting will be duplicated here, just a few.

An alternative approach would be to attempt to load the Python file in Rust so that the values would be defined only once. This option was not pursued as it is messier and likely not the right long term direction for settings anyway. It has been a long time goal in Snuba to move away from the Python specific settings to something language agnostic that is easier for end users (including open source users) to configure. For this reason, I wanted to propose starting fresh with yaml settings rather than attempting to share the Python ones.

### DIFF
--- a/rust_snuba/Cargo.toml
+++ b/rust_snuba/Cargo.toml
@@ -14,4 +14,6 @@ path = "src/bin/consumer/querylog_consumer.rs"
 
 [dependencies]
 rust_arroyo = { path = "./rust_arroyo" }
+serde = { version = "1.0", features = ["derive"]}
+serde_yaml = "0.9"
 tokio = { version = "1.19.2", features = ["full"] }
--- a/rust_snuba/src/main.rs
+++ b/rust_snuba/src/main.rs
@@ -1,0 +1,1 @@
+mod settings;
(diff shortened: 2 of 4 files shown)


volokluev:
why not 

1. move the python settings into yaml
2. load the yaml settings in python
3. load the yaml settings in rust?

lynnagara:
> why not
> 
> 1. move the python settings into yaml
> 2. load the yaml settings in python
> 3. load the yaml settings in rust?

This is exactly what this PR is proposing long term, but it won't all be done in this one PR. Updating the Python code would be done as a follow up. This is only used for dev right now, so keeping them in sync is not a big concern. For now, all this PR seeks to do is build consensus on whether we should move to yaml for settings long term, or pursue the other main alternative: loading the Python settings as-is in Rust.

dbanda (approved):

lynnagara:
Superceded by https://github.com/getsentry/snuba/pull/3856

PR #3804, opened 2023-03-01
fix(subscriptions): Fix sessions subscriptions issue #3696

klboke:
@lynnagara Hi 

https://github.com/getsentry/snuba/pull/3697  did not fully restore the sessions subscription, try to fix it  #3696

### Legal Boilerplate


lynnagara (approved):
This seems ok to me to turn this back on. Afaik it's not running in Sentry prod anyway. But it may affect some ST customers. @onewland @jan-auer @getsentry/open-source, @getsentry/single-tenant, any concerns?

lynnagara:
Discussed offline, the migration over to metrics won't happen for a while so we need to unbreak sessions subscriptions for self hosted users.

PR #3807, opened 2023-03-01
fix(gocd): Automatically deploy after CI passes

untitaker:
There's a real discussion to be had as to which behavior is better, but
this is currently closer to how Freight behaved.


volokluev (approved):
👑

evanh (approved):

PR #3812, opened 2023-03-02
ref: Start using python types from sentry-kafka-schemas

untitaker:
after https://github.com/getsentry/sentry-kafka-schemas/pull/24, we have python types generated from schemas. those schemas are being validated in production at the moment

also requires https://github.com/getsentry/sentry-kafka-schemas/pull/29 to be merged

**NOTE**: Before deploying this, deploy the bump to sentry-kafka-schemas to ensure that the updated event schema passes validation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -381,7 +381,7 @@ jobs:
       - name: Checkout Snuba
         uses: actions/checkout@v3
       - name: Run Sentry self-hosted e2e CI
-        uses: getsentry/action-self-hosted-e2e-tests@627c9b50cb168c546ed6e2c461dae76d5d0c7cdb
+        uses: getsentry/action-self-hosted-e2e-tests@77805295ebff8a603def5970e18743ded72cb304
         with:
           project_name: snuba
           docker_repo: getsentry/snuba
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ python-rapidjson==1.8
 pytz==2022.2.1
 redis==4.3.4
 sentry-arroyo[json]==2.7.1
-sentry-kafka-schemas==0.0.3
+sentry-kafka-schemas==0.0.4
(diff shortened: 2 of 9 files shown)


snuba/datasets/processors/errors_processor.py
```diff
@@ -193,6 +218,8 @@ def extract_custom(
         output["type"] = _unicodify(data.get("type", ""))
         output["title"] = _unicodify(data.get("title", ""))
 
+        return contexts
```
evanh, line 221: Why did you change this return the contexts?

snuba/datasets/processors/querylog_processor.py
```diff
@@ -70,15 +75,16 @@ def __extract_query_list(
...
             consistent.append(int(query["stats"].get("consistent") or 0))
-            profile = query.get("profile") or {
+            fallback_profile: ClickhouseQueryProfile = {
                 "time_range": 0,
                 "all_columns": [],
                 "multi_level_condition": False,
                 "where_profile": {"columns": [], "mapping_cols": []},
                 "groupby_cols": [],
                 "array_join_cols": [],
-                "bytes_scanned": 0,
```
evanh, line 80: Why was this removed?

untitaker: because we do not need it later. it also seems that we have not encountered this value in prod when the dict is present

evanh: I believe the plan was to add this value, but the migrations necessary didn't happen. We can always add it back I guess.

evanh (approved):

snuba/datasets/processors/querylog_processor.py
untitaker: bytes scanned is extracted from a different subkey later on, so i have no idea if this was even correct to include in the first place

lynnagara (approved):

PR #3825, opened 2023-03-06
fix(profiling): Use correct field as origin timestamp for functions

phacops:
We're currently using the client's timestamp (start of the profile) for the origin timestamp and it's producing wild results (like suggesting we took 11 days to ingest some values).

It'd be better to use the timestamp taken from Kafka when we ingest the message. We started forwarding that value recently in https://github.com/getsentry/vroom/pull/190.

### DIFF
--- a/tests/datasets/test_functions_processor.py
+++ b/tests/datasets/test_functions_processor.py
@@ -49,6 +49,7 @@ class ProfileCallTreeEvent:
     os_name: str
     os_version: str
     retention_days: int
+    received: int
 
     def serialize(self) -> Mapping[str, Any]:
         return {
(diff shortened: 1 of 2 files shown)


tests/datasets/test_functions_processor.py
```diff
@@ -49,6 +49,7 @@ class ProfileCallTreeEvent:
     os_name: str
     os_version: str
     retention_days: int
+    received: int
```
Zylphrex, line 52: This should be an `Optional[int]` right?

Zylphrex (approved):
LGTM

evanh (approved):
Fine by me, but do you need to preserve the client timestamp as well?

phacops:
@evanh we're still preserving the client timestamp (and storing it), this is simply to use a different timestamp for origin (from when it enters our ingest pipeline to when it's done).

PR #3826, opened 2023-03-06
feat(replays): Add bloom filter index to replay_id column

cmanallen:
Intended to improve the performance of replay-details pages.

This addition was tested locally against a dataset of 10 million replays.  Multiple strategies were tested.  The bloom filter consistently returned the best results (on read).  I did not test write performance.  However, I did not notice significant regression when ingesting 10M records.

### DIFF
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -196,6 +196,7 @@ class ReplaysLoader(DirectoryLoader):
             "0006_add_is_archived_column",
             "0007_add_replay_type_column",
             "0008_add_sample_rate",
+            "0009_add_index_to_replay_id",
         ]
 
 
--- a/snuba/snuba_migrations/replays/0009_add_index_to_replay_id.py
+++ b/snuba/snuba_migrations/replays/0009_add_index_to_replay_id.py
@@ -1,0 +1,34 @@
+from typing import Iterator, Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return list(forward_columns_iter())
+
+    def backwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return list(backward_columns_iter())
+
+
+def forward_columns_iter() -> Iterator[operations.SqlOperation]:
+    yield operations.AddIndex(
+        storage_set=StorageSetKey.REPLAYS,
+        table_name="replays_local",
+        index_name="bf_replay_id",
+        index_expression="replay_id",
+        index_type="bloom_filter()",
+        granularity=1,
+    )
+
+
+def backward_columns_iter() -> Iterator[operations.SqlOperation]:
+    yield operations.DropIndex(
+        storage_set=StorageSetKey.REPLAYS,
+        table_name="replays_local",
+        index_name="bf_replay_id",
+        target=operations.OperationTarget.LOCAL,
+    )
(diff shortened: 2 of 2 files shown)


cmanallen:
PR fails because of:

```
snuba.clickhouse.errors.ClickhouseError: DB::Exception: Unexpected type UUID of bloom filter index.
```

I'm able to run the migration locally.  I'm wondering if this is a version conflict?  My local server is 21.6.1.  What is production?

evanh:
Replays is probably 21.8, but I believe CI runs against 20.3, since there are some clusters still on that version.

nikhars:
Snuba CI runs with version 20.3 because that is the minimal version which is supported across different environments like SaaS, Single Tenant and Self Hosted. The clickhouse version run in production (21.8) would support the migration you are writing. But if the replays product needs to run on self-hosted or single tenant, there would be a migration failure.

One thing you could try is to create a a new column of type string which is [materialized](https://clickhouse.com/docs/en/sql-reference/statements/alter/column/#materialize-column) from the [UUID](https://clickhouse.com/docs/en/sql-reference/functions/uuid-functions/#uuidnumtostring). When you want to run queries, you run them against this new column.

JoshFerge:
> Snuba CI runs with version 20.3 because that is the minimal version which is supported across different environments like SaaS, Single Tenant and Self Hosted. The clickhouse version run in production (21.8) would support the migration you are writing. But if the replays product needs to run on self-hosted or single tenant, there would be a migration failure.
> 
> One thing you could try is to create a a new column of type string which is [materialized](https://clickhouse.com/docs/en/sql-reference/statements/alter/column/#materialize-column) from the [UUID](https://clickhouse.com/docs/en/sql-reference/functions/uuid-functions/#uuidnumtostring). When you want to run queries, you run them against this new column.

since this is an optimization that only improves the performance of queries, could we instead specify some kind of version limitation for migrations? I feel like going forward as we want to use new clickhouse features / capabilities this seems like a continual issue we may run into and should probably start thinking about ways around this problem, e.g. "to use replays to its fullest potential you must be on clickhouse version XX or greater". additionally, no self hosted customers are using replays at the moment, so i do not think that is a concern.

evanh:
So this is a problem we will definitely have to solve, but we don't have an immediate solution for it at the moment.

@JoshFerge our understanding is that your team should have access to the replays cluster in prod. We recommend manually adding the index there and closing this PR. If you need help with that let us know.

nikhars:
> > Snuba CI runs with version 20.3 because that is the minimal version which is supported across different environments like SaaS, Single Tenant and Self Hosted. The clickhouse version run in production (21.8) would support the migration you are writing. But if the replays product needs to run on self-hosted or single tenant, there would be a migration failure.
> > One thing you could try is to create a a new column of type string which is [materialized](https://clickhouse.com/docs/en/sql-reference/statements/alter/column/#materialize-column) from the [UUID](https://clickhouse.com/docs/en/sql-reference/functions/uuid-functions/#uuidnumtostring). When you want to run queries, you run them against this new column.
> 
> since this is an optimization that only improves the performance of queries, could we instead specify some kind of version limitation for migrations? I feel like going forward as we want to use new clickhouse features / capabilities this seems like a continual issue we may run into and should probably start thinking about ways around this problem, e.g. "to use replays to its fullest potential you must be on clickhouse version XX or greater". additionally, no self hosted customers are using replays at the moment, so i do not think that is a concern.

Dealing with multiple versions of clickhouse within migrations isn't feasible. For example, till this migration replays would work fine with Clickhouse version 20.3. So lets assume self hosted customers are using replays with clickhouse version 20.3. Then this migration comes along and it puts a constraint that for replays product, Clickhouse needs to be version 21.8. On self-hosted and singlet tenant environments, customers are running 20.3. All the datasets run on a single Clickhouse instance. Without knowing whether something else breaks, we cannot recommend upgrading Clickhouse. The issue of new clickhouse features/capabilities goes beyond migrations. It involves Snuba as a whole. Here, you are only dealing with migration. But lets assume you want to use a new SQL statement introduced in the newer version. Then it becomes a Snuba concern. The right way to go about the issues you raise is being on a regular cadence of upgrade cycles.

PR #3833, opened 2023-03-08
feat(CapMan): Passthrough Allocation Policy

volokluev:
### Context

In order to start making policy decisions about capacity, we need a place to put them. This PR lays the groundwork for all future capacity management work. 

This PR:

* defines an `AllocationPolicy` and makes it the property of a `Storage`
* updates the `StorageQueryPlanBuilder` to inject the allocation policy into the `clickhouse_query`'s from clause
* updates `db_query.py` to check and update the allocation policy before and after the `db_query` 
* updates `snuba/web/views.py` to handle `AllocationPolicyViolation`s in the same way as rate limit violations

This PR *does not*:

* add allocation policy as a configurable item on the storage
* introduce any new limiting mechanisms

### Review Tips

The line count is high in this PR. A lot of that is tests and docstrings. If you find yourself getting lost, please ask (even if the feedback is vague)

Suggested file review order:

`snuba/query/allocation_policies/__init__.py` - see what an allocation policy is with usage examples
`tests/query/allocation_policies/test_allocation_policy_base.py` - tests for above

`snuba/datasets/storage.py` - where the allocation policy plugs into the configuration
`snuba/query/data_source/simple.py` - changes to the table data source to allow for allocation policy
`snuba/datasets/plans/storage_plan_builder.py` - where allocation policy gets injected in the query pipeline
`tests/datasets/plans/test_storage_plan_builder.py` - tests for above

`snuba/web/db_query.py` - how the allocation policy is actually used with queries
`tests/web/test__get_allocation_policy.py` - tests for above
`tests/web/test_db_query.py` - tests for above

`snuba/web/views.py` - handling of allocation policy violations at the top level
`tests/test_snql_api.py` - test for above

`snuba/web/query.py` - simple rename
`tests/query/processors/query_builders.py` - simple test update
`tests/pipeline/test_composite_planner.py` - simple test update

### Blast Radius

This is on the query critical path, all of snuba is affected by this change


snuba/datasets/storage.py
```diff
@@ -85,6 +86,9 @@ def get_mandatory_condition_checkers(self) -> Sequence[ConditionChecker]:
         """
         return []
 
+    def get_allocation_policy(self) -> AllocationPolicy:
```
volokluev, line 89: having this be configurable will come in a future PR

tests/datasets/plans/test_storage_plan_builder.py
```diff
@@ -116,18 +113,23 @@ def test_storage_query_plan_builder(
     )
     query, snql_anonymized = parse_snql_query(str(snql_query), dataset)
     assert isinstance(query, Query)
-    request = Request(
```
volokluev, line 119: I removed this because it was not necessary for the test. Can add it back in if people like

snuba/web/db_query.py
```diff
@@ -251,17 +264,18 @@ def _apply_thread_quota_to_clickhouse_query_settings(
...
         )
-        clickhouse_query_settings["max_threads"] = max(
-            1, maxt - project_rate_limit_stats.concurrent + 1
-        )
+        if project_rate_limit_stats:
+            clickhouse_query_settings["max_threads"] = max(
+                1, maxt - project_rate_limit_stats.concurrent + 1
+            )
+        else:
+            clickhouse_query_settings["max_threads"] = maxt
```
volokluev, line 278: this was a bug where if there is no project rate limiter in the query pipeline the max_threads does nothing to the query.

snuba/web/db_query.py
```diff
@@ -186,7 +195,11 @@ def execute_query(
 
     timer.mark("execute")
     stats.update(
-        {"result_rows": len(result["data"]), "result_cols": len(result["meta"])}
+        {
+            "result_rows": len(result["data"]),
+            "result_cols": len(result["meta"]),
+            "max_threads": clickhouse_query_settings.get("max_threads", None),
```
volokluev, line 201: keeping track of what the max threads we ran the query with, needed it for testing, could be useful for other things too

snuba/query/allocation_policies/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+            required_tenant_types, list
+        ), "required_tenant_types must be a list of strings"
+        assert isinstance(storage_set_key, str)
+        return cls(
+            required_tenant_types=required_tenant_types,
+            storage_set_key=StorageSetKey(storage_set_key),
+            **kwargs,
+        )
+
+    def get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowance:
```
dbanda, line 163: How often is this method called? Is the expectation that it is very fast? I'm wondering if there would be a huge perf impact if say someone had a complicated statistical algo here.

snuba/query/allocation_policies/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    @property
+    def explanation(self) -> dict[str, Any]:
+        return self.extra_data.get("quota_allowance", {}).get("explanation", {})  # type: ignore
+
+    def __str__(self) -> str:
+        return f"{self.message}, explanation: {self.explanation}"
+
+
+class AllocationPolicy(ABC, metaclass=RegisteredClass):
+    """This class should be the centralized place for policy decisions regarding
```
dbanda, line 60: nice docstrings btw 👍

snuba/web/db_query.py
```diff
@@ -607,3 +615,141 @@ def raw_query(
...
+    elif isinstance(from_clause, CompositeQuery):
+        return _get_allocation_policy(from_clause)
+    elif isinstance(from_clause, JoinClause):
+        # HACK (Volo): Joins are a weird case for allocation policies and we don't
+        # actually use them anywhere so I'm purposefully just kicking this can down the
+        # road
+        return DEFAULT_PASSTHROUGH_POLICY
+    else:
+        logger.exception(
+            f"Could not determine allocation policy for {clickhouse_query}"
```
rahul-kumar-saini, line 645: Could certain useful info from the Clickhouse query be passed as tags? I'm guessing this message would get truncated and we might miss the from clause.

snuba/web/db_query.py
```diff
@@ -607,3 +615,141 @@ def raw_query(
...
+    * applying rate limits which have been applied during the query pipeline
+    * collecting information about the query which will become part of the querylog entry for this request
+        * this is done with the stats, and query_metadata_list parameters
+
+
+    ** GOTCHAS **
+    --------------
+
+    * Whenever something goes wrong during the running of this function, it is wrapped in a QueryException,
+        that exception neeeds to have whatever stats were collected during this function's execution
```
rahul-kumar-saini, line 678:
```suggestion
        that exception needs to have whatever stats were collected during this function's execution
```

rahul-kumar-saini (commented):
Just a few nits but overall LGTM. I will hold off on approving to let others get in their reviews since this PR modifies the critical query path.

evanh (commented):
Architecture question:

Since the allocation policy is currently instantiated inside the storage, that implies the allocation policy is isolated to each storage. However that policy is also passed into a Table. In theory there can be multiple storages using the same table, and many tables shared across a storage set key.

My question is: are the allocation policies aware of each other? If not, how do we properly tune a policy that is sharing a resource with another policy?

snuba/query/allocation_policies/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+
+    def get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowance:
+        allowance = self._get_quota_allowance(tenant_ids)
+        if not allowance.can_run:
+            raise AllocationPolicyViolation.from_args(tenant_ids, allowance)
+        return allowance
+
+    @abstractmethod
+    def _get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowance:
```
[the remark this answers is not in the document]
volokluev: This structure separates two core functionalities:

`get_quota_allowance` wraps around `_get_quota_allowance` to allow changing the behavior of all allocation policies. This is in part an answer to your question: "how do we properly tune a policy that is sharing a resource with another policy?"


If/when we do need to make allocation policies aware of other policies or resources, `get_quota_allowance` is where that code will go. If we want to introduce more monitoring around all allocation policies, that will go in `get_quota_allowance` with the user logic not needing to change.

enochtangg (commented):
Looks good! I just have two clarifying questions. 

Firstly, you might have answered this already, but how would work from the user's perspective. With this new system, if I wanted to rate limit the threads of a particular tenant_id, and assuming I already have a custom policy in place, would I still input the rate limit values as part of a runtime config in redis?

Secondly, is the following statement correct: all tables that are part of the same storage_set have the same allocation policy? If so, my follow up question (which ties on to Evan's architecture question), how would this shape the management of storage_sets in the future? Would partitioning/creating more storage_sets mean more flexibility for this system? For example, all storages within the metrics dataset (sets, counters, dist, etc.) all share one storage_set (less flexible). On the other hand, generic metrics dataset contains 3 storage_sets (more flexible).

snuba/query/allocation_policies/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+    * An organization_id may only scan this many bytes in an hour
+    * A a referrer that has scanned this many bytes has this many max_threads to run their query
+
+    To make your own allocation policy:
+
+        >>> class MyAllocationPolicy(AllocationPolicy):
+
+        >>>     def _get_quota_allowance(
+        >>>         self, tenant_ids: dict[str, str | int]
```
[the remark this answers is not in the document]
volokluev: Not necessarily. For example, rate limiting is an allocation policy decision made without knowledge of how much capacity is available. the implemented allocation policy may use the load feedback loop to make its decisions but it does not have to

snuba/query/allocation_policies/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        self,
+        tenant_ids: dict[str, str | int],
+        result_or_error: QueryResultOrError,
+    ) -> None:
+        pass
+
+
+class PassthroughPolicy(AllocationPolicy):
+    def _get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowance:
+        from snuba.state import get_config
```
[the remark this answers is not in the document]
volokluev: good call, will do

snuba/query/allocation_policies/__init__.py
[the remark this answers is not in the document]
volokluev: > is each allocation policy responsible for figuring that all out?

I see what you're saying. Yes that is correct. Different policies may keep track of different things but that responsibility should be centralized there

snuba/query/allocation_policies/__init__.py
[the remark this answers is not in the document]
evanh: OK that makes sense.

I'm not sure it answers my question though. Each Storage is going to have its own instance of `AllocationPolicy`, so I don't understand how adding functionality to `get_quote_allowances` will let separate instances of `AllocationPolicy` work together?

volokluev:
@enochtangg @evanh 

> if I wanted to rate limit the threads of a particular tenant_id, and assuming I already have a custom policy in place, would I still input the rate limit values as part of a runtime config in redis?

Functionally, yes. Any overrides or configuration would be runtime configs in redis, however because these exist as part of one policy, we can make better tooling around the available overrides. That work isn't in this PR but it will come as well.


> are all tables that are part of the same storage_set have the same allocation policy?

not quite. I'm trying to thread the needle through a few (sometimes clashing) concepts so I understand that this is confusing. 


TL;DR: This PR will not solve multiple storages living on the same cluster, but hopefully makes enough room for us to adapt to new challenges that come up. 

**Here is the dilemma:**

1. Allocation policies should live on a configuration object, because how we allocate capacity is primarily a product concern. It's not hard for us to turn down the rate limits until the load is fine, but that breaks the product. The configuration object closest to the hardware that is running the query is a storage.
2. We are trying to manage the load on the clusters that run queries. The cluster that runs the query is only known at runtime (or in the ops repo) because different environments have different cluster configurations.

Enter the `storage_set_key`, which implies that [the storages must be located on the same cluster](https://github.com/getsentry/snuba/blob/master/snuba/clusters/storage_sets.py#L49-L50) if they have the same `storage_set_key`. In our settings we also specify [which storage sets live on which clusters](https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L79).

It is because of this structure of the code that I've decided to pass the `StorageSetKey` to the `AllocationPolicy`.  A `StorageSet` is effectively a tenant on the cluster but this information is outside of the product concern. 

With that in mind, here are some concrete answers to questions:

**Q: Do all tables that are part of the same storage_set have the same allocation policy?**

They can but do not have to. If storages are part of the same storage set, they can re-use the same allocation policy or create different ones, we don't really care. If different tables belong in the same storage set, it is known before runtime that their fates (and capacity) are tied together. How they manage access to themselves with relation to other tables in that set is not in scope because it's not really an issue right now. 

**Q: are the allocation policies aware of each other?**

Currently they are not but there is no reason they couldn't be in the future. 

**Q: how do we properly tune a policy that is sharing a resource with another policy?**

The answer to "how do we do it after this PR is merged" is we don't, in the beginning of the project we recorded an [anti-requirement](https://www.notion.so/sentry/Clickhouse-Capacity-Management-df585f5b7ae4453a9e635e1520b4a211?pvs=4#459d85d0c1e149a8b9133033ee523217) to not worry about this problem too much. 

The answer to "how will we do it in the future" is more dependent on what that future brings. Here is how the code as is leaves room for that future rather than ignoring it:

* The base class of the `AllocationPolicy` allows us to write whatever code we want before and after the product logic of the specific policy is applied
* We know the `StorageSet` that the allocation policy is being applied on behalf of. This gives us the ability to make more general decisions about storage sets sharing a cluster

For example:

* We may add a field to the `CLUSTERS` setting which says how much capacity of a cluster each storage set is supposed to get. The base class of the allocation policy could apply those limits before `_get_quota_allowance` is called.

volokluev:
> I ask because if that's the idea, then I think we will inevitably end up with each allocation policy being bespoke for each storage, and worse, requiring code changes in order to change those policies

I think it's actually fairly likely that allocation policies will be different based on the storage and the policy will require code changes when we change it because the code is the policy. I'd rather let the bespokeness happen for some amount of time as we see the patterns that are reusable emerge rather than trying to predict and avoid them

evanh:
> allocation policies will be different based on the storage

I agree this will be the case.

> rather than trying to predict and avoid them

I agree we shouldn't be trying to avoid special cases, but I think we understand enough about our system to have opinions from the start. For example we know we want policies around referrer and organization.

It also seems to me that if we move product configuration into code we are going backwards from all the work we did to take product configuration out of code. It seems reasonable to me to have some common building blocks that can be combined in configuration. Then we can always adapt to handle any new cases that come up.

volokluev:
> It seems reasonable to me to have some common building blocks that can be combined in configuration. Then we can always adapt to handle any new cases that come up.

I agree that we should have those. What I'm saying is that I don't know exactly how those blocks will look yet because we're introducing a new concept to the system. There's nothing that prevents us from having the allocation policy be made up of sub-components that are also referenced in configuration. At the moment though, we have two tables whose capacity we're trying to manage and any code we write around those will be limited to the policy, not sprinkled throughout the codebase. 

What I'm trying to do in this PR is give us that place to quickly iterate on policies for the pieces of infrastructure which are experiencing capacity management issues.

evanh (approved):

PR #3835, opened 2023-03-08
ref: Add typing to all messages in errors replacer

untitaker:
Requires https://github.com/getsentry/sentry-kafka-schemas/pull/32 and
upgrade to latest mypy, because only there can you have generic
namedtuples.

The replacement type dispatch is a bit wonky, but I don't want to do
refactoring in the same PR.


lynnagara (approved):
Seems fine for now, but let's follow up soon to either improve the replacer typing or file a task so we remember to address it later

PR #3837, opened 2023-03-08
fix: Try running dependabot daily for internal dependencies

untitaker:


untitaker:
apparently this wont work

PR #3841, opened 2023-03-08
fix(test): Add extra query editor test

john-z-yang:
### Overview

- Add another test to query editor to increase code coverage
- Fix a race condition in existing query editor tests

### DIFF
--- a/snuba/admin/static/tests/query_editor.spec.tsx
+++ b/snuba/admin/static/tests/query_editor.spec.tsx
@@ -62,8 +62,8 @@ describe("Query editor", () => {
           description: "descripton for query 2",
         },
       ];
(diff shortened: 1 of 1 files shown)


volokluev:
Where was the race condition? How does your test fix it?

john-z-yang:
> Where was the race condition? How does your test fix it?
```
predefinedQueries.forEach(async (predefinedQuery) => {
          await user.selectOptions(getByTestId("select"), predefinedQuery.name);
          expect(mockOnQueryUpdate).lastCalledWith(predefinedQuery.sql);
});
```
Actually fires off multiple asynchronous calls, it doesn't break right now but sets a bad example when others are referencing this test

evanh (approved):

PR #3842, opened 2023-03-09
test: Run all consumers' message processors against examples

untitaker:
sentry-kafka-schemas is supposed to contain more and more examples over
time for all kafka topics. Run those examples against the relevant snuba
consumers.

### DIFF
--- a/snuba/processor.py
+++ b/snuba/processor.py
@@ -88,23 +88,7 @@ class ReplacementType(str, Enum):
     EXCLUDE_GROUPS = "exclude_groups"
 
 
-REPLACEMENT_EVENT_TYPES = frozenset(
-    [
-        ReplacementType.START_DELETE_GROUPS,
-        ReplacementType.START_MERGE,
-        ReplacementType.START_UNMERGE,
-        ReplacementType.START_DELETE_TAG,
-        ReplacementType.END_DELETE_GROUPS,
-        ReplacementType.END_MERGE,
-        ReplacementType.END_UNMERGE,
-        ReplacementType.END_DELETE_TAG,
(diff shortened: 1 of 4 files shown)


snuba/processor.py
```diff
@@ -101,6 +101,8 @@ class ReplacementType(str, Enum):
         ReplacementType.TOMBSTONE_EVENTS,
         ReplacementType.EXCLUDE_GROUPS,
         ReplacementType.REPLACE_GROUP,
+        ReplacementType.START_UNMERGE_HIERARCHICAL,
```
untitaker, line 104: those message types definitely exist in Sentry, but the product logic is behind a featureflag nobody has. I suppose we would be crashing the consumer today if they aren't here.

tests/consumers/test_schemas.py
```diff
@@ -1,107 +1,84 @@
...
+    ReplacerProcessor,
+)
+
+
+# Since these tests are pure, we can override the run_migrations fixture to
+# skip all test isolation. This reduces the test runtime from 10 seconds to 0.6
+# seconds
+@pytest.fixture
+def run_migrations():
+    pass
```
lynnagara, line 27: Oh nice, we can probably speed up a whole lot of our test suites this way

snuba/processor.py
```diff
@@ -101,6 +101,8 @@ class ReplacementType(str, Enum):
...
+        ReplacementType.END_UNMERGE_HIERARCHICAL,
```
lynnagara, line 105: Good catch.. though this is literally just all the members of `ReplacementType` enum... Do we actually need to define this here at all?

lynnagara (approved):
Nice!

PR #3845, opened 2023-03-09
feat(replays): Add migration for new DOM indexing behavior

cmanallen:
The replays product wants to enable customers to search by user interactions.  To this end we're adding click tracking.  When a user clicks on a DOM node that node's metadata is indexed in Clickhouse where it can be retrieved.

To fetch this data our query will need to aggregate each queried field into an array of arrays.  Rows are related by their position within the sub-arrays.  Index 0 refers to a single "click" action on a "div" with id "my-id".

```
[
    ["click", "click"],
    ["div", "button"],
    ["my-id", ""],
]
```

From there we will use arrayFilter to filter elements which do not match a condition.  This operation is repeated until all conditions are applied.

```
// Filter by element = "div"
[
    ["click"],
    ["div"],
    ["my-id"],
]

// Filter by id = "something"
[
    [],
    [],
    [],
]
```

We chose to append these new columns to our existing table rather than create a new one.  This was done to offer more query functionality.  E.g. `I want clicks on "my-id" where the environment is "production" sorted by error count`.


cmanallen:
@evanh I think this should help https://www.notion.so/sentry/Replay-DOM-Search-ca1eea3e377845bfbec123cda0a068a9.  It wasn't written to be an explanation though.

Long story short, replay recordings track clicks.  We want to index those clicks so we can search against them.  We're trying to enable queries like: "Show me the replays where the "login" button was clicked".

evanh (commented):
Please separate the configuration changes into a separate PR from the migration. Otherwise this broadly looks good to me.

cmanallen:
@evanh Migrated three more columns.  Allowing nullable on user, sdk-name, and sdk-version.

cmanallen:
- Now using the Nested data type to contain click actions.
- Removed the action column.

snuba/snuba_migrations/replays/0009_add_dom_index_columns.py
```diff
@@ -0,0 +1,103 @@
...
+        Column("click_role", String(Modifiers(default="''", low_cardinality=True))),
+        "click_text",
+    ),
+    (Column("click_alt", String(Modifiers(default="''"))), "click_role"),
+    (Column("click_testid", String(Modifiers(default="''"))), "click_alt"),
+    (Column("click_aria_label", String(Modifiers(default="''"))), "click_testid"),
+    (Column("click_title", String(Modifiers(default="''"))), "click_aria_label"),
+]
+
+alters: Sequence[str] = ["title", "user", "sdk_name", "sdk_version"]
```
evanh, line 27: Can you split these changes out into a separate PR? They are causing the test failures.

The test failures are because the existing columns in the storage configuration don't match what is in CH. You'll have to change the storage configuration to have the correct type, but that will muddy this PR up even more.

evanh (commented):
The CI failures here are real, this is my fault.

evanh (approved):

PR #3848, opened 2023-03-10
Revert "fix(querylog): enable readonly 2 for querylog"

MeredithAnya:
Reverts getsentry/snuba#3847

This didn't help at all, but also cabin doesn't even have the `trace_readonly` user, so this change probably wouldn't have worked anyway


MeredithAnya:
not needed, we fixed it

PR #3854, opened 2023-03-13
feat(tx_processor): write replay_id as top level column

JoshFerge:
We're deprecating sending replay_id as a tag (https://github.com/getsentry/rfcs/blob/rfc/escalating-issues/text/0048-move-replayid-out-of-tags.md)

This PR modifies the transactions processor to:

- [x] if the replay_id is sent as a tag (replayId), then set the top level value on the processed dict so the replay_id column is written.
    - [x] ensure the replay_id is a valid UUID before setting the top level value, as current replayId tags are not validated
- [x] if the replay_id is sent as a context, then set the top level value, and write it as a tag as well

we can remove the dual writing once we conditionally query on the new column based on date and only write the column

Task: https://github.com/getsentry/team-replay/issues/32
Epic: https://github.com/getsentry/replay-backend/issues/248

Relies on a change to relay to validate the replay context, which i'll be putting up shortly.

### DIFF
--- a/snuba/datasets/processors/transactions_processor.py
+++ b/snuba/datasets/processors/transactions_processor.py
@@ -50,6 +50,7 @@ class TransactionsMessageProcessor(DatasetMessageProcessor):
         "sentry:release",
         "sentry:user",
         "sentry:dist",
+        "replay_id",
     }
 
     def __extract_timestamp(self, field: int) -> Tuple[datetime, int]:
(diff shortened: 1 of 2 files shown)


evanh (commented):
Can you please also add the replay_id column to the transactions storage/entity configurations?

JoshFerge:
> Can you please also add the replay_id column to the transactions storage/entity configurations?

We want to wait to do that, as we want to populate the column first / change the queries before making the new field queryable. primarily, we need to change the current queries to use tags[replayId] as to not be confusing.

JoshFerge:
@evanh see https://github.com/getsentry/replay-backend/issues/248 for more information

evanh:
> We want to wait to do that, as we want to populate the column first / change the queries before making the new field queryable.

I don't see the harm in adding the column to the configuration, as long you aren't querying it in Sentry. But if you're worried about it I guess you can wait.

evanh (approved):

PR #3855, opened 2023-03-13
feat(replays): Add DOM indexing processor and configuration

cmanallen:
Depends on: https://github.com/getsentry/snuba/pull/3845

Updates the entity and storage configurations to include the new DOM indexing columns.  Also updates the processor to handle the new DOM indexing messages.

- For nullable columns do I need to update the processor to manually specify `None` for each of those columns?


snuba/datasets/processors/replays_processor.py
```diff
@@ -185,25 +185,76 @@ def process_message(
...
+            "sdk_version": None,
+            # Kafka columns.
+            "retention_days": processed["retention_days"],
+            "partition": metadata.partition,
+            "offset": metadata.offset,
+            # DOM Index fields.
+            "click_node_id": _collapse_or_err(_collapse_uint32, int(click["node_id"])),
+            "click_tag": to_string(click["tag"])[:32],
+            "click_id": to_string(click["id"])[:64],
+            "click_class": to_typed_list(to_string, click["class"][:10]),
```
evanh, line 246: This is limiting just the number of elements right? Since this is arbitrary user input I would cap each element as well.

snuba/datasets/processors/replays_processor.py
```diff
@@ -185,25 +185,76 @@ def process_message(
...
-            self._process_kafka_metadata(metadata, processed)
-
-            # # the following operation modifies the event_dict and is therefore *not* order-independent
-            self._process_user(processed, replay_event)
-            self._process_event_hash(processed, replay_event)
-            self._process_contexts(processed, replay_event)
-            return InsertBatch([processed], None)
+            if replay_event["type"] == "replay_actions":
+                actions = process_replay_actions(replay_event, processed, metadata)
+                return InsertBatch(actions, None)
```
evanh, line 190: @lynnagara Is there some internal limit to how many messages can be in an InsertBatch?

cmanallen:
Do I need to set a default value on the storages yaml?

evanh:
> For columns with defaults do I need to set a default value on the storages yaml?

No you don't.

evanh (commented):
If you're exposing the entity, you should also add an E2E test in test_replays_api.py as well.

snuba/datasets/processors/replays_processor.py
```diff
@@ -188,25 +189,78 @@ def process_message(
...
+    payload: Mapping[Any, Any],
+    processed: Mapping[Any, Any],
+    metadata: KafkaMessageMetadata,
+) -> list[dict[str, Any]]:
+    """Process replay_actions message type."""
+    return [
+        {
+            # Primary-key.
+            "project_id": processed["project_id"],
+            "timestamp": default(
```
JoshFerge, line 226: do we need to have a default for timestamp? when would it ever not be defined?

snuba/datasets/processors/replays_processor.py
```diff
@@ -188,25 +189,78 @@ def process_message(
...
+            ),
+            "replay_id": to_uuid(payload["replay_id"]),
+            "segment_id": None,
+            "event_hash": click["event_hash"],
+            # Default values for non-nullable columns.
+            "trace_ids": [],
+            "error_ids": [],
+            "urls": [],
+            "title": None,
+            "platform": "javascript",
```
JoshFerge, line 238: i'm wondering if hardcoding platform here to `javascript` would ever be an issue e.g. if the rest of the events are `react`. i don't think we use platform for anything at the moment so probably is fine. could also set to empty string?

snuba/datasets/processors/replays_processor.py
```diff
@@ -188,25 +189,78 @@ def process_message(
...
+            "user": None,
+            "sdk_name": None,
+            "sdk_version": None,
+            # Kafka columns.
+            "retention_days": processed["retention_days"],
+            "partition": metadata.partition,
+            "offset": metadata.offset,
+            # DOM Index fields.
+            "click_node_id": _collapse_or_err(_collapse_uint32, int(click["node_id"])),
+            "click_tag": to_string(click["tag"])[:32],
```
JoshFerge, line 248: ah i see we set string limits here. wonder if its worth doing upstream as well just to transmit less bytes in these cases? but not a huge deal to start unless we end up with problematic payloads

JoshFerge: also would `to_capped_string` work for all of these?

snuba/datasets/processors/replays_processor.py
```diff
@@ -188,25 +189,78 @@ def process_message(
...
+            "replay_id": to_uuid(payload["replay_id"]),
+            "segment_id": None,
+            "event_hash": click["event_hash"],
+            # Default values for non-nullable columns.
+            "trace_ids": [],
+            "error_ids": [],
+            "urls": [],
+            "title": None,
+            "platform": "javascript",
+            "user": None,
```
JoshFerge, line 239: just thinking ahead if we end up trying to go the subquery route it would be less headache if we stored all of the metadata on these rows as well, but can also be solved with a second query and joining so i think its fine as is, e.g. for the case of

"show me replays where it was user X and they clicked Y". i'm guessing we did this to save data and it makes sense.

snuba/datasets/processors/replays_processor.py
JoshFerge: do we limit the number of replay actions upstream at all in recordings consumer or sdk? cc @billyvg @cmanallen

billyvg: There are no limits on the SDK side

JoshFerge: i feel like we should put a bounds for bot cases, etc. something like 100 or 1000 per segment?

snuba/datasets/processors/replays_processor.py
cmanallen: We're doing it on both sides now.

snuba/datasets/processors/replays_processor.py
cmanallen: Unfortunately we don't have that information.  We would have to read on ingest which has two problems:

1. It doesn't scale.
2. The row is not guaranteed to exist in Clickhouse.

Merging the two events into one in Relay and/or the SDK will solve this problem as we would have both the event and the segment-data.

snuba/datasets/processors/replays_processor.py
cmanallen: I'm using `javascript` as a sentinel value because I was under the impression it was the only valid platform for the product.  To prevent weird aggregation errors we would need to store this as null

cmanallen:
https://github.com/getsentry/sentry/issues/45524

PR #3856, opened 2023-03-13
feat: Share Python settings with Rust and add Rust consumer entrypoint

lynnagara:
Alt to https://github.com/getsentry/snuba/pull/3802

I prefer this approach at least in the short term as it keeps the settings in Rust aligned with the Python version. It requires more minimal changes to the Python settings than the other option: only that the timedelta types need to change as these are not easily json serializable. Whether Python settings should ultimately move to yaml is a larger, more consequential topic - let's not try to solve it within the Rust consumer prototype.

The Rust consumer entrypoint loads the Python settings, writes the json file and replaces the current process with the Rust program passing the path to the JSON file through to the Rust consumer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ snuba.egg-info/
 node_modules
 .vscode/*.log
 snuba/admin/dist/bundle.js*
+tmp/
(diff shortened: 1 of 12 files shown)


untitaker (approved):
lgtm, can we merge this safely, or do we need to update our own prod settings? since you changed a few types from timedelta to something else

otherwise I guess you could also handle timedeltas in your custom json encoder

PR #3858, opened 2023-03-13
feat(schemas): Partially type the data in querylog producer

lynnagara:
It cannot be fully typed yet as schema is not correct. The SLO fields are marked as part of the `clickhouse_query` in the schema but are at the top level in the producer code.


evanh:
Schema change: https://github.com/getsentry/sentry-kafka-schemas/pull/42

untitaker (approved):
@lynnagara feel free to make further changes, but the current state looks like an improvement

PR #3861, opened 2023-03-14
feat(migrations): detect inactive replicas SNS-2131

dbanda:
## context
We implement Snuba migrations as loop that visits all the nodes and applies a sequence of DDL changes to the tables. In order for this to work properly, we need the DDL changes (ALTER statements) to execute synchronously where possible to reduce the chance of race conditions. The way this is done is by setting `replication_alter_partitions_sync=2` . This will make Clickhouse wait for all replicas to complete before moving on to the next statement.

 If there’s an inactive replica Clickhouse 20.x will throw the exception  `DB::Exception: Some replicas doesn’t finish metadata alter`. Clickhouse 21.xx has an option to pass a timeout and wait.

## changes
This PR adds a check that will run before run/reverse to query each node for inactive replicas. If an inactive replica is found, it will throw `InactiveClickhouseReplica` exception.


snuba/migrations/connect.py
```diff
@@ -55,3 +58,45 @@ def check_clickhouse(clickhouse: ClickhousePool) -> None:
...
+            checked_nodes.add((node.host_name, node.port))
+
+            conn = cluster.get_node_connection(ClickhouseClientSettings.MIGRATE, node)
+            tables_with_inactive = conn.execute(
+                f"SELECT table, total_replicas, active_replicas FROM system.replicas "
+                f"WHERE active_replicas < total_replicas AND database ='{conn.database}'",
+            ).results
+
+            for table, total_replicas, active_replicas in tables_with_inactive:
+                raise InactiveClickhouseReplica(
```
nikhars, line 99: One thing to consider is how a user might be able to help themselves when they see this problem. Should we point them to the clickhouse documentation on how to remove it?

Also, since you are checking for all tables, this will only tell them the first cluster to fix. What if there are 3 other clusters which need to be fixed? They would be in a loop where they fix one, try again, fix second, try again, fix third, try again and succeed. Can we raise an exception which lists all clusters which need to be fixed?

dbanda: I fixed it to list all inactive replicas. 

But I think it may be tricky to point to the Clickhouse documentation because there are many reasons why a replica may be inactive and there is not exactly one solution to fix them. It's better to take it case by case.

nikhars (approved):

PR #3862, opened 2023-03-14
Allow to specify redis username

HydrofinLoewenherz:
Takes the redis username like the password from the env and provides it to the redis client constructor. 
If no username is set in the env, the env getter (`os.environ.get("REDIS_USERNAME")`) gives `None`which is also the default value for the constructor. So it shouldn't break anything.

### Legal Boilerplate


evanh:
This caused errors in our production environment, I'll have to debug that.

PR #3810, opened 2023-03-02
Add exception_main_thread to Snuba

marandaneto:
### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #3863, opened 2023-03-14
Add exception_main_thread to storage and entity schemas

marandaneto:
Follow up of https://github.com/getsentry/snuba/pull/3810 and https://github.com/getsentry/snuba/pull/3840


snuba/datasets/configuration/discover/entities/discover.yaml
```diff
@@ -313,6 +318,7 @@ storages:
               - exception_frames.colno
               - exception_frames.lineno
               - exception_frames.stack_level
+              - exception_main_thread
```
marandaneto, line 321: I don't know if `DefaultNoneColumnMapper` needs this or not, just adding because `exception_frames` also has it.

snuba/datasets/configuration/discover/entities/discover_transactions.yaml
```diff
@@ -244,6 +244,7 @@ storages:
               - exception_frames.colno
               - exception_frames.lineno
               - exception_frames.stack_level
+              - exception_main_thread
```
marandaneto, line 247: Same as https://github.com/getsentry/snuba/pull/3863/files#r1135875859

marandaneto:
I added to all configuration files that `exception_frames` is also there, I hope this is right.

marandaneto:
@evanh should the processor be in the same PR or a different PR?
Can you do a quick review for the next steps?

snuba/datasets/configuration/discover/entities/discover.yaml
evanh: Yes, this is a list of columns that are in errors but not transactions. We need them listed in Discover.

evanh (approved):

marandaneto:
> @evanh should the processor be in the same PR or a different PR?
> 
> Can you do a quick review for the next steps?

Thanks for the approval, besides fixing the CI (failing test), should I just merge this and do the next PR or all in this one is fine?

evanh:
The only important part to be separate is the migration. You need to make sure any migrations have been run in production before you merge any of this configuration code.

snuba/datasets/processors/errors_processor.py
```diff
@@ -133,7 +134,16 @@ def process_insert(
...
+
+        thread: SentryThreadChain = data.get(
+            "threads",
+            cast(
+                SentryThreadChain, data.get("sentry.interfaces.threads.Threads", None)
+            ),
+        ) or {"values": []}
+        threadsStack = thread.get("values", None) or []
+
+        self.extract_stacktraces(processed, stacks, threadsStack)
```
marandaneto, line 146: @evanh I have no idea if this is right, I just kinda copied based on the `exception` example.
Does `sentry.interfaces.threads.Threads` even exist? I have not found the usage anywhere else.

snuba/datasets/processors/errors_processor.py
```diff
@@ -327,6 +343,23 @@ def extract_stacktraces(
...
+                        ## if it's not the main thread, continue
+                        main = thread.get("main", None)
+                        if main is None or main is False:
+                            continue
+
+                        id = thread.get("id", None)
+                        if id is not None and id == thread_id:
+                            ## if it's the main thread, mark it
+                            exception_main_thread = True
+                            break
```
marandaneto, line 361: @evanh how do I know if this is ok and not heavy? usually, there are not a lot of exceptions or threads, but I don't know what's ok and what's not here.

snuba/datasets/processors/errors_processor.py
marandaneto: Where is the definition of `sentry_kafka_schemas.schema_types.events_v1`?

untitaker: That module is code-generated from https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/events.v1.schema.json -- we can pair tomorrow on fixing it up to contain threads. You should be able to add a minimal type that covers the fields you need.

untitaker: > Does sentry.interfaces.threads.Threads even exist? I have not found the usage anywhere else.

no, this is dead code. you can remove it. everything with `sentry.interfaces` is legacy

marandaneto: Alright, let's pair up when you can.
Started something here https://github.com/getsentry/sentry-kafka-schemas/pull/45/files

snuba/datasets/processors/errors_processor.py
evanh: We monitor that performance pretty well, so I think mostly what you should do is let the on-call person know when this change is going out so they know to keep an extra eye out.

If we need too we can also add a metric to track the numbers later.

untitaker (approved):
if the migration is applied feel free to merge

marandaneto:
> if the migration is applied feel free to merge

thanks, changed a little bit to set the `flag` to false as well and added one more test case, if CI is happy, I will merge it.

snuba/datasets/processors/errors_processor.py
```diff
@@ -327,6 +341,27 @@ def extract_stacktraces(
                     frame_linenos.append(_collapse_uint32(frame.get("lineno", None)))
                     frame_stack_levels.append(stack_level)
 
+                ## mark if at least one of the exceptions happened in the main thread
+                if thread_id is not None and exception_main_thread is not True:
```
MeredithAnya, line 345: @marandaneto can we put this logic behind a snuba config option? 

if `get_config` is not already in this file you can import:  `from snuba.state import get_config` and then just set some flag with a default value so that we can turn this off without needing code changes if this ends up causing issues

```python
if get_config("check_exception_threads", 0):
   # your logic
```

marandaneto: [`3186e1a` (#3863)](https://github.com/getsentry/snuba/pull/3863/commits/3186e1a3527c9de7b6a4f62f662b0f69654dc9f1)

snuba/datasets/processors/errors_processor.py
```diff
@@ -311,6 +312,8 @@ def extract_stacktraces(
         frame_stack_levels = []
         exception_main_thread = None
 
+        check_exception_main_thread = get_config("check_exception_main_thread", True)
```
evanh, line 315: Configs should either be numbers or strings. So use 0/1 instead for this.

marandaneto: Done but Booleans are used multiple types in the code base, for example
`if get_config("skip_seen_offsets", False):`
`state.get_config("enable_events_readonly_table", False)`
Etc...

evanh (approved):

MeredithAnya:
I'm defaulting the config option to `0` for safety incase something were to happen with redis

PR #3864, opened 2023-03-14
fix(issue-platform): map timestamp to receive_timestamp

barkbarkimashark:
Primary key on this table is actually `receive_timestamp` so we should be querying on `receive_timestamp` instead of `client_timestamp`.

### DIFF
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -72,7 +72,7 @@ storages:
             from_table_name:
             from_col_name: "timestamp"
             to_table_name:
-            to_col_name: "client_timestamp"
+            to_col_name: "receive_timestamp"
         - mapper: ColumnToColumn
           args:
             from_table_name:


wedamija (commented):
Timestamp should map to `client_timestamp` though, since that's when the event actually occurred.

I forget if there was a reason that we made `received_timestamp` part of the pk. We should probably drop/recreate the table now before we fully release and fix the key, unelss there's some reason not to

barkbarkimashark:
> Timestamp should map to `client_timestamp` though, since that's when the event actually occurred.
> 
> I forget if there was a reason that we made `received_timestamp` part of the pk. We should probably drop/recreate the table now before we fully release and fix the key, unelss there's some reason not to

Bleh, yeah was hoping we didn't have to rebuild but I think it can't be avoided. I think we'll have to find a way to move over the data in the current table to the recreated one.

wedamija:
> > Timestamp should map to `client_timestamp` though, since that's when the event actually occurred.
> > I forget if there was a reason that we made `received_timestamp` part of the pk. We should probably drop/recreate the table now before we fully release and fix the key, unelss there's some reason not to
> 
> Bleh, yeah was hoping we didn't have to rebuild but I think it can't be avoided. I think we'll have to find a way to move over the data in the current table to the recreated one.

Are any profile issues actually released to customers? We might be able to get away with not retaining data (although worth checking with SNS if there's an easy way to just dump the data in)

barkbarkimashark:
> > > Timestamp should map to `client_timestamp` though, since that's when the event actually occurred.
> > > I forget if there was a reason that we made `received_timestamp` part of the pk. We should probably drop/recreate the table now before we fully release and fix the key, unelss there's some reason not to
> > 
> > 
> > Bleh, yeah was hoping we didn't have to rebuild but I think it can't be avoided. I think we'll have to find a way to move over the data in the current table to the recreated one.
> 
> Are any profile issues actually released to customers? We might be able to get away with not retaining data (although worth checking with SNS if there's an easy way to just dump the data in)

We have like two orgs that we've turned on profile occurrence ingestion for:
https://github.com/getsentry/getsentry/pull/9697/files

they're not seeing the issues in the FE, but we are ingesting the occurrences. The idea was to ingest it and hide it for customers but still let sentry employees 'preview' the issues in the FE.

barkbarkimashark:
> 
Oh yeah, I think you were on PTO when we made this change. 

We are processing the issue occurrences for profiling issues for those two orgs. 
And we made a change here to selectively enable showing Profiling issues to only sentry employees. 
https://github.com/getsentry/sentry/blob/225918c670ad717f290ab11f7d21020ec1243fa8/src/sentry/issues/search.py#L210-L213

As far as I can tell, once we EA/GA this feature https://flagr.getsentry.net/#/flags/314
The two orgs who are on that list should 'all of a sudden' be able to see profiling issues that were created when we 'secretly' started ingesting their profile occurrences.

wedamija:
> > 
> 
> Oh yeah, I think you were on PTO when we made this change.
> 
> We are processing the issue occurrences for profiling issues for those two orgs. And we made a change here to selectively enable showing Profiling issues to only sentry employees. https://github.com/getsentry/sentry/blob/225918c670ad717f290ab11f7d21020ec1243fa8/src/sentry/issues/search.py#L210-L213
> 
> As far as I can tell, once we EA/GA this feature https://flagr.getsentry.net/#/flags/314 The two orgs who are on that list should 'all of a sudden' be able to see profiling issues that were created when we 'secretly' started ingesting their profile occurrences.

Right, I guess what I mean is that maybe it'd be easier to just drop the events for these customers. If there are no events, then they likely won't show up in issue search. The other option could be to just drop all the events, and then delete all of the profile issues from these customers via a backfill in getsentry.

barkbarkimashark:
> > > 
> > 
> > 
> > Oh yeah, I think you were on PTO when we made this change.
> > We are processing the issue occurrences for profiling issues for those two orgs. And we made a change here to selectively enable showing Profiling issues to only sentry employees. https://github.com/getsentry/sentry/blob/225918c670ad717f290ab11f7d21020ec1243fa8/src/sentry/issues/search.py#L210-L213
> > As far as I can tell, once we EA/GA this feature https://flagr.getsentry.net/#/flags/314 The two orgs who are on that list should 'all of a sudden' be able to see profiling issues that were created when we 'secretly' started ingesting their profile occurrences.
> 
> Right, I guess what I mean is that maybe it'd be easier to just drop the events for these customers. If there are no events, then they likely won't show up in issue search. The other option could be to just drop all the events, and then delete all of the profile issues from these customers via a backfill in getsentry.

Ahh, I see what you mean. The Group row would exist but the events that mapped back to those groups would just be wiped out. Yeah, that's a good suggestion about just dropping the Group rows in a backfill. Way easier than backfilling the actual events ..

PR #3869, opened 2023-03-16
feat: Start building the config for Rust consumer

lynnagara:
We don't want to reimplement the slice/cluster resolution in Rust, call it in Python when the Rust consumer starts and write it in a file for Rust.

This is currently incomplete, there are some additional fields we need to add here such as the reference to the message processor.

### DIFF
--- a/snuba/cli/rust_consumer.py
+++ b/snuba/cli/rust_consumer.py
@@ -122,8 +122,6 @@ def rust_consumer(
         consumer_config_path,
     ]
 
-    rust_consumer_args = ["--", "--settings-path", settings_path]
-
     for storage_name in storage_names:
         rust_consumer_args.extend(["--storage", storage_name])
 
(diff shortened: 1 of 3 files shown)


snuba/cli/rust_consumer.py
```diff
@@ -69,3 +130,175 @@ def rust_consumer(*, storage_names: Sequence[str], log_level: str) -> None:
...
+
+    # TODO: Temporarily hardcoded. To be properly resolved based on the storage set.
+    hardcoded_clickhouse_cluster = ClickhouseClusterConfig(
+        host="127.0.0.1", port=9000, user="default", password="", database="default"
+    )
+
+    return RustConsumerConfig(
+        storages=[
+            StorageConfig(
+                name=storage_name, clickhouse_cluster=hardcoded_clickhouse_cluster
```
untitaker, line 233: can you resolve table here as well? or should I continue parsing yaml

dbanda (approved):

snuba/cli/rust_consumer.py
lynnagara: Done!

PR #3871, opened 2023-03-16
feat: The Rust consumer calls Python message processors

untitaker:
* Invoke Python message processor, get back the encoded rows and print them.
* Move away from os.exec as per armin's recommendation. It causes the Rust part to call into the right Python interpreter in the right virtualenv.
* TBD: parallelism and clickhouse inserts. What should we focus on?

In order to build and install the Rust consumer, one now can no longer run `cargo build`, it has to be `maturin develop`.


rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
```diff
@@ -67,7 +67,7 @@ fn create_kafka_message(msg: BorrowedMessage) -> ArroyoMessage<KafkaPayload> {
             headers: msg.headers().map(BorrowedHeaders::detach),
             payload: msg.payload().map(|p| p.to_vec()),
         },
-        DateTime::from_utc(NaiveDateTime::from_timestamp(time_millis, 0), Utc),
```
untitaker, line 70: bugfix in arroyo: `NaiveDateTime::from_timestamp`'s first argument is seconds, not milliseconds. This caused the conversion to Python datetime to fail because the year was too large.

rust_snuba/Cargo.toml
```diff
@@ -23,4 +24,5 @@ env_logger = "0.10.0"
 serde = { version = "1.0", features = ["derive"]}
 serde_json = { version = "1.0" }
 glob = "0.3.1"
-pyo3 = { version = "0.18.1", features = ["auto-initialize"] }
+pyo3 = { version = "0.18.1", features = ["chrono", "extension-module"] }
+clickhouse = "0.11.3"
```
lynnagara, line 28: Is this needed? Afaict it's not used. There's a clickhouse client example in rust_arroyo already that doesn't use a library.

lynnagara (commented):
Overall approach LGTM. Mind rebasing though? A few things such as the broker config and the arguments passed from Python changed quite a bit.

rust_snuba/Cargo.toml
untitaker: removed

lynnagara (approved):

PR #3872, opened 2023-03-17
feat(replays): store replay_id as top level column value in clickhous…

JoshFerge:
following up on https://github.com/getsentry/snuba/pull/3854, this PR does the same thing for the errors processor, so replay_id is written as a top level column in addition to the tag. RFC here and top level epic here 

Task: https://github.com/getsentry/team-replay/issues/32
Epic: https://github.com/getsentry/replay-backend/issues/248

### DIFF
--- a/snuba/datasets/processors/errors_processor.py
+++ b/snuba/datasets/processors/errors_processor.py
@@ -230,6 +230,16 @@ class ErrorsProcessor(DatasetMessageProcessor):
         output["release"] = tags.get("sentry:release")
         output["dist"] = tags.get("sentry:dist")
         output["user"] = tags.get("sentry:user", "") or ""
+
+        replay_id = tags.get("replayId")
+        if replay_id:
+            try:
+                # replay_id as a tag is not guarenteed to be UUID (user could set value in theory)
+                # so simply continue if not UUID.
+                output["replay_id"] = str(uuid.UUID(replay_id))
+            except ValueError:
+                pass
+
         # The table has an empty string default, but the events coming from eventstream
         # often have transaction_name set to NULL, so we need to replace that with
         # an empty string.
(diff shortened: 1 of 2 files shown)


tests/datasets/test_errors_processor.py
```diff
@@ -121,6 +122,7 @@ def test_error_processor() -> None:
                             "build": "3.7.6",
                         },
                         "trace": {"trace_id": trace_id, "span_id": span_id},
+                        "replay": {"replay_id": str(replay_id)},
```
JoshFerge, line 125: feel like i need to add a few more tests cases to this file for my code -- also interesting we only seem to have one? is there some place else my tests should go?

evanh: Another place to put the test would be `test_replays_api.py`. That writes replays and then reads them. You could also test the case of passing a non-UUID replay in the `replayId` tag.

snuba/datasets/processors/errors_processor.py
```diff
@@ -238,13 +248,22 @@ def extract_tags_custom(
...
         trace_id = transaction_ctx.get("trace_id", None)
         span_id = transaction_ctx.get("span_id", None)
 
+        replay_ctx = contexts.get("replay") or {}
+        replay_id = replay_ctx.get("replay_id", None)
+
+        if replay_id:
+            replay_id_uuid = uuid.UUID(replay_id)
+            output["replay_id"] = str(replay_id_uuid)
+            tags["replayId"] = replay_id_uuid.hex
```
evanh, line 264: Why do you overwrite this? If the customer is passing it in, shouldn't we keep it the same?

tests/datasets/test_errors_processor.py
JoshFerge: @evanh this is in the errors processor -- would i do `test_errors_api.py`?

snuba/datasets/processors/errors_processor.py
JoshFerge: i'll add a check to make sure we're not overwriting as to not interfere. i did a similar thing on the transactions processor PR.

tests/datasets/test_errors_processor.py
JoshFerge: it seems like that is mainly testing querying, not insertion?

evanh: You're kind of bumping into some old code that hasn't really changed much, and so has older testing.

Options:
1) Make `test_errors_processing` more modern. Move the test event out of the test and allow it to be customized in different tests (see `test_transactions_processor`). Then you can add tests for your specific case.
2) Leverage some E2E tests as in `test_snql_api` to ensure your data is written able to be read.

Option 1 is the more correct option and if you were on the SnS team I would push you to that.
Option 2 is faster, and I'm OK with it because I don't think it's fair to make product team members upgrade our tests to ship their products (as much as I would like that 😀)

JoshFerge:
@evanh i've went ahead and spent the day refactoring the tests to be more like `test_transactions_processor` and added new tests. In the time since you last commented, it seems someone else has just gone ahead and added more tests like the original error test https://github.com/getsentry/snuba/pull/3863 (which is what my refactoring was assumed to be avoiding and what i could have done in a couple of minutes). I would really like to be done with this and i'd rather not work extra if i don't need to, so please advise on next steps.

I've simply copy/pasted the tests that were written into the branch since my refactoring.

JoshFerge:
I've just gone ahead and just refactored those new tests to use the refactored version. let me know if that works!

evanh (approved):

evanh (commented):
Thanks for doing all that!

PR #3873, opened 2023-03-17
feat(rust-consumer): Avoid config files overwriting each other

lynnagara:
Better to give each config a unique path


untitaker (approved):
We will soon no longer have to do IPC and can just pass the JSON blob as a gigantic string to Rust.

PR #3874, opened 2023-03-17
feat(rust-consumer): Add python processor info to consumer config

lynnagara:
So Rust consumer can import the Python module

### DIFF
--- a/snuba/cli/rust_consumer.py
+++ b/snuba/cli/rust_consumer.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 import os
 from dataclasses import asdict, dataclass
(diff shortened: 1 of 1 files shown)


snuba/cli/rust_consumer.py
```diff
@@ -153,11 +148,20 @@ class ClickhouseClusterConfig:
     database: str
 
 
+@dataclass(frozen=True)
+class MessageProcessorConfig:
+    # TODO: Probably won't end up needing all this
+    python_class_name: str
+    python_module: str
+    python_file_path: str
```
fpacifici, line 156: Does this mean the whole processor needs to be in one file ?

untitaker: we only need `name` and args from the storage yaml, the `DatasetMessageProcessor` from the baseclass can load by classname: https://github.com/getsentry/snuba/blob/7e8df53a16171dcd3ff06f610ef1d0aae8b84006/rust_snuba/src/strategies/python.rs#L33-L35

alternatively we can also pass the full classpath `snuba.processors....Processor` so we don't rely on this class registry in Python.

`file_path` I can't use for anything.

We are not constrained to a single file... we launch a real Python interpreter and can import the entirety of Snuba

untitaker (approved):
lgtm but i would remove `file_name`. I don't have strong opinion on whether we should provide the full classpath, or just the classname and rely on `get_from_name` from within Rust

snuba/cli/rust_consumer.py
fpacifici: I see. Cool

lynnagara: Removed the file path

PR #3876, opened 2023-03-17
Remove filtering only local nodes for tracing tool

enochtangg:
Currently, snuba admin tracing tool does not have access to discover tables. This is because discover is deployed on query nodes and we filter by local nodes. This PR is responsible for exposing query nodes and distributed nodes to snuba admin tracing tool.


dbanda (approved):

evanh (approved):
Does the widget that shows the storage list already use the query node/dist nodes?

PR #3878, opened 2023-03-17
fix(replays): set transaction as empty string instead of null

JoshFerge:
on clickhouse 20 it seems that we have to specify an empty string for `title` as it is a non-nullable column. On newer versions of clickhouse this does not cause problems, but we were seeing issues in self-hosted. Sentry employees did not see this as generally everyone is on an M1 and uses clickhouse 21.

In CI we run clickhouse 20, but we also do not have adequate test coverage for actual DB insertion, and this is something i plan to follow up on. additionally, we don't use this column anywhere so i'm thinking we should plan on removing it / deprecating it.

Solves https://github.com/getsentry/self-hosted/issues/2002

### DIFF
--- a/snuba/datasets/processors/replays_processor.py
+++ b/snuba/datasets/processors/replays_processor.py
@@ -80,7 +80,10 @@ class ReplaysProcessor(DatasetMessageProcessor):
         self, processed: MutableMapping[str, Any], replay_event: ReplayEventDict
     ) -> None:
         tags = process_tags_object(replay_event.get("tags"))
-        processed["title"] = tags.transaction
+
+        # we have to set title to empty string as it is non-nullable,
+        # and on clickhouse 20 this throws an error.
+        processed["title"] = tags.transaction or ""
         processed["tags.key"] = tags.keys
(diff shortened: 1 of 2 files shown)


MeredithAnya:
I might not know this code well enough to know if this is a relevant suggestion but if the `title` is the problem why not change 

```python
processed["title"] = tags.transaction
```

to be 

```python
processed["title"] = tags.transaction or "" 
```

from https://github.com/getsentry/snuba/blob/master/snuba/datasets/processors/replays_processor.py#L83

JoshFerge:
> I might not know this code well enough to know if this is a relevant suggestion but if the `title` is the problem why not change
> 
> ```python
> processed["title"] = tags.transaction
> ```
> 
> to be
> 
> ```python
> processed["title"] = tags.transaction or "" 
> ```
> 
> from https://github.com/getsentry/snuba/blob/master/snuba/datasets/processors/replays_processor.py#L83

great suggestion @MeredithAnya, thank you! this simplifies the code a bit. i've also added a comment explaining that line.

PR #3884, opened 2023-03-21
ci: Run Rust linter

lynnagara:


untitaker (commented):
wdyt about putting this into precommit? this way, rust linter also only runs if rust files are changed

dbanda (approved):

Makefile
```diff
@@ -66,3 +66,7 @@ watch-rust-snuba:
 generate-config-docs:
 	pip install -U -r ./docs-requirements.txt
 	python3 -m snuba.datasets.configuration.generate_config_docs
+
+lint-rust:
+	cd rust_snuba && cargo clippy -- -W clippy::pedantic
```
untitaker, line 71: This just warns FWIW, the exit code is still 0. we should fix it to fail CI using `-D`

untitaker (approved):

PR #3885, opened 2023-03-21
feat(rust-arroyo): Split the message interface to support batching

lynnagara:
Same as we did in Python Arroyo https://github.com/getsentry/arroyo/pull/134.

The processing strategy interface needs to support messages that represent a batch or a filter message - i.e. not necessarily a 1:1 relationship between messages from the Kafka broker and a message passed to the processing strategy. This means a message can be associated with a number of partitions and offsets, not just a single one.

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/mod.rs
@@ -217,7 +217,7 @@ mod tests {
     use crate::backends::local::broker::LocalBroker;
     use crate::backends::local::LocalConsumer;
     use crate::backends::storages::memory::MemoryMessageStorage;
-    use crate::types::{Message, Partition, Position, Topic};
+    use crate::types::{Message, Partition, Topic};
     use crate::utils::clock::SystemClock;
     use std::collections::HashMap;
     use std::time::Duration;
@@ -234,10 +234,7 @@ mod tests {
                 Some(message) => Some(CommitRequest {
                     positions: HashMap::from([(
                         message.partition.clone(),
-                        Position {
-                            offset: message.offset,
-                            timestamp: message.timestamp,
-                        },
+                        message.offset,
                     )]),
                 }),
             }
--- a/rust_snuba/rust_arroyo/src/processing/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/mod.rs
@@ -115,15 +115,11 @@ impl<'a, TPayload: 'static + Clone> StreamProcessor<'a, TPayload> {
             let msg = self.consumer.poll(Some(Duration::ZERO));
             //TODO: Support errors properly
             match msg {
-                Ok(m) => {
-                    match m {
-                        None => {
-                            self.message = None;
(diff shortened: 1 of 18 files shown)


rust_snuba/rust_arroyo/src/processing/mod.rs
```diff
@@ -115,7 +115,16 @@ impl<'a, TPayload: 'static + Clone> StreamProcessor<'a, TPayload> {
             let msg = self.consumer.poll(Some(Duration::ZERO));
             //TODO: Support errors properly
             match msg {
-                Ok(m) => self.message = m,
+                Ok(m) => {
+                    match m {
+                        None => {
```
untitaker, line 120: *nit* you can fold this inner match into the outer one like so:

```
match msg {
    Ok(None) => { ... },
    Ok(Some(inner)) => { ... }
}
```

rust_snuba/rust_arroyo/src/types/mod.rs
```diff
@@ -35,36 +36,125 @@ pub enum TopicOrPartition {
...
+            InnerMessage::BrokerMessage(BrokerMessage{payload, ..}) => payload.clone(),
+            InnerMessage::AnyMessage(AnyMessage{payload, ..}) => payload.clone(),
+        }
+    }
+
+    pub fn committable(&self) -> BTreeMap<Partition, u64> {
+        match &self.inner_message {
+            InnerMessage::BrokerMessage(BrokerMessage{partition, offset, ..}) => {
+                let mut map = BTreeMap::new();
+                // TODO: Get rid of the clone
```
untitaker, line 111: Partition IMO should be Copy, but you can't do that because the topic is contained as a String. You could make it so that the topic names are statically allocated somewhere (easily achievable using `Box::leak`), and then all the Partitions can hold a `&'static str` for the topic name which is just a pointer to some string that Rust now knows will live forever

rust_snuba/rust_arroyo/src/types/mod.rs
```diff
@@ -35,36 +36,125 @@ pub enum TopicOrPartition {
...
+                map
+            },
+            InnerMessage::AnyMessage(AnyMessage{committable, ..}) => {
+                committable.clone()
+            }
+        }
+
+    }
+
+    pub fn replace<TReplaced: Clone>(&self, replacement: TReplaced) -> Message<TReplaced> {
```
untitaker, line 122: I believe this should also be `self`. If people need to retain the original message, they can `.clone().replace()` IMO

untitaker (changes requested):
what is the reason for making this an enum? I don't think broker vs transformed message is represented as enum in python either.

also please run `cargo fmt` and `cargo clippy`, i will add CI shortly

comments marked with `*nit*` are just stilistic things that shouldn't block merge

untitaker (approved):

untitaker:
> what is the reason for making this an enum? 

tldr from slack, this split also exists in Python in the form of Value vs BrokerValue

PR #3886, opened 2023-03-21
ref: Include rust_arroyo in Dockerfile

untitaker:
This will allow us to run `snuba rust-consumer` from docker.

The build now happens in roughly four steps:

1. Initialize `build_base` layer, which contains common debian packages we need
  to have around for building everything else + requirements-build.txt

2. Then, based on `build_base`, there are _three_ build stages that (at
  least on my machine) run in parallel:

   * `base` is just `build_base` with requirements.txt + additional runtime
     deps installed, and all the build dependencies stripped out.

   * `build_admin_ui` is `build_base` with JS dependencies installed, and
    the admin UI built

   * `build_rust_snuba` is `build_base` with Rust installed, and the
    `rust_snuba` package built.

3. Based on `base`, `application` is built. Build artifacts from
   `build_admin_ui` and `build_rust_snuba` are selectively copied into
   the container. The entire git repo is copied into the container so
   that `snuba`-the-python-package can be installed.

4. The `testing` image is built from `application` as before.

The main changes are in step 2, which used to be synchronous
installation of additional runtime dependencies + admin UI build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,7 +75,7 @@ RUN set -ex; \
     cd ./rust_snuba/; \
     export PATH="$HOME/.cargo/bin/:$PATH"; \
     # use git CLI to avoid OOM on ARM64
-    echo -e "[net]\ngit-fetch-with-cli = true" > $CARGO_HOME/config; \
+    echo -e "[net]\ngit-fetch-with-cli = true" > ~/.cargo/config; \
     maturin build --release --compatibility linux --locked --strip
 
 # Install nodejs and yarn and build the admin UI
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,7 +75,8 @@ RUN set -ex; \
     cd ./rust_snuba/; \
     export PATH="$HOME/.cargo/bin/:$PATH"; \
     # use git CLI to avoid OOM on ARM64
-    echo -e "[net]\ngit-fetch-with-cli = true" > ~/.cargo/config; \
+    echo '[net]' > ~/.cargo/config; \
+    echo 'git-fetch-with-cli = true' >> ~/.cargo/config; \
     maturin build --release --compatibility linux --locked --strip
 
 # Install nodejs and yarn and build the admin UI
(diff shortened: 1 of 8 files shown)


cloudbuild.yaml
```diff
@@ -12,7 +12,7 @@ steps:
         "-f",
         "./Dockerfile",
       ]
-    timeout: 180s
+    timeout: 1200s
```
untitaker, line 15: this timeout is necessary for a full rebuild. Rust compilation is very slow. It should however be cached for almost all PRs

lynnagara (approved):

Dockerfile
```diff
@@ -52,20 +90,23 @@ RUN set -ex; \
...
     yarn run build && \
     yarn cache clean && \
-    rm -rf node_modules && \
-    apt-get purge -y --auto-remove yarn curl nodejs gnupg && \
-    rm -rf /var/lib/apt/lists/* /etc/apt/sources.list.d/yarn.list
+    rm -rf node_modules
 
 # Layer cache is pretty much invalidated here all the time,
 # so try not to do anything heavy beyond here.
-FROM $BUILD_BASE AS application
```
dbanda, line 61: This was a parameter because we wanted to allow switching the base between `build_admin_ui` and `base`. This was because we wanted to allow skipping the `build_admin_ui` layers when we were building on ARM, because currently the ARM image is built inside an emulator and is really slow. AFAIK github doesn't provide us with ARM vms to speed this up. The trick we used was to build the admin `bundle.js` outside of docker then copy it in for the ARM image. I'd imagine if we are building rust it would take a long time in the emulator so we may want to do something similar.

untitaker: OMG @dbanda thank you for pointing this out!! I haven't gotten into debugging the ARM failure yet but it makes perfect sense.

asottile-sentry:
ah yeah this is the OOM thingy -- https://github.com/getsentry/relay/blob/011e56bfc95b73bd3e8f3790a509ca1089039fe0/Dockerfile#L28

PR #3887, opened 2023-03-21
fix(admin): Avoid get_local and get_dist nodes for discover in admin

enochtangg:
The Discover dataset is now exposed in snuba admin tracing tool. However, in production, the discover dataset does not follow the typical snuba cluster pattern. Internally, the discover cluster identifies itself as a non-single node, but is not part of a cluster. 

As a result, this [check](https://github.com/getsentry/snuba/blob/c7d920648a24d2f66c0248cf91423b33ba110f60/snuba/clusters/cluster.py#L325) will always fail for discover. To make this work in admin, we could either refactor `ClickhouseCluster` to work with discover, or we can just update the admin API to accommodate for this one-off case which is what this PR does.

### DIFF
--- a/snuba/admin/clickhouse/nodes.py
+++ b/snuba/admin/clickhouse/nodes.py
@@ -40,12 +40,20 @@ def _get_nodes(storage_key: StorageKey, local: bool = True) -> Sequence[Node]:
     try:
         storage = get_storage(storage_key)
         cluster = storage.get_cluster()
-        return [
-            {"host": node.host_name, "port": node.port}
-            for node in (
-                cluster.get_local_nodes() if local else cluster.get_distributed_nodes()
-            )
-        ]
+        if storage_key == StorageKey.DISCOVER:
+            # Discover does not follow typical cluster pattern.
+            # The get_nodes cluster methods would result in an error because
+            # discover is not a single node, but also does not belong to any cluster.
+            return []
+        else:
+            return [
+                {"host": node.host_name, "port": node.port}
+                for node in (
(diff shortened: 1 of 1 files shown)


evanh (commented):
What is the actual error this is fixing?

enochtangg:
> What is the actual error this is fixing?

@evanh The one linked in the description: https://github.com/getsentry/snuba/blob/c7d920648a24d2f66c0248cf91423b33ba110f60/snuba/clusters/cluster.py#L325. For discover prod settings, we specify that discover is not a single node cluster, but it isn't part of any cluster name. Discover only has query nodes.

evanh:
Ah right, sorry. Missed the link.

> it isn't part of any cluster name

Should it be? Could we add a cluster name to make it consistent, instead of adding more exceptions?

dbanda:
Will this still allow us to run system queries on discover? How is that still working if this assert was failing?

enochtangg:
> Should it be? Could we add a cluster name to make it consistent, instead of adding more exceptions?

From my understanding, the `cluster_name` maps to real provisioned clusters with specified shards. So I'm not sure if we can just create a new name. To me, it feels like discover doesn't fit the snuba cluster model properly. If this discover use-case is something we'd like to support for future datasets, then maybe these are grounds for changing the snuba cluster model. But if not, I think adding an exception in admin can work for now.

> Will this still allow us to run system queries on discover? How is that still working if this assert was failing?
It should. System queries tool and tracing tool both use the `clickhouse_nodes` endpoint.

evanh (approved):

PR #3890, opened 2023-03-21
migration(search_issues): rebuild search_issues with different primary key

barkbarkimashark:
We should be using `client_timestamp` as part of the primary key instead of `receive_timestamp`. 

This migration is nearly idential to `0004_rebuild_search_issues_with_version`, where we had to rebuild the table. The only difference should be the primary key change:
Screenshot 2023-03-21 at 2 33 20 PM

### DIFF
--- a/snuba/datasets/processors/search_issues_processor.py
+++ b/snuba/datasets/processors/search_issues_processor.py
@@ -194,6 +194,9 @@ class SearchIssuesMessageProcessor(DatasetMessageProcessor):
             event_occurrence_data["detection_time"]
         )
         receive_timestamp = datetime.utcfromtimestamp(event_data["received"])
+        retention_days = enforce_retention(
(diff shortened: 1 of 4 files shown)


snuba/datasets/processors/search_issues_processor.py
```diff
@@ -215,6 +212,10 @@ def process_insert_v1(
                 )
             client_timestamp = _client_timestamp
 
+        retention_days = enforce_retention(
```
MeredithAnya, line 215: does this need to be part of the migration PR? can this be merged independently?

fpacifici (changes requested):

barkbarkimashark:
Closing this in favor of the following PRs:
https://github.com/getsentry/snuba/pull/3894
https://github.com/getsentry/snuba/pull/3895
https://github.com/getsentry/snuba/pull/3896

PR #3892, opened 2023-03-21
feat(entities): Add xproj_errors entity for the errors dataset

NisanthanNanthakumar:
## Objective:

We want to create a new metrics entity for `errors` that we can query for the new Escalating Forecasts feature.

I tested manually with the following snql query:

```
{
    "query": "MATCH (xproj_errors) SELECT event_id WHERE timestamp >= toDateTime('2023-03-17 00:00:00') AND timestamp < toDateTime('2023-03-23 20:00:00') AND project_id = 2",
    "dataset": "errors",
    "turbo": false,
    "consistent": false,
    "debug": false,
    "dry_run": false,
    "legacy": false,
    "app_id": "default"

}
```

And was able to validate that events from my local clickhouse was returned.


onewland:
The tests are not passing (the failure is very simple to remedy but if you have to make more snuba changes I think it's worth getting used to the local testing flow)

evanh (commented):
Can you provide some context on this change? Why do you need a new entity?

armenzg:
> Can you provide some context on this change? Why do you need a new entity?

@onewland Could you please handle this question as you have more context?

PR #3893, opened 2023-03-22
feat(rust-consumer): Add basic batching strategy

lynnagara:
Not yet used in the Rust consumer, will be integrated as a follow up

### DIFF
--- a/rust_snuba/Cargo.toml
+++ b/rust_snuba/Cargo.toml
@@ -25,3 +25,6 @@ serde = { version = "1.0", features = ["derive"]}
 serde_json = { version = "1.0" }
 glob = "0.3.1"
 pyo3 = { version = "0.18.1", features = ["chrono", "extension-module"] }
+
+[dev-dependencies]
+chrono = "0.4.11"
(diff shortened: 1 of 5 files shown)


rust_snuba/Cargo.toml
```diff
@@ -25,3 +25,6 @@ serde = { version = "1.0", features = ["derive"]}
 serde_json = { version = "1.0" }
 glob = "0.3.1"
 pyo3 = { version = "0.18.1", features = ["chrono", "extension-module"] }
+
+[dev-dependencies]
+chrono = "0.4.11"
```
lynnagara, line 30: I needed this for the test since it's part of the Message interface in rust_arroyo. Though I wonder if we can get rid of it in arroyo too.

untitaker: I don't know at the moment, I don't think we need to solve this problem atm though. I am not sure if chrono is the best library for our needs

lynnagara:
Superceded by https://github.com/getsentry/snuba/pull/4061

PR #3894, opened 2023-03-22
migration(search-issues): create a new table with proper primary key

barkbarkimashark:
We incorrectly have the primary key set to `receive_timestamp` when it should be `client_timestamp`. 

We'll create a new table with the proper key and look to move over writes and reads to this table in steps.

### DIFF
--- a/snuba/snuba_migrations/search_issues/0005_search_issues_v2.py
+++ b/snuba/snuba_migrations/search_issues/0005_search_issues_v2.py
@@ -1,0 +1,132 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import (
+    UUID,
+    Array,
+    Column,
+    DateTime,
+    IPv4,
+    IPv6,
+    Nested,
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/search_issues/0005_search_issues_v2.py
```diff
@@ -0,0 +1,132 @@
...
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.datasets.storages.tags_hash_map import TAGS_HASH_MAP_COLUMN
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
```
fpacifici, line 21: What do you need organization id for? It is not in the primary key.

fpacifici (approved):

snuba/snuba_migrations/search_issues/0005_search_issues_v2.py
barkbarkimashark: I think this was carry over and probably not used but I don't see any harm in keeping it.

PR #3895, opened 2023-03-22
migration(search-issues): move over writes to the newly created table with proper primary key

barkbarkimashark:

PR #3896, opened 2023-03-22
migration(search-issues): finally drop old table

barkbarkimashark:
After moving all writes and reads to the newly created table we need to drop the old table as its no longer being used.


evanh (approved):
Please do not merge this before https://github.com/getsentry/snuba/pull/3895 is merged to master.

PR #3897, opened 2023-03-22
feat: Type querylog producer based on type generated from schema

lynnagara:
This should work but it doesn't.


evanh:
> This should work but it doesn't.

Is this ready for review or are you still working on it lol.

lynnagara:
I would like to merge it but there will be some prerequisites. If you look at the failing mypy check, it illustrates a couple of issues with the schema currently.

lynnagara:
@untitaker helped me, this should be ready for review now.

PR #3900, opened 2023-03-22
feat(CI): add check to detect migrations changes coupled with others

dbanda:
There's an ongoing effort to make migrations run in a more automated manner. As such, there's a need to add some guardrails to check that migrations aren't coupled with other code changes. This makes them more easily reversible if something goes wrong during when apply a migration.

This PR checks that when changes are made to migrations (in `snuba_migrations`), no other files outside the `snuba_migrations` are touched. It adds two scripts to run in CI:

* `check_migrations.py` :  a script that diff HEAD with a given sha and throws a `CoupledMigrations` exceptions if the diff has a mixture of migrations and non-migrations changes. 
* `fetch_service_ref.py` :  a script that uses the GoCD REST api to fetch the most recent successfully deployed sha from the GoCD pipeline. 

To first test this, we are adding it to the `snuba-sns-test.yaml` for test deployments, then in a follow up PR will will add it to the main gocd pipeline.

### DIFF
--- a/gocd/pipelines/snuba-sns-test.yaml
+++ b/gocd/pipelines/snuba-sns-test.yaml
@@ -24,6 +24,18 @@ pipelines:
                 branch: master
                 destination: snuba
         stages:
+            - checks:
+                  approval:
+                      type: manual
+                  fetch_materials: true
+                  jobs:
+                      checks-migrations:
+                          timeout: 1800
+                          elastic_profile_id: snuba
+                          tasks:
+                              - script: |
+                                    deploy_sha=`python scripts/fetch_service_refs.py --pipeline "deploy-snuba-sns-test"` && \
+                                    python scripts/check-migrations.py --to $deploy_sha
             - deploy:
                   fetch_materials: true
                   jobs:
--- a/gocd/pipelines/snuba-sns-test.yaml
+++ b/gocd/pipelines/snuba-sns-test.yaml
@@ -15,6 +15,7 @@ pipelines:
             GKE_REGION: us-central1
             GKE_CLUSTER_ZONE: a
             GKE_BASTION_ZONE: a
+            MIGRATIONS_READINESS: complete
         group: snuba
         lock_behavior: unlockWhenFinished
         materials:
@@ -108,7 +109,7 @@ pipelines:
                                     --label-selector="service=snuba-admin" \
                                     --container-name="snuba-admin" \
                                     "snuba-migrate" "ghcr.io/getsentry/snuba-ci:${GO_REVISION_SNUBA_REPO}" \
-                                    -- snuba migrations migrate --force
+                                    -- snuba migrations migrate -r ${MIGRATIONS_READINESS}
                               - plugin:
                                     options:
                                         script: |
--- a/gocd/pipelines/snuba-sns-test.yaml
+++ b/gocd/pipelines/snuba-sns-test.yaml
@@ -20,7 +20,7 @@ pipelines:
         materials:
             snuba_repo:
                 git: https://github.com/getsentry/snuba.git
-                shallow_clone: true
+                shallow_clone: false
                 branch: master
                 destination: snuba
         stages:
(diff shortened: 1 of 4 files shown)


gocd/pipelines/snuba-sns-test.yaml
```diff
@@ -15,15 +15,28 @@ pipelines:
...
             GKE_CLUSTER_ZONE: a
             GKE_BASTION_ZONE: a
+            MIGRATIONS_READINESS: complete
         group: snuba
         lock_behavior: unlockWhenFinished
         materials:
             snuba_repo:
                 git: https://github.com/getsentry/snuba.git
-                shallow_clone: true
+                shallow_clone: false
```
nikhars, line 24: Is this set to false because you need reference to other SHA's?

scripts/check-migrations.py
```diff
@@ -0,0 +1,57 @@
...
+not coupled with other changes.
+"""
+
+
+class CoupledMigrations(Exception):
+    pass
+
+
+def _get_migration_changes(workdir: str = ".", to: str = "origin/master") -> str:
+    return _get_changes("snuba/snuba_migrations/*/[0-9]*.py", workdir, to)
```
evanh, line 17: This needs to also allow changing the migration group to add the migration (https://github.com/getsentry/snuba/blob/master/snuba/migrations/groups.py).

You might also want to allow test files.

scripts/check-migrations.py
```diff
@@ -0,0 +1,57 @@
...
+This script is meant to be run in CI to check that migrations changes are
+not coupled with other changes.
+"""
+
+
+class CoupledMigrations(Exception):
+    pass
+
+
+def _get_migration_changes(workdir: str = ".", to: str = "origin/master") -> str:
```
nikhars, line 16: Nit: Since you have default parameters defined in main itself, you can remove the defaults from all other places.

scripts/fetch_service_refs.py
```diff
@@ -0,0 +1,96 @@
...
+import json
+import os
+import sys
+import time
+import urllib.error
+import urllib.request
+from typing import Any, Dict, Optional
+
+GO_SERVER_URL = os.environ.get(
+    "GO_SERVER_URL", "http://gocd-server.gocd.svc.cluster.local:8153/go"
```
nikhars, line 14: Probably you should rely on environment being set only and not provide a default. The defaults are something which are useful only in our SaaS environments. If we want to use this on Single Tenant or EU regions, these would be wrong.

scripts/fetch_service_refs.py
```diff
@@ -0,0 +1,96 @@
...
+                        print(rev)
+                        return 0
+                    elif (
+                        f"https://github.com/getsentry/{repo}.git"
+                        in r["material"]["description"]
+                    ):
+                        rev = r["modifications"][0]["revision"]
+                        print(rev)
+                        return 0
+        time.sleep(1)
```
nikhars, line 83: Are there cases where you might never exit this while loop? Should we have a maximum number of retries or maximum time for the job to check?

nikhars (commented):
Overall looks ok to me overall. Left some comment. You should probably get someone from dev-infra to also review this PR.

gocd/pipelines/snuba-sns-test.yaml
dbanda: yes

scripts/fetch_service_refs.py
dbanda: removed it

scripts/fetch_service_refs.py
dbanda: we should always exist since we fetch the pagination link for the response and break if it is null. I added a max fetches parameter anyway just in case.

scripts/check-migrations.py
dbanda: fixed. I updated it to use an `ALLOWED_MIGRATIONS` list that ignores tests, docs and groups.py in the git diff.

scripts/check-migrations.py
dbanda: removed

evanh (approved):

PR #3903, opened 2023-03-23
ref(db_query): Stop passing SnubaQueryMetada into `raw_query`

volokluev:
### Context

While adding the [allocation policy hook](https://github.com/getsentry/snuba/pull/3833) and trying to write a test, a few things became clear:

1. The `raw_query` function has a lot of untested behavior
2. The `raw_query` function is passed a lot of things through `SnubaQueryMetadata` such that in order to write a test for it, I would have to instantiate an entire `Request` 

This PR makes the following changes:

1. Passes `AttributionInfo` to `raw_query` (it was already being passed and used before, just not explicitly)
2. Passes `dataset_name` to `raw_query` (it was already being passed and used before, just not explicitly)
3. Passes the part `query_list` attribute of `SnubaQueryMetadata` to `raw_query` instead of all of `SnubaQueryMetadata`
4. Writes the first test for the success case and failure case of `raw_query` 

(3) was done in order to limit and make explicit the state being modified by `db_query`. 

#### Other small changes

* `MutableMapping` was changed to `dict` in `db_query.py` because I kept getting mypy errors and they were all dicts anyways
* QueryRunner can now be used with kwargs. I did this because when changing the signatures of these functions, it was very confusing to count positional arguments
* Documents some gotchas of the code that were not obvious until I dove deep into it

### Blast Radius

This function and code is on the critical path, all snuba functionality is affected by this


snuba/web/db_query.py
```diff
@@ -154,8 +152,8 @@ def execute_query(
     formatted_query: FormattedQuery,
     reader: Reader,
     timer: Timer,
-    stats: MutableMapping[str, Any],
-    clickhouse_query_settings: MutableMapping[str, Any],
+    stats: Dict[str, Any],
```
untitaker, line 155: why the change from MutableMapping to Dict?

untitaker (changes requested):

untitaker (commented):
wrong button

snuba/web/query.py
```diff
@@ -164,15 +165,21 @@ def _run_query_pipeline(
...
+            concurrent_queries_gauge=concurrent_queries_gauge,
         )
 
     record_missing_tenant_ids(request)
+    # NOTE: Something which is not immediately clear from looking at the code is that the
+    # query_runner can be run multiple times during the execution of the pipeline.
+    # The execution pipeline may choose to break up a query into multiple subqueries. And
+    # then assemble those together into one resut
 
+    # Throughout those executions, the query_metadata.query_list is appended to every time a query runs
```
enochtangg, line 181: Nice! One question I have is: are there any differences between passing `query_metadata. query_list` by reference into `raw_query()` and passing the entire object when it comes to safety with multiple executions? Do we parallelize this process?

snuba/web/query.py
```diff
@@ -164,15 +165,21 @@ def _run_query_pipeline(
...
+    # within `db_query.py` with metadata about the query. That metada then goes into the querylog
```
enochtangg, line 182: nit: metadata

tests/web/test_db_query.py
```diff
@@ -93,8 +108,107 @@ def test_apply_thread_quota(
...
+            final=False,
+        ),
+        selected_columns=[
+            SelectedExpression(
+                "bad_column", parse_clickhouse_function("count(non_existent_column)")
+            )
+        ],
+    )
+
+    query_metadata_list: list[ClickhouseQueryMetadata] = []
```
enochtangg, line 183: nit `list` -> `MutableSequence`

enochtangg (commented):
Looks good. Just had one question.

snuba/web/db_query.py
evanh: Typing general principle is to be as relaxed as possible for inputs and strict as possible for outputs.

And if you are going to use `Dict`, please use `dict` instead.

snuba/web/query.py
```diff
@@ -164,15 +165,21 @@ def _run_query_pipeline(
...
-            concurrent_queries_gauge,
+            timer=timer,
+            query_metadata=query_metadata,
+            attribution_info=request.attribution_info,
+            robust=robust,
+            concurrent_queries_gauge=concurrent_queries_gauge,
         )
 
     record_missing_tenant_ids(request)
+    # NOTE: Something which is not immediately clear from looking at the code is that the
```
evanh, line 176: This should probably be in the doc string.

tests/web/test_db_query.py
```diff
@@ -93,8 +108,107 @@ def test_apply_thread_quota(
...
-    clickhouse_query_settings: MutableMapping[str, Any] = {}
+    clickhouse_query_settings: dict[str, Any] = {}
     _apply_thread_quota_to_clickhouse_query_settings(
         settings, clickhouse_query_settings, rate_limit_stats
     )
     assert clickhouse_query_settings == expected_query_settings
+
+
+def test_raw_query_success() -> None:
+    storage = get_storage(StorageKey("errors"))
```
evanh, line 119: I would move all of this object building into a helper function(s) that can be called/overriden by both tests.

evanh (commented):
Are most of these changes just to avoid having a test that instantiates a Request? That seems like something that could be easily done, and would only need to be done once and then all the tests could reuse that functionality.

tests/web/test_db_query.py
rahul-kumar-saini: `list` from future annotations > `MutableSequence` imo

snuba/web/query.py
volokluev: Good spot! The answer to your question is Yes and No.

There is the [MultipleConcurrentPipeline](https://github.com/getsentry/snuba/blob/master/snuba/pipeline/pipeline_delegator.py#L73) which takes a `runner` callable and runs different versions of the pipeline in different threads. In the case that is used, this would indeed be a race condition.

Currently [`SnubaQueryMetadata`](https://github.com/getsentry/snuba/blob/master/snuba/querylog/query_metadata.py#L209) does not do any sort of locking on the `query_list` so there is nothing preventing this race condition.

All that being said, the `MultipleConcurrentPipeline` is not actually used anywhere currently but even when it was, we did have this problem. Also, python threads are not truly parallel so we are fine on this front overall

volokluev: I updated the comment to provide more info. Let me know if that makes sense

tests/web/test_db_query.py
volokluev: gonna ignore this nit. it is a list after all

snuba/web/db_query.py
volokluev: I undid these changes because they're just adding noise

snuba/web/query.py
volokluev: done

volokluev:
Hi everyone, could use another review on this. Pretty sure I've addressed everyone's comments. To reply to @evanh about whether this change is just to avoid instantiating requests in tests. It is not. It is about:

1. Separation of concerns of what runs clickhouse queries from the rest of the code
2. Making explicit the state that is carried between different parts of the code

snuba/web/query.py
enochtangg: Cool, thanks for the clarification and the added comment 👍

PR #3904, opened 2023-03-23
feat(DRS): Add SUPPORTED_STATES settings in dev and self-hosted

enochtangg:
### Overview
This PR is responsible for adding `SUPPORTED_STATES` settings in dev and self-hosted settings files. Right now, there are four states: `deprecate`, `limited`, `partial`, and `complete`. These readiness states map to snuba specific environments within Sentry are used to determine which storages and migrations groups are available within an environment.

### Blast Radius
These settings will not do anything after being merged because Dataset Readiness States have not been implemented yet. Side note, the rollout for readiness states will be control by a runtime config and updated for less critical datasets first.

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -4,7 +4,6 @@ import os
 from pathlib import Path
 from typing import (
     Any,
-    List,
     Mapping,
     MutableMapping,
     Optional,
@@ -263,7 +262,7 @@ if os.environ.get("ENABLE_AUTORUN_MIGRATION_SEARCH_ISSUES", False):
     SKIPPED_MIGRATION_GROUPS.remove("search_issues")
 
 # Dataset readiness states supported in this environment
-SUPPORTED_STATES: List[str] = ["deprecate", "limited", "partial", "complete"]
+SUPPORTED_STATES: Set[str] = {"deprecate", "limited", "partial", "complete"}
 
 MAX_RESOLUTION_FOR_JITTER = 60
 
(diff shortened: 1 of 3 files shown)


snuba/settings/__init__.py
```diff
@@ -261,6 +262,8 @@ class RedisClusters(TypedDict):
 if os.environ.get("ENABLE_AUTORUN_MIGRATION_SEARCH_ISSUES", False):
     SKIPPED_MIGRATION_GROUPS.remove("search_issues")
 
+# Dataset readiness states supported in this environment
+SUPPORTED_STATES: List[str] = ["deprecate", "limited", "partial", "complete"]
```
evanh, line 266: This can probably be a set.

snuba/settings/settings_self_hosted.py
```diff
@@ -16,4 +17,7 @@
 DOGSTATSD_HOST = env("DOGSTATSD_HOST")
 DOGSTATSD_PORT = env("DOGSTATSD_PORT")
 
+# Dataset readiness states supported in this environment
+SUPPORTED_STATES: List[str] = ["complete"]
```
evanh, line 21: I think it's better to leave this the same as the other environments for now.

If we make this setting different at first, then this will change for ST _implicitly_ when we deploy the backend support for the setting.

I think it's better to deploy the support on the backend first, and the _explicitly_ change the setting to restrict things.

enochtangg: That makes sense to me. Updated self-hosted settings to support all readiness states for now (until support for backend is done).

evanh (approved):

PR #3906, opened 2023-03-23
ref(migrations): lower CLICKHOUSE_SERVER_MIN_VERSION

MeredithAnya:
**context:**
We are trying to enable metrics on ST instances to replace sessions and to do so we need to run migrations for the metrics group. We are using the migrations tooling to do so, but we ran into https://github.com/getsentry/snuba/blob/ee7032dace7efa7b8035af2a18d570539fd36ddb/snuba/migrations/connect.py#L52-L57 
since one of our ST instances runs on version `20.3.5.21`.

We would like to upgrade clickhouse for this instance but that is out of scope for the project and so we are lowering the version for the time being to allow the tool to work. 

**testing:**
I've tested that this works by using https://github.com/getsentry/clickhouse-setup/tree/master/examples/new_replica. I changed the clickhouse image versions to `20.3.5.21` for the nodes in that set up and then I ran `snuba migrations migrate --force` and it ran successfully.


onewland (approved):
I begrudgingly approve

PR #3908, opened 2023-03-23
migration(replays): Allow null for non-required columns

cmanallen:
When the replays service internally generates rows some field values are not available without reading from the dataset.  We would rather mark these fields as null to not interfere with the aggregation process.

Note: some of the files were auto-formatted so the diff appears large than it is.  The following fields are now allowed to be null:

- title
- user
- sdk_name
- sdk_version


lynnagara:
The change itself LGTM but I think it's safer to split the migration into a separate PR and ensure that gets run first. Otherwise we may be attempting to write/access the field before it exists.

lynnagara (approved):

evanh (approved):
Per offline discussion, in this case the storage schema needs to be updated in the same PR as the migration in order to avoid schema drift.

PR #3914, opened 2023-03-23
ref: Opt into database usage in tests

untitaker:
We determined that CI is too slow, and this idea explores what it would
take to opt into clickhouse and redis usage per-test, like we do in
sentry for django_db


tests/clusters/test_cluster.py
```diff
@@ -102,12 +103,15 @@
 ]
 
 
-def teardown_function() -> None:
+@pytest.fixture(autouse=True)
+def setup_teardown(clickhouse_db: None) -> Generator[None, None, None]:
```
untitaker, line 107: stuff like this really sucks: I have to explicitly depend on the clickhouse_db fixture so that it is setup before this fixture, and torn down after it

this matters a lot for fixtures where some test data is inserted

evanh (commented):
Does it make more sense to try the opt out instead of opt in? It seems like this is changing a lot of tests, but it's hard to tell if opt in would require less.

untitaker:
@evanh I would say that opt-in has the significant advantage of making it harder to write accidentally slow tests (by unnecessarily depending on the DB)

(note that this is essentially the same model as we have in sentry with pytest-django's django_db decorator)

evanh:
> significant advantage of making it harder to write accidentally slow tests

OK that makes sense to me then.

evanh (approved):

lynnagara (approved):

PR #3920, opened 2023-03-24
fix(migrations): skip checking local/dist nodes for discover cause it doesn't have them

dbanda:
Another follow up to #3887 

Discover doesn't have dist/local nodes so it was breaking the the check and blocked migrations.

### DIFF
--- a/snuba/migrations/connect.py
+++ b/snuba/migrations/connect.py
@@ -1,13 +1,15 @@
 import logging
 import re
 import time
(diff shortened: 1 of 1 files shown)


untitaker (approved):

MeredithAnya:
I'm confused doesn't discover still have dist nodes? or dont we still need to be able to run migrations on the `discover_dist` tables?

PR #3924, opened 2023-03-24
ref(migrations): run multiple migrations through a specifc migration id

MeredithAnya:
Specifying the group and prefix will migration through that migration (includes the matching migration)

this should also work for `--fake`

```shell
12:35 $ snuba migrations migrate -g generic_metrics 0012
2023-03-24 12:35:29,689 Initializing Snuba...
2023-03-24 12:35:31,869 Snuba initialization took 2.181490815s
2023-03-24 12:35:31,895 Running migration: 0012_counters_mv
2023-03-24 12:35:31,957 Finished: 0012_counters_mv
2023-03-24 12:35:31,965 Ran through: 0012_counters_mv
Finished running migrations
```

If the migration is already run or the arg doesn't match, get an error
```shell
12:34 $ snuba migrations migrate -g generic_metrics 0011
2023-03-24 12:35:23,842 Initializing Snuba...
2023-03-24 12:35:26,178 Snuba initialization took 2.337297151s
Error: No pending migration matching: 0011
```

```
12:28 $ snuba migrations migrate -g generic_metrics 0089
2023-03-24 12:34:41,059 Initializing Snuba...
2023-03-24 12:34:43,528 Snuba initialization took 2.470212598s
Error: No migration matching: 0089
```

### DIFF
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -59,7 +59,11 @@ def list() -> None:
     "--log-level", help="Logging level to use.", type=click.Choice(LOG_LEVELS)
 )
 def migrate(
-    group: str, through: str, force: bool, fake: bool, log_level: Optional[str] = None
+    group: Optional[str],
+    through: str,
+    force: bool,
+    fake: bool,
+    log_level: Optional[str] = None,
 ) -> None:
     """
     If group is specified, runs all the migrations for a group (including any pending
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -51,13 +51,15 @@ def list() -> None:
 
 
 @migrations.command()
+@click.argument("group", default="all")
+@click.argument("through", default="all")
 @click.option("--force", is_flag=True)
-@click.option("--group", help="Migration group")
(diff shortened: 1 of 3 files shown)


snuba/cli/migrations.py
```diff
@@ -71,11 +73,13 @@ def migrate(
     runner = Runner()
 
     try:
-        if group is not None:
+        if group != "all":
```
dbanda, line 76: nit: maybe its just me, but I'd rather use `group=None` to signal that all migrations should be run rather than `group='all'` makes it clear that its the default special case by having a null type, and avoid any bugs if we ever decide to make a group named 'all'

dbanda (approved):

snuba/cli/migrations.py
MeredithAnya: @dbanda It seems like to me that positional arguments don't work if you have the default be `None`, I get 

`Error: Missing argument 'GROUP'.` 

So if we want it to be an option I can change it, but I was trying to make it so you didn't have to type like `-g` or `--group`.  What are your thoughts?

PR #3931, opened 2023-03-27
fix(migrations): add passing in yes to migrations cli prompt for headless cli running

dbanda:
allow passing `--yes` to the migrations fake prompt so we can run in headless mode. Needed for k8s jobs


mwarkentin (approved):
This allowed me to run migrations from a k8s job in disney 👍🏻

MeredithAnya (approved):

PR #3933, opened 2023-03-27
ref: Remove legacy dashboard

lynnagara:
We have Snuba admin for this now


volokluev (approved):
love to see code get removed 😊

PR #3935, opened 2023-03-28
ref: Remove more dead code

lynnagara:


evanh (changes requested):
There are people who still use the web dashboard for local development, I don't think we can just get rid of it without making sure that the admin tool supports their use cases.

snuba/state/__init__.py
```diff
@@ -88,26 +87,6 @@ class MismatchedTypeException(Exception):
     new_type: Type[Any]
 
 
-def get_concurrent(bucket: str) -> Any:
```
evanh, line 91: These functions are used by the web dashboard, I don't think you can just remove them.

lynnagara:
@evanh Not sure I understand the concern. Which features are these specifically? I really don't think we should keep supporting all this legacy code. Some of it like the way options are configure is wrong.

evanh:
> If people are incorrectly using this dashboard, can we please tell them to move to Snuba admin rather than keep this around.

I agree but I don't know how that dashboard is being used 100% and I don't think the admin tool covers the same use cases. For example the dashboard has a query tool that returns actual results and we don't allow that in the admin tool right now.

evanh (approved):
I missed the previous PR that removed the rest of the dashboard. Since it's already gone, I guess let's remove this and then hope no one wants us to add it back.

PR #3939, opened 2023-03-28
feat(DRS): Add readiness_state support for storages

enochtangg:
### Overview
This PR is responsible for adding the readiness_state attribute to the storage definition. This includes the following:
* Adding support to jsonschema for storage schema
* Adding `readiness_state` label to all existing storages in config
* Updating and adding new tests  

## Blast Radius
This change will not affect the current state of snuba because snuba currently does not use these `readiness_states` yet.


evanh (approved):
Nice

PR #3942, opened 2023-03-28
feat(rate-limiting): Add an Organization + Referrer thread rate-limiter

rahul-kumar-saini:
### Overview
- We now have Organization IDs for >99% of queries
- Sometimes multiple projects of an organization are abusing a certain referrer (as is happening right now with `api.auth-token.events`)
- This PR introduces organization ID to `QuerySettings` which is picked up by the Resource Quota rate limiter
- Adding a runtime config of the form `referrer_organization_thread_quota_<referrer>_<organization_id>` will now add a thread limit to all requests for a certain referrer coming from a certain organization

End to end tested locally by setting org thread limit and querying.


volokluev:
This is okay but can you file a ticket in the capacity management epic to revert this PR once the allocation policy is in place. There should be no reason to do this in the query pipeline

volokluev (approved):
Approved but this should be reverted within a week or so

evanh (approved):

PR #3945, opened 2023-03-29
fix(outcome): Update outcome schema

john-z-yang:
### Overview
Update `outcomes` schema by bumping `sentry-kafka-schemas` version


lynnagara (approved):
One day we'll get around to automating this so they don't need a second approval

PR #3947, opened 2023-03-29
ci: add canary deployments to snuba pipelines

mattgauntseo-sentry:
With this change, the snuba pipeline will perform the following steps:

1. Deploy new changes to the canary replica
1. Wait for 5 minutes (determined by the canarychecks script flag)
	- NOTE: This delay is intended to give the deployer a chance to check for any early signs of issues and our hope is to add some automated checks to this script in the future (i.e. if any alerts trigger for the canary instances, mark the deploy as a failure)
1. Deploy new changes to all replicas (i.e. canary and primary replicas a.k.a all of prod).

For the snuba stable pipeline, this change will:

1. Deploy new changes to the canary replica *after the deployer has clicked the next step button in GoCD.
1. Wait for 5 minutes (determined by the canarychecks script flag)
1. Deploy new changes to all replicas *after the deployer has clicked the next step button in GoCD.


gocd/pipelines/snuba-stable.yaml
```diff
@@ -60,6 +60,67 @@ pipelines:
...
+                                    --container-name="replacer" \
+                                    --container-name="snuba-admin" \
+                                    --container-name="transactions-consumer-new" \
+                                    --container-name="transactions-subscriptions-executor" \
+                                    --container-name="transactions-subscriptions-scheduler" \
+                                    && /devinfra/scripts/k8s/k8sdeploy.py \
+                                    --context="gke_${GCP_PROJECT}_${GKE_REGION}-${GKE_CLUSTER_ZONE}_${GKE_CLUSTER}" \
+                                    --label-selector="service=snuba,is_canary=true" \
+                                    --image="us.gcr.io/sentryio/snuba:${GO_REVISION_SNUBA_REPO}" \
+                                    --type="cronjob" \
```
dbanda, line 106: Do canaries apply to cron jobs? If so, how do we prevent duplicate jobs?

mattgauntseo-sentry: Ah good catch, yeah I don't have any intention of adding canaries to cronjobs

evanh (commented):
Can you provide some more context for this change? Are we supposed to use this deploy now? Will it be automatic? The PR description isn't particularly helpful.

mattgauntseo-sentry:
@evanh apologies you are right. I've added a description to the PR that hopefully explains what this change will do.

Please let me know if you have any questions.

evanh (commented):
Why are there two different flows for snuba and snuba-stable?

mattgauntseo-sentry:
@evanh I'm just following the patterns of each pipeline as they were before, but happy to tweak any way you think is best.

I would say the extra manual step between canary and prod in the stable pipeline could be removed

mattgauntseo-sentry:
@evanh gotcha, this is the difference that exists at the moment: https://github.com/getsentry/snuba/blob/master/gocd/pipelines/snuba-stable.yaml#L50-L51

Shall I remove that so the two pipelines follow the same deployment process (just with different label selectors)?

And no there are no docs for what to do when a canary rollout encounters an error as we haven't got enough cases of failures to determine what to do. One team is adding alerts to datadog for canary after spotting an issue in prod that could have been caught early in canary, but that's the only anecdote I can give you at the moment.

mattgauntseo-sentry:
@evanh I've remove the additional approval steps for deploy steps in the stable pipeline (chances are this step was left over from when we first introduced the pipelines)

evanh (approved):

PR #3949, opened 2023-03-30
testing dont merge: querying exception main thread column doesn't work consistently across datasets

barkbarkimashark:


marandaneto:
Replaced by https://github.com/getsentry/snuba/pull/3958

PR #3950, opened 2023-03-30
feat: DLQ performance test

lynnagara:
We want to check if the DLQ proposal in Arroyo is viable by testing the performance impact on Snuba in production.

There are no functional changes here, however if the runtime config `run_dlq_test_{storage_name}` is set, the consumer for that storage will start buffering messages as would be needed for the DLQ.

Note that this config is only checked once upon start up of the consumer so a deploy is needed for the runtime config value to take effect.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ python-dateutil==2.8.2
 python-rapidjson==1.8
 pytz==2022.2.1
 redis==4.3.4
-sentry-arroyo[json]==2.7.1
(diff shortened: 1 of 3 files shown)


untitaker (approved):
This only sets the sample rate at the start of the consumer right? I guess there is no way around that.

lynnagara:
Yeah, it needs a deploy for changes to take effect. But it's the easiest way since I don't want to write a custom strategy or have arroyo strategies try and load from Snuba's runtime config.

PR #1968, opened 2021-06-23
ref: Refactor unmerge replacement to be class-based

untitaker:
See #1912 and #1861

PR #3957, opened 2023-03-31
ref: Refactor replace_group to class-based types

untitaker:
Also improve typing around _build_event_set_filter

Previously: https://github.com/getsentry/snuba/pull/1968

why now again? well, because I left the job half-finished and it really bugged me when I tried to add types from sentry-kafka-schemas

### DIFF
--- a/snuba/replacers/errors_replacer.py
+++ b/snuba/replacers/errors_replacer.py
@@ -278,10 +278,9 @@ class ErrorsReplacer(ReplacerProcessor[Replacement]):
                 self.__state_name,
             )
         elif type_ == ReplacementType.REPLACE_GROUP:
-            processed = process_replace_group(
+            processed = ReplaceGroupReplacement.parse_message(
                 cast(ReplacementMessage[ReplaceGroupMessageBody], message),
-                self.__all_columns,
-                self.__state_name,
+                self.__replacement_context,
             )
         elif type_ == ReplacementType.EXCLUDE_GROUPS:
             processed = ExcludeGroupsReplacement.parse_message(
@@ -447,14 +446,12 @@ def _build_group_replacement(
 
 
 def _build_event_set_filter(
-    message: Mapping[str, Any], state_name: ReplacerState
-) -> Optional[Tuple[List[str], List[str], MutableMapping[str, str]]]:
-    event_ids = message["event_ids"]
-    if not event_ids:
-        return None
-
-    def get_timestamp_condition(msg_field: str, operator: str) -> str:
-        msg_value = message.get(msg_field)
+    project_id: int,
+    event_ids: Sequence[str],
+    from_timestamp: Optional[str],
+    to_timestamp: Optional[str],
+) -> Tuple[List[str], List[str], MutableMapping[str, str]]:
+    def get_timestamp_condition(msg_value: Optional[str], operator: str) -> str:
         if not msg_value:
             return ""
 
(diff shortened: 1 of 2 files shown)


snuba/replacers/errors_replacer.py
```diff
@@ -478,50 +475,103 @@ def get_timestamp_condition(msg_field: str, operator: str) -> str:
...
+    def get_replacement_type(self) -> ReplacementType:
+        return self.replacement_type
 
-    full_where = f"PREWHERE {' AND '.join(prewhere)} WHERE {' AND '.join(where)}"
-    project_id: int = message.data["project_id"]
-    query_time_flags = (None, project_id)
+    def get_message_metadata(self) -> ReplacementMessageMetadata:
+        return self.replacement_message_metadata
 
-    return _build_group_replacement(
```
untitaker, line 517: _build_group_replacement was a bad abstraction that got in the way of refactoring, so I just inlined the relevant bits.

lynnagara:
This seems a lot cleaner in general though all the Replacement vs LegacyReplacement stuff makes it confusing... is there any reason we can't just blow away the legacy now? It's been 2 years since it was introduced so it feels like time.

untitaker:
> is there any reason we can't just blow away the legacy now?

that's the goal, but at the time it was decided that it shouldn't be one PR effectively rewriting this entire file, and that we should make incremental progress towards getting rid of function-based replacements. this is the reason LegacyReplacement was built as a compat layer.

lynnagara:
I see... so this is a step on the way to removing LegacyReplacement?

snuba/replacers/errors_replacer.py
```diff
@@ -478,50 +475,106 @@ def get_timestamp_condition(msg_field: str, operator: str) -> str:
 
     query_args = {
         "event_ids": event_id_list,
-        "project_id": message["project_id"],
+        "project_id": str(project_id),
```
lynnagara, line 478: Is this changing to string? How come?

snuba/replacers/errors_replacer.py
```diff
@@ -478,50 +475,106 @@ def get_timestamp_condition(msg_field: str, operator: str) -> str:
...
+            from_timestamp=self.from_timestamp,
+            to_timestamp=self.to_timestamp,
+        )
+
+        return (
+            f"PREWHERE {' AND '.join(prewhere)} WHERE {' AND '.join(where)}"
+            % query_args
+        )
+
+    def get_count_query(self, table_name: str) -> Optional[str]:
```
lynnagara, line 554: Curious about this interface - how come `table_name` isn't just a property on the replacement

untitaker:
> I see... so this is a step on the way to removing LegacyReplacement?

yeah the goal is to get rid of all uses of LegacyReplacement for sure

snuba/replacers/errors_replacer.py
untitaker: I did that just to satisfy typechecking, and I don't think it matters for runtime. those parameters are string-interpolated

snuba/replacers/errors_replacer.py
```diff
@@ -478,50 +475,106 @@ def get_timestamp_condition(msg_field: str, operator: str) -> str:
...
+            SELECT count()
+            FROM {table_name} FINAL
+            {self._where_clause}
+        """
+
+    def get_insert_query(self, table_name: str) -> Optional[str]:
+        all_column_names = [c.escaped for c in self.all_columns]
+        select_columns = ", ".join(
+            map(
+                lambda i: i if i != "group_id" else str(self.new_group_id),
```
lynnagara, line 565: Similar question to the project id, how come it's `str`?

lynnagara (approved):

snuba/replacers/errors_replacer.py
untitaker: tbh i think this was already like this?

PR #3958, opened 2023-03-31
Fix check_exception_main_thread default config

marandaneto:
Follow up from https://github.com/getsentry/sentry/pull/46183
Test case https://github.com/getsentry/snuba/pull/3949

### DIFF
--- a/snuba/datasets/configuration/discover/entities/discover_transactions.yaml
+++ b/snuba/datasets/configuration/discover/entities/discover_transactions.yaml
@@ -244,7 +244,6 @@ storages:
               - exception_frames.colno
               - exception_frames.lineno
               - exception_frames.stack_level
-              - exception_main_thread
               - modules.name
               - modules.version
       subscriptables:
--- a/snuba/datasets/processors/errors_processor.py
+++ b/snuba/datasets/processors/errors_processor.py
@@ -332,7 +332,7 @@ class ErrorsProcessor(DatasetMessageProcessor):
         frame_stack_levels = []
         exception_main_thread = None
 
-        check_exception_main_thread = get_config("check_exception_main_thread", 0)
+        check_exception_main_thread = get_config("check_exception_main_thread", 1)
 
         if output["project_id"] not in settings.PROJECT_STACKTRACE_BLACKLIST:
             stack_level = 0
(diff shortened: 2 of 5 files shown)


snuba/datasets/configuration/discover/entities/discover_transactions.yaml
```diff
@@ -244,7 +244,6 @@ storages:
               - exception_frames.colno
               - exception_frames.lineno
               - exception_frames.stack_level
-              - exception_main_thread
```
marandaneto, line 247: @barkbarkimashark I noticed you removed this only from `discover.yaml`, should I remove it from here as well?

snuba/datasets/processors/errors_processor.py
```diff
@@ -332,7 +332,7 @@ def extract_stacktraces(
         frame_stack_levels = []
         exception_main_thread = None
 
-        check_exception_main_thread = get_config("check_exception_main_thread", 0)
+        check_exception_main_thread = get_config("check_exception_main_thread", 1)
```
marandaneto, line 335: @MeredithAnya I rolled back the default so I can do an integration test [here](https://github.com/getsentry/sentry/pull/46183).
As of now, there's no way to enable the config for the test pipeline, so this is needed.

untitaker: since this is rolled out you can just remove the config check altogether

marandaneto: done

untitaker (approved):

snuba/datasets/configuration/discover/entities/discover_transactions.yaml
marandaneto: I guess this should stay, `exception_main_thread` should always be nullable for transactions

tests/datasets/test_errors_processor.py
```diff
@@ -659,18 +656,21 @@ def test_exception_main_thread_false(self) -> None:
                     },
                     {
                         "id": 2,
-                        "main": False,
+                        "main": True,
                     },
                     {
                         "id": 3,
-                        "main": True,
+                        "main": False,
```
marandaneto, line 663: Test regression introduced by https://github.com/getsentry/snuba/commit/6215a99e89360c899157c119a445a728971e3651

PR #3960, opened 2023-03-31
feat(DRS): Add readiness_states support for MigrationGroup

enochtangg:
### Overview
In order to support the newly defined dataset lifecycle, the `readiness_state` flag needs to be set on the MigrationGroup. This PR is responsible for the following:
* Updating all migration groups with the appropriate readiness_states
* Splitting up `snuba/migrations/group_loader.py` and `snuba/migrations/groups.py`. 

For `snuba/migrations/group_loader.py`:
* No changes. This new file was created to house all `GroupLoader` code.

For `snuba/migrations/groups.py`:
* Contains the original `MigrationGroup: Enum` which was unchanged.
* Created new class called `_MigrationGroup` which contains loader and readiness_state attributes
* The `REGISTED_MIGRATION_GROUPS: dict` was changed to map `MigrationGroup: Enum` to `_MigrationGroup` instead.

### Blast Radius
This change will not affect the current state of snuba because snuba currently does not use these readiness_states yet.


snuba/migrations/group_loader.py
```diff
@@ -0,0 +1,308 @@
...
+        raise NotImplementedError
+
+
+class DirectoryLoader(GroupLoader, ABC):
+    """
+    Loads migrations that are defined as files of a directory. The file name
+    represents the migration ID.
+    """
+
+    def __init__(self, module: str) -> None:
```
volokluev, line 35: is this a `module_name` or `module_path`?

snuba/migrations/group_loader.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def get_migrations(self) -> Sequence[str]:
+        return ["0001_migrations"]
+
+
+class EventsLoader(DirectoryLoader):
+    def __init__(self) -> None:
+        super().__init__("snuba.snuba_migrations.events")
+
+    def get_migrations(self) -> Sequence[str]:
```
volokluev, line 62: suggestion: what if we just used the `import_modules_in_subdirectory` function that we use for loading registered classes in order to do this and avoid the extra step of having to populate these lists?

snuba/migrations/groups.py
```diff
@@ -41,323 +52,39 @@ class MigrationGroup(Enum):
...
-    MigrationGroup.GENERIC_METRICS: GenericMetricsLoader(),
-    MigrationGroup.TEST_MIGRATION: TestMigrationLoader(),
-    MigrationGroup.SEARCH_ISSUES: SearchIssuesLoader(),
+class _MigrationGroup:
+    def __init__(self, loader: GroupLoader, readiness_state: ReadinessState) -> None:
+        self.loader = loader
+        self.readiness_state = readiness_state
+
+
+_REGISTERED_MIGRATION_GROUPS = {
```
volokluev, line 61: yeah this seems ripe for RegisteredClass, but I see now that you're just moving these files around, not changing them?

MIGRATIONS.md
```diff
@@ -98,7 +98,7 @@ The `snuba migrations` CLI tool should be used to manage migrations.
...
 
-If you need to create a new group, add the group to `migrations.groups.MigrationGroup` and a loader for the group defining the path to the directory where that group's migrations will be located. Register these to `migrations.groups._REGISTERED_GROUPS` - note the position of the group in this list determines the order the migrations will be executed in.
+If you need to create a new group, add the group to `migrations.groups.MigrationGroup` and a loader to `migrations.group_loader` for the group defining the path to the directory where that group's migrations will be located. Register these to `migrations.groups._REGISTERED_MIGRATION_GROUPS` - note the position of the group in this list determines the order the migrations will be executed in.
```
MeredithAnya, line 101: I know that right now the `readiness_state` is complete but should we say that when creating a new migration group it should have the limited `readiness_state`?

MeredithAnya (approved):
Looks good to me - just a small comment for the docs

volokluev (approved):
My comments are pretty unrelated to the change. Question for you: what can be the first change that actually does have impact on the system? How can we start actually using readiness states and what does that interface look like?

MIGRATIONS.md
enochtangg: I documented that in [snuba docs](https://getsentry.github.io/snuba/configuration/writable_storage.html), but I'll add it here in README as well.

snuba/migrations/group_loader.py
enochtangg: Updated to `module_path`

snuba/migrations/group_loader.py
enochtangg: Seems like that would make sense. Looking at how `get_migrations() -> Sequence[str]` works, I think the refactor would actually be a big change. We don't want to completely remove this string list because the index in this list is used to track reverse migrations.

snuba/migrations/groups.py
enochtangg: Yes, one of the reason for decoupling the loaders and MigrationGroups files is to be able to build guardrails to check that migration changes (include `readiness_state` changes) are not deployed alongside code changes.

evanh (approved):
Looks good to me.

PR #3962, opened 2023-04-01
fix: Subscription result data

lynnagara:
`trace_output` and `profile` are internal Snuba things. Sentry does nothing with them and they are not required by the schema. Let's not produce this data unnecessarily.

### DIFF
--- a/tests/subscriptions/test_codecs.py
+++ b/tests/subscriptions/test_codecs.py
@@ -164,7 +164,7 @@ def test_subscription_task_result_encoder() -> None:
         task_result.task.task.subscription.identifier
     )
     assert payload["request"] == request.original_body
-    assert payload["result"] == result
+    assert payload["result"]["data"] == result["data"]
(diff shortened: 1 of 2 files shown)


lynnagara:
Related: https://github.com/getsentry/sentry-kafka-schemas/pull/88

tests/subscriptions/test_codecs.py
```diff
@@ -164,7 +164,7 @@ def test_subscription_task_result_encoder() -> None:
         task_result.task.task.subscription.identifier
     )
     assert payload["request"] == request.original_body
-    assert payload["result"] == result
+    assert payload["result"]["data"] == result["data"]
```
evanh, line 167: Why not remove the `profile` and `trace_output` from the expected value?

evanh (approved):

tests/subscriptions/test_codecs.py
lynnagara: Probably could've but I don't think it makes too much difference since only this one actually gets used

PR #3966, opened 2023-04-03
feat(profiling): Suspect functions v2

Zylphrex:
This change adds some new columns to the raw functions table to align with the same columns indexed by metrics and rematerializes the views with these new columns. We WILL be deleting existing data when this switchover happens.

### DIFF
--- a/snuba/snuba_migrations/functions/0002_functions_v2.py
+++ b/snuba/snuba_migrations/functions/0002_functions_v2.py
@@ -1,0 +1,266 @@
+from dataclasses import dataclass
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import (
+    UUID,
+    AggregateFunction,
+    Column,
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/functions/0002_functions_v2.py
```diff
@@ -0,0 +1,266 @@
...
+    Column("examples", AggregateFunction("groupUniqArray(5)", [UUID()])),
+]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    storage_set = StorageSetKey.FUNCTIONS
+
+    materialization_version = 1
```
Zylphrex, line 103: bumping the materialization version as we'll be changing the data that is sent to snuba as well

phacops (approved):

PR #3971, opened 2023-04-04
fix(migrations): Enable profiles, functions and search-issues migrations in dev

enochtangg:
### Overview
This PR is responsible for enabling migrations for profiles, functions, and search-issues to run in local ClickHouse. We already run these migrations in production.


evanh (commented):
I just realized why this might be this way: self hosted users. However I'm not sure it will break anything to have these run.

dbanda (approved):

PR #3974, opened 2023-04-04
fix(migrations): add filtering by readiness on run all

dbanda:
Adds filtering by readiness state to the migrations cli


snuba/cli/migrations.py
```diff
@@ -52,6 +55,12 @@ def list() -> None:
 
 @migrations.command()
 @click.option("-g", "--group", default=None)
+@click.option(
+    "-r",
+    "--readiness_state",
```
MeredithAnya, line 60: nit: i would make this `--readiness-state` since we use `-` over `_` in other options

MeredithAnya (approved):
just a smol change and then it LGTM

PR #3975, opened 2023-04-04
Update MIGRATIONS.md

dbanda:
Add description of the correct query node value

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -48,6 +48,8 @@ For each cluster, `single_node` can be set to either True or False. If False, `c
 
 For a single node cluster, do not change the default configuration.
 
+For multi node clusters (`single_node=False`), the `host` and `port` of the cluster mappings must point to a node that will have the distributed tables. Typically, this is a node that belongs to the `distributed_cluster_name` cluster in clickhouse. This node is used as the query node. In our production environment, the `host` value is a loadbalancer proxy over the nodes is the the `distributed_cluster_name`
+
 ## Running migrations
 
 The `snuba migrations` CLI tool should be used to manage migrations.


MIGRATIONS.md
```diff
@@ -48,6 +48,8 @@ For each cluster, `single_node` can be set to either True or False. If False, `c
 
 For a single node cluster, do not change the default configuration.
 
+For multi node clusters (`single_node=False`), the `host` and `port` of the cluster mappings must point to a node that will have the distributed tables. Typically, this is a node that belongs to the `distributed_cluster_name` cluster in clickhouse. This node is used as the query node. In our production environment, the `host` value is a loadbalancer proxy over the nodes is the the `distributed_cluster_name`
```
evanh, line 51: > the `host` value is a loadbalancer proxy over the nodes is the the `distributed_cluster_name`

I'm not sure what this means.

lynnagara: `Typically, this is a node that belongs to the `distributed_cluster_name` cluster in clickhouse.`

I question whether this is actually typical. At least it seems not at Sentry. Afaik we often have no distributed cluster at all (local, test, small self hosted envs) or multiple distributed tables with a loadbalancer in front, but not only one node in the distributed cluster like you described

dbanda: For cases where there's no distributed cluster, single_node = True and everything is in one host. What I was trying to communicate is that the query node has to have access to the dist tables in the non single node case because we don't make that clear anywhere.

MeredithAnya: Does this make anything clearer?

> For multi node clusters (`single_node=False`), the `host` and `port` of the cluster mappings must point to a node that will have the distributed tables. If distributed tables are separated onto their own node (query node) and you have more than one of these query nodes, you should have a ClickHouse cluster defined for those nodes. That cluster name is used for the `distributed_cluster_name`. The `host` value is this scenario could either be the query node directly or a load balancer proxy in front of the query nodes.

onkar:
Going to close this stale (opened 1 year ago) PR.

PR #3978, opened 2023-04-05
feat(DRS): Validate readiness states

enochtangg:
### Overview
With the introduction of readiness states defined on storages and migration groups, the system needs semantic validation to ensure that the configurations make sense. This PR is responsible for adding a check to ensure that the `readiness_state` of a storage is not greater than its migration group counterpart. This validation takes place when each storage is built in factory.

To accomplish this validation, there were a few other changes required:
* Adding a `level` attribute to the `ReadinessState` class
* Defining `storage_set_keys` as an attribute of `_MigrationGroup`
* Enforce that each storage_set can only be associated in one migration group. In other words, two migration groups cannot run an operation on the same storage_set. This pattern is already true today, but is not a written rule. The implication of this pattern ensures that the ordering of migrations groups does not matter (with the exception of `discover`). More details found in [Dataset Readiness States](https://www.notion.so/sentry/Dataset-Readiness-States-da12e741fe024a7c8916c7b36b94ed8f?pvs=4#f5f4baf48fca4b6ab3636bddb7299554).
* Create tests

### Blast Radius
This change impacts all storages in snuba as the validator will run for each storage in factory initialization.


volokluev:
Is it necessary to have the validation be built as part of the storage factory? We could put this as a check in CI but the idea of breaking all snuba pods because someone got a migration readiness state wrong seems too heavy handed

snuba/datasets/storages/validator.py
```diff
@@ -0,0 +1,37 @@
...
+        For example, there is no case where a storage is `complete` and the associated migration group is in `limited`.
+        A storage should not be able to query a non-existing table in ClickHouse.
+        This check ensures that the migrations are already deployed before the storage is even loaded.
+        """
+        storage_set_key = self.storage.get_storage_set_key()
+        storage_readiness_state = self.storage.get_readiness_state()
+        group_readiness_state = get_group_readiness_state_from_storage_set(
+            storage_set_key
+        )
+        if storage_readiness_state.level > group_readiness_state.level:
```
evanh, line 30: Was the Enum `__new__` override to support this check? You could also add a `__gt__` operator override I think, which would then mean you don't need the level parameter.

snuba/datasets/readiness_state.py
```diff
@@ -1,8 +1,21 @@
...
 from enum import Enum
+from typing import Any
 
 
 class ReadinessState(Enum):
-    LIMITED = "limited"
-    DEPRECATE = "deprecate"
-    PARTIAL = "partial"
-    COMPLETE = "complete"
+    LIMITED = "limited", 1
```
dbanda, line 8: Can we get away with not having level here? Sometimes we want to do enum comparison with strings e.g `x.state == `limited` . This will make it harder since you have to remember the level. If level is already bound to the enum string I see no point in keeping the enum as a tuple.

snuba/datasets/storages/validator.py
enochtangg: I like the `__gt__` override better. Updated.

snuba/datasets/readiness_state.py
enochtangg: Makes sense. Addressed in comment above.

enochtangg:
> Is it necessary to have the validation be built as part of the storage factory? We could put this as a check in CI but the idea of breaking all snuba pods because someone got a migration readiness state wrong seems too heavy handed


@volokluev Good point. My initial thought was to put this in the factory because that's where we do syntax validation (via `load_configuration_data()` too. So, the process I had in mind would be syntax validation -> load configs in python -> semantic validation. Also, if we get a migration readiness state wrong, it should fail in CI (tests) when initializing the factory before it breaks all snuba pods. What do you think about this process?

volokluev:
yeah that makes sense to me

evanh (approved):

PR #3979, opened 2023-04-05
feat: Throttle commits, add logging to help debug

lynnagara:


onewland (approved):

snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -454,7 +454,7 @@ def submit(self, message: Message[CommittableTick]) -> None:
         # to commit is provided.
         if len(encoded_tasks) == 0 and message.payload.offset_to_commit is not None:
             offset = {message.value.partition: message.value.payload.offset_to_commit}
-            logger.info("Committing offset: %r", offset)
+            logger.info("Committing offset - no subscriptions: %r", offset)
             self.__commit(offset)
```
untitaker, line 458: can you add `force=True` in join() of this strategy?

untitaker (approved):

snuba/subscriptions/scheduler_processing_strategy.py
lynnagara: Good point, will update

untitaker: there's generally no guarantee of exactly-once so I am not sure if it is worth restarting CI over rn though

lynnagara: race condition with you posting comment... already pushed

PR #3982, opened 2023-04-06
feat(DRS): Use readiness_states in health-checks and active migration groups

enochtangg:
### Overview
This PR is responsible for updating `get_active_migration_groups()` and `check_clickhouse()` functions to use readiness_states to determine:
1. Whether or not a health-check should be executed 
2. Whether or not a migration group is active

The roll-out process is controlled by two new settings variables called `READINESS_STATE_MIGRATION_GROUPS_ENABLED: set[str]` and `READINESS_STATE_STORAGES_ENABLED: set[str]`.
If the storage or migration group exists in these sets, then use the readiness_state to determine whether or not the health-check or migration group is active. If not, then use the existing system (is_experimental and SKIPPED_MIGRATION_GROUPS). The goal is to incrementally roll this out to all storages and migration groups. Once this is completely rolled out and there is parity across all environments, we can deprecate the old settings. More details in [Dataset Readiness State Rollout Plan](https://www.notion.so/sentry/Dataset-Readiness-State-Rollout-Plan-d38deffe77534856a4e4f4a3b0cfb110?pvs=4#7c8bd102fbd2467ca40df7029a29e448).

### Blast Radius
This change impacts all storage health-checks and how migrations_groups are loaded. Supposedly, there should be no change until we add to `READINESS_STATE_MIGRATION_GROUPS_ENABLED` and `READINESS_STATE_STORAGES_ENABLED`

### DIFF
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -21,7 +21,12 @@ from snuba.migrations.errors import (
     MigrationError,
     MigrationInProgress,
 )
-from snuba.migrations.groups import OPTIONAL_GROUPS, MigrationGroup, get_group_loader
+from snuba.migrations.groups import (
+    OPTIONAL_GROUPS,
+    MigrationGroup,
+    get_group_loader,
+    get_readiness_state,
+)
 from snuba.migrations.migration import ClickhouseNodeMigration, CodeMigration, Migration
 from snuba.migrations.operations import OperationTarget, SqlOperation
 from snuba.migrations.status import Status
@@ -33,15 +38,19 @@ DIST_TABLE_NAME = "migrations_dist"
 
 
 def get_active_migration_groups() -> Sequence[MigrationGroup]:
-
-    return [
-        group
-        for group in MigrationGroup
-        if not (
-            group in OPTIONAL_GROUPS
-            and group.value in settings.SKIPPED_MIGRATION_GROUPS
-        )
-    ]
+    groups = []
+    for group in MigrationGroup:
+        if group.value in settings.READINESS_STATE_MIGRATION_GROUPS_ENABLED:
+            readiness_state = get_readiness_state(group)
+            if readiness_state.value in settings.SUPPORTED_STATES:
+                groups.append(group)
+        else:
+            if not (
+                group in OPTIONAL_GROUPS
+                and group.value in settings.SKIPPED_MIGRATION_GROUPS
+            ):
+                groups.append(group)
+    return groups
 
 
 class MigrationKey(NamedTuple):
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -261,6 +261,8 @@ if os.environ.get("ENABLE_AUTORUN_MIGRATION_SEARCH_ISSUES", False):
 
 # Dataset readiness states supported in this environment
 SUPPORTED_STATES: Set[str] = {"deprecate", "limited", "partial", "complete"}
+READINESS_STATE_MIGRATION_GROUPS_ENABLED: set[str] = set()
+READINESS_STATE_STORAGES_ENABLED: set[str] = set()
 
 MAX_RESOLUTION_FOR_JITTER = 60
 
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -56,6 +56,7 @@ from snuba.datasets.factory import (
     get_enabled_dataset_names,
 )
 from snuba.datasets.schemas.tables import TableSchema
+from snuba.datasets.storage import ReadableTableStorage
 from snuba.query.allocation_policies import AllocationPolicyViolation
 from snuba.query.exceptions import InvalidQueryException
 from snuba.query.query_settings import HTTPQuerySettings
@@ -120,6 +121,7 @@ def check_clickhouse(
 ) -> bool:
     """
     Checks if all the tables in all the enabled datasets exist in ClickHouse
+    TODO: Eventually, when we fully migrate to readiness_states, we can remove is_experimental and DISABLED_DATASETS.
     """
     try:
         if ignore_experimental:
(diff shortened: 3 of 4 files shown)


evanh (approved):

snuba/migrations/runner.py
```diff
@@ -33,15 +38,19 @@
...
+        if group.value in settings.READINESS_STATE_MIGRATION_GROUPS_ENABLED:
+            readiness_state = get_readiness_state(group)
+            if readiness_state.value in settings.SUPPORTED_STATES:
+                groups.append(group)
+        else:
+            if not (
+                group in OPTIONAL_GROUPS
+                and group.value in settings.SKIPPED_MIGRATION_GROUPS
+            ):
+                groups.append(group)
```
volokluev, line 52: if a migration is skipped it's active? I'm not sure I understand that logic

snuba/settings/__init__.py
```diff
@@ -261,6 +261,8 @@ class RedisClusters(TypedDict):
 
 # Dataset readiness states supported in this environment
 SUPPORTED_STATES: Set[str] = {"deprecate", "limited", "partial", "complete"}
+READINESS_STATE_MIGRATION_GROUPS_ENABLED: set[str] = set()
```
volokluev, line 264: When putting in "temporary" config values I like to leave a comment saying at which condition/date I would expect to remove them in case some big re-prioritization in the company comes through

volokluev (commented):
given the blast radius of this, how did you test it? can it be tested with unit tests?

snuba/web/views.py
```diff
@@ -134,9 +136,18 @@ def check_clickhouse(
         entities = itertools.chain(
             *[dataset.get_all_entities() for dataset in datasets]
         )
-        storages = list(
-            itertools.chain(*[entity.get_all_storages() for entity in entities])
-        )
+
```
nikhars, line 139: I would recommend moving the logic of filtering which storage/cluster needs to be checked(including the experimental and disabled_datasets) into a method of its own. The `check_connection` method is doing too much right now and has too much context baked into it. Moving it out would help code readability and help write tests for that logic.

snuba/web/views.py
```diff
@@ -134,9 +136,18 @@ def check_clickhouse(
...
+        storages = []
+        for entity in entities:
+            entity_storages = entity.get_all_storages()
+            for storage in entity_storages:
+                assert isinstance(storage, ReadableTableStorage)
```
nikhars, line 144: Why does the storage need to be `ReadableTableStorage`? Mind providing a bit more context?

snuba/settings/__init__.py
enochtangg: Makes sense. Added!

snuba/web/views.py
enochtangg: Agreed. Moved into `filter_checked_storages` function.

snuba/web/views.py
enochtangg: The type checking is used to access the `get_storage_key()` method which is not implemented on the base Storage class.

enochtangg:
Added unit test for `get_active_migration_groups()`.

snuba/migrations/runner.py
volokluev: okay I think my confusion here is that it's not clear what "active" means or there's a logic bug

> If it isn't then default to using the existing SKIPPED_MIGRATIONS_GROUPS to determine if it's active.

But shouldn't it be

```python3
            if not (
                group in OPTIONAL_GROUPS
                and group.value **not** in settings.SKIPPED_MIGRATION_GROUPS
            ):
                groups.append(group)
```

otherwise you're going to mark a migration that is skipped in this environment as active

snuba/web/views.py
```diff
@@ -115,29 +117,41 @@ def check_down_file_exists() -> bool:
             return False
 
 
+def filter_checked_storages(ignore_experimental: bool) -> List[Storage]:
```
volokluev, line 120: test for this?

snuba/migrations/runner.py
volokluev: I misread the code and just didn't see the `not` in front of the condition 😅

tests/migrations/test_runner.py
```diff
@@ -358,6 +358,23 @@ def get_total_migration_count() -> int:
...
+        MigrationGroup.SEARCH_ISSUES not in active_groups
+    )  # should be skipped by SKIPPED_MIGRATION_GROUPS
+
+    settings.READINESS_STATE_MIGRATION_GROUPS_ENABLED = {"search_issues"}
+    active_groups = get_active_migration_groups()
+    assert (
+        MigrationGroup.SEARCH_ISSUES in active_groups
+    )  # should be active by readiness_state
+
+    importlib.reload(settings)
```
volokluev, line 375: It's good practice to use a fixture for stuff like this in case your test throws (since you are affecting global state) for example:

```python3
@pytest.fixture(scope="function")
def temp_settings():
    from snuba import settings
    yield settings
    importlib.reload(settings)
    
def test_get_active_migration_groups(temp_settings) -> None:
    pass    
```

volokluev (commented):
add a test for `filter_checked_storages` and we're good

tests/migrations/test_runner.py
enochtangg: Updated 👍

tests/web/test_check_clickhouse.py
```diff
@@ -88,3 +88,16 @@ def test_dataset_undefined_storage_set(
...
         assert isinstance(v, str)
+
+
+@mock.patch(
+    "snuba.web.views.get_enabled_dataset_names",
+    return_value=["events", "experimental"],
+)
+@mock.patch("snuba.web.views.get_dataset", side_effect=fake_get_dataset)
+@pytest.mark.clickhouse_db
+def test_filter_checked_storages(mock1: mock.MagicMock, mock2: mock.MagicMock) -> None:
```
volokluev, line 99: does this test actually test your changes? I don't see anything about readiness states in here

enochtangg: It didn't initially because I thought the test could be simplified when we eventually remove `is_experimental`. But, it's better to have a test for it even when this code isn't used yet. Added!

PR #3983, opened 2023-04-06
fix(admin) Fix regex to allow SelectExecutor tracing lines

evanh:
These lines changed in the tracing tool and were being excluded by a regex.
Change the regex to be more permissive.


nikhars:
> These lines changed in the tracing tool and were being excluded by a regex.
> Change the regex to be more permissive.

Not too familiar with frontend code so will let others approve. But curious to know what the regex was before and after your change.

evanh:
> But curious to know what the regex was before and after your change.

So the incoming lines had the form 

`[ snuba-errors-tiger-mz-2-2 ] [ 347199 ] {2dfa098f-01e8-45a0-85ea-4980f682e723} <Debug> default.errors_local (4cfa084e-1cad-4cb2-8cfa-084e1cadfcb2) (SelectExecutor): Key condition: (column 1 in [1680307200, +inf)), (column 1 in (-inf, 1680393600]), and, (column 0 in [1, 1]), and, (column 1 in [1680307200, +inf)), (column 1 in (-inf, 1680393600]), and, and`

There is parsing that pulled out a section that looks like `default.errors_local (4cfa084e-1cad-4cb2-8cfa-084e1cadfcb2) (SelectExecutor):`

The regex would then check for `/^(default.(\S+) \(SelectExecutor\)|MergeTreeSelectProcessor)/` and show rows that matched. However at some point the UUID was added between the table name and the header, which throws off the regex. So I simplified it a bit.

PR #3987, opened 2023-04-07
Update to new k8s deploy script

mattgauntseo-sentry:
Committed via https://github.com/asottile/all-repos


mattgauntseo-sentry:
@lyn this change will add retries to deployments which should alleviate the gateway timeout issues snuba has seen.

Josh's fix for conflicts were applied to both old and new scripts so you should see that affect immediately.

dbanda (approved):

PR #3993, opened 2023-04-11
migration(search-issues): add missing columns 

barkbarkimashark:
These columns were overlooked when creating this table. We need to add them to support various search queries.

### DIFF
--- a/snuba/datasets/configuration/issues/storages/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/storages/search_issues.yaml
@@ -21,9 +21,6 @@ schema:
       { name: occurrence_type_id, type: UInt, args: { size: 16 } },
       { name: detection_timestamp, type: DateTime },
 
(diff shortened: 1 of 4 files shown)


MeredithAnya:
@dbanda do we want to allow the yaml changes to be with the migrations? if we do, then the check you added might need to be changed

barkbarkimashark:
> @dbanda do we want to allow the yaml changes to be with the migrations? if we do, then the check you added might need to be changed



> @dbanda do we want to allow the yaml changes to be with the migrations? if we do, then the check you added might need to be changed

Actually, I think this is my bad. We should do the migration first, then add it to the yaml.. I think

barkbarkimashark:
Hmm, now I'm not so sure:
```
All changes: 
snuba/migrations/group_loader.py
snuba/snuba_migrations/search_issues/0006_add_subtitle_culprit_level.py
```

MeredithAnya (approved):

PR #3994, opened 2023-04-11
fix(search-issues): make new columns assessible in storage

barkbarkimashark:
Will only merge this once https://github.com/getsentry/snuba/pull/3993 is merged.

PR #3995, opened 2023-04-11
fix(search-issues): make resource_id, subtitle, culprit, and level fields available in entity

barkbarkimashark:
Expose the fields to the entity. Will merge after https://github.com/getsentry/snuba/pull/3994


enochtangg (commented):
LGTM! Just one small thing, could you please add a test to `test_search_issues_api.py` which inserts and queries the new column.

enochtangg (approved):

PR #3999, opened 2023-04-12
feat(CapMan): AllocationPolicy for Errors

volokluev:
This PR introduces the first iteration of the allocation policy for the errors storage

### The Policy

For every organization, keep track of how many bytes they've scanned in the last 10 minutes, when they reach a limit, throttle them to one thread

### How to turn it on/off

In runtime config:

```
# completely bypasses the code path
SET ErrorsAllocationPolicy.is_active 0
```

```
# code runs and emits metrics but no action is taken
SET ErrorsAllocationPolicy.is_enforced 0
```

```
# how many bytes the org can scan in ten minutes before being throttled
SET ErrorsAllocationPolicy.org_scan_limit 42069
```

### Intent

This is not meant to be the final policy, this is just a quick way to get something working end to end and then iterate on it. Most of this code will be rewritten

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
+import os
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass
 from typing import Any, cast
 
 from snuba.clusters.storage_sets import StorageSetKey
 from snuba.state import get_config
-from snuba.utils.registered_class import RegisteredClass
+from snuba.utils.registered_class import RegisteredClass, import_submodules_in_directory
 from snuba.utils.serializable_exception import SerializableException
 from snuba.web import QueryException, QueryResult
 
@@ -204,4 +205,8 @@ class PassthroughPolicy(AllocationPolicy):
 
 DEFAULT_PASSTHROUGH_POLICY = PassthroughPolicy(
     StorageSetKey("default.no_storage_set_key"), required_tenant_types=[]
+)
+
+import_submodules_in_directory(
+    os.path.dirname(os.path.realpath(__file__)), "snuba.query.allocation_policies"
 )
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -10,7 +10,7 @@ from snuba import settings
 from snuba.clusters.storage_sets import StorageSetKey
 from snuba.state import get_config
 from snuba.utils.registered_class import RegisteredClass, import_submodules_in_directory
-from snuba.utils.serializable_exception import SerializableException
+from snuba.utils.serializable_exception import JsonSerializable, SerializableException
 from snuba.web import QueryException, QueryResult
 
 logger = logging.getLogger("snuba.query.allocation_policy_base")
@@ -55,11 +55,13 @@ class AllocationPolicyViolation(SerializableException):
         )
 
     @property
-    def quota_allowance(self):
-        return self.extra_data.get("quota_allowance", {})
+    def quota_allowance(self) -> dict[str, JsonSerializable]:
+        return cast(
+            "dict[str, JsonSerializable]", self.extra_data.get("quota_allowance", {})
+        )
 
     @property
-    def explanation(self) -> dict[str, Any]:
+    def explanation(self) -> dict[str, JsonSerializable]:
         return self.extra_data.get("quota_allowance", {}).get("explanation", {})  # type: ignore
 
     def __str__(self) -> str:
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -55,6 +55,10 @@ class AllocationPolicyViolation(SerializableException):
         )
 
     @property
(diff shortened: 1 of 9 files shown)


snuba/query/allocation_policies/errors_allocation_policy.py
```diff
@@ -0,0 +1,186 @@
...
+                10000,
+            )
+
+            timestamp, granted_quotas = self._rate_limiter.check_within_quotas(
+                [
+                    RequestedQuota(
+                        self.rate_limit_prefix,
+                        # request a big number because we don't know how much we actually
+                        # will use in this query. this doesn't use up any quota, we just want to know how much is left
+                        int(1e10),
```
dbanda, line 110: What are the units for this value? Is it bytes scanned?

snuba/query/allocation_policies/errors_allocation_policy.py
```diff
@@ -0,0 +1,186 @@
...
+                            granularity_seconds=self.WINDOW_GRANULARITY_SECONDS,
+                            limit=int(org_scan_limit),  # type: ignore
+                            prefix_override=f"{self.rate_limit_prefix}-organization_id-{tenant_ids['organization_id']}",
+                        )
+                    ],
+                )
+            ],
+            grants=[
+                GrantedQuota(
+                    "{self.rate_limit_prefix}-organization_id-{tenant_ids['organization_id']}",
```
rahul-kumar-saini, line 180:
```suggestion
                    f"{self.rate_limit_prefix}-organization_id-{tenant_ids['organization_id']}",
```

rahul-kumar-saini (commented):
Just some nits, also how easy/clean would it be to pass quota data down to querylog in one of the existing json blobs like clickhouse stats?

snuba/query/allocation_policies/__init__.py
```diff
@@ -162,7 +167,15 @@ def from_kwargs(cls, **kwargs: str) -> "AllocationPolicy":
         )
 
     def get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowance:
-        allowance = self._get_quota_allowance(tenant_ids)
+        try:
+            allowance = self._get_quota_allowance(tenant_ids)
+        except Exception:
+            logger.exception(
+                "Allocation policy failed to get quota allowance, this is a bug, fix it"
+            )
```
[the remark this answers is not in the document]
volokluev: logger.exception should propagate to sentry and logs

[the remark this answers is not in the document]
rahul-kumar-saini: I meant the message from the original exception that was caught, this looks like it'd just say "Allocation policy failed" without any context no?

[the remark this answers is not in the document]
volokluev: I understand your question now. `logging.exception` captures the exception information as well, example:

```
>>> import logging
>>> try:
...     raise Exception("This is an error")
... except Exception:
...     logging.exception("We messed up")
...
ERROR:root:We messed up
  File "<stdin>", line 2, in <module>
Exception: This is an error
```

snuba/query/allocation_policies/errors_allocation_policy.py
volokluev: good question. Yes but I will clarify that

volokluev:
@rahul-kumar-saini Addressed all comments, also policy decisions are now captured in `stats` in the querylog

snuba/query/allocation_policies/__init__.py
```diff
@@ -162,7 +173,15 @@ def from_kwargs(cls, **kwargs: str) -> "AllocationPolicy":
         )
 
     def get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowance:
-        allowance = self._get_quota_allowance(tenant_ids)
+        try:
+            allowance = self._get_quota_allowance(tenant_ids)
+        except Exception:
+            logger.exception(
+                "Allocation policy failed to get quota allowance, this is a bug, fix it"
```
evanh, line 180: It's probably worth capturing the tenant_ids here too.

snuba/query/allocation_policies/errors_allocation_policy.py
```diff
@@ -0,0 +1,205 @@
...
+    RequestedQuota,
+)
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+metrics = MetricsWrapper(environment.metrics, "errors_allocation_policy")
+
+
+# A hardcoded list of referrers which do not have an organization_id associated with them
+# purposefully not in config because we don't want that to be easily changeable
+_ORG_LESS_REFERRERS = set(
```
evanh, line 29: Why have this list? Why not skip any query that doesn't have an `organization_id` in it?

rahul-kumar-saini (approved):

snuba/query/allocation_policies/errors_allocation_policy.py
volokluev: some of them are valid, e.g. subscriptions. Some of them are maybe valid but I don't have enough knowledge. They don't constitute even a percent of traffic but I think you're right that we should make this list go away eventually

PR #4001, opened 2023-04-12
feat: Use new DLQ from Arroyo

lynnagara:
Depends on https://github.com/getsentry/snuba/pull/4003

These are the Snuba changes needed once https://github.com/getsentry/arroyo/pull/202 lands.

Also fixes a bug where the DLQ topic name is not resolved from the logical to physical names or to the correct slice if the storage is sliced.

I kept the DLQ policy definitions in the storage yaml files unchanged, we may want to revisit at some later stage

### DIFF
--- a/snuba/cli/multistorage_consumer.py
+++ b/snuba/cli/multistorage_consumer.py
@@ -292,7 +292,7 @@ def multistorage_consumer(
         resolved_topic = Topic(dlq_topic_spec.get_physical_topic_name(slice_id))
 
         dlq_policy = DlqPolicy(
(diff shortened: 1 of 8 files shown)


untitaker (approved):
this makes sense to me. should we add something that raises InvalidMessage? I'm thinking an additional bool config option like `dlq_schema_EVENTS` that turns schema warnings into errors. Then we can send in bad messages for any topic.

lynnagara:
> I'm thinking an additional bool config option like dlq_schema_EVENTS that turns schema warnings into errors. Then we can send in bad messages for any topic.

Sgtm. Let's do it as a separate PR after this.

PR #4002, opened 2023-04-12
fix: update validation from org_countries for granularity

andriisoldatenko:
This PR synchronize validation for org_countries for granularity,

Based on:

https://github.com/getsentry/snuba/blob/51062e0cf93e5dd54fd596a276a774af71f8b197/snuba/datasets/configuration/generic_metrics/entities/org_counters.yaml#L59-L64

Some more detailed context where we use it:

In Dynamic sampling we use query for last 5 mins to calibrate dynamic sampling rules, so we need granularity=60, otherwise we'll see a validation error: 

```
E   sentry.utils.snuba.SnubaError: validation failed for entity generic_org_metrics_counters: granularity must be multiple of 3600
```

### DIFF
--- a/snuba/datasets/configuration/generic_metrics/entities/org_counters.yaml
+++ b/snuba/datasets/configuration/generic_metrics/entities/org_counters.yaml
@@ -86,4 +86,4 @@ query_processors:
 validators:
   - validator: GranularityValidator
     args:
-      minimum: 3600
+      minimum: 60


RaduW (approved):

onewland:
This is potentially going to increase the number of rows queried by a factor of 60, right? Or is there a new query path?

andriisoldatenko:
> This is potentially going to increase the number of rows queried by a factor of 60, right? Or is there a new query path?

we are querying data for last 5 mins only in this case with wrong granularity which cause no data.  

here is related sentry PR https://github.com/getsentry/sentry/pull/47220

onewland (approved):
I'm fine with this, but please verify that your existing queries that do wider time windows explicitly ask for the hour granularity

PR #4004, opened 2023-04-12
feat(profiling): Add new columns to raw functions table

Zylphrex:
This adds some new columns to the raw functions table that we want to explore the cardinality of. These columns were chosen based on the same columns the metrics datasets had chosen plus some profiling specific ones. This will be used to create a new materialization from the findings. And any high cardinality columns will be removed.

### DIFF
--- a/snuba/migrations/group_loader.py
+++ b/snuba/migrations/group_loader.py
@@ -270,7 +270,7 @@ class FunctionsLoader(DirectoryLoader):
         super().__init__("snuba.snuba_migrations.functions")
 
     def get_migrations(self) -> Sequence[str]:
-        return ["0001_functions"]
+        return ["0001_functions", "0002_add_new_columns_to_raw_functions"]
 
 
 class GenericMetricsLoader(DirectoryLoader):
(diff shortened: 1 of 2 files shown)


MeredithAnya:
> This adds some new columns to the raw functions table that we want to explore the cardinality of.

Do you have some more context about this? Why these columns in particular? Does this mean that these columns may be dropped in the future?

evanh:
I think some more context is necessary. If this is exploratory, why not use DataDog or logging or leverage the data team to run some analysis for you?

Zylphrex:
> I think some more context is necessary. If this is exploratory, why not use DataDog or logging or leverage the data team to run some analysis for you?

One of the action items from the discussion with various sns team members was to perform a cardinality analysis of these new fields were introducing. I don't mind doing this another way but writing to the raw table was the recommendation I received.

For datadog, I'm not sure we'll be able to do this without potentially incurring high costs.

For logging, is there some easy mechanism to calculate the cardinality that we're looking for here?

We've talked to the data team about surfacing this data in big query but that won't be ready until next quarter at least.

Given all this, is there some easy way to do the cardinality analysis without writing the data to clickhouse?

evanh (approved):

PR #1861, opened 2021-05-07
ref: Redefine Replacement as ABC

untitaker:
This change has been motivated by two comments:

1. https://github.com/getsentry/snuba/pull/1476#discussion_r521151540
2. https://github.com/getsentry/snuba/pull/1795#discussion_r606420996

<details><summary>Old description</summary>
...though the interface may have to change in a bunch of ways to
actually solve the problems that Filippo complains about wrt poor typing
of queries. The important bit is that replacements now encapsulate their
own logic.

The old definition of replacement is implemented in terms of the new
one, and acts as a compatibility layer until all replacements have been
migrated.

Only two replacements have been actually refactored. exclude_groups has
been refactored to be a "modern" replacement, replace_groups has been
refactored just to be a type.

If we are fine with the Replacement interface, we can gradually:

1. refactor all replacements to look like ReplaceGroupReplacement
   (direct subtypes of LegacyReplacement)
2. get rid of LegacyReplacement

</details>

PR #4009, opened 2023-04-13
ref: Make all errors replacements class-based

untitaker:
decided to just rip the bandaid off and do all replacements at once instead of
continuing to make tiny PRs. The diff is massive but this refactor has become
so mechanical to me that I _think_ there'd be not that much risk in rolling it
out at once. Test coverage is OK.


evanh (commented):
What's the context behind this change? Why make everything a class?

untitaker:
It is part of a series of PRs such as:

https://github.com/getsentry/snuba/pull/3996
https://github.com/getsentry/snuba/pull/3957


original context here: https://github.com/getsentry/snuba/pull/1861

evanh (approved):
If you're confident in this, then I'm confident.

PR #4013, opened 2023-04-13
feat(CI): add migrations check to main pipeline

dbanda:
We've tested the migrations check on the test pipeline. This PR transfers it to the main pipeline.

The check compares against the last deployed commit to check that there are not too many undeployed migrations commits stacked up . We are ignoring migrations status for now.

### DIFF
--- a/gocd/pipelines/snuba.yaml
+++ b/gocd/pipelines/snuba.yaml
@@ -17,7 +17,7 @@ pipelines:
         materials:
             snuba_repo:
                 git: git@github.com:getsentry/snuba.git
-                shallow_clone: true
+                shallow_clone: false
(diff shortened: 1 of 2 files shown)


MeredithAnya:
What is the action someone can take if too many migration commits are stacked up? Are we able to deploy specific commits?

dbanda:
> What is the action someone can take if too many migration commits are stacked up? Are we able to deploy specific commits?

Yes. in GoCD you would deploy the commit before your earliest undeployed migration commit, then deploy your migration commit, then deploy the rest. The intent is to give you a pause to actually run the migration.

MeredithAnya (approved):

PR #4016, opened 2023-04-13
feat(CapMan): Errors Allocation Policy V0 (fixed)

volokluev:
Same as #3999 except we make sure the allocation policy is pickleable (and all future allocation policies are pickleable)

### The Policy

For every organization, keep track of how many bytes they've scanned in the last 10 minutes, when they reach a limit, throttle them to one thread

### How to turn it on/off

In runtime config:

```
# completely bypasses the code path
SET ErrorsAllocationPolicy.is_active 0
```

```
# code runs and emits metrics but no action is taken
SET ErrorsAllocationPolicy.is_enforced 0
```

```
# how many bytes the org can scan in ten minutes before being throttled
SET ErrorsAllocationPolicy.org_scan_limit 42069
```

### Intent

This is not meant to be the final policy, this is just a quick way to get something working end to end and then iterate on it. Most of this code will be rewritten

### DIFF
--- a/snuba/datasets/configuration/events/storages/errors.yaml
+++ b/snuba/datasets/configuration/events/storages/errors.yaml
@@ -228,6 +228,13 @@ schema:
     - retention_days
     - date
   not_deleted_mandatory_condition: deleted
+allocation_policy:
+  name: ErrorsAllocationPolicy
+  args:
+    storage_set_key: events_ro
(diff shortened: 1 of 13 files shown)


rahul-kumar-saini (approved):
I'll just trust it's the same as #3999 🙏

PR #4022, opened 2023-04-14
fix(capman): make subscription referrer passthrough

volokluev:
It was assumed that subscriptions did not have an organization id attached to them. Turns out it's organization_id=1 for everything.

Make subscriptions queries not have a quota enforced for them (as is done currently with rate limiting)

### DIFF
--- a/snuba/query/allocation_policies/errors_allocation_policy.py
+++ b/snuba/query/allocation_policies/errors_allocation_policy.py
@@ -28,7 +28,7 @@ metrics = MetricsWrapper(environment.metrics, "errors_allocation_policy")
 # purposefully not in config because we don't want that to be easily changeable
 _ORG_LESS_REFERRERS = set(
     [
-        "subscription_executor",
+        "subscriptions_executor",
         "weekly_reports.outcomes",
         "reports.key_errors",
         "weekly_reports.key_transactions.this_week",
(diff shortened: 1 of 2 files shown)


snuba/query/allocation_policies/errors_allocation_policy.py
```diff
@@ -104,6 +114,8 @@ def _get_quota_allowance(self, tenant_ids: dict[str, str | int]) -> QuotaAllowan
                 return QuotaAllowance(
                     can_run=False, max_threads=0, explanation={"reason": why}
                 )
+        if tenant_ids["referrer"] in _PASS_THROUGH_REFERRERS:
+            return DEFAULT_PASSTHROUGH_POLICY.get_quota_allowance(tenant_ids)
```
rahul-kumar-saini, line 118: Shouldn't this early return check happen before tenant id validation? (Unless want to fail on bad tenant ids even if it's a passthrough referrer)

rahul-kumar-saini (approved):

snuba/query/allocation_policies/errors_allocation_policy.py
volokluev: The outcome is the same but knowing when we have bad tenant ids is good data. Also maybe there is no `referrer` either, don't want a KeyError happening

rahul-kumar-saini:
You could probably also just add this tag to the `db_request_rejected` metric in this PR:

`"referrer": str(tenant_ids.get("referrer", "no_referrer"))`

PR #4025, opened 2023-04-14
feat(CapMan): AllocationPolicy Config API

rahul-kumar-saini:
### Overview
- Related front end PR: https://github.com/getsentry/snuba/pull/4055
- Introduced data structure and interface that will store Policy configurations
- Introduced a few methods to `AllocationPolicy` which will be used by the UI to interact with the policies
- Replaced StorageSetKey with StorageKey across Policies
- Added `org_limit_bytes_scanned_override` optional config to `BytesScannedWindowAllocationPolicy` which can be used to override an individual organizations bytes scanned limit.

Updated tests.


snuba/query/allocation_policies/__init__.py
```diff
@@ -133,12 +145,22 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
 
     def __init__(
         self,
+        name: str,
         storage_set_key: StorageSetKey,
         required_tenant_types: list[str],
         **kwargs: str,
     ) -> None:
+        self._name = name
```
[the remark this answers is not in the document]
rahul-kumar-saini: Let's say we have some policy which we want to use on multiple storages but with different configs, without some way to uniquely identify an instance of the policy there's no way to "namespace" the policy to have different configs in Redis.

eg. `APolicy` would have a config `APolicy.enabled` and if we want to use it on multiple storages, it'll be enabled or disabled for all. This gets worse when we wanna customize more configs like different bytes scanned limits by storage.

With some `name` parameter we can make the keys in Redis unique so like `APolicy` can have an instance named `ErrorsAPolicy` and an instance named `TransactionsAPolicy` with different unique keys (`ErrorsAPolicy.enabled`, `TransactionsAPolicy.enabled`, etc)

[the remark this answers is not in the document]
rahul-kumar-saini: Actually, there's probably a better way to do this since the policy knows which storage it's part of, will update the PR

snuba/query/allocation_policies/errors_allocation_policy.py
```diff
@@ -57,15 +58,25 @@ class ErrorsAllocationPolicy(AllocationPolicy):
...
+                    "Number of threads any throttled query gets assigned.", int, 1
+                ),
+                "org_limit_bytes_scanned": AllocationPolicyConfig(
+                    "Number of bytes any org can scan in a 10 minute window.",
+                    int,
+                    10000,
+                ),
+            }
+        )
+        self._name = self._name or self.__class__.__name__
```
volokluev, line 79: This interface is a little confusing. You're updating state on the class in the init method rather than declaring what the configuration options are. There are two classes of configs:

1. Configs that apply to any policy (is_enforced, is_active)
2. Configs that apply to a specific policy

It shouldn't be the user's responsibility to correctly merge them, only to define them

snuba/query/allocation_policies/__init__.py
```diff
@@ -173,6 +195,41 @@ def from_kwargs(cls, **kwargs: str) -> "AllocationPolicy":
             **kwargs,
         )
 
+    def config_params(self) -> dict[str, AllocationPolicyConfig]:
```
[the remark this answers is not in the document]
rahul-kumar-saini: Updated how the configs are stored to also contain the current value, so this now returns an object containing an up to date current value per config.

snuba/datasets/configuration/storage_builder.py
```diff
@@ -92,7 +93,12 @@ def __build_readable_storage_kwargs(config: dict[str, Any]) -> dict[str, Any]:
         ),
         ALLOCATION_POLICY: AllocationPolicy.get_from_name(
             config[ALLOCATION_POLICY]["name"]
-        ).from_kwargs(**config[ALLOCATION_POLICY].get("args", {}))
+        ).from_kwargs(
+            **{
+                **config[ALLOCATION_POLICY].get("args", {}),
+                "storage_key": storage_key.value,
```
evanh, line 99: Why does this take the `.value` instead of just the storage key?

rahul-kumar-saini: The `from_kwargs()` method expects string kwargs, and the method picks up the string and converts it back to a storage key object. Just seems like the general pattern for Registered classes, @volokluev can probably chime in here?

rahul-kumar-saini:
Made a couple changes to make configs parameterized, will update tests and PR description next

evanh (commented):
Seems fine, just need to fix CI.

evanh (approved):

snuba/query/allocation_policies/__init__.py
```diff
@@ -133,12 +180,30 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
 
     def __init__(
         self,
-        storage_set_key: StorageSetKey,
+        storage_key: StorageKey,
         required_tenant_types: list[str],
         **kwargs: str,
     ) -> None:
         self._required_tenant_types = set(required_tenant_types)
-        self._storage_set_key = storage_set_key
```
volokluev, line 141: since you're removing the `storage_set_key`, please update the class docstring as well

tests/query/allocation_policies/test_bytes_scanned_window_allocation_policy.py
```diff
@@ -161,11 +161,47 @@ def test_reject_queries_without_tenant_ids(policy) -> None:
 
 
 @pytest.mark.redis_db
-def test_passthrough_subscriptions(policy) -> None:
+def test_bad_config_keys(policy: AllocationPolicy) -> None:
```
volokluev, line 164: I don't think this test is necessary, the base provides this already

volokluev (approved):
looks good, with a few questions

snuba/query/allocation_policies/__init__.py
```diff
@@ -155,24 +220,208 @@ def __eq__(self, other: Any) -> bool:
...
+
+    def config_definitions(self) -> dict[str, AllocationPolicyConfig]:
+        """Returns a dictionary of config definitions on this AllocationPolicy."""
+        return {
+            config.name: config
+            for config in self._default_config_definitions
+            + self._additional_config_definitions()
+        }
+
+    def get_parameterized_config_definitions(self) -> list[dict[str, Any]]:
```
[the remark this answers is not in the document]
rahul-kumar-saini: This is similar to my answer for this: https://github.com/getsentry/snuba/pull/4055#discussion_r1187825985

When the user goes to add an instance of an optional config they are greeted with a modal which needs to know what to ask the user. This gives a list of optional configs available and what types they expect for their params etc.

snuba/query/allocation_policies/__init__.py
```diff
@@ -173,9 +180,9 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
     * get_quota_allowance will throw an AllocationPolicyViolation if _get_quota_allowance().can_run is false.
         this is to keep with the pattern in `db_query.py` which communicates error states with exceptions. There is no other
         reason. For more information see snuba.web.db_query.db_query
-    * Every allocation policy takes a `storage_set_key` in its init. The storage_set_key is like a pseudo-tenant. In different
+    * Every allocation policy takes a `storage_key` in its init. The storage_key is like a pseudo-tenant. In different
         environments, storages may be co-located on the same cluster. To facilitate resource sharing, every allocation policy
-        knows which storage_set_key it is serving. This is currently not used
+        knows which storage_key it is serving. This is currently not used
```
volokluev, line 185: Seems like the storage key is actually used now and it's pretty important to every policy's operation

snuba/query/allocation_policies/__init__.py
```diff
@@ -199,6 +206,12 @@ def __init__(
                 value_type=int,
                 default=0,
             ),
+            AllocationPolicyConfig(
```
volokluev, line 209: This is not what I meant. I meant the maximum amount of threads the policy could allocate to a query.

tests/test_api.py
```diff
@@ -56,12 +56,12 @@ def setup_teardown(
         yield
 
         # Reset rate limits
-        state.delete_config("global_concurrent_limit")
-        state.delete_config("global_per_second_limit")
-        state.delete_config("project_concurrent_limit")
-        state.delete_config("project_concurrent_limit_1")
-        state.delete_config("project_per_second_limit")
-        state.delete_config("date_align_seconds")
+        state.delete_config_value("global_concurrent_limit")
```
volokluev, line 59: why did this name need to change?

rahul-kumar-saini: `{set/get/delete}_config()` could be slightly ambiguous in terms of the return value since other "get" functions return more than just value, eg `get_current_configs()`

snuba/query/allocation_policies/__init__.py
rahul-kumar-saini: I created a `max_threads` property on the base class for this now, let me know if it makes sense.

As far as `throttled_thread_number` goes, I can revert it back to being on the bytes scanned class only but would it make sense for all policies to have it as well? I assume each policy would have to have it's own number it would throttle threads to regardless of what kind of policy it is?

tests/test_api.py
volokluev: but this is a `snuba.state` function, why did it have to change?

snuba/query/allocation_policies/bytes_scanned_window_policy.py
```diff
@@ -81,20 +83,39 @@ class BytesScannedWindowAllocationPolicy(AllocationPolicy):
...
+                default=10000,
+            ),
+            AllocationPolicyConfig(
+                name="org_limit_bytes_scanned_override",
+                description="Number of bytes a specific org can scan in a 10 minute window.",
+                value_type=int,
+                default=DEFAULT_OVERRIDE_LIMIT,
+                params={"org_id": int},
+            ),
+        ]
```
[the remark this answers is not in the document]
rahul-kumar-saini: Not sure what you mean, this is the optional config you'd use to override the default number of bytes scanned for an org across the storage.

snuba/query/allocation_policies/__init__.py
volokluev: > I can revert it back to being on the bytes scanned class only but would it make sense for all policies to have it as well?

It does not make sense at this time. Policies will always throttle threads however it may not always be a simple constant value like the existing policy.

snuba/query/allocation_policies/bytes_scanned_window_policy.py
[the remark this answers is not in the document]
volokluev: Yeah I mean you've added a new feature to the policy. Which is fine, just expands the scope of the PR

tests/test_api.py
rahul-kumar-saini: Whooops bad grep

snuba/query/allocation_policies/__init__.py
```diff
@@ -221,16 +221,40 @@ def runtime_config_prefix(self) -> str:
...
+            if settings.RAISE_ON_ALLOCATION_POLICY_FAILURES:
+                raise
+            return False
 
     @property
     def throttled_thread_number(self) -> int:
         """Number of threads any throttled query gets assigned."""
-        return int(self.get_config_value(THROTTLED_THREAD_NUMBER))
+        try:
+            return int(self.get_config_value(THROTTLED_THREAD_NUMBER))
```
volokluev, line 250: you're duplicating this code over and over. Why not `try/accept` `get_config_value`?

rahul-kumar-saini: These all had different return values so try/excepting the get_config_value function would leave us with an unknown value to return. I removed these checks anyhow since they aren't necessary, the main try/excepts in the quota balance methods take care of it.

snuba/query/allocation_policies/__init__.py
```diff
@@ -155,24 +220,208 @@ def __eq__(self, other: Any) -> bool:
...
+
+        required_configs = set(
+            config_name
+            for config_name, config_def in definitions.items()
+            if not config_def.params
+        )
+
+        detailed_configs: list[dict[str, Any]] = []
+
+        for key in runtime_configs:
```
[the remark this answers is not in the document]
rahul-kumar-saini: Alright handled this case, if there's a key match but it doesn't make sense, the key is just skipped and an exception is logged to Sentry.

snuba/query/allocation_policies/bytes_scanned_window_policy.py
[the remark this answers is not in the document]
rahul-kumar-saini: Oh I hadn't realized, I guess I added that in the initial pass of this PR as part of it and just kept it since. I'll update the PR description.

snuba/query/allocation_policies/__init__.py
```diff
@@ -164,12 +166,45 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
...
-    The allocation policy base class has two public methods:
+    The allocation policy base class has two public methods for working with the actual quota:
         * get_quota_allowance
         * update_quota_balance
 
     These functions can be used to modify the behaviour of ALL allocation policies, use with care.
 
+    Configurations
+    --------------
+    The Allocation Policy base class comes with an abstraction on top of the runtime config which is used specifically
```
volokluev, line 177: * Remove the words that are not strictly necessary for the user.
* Since we're breaking from the existing patterns in the codebase, that should be stated
* examples > text

volokluev (approved):

PR #4030, opened 2023-04-17
fix mypy typing test

dbanda:
Recent dependency update broke mypy check. Fixing.


evanh (commented):
Sorry I didn't realize this PR was open already. I merged a fix for the same thing: https://github.com/getsentry/snuba/pull/4032

PR #4031, opened 2023-04-17
[wip] basic logic for fetching partition count using AdminClient

onewland:
Just wanted to share this approach before going further to wiring it in to the necessary classes (which I think is `KafkaTopicSpec` here): 

https://github.com/getsentry/snuba/blob/c6d549a96abfa45836867d688c61e8f6f9d40beb/snuba/datasets/table_storage.py#L61-L63 

which will be a little more complicated

### DIFF
--- a/snuba/cli/subscriptions_scheduler.py
+++ b/snuba/cli/subscriptions_scheduler.py
@@ -6,6 +6,7 @@ from typing import Any, Optional, Sequence
 import click
 from arroyo import configure_metrics
 from arroyo.backends.kafka import KafkaProducer
+from confluent_kafka.admin import AdminClient
 
 from snuba import environment
 from snuba.datasets.entities.entity_key import EntityKey
@@ -13,7 +14,10 @@ from snuba.datasets.entities.factory import get_entity
 from snuba.environment import setup_logging, setup_sentry
 from snuba.subscriptions.scheduler_consumer import SchedulerBuilder
 from snuba.utils.metrics.wrapper import MetricsWrapper
-from snuba.utils.streams.configuration_builder import build_kafka_producer_configuration
+from snuba.utils.streams.configuration_builder import (
+    build_kafka_producer_configuration,
+    get_default_kafka_configuration,
+)
 from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


snuba/subscriptions/scheduler_consumer.py
```diff
@@ -248,7 +249,9 @@ def __init__(
 
         self.__mode = mode
 
-        self.__partitions = stream_loader.get_default_topic_spec().partitions_number
+        self.__partitions = get_partition_count(
+            stream_loader.get_default_topic_spec().topic
```
lynnagara, line 253: Use `get_physical_topic_name(slice_id)`

lynnagara (changes requested):
Please resolve to the physical topic name + correct slice rather than looking up the partition count by the logical default name.

snuba/subscriptions/scheduler_consumer.py
onewland: Follow-up: will deploying this break something if I don't also update 

https://github.com/getsentry/snuba/blob/c12f247997b18662288aeecf93616b0e6d3a7cc9/snuba/subscriptions/partitioner.py#L14-L25

which still uses the old topic map?

lynnagara: Yeah, I think you need to migrate everything off this topic partition map here too if you want to avoid configuring it

snuba/subscriptions/scheduler_consumer.py
```diff
@@ -248,7 +249,9 @@ def __init__(
 
         self.__mode = mode
 
-        self.__partitions = stream_loader.get_default_topic_spec().partitions_number
+        self.__partitions = get_partition_count(
```
fpacifici, line 252: If this is only set at the start of the scheduler, it means that the scheduler will not be able to withstand an increase in the number of partitions, it will receive a tick for an unknown partition and crash. 
I think it would be ok, but Could you  find the places where we validate that the tick received is for a known partition and provide a clear error message that explains that the crash would be normal if due to an increase in partitions ?
The goal is to reduce the noise and the complexity of the partition resize process.

snuba/subscriptions/scheduler_consumer.py
onewland: Are you sure that this PR can't go out alone? I've started looking into replacing this logic and it requires changing a lot more code. If I can do it safely, I'd rather deploy the current change alone (taking into account slicing + Filippo's comment about a clearer error message)

lynnagara: I would prefer for this PR not to go out alone, and now I recall it was intentionally not done in the past.

This PR is essentially saying it is ok for the topic to not be set in `TOPIC_PARTITION_COUNTS` and will no longer error when it is wrong. But it is not really ok and will actually cause more subtle failures unless we replace the logic across the board. Specifically all new subscriptions will be inserted into one partition instead of being properly distributed. We will also break consistency - no more guarantees that all data is written before subscriptions are run if they are being put on different partitions. I'd much rather error in an obvious way even if it is one extra step to fix than risk it being subtly wrong in a way that is hard to spot.

PR #4034, opened 2023-04-18
feat(rust-consumer): Handle shutdown signal

lynnagara:


rust_snuba/rust_arroyo/src/processing/mod.rs
```diff
@@ -71,7 +72,7 @@ pub struct StreamProcessor<'a, TPayload: Clone> {
     consumer: Box<dyn Consumer<'a, TPayload> + 'a>,
     strategies: Arc<Mutex<Strategies<TPayload>>>,
     message: Option<Message<TPayload>>,
-    shutdown_requested: bool,
+    pub shutdown_requested: Arc<AtomicBool>,
```
lynnagara, line 75: I wasn't able to get `processor.signal_shutdown()` to work since StreamProcessor is not thread safe

rust_snuba/rust_arroyo/src/processing/mod.rs
```diff
@@ -201,7 +202,7 @@ impl<'a, TPayload: 'static + Clone> StreamProcessor<'a, TPayload> {
     }
 
     pub fn signal_shutdown(&mut self) {
```
untitaker, line 204:
```suggestion
    pub fn signal_shutdown(&self) {
```

then you should be able to use it

untitaker (approved):

rust_snuba/rust_arroyo/src/processing/mod.rs
untitaker: oh now i see what's goign on. ok. should we remove this method?

lynnagara: Updated to `ProcessorHandle.signal_shutdown()` as discussed offline

PR #4035, opened 2023-04-18
ref: Start using FailoverRedis in Snuba

untitaker:
There is a class called FailoverRedis in Sentry's sentry.utils that is
supposed to deal with networking errors when GCP memorystore falls over.
Snuba doesn't have an equivalent class, and as a result would be
susceptible to the same failures.

sentry-redis-tools is a new package that contains this code + maybe more
redis-related code in the future. For now, it only contains a copy of
Sentry's FailoverRedis.

Add FailoverRedis and make it the default redis client for everything.

Sentry does NOT use FailoverRedis as the default redis client, but I
believe we should change that as well to prevent incidents like INC-311.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,6 +26,7 @@ pytz==2022.2.1
 redis==4.3.4
 sentry-arroyo[json]==2.10.1
 sentry-kafka-schemas==0.0.28
+sentry-redis-tools==0.1.2
 sentry-relay==0.8.19
 sentry-sdk==1.18.0
 simplejson==3.17.6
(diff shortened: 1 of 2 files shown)


untitaker:
GCP memorystore is a single node with a replica and some odd network
routing in front of it. Yes, most single node setups are not memorystore,
but all memorystores are single nodes. We want this behavior for all
memorystores, and I believe it does no harm for the rest.

lynnagara (approved):

beezz (approved):

PR #4036, opened 2023-04-18
fix(migrations): add the actual run/reverse command to gocd

dbanda:
We tested the migrations pipeline with the list command and verified that it captures the correct migrations. This now adds the actual run/reverse command to the pipeline.


evanh (approved):
:shipit:

PR #4038, opened 2023-04-18
Revert "feat(admin): allow querylog to be queried by more threads"

volokluev:
Reverts getsentry/snuba#4037


volokluev:
not necessary

PR #4042, opened 2023-04-18
fix: Fix Rust tests and run in CI

lynnagara:
Tests are currently broken - this fixes them and runs in CI so we don't break them again


dbanda (approved):

rust_snuba/rust_arroyo/Cargo.lock
```diff
@@ -0,0 +1,1817 @@
+# This file is automatically @generated by Cargo.
```
untitaker, line 1: we should not check in this file, only for rust_snuba

rust_arroyo is a library, rust_snuba is an application

untitaker (approved):

PR #4045, opened 2023-04-19
fix(search-issue): map transaction and transaction_name column to tags[transaction]

barkbarkimashark:
Need for issue search.

Is it better to have this as an actual column in the storage?


volokluev (approved):
It is definitely faster to promote it to a column but you can see if your performance is suffering before you do that work

PR #4049, opened 2023-04-19
ref(consumers): process kafka headers

MeredithAnya:
Need a way to be able to pass headers through to `process_message` in https://github.com/getsentry/snuba/blob/master/snuba/consumers/consumer.py#L526-L532 if we want to use https://github.com/getsentry/sentry/pull/47577

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -532,6 +532,7 @@ def process_message(
                 message.value.offset,
                 message.value.partition.index,
                 message.value.timestamp,
+                decoded_headers,
             ),
         )
     except Exception as err:
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -500,6 +500,9 @@ def process_message(
         start = time.time()
 
         decoded = codec.decode(message.payload.value, validate=False)
+        decoded_headers = {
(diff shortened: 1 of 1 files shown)


snuba/consumers/consumer.py
```diff
@@ -529,6 +532,7 @@ def process_message(
                 message.value.offset,
                 message.value.partition.index,
                 message.value.timestamp,
+                decoded_headers,
```
ayirr7, line 535: Is the next step to extract the `decoded_headers` from the `result` (inside of the `MetricsBucketProcessor` and the `GenericMetricsBucketProcessor` and pass it to `BytesInsertBatch`?

MeredithAnya: yeah so like it would be passed in https://github.com/getsentry/snuba/blob/e86fe9a4f12f2641c3b0673ed5ae741c2a8f78a3/snuba/datasets/processors/metrics_bucket_processor.py#L69-L71 

and then we could just do `metadata.headers["sentry_received_timestamp"]` or whatnot in https://github.com/getsentry/snuba/blob/e86fe9a4f12f2641c3b0673ed5ae741c2a8f78a3/snuba/datasets/processors/metrics_bucket_processor.py#L116

PR #4050, opened 2023-04-19
ref(consumers): add optional headers to KafkaMessageMetadata

MeredithAnya:
So we want to be able to add a new header, `sentry_received_timestamp` in https://github.com/getsentry/sentry/pull/47577. But in order to use this header in the `process_message`, it would first need to be passed through in 
https://github.com/getsentry/snuba/blob/4e7e95dfd12c6c257dd0202f2d8b038beaf01bc7/snuba/consumers/consumer.py#L533-L540

I have https://github.com/getsentry/snuba/pull/4049 open that would basically decode the headers, but I thought I'd split up the change to the metadata incase we want to do this in a different way that maybe is more involved than what I have here. 

Ultimately the `sentry_received_timestamp` needs to be used in order to pass an `origin_timestamp` to the insert batch like the following: https://github.com/getsentry/snuba/blob/8427b03e514e4891cc490d33ee0aa1461abae766/snuba/datasets/processors/metrics_aggregate_processor.py#L102

### DIFF
--- a/snuba/consumers/types.py
+++ b/snuba/consumers/types.py
@@ -1,8 +1,9 @@
 from datetime import datetime
-from typing import NamedTuple
+from typing import Mapping, NamedTuple, Optional
 
 
 class KafkaMessageMetadata(NamedTuple):
     offset: int
     partition: int
     timestamp: datetime
+    headers: Optional[Mapping[str, str]] = None


MeredithAnya:
@ayirr7 @lynnagara The other option that seemed reasonable here was to add the headers as a separate arg for `process_message`. It could maybe be an optional arg, curious if yall have preferences for that solution, this one, or something else I haven't considered

ayirr7:
> @ayirr7 @lynnagara The other option that seemed reasonable here was to add the headers as a separate arg for `process_message`. It could maybe be an optional arg, curious if yall have preferences for that solution, this one, or something else I haven't considered

I think having it like this inside the message metadata makes sense. Maybe others will have thoughts?

lynnagara:
Generally speaking, I think of the header, key and value as part of the message itself rather than the metadata. It is passed through to the strategy via the KafkaPayload https://github.com/getsentry/arroyo/blob/69bfaac31fde8586633deee722fa4f4b2cbb3fca/arroyo/backends/kafka/consumer.py#L66-L69.

MeredithAnya:
> Generally speaking, I think of the header, key and value as part of the message itself rather than the metadata. It is passed through to the strategy via the KafkaPayload https://github.com/getsentry/arroyo/blob/69bfaac31fde8586633deee722fa4f4b2cbb3fca/arroyo/backends/kafka/consumer.py#L66-L69.

yeah but the problem is that only the payload.value and not payload.headers is passed through to process_message https://github.com/getsentry/snuba/blob/4e7e95dfd12c6c257dd0202f2d8b038beaf01bc7/snuba/consumers/consumer.py#L502

I think the options we have are: 
1. add the decoded headers to the metadata object
2. add another arg for the decoded headers in the process_message signature
3. change the process_message message arg to have Message be an object with payload and headers (and even key if we wanted), basically like DecodedKafkaPayload or something

I was trying to avoid changing the signature in too many places by going with option 1. , but i think we could go 2. or 3.

lynnagara:
This is making me wonder if the header option is too complex and we should have just gone ahead and put this on the payload itself. It seems a lot easier for Snuba that way. Also because this means we would also need to update the Rust consumer because this part is shared. https://github.com/getsentry/snuba/blob/6bc0c12e6ff09657fbd429d396086ba4b5f10f22/rust_snuba/src/strategies/python.rs#L43-L51

MeredithAnya:
https://github.com/getsentry/sentry/pull/47577 was changed to include the `sentry_received_timestamp` in the payload and therefore there is no need for this PR

PR #4051, opened 2023-04-19
feat(rust): add sentry dsn

dbanda:
Add sentry to the rust proj.

### DIFF
--- a/rust_snuba/src/lib.rs
+++ b/rust_snuba/src/lib.rs
@@ -3,16 +3,37 @@ mod consumer;
 mod strategies;
 mod types;
 
-use pyo3::prelude::*;
+use pyo3::{prelude::*};
 
+#[pymodule]
 fn rust_snuba(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
-    // let DSN
-    let sentry_dsn = Python::with_gil(|py| {
+    setup_sentry(_py);
+    m.add_function(wrap_pyfunction!(consumer::consumer, m)?)?;
+    m.add_function(wrap_pyfunction!(setup_sentry, m)?)?;
+    Ok(())
+}
+
+#[pyfunction]
+pub fn setup_sentry(_py: Python<'_>) {
+    env_logger::init();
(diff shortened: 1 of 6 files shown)


rust_snuba/src/lib.rs
```diff
@@ -7,6 +7,20 @@ use pyo3::prelude::*;
...
 #[pymodule]
 fn rust_snuba(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    // let DSN
+    let sentry_dsn = Python::with_gil(|py| {
+        let snuba_settings = PyModule::import(py, "snuba.settings")?;
+        let dsn: Result<String, PyErr> = snuba_settings
+            .getattr("SENTRY_DSN")?
+            .extract();
+        dsn
+    }).unwrap();
```
lynnagara, line 17: SENTRY_DSN is optional since not all environments send errors to Sentry. We should probably skip the Sentry initialization if it's None rather than crash.

dbanda: Updated the func. It should be safe now, but maybe using py import is overkill and I should just export the DSN to an environment variable and fetch it from Rust?

rust_snuba/src/lib.rs
```diff
@@ -3,10 +3,48 @@ mod consumer;
...
     m.add_function(wrap_pyfunction!(consumer::consumer, m)?)?;
+    m.add_function(wrap_pyfunction!(setup_sentry, m)?)?;
     Ok(())
 }
+
+#[pyfunction]
+pub fn setup_sentry(_py: Python<'_>) {
+    env_logger::init();
+    let sentry_dsn: Result<Option<String>, PyErr> = Python::with_gil(|py| {
+        let snuba_settings = PyModule::import(py, "snuba.settings")?;
```
untitaker, line 20: Instead of this I would only call `setup_sentry` in consumer::consumer, and use the existing config payload which we already use to pass consumer config, to pass the DSN.

dbanda: ok. I updated it to be that way.

lynnagara (approved):

PR #4055, opened 2023-04-20
feat(CapMan): Basic Snuba Admin UI

rahul-kumar-saini:
### Overview
- Allocation Policies need a better way to be interacted with than the runtime config
- Enter "Capacity Management" Snuba Admin panel
- This will enable users to modify the configs of each Allocation Policy
- The only "edit" you can make to any config is to its value

### Two kind of config
- Required
    - These are editable and can be reset to default but cannot be added or deleted
    - eg `is_enforced`, `is_enabled`
    - eg for errors: `throttled_thread_number`
- Parameterized / Optional
    - These are configs that can be added, edited, and deleted
    - eg some org + referrer specific bytes scanned could look like this:
        - key: `org_limit`, params: `{"organization": 123, "referrer": "something"}`
    - When you try to add a new config, only parameterized configs will be available to be added and you'll notice the delete option only exists on parameterized configs as well

### Notes
- This change contains the UI + API changes necessary to enable the panel, but not the backend changes that would make it actually do anything, that will be in a different PR
- All values you see in the panel are placeholders, nothing will get updated/deleted either
- This change introduced the `react-bootstrap` library and it only gets imported in the capacity management panel
    - Importing it imports a lot of styles that make everything else look "off" but in a future PR I'd like to reskin the entire admin console to use the library properly

Demo:

https://user-images.githubusercontent.com/23463457/236331397-94aedc31-0668-4a22-99c6-d352ab8e389e.mov


snuba/admin/static/capacity_management/types.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+type RowData = [
+  ReactNode,
+  ReactNode,
+  ReactNode,
+  ReactNode,
+  ReactNode,
+  ReactNode
+];
```
volokluev, line 36: this is a strange type. What's the story behind it?

snuba/admin/views.py
```diff
@@ -708,3 +711,109 @@ def snql_to_sql() -> Response:
...
+@check_tool_perms(tools=[AdminTools.CAPACITY_MANAGEMENT])
+def allocation_policies() -> Response:
+    return Response(
+        json.dumps(get_allocation_policies()), 200, {"Content-Type": "application/json"}
+    )
+
+
+@application.route("/allocation_policy_configs/<path:storage>", methods=["GET"])
+@check_tool_perms(tools=[AdminTools.CAPACITY_MANAGEMENT])
+def get_allocation_policy_configs(storage: str) -> Response:
```
volokluev, line 726: is this the current configuration values or what can be configured on the policy?

snuba/admin/views.py
```diff
@@ -708,3 +711,109 @@ def snql_to_sql() -> Response:
...
+            {"Content-Type": "application/json"},
+        )
+
+
+@application.route(
+    "/allocation_policy_parameterized_config_definitions/<path:storage>",
+    methods=["GET"],
+)
+@check_tool_perms(tools=[AdminTools.CAPACITY_MANAGEMENT])
+def get_allocation_policy_parameterized_config_definitions(storage: str) -> Response:
```
volokluev, line 744: why does this need a separate endpoint?

volokluev (commented):
Some questions. Also can you file a ticket to write tests for this once it's hooked up?

snuba/admin/static/capacity_management/types.tsx
rahul-kumar-saini: Copied it from runtime config code and added an extra `ReactNode` to work with our use case lol, it's just what the `Table` component requires:

https://github.com/getsentry/snuba/blob/21864c360af65cb9aac0b10866bdab1a3f49f53e/snuba/admin/static/runtime_config/types.tsx#L16

snuba/admin/views.py
rahul-kumar-saini: This is kind of for answering the question "What optional configs do we have and how are they defined?".

This informs the "Add new" modal with what can be configured and what it expects for types of the parameters etc. I'll change the "parameterized" nomenclature to "optional".

snuba/admin/views.py
rahul-kumar-saini: This gets the current configuration values, I'll update the name to be more clear.

volokluev (approved):

PR #4061, opened 2023-04-21
feat(rust-arroyo): Reduce strategy

lynnagara:
Currently the consumer does not do any batching, add a reduce strategy to help with this.

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
+++ b/rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
@@ -1,12 +1,12 @@
 use crate::processing::strategies::{CommitRequest, MessageRejected, ProcessingStrategy};
 use crate::types::{AnyMessage, InnerMessage, Message, Partition};
 use std::collections::BTreeMap;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
 struct BatchState<T, TResult> {
     value: Option<TResult>,
-    accumulator: Arc<Mutex<dyn FnMut(TResult, T) -> TResult + Send + Sync>>,
+    accumulator: Arc<dyn Fn(TResult, T) -> TResult + Send + Sync>,
     offsets: BTreeMap<Partition, u64>,
     batch_start_time: SystemTime,
     message_count: usize,
@@ -14,7 +14,7 @@ struct BatchState<T, TResult> {
 }
 
 impl<T: Clone, TResult: Clone> BatchState<T, TResult> {
-    fn new(initial_value: TResult, accumulator: Arc<Mutex<dyn FnMut(TResult, T) -> TResult + Send + Sync>>) -> BatchState<T, TResult> {
+    fn new(initial_value: TResult, accumulator: Arc<dyn Fn(TResult, T) -> TResult + Send + Sync>) -> BatchState<T, TResult> {
         BatchState {
             value: Some(initial_value),
             accumulator,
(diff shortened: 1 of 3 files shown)


rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
```diff
@@ -0,0 +1,195 @@
...
+            batch_start_time: SystemTime::now(),
+            message_count: 0,
+            is_complete: false,
+        }
+    }
+
+    fn add(&mut self, message: Message<T>) {
+        let mut tmp = None;
+        mem::swap(&mut tmp, &mut self.value);
+        self.value = Some((self.accumulator)(tmp.unwrap(), message.payload()));
```
lynnagara, line 31: Not a fan of this but didn't find a better solution, should I switch to using `take_mut` crate or something else?

untitaker: I think this is fine. Since you're now using Option, you can actually use `self.accumulator.take()` instead of `mem::swap`

rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+pub struct Reduce<T, TResult> {
+    next_step: Box<dyn ProcessingStrategy<TResult>>,
+    accumulator: fn(TResult, T) -> TResult,
+    initial_value: TResult,
+    max_batch_size: usize,
+    max_batch_time: Duration,
+    batch_state: BatchState<T, TResult>,
```
untitaker, line 46: I would suggest to have `batch_state: None`, then lazily initializing `BatchState to `Some()` when a new batch is started. I believe this will allow you to use `self.batch_state.take()` when flushing (which avoids clones)

lynnagara: The main reason I was cloning was for the `MessageRejected` case... where we would need to put back the batch state. Maybe we could send it back with `MessageRejected`?

rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
```diff
@@ -0,0 +1,201 @@
+use crate::processing::strategies::{CommitRequest, MessageRejected, ProcessingStrategy};
+use crate::types::{AnyMessage, InnerMessage, Message, Partition};
+use std::collections::BTreeMap;
+use std::time::{Duration, SystemTime};
+
+struct BatchState<T, TResult> {
+    value: Option<TResult>,
+    accumulator: fn(TResult, T) -> TResult,
```
untitaker, line 8:
```suggestion
struct BatchState<T, TResult, F: FnMut(TResult, T) -> TResult> {
    value: Option<TResult>,
    accumulator: F,
```

untitaker: you can also make `accumulator: Box<dyn FnMut(...) -> ...>` if it's easier, then the generic is not needed on the struct, but only in `new`

this is probably better if people want to write `BatchState<T, TResult>`, and they can't if there's a closure being passed.

untitaker (approved):

rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
```diff
@@ -0,0 +1,202 @@
...
+            offsets: Default::default(),
+            batch_start_time: SystemTime::now(),
+            message_count: 0,
+            is_complete: false,
+        }
+    }
+
+    fn add(&mut self, message: Message<T>) {
+        let tmp = self.value.take();
+        self.value = Some((self.accumulator.lock().unwrap())(tmp.unwrap(), message.payload()));
```
lynnagara, line 30: @untitaker Changed to `Arc<Mutex<dyn FnMut>>`... should I be handling the poisoned mutex error here?

untitaker: oh. uhh, please change `FnMut` to `Fn` to avoid the Mutex. Alternatively please try Rc instead of Arc.

lynnagara: Ah, thanks. Removed the FnMut/Mutex

PR #4062, opened 2023-04-21
ref(admin) Add a Tracing role to let replays access that tool

evanh:
Add a new role that gives access to the Tracing tool in the admin. Replays is
doing a lot of experimentation on that right now and it would help them.


volokluev (approved):
nice

PR #4063, opened 2023-04-21
feat(rust): add some metrics backends

dbanda:
Switches the metrics client to be a trait and introduces a testing and datadog backend of the metrics client. Datadog is still WIP but this impl uses the dogstatsd library.

### DIFF
--- a/rust_snuba/rust_arroyo/src/utils/metrics.rs
+++ b/rust_snuba/rust_arroyo/src/utils/metrics.rs
@@ -62,7 +62,7 @@ impl MetricsClientTrait for MetricsClient {
             match result {
                 Ok(_) => {}
                 Err(_err) => {
-                    println!("Failed to send metric {}: {}", key, _err)
+                    println!("Failed to send metric {key}: {_err}")
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


rust_snuba/src/utils/metrics/backends/datadog.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        match value {
+            Some(v) => {
+                for _ in 0..v.abs() {
+                    if v < 0 {
+                        self.client_sd.decr(key, &tags_str).unwrap();
+                    } else {
+                        self.client_sd.incr(key, &tags_str).unwrap();
+                    }
+                }
```
onewland, line 62: should this logic be replaced by https://docs.rs/dogstatsd/latest/dogstatsd/struct.Client.html#method.count ?

Seems inefficient to send `n` decr messages vs. subtract `n` if that's possible

dbanda: Yeah that was so dumb. Updated.

untitaker (commented):
i would recommend using cadence instead of dogstatsd, and take a look at relay for how it is set up there

rust_snuba/rust_arroyo/src/utils/metrics.rs
```diff
@@ -8,7 +8,12 @@ use std::collections::HashMap;
 use std::net::{ToSocketAddrs, UdpSocket};
 use std::sync::Arc;
 
-pub trait Metrics {
+pub trait MetricsClientTrait: Send + Sync  {
```
untitaker, line 11: it's not idiomatic to name traits as `Trait`

MeredithAnya:
I think we already have metrics now

PR #4065, opened 2023-04-21
feat(chaos): bypass readthrough cache by table name

volokluev:
Provide the ability to not apply readthrough caching to specific clickhouse tables. This is necessary for a chaos engineering experiment but it's not how I would like this feature to be implemented (if we actually want it). This PR should be reverted after the experiment is complete.


untitaker (changes requested):
I believe this won't work because we generate the query ID as a kind of hash from the query input, and if two identical queries hit clickhouse, clickhouse will reject one of them for having the same query ID as the other one.

I was discovering this case when we moved cache to a different redis cluster.

In response to that very short-lived incident, I added an option to retry duplicate queries: https://github.com/getsentry/snuba/blob/c9c61cd5479803c884e0567f9fe63f6c2edbcae9/snuba/web/db_query.py#L376-L380

you might want to enable that first before trying this experiment, or make all query IDs random from the start to avoid the roundtrip from retry

volokluev:
Thanks for the tip Markus!

untitaker (approved):
that probably works. mind making this a sample rate instead of bool? not a big blocker, just suggestion

PR #4074, opened 2023-04-24
feat(DRS): Fail queries when storage is unavailable due to readiness state

enochtangg:
### Overview
This PR is responsible for failing queries to Snuba storages which are not ready yet. For example, a new experimental storage (`readiness_state=limited`) should not be queryable in Sentry prod environments. This PR adds `StorageNotAvailable` error messaging in the storage plan builder. Additionally, in order for this new error to not count against our SLO, a new error type called `QueryPlanException` was created to capture errors outside the query execution flow. More details on how to review this PR below.

### More Details 
I would recommend reviewing this PR in the following order:
* `snuba/datasets/plans/storage_plan_builder.py`: After the appropriate storage is selected for the incoming query, a readiness_state check was added here and throws an error if it is unavailable.
* `snuba/web/views.py`: The exception is bubble back to the original endpoint and the appropriate status code and tags are added.
* `snuba/query/exceptions.py`: Added new exception `QueryPlanException` to capture exceptions outside of query execution. 
* `snuba/web/query.py`: Catches the new exception type and forwards to querylog appropriately.
* * `snuba/querylog/__init__.py`:  Updated `_record_timer_metrics()` to accommodate for new error type which does not use query_list to determine status and SLO.
* `snuba/web/__init__.py`: Added `exception_type: str` to `QueryException` to track the name of which exception was thrown.

* `snuba/datasets/storage.py`: Move `readiness_state` attribute to base `Storage` class.
* `tests/datasets/plans/test_storage_plan_builder.py`: unit test
* `settings files`: Added `READINESS_STATE_FAIL_QUERIES` to control failing queries in each sentry env.

### Blast Radius
All incoming queries to storages which are not available in a given environment will now fail. However, right now, all sentry environments are currently configured to support all readiness states. Therefore, there should no immediate effect until we begin rollout. The settings variable `READINESS_STATE_FAIL_QUERIES` to control which sentry envs will use this change.

### DIFF
--- a/snuba/datasets/plans/storage_plan_builder.py
+++ b/snuba/datasets/plans/storage_plan_builder.py
@@ -43,7 +43,7 @@ from snuba.utils.metrics.util import with_span
 # dependency is a refactoring of the methods that return RawQueryResult to make them
 # depend on Result + some debug data structure instead. Also It requires removing
 # extra data from the result of the query.
-from snuba.web import QueryResult
+from snuba.web import QueryException, QueryResult
 
 
 class SimpleQueryPlanExecutionStrategy(QueryPlanExecutionStrategy[Query]):
@@ -182,9 +182,17 @@ class StorageQueryPlanBuilder(ClickhouseQueryPlanBuilder):
         assert isinstance(storage, ReadableTableStorage)
         readiness_state = storage.get_readiness_state()
         if readiness_state.value not in snuba_settings.SUPPORTED_STATES:
-            raise StorageNotAvailable(
+            cause = StorageNotAvailable(
                 f"The selected storage={storage.get_storage_key().value} is not available in this environment yet. To enable it, consider bumping the storage's readiness_state."
             )
+            raise QueryException.from_args(
+                str(cause),
+                extra={
+                    "stats": {},
+                    "sql": "",
+                    "experiments": {},
+                },
+            ) from cause
 
         with sentry_sdk.start_span(
             op="build_plan.storage_query_plan_builder", description="translate"
--- a/snuba/datasets/plans/storage_plan_builder.py
+++ b/snuba/datasets/plans/storage_plan_builder.py
@@ -28,7 +28,6 @@ from snuba.datasets.storage import (
 )
 from snuba.query.allocation_policies import AllocationPolicy
 from snuba.query.data_source.simple import Table
-from snuba.query.exceptions import QueryPlanException
 from snuba.query.logical import Query as LogicalQuery
 from snuba.query.processors.physical import ClickhouseQueryProcessor
 from snuba.query.processors.physical.conditions_enforcer import (
@@ -184,13 +183,10 @@ class StorageQueryPlanBuilder(ClickhouseQueryPlanBuilder):
             assert isinstance(storage, ReadableTableStorage)
             readiness_state = storage.get_readiness_state()
             if readiness_state.value not in snuba_settings.SUPPORTED_STATES:
-                cause = StorageNotAvailable(
-                    f"The selected storage={storage.get_storage_key().value} is not available in this environment yet. To enable it, consider bumping the storage's readiness_state."
+                raise StorageNotAvailable(
+                    StorageNotAvailable.__name__,
+                    f"The selected storage={storage.get_storage_key().value} is not available in this environment yet. To enable it, consider bumping the storage's readiness_state.",
                 )
-                raise QueryPlanException.from_args(
-                    exception_type=StorageNotAvailable.__name__,
-                    message=str(cause),
-                ) from cause
 
         with sentry_sdk.start_span(
             op="build_plan.storage_query_plan_builder", description="translate"
--- a/snuba/datasets/plans/storage_plan_builder.py
+++ b/snuba/datasets/plans/storage_plan_builder.py
@@ -2,6 +2,7 @@ from typing import List, Optional, Sequence
 
 import sentry_sdk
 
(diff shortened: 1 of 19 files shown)


snuba/datasets/plans/storage_plan_builder.py
```diff
@@ -176,6 +178,23 @@ def build_and_rank_plans(
         mappers = storage_connection.translation_mappers
         cluster = self.get_cluster(storage, query, settings)
 
+        # Return failure if storage readiness state is not supported in current environment
+        assert isinstance(storage, ReadableTableStorage)
```
evanh, line 182: Why this assert?

snuba/datasets/storage.py
```diff
@@ -205,5 +206,5 @@ class EntityStorageConnectionNotFound(Exception):
     pass
 
 
-class StorageNotFound(Exception):
```
evanh, line 208: Was this exception not being used?

evanh (commented):
My main question with this PR is: why does the `StorageNotAvailable` exception need to be wrapped in a `QueryException` at all? Why can't the code return that exception directly, and then web/views.py can be changed to handle it? That seems like it would simplify things.

snuba/datasets/storage.py
enochtangg: yup, it wasn't used.

enochtangg:
@evanh > Why does the StorageNotAvailable exception need to be wrapped in a QueryException at all? Why can't the code return that exception directly, and then web/views.py can be changed to handle it? That seems like it would simplify things.

I agree. By returning `StorageNotAvailable` directly and modifying views.py will simplify all the query metadata stuff. In my first pass, I found that the [QueryException docstring](https://github.com/getsentry/snuba/blob/5cf1fe0bd521db94646ac4ec978f084b0db9349c/snuba/web/__init__.py#L17-L21) clearly described the type of error the I wanted to throw which is why I decided to use it.. unaware of the metadata work.

What are your thoughts on creating a new exception class like `SystemException` where exceptions like clickhouse connection, storage unavailable, and etc. exceptions wrap around?

snuba/datasets/plans/storage_plan_builder.py
enochtangg: Moved to base class. But `assert` is still required to access to storage_key.

enochtangg:
Added new error called `QueryPlanException` to capture all errors outside the query execution phase. The summary of this PR was updated accordingly (cc @evanh).

snuba/datasets/plans/storage_plan_builder.py
```diff
@@ -176,6 +179,19 @@ def build_and_rank_plans(
         mappers = storage_connection.translation_mappers
         cluster = self.get_cluster(storage, query, settings)
 
+        # Return failure if storage readiness state is not supported in current environment
+        if snuba_settings.READINESS_STATE_FAIL_QUERIES:
+            assert isinstance(storage, ReadableTableStorage)
```
evanh, line 184: I thought you were removing this?

snuba/datasets/plans/storage_plan_builder.py
```diff
@@ -176,6 +179,19 @@ def build_and_rank_plans(
...
+            readiness_state = storage.get_readiness_state()
+            if readiness_state.value not in snuba_settings.SUPPORTED_STATES:
+                cause = StorageNotAvailable(
+                    f"The selected storage={storage.get_storage_key().value} is not available in this environment yet. To enable it, consider bumping the storage's readiness_state."
+                )
+                raise QueryPlanException.from_args(
```
evanh, line 190: Why do you wrap `StorageNotAvailable` in `QueryPlanException` instead of having `StorageNotAvailable` be a subclass of `QueryPlanException`?

snuba/querylog/__init__.py
```diff
@@ -21,29 +30,47 @@ def _record_timer_metrics(
...
+        "referrer": referrer,
+        "parent_api": parent_api,
+        "dataset": query_metadata.dataset,
+    }
+    if isinstance(result, QueryPlanException):
+        # The QueryPlanException is raised outside the query execution flow.
+        # As a result, its status and SLO values are not based on its query_list
+        if result.exception_type == StorageNotAvailable.__name__:
+            tags = {
+                "status": QueryStatus.ERROR.value,
```
evanh, line 60: I would rather you used the `get_request_status` for the `request_status` and `slo` parameters. I am trying to keep the logic that converts exceptions -> statuses central in one place.

snuba/datasets/plans/storage_plan_builder.py
enochtangg: Left a comment above. Apparently, we need it to access `get_storage_key()` since storage keys are not part of base.

snuba/datasets/plans/storage_plan_builder.py
enochtangg: Just following the pattern we use for QueryException. For example, https://github.com/getsentry/snuba/blob/6582b774e60065065271ef74b4720eef8767f155/snuba/web/query.py#L342-L349

enochtangg: I was following the pattern we use for QueryException, but making StorageNotAvailable inherit from `QueryPlanException` simplifies a few type checks. Updated!

snuba/querylog/__init__.py
enochtangg: Makes sense, updated.

evanh (approved):

PR #4075, opened 2023-04-24
fix: Rate-limit the number of invalid message payloads

untitaker:
Instead of sampling invalid messages, let's throttle them so we don't
miss spurious, small issues, and globally so we don't forget to do it
for a particular dataset.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -79,6 +79,8 @@ logger = logging.getLogger("snuba.consumer")
 
 commit_codec = CommitCodec()
 
+_LAST_INVALID_MESSAGE: MutableMapping[str, float] = {}
+
 
 class CommitLogConfig(NamedTuple):
     producer: ConfluentProducer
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

snuba/consumers/consumer.py
```diff
@@ -511,14 +513,14 @@ def process_message(
...
                 except Exception as err:
-                    config = state.get_config(
-                        f"log_validate_schema_{snuba_logical_topic.name}", 1.0
-                    )
-                    log_validate_sample_rate = float(
-                        config if config is not None else 1.0
+                    min_seconds_ago = (
+                        state.get_config("log_validate_schema_every_n_seconds", 1) or 1
                     )
-                    sentry_sdk.set_tag("invalid_message_schema", "true")
```
MeredithAnya, line 520: do you care about keeping this tag?

untitaker: yes, i was just typing too fast

PR #4077, opened 2023-04-24
feat(replays): Cap URL length and decrease the maximum length of array columns

cmanallen:
- Each URL is capped to a maximum of 128 bytes.
- The URLs array will only accept up to 10 URLs per segment.
- Error and trace ID arrays have the same length restriction.

These changes are intended to impose reasonable limits on our ingest process.  They may have a positive affect on our current OOM issues.  Reference: https://sentry.sentry.io/issues/3989847924/?referrer=resolved_activity-slack.  However, this change is not meant to solve that issue instead these limits exist to prevent badly behaving clients or extremely long URLs from degrading normal usage of the product.


evanh (approved):

JoshFerge (approved):
will want to see if any users have feedback about not finding array items / long strings they're looking for, but this lgtm.

PR #4081, opened 2023-04-24
Feat/spans dataset

nikhars:
WIP

### DIFF
--- a/snuba/snuba_migrations/spans/0001_spans_v1.py
+++ b/snuba/snuba_migrations/spans/0001_spans_v1.py
@@ -36,6 +36,7 @@ columns: List[Column[Modifiers]] = [
     Column("module", String(Modifiers(low_cardinality=True, nullable=True))),
     Column("action", String(Modifiers(low_cardinality=True, nullable=True))),
     Column("domain", String(Modifiers(nullable=True))),
+    Column("platform", String(Modifiers(low_cardinality=True, nullable=True))),
     Column("user", String(Modifiers(nullable=True))),
     Column("tags", Nested([("key", String()), ("value", String())])),
     Column(
--- a/snuba/snuba_migrations/spans/0001_spans_v1.py
+++ b/snuba/snuba_migrations/spans/0001_spans_v1.py
@@ -1,0 +1,94 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import UUID, Column, DateTime, Nested, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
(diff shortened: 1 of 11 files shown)


evanh (commented):
Also note you'll have to split this PR between migration code and everything else.

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, segment_name, segment_id, timestamp)",
```
fpacifici, line 65: If segment_id comes before timestamp, the timestamp field is not going to be used in a query across different segments. I think you will not want to do that. 
Do we have a segment start timestamp ? to do something like: `segment_start_ts, segment_id, teimstamp` ?
Or something like `toStartOfHour(timestamp), segment_id, timestamp`

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    Column("group", UInt(64)),
+    Column("span_status", UInt(8)),
+    Column("span_kind", String(Modifiers(low_cardinality=True))),
+    Column("description", String()),
+    Column("status", UInt(32, Modifiers(nullable=True))),
+    Column("module", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("action", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("domain", String(Modifiers(nullable=True))),
+    Column("platform", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("user", String(Modifiers(nullable=True))),
```
fpacifici, line 40: Are these actually all nullable ?

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+                ),
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=dist_table_name,
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name=local_table_name,
+                    sharding_key="cityHash64(segment_id)",
```
fpacifici, line 81: Why not sharding by trace id so that we can run complex subqueries on the same trace locally ?

fpacifici:
@nikhars  do we have a sense of the types of queries we are optimizing for ?

PR #4052, opened 2023-04-19
ref(consumer): Pass sentry_received_timestamp for e2e latency 

ayirr7:
Allows us to use a new field `sentry_received_timestamp` for both e2e (indexer and snuba) and indexer-only latency metrics

https://getsentry.atlassian.net/browse/SNS-2173

PR #4085, opened 2023-04-25
ref(consumers): record sentry_received_timestamp metrics

MeredithAnya:
**context**
We want to be able to include the latency from sentry indexer for metrics and generic metrics SLOs. We needed to add the `sentry_received_timestamp` in order to do so since the other timestamps we had were either the broker timestamp (so latency of the snuba consumer itself) or the relay timestamp (which we didn't want to include as part of SNS SLOs)

**pre reqs**
- [x] `sentry_received_timestamp` must be added to the payload in the indexer https://github.com/getsentry/sentry/pull/47577
- [x] `sentry_recieved_timestamp` must be passed through the processor in snuba ~https://github.com/getsentry/snuba/pull/4052~  **edit** this pr had to be re-done so we need https://github.com/getsentry/snuba/pull/4198

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -117,19 +117,24 @@ class LatencyRecorder:
     def __init__(self) -> None:
         self._sum = 0.0
         self._max: Optional[float] = None
+        self._msg_count = 0
 
-    def record(self, latency: float) -> None:
-        self._sum += latency
-        if self._max is None or latency > self._max:
-            self._max = latency
+    def record(self, latency_seconds: float) -> None:
+        self._msg_count += 1
+        self._sum += latency_seconds
+        if self._max is None or latency_seconds > self._max:
+            self._max = latency_seconds
+
+    def has_data(self) -> bool:
+        return self._max is not None
 
     @property
-    def sum(self) -> float:
-        return self._sum
+    def avg_ms(self) -> float:
+        return (self._sum / self._msg_count) * 1000
 
     @property
-    def max(self) -> Optional[float]:
-        return self._max
+    def max_ms(self) -> Optional[float]:
+        return self._max * 1000
 
 
 class InsertBatchWriter:
@@ -181,28 +186,26 @@ class InsertBatchWriter:
                 )
                 sentry_received_latency_recorder.record(sentry_received_latency)
 
-        if snuba_latency_recorder.max is not None:
-            self.__metrics.timing("max_latency_ms", snuba_latency_recorder.max * 1000)
+        if snuba_latency_recorder.has_data():
+            self.__metrics.timing("max_latency_ms", snuba_latency_recorder.max_ms)
+            self.__metrics.timing("latency_ms", snuba_latency_recorder.avg_ms)
+        if end_to_end_latency_recorder.has_data():
             self.__metrics.timing(
-                "latency_ms", (snuba_latency_recorder.sum / len(self.__messages)) * 1000
-            )
-        if end_to_end_latency_recorder.max is not None:
-            self.__metrics.timing(
-                "max_end_to_end_latency_ms", end_to_end_latency_recorder.max * 1000
+                "max_end_to_end_latency_ms", end_to_end_latency_recorder.max_ms
             )
             self.__metrics.timing(
                 "end_to_end_latency_ms",
-                (end_to_end_latency_recorder.sum / len(self.__messages)) * 1000,
+                end_to_end_latency_recorder.avg_ms,
             )
 
-        if sentry_received_latency_recorder.max is not None:
+        if sentry_received_latency_recorder.has_data():
             self.__metrics.timing(
                 "max_sentry_received_latency_ms",
-                sentry_received_latency_recorder.max * 1000,
+                sentry_received_latency_recorder.max_ms,
             )
             self.__metrics.timing(
                 "sentry_received_latency_ms",
-                (sentry_received_latency_recorder.sum / len(self.__messages)) * 1000,
+                sentry_received_latency_recorder.avg_ms,
             )
 
         self.__metrics.timing("batch_write_ms", (write_finish - write_start) * 1000)
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -125,15 +125,14 @@ class LatencyRecorder:
         if self._max is None or latency_seconds > self._max:
             self._max = latency_seconds
 
-    def has_data(self) -> bool:
-        return self._max is not None
-
     @property
     def avg_ms(self) -> float:
         return (self._sum / self._msg_count) * 1000
 
     @property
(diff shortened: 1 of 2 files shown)


snuba/consumers/consumer.py
```diff
@@ -135,22 +135,37 @@ def close(self) -> None:
         latency_sum = 0.0
         max_end_to_end_latency: Optional[float] = None
         end_to_end_latency_sum = 0.0
+        max_sentry_received_latency: Optional[float] = None
+        sentry_received_latency_sum = 0.0
+
+        def update_latency(value: float, sum: float, max: Optional[float]) -> None:
```
onewland, line 141: I think when you pass floats like this to methods in python, mutations to the arguments only apply within the context of the method:

```
➜  ~ python
Python 3.8.12 (default, Nov  8 2021, 15:11:59) 
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def update_val(value):
...     value += 1
... 
>>> x = 5
>>> update_val(x)
>>> x
5
```

So if you want to continue with this approach I think you'll need something like a `LatencyRecorder` object which holds onto a reference to something whose contents exist outside the scope of the method

MeredithAnya: yup updated

ayirr7:
Approach looks good to me!

snuba/consumers/consumer.py
```diff
@@ -140,39 +159,50 @@ def close(self) -> None:
...
-        if max_end_to_end_latency is not None:
+        if end_to_end_latency_recorder.max is not None:
             self.__metrics.timing(
-                "max_end_to_end_latency_ms", max_end_to_end_latency * 1000
+                "max_end_to_end_latency_ms", end_to_end_latency_recorder.max * 1000
             )
             self.__metrics.timing(
                 "end_to_end_latency_ms",
-                (end_to_end_latency_sum / len(self.__messages)) * 1000,
+                (end_to_end_latency_recorder.sum / len(self.__messages)) * 1000,
```
ayirr7, line 195: Maybe not important, but should we have the avg timing metric not be under the` if max is not None` condition?

snuba/consumers/consumer.py
```diff
@@ -140,39 +159,50 @@ def close(self) -> None:
...
-
-        if max_latency is not None:
-            self.__metrics.timing("max_latency_ms", max_latency * 1000)
+                sentry_received_latency_recorder.record(sentry_received_latency)
+
+        if snuba_latency_recorder.max is not None:
+            self.__metrics.timing("max_latency_ms", snuba_latency_recorder.max * 1000)
             self.__metrics.timing(
-                "latency_ms", (latency_sum / len(self.__messages)) * 1000
+                "latency_ms", (snuba_latency_recorder.sum / len(self.__messages)) * 1000
```
onewland, line 187: You could add to a counter inside of `LatencyRecorder` that increments on every `record`, and `avg_ms()` could be a method on that class that divides the sum by length, and then multiply by 1000.

snuba/consumers/consumer.py
```diff
@@ -140,39 +159,50 @@ def close(self) -> None:
...
             )
-        if max_end_to_end_latency is not None:
+        if end_to_end_latency_recorder.max is not None:
```
onewland, line 189: Rather than checking if `recorder.max is not None` maybe we could do `recorder.has_data()` (which can check if `max is not None`) which reveals less about the implementation to the caller

snuba/consumers/consumer.py
```diff
@@ -113,6 +113,25 @@ def __reduce_ex__(
             )
 
 
+class LatencyRecorder:
+    def __init__(self) -> None:
+        self._sum = 0.0
+        self._max: Optional[float] = None
+
+    def record(self, latency: float) -> None:
```
onewland, line 121: let's call it `latency_seconds` since we're assuming we have to multiply by 1000 to get ms later

onewland (approved):
I'm good with this as-is, but added my thoughts on some basically aesthetic changes.

snuba/consumers/consumer.py
MeredithAnya: added both `avg_ms` and `max_ms` s

onewland (approved):

lynnagara (dismissed):
What's the reason for all of the separate latency metrics (average, max, sum)? It feels a bit overkill, and I'd prefer to not add so much overhead (including instantiating 3 LatencyRecorder instances on every batch) if there's not a strong reason for having this.

I see that average may be useful for SLO purposes but I can't think of any use case for sum - by itself the sum isn't meaningful since it depends on the number of messages that just so happens to be in that batch. I also don't think max is particularly useful since messages are processed in order (within a partition). So we don't expect to see outliers or anything here.

fpacifici:
Our SLOs are based on `max` so that is needed. 
The avg is a good indicator about the variance. We keep it in many dashboards to see whether we have outliers or the normal behavior is unhealthy.

snuba/consumers/consumer.py
```diff
@@ -140,39 +163,48 @@ def close(self) -> None:
...
-                    max_end_to_end_latency = end_to_end_latency
-
-        if max_latency is not None:
-            self.__metrics.timing("max_latency_ms", max_latency * 1000)
+                sentry_received_latency_recorder.record(sentry_received_latency)
+
+        if snuba_latency_recorder.max_ms:
+            self.__metrics.timing("max_latency_ms", snuba_latency_recorder.max_ms)
+            self.__metrics.timing("latency_ms", snuba_latency_recorder.avg_ms)
+        if end_to_end_latency_recorder.max_ms:
```
untitaker, line 191: would it make sense to record this as a gauge instead of a timing?

if we record it as a timing, we get metrics like `max_latency_ms.max` and `max_latency.avg`, which are undesirable IMO

untitaker (approved):
seeing that we already record those metrics, and all this does is batch teh calls to `timing()`. I do think we should try and avoid recording too many metrics (considering that max is also exploded into a timeseries by veneur), but that's a bigger topic

snuba/consumers/consumer.py
```diff
@@ -140,39 +163,48 @@ def close(self) -> None:
...
             self.__metrics.timing(
-                "max_end_to_end_latency_ms", max_end_to_end_latency * 1000
+                "max_sentry_received_latency_ms",
+                sentry_received_latency_recorder.max_ms,
             )
             self.__metrics.timing(
-                "end_to_end_latency_ms",
-                (end_to_end_latency_sum / len(self.__messages)) * 1000,
+                "sentry_received_latency_ms",
+                sentry_received_latency_recorder.avg_ms,
```
fpacifici, line 207: Why did we change the name of this metric? 
- There are teams using this as an SLO like profiling. 
- There may not be sentry in front of snuba. Thus it was called ned to end.

PR #4087, opened 2023-04-25
fix(observability): Not all ClickHouse Errors need to be surfaced to Sentry

rahul-kumar-saini:
### Overview
- We're seeing a lot of redundant ClickHouse errors in Sentry
- The only ones we really want to see are internal errors - not rate limits or other user errors

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -44,9 +44,9 @@ from snuba.query.data_source.join import JoinClause
 from snuba.query.data_source.simple import Table
 from snuba.query.query_settings import QuerySettings
 from snuba.querylog.query_metadata import (
+    SLO,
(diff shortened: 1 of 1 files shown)


snuba/web/db_query.py
```diff
@@ -598,20 +598,15 @@ def _raw_query(
...
+        if request_status.status == RequestStatus.ERROR:
             logger.exception("Error running query: %s\n%s", sql, cause)
 
         with configure_scope() as scope:
             if scope.span:
                 sentry_sdk.set_tag("slo_status", request_status.status.value)
 
         stats = update_with_status(
-            status=status,
+            status=status or QueryStatus.ERROR,
```
volokluev, line 609: why was this line necessary?

volokluev (approved):

PR #4089, opened 2023-04-26
feat(spans): Initial spans migrations

nikhars:
This is the first iteration of building the spans schema based on the inputs provided by the product team.

Key considerations taken while designing the schema. 
1. Queries would always be made with a project id filter and hence it is the first entry of the sorting key.
2. The current transaction name is named as `segment_name` in the spans schema. `segment_name` would be added to each span so that data can be sorted based on `segment_name`. For spans (in the future) which will not be part of a transaction, those spans will have an empty `segment_name`.
3. The `event_id` from the transactions topic is named as `transaction_id` in the spans schema. So if a relationship needs to be made between what is stored in clickhouse and nodestore, `transaction_id` column can be used. We cannot use `segment_id` as `segment_id` is an 8 byte field whereas `event_id` is 16 byte field.
4. Information from the top level trace context forms the first span. For example, the data below would be used to fill in the first span. The `span_id` of the trace will become `segment_id`. `segment_id` would be added to each span belonging in the transaction payload so that data on clickhouse is sorted by `segment_id`.
```
{
   "event_id":"017a4b4199014c14ab6f523f9293c9fa",   <--- transaction_id (This is 16 bytes)
   "organization_id":1,
   "project_id":1,
   "message":"sentry.tasks.send_ping",                             <--- segment_name
   "data":{
      "contexts":{
	  "trace": {                                                                      <--- This will build the 1st span table row 
            "trace_id":"31f1c4f5fd60411389578c48639f358f",
            "span_id":"99733332b493047f",                           <--- span_id becomes segment_id for this span. It is copied to child spans for easier search by segment_id   
            "parent_span_id":"a53003ba84ed8b3c",
            "op":"queue.task.celery",
            "status":"ok",
            "exclusive_time":3.013,
            "type":"trace",
            "hash":"4a7e0ecc99a71a2a"
         }
      },
      "spans":[
         {                                                             <--- 2nd span table row
            "timestamp":1682188740.189884,
            "start_timestamp":1682188740.182367,
            "exclusive_time":7.517,
            "description":"connect",
            "op":"db",
            "span_id":"8288d24ef7c0d224",
            "parent_span_id":"99733332b493047f",
            "trace_id":"31f1c4f5fd60411389578c48639f358f",
            "same_process_as_parent":true,
            "tags":null,
            "data":null,
            "hash":"b640a0ce465fa2a4"
         },
         {                                                               <--- 3rd span table row
            "timestamp":1682188740.196926,
            "start_timestamp":1682188740.190089,
            "exclusive_time":6.837,
            "description":"UPDATE \"sentry_option\" SET \"value\" = %s, \"last_updated\" = %s WHERE \"sentry_option\".\"key\" = %s",
            "op":"db",
            "span_id":"bd4ab9cdfb2a53a2",
            "parent_span_id":"99733332b493047f",
            "trace_id":"31f1c4f5fd60411389578c48639f358f",
            "same_process_as_parent":true,
            "tags":null,
            "data":null,
            "hash":"6763cae24db9aa0e"
         },
         {                                                                 <--- 4th span table row
            "timestamp":1682188740.204828,
            "start_timestamp":1682188740.197253,
            "exclusive_time":3.329,
            "op":"transaction.atomic",
            "span_id":"b8f065b90845d390",
            "parent_span_id":"99733332b493047f",
            "trace_id":"31f1c4f5fd60411389578c48639f358f",
            "data":{
               "using":"default"
            },
            "same_process_as_parent":true,
            "description":null,
            "tags":null,
            "hash":"d41d8cd98f00b204"
         },
       }
   }
}
```


snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, segment_name, segment_id, timestamp)",
```
evanh, line 65: I'm not sure if it's a good idea to order by `segment_name` if it defaults to `''`. Also, would it make sense to have `trace_id` in this?

nikhars: Regarding `trace_id`, I would be curious to know between which columns we would put it in the existing order by key. 

One of the reasons I did not add `trace_id` is because the primary use case I considered is finding segments of interest. Traces span across different segments. With the current sorting key, parts of a trace would be in random `segment_names`. We could potentially add `trace_id` after `segment_id` and also add some sort of skip index so that finding all spans of a trace would be faster and the skip index could filter out lots of granules because `trace_id` is sorted. But I am not sure if there is any advantage to doing that compared with just adding a skip index on the column. The timestamp field in the sorting key would already be efficient in filtering out granules which would not have the `trace_id` since traces are short lived.

evanh: Ah, OK that all makes sense. If the primary thing is segment_id then leaving trace_id out makes sense to me too.

nikhars, line 65: > From https://github.com/getsentry/snuba/pull/4081#discussion_r1178527181, Filippo: If segment_id comes before timestamp, the timestamp field is not going to be used in a query across different segments. I think you will not want to do that.
Do we have a segment start timestamp ? to do something like: segment_start_ts, segment_id, teimstamp ?
Or something like toStartOfHour(timestamp), segment_id, timestamp

Not all spans would have a corresponding segment start timestamp. Only the `segment_name` and `segment_id` are copied to all child spans in the existing model. 

The current schema is optimized for searching for outlier `segment_id` of a `segment` based on different criteria's. Duration, module etc. @wmak, @shruthilayaj, @gggritso is this a fair assumption to make? Are your workflows around sampled spans going to be around this idea? If not, what would the workflow be?

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    Column("group", UInt(64)),
+    Column("span_status", UInt(8)),
+    Column("span_kind", String(Modifiers(low_cardinality=True))),
+    Column("description", String()),
+    Column("status", UInt(32, Modifiers(nullable=True))),
+    Column("module", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("action", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("domain", String(Modifiers(nullable=True))),
+    Column("platform", String(Modifiers(low_cardinality=True, nullable=True))),
+    Column("user", String(Modifiers(nullable=True))),
```
nikhars, line 40: > From https://github.com/getsentry/snuba/pull/4081#discussion_r1178527714, Filippo: Are these actually all nullable ?

Yes, it looks like except for `module` everything else can be nullable. @wmak @shruthilayaj, @gggritso thoughts?

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+                ),
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=dist_table_name,
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name=local_table_name,
+                    sharding_key="cityHash64(segment_id)",
```
nikhars, line 81: > From https://github.com/getsentry/snuba/pull/4081#discussion_r1178528521, Fillipo: Why not sharding by trace id so that we can run complex subqueries on the same trace locally ?

That is a good point. If we need to support querying by `trace_id`, sharding by `trace_id` would be better. Also, by virtue of a `segment_id` belonging to a `trace_id`, a given `segment_id` would always be on a given shard.I will modify it.

nikhars (commented):
Brining @fpacifici comment from draft PR to this PR.

snuba/snuba_migrations/spans/0001_spans_v1.py
wmak: @nikhars the current workflow for the modules is currently narrowing down to a specific span description(eg. a SQL query) using things like throughput, module, domain and action.
Then using that description to find corresponding transaction names
Then showing example span events (rows in this table) based on everything above

snuba/snuba_migrations/spans/0001_spans_v1.py
wmak: These can be nullable :+1:

wmak:
@nikhars regarding the current workflow its something like this:
1. Narrow down to a span description using a mix of domain, action, timestamps using Metrics Backed data
2. For the given span description along with domain etc. that could be selectively chosen find some samples using Indexed data
Here's an example of that query we're using right now on our hacked version of the table:

```sql
SELECT transaction_id,
      transaction,
      description,
      user,
      domain,
      span_id,
      sum(exclusive_time) as exclusive_time
 FROM spans_experimental_starfish
# This should be an int
WHERE group_id = '74580521-3cc2-9e7a-c5f3-0827db4fcb90'
  AND user = 'id:53791'
  AND greaterOrEquals(start_timestamp, '2023-04-13 14:05:26')
GROUP BY transaction_id,
         transaction,
         description,
         user,
         domain,
         span_id
ORDER BY exclusive_time desc
LIMIT 10

```

You can also watch the demos [here](https://sentry.rewatch.com/video/iq4qx12b1xergqrk-starfish-demos-feedback-april-27-2023) if you want to see the above with a UI

snuba/snuba_migrations/spans/0001_spans_v1.py
wmak: Normalized, the workflows actually depend on the description being normalized because otherwise we can't show useful groups of things. I think at most (we dont have anything for this yet) we just want to display the raw description next to the normalized one, but we can use eventstore for that

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+UNKNOWN_SPAN_STATUS = 2
+
+columns: List[Column[Modifiers]] = [
+    Column("project_id", UInt(64)),
+    Column("transaction_id", UUID(Modifiers(nullable=True))),
+    Column("trace_id", UUID()),
+    Column("span_id", UInt(64)),
+    Column("parent_span_id", UInt(64, Modifiers(nullable=True))),
+    Column("segment_id", UInt(64)),
+    Column("is_segment", UInt(8)),
```
dbanda, line 23: I'm a little confused as to why this is needed. In this model are spans and segments the same entity demarcated by this flag?

snuba/snuba_migrations/spans/0001_spans_v1.py
nikhars: Just reiterating to be on the same page. Please confirm
From Metrics - the current workflow for the modules is currently narrowing down to a specific span description(eg. a SQL query) using things like throughput, module, domain and action.
From Metrics - Then using that description to find corresponding transaction names
From Spans table - Then showing example span events (rows in this table) based on everything above

Now, for showing example span events, the workflows imply that most queries would have a where condition like this, correct?
```
WHERE project_id = A AND segment_name = B
```

snuba/snuba_migrations/spans/0001_spans_v1.py
nikhars: Yes. segments are the root spans from which a trace starts. It is not part of the list of spans sent from upstream. Look at the example trace in the description to see what a segment is.

nikhars:
> @nikhars regarding the current workflow its something like this:
> 
> 1. Narrow down to a span description using a mix of domain, action, timestamps using Metrics Backed data
> 2. For the given span description along with domain etc. that could be selectively chosen find some samples using Indexed data
>    Here's an example of that query we're using right now on our hacked version of the table:
> 
> ```sql
> SELECT transaction_id,
>       transaction,
>       description,
>       user,
>       domain,
>       span_id,
>       sum(exclusive_time) as exclusive_time
>  FROM spans_experimental_starfish
> # This should be an int
> WHERE group_id = '74580521-3cc2-9e7a-c5f3-0827db4fcb90'
>   AND user = 'id:53791'
>   AND greaterOrEquals(start_timestamp, '2023-04-13 14:05:26')
> GROUP BY transaction_id,
>          transaction,
>          description,
>          user,
>          domain,
>          span_id
> ORDER BY exclusive_time desc
> LIMIT 10
> ```
> 
> You can also watch the demos [here](https://sentry.rewatch.com/video/iq4qx12b1xergqrk-starfish-demos-feedback-april-27-2023) if you want to see the above with a UI

Do you want to store both raw and sanitized descriptions? I believe @fpacifici mentioned that storing raw descriptions is not a good idea here https://www.notion.so/Known-Needed-Data-66cdfbafb1bf43d58ddea5c4680eb4b8?d=2582b83f5cdd4daca898b6b0091a20ec#343cf218380c4cec9e741f27115f8a44. Do you still have a necessary use case to store both.

nikhars:
> ```sql
> WHERE group_id = '74580521-3cc2-9e7a-c5f3-0827db4fcb90'
> ```

@wmak - What is `group_id` in this query. Is it same as what `hash` is in this example span?
```
{                                                             
            "timestamp":1682188740.189884,
            "start_timestamp":1682188740.182367,
            "exclusive_time":7.517,
            "description":"connect",
            "op":"db",
            "span_id":"8288d24ef7c0d224",
            "parent_span_id":"99733332b493047f",
            "trace_id":"31f1c4f5fd60411389578c48639f358f",
            "same_process_as_parent":true,
            "tags":null,
            "data":null,
            "hash":"b640a0ce465fa2a4".   <--- is this group_id?
         },
```

nikhars:
> > ```sql
> > WHERE group_id = '74580521-3cc2-9e7a-c5f3-0827db4fcb90'
> > ```
> 
> @wmak - What is `group_id` in this example query. Is it same as what `hash` is in this example span? And are most queries from this table going to have this filter?
> 
> ```
> {                                                             
>             "timestamp":1682188740.189884,
>             "start_timestamp":1682188740.182367,
>             "exclusive_time":7.517,
>             "description":"connect",
>             "op":"db",
>             "span_id":"8288d24ef7c0d224",
>             "parent_span_id":"99733332b493047f",
>             "trace_id":"31f1c4f5fd60411389578c48639f358f",
>             "same_process_as_parent":true,
>             "tags":null,
>             "data":null,
>             "hash":"b640a0ce465fa2a4".   <--- is this group_id?
>          },
> ```


Based on the demo video linked by @wmak and the attached screenshots of the demo, sample events are shown for spans belonging to a specific group. Which implies that most queries would have this in the filter clause. Hence I have added group in the sorting key. A side effect of this would be that building the trace view using this table would be harder since spans belonging to a single transaction would be across different groups. Which would imply that building the transaction view or trace view using nodestore would most likely be much faster. That is the tradeoff which we would have to make today. cc: @fpacifici, @evanh 
Screenshot 2023-04-27 at 10 45 35 PM

Screenshot 2023-04-27 at 10 39 51 PM
Screenshot 2023-04-27 at 10 42 06 PM
Screenshot 2023-04-27 at 10 42 57 PM

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    Column("start_timestamp", DateTime()),
+    Column("end_timestamp", DateTime(Modifiers(codecs=["DoubleDelta"]))),
+    Column("duration", UInt(32)),
+    Column("exclusive_time", Float(64)),
+    Column("op", String(Modifiers(low_cardinality=True))),
+    Column("group", UInt(64)),
+    Column("span_status", UInt(8)),
+    Column("span_kind", String(Modifiers(low_cardinality=True))),
+    Column("description", String()),
+    Column("status", UInt(32, Modifiers(nullable=True))),
```
gggritso, line 35: I think this column would work better as a low cardinality string. In the metrics dataset, all tags are strings, and it'd be good if the data types matched.

nikhars: That is not how you should think about metrics and raw data. Metrics are pre-aggregated data so the types have to generic to support all use cases. With raw data, you need to answer questions based on the data model. The data model has knowledge of the system. For example, if you run a query against status as string versus int, your queries on raw data would be an oder of magnitude slower. I would recommend you do not try to match types between metrics and spans table.

gggritso: Okay, understood!

dbanda (approved):

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    Column("segment_name", String(Modifiers(default="''"))),
+    Column("start_timestamp", DateTime()),
+    Column("end_timestamp", DateTime(Modifiers(codecs=["DoubleDelta"]))),
+    Column("duration", UInt(32)),
+    Column("exclusive_time", Float(64)),
+    Column("op", String(Modifiers(low_cardinality=True))),
+    Column("group", UInt(64)),
+    Column("span_status", UInt(8)),
+    Column("span_kind", String(Modifiers(low_cardinality=True))),
+    Column("description", String()),
```
fpacifici, line 34: I think we should rename this to normalized_description as this is supposed to be normalized.

snuba/snuba_migrations/spans/0001_spans_v1.py
```diff
@@ -0,0 +1,99 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, group, end_timestamp, cityHash64(span_id))",
```
fpacifici, line 65: Does the `group` have a 1:1 relationship with a segment_name or with segment_id ? 
I think this was discussed in the draft, but I cannot find that PR anymore. Any change we can sort spans by transaction (or segment_id) within a time range? 

```
..... toStartOfHour(end_timestamp), segment_id, cityHash64(span_id)
```

This is meant to pack together the spans for a segment_id thus achieving a much better compression as data will be more homogeneous on very high entropy columns like transaction id and segment id

snuba/snuba_migrations/spans/0001_spans_v1.py
nikhars: I have kept only a single `description` field for now assuming that it would be the normalized description. If we need to store the un-normalized ones, we can create a new column called `raw_description`.

snuba/snuba_migrations/spans/0001_spans_v1.py
nikhars: The `group` does not bear any relationship with a segment_name or with a `segment_id`. `group` is like a hash of some common elements of a span (maybe like module, op, etc.). Different transactions (with different transaction_id but same transaction_name) would have the same common elements of the span. But it is also possible that the same common elements of the span can belong to completely different transactions. For example, a db call to get user count from a table can be triggered from different workflows and hence different transaction names. 

With `group` being the primary entity by which individual samples are found, the spans belonging to an individual segment_id are bound to be spread across in disk. My first proposal was aligned to have all spans belonging to a segment be close to each other. But the product workflow (see https://github.com/getsentry/snuba/pull/4089#issuecomment-1527056783) points towards use cases which led me to make group be so high in the sorting order.

fpacifici: Ok, we will have to watch out very carefully for disk usage then as, spreading all those segment/transactions unique ids will hit us on compression efficiency quite hard.

PR #4096, opened 2023-04-26
feat(geo): Add missing mappings and tests

udameli:
It looks like I missed geo.subdivision column mapping in a couple of places in https://github.com/getsentry/snuba/pull/4080. Also added geo.subdivision to tests 

This should help fix CI errors in https://github.com/getsentry/sentry/pull/47995


snuba/datasets/configuration/discover/entities/discover.yaml
```diff
@@ -28,6 +28,7 @@ schema:
     },
     { name: geo_region, type: String, args: { schema_modifiers: [nullable] } },
     { name: geo_city, type: String, args: { schema_modifiers: [nullable] } },
+    { name: geo_subdivision, type: String, args: { schema_modifiers: [nullable] } },
```
volokluev, line 31: why is this column necessary if you have a mapper. Did something fail because you didn't have this?

tests/datasets/entities/storage_selectors/test_errors.py
```diff
@@ -46,6 +46,9 @@
             None, "geo_region", None, "contexts", "geo.region", nullable=True
         ),
         ColumnToMapping(None, "geo_city", None, "contexts", "geo.city", nullable=True),
+        ColumnToMapping(
```
volokluev, line 49: hmm, you shouldn't have to do this, let me see if I can make this change unnecessary

tests/datasets/test_errors_processor.py
```diff
@@ -311,6 +312,7 @@ def build_result(self, meta: KafkaMessageMetadata) -> Mapping[str, Any]:
                 "geo.country_code",
                 "geo.region",
                 "geo.city",
+                "geo.subdivision",
```
volokluev, line 315: you added this field to the processor test but are there tests querying for them?

volokluev (commented):
just some questions

snuba/datasets/configuration/discover/entities/discover.yaml
udameli: i'm mostly replicating what we do for the rest of geo information tbh

udameli: I'm just trying to fix the errors I have here https://github.com/getsentry/sentry/pull/47995

udameli: https://github.com/getsentry/sentry/actions/runs/4813125117/jobs/8569256186?pr=47995

udameli: also when I tried querying for geo.subdivision it didn't return any data although there were events that had it set when I look at their json

udameli: so i'm wondering if i missed something in my last pr

volokluev: I see, I don't think this column addition is necessary and the way to check for sure would be to write an api test like [here](https://github.com/getsentry/snuba/blob/master/tests/test_snql_api.py#L55) which should give you the confidence that all the columns are where they should be.

Also side question, are these columns only going to be queried through discover?

tests/datasets/plans/test_storage_plan_builder.py
```diff
@@ -48,6 +48,9 @@
             None, "geo_region", None, "contexts", "geo.region", nullable=True
         ),
         ColumnToMapping(None, "geo_city", None, "contexts", "geo.city", nullable=True),
+        ColumnToMapping(
```
volokluev, line 51: not necessary

lynnagara:
Would you mind updating the schema as well? https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/events.v1.schema.json

volokluev (approved):

udameli:
> Would you mind updating the schema as well? https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/events.v1.schema.json

https://github.com/getsentry/sentry-kafka-schemas/pull/115 updated!

PR #4098, opened 2023-04-26
feat(capman): enforce allocation poliy by default

volokluev:
Allocation policies are turned on in production now so let's turn them on by default in CI and dev


evanh (approved):
Approved, pending CI fixes.

dbanda (approved):

PR #4099, opened 2023-04-27
bump sentry-kafka-schemas to 0.0.33

udameli:


lynnagara (approved):

udameli:
for some reason it doesn't see the latest release https://github.com/getsentry/publish/actions/runs/4822649488

untitaker:
> for some reason it doesn't see the latest release

this stuff does happen. I am not sure why, but it takes quite a while for the release to propagate to github.com/getsentry/pypi. however, it should not take an entire hour. But rerunning probably fixes it

PR #4100, opened 2023-04-27
fix: Replace TagsExpanderProcessor with column mapper

evanh:
This PR is a required step to being able to turn on the entity column validator.
The column validator fails on the magic columns (`tags_key` and `tags_value`)
that this processor changes. This would mean having to expose these columns in
the entity schema. That is a bad workflow, if every time a processor gets added
to an Entity, any relevant columns have to also be added to the schema.

Instead, expose the magic columns as ColumnMappers on the storage, which the
entity column validator can use to check incoming columns (this will be done in
a separate PR). Then this entire processor becomes unnecessary.

This also adds a new type of ColumnMapper that works with the YAML format, since
that format doesn't support complex expressions.


tests/query/processors/test_arrayjoin_optimizer.py
```diff
@@ -429,19 +431,24 @@ def parse_and_process(snql_query: str) -> ClickhouseQuery:
...
+    from_clause = query.get_from_clause()
+    assert isinstance(from_clause, QueryEntity)
+    entity = get_entity(from_clause.key)
     storage = entity.get_writable_storage()
     assert storage is not None
+    assert isinstance(query, LogicalQuery)
     for p in entity.get_query_processors():
         p.process_query(query, request.query_settings)
 
-    ArrayJoinKeyValueOptimizer("tags").process_query(query, request.query_settings)
```
evanh, line 438: When I made the change, this test started failing.

The `ArrayJoinKeyValueOptimizer` is a storage level processor, but in this test it was being called before the query even knew what storage it was using. This only worked because the `TagsExpanderProcessor` was an entity processor.  Moving this line below the plan builder (where it would actually be called) fixed the tests.

volokluev (approved):

PR #4102, opened 2023-04-28
feat(spans): Spans dataset definition and consumer processing

nikhars:
This PR depends on https://github.com/getsentry/snuba/pull/4089 to be merged first.

Added the dataset definition to access the spans entity. Also defined the spans processor necessary to populate data from the transactions topic to the spans table.

There are a bunch of hacks in the code right now. So there might be bugs in the code. But in order to get spans ingestion going some of these hacks are needed.


snuba/datasets/configuration/spans/entities/spans.yaml
```diff
@@ -0,0 +1,146 @@
...
+  ]
+
+storage_selector:
+  {
+    selector: DefaultQueryStorageSelector
+  }
+
+query_processors:
+  [
+    { processor: TagsExpanderProcessor },
```
evanh, line 116: You can remove this, I have a PR that deleted this processor.

snuba/datasets/processors/spans_processor.py
```diff
@@ -0,0 +1,484 @@
...
+        # The expected format is [project,project,...]
+        project_allowlist = project_allowlist[1:-1]
+        if project_allowlist:
+            rolled_out_projects = [int(p.strip()) for p in project_allowlist.split(",")]
+            if project_id in rolled_out_projects:
+                return True
+    return False
+
+
+def parse_query(query, is_savepoint):
```
evanh, line 53: This needs to be typed.

evanh: Also I do not understand what this function is doing.

snuba/datasets/processors/spans_processor.py
```diff
@@ -0,0 +1,484 @@
...
+        if op != "db.redis" and op.startswith("db"):
+            parse_state = None
+            table = []
+
+            # This is wrong for so many reasons lol, but just something simple to pull some values out
+            raw_parsed = sqlparse.parse(description)[0]
+            processed_span["description"] = parse_query(
+                raw_parsed, "SAVEPOINT" in description.upper()
+            )
+            parsed = sqlparse.parse(processed_span["description"])[0]
```
evanh, line 282: Why do we parse the query 3 times?

evanh (commented):
I think you should unit test your `parse_query` function.

Also, I would add a `test_spans_api.py` file that actually writes a SnQL query, sends it to Snuba, and gets data back to ensure this is all working as expected.

snuba/datasets/processors/spans_processor.py
```diff
@@ -0,0 +1,484 @@
...
+        data = event_dict["data"]
+        processed_spans: MutableSequence[MutableMapping[str, Any]] = []
+        for span in data.get("spans", []):
+            processed_span = self._process_span(span, common_span_fields)
+            if processed_span is not None:
+                processed_spans.append(processed_span)
+
+        return processed_spans
+
+    def process_message(
```
evanh, line 441: Since this is an experiment and the incoming data will probably change format a few times, I might wrap more of this in Exception handling code so the consumer doesn't die every time we get a bad message.

snuba/datasets/processors/spans_processor.py
nikhars: See https://github.com/getsentry/snuba/pull/4102#discussion_r1180738350. I don't want to spend time optimizing this code right now. I will ask the product team to plug in whatever additional business logic they want to add in here in the future.

snuba/datasets/processors/spans_processor.py
nikhars: That's a good point. You are correct that this has potential to blow up. We do not want to care about this blowing up just yet.

nikhars:
> I think you should unit test your `parse_query` function.
> 
> Also, I would add a `test_spans_api.py` file that actually writes a SnQL query, sends it to Snuba, and gets data back to ensure this is all working as expected.

I will add tests to `test_spans_api.py`. For the `parse_query`, I would just let it be as is. This is temporary code until we get sanitized data from upstream.

snuba/datasets/processors/spans_processor.py
nikhars: The processing code has been wrapped in a try except block [here](https://github.com/getsentry/snuba/pull/4102/files#diff-c76e01559784d4d35d2a011e4430e38f324f617fc24180607d8aec538241238cR470-R491)

nikhars:
> > I think you should unit test your `parse_query` function.
> > Also, I would add a `test_spans_api.py` file that actually writes a SnQL query, sends it to Snuba, and gets data back to ensure this is all working as expected.
> 
> I will add tests to `test_spans_api.py`. For the `parse_query`, I would just let it be as is. This is temporary code until we get sanitized data from upstream.

Added a few api tests in `test_spans_api.py`

requirements.txt
```diff
@@ -38,3 +38,4 @@ urllib3==1.26.12
 pyuwsgi==2.0.20
 Werkzeug==2.2.2
 PyYAML==6.0
+sqlparse==0.4.2
```
nikhars, line 41: Temporarily needed for application logic of product in order to classify spans. Should go away once the business logic moves to relay.

snuba/datasets/processors/spans_processor.py
```diff
@@ -0,0 +1,493 @@
...
+
+UNKNOWN_SPAN_STATUS = 2
+
+EventDict = MutableMapping[str, Any]
+SpanDict = MutableMapping[str, Any]
+CommonSpanDict = MutableMapping[str, Any]
+RetentionDays = int
+
+
+def is_project_in_allowlist(project_id: int) -> bool:
```
nikhars, line 42: Config option to allow spans from a fixed set of projects to be written to the table

snuba/datasets/processors/spans_processor.py
```diff
@@ -0,0 +1,493 @@
...
+                )
+            except Exception:
+                logger.error(
+                    "Invalid measurements field.",
+                    extra={"measurements": measurements},
+                    exc_info=True,
+                )
+
+    @staticmethod
+    def _process_module_details(
```
nikhars, line 247: The top level span in the trace context does not have enough details to fill these in. Setting them to defaults for now.

evanh (approved):

snuba/datasets/processors/spans_processor.py
```diff
@@ -0,0 +1,493 @@
...
+        TODO: For the top level span belonging to a transaction, we do not know how to fill these
+              values yet. For now lets just set them to their default values.
+        """
+        processed["module"] = ""
+        processed["action"] = ""
+        processed["domain"] = ""
+        processed["status"] = 0
+        processed["span_kind"] = ""
+
+    def _process_child_span_module_details(
```
nikhars, line 261: All of this code has been borrowed from https://github.com/getsentry/Starfish-sample-generator/blob/main/load-it-up.py#L40. This is application logic of how to categorize spans.

tests/test_spans_api.py
```diff
@@ -0,0 +1,306 @@
...
+                WHERE project_id = 2
+                AND end_timestamp >= toDateTime('{from_date}')
+                AND end_timestamp < toDateTime('{to_date}')
+            """
+        )
+        data = json.loads(response.data)
+        assert response.status_code == 200, response.data
+        assert data["data"][0]["aggregate"] == 0, data
+
+    def test_get_group_sorted_by_exclusive_time(self) -> None:
```
nikhars, line 267: This is the query which has been provided by product in https://github.com/getsentry/snuba/pull/4089#issuecomment-1526118481

tests/test_spans_api.py
```diff
@@ -0,0 +1,306 @@
...
+                AND end_timestamp >= toDateTime('{from_date}')
+                AND end_timestamp < toDateTime('{to_date}')
+                ORDER BY exclusive_time DESC
+            """
+        )
+        data = json.loads(response.data)
+        assert response.status_code == 200, response.data
+        assert len(data["data"]) > 2, data
+
+    def test_get_all_spans_belonging_to_trace_id(self) -> None:
```
nikhars, line 288: Example query on how to get all spans belonging to a trace.

nikhars (commented):
Leaving some comments to help with review.

PR #4103, opened 2023-04-28
feat(DRS): Rollout readiness states migration groups for [part 1]

enochtangg:
### Overview
This PR is responsible for enabling readiness_states for the following migration groups: `querylog`, `search_issues`, and `test_migration`. This [section of this document](https://www.notion.so/sentry/Dataset-Readiness-State-Rollout-Plan-d38deffe77534856a4e4f4a3b0cfb110?pvs=4#479d1451ca8a457e9542fedce8cdf87c) is used to track the roll out.

Additionally, this PR bring us closer to migration parity between sentry environments (s4s and saas). To do this, the following migrations groups were removed from `SKIPPED_MIRGATIONS_GROUPS` which enables them to run locally: `querylog`, `test_migration`, and `search_issues`. This was done in order for them to work properly with readiness_states because there is no readiness_state which skips local, but enabled in prod (which most of these migrations groups do).

### What will happen after this is deployed?
* The mentioned migration groups will now use its `readiness_state` flag to determine whether or not the migration group is skipped in particular sentry environments. 
* The migration groups: querylog, and search_issues will not be skipped in S4S. Even if these tables are not actively used, the goal is to have these schemas exist in S4S in order to be up to date with SaaS.

### Blast Radius
The selected migrations groups are all `partial` or `limited`. This means that this will only affect the local, saas, and s4s sentry environments.

### Blocker
The storage sets for these migration groups needs to be added to s4s before merging this. https://github.com/getsentry/ops/pull/6751


snuba/settings/__init__.py
```diff
@@ -241,13 +241,7 @@ class RedisClusters(TypedDict):
...
 # The migration groups that can be skipped are listed in OPTIONAL_GROUPS.
 # Migrations for skipped groups will not be run.
-SKIPPED_MIGRATION_GROUPS: Set[str] = {
-    "querylog",
-    "profiles",
-    "functions",
-    "test_migration",
-    "search_issues",
-}
+SKIPPED_MIGRATION_GROUPS: Set[str] = set()
```
volokluev, line 244: Should we just remove this?

volokluev (approved):
hell yeah brother

snuba/settings/__init__.py
enochtangg: Updated!

enochtangg: Rebased master, we just added `spans` migration group which will still use SKIPPED_MIGRATIONS_GROUPS for now.

volokluev: why not just set its readiness state since it's new?

enochtangg: Unfortunately, readiness states do not capture the current roll out process for `spans` which is a unique case.

enochtangg:
Closed in favour of https://github.com/getsentry/snuba/pull/4225.

PR #4104, opened 2023-04-28
fix(inc): use errors readonly storage by default

volokluev:
errors_ro is used in production by default, thus we should use it by default in all our code unless we need to do otherwise


nikhars (approved):

nikhars (changes requested):
Actually we should check whether that is true in single tenant environments or not

PR #4106, opened 2023-04-28
fix: Replace TagsExpanderProcessor with column mappers

evanh:
This PR is a required step to being able to turn on the entity column validator.
The column validator fails on the magic columns (tags_key and tags_value)
that this processor changes. This would mean having to expose these columns in
the entity schema. That is a bad workflow, if every time a processor gets added
to an Entity, any relevant columns have to also be added to the schema.

Instead, expose the magic columns as ColumnMappers on the storage, which the
entity column validator can use to check incoming columns (this will be done in
a separate PR). Then this entire processor becomes unnecessary.

This also adds a new type of ColumnMapper that works with the YAML format, since
that format doesn't support complex expressions.

Finally, this adds some tests that will run on the errors_ro storage as well, to
validate that the mapping is working correctly on the _ro version of the storage


evanh:
> Is this being tested any differently?

Yes, see the additional test classes in `test_snql_api.py`, `test_api.py` and `test_discover_api.py`. I also added a new test directly for the failure case that caused this to be reverted in `test_snql_api.py`.

tests/test_api.py
```diff
@@ -2372,3 +2373,16 @@ def test_invalid_dataset_and_entity_combination(self) -> None:
                 "type": "subscription",
             }
         }
+
+
+@pytest.mark.clickhouse_db
```
volokluev, line 2378: clever!

volokluev (approved):

PR #4111, opened 2023-05-01
feat(ci): allow skip check migrations

dbanda:
skip check migrations if the skip-migrations-check label is used and add a note so that gocd is aware


evanh (approved):
Please make sure that why this flag is needed and how to use it is documented in our runbooks.

dbanda:
> skip-check-migrations



> Please make sure that why this flag is needed and how to use it is documented in our runbooks.

Added it to the runbooks.

PR #4122, opened 2023-05-01
fix: Move topic compression into topic definition not producer config

lynnagara:
This is consistent with Sentry prod where the compression type is always defined on the topic not in the producer config


volokluev (approved):

bmckerry:
Sorry, please don't merge yet, I still haven't applied `compression.type=lz4` everywhere
Will let you know when

lynnagara:
After thinking about this some more, I think we should keep the compression on the producer as well. Closing this PR.

PR #4125, opened 2023-05-02
ref: Bump sentry-kafka-schemas to 0.1.3

untitaker:
Use codecs from sentry-kafka-schemas instead of Arroyo.

This change would allow us to transparently handle decoding of msgpack
topics, solely based off of configuration in sentry-kafka-schemas. Snuba
doesn't have to know about the underlying encoding format anymore. We
don't want to do that, but I think it's pretty cool.

Unfortunately, when there's no schema, we get no codec back from
sentry-kafka-schemas. So we have to define our own noop subclass.
Eventually we should be able to get definitions for all topics. We might
already have them, actually.

### DIFF
--- a/snuba/consumers/schemas.py
+++ b/snuba/consumers/schemas.py
@@ -26,7 +26,7 @@ def _get_codec_impl(topic: Topic) -> Optional[Codec[Any]]:
         return None
 
 
-_NOOP_CODEC = JsonCodec(json_schema=None)
+_NOOP_CODEC: Codec[Any] = JsonCodec(json_schema=None)
 _cache: MutableMapping[Topic, Codec[Any]] = {}
 
 
--- a/snuba/consumers/schemas.py
+++ b/snuba/consumers/schemas.py
@@ -4,7 +4,7 @@ from typing import Any, MutableMapping, Optional
 import sentry_kafka_schemas
 import sentry_sdk
 from sentry_kafka_schemas.codecs import Codec
-from sentry_kafka_schemas.codecs.json import JsonCodec, T
+from sentry_kafka_schemas.codecs.json import JsonCodec
 
 from snuba.utils.streams.topics import Topic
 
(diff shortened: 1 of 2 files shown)


snuba/consumers/schemas.py
```diff
@@ -25,11 +26,20 @@ def get_schema(topic: Topic) -> Optional[Mapping[str, Any]]:
...
+class NoopCodec(JsonCodec[T]):
+    def __init__(self):
+        pass
 
+    def validate(self, data: T) -> None:
+        pass
 
-def get_json_codec(topic: Topic) -> JsonCodec[Any]:
+
+_NOOP_CODEC = NoopCodec()
```
lynnagara, line 37: Do we really need the user to define this NoopCodec? Isn't that the default behavior for JsonCodec if a schema isn't provided?

untitaker: The schema argument in codecs is now non-nullable. I thought it made sense to change that. We can however revert it. I'm just not sure if we want to keep the consumer entirely clueless about whether there is a schema at all

untitaker: I guess my question is, what is the default return value for get-codec if there is no schema? is it specifically JsonCodec?

lynnagara: Makes sense. We would need to put both a JsonCodec and MsgPackCodec in there. Not sure if it is worth it.

lynnagara: What about if we bring over this from the arroyo implementation. So we don't need to define the noop codec in many places: https://github.com/getsentry/sentry-kafka-schemas/pull/123

untitaker: merged and released

lynnagara: thanks

lynnagara (approved):

lynnagara:
Merging this to get some of those ValidationErrors cleaned up :)

lynnagara:
oh wait... it needs 0.1.4

PR #4129, opened 2023-05-03
feat: Turn on DLQ by default

lynnagara:
DLQ rebalancing bug was fixed. Turn on producing to the DLQ by default now.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -546,7 +546,7 @@ def process_message(
             scope.set_tag("invalid_message", "true")
             logger.warning(err, exc_info=True)
             value = message.value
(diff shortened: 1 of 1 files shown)


untitaker (commented):
Can we actually DLQ things by default if we don't have a way to replay items?

Also ValidationErrors are still only warnings, right?

lynnagara:
> Can we actually DLQ things by default if we don't have a way to replay items?
> 
> Also ValidationErrors are still only warnings, right?

Yes - we've been doing that for years with the old DLQ implementation. This was only changed last week (via this config) because of the DLQ rebalancing bug

untitaker (approved):
i assume we are not hitting the DLQ case too often, otherwise I'd be concerned about the performance by the new DLQ. old DLQ probably performed better than the new one right now

PR #4135, opened 2023-05-04
feat(spans): Change required time column to be timestamp

nikhars:
Changes the time required column from `end_timestamp` to `timestamp`. There is already a column mapper defined which maps timestamp to end_timestamp [here](https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/spans/entities/spans.yaml#L72-L77). Also adds the timeseries processor to spans entity


volokluev (approved):
wfm

PR #4137, opened 2023-05-04
fix(replays): use uuid with dashes for event hash

JoshFerge:
- on clickhouse 20.3, uuids need to have dashes on them. We didn't have this covered in a test case but we now see it occuring in a few different places with new behavior we're trying to add.


volokluev:
should you have a test case for it now?

JoshFerge:
> should you have a test case for it now?

It will be covered in our sentry tests, if that works.

volokluev (approved):

PR #4138, opened 2023-05-04
chore(capman): add user reports to single thread referrers

volokluev:
There is a user report task which calls `eventstore.get_events` and it's not very important (because it has been rejected for a straight week and nobody has said anything) but probably shouldn't leave it broken. We throttle it to one thread and call it a day.

### DIFF
--- a/snuba/query/allocation_policies/bytes_scanned_window_policy.py
+++ b/snuba/query/allocation_policies/bytes_scanned_window_policy.py
@@ -57,6 +57,7 @@ _SINGLE_THREAD_REFERRERS = set(
         "delete-events-by-tag-value",
         "delete.fetch_last_group",
         "forward-events",
+        "tasks.update_user_reports",
     ]
 )
 


snuba/query/allocation_policies/bytes_scanned_window_policy.py
```diff
@@ -57,6 +57,7 @@
         "delete-events-by-tag-value",
         "delete.fetch_last_group",
         "forward-events",
+        "tasks.update_user_reports",
```
nikhars, line 60: Do you envision the allocation policy being enabled in environments like single tenant? Or it would be used only on the SaaS farms?

The reason I ask is that if it would be used in Single Tenant as well, then it might be better to have this configured in the settings file versus hardcoding them.

volokluev: I think that's a pretty easy change to make later. Currently the priority is to just let everything through that needs to be let through. It's also nice to have all of the information related to the allocation policy in one place.

Also this doesn't change per-environment and I don't think it should. If it does we can definitely put it there.

nikhars (approved):

PR #4141, opened 2023-05-05
fix(optimize): Remove cutoff for parally optimize

nikhars:
With bigger clusters, we have capacity to run 2 merges in parallel for longer. Change the parallel cutoff time to be the same as regular optimize cutoff time.


volokluev (approved):
let's do it

PR #4158, opened 2023-05-10
fix(clickhouse): Add a check for the max supported clickhouse version

davidtsuk:
Closes SNS-1539

Add a check to ensure the clickhouse version is at most the max supported version.


snuba/migrations/clickhouse.py
```diff
@@ -1 +1,2 @@
 CLICKHOUSE_SERVER_MIN_VERSION = "20.3.5.21"
+CLICKHOUSE_SERVER_MAX_VERSION = "21.8"
```
davidtsuk, line 2: Not sure if this needs to be 21.8.0.0 or if this works.

nikhars (changes requested):
I would advise against blocking migrations altogether if the clickhouse version is higher than 21.8. Sentry and Snuba are both open source projects. While internally all our clickhouse clusters have a maximum version of 21.8, we do not know which version of clickhouse self hosted customers might be using. They could be on versions higher than 21.8. In those cases, migrations would start failing for them abruptly.

We should document what clickhouse versions we know work (happening in this [PR](https://github.com/getsentry/snuba/pull/4155) and leave it as is. Now it is upto self hosted customers to decide what to do with this information.

davidtsuk:
@nikhars Instead of blocking the migration, would it be better to just log a warning if they are on versions higher than 21.8?

nikhars:
> @nikhars Instead of blocking the migration, would it be better to just log a warning if they are on versions higher than 21.8?

Yes, logging a warning would be ok.

snuba/migrations/connect.py
```diff
@@ -66,6 +72,11 @@ def check_clickhouse(clickhouse: ClickhousePool) -> None:
             f"Snuba requires minimum Clickhouse version {CLICKHOUSE_SERVER_MIN_VERSION} ({clickhouse.host}:{clickhouse.port} - {ver})"
         )
 
+    if version.parse(ver.group()) > version.parse(CLICKHOUSE_SERVER_MAX_VERSION):
+        logger.warning(
+            "Snuba has only been tested on Clickhouse versions up to {CLICKHOUSE_SERVER_MAX_VERSION} ({clickhouse.host}:{clickhouse.port} - {ver}). Higher versions might not be supported."
```
rahul-kumar-saini, line 77:
```suggestion
            f"Snuba has only been tested on Clickhouse versions up to {CLICKHOUSE_SERVER_MAX_VERSION} ({clickhouse.host}:{clickhouse.port} - {ver}). Higher versions might not be supported."
```

rahul-kumar-saini (approved):
Looks good pending @nikhars review

nikhars:
Were you able to test this in any way? If not, I would recommend that you test it out. You can do that by installing a clickhouse docker version locally by pulling it from https://hub.docker.com/r/clickhouse/clickhouse-server/ and testing against it

davidtsuk:
> Were you able to test this in any way? If not, I would recommend that you test it out. You can do that by installing a clickhouse docker version locally by pulling it from https://hub.docker.com/r/clickhouse/clickhouse-server/ and testing against it

```
snuba migrations list
2023-05-19 11:24:28,421 Initializing Snuba...
2023-05-19 11:24:29,673 Snuba initialization took 1.252941208s
2023-05-19 11:24:29,681 Snuba has only been tested on Clickhouse versions up to 21.8 (127.0.0.1:9000 - 23.4.2.11). Higher versions might not be supported.
system (readiness_state: complete)
[ ]  0001_migrations
...
```

nikhars (approved):

PR #4160, opened 2023-05-11
fix(consumers): Add timeout for http batch writes

nikhars:
When a join is being called on HTTPWriteBatch we should limit the time spent waiting for the operation to be successful. If we do not limit the time spent the consumer can stay hung for any amount of time. It would stay hung till the kafka broker realizes that the consumer is dead (which today is 5 minutes). Instead of waiting so long, we should fail fast and abort earlier.

### DIFF
--- a/snuba/clickhouse/http.py
+++ b/snuba/clickhouse/http.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 import re
-import socket
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
 from queue import Queue, SimpleQueue
@@ -243,9 +242,6 @@ class HTTPWriteBatch:
                 "http_batch.timeout",
                 tags={"table": str(self.__statement.get_qualified_table())},
             )
-            # Container names are available as the hostnames within the container. Let's use that
-            # when logging to Sentry.
-            sentry_sdk.set_tag("hostname", socket.gethostname())
             raise
 
         logger.debug("Received response for %r.", self)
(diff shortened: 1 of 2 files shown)


untitaker (approved):

snuba/clickhouse/http.py
```diff
@@ -235,7 +236,18 @@ def close(self) -> None:
...
+        try:
+            response = self._result.result(timeout)
+        except TimeoutError:
+            self.__metrics.increment(
+                "http_batch.timeout",
+                tags={"table": str(self.__statement.get_qualified_table())},
+            )
+            # Container names are available as the hostnames within the container. Let's use that
+            # when logging to Sentry.
+            sentry_sdk.set_tag("hostname", socket.gethostname())
```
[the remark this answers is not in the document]
nikhars: You are right. Removed the code. Do not want duplicate tags.

snuba/clickhouse/http.py
```diff
@@ -325,4 +341,6 @@ def write(self, values: Iterable[bytes]) -> None:
             batch.append(value)
 
         batch.close()
-        batch.join()
+        # IMPORTANT: Please read the docstring of this method if you ever decide to remove the
+        # timeout argument from the join method.
+        batch.join(timeout=10)
```
fpacifici, line 346: Can we make this configurable ? I feel we will eventually shoot ourselves in the feet without the ability to tweak it without merging a snuba change. 
It happened once with replays (the thread wait timeout on the parallel collector was hardcoded and broke during a weekend).

nikhars: Made it configurable with the default timeout of 10 seconds.

PR #4174, opened 2023-05-15
ref(dashboard): Add a setting that restricts the tools visible in the dashboard

evanh:
This is a precursor to having a separate dashboard app for product developers.
In the future we can run a version of the dashboard with this developer mode on
so that product developers only have access to the tools we want them to see.


snuba/admin/tool_policies.py
```diff
@@ -60,6 +64,15 @@ def check_tool_perms(
     def decorator(f: Callable[..., Response]) -> Callable[..., Response]:
         @wraps(f)
         def check_perms(*args: Any, **kwargs: Any) -> Response:
+            if settings.ADMIN_DEVELOPER_MODE:
+                for tool in tools:
+                    if tool in DEVELOPER_TOOLS:
```
enochtangg, line 69: Looks good, only question I have is what happens if an admin endpoint requires two [different roles](https://github.com/getsentry/snuba/blob/76420cf365a7429333fd31f1a5f730e462c2494b/snuba/admin/views.py#L675). Should we check all required tools exist in `DEVELOPER_TOOLS` set?

enochtangg (commented):
Nice, just have one question.

snuba/admin/tool_policies.py
evanh: No, that filter is an "OR" filter. So people with access to the system queries tool OR the tracing tool can access that endpoint.

enochtangg (approved):

PR #4176, opened 2023-05-15
fix(CapMan): Validate Tenant IDs

rahul-kumar-saini:
### Overview
- Queries with bad Tenant IDs are showing up as errors in Sentry
- We should start rejecting these queries
- Modified existing validation to check for `None`s


volokluev (changes requested):
This will break many things. There is a validation function on the bytes scanned policy already. See how it works:

https://github.com/getsentry/snuba/blob/master/snuba/query/allocation_policies/bytes_scanned_window_policy.py#L122-L133

volokluev (approved):

volokluev (approved):
approved but looks like you'll have to fix some sentry tests

PR #4177, opened 2023-05-15
fix: All consumers run with --no-strict-offset-reset in dev

lynnagara:
This was missing in some of the newer consumers


rahul-kumar-saini:
How is that flag used in the CLI?

As far as I can see it's not in the consumer CLI [args](https://github.com/getsentry/snuba/blob/a2a40b395608d8d7630198d54177bc9043f2feb1/snuba/cli/consumer.py#L126-L150) or used anywhere in the function itself, is it propagated forward to the consumer in another way?

untitaker:
it's here: https://github.com/getsentry/snuba/blob/a2a40b395608d8d7630198d54177bc9043f2feb1/snuba/cli/consumer.py#L88

rahul-kumar-saini (approved):

PR #4186, opened 2023-05-16
fix(tech-debt): Fix Column Validator Bug

enochtangg:
### Overview
This PR is responsible for fixing the EntityContainsColumnsValidator where a selected nested column (e.g. `tags`) would return missing in the validator when it is present.

There is a bug in populating the `_lookup` and `_nested` dictionaries in `ColumnSet`. The `_lookup` only contains the flattened columns (e.g. tags.keys, tags.values). As a result, `tags` does not exist. To fix this, this PR separates one lookup for base name columns and one for all flattened columns.


volokluev (commented):
test?

enochtangg:
> test?

Added, thanks.

rahul-kumar-saini (approved):

PR #4189, opened 2023-05-16
fix(parsing): Update error message for missing datetime column comparison condition

davidtsuk:
Currently, if a query is sent in which a required datetime column is compared to a string literal instead of a datetime literal, the error message that is returned is misleading. This new error message should make it more clear that the datetime column needs to be compared with a datetime literal.

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -1297,11 +1297,11 @@ def _align_max_days_date_align(
     )
     if not lower:
         raise ParsingException(
-            f"missing >= condition on column {entity.required_time_column} for entity {key.value}"
+            f"missing >= condition with a datetime literal on column {entity.required_time_column} for entity {key.value}"
         )
     elif not upper:
         raise ParsingException(
-            f"missing < condition on column {entity.required_time_column} for entity {key.value}"
+            f"missing < condition with a datetime literal on column {entity.required_time_column} for entity {key.value}"
         )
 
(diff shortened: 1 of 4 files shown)


snuba/query/snql/parser.py
```diff
@@ -1297,11 +1297,11 @@ def _align_max_days_date_align(
     )
     if not lower:
         raise ParsingException(
-            f"missing >= condition on column {entity.required_time_column} for entity {key.value}"
+            f"missing >= condition with a datetime literal on column {entity.required_time_column} for entity {key.value}"
```
volokluev, line 1300: we could probably put an example here of what a "datetime literal" looks like and make this error message easier to understand
```suggestion
            f"""missing >= condition with a datetime literal on column {entity.required_time_column} for entity {key.value}
            
            Example:
                {entity.required_time_column} >=  toDateTime('{datetime.now().replace(microsecond=0).isoformat()}')
                AND {entity.required_time_column} <  toDateTime('{datetime.now().replace(microsecond=0).isoformat()}')
            """
```

davidtsuk: It seems there are some issues with multi-line error messages, but I agree that an example would make it easier to understand. Maybe, we could condense it into a single line.
image

volokluev: that would work

evanh (approved):

PR #4192, opened 2023-05-16
feat: Snuba consumer takes join_timeout as CLI arg

lynnagara:
If a join timeout is passed, any remaining  unprocessed or partially processed messages are dropped. Since they aren't committed yet, we will pick them up again when the consumer restarts.

This is an experiment to see if passing a join timeout fixes the rebalancing issues we are seeing on the generic metrics consumer.


volokluev (commented):
could we do this with runtime config?

onewland:
> could we do this with runtime config?

A setting might make more sense since it won't take effect without a restart, but I agree with this sentiment (don't specifically have a condition for `generic_metrics_distributions`).

Also I think we'd want it on the raw storage: https://github.com/getsentry/ops/blob/master/k8s/services/snuba/_deployment.generic-metrics-distributions-consumer.yaml#L61

untitaker (approved):
i don't think we _can_ make this work with dynamic config. we should expose it as CLI parameter at most so it can be applied via ops pr

lynnagara:
It doesn't quite work with runtime config since it needs a restart. The approach taken in this PR is just to confirm whether it actually has an impact without involving ops. If it works we will revert this code an implement as a CLI arg.

PR #4196, opened 2023-05-17
feat(clickhouse-settings): Add ability to override clickhouse query settings via query processor

davidtsuk:
This allows custom clickhouse query settings to be specified in the dataset config through a physical query processor.

### DIFF
--- a/snuba/query/processors/physical/clickhouse_settings_override.py
+++ b/snuba/query/processors/physical/clickhouse_settings_override.py
@@ -1,0 +1,17 @@
+from typing import Any, MutableMapping
+
+from snuba.clickhouse.query import Query
+from snuba.query.processors.physical import ClickhouseQueryProcessor
+from snuba.query.query_settings import QuerySettings
+
+
+class ClickhouseSettingsOverride(ClickhouseQueryProcessor):
+    """
+    Overrides arbitrary clickhouse settings via a dictionary specifying the clickhouse settings to override.
+    """
+
+    def __init__(self, settings: MutableMapping[str, Any]) -> None:
+        self.__settings = settings
+
+    def process_query(self, query: Query, query_settings: QuerySettings) -> None:
+        query_settings.set_clickhouse_settings(self.__settings)
--- a/snuba/query/query_settings.py
+++ b/snuba/query/query_settings.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import List, Optional, Sequence
+from typing import Any, List, MutableMapping, Optional, Sequence
 
 from snuba.state.quota import ResourceQuota
 from snuba.state.rate_limit import RateLimitParameters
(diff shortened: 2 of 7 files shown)


snuba/query/query_settings.py
```diff
@@ -165,3 +180,9 @@ def get_resource_quota(self) -> Optional[ResourceQuota]:
 
     def set_resource_quota(self, quota: ResourceQuota) -> None:
         pass
+
+    def get_clickhouse_settings(self) -> MutableMapping[str, Any]:
+        return {}
+
+    def set_clickhouse_settings(self, settings: MutableMapping[str, Any]) -> None:
+        pass
```
volokluev, line 188: why does this not take effect for subscriptions?

snuba/query/processors/physical/clickhouse_settings_override.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any, MutableMapping
```
volokluev, line 1: I know this is a very simple query processor so it doesn't feel like it needs a test but it should have a test (event if it is simple)

davidtsuk: Yeah, I was debating whether this needs a test, but I'll add a test for the sake of it

snuba/query/query_settings.py
davidtsuk: I wasn't sure whether subscriptions needed this feature so I didn't add it, but I'll add it to subscriptions as well since there's no harm in adding it.

snuba/query/processors/physical/clickhouse_settings_override.py
```diff
@@ -0,0 +1,17 @@
...
+class ClickhouseSettingsOverride(ClickhouseQueryProcessor):
+    """
+    Overrides arbitrary clickhouse settings via a dictionary specifying the clickhouse settings to override.
+    """
+
+    def __init__(self, settings: MutableMapping[str, Any]) -> None:
+        self.__settings = settings
+
+    def process_query(self, query: Query, query_settings: QuerySettings) -> None:
+        query_settings.set_clickhouse_settings(self.__settings)
```
nikhars, line 17: Can we restrict the query settings which are modifiable by this processor? Snuba should have control over which settings can be modified rather than product teams.

Also, this should not be made generic. We should put additional safeguards in place like known referrers etc. on which we allow clickhouse setting override. That would allow better control over clickhouse usage and narrowing down bad referrers who might abuse clickhouse with some sort of query settings

lynnagara:
What's the motivation for this? Which query settings do you want to be able to change in query processors?

davidtsuk:
> What's the motivation for this? Which query settings do you want to be able to change in query processors?

Here is an example of settings that have been requested to be changed:
```
Set max_rows_to_group_by to 1,000,000.
Set group_by_overflow_mode to any.
```

volokluev:
@lynnagara the context is that replays wanted to change these query settings on the clickhouse config. In order to let them experiment with their clickhouse settings we're providing this query processor which is easier to revert and modify than clickhouse profile configurations

nikhars (approved):

snuba/query/processors/physical/clickhouse_settings_override.py
volokluev: Nikhar and I spoke offline. This query processor is no different to changing the clickhouse configuration for the user, but easier to manage and revert if need be than an ops repo change. Initially there was some confusion which made it seem like we were accepting arbitrary query settings from the API user (which we are not)

PR #4198, opened 2023-05-17
ref(consumer): pass sentry_received_timestamp again

MeredithAnya:
Same as https://github.com/getsentry/snuba/pull/4052 (which got reverted in https://github.com/getsentry/snuba/commit/0d1b0cfa4540e99ff6689e529961eb6d2fb8a3b4 because sentry didn't for some reason have the previous changes that fixed the tests) Fixed tests were re-added in https://github.com/getsentry/sentry/pull/49348

### DIFF
--- a/snuba/datasets/processors/generic_metrics_processor.py
+++ b/snuba/datasets/processors/generic_metrics_processor.py
@@ -141,7 +141,13 @@ class GenericMetricsBucketProcessor(DatasetMessageProcessor, ABC):
                 GRANULARITY_ONE_DAY,
             ],
         }
-        return InsertBatch([processed], None)
+        return InsertBatch(
+            [processed],
+            None,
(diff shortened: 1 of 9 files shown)


snuba/datasets/processors/generic_metrics_processor.py
```diff
@@ -141,7 +141,13 @@ def process_message(
                 GRANULARITY_ONE_DAY,
             ],
         }
-        return InsertBatch([processed], None)
+        return InsertBatch(
+            [processed],
+            None,
+            sentry_received_timestamp=datetime.utcfromtimestamp(
+                message["sentry_received_timestamp"]
```
untitaker, line 148: I would suggest to make it optional for now to minimize risk

MeredithAnya: ok sure, can do

untitaker (approved):

PR #4201, opened 2023-05-17
feat(capman): Allow overriding policy defaults in configuration

volokluev:
### Context

Currently, a policy has configured defaults and whenever it is reused, those defaults are used as well. That runs into a problem when the policy is used across two storages with different needs (for example, the `org_bytes_scanned_limit` default is different for errors vs transactions.

### This PR:

* Allows for the policy defaults to be overridden at instantiation
* inputs the existing defaults we have in production into configuration
* adds validation to catch mistyped defaults on AllocationPolicyConfig(s)

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -315,7 +315,7 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
         self,
         storage_key: StorageKey,
         required_tenant_types: list[str],
-        default_config_overrides: dict[str, Any] = {},
+        default_config_overrides: dict[str, Any],
         **kwargs: str,
     ) -> None:
         self._required_tenant_types = set(required_tenant_types)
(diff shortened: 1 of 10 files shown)


snuba/query/allocation_policies/__init__.py
```diff
@@ -261,6 +269,22 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
...
+    Every AllocationPolicyConfig comes with a default value specified in code. That can be overridden for any specific instance of the policy
+    Example:
+
+        >>> policy = MyAllocationPolicy(
+        >>>     storage_key=StorageKey("some_storage"),
+        >>>     required_tenant_types=["foo"],
+        >>>     # This dictionary overrides whatever defaults are set
+        >>>     # for this class
+        >>>     default_config_overrides={
+        >>>         "is_enforced": False, "fart_noise_level": 100
```
rahul-kumar-saini, line 283: rofl

rahul-kumar-saini (approved):

PR #4204, opened 2023-05-18
Fix slack notifications for when allocation policy changes

volokluev:
Current message does not tell us what change fix it and also fix the default behavior to be less useless. Add tests for slack message generation

image

### DIFF
--- a/snuba/admin/audit_log/base.py
+++ b/snuba/admin/audit_log/base.py
@@ -31,7 +31,7 @@ class AuditLog:
             timestamp=timestamp,
             **data,
         )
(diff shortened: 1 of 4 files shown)


snuba/admin/audit_log/base.py
```diff
@@ -31,7 +31,7 @@ def record(
             timestamp=timestamp,
             **data,
         )
+        blocks = build_blocks(data, action, timestamp, user)
```
volokluev, line 34: did this so that the code would get run even if the client wasn't configured (like in tests)

rahul-kumar-saini (approved):

PR #4205, opened 2023-05-19
feat: Python and Rust consumers share common cluster resolution code

lynnagara:
Centralize the code for validating and resolving the Kafka and ClickHouse clusters and other consumer config to be passed to the Python and Rust consumers. Will likely also be shared by the DLQ consumer and multistorage consumer.


snuba/consumers/consumer_config.py
```diff
@@ -0,0 +1,218 @@
+from dataclasses import dataclass
```
lynnagara, line 1: The contents of this file are essentially unchanged - just moved from the Rust consumer file so it can be used in other consumers

volokluev:
how does the rust consumer use the python code?

lynnagara:
The Rust consumer is packaged as a Python module. It gets passed the consumer config (as a string) which is then deserialized in Rust. This is to avoid the need to duplicate things like resolving ClickHouse clusters, Kafka topics, settings overrides, slicing both Rust and python This is a small step towards trying to centralize this code.

untitaker (approved):

PR #4214, opened 2023-05-22
chore(spans): Allow spans ingestion based on project sampling

nikhars:
In order to perform a load test on how much the spans consumer can scale allow an option to configure spans ingestion based on a sampling percentage.


snuba/datasets/processors/spans_processor.py
```diff
@@ -40,6 +40,15 @@
 
 
 def is_project_in_allowlist(project_id: int) -> bool:
+    """
+    Allow spans to be written to Clickhouse if the project falls into one of the following categories:
+    1. The project falls in the configured sample rate
+    2. The project is in the allowlist
+    """
+    spans_sample_rate = state.get_config("spans_sample_rate", None)
+    if spans_sample_rate and project_id % 100 <= spans_sample_rate:
```
dbanda, line 49: Are the last two digits of the project_id always the sample rate? It seems really weird to encode sample rate within the ID. Maybe it would be better to have a mapping somewhere of proj_id to sample rate.

nikhars: I updated the comments and the tests to reflect what is happening. There are 2 ways now for accepting projects to ingest. Sample rate and project list. Sample rate takes precedence over project list. The tests should hopefully make things a bit clearer

dbanda: What's not clear to me is what `project_id % 100` represents. Does it mean the last two digits of the id are the sample rate? Or were you attempting to get a random sample of projects to allow? If it's the second case maybe it's clearer to do `random() < sample_rate/100.0`.  It makes it clearer what's going on and avoids bugs arising from unbalanced samples if the proj_id is not truly random.

untitaker: > Or were you attempting to get a random sample of projects to allow?

Yup that's what's happening here. For what it's worth, `project_id % 100` or a variation of it has been used in sentry for a long time. In practice all of our primary keys are sufficiently random because there's no correlation between the order in which projects are created vs how much traffic they send.

dbanda (approved):

snuba/datasets/processors/spans_processor.py
dbanda: ok if that's a well understood convention then LGTM. approved.

nikhars: > Or were you attempting to get a random sample of projects to allow?
That is correct. I do not want random spans from any project. I would like spans from a fixed number of projects to be ingested. Since project_id is already an int field, doing a modulo operation on it would give the same result versus hashing the project_id and performing the modulo operation. So not doing a hash on the project_id

PR #4220, opened 2023-05-23
ref: Set compression in `build_kafka_producer_configuration()` function

lynnagara:
All Snuba producers use lz4, this is set in the producer configuration building function rather than in independent producer code.


evanh:
> All Snuba producers use lz4

Is this true even in single tenant/self hosted?

lynnagara:
> > All Snuba producers use lz4
> 
> Is this true even in single tenant/self hosted?

Sorry this comment was a bit misleading - meant to say that they *should* all use lz4 like prod, which is what this change is about.

As of today though single tenant does since it's managed similar to prod, but self hosted doesn't and would be using the default which is no compression.

evanh (approved):

untitaker (approved):

PR #4226, opened 2023-05-23
ref(process_message): Tag all metrics by consumer_group

untitaker:
It's basically the same information as consumer_group, however, it is
inconvenient to query snuba_logical_topic due to lack of global filter.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -499,6 +499,13 @@ def process_message(
     snuba_logical_topic: SnubaTopic,
     message: Message[KafkaPayload],
 ) -> Union[None, BytesInsertBatch, ReplacementBatch]:
+    local_metrics = MetricsWrapper(
+        metrics,
+        tags={
+            "consumer_group": consumer_group,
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

snuba/consumers/consumer.py
```diff
@@ -499,6 +499,13 @@ def process_message(
     snuba_logical_topic: SnubaTopic,
     message: Message[KafkaPayload],
 ) -> Union[None, BytesInsertBatch, ReplacementBatch]:
+    local_metrics = MetricsWrapper(
+        metrics,
+        tags={
+            "consumer_group": consumer_group,
+            "snuba_logical_topic": snuba_logical_topic.name,
```
lynnagara, line 506: The physical topic names would be more consistent with those on the dashboard, if we have a way to easily resolve those

untitaker: Possibly but it would be more data to be serialized into subprocesses. I was actually thinking of getting rid of `snuba_logical_topic` too

PR #4235, opened 2023-05-23
feat(dlq): Consumer builder can build a DLQ strategy

lynnagara:
DLQ consumer strategy is similar to the main consumer strategy with 2 notable differences:
- it doesn't commit to the commit log
- it subtitutes ExitAfterNMessages for the CommitOffsets strategy. The consumer exits after the requested number of messages is processed and committed.

### DIFF
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -1,7 +1,7 @@
 import functools
 import logging
 from dataclasses import dataclass
-from typing import Any, Callable, Optional
+from typing import Callable, Optional
 
 from arroyo.backends.kafka import (
     KafkaConsumer,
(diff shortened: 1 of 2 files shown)


snuba/consumers/consumer_builder.py
```diff
@@ -280,6 +282,56 @@ def build_streaming_strategy_factory(
...
+        - never commits to the commit log (the commit log should always be monotonic)
+        """
+        table_writer = self.storage.get_table_writer()
+        stream_loader = table_writer.get_stream_loader()
+
+        logical_topic = stream_loader.get_default_topic_spec().topic
+
+        processor = stream_loader.get_processor()
+
+        def build_commit_strategy(commit: Commit) -> ProcessingStrategy[Any]:
```
untitaker, line 300: I think it would make more sense to add `max_messages_to_process: Optional[int]` as a parameter to `KafkaConsumerStrategyFactory` -- since for other strategies like multiprocessing we also forward args

untitaker (approved):

snuba/consumers/consumer_builder.py
lynnagara: yeah you're right, will update

PR #4240, opened 2023-05-24
feat(DRS): Enable readiness state for storages

enochtangg:
### Overview
This PR is responsible for enabling readiness_states for all storages. This [section of this document](https://www.notion.so/sentry/Dataset-Readiness-State-Rollout-Plan-d38deffe77534856a4e4f4a3b0cfb110?pvs=4#479d1451ca8a457e9542fedce8cdf87c) is used to track the roll out.

### What will happen after this is deployed?
* Existing storages will use their `readiness_state` flag to determine whether or not to execute clickhouse healthchecks in a particular environment.
* Note the `DISABLED_DATASETS` and `is_experimental` still takes precedence over readiness_states.

### Blast Radius
The following storages will use readiness states for each environment:
* In SaaS: All storages except for `groupassignees`, `groupedmessages`, `search_issues`, `querylog`, `sessions`, `spans`.
* In S4S: All storages except for `functions`, `groupassignees`, `groupedmessages`, `search_issues`, `profiles`, `querylog`, `spans`.
* In Self-hosted: All storages except for `search_issues`, `spans`.
* In ST: All storages except for `functions`, `groupassignees`, `groupedmessages`, `search_issues`, `profiles`, `querylog`, `spans`.

These storages are excluded because they still depend on the `DISABLED_DATASETS` and `is_experimental` flags. Eventually, when we obtain parity across environments, we can migrate the rest off these flags and into readiness states.

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -131,8 +131,7 @@ def filter_checked_storages(ignore_experimental: bool) -> List[Storage]:
 
     storages: List[Storage] = []
     for entity in entities:
-        entity_storages = entity.get_all_storages()
-        for storage in entity_storages:
+        for storage in entity.get_all_storages():
             if storage.get_readiness_state().value in settings.SUPPORTED_STATES:
                 storages.append(storage)
     return storages
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -57,7 +57,7 @@ from snuba.datasets.factory import (
     get_enabled_dataset_names,
 )
 from snuba.datasets.schemas.tables import TableSchema
-from snuba.datasets.storage import ReadableTableStorage, Storage, StorageNotAvailable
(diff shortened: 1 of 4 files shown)


snuba/web/views.py
```diff
@@ -133,12 +133,7 @@ def filter_checked_storages(ignore_experimental: bool) -> List[Storage]:
     for entity in entities:
         entity_storages = entity.get_all_storages()
         for storage in entity_storages:
```
rahul-kumar-saini, line 135: nit: Let's refactor this [inline variable](https://refactoring.com/catalog/inlineVariable.html) but otherwise looks good.
```suggestion
        for storage in entity.get_all_storages():
```

rahul-kumar-saini (approved):

PR #4244, opened 2023-05-25
fix(spans): use system instead of platform

dbanda:
Extract platform from `spans.system`

### DIFF
--- a/snuba/datasets/processors/spans_processor.py
+++ b/snuba/datasets/processors/spans_processor.py
@@ -263,7 +263,7 @@ class SpansMessageProcessor(DatasetMessageProcessor):
         else:
             processed_span["span_status"] = UNKNOWN_SPAN_STATUS
         processed_span["domain"] = _unicodify(span_data.get("span.domain", ""))
-        processed_span["platform"] = _unicodify(span_data.get("span.platform", ""))
+        processed_span["platform"] = _unicodify(span_data.get("span.system", ""))
         processed_span["action"] = _unicodify(span_data.get("span.action", ""))
         processed_span["status"] = span_data.get("span.status_code", 0)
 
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/spans_processor.py
```diff
@@ -357,6 +355,7 @@ def process_message(
             "retention_days": retention_days,
             "partition": metadata.partition,
             "offset": metadata.offset,
+            "platform": "",
```
nikhars, line 358: The `common_span_fields` are used to hold fields which need to be pushed down to each individual span. I would avoid adding `platform` here.

snuba/datasets/processors/spans_processor.py
```diff
@@ -382,6 +381,11 @@ def process_message(
             processed_rows.append(processed)
             processed_spans = self._process_spans(event_dict, common_span_fields)
             processed_rows.extend(processed_spans)
+
+            # extract platform from spans
+            if processed_spans:
+                processed["platform"] = processed_spans[0].get("platform", "")
```
nikhars, line 387: `processed` dictionary is not useful here since all processed spans have already been added to the `processed_rows` before this. For the root span, I would recommend that you set the `platform` to empty string in `_process_module_details`. When you make that change, the tests would change to reflect that `platform` is set to empty string in the first span of the result

nikhars: For now yes. But it is something we should follow up on. I can initiate the conversation to figure it out and tag you

snuba/datasets/processors/spans_processor.py
dbanda: removed

snuba/datasets/processors/spans_processor.py
dbanda: ok I set to empty for now.

PR #4245, opened 2023-05-25
feat(dashboard): Add settings file to frontend code

evanh:
In order to integrate Sentry into the dashboard frontend, we need a way to have
environment specific settings in the frontend. Add a file that can be overriden
in our production environments.

Also add a Sentry integration that uses that file. This is a noop right now,
but we can have it start working once the appropriate change has been made
in the prod environment.


evanh:
This might not be the right solution after doing some further research.

PR #4246, opened 2023-05-25
config(replays): change group_by_overflow to break

JoshFerge:
Still observing memory usage at high volume queries. changing the overflow mode to `break` should fix this. It's possible small numbers of replays could be broken as it will stop all scanning once the key limit is reached, but i think this is a very unlikely edge case, all though will have to verify in production.

Relevant docs: https://clickhouse.com/docs/en/operations/settings/query-complexity

### DIFF
--- a/snuba/datasets/configuration/replays/storages/replays.yaml
+++ b/snuba/datasets/configuration/replays/storages/replays.yaml
@@ -162,7 +162,7 @@ query_processors:
     args:
       settings:
         max_rows_to_group_by: 1000000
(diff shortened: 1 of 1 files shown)


cmanallen (approved):

rahul-kumar-saini (approved):

JoshFerge:
It seems like we fixed the OOMs in the previous PR https://github.com/getsentry/snuba/pull/4232 closing this!

PR #4247, opened 2023-05-25
ref(dashboard) Add frontend settings endpoint

evanh:
The frontend also needs access to some production only values. Unfortunately
those can't be added to the frontend code automatically, since the bundle is
built offline and React recommends not exposing secrets in the app anyways, for
security reasons.

Instead, add an endpoint that the frontend can use to fetch relevant settings
from the backend. These can then be used by the frontend code.


snuba/admin/static/index.tsx
```diff
@@ -19,6 +20,19 @@ const bodyStyle = {
 };
 
 let client = Client();
+client.getSettings().then((settings) => {
+  if (settings.dsn != "") {
+    Sentry.init({
+      dsn: settings.dsn,
+      integrations: [new Sentry.BrowserTracing(), new Sentry.Replay()],
+      // Performance Monitoring
+      tracesSampleRate: 1.0, // Capture 100% of the transactions, reduce in production!
```
rahul-kumar-saini, line 29: How do we reduce this in production, shouldn't we just set the right value here?

evanh: I wasn't planning to reduce it at all. This isn't going to generate a lot of traces. However I will change it to be configurable just in case.

rahul-kumar-saini (approved):

PR #4249, opened 2023-05-25
fix(logs): use structlog for optimize module

dbanda:
## Context
Were in the process of migrating over the logs to use stuctured logging where possible. This makes them easier to parse in tools like GCP. The plan is to implement this module by module. This PR enables structured logging for the optimize module.

### DIFF
--- a/snuba/clickhouse/optimize/optimize.py
+++ b/snuba/clickhouse/optimize/optimize.py
@@ -1,8 +1,12 @@
-import logging
+import multiprocessing
+import os
 import threading
 import time
 from datetime import datetime, timedelta
 from typing import Mapping, MutableSequence, Optional, Sequence
+
+import structlog
+from structlog.types import EventDict, WrappedLogger
 
 from snuba import environment, util
 from snuba.clickhouse.native import ClickhousePool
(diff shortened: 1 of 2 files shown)


snuba/clickhouse/optimize/optimize.py
```diff
@@ -22,17 +26,20 @@
...
 # include thread and process info in log messages
-log_handler = logging.StreamHandler()
-log_handler.setFormatter(
-    logging.Formatter(
-        "%(levelname)s %(asctime)s %(name)s %(process)d-%(threadName)s %(message)s"
-    )
-)
-logger.addHandler(log_handler)
-logger.propagate = False
+def thread_info_processor(
```
evanh, line 31: Why is this all necessary? Was this information automatically collected by the previous logger?

dbanda: Previous logger used the `logging.Formatter` class for format to add the thread info. `structlog` add this info by using the processor wrapper. We spin up threads for running the optimizer so we want visibility on thread id and thread name in all the logs.

evanh (approved):

PR #4251, opened 2023-05-26
ref: Decouple processing batches and clickhouse batches

untitaker:
Add two new parameters to be able to control the size of INSERT
statements separately from the size of multiprocessing batches.

Also change commit policy so that committing never happens more than
once per second. Committing is still inherently tied to how the
preceding processing steps batch.

### DIFF
--- a/snuba/consumers/consumer_builder.py
+++ b/snuba/consumers/consumer_builder.py
@@ -10,7 +10,7 @@ from arroyo.backends.kafka import (
     build_kafka_configuration,
     build_kafka_consumer_configuration,
 )
-from arroyo.commit import IMMEDIATE
+from arroyo.commit import ONCE_PER_SECOND
 from arroyo.dlq import DlqLimit, DlqPolicy, KafkaDlqProducer
 from arroyo.processing import StreamProcessor
 from arroyo.processing.strategies import ProcessingStrategyFactory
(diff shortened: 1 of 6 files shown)


lynnagara:
I'm a bit wary of adding parameters to the consumer unless absolutely needed, as there's more things to think about when tuning them. What would be the benefit of being able to separately configure the 2 batch sizes?

untitaker:
We are currently struggling to use more than two cores in snuba-generic-metrics-sets, and the only way I see that improving is if we reduce multiprocessing batch size. At the same time, we cannot reduce clickhouse http batch size

lynnagara:
Sorry, i lost track of this earlier. Not sure if you're still planning to do these changes..

untitaker:
yeah I think we should give it a try for some snuba consumers and see if it improves CPU saturation

lynnagara (approved):

snuba/consumers/consumer_builder.py
```diff
@@ -132,6 +134,8 @@ def __init__(
         self.metrics = metrics
         self.max_batch_size = max_batch_size
         self.max_batch_time_ms = max_batch_time_ms
+        self.max_insert_batch_size = max_insert_batch_size
+        self.max_insert_batch_time_ms = max_insert_batch_time_ms
```
lynnagara, line 138: Should these default to the original max_batch_size/max_batch_time values?

untitaker: yeah that fallback happens later in the strategy factory

PR #4270, opened 2023-05-31
feat(capman): replays allocation policy

volokluev:


rahul-kumar-saini:
Broke getsentry tests this time https://github.com/getsentry/getsentry/actions/runs/5136968047/jobs/9244411714

PR #4276, opened 2023-05-31
migration(events): add trace_sampled and num_processing_errors columns to errors and errors_ro

barkbarkimashark:
Adds the following columns to errors:
* `trace_sampled` (1/0/NULL) - bool whether the trace is sampled or not.
* `num_processing_errors` (number/NULL) - [number of errors while processing the error event](https://develop.sentry.dev/sdk/event-payloads/#core-interfaces). 

This functionality is meant to support surfacing up a 'more useful event' in the issue details page. Currently, the event shown when a user goes to the issue detail page is the latest event that groups to that issue. We want to help users debug and find the root cause of the issue by up-ranking more useful events with extra contextual information that can help them resolve the issue.

Conceptually, the sort might look something like this: `order by if(replay_id is not null, 1, 0) desc, num_processing_errors asc, trace_sampled desc, timestamp desc`

Related to https://github.com/getsentry/relay/pull/2026/files


evanh (commented):
Can you give some additional context on where this column will be used?

evanh (approved):
My concern with this is that if you start sorting the main issues page by this column, the performance could drop significantly because it's not an indexed column.

So something to be aware of as you test that feature.

barkbarkimashark:
> My concern with this is that if you start sorting the main issues page by this column, the performance could drop significantly because it's not an indexed column.
> 
> So something to be aware of as you test that feature.

Yeah, that's fair. We don't have any plans to use this column as a sort column for issues. But I'll keep that in mind if we do end up adding any features that would use this column.

PR #4280, opened 2023-06-02
feat(functions): Add timeseries processor for functions entity

Zylphrex:
To support timeseries on functions data, this adds the timeseries processor on the functions entity.

### DIFF
--- a/snuba/datasets/configuration/functions/entities/functions.yaml
+++ b/snuba/datasets/configuration/functions/entities/functions.yaml
@@ -90,6 +90,12 @@ query_processors:
   - processor: ResourceQuotaProcessor
     args:
       project_field: project_id
+  - processor: TimeSeriesProcessor
+    args:
+      time_group_columns:
+        time: timestamp
+      time_parse_columns:
+      - timestamp
 
 validators:
   - validator: EntityRequiredColumnValidator


evanh (changes requested):
This processor is deprecated and shouldn't be necessary any more. What are you trying to do?

Zylphrex:
> This processor is deprecated and shouldn't be necessary any more. What are you trying to do?

We're looking to be able to generate timeseries data from the functions dataset. So a query might look like

```
MATCH (functions)
SELECT arrayElement(quantilesMerge(0.95)(percentiles), 1) AS `p95`
BY time
WHERE timestamp >= toDateTime('2023-05-29T14:30:56')
  AND timestamp < toDateTime('2023-06-05T14:30:56')
  AND project_id IN array(1)
ORDER BY time ASC
LIMIT 10000
GRANULARITY 21600
```

This previously was done using the time series processor that handled the `time` column and granularity. Is there a new way of doing this? If so, what is it?

Zylphrex:
> > Is there a new way of doing this? If so, what is it?
> 
> The TimeSeriesProcessor was just syntactic sugar that converted `time` into `toStartOfHour(timestamp)` or whatever granularity you wanted. Now you can do all of that in SnQL, so the processor isn't necessary.
> 
> The query you have above would be
> 
> ```
> MATCH (functions)
> SELECT arrayElement(quantilesMerge(0.95)(percentiles), 1) AS `p95`
> BY toDateTime(multiply(intDiv(toUInt32(timestamp), 21600), 21600)) AS `rollup`
> WHERE timestamp >= toDateTime('2023-05-29T14:30:56')
>   AND timestamp < toDateTime('2023-06-05T14:30:56')
>   AND project_id IN array(1)
> ORDER BY `rollup` ASC
> LIMIT 10000
> ```

Understood, I'll close this PR and implement this functionality in sentry instead.

PR #4290, opened 2023-06-05
rfc(CapMan): New Capacity Management API for multiple storages

rahul-kumar-saini:


volokluev (approved):
This looks good to me, I might think about it a little bit more but sleep on it first before implementing it

PR #4294, opened 2023-06-05
feat(CapMan): Multiple Allocation Policies on a Storage [backend]

rahul-kumar-saini:
### Overview
- In order to bring rate-limiting and other forms of capacity management into the new capacity management system, storages should be able to have multiple allocation policies
- These policies should be chained when calculating resources
- <strong>Ignore \*.tsx when reviewing this PR</strong>
    - Those changes were reviewed in a different PR mentioned below

### Breaking changes
- ~This PR updates the API for the capacity management admin tool, the UI *will* break if not updated~ 
- #4316 has been merged into this branch, there should be no issues

#### Important new stuff
- `AllocationPolicyViolations`
    - Previously, db_query ran the get_quota method of a single allocation policy, which could raise an exception
    - Now it runs multiple policies methods and catches all exceptions, raising a single `AllocationPolicyViolations` exception containing all of them
- Throttling
    - The minimum `max_threads` of all allocation policies quota allowance will be used to throttle queries
    - All quota allowances are now added to the `stats` of a query

#### Changes
- Replaced the concept of a single allocation policy on a storage with a list of allocation policies
- Updated all storage yamls to use a list of policies instead of singular
- Updated admin API for capacity management endpoints with API agreed upon here #4290
- Introduced new AllocationPolicyViolations exception
- Updated Slack message for config updates to also show policy name
- Updated tests


volokluev:
Can we make any minimal changes to the UI in this PR so we don't break it?

rahul-kumar-saini:
> Can we make any minimal changes to the UI in this PR so we don't break it?

I don't think it can be too minimal, updating how the API is used so the front end doesn't break will encompass almost everything that needs to be updated. It could be easier to just branch off of this PR to create the frontend and create a new PR back into this branch so the diff can be reviewed separately.

evanh (approved):
Seems good to me.

PR #4296, opened 2023-06-06
ref(validation): Turn on column validation by default

evanh:
This enables the column validation by default on all entities.

The code itself has been tested on multiple datasets, so it should
be safe to turn on across the board. This PR will be validated
against all Sentry tests.

- [X] Snuba CI passes
- [X] Sentry tests pass (https://github.com/getsentry/sentry/pull/50446)
- [X] Getsentry tests pass (https://github.com/getsentry/getsentry/pull/10769)


rahul-kumar-saini (approved):
🔥

PR #4298, opened 2023-06-06
ref: Don't pass unnecessary argument to process_message

lynnagara:
This was originally passed in due to the old DLQ implementation which wrote the consumer group into the invalid message that was produced. It's no longer needed.

### DIFF
--- a/snuba/consumers/consumer.py
+++ b/snuba/consumers/consumer.py
@@ -521,14 +521,12 @@ MultistorageProcessedMessage = Sequence[
 
 def process_message(
     processor: MessageProcessor,
-    consumer_group: str,
     snuba_logical_topic: SnubaTopic,
     message: Message[KafkaPayload],
 ) -> Union[None, BytesInsertBatch, ReplacementBatch]:
     local_metrics = MetricsWrapper(
         metrics,
         tags={
-            "consumer_group": consumer_group,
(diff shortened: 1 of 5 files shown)


snuba/consumers/consumer.py
```diff
@@ -521,14 +521,12 @@ class MultistorageKafkaPayload(NamedTuple):
...
 def process_message(
     processor: MessageProcessor,
-    consumer_group: str,
     snuba_logical_topic: SnubaTopic,
     message: Message[KafkaPayload],
 ) -> Union[None, BytesInsertBatch, ReplacementBatch]:
     local_metrics = MetricsWrapper(
         metrics,
         tags={
-            "consumer_group": consumer_group,
```
MeredithAnya, line 531: so we don't need the consumer group in the metrics anymore?

lynnagara: I don't think it's all that useful here, it just tells us about invalid messages on the source topic which will be the same regardless of consumer group. It's not worth passing it in here for.

untitaker: I had to add it so that the global dashboard filters work, it's actually there not to break down by consumer group but to filter approximately by dataset (since it's the same metric across datasets)

`snuba_logical_topic` can go, on the other hand. It's not used in dashboards at all, and introducing it would be cumbersome.

PR #4302, opened 2023-06-07
ref: Clean up the way DLQ is specified in storage config files

lynnagara:
The only thing the user should specify is the topic to be used


untitaker (approved):
i think we had various different dlq policies in mind, including one where the consumer would just crash. how will that look?

lynnagara:
Policy has a different meaning here, and maybe we shouldn't use that word. The other one relates to the replaying policy, number of messages, etc, and that is specified by the user when they provide the replay instruction.

PR #4303, opened 2023-06-07
feat: Add logical topic to shared consumer config

lynnagara:
So it can be easily resolved for all of our consumers without duplicating code in each.

### DIFF
--- a/rust_snuba/src/config.rs
+++ b/rust_snuba/src/config.rs
@@ -18,6 +18,7 @@ pub struct ConsumerConfig {
 #[derive(Deserialize)]
 pub struct TopicConfig {
     pub physical_topic_name: String,
+    pub logical_topic_name: String,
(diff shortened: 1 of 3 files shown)


lynnagara:
@untitaker It came up in my refactor to move multistorage consumer onto the shared consumer config... but i think it would also definitely be needed for schema validation?

lynnagara:
On second thoughts - I was able to rewrite the multistorage consumer without this (https://github.com/getsentry/snuba/pull/4305). I think it might still be needed later for schema validation, but we can close this right now if you think it's better.

untitaker:
> I think it might still be needed later for schema validation

you're right. i think multistorage consumer PR is nice though. i'd just like to avoid using logical topic for topic resolution. schema validation is fine

untitaker (approved):

PR #4304, opened 2023-06-07
ref(admin) Allow access to tracing/snql-to-sql by default

evanh:
Create a new role for product developers that gives access to the tracing and
snql-to-sql tool. Make this role a default role, so everyone with access to the
admin tool will have access to both tools.

Small additional changes: add David to list of users with access to all tools,
and have the product tool role reflect the actual google group we will use to
grant access going forward.

### DIFF
--- a/snuba/admin/iam_policy/iam_policy.json
+++ b/snuba/admin/iam_policy/iam_policy.json
@@ -39,25 +39,16 @@
           "user:oliver.newland@sentry.io",
           "user:rahul.saini@sentry.io",
           "user:riya.chakraborty@sentry.io",
-          "user:volo.kluev@sentry.io"
+          "user:volo.kluev@sentry.io",
+          "user:david.tsukernik@sentry.io"
         ],
         "role": "roles/AllTools"
       },
       {
         "members": [
-            "user:colton.allen@sentry.io",
-            "user:josh.ferge@sentry.io",
-            "user:kevan.fisher@sentry.io"
+            "group:access-snuba-admin@sentry.io"
         ],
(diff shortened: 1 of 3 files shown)


snuba/admin/iam_policy/iam_policy.json
```diff
@@ -39,25 +39,16 @@
...
+          "user:david.tsukernik@sentry.io"
         ],
         "role": "roles/AllTools"
       },
       {
         "members": [
-            "user:colton.allen@sentry.io",
-            "user:josh.ferge@sentry.io",
-            "user:kevan.fisher@sentry.io"
+            "group:access-snuba-admin@sentry.io"
```
volokluev, line 49: why not the engineering team group?

snuba/admin/iam_policy/iam_policy.json
```diff
@@ -39,25 +39,16 @@
           "user:oliver.newland@sentry.io",
           "user:rahul.saini@sentry.io",
           "user:riya.chakraborty@sentry.io",
-          "user:volo.kluev@sentry.io"
+          "user:volo.kluev@sentry.io",
+          "user:david.tsukernik@sentry.io"
```
lynnagara, line 43: can we have the search and storage group here too rather than listing our individual emails?

evanh: That part is being worked on. Once we can fetch group membership from the Google API, we can get rid of the individual emails. That's separate from this change however.

enochtangg (approved):

PR #4305, opened 2023-06-07
ref: Refactor multistorage consumer to use consumer config

lynnagara:
Multistorage consumer now resolves topics and storages in the same way as the other consumers


untitaker (approved):
nice

PR #4306, opened 2023-06-08
Use Google Groups API for authorization

davidtsuk:
This change calls the Groups API to check that a user has membership in a given group. This will allow the iam_policy file to use group-based role assignment.

### DIFF
--- a/snuba/admin/google.py
+++ b/snuba/admin/google.py
@@ -3,6 +3,7 @@ from urllib.parse import urlencode
 
 import structlog
 from googleapiclient.discovery import Resource, build
+from sentry_sdk import capture_exception
 
 logger = structlog.get_logger().bind(module=__name__)
 
@@ -21,6 +22,7 @@ class CloudIdentityAPI:
             self.initialized = True
         except Exception as e:
             logger.exception(e)
+            capture_exception(e)
 
     def _get_group_id(self, group_email: str) -> Optional[str]:
         if not self.initialized:
--- a/snuba/admin/google.py
+++ b/snuba/admin/google.py
@@ -1,4 +1,3 @@
-from abc import ABC
 from typing import Optional
 from urllib.parse import urlencode
 
@@ -8,15 +7,9 @@ from googleapiclient.discovery import Resource, build
 logger = structlog.get_logger().bind(module=__name__)
 
 
-class GoogleAPIWrapper(ABC):
-    """
-    An abstract class defining an interface for google cloud APIs.
-    """
-
-
 class CloudIdentityAPI:
     """
-    An interface for the Google Cloud Identity API.
+    A class for interfacing with the Google Cloud Identity API.
     """
 
     def __init__(self, service: Resource = None) -> None:
--- a/snuba/admin/google.py
+++ b/snuba/admin/google.py
@@ -19,9 +19,11 @@ class CloudIdentityAPI:
     An interface for the Google Cloud Identity API.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, service: Resource = None) -> None:
         try:
-            self.service: Resource = build("cloudidentity", "v1")
+            self.service: Resource = (
+                service if service else build("cloudidentity", "v1")
+            )
         except Exception as e:
             logger.exception(e)
 
@@ -44,12 +46,10 @@ class CloudIdentityAPI:
 
         return None
 
-    def check_group_membership(self, group_email: str, member: str) -> bool:
+    def _check_transitive_membership(
+        self, group_resource_name: str, member: str
+    ) -> bool:
         try:
-            group_resource_name = self._get_group_id(group_email)
-            if not group_resource_name:
-                return False
-
             query_params = urlencode({"query": "member_key_id == '{}'".format(member)})
             request = (
                 self.service.groups()
@@ -63,9 +63,17 @@ class CloudIdentityAPI:
                     f"An HTTP error occured when checking if user {member} is a member of group {group_resource_name}",
                     google_api_error=response["error"],
                 )
-
             return bool(response.get("hasMembership", False))
         except Exception as e:
             logger.exception(e)
+
+        return False
+
+    def check_group_membership(self, group_email: str, member: str) -> bool:
+        group_resource_name = self._get_group_id(group_email)
+        if group_resource_name:
+            return self._check_transitive_membership(
+                group_resource_name=group_resource_name, member=member
+            )
 
         return False
--- a/snuba/admin/google.py
+++ b/snuba/admin/google.py
@@ -1,21 +1,9 @@
 from urllib.parse import urlencode
 
+import structlog
 from googleapiclient.discovery import Resource
 
-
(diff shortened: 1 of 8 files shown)


volokluev (commented):
Can you write a test for this? Also, when this is merged does the snuba-admin immediately start using the google groups API to authenticate or is there some other work down the road to be done?

davidtsuk:
> Can you write a test for this? Also, when this is merged does the snuba-admin immediately start using the google groups API to authenticate or is there some other work down the road to be done?

Since I'm just calling the API and directly using the response, I don't think this can be tested with unit tests. Even if I write unit tests for the functions I added, they will just use mock API responses so they won't actually be testing anything. We should add additional tests for authorization, but I think that should be done in a separate PR.

When this is merged, snuba-admin should in theory immediately start using the google groups API, but there is still work to be done with actually updating the iam_policy file to only use groups. Whether or not additional work needs to be done outside of that depends on whether the service account for snuba is properly set up to be able to read group information. I am not sure if that's the case right now.

volokluev:
> When this is merged, snuba-admin should in theory immediately start using the google groups API

Okay, can we make a test that would make sure this happened in practice rather than in theory?

snuba/admin/google.py
```diff
@@ -0,0 +1,31 @@
...
+    try:
+        query_params = urlencode({"query": "member_key_id == '{}'".format(member)})
+        request = (
+            service.groups().memberships().checkTransitiveMembership(parent=parent)
+        )
+        request.uri += "&" + query_params
+        response = request.execute()
+        return bool(response["hasMembership"])
+    except Exception as e:
+        print(e)
```
volokluev, line 16: This swallows the error and makes sure it goes nowhere. `logging.exception` will at least send it to sentry.

Also a blanket try catch like this is too permissive. This will catch everything like `KeyError` and `AttributeError` which you probably want to fail loudly. 

Combine this with the fact that there are no tests for this function (or `get_group_id`) means that you don't have a reliable authentication mechanism

davidtsuk: You're right, I'll refactor this to log the error.

snuba/admin/google.py
```diff
@@ -0,0 +1,31 @@
...
+
+def get_group_id(service: Resource, group_email: str) -> str:
+    try:
+        query_params = f"groupKey.id={group_email}"
+        request = service.groups().lookup()
+        request.uri += "&" + query_params
+        response = request.execute()
+        return str(response["name"])
+    except Exception as e:
+        print(e)
```
evanh, line 29: This was mentioned before, but don't put `print` statements in production code, and make sure errors are being logged.

evanh (commented):
This requires some tests. The google library provides a mocking system that could be used for this. https://googleapis.github.io/google-api-python-client/docs/mocks.html

There are a number of tests based around the iam authentication that you could use as inspiration to test: if a user is a member of a group, they are granted access to an endpoint, and the negative of that as well. See [here](https://github.com/getsentry/snuba/blob/master/tests/admin/clickhouse_migrations/test_api.py#L284) for an example that overrides the iam_policy.json file to test different permissions.

davidtsuk:
@volokluev @evanh I've added some tests using a mock http object

snuba/admin/google.py
```diff
@@ -0,0 +1,49 @@
...
+        request = service.groups().lookup()
+        request.uri += "&" + query_params
+        response = request.execute()
+        if "error" in response:
+            logger.exception(
+                f"An HTTP error occured when fetching group id for email {group_email}.",
+                google_api_error=response["error"],
+            )
+
+        return str(response.get("name", ""))
```
lynnagara, line 21:
```suggestion
        return response["name"]
```
When would `name` not be in the response? Should we log the error if that happens?

davidtsuk: This can happen when there's an error in the response in which case we log the error anyway. This ensures that we only log the error once in that case and not again when a KeyError is raised.

snuba/admin/google.py
```diff
@@ -0,0 +1,49 @@
...
+    except Exception as e:
+        logger.exception(e)
```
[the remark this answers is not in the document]
davidtsuk: Essentially, there are only a few ways these can fail at all. Either something is misconfigured (e.g. the group email, infra-related config) or the google api is failing. For the user, if the check fails, they will simply not see the tool that they should otherwise see. This way, the worst that could happen is that they do not have access to a tool they should have access to which is preferable over accidentally giving someone access to a role that they don't have access to.

It might be a good idea to add a caching mechanism that caches a group's members and in case of failure, the function can check the cache. The only downside to doing this that I can think of is that in the case where someone is revoked access to a group and something fails, they will have access to a role they shouldn't have access to for the duration that the function is failing, but I think this case is extremely unlikely. We can also add caching for group IDs which I think is a good idea, but I am not sure if a google group ID ever changes, so this is something that we'd have to look into.

snuba/admin/google.py
```diff
@@ -1,49 +1,71 @@
...
+class CloudIdentityAPI:
+    """
+    An interface for the Google Cloud Identity API.
+    """
+
+    def __init__(self) -> None:
+        try:
+            self.service: Resource = build("cloudidentity", "v1")
+        except Exception as e:
+            logger.exception(e)
```
evanh, line 28: Small nit: I might set a `self.initialized = False` here and `self.initialized = True` on success. Then in the other functions you can check `self.initialized` instead of catching Exceptions.

snuba/admin/google.py
```diff
@@ -1,49 +1,71 @@
...
-        query_params = f"groupKey.id={group_email}"
-        request = service.groups().lookup()
-        request.uri += "&" + query_params
-        response = request.execute()
-        if "error" in response:
-            logger.exception(
-                f"An HTTP error occured when fetching group id for email {group_email}.",
-                google_api_error=response["error"],
-            )
+class GoogleAPIWrapper(ABC):
```
evanh, line 11: This should define functions using the `@abstractmethod` decorator. Otherwise this doesn't define any interface.

evanh (commented):
The purpose of `GoogleAPIWrapper` is to actually provide an interface. If you don't define any functions in it, then any class technically would satisfy the interface. Also there's no "dummy" API here to use in tests.

Since this code isn't actually using the interface model, I would ask whether you really want it or not. If what you have here works, then the `GoogleAPIWrapper` class doesn't provide any value.

snuba/admin/google.py
davidtsuk: Oh oops, I meant to delete this.

snuba/admin/google.py
davidtsuk: I've added this flag, but I think it's a good idea to catch exceptions here, so that they're silently caught and the end user will not necessarily know about it. I think this might be a good place to send this exception to sentry as well, so we can be alerted if something silently breaks. @evanh What do you think?

evanh (approved):

PR #4307, opened 2023-06-08
feat: Always record scheduling latency

lynnagara:
We want to ensure the scheduling latency is recorded even for low volume environments where there are barely any subscriptions registered (e.g. single tenants).


untitaker (approved):

mwarkentin (approved):
LGTM

mwarkentin:
This should make it out to all of our ST customers by June 29, assuming deployments go smoothly.

PR #4311, opened 2023-06-08
fix(capman): Change max_threads to use allocation policy config

evanh:
Change the `max_threads` attribute of an AllocationPolicy to use the config of
the policy instead of the state config. This means it can be overridden in the
dataset configuration files as well.

This means that the max threads in the allocation policy will no longer use the
state setting. The state setting still is used in db_query.py however. The
default value for the allocation policy max threads has been set to match what
is in prod currently.


rahul-kumar-saini (approved):
Cool, now we can set search_issues max_threads to 4 instead of the hacky way we're doing it right now.

PR #4315, opened 2023-06-08
fix(tests): Skip flaky optimize tests

nikhars:
Disabling flaky test for now to improve everyone's life

### DIFF
--- a/tests/clickhouse/optimize/test_optimize_tracker.py
+++ b/tests/clickhouse/optimize/test_optimize_tracker.py
@@ -44,6 +44,7 @@ redis_client = get_redis_client(RedisClientKey.REPLACEMENTS_STORE)
 )
 @pytest.mark.clickhouse_db
 @pytest.mark.redis_db
+@pytest.mark.skip(reason="This test is flaky and needs to be fixed.")
 def test_optimized_partition_tracker(tracker: OptimizedPartitionTracker) -> None:
     def assert_partitions(
         *,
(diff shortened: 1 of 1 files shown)


lynnagara:
Is there a way to skip for distributed clickhouse only?

evanh (approved):

nikhars:
> Is there a way to skip for distributed clickhouse only? We could add a `@pytest.mark.skip_distributed` or something?

Not sure how easy it is to do that. Haven't done it before. I don't see an example of this as well. I will just skip this test for now and we can revisit it hopefully with the right fix.

lynnagara:
If you take a look at how we did `pytest.mark.ci_only`, it should be the same as that

nikhars:
> If you take a look at how we did `pytest.mark.ci_only`, it should be the same as that

It doesn't seem as simple as `ci_only` mark. The only difference between `tests` and `tests distributed` are separate settings file as can be seen [here](https://github.com/getsentry/snuba/blob/master/.github/workflows/ci.yml#L160-L212). If we had to make this work to differentiate between those 2 the effort required would probably be more than I originally intended. So I would prefer to skip the test entirely since there is no active work happening in the codebase for the test which is being disabled.

untitaker:
I think you can just do `pytest.mark.skipif(os.environ.get("SNUBA_SETTINGS") == "test_distributed")` or something gross like that

PR #4319, opened 2023-06-09
fix(envrc): Try to grab the right Python version

untitaker:
Just migrated to new laptop, and clickhouse-driver does not install with
the random Python 3.11 I happen to have installed, so this envrc does
not work out of the box.

A lot of people who try to contribute to Snuba likely already have
sentry installed, so I think we can just piggyback off of their Python
3 install. We practically already do for devservices.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,21 +1,22 @@
 #!/bin/bash
 # shellcheck disable=SC1091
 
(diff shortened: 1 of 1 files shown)


untitaker:
actually ignore me. I just don't use pyenv shell hooks, that's why it's broken. this envrc works fine 🤦

PR #4320, opened 2023-06-09
fix(capman): Update allocation policy variable descriptions

enochtangg:
This PR is responsible for updating the variable UI  descriptions for `is_active` and `is_enforced` for the allocation policy (SNS-2369).


snuba/query/allocation_policies/__init__.py
```diff
@@ -323,13 +323,13 @@ def __init__(
         self._default_config_definitions = [
             AllocationPolicyConfig(
                 name=IS_ACTIVE,
-                description="Whether or not this policy is active.",
+                description="Toggles whether or not this policy is active. If inactive, the system will default to a base policy.",
```
volokluev, line 326: >  If inactive, the system will default to a base policy

What is the base policy? Think of yourself as oncall getting paged for this at 2am. does this description help you while you're half asleep?

snuba/query/allocation_policies/__init__.py
```diff
@@ -323,13 +323,13 @@ def __init__(
...
                 value_type=int,
                 default=default_config_overrides.get(IS_ACTIVE, 1),
             ),
             AllocationPolicyConfig(
                 name=IS_ENFORCED,
-                description="Whether or not this policy is enforced.",
+                description="Toggles whether or not this policy is enforced. If enforced, policy will be able to throttle/reject incoming queries.",
```
volokluev, line 332: what does it mean for it not to be enforced?

snuba/query/allocation_policies/__init__.py
enochtangg: From what I understand, the base policy could be different for each type of allocation policy. For example, the `BytesScannedWindowAllocationPolicy` default is `DEFAULT_PASSTHROUGH_POLICY`. In the future, when we eventually add new policies, I'd imagine their defaults would be different?

On a side note:  one recommendation is what is we defined the default/base policy as an attribute of those inheriting `AllocationPolicy` rather than calling them inline?

snuba/query/allocation_policies/__init__.py
```diff
@@ -323,13 +323,13 @@ def __init__(
...
                 name=IS_ACTIVE,
-                description="Whether or not this policy is active.",
+                description="Toggles whether or not this policy is active. If inactive, the system will default to a base policy.",
                 value_type=int,
                 default=default_config_overrides.get(IS_ACTIVE, 1),
             ),
             AllocationPolicyConfig(
                 name=IS_ENFORCED,
-                description="Whether or not this policy is enforced.",
+                description="Toggles whether or not this policy is enforced. If enforced, policy will be able to throttle/reject incoming queries. If not enforced, this policy will not throttle/reject queries if policy is triggered.",
```
volokluev, line 332:
```suggestion
                description="Toggles whether or not this policy is enforced. If enforced, policy will be able to throttle/reject incoming queries. If not enforced, this policy will not throttle/reject queries if policy is triggered. But all the policy code will still run.",
```

snuba/query/allocation_policies/__init__.py
enochtangg: Discussed offline: Will move is_active logical to base `AllocationPolicy` class

volokluev (approved):

PR #4323, opened 2023-06-09
feat(spans): Allow spans migrations to run locally

enochtangg:
This PR allows spans migrations to run locally in snuba since this already runs in SaaS.


volokluev (commented):
nice

dbanda (approved):

PR #4324, opened 2023-06-09
fix(gen-metrics): Add tags hash column to the generic metrics dist ta…

nikhars:
We have the tags hash on the raw tags table on generic metrics defined on the local tables but do not have it on the dist tables. This PR adds it to the dist tables for distributions. A separate PR will create it for counters and sets dist tables.


onewland (approved):

onewland:
the change looks fine but tests_distributed_migrations is failing which may matter

PR #4327, opened 2023-06-12
feat: Allow parallel replica read for configured datasets

dmajere:
On Disney ST enviroment we observed improved query latencies for long queries when executed with [max_parallel_replicas](https://clickhouse.com/docs/en/operations/settings/settings#settings-max_parallel_replicas) setting. With 4 replicas available we observed 4x improvement in latencies.

For older clickhouse clusters only data sets with SAMPLE BY supported.

This feature allows to configure max_parallel_replicas and the list of datasets where parallel execution should be allowed.

3 configuration parameters added to state:
* parallel_replica_reading_enabled: int - main feature toggle
* parallel/max_parallel_replicas: int - passed as clickhouse query setting to clickhouse 
* parallel/tables: str - comma separated list of datasets, for example transsactions_dist,errors_dist

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -14,7 +14,6 @@ from typing import (
     MutableMapping,
     MutableSequence,
     Optional,
-    Tuple,
     Union,
     cast,
 )
@@ -25,7 +24,7 @@ from clickhouse_driver.errors import ErrorCodes
 from sentry_sdk import Hub
 from sentry_sdk.api import configure_scope
 
-from snuba import environment, state
+from snuba import environment, settings, state
 from snuba.attribution.attribution_info import AttributionInfo
 from snuba.clickhouse.errors import ClickhouseError
 from snuba.clickhouse.formatter.nodes import FormattedQuery
@@ -80,8 +79,6 @@ from snuba.web import QueryException, QueryResult, constants
 metrics = MetricsWrapper(environment.metrics, "db_query")
 
 redis_cache_client = get_redis_client(RedisClientKey.CACHE)
-
-ENABLE_PARALLEL_READING: str = "parallel_replica_reading_enabled"
 
 
 class ResultCacheCodec(ExceptionAwareCodec[bytes, Result]):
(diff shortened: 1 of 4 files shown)


snuba/web/db_query.py
```diff
@@ -80,6 +81,8 @@
 
 redis_cache_client = get_redis_client(RedisClientKey.CACHE)
 
+ENABLE_PARALLEL_READING: str = "parallel_replica_reading_enabled"
```
evanh, line 84: Please move this into `settings.__init__.py`

snuba/web/db_query.py
```diff
@@ -191,6 +194,8 @@ def execute_query(
         clickhouse_query_settings["load_balancing"] = "in_order"
         clickhouse_query_settings["max_threads"] = 1
 
+    stats["parallel_read"] = "max_parallel_replicas" in clickhouse_query_settings
```
evanh, line 197: Why did you call this `parallel_read`? Why not leave it `max_parallel_replicas` for clarity? Also, why not set the actual value?

snuba/web/db_query.py
```diff
@@ -520,17 +525,34 @@ def _get_query_settings_from_config(
...
     if override_prefix:
         for k, v in all_confs.items():
-            if k.startswith(f"{override_prefix}/query_settings/"):
+            if k.startswith(f"{override_prefix}/{settings_key}/"):
                 clickhouse_query_settings[k.split("/", 2)[2]] = v
 
     return clickhouse_query_settings
 
 
+def _get_parallel_read_settings_from_config(
```
evanh, line 539: I don't think you really need this function. The `override_prefix` in snuba is actually the storage name (e.g. `transactions`). If you set `transactions/query_settings/max_parallel_replicas=4` it would only apply that setting to that storage. I believe that's what you want?

dmajere: oh, didn't know that. hm, this means that no code change might be necessary at all. well, except adding a stat for visibility.

dmajere: Ah, no.
Seems that prefix is configured for cluster (and actually seems to be never set) 
and reader inherits it
https://github.com/getsentry/snuba/blob/1b693962ee0bfaea0fdc52c5bc0db80efa0ca0e0/snuba/clusters/cluster.py#L278

https://github.com/getsentry/snuba/blob/1b693962ee0bfaea0fdc52c5bc0db80efa0ca0e0/snuba/clusters/cluster.py#L382

dmajere: @evanh But it seems like a good simplification, can I change what is passed as prefix to dataset?

snuba/web/db_query.py
dmajere: done

PR #4332, opened 2023-06-12
feat(errors): process trace_sampled and num_processing_errors from event

barkbarkimashark:
Related to this https://github.com/getsentry/snuba/pull/4276

Extracts `trace_sampled` from the error event `data.context.trace` map.
Extracts `num_processing_errors` from the error event `data.errors` list of processing errors.

A follow-up PR will be created to expose these two new columns in the entity and discover.

### DIFF
--- a/snuba/datasets/processors/errors_processor.py
+++ b/snuba/datasets/processors/errors_processor.py
@@ -146,8 +146,6 @@ class ErrorsProcessor(DatasetMessageProcessor):
         processing_errors = data.get("errors", None)
         if processing_errors is not None and isinstance(processing_errors, list):
             processed["num_processing_errors"] = len(processing_errors)
-        # else:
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/errors_processor.py
```diff
@@ -143,6 +143,12 @@ def process_insert(
 
         self.extract_stacktraces(processed, stacks, threads)
 
+        processing_errors = data.get("errors", None)
+        if processing_errors is not None and isinstance(processing_errors, list):
+            processed["num_processing_errors"] = len(processing_errors)
+        # else:
```
evanh, line 149: ?

evanh (commented):
Remove comments and fix CI and this should be good to go.

evanh (approved):

PR #4334, opened 2023-06-13
feat(dlq): DLQ topics API for admin

lynnagara:
Moves the hardcoded DLQ topics out of the admin UI and generate them from storage config definitions

### DIFF
--- a/tests/admin/dead_letter_queue/test_dlq.py
+++ b/tests/admin/dead_letter_queue/test_dlq.py
@@ -1,0 +1,5 @@
+from snuba.admin.dead_letter_queue import get_dlq_topics
+
+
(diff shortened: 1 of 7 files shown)


tests/admin/dead_letter_queue/test_dlq.py
```diff
@@ -0,0 +1,5 @@
+from snuba.admin.dead_letter_queue import get_dlq_topics
+
+
+def test_dlq() -> None:
+    assert len(get_dlq_topics()) == 7
```
evanh, line 5: I would probably test the structure of the response rather than this number. Otherwise this test will fail every time a new DLQ is added.

PR #4335, opened 2023-06-13
fix: All consumers have unique DLQ topics

lynnagara:
All Snuba consumers should have independent DLQ topics even if they share a source input topic so that a failure in one can be handled without affecting the other.

This change requires these topics to be created before merging.


untitaker (approved):
makes sense. in this particular scenario it is not necessary, but i believe for eventstream or outcomes it would have to be separate topics. so coupling one dlq topic to one consumer group makes sense

PR #4336, opened 2023-06-13
ref(gen_met): Add mapping optimizer to generic metrics

evanh:
Modify the MappingOptimizer and the patterns it uses to match against tag
conditions with integer keys, so that it can be used with the generic metrics
datasets.

Add the MappingOptimizer to the generic metrics storage configs.


snuba/datasets/configuration/generic_metrics/storages/counters.yaml
```diff
@@ -71,5 +71,11 @@ allocation_policy:
       org_limit_bytes_scanned: 100000
 
 query_processors:
+  - processor: MappingOptimizer
+    args:
+      column_name: tags
+      hash_map_name: _raw_tags_hash
+      killswitch: generic_metrics/tags_hash_map_enabled
+      value_subcolumn_name: raw_value
```
nikhars, line 79: Counters doesn't have the hash map column yet so i would remove this for now. The intention is to verify this works on one dataset and then apply it to others.

snuba/datasets/configuration/generic_metrics/storages/sets.yaml
```diff
@@ -71,5 +71,11 @@ allocation_policy:
       org_limit_bytes_scanned: 100000
 
 query_processors:
+  - processor: MappingOptimizer
+    args:
+      column_name: tags
+      hash_map_name: _raw_tags_hash
+      killswitch: generic_metrics/tags_hash_map_enabled
+      value_subcolumn_name: raw_value
```
nikhars, line 79: Same comment about this column not being available yet. We can remove this for now

nikhars (approved):

PR #4342, opened 2023-06-13
ref(capman): Move is_active check to base allocation policy

enochtangg:
### Overview
Minor refactor to move `is_active` logic to be contained within the Allocation Policy base class. This makes it such that any new policies being added would not need to reimplement is_active logic.

### DIFF
--- a/tests/query/allocation_policies/test_allocation_policy_base.py
+++ b/tests/query/allocation_policies/test_allocation_policy_base.py
@@ -17,6 +17,7 @@ from snuba.query.allocation_policies import (
     QuotaAllowance,
 )
 from snuba.state import set_config
+from snuba.web import QueryResult
 
 
 def test_eq() -> None:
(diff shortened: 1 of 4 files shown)


volokluev (approved):

volokluev (changes requested):
Add a test plz!

tests/query/allocation_policies/test_bytes_scanned_window_allocation_policy.py
```diff
@@ -114,6 +119,9 @@ def test_killswitch(policy: AllocationPolicy) -> None:
         ),
     )
     allowance = policy.get_quota_allowance(tenant_ids)
+    # assert that child methods were not called
+    mock_get_quota_allowance._update_quota_balance.assert_not_called()
```
volokluev, line 123: there's a test for the base class, you're changing functionality in the base class. Please put it there.

tests/query/allocation_policies/test_allocation_policy_base.py
```diff
@@ -385,3 +386,35 @@ def test_bad_defaults() -> None:
         SomeParametrizedConfigPolicy(
             StorageKey("some_storage"), [], {"my_param_config": False}
         )
+
+
+@pytest.mark.redis_db
+@mock.patch("snuba.query.allocation_policies.AllocationPolicy")
```
volokluev, line 392: You're mocking the base class (which is under test) in the test for the base class. It's hard for me to understand how this test works or why it works. I think it would be much easier to do the following:

1. Make a policy for this test that errors on `_get_quota_allowance` and `_update_quota_balance` 
2. Make the policy inactive
3. see that you can run the inactive policy and it won't raise

For an example see `test_passes_through_on_error`

enochtangg: By mocking the base class, we can assert if a method of the class (`_get_quota_allowance` and `_update_quota_balance`) is called inside the method we are testing. The goal is to ensure that these private methods (which are overridden by the child policies) are not called when the policy is inactive which this test checks. Does that make sense?

volokluev: but if you're mocking the whole base class, why does any of the functionality of the base class work?

enochtangg: Ah you're right, I'm mocking the base class not the sub-class and since we've moved the test outside the sub-class (to the base) it doesn't make sense anymore. I've changed the test to what you've have suggested.

volokluev (approved):
lit

PR #4347, opened 2023-06-15
feat(dlq): Redesign admin UI, add API endpoints for replaying

lynnagara:
Tried to make it look a bit cleaner/minimalistic. Added ability to get/post/delete the replay instruction.

Screenshot 2023-06-14 at 5 02 54 pm


untitaker (approved):
this seems to work!

one thing i noticed is that when I replay from an empty topic, the consumer gets stuck because it just polls forever

Now I wonder whether putting a limit on the `_run_once` calls might have been a good idea actually

PR #4349, opened 2023-06-15
fix(admin): Pin yarn and node using volta

untitaker:
With the way we manage yarn and node at sentry, you'd otherwise get yarn
v2, which will create a bunch of new files in the admin folder.

I wish we could deduplicate the volta pins with the pins in Docker like
we do in Sentry, but volta does not support Linux ARM.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         name: Checkout code
+      - uses: volta-cli/action@v4
       - name: Set up and run tests through yarn
         run: cd snuba/admin && yarn install && yarn run test --coverage
       - name: Upload to codecov


snuba/admin/package.json
```diff
@@ -29,5 +29,9 @@
     "ts-jest": "^29.0.5",
     "webpack": "^5.74.0",
     "webpack-cli": "^4.10.0"
+  },
+  "volta": {
+    "node": "19.9.0",
+    "yarn": "1.22.19"
```
lynnagara, line 35: How are these versions chosen? Is that what we run in CI?

untitaker: no idea what is being run in CI, i presume a completely random version. I added the volta github action so this becomes the source of truth

lynnagara (approved):

PR #4353, opened 2023-06-15
fix(spans): set parent_span_id to 0 if missing

dbanda:
set parent_span_id to 0 if missing, log and fail gracefully if span_id is missing


nikhars (approved):

evanh (approved):
This might seem weird, but I would consider making the `parent_span_id` equal to its own span_id itself instead of 0. It sort of depends on how (or if) we want to build the span trees.

I'm going to approve this anyways because I know it's an on-going bug.

PR #4358, opened 2023-06-16
feat(admin-auth): Add TTL cache for admin roles

davidtsuk:
Currently, we set up a user's roles every time we make a request which is inefficient as it requires reading from a file and making multiple API requests to google. In order to improve performance, I've added a TTL cache with a 10 min TTL. Since it's unlikely for roles and groups to change, we should consider increasing this TTL in the future.

### DIFF
--- a/snuba/admin/auth.py
+++ b/snuba/admin/auth.py
@@ -45,7 +45,7 @@ def _is_member_of_group(user: AdminUser, group: str) -> bool:
     return google_api.check_group_membership(group_email=group, member=user.email)
 
 
-def get_iam_roles_from_file(user: AdminUser) -> Sequence[str]:
+def get_iam_roles_from_user(user: AdminUser) -> Sequence[str]:
     iam_roles = []
     try:
         with open(settings.ADMIN_IAM_POLICY_FILE, "r") as policy_file:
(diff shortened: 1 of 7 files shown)


volokluev (commented):
This looks like it's caching the reading of the file (which shouldn't be that slow). Do we have evidence that this is the thing that needs to be cached? Is it caching the API calls somewhere that I don't see?

snuba/admin/auth.py
```diff
@@ -68,7 +72,14 @@ def get_iam_roles_from_file(user: AdminUser) -> Sequence[str]:
 def _set_roles(user: AdminUser) -> AdminUser:
     # todo: depending on provider convert user email
     # to subset of DEFAULT_ROLES based on IAM roles
-    iam_roles = get_iam_roles_from_file(user)
+    key = f"roles-{user.email}"
```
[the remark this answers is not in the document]
davidtsuk: I think the function name is a bit misleading here, but this function reads the file and also makes a call to check whether a member is in a group and that's where the API is being called.

snuba/admin/auth.py
```diff
@@ -68,7 +72,14 @@ def get_iam_roles_from_file(user: AdminUser) -> Sequence[str]:
...
+    iam_roles_str = redis_client.get(key)
+    iam_roles = rapidjson.loads(iam_roles_str) if iam_roles_str else None
+    if not iam_roles:
+        iam_roles = get_iam_roles_from_file(user)
+        redis_client.set(key, rapidjson.dumps(iam_roles))
+        redis_client.expire(key, settings.ADMIN_ROLES_REDIS_TTL)
+
```
volokluev, line 82: This implementation tightly couples authentication to caching. If your cache fails to update, you'll fail the request. this should be done in such a way that:

1. The distinction between cache and auth is clear
2. If the caching fails the app still works

snuba/admin/auth.py
```diff
@@ -68,7 +72,14 @@ def get_iam_roles_from_file(user: AdminUser) -> Sequence[str]:
 def _set_roles(user: AdminUser) -> AdminUser:
     # todo: depending on provider convert user email
     # to subset of DEFAULT_ROLES based on IAM roles
-    iam_roles = get_iam_roles_from_file(user)
+    key = f"roles-{user.email}"
+    iam_roles_str = redis_client.get(key)
+    iam_roles = rapidjson.loads(iam_roles_str) if iam_roles_str else None
+    if not iam_roles:
+        iam_roles = get_iam_roles_from_file(user)
+        redis_client.set(key, rapidjson.dumps(iam_roles))
```
volokluev, line 80: slight performance nit: [set already has an expiration parameter](https://redis.io/commands/set/)

davidtsuk:
Added additional tests and handling for redis failure.

snuba/redis.py
```diff
@@ -137,6 +138,9 @@ class RedisClientKey(Enum):
     RedisClientKey.OPTIMIZE: _initialize_specialized_redis_cluster(
         settings.REDIS_CLUSTERS["optimize"]
     ),
+    RedisClientKey.ADMIN_AUTH: _initialize_specialized_redis_cluster(
+        settings.REDIS_CLUSTERS["admin_auth"]
```
evanh, line 142: Is there a PR to add `admin_auth` to the prod settings as well?

davidtsuk: not yet, I'll make a PR right now

evanh: That will need to go out before this can be merged.

untitaker: this stuff is super inconvenient. I wonder if we can make it so that the prod_settings file can override individual keys of the default settings file, and does not have to redefine the entire variable

davidtsuk:
@evanh ops pr has been merged

evanh (approved):

PR #4359, opened 2023-06-16
hack(starfish): cardinality analyzer

volokluev:
This is a hacky tool which allows people to query the generic metrics distribution storage directly. Currently that ability only lies with specific people. In order to better understand how to navigate these challenges, the devs working on starfish need access to the data. 

This is mostly copy/paste of the querylog code

Compliance concerns:

* This tool can only query `generic_metrics_distributions` storage, it's enforced on the backend. This is a pretty low risk dataset for leaking personal information
* All queries are auditlogged
* @mcannizz said I could do it

image


enochtangg (approved):
As discussed offline, this is a temporary tool and will be reverted.

PR #4360, opened 2023-06-16
ref(sessions): Demote sessions migration group to deprecate

enochtangg:
The sessions storages are already set to `deprecate` and this migration group should follow. Note that session this to `deprecate` means that the migrations will only be executable locally and in self-hosted. These migrations will not be executable in any other sentry environments anymore (include ST).

### DIFF
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -107,7 +107,7 @@ _REGISTERED_MIGRATION_GROUPS: Dict[MigrationGroup, _MigrationGroup] = {
     MigrationGroup.SESSIONS: _MigrationGroup(
         loader=SessionsLoader(),
         storage_sets_keys={StorageSetKey.SESSIONS},
-        readiness_state=ReadinessState.COMPLETE,
+        readiness_state=ReadinessState.DEPRECATE,
     ),
     MigrationGroup.QUERYLOG: _MigrationGroup(
         loader=QuerylogLoader(),


volokluev (approved):

enochtangg:
@mwarkentin About 1-2 weeks, checked with ops that sessions was no longer used in ST. Do you mind confirming again?

mwarkentin:
That's my understanding, but I'm wondering if there's a way to confirm. Maybe query the Clickhouse table and see when latest data was written?

mwarkentin:
Leaving some notes on my investigation:

`sessions_raw_local` seems to have stopped being updated on `s4s` in April: 

```
mwarkentin@clickhouse-1:~$ sudo ls -la /var/lib/clickhouse/data/default/ | grep local
drwxr-x--- 414 clickhouse clickhouse  200704 Jun 19 14:07 errors_local
drwxr-x--- 349 clickhouse clickhouse  958464 Jun 19 14:07 outcomes_hourly_local
drwxr-x--- 725 clickhouse clickhouse 1146880 Jun 19 14:07 outcomes_raw_local
drwxr-x---  78 clickhouse clickhouse  434176 Jun 19 00:10 sessions_hourly_local
drwxr-x--- 167 clickhouse clickhouse  557056 Apr 28 00:41 sessions_raw_local
drwxr-x--- 694 clickhouse clickhouse  323584 Jun 19 14:07 transactions_local
```

`sessions_hourly_local` is still being updated but I suspect because that's a "materialized view" or similar that it just running against the raw table?

mwarkentin:
`ingest-sessions` showing all traffic stopped:

image

enochtangg:
Thanks for looking into this. It also looks like queries to the session dataset in ST stopped around Apr 18 - May 8.
image

enochtangg:
Discussed offline with @mwarkentin. Ingestion and querying to the sessions dataset stopped late April to early May. We no longer need to run migrations in ST for this dataset.

PR #4366, opened 2023-06-16
ref(db_query): Refactor db_query.py for readability and earlier cache hit results

rahul-kumar-saini:
### Overview / Motivation
- We want to bring rate limiting into Allocation Policies to unify different types of Capacity Management
- Currently, allocation policies are executed, then cache is read, then rate limits are applied
- Bringing rate limits up to the allocation policy level means there will inevitably be some amount of queries that will be rate-limited which could have been served from the cache
- Therefore, we need to refactor the system to read cache, then execute allocation policies, then apply rate limits
- This brings us to `db_query.py`:
  - It was horrendous to read and understand
  - Needed a full refactor so future or unfamiliar S<sup>3</sup> engineers do not have to go through what I did

### Changes
- `DBQuery` class refactor
  - This replaces the old `db_query()` function
  - New usage is `DBQuery(...).db_query()`
  - The existing code passes around a *lot* of variables into many sub functions and the execution looks like:
    - fn1 calls fn2 calls ... calls fnN which may early exit or call ... calls fnM which executes the query
  - The new version executes in more of a "tree" of methods than a linked list of functions
  - This should be significantly easier to read or trace by hand
  - Added some spans to methods for instrumentation
- Readthrough caching
  - The Cache class is no longer implementing the readthrough caching pattern, DBQuery.db_query() is instead
  - The new db_query method now checks for a result in the cache first, then executes everything else one by one
  - Renamed a method of the Cache class as it aren't technically readthrough but can be readthrough when methods are used together

### Reviewing tips
- This is a critical file to be manipulating - carefully review the logic in `db_query.py` on it's own and compare to the old version
  - Viewing a diff for this file will likely be much more confusing than understanding the previous logic and reading the file for new logic directly - maybe even check out this branch to review rigorously
- If you're familiar with or have written any of the tests I've modified, please ensure my changes make sense!


snuba/state/cache/redis/scripts/get.lua
```diff
@@ -1,25 +1,15 @@
...
+local wait_queue_key = KEYS[1]
+local task_id_key = KEYS[2]
 local task_timeout = ARGV[1]
 local task_id = ARGV[2]
 
-local CODE_RESULT_VALUE = 0
 local CODE_RESULT_EXECUTE = 1
 local CODE_RESULT_WAIT = 2
 
--- Check to see if a value already exists at the result key. If one does, we
```
evanh, line 16: I'm not sure I understand why removing this is necessary.

evanh (commented):
Generally this seems pretty clean. My questions are about the caching changes.

snuba/state/cache/redis/backend.py
```diff
@@ -73,7 +73,25 @@ def set(self, key: str, value: TValue) -> None:
...
+        key: str,
+        timer: Optional[Timer] = None,
+    ) -> Optional[TValue]:
+        if get_config("read_through_cache.short_circuit", 0):
+            return None
+
+        if timer is not None:
+            timer.mark("cache_get")
+
+        result = self.get(key)
```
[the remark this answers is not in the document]
rahul-kumar-saini: Yep, previously on cache hit there was that logger.debug statement so I wanted to preserve the same logging to minimize functional changes. If this wasn't a useful log to begin with I can remove and simply `return self.get(key)`

snuba/state/cache/redis/scripts/get.lua
rahul-kumar-saini: Previously, `execute_with_readthrough_caching` was responsible for:
- run get.lua which does:
  - check cache 
  - run value generation if cache miss
  - store cache

Now that we do the cache check in the very beginning in `DBQuery.db_query()`, we only get to executing this script if the value did not exist already. So the "readthrough" pattern is on the DBQuery class itself instead of within this script.
New order of ops is:

db_query():
- check cache

if cache miss:

_execute_and_cache_query():
- run get.lua which does:
  - run value generation
  - store cache

snuba/state/cache/redis/backend.py
```diff
@@ -87,14 +105,12 @@ def get_readthrough(
             return function()
 
         # This method is designed with the following goals in mind:
-        # 1. The value generation function is only executed when no value
```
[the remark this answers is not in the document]
rahul-kumar-saini: I think my comment on the get.lua script diff should cover this question as well.

snuba/state/cache/redis/scripts/get.lua
evanh: `get.lua` does not do the value generation. Nor does it store the cache. The "value generation" is the actual Clickhouse query execution, and the cache storage happens in `set.lua`.

The get script
- Checks the cache
- Checks to see how many people are waiting on the wait queue
- If there is no one else besides the current query waiting, set a marker showing that the query is being executed and return
- Otherwise return the task timeout of the executing task

After this the "value generation" is run, which is a fancy word for the Clickhouse query. Then the `set.lua` script is run, which updates all the various keys.

PR #4376, opened 2023-06-20
Update settings loader

davidtsuk:
Currently, dictionaries defined in the default settings file are fully overwritten by prod/test settings files if the variable is redefined. This change allows specific keys in the dictionary to be overwritten, so if a key is added to the default settings file, it doesn't necessarily have to be added to other settings files.

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -424,8 +424,12 @@ def _load_settings(obj: MutableMapping[str, Any] = locals()) -> None:
             )
 
         for attr in dir(settings_module):
+            new_attr = getattr(settings_module, attr)
             if attr.isupper():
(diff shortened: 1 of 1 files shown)


volokluev (commented):
add a test please

untitaker:
in sentry we do it like this:


```
from default_settings import *

KAFKA_CLUSTERS["foo"] = <overridden value>
```

I think this is a bit nicer because it plays well with type checking

wdyt?

davidtsuk:
> in sentry we do it like this:
> 
> ```
> from default_settings import *
> 
> KAFKA_CLUSTERS["foo"] = <overridden value>
> ```
> 
> I think this is a bit nicer because it plays well with type checking
> 
> wdyt?

oh interesting, I agree that this is nicer

PR #4381, opened 2023-06-20
feat(snubsplain): Use request context to capture query processors

evanh:
This change is a framework for how meta information about the query execution
will be captured and surfaced through the tool. It starts with one simple data
point, the entity query processors that ran. This can be expanded in subsequent
PRs to add more information.

Since the explain functionality is very similar to what the SnQL-to-SQL tool
already does, the two endpoints were collapsed. The tools are still separate,
and use different parts of the output. In the future the tools can be
 consolidated.

The meta module that was added is the main interface for capturing the query
steps. The data is stored as a list of steps, to preserve the order of execution
during the pipeline.

Changes:
- Add a meta module to provide an interface for capturing steps during query
execution.
- Setup the meta capture based on the `dry_run` flag using the existing
  snql-to-sql tool


rahul-kumar-saini (approved):
Tested locally - I can see a list of entity processors in the response 👍

snuba/state/explain_meta.py
```diff
@@ -0,0 +1,57 @@
...
+
+        g.explain_meta.add_step(step)
+    except RuntimeError:
+        # Code is executing outside of a flask context
+        return
+
+
+def get_explain_meta() -> ExplainMeta | None:
+    try:
+        if hasattr(g, "explain_meta"):
```
volokluev, line 43: isn't `g` a global context between requests? I think you want `from flask import request`

untitaker: `g` is scoped per request. patching stuff on `request` also works but is not recommended. `session` persists between requests.

PR #4385, opened 2023-06-20
feat(search-issues): Add message column to search issues

wedamija:
This adds the `message` column to search issues, since we missed including it a column initially.

Related to https://github.com/getsentry/sentry/issues/50345

PR #4387, opened 2023-06-21
feat(search-issues): Start writing message column in search issues

wedamija:
This starts writing to the new `message` column. `message` is already being sent from sentry, so we just need to start writing it here.

I will let this write for a week or two before switching over to use it. We don't need the full 90 days of data for this to be effective.

Relies on https://github.com/getsentry/snuba/pull/4385
Related to https://github.com/getsentry/sentry/issues/50345

### DIFF
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -9,6 +9,7 @@ schema:
     { name: group_id, type: UInt, args: { size: 64 } },
     { name: search_title, type: String },
     { name: resource_id, type: String },
+    { name: message, type: String },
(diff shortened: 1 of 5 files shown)


evanh:
I think you need to remove the `__init__` file in the migrations folder to have the DDL changes CI pass.

> I will let this write for a week or two before switching over to use it.

Is this true for single tenant and self hosted as well?

wedamija:
> I think you need to remove the `__init__` file in the migrations folder to have the DDL changes CI pass.
> 
> > I will let this write for a week or two before switching over to use it.
> 
> Is this true for single tenant and self hosted as well?

Fortunately it's not released there :)

evanh (approved):

PR #4388, opened 2023-06-21
show nodes migration is applied on

dbanda:
Log what nodes we are applying migrations on to help with debugging.

### DIFF
--- a/snuba/migrations/operations.py
+++ b/snuba/migrations/operations.py
@@ -3,11 +3,15 @@ from abc import ABC, abstractmethod
 from enum import Enum
 from typing import Any, Callable, Mapping, Optional, Sequence, Tuple
 
+import structlog
+
 from snuba.clickhouse.columns import Column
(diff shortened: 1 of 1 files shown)


evanh (approved):

snuba/migrations/operations.py
```diff
@@ -49,12 +53,13 @@ def execute(self) -> None:
...
             raise ValueError(f"Target not set for {self}")
-
+        if nodes:
+            logger.info(f"Executing op: {self.format_sql()[:32]}...")
         for node in nodes:
             connection = cluster.get_node_connection(
                 ClickhouseClientSettings.MIGRATE, node
             )
-
+            logger.info(f"Executing on node: {node}")
```
MeredithAnya, line 62: @dbanda do you think it would be worth adding whether it's a local or dist node here as well? i.e `Executing on local node:...`

dbanda: added

PR #4390, opened 2023-06-21
ci: Fail faster

lynnagara:
This sometimes just gets stuck and times out. Let's fail faster.


dbanda:
Maybe we should also add `pytest -x` so that we fail on first error.

untitaker (approved):

PR #4391, opened 2023-06-21
feat(search-issues): Make `message` column available for querying

wedamija:
This allows the message column on search issues to be queried, and stops using `issue_title`. We won't merge this until we've dual written data for a week or two.

Relies on https://github.com/getsentry/snuba/pull/4385, https://github.com/getsentry/snuba/pull/4387 Related to https://github.com/getsentry/sentry/issues/50345

### DIFF
--- a/snuba/datasets/configuration/issues/entities/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/entities/search_issues.yaml
@@ -84,12 +84,6 @@ storages:
         - mapper: ColumnToColumn
           args:
             from_table_name:
-            from_col_name: "message"
-            to_table_name:
-            to_col_name: "search_title"
-        - mapper: ColumnToColumn
-          args:
-            from_table_name:
             from_col_name: "title"
             to_table_name:
             to_col_name: "search_title"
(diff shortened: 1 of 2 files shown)


evanh (approved):

wedamija:
We started dual writing today, can merge in a couple weeks

wedamija:
> I think it's possible to add a query processor that falls back to the other column. So you can switch over immediately and keep queries simpler.

Oh like a coalesce or something along those lines?

lynnagara:
yeah, something like that. Pretty sure you can use the ColumnToFunction mapper than map the `messages` column to something like `coalesce(nullif(message, ""), "search_title")` @evanh / search team might have more opinions on it though.

evanh:
> Pretty sure you can use the ColumnToFunction mapper to map the messages column to something like coalesce(nullif(message, ""), "search_title")

That isn't currently supported in the YAML right now, so you'd need to write a custom query processor and add it to the config that way. If that's something you want to do we can help with that. As long as it gets removed when the cutover is complete.

PR #4394, opened 2023-06-21
fail tests on first error

dbanda:
fails pytest on the first error


lynnagara (approved):
💯

lynnagara:
Can we add this for the rest of the tests too?

PR #4396, opened 2023-06-21
ref(generic-metrics): Remove 10 seconds granularity

john-z-yang:
### Overview
Create migrations for materialized view of sets, counters, and distributions to remove granularity of 10 seconds.

### Testing
Ran snuba locally with migrations applied and sent 9 messages in total, a single (set, counter, distribution) kafka payload for 3 use case IDs.

```
SELECT
    use_case_id,
    org_id,
    project_id,
    metric_id,
    granularity
FROM generic_metric_sets_local
WHERE arrayExists(v -> match(v, 'gen_metric_e2e_.*FJ1D6416'), tags.raw_value)
UNION ALL
SELECT
    use_case_id,
    org_id,
    project_id,
    metric_id,
    granularity
FROM generic_metric_distributions_aggregated_local
WHERE arrayExists(v -> match(v, 'gen_metric_e2e_.*FJ1D6416'), tags.raw_value)
UNION ALL
SELECT
    use_case_id,
    org_id,
    project_id,
    metric_id,
    granularity
FROM generic_metric_counters_aggregated_local
WHERE arrayExists(v -> match(v, 'gen_metric_e2e_.*FJ1D6416'), tags.raw_value)

Query id: 3554e50f-e92d-4333-949e-4c8e00e36cef

┌─use_case_id──┬─org_id─┬─project_id─┬─metric_id─┬─granularity─┐
│ transactions │      1 │          3 │     65551 │           1 │
│ transactions │      1 │          3 │     65551 │           2 │
│ transactions │      1 │          3 │     65551 │           3 │
│ spans        │      1 │          3 │     65559 │           1 │
│ spans        │      1 │          3 │     65559 │           2 │
│ spans        │      1 │          3 │     65559 │           3 │
│ hello        │      1 │          3 │     65641 │           1 │
│ hello        │      1 │          3 │     65641 │           2 │
│ hello        │      1 │          3 │     65641 │           3 │
└──────────────┴────────┴────────────┴───────────┴─────────────┘
┌─use_case_id──┬─org_id─┬─project_id─┬─metric_id─┬─granularity─┐
│ transactions │      1 │          3 │     65553 │           1 │
│ transactions │      1 │          3 │     65553 │           2 │
│ transactions │      1 │          3 │     65553 │           3 │
│ spans        │      1 │          3 │     65556 │           1 │
│ spans        │      1 │          3 │     65556 │           2 │
│ spans        │      1 │          3 │     65556 │           3 │
│ hello        │      1 │          3 │     65643 │           1 │
│ hello        │      1 │          3 │     65643 │           2 │
│ hello        │      1 │          3 │     65643 │           3 │
└──────────────┴────────┴────────────┴───────────┴─────────────┘
┌─use_case_id──┬─org_id─┬─project_id─┬─metric_id─┬─granularity─┐
│ transactions │      1 │          3 │     65546 │           1 │
│ transactions │      1 │          3 │     65546 │           2 │
│ transactions │      1 │          3 │     65546 │           3 │
│ spans        │      1 │          3 │     65562 │           1 │
│ spans        │      1 │          3 │     65562 │           2 │
│ spans        │      1 │          3 │     65562 │           3 │
│ hello        │      1 │          3 │     65649 │           1 │
│ hello        │      1 │          3 │     65649 │           2 │
│ hello        │      1 │          3 │     65649 │           3 │
└──────────────┴────────┴────────────┴───────────┴─────────────┘

```

### DIFF
--- a/snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
+++ b/snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
@@ -16,7 +16,7 @@ from snuba.migrations.operations import OperationTarget
 
 class Migration(migration.ClickhouseNodeMigration):
     blocking = False
-    view_name = "generic_metric_sets_aggregation_mv"
+    view_name = "generic_metric_sets_aggregation_mv_2"
     dest_table_columns: Sequence[Column[Modifiers]] = [
         Column("org_id", UInt(64)),
         Column("project_id", UInt(64)),
@@ -81,4 +81,10 @@ class Migration(migration.ClickhouseNodeMigration):
         ]
 
     def backwards_ops(self) -> Sequence[operations.SqlOperation]:
-        return []
+        return [
+            operations.DropTable(
+                storage_set=self.storage_set_key,
+                table_name=self.view_name,
+                target=OperationTarget.LOCAL,
+            )
+        ]
--- a/snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
+++ b/snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
@@ -1,0 +1,84 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import (
+    AggregateFunction,
+    Column,
+    DateTime,
+    Nested,
+    String,
+    UInt,
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    view_name = "generic_metric_sets_aggregation_mv"
+    dest_table_columns: Sequence[Column[Modifiers]] = [
+        Column("org_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("granularity", UInt(8)),
+        Column("timestamp", DateTime(modifiers=Modifiers(codecs=["DoubleDelta"]))),
+        Column("retention_days", UInt(16)),
+        Column(
+            "tags",
+            Nested(
+                [
+                    ("key", UInt(64)),
+                    ("indexed_value", UInt(64)),
+                    ("raw_value", String()),
+                ]
+            ),
+        ),
+        Column("value", AggregateFunction("uniqCombined64", [UInt(64)])),
+        Column("use_case_id", String(Modifiers(low_cardinality=True))),
+    ]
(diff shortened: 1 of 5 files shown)


snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
```diff
@@ -0,0 +1,84 @@
...
+)
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    view_name = "generic_metric_sets_aggregation_mv"
```
nikhars, line 19: The view name should encode the version otherwise you wouldn't be creating this view now since the generated SQL has an `IF EXISTS` clause.

You should change this to something like `generic_metric_sets_aggregation_mv_v2`. That way you will have both the original materialization view and the new materialization view.

snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
```diff
@@ -0,0 +1,84 @@
...
+                    tags.raw_value,
+                    timestamp,
+                    granularity,
+                    retention_days
+                """,
+            ),
+        ]
+
+    def backwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return []
```
nikhars, line 84: Shouldn't we drop the materialization view which we are creating above in the backward ops?

snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
john-z-yang: Fixed, thank you!

snuba/snuba_migrations/generic_metrics/0014_sets_mv_2.py
john-z-yang: Fixed, thank you!

onewland (approved):

john-z-yang:
Updated the pr to remove the `0` from granularities in the consumer, let me know if this needs to be split up. One of the action is complaining about `Migration changes are coupled with other changes`

evanh (commented):
I don't think the migration plan makes sense here.

If you have to change the processor in the same PR as the migration, then that means there will be a period of time where the processor is broken.

john-z-yang:
> I don't think the migration plan makes sense here.
> 
> If you have to change the processor in the same PR as the migration, then that means there will be a period of time where the processor is broken.

I think I'm going to close this due to some additional changes I like to make to the matview.

But quick (and pedantic) question, would this really break? Wouldn't this need another pr to set the materialized version in the processor to the one specified in the migration for the matview to kick in?

PR #4398, opened 2023-06-22
feat(prod-queries): Implement backend for prod queries

davidtsuk:
**Frontend Changes**
I revamped the UI a bit to take in SnQL queries with a dataset and show the most recent result. I plan to add support for result history in the future, but right now I'm just keeping it simple.

**Backend Changes**
I refactored the audit log decorator a bit so that I can use it for prod queries. I also added an api endpoint for snql queries. In terms of validation, I first validate the query by using a dry run and then validate the projects in the query.

**Testing**
I added several tests for the api to ensure that it works correctly and handles errors.


snuba/admin/static/api_client.tsx
```diff
@@ -200,6 +201,24 @@ function Client() {
...
+      return fetch(url, {
+        headers: { "Content-Type": "application/json" },
+        method: "POST",
+        body: JSON.stringify(query),
+      }).then((res) => {
+        if (res.ok) {
+          return Promise.resolve(res.json());
+        } else {
+          return res.json().then((err) => {
+            let errMsg = err?.error.message || "Could not convert SnQL";
```
evanh, line 215: "Could not convert SnQL" is probably not the right error message.

evanh (approved):

snuba/admin/static/api_client.tsx
davidtsuk: lol good catch

PR #4399, opened 2023-06-22
fix(validation): EntityContainsColumnValidator wasn't actually enabled

rahul-kumar-saini:
### Overview
- Although by default we specify `ColumnValidationMode.ERROR` for the `EntityContainsColumnValidator` on the `PluggableEntity`, we overwrite this value to `ColumnValidationMode.DO_NOTHING` upon reading an entity config that doesn't explicitly contain the validator
- This means there is still no entity column validation happening in prod!
- We want the default behaviour to be ERROR and this PR modifies how the validation mode is initialized


evanh (approved):

0Calories:
🔥

evanh (approved):

enochtangg:
I made the following changes:
* Set default entity column validator to `WARN`. 
* Updated all units tests which used improper/arbitrary column names
* Added `use_case_id` and `granularity` to entity schema of both metrics and generic_metrics datasets
* Added `group_raw` column to spans storage and entity.

evanh (approved):

PR #4403, opened 2023-06-22
fix(CapMan): Allocation Policies causing potentially timeout errors on ST

rahul-kumar-saini:
### Overview
- Currently there is some strong correlation between the latest query errors and timeouts on Disney ST and Allocation Policy throttles on the instance
- Disney ST max threads are 30 by default, the throttling sets the queries to 1 thread - a massive difference
- I don't believe we have a huge load problem on ST so we can probably disable the policy enforcement altogether
- Ideally this is done through Snuba Admin but since that does not yet exist for ST this is a good enough solution

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -395,7 +395,10 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
 
     @property
     def is_enforced(self) -> bool:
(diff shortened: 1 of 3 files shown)


evanh (approved):
Does this setting need to be enabled in prod_settings as well?

volokluev (commented):
This is a hack that is necessary because single tenant does not have snuba-admin deployed so we can't change the configs there. Please leave a comment explaining that this is a hack and at what point it should be removed.

PR #4404, opened 2023-06-22
test(spans): add some validation to the test json

dbanda:
Add some type checking to the spans tests to make sure our test data schemas make sense.

### DIFF
--- a/tests/datasets/test_spans_processor.py
+++ b/tests/datasets/test_spans_processor.py
@@ -1,9 +1,11 @@
+import json
 import uuid
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Any, Mapping, Optional, Sequence, Tuple
 
 import pytest
+from sentry_kafka_schemas import get_codec
 from sentry_relay.consts import SPAN_STATUS_NAME_TO_CODE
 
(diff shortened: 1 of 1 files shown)


lynnagara:
Have you seen https://github.com/getsentry/snuba/blob/ce8470829bc683345695bcd3ef26492757727c14/tests/consumers/test_schemas.py?

It attempts to run all message processors against the example payloads in sentry-kafka-schemas, including spans. Not sure if this overlaps with the intention here, or this targets something different.

dbanda:
> Have you seen https://github.com/getsentry/snuba/blob/ce8470829bc683345695bcd3ef26492757727c14/tests/consumers/test_schemas.py?
> 
> It attempts to run all message processors against the example payloads in sentry-kafka-schemas, including spans. Not sure if this overlaps with the intention here, or this targets something different.

Does this require that we put stuff in the examples folder? the only examples we have so far are the ones in this test, but I can craft some if that helps.

lynnagara:
> Does this require that we put stuff in the examples folder? the only examples we have so far are the ones in this test, but I can craft some if that helps.

Yeah it does. If there's an example message that looks different to what's already covered, i think that would be good

PR #4405, opened 2023-06-23
feat: A small step towards schema enforcement

lynnagara:
Devserver now enforces the schema for the events topic


untitaker (approved):

snuba/consumers/consumer.py
```diff
@@ -573,6 +574,8 @@ def process_message(
                         _LAST_INVALID_MESSAGE[snuba_logical_topic.name] = start
                         sentry_sdk.set_tag("invalid_message_schema", "true")
                         logger.warning(err, exc_info=True)
+                    if enforce_schema:
+                        raise
```
dbanda, line 578: What will this raise?

lynnagara: It will reraise the ValidationError https://github.com/getsentry/sentry-kafka-schemas/blob/b1385c004934060cb329a9e47a5234dbd53e05b9/python/sentry_kafka_schemas/codecs/__init__.py#L36C1-L41

lynnagara:
Nice test failure, tells us our tests are outdated :)

snuba/cli/consumer.py
```diff
@@ -116,6 +116,13 @@
     type=int,
 )
 @click.option("--join-timeout", type=int, help="Join timeout in seconds.", default=5)
+@click.option(
+    "--enforce-schema",
+    type=bool,
+    is_flag=True,
+    default=False,
```
ayirr7, line 123: Maybe a silly question, but why is this False by default?

lynnagara: False is the current behaviour. Eventually we'll want enforcement to be default but we're slow rolling to reduce risk: enforcing a schema that isn't guaranteed to be correct might mean messages get incorrectly dropped.

ayirr7: That makes sense. Thanks!

PR #4419, opened 2023-06-26
ref: Emit more specific FINAL causes, log specific replacement types

untitaker:
It's most likely unmerge replacement, but it's not definitive.


untitaker:
I got confused. It seems we are already tracking this as one-tag-per-replacement

PR #4424, opened 2023-06-27
Add GoCD role to migrations stage

dbanda:
Adding a role permission to the migration stage. This will limit the migration stage access to only people in the `role-deploy-snuba-operator` google group. Ideally, this group should have been called `role-deploy-snuba-migrations-operator` but it's hard to change names on the dev-infra GoCD side.

### DIFF
--- a/gocd/pipelines/snuba.yaml
+++ b/gocd/pipelines/snuba.yaml
@@ -191,6 +191,9 @@ pipelines:
             - migrate:
                   approval:
                       type: manual
+                      allow_only_on_success: true
+                      roles:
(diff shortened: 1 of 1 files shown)


mattgauntseo-sentry:
Going to hold off on approving - working with @bukzor-sentryio to fix this, I tried deploying gocd to get the updated roles, but it didn't work.

mattgauntseo-sentry (approved):

mattgauntseo-sentry:
I've rolled out the updates to deploy.getsentry.net, so this should be good to go.

MeredithAnya (approved):

PR #4430, opened 2023-06-27
feat(admin): Add syntax highlighting to snuba admin

enochtangg:
### Overview
This PR is responsible for adding SQL syntax highlighting for all query tools in Snuba admin. To accomplish this, the typescript based, open source [Mantine](https://mantine.dev/) react component library along with [Prism](https://github.com/FormidableLabs/prism-react-renderer) was used to reorganize text boxes and highlight code.

Note that Mantine does contain a rich text editor that [highlights code in-place](https://mantine.dev/others/tiptap/#code-highlight), however, the highlighter breaks for ClickHouse SQL queries which contains numbers as parameter of a function. As a result, the Prism library was used to highlight code in a separate box much like the querylog resultant query box (after variables are applied).

### Screenshots
Screenshot 2023-06-29 at 11 37 01 AM

Screenshot 2023-06-29 at 11 37 19 AM

### Tests
Tests cases were updated to reference the new react text editor components.


enochtangg:
> From the screenshots, it looks like the styling for the querylog and tracing editors is different. Is it possible to make the editors have the same styling?

I think this because I copied the query from a styled text editor. Typing or copying unformatted text seems to produce similar styles.

davidtsuk (approved):

evanh (commented):
I think we should standardize all our tools to use the `QueryEditor` class. However I think that should happen in a separate PR to this. My suggestion:

- Keep this PR, but only add highlighting to the `QueryEditor` class. Then the tools that use that component will get syntax highlighting
- Open a new PR that migrates all the other tools to use the `QueryEditor` class

enochtangg:
> I think we should standardize all our tools to use the `QueryEditor` class. However I think that should happen in a separate PR to this.

@evanh Sounds good to me. It would be nice to have the consistency and substring variable functionality in tracing tool. Updated the PR.

evanh (approved):

PR #4434, opened 2023-06-27
Use a single GoCD group and order pipelines

mattgauntseo-sentry:
This does *not* change the functionality of the piplines, it puts all the pipelines in one group and it defines an order for the pipelines.


lynnagara:
curious about this comment: `This does change the functionality of the piplines`. Any changes to call out?

mattgauntseo-sentry:
Sorry, thats a typo - was meant to say "does not* change the functionality of the pipelines"

lynnagara (approved):

PR #4437, opened 2023-06-28
feat(rust): Connect consumer to clickhouse

dbanda:
This is an attempt to get the consumers to actually write to clickhouse on submit. A challenge here was our clickhouse client is async, so I had to modify the ProcessingStrategy to use an async submit function. This change had ripple effects all across the board. 

The quick solution in many places where the submit function was called in sync functions was to use `block_on` but this would sometimes hang, so I avoided `block_on` and used `async await` + `tokio::spawn`. 

In some parts of the code, I had to replace mutexes with async-aware mutexes to make sure that we are not holding the locks synchronously in the submit async function.


rust_snuba/src/consumer.rs
```diff
@@ -14,53 +15,84 @@ use rust_arroyo::processing::StreamProcessor;
...
     }
 
-    fn submit(&mut self, message: Message<Vec<BytesInsertBatch>>) -> Result<(), MessageRejected> {
+    async fn submit(&mut self, message: Message<Vec<BytesInsertBatch>>) -> Result<(), MessageRejected> {
+        log::trace!("submitting clickhouse writer step");
         for batch in message.payload() {
             for row in batch.rows {
                 let decoded_row = String::from_utf8_lossy(&row);
-                log::debug!("insert: {:?}", decoded_row);
+                let res = self.clickhouse_client.send( decoded_row.to_string()).await;
```
untitaker, line 58: Spawning via https://docs.rs/tokio/latest/tokio/task/fn.spawn.html or https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.spawn would allow you to run a future without making all processing strategies async

dbanda: yes I considered it but the tricky part with spawn is it runs the futures at its own time and doesn't guarantee completion. My understanding is we want to move on to the next batch once we get an ack of the current batch written in Clickhouse. That way we don't commit and  then later on figure out the spawn job failed. This approach is slower, but has more guarantees that at commit time the data has been written to Clickhouse.

dbanda: Two issues with that

1. Tokio's JoinHandle still needs to be awaited on to actually join. So we still need async context for that.
2. Channels would work, but then you lose out on concurrency. The standard channel primitive will block the thread waiting for a response.

untitaker: no, you have `try_recv` and `recv_timeout` on the channel. I believe you can build this internally similarly to how RunTaskWithMultiprocessing in Python deals with the Future objects of the processing pool: Keep all those channels in a list ordered by offset/submission order, `recv_timeout(1 second)` the channels in the list in `Strategy.poll/join`, etc

dbanda: I'm not sure I understand the benefit of doing it that way. If we do `recv_timeout` we risk timing out right before the msg is sent to Clickhouse and report an incorrect failure. Using `try_recv` has the same issue. 

Keeping the channels in a list raises challenges with bookkeeping and spinning to check channels for msgs. I guess I'm struggling to understand why that would be better vs using async and having Tokio take care of all that for us. 

The limitation of the Tokio approach is that Kafka isn't async aware so the callbacks have to be run synchronously. To fix that, I'd spawn threads within the callbacks. Given that the callbacks are only executed infrequently, it shouldn't be too expensive.

untitaker: > If we do recv_timeout we risk timing out right before the msg is sent to Clickhouse and report an incorrect failure

PTAL at how the multiprocessing strategy works in Python. Yes we do a lot of bookkeeping there for pending futures, but this is not a concern. In every call to `Strategy.poll()`, we wait for a future with a timeout of 1 second, and if the future completes, take its result and forward it to the next strategy. In `join()` we wait for the future to complete indefinitely. I propose we do the same in Rust here, except we wait for channels instead of futures.

> Keeping the channels in a list raises challenges with bookkeeping and spinning to check channels for msgs. I guess I'm struggling to understand why that would be better vs using async and having Tokio take care of all that for us.

I want to ensure we do not introduce Tokio into the core arroyo architecture just because one strategy requires it. It's a significant amount of complexity on top of everything else, and even if we did introduce it, I would expect that each strategy has its own separate runtime either way

MeredithAnya:
looks like this was completed in https://github.com/getsentry/snuba/pull/4490

PR #4441, opened 2023-06-29
ref(admin): Upload sourcemaps to Sentry for the admin tool

evanh:
Add the webpack integration for Sentry into webpack, and setup the integration
key to be passed down into the build step.

NOTE: The secret has already been created for this.


.github/workflows/image.yml
```diff
@@ -42,6 +43,7 @@ jobs:
             --build-arg BUILDKIT_INLINE_CACHE=1 \
             --build-arg SHOULD_BUILD_ADMIN_UI="$SHOULD_BUILD_ADMIN_UI" \
             --build-arg SHOULD_BUILD_RUST=false \
+            --build-arg ADMIN_SOURCEMAP_KEY="$ADMIN_SOURCEMAP_KEY" \
```
asottile-sentry, line 46: as written this value will be in the image -- if it's actually secret this will leak the secret

evanh: I didn't think of that. How would you recommend doing this?

asottile-sentry: I think getsentry does this through the deploy pipeline -- I'm not entirely sure would have to check that

evanh: I was thinking that too, but our images are built during CI, and I need this value during the build steps I think.

asottile-sentry: is the value an actual secret? if not then we can just inline it

does it need to be in the source?

evanh: I think it should be a secret? It's an internal integration API key for our Sentry projects. It would allow anyone with the key at minimum to upload files to our internal projects (that's what I plan to do with it).

mwarkentin: Can we use `docker build --secret id=mysecret,env=MYSECRET .`?

asottile-sentry: I don't think that will help here -- the value is exposed in `ENV` and appears to get baked in by webpack

evanh: Maybe I can have a separate step that builds webpack outside of the docker image just to get the sourcemaps uploaded. I don't need this value after the image is built.

.github/workflows/admin-sourcemaps.yml
```diff
@@ -0,0 +1,22 @@
+on:
+  pull_request:
+  push:
+    branches: [master]
+jobs:
+  build:
+    name: "build sourcemaps"
+    runs-on: ubuntu-latest
+    env:
+      SENTRY_AUTH_TOKEN: ${{ secrets.SNUBA_SENTRY_SOURCEMAP_KEY }}
```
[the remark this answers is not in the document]
asottile-sentry: the team which owns the repo (I believe sns) should have admin to add / alter repository-specific secrets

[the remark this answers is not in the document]
evanh: Perfect, thank you for all your help!

volokluev (approved):

PR #4442, opened 2023-06-29
feat(consumer): Add transactions consumer SLO

ayirr7:
The `received` field is (most likely) always present in transaction payloads, but we currently don't do anything with it in the relevant MessageProcessor. This PR allows us to grab the timestamp at which Relay received the transaction and pass it into the SLO calculation, to help us get the ingest transactions SLO. 

I think, if we are confident that we are always containing the `received` field, then we should change the Kafka schema to make `received` a required field. Need to verify this is the case though.

### DIFF
--- a/snuba/datasets/processors/transactions_processor.py
+++ b/snuba/datasets/processors/transactions_processor.py
@@ -476,6 +476,7 @@ class TransactionsMessageProcessor(DatasetMessageProcessor):
 
         try:
             raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+            assert raw_received is not None
             received = datetime.utcfromtimestamp(raw_received)
             return InsertBatch([processed], received)
         except Exception:
--- a/snuba/datasets/processors/transactions_processor.py
+++ b/snuba/datasets/processors/transactions_processor.py
@@ -24,6 +24,7 @@ from snuba.processor import (
     InsertBatch,
     ProcessedMessage,
     _as_dict_safe,
+    _collapse_uint32,
     _ensure_valid_date,
     _ensure_valid_ip,
     _unicodify,
@@ -473,4 +474,11 @@ class TransactionsMessageProcessor(DatasetMessageProcessor):
         # the following operation modifies the event_dict and is therefore *not* order-independent
         self._process_contexts_and_user(processed, event_dict)
 
-        return InsertBatch([processed], event_dict["data"]["received"])
+        raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+        received = (
+            datetime.utcfromtimestamp(raw_received)
+            if raw_received is not None
+            else None
+        )
+
+        return InsertBatch([processed], received)
--- a/snuba/datasets/processors/transactions_processor.py
+++ b/snuba/datasets/processors/transactions_processor.py
@@ -474,11 +474,11 @@ class TransactionsMessageProcessor(DatasetMessageProcessor):
         # the following operation modifies the event_dict and is therefore *not* order-independent
         self._process_contexts_and_user(processed, event_dict)
 
-        raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
-        received = (
-            datetime.utcfromtimestamp(raw_received)
-            if raw_received is not None
-            else None
-        )
+        if event_dict["data"]["received"] is not None:
+            raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+            assert raw_received is not None
+            received = datetime.utcfromtimestamp(raw_received)
+        else:
+            received = None
 
         return InsertBatch([processed], received)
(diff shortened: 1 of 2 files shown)


lynnagara:
I think you can go ahead and change `received` in the transactions schema to required. It should fail Sentry / Snuba CI if there are tests where the payload is missing it. In prod, it should be there since Relay is processing events/transactions the same - but even if it's not, it's just a warning.

ayirr7:
> It should fail Sentry / Snuba CI if there are tests where the payload is missing it

There was a test where the sample payload didn't include the `received` field, and I had to add it for the test to pass. @lynnagara

snuba/datasets/processors/transactions_processor.py
```diff
@@ -472,4 +474,11 @@ def process_message(
         # the following operation modifies the event_dict and is therefore *not* order-independent
         self._process_contexts_and_user(processed, event_dict)
 
-        return InsertBatch([processed], None)
+        if event_dict["data"]["received"] is not None:
```
lynnagara, line 477: In the schema it's defined as a number, and not nullable, so why do we need this check here?

ayirr7: I was thinking of merging this PR first (before we make the change to the schema to make `received` a required field), and then making the schema change after?

hubertsentry: from schema point of view, if it's required and non-nullable, you probably don't need to check it. @lynnagara do you know what will happen if a payload is not schema compatible, will we drop the payload? Do you know what are the processes to ensure the payloads matched the schema?

lynnagara: right now, you'll just get a sentry warning if the schema doesn't match. i'd merge the schema change first to validate that if you're unsure.

hubertsentry: sounds good. Once the payload is always schema compatible, then we probably don't need this extra check.

snuba/datasets/processors/transactions_processor.py
```diff
@@ -472,4 +474,10 @@ def process_message(
         # the following operation modifies the event_dict and is therefore *not* order-independent
         self._process_contexts_and_user(processed, event_dict)
 
-        return InsertBatch([processed], None)
+        try:
+            raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+            assert raw_received is not None
+            received = datetime.utcfromtimestamp(raw_received)
+            return InsertBatch([processed], received)
+        except Exception:
```
hubertsentry, line 482: I would do

```
except KeyError:
```

Also, maybe looking into https://docs.python.org/3/tutorial/errors.html#exception-chaining

snuba/datasets/processors/transactions_processor.py
```diff
@@ -472,4 +474,10 @@ def process_message(
...
         self._process_contexts_and_user(processed, event_dict)
 
-        return InsertBatch([processed], None)
+        try:
+            raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+            assert raw_received is not None
+            received = datetime.utcfromtimestamp(raw_received)
+            return InsertBatch([processed], received)
+        except KeyError:
+            raise KeyError("Missing received timestamp field in transaction")
```
lynnagara, line 483: Wasn't the point of the try/except to fall back to the old behavior in case we hit the KeyError?

ayirr7: Just to be clear: today, what happens if there is a payload missing the `received` field is that the `MessageProcessor` will just drop it - is this correct?

So I guess we can fall back to the old behavior, but we also want to see the error show up in Sentry so we can know if there were any payloads missing the field. What's the right way of getting that error to show up?

ayirr7: Alright I can do that then, thanks

snuba/datasets/processors/transactions_processor.py
```diff
@@ -472,4 +474,11 @@ def process_message(
...
         self._process_contexts_and_user(processed, event_dict)
 
-        return InsertBatch([processed], None)
+        try:
+            raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+            assert raw_received is not None
+            received = datetime.utcfromtimestamp(raw_received)
+            return InsertBatch([processed], received)
+        except KeyError as err:
+            logger.exception("Missing required field received in payload", err)
```
lynnagara, line 483:
```suggestion
            logger.exception(e)
```

snuba/datasets/processors/transactions_processor.py
```diff
@@ -472,4 +474,11 @@ def process_message(
         # the following operation modifies the event_dict and is therefore *not* order-independent
         self._process_contexts_and_user(processed, event_dict)
 
-        return InsertBatch([processed], None)
+        try:
+            raw_received = _collapse_uint32(int(event_dict["data"]["received"]))
+            assert raw_received is not None
```
lynnagara, line 479: If received is not there, won't you get an AssertionError that isn't being caught?

ayirr7: Right, but wouldn't the first line (which throws KeyError) be thrown first?

ayirr7: But I guess the AssertionError could be even easier, since it'd be guaranteed that it's being thrown at that line

PR #4458, opened 2023-07-03
perm(cardinality-analyzer): Add ingest team

jjbayer:
Give the entire ingest team permission to use the cardinality analyzer.

### DIFF
--- a/snuba/admin/iam_policy/iam_policy.json
+++ b/snuba/admin/iam_policy/iam_policy.json
@@ -33,7 +33,7 @@
     {
       "members": [
           "group:team-starfish@sentry.io",
-          "user:iker.barriocanal@sentry.io"
+          "group:team-ingestion-pipeline@sentry.io"
       ],
       "role": "roles/CardinalityAnalyzer"
     }


snuba/admin/iam_policy/iam_policy.json
```diff
@@ -33,7 +33,7 @@
     {
       "members": [
           "group:team-starfish@sentry.io",
-          "user:iker.barriocanal@sentry.io"
+          "group:team-ingestion-pipeline@sentry.io"
```
jjbayer, line 36: Not sure if this will work. Where does this team have to be declared? It's listed as a subteam [here](https://github.com/getsentry/security-as-code/blob/cc7c5bce266983a3968f4f9ceccde877ede814d6/rbac/env/prod/team/engineering/infra.tf#L13) but I'm not sure whether that's connected.

iker-barriocanal (approved):

snuba/admin/iam_policy/iam_policy.json
evanh: It's coming from Google Groups, so if it's a valid google group then it should work.

evanh (approved):

PR #4464, opened 2023-07-05
experiment(capman,starfish): add ability to turn off concurrent throttling

volokluev:
The concurrent throttler is possibly cramping our performance too much. It's also a hidden part of our allocation policy which needs to become explicit. 

This PR allows turning off that functionality so we can run an experiment to tell us how much it actually impacts things

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -315,7 +315,7 @@ def execute_query_with_rate_limits(
         # depending on the `stats` dict to be populated ahead of time
         # is not great style, but it is done in _format_storage_query_and_run.
         # This should be removed by 07-15-2023. Either the concurrent throttling becomes
-        # another allocation policy
+        # another allocation policy or we remove this mechanism entirely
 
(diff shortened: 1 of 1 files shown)


snuba/web/db_query.py
```diff
@@ -307,9 +307,23 @@ def execute_query_with_rate_limits(
...
-        )
+        # -----------------------------------------------------------------
+        # HACK (Volo): This is a hack experiment to see if we can
+        # stop doing  concurrent throttling in production on a specific table
+        # and still survive.
+
+        # depending on the `stats` dict to be populated ahead of time
+        # is not great style, but it is done in _format_storage_query_and_run.
+        # This should be removed by 07-15-2023. Either the concurrent throttling becomes
+        # another allocation policy
```
ayirr7, line 318: Just a question about the comment here, is there supposed to be more in this last part of the description?

volokluev: woops, updated. Good catch!

ayirr7 (approved):

PR #4466, opened 2023-07-06
fix(metrics): Fix granularity processor to allow multiple manual conditions

enochtangg:
### Overview
The following PR is responsible for fixing `GranularityProcessor` and `MappedGranularityProcessor` to handle multiple granularities that is passed as a condition in the `WHERE` clause. Currently, the processor only looks at the `GRANULARITY` clause to determine whether a granularity has been specified. As a result, the processor 

This change adds the following functionalities:
1. Fixes bug where processor incorrectly adds another granularity condition into the WHERE clause even when one already exists in GRANULARITY clause. 
2. Throws an error when both granularities are specified in both the WHERE clause and GRANULARITY clause
3. Allow for multiple granularities to be specified in WHERE conditions.  This was technically broken before since mappings were only applied to GRANULARITY clause. 

### Testing
Unit tests were added to check the functionality of the two processors when different queries are passed in.

### DIFF
--- a/snuba/query/processors/logical/granularity_processor.py
+++ b/snuba/query/processors/logical/granularity_processor.py
@@ -1,3 +1,4 @@
+from abc import abstractmethod
 from typing import Callable, Mapping, NamedTuple, Optional, Tuple
 
 from snuba.query.conditions import (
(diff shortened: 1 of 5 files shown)


snuba/query/processors/logical/granularity_processor.py
```diff
@@ -12,36 +31,155 @@
...
+            )
+            rhs_granularity = self.find_granularity_in_expression(
+                match.expression("rhs")
+            )
+            if lhs_granularity is None:
+                return rhs_granularity
+            elif rhs_granularity is None:
+                return lhs_granularity
+            else:
+                raise InvalidGranularityException(
```
evanh, line 83: I don't think this is quite what the product team is looking for. Part of what they want to do is specify multiple granularities in a query, so they can do a single query that fetches a different granularity for different time periods. So something like

```
MATCH (...)
SELECT ...
WHERE (
    (granularity = 1 AND time > now() - INTERVAL 1 DAY)*
    OR (granularity = 2 AND time < now() - INTERVAL 1 DAY)
)
AND ...
```

In this example they can have the hourly granularity for the most recent 24 hours and the daily granularity for everything else. Note that they would still need top level time conditions that cover the entire range (`time >= ... AND time < ...`).

*Note that we don't actually support the INTERVAL syntax in SnQL, I was just using it to make the example easier to read.

enochtangg: This PR isn't meant to support multiple granularities in a query. The objective is to first allow granularities to work inside the WHERE clause since this is currently broken. For the use-case you mentioned (supporting multiple granularities in a single query for different time periods), I think would build off of this fix and maybe should be described in a new ticket.

enochtangg: Update the PR and its description to include this change. The use-case you describe is captured in this test https://github.com/getsentry/snuba/pull/4466/files#diff-0111cefeaef97e9d9b44d01013f69d4b529e1b1cbe1d9136e32268683f7e6311R449

evanh (approved):
Loving all the new tests.

PR #4467, opened 2023-07-06
feat(generic-metrics): Add aggregation options for generic metrics raw tables

john-z-yang:
| [#1](https://github.com/getsentry/snuba/pull/4467) | ➤ Add aggregation options for generic metrics raw tables | Ready for review          | master                                               |
| [#2](https://github.com/getsentry/snuba/pull/4477) | Add new MatViews for generic metrics | Ready for review | [#1](https://github.com/getsentry/snuba/pull/4467) |
| [#3](https://github.com/getsentry/snuba/pull/4478) | Bump `materialization_version` for generic metrics | Ready for review | [#2](https://github.com/getsentry/snuba/pull/4477) |

### Add new options in raw tables of distribution, sets, and counters

- `decasecond_retention_days`: the number of days 10 seconds granularity aggregation will be stored for
- `min_retention_days`: the number of days 60 seconds granularity aggregation will be stored for
- `hr_retention_days`: the number of days 3600 seconds granularity aggregation will be stored for
- `enable_histogram`: Whether to enable histogram or not (distributions only)

### DIFF
--- a/snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
+++ b/snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
@@ -46,7 +46,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 "day_retention_days",
                 UInt(8, MigrationModifiers(default=str("retention_days"))),
             ),
-            "min_retention_days",
+            "hr_retention_days",
         ),
     ]
 
--- a/snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
+++ b/snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
@@ -1,0 +1,82 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    storage_set_key = StorageSetKey.GENERIC_METRICS_DISTRIBUTIONS
+
+    local_table_name = "generic_metric_distributions_raw_local"
+    dist_table_name = "generic_metric_distributions_raw_dist"
+
+    columns = [
+        (
+            Column("enable_histogram", UInt(8, MigrationModifiers(default=str("0")))),
+            "granularities",
+        ),
+        (
+            Column(
+                "decasecond_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "enable_histogram",
+        ),
+        (
+            Column(
+                "min_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "decasecond_retention_days",
+        ),
+        (
+            Column(
+                "hr_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
+        ),
+        (
+            Column(
+                "day_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
+        ),
+    ]
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=self.storage_set_key,
+                table_name=table_name,
+                column=column,
+                after=after,
+                target=target,
+            )
+            for column, after in self.columns
+            for table_name, target in [
+                (self.dist_table_name, OperationTarget.DISTRIBUTED),
+                (self.local_table_name, OperationTarget.LOCAL),
+            ]
+        ]
+
+    def backwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.DropColumn(
+                storage_set=self.storage_set_key,
+                table_name=table_name,
+                column_name=column.name,
+                target=target,
+            )
+            for column, _ in self.columns
+            for table_name, target in [
+                (self.dist_table_name, OperationTarget.DISTRIBUTED),
+                (self.local_table_name, OperationTarget.LOCAL),
+            ]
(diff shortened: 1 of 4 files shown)


snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
```diff
@@ -0,0 +1,82 @@
...
+    blocking = False
+
+    storage_set_key = StorageSetKey.GENERIC_METRICS_DISTRIBUTIONS
+
+    local_table_name = "generic_metric_distributions_raw_local"
+    dist_table_name = "generic_metric_distributions_raw_dist"
+
+    columns = [
+        (
+            Column("enable_histogram", UInt(8, MigrationModifiers(default=str("0")))),
```
nikhars, line 20: You should probably make MigrationModifiers default value accept `ints` in a separate PR first. I don't see a reason why it has to be restricted to strings only. Then once that PR is merged, you can use the int value itself over here. It seems a bit odd that a UInt8 column is being populated from a string.

snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
```diff
@@ -0,0 +1,82 @@
...
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
+        ),
+        (
+            Column(
+                "day_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
```
nikhars, line 49: This should probably be `hr_retention_days`. `min_retention_days` is already used above.

snuba/snuba_migrations/generic_metrics/0015_sets_add_options.py
```diff
@@ -0,0 +1,78 @@
...
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
+        ),
+        (
+            Column(
+                "day_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
```
nikhars, line 45: `hr_retention_days` here as well

snuba/snuba_migrations/generic_metrics/0016_counters_add_options.py
```diff
@@ -0,0 +1,78 @@
...
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
+        ),
+        (
+            Column(
+                "day_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "min_retention_days",
```
nikhars, line 45: Same comment as above

snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
john-z-yang: I was really confused about the `str` type hints there as well. I took a look at the underlying code for `MigrationModifiers`, looks like it's just generating SQL strings [under the hood](https://github.com/getsentry/snuba/blob/b0f7e434459fb5cad5a1b6a8ec9e460f9e8c1cd4/snuba/migrations/columns.py#L72) so the str here is technically an SQL expression
```python
def for_schema(self, content: str) -> str:
        return "{} DEFAULT {}".format(content, self.default)
```
I could make another pr that let default be some other type, but I think that type will need to support construction of arbitrary sql queries instead of `ints` and it needs to be turned into a string anyway.

snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
john-z-yang: Fixed, good catch

nikhars (approved):

snuba/snuba_migrations/generic_metrics/0014_distribution_add_options.py
```diff
@@ -0,0 +1,82 @@
...
+        ),
+        (
+            Column(
+                "day_retention_days",
+                UInt(8, MigrationModifiers(default=str("retention_days"))),
+            ),
+            "hr_retention_days",
+        ),
```
fpacifici, line 50: If we use adedicated column per retention, we would have to run through this process (that includes the change to the matview) if we wanted to introduce a new granularity. Did you consider having an array of nested structures where you could store pairs of granularity and retention? That would not require structural changes to the table to introduce a new granularity.

fpacifici: Also how are we going to disable a granularity entirely? do we set the column to 0? 
It is important we do not rely on a retention = 0 to disable a granularity entirely. even a 0 retention would mean data gets materialized and then dropped.

john-z-yang: I did consider using an array of nested structures, and did not go forward with it because
- From discussions with Nikhar and Oliver it looks like we are unlikely to introduce new granularity, and if so, not very frequently
- With the nested structures, we still need to run migration on the MatViews since we need to update the integer to seconds mapping

We will not be relying on setting retention to zero. The messages sent from Sentry will have some enum that represents the metric aggregation type. The snuba consumers will read that enum and only insert granularities that should be enabled into the granularities columns.

fpacifici: > With the nested structures, we still need to run migration on the MatViews since we need to update the integer to seconds mapping

I am not sure this is accurate:
I am talking about this schema 
```
retentions: Nested(
     granularity: UInt,
     retention: Int
)
```

Each row of that array defines a rule you do not need to change the table to add new rules or remove some

john-z-yang: I see, but wouldn't the MatView still not understand what granularity maps to in seconds?
I.e.
```sql
toDateTime(multiIf(granularity=0,10,
                                 granularity=1,60,
                                 granularity=2,3600,
                                 granularity=3,86400,-1)
...
as timestamp,
```
Don't we need to run another migration to the MatView to add another condition for `multiIf` even if we use the nested structure?

PR #4469, opened 2023-07-06
feat(replays): Add migration for click_is_dead and click_is_rage columns

cmanallen:


snuba/snuba_migrations/replays/0011_add_is_dead_rage.py
```diff
@@ -0,0 +1,57 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return list(forward_columns_iter())
+
+    def backwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return list(backward_columns_iter())
+
+
+def forward_columns_iter() -> Iterator[operations.SqlOperation]:
```
evanh, line 24: Why is this a separate function that `yield`s ? Why not a simple list?

cmanallen: @evanh Its just the pattern I've established previously and am carrying over.  I'm aware that I can append to and return a list but yielding works just as well for me.  The two are functionally equivalent but yield requires less indentation and (for me) requires less thought when reading and writing migration operations.

cmanallen:
@evanh Will these new columns default to 0?

snuba/snuba_migrations/replays/0011_add_is_dead_rage.py
evanh: Right, this is a stylistic nit: this is needlessly complicated. Simple is better than complex and flat is better than nested.

evanh (approved):
Just double checked in the docs to be sure: yes it will default the empty values to 0.

> If the data is missing for a column when reading from the table, it is filled in with default values (by performing the default expression if there is one, or using zeros or empty strings).

PR #4470, opened 2023-07-06
feat(replays): Add click_is_dead and click_is_rage columns

cmanallen:


evanh (commented):
Are these fields already being sent to the consumer? If there are any messages that don't have these values, your consumer will crashloop. Also, is there a kafka schema for the replays processor? Have these been added there?

cmanallen:
@evanh I'll merge the consumer changes first.  Also, I'll update the kafka schema now.

cmanallen:
Ah, because of the way we encode the payload value there's no need (read: ability) to update the schema.

evanh (approved):
Don't merge this until the migration is merged and confirmed to have run.

PR #4472, opened 2023-07-06
feat(snubsplain): Add basic frontend for snubsplain

evanh:
Add a basic frontend for the explain data that is being sent by the endpoint.
This adds a new tool called Snubsplain to the admin tool, which when provided a
SnQL query will show the list of query processors that ran.

Frontend Changes:
- Create a new Snuba Explain tool that uses the meta data returned from calling
  the debugSnQLQuery endpoint
- Build a basic framework for rendering different steps of the query pipeline
- Show the names of the query processors that ran and in what order
- Show the final Clickhouse SQL that ran

Backend Changes:
- Capture the storage query processor names
- Set up some standard dataclasses for capturing explain meta data

snubsplain-v1

### DIFF
--- a/snuba/state/explain_meta.py
+++ b/snuba/state/explain_meta.py
@@ -1,14 +1,17 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Any, cast
+from typing import Any, Literal, cast
 
 from flask import g
+
+ExplainType = Literal["query_transform"]
 
 
 @dataclass
 class ExplainStep:
-    category: str  # The type of step e.g. "processor"
+    category: str  # The class of step e.g. "processor"
+    type: ExplainType  # A value that tells the frontend what data the step has
     name: str  # The specific name for the step e.g. "TimeSeriesProcessor"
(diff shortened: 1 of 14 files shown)


snuba/state/explain_meta.py
```diff
@@ -27,7 +31,7 @@ def add_step(category: str, name: str, data: dict[str, Any] | None = None) -> No
     try:
         if data is None:
             data = {}
-        step = ExplainStep(category, name, data)
+        step = ExplainStep(category, "query_transform", name, data)
```
enochtangg, line 34: What is this type used for if it's always defaulted to `query_transform`? In the future, what other types would you imagine?

evanh: Good question! I have a couple other types I thought of:
- validation steps, that record information about why a query isn't valid
- splitter steps, that show how the query was split (if at all)
- caching step, just shows whether the query was cached, waited etc.
- allocation step, would this query have been throttled? Why?
- rate limit steps, would this query have been rate limited? Why?

enochtangg (approved):

PR #4474, opened 2023-07-06
feat(capman): add project_id as implicit tenant

volokluev:
In order to port the project based rate limiters to the capman framework, we have to have project ids as tenants, we currently do not. Add logic in the `build_request` function to add `project_id` to `AttributionInfo.tenant_ids`

This logic makes an explicit decision that when there are multiple projects being searched in one request, there is no project tenant, instead the organization_id becomes the tenant.

### DIFF
--- a/snuba/request/validation.py
+++ b/snuba/request/validation.py
@@ -98,8 +98,10 @@ def build_request(
             )
 
             project_ids = get_object_ids_in_query_ast(query, "project_id")
+            query_project_id = None
             if project_ids is not None and len(project_ids) == 1:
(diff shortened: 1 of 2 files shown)


evanh:
> we have to have project ids as tenants

> when there are multiple projects being searched in one request, there is no project tenant

These two statements seem hard to reconcile. Can you elaborate on how rate limiting will work with multiple project queries?

volokluev:
> we will use the organization id in the tenant ids instead

evanh (approved):

PR #4475, opened 2023-07-07
Fix MultistorageCollector as #4297

qbx2:
I was digging the off-by-one error in snuba commit log issue. I prepared two-line fix for 23.3.1 but soon I found the one-line fix (#4297) was merged in last month. I think the rest should be fixed too.

### Legal Boilerplate


lynnagara (approved):
Thanks @qbx2!

PR #4479, opened 2023-07-07
ref(snubsplain): Show what changed in the query when a processor runs

evanh:
Track the changes in the query (using the AST formatter) to show what changes a processor made to the query. Display that in the code blocks.

Other changes:
- Change the query display component to not show the predefined queries option if none are provided
- Format the final SQL using Mantine

snubsplain-query-diff

### DIFF
--- a/snuba/datasets/plans/storage_plan_builder.py
+++ b/snuba/datasets/plans/storage_plan_builder.py
@@ -39,6 +39,7 @@ from snuba.query.processors.physical.mandatory_condition_applier import (
     MandatoryConditionApplier,
 )
 from snuba.query.query_settings import QuerySettings
+from snuba.state import explain_meta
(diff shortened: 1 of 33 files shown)


snuba/datasets/plans/storage_plan_builder.py
```diff
@@ -75,10 +75,13 @@ def process_and_run_query(
                     description=type(processor).__name__, op="processor"
                 ):
                     if query_settings.get_dry_run():
-                        explain_meta.add_step(
-                            "storage_processor", type(processor).__name__
-                        )
-                    processor.process_query(query, query_settings)
+                        with explain_meta.with_query_differ(
```
volokluev, line 78: does the differ run whether or not we are using snubsplain?

volokluev (approved):

PR #4485, opened 2023-07-10
feat: Validate schemas for all topics in the test api

lynnagara:


snuba/web/views.py
```diff
@@ -619,9 +619,7 @@ def eventstream(*, entity: Entity) -> RespTuple:
 
         storage = entity.get_writable_storage()
         assert storage is not None
-        storage_name = storage.get_storage_key().value
-        # TODO: A few storages are currently excluded from schema validation
-        should_validate_schema = storage_name not in ["search_issues", "replays"]
+        should_validate_schema = True
```
evanh, line 622: Is this flag still necessary?

lynnagara: It could be inlined, but i figured keeping the variable is more readable

untitaker: for the purposes of a smaller diff? 🤔

untitaker (approved):

PR #4488, opened 2023-07-10
fix(rust-consumer): Fix batching

lynnagara:
The reduce step should accumulate rows inside the BytesInsertBatch not produce a batch of batches.


lynnagara:
Done in https://github.com/getsentry/snuba/pull/4490

PR #4490, opened 2023-07-11
feat(rust-consumer): Write to ClickHouse

lynnagara:
feat(rust-consumer): Add basic clickhouse strategy

Same idea as https://github.com/getsentry/snuba/pull/4437 with a few differences:
- Does not switch the arroyo interface to async
- Does not write to ClickHouse by default (only if --no-skip-write is passed)
- Batches inserted rows into fewer writes with multiple rows


rust_snuba/src/strategies/clickhouse.rs
```diff
@@ -0,0 +1,68 @@
...
+        let mut decoded_rows = vec![];
+        for row in message.payload().rows {
+            let decoded_row = String::from_utf8_lossy(&row);
+            decoded_rows.push(decoded_row.to_string());
+        }
+
+        if !self.skip_write {
+            let data = decoded_rows.join("\n").to_string();
+            // TODO: Handle error
+            let _result = self.runtime.block_on(self.clickhouse_client.send(data)).unwrap();
```
dbanda, line 49: If we do it this way, aren't we losing out on the benefits of concurrency/asynchrony? The thread will block on sending to Clickhouse. Is this intentional?

untitaker: yes. I would use `self.runtime.spawn`, which returns a `Handle` you can call `is_finished` on in `Strategy::poll`. if you want to be even fancier you can use a oneshot channel to block with timeout

dbanda (approved):
This scheme will have only one outstanding write request. Shouldn't we have handles be stored in some kind of list if we want to have multiple writes? Seems like we are giving up a lot of performance and not taking advantage of the runtime.

Approving to unblock since I think the experiment with one outstanding write might still be worth doing.

lynnagara:
> This scheme will have only one outstanding write request. Shouldn't we have handles be stored in some kind of list if we want to have multiple writes? Seems like we are giving up a lot of performance and not taking advantage of the runtime.
> 
> Approving to unblock since I think the experiment with one outstanding write might still be worth doing.

Yeah. I did it this way for compatibility with the Python consumer which also does not support multiple writes at once and strictly enforces ordering of inserts. https://github.com/getsentry/snuba/blob/master/snuba/consumers/consumer.py#L828

Though, I think you could be right right and we should reconsider this in both Python and Rust.

PR #4495, opened 2023-07-11
Update jsonnet lib and add rollback pipeline

mattgauntseo-sentry:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ lint-rust:
 gocd:
 	rm -rf ./gocd/generated-pipelines
 	mkdir -p ./gocd/generated-pipelines
-	cd ./gocd/templates && jb install
+	cd ./gocd/templates && jb install && jb update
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -88,7 +88,7 @@ lint-rust:
 gocd:
 	rm -rf ./gocd/generated-pipelines
 	mkdir -p ./gocd/generated-pipelines
-	cd ./gocd/templates && jb install
+	cd ./gocd/templates && jb install && jb update
```
fpacifici, line 91: Why do you need to update right after `jb install` ?

mattgauntseo-sentry: If someone deletes the `jsonnetfile.lock` then `jb install` is needed.

If `jsonnetfile.lock` exists but its out of date with `jsonnetfile.json`, i.e. a version change like this PR, then `jb update` is needed to update the lock file and download the new version.

It would be fine 99% of the time to remove the `jb install`, but there is no harm in having both in the name of making things easy for folks.

PR #4496, opened 2023-07-11
migration(group_attributes): create new GroupAttributes table that mirrors a subset of columns from sentry

barkbarkimashark:
This new dataset replicates a subset of crucial values from the Groups, GroupAssignee and GroupOwners postgres tables. 

The plan is to stream the latest 'snapshot' values described in the schema every time any of the postgres column values are updated/created/deleted. 

The latest snapshot will always replace the oldest snapshot and eventually get replaced by clickhouse. 

Note: for the time being, this table should only be available in SaaS and in 'prototype' mode. We want to verify joins between the events/search_issues and this table works as expected and that most straight-forward way was to build the dataset, insert some data, and execute some parallel issue search queries to verify.

### DIFF
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -158,7 +158,7 @@ _REGISTERED_MIGRATION_GROUPS: Dict[MigrationGroup, _MigrationGroup] = {
     ),
     MigrationGroup.GROUP_ATTRIBUTES: _MigrationGroup(
         loader=GroupAttributesLoader(),
-        storage_sets_keys={StorageSetKey.Group_ATTRIBUTES},
+        storage_sets_keys={StorageSetKey.GROUP_ATTRIBUTES},
         readiness_state=ReadinessState.PARTIAL,
     ),
 }
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -8,6 +8,7 @@ from snuba.migrations.group_loader import (
     EventsLoader,
     FunctionsLoader,
     GenericMetricsLoader,
+    GroupAttributesLoader,
     GroupLoader,
     MetricsLoader,
     OutcomesLoader,
@@ -153,6 +154,11 @@ _REGISTERED_MIGRATION_GROUPS: Dict[MigrationGroup, _MigrationGroup] = {
     MigrationGroup.SPANS: _MigrationGroup(
         loader=SpansLoader(),
         storage_sets_keys={StorageSetKey.SPANS},
+        readiness_state=ReadinessState.PARTIAL,
+    ),
+    MigrationGroup.GROUP_ATTRIBUTES: _MigrationGroup(
+        loader=GroupAttributesLoader(),
+        storage_sets_keys={StorageSetKey.Group_ATTRIBUTES},
         readiness_state=ReadinessState.PARTIAL,
     ),
 }
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -39,6 +39,7 @@ class MigrationGroup(Enum):
     TEST_MIGRATION = "test_migration"
     SEARCH_ISSUES = "search_issues"
     SPANS = "spans"
+    GROUP_ATTRIBUTES = "group_attributes"
 
 
 # Migration groups are mandatory by default. Specific groups can
@@ -54,6 +55,7 @@ OPTIONAL_GROUPS = {
     MigrationGroup.TEST_MIGRATION,
     MigrationGroup.SEARCH_ISSUES,
     MigrationGroup.SPANS,
+    MigrationGroup.GROUP_ATTRIBUTES,
 }
 
 
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -159,7 +159,7 @@ _REGISTERED_MIGRATION_GROUPS: Dict[MigrationGroup, _MigrationGroup] = {
     MigrationGroup.GROUP_ATTRIBUTES: _MigrationGroup(
         loader=GroupAttributesLoader(),
         storage_sets_keys={StorageSetKey.GROUP_ATTRIBUTES},
-        readiness_state=ReadinessState.PARTIAL,
+        readiness_state=ReadinessState.LIMITED,
     ),
 }
 
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -274,7 +274,7 @@ COLUMN_SPLIT_MAX_RESULTS = 5000
 
 # The migration groups that can be skipped are listed in OPTIONAL_GROUPS.
 # Migrations for skipped groups will not be run.
-SKIPPED_MIGRATION_GROUPS: Set[str] = {"group_attributes"}
+SKIPPED_MIGRATION_GROUPS: Set[str] = set()
 
 # Dataset readiness states supported in this environment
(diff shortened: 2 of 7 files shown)


snuba/snuba_migrations/group_attributes/0001_group_attributes.py
```diff
@@ -0,0 +1,71 @@
...
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_local",
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, group_id",
+                    settings={"index_granularity": "8192"},
+                    storage_set=StorageSetKey.GROUP_ATTRIBUTES,
```
barkbarkimashark: I omitted certain common table engine parameters like `version_column`, `partition_by`, `sample_by`, and `ttl`.

Since this data is a mirror of the postgres Groups tables, I think we'll need to retain the data indefinitely.

lynnagara: Don't we need the ability to delete the group and null out the data though, for which a version column may be helpful? If a column is deleted from postgres, we should remove it here too?

barkbarkimashark: Oh, right. Forgot about when we actually delete the Group row. In the case of `GroupAssignee` and `GroupOwner` rows getting deleted, I was planning to simply send a new snapshot with the appropriate column values nulled out. But that doesn't really work when the Group is totally deleted. Will include a deleted column to account for this.

On a somewhat related note, does clickhouse store timestamp metadata for every row insert? I'm wondering how the replacing merge tree knows which duplicate rows are 'latest' without us having to specify a timestamp to impose ordering for de-duping:

```
CREATE TABLE test_table(`key` Int64, `time` DateTime) ENGINE = ReplacingMergeTree() ORDER BY key;
INSERT INTO test_table Values (1, now());
INSERT INTO test_table Values (1, now());
INSERT INTO test_table Values (1, now());
INSERT INTO test_table Values (1, yesterday());

----

select * from test_table

┌─key─┬────────────────time─┐
│   1 │ 2023-07-10 00:00:00 │
└─────┴─────────────────────┘
┌─key─┬────────────────time─┐
│   1 │ 2023-07-11 23:27:48 │
└─────┴─────────────────────┘
┌─key─┬────────────────time─┐
│   1 │ 2023-07-11 23:27:47 │
└─────┴─────────────────────┘
┌─key─┬────────────────time─┐
│   1 │ 2023-07-11 23:27:45 │
└─────┴─────────────────────┘

----

select * from test_table final

┌─key─┬────────────────time─┐
│   1 │ 2023-07-10 00:00:00 │
└─────┴─────────────────────┘
```

notice the first three inserts results in dupe rows the increasing `time` value, and the final insert is out of order, but when we apply `FINAL` clickhouse knows that the row with the `time=yesterday()` is the 'latest'.

lynnagara: Since you didn't specify a version column, ClickHouse defaults to keeping the last inserted row. See https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree#ver

lynnagara:
There is a mechanism that controls which environments migrations get rolled out into. I'm guessing we need this in SaaS first (after new cluster is available) but not anywhere else for the test. I don't recall how to set the states though - I think @enochtangg can advise you on it.

lynnagara:
Since we are going to put this on a 22.8 cluster, we need to also run all Snuba tests against 22.8 before deploying it there (and increase the max supported version).

barkbarkimashark:
> There is a mechanism that controls which environments migrations get rolled out into. I'm guessing we need this in SaaS first (after new cluster is available) but not anywhere else for the test. I don't recall how to set the states though - I think @enochtangg can advise you on it.

Right, I believe we'll need to make sure this table is joinable against the events and search_issues dataset as well. 

I glanced over this bit of code https://github.com/getsentry/snuba/blob/206c138d3eb71c6f84e5f49f32393939edc61387/snuba/clusters/storage_sets.py#L82-L89
but it's not completely clear to me what changes we'd have to make respect to the various environments.

evanh:
> what changes we'd have to make respect to the various environments

There will be a couple changes needed in the configuration to allow this, but I think this table needs to be located on the same cluster as the errors table. I don't believe our join system allows for joins across clusters.

I would maybe use the same storage set key as errors just to make that dependency clear.

evanh:
> There is a mechanism that controls which environments migrations get rolled out into. I'm guessing we need this in SaaS first (after new cluster is available) but not anywhere else for the test. I don't recall how to set the states though - I think @enochtangg can advise you on it.

The migration group needs to be added to `_REGISTERED_MIGRATION_GROUPS` with `ReadinessState.PARTIAL`. See the link below.
https://github.com/getsentry/snuba/pull/4496/files#diff-a105aae1421f2479451fdcac276a25d3715cce152b52ecfbbe50f4b47896f500R74

lynnagara:
> I would maybe use the same storage set key as errors just to make that dependency clear.

It should stay a different one as we need to map this to different cluster

lynnagara:
please see the ci failure

snuba/settings/__init__.py
```diff
@@ -273,7 +274,7 @@ class RedisClusters(TypedDict):
 
 # The migration groups that can be skipped are listed in OPTIONAL_GROUPS.
 # Migrations for skipped groups will not be run.
-SKIPPED_MIGRATION_GROUPS: Set[str] = set()
+SKIPPED_MIGRATION_GROUPS: Set[str] = {"group_attributes"}
```
enochtangg, line 277: Adding the migration to this set will not be necessary. The readiness state will handle whether or not the migration is executed in specific environments.

snuba/migrations/groups.py
```diff
@@ -153,6 +156,11 @@ def __init__(
         storage_sets_keys={StorageSetKey.SPANS},
         readiness_state=ReadinessState.PARTIAL,
     ),
+    MigrationGroup.GROUP_ATTRIBUTES: _MigrationGroup(
+        loader=GroupAttributesLoader(),
+        storage_sets_keys={StorageSetKey.GROUP_ATTRIBUTES},
```
lynnagara, line 161: I believe this new storage set needs to be added in this PR so we can use it

barkbarkimashark: Specifically, do you mean adding it here: https://github.com/getsentry/snuba/pull/4496/commits/c3ea983704e35f503fed48a91b09fbdfd8569642

lynnagara: yup

snuba/snuba_migrations/group_attributes/0001_group_attributes.py
```diff
@@ -0,0 +1,74 @@
...
+                ),
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_dist",
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name="group_attributes_dist",
```
evanh, line 56:
```suggestion
                    local_table_name="group_attributes_local",
```

lynnagara (approved):

PR #4497, opened 2023-07-11
check connections on consumer startup

dbanda:
Sometimes it takes envoy sidecar a while to load. Instead of crashlooping immediately, this lets snuba consumers spin for 60 clickhouse connection attempts at startup to check connectivity before running.

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -16,6 +16,7 @@ from snuba.consumers.consumer_config import resolve_consumer_config
 from snuba.datasets.storages.factory import get_writable_storage_keys
 from snuba.datasets.storages.storage_key import StorageKey
 from snuba.environment import setup_logging, setup_sentry
+from snuba.migrations.connect import check_clickhouse_connections
 from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.streams.metrics_adapter import StreamMetricsAdapter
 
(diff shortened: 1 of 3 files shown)


snuba/cli/consumer.py
```diff
@@ -179,6 +180,10 @@ def consumer(
 
     setup_logging(log_level)
     setup_sentry()
+
+    logger.info("Checking Clickhouse connections")
+    check_clickhouse_connections()
```
lynnagara, line 185: Consumers generally only connect to 1 cluster, mind just checking the one instead of the whole lot?

dbanda: ok. update the PR to allow passing in cluster as a parameter

snuba/settings/__init__.py
```diff
@@ -73,6 +73,7 @@
 
 # Clickhouse Options
 CLICKHOUSE_MAX_POOL_SIZE = 25
+CLICKHOUSE_STARTUP_CHECK_RETRIES = 60
```
lynnagara, line 76: Are you planning to override this value? The default seems ok to me since 1 minute is a long time... and consumers won't make any progress if they can't connect anyway.

I wonder if we actually need to move it here and make settings more complex if there's not a use case to change it.

lynnagara (approved):

snuba/settings/__init__.py
dbanda: prob not. I removed it

PR #4504, opened 2023-07-13
feat(generic-metrics): Add new MatView for generic distribution metrics

john-z-yang:
| PR                                                   | Changes                                                  | Status        | Branch From                                          |
| ---------------------------------------------------- | -------------------------------------------------------- | ------------- | ---------------------------------------------------- |
| [#1](https://github.com/getsentry/snuba/pull/4467) | Add aggregation options for generic metrics raw tables | Merged         | master                                               |
| [#2](https://github.com/getsentry/snuba/pull/4504) | ➤ Add new MatView for generic distribution metrics | Approved | master |
| [#3](https://github.com/getsentry/snuba/pull/4509) | Add facilities to insert aggregation options in processor | Ready for review | [#2](https://github.com/getsentry/snuba/pull/4504) |
| [#4](https://github.com/getsentry/snuba/pull/4505) | Bump `materialization_version` to 2 for generic distribution metrics | Ready for review | [#3](https://github.com/getsentry/snuba/pull/4509) |

### Add new MatView for distribution dataset that uses options added in the distribution raw table

This breaks up [4477](https://github.com/getsentry/snuba/pull/4477), where it adds all 3 Matview migrations in one pr. This PR only adds new Matview for distribution

PR #4505, opened 2023-07-13
feat(generic-metrics): Bump `materialization_version` to 2 for generic distribution metrics

john-z-yang:
| [#1](https://github.com/getsentry/snuba/pull/4467) | Add aggregation options for generic metrics raw tables | Merged         | master                                               |
| [#2](https://github.com/getsentry/snuba/pull/4504) | Add new MatView for generic distribution metrics | Merged | master |
| [#3](https://github.com/getsentry/snuba/pull/4509) | Add facilities to insert aggregation options in processor | Merged | [#2](https://github.com/getsentry/snuba/pull/4504) |
| [#4](https://github.com/getsentry/snuba/pull/4505) | ➤ Bump `materialization_version` to 2 for generic distribution metrics | Approved | [#3](https://github.com/getsentry/snuba/pull/4509) |

### Bump `materialization_version` to 2 for generic distribution metrics

This kicks in the new MatView for generic distribution metrics.

**Note**: We're setting `enable_histogram` to `1` because it is set to `0` as default in clickhouse. There's a single pesky endpoint still using histogram. Once we have the rest of the system built, we will tag that specific metric ID with an aggregation option enum to enable histogram for that specific metric only. However, for now, we will have to make do with enabling histogram for all.

### DIFF
--- a/snuba/datasets/metrics_messages.py
+++ b/snuba/datasets/metrics_messages.py
@@ -83,7 +83,8 @@ def aggregation_options_for_distribution_message(
     _: Mapping[str, Any]
 ) -> Mapping[str, Any]:
     return {
-        "materialization_version": 1,
+        "materialization_version": 2,
+        "enable_histogram": 1,
         "granularities": [
             GRANULARITY_ONE_MINUTE,
             GRANULARITY_ONE_HOUR,


snuba/datasets/metrics_messages.py
```diff
@@ -61,3 +66,38 @@ def value_for_counter_message(message: Mapping[str, Any]) -> Mapping[str, Any]:
...
+
+
+def aggregation_options_for_set_message(_: Mapping[str, Any]) -> Mapping[str, Any]:
+    return {
+        "materialization_version": 1,
+        "granularities": [
+            GRANULARITY_ONE_MINUTE,
+            GRANULARITY_ONE_HOUR,
+            GRANULARITY_ONE_DAY,
+        ],
```
nikhars, line 78: Does this need to be overridable for different metric types?

snuba/datasets/metrics_messages.py
```diff
@@ -61,3 +66,38 @@ def value_for_counter_message(message: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    }
+
+
+def aggregation_options_for_distribution_message(
+    _: Mapping[str, Any]
+) -> Mapping[str, Any]:
+    return {
+        "materialization_version": 2,
```
nikhars, line 86: I would break up the PR's so that the refactoring happens in 1 PR and changing only the materialization version happens in a separate PR

snuba/datasets/metrics_messages.py
john-z-yang: Yes, I was thinking that we can parse the aggregation option enum here, and some enum might specify that we don't need GRANULARITY_ONE_MINUTE, so we can just omit it here

snuba/datasets/metrics_messages.py
john-z-yang: Sounds good

john-z-yang: https://github.com/getsentry/snuba/pull/4509

nikhars (approved):

PR #4506, opened 2023-07-13
feat(db_query): Readthrough Cache entire `db_query()` pipeline

rahul-kumar-saini:
### Overview
- This PR brings the readthrough cache hit check to the beginning of db_query
- Previously, db_query did the following:
  - Check Allocation policy quotas
  - ...
  - Check cache
  - ...
- New behavior:
  - Check cache
  - Check allocation policy quotas
  - ...

This will enable bringing rate limits to Allocation policies without rate-limiting queries that would have otherwise been cached

### Other Changes
- Entire `QueryResult` is now cached and returned instead of just `Result`
  - This means `QueryExtraData` is cached with each query containing:
    - `stats`
    - `sql`
    - `experiments`

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -34,7 +34,7 @@ class QueryResultOrError:
     differently"""
 
     query_result: QueryResult | None
-    error: Exception | None
+    error: SerializableException | None
 
     def __post_init__(self) -> None:
         assert self.query_result is not None or self.error is not None
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -15,7 +15,7 @@ from snuba.state import set_config as set_runtime_config
 from snuba.utils.metrics.wrapper import MetricsWrapper
 from snuba.utils.registered_class import RegisteredClass, import_submodules_in_directory
 from snuba.utils.serializable_exception import JsonSerializable, SerializableException
-from snuba.web import QueryException, QueryResult
+from snuba.web import QueryResult
 
 logger = logging.getLogger("snuba.query.allocation_policy_base")
 CAPMAN_PREFIX = "capman"
@@ -34,7 +34,7 @@ class QueryResultOrError:
     differently"""
 
     query_result: QueryResult | None
-    error: QueryException | None
+    error: Exception | None
 
     def __post_init__(self) -> None:
         assert self.query_result is not None or self.error is not None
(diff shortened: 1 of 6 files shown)


snuba/query/allocation_policies/__init__.py
```diff
@@ -34,7 +34,7 @@ class QueryResultOrError:
     differently"""
 
     query_result: QueryResult | None
-    error: QueryException | None
+    error: Exception | None
```
volokluev, line 37: This should probably be a SerializableException, any Exception that gets raised by db_query should be that.

snuba/web/db_query.py
```diff
@@ -547,7 +423,37 @@ def _get_query_settings_from_config(
     return clickhouse_query_settings
 
 
-def _raw_query(
```
volokluev, line 550: the `_raw_query` docstring is no longer accurate, please update it

tests/web/test_db_query.py
```diff
@@ -294,6 +254,8 @@ def _update_quota_balance(
         ) -> None:
             return
 
+    query, storage, attribution_info = _build_test_query("count(distinct(project_id))")
```
volokluev, line 257: the comment at the top of this test no longer applies

volokluev (commented):
Good PR overall but there's some crucial test coverage missing from here. We should have a test confirm our new behavior: 

When the query is cached, the allocation policy does not run

snuba/web/db_query.py
rahul-kumar-saini: Let me know if the docstrings are appropriate, I moved most of it to `db_query()`

snuba/query/allocation_policies/__init__.py
rahul-kumar-saini: I'm not sure this works, eg `TimeoutError` is a builtin OSError that we raise and handle. We only encapsulate it into a QueryException at the top level so somehow we'd need to bring allocation policy update to happen after it is wrapped. Other solution is just to pass any `Exception` in `QueryResultOrError`, would this break something?

volokluev: Wait a minute, why does this need to be updated at all? If the cache pre-empts the allocation policy, why would we need to pass cache errors to the allocation policy?

volokluev: We talked offline, this behavior actually did change. This should be okay

volokluev (approved):

volokluev (approved):

PR #4507, opened 2023-07-14
feat(group-attributes): expose storage and entity, process messages

barkbarkimashark:
Part 2 of getting group_attributes dataset stood up. This should only be made available in CI and dev. 

* defines the topic for consuming write messages and the dead letter topic 
* defines the storage for the dataset
* defines the entity
* implements a basic processor to start writing to the table

I'll create a follow-up PR to define the join relationships between the entities along with appropriate tests. Just wanna start processing the messages and begin writing to the table in this PR.

I verified this locally by doing the following:
* Running snuba locally with this consumer enabled: `ENABLE_GROUP_ATTRIBUTES_CONSUMER="1" snuba devserver` 
* Produce a message to the topic using an example json message from [sentry-kafka-schemas](https://github.com/getsentry/sentry-kafka-schemas/blob/main/examples/group-attributes/1/group_created.json): `kcat -b localhost:9092 -X broker.address.family=v4 -t group-attributes -T -P examples/group-attributes/1/group_created.json` 
* Verifying the row gets created in clickhouse: `select * from group_attrbiutes_local`

### DIFF
--- a/snuba/datasets/configuration/group_attributes/entities/group_attributes.yaml
+++ b/snuba/datasets/configuration/group_attributes/entities/group_attributes.yaml
@@ -48,9 +48,9 @@ query_processors:
   - processor: BasicFunctionsProcessor
 
 validate_data_model: error
-validators: []
-#  - validator: EntityRequiredColumnValidator
-#    args:
-#      required_filter_columns: ["project_id"]
+validators:
+  - validator: EntityRequiredColumnValidator
+    args:
+      required_filter_columns: ["project_id"]
 
 required_time_column: null
--- a/snuba/datasets/configuration/group_attributes/entities/group_attributes.yaml
+++ b/snuba/datasets/configuration/group_attributes/entities/group_attributes.yaml
@@ -1,0 +1,56 @@
+version: v1
+kind: entity
+name: group_attributes
+
+schema:
+  [
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: group_id, type: UInt, args: { size: 64 } },
+
+    { name: group_status, type: UInt, args: { size: 8 } },
+    { name: group_substatus, type: UInt, args: { size: 8, schema_modifiers: [ nullable ] } },
+    { name: group_first_seen, type: DateTime },
+    { name: group_num_comments, type: UInt, args: { size: 64 } },
+
+    { name: assignee_user_id, type: UInt, args: { size: 64, schema_modifiers: [ nullable ] } },
+    { name: assignee_team_id, type: UInt, args: { size: 64, schema_modifiers: [ nullable ] } },
+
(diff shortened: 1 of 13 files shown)


snuba/datasets/configuration/group_attributes/entities/group_attributes.yaml
```diff
@@ -0,0 +1,56 @@
...
+#  - processor: ResourceQuotaProcessor
+#    args:
+#      project_field: project_id
+  - processor: BasicFunctionsProcessor
+
+validate_data_model: error
+validators: []
+#  - validator: EntityRequiredColumnValidator
+#    args:
+#      required_filter_columns: ["project_id"]
```
barkbarkimashark, line 54: Not totally sure how these processors interacts with joins, so decided to leave them out for now and modify it when we test joins.

rahul-kumar-saini: You can probably still keep `project_id` as required for querying, I believe this is a required column on most if not all entities

snuba/datasets/configuration/group_attributes/storages/group_attributes.yaml
```diff
@@ -0,0 +1,55 @@
...
+      { name: offset, type: UInt, args: { size: 64 } },
+    ]
+  local_table_name: group_attributes_local
+  dist_table_name: group_attributes_dist
+
+allocation_policies:
+  - name: PassthroughPolicy
+    args:
+      required_tenant_types:
+        - blank
```
rahul-kumar-saini, line 43: In full prod version you should use the BytesScannedWindowAllocationPolicy to throttle queries by org above a certain threshold of bytes scanned. We can discuss further what this means if you'd like.

snuba/settings/validation.py
```diff
@@ -72,6 +73,7 @@ def validate_settings(locals: Mapping[str, Any]) -> None:
         "snuba-dead-letter-metrics-counters",
         "snuba-dead-letter-generic-events",
         "snuba-dead-letter-querylog",
+        "snuba-dead-letter-group-attributes",
```
rahul-kumar-saini, line 76: Ensure all of your topics exist in production - and the ops repo is updated with the topics in [prod_settings.py](https://github.com/getsentry/ops/blob/master/k8s/services/snuba/templates/prod_settings.py#L244). Without this, Snuba will fail to deploy due to topics mismatch.

rahul-kumar-saini (approved):
LGTM, left some notes

snuba/datasets/configuration/group_attributes/storages/group_attributes.yaml
barkbarkimashark: Yup, that makes sense to me. Will create a follow-up PR with more robust settings for prod when we're closing to launching this dataset.

PR #4508, opened 2023-07-14
feat(consumer): add health check step to consumers

dbanda:
Add a health check step to consumers. The health check touches a file `tmp/health.txt`  at regular intervals. In k8s, we can check for this file to see if our consumer is up and running and healthy. This can be use avoid k8s rolling over to consumer pods that are unhealthy or haven't yet completed their setup.

needs: https://github.com/getsentry/arroyo/pull/274

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -149,6 +149,12 @@ logger = logging.getLogger(__name__)
     "--max-poll-interval-ms",
     type=int,
 )
+@click.option(
+    "--health-check-file",
+    default=None,
+    type=str,
+    help="Arroyo will touch this file at intervals to indicate health. If not provided, no health check is performed.",
+)
 def consumer(
     *,
     storage_name: str,
@@ -176,6 +182,7 @@ def consumer(
     log_level: Optional[str] = None,
     profile_path: Optional[str] = None,
     max_poll_interval_ms: Optional[int] = None,
+    health_check_file: Optional[str] = None,
 ) -> None:
 
     setup_logging(log_level)
@@ -238,6 +245,7 @@ def consumer(
         slice_id=slice_id,
         join_timeout=join_timeout,
         max_poll_interval_ms=max_poll_interval_ms,
+        health_check_file=health_check_file,
         enforce_schema=enforce_schema,
     )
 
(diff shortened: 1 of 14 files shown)


tests/test_consumer.py
```diff
@@ -72,6 +74,7 @@ def write_step() -> ProcessedMessageBatchWriter:
             ),
         )
 
+    health_check_file = tempfile.mktemp()
```
untitaker, line 77: you can use the pytest tmpdir fixture here

snuba/consumers/consumer_builder.py
```diff
@@ -239,6 +239,7 @@ def build_streaming_strategy_factory(
             input_block_size=self.input_block_size,
             output_block_size=self.output_block_size,
             initialize_parallel_transform=setup_sentry,
+            health_check_file="/tmp/health.txt",
```
untitaker, line 242: I would not hardcode this sort of thing in the Snuba codebase, and make it configurable. If our ops setup is not generic enough to deal with automatically adding this parameter to all snuba consumers (and configuring health checks appropriately), we should fix that instead.

snuba/cli/consumer.py
```diff
@@ -149,6 +149,11 @@
     "--max-poll-interval-ms",
     type=int,
 )
+@click.option(
+    "--health-check-file",
+    type=str,
+    help="Arroyo will touch this file at intervals to indicate health. If not provided, no health check is performed.",
```
lynnagara, line 155:
```suggestion
    help="The consumer will touch this file at intervals to indicate health. If not provided, no health check is performed.",
```

snuba/consumers/strategy_factory.py
```diff
@@ -166,4 +169,7 @@ def flush_batch(
                 strategy, self.__max_messages_to_process, 10.0
             )
 
+        if self.__health_check_file is not None:
+            strategy = Healthcheck(self.__health_check_file, strategy)
```
ayirr7, line 173: Just a question as I'm trying to understand this better, this means that the Healthcheck would be the first step of the ProcessingStrategy?

dbanda: yes

tests/test_consumer.py
dbanda: fixed

snuba/consumers/consumer_builder.py
dbanda: fixed

untitaker (approved):

PR #4510, opened 2023-07-14
feat(admin): Add new trace logs view

enochtangg:
### Overview
The following PR is responsible for adding a formatted view for the tracing tool in Snuba admin. The purpose of this view is to enable users to quickly analyze the performance of their query at a quick glance without parsing through raw trace logs. This tool is used to help users can answer questions like: 
* Which part of my query is slow (aggregation step or filtering step)?
* Have I provided enough primary keys to ensure quicker filtering of parts?
* What is the read performance and memory consumption?
* How many thread did ClickHouse actually used behind each node?
* Were any skip indexes applied?
* What filtering algorithm did ClickHouse choose to use?
* What aggregation method did ClickHouse use?

Demo:
30a49c3f-0e44-4b33-acd7-67ed313d77f5

### Testing
A test case was added to validate the tracing formatter.


dbanda:
Will this show the index range? I noticed that was missing sometimes in the old one. Not sure if that's easy to add.

dbanda (approved):

PR #4512, opened 2023-07-14
feat(subscriptions): add clickhouse connection check

dbanda:
Our subscriptions containers are starting up before the connections to clickhouse are ready. This add a check to attempt to connect to clickhouse for 60s before starting up the consumers.

### DIFF
--- a/snuba/cli/subscriptions_scheduler.py
+++ b/snuba/cli/subscriptions_scheduler.py
@@ -4,6 +4,7 @@ from contextlib import closing
 from typing import Any, Optional, Sequence
 
 import click
+import structlog
 from arroyo import configure_metrics
 from arroyo.backends.kafka import KafkaProducer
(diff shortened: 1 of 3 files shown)


snuba/cli/subscriptions_scheduler.py
```diff
@@ -146,6 +150,10 @@ def subscriptions_scheduler(
         storage is not None
     ), f"Entity {entity_name} does not have a writable storage by default."
 
+    logger.info("Checking Clickhouse connections")
+    cluster = storage.get_cluster()
+    check_clickhouse_connections([cluster])
```
lynnagara, line 155: nit: I'd move this to the line before `build_consumer`, rather than in the middle of validating the CLI args

lynnagara (approved):
thanks!

PR #4513, opened 2023-07-14
feat: Add group attributes storage set

lynnagara:
CI didn't like it being part of the same PR as https://github.com/getsentry/snuba/pull/4496/, so let's try this?


lynnagara:
it will be allowed

PR #4520, opened 2023-07-14
ref: Simplify querylog processor more

lynnagara:
We don't need this all this fallback logic, "profile" is not optional: https://github.com/getsentry/snuba/blob/e369d972756d7cfb4415be942f1b38b371844a8f/snuba/querylog/query_metadata.py#L192


evanh (approved):

lynnagara:
We don't need this all this fallback logic, "profile" is not optional: https://github.com/getsentry/snuba/blob/e369d972756d7cfb4415be942f1b38b371844a8f/snuba/querylog/query_metadata.py#L192

PR #4521, opened 2023-07-15
fix(ci): improve caching

dbanda:
Use the docker build step to use more aggressive caching. Also fixes a bug where we were not always pulling images into the cache. I found this shaved off a 2-3 minutes in CI since we were not always building the image at each step.


evanh (commented):
Looks good but CI is failing.

dbanda:
> Looks good but CI is failing.

it was a flake. Passing now.

evanh (approved):

PR #4532, opened 2023-07-18
ref(capman): split the rate limit function into a start and finish phase

volokluev:
The allocation policy infrastructure assumes theres a start and a finish of a query, the rate limit code is a context manager which assumes everything happens in a specific block.

This PR:

* Splits apart the rate limit function into two reusable components
* Adds a test for the fail open case for the rate limiter (which wasn't tested before)
* maintains the context manager abstraction so everything still works as before

### DIFF
--- a/snuba/state/rate_limit.py
+++ b/snuba/state/rate_limit.py
@@ -185,7 +185,7 @@ def rate_limit_start_request(
     #
     # it is fine to only perform this cleanup for the shard of the current
     # query, because on average there will be many other queries that hit other
-    # shards and perform cleanup therea
+    # shards and perform cleanup there
     pipe.zremrangebyscore(query_bucket, "-inf", "({:f}".format(now - rate_history_sec))
 
     # Now for the tricky bit:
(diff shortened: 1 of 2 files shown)


snuba/state/rate_limit.py
```diff
@@ -166,60 +166,16 @@ def _get_bucket_key(prefix: str, bucket: str, shard_id: int) -> str:
...
+    rate_limit_shard_factor,
+) -> RateLimitStats:
     now = time.time()
 
-    query_id_uuid = uuid.uuid4()
-    query_id = str(query_id_uuid)
-
     # Compute the set shard to which we should add and remove the query_id
-    bucket_shard = int(query_id_uuid) % rate_limit_shard_factor
+    bucket_shard = hash(query_id) % rate_limit_shard_factor
```
volokluev, line 178: this change is necessary to allow any arbitrary string to be used as the query id. If people feel really strongly about this I can change it but I would also have to change the allocation policy to require UUIDs as well and I don't think there's a meaningful difference

evanh (approved):

PR #4544, opened 2023-07-19
Update gocd jsonnet libs and fix rollbacks

mattgauntseo-sentry:


mattgauntseo-sentry:
The error with pipeline validation is something we hadn't seen prior to changing our validation logic, we'll work on fixing that shortly, but it's safe to ignore.

rahul-kumar-saini (approved):

PR #4546, opened 2023-07-20
fix(cardinality-analyzer): Fix CSV escaping

gggritso:
The "Copy to clipboard (CSV)" button is great, but it doesn't escape double quotes in the CSV, so the content is sometimes mangled. This is a fix.

Screenshot 2023-07-20 at 12 04 37 PM

## Changes

- Prettier
- Remove unused variables
- Add some more typing
- Extract CSV class
- Add spec


volokluev (approved):
You did a good job George

PR #4548, opened 2023-07-20
feat(replay): pass start_time as received header

bmckerry:
The [ReplayEventKafkaMessage](https://github.com/getsentry/relay/blob/master/relay-server/src/actors/store.rs#L820-L832) struct doesn't have a `received` header - but it does have `start_time`, which is set [here](https://github.com/getsentry/relay/blob/master/relay-server/src/actors/store.rs#L651C13-L651C74):
```rust
start_time: UnixTimestamp::from_instant(start_time).as_secs()
```

This is the same as the Replay Recording `received` header: https://github.com/getsentry/relay/blob/master/relay-server/src/actors/store.rs#L687
As well as the profiles `received` header (which an E2E SLO is already tracking): https://github.com/getsentry/relay/blob/master/relay-server/src/actors/store.rs#L623


tests/datasets/test_replays_processor.py
```diff
@@ -162,7 +162,9 @@ def serialize(self) -> Mapping[Any, Any]:
 
         return {
             "type": "replay_event",
-            "start_time": datetime.now().timestamp(),
+            "start_time": self.timestamp
+            if isinstance(self.timestamp, (float, int))
+            else datetime.now().timestamp(),
```
bmckerry, line 167: Previously, the test `serialize()` function was just ignoring whatever was passed as a `timestamp` and sending it's own timestamp value.
Actually using what was passed as `timestamp` caused the tests involving bad timestamps to fail, so this keeps the behaviour of 'if timestamp is bad just chuck it & send a real timestamp'.

tests/datasets/test_replays_processor.py
```diff
@@ -160,11 +162,12 @@ def serialize(self) -> Mapping[Any, Any]:
             "extra": {},
         }
 
+        start_time = datetime.now().timestamp()
+        if header_overrides and header_overrides["start_time"]:
+            start_time = header_overrides["start_time"]
         return {
```
hubertsentry, line 168:
```
headers = {
    "start_time": datetime.now().timestamp(),
    "type": "replay_event",
    "project_id": 1,
    "retention_days": 30
}.update(header_overrides or {})

return {**headers, **{"payload": list(bytes(json.dumps(replay_event).encode()))}}
```

tests/datasets/test_replays_processor.py
```diff
@@ -162,16 +162,16 @@ def serialize(
...
 
-        start_time = datetime.now().timestamp()
-        if header_overrides and header_overrides["start_time"]:
-            start_time = header_overrides["start_time"]
-        return {
+        headers = {
+            "start_time": datetime.now().timestamp(),
             "type": "replay_event",
-            "start_time": start_time,
-            "replay_id": self.replay_id,
```
hubertsentry, line 166: oh, sorry, I probably missed `replay_id` in my previous comment

hubertsentry (approved):

JoshFerge (approved):

ayirr7 (approved):

PR #4560, opened 2023-07-24
Rename monitor to s4s and add first customer

mattgauntseo-sentry:
This switches the name of monitor to s4s (just easier to understand for sentaurs).

The `customer-1` pipeline is the first customer pipeline being added, this should follow the same process as s4s.

NOTE: This uses a new helper to determine a ST deploy or not, this means naming changes like the monitor -> s4s change will be less error prone.

#skip-changelog


joshuarli (approved):

mattgauntseo-sentry:
Closing in favor of https://github.com/getsentry/super-big-consumers/pull/169 which includes this change.

PR #4574, opened 2023-07-27
feat(prod-queries): Add button to view all allowed projects

davidtsuk:
This adds the ability to view all allowed project ids.
image


rahul-kumar-saini (approved):
I'd maybe add a simple api test

PR #4576, opened 2023-07-27
cleanup(capman): remove project throttler

volokluev:
This is an allocation policy decision that is not necessary given that we have the BytesScannedWindow policy. it's also fairly opaque. If we decide we want it back we can reintroduce it as its own policy.

Currently this code looks at how many concurrent queries the project is doing and then assigns max_threads using this formula:

```
max_threads = max(1, table_max_threads - project_concurrent_queries)
```

This has been disabled all of today and we're [fine](https://www.notion.so/sentry/Experiment-5-5-Turn-of-concurrent-throttling-across-datasets-3e9cdcf441a44804876dfe4b7fcac319)


rahul-kumar-saini (approved):
This makes sense to me, maybe should be looked at by others as well

PR #4577, opened 2023-07-27
fix(replays): add uuid processor to replay_id column

JoshFerge:
This will make it so replay_ids dashes are stripped, akin to how our other uuid field dashes are stripped.

### DIFF
--- a/snuba/datasets/configuration/discover/storages/discover.yaml
+++ b/snuba/datasets/configuration/discover/storages/discover.yaml
@@ -132,6 +132,7 @@ query_processors:
         - transaction_name
         - environment
         - project_id
+        - replay_id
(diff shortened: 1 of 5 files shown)


volokluev:
Is there anyone using the replay id in queries at the moment that this would break?

JoshFerge:
@volokluev  no. this field isn't available to be queried through any of our APIs yet.

volokluev (approved):

mcannizz:
@JoshFerge this PR has been open for a long time. Do you still want to make this change or should we close it?

JoshFerge:
can go ahead and close it, thanks. I don't know that this would have done much perf wise..

PR #4579, opened 2023-07-29
Update gocd-jsonnet to v1.4.1

mattgauntseo-sentry:
This change adds customer 3 and 4 to pipedream and test deployments for customers 5 and 6

NOTE: This is the final addition of deployments that I've planned

### DIFF
--- a/gocd/generated-pipelines/rollback-snuba.yaml
+++ b/gocd/generated-pipelines/rollback-snuba.yaml
@@ -3,16 +3,16 @@ pipelines:
   rollback-snuba:
     display_order: 1
     environment_variables:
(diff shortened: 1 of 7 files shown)


joshuarli (approved):

rahul-kumar-saini (approved):
I'm not gonna block this as we already have a couple ST customers in the pipeline but I do have a concern:

- Currently if something goes wrong in prod we have the runtime config to fall back on to apply quick changes in a feature flag fashion
- Eg, today I used it to disable the cache live to deal with an incident
- This same functionality isn't available for ST as we don't have a Snuba Admin deployment there and would rely on ops intervention to do something similar

Now that deployments go to SaaS and then ST right away, we have less time to deal with things breaking before ST deployment starts with no way for us to modify live configs there. Maybe the ST deployment step should be a manual start?

Eg.
- Start deploy
- S4S + SaaS goes out automatically
- Pipeline stops there till manually unblocked to start ST deployments

PR #4584, opened 2023-08-01
feat(async-queries): Add ability to override query settings for async queries

davidtsuk:
Adding the ability to override query settings using the `async_query_settings/` prefix in runtime config. These async query settings take priority over regular query settings, but cluster specific query settings take priority over async query settings.

### DIFF
--- a/snuba/query/query_settings.py
+++ b/snuba/query/query_settings.py
@@ -60,6 +60,10 @@ class QuerySettings(ABC):
     def set_clickhouse_settings(self, settings: MutableMapping[str, Any]) -> None:
         pass
 
+    @abstractmethod
+    def get_asynchronous(self) -> bool:
+        pass
+
 
 # TODO: I don't like that there are two different classes for the same thing
 # this could probably be replaces with a `source` attribute on the class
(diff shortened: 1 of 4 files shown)


volokluev:
Is this just so that you can change the `max_threads` for an async query to be different?

davidtsuk:
@volokluev For now, yes, but I also added async as a flag in the query settings which will be used when workers run the query.

snuba/query/query_settings.py
```diff
@@ -79,6 +83,7 @@ def __init__(
         # TODO: is this flag still relevant?
         legacy: bool = False,
         referrer: str = "unknown",
+        asynchronous: bool = False,
```
evanh, line 86: If you want this flag to be populated from the Request itself, you'll need to add it to here: https://github.com/getsentry/snuba/blob/master/snuba/request/schema.py#L157

Otherwise this will error. Also, this field should probably be added to the SnQL SDK request `Flags` class at some point.

davidtsuk: yeah, I'm working on a PR that uses the flag and I've added it in the schema there.

evanh (approved):

PR #4587, opened 2023-08-02
feat(capman): pass through subscriptions for new concurrent rate limiter

volokluev:
Our current policy is to not rate limit subscriptions, this puts that policy in code

### DIFF
--- a/tests/query/allocation_policies/test_concurrent_rate_limit_policy.py
+++ b/tests/query/allocation_policies/test_concurrent_rate_limit_policy.py
@@ -148,7 +148,11 @@ def test_tenant_selection(policy):
 
 def test_pass_through(policy: ConcurrentRateLimitAllocationPolicy) -> None:
     ## should not be blocked because the subscriptions_executor referrer is not rate limited
-    for i in range(MAX_CONCURRENT_QUERIES * 2):
-        policy.get_quota_allowance(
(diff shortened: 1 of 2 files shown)


tests/query/allocation_policies/test_concurrent_rate_limit_policy.py
```diff
@@ -144,3 +144,11 @@ def test_tenant_selection(policy):
     )
     with pytest.raises(AllocationPolicyViolation):
         policy._get_tenant_key_and_value({})
+
+
+def test_pass_through(policy: ConcurrentRateLimitAllocationPolicy) -> None:
```
nikhars, line 149: I don't see an assertion in this test. Is that intended?

volokluev: yes it is intended. If this was not a passthrough referrer, the request would be rejected

nikhars: Looking at the code it seems like to reject a request an exception is raised. If that is true, you could potentially explicitly check that exception is not raised in the test. Just makes things a bit more clearer. Something along the lines of the following. But this is not blocking so will approve the change
```
try:
   policy.get_quota_allowance()
except Exception:
   pytest.fail("Reason")
```

nikhars (approved):

PR #4589, opened 2023-08-03
feat(async-queries): Add celery workers to run snuba queries and an API

davidtsuk:
# API
1. Updated existing endpoint so that it can receive asynchronous queries by specifying an asynchronous flag in the body. When an asynchronous query is received, it is queued and the query id (task id) is returned.
2. Added a new polling endpoint used to check the status of the query via query id and get the results when the query is complete.

# Celery
Added a simple CLI helper that runs a worker. Also, added some initialization logic to determine the redis connection.

# Note
Can't be merged until prod_settings are updated with redis instance.

### DIFF
--- a/snuba/workers/tasks.py
+++ b/snuba/workers/tasks.py
@@ -1,0 +1,89 @@
+import logging
+from typing import Any, Dict, Mapping, MutableMapping
+
+from snuba import settings
+from snuba.clickhouse.errors import ClickhouseError
+from snuba.datasets.dataset import Dataset
+from snuba.datasets.storage import StorageNotAvailable
+from snuba.query.allocation_policies import AllocationPolicyViolations
+from snuba.query.exceptions import QueryPlanException
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.request.schema import RequestSchema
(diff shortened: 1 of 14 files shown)


snuba/workers/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+
+logger = logging.getLogger("snuba.tasks")
+
+
+@celery_app.task(serializer="pickle")
+def send_query(
+    dataset: Dataset, body: Dict[str, Any], timer: Timer, referrer: str
+) -> Mapping[str, Any]:
+    schema = RequestSchema.build(HTTPQuerySettings)
+    request = build_request(
```
evanh, line 28: Since this is happening outside the normal flask app, all the handlers we have to catch exceptions and format them properly are not going to work. I think you will have to add something similar to ensure that the results stored in redis are a consistent format.

evanh (commented):
This looks generally good but needs tests.

volokluev:
Can we have two different endpoints for synchronous/asynchronous queries? I don't like the payload and behavior of an endpoint drastically changing due to a field in the JSON

PR #4597, opened 2023-08-07
feat(replays): Remove nullability requirement from most columns

cmanallen:
- I re-ordered the entity and storage yaml files to be in alphabetical order.  The diff is very large...  if this hurts code review too much I can remove it.
- I left the processor changes on this branch.  It seems like the processor should go first and then the migration second.  I don't see a need to break these two changes out since migrations are manually run.
- There's an API change that needs to be merged first.  I'll link it when its up.


cmanallen:
Closing in favor of a split PR.

PR #4598, opened 2023-08-07
feat(replays): Update processor to send the empty state rather than null

cmanallen:
Storing columns as nullable has performance and storage implications.  So lets remove it!

There's an API change that needs to be merged first: https://github.com/getsentry/sentry/pull/54316


nikhars (commented):
Do you have any queries which perform NULL checks on any of these columns which you are converting to non-NULL? You might want to check.

Other thing, I do not know if making a nullable column non-nullable perform's a whole table rewrite or not. Basically when the migration is applied, does it apply to new data only or also to the old data. Might want to check on that as well.

If it does a whole table rewrite, you would have to wait 90 days before you can actually apply the migration to make the columns non-nullable.

evanh:
> you would have to wait 90 days before you can actually apply the migration to make the columns non-nullable.

You will also have to figure out how this will work for self-hosted customers.

cmanallen:
@nikhars @evanh 

> Do you have any queries which perform NULL checks on any of these columns which you are converting to non-NULL? You might want to check.

I have checked.  I've replaced usage of `any` with `anyIf` which allows me to check for the empty state.  I will check again to make sure that nothing has slipped by.

> If it does a whole table rewrite, you would have to wait 90 days before you can actually apply the migration to make the columns non-nullable.

Its my understanding that Clickhouse does not store nulls on the column file.  It stores "nulls" in a separate file (uint8s either 0 or 1).  Dropping the null constraint should leave the empty state in the primary column and just drop the nullable dictionary mapping.

That being said I have no idea what ClickHouse will do so I will research and find documentation for this case.

nikhars (approved):

PR #4600, opened 2023-08-07
feat(replays): Remove nullability from most columns

cmanallen:
I re-ordered the entity and storage yaml files to be in alphabetical order.  The diff is very large...  if this hurts code review too much I can remove it.

Depends on: https://github.com/getsentry/snuba/pull/4598

The following columns are now `NOT NULL`.

- browser_name
- browser_version
- device_brand
- device_family
- device_model
- device_name
- dist
- environment
- error_sample_rate
- ip_address_v4
- ip_address_v6
- is_archived
- os_name
- os_version
- platform
- release
- replay_type
- sdk_name
- sdk_version
- session_sample_rate
- user_email
- user_id
- user_name
- user


evanh:
> I re-ordered the entity and storage yaml files to be in alphabetical order.

I believe these are currently ordered the same way the columns are added (new columns require an `after` keyword). Up to you how you want the sorting though. That also has the added benefit of showing things in the same order they are in the primary sorting key. It's important to remember this isn't a postgres table, and the data isn't meant to be filtered arbitrarily. The order the columns are in the primary key determines how you should filter the table.

onewland:
@cmanallen before merging this, we should verify that the column doesn't get rewritten. 

We're tracking the work internally here: https://getsentry.atlassian.net/browse/SNS-2451

cmanallen:
Conclusion seems to be: update the processor to insert empty values, let the null values drain (90 day retention), when the null values have flushed the nullable constraint will be dropped.

PR #4603, opened 2023-08-07
feat(replays): Add materialized counts for array columns

cmanallen:
The replays' queries have had trouble controlling their memory usage.  By materializing the length of the urls and error_ids arrays we can order by those values with having to bring the contents of the column into memory.

Example of the speed up summing the materialized column provides (using contrived data in a local environment):

```
┌─memory_usage─┬─query_duration_ms─┬─query─────────────────────────────────────────────────────────┐
│      5034369 │                26 │ select sum(length(urls)) from replays_test group by replay_id │
│      5034369 │                12 │ select sum(length(urls)) from replays_test group by replay_id │
│      5034369 │                29 │ select sum(length(urls)) from replays_test group by replay_id │
│      5034369 │                29 │ select sum(length(urls)) from replays_test group by replay_id │
│            0 │                 5 │ select sum(count_urls) from replays_test group by replay_id   │
│            0 │                 6 │ select sum(count_urls) from replays_test group by replay_id   │
│            0 │                17 │ select sum(count_urls) from replays_test group by replay_id   │
│            0 │                10 │ select sum(count_urls) from replays_test group by replay_id   │
└──────────────┴───────────────────┴───────────────────────────────────────────────────────────────┘
```


snuba/migrations/group_loader.py
```diff
@@ -172,6 +172,7 @@ def get_migrations(self) -> Sequence[str]:
             "0009_add_dom_index_columns",
             "0010_add_nullable_columns",
             "0011_add_is_dead_rage",
+            "0015_materialize_counts",
```
evanh, line 175: So this is a bit of a nit:

If these PRs are merged out of order, this list will be out of order. As in you'll have `0015_` before `0014_`. I know these migrations are separate and the ordering doesn't really matter. But having the numbers be out of order would drive me crazy, and it's not trivial to re-order these after the fact, because we store them in order in a table in production. That's what we use to rollback migrations.

I know it's more work, but it would be great if you could make sure that these numbers are ordered correctly as you merge them. I know it will cause some merge conflicts but I think it will be worth it.

cmanallen: @evanh Yeah I can do that.  I was thinking we'd get them all approved and merge in order.  But I can update this so we can merge it "out of order" if it makes life easier.

evanh: OK got it. I wasn't sure if you planned to merge piecemeal or all at once.

evanh (approved):

cmanallen:
@evanh It seems the other pulls are going to move much more slowly.  So I'm adjusting this migration's number to be in the correct order.

JoshFerge (approved):

PR #4607, opened 2023-08-09
feat(spans): switch to only spans consumer

dbanda:
This switches to the only spans consumer pipeline. This is to be merged after  sentry consumer and kafka schemas PRs are merged in.

### DIFF
--- a/snuba/datasets/processors/spans_processor.py
+++ b/snuba/datasets/processors/spans_processor.py
@@ -200,8 +200,8 @@ class SpansMessageProcessor(DatasetMessageProcessor):
 
         status = sentry_tags.get("status", None)
         if status:
-            processed["status"] = status
             int_status = SPAN_STATUS_NAME_TO_CODE.get(status, UNKNOWN_SPAN_STATUS)
+            processed["status"] = int_status
         else:
             int_status = UNKNOWN_SPAN_STATUS
         processed["span_status"] = int_status
--- a/snuba/datasets/processors/spans_processor.py
+++ b/snuba/datasets/processors/spans_processor.py
@@ -1,11 +1,11 @@
-import copy
 import logging
 import numbers
 import random
 import uuid
 from datetime import datetime
-from typing import Any, Dict, Mapping, MutableMapping, MutableSequence, Optional, Tuple
+from typing import Any, Mapping, MutableMapping, MutableSequence, Optional, Tuple
 
+from sentry_kafka_schemas.schema_types.snuba_spans_v1 import SpanEvent
 from sentry_relay.consts import SPAN_STATUS_NAME_TO_CODE
 
 from snuba import environment, state
@@ -77,10 +77,10 @@ class SpansMessageProcessor(DatasetMessageProcessor):
     from the transactions topic and de-normalize it into the spans table.
     """
 
-    def __extract_timestamp(self, field: int) -> Tuple[datetime, int]:
+    def __extract_timestamp(self, timestamp_sec: float) -> Tuple[datetime, int]:
         # We are purposely using a naive datetime here to work with the rest of the codebase.
         # We can be confident that clients are only sending UTC dates.
-        timestamp = _ensure_valid_date(datetime.utcfromtimestamp(field))
+        timestamp = _ensure_valid_date(datetime.utcfromtimestamp(timestamp_sec))
         if timestamp is None:
             timestamp = datetime.utcnow()
         milliseconds = int(timestamp.microsecond / 1000)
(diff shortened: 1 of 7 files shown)


snuba/datasets/processors/spans_processor.py
```diff
@@ -224,165 +181,53 @@ def _process_measurements(
...
     ) -> None:
         """
         TODO: For the top level span belonging to a transaction, we do not know how to fill these
               values yet. For now lets just set them to their default values.
         """
-        processed["module"] = ""
-        processed["action"] = ""
-        processed["domain"] = ""
-        processed["status"] = 0
+        sentry_tags = span_event["sentry_tags"]
```
nikhars, line 192: Let's assume all data in `sentry_tags` to be optional for now. But the data model has a few required fields. Can you confirm that we are mandating all required fields here and all optional fields are populated with safe defaults in case they are missing?

snuba/datasets/processors/spans_processor.py
```diff
@@ -224,165 +181,53 @@ def _process_measurements(
                 )
 
     @staticmethod
-    def _process_module_details(
+    def _process_sentry_tags(
```
nikhars, line 184: We would need to do the following as well. Do you want to merge this PR and handle this case later? Or do you want to do the migration of creating the column first and then handle the case in this PR itself?
1. Add `sentry_tags` column to the spans clickhouse cluster.
2. When 1 is done, remove all fields from `sentry_tags` which are promoted to top level columns.
3. Remaining `sentry_tags` need to be populated in the clickhouse `sentry_tags` column.

snuba/datasets/processors/spans_processor.py
```diff
@@ -393,13 +238,11 @@ def process_message(
...
-            self._process_module_details(processed, event_dict)
+            self._process_span_event(processed, span_event)
+            self._process_tags(processed, span_event)
+            self._process_measurements(processed, span_event)
+            self._process_sentry_tags(processed, span_event)
             processed_rows.append(processed)
-            processed_spans = self._process_spans(event_dict, common_span_fields)
-            processed_rows.extend(processed_spans)
 
         except Exception as e:
```
nikhars, line 247: At some point we would need to get rid of this catch all exception? Can you add a ticket in JIRA to track it? When we go to GA then we would need to clean this up.

snuba/datasets/processors/spans_processor.py
dbanda: Yes I'll handle in a later PR after a migration PR is done.

snuba/datasets/processors/spans_processor.py
```diff
@@ -77,131 +77,88 @@ class SpansMessageProcessor(DatasetMessageProcessor):
...
+        processed["is_segment"] = span_event["is_segment"]
+        parent_span_id: Optional[str] = span_event.get("parent_span_id", None)
+        if parent_span_id:
+            processed["parent_span_id"] = int(parent_span_id, 16)
+        transaction_id: Optional[str] = span_event.get("event_id", None)
+        if transaction_id:
+            processed["transaction_id"] = str(uuid.UUID(transaction_id))
+
+        # descriptions
+        processed["description"] = _unicodify(span_event.get("description", ""))
```
phacops, line 126: I don't think there's such a field in the schema https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-spans.v1.schema.json

snuba/datasets/processors/spans_processor.py
```diff
@@ -77,131 +77,88 @@ class SpansMessageProcessor(DatasetMessageProcessor):
...
+        processed["trace_id"] = str(uuid.UUID(span_event["trace_id"]))
+        processed["span_id"] = int(span_event["span_id"], 16)
+        processed["segment_id"] = processed["span_id"]
+        processed["is_segment"] = span_event["is_segment"]
+        parent_span_id: Optional[str] = span_event.get("parent_span_id", None)
+        if parent_span_id:
+            processed["parent_span_id"] = int(parent_span_id, 16)
+        transaction_id: Optional[str] = span_event.get("event_id", None)
+        if transaction_id:
+            processed["transaction_id"] = str(uuid.UUID(transaction_id))
```
phacops, line 123: We're not going to have the `transaction_id` anymore, are we?

snuba/datasets/processors/spans_processor.py
dbanda: Its an additional property and not required. It will be extracted if it exists otherwise we use empty string.

snuba/datasets/processors/spans_processor.py
dbanda: Looks like it. If the `event_id` is supplied we use it otherwise its null.

nikhars: I believe this was needed to link spans to transactions since at one point there was a requirement to link spans to their transactions. We can followup in a separate PR if that requirement is still true or not.

nikhars (approved):

snuba/datasets/processors/spans_processor.py
dbanda: Added a test with the empty `sentry_tags` and missing optional fields in `test_spans_payloads.py`

snuba/datasets/processors/spans_processor.py
phacops: It's in the tags though as `span.description`.

phacops (approved):

PR #4617, opened 2023-08-14
feat(cache): More generalized cache fail-open

rahul-kumar-saini:
### Overview
- Currently, the Redis Cache fails open on a select few Exceptions
- Differentiating between a redis failure vs clickhouse failure isn't really established in the fail-open logic
- This PR adds a way to identify errors raised by the value generation function execution step within the readthrough cache, since those are the only errors should actually fail the query
    - Any exception raised during value gen function is wrapped with `ValueGenerationError` which is then later caught and the original exception is raised
- All other errors fail open, ie executes the value generation function which routes the query to ClickHouse instead
    - Previously only a metric `snuba.read_through_cache.fail_open` was incremented upon cache fail-opens
    - Now a warning is also logged with the exception stack attached to diagnose why the cache failed

### DIFF
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -6,7 +6,7 @@ from typing import Callable, Optional
 
 from pkg_resources import resource_string
 
-from redis.exceptions import RedisError, ResponseError
+from redis.exceptions import ResponseError
 from snuba import environment, settings
 from snuba.redis import RedisClientType
 from snuba.state import get_config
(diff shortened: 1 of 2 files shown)


evanh (approved):

rahul-kumar-saini:
Reverting as result of Volo's [investigation](https://www.notion.so/sentry/Concurrent-Rate-Limit-Mondays-6aeccd4b556a404f801d9a36197064c0) for now.

PR #4619, opened 2023-08-15
feat(ci): Test Clickhouse 21

dbanda:
We had an issue today where a migration ran on Clickhouse 20 but failed on Clickhouse 21. This adds a test for Clickhouse 21. This is kept outside the matrix to allow it to be optional. This PR relies on using cached migrations in ci #4622  so that we can finish the tests in a reasonable amount of time.

### test updates
* For some discover API tests, Clickhouse 21 returns no rows whereas Clickhouse 20 would return a count of 0. To fix this, we try the asserts for both cases. 
* Because migrations seem to run faster on 21, we also make the setup function in the discover api test a fixture that runs after the clickhouse_db fixture to guarantee the ordering and avoid a race.
* In the cleanup test, we use a future Sunday date  as the base to avoid hitting the TTLs. Clickhouse 21 tends to work quicker and cleans out the TTL expired rows before the tests could finish.
* Invalid dates throw a 500 in Clickhouse 20 but error 400 in Clickhouse 21, so we are assert for both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -375,3 +375,51 @@ jobs:
           docker_repo: getsentry/snuba
           image_url: us.gcr.io/sentryio/snuba:${{ github.event.pull_request.head.sha || github.sha }}
           docker_password: ${{ secrets.DOCKER_HUB_RW_TOKEN }}
+
+  clickhouse-21:
+    needs: [linting, snuba-image]
+    name: Tests on Clickhouse 21
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout code
+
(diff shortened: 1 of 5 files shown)


nikhars (approved):
Nice. Assuming that you would update this PR since the cached migration is a different PR now. Rest of the changes look good to me.

lynnagara:
Nice! Should we make this a matrix where we can plug in various clickhouse versions? I'd love to run multiple clickhouse versions on the migrations at least in CI

dbanda:
> Nice! Should we make this a matrix where we can plug in various clickhouse versions? I'd love to run multiple clickhouse versions on the migrations at least or a subset of tests in CI

Thats the goal. I set it outside the matrix so that it optional for now so that we can test it for a while,  but in a subsequent PR I can move.

PR #4620, opened 2023-08-15
feat: Don't retry subscriptions queries which get timeout or OOM from clickhouse

lynnagara:
This is a follow up from INC-395. We should not constantly retry when we get these errors, as it blocks execution of all other subscriptions queries.


dbanda (commented):
For the queries that fit those error code, do we reschedule to be retried after a long delay or does that mean those queries are permanently dropped?

lynnagara:
The query will be permanently dropped for that window, but will be rerun on the next window, which I believe is most often at the next minute anyway

dbanda (approved):

PR #4621, opened 2023-08-15
fix(gocd): Use deploy-snuba-us as reference for last deployed

dbanda:
When checking for migrations changes between HEAD and the last known good deploy, use the `deploy-snuba-us` pipeline, which is the saas. In the new pipedream pipeline `deploy-snuba` is empty.

### DIFF
--- a/scripts/fetch_service_refs.py
+++ b/scripts/fetch_service_refs.py
@@ -27,7 +27,7 @@ def pipeline_passed(pipeline: Dict[str, Any]) -> bool:
 
 
 # print the most recent passing sha for a repo
-def main(pipeline_name: str = "deploy-snuba", repo: str = "snuba") -> int:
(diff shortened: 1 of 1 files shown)


nikhars (commented):
Should we have S4S as the first deploy to be checked? I am wondering if we do this in the US deploy, would it lead to inconsistencies between different stages of the pipeline? Because S4S would have run the migration and SaaS would have blocked it.

dbanda:
> Should we have S4S as the first deploy to be checked? I am wondering if we do this in the US deploy, would it lead to inconsistencies between different stages of the pipeline? Because S4S would have run the migration and SaaS would have blocked it.

I'm not sure about that. The incident with the replays migrations went through S4S without issue because the tables are smaller. SaaS is the biggest instance so we use it as a reference for the LKG. If a migration goes through on SaaS then it likely will go through the rest.

PR #4622, opened 2023-08-15
fix(ci): cache migrations in ci

dbanda:
Instead of running migrations from start to finish each time before a clickhouse, we can cache the end result of applying all migrations and use that instead. We are not adding new migrations within test so that should be fine. This will speed up CI.


volokluev (approved):
clever! I like it

PR #4623, opened 2023-08-15
feat(rust-consumer): Handle failed inserts

lynnagara:
Previously we were treating successful and failed ClickHouse inserts the same, and we simply commit the offset once we get a result from ClickHouse. It's probably better to retry in at least some scenarios. Now the thread will panic. That is probably also not quite right but it's better than silently failing at least for testing purposes.

There was also a bug in the ClickHouse test uncovered by this change. Now we run Snuba migrations in CI before the Rust tests, as it depends on the querylog table being there.

### DIFF
--- a/docker-compose.gcb.yml
+++ b/docker-compose.gcb.yml
@@ -1,5 +1,5 @@
 ---
-version: '3.4'
+version: "3.4"
 
 x-test-common: &test-common
   depends_on:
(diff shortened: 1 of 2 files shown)


nikhars (approved):
Please check if the changes to docker-compose.gcb.yml were intentional or not. The rust code changes look good.

lynnagara:
Yeah, the Clickhouse actually caught a bug in the test. It was failing since the querylog table doesn't exist. Now we run Snuba migrations in CI before the rust tests.

PR #4637, opened 2023-08-17
feat: Add support for SELECT DISTINCT

cmanallen:
Adds support for SELECT DISTINCT queries and adds new `DISTINCT` SnQL keyword.

ref: https://github.com/getsentry/snuba-sdk/issues/116

### DIFF
--- a/snuba/query/expressions.py
+++ b/snuba/query/expressions.py
@@ -17,6 +17,7 @@ TVisited = TypeVar("TVisited")
 # Sometimes however, we want the raw data, and this allows us to print that out
 _AUTO_REPR = not settings.PRETTY_FORMAT_EXPRESSIONS
 
+
 # This is a workaround for a mypy bug, found here: https://github.com/python/mypy/issues/5374
 @dataclass(frozen=True, repr=_AUTO_REPR)
 class _Expression:
--- a/snuba/clickhouse/query.py
+++ b/snuba/clickhouse/query.py
@@ -2,7 +2,7 @@ from typing import Callable, Iterable, Optional, Sequence
 
 from snuba.query import LimitBy, OrderBy
 from snuba.query import ProcessableQuery as AbstractQuery
-from snuba.query import SelectedExpression
+from snuba.query import SelectionType
 from snuba.query.data_source.simple import Table
 from snuba.query.expressions import Expression as SnubaExpression
 from snuba.query.expressions import ExpressionVisitor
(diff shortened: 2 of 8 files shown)


snuba/query/expressions.py
```diff
@@ -17,6 +17,7 @@
 # Sometimes however, we want the raw data, and this allows us to print that out
 _AUTO_REPR = not settings.PRETTY_FORMAT_EXPRESSIONS
 
+
```
evanh, line 20: ?

snuba/query/types.py
evanh, line 1: Is this needed?

evanh (changes requested):
The general approach of wrapping the select expressions in a class makes sense to me. However I wouldn't use two different classes (`Select` and `SelectDistinct`). I would suggest having one class with a `distinct` flag. That simplifies the typing quite a bit.

Another way to do this that might make some of the other steps easier would be to add a `distinct` flag to the query itself. Since the DISTINCT keyword applies to the entire SELECT expression, I don't think it's wrong to keep that flag at the query level.

Regardless this is missing a lot of logic.

While this looks good for the Logical and Composite query classes, there is another query class that needs to be modified: the Clickhouse query class found [here](https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/query.py#L15).

We translate the logical queries into a Clickhouse query (which is also where we go from Entities to Storages). That class will also need to support the new `Select` class. We translate from one to the other using [this logic](https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/translators/snuba/mapping.py#L77).

Next, those Clickhouse queries get translated into a SQL string, and you will need to add logic [there](https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/formatter/query.py#L31) as well in order to get the SQL you are hoping for.

I would also like to maintain a consistent way of working with queries. The way you changed `set_ast_selected_columns` allows either using the new `Select` class or not, which means we have an inconsistent way of building queries. I think the function should accept a `Select` object or a list of `SelectedExpression` but not both.

This is also missing a _lot_ of testing. We use `SelectedExpression` liberally in our tests, so you would need to update all of those to use the new class. And you need to thoroughly test the pipeline and changes above, which means adding tests for each of the steps above, plus some E2E tests in the `test_snql_api.py` that ensures data is being returned correctly.

This is a great start but there is definitely more work here to do.

PR #4641, opened 2023-08-17
Update GoCD jsonnet

mattgauntseo-sentry:
- Update the GoCD jsonnet library (this allows excluding a region from pipedream should it be needed during an incident)
- GoCD is able to render the jsonnet files directly, so remove the rendered yaml
- Remove the old snuba pipeline since pipedream has been in use for a while now


IanWoodard (approved):

gocd/templates/pipelines/snuba.libsonnet
```diff
@@ -59,7 +59,6 @@ function(region) {
     GITHUB_TOKEN: '{{SECRET:[devinfra-github][token]}}',
     GOCD_ACCESS_TOKEN: '{{SECRET:[devinfra][gocd_access_token]}}',
   },
-  group: 'snuba-next',
```
mattgauntseo-sentry, line 62: This isn't used.

gocd/pipelines/snuba-stable.yaml
```diff
@@ -13,6 +13,7 @@ pipelines:
             # Required for checkruns.
             GITHUB_TOKEN: "{{SECRET:[devinfra-github][token]}}"
         group: snuba
+        display_order: 100 # Ensure it's last pipeline in UI
```
mattgauntseo-sentry, line 16: This pipeline hasn't been run before, so it's safe to move to the end.

gocd/pipelines/snuba-sns-test.yaml
```diff
@@ -17,6 +17,7 @@ pipelines:
             GKE_BASTION_ZONE: a
             MIGRATIONS_READINESS: complete
         group: snuba
+        display_order: 100 # Ensure it's last pipeline in UI
```
mattgauntseo-sentry, line 20: The last run of this was in April, so moving to the end should be ok.

.github/workflows/validate-pipelines.yml
```diff
@@ -26,21 +26,6 @@ jobs:
                 gocd:
                   - 'gocd/**'
 
-    render:
```
mattgauntseo-sentry, line 29: The validate step renders the YAML and checks the render can be accepted by GoCD.

This action was only needed to check that the committed pipelines matched the rendered pipelines, which is no longer needed since we don't need to commit the pipelines.

evanh (approved):

PR #4649, opened 2023-08-28
feat(replays): Add data-skipping index to urls column

cmanallen:
These are the most heavily queried columns in the replays schema.


cmanallen:
Whoops misclick.

snuba/snuba_migrations/replays/0014_urls_index.py
```diff
@@ -0,0 +1,116 @@
...
+            f"_{column}_hashed",
+            operations.OperationTarget.DISTRIBUTED,
+        )
+
+        yield operations.DropIndex(
+            StorageSetKey.REPLAYS,
+            "replays_local",
+            f"bf_{column}_hashed",
+            target=operations.OperationTarget.LOCAL,
+        )
```
nikhars, line 113: I believe you would need to drop the bloom filter index first before dropping the column the index refers to. You can try running the reverse migration locally to verify.

cmanallen: Fixed :D

cmanallen:
Dropped the materialization.  There was no need to.

snuba/snuba_migrations/replays/0015_index_frequently_accessed_columns.py
```diff
@@ -0,0 +1,40 @@
...
+    for column in materialized_columns:
+        yield operations.DropIndex(
+            StorageSetKey.REPLAYS,
+            "replays_local",
+            f"bf_{column}",
+            target=operations.OperationTarget.LOCAL,
+        )
+
+
+materialized_columns = ["urls", "user_id", "user_name", "user_email", "ip_address_v4"]
```
dbanda, line 40: It seems excessive to have indexes on all these columns. For example, I'd imaging user_name and user_email are tied to the same user_id. Do we need indexes on user_name and user_email? Adding indexes incurs a cost on ingest and queries that don't use them.

cmanallen: I could remove urls and user_name.  I think urls wouldn't benefit too much and user_name is not regularly searched.  user id/ip/email are three of our most poplar search fields so they should stay.

nikhars (approved):

PR #4656, opened 2023-08-29
chore(self-hosted): Dogfood self-hosted e2e test action

hubertdeng123:


lynnagara (approved):

hubertdeng123:
Decided that we would remove these e2e tests in favor of running them in self-hosted on a more regular cadence, so closing

PR #4657, opened 2023-08-29
build: Cargo workspace

lynnagara:
So that rust_snuba and rust_arroyo use the same (pinned) library versions of packages

### DIFF
--- a/rust_snuba/Cargo.toml
+++ b/rust_snuba/Cargo.toml
@@ -5,6 +5,11 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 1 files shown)


volokluev (approved):

untitaker (approved):
i think normally one would move rust_snuba into a separate crate that is also just a member of the workspace. but if it works it works!

lynnagara:
> i think normally one would move rust_snuba into a separate crate that is also just a member of the workspace. but if it works it works!

https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package seems to suggest that this configuration is also ok, with rust_snuba being the root package and only rust_arroyo is the workspace member.

PR #4663, opened 2023-08-30
fix(cache): Cache fail open logic missed edge case

rahul-kumar-saini:
### Overview
- If a previously failed query has its error stored as cache, the fail open logic treated it as a Redis error
- The stored error was received from cache, raised by cache, but since it was not a `ValueGenerationError` raised within the cache logic it was treated as a cache error as per the "else" logic
- Fixed this edge case and added more data to the fail-open log line

### Real changes
- This PR is exactly the same as #4617 with one exception, there is an additional check for `SerializableException` raised (which should be bubbled up instead of retrying the query)

`backend.py`
```
        except ValueGenerationError as e:
            # Not a Redis Cache Error -> bubble up
            assert e.__cause__ is not None
            raise e.__cause__
        except SerializableException: # <------------------ This is the new check
            raise
        ...
```


rahul-kumar-saini:
Reverting because it's causing rate limit exceeded [again](https://sentry.sentry.io/issues/4403056293/?project=300688&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=2d&stream_index=0)

PR #4664, opened 2023-08-31
ref: Message processors don't take args

lynnagara:
Message processors shouldn't need to take args. Let's simplify to avoid having to port this logic to Rust.


untitaker (commented):
if that is the case then `from_kwargs` can be removed as a concept in DatasetMessageProcessor...

untitaker (changes requested):
there is one left in the rust python processor. i'd do a grep through the codebase for from_kwargs... though i suppose there are a lot of false-positives

untitaker:
preemptively approving bc all that's left to do is fix tests

PR #4665, opened 2023-08-31
fix(rust_consumer): Fix wrong default for --skip-write

untitaker:
A recent PR changed the default, and until this is fixed we may be
writing additional rows into querylog.


lynnagara (approved):

untitaker:
i'm going to force-merge this since any deploy right now would corrupt our querylog dataset, and it appears CI is going to be flaky on this PR

PR #4668, opened 2023-08-31
feat(capman): Add concurrent rate limit policy to all storages

volokluev:
This policy is already running on transactions, this PR:

* Adds a concurrent limit policy to every other storage we care about
* Does not enforce it (we will turn on enforcement on a one-by-one basis)
* Paves the way to remove all the existing rate limit processors we have in production


rahul-kumar-saini (approved):
🚀 🚀 🚀

PR #4669, opened 2023-08-31
feat(replays): migration for event link columns

JoshFerge:
As part of https://github.com/getsentry/team-replay/issues/145, we're adding new columns that can represent replay / event links to verfied-to-exist Sentry events. This PR should create 5 new columns.

Will be used in an upcoming processor PR.

### DIFF
--- a/snuba/snuba_migrations/replays/0014_add_id_event_columns.py
+++ b/snuba/snuba_migrations/replays/0014_add_id_event_columns.py
@@ -1,0 +1,43 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import UUID, Column
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.operations import OperationTarget
+
+LOG_LEVELS = ["fatal", "error", "warning", "info", "debug"]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.REPLAYS,
+                table_name=table_name,
+                column=Column(log_level + "_id", UUID()),
+                after="replay_id",
+                target=target,
+            )
+            for log_level in LOG_LEVELS
+            for table_name, target in [
+                ("replays_local", OperationTarget.LOCAL),
+                ("replays_dist", OperationTarget.DISTRIBUTED),
+            ]
+        ]
+
+    def backwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/replays/0014_add_id_event_columns.py
```diff
@@ -0,0 +1,43 @@
...
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.REPLAYS,
+                table_name=table_name,
+                column=Column(log_level + "_id", UUID()),
```
evanh, line 19: Are these guaranteed to always exist? Shouldn't this be a Nullable column?

JoshFerge: I believe that we want to get a way from making nullable columns b.c. of performance -- cc @cmanallen. https://stackoverflow.com/questions/63057886/clickhouse-string-field-disk-usage-null-vs-empty

cmanallen: Yeah I think its fine to leave as not null.  UUIDs support the `empty` and `notEmpty` functions.

snuba/snuba_migrations/replays/0014_add_id_event_columns.py
```diff
@@ -0,0 +1,43 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import UUID, Column
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.operations import OperationTarget
+
+LOG_LEVELS = ["fatal", "error", "warning", "info", "debug"]
```
cmanallen, line 8: An alternative idea for this schema.  Instead of creating multiple columns for each log level we should instead add a single `event_id` column and a `event_log_level` column?  `event_log_level` encoded as low-cardinality.

When you query for an `event_id` you target a singular column without needing to find out which log-level it is first.  When you count different log-levels you target the `event_log_level` column with a `countIf` function.  If in the future you want to support more log levels it does not require a migration.

cmanallen (changes requested):

snuba/snuba_migrations/replays/0014_add_id_event_columns.py
JoshFerge: i don't think we'd ever want to support more log levels. we've had the same 5 hard coded in sentry for forever. https://develop.sentry.dev/sdk/event-payloads/types/

cmanallen (approved):

evanh (approved):

PR #4670, opened 2023-08-31
feat(replays): add processor for replay event link type

JoshFerge:
Following up to https://github.com/getsentry/snuba/pull/4669, as part of https://github.com/getsentry/team-replay/issues/145, adds a processor for the replay event link new message type.

This message will come in from this PR https://github.com/getsentry/sentry/pull/54986 (still in draft).

### DIFF
--- a/snuba/datasets/processors/replays_processor.py
+++ b/snuba/datasets/processors/replays_processor.py
@@ -36,6 +36,7 @@ LIST_ELEMENT_LIMIT = 1000
 MAX_CLICK_EVENTS = 20
 
 USER_FIELDS_PRECEDENCE = ("user_id", "username", "email", "ip_address")
+LOG_LEVELS = ["fatal", "error", "warning", "info", "debug"]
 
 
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/replays_processor.py
```diff
@@ -278,6 +282,37 @@ def process_replay_actions(
...
+) -> dict[str, Any]:
+    event_id = None
+    event_severity = None
+    for level in LOG_LEVELS:
+        if payload.get(level + "_id") is not None:
+            event_id = to_uuid(payload[level + "_id"])
+            event_severity = level
+            break
+
+    if event_id is None or event_severity is None:
```
evanh, line 298: How confident are you that this clause won't ever be hit? Even one bad event as you test this feature will shut down your consumer.

evanh: Right, fair enough.

evanh:
Looks good to me, pending the migration and CI passing.

evanh (approved):

PR #4673, opened 2023-09-01
feat(rust): Multiprocessing with Python message processors, second attempt

untitaker:
reapply https://github.com/getsentry/snuba/pull/4655, the only notable change here is that with processes=1, there is no process being spawned

### DIFF
--- a/rust_snuba/src/strategies/python.rs
+++ b/rust_snuba/src/strategies/python.rs
@@ -278,7 +278,12 @@ mod tests {
 
     #[test]
     fn test_basic() {
-        run_basic(1);
+        let handle = procspawn::spawn((), |()| {
+            run_basic(1);
+        });
+        // on macos, this test is polluting something in the global process state causing
+        // test_basic_two_processes to hang, therefore isolate it in another subprocess of its own.
+        handle.join().unwrap();
     }
 
     #[test]
--- a/rust_snuba/src/strategies/python.rs
+++ b/rust_snuba/src/strategies/python.rs
@@ -15,16 +15,22 @@ use crate::types::BytesInsertBatch;
 
 use crate::config::MessageProcessorConfig;
 
-type TaskHandle = (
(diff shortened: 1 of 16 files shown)


snuba/cli/rust_consumer.py
```diff
@@ -84,12 +84,15 @@
...
 )
 @click.option(
-    "--no-skip-write",
-    "no_skip_write",
-    is_flag=True,
-    help="Writes to ClickHouse.",
+    "--skip-write/--no-skip-write",
+    "skip_write",
+    help="Skip the write to clickhouse",
     default=False,
```
lynnagara, line 90: Does this need to get flipped too?

untitaker: ...good catch

lynnagara (approved):

rust_snuba/src/strategies/python.rs
```diff
@@ -11,133 +15,178 @@ use crate::types::BytesInsertBatch;
...
-        let py_process_message = Python::with_gil(|py| -> PyResult<Py<PyAny>> {
-            let fun: Py<PyAny> = PyModule::from_code(py, &code, "", "")?
-                .getattr("_wrapped")?
-                .into();
-            Ok(fun)
-        })?;
+        let processing_pool = if processes <= 1 {
+            Some(procspawn::Pool::builder(processes)
+                .env("RUST_SNUBA_PROCESSOR_MODULE", python_module)
+                .env("RUST_SNUBA_PROCESSOR_CLASSNAME", python_class_name)
```
lynnagara, line 52: Can we get rid of these eventually? Would be cool if the python strategy could be agnostic of Snuba

untitaker: i think if we move this out of rust_snuba there will have to be a rethinking of the entire approach

PR #4680, opened 2023-09-05
feat(COGS): Add `UsageAccumulator` and use in Generic Metrics

rahul-kumar-saini:
### Overview
- Adds `UsageAccumulator` from the [usage-accountant](https://github.com/getsentry/usage-accountant) library (not yet published so I had to copy/paste it in for now)
- Uses said UsageAccumulator class to start recording consumer throughput usage for Generic Metrics consumers
- Rolled out percentage-wise via runtime config: `gen_metrics_processor_cogs_probability` (a float [0-1])
- What this will do is start publishing throughput (in bytes) of the generic metrics consumers to a new topic `shared-resources-usage` to start sending usage data for COGS

### Notes
- Depends on a Kafka Config existing in prod settings for the new topic (open [PR](https://github.com/getsentry/ops/pull/7927))


snuba/cogs/accountant.py
```diff
@@ -0,0 +1,250 @@
+from __future__ import annotations
```
fpacifici, line 1: The library is now public. You do not need to vendor in the accumulator.
https://pypi.devinfra.sentry.io/simple/sentry-usage-accountant/index.html

snuba/datasets/processors/generic_metrics_processor.py
```diff
@@ -146,10 +156,20 @@ def process_message(
...
         return InsertBatch(
             [processed], None, sentry_received_timestamp=sentry_received_timestamp
         )
 
+    def __record_cogs(self, message: GenericMetric) -> None:
+        if get_config("enable_gen_metrics_processor_cogs", 0):
+            record_cogs(
+                resource_id=self._resource_id,
+                app_feature=message["use_case_id"],
+                amount=len(json.dumps(message).encode("utf-8")),
```
nikhars, line 169: I am not sure how much overhead serializing the message here would cost. You might want to be aware of it when you do the rollout. Maybe make the config percentage based to be safer.

snuba/cogs/accountant.py
```diff
@@ -0,0 +1,250 @@
...
+        `app_feature` identifies the product feature.
+        `amount`  is the amount of resource used.
+        `usage_type` is the unit of measure for `amount`.
+        """
+        now = time.time()
+        if (
+            self.__first_timestamp is not None
+            and now - self.__first_timestamp >= self.__granularity_sec
+        ):
+            self.flush()
```
nikhars, line 142: `flush` seems to be a blocking call according to [this](https://github.com/getsentry/snuba/pull/4680/files#diff-2cfdc12931a5cd6ce27d7c7103521551f758ae2c6995cad2dc750989a3ec0a32R176-R177). That would imply that the consumer would be blocked when doing this work. Can the accounting be done from a separate thread instead in order to not impact the throughput of the consumer?

snuba/datasets/processors/generic_metrics_processor.py
```diff
@@ -146,10 +158,20 @@ def process_message(
...
+        self.__record_cogs(message)
         return InsertBatch(
             [processed], None, sentry_received_timestamp=sentry_received_timestamp
         )
 
+    def __record_cogs(self, message: GenericMetric) -> None:
+        if random() < (get_config("gen_metrics_processor_cogs_probability") or 0):
+            record_cogs(
+                resource_id=self._resource_id,
+                app_feature=message["use_case_id"],
```
fpacifici, line 170: Please prefix the use_case_id with `genericmetrics_`.
Since the use case ids will be owned by product they are likely to eventually conflict with other app_feature labels.

snuba/datasets/processors/generic_metrics_processor.py
fpacifici: pickle is generally serializing faster than JSON.
You can do that.

snuba/cogs/accountant.py
```diff
@@ -0,0 +1,36 @@
...
+from arroyo.backends.kafka.consumer import KafkaProducer
+from usageaccountant import UsageAccumulator, UsageUnit
+
+from snuba.utils.streams.configuration_builder import build_kafka_producer_configuration
+from snuba.utils.streams.topics import Topic as StreamTopic
+
+logger = logging.getLogger("usageaccountant")
+
+
+accumulator: UsageAccumulator | None = None
```
fpacifici, line 14: I'll let the Storage team chime in on whether to use a module attribute or whether to tie the accumulator to the MessageProcessor lifecycle. 
Just remind that, as this contains a `KafkaProducer` that has a threadpool that has to be closed when the application terminates.

fpacifici: cc @lynnagara, @untitaker may know more.

snuba/cogs/accountant.py
rahul-kumar-saini: The flush call happens at most once every 60s, so the consumer would be blocked once a minute on the flush. If we still want it in it's own thread I can look into it.

snuba/cogs/accountant.py
untitaker: I would also prefer it as part of the processor's lifecycle. Doing side-effects in MessageProcessor like this will conflict with the design of Rust consumers where we mostly assume that they are side-effect free transformation functions, but i don't really know right now if there is a better way to achieve this.

rahul-kumar-saini: Another way to do this would be to spin up an independent consumer on the gen-metrics topic which has the sole job of reporting this data using the UsageAccumulator. It would mean introducing a new piece of infra which just has the new consumer and it would be another consumer to maintain, but it would mean we don't have to touch the existing logic.

I'm not sure if we're ok with doing that (it would mean introducing a new consumer per shared resource consumer when accounting cogs), but if the conflict with the rust consumer design is too great it could be something to consider.

untitaker: let's just put it in the metrics processor for now. would prefer avoiding globals. if we have a need to move the generic metrics dataset to pure rust, we can create a separate consumer

fpacifici: I would discourage adding a consumer for that. This library is supposed to be really lightweight.
Also having a separate consumer would be very fragile as you will have to keep the code in sync between the two if a pre-filter is added to the snuba main consumer that drops more events.

This is python, there is no issue in having side effects here as this is equivalent to recording metrics or logging. 

The alternative is to make a change to the Snuba consumer to add a place to record COGS like we do for recording end to end latency (that is recorded in the flush step).

untitaker: > This is python, there is no issue in having side effects here as this is equivalent to recording metrics or logging.

there is no concern with having the python logic do more, but should we at some point have a need to port the message processor to rust, we will have to port all of this logic to Rust, and we'd be rolling out our first rust producer (next to relay) at the same time.

i'm not thrilled to allow for (potentially async) I/O in message processors, but as said I think having it in the message processor is fine for now, it's just an architectural decision I think we should revisit in case we port to rust

untitaker: > The alternative is to make a change to the Snuba consumer to add a place to record COGS like we do for recording end to end latency (that is recorded in the flush step).

if this logic is supposed to be reusable across datasets then this would be preferrable

requirements.txt
```diff
@@ -39,3 +39,4 @@ Werkzeug==2.2.3
 PyYAML==6.0
 sqlparse==0.4.2
 google-api-python-client==2.88.0
+sentry-usage-accountant==0.0.9
```
fpacifici, line 42: We have 0.0.10 now that fixes a bug

snuba/cogs/accountant.py
fpacifici: One problem with the global variable approach. 
As all Arroyo producer. You need to close it before stopping the application or your application will hang till it gets a SIGTERM. 
(try to use the accumulator and flush it from a python repl then exit the repl without calling close(), it will hang there).

I don't see a call to `close()` in this approach. 
If you tie that to the lifecycle of the processor you can ensure it gets closed. 
If your consumer hang when closing you will have problems in production.

fpacifici: The issue of a SIGTERM is that the Arroyo producer keeps a threadpool around. a threadpool needs to be manually closed before leaving the application.

rahul-kumar-saini: Updated the Consumer SIGINT/SIGTERM handler with a call to flush and close the cogs recorder (if one exists). The consumer no longer hangs when killed when killed.

nikhars (approved):
LGTM

snuba/datasets/processors/generic_metrics_processor.py
```diff
@@ -146,10 +158,20 @@ def process_message(
                 message["sentry_received_timestamp"]
             )
 
+        self.__record_cogs(message)
         return InsertBatch(
             [processed], None, sentry_received_timestamp=sentry_received_timestamp
         )
 
+    def __record_cogs(self, message: GenericMetric) -> None:
+        if random() < (get_config("gen_metrics_processor_cogs_probability") or 0):
```
lynnagara, line 167: Curious, why do you sample this?

rahul-kumar-saini: This is just to have a gradual rollout, it won't be sampled in its final state.

lynnagara: You should buffer the counts in memory if you want to produce less messages. Then you don't need a gradual rollout or to produce one message per cogs item recorded in the future

PR #4681, opened 2023-09-06
feat(rust-consumer): Configure metrics for Rust consumer

lynnagara:
Add statsd and testing metric backends to rust_snuba. Remove the metric implementation and dependency on cadence from Arroyo - users of Arroyo should provide the trait implementation.

Pass the dogstatsd host and port from Python to Rust via config.

Also start logging one metric ("arroyo.consumer.join.time") to verify it works.

### DIFF
--- a/rust_snuba/src/metrics/statsd.rs
+++ b/rust_snuba/src/metrics/statsd.rs
@@ -49,19 +49,19 @@ impl ArroyoMetrics for StatsDBackend {
         if let Err(e) =
             self.send_with_tags(self.client.count_with_tags(key, value.unwrap_or(1)), tags)
         {
-            log::error!("Error sending metric: {}", e);
+            log::debug!("Error sending metric: {}", e);
         }
     }
 
     fn gauge(&self, key: &str, value: u64, tags: Option<HashMap<&str, &str>>) {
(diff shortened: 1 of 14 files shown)


rust_snuba/src/metrics/statsd.rs
```diff
@@ -0,0 +1,82 @@
...
+        builder.try_send()
+    }
+}
+
+impl ArroyoMetrics for StatsDBackend {
+    fn increment(&self, key: &str, value: Option<i64>, tags: Option<HashMap<&str, &str>>) {
+        if let Err(e) =
+            self.send_with_tags(self.client.count_with_tags(key, value.unwrap_or(1)), tags)
+        {
+            log::error!("Error sending metric: {}", e);
```
untitaker, line 52: I would not send sentry errors for individual metric failures, which is what I think this might do?

untitaker (approved):

rust_snuba/src/metrics/statsd.rs
lynnagara: It should but suspect it won't yet since we don't have a sentry logging integration. Changed to debug as discussed elsewhere.

PR #4683, opened 2023-09-06
fix(profiling): Handle null timestamps for functions

phacops:
Since we started accepted timestamps for Android profiles as an optional field, this started to crash. We have a bunch of profiles without timestamps on the topic so this is a workaround to still ingest them.

We're also fixing the pipeline upstream to fall back on the older value if the timestamp is not sent.

https://github.com/getsentry/vroom/pull/307

### DIFF
--- a/snuba/datasets/processors/functions_processor.py
+++ b/snuba/datasets/processors/functions_processor.py
@@ -24,7 +24,8 @@ class FunctionsMessageProcessor(DatasetMessageProcessor):
 
         profile_id = str(uuid.UUID(message["profile_id"]))
         now = datetime.utcnow().timestamp()
-        timestamp = datetime.utcfromtimestamp(message.get("timestamp", now) or now)
+        raw_timestamp = max(message.get("timestamp", 0), 0)
+        timestamp = datetime.utcfromtimestamp(raw_timestamp or now)
 
         if "call_trees" in message:
             functions = {}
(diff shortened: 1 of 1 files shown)


Zylphrex (approved):

volokluev (approved):

snuba/datasets/processors/functions_processor.py
```diff
@@ -23,7 +23,8 @@ def process_message(
         max_depth_reached = False
 
         profile_id = str(uuid.UUID(message["profile_id"]))
-        timestamp = datetime.utcfromtimestamp(message["timestamp"])
+        now = datetime.utcnow().timestamp()
```
Zylphrex, line 26: this should be a datetime, not converted to a timestamp

Zylphrex (approved):

snuba/datasets/processors/functions_processor.py
```diff
@@ -23,7 +23,9 @@ def process_message(
...
+        raw_timestamp = max(message.get("timestamp", 0), 0)
+        timestamp = datetime.utcfromtimestamp(raw_timestamp or now)
```
lynnagara, line 28: I don't think this every applies now, it falls back to 1st January 1970

lynnagara (changes requested):

snuba/datasets/processors/functions_processor.py
phacops: What do you mean? if raw_timestamp is 0 because we received null or a negative value, it will get the `now` value.

lynnagara: Sorry, meant to dismiss this. Please ignore.

PR #4684, opened 2023-09-06
feat(rust-consumer): Consumer can run Python or Rust message processor

lynnagara:
Supports a CLI arg --use-rust-processor. If passed and if there is a one available, the Rust processing function is used instead of the default Python one.


rust_snuba/src/consumer.rs
```diff
@@ -67,25 +75,79 @@ pub fn consumer_impl(
...
+                                        }),
+                                    }),
+                                    Err(e) => Err(e),
+                                }
+                            })
+                        }
+                    }
+
+                    let task_runner = MessageProcessor { func };
+                    Box::new(RunTaskInThreads::new(next_step, Box::new(task_runner), 100))
```
untitaker, line 140: I don't think concurrency=100 is appropriate

rust_snuba/src/consumer.rs
```diff
@@ -67,25 +75,79 @@ pub fn consumer_impl(
...
+                self.max_batch_time,
+            );
+
+            match (
+                self.use_rust_processor,
+                processors::get_processing_function(&self.processor_config.python_class_name),
+            ) {
+                (true, Some(func)) => {
+                    struct MessageProcessor {
+                        func: fn(
```
untitaker, line 98: I think you can write a `impl<T> TaskRunner for T where T: FnMut` inside of `run_task_in_threads`, so you can use the `fn` directly as a TaskRunner. Or alternatively I think it'd be best if the MessageProcessor here would be a general-purpose utility.

untitaker (changes requested):

untitaker (approved):
turns out to be too hard, please fix concurrency=100 tho

PR #4686, opened 2023-09-07
fix(rust): CommitRequest must be used

lynnagara:
We are skipping committing offsets in a fewplaces since we are not doing anything with the CommitRequest in some strategies. `ProcessingStrategy.poll()` and `ProcessingStrategy.join()` are annotated must_use to avoid this happening in future.

Added a TODO for committing offsets in the revocation callback, we are not doing anything with them currently.


untitaker (approved):
preliminary approval

untitaker (approved):

PR #4687, opened 2023-09-07
fix(monitoring): Add dataset tag to bytes scanned metric

enochtangg:
This PR is responsible for adding missing dataset tag to the new bytes scanned metric.


volokluev (commented):
I'm not sure I want to update the interface to all these policies just for this reporting, the storage is already there as a tag, what is the reason for adding dataset?

enochtangg:
@volokluev That's fair. I was comparing our old dashboard for bytes scanned and it contained dataset tag so I was looking to consolidate. Maybe the storage_key is sufficient. It just takes a little bit more mental energy to switch contexts since cumulative time dashboard is grouped by dataset.

volokluev:
which dashboard are you looking at?

enochtangg:
There are a few. Let's discuss offline so I can avoid sharing DD links.

PR #4690, opened 2023-09-08
fix(gocd): Also deploy rust-spans-consumer

untitaker:


untitaker:
@lynnagara @mattgauntseo-sentry does rust-spans also need to be added here? https://github.com/getsentry/snuba/blob/fix/gocd-rust-spans-consumer/gocd/pipelines/snuba-stable.yaml

lynnagara (approved):

lynnagara:
> @lynnagara @mattgauntseo-sentry does rust-spans also need to be added here? https://github.com/getsentry/snuba/blob/fix/gocd-rust-spans-consumer/gocd/pipelines/snuba-stable.yaml

No, that was supposed to be only for the non experimental consumers (though i don't think it's useful anyway)

PR #4694, opened 2023-09-08
ref(snql) Aliases in SELECT should be pushed to expressions

evanh:
Currently the alias in the SELECT expressions is used only for the results in
the SelectedExpression. The alias also needs to be pushed down to the
expression inside the SelectedExpression as well though, to ensure that other
references to the alias get expanded properly.


lynnagara (approved):

evanh:
Out of date

PR #4695, opened 2023-09-08
fix(rust-arroyo): Simplify AnyMessage::committable() to avoid off by one error

lynnagara:
There was a bug in the Reduce strategy - since it was passing the seen offsets, rather than the committable offsets (which should be the message offset + 1), it was always committing one behind what it should be. Hopefully this change makes it easier for users of AnyMessage and harder to commit the wrong offset now.


untitaker (approved):

untitaker:
from slack (realized after approval)

```
at some point, every message is a BrokerMessage
so there we already add +1
in reduce strategy, we merge message.committable() from many messages together
based on that we create a new kind of message
so now when we get committable() of that message, we add another +1
```

PR #4698, opened 2023-09-10
feat(profiling): Add a profiles processor in Rust

phacops:


rust_snuba/src/processors/mod.rs
```diff
@@ -9,6 +11,7 @@ type ProcessingFunction =
 pub fn get_processing_function(name: &str) -> Option<ProcessingFunction> {
     match name {
         "QuerylogProcessor" => Some(querylog::process_message),
+        "ProfilesProcessor" => Some(profiles::process_message),
```
lynnagara, line 14: This will probably change in the future but right now this name should match https://github.com/getsentry/snuba/blob/master/snuba/datasets/processors/profiles_processor.py#L15 since it swaps out the Python processor for a Rust one if it finds one matching the class name.

```suggestion
        "ProfilesMessageProcessor" => Some(profiles::process_message),
```

lynnagara:
You might find it helpful to parametrize https://github.com/getsentry/snuba/blob/0d49f46055269e59874b9669292b58c6b84b1b73/tests/consumers/test_message_processors.py#L13 and add profiles to this test.

It loads all the examples from sentry kafka schemas, runs the Python + Rust processing function and fails if the bytes don't match on any example.

tests/consumers/test_message_processors.py
```diff
@@ -10,13 +11,17 @@
 from snuba.processor import InsertBatch
 
 
-def test_message_processors() -> None:
+@pytest.mark.parametrize(
+    "topic,processor_name",
+    [
+        ("snuba-queries", "QuerylogProcessor"),
+        ("profile-metadata", "ProfilesMessageProcessor"),
```
lynnagara, line 18: should it be `processed-profiles`? Needs to match the (default) topic name otherwise the examples won't be loaded

untitaker (approved):

PR #4699, opened 2023-09-10
fix(profiling): Support only one format as we only emit and validate one format

phacops:
We used to expect 2 different formats and would ingest one or the other but since we switched to validating schemas, we only have one format (documented here https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/profile-metadata.v1.schema.json).

It is not useful to have a condition to switch to one or the other.


untitaker (approved):
(pls deploy when sfo is awake)

PR #4706, opened 2023-09-11
feat(spans): write tags to sentry_tags col

dbanda:
Dual write leftover sentry extracted tags into the sentry_tags and tags columns. Eventually, this will be switched to writing to only sentry_tags to avoid any conflicts with user tags.


snuba/datasets/processors/spans_processor.py
```diff
@@ -187,6 +187,8 @@ def _process_sentry_tags(
         leftover_tag_keys, leftover_tag_values = extract_extra_tags(sentry_tags)
         processed["tags.key"].extend(leftover_tag_keys)
         processed["tags.value"].extend(leftover_tag_values)
+        processed["sentry_tags.key"].extend(leftover_tag_keys)
```
phacops, line 190: I'm confused, why don't we just write the tags contained in `sentry_tags`?

phacops (approved):

evanh (approved):

PR #4712, opened 2023-09-11
feat(spans): Port the processor to Rust

john-z-yang:


rust_snuba/src/processors/spans.rs
```diff
@@ -0,0 +1,222 @@
...
+fn enforce_retention(_: &SpanMessage) -> Option<u32> {
+    Some(90)
+}
+
+fn process_span_meta(event: &SpanMessage, processed: &mut BTreeMap<&str, Value>) {
+    processed.insert("trace_id", json!(event.trace_id));
+    processed.insert(
+        "span_id",
+        json!(u128::from_str_radix(event.span_id.as_str(), 16).unwrap()),
+    );
```
john-z-yang, line 103: Maybe these need to be validated during deserialization? I think unwrap would panic when it's None, right?

untitaker: you can try but you can also just return InvalidMessage like you already do upon failed deserialization

rust_snuba/src/processors/spans.rs
```diff
@@ -0,0 +1,222 @@
...
+            let s = err.to_string();
+            log::error!("Failed to deserialize message: {}, {}", err, s);
+            InvalidMessage
+        })?;
+        let retention_days = enforce_retention(&event).ok_or_else(|| {
+            log::error!("Failed to process message: event too old");
+            InvalidMessage
+        })?;
+
+        let mut processed: BTreeMap<&str, Value> = BTreeMap::from([
```
untitaker, line 142: instead of btreemap I would create a new struct, similar to how querylog has FromQuerylog for input data and a Querylog for output data

untitaker (changes requested):

rust_snuba/src/processors/spans.rs
john-z-yang: I see, I'm assuming you mean something like
``` rust

#[derive(Debug, Deserialize)]
struct Span {
...
}

impl TryFrom<SpanMessage> for Span {
    fn try_from(from: SpanMessage) -> Result<QueryList, InvalidMessage> {
      ...
    }
}
```

untitaker: yup!

snuba/datasets/processors/spans_processor.py
```diff
@@ -100,12 +100,14 @@ def _process_span_event(
...
         # timestamps
-        processed["start_timestamp"], processed["start_ms"] = self.__extract_timestamp(
-            span_event["start_timestamp_ms"] / 1000,
-        )
-        processed["end_timestamp"], processed["end_ms"] = self.__extract_timestamp(
-            (span_event["start_timestamp_ms"] + span_event["duration_ms"]) / 1000,
-        )
+        start_timestamp_sec = span_event["start_timestamp_ms"] / 1000
+        processed["start_timestamp"] = int(start_timestamp_sec)
+        _, processed["start_ms"] = self.__extract_timestamp(start_timestamp_sec)
```
lynnagara, line 105: Mind modifying `__extract_timetamp()` to return int instead of datetime if that's what we need now? (these are the only 2 places it's used)

snuba/datasets/processors/spans_processor.py
```diff
@@ -46,14 +46,15 @@ class SpansMessageProcessor(DatasetMessageProcessor):
...
+    def __extract_timestamp(self, timestamp_ms: int) -> Tuple[int, int]:
         # We are purposely using a naive datetime here to work with the rest of the codebase.
         # We can be confident that clients are only sending UTC dates.
-        timestamp = _ensure_valid_date(datetime.utcfromtimestamp(timestamp_sec))
+        timestamp = _ensure_valid_date(datetime.utcfromtimestamp(timestamp_ms / 1000))
         if timestamp is None:
             timestamp = datetime.utcnow()
         milliseconds = int(timestamp.microsecond / 1000)
-        return timestamp, milliseconds
+        unix_timestamp = int(timestamp.replace(tzinfo=timezone.utc).timestamp())
```
phacops, line 56: I have to replace the timezone as our dev environments are not on UTC time usually but it's not something a production system need.

https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

rust_snuba/src/processors/spans.rs
```diff
@@ -0,0 +1,406 @@
...
+            .map(|v| match v {
+                Value::String(v) => Value::String(v),
+                Value::Bool(b) => {
+                    let mut v: Vec<char> = b.to_string().as_str().chars().collect();
+                    v[0] = v[0].to_uppercase().nth(0).unwrap();
+                    Value::String(v.into_iter().collect())
+                }
+                _ => Value::String(v.to_string()),
+            })
+            .collect();
```
phacops, line 187: This is horrible but it's needed to match the exact Python output as it's making everything a string (so a `bool` becomes `True` where it's would be `true` if transforming the JSON equivalent to a string).

Ideas to improving this situation are welcome.

rust_snuba/src/processors/spans.rs
```diff
@@ -0,0 +1,406 @@
...
+    D: Deserializer<'de>,
+{
+    let hex = String::deserialize(deserializer)?;
+    u64::from_str_radix(&hex, 16).map_err(serde::de::Error::custom)
+}
+
+#[derive(Clone, Copy, Default, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+#[repr(u8)] // size limit in clickhouse
+pub enum SpanStatus {
```
phacops, line 248: I want to get that from `relay` directly but I don't think we can import the crate just yet.

@jan-auer Any ideas on how we could share this?

snuba/datasets/processors/spans_processor.py
lynnagara: I'd skip the conversion from int to datetime and back again, then you won't have to deal with timezone at all.

Something like:
```
is_valid = _ensure_valid_date(datetime.utcfromtimestamp(timestamp_ms / 1000)) is not None
timestamp_ms = timestamp_ms if is_valid else int(time.time() * 1000)
return int(timestamp_ms / 1000), int(timestamp_ms % 1000)
```

rust_snuba/src/processors/spans.rs
lynnagara: It's easier for my brain as:
```
match b {
  True => "True",
  False => "False"
}
```

rust_snuba/src/processors/spans.rs
lynnagara: I think it's fine to leave it as is, not a huge fan of adding a relay dependency here

rust_snuba/src/processors/spans.rs
lynnagara: Aside, should this conversion be happening in Snuba at all? Tag values are only supposed to be strings, shouldn't we be checking that somewhere upstream of here? In datasets where data is written to nodestore or other storages, it should be consistent everywhere.

rust_snuba/src/processors/spans.rs
```diff
@@ -0,0 +1,406 @@
...
+            self.transaction_op.clone().into(),
+        ]
+    }
+}
+
+#[derive(Debug, Default, Serialize)]
+struct Span {
+    deleted: u8,
+    description: String,
+    duration: u64,
```
lynnagara, line 127: Please review all of the types here. I just double checked this one and it does not match the ClickHouse storage definition which says this is a u32. I did not review at the rest of the list.

rust_snuba/src/processors/spans.rs
phacops: Good point. https://github.com/getsentry/sentry/pull/56362

untitaker (approved):

lynnagara (approved):

PR #4722, opened 2023-09-14
fix(on-call): Negative org ID in Tenant IDs crashes Querylog consumer

rahul-kumar-saini:
Querylog consumer is crashlooping due to negative org ID's in `tenant_ids`, (Clickhouse insert rejects)

### DIFF
--- a/snuba/querylog/query_metadata.py
+++ b/snuba/querylog/query_metadata.py
@@ -258,7 +258,7 @@ class SnubaQueryMetadata:
         org_id = self.request.attribution_info.tenant_ids.get("organization_id")
         if org_id is not None and isinstance(org_id, int):
             if org_id <= 0:
-                logger.warning(f"Invalid Org ID in tenant_ids: {org_id}")
+                logger.warning(f"Invalid Org ID in tenant_ids: {org_id}", request_dict)
                 org_id = 0
             request_dict["organization"] = org_id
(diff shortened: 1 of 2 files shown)


snuba/querylog/query_metadata.py
```diff
@@ -254,7 +254,7 @@ def to_dict(self) -> snuba_queries_v1.Querylog:
         # TODO: Remove check once Org IDs are required
         if org_id := self.request.attribution_info.tenant_ids.get("organization_id"):
             if isinstance(org_id, int):
-                request_dict["organization"] = org_id
+                request_dict["organization"] = max(org_id, 0)
```
onewland, line 257: if we aren't sending to the DLQ, maybe we should emit a metric or a sentry warning?

onewland (approved):
it would be best to put this on the DLQ, but fine with this change so that we can keep the data flowing

snuba/querylog/query_metadata.py
lynnagara: If you did want to put it in DLQ instead of writing to clickhouse and treating it as valid, you can just raise an exception and it should automatically happen

onewland: we figured out that this unfortunately is done on the produce side, so it's not a good place to raise that exception

PR #4726, opened 2023-09-14
feat(spans): add migration to add compression to spans

dbanda:
Add ZSTD compression to most spans columns.


nikhars (commented):
Are these codec changes already applied in US SaaS or they need to be applied via the migration framework? I remember you applied some compression enhancements earlier. Wanted to know if these are new ones or migrations which are missing from prod.

dbanda:
> Are these codec changes already applied in US SaaS or they need to be applied via the migration framework? I remember you applied some compression enhancements earlier. Wanted to know if these are new ones or migrations which are missing from prod.

Yes these are already applied in SaaS.

nikhars (approved):

PR #4728, opened 2023-09-15
feat(capman): Cross-organization allocation policy

volokluev:
# Introduction

Before this PR all allocation policies have used a customer id (either `project_id` or `organization_id`) as the tenants for allocation policies. This has worked because most queries that have come in to snuba have been made on behalf of a specific customer. Recently however, new features are being developed where one query can issue scan data across organizations and are done outside of the customer’s control (e.g. escalating issues), as such, capacity to these queries will need to be managed differently.

# Blast Radius of this PR
This PR will roll out to production **inactive**. This means none of its code will run until it is turned on by a human being. The only storage this PR currently affects is `generic_metrics_counters`

# Goals

- Cross-org queries should not be able to monopolize the entire capacity of the cluster
- It should be easy to adjust the capacity allocation for a cross org referrer
- Simplicity of the policy
    - These kinds of queries are pretty new, we don’t have a good way of thinking about them yet, we just want some basic safeguards in place
    
    
 # Design

The `CrossOrgQueryPolicy` is implemented as a separate policy because its concerns are orthogonal to that of the customer based policies. Unlike the customer based policies, the amount of possible tenants is very small but each one has its own specific consideration. Therefore **the limits will be tuned on a case-by-case basis for the first iteration.**

## Interface

Every cross-org query referrer will have to be added manually to the configuration of that specific allocation policy. The config looks like:

```yaml
allocation_policies:
  - CrossOrgQueryPolicy
  cross_org_referrer_limits:
    statistical_detectors.get_top_transactions:
      max_threads: 4
      concurrent_limit: 2
    escalating_issues.scan_issues:
      max_threads: 10
      concurrent_limit: 1
```

The client will add the following field to the `tenant_ids` of the snuba request:

```yaml
cross_org_query: 1
```

This is done so that the other allocation policies which are for meant for customers will not try to enforce themselves on this query.

## Knobs and Levers

Both concurrent limit and max threads for each referrer can be overridden at runtime without deploy. Any referrers not defined in the config in source code will not be affected by the policy. 

Any query with `cross_org_query=1` in the `tenant_ids` dictionary will be rejected by the `CrossOrgQueryPolicy` . Since `cross_org_query=1` bypasses the customer specific rate limiters, it’s important to make sure they do not bypass this one

### DIFF
--- a/snuba/datasets/configuration/generic_metrics/storages/counters.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/counters.yaml
@@ -83,6 +83,7 @@ allocation_policies:
         - referrer
       default_config_overrides:
         is_enforced: 0
+        is_active: 0
       cross_org_referrer_limits:
         dynamic_sampling.counters.get_org_transaction_volumes:
           max_threads: 4
--- a/snuba/datasets/configuration/generic_metrics/storages/counters.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/counters.yaml
@@ -77,7 +77,16 @@ allocation_policies:
         is_enforced: 1
         throttled_thread_number: 1
         org_limit_bytes_scanned: 100000
-
+  - name: CrossOrgQueryAllocationPolicy
+    args:
+      required_tenant_types:
+        - referrer
+      default_config_overrides:
+        is_enforced: 0
+      cross_org_referrer_limits:
+        dynamic_sampling.counters.get_org_transaction_volumes:
+          max_threads: 4
+          concurrent_limit: 10
 query_processors:
   - processor: TableRateLimit
   - processor: TupleUnaliaser
--- a/snuba/datasets/configuration/generic_metrics/storages/counters.yaml
+++ b/snuba/datasets/configuration/generic_metrics/storages/counters.yaml
@@ -87,6 +87,16 @@ allocation_policies:
         dynamic_sampling.counters.get_org_transaction_volumes:
           max_threads: 4
           concurrent_limit: 10
+        dynamic_sampling.counters.fetch_projects_with_count_per_transaction_volumes:
+            max_threads: 4
+            concurrent_limit: 10
+        dynamic_sampling.counters.fetch_projects_with_transaction_totals:
+            max_threads: 4
+            concurrent_limit: 10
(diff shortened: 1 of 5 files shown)


evanh:
> Any query with cross_org_query=1 in the

cross_org_query=0?

snuba/datasets/configuration/generic_metrics/storages/counters.yaml
```diff
@@ -77,6 +77,26 @@ allocation_policies:
...
         org_limit_bytes_scanned: 100000
+  - name: CrossOrgQueryAllocationPolicy
+    args:
+      required_tenant_types:
+        - referrer
+      default_config_overrides:
+        is_enforced: 0
+        is_active: 0
+      cross_org_referrer_limits:
+        dynamic_sampling.counters.get_org_transaction_volumes:
```
evanh, line 88: I understand wanting to use these fields as specific overrides, but why not use the `cross_org_query` flag to determine if the referrer should be included or not? 

In other words, if a referrer comes in with `cross_org_query=1`, automatically use the cross org policy with `max_threads=4` and `concurrent_limit=10`?

snuba/query/allocation_policies/cross_org.py
```diff
@@ -0,0 +1,172 @@
...
+        self._registered_cross_org_referrers = cast(
+            "dict[str, dict[str, int]]", kwargs.get("cross_org_referrer_limits", {})
+        )
+        self._validate_cross_org_referrer_limits(self._registered_cross_org_referrers)
+
+    def _get_quota_allowance(
+        self, tenant_ids: dict[str, str | int], query_id: str
+    ) -> QuotaAllowance:
+        referrer = str(tenant_ids.get("referrer", "no_referrer"))
+        referrer_is_registered = referrer in self._registered_cross_org_referrers
```
evanh, line 112: I mentioned this in a comment above, but couldn't the referrer be automatically included if `cross_org_query=1`? Why do we need that flag and this check?

volokluev:
@evanh I thought about it more and I think you're right about this being too restrictive. added a default of 1 max thread and 1 concurrent query for cross org queries by default.

evanh (approved):
Pending the CI passing, looks good.

PR #4729, opened 2023-09-15
feat(deploy-gates): Add SaaS -> ST deploy gates

rahul-kumar-saini:
### Overview
- Adds 2 Sentry checks
    - One checks latest Release for any new issues
    - One checks latest Release for total error events > 500
- Adds a Datadog Monitor check
    - Simple check of two existing monitors
    - Currently this is set to dry-run, which means it won't block since I don't believe these monitors are appropriate for this check
    - We need to add some "current" status checks, all existing monitors check for ~5 minutes of data which could block a deploy fixing said monitor to begin with

### DIFF
--- a/gocd/templates/bash/saas-ddog-health-check.sh
+++ b/gocd/templates/bash/saas-ddog-health-check.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+/devinfra/scripts/checks/datadog/monitor_status.py --dry-run=true \
+  113296727 \
+  42722121
+
+
+# Above monitor IDs map to following monitors respectively:
+# Snuba - SLO - High API error rate
+# Snuba - Too many restarts on Snuba pods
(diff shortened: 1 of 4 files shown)


gocd/templates/bash/saas-ddog-health-check.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+/devinfra/scripts/checks/datadog/monitor_status.py --dry-run=true \
+  113296727 \
+  42722121
```
rahul-kumar-saini, line 5: This can be merged as it's a dry-run and won't block but once we've figured out what Datadog monitors to actually use, we should swap these values out and disable dry-run

mattgauntseo-sentry (approved):

evanh (approved):
We talked about this offline, but we can't enable DD check as things stand. The rollback mechanism uses the same deploy flow as the normal deploy, and the DD alerts aren't scoped to a release. For example:

1) We deploy a bad change
2) The monitor goes red
3) We rollback the change
4) The monitor is still red, so the rollback fails
5) We are stuck

We need to figure this out before any DD alerts can be enabled.

PR #4734, opened 2023-09-18
feat: Experiment using statsdproxy for aggregation

lynnagara:


untitaker:
I pushed some fixes to statsdproxy master, but I couldn't figure out any problem that is actually reproducible locally.

@lynnagara I remember vaguely that metrics didn't arrive on your machine, and I am seeing some issues locally as well where sometimes metrics are missing with `socat`, but those issues are also present with cadence's own metrics sink. i think i just want to merge this, after we have merged some metrics into master, and see if they appear in datadog.

lynnagara:
Seems fine to me, just remember to revert the settings file changes

Swatinem (approved):

untitaker:
for the record, I want to do this because I want to get statsdproxy in a prod-ready state, and because I want to avoid manual metrics buffering like we have to do it in python-arroyo all over the place.

the networking code in statsdproxy is now identical to cadence, so I think it's now unlikely we'd miss/lose any data because of this experimental sink. but if we do, let's just revert and not think about it again. it shouldn't block our milestones.

lynnagara:
i'm going to revert since this appears to have cut throughput by approx 6x

PR #4736, opened 2023-09-18
feat(ci): make sure possible API changes are tested against sentry

volokluev:
If we change the flask app, query processors, or the allocation policies, that affects the API. Therefore we should have changes to these files run the sentry test suite

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     timeout-minutes: 3
     # Map a step output to a job output
     outputs:
-      datasets: ${{ steps.changes.outputs.datasets }}
+      api_changes: ${{ steps.changes.outputs.api_changes }}
     steps:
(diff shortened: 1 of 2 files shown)


lynnagara (approved):

.github/workflows/ci.yml
```diff
@@ -376,14 +376,14 @@ jobs:
...
 
       - name: Run snuba tests
-        if: needs.files-changed.outputs.datasets == 'false'
+        if: needs.files-changed.outputs.api_changes == 'false'
         working-directory: sentry
         run: |
           make test-snuba
 
       - name: Run full tests
         # TODO: Adjust which changed files we care about
```
lynnagara, line 385: You can remove this line

volokluev: because it's the default behavior?

lynnagara: It was intented as note to myself to confirm with search team which files we want to run the full tests on in addition to the *.yamls. I think this PR does that now.

PR #4740, opened 2023-09-19
feat(consumer): allow group instance id to enable static membership

dbanda:
We would like to experiment with running Kafka with static membership. This mode is enabled when consumer passing in the `group.instance.id` parameter.  This PR adds passing in the group instance id as a CLI parameter.

### DIFF
--- a/snuba/consumers/consumer_config.py
+++ b/snuba/consumers/consumer_config.py
@@ -64,6 +64,25 @@ class ConsumerConfig:
     env: Optional[EnvConfig]
 
 
+def _add_to_topic_broker_config(
+    topic_config: TopicConfig, param_key: str, param_value: str
+) -> TopicConfig:
+    """
+    Add a parameter to the broker configuration of a topic.
+    Returns a new TopicConfig with the added parameter to
+    the broker configuration.
+    """
+    assert isinstance(param_key, str)
+    # copy the broker config to avoid modifying the original
+    broker_config = {k: v for k, v in topic_config.broker_config.items()}
+    broker_config[param_key] = param_value
+    return TopicConfig(
+        broker_config=broker_config,
+        logical_topic_name=topic_config.logical_topic_name,
+        physical_topic_name=topic_config.physical_topic_name,
+    )
+
+
 def _resolve_topic_config(
     param: str,
     topic_spec: Optional[KafkaTopicSpec],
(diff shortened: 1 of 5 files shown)


rust_snuba/src/consumer.rs
```diff
@@ -215,6 +215,10 @@ pub fn consumer_impl(
         })
         .collect();
 
+    if let Some(group_instance_id) = consumer_config.group_instance_id {
+        broker_config.insert("group.instance.id".to_owned(), group_instance_id);
+    }
+
```
lynnagara, line 221: Can you have the `resolve_consumer_config()` function handle this? It's supposed to be the singular place where consumer config is resolved before passing it to the Rust or Python consumer so we can avoid needing to do stuff like this twice in Rust and Python

lynnagara (changes requested):

rust_snuba/src/consumer.rs
dbanda: I can do it on the python side, but I still need to a way to get that into `KafkaConfig::new_consumer_config`? Should I make `group_instance_id` a parameter to that function? Here I resolved to just adding it into the overrides hashmap.

rust_snuba/rust_arroyo/src/backends/kafka/config.rs
```diff
@@ -85,6 +90,10 @@ mod tests {
                 "queued.max.messages.kbytes".to_string(),
                 "1000000".to_string(),
             )])),
+            Some(HashMap::from([(
+                "max.poll.interval.ms".to_string(),
+                "1000000".to_string(),
```
lynnagara, line 95: What's this for?

dbanda: It's a unit test. Refactored it to allow passing in key value pairs to the kafka config from the resolver.

rust_snuba/rust_arroyo/src/backends/kafka/config.rs
```diff
@@ -23,6 +24,7 @@ impl KafkaConfig {
         auto_offset_reset: String,
         _strict_offset_reset: bool, // TODO: Implement this
         override_params: Option<HashMap<String, String>>,
+        consumer_config_params: Option<HashMap<String, String>>,
```
lynnagara, line 27: This argument should not be necessary - there is already a way to pass consumer consumer config via `override_params` in the argument above this

snuba/consumers/consumer_config.py
```diff
@@ -170,6 +175,7 @@ def resolve_consumer_config(
         max_batch_size=max_batch_size,
         max_batch_time_ms=max_batch_time_ms,
         env=resolved_env_config,
+        kafka_consumer_config=kafka_consumer_config,
```
lynnagara, line 178: Please don't add a new item here, Kafka consumer config should be part of the existing`raw_topic` configuration item

lynnagara (approved):
thanks!

PR #4742, opened 2023-09-20
feat(replays): add new materialized columns for event linking

JoshFerge:
as part of https://github.com/getsentry/team-replay/issues/145 -- add new materialized columns for event counts. we add three -- info, warning, and error.

### DIFF
--- a/snuba/snuba_migrations/replays/0016_materialize_new_event_counts.py
+++ b/snuba/snuba_migrations/replays/0016_materialize_new_event_counts.py
@@ -57,21 +57,31 @@ columns: List[Tuple[str, Column[Modifiers]]] = [
         "debug_id",
         Column(
             "count_info_events",
-            UInt(8, Modifiers(materialized="(debug_id > 0) + (info_id > 0)")),
+            UInt(
+                8,
+                Modifiers(
+                    materialized="(toInt128(debug_id) > 0) + (toInt128(info_id) > 0)"
+                ),
+            ),
         ),
     ),
     (
         "count_info_events",
         Column(
             "count_warning_events",
-            UInt(8, Modifiers(materialized="warning_id > 0")),
+            UInt(8, Modifiers(materialized="toInt128(warning_id) > 0")),
         ),
     ),
     (
         "count_warning_events",
         Column(
             "count_error_events",
-            UInt(8, Modifiers(materialized="(error_id > 0) + (fatal_id > 0)")),
+            UInt(
+                8,
+                Modifiers(
+                    materialized="(toInt128(error_id) > 0) + (toInt128(fatal_id) > 0)"
+                ),
+            ),
         ),
     ),
 ]
--- a/snuba/snuba_migrations/replays/0016_materialize_new_event_counts.py
+++ b/snuba/snuba_migrations/replays/0016_materialize_new_event_counts.py
@@ -60,7 +60,7 @@ columns: List[Tuple[str, Column[Modifiers]]] = [
             UInt(
                 8,
                 Modifiers(
-                    materialized="(toInt128(debug_id) > 0) + (toInt128(info_id) > 0)"
+                    materialized="(debug_id != '00000000-0000-0000-0000-000000000000') + (info_id != '00000000-0000-0000-0000-000000000000')"
                 ),
             ),
         ),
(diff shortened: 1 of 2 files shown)


cmanallen:
I was curious about the performance of computing the value at read-time versus using the cached value.  Roughly 25% faster.  This is the output of ClickHouse benchmark:

```sql
-- Raw Query

SELECT
    parent_id,
    sum(error_id > 0)
FROM mat_col_scalar
GROUP BY parent_id
LIMIT 1

Queries executed: 1054.

localhost:9000, queries 1054, QPS: 60.440, RPS: 60440298.217, MiB/s: 1383.369, result RPS: 60.440, result MiB/s: 0.001.

0.000%          0.012 sec.
10.000%         0.014 sec.
20.000%         0.015 sec.
30.000%         0.015 sec.
40.000%         0.016 sec.
50.000%         0.016 sec.
60.000%         0.017 sec.
70.000%         0.017 sec.
80.000%         0.018 sec.
90.000%         0.019 sec.
95.000%         0.020 sec.
99.000%         0.023 sec.
99.900%         0.029 sec.
99.990%         0.030 sec.

-- Materialized Query

SELECT
    parent_id,
    sum(count_error)
FROM mat_col_scalar
GROUP BY parent_id
LIMIT 1

Queries executed: 1039.

localhost:9000, queries 1039, QPS: 77.360, RPS: 77359520.195, MiB/s: 663.982, result RPS: 77.360, result MiB/s: 0.001.

0.000%          0.009 sec.
10.000%         0.011 sec.
20.000%         0.012 sec.
30.000%         0.012 sec.
40.000%         0.012 sec.
50.000%         0.013 sec.
60.000%         0.013 sec.
70.000%         0.013 sec.
80.000%         0.014 sec.
90.000%         0.015 sec.
95.000%         0.016 sec.
99.000%         0.019 sec.
99.900%         0.022 sec.
99.990%         0.023 sec.
```

cmanallen (approved):

dbanda:
Has team evaluated how wide/compact we want these table parts? Although adding columns in Clickhouse is inexpensive, eventually there can be a penalty. If we are not using compact part i.e setting `min_bytes_for_wide_part`, each column will be adding in a separate mark file. If you're selecting many columns once it could get expensive. Also, you might need a bigger mark cache.

JoshFerge:
> Has team evaluated how wide/compact we want these table parts? Although adding columns in Clickhouse is inexpensive, eventually there can be a penalty. If we are not using compact part i.e setting `min_bytes_for_wide_part`, each column will be adding in a separate mark file. If you're selecting many columns once it could get expensive. Also, you might need a bigger mark cache.

we have not evaluated this. we generally don't select many columns at once if we're scanning lots of data though. we don't believe we have performance issues w/ this at this time.

snuba/snuba_migrations/replays/0016_materialize_new_event_counts.py
```diff
@@ -0,0 +1,77 @@
...
+            operations.OperationTarget.LOCAL,
+        )
+
+
+columns: List[Tuple[str, Column[Modifiers]]] = [
+    (
+        "debug_id",
+        Column(
+            "count_info_events",
+            UInt(8, Modifiers(materialized="(debug_id > 0) + (info_id > 0)")),
```
dbanda, line 60: This raises and error `DB::Exception: Illegal types of arguments (UUID, UInt8) of function greater: While processing CAST((debug_id > 0) + (info_id > 0)`

Are you trying to check if the ID is not null or explicitly greater than 0?

JoshFerge:
@dbanda does that comment need to block the merge here?

JoshFerge:
@cmanallen @dbanda it seems like all versions of clickhouse are satisfied with this version of the migration.

cmanallen:
@dbanda Is the concern write performance?  My understanding is on the read side a compact part does not give us anything.  I can try to benchmark the write impact of these new columns to get some idea of overhead.

@JoshFerge Do we have any metrics on the total number of error_ids being inserted?  What I'd really like to know is what percentage of rows in this table are dedicated to error_ids (extrapolated over 90 days)?

---

As this table becomes more sparse we may want to add sparse serialization to the table.  If you remove the duplicate data we store, I think we're already in a place where sparse serialization could make sense.

dbanda:
@cmanallen the concern was with memory. More columns on wide parts mean more mark files. If say you are doing an aggregation, you'll have more stuff to put in memory. The mark cache is shared across all tables. So as you add columns you could hit a point where you get performance degradation. Just wanted to raise that point. Essentially, the cost of materialized columns is memory and that can build up overtime.

dbanda (approved):

cmanallen (approved):

PR #4743, opened 2023-09-21
ref(rust): Set up logging integration correctly

untitaker:
Switch to pretty-env-logger, since that's what we have docs for: https://docs.rs/sentry/0.31.7/sentry/integrations/log/index.html


lynnagara:
Does it matter that arroyo uses env_logger?

untitaker:
> Does it matter that arroyo uses env_logger?

It seems to me that arroyo only uses env_logger in examples and tests. the library code should not use env_logger at all

lynnagara (approved):

PR #4747, opened 2023-09-21
feat(generic-metrics): Add support for 10s granularity

ayirr7:
`CUSTOM` use case needs 10 second granularity support. Need to double check if this is the only metric type they need it for.

This would go in after the corresponding Sentry changes: https://github.com/getsentry/sentry/pull/56707

UPDATE: confirmed that `CUSTOM` use case wants it for all metric types

### DIFF
--- a/snuba/datasets/metrics_messages.py
+++ b/snuba/datasets/metrics_messages.py
@@ -17,9 +17,6 @@ class OutputType(Enum):
 class AggregationOption(Enum):
     HIST = "hist"
     TEN_SECOND = "ten_second"
-    ONE_MINUTE = "one_minute"
-    ONE_HOUR = "one_hour"
-    ONE_DAY = "one_day"
 
 
 ILLEGAL_VALUE_IN_SET = "Illegal value in set."
--- a/snuba/datasets/metrics_messages.py
+++ b/snuba/datasets/metrics_messages.py
@@ -16,6 +16,7 @@ class OutputType(Enum):
 
 class AggregationOption(Enum):
     HIST = "hist"
+    TEN_SECOND = "ten_second"
 
 
 ILLEGAL_VALUE_IN_SET = "Illegal value in set."
@@ -25,6 +26,7 @@ ILLEGAL_VALUE_IN_COUNTER = "Illegal value in counter."
 INT_FLOAT_EXPECTED = "Int or Float expected"
 
 # These are the hardcoded values from the materialized view
+GRANULARITY_TEN_SECONDS = 0
 GRANULARITY_ONE_MINUTE = 1
 GRANULARITY_ONE_HOUR = 2
 GRANULARITY_ONE_DAY = 3
@@ -91,11 +93,6 @@ def aggregation_options_for_distribution_message(
     aggregation_options = {
         "min_retention_days": retention_days,
         "materialization_version": 2,
-        "granularities": [
-            GRANULARITY_ONE_MINUTE,
-            GRANULARITY_ONE_HOUR,
-            GRANULARITY_ONE_DAY,
-        ],
     }
 
     if aggregation_setting := message.get("aggregation_option"):
@@ -103,7 +100,22 @@ def aggregation_options_for_distribution_message(
         if parsed_aggregation_setting is AggregationOption.HIST:
             return {
                 **aggregation_options,
+                "granularities": [
+                    GRANULARITY_ONE_MINUTE,
+                    GRANULARITY_ONE_HOUR,
+                    GRANULARITY_ONE_DAY,
+                ],
                 "enable_histogram": 1,
+            }
+        if parsed_aggregation_setting is AggregationOption.TEN_SECOND:
+            return {
+                **aggregation_options,
+                "granularities": [
+                    GRANULARITY_ONE_MINUTE,
+                    GRANULARITY_ONE_HOUR,
+                    GRANULARITY_ONE_DAY,
+                    GRANULARITY_TEN_SECONDS,
+                ],
             }
 
     return aggregation_options
--- a/snuba/datasets/metrics_messages.py
+++ b/snuba/datasets/metrics_messages.py
@@ -17,6 +17,9 @@ class OutputType(Enum):
 class AggregationOption(Enum):
     HIST = "hist"
     TEN_SECOND = "ten_second"
+    ONE_MINUTE = "one_minute"
+    ONE_HOUR = "one_hour"
+    ONE_DAY = "one_day"
 
 
 ILLEGAL_VALUE_IN_SET = "Illegal value in set."
--- a/snuba/datasets/metrics_messages.py
+++ b/snuba/datasets/metrics_messages.py
@@ -90,6 +90,13 @@ def aggregation_options_for_set_message(
 def aggregation_options_for_distribution_message(
     message: Mapping[str, Any], retention_days: int
 ) -> Mapping[str, Any]:
+    granularity_dict = {
+        "granularities": [
+            GRANULARITY_ONE_MINUTE,
+            GRANULARITY_ONE_HOUR,
+            GRANULARITY_ONE_DAY,
+        ],
+    }
     aggregation_options = {
         "min_retention_days": retention_days,
         "materialization_version": 2,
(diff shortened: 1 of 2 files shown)


john-z-yang (approved):
LGTM, might be worth adding a small test for sanity

ayirr7:
Will look into adding a small test

snuba/datasets/metrics_messages.py
```diff
@@ -16,6 +16,7 @@ class OutputType(Enum):
 
 class AggregationOption(Enum):
     HIST = "hist"
+    TEN_SECOND = "ten_second"
```
fpacifici, line 19: This seems it would make the configuration incongruent. Why is 10 second special enough to have its own flag while the others do not? 
Possible options:
- Have a flag per granularity so as I can turn on and off any granularity, not just the 10 seconds.
- Have a structured config where we have a field "granularity" that is a set of granularities that I want.

snuba/datasets/metrics_messages.py
```diff
@@ -25,6 +26,7 @@ class AggregationOption(Enum):
 INT_FLOAT_EXPECTED = "Int or Float expected"
 
 # These are the hardcoded values from the materialized view
+GRANULARITY_TEN_SECONDS = 0
```
fpacifici, line 29: Is the retention going to stay 90 days for the 10 seconds granularity?
This should not be a valid product requirement.  The usefulness of 10 seconds aggregated data from 80 days ago is very questionable. While the cost is substantial. 
I would advise being able to tune the retention time in function of granularity. So the user can the retention time they want for each granularity. 
Preferably, we would limit the max retention for 10 seconds granularity.

snuba/datasets/metrics_messages.py
```diff
@@ -75,48 +77,68 @@ def value_for_counter_message(message: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    granularity_dict = {
         "granularities": [
             GRANULARITY_ONE_MINUTE,
             GRANULARITY_ONE_HOUR,
             GRANULARITY_ONE_DAY,
         ],
     }
+    aggregation_options = {
+        "min_retention_days": retention_days,
+        "materialization_version": 2,
```
fpacifici, line 109: Why is this different from sets ? 
Also has the new matview been applied everywhere ?

fpacifici (commented):
I'd add test coverage for this change.

snuba/datasets/metrics_messages.py
ayirr7: Yep, we talked about this for the upcoming product usage of 10s granularity. They will not be using 90 days retention. 

We can look into limiting the max retention in a subsequent PR so that we can unblock the team for now, if that sounds fair to you?

snuba/datasets/metrics_messages.py
ayirr7: Materialization version 2 was planned to be added for all metric types I believe (see https://github.com/getsentry/snuba/pull/4477/files) but it seems that never happened for sets and counters. (@john-z-yang maybe you have some information on this?)

The new materialization version would allow us to restrict the `retention_days` based on the granularity being stored. So this would address your above comment as well, once the matview is applied to all metric types. 

One option is to apply the new matviews to all the metric types now, or, we could do it as a follow-up from this PR. Do you have a preference for the order in which we do these?

snuba/datasets/metrics_messages.py
ayirr7: I think we would be able to customize the retention based on granularity, after applying the new matviews (described in my comment below).

snuba/datasets/metrics_messages.py
ayirr7: I think the reason why 10 second granularity is singled out in this way is because the other granularities are all enabled by default, but 10s is not. 

But yea, I think this makes sense. I can try to reconfigure this (might go with option #2)

john-z-yang: @fpacifici Originally this `AggregationOption` thing was designed to be highly opinionated. The enums are suppose to be things like `ALERT` etc. Where snuba would parse it and apply the correct aggregations (for the case of `ALERT` it would be only producing a `[0]` in the `granularities` column and then set the retention to 7 days or less). But for various reasons, those requirements never came.

@nikhars I'm thinking of rebuilding this thing since people using it in a clearly different way than we intended. I think perhaps we could tie in with your idea of "having a better metrics use case creation experience" somehow.

snuba/datasets/metrics_messages.py
john-z-yang: > Why is this different from sets ?
Also has the new matview been applied everywhere ?

It's only for distribution because we use v2 to disable histogram

john-z-yang: > (also, side question for John, do we know why the default values for the granularity-specific retention days columns, such as decasecond_retention_days, are different between distributions and sets/counters in this PR? For decasecond_retention_days I see 7 for distributions but retention_days for sets/counters. Am I missing something?)

It was a mistake, we were going to change it but since mat view v2 is not happening for the other 2, those values never did anything.

snuba/datasets/metrics_messages.py
nikhars: I agree that the system for allowing different types of configurations needs a refactor. What we did earlier was very specific focusing on one detail only. If we want to allow more types of configurations to be possible, we should make the workflows more generic.

snuba/datasets/metrics_messages.py
nikhars: Personally, I do question the value 10s granularity provides. It means the product cares about deviances which might happen within 10s boundary intervals. Which I am not sure is needed or not. If there was a sustained deviance then a 60s granularity would provide some of the same signals that a 10s granularity would provide but at a very substantial cheaper cost. There could even be some middle ground where we provide 20s or 30s granularity but at a cheaper cost. But I do not know how the cost model is going to work for the custom use cases so we can revisit this topic in the future.

> Also, we had discussed this in the product requirements and I believe we're all on the same page that 10s granularity shouldn't be stored at 90 day retention.
Do we know how long of a retention is needed for the 10s granularity?

snuba/datasets/metrics_messages.py
nikhars: > Materialization version 2 was planned to be added for all metric types I believe (see https://github.com/getsentry/snuba/pull/4477/files) but it seems that never happened for sets and counters.
Yeah it was supposed to get applied on the other metric types as well. But priorities changed and we weren't able to apply it everywhere. We should fix this since now that there is a need to control configurations across all datasets.

ayirr7: Yeah, I agree. Do we have preferences on how this work should be done? Could this PR go in first and open up 10s granularity for a bit (and also advise the relevant product team to send in a lower `retention_days` in their metrics)? Or do we think these re-configurations should happen first?

snuba/datasets/metrics_messages.py
ayirr7: Yeah, I think it's valid to wonder what the exact need is for 10s granularity here. I think we can try to unblock the team for now, and maybe revisit this question once we all have a sense of the cost involved.

I'm double checking with the team on how much retention they need. 90 days was not a requirement, but we didn't talk about how long of a retention is required exactly.

snuba/datasets/metrics_messages.py
ayirr7: (I asked a similar question below, but) Do we have a preference for the order in which this work should be done? Do we think that this could be follow-up work after we unlock 10s granularity for the relevant product team?

snuba/datasets/metrics_messages.py
nikhars: Yes, we can revisit the topic later. No need to block merging this PR because of it.

snuba/datasets/metrics_messages.py
nikhars: > Yeah, I agree. Do we have preferences on how this work should be done? Could this PR go in first and open up 10s granularity for a bit (and also advise the relevant product team to send in a lower retention_days in their metrics)? Or do we think these new materialized views should go in first?

Those 2 things can be decoupled from each other. There isn't a need to sequence them.

snuba/datasets/metrics_messages.py
fpacifici: I think you could keep the configuration congruent and consistent without redesigning the whole configuration schema. This could be one by adding one flag per granularity interval. At least this makes the 10 seconds granularity not special. 
Then you can rearchitect it with a better overall schema separately.

@john-z-yang, I like the idea of having configuration templates. But it must be possible to introduce new ones easily, otherwise you will be in a position where the platform would need a lot of product knowledge to be useful.

ayirr7: Makes sense. I can add flags so that 10s granularity isn't a one-off in the configuration.

john-z-yang: Thanks, @fpacifici, how would you feel about
1. Sentry sending aggregation options in a list instead of a single one like I originally planned. Say if someone wants both histogram and 10 sec granularity, they could just add it to the list, so we do not get an explosion in this enum.
2. As for having a flag per granularity, I think for now we should only have a flag if it is disabled by default, and the existence of the flag means to enable it? It doesn't make a lot of sense to me why we want the other granularities as flag as that would imply there would be a product need to disable them.

cc @nikhars @ayirr7 on your thoughts

nikhars (approved):
Discussed offline that some of the refactoring will happen later. Approving the PR to make progress.

snuba/datasets/metrics_messages.py
fpacifici: > I think for now we should only have a flag if it is disabled by default, and the existence of the flag means to enable it? It doesn't make a lot of sense to me why we want the other granularities as flag as that would imply there would be a product need to disable them

With this approach how would you disable retentions that are enabled by default ? 
You have three options:
- treat everything as disabled by default and add enabled flags 
- the opposite and treat everything as enabled by default and have a flag to disable
- Have negative flags for enabled by default and positive flags for disabled by default.

I htink both 1 and 2 are ok. I like 3 less as it adds cognitive load in knowhing which flags are positive and which ones are negative.

PR #4750, opened 2023-09-23
feat(CoGS): Close CoGS Producer on Consumer shutdown

rahul-kumar-saini:
Consumer shutdown will close CoGS producer if it exists.


volokluev (commented):
Can you test this?

PR #4752, opened 2023-09-25
feat(CoGS): Add `shared-resources-usage` topic to Snuba validation

rahul-kumar-saini:
### Overview
- This is the topic that we will be producing cogs data to
- https://github.com/getsentry/ops/pull/7927 is blocked by this PR


lynnagara (approved):
Would you mind adding a schema to document what we can expect to see on this topic? 😀

PR #4765, opened 2023-09-26
CI: add clickhouse 21 checks to gocd 

dbanda:
* Adds the clickhouse 21 test to be checked in gocd
* Fixes a bug were we were checking shas against the old pipeline


lynnagara:
Does it actually need to block deploy? This test is sometimes flaky still.

evanh (approved):

PR #4769, opened 2023-09-26
bugfix(parser): allow multiple or conditions in `or` and `and` clauses

volokluev:
### Context

Even though `or(1, 1, 1, 1)` is valid snql, the clickhouse formatter chokes on it because there's an assumption in [other parts of the code](https://github.com/getsentry/snuba/blob/master/snuba/query/conditions.py#L237-L243) which assumes that ands and ors have two arguments. Breaking that assumption causes an infinite recursion error

### Solution

Rather than changing the matching infrastructure and making sure all the query processors using that assumption work afterwards, I just massaged the AST post-parsing to make sure that all `or` and `and` functions had at most two arguments

### DIFF
--- a/snuba/query/snql/parser.py
+++ b/snuba/query/snql/parser.py
@@ -560,9 +560,6 @@ class SnQLVisitor(NodeVisitor):  # type: ignore
     def visit_or_expression(
         self, node: Node, visited_children: Tuple[Any, Expression, Node]
     ) -> Expression:
-        import pdb
-
-        pdb.set_trace()
         _, left_condition, or_condition = visited_children
         args = [left_condition]
         # in the case of one Condition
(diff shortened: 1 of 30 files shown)


snuba/query/snql/parser.py
```diff
@@ -1478,12 +1513,16 @@ def parse_snql_query(
...
 ) -> Tuple[Union[CompositeQuery[QueryEntity], LogicalQuery], str]:
+
     with sentry_sdk.start_span(op="parser", description="parse_snql_query_initial"):
         query = parse_snql_query_initial(body)
 
     if settings and settings.get_dry_run():
         explain_meta.set_original_ast(str(query))
 
+    with sentry_sdk.start_span(op="processor", description="treeify_conditions"):
+        _post_process(query, [_treeify_or_and_conditions], settings)
```
enochtangg, line 1524: Looks good! Quick question, is there a reason why we shouldn't just shove this into `POST_PROCESSORS`?

volokluev: Good question, explained in comment

snuba/query/snql/parser.py
```diff
@@ -988,6 +988,40 @@ def transform(exp: Expression) -> Expression:
...
+    Adding this post-process step is easier than changing the rest of the query pipeline
+    """
+
+    def transform(exp: Expression) -> Expression:
+        if not isinstance(exp, FunctionCall):
+            return exp
+
+        if exp.function_name not in ("and", "or"):
+            return exp
+        args = exp.parameters
```
evanh, line 1010: There are helper functions that will do this for you already: https://github.com/getsentry/snuba/blob/master/snuba/query/conditions.py#L278

snuba/query/snql/parser.py
```diff
@@ -1478,12 +1513,19 @@ def parse_snql_query(
...
     settings: QuerySettings | None = None,
 ) -> Tuple[Union[CompositeQuery[QueryEntity], LogicalQuery], str]:
+
     with sentry_sdk.start_span(op="parser", description="parse_snql_query_initial"):
         query = parse_snql_query_initial(body)
 
     if settings and settings.get_dry_run():
         explain_meta.set_original_ast(str(query))
 
+    # NOTE (volo): The anonimizer that runs after this function call chokes on
```
evanh, line 1523:
```suggestion
    # NOTE (volo): The anonymizer that runs after this function call chokes on
```

snuba/query/snql/parser.py
```diff
@@ -1478,12 +1513,19 @@ def parse_snql_query(
...
+    # OR and AND clauses with multiple parameters so we have to teeify them
```
evanh, line 1524:
```suggestion
    # OR and AND clauses with multiple parameters so we have to treeify them
```

enochtangg (approved):

PR #4775, opened 2023-09-27
feat(COGS): `create_subscription` requests now support Tenant IDs

rahul-kumar-saini:
### Overview
- As part of figuring out what % of generic metrics subscriptions queries originate from which use case IDs, we need to add extra data to the create subscription request
- This has a side effect of also enabling us to categorize subscriptions by Organization ID once Sentry starts sending `tenant_ids` as part of the request
- Sentry change [here](https://github.com/getsentry/sentry/pull/57101)

### Changes
- `POST /<dataset>/<entity>/subscriptions` route (`create_subscription`) now supports `tenant_ids` in the request
- These Tenant IDs are stored in `SubscriptionData` but *not* serialized into Redis when the subscription is saved
    - For now, this PR will increment a simple metric checking how many subscription are vs aren't being created with `tenant_ids`

### DIFF
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -236,13 +236,16 @@ def record_subscription_created_missing_tenant_ids(request: Request) -> None:
     Used to track how often new subscriptions are created without Tenant IDs.
     """
     if request.referrer == SUBSCRIPTION_REFERRER:
-        if (
-            not (tenant_ids := request.attribution_info.tenant_ids)
-            or "organization_id" not in tenant_ids
-        ):
+        if not (tenant_ids := request.attribution_info.tenant_ids):
             metrics.increment("subscription_created_without_tenant_ids")
         else:
-            metrics.increment("subscription_created_with_tenant_ids")
+            metrics.increment(
+                "subscription_created_with_tenant_ids",
+                tags={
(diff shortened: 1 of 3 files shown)


evanh (commented):
Maybe a test?

snuba/web/query.py
```diff
@@ -229,6 +231,23 @@ def record_missing_use_case_id(request: Request, dataset: Dataset) -> None:
             )
 
 
+def record_subscription_created_missing_tenant_ids(request: Request) -> None:
+    """
+    Used to track how often new subscriptions are created without Tenant IDs.
+    """
+    if request.referrer == SUBSCRIPTION_REFERRER:
+        if not (tenant_ids := request.attribution_info.tenant_ids):
+            metrics.increment("subscription_created_without_tenant_ids")
```
lynnagara, line 240: Why would it ever be the case? Did we not update the code in Sentry to send them already?

rahul-kumar-saini: We have Tenant IDs on almost everything _except_ subscriptions because we didn't go down the route of updating subscriptions during capacity management work. Executor queries have no Tenant IDs yet because of this.

evanh (approved):
Pending CI passing looks good.

PR #4776, opened 2023-09-27
feat(COGS): Enable Subscriptions to save and execute with `tenant_ids`

rahul-kumar-saini:
### Overview
- Builds upon #4775
- Introduces the ability to actually save Tenant IDs into Redis and use them when executing the subscription queries

### How does it work?
- Enabling the `save_subscription_with_tenant_ids` will make the `SubscriptionData.to_dict()` method to start returning `tenant_ids` in the dict representation of the subscription data, if my understanding is correct this will:
    - Update the message posted to scheduled-subscriptions to also include a `tenant_ids` field
    - Update the data stored in Redis to include the same
    - Update the request made by `subscriptions_executor` to start including a `tenant_ids` field
    - Update the original request in messages produced to the subscription result topic to include `tenant_ids`
- All in all, it _should_ just work, ie it should just add Tenant IDs and have them accessible at query process time for us to analyze subscriptions queries by organization in querylog

### Gotchas
- Enabling this new form of subscription will require some sort of backfill of existing subscriptions to now include the new Tenant IDs so the stored subscriptions do not differ pre/post change, and this will likely need to be done every time a new Tenant ID (eg `use_case_id`) is introduced and is required for subscriptions query analysis

### DIFF
--- a/snuba/subscriptions/data.py
+++ b/snuba/subscriptions/data.py
@@ -45,6 +45,7 @@ from snuba.utils.metrics import MetricsBackend
 from snuba.utils.metrics.timer import Timer
 
 SUBSCRIPTION_REFERRER = "subscription"
+ENABLE_TENANT_IDS_RUNTIME_CONFIG = "save_subscription_with_tenant_ids"
 
 # These are subscription payload keys which need to be set as attributes in SubscriptionData.
 SUBSCRIPTION_DATA_PAYLOAD_KEYS = {
@@ -222,7 +223,7 @@ class SubscriptionData:
             "query": self.query,
         }
 
-        if get_config("save_subscription_with_tenant_ids", 0):
+        if get_config(ENABLE_TENANT_IDS_RUNTIME_CONFIG, 0):
             subscription_data_dict["tenant_ids"] = self.tenant_ids
 
         subscription_processors = self.entity.get_subscription_processors()
--- a/snuba/subscriptions/data.py
+++ b/snuba/subscriptions/data.py
@@ -39,7 +39,6 @@ from snuba.reader import Result
 from snuba.request import Request
 from snuba.request.schema import RequestSchema
 from snuba.request.validation import build_request, parse_snql_query
-from snuba.state import get_config
 from snuba.subscriptions.utils import Tick
 from snuba.utils.metrics import MetricsBackend
 from snuba.utils.metrics.timer import Timer
@@ -221,9 +220,6 @@ class SubscriptionData:
             "resolution": self.resolution_sec,
             "query": self.query,
         }
-
-        if get_config("save_subscription_with_tenant_ids", 0):
-            subscription_data_dict["tenant_ids"] = self.tenant_ids
 
         subscription_processors = self.entity.get_subscription_processors()
         if subscription_processors:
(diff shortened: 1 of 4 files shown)


fpacifici:
Two general comments.

I would change the name TenantId to something else. TenantId has a specific meaning in our infra. It is the tenant in Single Tenant. Using that name will be quite confusing.

> Enabling this new form of subscription will require some sort of backfill of existing subscriptions to now include the new Tenant IDs so the stored subscriptions do not differ pre/post change, and this will likely need to be done every time a new Tenant ID (eg use_case_id) is introduced and is required for subscriptions query analysis

Why does this need to be done again when a new use_case_id is added ?

snuba/subscriptions/data.py
```diff
@@ -163,13 +172,16 @@ def build_request(
                 custom_processing.append(validator.validate)
         custom_processing.append(partial(self.add_conditions, timestamp, offset))
 
+        tenant_ids = {**self.tenant_ids}
+        tenant_ids["referrer"] = referrer
+        if "organization_id" not in tenant_ids:
+            # TODO: Subscriptions queries should have an org ID
+            tenant_ids["organization_id"] = 1
```
fpacifici, line 179: Is this related with this PR or is it an unrelated bug ?

rahul-kumar-saini:
> I am not sure I understand the relationship between tenantId and the use case id for generic metrics subscriptions.

It would be very small code change to enforce `use_case_id` requirement on generic metrics queries this way, the other way to do it would be to introduce a top level field in the [request](https://github.com/getsentry/snuba-sdk/blob/main/snuba_sdk/request.py#L44-L50) itself and enforce that way.

@volokluev what do you think about `use_case_id` as a Tenant ID but not necessarily used to allocate resources?


> does it mean the capacity management is applied only per use case and not per organization id on the generic metrics platform?

It won't be automatically used by capacity management to start allocating resources, we'd have to create some policy for it (which we don't plan on doing as of now).

> Why does this need to be done again when a new use_case_id is added ?

Unless new use case IDs are introduced that divide and replace a previous one, we won't need to update all subscriptions. It needs to be done again if/when a new Tenant ID itself is added, requiring updating all previously saved subscriptions to attribute executed subscriptions queries appropriately.

In general, we're missing org id from subscriptions entirely at the moment. This solution solves that problem along with the `use_case_id` problem which is why I'm inclined to do it this way.

snuba/subscriptions/data.py
rahul-kumar-saini: We have no Tenant IDs currently for any subscription query, so for now it's just been Org ID 1 but if we actually wanted to get an insight into which orgs are using how much, we need the Tenant IDs to be propagated through from Sentry instead of just hardcoding values here.

This change is just cleaning this up to now accept org id if it exists (since after [this](https://github.com/getsentry/sentry/pull/57101) and [this](https://github.com/getsentry/snuba/pull/4775) we will start accepting Tenant IDs)

volokluev:
> @volokluev what do you think about use_case_id as a Tenant ID but not necessarily used to allocate resources?

I think it's a fine thing to do. The policy for `use_case_id` can just be that we are tracking its usage. We can decide later if we want to enforce anything

rahul-kumar-saini:
Dig this PR back up if/when one day Subscriptions need to be COGS'd 🫡

PR #4777, opened 2023-09-28
fix(spans): Allow span status to be optional

phacops:
Turns out span status is optional: https://develop.sentry.dev/sdk/event-payloads/span/#:~:text=Optional.%20Describes%20the%20status%20of%20the%20Span/Transaction.


lynnagara (approved):
Thanks!

PR #4778, opened 2023-09-28
Update GoCD Jsonnet library

mattgauntseo-sentry:
This fixes an issue in rollbacks that can cause multiple deploys of the same commit. Now rollbacks have a final manual stage that unpauses and unlocks the pipelines once an incident is complete.

For the curious: The I removed the pipeline-complete stage in a previous version, this caused pipelines to be unlocked after a rollback is complete (a problem folks were running into), however this meant the final stage in the deploy was the stage the rollback re-ran. Since this final stage was run, it triggered the downstream pipeline to run which was unintentional.

#skip-changelog

### DIFF
--- a/gocd/templates/jsonnetfile.json
+++ b/gocd/templates/jsonnetfile.json
@@ -8,7 +8,7 @@
           "subdir": "libs"
         }
       },
-      "version": "v2.2.2"
+      "version": "v2.3.1"
     }
   ],
   "legacyImports": true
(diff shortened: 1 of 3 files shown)


IanWoodard (approved):

rahul-kumar-saini (approved):
Let our team know what exactly is changing in eng-sns channel! ie what happened before vs now upon rollback

PR #4779, opened 2023-09-28
spans: add profile_id column

dbanda:
add profile_id column to spans.


snuba/snuba_migrations/spans/0007_spans_add_profile_id.py
```diff
@@ -0,0 +1,61 @@
...
+                ),
+                target=OperationTarget.LOCAL,
+                after="span_id",
+            ),
+            operations.AddColumn(
+                storage_set=storage_set_name,
+                table_name=dist_table_name,
+                column=Column(
+                    "profile_id",
+                    UInt(64),
```
nikhars, line 40: On the transactions table, profile_id field is a UUID [field](https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/transactions/0018_transactions_add_profile_id.py). This should probably be the same. Can you confirm @phacops ?

nikhars: As per conversation with Pierre offline, profile_id needs to be UUID field. Let's change it accordingly.

phacops: Indeed, it should be a `Uuid` column.

dbanda: ok fixed it.

snuba/snuba_migrations/spans/0007_spans_add_profile_id.py
```diff
@@ -0,0 +1,57 @@
...
+
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                column=Column(
+                    "profile_id", UUID(Modifiers(nullable=True, codecs=["ZSTD(1)"]))
```
nikhars, line 27: Not sure how much zstd compression would benefit on a UUID column. Not a blocker for the PR. Was just curious to know.

nikhars (approved):

snuba/snuba_migrations/spans/0007_spans_add_profile_id.py
dbanda: Yeah there isn't much benefit. In the docs they mentioned using ZSTD over LZ4 by default is better.

PR #4780, opened 2023-09-28
fix(generic-metrics): Update default `decasecond_retention_days` and `min_retention_days` columns for distributions

john-z-yang:
### Stack PR Chain

| [#1](https://github.com/getsentry/snuba/pull/4780) | ➤ Update default `decasecond_retention_days` and `min_retention_days` columns for distributions | In review          | Master                                               |
| [#2](https://github.com/getsentry/snuba/pull/4782) | Update default `decasecond_retention_days` and `min_retention_days` columns for sets | In review | [#1](https://github.com/getsentry/snuba/pull/4780) |
| [#3](https://github.com/getsentry/snuba/pull/4781) | Update default `decasecond_retention_days` and `min_retention_days` columns for counters | In review | [#2](https://github.com/getsentry/snuba/pull/4782) |

### Overview

Updates the default values for `decasecond_retention_days` and `min_retention_days` columns to 30 and `retention_days`

### DIFF
--- a/snuba/migrations/group_loader.py
+++ b/snuba/migrations/group_loader.py
@@ -304,6 +304,7 @@ class GenericMetricsLoader(DirectoryLoader):
             "0015_sets_add_options",
             "0016_counters_add_options",
             "0017_distributions_mv2",
+            "0018_distributions_update_opt_default",
         ]
 
 
--- a/snuba/snuba_migrations/generic_metrics/0018_distributions_update_opt_default.py
+++ b/snuba/snuba_migrations/generic_metrics/0018_distributions_update_opt_default.py
@@ -1,0 +1,68 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    storage_set_key = StorageSetKey.GENERIC_METRICS_DISTRIBUTIONS
+
+    local_table_name = "generic_metric_distributions_raw_local"
+    dist_table_name = "generic_metric_distributions_raw_dist"
+
+    before = [
+        Column(
+            "decasecond_retention_days",
+            UInt(8, MigrationModifiers(default=str("7"))),
+        ),
+        Column(
+            "min_retention_days",
+            UInt(8, MigrationModifiers(default=str("30"))),
+        ),
+    ]
+
+    after = [
+        Column(
+            "decasecond_retention_days",
+            UInt(8, MigrationModifiers(default=str("30"))),
+        ),
+        Column(
+            "min_retention_days",
+            UInt(8, MigrationModifiers(default=str("retention_days"))),
+        ),
+    ]
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.ModifyColumn(
+                storage_set=self.storage_set_key,
+                table_name=table_name,
+                column=column,
+                target=target,
+            )
+            for column in self.after
+            for table_name, target in [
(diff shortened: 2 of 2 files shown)


nikhars:
Is there a way to get the desired retention for the product team without modifying the default retention days? If we are modifying the default for 10s to 30 days, I would expect that the default retention days for 1 min should be higher than that maybe 60 days. And then there is a ripple effect of having increased default retention days.

Is the change necessary because of the condition we use in the materialized views?

john-z-yang:
> Is there a way to get the desired retention for the product team without modifying the default retention days? If we are modifying the default for 10s to 30 days, I would expect that the default retention days for 1 min should be higher than that maybe 60 days. And then there is a ripple effect of having increased default retention days.
> 
> Is the change necessary because of the condition we use in the materialized views?

The change is not necessary, the main reason I am changing it is because they are not consistent across the tables.

We can control this by updating the consumer to write to that column directly, like what we are doing with minute granularity (the default is 30 days but are we writing `retention_days` to it). I just wanted a reasonable value for them so they're all on the same page. Another option is just set everything to `retention_days` by default, and write our own values to it.

nikhars:
> Not at all, The main reason I am changing it is because they are not consistent across the tables, and it could get confusing. (I think the retention days for the other ones are all `retention_days`). I just wanted a reasonable value for them so they're all on the same page. Another option is just set everything to `retention_days` by default, and write our own values to it.

I see. Yeah having a consistent behavior across all types of metrics is desirable. What if we changed the other tables like `sets` and `counters` to have their defaults as they are present today on the distributions table? I believe the values on the distributions table make a lot of sense to me. What do you think?

john-z-yang:
> > Not at all, The main reason I am changing it is because they are not consistent across the tables, and it could get confusing. (I think the retention days for the other ones are all `retention_days`). I just wanted a reasonable value for them so they're all on the same page. Another option is just set everything to `retention_days` by default, and write our own values to it.
> 
> I see. Yeah having a consistent behavior across all types of metrics is desirable. What if we changed the other tables like `sets` and `counters` to have their defaults as they are present today on the distributions table? I believe the values on the distributions table make a lot of sense to me. What do you think?

Right now for distributions we have
7 days retention for 10 sec
30 days retention for 60 sec

For the others (sets/counters) we have
`retention_days` days retention for 10 sec
`retention_days` days retention for 60 sec

The migration proposed in this pr
30 days retention for 10 sec
`retention_days` days retention for 60 sec


I'm not sure how I feel applying 30 days retention for 60 sec, since we are manually writing `retention_days` to that column after we floated the idea of 30 days retention. Same goes with setting 7 days for 10 sec, because DDM will want 30 days. 
So we will need to manually write 30 to 10 sec retention days and `retention_days` to 60 sec retention days if we change all the other tables to be the same as distribution.


With migration proposed in this pr we don't need to write any values. But I honestly don't feel very strongly either way, I just want them to be the same, let me know what you think!

nikhars:
> With migration proposed in this pr we don't need to write any values.

My current thinking is along these lines. The clickhouse tables should have sane defaults. Defaults are just an idea of the action which needs to be applied if a column is empty. Whether an application relies on the default behavior or not is configurable. But the rules for applying restrictions should be standardized. In this specific scenario, the question is whether all metrics should always write some values for each granular retention or only provide overrides. If our defaults are not going to be sane, like the 10s granularity taking 30days by default, then where would the restriction of a metric being added needing a 10s granularity with a default retention of 7 days be enforced?

john-z-yang:
> > With migration proposed in this pr we don't need to write any values.
> 
> My current thinking is along these lines. The clickhouse tables should have sane defaults. Defaults are just an idea of the action which needs to be applied if a column is empty. Whether an application relies on the default behavior or not is configurable. But the rules for applying restrictions should be standardized. In this specific scenario, the question is whether all metrics should always write some values for each granular retention or only provide overrides. If our defaults are not going to be sane, like the 10s granularity taking 30days by default, then where would the restriction of a metric being added needing a 10s granularity with a default retention of 7 days be enforced?

That make sense to me, I will be closing this pr, and update the pr that change the other tables

PR #4782, opened 2023-09-29
fix(generic-metrics): fix(generic-metrics): Update default `decasecond_retention_days` and `min_retention_days` columns for sets

john-z-yang:
### Stack PR Chain

| [#1](https://github.com/getsentry/snuba/pull/4780) | ~~Update default `decasecond_retention_days` and `min_retention_days` columns for distributions~~ | closed          | master  |
| [#2](https://github.com/getsentry/snuba/pull/4782) | ➤ Update default `decasecond_retention_days` and `min_retention_days` columns for sets | In review | master |
| [#3](https://github.com/getsentry/snuba/pull/4781) | Update default `decasecond_retention_days` and `min_retention_days` columns for counters | In review | [#2](https://github.com/getsentry/snuba/pull/4782) |

### Overview

Right now, only the default values for `decasecond_retention_days` and `min_retention_days` in the distribution table is correct. This is because the distribution table is the only table that is using mat view V2 (which uses those 2 columns). As we start applying mat view V2 to the other 2 tables (counter/sets), we need to make sure the default values are correct in those tables.

Updates the default values for `decasecond_retention_days` and `min_retention_days` columns to 7 and 30 so that it is the same as the distribution table. Before, those values are set to `retention_days`.

### DIFF
--- a/snuba/migrations/group_loader.py
+++ b/snuba/migrations/group_loader.py
@@ -304,6 +304,7 @@ class GenericMetricsLoader(DirectoryLoader):
             "0015_sets_add_options",
             "0016_counters_add_options",
             "0017_distributions_mv2",
+            "0018_sets_update_opt_default",
         ]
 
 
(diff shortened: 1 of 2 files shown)


evanh (approved):
These PRs would be easier for other people to review if you added some more context about why these changes are being made.

john-z-yang:
> These PRs would be easier for other people to review if you added some more context about why these changes are being made.

Sounds good, I updated the description. Please let me know if it makes sense

nikhars (approved):

PR #4785, opened 2023-09-29
feat: Remove sessions datasets, storages, APIs and message processors

lynnagara:


hubertdeng123:
We've begun writing to the metrics backend for self-hosted as of 23.9.1 this month but want to give people a bit more time before a cutover. Let's plan to get this merged in November and say goodbye to sessions once and for all

hubertdeng123:
@lynnagara we can revisit this now that sessions is gone

mwarkentin:
Back from the dead? 🧟

PR #4786, opened 2023-09-29
feat: Port the commit log encoder/decoder to Rust

lynnagara:
Ports https://github.com/getsentry/arroyo/blob/f43a9d69ab0e504869f88fcc39ede98bc0cc6807/arroyo/backends/kafka/commit.py to Rust. The codec is not used yet.

### DIFF
--- a/rust_snuba/src/strategies/commit_log.rs
+++ b/rust_snuba/src/strategies/commit_log.rs
@@ -18,12 +18,13 @@ struct Payload {
 }
 
 #[derive(Error, Debug)]
-#[error(transparent)]
 enum CommitLogError {
-    #[error("encode error")]
-    EncodeError,
-    #[error("decode error")]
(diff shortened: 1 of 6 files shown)


rust_snuba/src/strategies/commit_log.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+struct Payload {
+    offset: u64,
+    orig_message_ts: f64,
+}
+
+impl TryFrom<KafkaPayload> for Commit {
+    type Error = anyhow::Error;
```
untitaker, line 20: can you define your own error type here?

you can use `thiserror` to derive the necessary traits automatically

I don't think we should use `anyhow` too pervasively in libraries such as rust-arroyo

PR #4787, opened 2023-09-29
fix(metrics): Update metrics processor to write use_case_id

enochtangg:
This PR is responsible for updating the metrics bucket processor to write use_case_id in ClickHouse. Currently in the metrics dataset (release-health), we are writing empty strings (`""`) as part of the `use_case_id`. As a result, this has caused some [confusion around querying for use_case_id](https://github.com/getsentry/sentry/pull/57005) in release-health. Since `use_case_id` is required in [kafka schema](https://github.com/getsentry/sentry-kafka-schemas/blob/8a28e031579e2d40e49a3086c74fdfbb93ac6fc7/schemas/snuba-metrics.v1.schema.json#L74), this field should be already be sent as part of the kafka message. Note: the metrics dataset only contains the `sessions` use_case_id anyways. A data migration might be necessary to follow this PR to update all the existing rows.

### DIFF
--- a/snuba/datasets/processors/metrics_aggregate_processor.py
+++ b/snuba/datasets/processors/metrics_aggregate_processor.py
@@ -83,6 +83,7 @@ class MetricsAggregateProcessor(DatasetMessageProcessor, ABC):
                 "org_id": _literal(message["org_id"]),
                 "project_id": _literal(message["project_id"]),
                 "metric_id": _literal(message["metric_id"]),
+                "use_case_id": _literal(message.get("use_case_id", "sessions")),
                 "timestamp": _call(
                     "toDateTime",
                     (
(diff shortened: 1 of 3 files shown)


ayirr7:
Do we need to make the change to the `MetricsAggregateProcessor` as well here?

enochtangg:
> Do we need to make the change to the `MetricsAggregateProcessor` as well here?

@ayirr7 Seems like we do? I can see the `use_case_id` column exist in ClickHouse for the tables in which `MetricsAggregateProcessor` writes to, but it seems like it's missing from the storages (e.g. https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/metrics/storages/sets.yaml). Is that an oversight?

evanh (approved):

snuba/datasets/processors/metrics_aggregate_processor.py
```diff
@@ -83,6 +83,7 @@ def process_message(
                 "org_id": _literal(message["org_id"]),
                 "project_id": _literal(message["project_id"]),
                 "metric_id": _literal(message["metric_id"]),
+                "use_case_id": _literal(message.get("use_case_id", "sessions")),
```
ayirr7, line 86: Just for documentation purposes -- I don't think we actually use the `AggregateProcessor` anywhere. But it's good to add for safety. We'll likely deprecate `AggregateProcessor` soon

PR #4799, opened 2023-09-29
feat(admin): Link to available Snuba Admin regions

rahul-kumar-saini:
### Overview
- Earlier today I was on s4s admin console and was confused for a split second as to why there were no runtime configs

### Changes
- Display current region
- Link to notion doc with all available snuba admin deployments

### Notes
- These could just be linked directly but I didn't want to put those regions in our codebase
- Alternative is to set a runtime config of all regions and have an API call to get them and display 
    -  I could do this if we really care but I think linking to the doc is sufficient

### Screenshot
Screenshot 2023-09-29 at 4 11 23 PM

### DIFF
--- a/snuba/admin/static/welcome/index.tsx
+++ b/snuba/admin/static/welcome/index.tsx
@@ -1,0 +1,48 @@
+import React from "react";
+
+function Welcome() {
+  const PROD_URL_START = "https://snuba-admin";
+
+  const AVAILABLE_REGIONS =
+    "https://www.notion.so/sentry/Snuba-Admin-8344d5d508014d4a867cdb1b8e8d22cf";
+
+  const URL = window.location.origin;
+
+  function urls() {
+    let current_region = "localhost";
+    if (URL.startsWith(PROD_URL_START)) {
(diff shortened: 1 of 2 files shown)


volokluev:
I like where this is going! Is it possible to just put the environment at the top bar like so

image

volokluev:
I'm also ok with the notion doc as a stopgap but we should put it in an ops setting to make life easier

snuba/admin/static/welcome/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+
+function Welcome() {
+  const PROD_URL_START = "https://snuba-admin";
+
+  const AVAILABLE_REGIONS =
+    "https://www.notion.so/sentry/Snuba-Admin-8344d5d508014d4a867cdb1b8e8d22cf";
```
dbanda, line 7: Maybe we should move this to settings. Feels weird having this in a public repo.

volokluev (approved):

dbanda (approved):

PR #4812, opened 2023-10-04
fix(spans): Handle no or null values for retention_days for the spans Rust consumer

phacops:


rust_snuba/src/processors/spans.rs
```diff
@@ -12,6 +12,7 @@ pub fn process_message(
     if let Some(payload_bytes) = payload.payload {
         let msg: FromSpanMessage = serde_json::from_slice(&payload_bytes).map_err(|err| {
             log::error!("Failed to deserialize message: {}", err);
+            println!("{:#?}", err);
```
lynnagara, line 15: Wanna combine that into the log line above?

lynnagara (approved):

PR #4815, opened 2023-10-04
feat(capman): update policies to skip cross org queries

volokluev:
in order to allow the cross org policy to take effect, we have to keep the non-cross org policies from rejecting the query. Add and test that functionality

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -437,6 +437,9 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
             and self._required_tenant_types == other._required_tenant_types
         )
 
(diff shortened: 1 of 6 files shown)


snuba/query/allocation_policies/__init__.py
```diff
@@ -635,7 +638,7 @@ def __validate_config_params(
 
         class_name = self.__class__.__name__
 
-        # config doesn't exist
+        # config doesn't existhttps://riseup.net/en
```
evanh, line 641: ?

evanh (approved):

snuba/query/allocation_policies/__init__.py
volokluev: oops, vim slip

PR #4817, opened 2023-10-04
fix(dlq): Add sane defaults to the dlq policy

nikhars:
The current dlq policy allows for 100% of problematic messages to be swallowed by the dlq. That would not be ideal since in effect we could be sending 100% of the messages to the dlq. Instead, we configure a max invalid ratio of 0.01 and max consecutive count of 1000 per partition. We set them both because the counts are not persisted across consumer restarts. So when a consumer crashes when the dlq policy limit is hit, the counters are reset.


lynnagara:
This only targets the multistorage consumer, can you update the main consumer too?

lynnagara (approved):

PR #4827, opened 2023-10-06
spans: add profile_id to tests 

dbanda:
Adds test for spans profile_id and fix bug where test was dependent on local timezone https://stackoverflow.com/questions/34622302/python-datetime-timestamp-issue

### DIFF
--- a/snuba/datasets/processors/spans_processor.py
+++ b/snuba/datasets/processors/spans_processor.py
@@ -1,4 +1,3 @@
-import logging
 import numbers
 import random
 import time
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/spans_processor.py
```diff
@@ -27,7 +27,7 @@
 )
 from snuba.utils.metrics.wrapper import MetricsWrapper
 
-logger = logging.getLogger(__name__)
+logger = structlog.get_logger(__name__)
```
nikhars, line 30: Have we used structlog in any of our consumer processors before? Not sure if there is a perf impact of using it or not. Do you happen to know?

dbanda: We use it subscriptions. Not sure if we've have done performance testing on ti.

PR #4834, opened 2023-10-09
feat(processor): Add 10s to generic metrics granularity processors

ayirr7:
Since we are now able to store metrics at the 10s granularity, we also want to make that data queryable. Adding 10s to the allowed granularities for the query processors.


ayirr7:
Will test locally next

evanh (approved):

PR #4836, opened 2023-10-10
ref: Remove use_progress_bytes_scanned option

evanh:
The migration to move to the new value has been completed, so this option can
be removed. Everything should be using this correct value now.


evanh:
Not merging this until we finish migrating ST customers.

volokluev (approved):
This policy is not enabled in single tenant, you can merge whenever

PR #4838, opened 2023-10-10
fix: Fix committing in the no-skip-write python consumer

lynnagara:
The consumer now always performs the Reduce step even if it's skipping writes. This is important since the commit strategy is IMMEDIATE and we would attempting to commit every offset otherwise.


untitaker (approved):
oh. i thought it was ONCE_PER_SECOND and discarded that theory!!! thanks for findign this

PR #4842, opened 2023-10-11
fix: fix 0001_functions and 0001_querylog for clickhouse 23

untitaker:
In order to upgrade to ClickHouse 23, we need to ensure all migrations run.

## functions

0001_functions uses a nullable column in the primary key. version 23 judges us for this, and one has to use allow_nullable_key to explicitly opt into permitting this `CREATE TABLE` statement.

To make matters worse, version 20.3 (a version we support) does not support the `allow_nullable_key` setting and will throw an error if it is set. So there is no table schema that works on both versions of ClickHouse.

However, the upside is that upgrading from ClickHouse version 21.6 to 23 appears to permit this schema regardless for old tables, but refuses to create new tables with the same schema. I tested this by recreating the container with the new image like so:

0. Be on `ghcr.io/getsentry/image-mirror-altinity-clickhouse-server:21.6.1.6734-testing-arm` due to MacOS M1
1. `docker container rm sentry_clickhouse`
2. patch sentry to use a different image `altinity/clickhouse-server:23.3.8.22.altinitystable`
3. `sentry devservices up`

`SHOW CREATE TABLE functions_mv_local` in version 23 will now return a piece of SQL that is not valid under the same version (as the `allow_nullable_key` setting is missing), but queries work.

Action plan:

1. make migrations on empty cluster pass by fixing the initial migration to conditionally set `allow_nullable_key` depending on the clickhouse version (this PR)
2. in production, upgrade to version 23 (TBD)
3. at this point, clusters may have inconsistent schemas (the setting is there or it is not, depending on when the table was created, NOT depending on the clickhouse version) -- but doesn't matter, the same set of queries work.
4. at a later point, when we are able to raise the minimum supported version to 23, check in a migration that adds the setting to existing tables (just for cleanliness and to make the output of `SHOW CREATE TABLE` valid again)

Note: **I don't think this is how ClickHouse intended us to upgrade versions at all.** It seems that they expected that applications only ever support two consecutive major versions at a time, so that we can instead:

1. upgrade everything from 20 to 21 without code changes
2. drop support for 20 and start setting `allow_nullable_key` unconditionally
3. upgrade everything from 21 to 22
4. drop support for 21
5. and so on until we reach version 23

## querylog

Very similar to functions, except a lot simpler.

0001_querylog has a `SAMPLE BY` clause that is invalid under ClickHouse version 23, as it is invalid to sample by a UUID. We don't use `SAMPLE BY` on querylog in practice, and it turns out that any use of `SAMPLE BY` already does not work today on version 20, as UUIDs were never intended to be supported.

So, removing the `SAMPLE BY` key in querylog has no downsides. In-place upgrades work, in that tables created on version 21 will just work in version 23, even though `SHOW CREATE TABLE` will return a schema that is invalid for new tables under version 23.

## risks

Diverging schemas between clickhouse versions is tricky. The fact that `SHOW CREATE TABLE` returns invalid SQL for newer versions of clickhouse might impact our ability to recreate a table in prod, should we choose to do so without a migration.

There is a gap in test coverage for our migrations. Ideally there would be a CI job that runs half of the migrations on version 20 or 21, performs an in-place upgrade in CI, then runs the other half.

### DIFF
--- a/snuba/migrations/migration.py
+++ b/snuba/migrations/migration.py
@@ -5,7 +5,7 @@ from typing import Optional, Sequence
 from snuba.clusters.cluster import get_cluster
 from snuba.migrations.check_dangerous import check_dangerous_operation
 from snuba.migrations.context import Context
-from snuba.migrations.operations import OperationTarget, RunPython, SqlOperation
+from snuba.migrations.operations import GenericOperation, OperationTarget, SqlOperation
 from snuba.migrations.status import Status
 from snuba.utils.types import ColumnStatesMapType
 
@@ -55,11 +55,11 @@ class CodeMigration(Migration, ABC):
     """
 
     @abstractmethod
-    def forwards_global(self) -> Sequence[RunPython]:
+    def forwards_global(self) -> Sequence[GenericOperation]:
         raise NotImplementedError
 
     @abstractmethod
-    def backwards_global(self) -> Sequence[RunPython]:
+    def backwards_global(self) -> Sequence[GenericOperation]:
         raise NotImplementedError
 
     def forwards(self, context: Context, dry_run: bool) -> None:
--- a/snuba/migrations/migration_utilities.py
+++ b/snuba/migrations/migration_utilities.py
@@ -1,4 +1,4 @@
-from typing import Tuple
+from typing import Set, Tuple
 
 from snuba.clusters.cluster import ClickhouseClientSettings, get_cluster
 from snuba.clusters.storage_sets import StorageSetKey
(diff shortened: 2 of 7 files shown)


lynnagara (changes requested):
I don't think we can just change historical migrations otherwise new and older installations are out of sync. Don't we need a plan for existing users who have already run this?

evanh:
Is it possible to make this changes in a new migration? Or do we have to have these changes in from the beginning?

evanh:
> We need both a plan for users who have already run this (who migrate from an existing cluster) and for new clusters.

Is this plan laid out in a doc somewhere? It would be helpful I think to lay out the cases and come up with a plan for them.

lynnagara (approved):

PR #4847, opened 2023-10-12
feat(generic-metrics): Migrations for all gauges tables

ayirr7:
Similar aggregations to other mat views (such as distributions) for the simple Aggregate Functions (`min`, `max`, `sum`, `count`). The main difference is the way `last` is aggregated. It utilizes the `ArgMax` function to get the latest running `last` value. Since it requires the `raw_timestamp` column (raw timestamps for individual data points from the raw table), we perform a `Max` aggregation on the `raw_timestamp` column in the Aggregated table. Regardless, we plan on making the `raw_timestamp` column non-queryable in the entity. 

To represent the fields of a Gauge value, we've decided on `Nested` since it's something we've used before in other mat views.


snuba/snuba_migrations/generic_metrics/0024_gauges_mv.py
```diff
@@ -0,0 +1,90 @@
...
+                    project_id,
+                    metric_id,
+                    arrayJoin(granularities) as granularity,
+                    tags.key,
+                    tags.indexed_value,
+                    tags.raw_value,
+                    toDateTime(multiIf(granularity=0,10,granularity=1,60,granularity=2,3600,granularity=3,86400,-1) *
+                      intDiv(toUnixTimestamp(timestamp),
+                             multiIf(granularity=0,10,granularity=1,60,granularity=2,3600,granularity=3,86400,-1))) as timestamp,
+                    retention_days,
```
john-z-yang, line 58: I think this retention days should be a multi-if to support different retention days for different granularity

ayirr7: Yep, gonna copy that logic for the new MVs. thanks!

ayirr7:
These are almost good to go. Testing various scenarios locally atm. I think the last piece we'll want is an AggregateFunction over `raw_timestamp` in the aggregate table.

snuba/snuba_migrations/generic_metrics/0024_gauges_mv.py
```diff
@@ -0,0 +1,94 @@
...
+                        multiIf(granularity=0,decasecond_retention_days,
+                                granularity=1,min_retention_days,
+                                granularity=2,hr_retention_days,
+                                granularity=3,day_retention_days,
+                                0)) as retention_days,
+                    minState(arrayJoin(gauges_values.min)) as min,
+                    maxState(arrayJoin(gauges_values.max)) as max,
+                    avgState(arrayJoin(gauges_values.avg)) as avg,
+                    sumState(arrayJoin(gauges_values.sum)) as sum,
+                    sumState(arrayJoin(gauges_values.count)) as count,
```
nikhars, line 67: Is `sumState` the right aggregation for count? What does count represent in the aggregated dataset?

snuba/snuba_migrations/generic_metrics/0024_gauges_mv.py
```diff
@@ -0,0 +1,94 @@
...
+                    argMaxState(arrayJoin(gauges_values.last), raw_timestamp) as last
+                FROM generic_metric_gauges_raw_local
+                WHERE materialization_version = 1
```
nikhars, line 70: Nit: Since all the other tables are on materialization_version = 2, it might be a good idea to keep this at version 2 as well to keep all materialization versions consistent. The versions could mismatch in the future but do not need to differ right now.

snuba/snuba_migrations/generic_metrics/0024_gauges_mv.py
nikhars: > Is `sumState` the right aggregation for count? What does count represent in the aggregated dataset?

I guess the answer is that `sum` column is the sum of all values and `count` column in the number of values which are seen? Is that correct?

ayirr7: Yes that's correct. The `count` value we will be receiving in the payload will indicate the number of times something was seen. So to aggregate those buckets, we need to `sum` the counts to get the aggregate count. Does this make sense?

snuba/snuba_migrations/generic_metrics/0024_gauges_mv.py
ayirr7: Done

nikhars (approved):

PR #4855, opened 2023-10-13
feat: Make subscription scheduler invalid interval a metric not warning

lynnagara:
This gets logged all the time but it's considered somewhat normal. It's about to get a lot more common if we move to a "received" timestamp instead of the timestamp from the Kafka topic which is largely monotonic. So let's remove the noisy.

Also moves the Comparable/Interval/InvalidRange types into the subscriptions module since they are fairly useless anywhere else. These haven't changed functionally at all.


snuba/subscriptions/scheduler_consumer.py
```diff
@@ -149,12 +154,7 @@ def poll(self, timeout: Optional[float] = None) -> Optional[BrokerValue[Tick]]:
                 )
                 offset_interval = Interval(previous_message.offset, commit.offset)
             except InvalidRangeError:
-                logger.warning(
-                    "Could not construct valid interval between %r and %r!",
-                    previous_message,
-                    MessageDetails(commit.offset, commit.orig_message_ts),
-                    exc_info=True,
-                )
+                self.__metrics.increment("invalid_interval")
```
nikhars, line 157: Would it be valuable to have atleast some messages being logged into Sentry? If we were just seeing this metric get incremented, do we have everything needed to troubleshoot the problem?

lynnagara: It won't be considered a problem anymore, as metric received time stamps are allowed to be out of order.

nikhars (approved):

PR #4863, opened 2023-10-16
ref(23.3) Fix test so it runs on 23.3

evanh:
23.3 has an issue with having trailing escape charaters in LIKE conditions.
Changing the test allows it to work on 23.3.

In the future it will probably be worth having validation on this specific case
but in the interest of unblocking the upgrade, that work will be tracked
separately.

### DIFF
--- a/tests/test_transactions_api.py
+++ b/tests/test_transactions_api.py
@@ -539,7 +539,7 @@ class TestTransactionsApi(BaseApiTest):
                     "tenant_ids": {"referrer": "r", "organization_id": 1234},
                     "project": 1,
                     "selected_columns": ["event_id"],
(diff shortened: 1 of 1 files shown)


dbanda (approved):

nikhars (approved):
Should we track somewhere the types of queries which can fail in the future so that we can come back to them later when we need to fix the root cause? I am unsure if product is making calls like this to Snuba right now. If they are, this most likely would break the product queries when the upgrade happens.

evanh:
> Should we track somewhere the types of queries which can fail in the future so that we can come back to them later when we need to fix the root cause?

I have a backlog task open to track this. https://getsentry.atlassian.net/browse/SNS-2501

PR #4864, opened 2023-10-16
ref(metrics): Add tests for the metrics SDK

evanh:
Add a beginner test suite that uses the MetricsSDK, which can be expanded as
the Metrics API changes and the SDK changes.


requirements-test.txt
```diff
@@ -1,6 +1,6 @@
 --index-url https://pypi.devinfra.sentry.io/simple
 
-snuba-sdk==2.0.1
+snuba-sdk==2.0.5
```
evanh, line 3: This isn't released yet, so this will initially fail.

nikhars:
I am unsure hence asking. Do we normally add SDK tests in Snuba in this manner for some of the other SDK features as well? Or is Metrics SDK special that its tests need to be defined in Snuba?

evanh:
> Or is Metrics SDK special that its tests need to be defined in Snuba?

It's not special. We do have other SDK tests in Snuba (see [here](https://github.com/getsentry/snuba/blob/master/tests/test_snql_sdk_api.py)).

These tests are basically a way to do an E2E test on querying Snuba using the SDK. Similar tests exist in Sentry, but having tests here is faster to iterate on and speeds up development.

nikhars (approved):

PR #4866, opened 2023-10-16
lint(generic-metrics): Remove feature flag for gen-metrics sets mat view version

john-z-yang:
Remove the flag that controls the materialization version


ayirr7:
Is there any chance we might want to keep this in case we ever wanted to toggle back to version 1?

nikhars (approved):

john-z-yang:
> Is there any chance we might want to keep this in case we ever wanted to toggle back to version 1?

I don't see why we would, we can always create a revert of this pr

PR #4892, opened 2023-10-18
feat(rust): Update all dependencies in lockfile

lynnagara:


nikhars (approved):
Should we have dependabot be configured to update rust dependencies as well?

PR #4893, opened 2023-10-18
feat(cardinality): Cardinalily analysis reporting

nikhars:
As we introduce more modules for metrics extraction from spans there is a bigger need to keep track of how cardinality evolves for different modules. This PR adds the queries which are used to perform cardinality analysis as a snuba cli command. The intention is for the snuba cli command to run as a cron job. The cron job would run the desired queries, dump the results in a CSV file locally and then send the file to a slack channel.

### Testing
Tested locally with some fake data and verified that the files get sent to slack

Screenshot 2023-10-18 at 3 40 53 PM

### DIFF
--- a/snuba/cli/spans_cardinality_analyzer.py
+++ b/snuba/cli/spans_cardinality_analyzer.py
@@ -1,0 +1,74 @@
+import csv
+from typing import Optional, Sequence
+
+import click
+import structlog
+
+from snuba import settings
+from snuba.admin.notifications.slack.client import SlackClient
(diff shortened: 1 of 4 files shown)


evanh (approved):
Would any of these also be helpful in the cardinality tool in the admin tool as well?

nikhars:
> Would any of these also be helpful in the cardinality tool in the admin tool as well?

These queries are already provided as pre defined queries in cardinality analysis of snuba admin

PR #4896, opened 2023-10-19
feat(generic-metrics): Add gauges to dataset and processor

ayirr7:


snuba/datasets/metrics_messages.py
```diff
@@ -74,6 +81,30 @@ def value_for_counter_message(message: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    sum = values["sum"]
+    count = values["count"]
+
+    return {
+        "metric_type": OutputType.GAUGE.value,
+        "gauges_values.min": [values["min"]],
+        "gauges_values.max": [values["max"]],
+        "gauges_values.sum": sum,
+        "gauges_values.count": count,
+        "gauges_values.avg": [sum / count],
```
nikhars, line 103: Not sure if count can be 0. But you might want to be safe and protect against division by 0.

snuba/datasets/metrics_messages.py
```diff
@@ -141,3 +172,23 @@ def aggregation_options_for_counter_message(
...
+    message: Mapping[str, Any], retention_days: int
+) -> Mapping[str, Any]:
+    settings = {
+        "granularities": [
+            GRANULARITY_ONE_MINUTE,
+            GRANULARITY_ONE_HOUR,
+            GRANULARITY_ONE_DAY,
+        ],
+        "min_retention_days": retention_days,
+        "materialization_version": 2,
```
nikhars, line 187: The materialization version on the migration is 1. Here you are using 2. Lets fix the migration to have the materialization version of 2.

snuba/datasets/metrics_messages.py
```diff
@@ -141,3 +172,23 @@ def aggregation_options_for_counter_message(
...
+    }
+
+    if aggregation_setting := message.get("aggregation_option"):
+        parsed_aggregation_setting = AggregationOption(aggregation_setting)
+        apply_aggregation_option(settings, parsed_aggregation_setting)
```
nikhars, line 192: What if upstream provides an aggregation option of HIST? We should make this more robust such that certain aggregation options only apply to certain datasets.

PR #4900, opened 2023-10-19
feat(rust): Invalid message has partition and offset properties

lynnagara:
This is needed so we can DLQ these messages later - this matches how InvalidMessage works in Python Arroyo

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/strategies/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/strategies/mod.rs
@@ -14,7 +14,10 @@ pub struct MessageRejected<T> {
 }
 
 #[derive(Debug, Clone)]
-pub struct InvalidMessage;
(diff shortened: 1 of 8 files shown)


dbanda:
Is the plan to begin using `InvalidMessage` in a follow up PR? What's the difference in `InvalidMessage` and `BadMessage` usecases?

lynnagara:
> Is the plan to begin using `InvalidMessage` in a follow up PR? What's the difference in `InvalidMessage` and `BadMessage` usecases?

InvalidMessage is the Arroyo contraption that get's used for DLQ, etc. This needs the partition and offset.

`BadMessage` is non Arroyo and it's part of the MessageProcessor interface (this is the part that is owned and implemented by product teams). It doesn't need the partition/offset on it.

dbanda (approved):

PR #4901, opened 2023-10-19
feat(rust): Add strategy that does json schema validation

lynnagara:


rust_snuba/src/strategies/validate_schema.rs
```diff
@@ -0,0 +1,185 @@
...
+        SchemaValidator {
+            schema,
+            enforce_schema,
+        }
+    }
+}
+
+impl TaskRunner<KafkaPayload, KafkaPayload> for SchemaValidator {
+    fn get_task(&self, message: Message<KafkaPayload>) -> RunTaskFunc<KafkaPayload> {
+        let mut errorred = false;
```
dbanda, line 40:
```suggestion
        let mut errored = false;
```

rust_snuba/src/strategies/validate_schema.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        if let Some(schema) = &self.schema {
+            let payload = message.payload().payload.unwrap();
+
+            let res = schema.validate_json(&payload);
+
+            if let Err(err) = res {
+                log::error!("Validation error {}", err);
+                if self.enforce_schema {
+                    errorred = true;
```
dbanda, line 50:
```suggestion
                    errored = true;
```

rust_snuba/src/strategies/validate_schema.rs
```diff
@@ -0,0 +1,185 @@
...
+                };
+            }
+        }
+
+        Box::pin(async move {
+            if errorred {
```
dbanda, line 56:
```suggestion
            if errored {
```

rust_snuba/src/strategies/validate_schema.rs
```diff
@@ -0,0 +1,185 @@
...
+    inner: Box<dyn ProcessingStrategy<KafkaPayload>>,
+}
+
+impl ValidateSchema {
+    #[allow(dead_code)]
+    pub fn new<N>(next_step: N, topic: String, enforce_schema: bool) -> Self
+    where
+        N: ProcessingStrategy<KafkaPayload> + 'static,
+    {
+        let concurrency = 5;
```
dbanda, line 75: Should this be hardcoded?

lynnagara: Probably not, updated to an arg

untitaker (approved):

PR #4903, opened 2023-10-20
feat(oncall): add error handling for corrupted payloads

volokluev:
A single tenant customer has some garbage data in their releases that causes the response to crash

if this happens we go through the payload and hex-encode the non-string byte values

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -418,6 +418,47 @@ def snql_dataset_query_view(*, dataset: Dataset, timer: Timer) -> Union[Response
         assert False, "unexpected fallthrough"
 
 
+def _sanitize_payload(payload: Dict[str, Any], res: Dict[str, Any]) -> None:
+    def hex_encode_if_bytes(value: Any) -> Any:
+        if isinstance(value, bytes):
+            try:
+                return value.decode("utf-8")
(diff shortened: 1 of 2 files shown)


snuba/web/views.py
```diff
@@ -418,6 +418,47 @@ def snql_dataset_query_view(*, dataset: Dataset, timer: Timer) -> Union[Response
...
+                    _sanitize_payload(item, res[hex_encode_if_bytes(k)][-1])
+                else:
+                    res[hex_encode_if_bytes(k)].append(hex_encode_if_bytes(item))
+        else:
+            res[hex_encode_if_bytes(k)] = hex_encode_if_bytes(v)
+
+
+def dump_payload(payload: Dict[str, Any]) -> str:
+    try:
+        return json.dumps(payload, default=str)
```
evanh, line 450: Doesn't adding `ensure_ascii=True` to this do the same thing as the `_sanitize_payload` function?

volokluev: no it doesn't. it still fails with 

```
FAILED tests/web/test_views.py::test_response_dumping - UnicodeDecodeError: 'utf-8' codec can't decode byte 0x83 in position 2: invalid start byte
```

evanh (approved):

PR #4905, opened 2023-10-20
feat(COGS): All Snuba API querying is COGS'd

rahul-kumar-saini:
### Overview
- Moving forward with recording cogs for all Snuba API and not just Generic Metrics queries
- Generic metrics queries are currently 100% sampled for cogs recording
- This will roll out similarly to 100% of all queries sampled for cogs recording


snuba/querylog/__init__.py
```diff
@@ -99,26 +99,29 @@ def _record_cogs(
...
+    if not isinstance(result, QueryResult):
         return
 
     profile = result.result.get("profile")
     if not profile or (bytes_scanned := profile.get("progress_bytes")) is None:
         return
 
-    if (use_case_id := request.attribution_info.tenant_ids.get("use_case_id")) is None:
-        return
+    app_feature = query_metadata.dataset.replace("_", "")
```
volokluev, line 112: would we want to add the referrer to this?

rahul-kumar-saini: I don't believe so, as far as cogs is concerned I think we just need a split of the API usage by dataset (and UseCaseID if applicable).

It wouldn't be hard to modify this to add referrer in future but as far as the requirements of the project go it's just dataset level.

volokluev (commented):
can we add a test?

snuba/querylog/__init__.py
evanh: Is there a way to preserve the dataset names so that we don't have weird data inconsistency ("genericmetrics" vs. "generic_metrics")? Maybe use a different separator character for the use case?

rahul-kumar-saini: cc @fpacifici 

This may be because other sources recording cogs call it "genericmetrics", but I could be wrong and we can change it to `"generic_metrics:transactions"` or similar pattern.

snuba/querylog/__init__.py
```diff
@@ -101,26 +101,29 @@ def _record_cogs(
...
+
+    if (
+        query_metadata.dataset == "generic_metrics"
+        and (use_case_id := request.attribution_info.tenant_ids.get("use_case_id"))
+        is not None
+    ):
+        app_feature += f"_{use_case_id}"
```
fpacifici, line 121: I think this approach is missing something. 
There are two independent resources that the snuba api should account for:
- the snuba api. Which should probably not be based on the bytes scanned 
- the clickhouse compute for shared datasets which should be based on bytes scanned.

Rationale:
- There are two shared resources to break down by feature from the snuba api: the snuba api itself (the k8s deployment). The shared clickhosue clusters (generic metrics clusters only basically today)
- When recording usage broken down by app_feature we provide three fields: the app_feature (which is the product); the resource_id (which identifies what we are breaking down. This can be provided by anybody involved with the cogs project); the amount.
- Ideally we would account clickhouse usage based on the CPU time spent on each query as reported by clickhouse. That is tricky to do and requires us to re-architect the way we query. So we can build a good enough approximation based on bytes scanned.
- The bytes scanned has very little to do with the snuba api usage. Snuba api could be better measured with the cpu time spent in snuba, which means api time - clickhosue time.

The approach outlined in this PR mixes the issues above. It provides a breakdown for the snuba api only (resource_id="snuba_api_bytes_scanned",) but based on bytes scanned that is not related with the snuba api cpu time. 

We should record either one or two usages per query (depending on the query):
- for every query record the time spent in snuba. app_feature can be what we do here: for shared datasets do what we do for generic metrics, for the rest use dataset name. 
- for the queries that hit a shared dataset also record the bytes scanned with a different resource_id. There would be one per cluster. I can provide the names, but those can be adjusted later. The existing ones are (`generic_metrics_clickhouse_gauges`, `generic_metrics_clickhouse_distributions`, `generic_metrics_clickhouse_sets`, `generic_metrics_clickhouse_counters`).

evanh: So two questions:

What is the resource_id for measuring the api time?
What is the app_feature for measuring the clickhouse compute bytes scanned?

evanh: Also I updated this to capture the bytes_scanned for a cluster. I will circle back on the API cogs.

fpacifici: > What is the resource_id for measuring the api time?
It is `snuba_api_bytes_scanned`. 
The issue there is not in the resource id but in using bytes_scanned to measure utilization. That should be query time.
Let's use `snuba_api`. I will take care of relabeling the kubernetes deployment

> What is the app_feature for measuring the clickhouse compute bytes scanned?
This is needed only for shared clusters, which afaik is only genericmetrics at this stage. 
For those I would use: `<CLUSTER_NAME>_<USE_CASE>` 
For generic metrics use case id should be in the query (but I am not sure this was actually provided everywhere before Rahul left). Anyway `genericmetrics_<USE_CASE_ID>`.

Does it answer your question ?

evanh (approved):

rahul-kumar-saini:
🚀 🚀 🚀

PR #4906, opened 2023-10-20
feat(CI): Shard the sentry tests more

john-z-yang:
Can't be that easy, can it?


volokluev (approved):

MeredithAnya:
I mean we basically do the same thing in sentry so I don't know why this would be a problem

PR #4912, opened 2023-10-23
feat(generic-metrics): Add support for gauges in dataset and processors

ayirr7:
This PR:

Storage + Entity Changes
- Adds both the readable and writeable storages for Gauges
- Adds the entity (for querying) 
- DOES NOT currently aim to support subscriptions (this will be in subsequent PRs)

Processor Changes
- Adds support within the generic metrics processor for gauge schema

**Local testing**
- Verified that gauges flow from `ingest-performance-metrics` Kafka topic, all the way to the Gauges aggregate table


snuba/datasets/metrics_messages.py
```diff
@@ -74,6 +81,29 @@ def value_for_counter_message(message: Mapping[str, Any]) -> Mapping[str, Any]:
...
+    assert isinstance(values, Mapping), "expected mapping for gauges"
+
+    for k in values:
+        assert isinstance(
+            values[k], (int, float)
+        ), f"{ILLEGAL_VALUE_IN_GAUGE} {INT_FLOAT_EXPECTED}: {values}"
+
+    # Build the nested values structure Clickhouse expects
+    sum = values["sum"]
+    count = values["count"]
```
nikhars, line 95: Nit: You can move this directly to lines 101-102

tests/datasets/processors/test_generic_metrics_processor.py
```diff
@@ -20,11 +21,37 @@
...
+                "metric_id": 1,
+                "value": {"min": 0, "max": 3, "sum": 3, "count": 2, "last": 3},
+            },
+            True,
+            id="gauge",
+        ),
+        pytest.param(
+            {"type": "d", "metric_id": 2, "value": 20},
+            False,
+            id="distribution",
```
nikhars, line 46: Is there a reason you added distribution to the gauge metrics processor test? I would imagine we would never want to do this in production.

nikhars (approved):
Overall LGTM

tests/datasets/processors/test_generic_metrics_processor.py
ayirr7: Oh, I think I mainly added it just to check that the `gauges` processor does not process a message with any other type other than `g`. 

I can remove it if that seems redundant.

evanh (approved):
LGTM too.

PR #4914, opened 2023-10-23
feat(scheduler): Scheduler can switch between different timestamps

lynnagara:
Scheduler can switch between using orig_message_ts and received_p99 for scheduling. Note this does not change anything functionally yet, as everything is still configured to orig_message_ts.

### DIFF
--- a/snuba/subscriptions/scheduler_consumer.py
+++ b/snuba/subscriptions/scheduler_consumer.py
@@ -41,7 +41,7 @@ class MessageDetails(NamedTuple):
     orig_message_ts: float
     # The timestamp the message was first received by Sentry (Relay)
     # It is optional since it is not currently present on all topics
-    received_ts: Optional[float]
+    received_p99: Optional[float]
 
 
 class CommitLogTickConsumer(Consumer[Tick]):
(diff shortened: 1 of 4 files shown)


snuba/subscriptions/scheduler_consumer.py
```diff
@@ -41,7 +41,7 @@ class MessageDetails(NamedTuple):
     orig_message_ts: float
     # The timestamp the message was first received by Sentry (Relay)
     # It is optional since it is not currently present on all topics
-    received_ts: Optional[float]
+    received_p99: Optional[float]
```
MeredithAnya, line 44: maybe im missing context but why the change from `received_ts` to `received_p99`?

MeredithAnya:
Also can you explain what the new timestamp is for and when we'd use it, just for a bit more context?

snuba/subscriptions/scheduler_consumer.py
lynnagara: Naming consistency. This is called received_p99 everywhere. Also, this was never used before and was introduced only in a recent commit.

lynnagara:
> Also can you explain what the new timestamp is for and when we'd use it, just for a bit more context?

We want to be able to use the p99 of "received" for subscriptions scheduling - that is the timestamp from when Relay first gets an event rather than the timestamp that Snuba receives the message. That way any delays/backlogs in the ingest consumer are also taken into account when we are scheduling subscriptions.

MeredithAnya (approved):

PR #4916, opened 2023-10-23
ref(migrations): raise error if no node found

MeredithAnya:
Ran into this earlier today where the cluster was misnamed so while there were no errors, the migrations did not actually run on the nodes we wanted, and in fact, updated the migrations table as if we had.

Assuming you've made it to the `execute` step of a migration, I think it's safe to say that if there are no nodes to run the migration on, something is wrong. 

cc @rgibert

### DIFF
--- a/snuba/migrations/operations.py
+++ b/snuba/migrations/operations.py
@@ -53,9 +53,19 @@ class SqlOperation(ABC):
             cluster.get_distributed_nodes(),
         )
 
+        cluster_name = f"{cluster.get_host()}:{cluster.get_port()}"
+        readable_name = cluster.get_clickhouse_cluster_name()
+        if readable_name:
+            cluster_name += readable_name
+
         if self.target == OperationTarget.LOCAL:
+            if not local_nodes:
+                print("MEEPS", local_nodes)
+                raise NodesNotFound(f"No nodes found for {cluster_name}")
             nodes = local_nodes
         elif self.target == OperationTarget.DISTRIBUTED:
+            if not cluster.is_single_node() and not dist_nodes:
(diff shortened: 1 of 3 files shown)


dbanda (approved):

snuba/migrations/operations.py
```diff
@@ -52,9 +53,19 @@ def get_nodes(self) -> Sequence[ClickhouseNode]:
...
         )
 
+        cluster_name = f"{cluster.get_host()}:{cluster.get_port()}"
+        readable_name = cluster.get_clickhouse_cluster_name()
+        if readable_name:
+            cluster_name += readable_name
+
         if self.target == OperationTarget.LOCAL:
+            if not local_nodes:
+                print("MEEPS", local_nodes)
```
dbanda, line 63: MEEPS?

snuba/migrations/operations.py
```diff
@@ -52,9 +53,19 @@ def get_nodes(self) -> Sequence[ClickhouseNode]:
             cluster.get_distributed_nodes(),
         )
 
+        cluster_name = f"{cluster.get_host()}:{cluster.get_port()}"
+        readable_name = cluster.get_clickhouse_cluster_name()
```
dbanda, line 57: do we allow this to be empty? is `cluster_name` only used for logs?

snuba/migrations/operations.py
MeredithAnya: oop thought I deleted haha

snuba/migrations/operations.py
MeredithAnya: we allow this to be optional, I'm not sure why

PR #4919, opened 2023-10-24
feat: Remove sessions consumer from devserver

lynnagara:
Has been deprecated for 2 years, will be fully removed shortly

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -54,18 +54,6 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
             ],
         ),
         (
-            "sessions-consumer",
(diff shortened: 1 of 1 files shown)


enochtangg:
Are we able to remove this? What about self-hosted?

enochtangg (commented):
Are we able to remove this? What about self-hosted?

lynnagara:
> Are we able to remove this? What about self-hosted?

Yes - this is safe to merge now as the devserver is not used in self hosted. Self hosted will be tearing down sessions completely in the November release. See https://github.com/getsentry/self-hosted/pull/2514

enochtangg (approved):

PR #4924, opened 2023-10-25
clickhouse upgrades: 22.8 fixes

MeredithAnya:
Ran test locally against 22.8 and I _think_ these two tests were the only troublesome ones?

* `test_clickhouse_type_mismatch_error`
* `test_error_handling` 

But wanted to put up a PR to verify that's the case

For the `test_error_handling` one, the format of the error has seemingly changed from something like:
```
Code: 60, e.displayText() = DB::Exception: Table default.not_existing doesn't exist. (version 19.11.4.24 (official build))
```
to
```
Code: 41. DB::ParsingException: Cannot parse datetime: (while reading the value of key timestamp): While executing ParallelParsingBlockInputFormat: (at row 2). (CANNOT_PARSE_DATETIME) (version 22.8.15.25.altinitystable (altinity build))
```

As for the `test_clickhouse_type_mismatch_error` one... when I put in the following snql from the test we can update the test to still be relevant https://github.com/getsentry/snuba/pull/4926

But the reason it was passing with the previous way it was written was due to

`
ClickHouse allows types to differ in the left and the right parts of IN subquery. In this case it converts the left side value to the type of the right side, as if the [accurateCastOrNull](https://clickhouse.com/docs/en/sql-reference/functions/type-conversion-functions#type_conversion_function-accurate-cast_or_null) function is applied. That means, that the data type becomes [Nullable](https://clickhouse.com/docs/en/sql-reference/data-types/nullable), and if the conversion cannot be performed, it returns [NULL](https://clickhouse.com/docs/en/sql-reference/syntax#null-literal).
`
which is documented in https://clickhouse.com/docs/en/sql-reference/operators/in#select-in-operators

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -153,6 +153,8 @@ def _build_migrations_cache() -> None:
         cluster = storage.get_cluster()
         database = cluster.get_database()
         nodes = [*cluster.get_local_nodes(), *cluster.get_distributed_nodes()]
+        mv_tables = []
+        non_mv_tables = []
         for node in nodes:
             if (cluster, node) not in MIGRATIONS_CACHE:
                 connection = cluster.get_node_connection(
@@ -162,6 +164,14 @@ def _build_migrations_cache() -> None:
                     f"SELECT name, create_table_query FROM system.tables WHERE database='{database}'"
                 )
                 for table_name, create_table_query in rows.results:
+                    if "MATERIALIZED VIEW" in create_table_query:
+                        mv_tables.append((table_name, create_table_query))
+                    else:
+                        non_mv_tables.append((table_name, create_table_query))
+                # when using the MIGRATIONS_CACHE we should make sure local
+                # tables are created before materialized views that depend on them
+                all_tables = mv_tables + non_mv_tables
+                for table_name, create_table_query in all_tables:
                     MIGRATIONS_CACHE.setdefault((cluster, node), {})[
                         table_name
                     ] = create_table_query
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -153,8 +153,6 @@ def _build_migrations_cache() -> None:
         cluster = storage.get_cluster()
         database = cluster.get_database()
         nodes = [*cluster.get_local_nodes(), *cluster.get_distributed_nodes()]
-        mv_tables = []
-        non_mv_tables = []
         for node in nodes:
             if (cluster, node) not in MIGRATIONS_CACHE:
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -162,6 +164,14 @@ def _build_migrations_cache() -> None:
                     f"SELECT name, create_table_query FROM system.tables WHERE database='{database}'"
                 )
                 for table_name, create_table_query in rows.results:
+                    if "MATERIALIZED VIEW" in create_table_query:
+                        mv_tables.append((table_name, create_table_query))
+                    else:
+                        non_mv_tables.append((table_name, create_table_query))
+                # when using the MIGRATIONS_CACHE we should make sure local
+                # tables are created before materialized views that depend on them
+                all_tables = mv_tables + non_mv_tables
```
MeredithAnya, line 173: I don't know why this hasn't caused a problem until now tbh since I kept getting `DB::Exception: Table snuba_test.generic_metric_counters_raw_local doesn't exist. Stack trace:` when running this for 22.8 locally too - but the other versions, including 22.3 seemed to work? anyhoo I think is just a weirdness of our tests cause running all the tests for `test_runner` individually work fine.

evanh:
For `test_clickhouse_type_mismatch_error`, does that cause an actual error at all? The condition optimizer is removing the `type != transaction` condition since it will always be true on the errors table. No need to actually spend CPU cycles checking.

tests/conftest.py
```diff
@@ -153,6 +153,8 @@ def _build_migrations_cache() -> None:
         cluster = storage.get_cluster()
         database = cluster.get_database()
         nodes = [*cluster.get_local_nodes(), *cluster.get_distributed_nodes()]
+        mv_tables = []
+        non_mv_tables = []
```
dbanda, line 157: These should be inside the loop otherwise we are creating all tables on all the nodes. It will pass CI but will be an inaccurate representation of the clusters.

tests/conftest.py
dbanda: Interesting. Maybe old versions of clickhouse maintained the order of creation in the query result.

dbanda (changes requested):

tests/conftest.py
MeredithAnya: ah good catch

tests/conftest.py
MeredithAnya: yeah perhaps

dbanda (approved):

PR #4926, opened 2023-10-25
fix: Test failed in specific versions of Clickhouse

evanh:
This test was failing in certain versions of Clickhouse because some versions
automatically cast the `type` condition to match types, meaning the test didn't
fail. Use a different column instead and don't try to match the exact error
message (since that changes between versions too).


volokluev (approved):
💯

PR #4928, opened 2023-10-25
Try Clickhouse 23.3 and 23.8

MeredithAnya:
Problems so far:

1. `tests/admin/test_api.py::test_query_trace`
    Seems like the `formatted_trace_output` maybe have changed for this specific query - so we maybe have to update the tracing tool. Or maybe the test just needs to be updated

    `{'5b9af86924cb': {'aggregation_performance': ['Aggregator: Merging partially aggregated single-level data.', 'Aggregator: Converted aggregated data to blocks. 1 rows, 8.00 B in 3.167e-06 sec. (315756.236 rows/sec., 2.41 MiB/sec.)'], 'memory_performance': ['MemoryTracker: Peak memory usage (for query): 225.13 KiB.'], 'node_name': '5b9af86924cb', 'node_type': 'query', 'number_of_storage_nodes_accessed': 0, 'read_performance': [], 'storage_nodes_accessed': [], 'thread_ids': ['399', '390'], 'threads_used': 2}}`

1. `tests/clickhouse/optimize/test_optimize.py`
    The following tests cases that fail are:
    * `TestOptimize.test_optimize[errors parallel]`
    * `TestOptimize.test_optimize[transactions parallel] `
    
    Not sure if related to the changes listed in https://docs.altinity.com/releasenotes/altinity-stable-release-notes/23.3/altinity-stable-23.3.8/#other-important-changes

### DIFF
--- a/tests/admin/test_api.py
+++ b/tests/admin/test_api.py
@@ -218,13 +218,9 @@ def test_predefined_system_queries(admin_api: FlaskClient) -> None:
     assert data[0]["name"] == "CurrentMerges"
 
 
-@pytest.mark.xfail
 @pytest.mark.redis_db
 @pytest.mark.clickhouse_db
 def test_query_trace(admin_api: FlaskClient) -> None:
-    """
-    TODO-23.3: Looks like the tracing output has changed:
-    """
     table, _, _ = get_node_for_table(admin_api, "errors_ro")
     response = admin_api.post(
         "/clickhouse_trace_query",
--- a/tests/admin/test_api.py
+++ b/tests/admin/test_api.py
@@ -218,9 +218,13 @@ def test_predefined_system_queries(admin_api: FlaskClient) -> None:
     assert data[0]["name"] == "CurrentMerges"
 
 
+@pytest.mark.xfail
 @pytest.mark.redis_db
 @pytest.mark.clickhouse_db
 def test_query_trace(admin_api: FlaskClient) -> None:
+    """
+    TODO-23.3: Looks like the tracing output has changed:
+    """
     table, _, _ = get_node_for_table(admin_api, "errors_ro")
     response = admin_api.post(
         "/clickhouse_trace_query",
--- a/tests/clickhouse/optimize/test_optimize.py
+++ b/tests/clickhouse/optimize/test_optimize.py
@@ -129,6 +129,7 @@ class TestOptimize:
         partitions = optimize.get_partitions_to_optimize(
             clickhouse, storage, database, table
         )
+        # XXX(FIX-LATER): This assertion breaks on ClickHouse 23.3
         assert partitions == []
 
         base = datetime(1999, 12, 26)  # a sunday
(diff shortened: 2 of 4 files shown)


tests/admin/test_api.py
```diff
@@ -218,9 +218,13 @@ def test_predefined_system_queries(admin_api: FlaskClient) -> None:
     assert data[0]["name"] == "CurrentMerges"
 
 
+@pytest.mark.xfail
 @pytest.mark.redis_db
 @pytest.mark.clickhouse_db
 def test_query_trace(admin_api: FlaskClient) -> None:
+    """
+    TODO-23.3: Looks like the tracing output has changed:
```
evanh, line 226: What was the problem with this? I can try to fix it.

MeredithAnya: @evanh The test assumes that the `formatted_trace_output` has data in the `read_performance` but from the payload of the test is the following

`{'5b9af86924cb': {'aggregation_performance': ['Aggregator: Merging partially aggregated single-level data.', 'Aggregator: Converted aggregated data to blocks. 1 rows, 8.00 B in 3.167e-06 sec. (315756.236 rows/sec., 2.41 MiB/sec.)'], 'memory_performance': ['MemoryTracker: Peak memory usage (for query): 225.13 KiB.'], 'node_name': '5b9af86924cb', 'node_type': 'query', 'number_of_storage_nodes_accessed': 0, 'read_performance': [], 'storage_nodes_accessed': [], 'thread_ids': ['399', '390'], 'threads_used': 2}}`

specifically `'read_performance': [],`

evanh: https://github.com/getsentry/snuba/pull/4939

tests/clickhouse/optimize/test_optimize.py
```diff
@@ -142,7 +142,8 @@ def test_optimize(
         partitions = optimize.get_partitions_to_optimize(
             clickhouse, storage, database, table
         )
-        assert partitions == []
+        # TODO: 23.3 Figure out why this fails for the parallel tests
```
MeredithAnya, line 145: So after doing some debugging it seems like for whatever reason, the parts that are created in the first run (e.g. `"transactions non-parallel"` test) hang around for the next test. Which is causing problems because it expects that you start out with no parts.

MeredithAnya:
I'm leaving the `test_optimize` test as xfail and we will come back to this later. The problem is not clickhouse, but rather how we are going testing this functionality (and also assumptions with clickhouse's internals) More info documented in https://getsentry.atlassian.net/browse/OPS-5105 

We will come back to this, right now we should have 23.3 in CI. 

Also removing 22.3 in the test suite since we are not planning to upgrade to that version, going to 22.8 instead

nikhars (approved):

nikhars (changes requested):
Actually lets put the next version as 23.8. Since most likely we will jump from 22.8 to 23.8

MeredithAnya:
> Actually lets put the next version as 23.8. Since most likely we will jump from 22.8 to 23.8

Also realized we didn't add that version in the image-mirror repo so adding just the 23.8 one https://github.com/getsentry/image-mirror/pull/18

nikhars (approved):

PR #4932, opened 2023-10-26
fix(migrations): Discover's migrations don't reflect our Prod setup.

rgibert:
Our current `discover` migrations don't leave Discover in a working state for our Production setup.  Discover is a `Merge` table of our `errors` and `transactions` clusters, the current migrations make `discover_dist` a `Distributed` table for `discover_local` which in turn is trying to be a `Merge` table of `errors_local` and `transactions_local`.  Since we don't have `errors_local` and `transactions_local` in the same cluster as `discover_local`, this leaves a new setup in a state where Discover is pointing to non-existent tables.

This latest (blocking) migration should get new setups to match our existing Production setup, it drops the existing `discover_dist` as a `Distributed` table and replaces it with a `Merge` table pointing to the `errors_dist` and `transactions_dist` tables which are then `Distributed` tables to their respective clusters.

### DIFF
--- a/snuba/snuba_migrations/discover/0008_discover_fix_dist_merge_table.py
+++ b/snuba/snuba_migrations/discover/0008_discover_fix_dist_merge_table.py
@@ -79,6 +79,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 storage_set=StorageSetKey.DISCOVER,
                 table_name="discover_dist",
                 columns=columns,
+                target=OperationTarget.DISTRIBUTED,
                 engine=table_engines.Distributed(
                     local_table_name="discover_local", sharding_key=None,
                 ),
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/discover/0008_discover_fix_dist_merge_table.py
```diff
@@ -0,0 +1,89 @@
...
+    Column("_tags_hash_map", Array(UInt(64))),
+    Column("contexts", Nested([("key", String()), ("value", String())])),
+    Column("trace_id", UUID(Modifiers(nullable=True))),
+    Column("span_id", UInt(64, Modifiers(nullable=True))),
+    Column("deleted", UInt(8)),
+]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = True
```
dbanda, line 52: Is there a reason this should be blocking? I think this migration would be fast. There will be a few dropped connections for a few seconds but the switch should be very quick. Nothing more than a network glitch. Ideally we could use https://clickhouse.com/docs/en/sql-reference/statements/exchange to make the rename atomic but it's not support on 20.3

dbanda (approved):

MeredithAnya:
I think https://github.com/getsentry/snuba/pull/5788 covered this so closing this PR

PR #4936, opened 2023-10-27
fix: Commit offsets from strategy join

untitaker:
The return value of Strategy.join() is currently ignored, because we
have no access to the consumer from within the assignment callbacks.

Shuffle a few things around so that consumer callbacks contain Arc to
the consumer internally.

SNS-2473

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/mod.rs
@@ -34,6 +34,13 @@ struct Strategies<TPayload: Clone> {
 
 struct Callbacks<TPayload: Clone> {
     strategies: Arc<Mutex<Strategies<TPayload>>>,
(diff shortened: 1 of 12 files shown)


rust_snuba/rust_arroyo/src/processing/mod.rs
```diff
@@ -69,7 +76,11 @@ impl<TPayload: 'static + Clone> AssignmentCallbacks for Callbacks<TPayload> {
                 s.close();
                 // TODO: We need to actually call consumer.commit() with the commit request.
                 // Right now we are never committing during consumer shutdown.
-                let _ = s.join(None);
+                if let Some(commit_request) = s.join(None) {
```
untitaker, line 77: this is the actual fix we're enabling

lynnagara (approved):
nice!

lynnagara:
This has stopped committing anything in prod, reverting

PR #4941, opened 2023-10-28
feat: Add the Snuba commit log compaction settings

lynnagara:
Add the compaction settings to match what we do in production.

Note: We currently have "compact" instead of "compact,delete" defined in prod, but this should be fixed.


evanh:
> Note: We currently have "compact" instead of "compact,delete" defined in prod, but this should be fixed.

Why not have this PR line up with production? Or is the plan to change production to "compact,delete"?

lynnagara:
Yeah, we need to change production. This caused an issue last week since we have changed the format of this topic and some years old data is never expired. Kicking this off in https://github.com/getsentry/ops/pull/8361.

evanh (approved):

PR #4945, opened 2023-10-30
fix: Fix SNUBA-3XE

lynnagara:
INC-548

### DIFF
--- a/snuba/subscriptions/types.py
+++ b/snuba/subscriptions/types.py
@@ -60,5 +60,7 @@ class Interval(Generic[T]):
     upper: T
 
     def __post_init__(self) -> None:
+        if self.lower is None or self.upper is None:
+            raise InvalidRangeError(self.lower, self.upper)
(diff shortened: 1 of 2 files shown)


mwarkentin (approved):
LGTM

snuba/subscriptions/types.py
```diff
@@ -60,5 +60,7 @@ class Interval(Generic[T]):
     upper: T
 
     def __post_init__(self) -> None:
+        if self.lower is None or self.upper is None:
+            raise InvalidRangeError(self.lower, self.upper)
```
mwarkentin, line 64: Add a matching test?

https://github.com/getsentry/snuba/blob/595c4eddac2c05d321b576841faea3d80ad2e333/tests/subscriptions/test_types.py#L10

lynnagara: done

evanh (approved):

PR #4951, opened 2023-10-30
ref(rust): Split factory into separate module

lynnagara:
consumer.rs is getting unwieldy, split the factory into a separate file. Also passes the whole StorageConfig object to the factory - we will need more data from this object later to produce to the commit log.

Most of this diff is just copy and paste, without functional changes

### DIFF
--- a/rust_snuba/src/factory.rs
+++ b/rust_snuba/src/factory.rs
@@ -1,0 +1,139 @@
+use crate::config;
+use crate::processors;
+use crate::strategies::clickhouse::ClickhouseWriterStep;
+use crate::strategies::python::PythonTransformStep;
+use crate::types::{BadMessage, BytesInsertBatch, KafkaMessageMetadata};
+use rust_arroyo::backends::kafka::types::KafkaPayload;
+use rust_arroyo::processing::strategies::commit_offsets::CommitOffsets;
+use rust_arroyo::processing::strategies::reduce::Reduce;
+use rust_arroyo::processing::strategies::run_task_in_threads::{
+    RunTaskFunc, RunTaskInThreads, TaskRunner,
+};
+use rust_arroyo::processing::strategies::InvalidMessage;
+use rust_arroyo::processing::strategies::{ProcessingStrategy, ProcessingStrategyFactory};
+use rust_arroyo::types::{BrokerMessage, InnerMessage, Message};
+use std::sync::Arc;
+use std::time::Duration;
+
+pub struct ConsumerStrategyFactory {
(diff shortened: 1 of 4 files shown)


rust_snuba/src/factory.rs
```diff
@@ -0,0 +1,139 @@
...
+use rust_arroyo::processing::strategies::run_task_in_threads::{
+    RunTaskFunc, RunTaskInThreads, TaskRunner,
+};
+use rust_arroyo::processing::strategies::InvalidMessage;
+use rust_arroyo::processing::strategies::{ProcessingStrategy, ProcessingStrategyFactory};
+use rust_arroyo::types::{BrokerMessage, InnerMessage, Message};
+use std::sync::Arc;
+use std::time::Duration;
+
+pub struct ConsumerStrategyFactory {
```
fpacifici, line 18: I noticed there are no docstrings anywhere here.
Is it intentional ?

rust_snuba/src/factory.rs
```diff
@@ -0,0 +1,139 @@
...
+                    }
+                }
+
+                let task_runner = MessageProcessor { func };
+                Box::new(RunTaskInThreads::new(
+                    next_step,
+                    Box::new(task_runner),
+                    self.concurrency,
+                    Some("process_message"),
+                ))
```
fpacifici, line 127: Am I reading this wrong or are you dispatching each individual message to a thread without any batching ? 
Wouldn't that imply taking and releasing a lock for every message? Otherwise how are threads going to synchronize? I think there can be a huge performance gain in not having suck lock contention.

fpacifici (approved):
Two comments that are not specifically related with the content of the PR, but I think they should be addressed.

rust_snuba/src/factory.rs
lynnagara: Mostly they are 1:1 with Snuba and Arroyo and I don't want to maintain both yet. Still haven't figured out how we'll generate shared docs for Python and Rust in the arroyo repo (the goal is both implementations will live together). But will make sure these get populated before this is actually used for real.

rust_snuba/src/factory.rs
lynnagara: Haven't noticed any performance issues with this approach yet.

From the tokio docs:
> Tasks are the unit of execution managed by the scheduler. Spawning the task submits it to the Tokio scheduler, which then ensures that the task executes when it has work to do. The spawned task may be executed on the same thread as where it was spawned, or it may execute on a different runtime thread. The task can also be moved between threads after being spawned.

PR #4955, opened 2023-10-31
ci: Switch default ClickHouse version in CI to 21.8

lynnagara:
We are about to drop support for 20.3. Let's run 21.8 as default in CI


evanh (commented):
What about 20.7? Is that no longer being used?

onewland:
20.7 is no longer in use anywhere.

evanh (approved):

PR #4970, opened 2023-11-02
ref(gauges) Expose timestamp on the gauges entity

evanh:
The gauges entity was exposing the rounded_timestamp field as the default time
column. Change that to timestamp to keep consistency with the other entities.

### DIFF
--- a/snuba/datasets/configuration/generic_metrics/entities/gauges.yaml
+++ b/snuba/datasets/configuration/generic_metrics/entities/gauges.yaml
@@ -7,7 +7,7 @@ schema:
     { name: org_id, type: UInt, args: { size: 64 } },
     { name: project_id, type: UInt, args: { size: 64 } },
     { name: metric_id, type: UInt, args: { size: 64 } },
-    { name: rounded_timestamp, type: DateTime },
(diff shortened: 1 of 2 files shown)


iambriccardo:
Thanks so much for doing this, it simplifies our work a lot

snuba/datasets/configuration/generic_metrics/entities/gauges.yaml
```diff
@@ -130,6 +137,13 @@ storages:
   - storage: generic_metrics_gauges_raw
     is_writable: true
     translation_mappers:
+      columns:
+        - mapper: ColumnToColumn
+          args:
+            from_table_name:
+            from_col_name: timestamp
+            to_table_name:
+            to_col_name: rounded_timestamp
```
enochtangg, line 146: For `generic_metrics_gauges_raw`, it looks like the column name is `timestamp`. Do we still need this mapper?

ayirr7 (approved):

enochtangg (commented):
How come we don't need the `TimeseriesProcessor`, isn't `bucketed_time` still exposed?

PR #4976, opened 2023-11-03
feat(rust): Add strategy that produces the commit log topic

lynnagara:
Not yet integrated into the consumer

### DIFF
--- a/rust_snuba/src/strategies/commit_log.rs
+++ b/rust_snuba/src/strategies/commit_log.rs
@@ -210,19 +210,22 @@ mod tests {
             pub payloads: Vec<KafkaPayload>,
         }
         impl ProcessingStrategy<KafkaPayload> for Noop {
-            fn poll(&mut self) -> Option<CommitRequest> {
+            fn poll(&mut self) -> Result<Option<CommitRequest>, InvalidMessage> {
                 None
             }
             fn submit(
                 &mut self,
                 message: Message<KafkaPayload>,
-            ) -> Result<(), MessageRejected<KafkaPayload>> {
+            ) -> Result<(), SubmitError<KafkaPayload>> {
                 self.payloads.push(message.payload());
                 Ok(())
             }
             fn close(&mut self) {}
             fn terminate(&mut self) {}
-            fn join(&mut self, _timeout: Option<Duration>) -> Option<CommitRequest> {
+            fn join(
+                &mut self,
+                _timeout: Option<Duration>,
+            ) -> Result<Option<CommitRequest>, InvalidMessage> {
                 None
             }
         }
--- a/rust_snuba/src/strategies/commit_log.rs
+++ b/rust_snuba/src/strategies/commit_log.rs
@@ -186,6 +186,9 @@ impl ProcessingStrategy<KafkaPayload> for ProduceCommitLog {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rust_arroyo::backends::ProducerError;
+    use rust_arroyo::types::Topic;
+    use std::collections::BTreeMap;
 
     #[test]
     fn commit() {
@@ -211,7 +214,7 @@ mod tests {
         }
         impl ProcessingStrategy<KafkaPayload> for Noop {
             fn poll(&mut self) -> Result<Option<CommitRequest>, InvalidMessage> {
-                None
+                Ok(None)
             }
             fn submit(
                 &mut self,
@@ -226,7 +229,7 @@ mod tests {
                 &mut self,
                 _timeout: Option<Duration>,
             ) -> Result<Option<CommitRequest>, InvalidMessage> {
-                None
+                Ok(None)
             }
         }
 
@@ -274,7 +277,7 @@ mod tests {
 
         let strategy = ProduceCommitLog::new(next_step, producer, 1, &topic, false);
 
-        for p in payloads {
+        for payload in payloads {
             strategy.submit(Message::new_any_message(payload, BTreeMap::new()));
             strategy.poll();
         }
--- a/rust_snuba/src/strategies/commit_log.rs
+++ b/rust_snuba/src/strategies/commit_log.rs
@@ -93,7 +93,7 @@ impl TryFrom<Commit> for KafkaPayload {
 
 struct ProduceMessage {
     producer: Arc<dyn Producer<KafkaPayload>>,
-    topic: Arc<TopicOrPartition>,
+    topic: &'static TopicOrPartition,
     skip_produce: bool,
 }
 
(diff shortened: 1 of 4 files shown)


rust_snuba/src/strategies/commit_log.rs
```diff
@@ -81,6 +92,105 @@ impl TryFrom<Commit> for KafkaPayload {
     }
 }
 
+struct ProduceMessage {
+    producer: Arc<dyn Producer<KafkaPayload>>,
+    topic: Arc<TopicOrPartition>,
```
fpacifici, line 97: Why do you need an Arc for `TopicOrPartition`? If I remember correctly `TopicOrPartition` is an immutable data structure with static lifetime. 
It should be possible to jsut borrow it everywhere.

rust_snuba/src/strategies/commit_log.rs
```diff
@@ -81,6 +92,105 @@ impl TryFrom<Commit> for KafkaPayload {
...
+    }
+}
+
+impl TaskRunner<KafkaPayload, KafkaPayload> for ProduceMessage {
+    fn get_task(&self, message: Message<KafkaPayload>) -> RunTaskFunc<KafkaPayload> {
+        let producer = self.producer.clone();
+        let topic = self.topic.clone();
+        let skip_produce = self.skip_produce;
+
+        Box::pin(async move {
```
fpacifici, line 122: What do you achieve with this? (pinning the function here)

rust_snuba/src/strategies/commit_log.rs
```diff
@@ -81,6 +92,105 @@ impl TryFrom<Commit> for KafkaPayload {
...
+
+            let mut retries = 3;
+            while retries > 0 {
+                match producer.produce(&topic, message.payload()) {
+                    Ok(_) => {
+                        return Ok(message);
+                    }
+                    Err(e) => {
+                        retries -= 1;
+                        sleep(Duration::from_millis(100));
```
fpacifici, line 135: Shouldn't this be a parameter ?

rust_snuba/src/strategies/commit_log.rs
```diff
@@ -81,6 +92,105 @@ impl TryFrom<Commit> for KafkaPayload {
...
+                        log::warn!("{}", e);
+                    }
+                }
+            }
+            panic!("Failed to produce message");
```
fpacifici, line 140: I think we should return a Result from this function.

rust_snuba/src/strategies/commit_log.rs
```diff
@@ -81,6 +92,105 @@ impl TryFrom<Commit> for KafkaPayload {
...
+        })
+    }
+}
+
+pub struct ProduceCommitLog {
+    inner: Box<dyn ProcessingStrategy<KafkaPayload>>,
+}
+
```
fpacifici, line 148: What do we need this outer structure for ?

rust_snuba/src/strategies/commit_log.rs
```diff
@@ -101,4 +211,83 @@ mod tests {
...
+            fn close(&mut self) {}
+            fn terminate(&mut self) {}
+            fn join(&mut self, _timeout: Option<Duration>) -> Option<CommitRequest> {
+                None
+            }
+        }
+
+        struct MockProducer {
+            pub payloads: Vec<KafkaPayload>,
+        }
```
fpacifici, line 240: I think we built an in memory producer for tests. You could use that one.

rust_snuba/src/strategies/commit_log.rs
lynnagara: Removed, this was supposed to be temporary code

rust_snuba/src/strategies/commit_log.rs
lynnagara: updated RunTaskInThreads to support returning `RetryableError` as well as `InvalidMessage`

rust_snuba/src/strategies/commit_log.rs
lynnagara: All the strategies do this since the nested processing step doesn't have a statically known size. I think this is actually your code originally.

rust_snuba/src/strategies/commit_log.rs
lynnagara: Needs a refactor, will tackle it separately.

rust_snuba/src/strategies/commit_log.rs
lynnagara: We cannot spawn the future without pin, and it's part of the RunTaskInThreads interface.

https://rust-lang.github.io/async-book/04_pinning/01_chapter.html is probably a better explanation than I am able to give.

rust_snuba/src/strategies/commit_log.rs
lynnagara: Updated

lynnagara: I had to revert this, it is not static currently and would require a large change.

fpacifici: Mh do you remember why it is not static ?

rust_snuba/src/strategies/commit_log.rs
fpacifici: >  I think this is actually your code originally.

I would not trust it to be doing the right thing.

rust_snuba/src/strategies/commit_log.rs
lynnagara: I am not sure, it's been that way since the beginning. We can look into changing it but I tried and it doesn't seem like an easy change so i'm not going to bundle it in here.

rust_snuba/src/strategies/commit_log.rs
lynnagara: Actually, it doesn't even exist. We need to build it.

rust_snuba/src/strategies/commit_log.rs
lynnagara: Still, it's not a statically sized thing.. do you have another suggestion?

fpacifici: > All the strategies do this since the nested processing step doesn't have a statically known size.

Not sure why this helps. Let's say the strategy was not there as a wrapper, where would we encounter problems with processing steps not having a statically known size ?

fpacifici (commented):
Thanks for addressing the error management. 
Overall ok to me, though please address the test failure.

Also a few more questions inline.

rust_snuba/src/strategies/commit_log.rs
fpacifici: Ah now I remember. 
The consumer knows the partitions only after rebalancing, up until the revocation. 
That makes the partitions known to the consumer non static.

rust_snuba/src/strategies/commit_log.rs
lynnagara: Actually, the inner thing is always RunTaskInThreads in this case. So I can change the annotation and get rid of the box

fpacifici (approved):

PR #4981, opened 2023-11-03
fix(slo): Track exception class name on errors

evanh:
Add the exception class name to the metrics we emit on failure, so we can get a
better idea of what the actual bugs are. `get_request_status` didn't seem to be
correctly classifying invalid requests, so also change that.


volokluev (commented):
would we want to put the whole exception message in there?

volokluev (approved):

Issue #4596, opened 2023-08-04
Connection pool is full, discarding connection: sentry-clickhouse. Connection pool size: 1


rdziwinski:
### Self-Hosted Version

23.7.1

### CPU Architecture

x86_64

### Docker Version

containerd://1.6.21

### Docker Compose Version

Kubernetes v1.26.5

### Steps to Reproduce

I installed Sentry via Helm chart, configure project and it's working for 1 week. After that, I updated it and get an error as in title in two snuba conainers.
 Generally, Sentry is working but errors cause container restarting, and long term should make some problems.and 

### Expected Result

How can i fix this error?

### Actual Result

sentry-snuba-outcomes-consumer and sentry-sessions-consumer containers logs:
```
Connection pool is full, discarding connection: sentry-clickhouse. Connection pool size: 1
...
```

cabrinha:
Anyone know of a way to solve this issue?

cabrinha:
Is this issue with Snuba or with ClickHouse? https://github.com/sentry-kubernetes/charts/issues/977

cabrinha:
@rdziwinski any update on your side?

cabrinha:
> @rdziwinski afaik sentry-self-hosted only supports deployments via docker compose, not via kubernetes. If you want to have any luck of getting help, you should probably add which helm chart you used for deployment.

That doesn't really matter, it's all containerized and the configuration settings should be the same.

hubertdeng123:
Yep, we only support deployments via docker compose and this it's difficult to debug this based on our familiarity with your setup. We'd advise you to open an issue in the repo of the helm chart you're using.

rojinebrahimi:
I am getting the direct log:
2024-01-07 17:21:25,682 Connection pool is full, discarding connection: clickhouse. Connection pool size: 1

Which makes my containers shutdown after a while.
Using Sentry v23.12.1.

ajith-nataraj:
This issue still exists , need to re-open this issue

rdziwinski:
Hello,
i upgraded Sentry helm chart to 22.2.0 version and I confirm that env CLICKHOUSE_MAX_CONNECTIONS is available in deployments but seem it's not working. At many pods i still have error
` Connection pool is full, discarding connection: sentry-clickhouse. Connection pool size: 1`

PR #4988, opened 2023-11-06
feat(urllib3): allow to override connection pool settings for clickhouse

adonskoy:
urllib3 by default preserves only 1 connection to a specific host. Also if pool queue is full, it will discard new connections. This pull request adds an option that allow us to specify connection pull size and block new connections until a connection is available from the pool.

Fixes https://github.com/getsentry/snuba/issues/4596

### Legal Boilerplate


snuba/settings/__init__.py
```diff
@@ -81,6 +81,7 @@
     {
         "host": os.environ.get("CLICKHOUSE_HOST", "127.0.0.1"),
         "port": int(os.environ.get("CLICKHOUSE_PORT", 9000)),
+        "max_connections": int(os.environ.get("CLICKHOUSE_MAX_CONNECTIONS", 100)),
```
volokluev, line 84: where did this number come from?

adonskoy: Just a proposal. I have no information about how many requests snuba can process simultaneously. 

I don’t think that more than 100 connections to the clickhouse will be required for one api replica; with such values, it makes sense to scale horizontally rather than vertically.

evanh: I would leave this as `1` (which is what it is now).

evanh (commented):
I'm not sure I understand the motivation behind this PR. Can you give more context on what this improves?

adonskoy:
I looked into the problem in more detail and found that by default, urllib3 discards connections only when there are no free resources for a new connection. In cases where there are enough resources, the error does not appear.

Therefore, there is no point in overriding the default behavior, but it would be nice to leave the ability to change these values for self-hosted installations with a small number of available resources.

adonskoy:
> I'm not sure I understand the motivation behind this PR. Can you give more context on what this improves?

In my installation, in almost every snuba container there is at least one error in the log that the connection was rejected, but most of the errors occur in the outcomes consumer.

The urllib3 error in question is: `Connection pool is full, discarding connection: clickhouse. Connection pool size: 1`

snuba/clickhouse/http.py
```diff
@@ -291,6 +291,8 @@ def __init__(
         self,
         host: str,
         port: int,
+        max_connections: int,
```
evanh, line 294: Please make these optional parameters with defaults, both for this class and the `ClickhouseCluster` class. We have settings files outside of this repo, and if we merge this all those clusters will break.

adonskoy: Done

artemistomaras:
I have the same problem in my installation

evanh (approved):
Please fixing the typing issue and then I can merge this.

adonskoy:
I don't understand why the linter doesn't show these errors locally...

evanh:
Just FYI, there seems to be a bug or something in GIthub, I can't get it to run the required checks. I'm digging into it.

adonskoy:
If this happens due to unsuccessful image builds, then it is worth removing the cache push in the registry for non-members or not using the GITHUB_TOKEN, but generating a static one and using it in CI. Or create two separate build steps for different types of contributors.

In general, there is no ideal solution here

hd-deman:
any updates?

untitaker:
discussed two changes with @evanh yesterday:

* we're going to make google cloud build not-required because 1) doesn't work for external contributors 2) most of it already runs within github actions -- additionally we may want to strip it down because it duplicates what GHA does
* i'm looking into fixing `ci.yaml` because of the failed cache push

untitaker (approved):
since this PR was opened we introduced `snuba rust-consumer` command which may eventually replace `snuba consumer`. It uses something other than urllib3 to connect to clickhouse, and will not have this setting applied. if you encounter issues there we're happy to support this setting there as well, but i suspect that consumers should generally never run concurrent INSERTs against clickhouse anyway

snuba/settings/__init__.py
bkk-bcd: As far as a few of us can tell, this isn't working: https://github.com/sentry-kubernetes/charts/pull/1175

bkk-bcd: Can someone show me where the code is that actually respects/utilizes this setting?

PR #4990, opened 2023-11-06
feat(rust): Commit offsets during partition recovation

lynnagara:
Based on https://github.com/getsentry/snuba/pull/4936

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/mod.rs
@@ -34,6 +34,7 @@ struct Strategies<TPayload: Clone> {
 
 struct Callbacks<TPayload: Clone> {
     strategies: Arc<Mutex<Strategies<TPayload>>>,
+    consumer: Arc<Mutex<dyn Consumer<TPayload>>>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


rust_snuba/rust_arroyo/src/processing/mod.rs
```diff
@@ -67,9 +68,11 @@ impl<TPayload: 'static + Clone> AssignmentCallbacks for Callbacks<TPayload> {
             None => {}
             Some(s) => {
                 s.close();
-                // TODO: We need to actually call consumer.commit() with the commit request.
-                // Right now we are never committing during consumer shutdown.
-                let _ = s.join(None);
+                if let Ok(Some(commit_request)) = s.join(None) {
+                    let mut consumer = self.consumer.lock().unwrap();
+                    consumer.stage_offsets(commit_request.positions).unwrap();
+                    consumer.commit_offsets().unwrap();
```
fpacifici, line 74: I think you should really start raising errors appropriately instead of panicking sooner than later. It will be really hard to hunt down all these places later, the more logic is added. The likely result is that some consumers will panic in prod.

fpacifici (commented):
Not following something about this PR. The title says that you commit upon join but it seems that the PR implements committing upon revocation. Is the title wrong ?

lynnagara:
> Not following something about this PR. The title says that you commit upon join but it seems that the PR implements committing upon revocation. Is the title wrong ?

Updated comment

rust_snuba/rust_arroyo/src/processing/mod.rs
lynnagara: Will do it soon

fpacifici (approved):

rust_snuba/rust_arroyo/src/processing/mod.rs
untitaker: it is semantically more correct to propagate as `Result::Err`, but even in python we have no concrete idea what to do with those errors except crash the process.

PR #4993, opened 2023-11-08
feat(rust): Port the buffered messages class to Rust

lynnagara:
This is the start of the DLQ implementation. It adds the struct that keeps track of seen messages from the broker so they can be produced to the DLQ if needed later.
It is basically a 1:1 port of current Arroyo code here ( https://github.com/getsentry/arroyo/blob/112667abdc2ef5c9137e361c0b2bbf69703bc524/arroyo/dlq.py#L215-L269) that we run in the Python consumer. Actual DLQing of messages will be implemented as a follow up.

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/dlq.rs
+++ b/rust_snuba/rust_arroyo/src/processing/dlq.rs
@@ -47,8 +47,6 @@ impl<TPayload> BufferedMessages<TPayload> {
                     }
                 };
             }
-
-            return None;
         }
 
         None
--- a/rust_snuba/rust_arroyo/src/processing/dlq.rs
+++ b/rust_snuba/rust_arroyo/src/processing/dlq.rs
@@ -1,0 +1,95 @@
+#![allow(dead_code)]
+use crate::types::{BrokerMessage, Partition};
+use std::cmp::Ordering;
+use std::collections::{BTreeMap, VecDeque};
+
+/// Stores messages that are pending commit. This is used to retreive raw messages
+// in case they need to be placed in the DLQ.
+pub struct BufferedMessages<TPayload> {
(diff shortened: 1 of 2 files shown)


rust_snuba/rust_arroyo/src/processing/dlq.rs
```diff
@@ -0,0 +1,95 @@
...
+                    Ordering::Greater => {
+                        break;
+                    }
+                    Ordering::Less => {
+                        messages.pop_front();
+                    }
+                };
+            }
+
+            return None;
```
untitaker, line 51: i think this line can be removed

rust_snuba/rust_arroyo/src/processing/dlq.rs
```diff
@@ -0,0 +1,95 @@
...
+                    .insert(message.partition.clone(), VecDeque::from([message]));
+            }
+        };
+    }
+
+    // Return the message at the given offset or None if it is not found in the buffer.
+    // Messages up to the offset for the given partition are removed.
+    pub fn pop(&mut self, partition: &Partition, offset: u64) -> Option<BrokerMessage<TPayload>> {
+        if let Some(messages) = self.buffered_messages.get_mut(partition) {
+            #[allow(clippy::never_loop)]
```
untitaker, line 36: are you sure this lint is alerting on the code? can't see why it would, is it possible it alerted on an earlier version of it?

untitaker (approved):

rust_snuba/rust_arroyo/src/processing/dlq.rs
lynnagara: fixed

rust_snuba/rust_arroyo/src/processing/dlq.rs
lynnagara: yeah.. i've tried everything but it is alerting

PR #4994, opened 2023-11-08
fix(generic-metrics): Fix function names in gauges entity

ayirr7:
There were some typos where the `func` field in the entity definition were incorrect and need to point to the correct Clickhouse AggregateFunction we are using. From my understanding this should not be a functional change since `func` is just representing the name of the function


evanh (approved):

ayirr7:
@evanh just wanted to double check that this shouldn't really have any functional impact?

evanh:
> @evanh just wanted to double check that this shouldn't really have any functional impact?

Sorry for the late reply, yes @ayirr7 this should be fine.

PR #4996, opened 2023-11-08
fix(rust): Simplify querylog UUID serialization

lynnagara:
No need to implement our own way of serializing uuid


volokluev (approved):
nice

PR #4998, opened 2023-11-08
fix(metrics): add _raw_tags_hash column to distributed table (counters)

volokluev:
This already exists on the local tables, now we need to add it to the distributed ones

### DIFF
--- a/tests/test_snql_api.py
+++ b/tests/test_snql_api.py
@@ -77,6 +77,26 @@ class TestSnQLApi(BaseApiTest):
             [get_raw_transaction()],
         )
 
+    def test_raw_tags_hash(self) -> None:
+        query = """MATCH (generic_metrics_counters) SELECT countIf(value, equals(metric_id, 87269613)) AS `count(g:custom/gauge@none)` WHERE org_id = 447951 AND project_id IN array(5572016) AND timestamp >= toDateTime('2023-10-25T08:00:00') AND timestamp < toDateTime('2023-11-08T12:00:00') AND tags_raw[9223372036854776010] = 'dev' AND metric_id IN array(87269613) LIMIT 21 OFFSET 0 GRANULARITY 14400"""
+
+        response = self.post(
+            "/metrics/snql",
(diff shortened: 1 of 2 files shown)


tests/test_snql_api.py
```diff
@@ -77,6 +77,26 @@ def setup_teardown(self, clickhouse_db: None, redis_db: None) -> None:
             [get_raw_transaction()],
         )
 
+    def test_raw_tags_hash(self) -> None:
```
volokluev, line 80: will remove, just need to make sure this works on the distributed test

ayirr7:
An initial question, but is there a reason why these PRs need to be applied in a particular sequence?

volokluev:
because of they are all under the same dataset and we name things in numerical sequence `0024_blah_blah`, `0026_blah_blah`

Other than that no reason

tests/test_snql_api.py
```diff
@@ -77,6 +77,26 @@ def setup_teardown(self, clickhouse_db: None, redis_db: None) -> None:
...
+        query = """MATCH (generic_metrics_counters) SELECT countIf(value, equals(metric_id, 87269613)) AS `count(g:custom/gauge@none)` WHERE org_id = 447951 AND project_id IN array(5572016) AND timestamp >= toDateTime('2023-10-25T08:00:00') AND timestamp < toDateTime('2023-11-08T12:00:00') AND tags_raw[9223372036854776010] = 'dev' AND metric_id IN array(87269613) LIMIT 21 OFFSET 0 GRANULARITY 14400"""
```
ayirr7, line 81: Small note, should that MRI (`g:custom/gauge@none`) be a counter metric instead?

volokluev: it doesn't really matter, first of all it's just an alias. Also I should just remove that because what I really wanted was for CI to run with the distributed tables and actually hit the table and make sure I applied the migration correctly

evanh (approved):

PR #5001, opened 2023-11-09
fix(metrics): Add _tags_raw_hash to sets

volokluev:


snuba/datasets/configuration/generic_metrics/storages/sets.yaml
```diff
@@ -81,3 +81,9 @@ allocation_policies:
 query_processors:
   - processor: TableRateLimit
   - processor: TupleUnaliaser
+  - processor: MappingOptimizer
```
volokluev, line 84: we don't want to apply this with the migration. Remove this before merging once the distributed test passes

evanh (approved):

PR #5005, opened 2023-11-09
feat(settings): Allow runtime configurable referrer overrides

nikhars:
In certain cases we would like to allow a referrer to override certain clickhouse settings. For example, in single tenant environments where the cluster is completely owned by a tenant we might want to allow them to use all replicas in parallel using setting like max_parallel_replicas. Making it referrer based allows the setting to only kick in for certain high volume bursty referrers rather than applying the settings at all times on the whole dataset.

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -465,7 +465,7 @@ def _get_query_settings_from_config(
 ) -> MutableMapping[str, Any]:
     """
     Helper function to get the query settings from the config. Order of precedence
-    for overlapping config wihtin this method is:
+    for overlapping config within this method is:
     1. referrer/<referrer>/query_settings/<setting>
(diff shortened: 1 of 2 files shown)


snuba/web/db_query.py
```diff
@@ -461,9 +461,14 @@ def _get_cache_wait_timeout(
 def _get_query_settings_from_config(
     override_prefix: Optional[str],
     async_override: bool,
+    referrer: Optional[str],
 ) -> MutableMapping[str, Any]:
     """
-    Helper function to get the query settings from the config.
+    Helper function to get the query settings from the config. Order of precedence
+    for overlapping config wihtin this method is:
```
evanh, line 468:
```suggestion
    for overlapping config within this method is:
```

evanh (approved):

PR #5006, opened 2023-11-09
Add support for join queries to the subscription system

wedamija:
Issues team wants to create error metric alerts that filter out archived issues. To do this we need to join to the group attributes table. This pr adds the ability to support joins to the subscription system.

### DIFF
--- a/snuba/subscriptions/data.py
+++ b/snuba/subscriptions/data.py
@@ -9,11 +9,11 @@ from functools import partial
 from typing import (
     Any,
     Iterator,
+    List,
     Mapping,
     NamedTuple,
     NewType,
     Optional,
-    Sequence,
     Tuple,
     Union,
 )
@@ -96,7 +96,7 @@ class SubscriptionData:
     ) -> None:
         added_timestamp_column = False
         from_clause = query.get_from_clause()
-        entities = []
+        entities: List[Tuple[Optional[str], Entity]] = []
         if isinstance(from_clause, JoinClause):
             for alias, node in from_clause.get_alias_node_map().items():
                 assert isinstance(node.data_source, EntityDS), node.data_source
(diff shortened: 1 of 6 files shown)


wedamija:
I'll add tests here if this approach makes sense

snuba/subscriptions/data.py
```diff
@@ -94,49 +94,68 @@ def add_conditions(
...
+
+            subscription_processors = self.entity.get_subscription_processors()
+            if subscription_processors:
+                for processor in subscription_processors:
+                    processor.process(query, self.metadata, offset)
+
+        if not added_timestamp_column:
+            raise InvalidSubscriptionError(
+                "At least one Entity must have a timestamp column for subscriptions"
             )
```
wedamija, line 157: The group attributes table doesn't have an appropriate timestamp col here, which makes sense since it's not a timeseries table

snuba/subscriptions/data.py
```diff
@@ -94,49 +94,68 @@ def add_conditions(
...
+                new_condition = binary_condition(
+                    BooleanFunctions.AND, condition, new_condition
+                )
+
+            query.set_ast_condition(new_condition)
+
+            subscription_processors = self.entity.get_subscription_processors()
+            if subscription_processors:
+                for processor in subscription_processors:
+                    processor.process(query, self.metadata, offset)
```
evanh, line 152: I think you'll need to rework the processors here, because I suspect they expect simple queries instead of `CompositeQuery`. If any of the processors use the from clause you'll need to refactor them as well.

evanh (commented):
Yes, please add some tests as well.

snuba/subscriptions/data.py
wedamija: The only processor here is https://github.com/getsentry/snuba/blob/7d2de78aac9e8c6948b6c0c57c28c3245fac6f0f/snuba/datasets/entity_subscriptions/processors.py#L40. It expects

evanh (approved):

PR #5016, opened 2023-11-13
Intern `Topic` and make it `Copy`

Swatinem:
This uses a static append-only `IndexSet` to intern `Topic` names and turn them into tiny `u16` indices, making it trivially `Copy`.

Usage of the `RwLock` however means that reading via `as_str` needs to acquire a read lock, which might be expensive.

I’m not quite sure of the tradeoffs here. Having `Topic` (and `Partition`) be tiny and trivially `Copy` is nice, but needing to lock in order to turn it back into a `&str` can have a non-trivial cost as well.

This is on top of #5014.


untitaker (commented):
how much perf would we lose if we stored a `&'static str` in `Topic` instead of a u16?

That would:

* move all the unsafety into one place (right now you assume on L36 that other functions in the module do not call e.g. drain)
* get rid of indexmap dep because we can just store a `RwLock<BTreeSet<&'static str>>` for interning

but Topic is now u128-sized which is a lot, but I wonder if 99% of the perf-gains would come from being Copy-able at all

untitaker:
also as mentioned by @loewenheim in slack,  `RwLock.read()` can be slow.

could be avoided if `&'static str` is already in the `Topic`, but at this point we'd have to measure

rust_snuba/rust_arroyo/Cargo.toml
```diff
@@ -10,7 +10,9 @@ chrono = "0.4.26"
 clap = "2.18.0"
 env_logger = "0.9.0"
 futures = "0.3.21"
+indexmap = "2.1.0"
```
untitaker, line 13: don't think this is necessary anymore

untitaker (approved):
let's just do it

untitaker (approved):

PR #5022, opened 2023-11-14
fix(rust): Fix duration underflow

lynnagara:
Duration can't be negative so it is not safe to subtract elapsed from timeout like the prior approach did

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
+++ b/rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
@@ -84,24 +84,35 @@ impl<T: Clone + Send + Sync, TResult: Clone + Send + Sync> ProcessingStrategy<T>
 
     fn join(&mut self, timeout: Option<Duration>) -> Result<Option<CommitRequest>, InvalidMessage> {
         let start = Instant::now();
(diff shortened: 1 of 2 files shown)


rust_snuba/rust_arroyo/src/processing/strategies/reduce.rs
```diff
@@ -84,24 +84,35 @@ impl<T: Clone + Send + Sync, TResult: Clone + Send + Sync> ProcessingStrategy<T>
...
         }
+
+        let remaining = match timeout {
+            Some(t) => {
+                let elapsed = start.elapsed();
+                if elapsed > t {
+                    Some(Duration::ZERO)
+                } else {
+                    Some(t - elapsed)
+                }
```
Swatinem, line 111: you can also use `checked_sub(…).unwrap_or(Duration::ZERO)` here.
plus, using `timeout.map(|t| …)` as well instead of `match`

lynnagara: fixed, thanks

Swatinem (approved):

untitaker (approved):

PR #5024, opened 2023-11-14
Add a benchmark to test Rust consumers

Swatinem:


rust_snuba/rust_arroyo/Cargo.toml
```diff
@@ -12,9 +12,9 @@ env_logger = "0.9.0"
 futures = "0.3.21"
 log = "0.4.17"
 once_cell = "1.18.0"
-rdkafka = { version = "0.36", features = ["cmake-build"] }
+rdkafka = { version = "0.36", default-features = false, features = ["cmake-build"] }
```
Swatinem, line 15: the reason for this is that `rdkafka` has a defauft feature for zlib compression, which for reasons I do not understand caused some linker problems locally for me

Swatinem:
The good news here is that profiling with `samply` did not reveal anything too out of place.

The majority of time is spent is ser/de, and some `String` cloning within the `functions` processor.
I was surprised to see that the various `SystemTime::now` and `elapsed` show up in profiles, I did not expect that.

Then there is a bunch of async / tokio overhead, which is possibly also relate to the fact that the benchmark creates a new runtime for each sample it runs.

It might surely be a good idea to find a solution in which you don’t have to tear down and restart the runtime on each rebalance, though that would probably not help with getting more readable profiles in the benchmarking script, unless we change the way the benchmarks are run.

untitaker:
> It might surely be a good idea to find a solution in which you don’t have to tear down and restart the runtime on each rebalance, though that would probably not help with getting more readable profiles in the benchmarking script, unless we change the way the benchmarks are run.

with the ConcurrencyConfig we should have that now, right?

rust_snuba/benches/processors.rs
```diff
@@ -0,0 +1,364 @@
...
+                let res = processor.run_once();
+                if matches!(res, Err(RunError::Poll(ConsumerError::EndOfPartition))) {
+                    // FIXME: this pretty much means that we *polled* the partition to the end,
+                    // it does not mean that we actually *committed* everything
+                    // (aka we finished actually processing everything).
+                    return;
+                }
+            }
+            // FIXME: this seems to deadlock?
+            //processor.shutdown();
```
lynnagara, line 85: I guess you are wanting to run this benchmark with the StreamProcessor here, since it's more realistic. Though it won't fully work until the assignment callback locking problem is fixed. If you wanted to benchmark the strategies in isolation and submit messages directly to them, that could work already.

lynnagara (approved):
this is neat! approving as i think we can continue to iterate on this once it's in master

nvzqz:
> Then there is a bunch of async / tokio overhead, which is possibly also relate to the fact that the benchmark creates a new runtime for each sample it runs.

It looks like you're cloning the handle for each sample, where I believe handles are backed by `Arc`. So it doesn't look like you're making a new runtime each time. Once Divan supports async benchmarks, these kinds of things should hopefully not be an issue. Planning to have async support sometime in December.

Swatinem:
Yes that was an older comment, as `snuba` was creating a separate runtime for each factory invocation, which was fixed separately.

Being able to run async benchmarks is `divan` is definitely a nice thing to add, then I can just grab the `Handle::current()` anytime I need it.

PR #5025, opened 2023-11-14
Avoid Mutex on `AssignmentCallbacks`

Swatinem:
The `AssignmentCallbacks` don’t need to take `&mut self`, which can remove one `Mutex`.

The reason for this is that the only non-test usage has internal `Mutex` usage anyway.


untitaker (approved):
I believe we will at _some_ point have to mutate _something_ in callbacks, but I think it's probably better if ppl use finer-grained locking for that

PR #5032, opened 2023-11-15
Remove `BadMessage` in favor of `anyhow::Error`

Swatinem:
This removes the `BadMessage` error type and related manual logging and error conversion. It also centralizes logging of those errors, configuring anyhow in such a way that it captures stack traces in those cases, and sends those errors to Sentry, however without the failing payloads for now.

Apart from this, it also streamlines some of the processors, in particular removing manual `Uuid` parsing, and avoiding a bunch of `Clone`s.

fixes SNS-2523
however, SNS-2529 is not quite fixed by this


untitaker (approved):
oof yeah. i thought BadMessage was a rust-arroyo type, but this makes a lot of sense

PR #5036, opened 2023-11-15
fix(rust): Rename metrics namespace, and implement one metric from python

untitaker:
In order to make the Rust consumer emit the same metrics as Python
consumer, we need to use the same namespace. This is probably risky as
it could break alerts that do not apply the correct tag filters. We'll
just find out, I'm oncall anyway.

Also add a metric that already exists in Python, to see if it will show
up correctly.

### DIFF
--- a/rust_snuba/src/consumer.rs
+++ b/rust_snuba/src/consumer.rs
@@ -92,7 +92,7 @@ pub fn consumer_impl(
         configure_metrics(Box::new(StatsDBackend::new(
             &host,
             port,
-            "snuba.rust_consumer",
+            "snuba.consumer",
             tags,
         )));
     }
--- a/rust_snuba/rust_arroyo/src/processing/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/mod.rs
@@ -150,6 +150,9 @@ impl<TPayload: 'static> StreamProcessor<TPayload> {
     }
 
     pub fn run_once(&mut self) -> Result<(), RunError> {
+        let metrics = get_metrics();
+        metrics.increment("arroyo.consumer.run_once", 1, None);
+
         if self.is_paused {
             // If the consumer waas paused, it should not be returning any messages
             // on ``poll``.


rust_snuba/src/consumer.rs
```diff
@@ -92,7 +92,7 @@ pub fn consumer_impl(
         configure_metrics(Box::new(StatsDBackend::new(
             &host,
             port,
-            "snuba.rust_consumer",
+            "snuba.consumer",
```
lynnagara, line 95: How about swapping this out only if `--skip-write` is False (implying it is the primary consumer actually writing to clickhouse) and keeping the other prefix if not.

Otherwise are are definitely going to ruin our existing metrics/dashboards/monitors

untitaker: I can't reuse our existing dashboard widgets to validate when the prefix is not the same

I think our dashboard widgets _should_ always filter by consumer group as well, I haven't checked everything but I was planning to fix it that way.

lynnagara: Ok. Keep in mind that our main consumer receive latency alert/SLO does not though. Are you planning to change that too? It's true that we're not emitting this particular metric from Rust right now but I expect this to change really soon.

untitaker: I didn't know that.. Ok, let me see if i can fix it tomorrow. If you think those are the only ones it seems manageable, I was definitely gonna just roll this out and see what breaks :D

lynnagara: Now i wonder if any python reference consumers we are running also causes this to be wrong 🤔

untitaker: I checked and our main receive latency SLO _does_ filter by consumer group. I think I may have added it when it went off because of the python reference consumer deployment, and then forgot about it.

untitaker:
I think this is ready to merge, unless we know of a concrete alert where this is broken, I think I'll just embrace the chaos (since i'm oncall anyway)

lynnagara (approved):

PR #5037, opened 2023-11-15
Move `SchemaValidator` into async task 

Swatinem:
Thus far, the schema validation was running in the "main" thread (the one calling `get_task`), and not on the thread pool actually executing the resulting Future.

Now, the schema validation itself will happen inside of the spawned Task.


untitaker (approved):
good catch

PR #5042, opened 2023-11-15
fix(referrer): Only use referrer passed in tenant_ids

enochtangg:
Currently, ST datadog is not displaying metrics with the proper tagged referrers. This is because in ST, the referrer (as part of the HTTP request) is not being sent. However, the referrer within the tenant_id (part of the body) is being sent. We enforce this in our allocation policies. This PR is responsible for only using the referrer being passed in the tenant_id. A runtime config `only_use_tenant_id_referrer` is used in case of unexpected behaviours.

### DIFF
--- a/tests/request/test_build_request.py
+++ b/tests/request/test_build_request.py
@@ -5,6 +5,7 @@ from typing import Any, Dict
 
 import pytest
 
+from snuba import state
 from snuba.datasets.entities.entity_key import EntityKey
 from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.factory import get_dataset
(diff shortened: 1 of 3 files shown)


tests/request/test_build_request.py
```diff
@@ -135,18 +138,47 @@ def test_build_request(body: Dict[str, Any], condition: Expression) -> None:
             "tenant_ids": {"organization_id": 1, "referrer": "test"},
         },
         {"organization_id": 1, "referrer": "test"},
+        "my_request",
+        0,
+        id="multiple projects, no project tenant",
```
volokluev, line 143: ids of the tests are the same, change it to make it accurate to what it is testing

volokluev (approved):
one small comment

PR #5043, opened 2023-11-15
meta: Make processing team codeowners

lynnagara:
This is likely temporary for the duration of the rust consumer project

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,6 +1,9 @@
 # Use owners-snuba by default
 * @getsentry/owners-snuba
 
+# rust_snuba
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,6 +1,9 @@
 # Use owners-snuba by default
 * @getsentry/owners-snuba
 
+# rust_snuba
+* @getsentry/owners-snuba @getsentry/owners-processing
```
untitaker, line 5: I don't think this is right, this makes them codeowners for the entire repo

lynnagara: oops, fixed

untitaker (approved):

PR #5044, opened 2023-11-15
feat(rust): Start buffering messages for DLQ

lynnagara:
This implements buffering of messages that are pending commit. Later, it can be used by the DLQ to retrieve untransformed messages which are deemed invalid and produced to the DLQ, (and logged to Sentry).

I'm adding this functionality now (even though rest of DLQ is not really done) because:
1. I think it might have some implications for how we've defined TPayload in a bunch of places (it probably needs to be Clone), and
2. It is likely to have performance implications that we should start understanding better now.
3. The first strategy must now take `Arc<TPayload>` instead of `TPayload`

Might need to bump requested memory a bit when we roll this out.

### DIFF
--- a/rust_snuba/rust_arroyo/examples/base_processor.rs
+++ b/rust_snuba/rust_arroyo/examples/base_processor.rs
@@ -11,8 +11,8 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 struct TestFactory {}
-impl ProcessingStrategyFactory<KafkaPayload> for TestFactory {
-    fn create(&self) -> Box<dyn ProcessingStrategy<KafkaPayload>> {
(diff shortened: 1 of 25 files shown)


untitaker (commented):
actually, arpad brought up a good point in slack... it's probably better and less invasive to change `KafkaPayload.payload` to be an `Option<Arc<[u8]>>` instead of adding the Arc everywhere else.

Swatinem:
That is still cloning the `key` and `headers`. But maybe moving the `Arc` internally in some way would be a good idea. not quite sure what we are cloning exactly in which places

lynnagara:
Superceded by https://github.com/getsentry/snuba/pull/5053

PR #5051, opened 2023-11-16
feat(rust): Buffering messages for DLQ - take 2

lynnagara:
Alternate implementation to https://github.com/getsentry/snuba/pull/5044 which uses Arc internally in KafkaPayload rather than forcing the outer strategy to be `Arc<KafkaPayload>`.

### DIFF
--- a/rust_snuba/rust_arroyo/src/backends/kafka/types.rs
+++ b/rust_snuba/rust_arroyo/src/backends/kafka/types.rs
@@ -1,8 +1,9 @@
 use rdkafka::message::OwnedHeaders;
+use std::sync::Arc;
 
 #[derive(Clone, Debug)]
 pub struct KafkaPayload {
-    pub key: Option<Vec<u8>>,
-    pub headers: Option<OwnedHeaders>,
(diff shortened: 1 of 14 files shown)


lynnagara:
Might try and make this 1 arc instead of 3

rust_snuba/rust_arroyo/src/backends/kafka/types.rs
```diff
@@ -1,8 +1,9 @@
 use rdkafka::message::OwnedHeaders;
+use std::sync::Arc;
 
 #[derive(Clone, Debug)]
 pub struct KafkaPayload {
-    pub key: Option<Vec<u8>>,
-    pub headers: Option<OwnedHeaders>,
-    pub payload: Option<Vec<u8>>,
+    pub key: Option<Arc<Vec<u8>>>,
```
Swatinem, line 6:
```suggestion
    pub key: Option<Arc<[u8]>>,
```
That would avoid one indirection

lynnagara:
Went with https://github.com/getsentry/snuba/pull/5053 instead

PR #5053, opened 2023-11-16
feat(rust): Buffered messages option 3

lynnagara:
Another take on https://github.com/getsentry/snuba/pull/5044
This one uses a single Arc inside KafkaPayload instead of one for
each of the key, headers and payload. Goal was to make it even
cheaper to clone than v2.


untitaker (approved):
i'm gonna approve this but if you wanna play around with having just one Arc, etc, feel free to. I just think for a first version this is ok and would probably be fine in prod

PR #5059, opened 2023-11-17
fix(generic-metrics): Add raw tags hash to sets dist table

enochtangg:
This column already exists on the local tables, adding to the sets dist tables.

Requires: https://github.com/getsentry/snuba/pull/5049


ayirr7 (approved):

ayirr7:
One note (that might already be on your radar): the Distributions dist table does not yet have the `_indexed_tags_hash` column present

PR #5060, opened 2023-11-17
Spans: Add hashmap col of spans measurements keys

dbanda:
Performance plans to begin querying measurements in the `measurements` nested column. This add hashmap column to `measurements`, similar to the one we have for `tags`. Unlike for tags however, this only hashes the keys

### DIFF
--- a/snuba/migrations/group_loader.py
+++ b/snuba/migrations/group_loader.py
@@ -352,6 +352,7 @@ class SpansLoader(DirectoryLoader):
             "0006_spans_add_profile_id",
             "0007_spans_add_metrics_summary",
             "0008_spans_add_index_on_span_id",
+            "0009_spans_add_measure_hashmap",
         ]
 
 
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/spans/0007_spans_add_measure_hashmap.py
```diff
@@ -0,0 +1,75 @@
...
+)
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    """
+    Adds measurements tags hash map column defined as Array(Int64)
+    Materialized with MEASUREMENTS_HASH_MAP_COLUMN expression.
+    This allows us to quickly find measure tag keys and we can add an
+    index on this column. Unlike sentry tags, we don't hash the
+    tag values since we don't need to query on them.
```
evanh, line 30: Why doesn't starfish want to filter based on the tag values?

nikhars: I am assuming you meant the measurements value here? If so, the measurements value is a float64 column and finding exact matches on float values is unnecessary. The use case right now is for finding spans with specific measurement keys. Also, most likely the use case is going to be finding measurement values with range checks for which the hash map column would not be useful.

nikhars (approved):

PR #5061, opened 2023-11-17
ci: Reduce Sentry test concurrency

lynnagara:
This is wasteful since a lot of time is spent just setting up the tests (which has to be done on every shard) and not actually running the test suite.

Sentry org is hitting max concurrency on GitHub so this aims to help in other projects too.


ayirr7:
Do we have a sense of how long Snuba CI takes now (before this change) so we have a point of comparison after this change?

ayirr7 (approved):

PR #5064, opened 2023-11-17
Revert "Revert "feat: Experiment using statsdproxy for aggregation (#…

lynnagara:
…4734)""

I reverted the original PR was reverted since it seems to cause a 6x reduction in throughput. Re-opening in case we want to try it again.

This reverts commit 201e07b09de421e3d9e6109a57e43d0b108abe52.


untitaker:
I think it's just this mutex causing trouble, and we'd probably be able to fix this by making the entire metrics client thread-local: https://github.com/getsentry/statsdproxy/blob/07b51e05107882226d94c27207bb968c1181e3d9/src/cadence.rs#L38

Swatinem (approved):

untitaker:
https://github.com/getsentry/statsdproxy/pull/38

PR #5071, opened 2023-11-20
feat(rust): Add snuba latency SLO metric

untitaker:


rust_snuba/src/processors/querylog.rs
```diff
@@ -20,9 +20,7 @@ pub fn process_message(
 
     let serialized = serde_json::to_vec(&querylog_msg)?;
 
-    Ok(BytesInsertBatch {
-        rows: vec![serialized],
-    })
+    Ok(BytesInsertBatch::from_rows(metadata.timestamp, vec![serialized]))
```
lynnagara, line 23: Since Snuba latency is calculated in a common way for all proessors, can we do this without touching the processor code or relying on the processor to pass the right timestamp into `BytesInsertBatch` here?

In the Python version, only the `recieved` timestamp, i.e. the original timestamp used to calculate end to end pipeline latency is done in message processor in this way. Can we do it the same way here?

untitaker: I'm porting this: https://github.com/getsentry/snuba/blob/5c40e08874cb3a3174d50c71f565adaae56b70cc/snuba/consumers/consumer.py#L190-L191

which uses the kafka message timestamp

the other metrics will follow, but i think this one is easiest to start with

> Since Snuba latency is calculated in a common way for all proessors, can we do this without touching the processor code or relying on the processor to pass the right timestamp into BytesInsertBatch here?

I can try that, yeah

PR #5076, opened 2023-11-21
fix(rust): Fix the ProduceCommitLog strategy

lynnagara:
The ProduceCommitLog should get a message with type `BytesInsertBatch` from the prior processing step, and produces one message for each partition contained within that batch. The prior version where it got a `KafkaPayload` and produced a single message was wrong.

In order to achieve this, this splits the `BytesInsertBatch` into two separate structs. Now we have `RowData` which we require the message processors to return, as well as `BytesInsertBatch` which is now internal to Snuba and contains other metadata that we will add for producing to the commit log.


untitaker (approved):

untitaker:
I have a similar refactor in #5071 but i think i'll revert it then for this one

PR #5083, opened 2023-11-21
fix(replacegroup): Catch bad datetime formats in ReplaceGroup

evanh:
A bad message was sent to ReplaceGroup that used the `+00:00` format instead of
`.000Z` format. In order to continue processing, handle that edge case just for
the ReplaceGroup class. As a follow up the emitter of the bad message will be
fixed.

### DIFF
--- a/snuba/replacers/errors_replacer.py
+++ b/snuba/replacers/errors_replacer.py
@@ -195,7 +195,6 @@ class ErrorsReplacer(ReplacerProcessor[Replacement]):
         ):
             return None
         elif type_ in _REPLACEMENT_BY_TYPE:
-
             processed = _REPLACEMENT_BY_TYPE[type_].parse_message(
                 message,
                 self.__replacement_context,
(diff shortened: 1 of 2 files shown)


snuba/replacers/errors_replacer.py
```diff
@@ -263,6 +262,10 @@ def get_timestamp_condition(msg_value: Optional[str], operator: str) -> str:
         if not msg_value:
             return ""
 
+        if "+" in msg_value:  # e.g. "2023-08-28T03:05:38+00:00"
```
volokluev, line 265: I would feel more comfortable if we tried to parse the timestamp with the other format instead and then re-serialize it into what we want. This is a little too hacky for my taste

snuba/replacers/errors_replacer.py
```diff
@@ -263,6 +262,10 @@ def get_timestamp_condition(msg_value: Optional[str], operator: str) -> str:
...
+            msg_value, _ = msg_value.rsplit("+", 1)
+            msg_value += ".000Z"
```
lynnagara: how about:
```suggestion
            datetime.datetime.strftime(datetime.datetime.strptime(msg_value, other_format), settings.PAYLOAD_DATETIME_FORMAT)
```

where:
`other_format = "%Y-%m-%dT%H:%M:%S.%f%z"`

We do have users running this in dev, self hosted where it wont work if it's not UTC

lynnagara (approved):

PR #5085, opened 2023-11-21
fix: make sure SNUBA_RELEASE is set

MeredithAnya:
The events in sentry for the snuba project are no longer being tagged with a release, I think that it _might_ have been due to https://github.com/getsentry/snuba/pull/4551/ since changes to the Dockerfile were made.

I'm moving the ENV chunk before running the `api` since I'm assuming that's why the `SNUBA_RELEASE` isn't being picked up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,6 +115,16 @@ RUN set -ex; \
     pip install -e .; \
     snuba --help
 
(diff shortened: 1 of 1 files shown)


untitaker (approved):
makes sense to me

PR #5088, opened 2023-11-22
fix(rust-python): Join handles less frequently, fix backpressure bug

untitaker:
max_queue_depth does not work as it is implemented right now, but now
that we expose it as a parameter, we might as well use it to determine
how many handles are supposed to be in-flight.

The original theory was correct though: We are blocking too much in
check_for_results, and that's the culprit for our performance problems.

Now the next performance problem is the serialization overhead for
individual messages between processes, and we might have to implement
batching to fix that. But at the very least the processor can now
saturate 5 cores on my machine.

### DIFF
--- a/rust_snuba/src/strategies/python.rs
+++ b/rust_snuba/src/strategies/python.rs
@@ -81,56 +81,52 @@ impl PythonTransformStep {
     }
 
     fn check_for_results(&mut self) {
-        // procspawn has no join() timeout that does not consume the handle on timeout.
-        //
-        // Additionally we have observed, at least on MacOS, that procspawn's active_count() only
-        // decreases when the handle is consumed. Therefore our count of actually saturated
-        // processes is `self.processing_pool.active_count() - self.handles.len()`.
-        //
-        // If no process is saturated (i.e. above equation is <= 0), we can conclude that all tasks
-        // are done and all handles can be joined and consumed without waiting.
-        while {
-            let active_count = self
-                .processing_pool
-                .as_ref()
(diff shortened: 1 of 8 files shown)


rust_snuba/src/strategies/python.rs
```diff
@@ -81,82 +81,84 @@ impl PythonTransformStep {
...
+                tracing::error!(error, "Invalid message");
+            }
         }
     }
 }
 
 impl ProcessingStrategy<KafkaPayload> for PythonTransformStep {
     fn poll(&mut self) -> Result<Option<CommitRequest>, InvalidMessage> {
-        self.check_for_results();
+        if self.handles.len() > self.max_queue_depth {
```
lynnagara, line 137: Doesn't this mean we are *never* checking for results if we never hit max queue depth? How would that work for low volume topics/environments?

untitaker: ah you're right. i was relying on `join()` to eventually flush things out, but this doesn't work. how about a parameter called `max_queue_latency`?

lynnagara: I think that approach sounds good. Wdyt about not even making it a parameter and just hardcoding 1 second or something? I worry that having too many parameters for stuff makes it confusing/difficult for users to pick good values and it seems that a constant value here would be just fine at any volume really.

untitaker: done. i set it to 10 seconds because otherwise it has a lot of impact on throughput

lynnagara (approved):

PR #5091, opened 2023-11-22
fix: Skip warnings in gocd error checks

evanh:
The rate limit warnings that go to Sentry was causing this check to trip up and
prevent the deploy pipeline from working. Skip warning error events when
running this check.

Depends on https://github.com/getsentry/devinfra-deployment-service/pull/503


volokluev (approved):

IanWoodard (approved):
We should be good to go

PR #5093, opened 2023-11-22
feat(rust): Pass the DLQ policy to stream processor

lynnagara:
This change starts wiring up the DLQ policy and producer.

- The StreamProcessor now takes an optional dlq_policy parameter, which defines the policy to be applied to invalid messages.

- The dlq_policy is applied for consumers/topics where a value for `dlq_topic` is defined (this comes from storage .yaml files), otherwise it is None and no messages will be DLQ'ed.

- The policy is not being used by the StreamProcessor yet

### DIFF
--- a/rust_snuba/src/consumer.rs
+++ b/rust_snuba/src/consumer.rs
@@ -122,27 +122,24 @@ pub fn consumer_impl(
     let consumer = Arc::new(Mutex::new(KafkaConsumer::new(config)));
     let logical_topic_name = consumer_config.raw_topic.logical_topic_name;
 
-    let dlq_policy = match consumer_config.dlq_topic {
-        None => None,
-        Some(dlq_topic_config) => {
-            let producer_config =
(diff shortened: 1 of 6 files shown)


untitaker (approved):

rust_snuba/src/consumer.rs
```diff
@@ -119,6 +121,29 @@ pub fn consumer_impl(
 
     let consumer = Arc::new(Mutex::new(KafkaConsumer::new(config)));
     let logical_topic_name = consumer_config.raw_topic.logical_topic_name;
+
+    let dlq_policy = match consumer_config.dlq_topic {
+        None => None,
+        Some(dlq_topic_config) => {
```
Swatinem, line 127:
```suggestion
    let dlq_policy = consumer_config.dlq_topic.map(|dlq_topic_config| {
```

Swatinem (approved):

PR #5105, opened 2023-11-27
fix(rust): Fix processing time metric

lynnagara:
Processing time was understated since we were not counting the time spent in strategy.poll() which is where a lot of consumers do processing work. This is now aligned with the Python Arroyo implementation.

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/mod.rs
+++ b/rust_snuba/rust_arroyo/src/processing/mod.rs
@@ -191,6 +191,9 @@ impl<TPayload: Clone + 'static> StreamProcessor<TPayload> {
                     return Err(RunError::Poll(error));
                 }
             }
+
+            self.metrics_buffer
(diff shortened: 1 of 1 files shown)


rust_snuba/rust_arroyo/src/processing/mod.rs
```diff
@@ -204,6 +207,8 @@ impl<TPayload: Clone + 'static> StreamProcessor<TPayload> {
                 Some(_) => return Err(RunError::InvalidState),
             }
         };
+
+        let poll_start = Instant::now();
```
untitaker, line 207: isnt this just an unused variable now?

lynnagara: sorry, put the metric in the wrong place earlier. pushed a fix

untitaker (approved):

PR #5107, opened 2023-11-27
fix(rust-python): Handle none payload

lynnagara:
None payload is valid in kafka, make sure we do not panic

### DIFF
--- a/rust_snuba/src/strategies/python.rs
+++ b/rust_snuba/src/strategies/python.rs
@@ -222,8 +222,7 @@ impl ProcessingStrategy<KafkaPayload> for PythonTransformStep {
                 partition,
                 timestamp,
             }) => {
-                // TODO: Handle None payload
-                let payload_bytes = (payload.payload().unwrap()).clone();
+                let payload_bytes = (payload.payload().unwrap_or_default()).clone();
 
                 let args = (payload_bytes, offset, partition.index, timestamp);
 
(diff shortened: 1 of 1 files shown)


rust_snuba/src/strategies/python.rs
```diff
@@ -222,8 +222,7 @@ impl ProcessingStrategy<KafkaPayload> for PythonTransformStep {
                 partition,
                 timestamp,
             }) => {
-                // TODO: Handle None payload
-                let payload_bytes = (payload.payload().unwrap()).clone();
+                let payload_bytes = payload.payload().cloned().unwrap_or_default();
```
Swatinem, line 225: is there a way to early-return rather than to pass an empty `Vec` through python which will then either early-return or error anyway?

Swatinem:
On a related note, it would be nice to also clarify what the intended behavior is in these edge-cases. It would be good to just filter these out as early as possible and then have properly validated payloads down the line.

untitaker:
what is `None` actually representing?

untitaker (approved):
what is `None` actually used for

rust_snuba/src/strategies/python.rs
lynnagara: We could early return here, but I'm not sure it's worth adding this optimization -- I think it's easier to let the Python code deal with it like any other invalid message and let the message be DLQ'ed and it's offsets committed as normal. It's also technically valid so if I guess someone wanted to write a Python processor with some headers/key and no payload then this way lets them do it (even though that doesn't necessarily make much sense in this particular scenario).

lynnagara:
> what is `None` actually representing?

Kafka payload is allowed to be null (same as keys or headers). This handles the case that a producer has produced a message with no value. Though it's not a pattern we use in Snuba (we put all important information in the payload and not headers or keys), it's valid.

PR #5108, opened 2023-11-27
feat(gocd): Add `de` region deployments

rgibert:
Bump `gocd-jsonnet` release to include `de` deployments.


onkar:
@rgibert is this PR still relevant? If not, can you close it?

PR #5118, opened 2023-11-29
feat(rust): Implement base logic for SLO metrics

untitaker:
Implement all the logic necessary for recording SLOs. What is missing is returning the right numbers in the Rust message processor.

SNS-2519

### DIFF
--- a/rust_snuba/src/types.rs
+++ b/rust_snuba/src/types.rs
@@ -33,6 +33,10 @@ impl LatencyRecorder {
     }
 
     fn send_metric(&self, metrics: &BoxMetrics, write_time: DateTime<Utc>, metric_name: &str) {
+        if self.num_values == 0 {
+            return;
+        }
+
         let into_latency = |ts: DateTime<Utc>| (write_time - ts).num_seconds().try_into().ok();
 
         if let Some(ts) = DateTime::from_timestamp(self.max_secs, 0).and_then(into_latency) {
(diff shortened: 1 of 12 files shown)


lynnagara (approved):

rust_snuba/src/types.rs
```diff
@@ -7,26 +7,116 @@ use std::collections::BTreeMap;
...
+    }
+
+    fn send_metric(&self, metrics: &BoxMetrics, write_time: DateTime<Utc>, metric_name: &str) {
+        if self.num_values == 0 {
+            return;
+        }
+
+        let into_latency = |ts: DateTime<Utc>| (write_time - ts).num_seconds().try_into().ok();
+
+        if let Some(ts) = DateTime::from_timestamp(self.max_secs, 0).and_then(into_latency) {
```
Swatinem, line 42: converting to `DateTime`, just to take the difference as seconds (is that `f64`?) is a bit weird. Why not convert `write_time` to seconds first and then just do straight up math?

Also, do you really want to explicitly log overflows here, or just do saturating math?

untitaker: > is that f64?

yes, i need a type where the sum of multiple timestamps will fit. u64 will overflow easily, and in theory with sums and averages I don't really need that much precision. i could also use u128 i believe

i'll rewrite the function, i agree with your other points

Swatinem (approved):

PR #5120, opened 2023-11-29
fix(joins) Push granularity down to the subqueries

evanh:
The granularity was not being pushed down to the subqueries created in a
CompositeQuery join. That meant that the TimeSeriesProcessor (an entity level
processor) didn't have access to the granularity and wasn't being executed
correctly.

Push the granularity down to all the subqueries so they can be properly
handled.

### DIFF
--- a/snuba/query/joins/subquery_generator.py
+++ b/snuba/query/joins/subquery_generator.py
@@ -1,5 +1,7 @@
+from __future__ import annotations
+
 from dataclasses import replace
-from typing import Generator, List, Mapping, Set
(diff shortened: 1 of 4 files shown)


volokluev (commented):
Thanks for doing this. Is it possible for subqueries to set different granularities? What happens in that case?

evanh:
> Is it possible for subqueries to set different granularities?

No, because the granularity is set at the upper query level in SnQL. We don't allow queries to specify a granularity for each Entity in the JOIN.

PR #5131, opened 2023-12-01
feat(rust): Produce invalid messages to DLQ

lynnagara:
This adds the basic DLQ implementation. Invalid messages are automatically produced to the DLQ if a topic is defined for the storage.

The feature is not enabled unless `skip_write` is set to False, as we should not be producing any data to actual topics in prod otherwise.

There are a couple of features that are not yet implemented yet and will need to be done as a followup:
- DLQ limits are currently ignored and still need to be wired up
- The DLQ producer should be flushed prior to committing offsets in the revocation callback

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/dlq.rs
+++ b/rust_snuba/rust_arroyo/src/processing/dlq.rs
@@ -184,8 +184,7 @@ impl<TPayload: Clone + Send + Sync + 'static> DlqPolicyWrapper<TPayload> {
     // they are safe to commit.
     pub fn flush(&mut self, committable: HashMap<Partition, u64>) {
         for (p, values) in self.futures.iter_mut() {
-            while !values.is_empty() {
-                let (offset, future) = &mut values[0];
+            while let Some((offset, future)) = values.front_mut() {
                 if let Some(committable_offset) = committable.get(p) {
                     // The committable offset is message's offset + 1
                     if *committable_offset > *offset {
--- a/rust_snuba/rust_arroyo/src/processing/dlq.rs
+++ b/rust_snuba/rust_arroyo/src/processing/dlq.rs
@@ -183,11 +183,11 @@ impl<TPayload: Clone + Send + Sync + 'static> DlqPolicyWrapper<TPayload> {
     // Blocks until all messages up to the committable have been produced so
     // they are safe to commit.
     pub fn flush(&mut self, committable: HashMap<Partition, u64>) {
-        for (p, values) in self.futures.iter_mut() {
-            while let Some((offset, future)) = values.front_mut() {
-                if let Some(committable_offset) = committable.get(p) {
+        for (p, committable_offset) in committable {
+            if let Some(values) = self.futures.get_mut(&p) {
+                if let Some((offset, future)) = values.front_mut() {
                     // The committable offset is message's offset + 1
-                    if *committable_offset > *offset {
+                    if committable_offset > *offset {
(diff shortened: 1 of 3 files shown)


rust_snuba/rust_arroyo/src/processing/dlq.rs
```diff
@@ -121,6 +126,84 @@ impl<TPayload> DlqPolicy<TPayload> {
...
+    max_pending_futures: usize,
+    futures: BTreeMap<Partition, Futures<TPayload>>,
+}
+
+impl<TPayload: Clone + Send + Sync + 'static> DlqPolicyWrapper<TPayload> {
+    pub fn new(dlq_policy: Option<DlqPolicy<TPayload>>) -> Self {
+        let concurrency_config = ConcurrencyConfig::new(10);
+        DlqPolicyWrapper {
+            dlq_policy,
+            runtime: concurrency_config.handle(),
```
Swatinem, line 145: this will start a new runtime. Better to take the `ConcurrencyConfig` from the outside.

rust_snuba/rust_arroyo/src/processing/dlq.rs
```diff
@@ -121,6 +126,84 @@ impl<TPayload> DlqPolicy<TPayload> {
...
+                .push_back((message.offset, handle));
+        }
+    }
+
+    // Blocks until all messages up to the committable have been produced so
+    // they are safe to commit.
+    pub fn flush(&mut self, committable: HashMap<Partition, u64>) {
+        for (p, values) in self.futures.iter_mut() {
+            while !values.is_empty() {
+                let (offset, future) = &mut values[0];
```
Swatinem, line 188:
```suggestion
            while let Some((offset, future)) = values.front_mut() {
```

rust_snuba/rust_arroyo/src/processing/dlq.rs
```diff
@@ -121,6 +126,84 @@ impl<TPayload> DlqPolicy<TPayload> {
...
+                if let Some(committable_offset) = committable.get(p) {
```
Swatinem, line 189: I would actually turn this around, and iterate over `committable` outside, and then pick the inner `self.futures` to join. As the futures are ordered by offset, you can then break out of that loop early.

rust_snuba/rust_arroyo/src/processing/dlq.rs
lynnagara: Is there a performance downside to starting a new runtime that does not actually do any work (unless we are getting invalid messages in which case performance is not our primary concern as something is broken)? I did it this way because I think it's simpler for the user to just not think about what config they need here and pass it to the stream processor, and it seems like it wouldn't impact perf under normal operation. Unlike the other examples where we share the config this isn't part of the user defined processing strategy.

rust_snuba/rust_arroyo/src/processing/dlq.rs
lynnagara: Checking `values.len()` inside the loop no longer works this way as `values` is borrowed

lynnagara: Actually, no. That's the other loop. Fixed this one.

rust_snuba/rust_arroyo/src/processing/dlq.rs
lynnagara: updated

PR #5152, opened 2023-12-04
feat(spans): Add a metrics summaries dataset

phacops:
For DDM, we'd like to add a way to provide examples of spans which generated some metrics. The idea is, when we're visualizing a DDM metric, the user would be able to select a square on the chart and we'd show them a list of spans associated with those values. This would add a new dataset called `metrics_summaries` in order to correlate DDM metrics with spans.

In order to correlate DDM metrics and spans, [RFC 123](https://github.com/getsentry/rfcs/blob/main/text/0123-metrics-correlation.md) adds a metrics summary to spans. The relationship between spans and summaries is like this:
```
1 span -> M metrics -> S metrics summaries -> T tags
```

To display metrics information when querying spans, we'd like to add a column called `metrics_summary` to the `spans` dataset to purely **store** the needed information. When rendering those spans, we'd be able to show information about the metrics and the different values.

In order to do the reverse, showing spans related to some specific metrics, we'd need a way to filter through these spans' metrics summaries. Given the multi-dimensional nature of metrics summaries (we could have many summaries attached to 1 metric attached to 1 span), using the same way we do to filter on `tags` is not possible without flattening the tags as well and it's less efficient to fully filter that way.
One efficient way to do this is have a separate dataset. We'd store enough information about the spans so we don't have to join (in ClickHouse or in the app). We could keep tags as an array column or even insert 1 row per tag value. This way, selecting a list of spans corresponding to certain values for given metric would be a simple `SELECT` query away.

### DIFF
--- a/snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
+++ b/snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
@@ -7,9 +7,9 @@ from snuba.migrations.columns import MigrationModifiers as Modifiers
 from snuba.migrations.operations import OperationTarget, SqlOperation
 from snuba.utils.schemas import Float
 
-storage_set_name = StorageSetKey.SPANS
-local_table_name = "spans_local"
-dist_table_name = "spans_dist"
+storage_set_name = StorageSetKey.METRICS_SUMMARIES
+local_table_name = "metrics_summaries_local"
+dist_table_name = "metrics_summaries_dist"
 
 UNKNOWN_SPAN_STATUS = 2
 
--- a/snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
+++ b/snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
@@ -14,23 +14,28 @@ dist_table_name = "spans_dist"
 UNKNOWN_SPAN_STATUS = 2
 
 columns: List[Column[Modifiers]] = [
+    # ids
     Column("project_id", UInt(64)),
     Column("trace_id", UUID()),
     Column("span_id", UInt(64)),
     Column("segment_id", UInt(64)),
-    Column("start_timestamp", DateTime()),
-    Column("exclusive_time", Float(64)),
-    Column("op", String(Modifiers(low_cardinality=True))),
-    Column("group", UInt(64)),
-    Column("tags", Nested([("key", String()), ("value", String())])),
-    Column("partition", UInt(16)),
-    Column("offset", UInt(64)),
-    Column("retention_days", UInt(16)),
-    Column("deleted", UInt(8)),
+    # metrics summary
     Column("min", Float(64)),
     Column("max", Float(64)),
     Column("sum", Float(64)),
     Column("count", Float(64)),
+    # metrics metadata
+    Column("tags", Nested([("key", String()), ("value", String())])),
+    # span metadata
+    Column("start_timestamp", DateTime()),
+    Column("exclusive_time", Float(64)),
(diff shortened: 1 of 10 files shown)


evanh:
Some high level questions:

- Have you done any analysis on the size/type of data you'll be storing in this table?
- How are you measuring the success of this? How do you know what part of this solution is working/not working?
- Is there any documentation of the decisions that lead to this particular solution?
- Is there some documentation on what the product use cases of this are?
    - On a related note, I'm not sure I understand the utility of seeing the metrics associated with a specific span. Metrics are generally useful in aggregate, and the metrics attached to one singular span would basically be raw values.
    - I think the more complex problem is actually the second one you mentioned, going from metrics -> spans. That will require a completely different table layout than what you propose here.

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
```diff
@@ -0,0 +1,86 @@
...
+from snuba.migrations.operations import OperationTarget, SqlOperation
+from snuba.utils.schemas import Float
+
+storage_set_name = StorageSetKey.METRICS_SUMMARIES
+local_table_name = "metrics_summaries_local"
+dist_table_name = "metrics_summaries_dist"
+
+UNKNOWN_SPAN_STATUS = 2
+
+columns: List[Column[Modifiers]] = [
```
evanh, line 16: This is missing any connection to our actual metrics dataset (metric_id, use case, mri etc.) Is that intended?

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
```diff
@@ -0,0 +1,86 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, group, cityHash64(span_id))",
```
evanh, line 52: Is every query to this table going to have a `group` and `span_id`?

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
```diff
@@ -0,0 +1,86 @@
...
+    Column("max", Float(64)),
+    Column("sum", Float(64)),
+    Column("count", Float(64)),
+    # metrics metadata
+    Column("tags", Nested([("key", String()), ("value", String())])),
+    # span metadata
+    Column("start_timestamp", DateTime()),
+    Column("exclusive_time", Float(64)),
+    Column("op", String(Modifiers(low_cardinality=True))),
+    Column("group", UInt(64)),
```
evanh, line 33: What is `group` in this context?

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
phacops: No, that was a mistake.

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
phacops: `group` is one way to select span examples. We might want to target exact spans producing those specific metric values or spans looking like them.

We might be able to not have that and just focus on returning span IDs.

phacops:
As asked by @evanh , I split this PR:
- https://github.com/getsentry/snuba/pull/5159
- https://github.com/getsentry/snuba/pull/5160
- https://github.com/getsentry/snuba/pull/5161

PR #5159, opened 2023-12-05
feat(spans): Add a metrics summary column to link spans with DDM metrics

phacops:

This is meant to show a metrics summary when we're showing span details.

### DIFF
--- a/snuba/snuba_migrations/spans/0007_spans_add_metrics_summary.py
+++ b/snuba/snuba_migrations/spans/0007_spans_add_metrics_summary.py
@@ -1,0 +1,69 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, String
+from snuba.clusters.storage_sets import StorageSetKey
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/spans/0007_spans_add_metrics_summary.py
```diff
@@ -0,0 +1,69 @@
...
+
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                column=Column(
+                    "metrics_summary",
```
nikhars, line 27: Ideally, since you don't want to search on these in the dataset, but just visualize them in a trace view, this column is good candidates for putting into a place like nodestore (similar to contexts). But since that is not available now, this should be ok.

nikhars (approved):

PR #5160, opened 2023-12-05
feat(spans): Add a metrics_summaries table

phacops:


snuba/migrations/groups.py
```diff
@@ -162,6 +165,11 @@ def __init__(
         storage_sets_keys={StorageSetKey.GROUP_ATTRIBUTES},
         readiness_state=ReadinessState.PARTIAL,
     ),
+    MigrationGroup.METRICS_SUMMARIES: _MigrationGroup(
+        loader=MetricsSummariesLoader(),
+        storage_sets_keys={StorageSetKey.METRICS_SUMMARIES},
+        readiness_state=ReadinessState.PARTIAL,
```
enochtangg, line 171: One thing to note, if we're setting this new migration group to `partial` immediately, this requires the storage_key for metrics summary to be mapped to an existing clickhouse cluster in both SaaS and S4S. We should check that is true before this gets merged.

nikhars: Good point. On a similar note, do you know if there is a migration state which enables migrations only on S4S but not SaaS? I believe that is what would be needed here to begin with.

evanh: Enoch mentioned this offline, but we don't have a state that is specific to S4S yet. We'll have to add that before this can be merged.

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
```diff
@@ -0,0 +1,79 @@
...
+local_table_name = "metrics_summaries_local"
+dist_table_name = "metrics_summaries_dist"
+
+columns: List[Column[Modifiers]] = [
+    # ids
+    Column("project_id", UInt(64)),
+    Column("span_id", UInt(64)),
+    Column("trace_id", UUID()),
+    # metrics summary
+    Column("metric_name", String(Modifiers(low_cardinality=True))),
```
evanh, line 20: I would call this `metric_mri` to be clear about exactly what is stored here. Otherwise it might get confused with `public_name`.

evanh:
Have you done any simple math estimates about how big this table will be? Like how many rows will we store just for the Sentry project?

snuba/migrations/groups.py
evanh: https://github.com/getsentry/snuba/pull/5172

snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py
```diff
@@ -0,0 +1,79 @@
...
+    Column("min", Float(64)),
+    Column("max", Float(64)),
+    Column("sum", Float(64)),
+    Column("count", Float(64)),
```
nikhars, line 24: Nit: Not sure under what circumstances can count be a float. If it can't be a float then you can probably change it to int. Takes lesser space to store.

nikhars (commented):
Is it safe to assume that you want to deploy this dataset on the same cluster as the spans dataset is on in S4S?

snuba/migrations/groups.py
nikhars: Based on what I understand, is the next step adding the new state here in S4S https://github.com/getsentry/ops/blob/b484e0f8d5f41f9ecab61997d70de9530f3fd44b/single-tenant/k8s/customers/s4s.yaml#L403-L406?

nikhars: I see experimental is already added in s4s. Thanks @evanh

nikhars:
> Is it safe to assume that you want to deploy this dataset on the same cluster as the spans dataset is on in S4S? If yes, we would need to add the mapping of the new storage set to cluster in s4s, something similar to https://github.com/getsentry/ops/blame/b484e0f8d5f41f9ecab61997d70de9530f3fd44b/single-tenant/k8s/customers/s4s.yaml#L400

PR is here https://github.com/getsentry/ops/pull/8681

phacops:
> Have you done any simple math estimates about how big this table will be? Like how many rows will we store just for the Sentry project?

I'm not sure how this table will be. We don't really have an idea of usage really and part of the experiment is to understand how much storage will be required.

nikhars (approved):

PR #5161, opened 2023-12-05
feat(spans): Add a metrics summary processor

phacops:

### DIFF
--- a/snuba/datasets/processors/metrics_summaries_processor.py
+++ b/snuba/datasets/processors/metrics_summaries_processor.py
@@ -1,0 +1,120 @@
+import time
+import uuid
+from datetime import datetime, timezone
+from typing import Any, Mapping, MutableMapping, MutableSequence, Optional, Tuple
+
+import structlog
+
+from snuba import environment
+from snuba.consumers.types import KafkaMessageMetadata
+from snuba.datasets.events_format import (
+    EventTooOld,
+    enforce_retention,
+    extract_extra_tags,
+)
+from snuba.datasets.processors import DatasetMessageProcessor
+from snuba.processor import (
+    InsertBatch,
+    ProcessedMessage,
+    _as_dict_safe,
+    _ensure_valid_date,
+)
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+logger = structlog.get_logger(__name__)
+
+metrics = MetricsWrapper(environment.metrics, "metrics_summaries.processor")
+
+MetricsSummaryEvent = MutableMapping[str, Any]
+RetentionDays = int
+
+
+class MetricsSummariesMessageProcessor(DatasetMessageProcessor):
+    """
+    Message processor for writing metrics summary data to the metrics_summaries table.
+    """
+
+    def __extract_timestamp(self, timestamp_ms: int) -> Tuple[int, int]:
+        # We are purposely using a naive datetime here to work with the rest of the codebase.
+        # We can be confident that clients are only sending UTC dates.
+        timestamp_sec = timestamp_ms / 1000
+        if _ensure_valid_date(datetime.utcfromtimestamp(timestamp_sec)) is None:
+            timestamp_sec = int(time.time())
+        return int(timestamp_sec), int(timestamp_ms % 1000)
+
+    @staticmethod
+    def _structure_and_validate_message(
+        message: MetricsSummaryEvent,
+    ) -> Optional[Tuple[MetricsSummaryEvent, RetentionDays]]:
+        try:
+            # We are purposely using a naive datetime here to work with the
+            # rest of the codebase. We can be confident that clients are only
+            # sending UTC dates.
+            retention_days = enforce_retention(
+                message.get("retention_days"),
+                datetime.utcfromtimestamp(message["start_timestamp_ms"] / 1000),
+            )
+        except EventTooOld:
+            return None
+
+        return message, retention_days
+
+    def _process_metrics_summary_event(
+        self,
+        processed: MutableMapping[str, Any],
+        metrics_summary_event: MetricsSummaryEvent,
+    ) -> None:
+        processed["trace_id"] = str(uuid.UUID(metrics_summary_event["trace_id"]))
+        processed["span_id"] = int(metrics_summary_event["span_id"], 16)
+        processed["segment_id"] = int(metrics_summary_event["segment_id"], 16)
+        processed["project_id"] = metrics_summary_event["project_id"]
+        processed["metric_id"] = metrics_summary_event["metric_id"]
+
+        processed["start_timestamp"], processed["start_ms"] = self.__extract_timestamp(
+            metrics_summary_event["start_timestamp_ms"],
+        )
+        processed["exclusive_time"] = float(metrics_summary_event["exclusive_time_ms"])
+        processed["op"] = metrics_summary_event["op"]
+        processed["group"] = metrics_summary_event["group"]
+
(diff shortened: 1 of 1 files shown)


snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
```diff
@@ -0,0 +1,84 @@
...
+            to_nested_col_table: null
+            to_nested_col_name: tags
+            value_subcolumn_name: value
+            nullable: false
+
+storage_selector:
+  selector: DefaultQueryStorageSelector
+
+query_processors:
+  - processor: TimeSeriesProcessor
```
evanh, line 58: I would love for this processor to die. It is mainly syntactic sugar around the `GRANULARITY` clause of SnQL. If you specify a `GRANULARITY 3600` and reference the magic column `time`, it converts it to `toStartOfHour(timestamp)` or whatever. It causes confusion with people all the time and really provides little value.

If you really want it, then you can leave this in, but if you don't think you'll need it I would ask it be removed. Everyone keeps copy and pasting it from other entities.

snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
```diff
@@ -0,0 +1,84 @@
...
+    args:
+      project_column: project_id
+  - processor: ResourceQuotaProcessor
+    args:
+      project_field: project_id
+
+validators:
+  - validator: EntityRequiredColumnValidator
+    args:
+      required_filter_columns: [project_id]
```
evanh, line 79: Should this have `metric_mri` (see my note on your other PR) as well?

snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
```diff
@@ -0,0 +1,101 @@
...
+  key: metrics_summaries
+  set_key: metrics_summaries
+
+readiness_state: partial
+
+schema:
+  columns:
+    [
+      { name: project_id, type: UInt, args: { size: 64 } },
+      { name: metric_id, type: String },
```
evanh, line 15:
```suggestion
      { name: metric_mri, type: String },
```

snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
```diff
@@ -0,0 +1,84 @@
+version: v1
+kind: entity
+name: metrics_summaries
+
+schema:
+  [
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: span_id, type: UInt, args: { size: 64 } },
+    { name: trace_id, type: UUID },
+    { name: metric_name, type: String },
```
evanh, line 10:
```suggestion
    { name: metric_mri, type: String },
```

snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
```diff
@@ -0,0 +1,101 @@
...
+      hash_map_name: _tags_hash_map
+      killswitch: tags_hash_map_enabled
+  - processor: EmptyTagConditionProcessor
+  - processor: ArrayJoinKeyValueOptimizer
+    args:
+      column_name: tags
+  - processor: PrewhereProcessor
+    args:
+      prewhere_candidates:
+        [trace_id, span_id]
```
evanh, line 87: Why these choices?

snuba/datasets/processors/metrics_summaries_processor.py
```diff
@@ -0,0 +1,117 @@
...
+        common_fields["tags.key"], common_fields["tags.value"] = extract_extra_tags(
+            tags
+        )
+
+        processed_rows: MetricsSummaries = []
+
+        for metric_name, metric_values in span_event.get("metrics_summary", {}).items():
+            for metric_value in metric_values:
+                processed = deepcopy(common_fields)
+                processed["metric_id"] = metric_name
```
evanh, line 87:
```suggestion
                processed["metric_mri"] = metric_name
```

snuba/datasets/processors/metrics_summaries_processor.py
```diff
@@ -0,0 +1,117 @@
...
+            "span_id": int(span_event["span_id"], 16),
+            "trace_id": str(uuid.UUID(span_event["trace_id"])),
+        }
+
+        tags: Mapping[str, Any] = _as_dict_safe(span_event.get("tags", None))
+        common_fields["tags.key"], common_fields["tags.value"] = extract_extra_tags(
+            tags
+        )
+
+        processed_rows: MetricsSummaries = []
```
evanh, line 82: When do you append to this?

evanh (commented):
Please add some basic smoke tests, both for the consumer and for querying the entity.

snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
phacops: I removed it, this was definitely a copy/paste from another configuration.

snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
phacops: Carry over from other configuration. Since we'll query with `project_id` and `metric_mri` all the time, I suppose these are the values I need here?

snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
phacops: We can't have this as a required column because the validator will force the type of the column to be an `int` and `metric_mri` is a `str`.

https://github.com/getsentry/snuba/blob/master/snuba/query/validation/validators.py#L104

snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
evanh: Thanks!

snuba/datasets/configuration/spans/entities/metrics_summaries.yaml
evanh: Ah fair enough. OK good to know.

snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
evanh: Yes those are values I would do, as long as you're confident those will always be filtered in the query.

snuba/datasets/processors/metrics_summaries_processor.py
```diff
@@ -0,0 +1,128 @@
...
+            return None
+        try:
+            processed_rows = self._process_span_event(span_event, retention_days)
+        except Exception:
+            metrics.increment("message_processing_error")
+            return None
+
+        received = (
+            datetime.fromtimestamp(span_event["received"], tz=timezone.utc)
+            if "received" in span_event
```
evanh, line 125: @lynnagara @untitaker Is it valid to leave `origin_timestamp` as `None`? I can see that the class marks the value as optional, but is that something that we actually want?

untitaker: it is valid, and we already have it like that in some datasets. i would not say it's something we want, i would love to see that `received` attribute being required in the message

untitaker: also we don't have any datasets where this is only None *some of the time*, we probably end up having somewhat misleading metrics in this case because we're skipping over datapoints. but i think that's sort of up to pierre if he can live with that.

PR #5162, opened 2023-12-05
fix(spans): Add sentry_tags to spans entity

nikhars:
Added the sentry_tags column to the spans entity. Also added the hash map columns and subscriptable mappers similar to tags column.


evanh (approved):
Why is this necessary?

nikhars:
> Why is this necessary?

The spans data model has 2 sets of tags
1. Customer sent tags.
2. Sentry extracted tags.

We are duplicating the sentry extracted tags into customer sent tags because product wasn’t aware of this distinction. But we are having a storage cost impact of that. By exposing the sentry extracted tags in the entity we can stop duplicating data and get rid of the redundancy.

PR #5165, opened 2023-12-06
fix(rust): Fix SLO metrics again

untitaker:
This time around, all our SLOs are zero. This is because we subtracted
the larger timestamp from the smaller one instead of the other way
around, always leading to an underflow

### DIFF
--- a/rust_snuba/src/types.rs
+++ b/rust_snuba/src/types.rs
@@ -39,11 +39,11 @@ impl LatencyRecorder {
 
         let write_time = write_time.timestamp() as u64;
 
-        let latency = self.max_secs.saturating_sub(write_time) * 1000;
-        metrics.timing(&format!("insertions.max_{}_ms", metric_name), latency, None);
+        let max_latency = write_time.saturating_sub(self.max_secs) * 1000;
+        metrics.timing(&format!("insertions.max_{}_ms", metric_name), max_latency * 1000, None);
 
-        let latency = ((self.sum_secs * 1000.0 / self.num_values as f64) as u64)
-            .saturating_sub(write_time * 1000);
+        let latency = write_time as f64 - (self.sum_secs / self.num_values as f64);
+        let latency = (latency * 1000.0) as u64;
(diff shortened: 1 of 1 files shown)


rust_snuba/src/types.rs
```diff
@@ -39,11 +39,15 @@ impl LatencyRecorder {
...
-        metrics.timing(&format!("insertions.max_{}_ms", metric_name), latency, None);
-
-        let latency = ((self.sum_secs * 1000.0 / self.num_values as f64) as u64)
-            .saturating_sub(write_time * 1000);
+        let max_latency = write_time.saturating_sub(self.max_secs) * 1000;
+        metrics.timing(
+            &format!("insertions.max_{}_ms", metric_name),
+            max_latency * 1000,
+            None,
+        );
```
Swatinem, line 47: now you are doing `* 1_000` twice :-D

Swatinem (commented):
I somehow have a hard time to understand how these metrics are supposed to work fundamentally?

From the names, I think you would want to measure the largest latency of a message in the batch, plus an average latency for that batch.

For the max latency, I believe naming it `max` is wrong, as well as using `max` when merging. You would want the timestamp of the *earliest* message (and use `min` on merge).
For latency, things should work fine I believe, except you mentioned at some point that overflow when adding a ton of timestamps might be a problem, which I believe it is. There maybe using some form of rolling average which does not require the sum total might be a solution.

untitaker:
> except you mentioned at some point that overflow when adding a ton of timestamps might be a problem, which I believe it is

i have changed my mind about the overflow issue (again). i think when we overflow we should make this visible somehow, and i think the rust behavior of just wrapping around is actually desirable in that case as it will almost certainly blow up our SLOs.

the rest of the review i agree with. this is getting super messy now... i am doing things quite differently from python in general and am not using 100% of my brain, clearly

untitaker:
https://github.com/getsentry/snuba/pull/5169

PR #5172, opened 2023-12-07
ref: Add experimental readiness state

evanh:
This state is meant for experimental features that should only be used locally
or in S4S.


volokluev:
Do we need to update any documentation with this new readiness state?

evanh:
> Do we need to update any documentation with this new readiness state?

I've already updated this doc: https://www.notion.so/sentry/Snuba-readiness-states-3752e3358ed94e04a09f7917a0095a85
I didn't update the diagram, I'm going to wait for @enochtangg to come back so I don't have to recreate it from scratch.

volokluev (approved):
Thanks for updating the docs!

PR #5180, opened 2023-12-08
chore(gocd): bump gocd-jsonnet version

IanWoodard:
Bumping to v2.7 of gocd-jsonnet which should remove the de region temporarily while we figure out why it is broken and remove customer-5 as we are getting rid of it.


kneeyo1 (approved):
lgtm.

volokluev (approved):

PR #5188, opened 2023-12-10
ref(rust): Make sure we log if the DLQ future errors

lynnagara:


untitaker:
i think this is now also done in https://github.com/getsentry/snuba/pull/5189

PR #5193, opened 2023-12-11
feat(metrics) Add a MQL endpoint to Snuba API 

evanh:
Add an endpoint to the Snuba API that accepts MQL instead of SnQL. This
endpoint expects a MQL string in the `query` field of the request, and a new
field called `mql_context` that contains the meta information not included in
the MQL query.

This new endpoint deserializes the MQL string and the context into a
`LogicalQuery` object, and then resolves all the fields in that query. This is
a departure from how metrics were being handled before: the indexer resolve
happens in Snuba now instead of in Sentry. Since Snuba doesn't actually have
access to the indexer, the relevant information needs to be passed in from
Sentry so that the query can be properly transformed.

Once the `LogicalQuery` is created and the resolve steps are completed, the
pipeline is exactly the same as the SnQL endpoint. This endpoint differs only
in how the initial `LogicalQuery` object is created.


enochtangg (approved):
Awesome! Thanks for catching the `AGGREGATE_ALIAS` stuff too.

PR #5195, opened 2023-12-11
ref(rust): Remove concurrency_override

untitaker:
We now have a few consumers in prod, and this override makes things more
confusing to review rollouts.


untitaker:
from slack: we have a few more consumers in non-prod where we feel that this is useful. let's leave it for now

PR #5197, opened 2023-12-12
fix(rust): Increase poll interval

lynnagara:
We are seeing this time out in prod, it's probably just too low?


untitaker:
> We are seeing this time out in prod, it's probably just too low?

how do you see this time out? max.poll.intervall is unrelated imo

lynnagara:
> how do you see this time out? max.poll.intervall is unrelated imo

Is it possible that `consumer.poll(10)` returns None instead of the offset reset error since the timeout is so low? Then the assignment doesn't work.

untitaker (approved):
ah got it. worth a try

PR #5199, opened 2023-12-12
fix(cache): Slightly clean up cache set errors

evanh:
When the query takes longer than the cache thread timeout, the executor raises
a concurrent.futures.TimeoutError. This was mislabelled as a cache set error,
which it is not. The cache set happens further down in the execution, and a
warning is logged for that. This is a query simply taking too long to run.

As such, reword the error to be slightly more informative and create a new
status to track the difference between a query taking to long to run and a
query that fails to set the cache.


volokluev (approved):
Thank you for this!

PR #5200, opened 2023-12-12
perf: Increase consumer join timeout

lynnagara:
A few consumers (spans, generic metrics distributions) regularly exceed the 5 second timeout currently configured, and cannot finish processing pending work in time. When this is the case, the multiprocessing pool cannot be reused since there is state in the pool that cannot be carried over to the next assignment. This aims to avoid that scenario.

### DIFF
--- a/snuba/cli/consumer.py
+++ b/snuba/cli/consumer.py
@@ -136,7 +136,7 @@ logger = logging.getLogger(__name__)
     "--output-block-size",
     type=int,
 )
-@click.option("--join-timeout", type=int, help="Join timeout in seconds.", default=5)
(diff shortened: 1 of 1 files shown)


ayirr7:
> When this is the case, the multiprocessing pool cannot be reused since there is state in the pool that cannot be carried over to the next assignment

Just a question, what happens exactly when there is state in the pool? Like does that trigger creation of a new multiprocessing pool somehow?

ayirr7 (approved):

lynnagara:
> Just a question, what happens exactly when there is state in the pool? Like does that trigger creation of a new multiprocessing pool somehow?

Yep. We do the same process as before the change.. recreate the pool next assignment. This is an attempt to avoid that more of the time.

PR #5210, opened 2023-12-14
feat(metrics-summaries): Add a Rust consumer for metrics summaries

phacops:


rust_snuba/src/processors/spans.rs
```diff
@@ -46,17 +46,41 @@ struct FromSpanMessage {
...
+}
+
+impl FromSpanMessage {
+    pub fn metrics_summary(&self) -> BTreeMap<String, Vec<FromMetricsSummary>> {
+        match serde_json::from_value::<BTreeMap<String, Vec<FromMetricsSummary>>>(
+            self._metrics_summary.clone(),
+        ) {
+            Ok(v) => v,
+            Err(err) => {
+                println!("{:#?}", err);
```
untitaker, line 69: print stmt. if you want to keep it, let's make it `tracing::error!`?

rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl TryFrom<FromSpanMessage> for MetricsSummaries {
+    type Error = anyhow::Error;
+
+    fn try_from(from: FromSpanMessage) -> anyhow::Result<MetricsSummaries> {
+        let mut metrics_summaries: Vec<MetricsSummary> = Vec::new();
+
+        let end_timestamp_ms = from.start_timestamp_ms + from.duration_ms as u64;
+        for (metric_mri, summaries) in from.metrics_summary() {
```
untitaker, line 68: nit: if all you need from the span message is the metrics summary, you can define a partial struct that only contains the fields you need to read:

```
struct SpanMessage {
    _metrics_summary: ...
}
```

i think the performance gain should be fairly significant, and then the metrics summaries processor would not require any other fields to be well-formed (ignoring the schema validation)

untitaker (approved):

untitaker (changes requested):
actually, the print stmt needs to go, the rest lgtm

rust_snuba/src/processors/metrics_summaries.rs
phacops: I need a little more than that but that's a better idea.

untitaker (approved):

PR #5220, opened 2023-12-15
fix(rust): Fix `self.consumer_offsets`

lynnagara:
self.consumer_offsets is the latest offsets seen by the consumer, not the committed offsets.

### DIFF
--- a/rust_snuba/rust_arroyo/Cargo.toml
+++ b/rust_snuba/rust_arroyo/Cargo.toml
@@ -16,7 +16,7 @@ serde_json = "1.0.81"
 thiserror = "1.0"
 tokio = { version = "1.19.2", features = ["full"] }
 tracing = "0.1.40"
-uuid = "1.5.0"
+uuid = { version = "1.5.0", features = ["v4"] }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


untitaker (approved):
ugh, sorry about that. I recently added this, i can't remember what I thought there. I was somehow under the impression that I ported `self.__offsets` from python, but indeed it's read offsets.

lynnagara:
@untitaker Check out the failing test now?! Isn't that the offset reset error we actually saw in prod?

untitaker:
> @untitaker Check out the failing test now?! Isn't that the offset reset error we actually saw in prod?

the error is fine. we currently log errors even though we then later reset the offset. the logged error is normal. the test is failing because of the next few lines, where `consumer.tell()` is returning the wrong value. and I think it's because i misunderstood what it's for

lynnagara:
> the error is fine. we currently log errors even though we then later reset the offset. the logged error is normal. the test is failing because of the next few lines, where `consumer.tell()` is returning the wrong value. and I think it's because i misunderstood what it's for

ah you're right, didn't read error message properly

PR #5229, opened 2023-12-19
fix(rust): Handle rebalancing ourselves, remove usage of pre/post-rebalance callbacks

untitaker:
previously: https://github.com/getsentry/snuba/pull/5181 https://github.com/getsentry/snuba/pull/5186

As explained in the code comments, this leads to calling rdkafka_assign
twice: once with the invalid offsets, and once with the
fixed-up offsets. since we assign our offsets twice, the theory is that the first assignment (the one with invalid offsets) leads to spurious errors in consumer.poll() that eventually go away

Let's replace the entire flow with something of our own. It turns out
that the use of unsafe in the original implementation of
ConsumerContext::rebalance is also entirely unnecessary, since we use https://github.com/fede1024/rust-rdkafka/pull/636 where we have access to BaseConsumer and therefore to safe bindings of `rdkafka_assign`


rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
```diff
@@ -165,107 +165,113 @@ impl<C: AssignmentCallbacks + Send + Sync> ClientContext for CustomContext<C> {
...
+    // if we call rdkafka.assign multiple times, we have seen random AutoOffsetReset errors popping
+    // up in poll(), since we (briefly) assigned invalid offsets to the consumer
+    fn rebalance(
+        &self,
+        base_consumer: &BaseConsumer<Self>,
+        err: RDKafkaRespErr,
+        tpl: &mut TopicPartitionList,
+    ) {
+        match err {
+            RDKafkaRespErr::RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS => {
```
lynnagara, line 191: Does this correspond to pre-rebalance or post-rebalance?

rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
```diff
@@ -165,107 +165,113 @@ impl<C: AssignmentCallbacks + Send + Sync> ClientContext for CustomContext<C> {
...
-                    let resolved_offset = match self.initial_offset_reset {
-                        InitialOffset::Earliest => low_watermark,
-                        InitialOffset::Latest => high_watermark,
-                        InitialOffset::Error => {
-                            panic!("received unexpected offset");
+                base_consumer
+                    .unassign()
+                    .expect("failed to revoke partitions");
+            }
+            RDKafkaRespErr::RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS => {
```
lynnagara, line 213: Same question here - is it pre-rebalance or post-rebalance?

lynnagara (approved):

rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
Swatinem: our pre-rebalance matches for `revoke` directly, so this corresponds to pre-rebalance, and vice-versa down below.

rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
```diff
@@ -165,107 +165,113 @@ impl<C: AssignmentCallbacks + Send + Sync> ClientContext for CustomContext<C> {
...
+                let committed_offsets = base_consumer
+                    .committed_offsets((*tpl).clone(), None)
+                    .unwrap();
+
+                let mut offset_map: HashMap<Partition, u64> = HashMap::new();
+                let mut tpl = TopicPartitionList::with_capacity(offset_map.len());
```
Swatinem, line 219: The `with_capacity` is useless here, as the map is empty. Both can be pre-allocated with the length of `committed_offsets` though.

Swatinem (approved):
nice!

rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
untitaker: we are implementing the function that normally calls both of those callbacks. this function is the entire rebalancing, both pre and post. whatever runs before `consumer.assign` in this function is "pre-rebalance", whatever runs after is "post-rebalance"

rust_snuba/rust_arroyo/src/backends/kafka/mod.rs
untitaker: copy-paste error, thanks

PR #5232, opened 2023-12-19
feat(replays): Add react_component_name column

cmanallen:


lynnagara:
Can you bump the rust lib too? https://github.com/getsentry/snuba/blob/78783784ba319813655220f1753ec7e4ffd5bd2f/rust_snuba/Cargo.toml#L38

cmanallen:
@lynnagara I'm going to do it on another PR since its becoming more involved.

lynnagara (approved):

PR #5233, opened 2023-12-19
ref: Python 3.9.18

lynnagara:
This is the same version as sentry


untitaker:
it needs to be changed here: https://github.com/getsentry/snuba/blob/ce1e948489124215bface3bbcc319b88ea0727cf/Dockerfile#L1

and here: https://github.com/getsentry/snuba/blob/ce1e948489124215bface3bbcc319b88ea0727cf/.github/workflows/ci.yml#L42

untitaker (approved):

PR #5235, opened 2023-12-20
feat(profiling): Add ondemand profiler code

untitaker:
going to snuba admin and setting 'ondemand_profiler_hostnames' to a pod name
will now capture a profile and send it to sentry once the option is un-set again.

we can't really figure out how to fit sentry's transaction concept into
snuba consumers, so we start a "bogus" transaction for any user-defined
amount of seconds that runs on the main thread.

we should turn on profiling for all web workers by default (even at low
sample rate) to get a feeling for the overhead. overhead is not linear
with sample rate. there is a baseline cost of having a non-zero value,
and then there is a cost per captured profile.

once there is any sample rate > 0 deployed in Snuba for some pods,
ondemand profiles can be performed for those pods. those are forced into
100% sampling regardless of sample rate (again, for as long as the
sample rate is > 0)


snuba/utils/profiler.py
```diff
@@ -0,0 +1,63 @@
...
+            # should be obvious why.
+            logger.warn("starting ondemand profile for %s", own_hostname)
+
+            with sentry_sdk.Hub.main:
+                open_transaction = sentry_sdk.start_transaction(
+                    name=f"ondemand profile: {own_hostname}", sampled=True
+                )
+                assert isinstance(open_transaction, Transaction)
+                assert open_transaction._profile is not None
+                open_transaction._profile.sampled = True
```
Zylphrex, line 40: Instead of doing this explicitly, you could use the `profiles_sampler` to decide when the profiler is started/sampled. And I assume the reason you call `__enter__` and `__exit__` explicitly below is so you can add this logic before starting the transaction which autoruns the profiler.

untitaker: i have to call `__enter__` and `__exit__` because otherwise the transaction doesn't start, regardless of sampling decision, i think

`profiles_sampler` works, but i feel that i would have to pass some value into the transaction that tells the sampler what to do. i don't want to sample 100% of flask transactions for now

Zylphrex: Ahh, I see, so the whole thing is controlled by the while loop outside and that either starts a transaction/profile, stops it, or wait 5s.

One caveat to mention here is that the profiler has a 30s limit and will buffer the profile in memory until the transaction ends. So It might be better to start/stop a new transaction every N seconds to flush the profile.

Perhaps something like

```python
while True:
    queried_hostnames = get_config("ondemand_profiler_hostnames") or ""
    queried_hostnames = queried_hostnames.split(",")
    
    if own_hostname in queried_hostnames and open_transaction is None:
       transaction = ...
       # set up the profile
       with transaction:
           time.sleep(5)
    else:
        time.sleep(5)
```

untitaker: >Ahh, I see, so the whole thing is controlled by the while loop outside and that either starts a transaction/profile, stops it, or wait 5s.

yeah I don't even want to profile this thread, i'm interested in the main thread (on _some_ pods)

got it wrt in-memory buffering. i pushed a fix that runs a profile for up to the full 30 seconds before sending the profile to sentry. i think we want to max that time out fully unless we see impact of the overhead in prod and want to turn it off again.

snuba/environment.py
```diff
@@ -86,8 +86,15 @@ def setup_sentry() -> None:
         ],
         release=os.getenv("SNUBA_RELEASE"),
         traces_sample_rate=settings.SENTRY_TRACE_SAMPLE_RATE,
+        # "turn on" profiling without sending anything --
+        # nonzero value required for ondemand profiling as well
+        # profiles_sample_rate=0.00001,
```
Zylphrex, line 91: I believe this isn't necessary to turn on profiling. Your trick to set `transaction._profile.sampled = True` should suffice for it to start the profiler.

snuba/utils/profiler.py
Zylphrex: There is no way to opt a thread out of being profiled but setting the active thread id should surface whatever is running on the main thread instead. 👍 

I'll let you decide how long to buffer for as long as you're aware that the buffering lasts for the duration of the transaction.

Zylphrex (approved):
LGTM 👍

snuba/environment.py
untitaker: without setting either a profiling_sampler or a nonzero sample rate, there will be no global scheduler, and so this assertion will fail: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/profiler.py

i could run my own scheduler and pass it in, but then i might have two schedulers when combined with the auto-instrumentation for flask that we also want to deploy at some point.

snuba/utils/profiler.py
untitaker: i'm not overly concerned about memory usage atm, my biggest fear comes from impact on CPU (or rather, the scheduler hogging the GIL and making the main thread slow). in that sense it would be really nice to have a way to opt out of calling `extract_stack` for certain (or almost all) threads

MeredithAnya (approved):

PR #5236, opened 2023-12-20
Optimize ConsumerState

Swatinem:
This has a bunch of micro-optimizations:

- Moves the `Arc<Mutex>` into `ConsumerState`.
- Avoids two locking operations per `run_once` by moving the `is_paused` flag to an `AtomicBool`. Though this is pretty much academic since we never have contention on this lock anyway, as it is purely single-threaded.
- Optimizes global metrics a bit by avoiding the `Arc` completely, and instead just use a `&'static dyn Metrics`.

### DIFF
--- a/rust_snuba/benches/processors.rs
+++ b/rust_snuba/benches/processors.rs
@@ -1,6 +1,5 @@
 use parking_lot::Mutex;
-use std::collections::HashMap;
-use std::sync::{Arc, Once};
+use std::sync::Arc;
 use std::time::Duration;
 
 use divan::counter::ItemsCount;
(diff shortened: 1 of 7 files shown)


untitaker (approved):

untitaker:
> Avoids two locking operations per run_once by moving the is_paused flag to an AtomicBool. Though this is pretty much academic since we never have contention on this lock anyway, as it is purely single-threaded.

I suspect that you may be able to replace Mutex::lock with RefCell::borrow_mut

Swatinem:
> I suspect that you may be able to replace Mutex::lock with RefCell::borrow_mut

Not really, as all the Callbacks need to be `Send + Sync` :-(

If it weren’t for those, and the circular references they introduce, we wouldn’t need any locking or refcounting to begin with.

PR #5238, opened 2023-12-20
feat(metrics): Create new DataSource for running multiple metrics queries

evanh:
Effectively, the idea of how to run metrics queries with formulas would be to
run a separate subquery for each timeseries requested, and then combine the
results back together from each of these queries.

The `CompositeQuery` class already has a lot of utlities in place for handling
subqueries, so this PR attempts to leverage those existing utilities. This
means modifying the composite pipeline to create a query plan for each of the
subqueries in a CompositeQuery, run each of them, and then use the result
function to rebuild the results.

### Legal Boilerplate

### DIFF
--- a/snuba/pipeline/composite.py
+++ b/snuba/pipeline/composite.py
@@ -522,6 +522,7 @@ class CompositeExecutionPipeline(QueryExecutionPipeline):
     def execute(self) -> QueryResult:
         plan = CompositeQueryPlanner(self.__query, self.__settings).build_best_plan()
         root_processors, aliased_processors = plan.get_plan_processors()
+        # TODO: This is also run in the `CompositeExecutionStrategy`, do we need it twice?
         ProcessorsExecutor(
             root_processors,
             aliased_processors,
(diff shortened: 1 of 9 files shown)


evanh:
I tried to document the changes more here: https://www.notion.so/sentry/CompositeQuery-stuff-13ae4e7853ab409396b89c71d7bae928

I think we can use the `CompositeQuery` class for metric queries. Adding a new DataSource (`MultiQuery` in my PR) wasn't difficult, and that is really the main difference.

This was effectively a test to see if we could use the `CompositeQueryPipeline` class, instead of having to create a new type of `Pipeline` altogether. `CompositeQueryPipeline` already has the concept of subqueries, visitors for those queries, running processors etc.

It suffers from one flaw though: It assumes that the ultimate output of a `CompositeQuery` is a single SQL query that runs on a single Clickhouse table. That means it enforces a single storage set and the `runner` it uses isn't capable of running multiple SQL statements on different tables.

This is a blocker for metrics, since different metric types exist in different tables, each with different storage sets. It's not clear to me yet  if we can modify `CompositeQueryPipeline` to handle this case, or we need to build from scratch.

evanh:
I don't think this is the right solution, or at least I want to explore some other ideas.

PR #5247, opened 2023-12-21
ref: skip docker pull

asottile-sentry:
buildx caching does not need the images locally to use them for caches -- so pulling them just wastes time


asottile-sentry:
this saves about a minute on every snuba CI run :D

untitaker (approved):

PR #5248, opened 2023-12-21
Use `setup-sentry` defaults

Swatinem:


Swatinem:
I guess this is a duplicate of #5246

PR #5251, opened 2023-12-21
ref: skip unnecessary push/pull for buildx build

asottile-sentry:
in the fully-cached case, buildx can avoid downloading / uploading all image contents


asottile-sentry:
there's something weird happening with cache collisions -- will debug later

PR #5253, opened 2023-12-22
fix(rust): pause/resume bugfixes

untitaker:
Implement pause/seek/resume/tell properly, and attempt to reproduce a rdkafka bug mentioned in python-arroyo where seek + pause would ignore the seek, + write testcase

SNS-2556


Swatinem (approved):

untitaker:
@lynnagara I wrote a testcase for the behavior mentioned in python-arroyo but couldn't repro it. there's also no testcase like this in python-arroyo. i think what might have happened is a rebalancing event that would then reset to the most recently committed offsets.

lynnagara (approved):

PR #5254, opened 2023-12-22
feat(spans): Add index on span_id

phacops:
We're aiming to speed queries by `span_id` using this index.


iambriccardo:
Just to give more context to this PR. We are now building the spans and metrics correlation which requires us to query the spans table by a set of span ids. Since the product decisions are not well defined, we are aiming to build an MVP that allows us to play with the querying and get a feel for what the final user experience will look like.

The end goal would be to denormalize data in a way which suits our access patterns.

evanh (approved):
LGTM, thanks!

PR #5263, opened 2023-12-27
ref(rust): Add rust to devenv, and remove querylog python processor

untitaker:
we deployed the rust message processor to rust. in order to not maintain two codebases, use the rust processor where the python one is needed (sentry tests, etc)

**with this PR, compiling rust_snuba will become mandatory in snuba devenv.** there are ways to only require it when querylog consumer is being started, but eventually we will require more and more rust in snuba's fundamentals, so I'd like to merge as-is to see how S&S folks react to it. developers of sentry should be unaffected since they run snuba in docker.


tests/datasets/test_querylog_processor.py
```diff
@@ -153,87 +153,3 @@ def test_simple() -> None:
         ],
         None,
     )
-
```
untitaker, line 156: need to confirm with team-search that this can go -- negative ids arent' supported in rust and it doesn't seem to cause problems

config/hooks/post-merge
```diff
@@ -10,22 +10,26 @@ trap "rm -f ${files_changed_upstream}" EXIT
...
 
 EOF
+
+grep -E --quiet 'rust_snuba/' "$files_changed_upstream" && cat <<EOF
+
+[${red}${bold}!!!{$reset}] Rust code changed, you may need to run:
+
+  make install-rs-dev
+
+EOF
```
lynnagara, line 35: Can we avoid ever requiring the user to do this? What if the `snuba rust-consumer` CLI command just runs it before launching the consumer instead?

lynnagara:
Can you add a test for the compat processor? Is it only for tests, or do you actually see some use case where we might want to run Rust processing function with Python consumer. If only for tests, I wonder if those tests should actually be migrated instead, or run on both Rust and Python variants anyway.

untitaker:
> Is it only for tests, or do you actually see some use case where we might want to run Rust processing function with Python consumer.

we still have the /test/ endpoints in Snuba for inserting new data via HTTP, I don't see an easier way to enable that usecase.

> Can you add a test for the compat processor?

There's still some querylog tests that indirectly test the compat processor. How would a more narrow unittest look like? I can't think of a way to mock out the dataset from Python.

lynnagara:
> we still have the /test/ endpoints in Snuba for inserting new data via HTTP, I don't see an easier way to enable that usecase.

Ah, forgot about the test endpoint. Makes sense.

> There's still some querylog tests that indirectly test the compat processor. How would a more narrow unittest look like? I can't think of a way to mock out the dataset from Python.

I was thinking of a separate processor that is not one of our real datasets in prod. In general I think the pattern we have of writing tests against real datasets has not served us - we now can't delete our legacy datasets because of this. But I realise now that this is not very easy to achieve since we can't write Rust and Python test code together, so I think the querylog tests are fine.

lynnagara (approved):

untitaker:
> In general I think the pattern we have of writing tests of core Snuba functionality against real datasets has not served us 

I think we can totally add a new dummy dataset in Rust if that makes things easier

untitaker:
> Can we avoid ever requiring the user to do this? What if the snuba rust-consumer CLI command just runs it before launching the consumer instead?

let me try this though

config/hooks/post-merge
untitaker: @lynnagara i see you already approved the PR, but i think we could do this. the concern i'd see is that when we change the rust code, people will be forced to recompile everytime, even though they might be fine running an older version of the code

for example, i also sometimes skip on the `make install-python-dependencies` since the code likely runs fine for my purposes anyway

PR #5266, opened 2023-12-28
feat(rust): Alternative hybrid consumer that uses RunTaskWithMultiprocessing

lynnagara:
The current hybrid consumer that uses procspawn is deadlocking. This is an alternative approach to hybrid which uses the existing RunTaskWithMultiprocessing strategy from Python Arroyo from a Rust consumer.

All hybrid consumer code is considered highly experimental.

### DIFF
--- a/rust_snuba/src/strategies/python_v2.rs
+++ b/rust_snuba/src/strategies/python_v2.rs
@@ -13,6 +13,7 @@ use rust_arroyo::processing::strategies::{
 use rust_arroyo::types::{BrokerMessage, InnerMessage, Message, Partition, Topic};
 use rust_arroyo::utils::timing::Deadline;
 use std::collections::{BTreeMap, VecDeque};
+use std::env;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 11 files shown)


rust_snuba/src/strategies/python_v2.rs
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    fn submit(&mut self, message: Message<KafkaPayload>) -> Result<(), SubmitError<KafkaPayload>> {
+        if self.transformed_messages.len() > 100_000 {
+            return Err(SubmitError::MessageRejected(MessageRejected { message }));
+        }
+
+        let mut apply_backpressure = false;
+
+        let _ = Python::with_gil(|py| -> PyResult<()> {
```
untitaker, line 143: please handle this result. you can probably just unwrap here

snuba/consumers/rust_processor.py
```diff
@@ -66,3 +77,171 @@ def process_rust_message(
...
+        """
+        self.__inner.close()
+        try:
+            self.__inner.join(timeout=timeout)
+        except InvalidMessage:
+            # We cannot forward offsets here since the inner strategy is already
+            # marked as closed. Log the exception and move on. The message will get
+            # reprocessed properly when the consumer is restarted.
+            logger.warning("Invalid message in join", exc_info=True)
+        self.__pool.close()
```
untitaker, line 242: i think you need to close or terminate the pool in `terminate` too

untitaker (changes requested):
let's just deploy this for now since it's not running on any prod data. some comments inline but otherwise lgtm

rust_snuba/src/strategies/python_v2.rs
lynnagara: Updated

untitaker (approved):
approving so you can merge async but the above review comments are imo important

PR #5268, opened 2023-12-28
perf: replace divan with criterion

anonrig:
Replaces Divan with Criterion. This opens the possibility of integration codspeed.io into our workflow.

### DIFF
--- a/rust_snuba/benches/processors.rs
+++ b/rust_snuba/benches/processors.rs
@@ -1,9 +1,13 @@
+mod utils;
+
+use crate::utils::{functions_payload, profiles_payload, querylog_payload, spans_payload, RUNTIME};
+
 use parking_lot::Mutex;
 use std::sync::Arc;
 use std::time::Duration;
 
-use divan::counter::ItemsCount;
-use once_cell::sync::Lazy;
+use criterion::measurement::WallTime;
+use criterion::{black_box, criterion_group, BenchmarkGroup, BenchmarkId, Criterion, Throughput};
 use rust_arroyo::backends::kafka::types::KafkaPayload;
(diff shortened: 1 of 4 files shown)


rust_snuba/benches/processors.rs
```diff
@@ -176,199 +102,94 @@ fn create_stream_processor(
...
+                    black_box(MSG_COUNT),
+                ));
+                loop {
+                    let res = processor.run_once();
+                    if matches!(res, Err(RunError::Poll(ConsumerError::EndOfPartition))) {
+                        // FIXME: this pretty much means that we *polled* the partition to the end,
+                        // it does not mean that we actually *committed* everything
+                        // (aka we finished actually processing everything).
+                        break;
+                    }
```
Swatinem, line 131: I believe you can also properly `close` the processor now, as the underlying deadlock was fixed.

Swatinem (approved):
lgtm. it would be nice to add back the throughput measurement/output as well.

anonrig:
> lgtm. it would be nice to add back the throughput measurement/output as well.

Should I go for `KiB/s` or `Kelem/s`?

rust_snuba/benches/processors.rs
anonrig: Should I just call `processor.shutdown()` on EndOfPartition error, and break?

Swatinem:
I would go with messages/sec, as that is pretty much the metric we look for in prod.
However those few messages we have there to test probably do not fully reflect whatever we see in prod though.

Swatinem (approved):

untitaker (approved):

PR #5271, opened 2023-12-29
feat(rust): Port outcomes message processor to Rust

lynnagara:
This ports all of the functionality from the Python message processor, including recording metrics (even though they seem like they may be unnecessary?).

It involves a small refactor to the Python message processor to produce unix timestamp instead of Python datetime objects so that the outputs of the 2 message processors are easier to compare in tests.


untitaker (approved):

untitaker:
since #5263 was merged, you could already remove the python processor now

PR #5275, opened 2023-12-29
fix: Hybrid consumer stacktraces

lynnagara:


untitaker (approved):

Swatinem (approved):

Swatinem:
We could also take a middle way here, like we do in symbolicator and other services:
- Use the full debug info for debug builds (the default) so we can properly attach a debugger.
- Turn on line-only debug info for production builds, so we have proper stack traces with inlinees, but without the option to attach a debugger.

untitaker:
> Turn on line-only debug info for production builds, so we have proper stack traces with inlinees, but without the option to attach a debugger.

What do we gain in this scenario, smaller images, faster stacktrace capturing?

PR #5276, opened 2023-12-29
fix(rust): Ensure Python does not return naive datetime to Rust consumer

lynnagara:
This fixes crashloop in the hybrid spans consumer

### DIFF
--- a/snuba/consumers/rust_processor.py
+++ b/snuba/consumers/rust_processor.py
@@ -15,7 +15,7 @@ import importlib
 import logging
 import os
 from collections import deque
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Deque, Mapping, Optional, Sequence, Tuple, Type, Union, cast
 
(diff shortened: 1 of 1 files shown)


snuba/consumers/rust_processor.py
```diff
@@ -57,6 +57,12 @@ def initialize_processor(
 initialize_processor()
 
 
+def ensure_utc(value: Optional[datetime]) -> Optional[datetime]:
+    if value and value.tzinfo is None:
+        value.replace(tzinfo=timezone.utc)
```
untitaker, line 62: this returns a new datetime that you need to return
```suggestion
        return value.replace(tzinfo=timezone.utc)
```

untitaker (approved):

snuba/consumers/rust_processor.py
lynnagara: ugg, thanks

PR #5278, opened 2023-12-29
fix(ci): Use ghcr for self-hosted image

untitaker:
trying to mimick: https://github.com/getsentry/sentry/pull/61434/files

we already publish nightlies and whatnot to ghcr to serve the needs of
'sentry devservices'. it should be possible to use the same exact
multiarch image for self-hosted.

the rest in cloudbuild.yml seems to run unittests that we already run in GHA


untitaker:
nvm this is broken. we deploy from gcr. do we want to deploy from ghcr instead?

evanh:
> do we want to deploy from ghcr instead?

What's the difference?

untitaker:
> What's the difference?

I'm trying to get rid of google cloud build in order to remove the need to run `/gcbrun` on external PRs. I think that means i can't use Google Container Registry (gcr.io). So this PR now moves everything to GitHub Container Registry (ghcr.io)

untitaker:
feedback from @asottile-sentry is that we don't actually know whether we can stop publishing to `gcr.io` entirely. current idea is to build in GHA but still publish to GCR, same as we do with DockerHub.

untitaker:
we made the gcr/gcb build not required for merge and gutted it to only build the docker image in #5416 , so the pain for external contributors and flakiness should be mostly gone.

PR #5281, opened 2024-01-02
ref(rust-docker): Build deps separately from main app

untitaker:
this should make PRs where Rust consumers are touched is touched significantly faster, as only application code is being rebuilt rather than the entire rust deptree.

we considered using `cargo chef`, but it doesn't work because our build tool is `maturin`, not `cargo`.

we also saw that relay instead mounts the `target/` folder into the docker container for caching, but i felt that this was harder to implement though it's probably better

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,9 +28,13 @@ RUN set -ex; \
         g++ \
         gnupg \
     '; \
+    runtimeDeps=' \
+        libjemalloc2 \
+    '; \
     apt-get update; \
-    apt-get install -y $buildDeps --no-install-recommends; \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -28,9 +28,13 @@ RUN set -ex; \
         g++ \
         gnupg \
     '; \
+    runtimeDeps=' \
```
untitaker, line 31: i moved the installation of libjemalloc2 as well so that it is not being re-installed everytime there's a code change in ./snuba/

ayirr7 (approved):

evanh (approved):

PR #5289, opened 2024-01-03
fix(rust): Fix for clickhouse 21.8

lynnagara:
This ports the clickhouse encoding logic from the Python consumer to Rust. We are now using transfer-encoding: chunked like we do in Python.

Not using chunked caused issues when inserting large payloads in 21.8. The issue was not present in 22.8.

Fix SENTRY-FOR-SENTRY-4VD

### DIFF
--- a/rust_snuba/src/types.rs
+++ b/rust_snuba/src/types.rs
@@ -171,7 +171,7 @@ impl BytesInsertBatch {
         self.rows.encoded_rows.len()
     }
 
-    pub fn encoded_rows(&self) -> &Vec<Vec<u8>> {
+    pub fn encoded_rows(&self) -> &[u8] {
         &self.rows.encoded_rows
     }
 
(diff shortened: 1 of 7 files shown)


rust_snuba/src/types.rs
```diff
@@ -183,8 +182,7 @@ impl BytesInsertBatch {
 
 #[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq)]
 pub struct RowData {
-    encoded_rows: Vec<u8>,
-    num_rows: usize,
+    encoded_rows: Vec<Vec<u8>>,
```
untitaker, line 185: this is a rather inefficient encoding since you will have one allocation per row. I think it should be possible to let a row go across chunks, or if not, split by `\n` again in the clickhouse strategy. I'm 80% sure either way is more efficient

untitaker (changes requested):

untitaker (approved):

rust_snuba/src/types.rs
lynnagara: Good point, reverted  this

PR #5290, opened 2024-01-03
fix(docker): Use macos to build arm image

untitaker:
instead of using QEMU to build arm64, use macos-m1. in theory this could be faster, assuming GHA actually gives us a runner in time


evanh (approved):

untitaker:
it seems that installing colima in macos runners uses qemu to run the VM that hosts docker, meaning we're back to bad performance. and it doesn't even work at all right now.

PR #5295, opened 2024-01-03
fix(slo): Mark invalid UUID error code as an invalid request

evanh:
A user hit an edge case where they were filtering for `['replay_id', '!=', 'null']` in an events endpoint. This caused an error in Clickhouse which wasn't being classified correctly by the SLO.


untitaker (approved):

lynnagara:
does it not use the UUIDColumnProcessor? Doesn't that get rejected before hitting clickhouse usually?

evanh:
> does it not use the UUIDColumnProcessor?

Replays does not use that processor.

PR #5296, opened 2024-01-04
Add more Rust benchmarks

Swatinem:
This cycles through all example payloads from the kafka schemas repo that I copy-pasted.

Plus also adds benchmarks for the newly added processors, though `outcomes` is the only one that works.

Each benchmark now also defines a "processor fn only" benchmark which only processes the payloads without the whole `StreamProcessor`.

---

As a followup, maybe it is a good idea to add all the examples as static data into the kafka schemas rust crate?

### DIFF
--- a/rust_snuba/benches/processors.rs
+++ b/rust_snuba/benches/processors.rs
@@ -1,13 +1,10 @@
-mod utils;
-
-use crate::utils::{functions_payload, profiles_payload, querylog_payload, spans_payload, RUNTIME};
-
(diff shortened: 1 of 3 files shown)


untitaker:
> As a followup, maybe it is a good idea to add all the examples as static data into the kafka schemas rust crate?

i'd say go for it, i thought you already fixed that at some point but not sure the right APIs are exposed

untitaker (approved):
i'd follow up asap if you have time to do it?

PR #5300, opened 2024-01-04
fix: Re-add QueuingMetricSink to cadence setup

untitaker:
We are currently seeing only a handful of metrics being emitted and it
might be due to that.


lynnagara (approved):

untitaker:
this one was useless, reverting. the real bugfix is https://github.com/getsentry/statsdproxy/pull/41

PR #5302, opened 2024-01-04
hack(oncall): remove migration connection checks

volokluev:
Don't check all cluster connections when starting migrations. if you have a skipped cluster, it will fail the whole job. Make querylog deprecated (for now)

### DIFF
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -85,7 +85,6 @@ def migrate(
     Blocking migrations will not be run unless --force is passed.
     """
     setup_logging(log_level)
-    check_clickhouse_connections()
     check_for_inactive_replicas()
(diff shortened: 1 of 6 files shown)


dbanda (approved):

snuba/cli/migrations.py
```diff
@@ -85,7 +85,6 @@ def migrate(
     Blocking migrations will not be run unless --force is passed.
     """
     setup_logging(log_level)
-    check_clickhouse_connections()
```
volokluev, line 88: this is not strictly necessary. We should update this to only check clusters we will be migrating. Although it doesn't really guarantee that we won't lose connection while applying other migrations.

untitaker (approved):

PR #5305, opened 2024-01-05
Micro-optimize `functions` processor

Swatinem:
Avoids a bunch of clones by borrowing data from the input.

### DIFF
--- a/rust_snuba/src/processors/functions.rs
+++ b/rust_snuba/src/processors/functions.rs
@@ -13,7 +13,7 @@ pub fn process_message(
     _metadata: KafkaMessageMetadata,
 ) -> anyhow::Result<InsertBatch> {
     let payload_bytes = payload.payload().context("Expected payload")?;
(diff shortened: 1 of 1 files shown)


untitaker (approved):

rust_snuba/src/processors/functions.rs
```diff
@@ -13,38 +13,38 @@ pub fn process_message(
...
     };
     let device_classification = msg.device_class.unwrap_or_default();
 
-    let functions = msg.functions.into_iter().map(|from| {
+    let functions = msg.functions.iter().map(|from| {
         Function {
             profile_id: msg.profile_id,
             project_id: msg.project_id,
             // Profile metadata
-            browser_name: msg.browser_name.clone(),
```
untitaker, line 29: i'm curious why this clone was needed if we already used `into_iter` -- should we not have managed to just _move_ the `String` into the new struct?

Swatinem: You could have more than one `Function` per `msg`, thats why.

PR #5309, opened 2024-01-05
hack(oncall) Mark querylog as limited to avoid migration checks

evanh:


untitaker (approved):
in order to validate that this hack works on its own, let's revert all other hacks in the same PR?

volokluev:
The migration connection checks will run on all the clusters so this by itself won't be enough

PR #5314, opened 2024-01-05
fix(rust): Fix unbatched inserts

untitaker:
This was an accidental change done in https://github.com/getsentry/snuba/pull/5289 between review rounds. We were counting our batch sizes in bytes, not messages.

### DIFF
--- a/rust_snuba/src/types.rs
+++ b/rust_snuba/src/types.rs
@@ -149,6 +149,7 @@ impl BytesInsertBatch {
             .encoded_rows
             .extend_from_slice(&other.rows.encoded_rows);
         self.commit_log_offsets.extend(other.commit_log_offsets);
+        self.rows.num_rows += self.rows.num_rows;
         self.message_timestamp.merge(other.message_timestamp);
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

rust_snuba/src/types.rs
```diff
@@ -168,7 +169,7 @@ impl BytesInsertBatch {
     }
 
     pub fn len(&self) -> usize {
```
onewland, line 171: I don't know if this suggestion makes sense, but does the naming of this method potentially create the expectation that you're getting the length of the bytes array? e.g. can we prevent this from happening again by renaming this method itself to `pub fn row_count(&self) -> usize`?

untitaker: @onewland potentially? I think in the sequence of changes there wasn't any confusion about what this method is supposed to do. rather, we made a change to how the batch is internally represented, then decided against it, and accidentally left a few changes from the reverted change in. and i didn't review the final diff properly

PR #5318, opened 2024-01-05
fix(rust): Use the right origin timestamp for functions

phacops:


untitaker (approved):

lynnagara:
I noticed that timestamp is mandatory in the schema but is Option<u64> here, should we align these?

phacops:
Yes, correct, I pushed a commit for that as you were commenting on it.

lynnagara:
I mean `msg.timestamp` (as well as `msg.received`)

lynnagara:
https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/profile-functions.v1.schema.json#L68

phacops:
Oh, I see. Good point.

PR #5329, opened 2024-01-08
Reuse Tokio Handle in DlqPolicy

Swatinem:
Especially in benchmarks, creating a fresh runtime for every run leads to thousands of threads being spawned which makes local profiling difficult.

### DIFF
--- a/rust_snuba/src/consumer.rs
+++ b/rust_snuba/src/consumer.rs
@@ -132,9 +132,6 @@ pub fn consumer_impl(
 
     let logical_topic_name = consumer_config.raw_topic.logical_topic_name;
 
-    let concurrency = ConcurrencyConfig::new(concurrency);
-    let clickhouse_concurrency = ConcurrencyConfig::with_runtime(2, concurrency.handle());
-
     // DLQ policy applies only if we are not skipping writes, otherwise we don't want to be
     // writing to the DLQ topics in prod.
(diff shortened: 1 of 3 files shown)


rust_snuba/src/consumer.rs
```diff
@@ -132,6 +132,9 @@ pub fn consumer_impl(
 
     let logical_topic_name = consumer_config.raw_topic.logical_topic_name;
 
+    let concurrency = ConcurrencyConfig::new(concurrency);
+    let clickhouse_concurrency = ConcurrencyConfig::with_runtime(2, concurrency.handle());
```
untitaker, line 136: if we used the existing runtime for clickhouse, we'd limit the CPU usage we could use for processing pool, and it turns out that passing `concurrency` to `with_runtime` actually has no effect. this is bad because we want to limit the amount of concurrency we use against clickhouse

untitaker (changes requested):

rust_snuba/src/consumer.rs
Swatinem: I reverted this change. Now the normal processor, clickhouse and dlq get their own runtimes when initializing the consumer, but still in a way that does not wreck the whole "run benchmarks in profiler" usecase.

untitaker (approved):

PR #5330, opened 2024-01-08
ref(spans): Remove deprecated field from payload and set default value for insertion

phacops:


Swatinem (approved):

untitaker (commented):
do you want to update the schema definition as well?

phacops:
That's the thing, those fields are already not required.

untitaker (approved):

PR #5333, opened 2024-01-08
fix(rust): Port retention days logic from Python

lynnagara:
Validates that the retention days matches one of the valid retention days provided via settings. This is very important otherwise invalid values can create too many clickhouse parts. This function is used in most of our message processors as is also needed for metrics, errors, etc.

### DIFF
--- a/rust_snuba/src/processors/metrics_summaries.rs
+++ b/rust_snuba/src/processors/metrics_summaries.rs
@@ -1,4 +1,4 @@
-use crate::config::EnvConfig;
+use crate::config::ProcessorConfig;
 use anyhow::Context;
 use rust_arroyo::backends::kafka::types::KafkaPayload;
 use serde::{Deserialize, Serialize};
@@ -11,7 +11,7 @@ use crate::types::{InsertBatch, KafkaMessageMetadata};
 pub fn process_message(
     payload: KafkaPayload,
     _: KafkaMessageMetadata,
-    config: &EnvConfig,
+    config: &ProcessorConfig,
 ) -> anyhow::Result<InsertBatch> {
     let payload_bytes = payload.payload().context("Expected payload")?;
     let from: InputMessage = serde_json::from_slice(payload_bytes)?;
@@ -35,7 +35,7 @@ pub fn process_message(
                 metric_mri,
                 min: summary.min,
                 project_id: from.project_id,
-                retention_days: enforce_retention(from.retention_days, config),
+                retention_days: enforce_retention(from.retention_days, &config.env_config),
                 span_id: from.span_id,
                 sum: summary.sum,
                 tag_keys,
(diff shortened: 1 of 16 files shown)


rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -53,7 +55,7 @@ struct InputMessage {
     #[serde(default)]
     duration_ms: u32,
     project_id: u64,
-    #[serde(default = "default_retention_days")]
```
untitaker, line 56: `#[serde(default)]` should do the right thing AFAIK, not sure why we hooked up `default_retention_days` here to begin with

rust_snuba/src/processors/spans.rs
```diff
@@ -6,17 +6,20 @@ use serde::{Deserialize, Serialize};
...
 
-use crate::processors::utils::{default_retention_days, hex_to_u64, DEFAULT_RETENTION_DAYS};
+use crate::config::EnvConfig;
+use crate::processors::utils::{enforce_retention, hex_to_u64};
 use crate::types::{InsertBatch, KafkaMessageMetadata};
 
 pub fn process_message(
     payload: KafkaPayload,
     metadata: KafkaMessageMetadata,
+    config: &EnvConfig,
```
untitaker, line 16: can we wrap this in some kind of `MessageProcessorConfig` so that people start extending that struct instead of adding more and more arguments?

untitaker (changes requested):

rust_snuba/src/processors/metrics_summaries.rs
lynnagara: default_retention days was applying the 90 days previously

rust_snuba/src/processors/spans.rs
lynnagara: yep, updated

rust_snuba/src/processors/metrics_summaries.rs
untitaker: your change makes `retention_days` a mandatory argument. you need `serde(default)` in order to default it to `None`, which afaik gets coerced into 90d later

untitaker: wdym? we definitely have `serde(default)` in this processor, on duration_ms. where are we missing it?

untitaker: in metrics_summaries.rs it is. so i guess the two consumers we have don't agree on which fields are optional. but that's imo besides the point for this PR

lynnagara: There's examples outside of this topic though. In outcomes for example there are many optional fields: https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/outcomes.rs#L77-L80

There is no issue parsing them when they are missing, such as in this test: https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/outcomes.rs#L92-L98

untitaker: ugh, my bad, i think i misunderstood how this works. TIL

untitaker (approved):

PR #5336, opened 2024-01-08
ref(hybrid): Remove legacy procspawn-based implementation

lynnagara:


rust_snuba/Cargo.toml
```diff
@@ -58,9 +58,3 @@ tikv-jemallocator = "0.5"
 [[bench]]
 name = "processors"
 harness = false
-
-[[bin]]
-# this is not a proper benchmark since it runs forever.
-name = "python-processor-infinite"
```
untitaker, line 64: can we move that benchmark instead of deleting it?

untitaker (approved):

rust_snuba/Cargo.toml
lynnagara: sure, replaced it

untitaker (approved):

PR #5340, opened 2024-01-08
ref: Remove dead metrics processor code

lynnagara:
Sets/counters/distribution are no longer writable storages

Remove unused metrics processor code


onewland (approved):

untitaker (approved):

ayirr7:
It looks like the test that's failing is incorrectly returning the `sum(values)` across _all timestamps_ in the table instead of just those in the desired time period (between hour 2 and hour 3). This makes me wonder if the timestamps are not materializing properly with the right granularity when going from raw to aggregate table. Could there possibly be some logic relevant to the metrics materialized views that's accidentally being removed in this PR?

ayirr7:
Maybe we can try modifying the failing test by using the pattern [here](https://github.com/getsentry/snuba/blob/master/tests/test_metrics_api.py#L326-L332) and using the `timestamp_to_bucket` function to get the start and end timestamps for the query? Not sure if this would change anything, but it's a hunch since the test using that pattern seems to be passing.

PR #5342, opened 2024-01-09
fix(replays): Align Rust schema with Python schema and add additional test coverage

cmanallen:
1. Aligns the rust schema with the python schema.
2. Adds test coverage to assert the final Row type contains exactly the data we expect.
3. Fixes an issue in `tests/consumers/test_message_processor.py` where the test was not respecting the fact that the processor can return multiple rows.
4. I also segmented the deserialization and row building code from the consumer interfaces to ease unit testing.

Changes to the Python processor:

- Normalized UUID formatting.
- `platform` is now optional and will default to "javascript" if not provided.
- `is_archived` is coerced to its integer value
- `user` is now set to its empty state rather than `None`.


rust_snuba/src/processors/replays.rs
```diff
@@ -11,145 +11,166 @@ pub fn process_message(
...
+                &event.user.email,
+                &ip_address_v4,
+                &ip_address_v6,
+            ) {
+                (Some(uid), _, _, _, _) => uid.clone(),
+                (_, Some(username), _, _, _) => username.clone(),
+                (_, _, Some(email), _, _) => email.clone(),
+                (_, _, _, Some(ip), _) => ip.to_string().clone(),
+                (_, _, _, _, Some(ip)) => ip.to_string().clone(),
+                _ => String::new(),
```
cmanallen, line 103: I am very unhappy with this code.  Its a precedence check.  In python it looks like this `user = user_id or user_email or ""`.  Where the bool(user_id) evaluates to false if `None` or false if `""`.  The first non-empty string is accepted.  Any help cleaning this up is appreciated.

rust_snuba/src/processors/replays.rs
```diff
@@ -11,145 +11,166 @@ pub fn process_message(
...
-        }
+            })
+            .collect()),
         ReplayPayload::Event(event) => {
             let event_hash = match (event.event_hash, event.segment_id) {
                 (None, None) => Uuid::new_v4(),
                 (None, Some(segment_id)) => {
-                    Uuid::from_slice(md5::compute(segment_id.to_string().as_bytes()).as_slice())?
+                    Uuid::from_slice(md5::compute(segment_id.to_string().as_bytes()).as_slice())
+                        .expect("md5 hash never exceeds 16 bytes")
```
cmanallen, line 62: Is it cool to `expect` here if we know its impossible to trigger?

rust_snuba/src/processors/replays.rs
```diff
@@ -11,51 +11,59 @@ pub fn process_message(
     metadata: KafkaMessageMetadata,
 ) -> anyhow::Result<InsertBatch> {
     let payload_bytes = payload.payload().context("Expected payload")?;
+    let replay_row = deserialize_message(payload_bytes, metadata.partition, metadata.offset)?;
+    InsertBatch::from_rows([replay_row])
```
Swatinem, line 15: I don’t think this is what you want, rather:
```suggestion
    let replay_rows = deserialize_message(payload_bytes, metadata.partition, metadata.offset)?;
    InsertBatch::from_rows(replay_rows)
```

rust_snuba/src/processors/replays.rs
Swatinem: that should be fine, yes. Though you are only doing this because of the error type I believe?

rust_snuba/src/processors/replays.rs
```diff
@@ -69,90 +77,100 @@ pub fn process_message(
...
-            } else {
-                String::new()
+            // Unwrap the ip-address string.
+            let ip_address_string = event.user.ip_address.unwrap_or_default();
+            let (ip_address_v4, ip_address_v6) = match ip_address_string.parse::<IpAddr>() {
+                Err(_) => (None, None),
+                Ok(ip) => match ip {
+                    IpAddr::V4(ipv4) => (Some(ipv4), None),
+                    IpAddr::V6(ipv6) => (None, Some(ipv6)),
+                },
```
Swatinem, line 87:
```suggestion
                Ok(IpAddr::V4(ipv4)) => (Some(ipv4), None),
                Ok(IpAddr::V6(ipv6)) => (None, Some(ipv6)),
```

You can also match sub-patterns

rust_snuba/src/processors/replays.rs
Swatinem: You can chain `Option::or` without a problem, like so:

```rust
event.user.user_id.as_ref().or(event.user.username.as_ref())
.or(event.user.email.as_ref()).unwrap_or_default()
```

The compiler will probably complain about types and references, so worst case you would have to clone everything, which can also be cleaned up at a later point.

Swatinem:
I believe the reason you moved this into a function was to more easily run assertions on the results? Why not use the JSON output for that?

rust_snuba/src/processors/replays.rs
cmanallen: Yeah I didn't want to map the error to a serde type.  Anyhow didn't like `anyhow::Result<<Vec<ReplayRow>>>` for some reason.  It's saying "expected :: not >".  Which I'm not exactly sure where it wants the double colon but syntactically I didn't understand how to resolve in Rust.

untitaker: @cmanallen it's a syntax error, you have `<<` and `>>`, that doesn't work, need to replace with `<` and `>`

rust_snuba/src/processors/replays.rs
cmanallen: Yes you're right it did not need to be double wrapped.  Totally missed that.

rust_snuba/src/processors/replays.rs
cmanallen: It's still early lol.  Thank you.  Last night I was doing this `anyhow::Vec<ReplayRow>` which was failing because I didn't have the `Result` wrapping.  Which I didn't realize until 2 minutes ago when I left that comment and I still made an error... good grief.

rust_snuba/src/processors/replays.rs
```diff
@@ -588,6 +914,92 @@ mod tests {
                 "type": "replay_event"
             }}"#
         );
+
+        let rows = deserialize_message(data.as_bytes(), 0, 0).unwrap();
+        let replay_row = rows.first().unwrap();
```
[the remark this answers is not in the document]
cmanallen: For me its hard to see the differences between the two structs.  With this method I know the exact field that failed.  Is there a way to get verbose output similar to pytest's `-vv`?

cmanallen:
@Swatinem 

> I believe the reason you moved this into a function was to more easily run assertions on the results? Why not use the JSON output for that?

In production code a `ReplayRow` should never deserialize input so I opt not to implement the `Deserialize` trait to enforce that constraint.  I don't want to implement it for the sake of the test suite.  I'd prefer to keep forbidden usage impossible.

rust_snuba/src/processors/replays.rs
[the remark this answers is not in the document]
untitaker: you can pull in the crate `pretty_assertions` in theory

untitaker (approved):

PR #5352, opened 2024-01-09
fix(rust): Actually count num_rows correctly

untitaker:
In https://github.com/getsentry/snuba/pull/5314 I introduced another
bug: Rowcount is always zero, as we always add 0 to 0.

This means:

1. bad metrics and logs
2. bad batching

need to be careful when rolling out


Swatinem (approved):
😅

PR #5353, opened 2024-01-09
fix: Remove any panics in threads

untitaker:
This causes a lot of double-erroring, and at least within
RunTaskInThreads we can carry forward errors normally. We still have to
remove panics in the main thread, but this should already cut down a few
duplicate errors we are seeing today.

### DIFF
--- a/rust_snuba/rust_arroyo/src/processing/strategies/run_task_in_threads.rs
+++ b/rust_snuba/rust_arroyo/src/processing/strategies/run_task_in_threads.rs
@@ -15,16 +15,17 @@ use crate::utils::metrics::{get_metrics, BoxMetrics};
 use crate::utils::timing::Deadline;
 
 #[derive(Clone, Debug)]
-pub enum RunTaskError {
+pub enum RunTaskError<TError>{
     RetryableError,
     InvalidMessage(InvalidMessage),
+    Other(TError),
 }
 
-pub type RunTaskFunc<TTransformed> =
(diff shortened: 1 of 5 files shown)


rust_snuba/rust_arroyo/src/processing/strategies/run_task_in_threads.rs
```diff
@@ -165,8 +169,17 @@ impl<TPayload, TTransformed: Send + Sync + 'static> ProcessingStrategy<TPayload>
                     Ok(Err(RunTaskError::RetryableError)) => {
                         tracing::error!("retryable error");
                     }
+                    Ok(Err(RunTaskError::Other(error))) => {
+                        // XXX: at some point we should extend the error return type of poll() to
+                        // pass those errors through to the stream processor
+                        let error: Box<dyn std::error::Error> = error.into();
+                        tracing::error!(error, "the thread errored");
+                        panic!("the thread errored");
```
lynnagara, line 177: Are you actually expecting less error logging with this PR? It looks like we're still double reporting on this line (iirc this error happens on every deploy)

untitaker: we're no longer crashing the thread, so tokio at the very least should no longer report an error from within itself. i should followup with a larger refactor where we propagate the error all the way to the outside

lynnagara (approved):

PR #5358, opened 2024-01-09
feat(MQL): Bump snuba-sdk to 2.0.18 and support arbitrary functions in MQL parser

enochtangg:
### Overview
In [snuba-sdk 2.0.18](https://github.com/getsentry/snuba-sdk/releases/tag/2.0.18), support for arbitrary functions were added in the MQL grammar. This change needs to be reflected in the MQL to LogicalQuery visitor class as well. This PR is responsible for adding visitors to handle arbitrary functions and creates the appropriate expressions for them in the AST.

### DIFF
--- a/tests/query/parser/test_mql_query.py
+++ b/tests/query/parser/test_mql_query.py
@@ -1374,6 +1374,259 @@ mql_test_cases = [
         "metrics",
         id="Select metric with filter for metrics dataset",
     ),
+    pytest.param(
+        'apdex(sum(`d:transactions/duration@millisecond`), 500){dist:["dist1", "dist2"]}',
+        {
+            "entity": "generic_metrics_distributions",
+            "start": "2021-01-01T00:00:00",
+            "end": "2021-01-02T00:00:00",
+            "rollup": {
+                "orderby": "ASC",
+                "granularity": 60,
+                "interval": None,
+                "with_totals": None,
+            },
+            "scope": {
+                "org_ids": [1],
+                "project_ids": [1],
+                "use_case_id": "transactions",
+            },
+            "limit": None,
+            "offset": None,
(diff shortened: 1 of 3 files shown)


tests/query/parser/test_mql_query.py
```diff
@@ -1374,6 +1374,259 @@
...
+            },
+        },
+        Query(
+            QueryEntity(
+                EntityKey.GENERIC_METRICS_DISTRIBUTIONS,
+                get_entity(EntityKey.GENERIC_METRICS_DISTRIBUTIONS).get_data_model(),
+            ),
+            selected_columns=[
+                SelectedExpression(
+                    "apdex",
```
evanh, line 1408: This isn't quite right, the outer selected expression should still be called "aggregate_value" to ensure that the results have consistent column naming.

enochtangg: good catch, updated

tests/query/parser/test_mql_query.py
```diff
@@ -1374,6 +1374,259 @@
...
+        Query(
+            QueryEntity(
+                EntityKey.GENERIC_METRICS_DISTRIBUTIONS,
+                get_entity(EntityKey.GENERIC_METRICS_DISTRIBUTIONS).get_data_model(),
+            ),
+            selected_columns=[
+                SelectedExpression(
+                    "aggregate_value",
+                    FunctionCall(
+                        None,
```
evanh, line 1410: So this still isn't quite right:

The inner expression of the `SelectedExpression` should also be aliased as `aggregate_value` (which will show up as `_snuba_aggregate_value` here in the test).

That's because if someone references the alias `aggregate_value` elsewhere in the query, right now it won't get the correct expression.

PR #5360, opened 2024-01-09
feat(rust): Port generic metrics to rust consumer

nikhars:
Port over the logic of all of the generic metrics processors (counters, sets, distributions and gauges) to rust. There is intentional duplication of logic between the processors. We tend to make changes to processors one at a time in order to reduce the blast radius of the change. To that end, it is better that there be some logic duplication in order to avoid scenarios where intentional change in one processor causes unintentional change in another.


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1173 @@
...
+        buffer.extend_from_slice(key.as_bytes());
+        buffer.extend_from_slice(value.as_bytes());
+    }
+
+    adler32(&*buffer).unwrap()
+}
+
+#[derive(Debug, Deserialize)]
+struct FromGenericMetricsMessage {
+    // TODO: We don't seem to be performing any validation on the version. Should we?
```
lynnagara, line 48: Nah, I think we are good. We check it's 2 in the schema https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-generic-metrics.v1.schema.json#L10-L12, and I think this was mostly relevant in the past when we were migrating.

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1173 @@
...
+    version: u8,
+    use_case_id: String,
+    org_id: u64,
+    project_id: u64,
+    metric_id: u64,
+    #[serde(rename = "type")]
+    type_: String,
+    timestamp: u64,
+    sentry_received_timestamp: u64,
```
lynnagara, line 57: the examples in sentry-kafka-schemas indicate this might be a float. e.g. https://github.com/getsentry/sentry-kafka-schemas/blob/255f56ee8e130f75bf0c0938db08cf57c8efb455/examples/snuba-generic-metrics/1/snuba-generic-metrics1.json#L19.

I think it'd be best to handle both - we wouldn't want to drop the message on account of this

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1173 @@
...
+    tags: BTreeMap<String, String>,
```
lynnagara, line 58: Are the tags not indexed?

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1173 @@
...
+            tags_indexed_value: vec![0; tag_keys.len()],
+            tags_raw_value: tag_values,
+            count_value,
+            materialization_version: 2,
+            timeseries_id,
+            granularities,
+            min_retention_days: None,
+            decasecond_retention_days: None,
+            hr_retention_days: None,
+            day_retention_days: None,
```
lynnagara, line 172: What's all this retention days stuff? I don't see similar in the python version of the processor

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1173 @@
...
+                sentry_received_timestamp,
+            })
+        }
+        Err(err) => {
+            // If we get an error here, we should check if it is because of an unsupported
+            // metric type. If so, we should call InsertBatch::skip. Otherwise, we should
+            // return the original error.
+            if err.to_string().contains("Unsupported metric type") {
+                return Ok(InsertBatch::skip());
+            }
```
lynnagara, line 208: how about returning an `Option<CountersRawRow>` instead and not treating this case as an error

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1173 @@
...
+            Err(err)
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Default)]
+struct DistributionsRawRow {
```
lynnagara, line 346: Are the structs identical? If so, can we re-use. If they are slightly different and you want to keep separate, how about moving just the common fields into a shared parent struct?

rust_snuba/src/processors/generic_metrics.rs
nikhars: Changed sentry_received_timestamp to be an enum which can be either u64 or f64.

rust_snuba/src/processors/generic_metrics.rs
nikhars: The tag keys and tag values both seem to be string in the pipeline. https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-generic-metrics.v1.schema.json#L35-L37

rust_snuba/src/processors/generic_metrics.rs
nikhars: Good suggestion. Made the change.

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1138 @@
...
+    #[serde(default)]
+    distribution_values: Vec<f64>,
+}
+
+impl TryFrom<FromGenericMetricsMessage> for CountersRawRow {
+    type Error = anyhow::Error;
+
+    fn try_from(from: FromGenericMetricsMessage) -> anyhow::Result<CountersRawRow> {
+        if from.r#type != "c" {
+            return Err(anyhow::anyhow!("Unsupported metric type: {}", from.r#type));
```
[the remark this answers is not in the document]
nikhars: Good idea. Going to rework the implementation around this idea.

rust_snuba/src/processors/generic_metrics.rs
nikhars: The clickhouse tables basically have a way to customize retention of different granularities. There are defaults in some of the fields. But those can be overriden. Example here https://github.com/getsentry/snuba/blob/master/snuba/datasets/metrics_messages.py#L106-L112.

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -0,0 +1,1138 @@
...
+        last: f64,
+        max: f64,
+        min: f64,
+        sum: f64,
+    },
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+enum SentryReceivedTimestamp {
```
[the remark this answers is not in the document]
nikhars: My initial thought was that we would only receive u64. But Lyn correctly pointed out that the schema examples had a float in there. Reference here https://github.com/getsentry/snuba/pull/5360#discussion_r1446774263

rust_snuba/src/processors/generic_metrics.rs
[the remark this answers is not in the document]
nikhars: Changed the behavior as suggested.

untitaker (commented):
please add the processor to `tests/consumers/test_message_processors.py` for additional testing

rust_snuba/src/processors/generic_metrics.rs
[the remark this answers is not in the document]
phacops: You only need to parse a `f64` here. JSON doesn't really have an integer anyway, everything is a float, so it works.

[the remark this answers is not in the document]
nikhars: changed it to f64

nikhars:
> please add the processor to `tests/consumers/test_message_processors.py` for additional testing

Done. Had to change the python processor and the tests a bit in order to make things work.

rust_snuba/src/processors/generic_metrics.rs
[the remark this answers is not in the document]
untitaker: I now tested it locally, and indeed serde deserializes integers into floats without error.

I wasn't sure because Python certainly distinguishes between integer vs float:

```
>>> import json
>>> json.loads("123.0")
123.0
>>> json.loads("123")
123
>>>
```

snuba/datasets/metrics_messages.py
```diff
@@ -78,6 +78,7 @@ def value_for_counter_message(message: Mapping[str, Any]) -> Mapping[str, Any]:
         value, (int, float)
     ), f"{ILLEGAL_VALUE_IN_COUNTER} {INT_FLOAT_EXPECTED}: {value}"
 
+    value = float(value) if isinstance(value, int) else value
```
untitaker, line 81: unconditional `float(value)` should not be harmful and probably faster

untitaker (approved):

PR #5362, opened 2024-01-10
feat(rust): Cogs recording utility for generic metrics

lynnagara:


rust_snuba/Cargo.toml
```diff
@@ -45,6 +45,11 @@ tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter", "json"] }
 uuid = "1.5.0"
 parking_lot = "0.12.1"
+sentry_usage_accountant = "0.0.3"
+
+
+[patch.crates-io]
+rdkafka = { git = "https://github.com/getsentry/rust-rdkafka", branch = "base-consumer", features = ["cmake-build", "tracing"] }
```
fpacifici, line 52: Is this solving the dependency issue we saw yeaterday? (dependency on rdkafka 0.29.0)

rust_snuba/src/processors/utils.rs
```diff
@@ -39,3 +40,38 @@ where
...
+            accountant: UsageAccountant::new_with_kafka(config, None, None),
+        }
+    }
+
+    fn record_bytes(&mut self, resource_id: &str, app_feature: &str, amount_bytes: u64) {
+        if let Err(err) =
+            self.accountant
+                .record(resource_id, app_feature, amount_bytes, UsageUnit::Bytes)
+        {
+            tracing::warn!(?err, "error recording cogs");
```
fpacifici, line 63: I think it would be a good idea to debounce this logs. 
Let's say that something goes wrong with the accountant. Not that there is much that can go wrong on the record method consistently, but anyway .... 
In that case you do not want to log a warning, potentially for each message you consume. The cogs accountant should never compromise the stability of the consumer itself.

rust_snuba/src/processors/utils.rs
```diff
@@ -39,3 +40,38 @@ where
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_record_cogs() {
+        let mut accountant = CogsAccountant::new("127.0.0.1:9092");
+        accountant.record_bytes("generic_metrics_processor_sets", "ddm", 100)
```
fpacifici, line 75: fyi, the app_feature for ddm is `custom`

rust_snuba/Cargo.toml
untitaker: yes, i recommended this. it makes the dependency resolver pretend that "rdkafka from crates.io" actually comes from this repo. this is useful because cogs recorder depends on "rdkafka from crates.io", not "rdkafka from random git repo", so without this change cargo will attempt to build two separate versions of rdkafka, one for the cogs recorder, and one for the application. normally this works out great, particularly for things like `serde` or `log`, but for a library linking to a C++ codebase it plainly breaks at buildtime.

rust_snuba/src/processors/utils.rs
untitaker: i feel that this pattern (and the fact that we use it aggressively at sentry) is a holdover from the trauma of writing performance sensitive software in Python, and particularly the fact that the GIL prevents us from offloading instrumentation overhead fully off to another thread.

we are also in a situation where "get the current system time" is one of the most expensive things we do on the main thread, so debouncing instead of just writing to stdout might harm performance.

empirically we know that it does not harm stability, because we did deploy with `println()` statements accidentally left in (per-poll, per-message) and also ran our experimental consumers with log-level=debug at some point

we should have the opportunity to question whether we are performing this debouncing at the right layer of abstraction. I feel that for printing logs, we already get the current system time there, and so ideally we'd be able to debouce in the log stdout handler if we really have to.

rust_snuba/src/processors/utils.rs
```diff
@@ -39,3 +40,38 @@ where
...
+impl CogsAccountant {
+    #[allow(dead_code)]
+    fn new(bootstrap_servers: &str) -> Self {
+        let config = KafkaConfig::new_producer_config(bootstrap_servers, None);
+        Self {
+            accountant: UsageAccountant::new_with_kafka(config, None, None),
+        }
+    }
+
+    fn record_bytes(&mut self, resource_id: &str, app_feature: &str, amount_bytes: u64) {
```
untitaker, line 58: how will this be used -- is it going to be per-batch in snuba, or per-message?

I feel that batches are fairly homogenous when it comes to COGS and it should be possible to call it once per batch.

if it is per-message, perhaps we should attach COGS to `BytesInsertBatch` and only record them in clickhouse.rs? that would inherently be more performant and also address filippo's logging concern

lynnagara: Per message - the accountant library takes care of batching and flushing periodically so calling record is really cheap.

rust_snuba/src/processors/utils.rs
lynnagara: Alright, I changed this to only log to sentry once. Once it has logged, it will never log again until the next deployment (or rebalance).

This way we do not harm performance by getting the current system time. At the same time, we do not entirely miss warnings. Once is enough to tell us if something is wrong.

rust_snuba/src/processors/utils.rs
lynnagara: ok may as well update to reflect reality

rust_snuba/src/processors/utils.rs
lynnagara: though actually the library does call `Utc::now()` on every message so might still affect perf

untitaker (approved):

PR #5367, opened 2024-01-10
feat(spans): Set origin_timestamp for spans

phacops:
We're now setting a `received` field for a span message: https://github.com/getsentry/relay/blob/master/relay-server/src/actors/store.rs#L1168


untitaker (approved):

untitaker:
this made `received` required, so it broke sentry

PR #5370, opened 2024-01-10
fix(ci): Add ddm_meta test to "full tests"

untitaker:
The current set doesn't cover anything in relation to metrics/spans, and
it resulted in a few bad PR merges related to INC-604

Add this file since it failed in almost all of those PRs and doesn't
take long to run.


phacops:
I would add this file as well:
```
tests/sentry/api/endpoints/test_organization_profiling_functions.py
```

dbanda (approved):

PR #5375, opened 2024-01-10
feat(metrics-summaries): Parse only the minimum to see if we need to skip the message

phacops:


rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -34,27 +44,39 @@ pub fn process_message(
...
+        origin_timestamp,
+        sentry_received_timestamp: None,
+    })
 }
 
 #[derive(Debug, Default, Deserialize)]
 struct InputMessage {
-    #[serde(default)]
-    _metrics_summary: BTreeMap<String, Vec<FromMetricsSummary>>,
+    _metrics_summary: Option<Box<RawValue>>,
```
untitaker, line 69: I believe [IgnoredAny](https://docs.rs/serde/latest/serde/de/struct.IgnoredAny.html) might fit your purpose even more narrowly -- it isn't specific to `serde-json`, and AFAIK there's fewer hacks in there because it doesn't allow you to access the inner value as a string and really just discards. who knows if it is faster though

rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -34,27 +44,39 @@ pub fn process_message(
...
+}
+
+#[derive(Debug, Default, Deserialize)]
+struct FromSpanMessage {
     #[serde(default)]
     duration_ms: u32,
+    #[serde(rename = "_metrics_summary")]
+    metrics_summary: BTreeMap<String, Vec<FromMetricsSummary>>,
     project_id: u64,
+    received: f64,
```
untitaker, line 79: so this mandatory field is new, right?

rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -1,25 +1,35 @@
...
     payload: KafkaPayload,
     _: KafkaMessageMetadata,
     config: &ProcessorConfig,
 ) -> anyhow::Result<InsertBatch> {
     let payload_bytes = payload.payload().context("Expected payload")?;
-    let from: InputMessage = serde_json::from_slice(payload_bytes)?;
+    let msg: InputMessage = serde_json::from_slice(payload_bytes)?;
 
+    if msg._metrics_summary.is_none() {
+        return Ok(InsertBatch {
```
untitaker, line 22: you can shorten it to `Ok(InsertBatch::default())`, or i think in the generic-metrics PR there's an alias `InsertBatch::skip()` as well

rust_snuba/Cargo.toml
```diff
@@ -37,10 +37,12 @@ rust_arroyo = { path = "./rust_arroyo" }
 sentry = { version = "0.32.0", features = ["anyhow", "tracing"] }
 sentry-kafka-schemas = "0.1.40"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "1.0" }
+serde_json = { version = "1.0", features = ["raw_value"] }
 thiserror = "1.0"
 tokio = { version = "1.19.2", features = ["full"] }
-statsdproxy = { version = "0.1.0", features = ["cadence-adapter"], git = "https://github.com/getsentry/statsdproxy", default-features = false }
+statsdproxy = { version = "0.1.0", features = [
```
untitaker, line 43: can you revert the formatting changes here? I feel that between various people, this file gets reformatted a lot.

Swatinem (commented):
TBH, I don’t think this is worth the complexity. Parsing a bunch of trivial number types should not be as bad for perf.
If it really is, might be better to just stop emitting such messages from the source?

phacops:
As pointed out, not worth the effort, I'll focus on emitting the summaries on a different topic.

PR #5376, opened 2024-01-10
feat(mql): Support single type formula queries

evanh:
Support single type formula queries using the -If suffix. This embeds all the
filters for a formula parameter into the SELECT clause, by amending the
aggregate with `-If` and collapsing all the filters into the new expression.

This also can handle nested formulas correctly.

Once `snuba-sdk==2.0.21` (https://github.com/getsentry/snuba-sdk/pull/159) is released I will update this to remove the `xfail` on the formula test.


enochtangg (commented):
The latest changes on `master` contains a change to the `InitialParseResult` class where the `aggregate` attribute is changed to `expression` to capture arbitrary functions. See: https://github.com/getsentry/snuba/blob/7cdfc86d2e34c36c3982c1675bf38b5eb1e33d52/snuba/query/mql/parser.py#L386-L395

Now that we are introducing formulas queries, do you think it makes sense to move the arbitrary function into `InitialParseResult.formula`?

evanh:
> Now that we are introducing formulas queries, do you think it makes sense to move the arbitrary function into InitialParseResult.formula?

That's an interesting idea. I'll merge your changes and see if that works.

snuba/query/mql/parser.py
```diff
@@ -49,18 +49,27 @@
...
 logger = logging.getLogger("snuba.mql.parser")
 
 
 @dataclass
 class InitialParseResult:
     expression: SelectedExpression | None = None
-    groupby: List[SelectedExpression] | None = None
-    conditions: List[Expression] | None = None
+    formula: str | None = None
+    parameters: list[InitialParseResult] | None = None
```
enochtangg, line 60: If parameters are present, is it guaranteed that formula also exists? If so, do you think it makes sense to codify that? E.g.
```
@dataclass
class InitialParseResult:
    expression: SelectedExpression | None = None
    formula: ParseFormula | None = None

@dataclass
class ParseFormula
    name: str
    parameters: list[InitialParseResult]
```

evanh: I don't think it's guaranteed, we could have a Formula with no parameters. I think keeping things in a single structure keeps things a little simpler for now. Particularly the typing.

enochtangg (approved):

PR #5380, opened 2024-01-11
feat(replays): Replace python processor with a rust-based processor

cmanallen:
**Python to Rust Behavioral Changes:**

- Processor output is now exhaustive.
  - Every field is specified.
  - The previous implementation would omit fields allowing the database to default.
  - Now the processor explicitly defines the default value.
- Tags are no longer accepted as objects. They are exclusively accepted as list of list of string.
  - This aligns with the type Relay provides.
  - Tags are not provided in any other format.
- `datetime` formatted timestamps are no longer accepted.
  - f64 (and smaller) are accepted but will be coerced to u32.
  - This aligns with the schema provided by Relay and other producers.
- Dynamically typed fields have been removed. In most cases, we will no longer implicitly coerce integer values to string.
  - Types must be correctly provided.
  - This is current behavior for our producers.
  - Exception: The user.id field.

**Rust Processor Changes:**

- Event link ids are validated to ensure at least one non-zero UUID was provided.
- The transaction tag is no longer redundantly stored.
- `user_id` field was renamed to `id` to reflect the schema received from Relay.

Most of the test coverage that was dropped was unit coverage for Python helpers.  There was only one test which did not make it and it was asserting implicit coercion of integers to strings.  This is not used and not desired going forward.


untitaker:
> Python to Rust Behavioral Changes

please ensure that sentry CI passes, and if a test there fails, add it here: https://github.com/getsentry/snuba/blob/7cdfc86d2e34c36c3982c1675bf38b5eb1e33d52/.github/workflows/ci.yml#L359-L365

cmanallen:
@untitaker I have a feeling all of them will fail due to the tags change.  There will be a Sentry PR to fix those (since the existing python processor accepts list[list[str]]) that should be merged before this PR.  But I will add any non-backwards compatible to the workflow 👍

lynnagara:
There is an error we are seeing in the replays test consumer running the rust processor relating to the IP address syntax (SNUBA-44P). This error is not present in the existing python processor. Is this field being handled differently between the 2?

lynnagara:
As discussed offline, we won't merge this PR until replays consumer is deployed to prod. As we don't want to use this hybrid approach in production.

cmanallen:
Skipped the hybrid processor and went straight to rust 🥇

untitaker:
@cmanallen I think we should still do this and get rid of the python processor. otherwise the tests in sentry will continue to use the python processor

cmanallen:
@untitaker Keep this PR exclusively to maintain the python test coverage?  The rust processor has the same coverage and more.  I would argue the rust coverage is better.  There were some (positive) changes on the Python to Rust transition which makes the test suites somewhat incompatible.

cmanallen:
@untitaker I may have confused myself.  Its my understanding the current production processor is not using the Python processor at all.  Its all rust.  Is the opposite true?  The current production implementation is the hybrid processor?

cmanallen:
Handled offline.  Dropping the python coverage and keeping the python processor to support the Sentry test suite.

cmanallen:
@untitaker @lynnagara This sat idle for a while.  Since the rust consumer has been running in prod for a while (three months??) we should merge this so I don't have to maintain the python processor any more.

All tests pass in the Sentry repo.

lynnagara (approved):

volokluev (approved):

PR #5381, opened 2024-01-11
fix(docker): Actually stop publishing a multiarch image

untitaker:
Publishing a manifest with just one architecture (x86) actually somehow
prevents colima on MacOS from finding a suitable image, even though it is capable of x86 emulation.

Let's just re-tag the x86 image instead of wrapping it in a manifest and hope that
works.

followup to #5365


evanh (approved):

asottile-sentry:
here's a speedier version of this too -- https://github.com/getsentry/snuba/pull/5382

PR #5382, opened 2024-01-11
ref: speed up .github/workflows/image.yml

asottile-sentry:
makes the build basically instant when it hits the cache


untitaker:
oh shit, 9s instead of 5 mins

untitaker (approved):

PR #5388, opened 2024-01-12
ref: Upgrade pre-commit, fix PYTHONPATH hacks

untitaker:
bug to repro:

1. clone the snuba repo
2. run `direnv allow` to create virtualenv and pre-commit hooks
3. verify pre-commit works by running `pre-commit run --files  snuba/cogs/accountant.py`
4. `rm -rf .venv` (to deal with a python upgrade, for example)
5. run `direnv allow` again to recreate
6. `pre-commit` is broken, the command from step 3 breaks with errors like `Executable check-case-conflict not found`

~The PYTHONPATH was supposed to fix _something_, but I can't recall now what it is. `cargo test` works fine. If it resurfaces we could re-introduce it as `SNUBA_PYTHONPATH` and set it similarly to how we set `SNUBA_PYTHONEXECUTABLE`, if we don't have any better ideas.~ it happened again on my machine, so I've renamed everything to be prefixed with `SNUBA_TEST_`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ develop: install-python-dependencies install-brew-dev install-rs-dev setup-git
 
 setup-git:
 	mkdir -p .git/hooks && cd .git/hooks && ln -sf ../../config/hooks/* ./
-	pip install 'pre-commit==2.18.1'
+	pip install 'pre-commit==3.6.0'
 	pre-commit install --install-hooks
 
 test:
--- a/rust_snuba/src/strategies/python.rs
+++ b/rust_snuba/src/strategies/python.rs
@@ -281,21 +281,9 @@ mod tests {
     use super::*;
     use rust_arroyo::testutils::TestStrategy;
 
-    fn set_sys_executable() {
-        let python_executable = std::env::var("SNUBA_PYTHONEXECUTABLE").unwrap();
-        Python::with_gil(|py| -> PyResult<()> {
-            PyModule::import(py, "sys")?.setattr("executable", python_executable)?;
-            Ok(())
-        })
-        .unwrap();
-    }
-
     #[test]
-    // test is flaky in CI and fails 100% of the time locally with " signal only works in main
-    // thread of the main interpreter"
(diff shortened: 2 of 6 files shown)


asottile-sentry (approved):

Makefile
```diff
@@ -13,7 +13,7 @@ develop: install-python-dependencies install-brew-dev install-rs-dev setup-git
 
 setup-git:
 	mkdir -p .git/hooks && cd .git/hooks && ln -sf ../../config/hooks/* ./
-	pip install 'pre-commit==2.18.1'
+	pip install 'pre-commit==3.6.0'
```
untitaker, line 16: this was probably not necessary but let's just do it for good measure

rust_snuba/src/strategies/python.rs
```diff
@@ -282,7 +282,7 @@ mod tests {
     use rust_arroyo::testutils::TestStrategy;
 
     fn set_sys_executable() {
-        let python_executable = std::env::var("PYTHONEXECUTABLE").unwrap();
+        let python_executable = std::env::var("SNUBA_PYTHONEXECUTABLE").unwrap();
```
untitaker, line 285: apparently PYTHONEXECUTABLE is already a used envvar name for some pip (or python? or virtualenv?) internals

untitaker:
cargo test just broke again locally for me which I think is related to this PR. the offending test is:

```
failures:

---- runtime_config::tests::test_runtime_config stdout ----
thread 'runtime_config::tests::test_runtime_config' panicked at src/runtime_config.rs:24:27:
called `Result::unwrap()` on an `Err` value: ModuleNotFoundError: No module named 'snuba'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    runtime_config::tests::test_runtime_config
```

gonna take a closer look at this before I merge

evanh (approved):
Pending test fix this looks good. Thanks!

lynnagara (approved):

PR #5389, opened 2024-01-12
fix(rust-consumers): Define consumer group for functions consumer

Zylphrex:
The consumer group is a required arg here otherwise the consumer will not start crashing all of dev server.


untitaker (approved):
yikes, sorry!

PR #5390, opened 2024-01-12
add rustfmt to precommit

untitaker:


dbanda (approved):

.pre-commit-config.yaml
```diff
@@ -45,5 +45,13 @@ repos:
         additional_dependencies: [ 'fastjsonschema', 'pyyaml', 'sentry_sdk' ]
         pass_filenames: false
         files: 'snuba/datasets/configuration/.*'
+  - repo: local
+    language: system
+    hooks:
+      - id: cargo-fmt
+        name: format rust code
+        language: system
+        entry: make format-rust
```
asottile-sentry, line 54: this will call things like `make format-rust rust_snuba/foo.rs rust_snuba/bar.rs` which is probably not what you want

untitaker: https://github.com/getsentry/snuba/pull/5404

PR #5392, opened 2024-01-12
feat(rust-snuba): Validate schemas prior to running python processors

cmanallen:


rust_snuba/src/factory.rs
```diff
@@ -156,3 +177,53 @@ impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
...
 }
+
+#[derive(Clone)]
+struct SchemaValidator {
+    schema: Option<Arc<Schema>>,
+    enforce_schema: bool,
+}
+
+impl SchemaValidator {
+    async fn process_message(
```
lynnagara, line 188: A lot of this code is duplicated from the rust processor, is it possible to share some of the implementation?

rust_snuba/src/strategies/processor.rs
```diff
@@ -187,6 +139,72 @@ impl TaskRunner<KafkaPayload, BytesInsertBatch, anyhow::Error> for MessageProces
...
+        return Err(maybe_err);
+    };
+
+    if let Err(error) = validate_schema(schema, enforce_schema, payload) {
+        let error: &dyn std::error::Error = &error;
+        tracing::error!(error, "Failed schema validation");
+        return Err(maybe_err);
+    };
+
+    Ok((msg.clone(), payload.clone()))
```
cmanallen, line 172: @untitaker These are the clones I was referring to.

rust_snuba/src/factory.rs
cmanallen: Per message.

rust_snuba/src/factory.rs
```diff
@@ -156,3 +176,29 @@ impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
...
+        self,
+        message: Message<KafkaPayload>,
+    ) -> Result<Message<KafkaPayload>, RunTaskError<anyhow::Error>> {
+        extract_and_validate(&message, &self.schema, self.enforce_schema)?;
```
untitaker, line 191: here you are not using the return value at all. is it possible to either:

1. stop returning the message from `extract_and_validate`, eliminating the clone

or:

1. changing the function to take `Message` instead of `&Message`

?

cmanallen: I agree with the first change that we could exit early.  But a memory copy before calling into Python is immaterial.  My primary concern is on the rust validation side where you do want the return value and the memory copy has significant cost.

You can't pass ownership to `extract_and_validate` because the `MessageProcessor` uses `message` after the function call.

untitaker: ok, I suggest `extract_and_validate` change its return type to `Result<(), _>`, and the rust processor just still contains all the boilerplate it used to have to extract the message a second time. I don't think that overhead is significant and might even be optimized away.

cmanallen: In my mind I'm going a different direction.  I think we should revert this commit.  If the rust side is running the same code twice you might as well not run it with the `extract_and_validate` function.  And if you have two functions with duplicated code you might as well remove the `extract_and_validate` function and move it to the `SchemaValidator` struct.  We get the best performance for both and the same amount of duplicate code.

untitaker: I also believe that is better design, because `extract_and_validate` technically does not need to have a BrokerMessage for schema validation

untitaker: >  I think we should revert this commit.

If we keep schema validation duplicated across python vs rust consumers there are two code locations where we report errors to sentry, and this means two places to iterate on instrumentation.

I would prefer the actual call to schema validation and the error reporting to still be consolidated. If you want to avoid duplication of the extraction in Rust (which I think is not necessary because once the function is inlined at compile time, who knows if it still is duplicated), you could change `extract_and_validate` to take a `&[u8]` instead of a kafka message

rust_snuba/src/strategies/processor.rs
```diff
@@ -187,6 +139,72 @@ impl TaskRunner<KafkaPayload, BytesInsertBatch, anyhow::Error> for MessageProces
...
+            )))
+        }
+    };
+
+    let maybe_err = RunTaskError::InvalidMessage(InvalidMessage {
+        partition: msg.partition,
+        offset: msg.offset,
+    });
+
+    let kafka_payload = &msg.payload.clone();
```
untitaker, line 161: btw this clone is also problematic IMO, but i see it exists in main as well

untitaker (approved):

PR #5396, opened 2024-01-15
fix(spans): Handle null values in measurements

phacops:
For some reason, the spans consumer has to handle `null` values for `measurements`.

Note that the schema doesn't allow that: https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-spans.v1.schema.json#L184-L195


evanh (approved):

enochtangg (approved):

mitsuhiko (approved):

phacops:
This is not needed.

PR #5398, opened 2024-01-16
feat(accountant): Add accountant into processing

nikhars:
This PR adds the accountant during the processing strategy flow. This is not fully functional yet. And its very rough.


rust_snuba/src/processors/mod.rs
```diff
@@ -48,6 +52,45 @@ define_processing_functions! {
...
 }
 
+lazy_static! {
+    static ref COGS_MAP: HashMap<&'static str, COGSLabel> = {
+        let mut m = HashMap::new();
+        m.insert(
+            "GenericCountersMetricsProcessor",
+            COGSLabel {
+                resource_id: "generic_metrics_processor_counters".to_string(),
+                app_feature: "counter".to_string(),
```
lynnagara, line 62: what's the `app_feature` for? And why is it `counter`? Isn't this supposed to be stuff like "custom" (for ddm), "release health", etc?

rust_snuba/src/processors/mod.rs
```diff
@@ -48,6 +52,45 @@ define_processing_functions! {
     ("GenericGaugesMetricsProcessor", "snuba-generic-metrics", generic_metrics::process_gauge_message),
 }
 
+lazy_static! {
```
lynnagara, line 55: I think this crate isn't really needed, you could probably just write this as a match statement on the string literals

rust_snuba/src/processors/mod.rs
nikhars: yes. will fix that part

rust_snuba/src/processors/mod.rs
nikhars: the problem is that hashmap needs to get defined on the heap. so the hashmap needs to get built during runtime. then the elements in the hashmap needs to have the correct lifetimes annotated else rust was complaining about it. 

I looked into how this gets generally solved for rust and there are 2 options. lazy_static or defining a macro. decided to use lazy_static since its already a dependency which is getting pulled in as a transient dependency.

rust_snuba/src/processors/mod.rs
nikhars: with the right fix for `app_feature` it now even harder to get the right label in the message processor unless the interface of processing function changes. the use_case_id is only known within the processing function. 

i think the whole model of accounting has to be rethought

rust_snuba/src/processors/mod.rs
untitaker: you want to use [`OnceCell`](https://doc.rust-lang.org/core/cell/struct.OnceCell.html) here, it is the successor to lazy_static

rust_snuba/src/processors/functions.rs
```diff
@@ -15,35 +15,35 @@ pub fn process_message(
     let payload_bytes = payload.payload().context("Expected payload")?;
     let msg: InputMessage = serde_json::from_slice(payload_bytes)?;
 
-    let functions = msg.functions.iter().map(|from| {
-        Function {
-            profile_id: msg.profile_id,
-            project_id: msg.project_id,
+    let functions: Vec<Function> = msg
```
untitaker, line 18: these changes seem unrelated

rust_snuba/src/types.rs
```diff
@@ -70,28 +92,28 @@ impl LatencyRecorder {
...
     pub origin_timestamp: Option<DateTime<Utc>>,
     pub sentry_received_timestamp: Option<DateTime<Utc>>,
+    pub cogs_data: Option<CogsData>,
 }
 
 impl InsertBatch {
-    pub fn from_rows<T>(rows: impl IntoIterator<Item = T>) -> anyhow::Result<Self>
+    pub fn from_rows<T>(
+        rows: impl IntoIterator<Item = T>,
+        origin_timestamp: Option<DateTime<Utc>>,
```
untitaker, line 106: i think at some point we said that we wanted to get rid of `InsertBatch::from_rows` actually, but why make those changes in this PR?

rust_snuba/src/processors/functions.rs
lynnagara: The idea was to use InsertBatch::from_rows everywhere as it's now the preferred method of constructing the InsertBatch

untitaker (approved):

nikhars:
Nice. Thanks for helping out with this @lynnagara

untitaker:
precommit etc are fixed on master

lynnagara:
@untitaker precommit is still failing on the merge commit

untitaker:
oops yes. it's fixed in https://github.com/getsentry/snuba/pull/5392

rust_snuba/src/processors/snapshots/rust_snuba__processors__tests__schemas@profiles-call-tree-0.snap
```diff
@@ -1,113 +1,115 @@
...
 expression: snapshot_payload
 ---
 [
-  {
```
untitaker, line 7: that's not the right fix. those files must not change, and I think they do because you're passing a nested array to `InsertBatch` instead of a flat one

PR #5403, opened 2024-01-16
fix(rust): Generic metrics serde implementation

nikhars:
Use serde's adjacently tagged enum representation to handle different metric types


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -1,6 +1,7 @@
 use adler::Adler32;
 use anyhow::Context;
 use chrono::DateTime;
+use core::result::Result::Ok;
```
untitaker, line 4: this should not be necessary

untitaker (approved):

PR #5407, opened 2024-01-16
fix(docker): Tag snuba:amd64-latest again

untitaker:
We are seeing some weirdness on snuba:latest on MacOS where either
pulling or starting fails, presumably because snuba:latest used to be a
multiarch image and no longer is. snuba:amd64-latest works

also add --platform tag on the off chance that fixes anything, since it
appears to be present in Relay's dockerfile


joshuarli (approved):
nice

PR #5410, opened 2024-01-17
feat(gcs): Allow snuba to write to gcs for tooling

nikhars:
Some of the tools which we are building for clckhouse upgrades have a need for persistent blob storage. Introduce mechanism to upload and download files from GCS.


MeredithAnya:
@nikhars where do we plan on saving the credentials file? does that already exist ops repo somewhere?

nikhars:
> @nikhars where do we plan on saving the credentials file? does that already exist ops repo somewhere?

It will be passed as an environment variable to the pod. That's the next step I am going to work on. And yes, it will be in the ops repo.

MeredithAnya (approved):

evanh (approved):

PR #5412, opened 2024-01-17
ref(rust): Remove rust-arroyo

untitaker:
This has moved over into the git repository getsentry/arroyo as of https://github.com/getsentry/arroyo/pull/326 -- will follow up with proper workflow


lynnagara (approved):
✂️

PR #5413, opened 2024-01-17
chore(on-call): Make QueryTooLong exception an invalid request

enochtangg:
Right now, queries that are too long (in character size) count against our SLO. These queries should be marked as an invalid request and ignored by SLO.


volokluev:
If I look at [this sentry issue](https://sentry.sentry.io/issues/4880782233/?project=1&query=is%3Aunresolved+github_comment&referrer=issue-stream&statsPeriod=24h&stream_index=0), it comes from [this line](https://github.com/getsentry/snuba/blob/2d5bfc6d26348d6b6fa2950d85c2f7b424c11609/snuba/web/query.py#L377-L391). Something that is worth checking:

If we throw an error there, do we even get to the SLO reporting stage? If we do then this PR looks good to me

enochtangg:
@volokluev You're right, we needed to update the query_metadata for the SLO reporting stage output the correct result. Added this.

evanh (approved):

volokluev (approved):

PR #5414, opened 2024-01-17
fix(inc-602): check_clickhouse_connections should not check all clusters always

volokluev:


snuba/cli/migrations.py
```diff
@@ -30,7 +33,7 @@ def list() -> None:
     Lists migrations and their statuses
     """
     setup_logging()
-    check_clickhouse_connections()
+    check_clickhouse_connections(CLUSTERS)
```
evanh, line 36: Doesn't this also have to filter CLUSTERS based on readiness state?

volokluev: This command does not take readiness states as an rgument and is only really used locally which is why I didn't bother.

evanh (approved):

PR #5419, opened 2024-01-18
feat(rust): Port PolymorphicMetricsProcessor to Rust

ayirr7:


ayirr7:
I think the next step is to start porting over the Release Health tests so we can verify the processor behavior

untitaker:
next steps IMO: add it in `processors/mod.rs`'s "declare" macro (and uncomment the `mod ` line), then you will see compilation errors and some tests will run out of the box

rust_snuba/src/processors/release_health_metrics.rs
```diff
@@ -0,0 +1,754 @@
...
+const GRANULARITY_ONE_DAY: u8 = 3;
+
+/// Generate a timeseries ID from the given parameters. Timeseries IDs are used to
+/// uniquely identify a timeseries in the database. This implemenation is based on
+/// the python implementation in order to ensure that the same timeseries ID is
+/// generated for the same input.
+///
+/// Upstream sends the tag keys and values as strings. Which is surprising since the
+/// tag keys are indexed so they should be integers. But we have to deal with it.
+fn generate_timeseries_id(
```
lynnagara, line 27: Since there's quite a bit of overlap between generic metrics and metrics, how about putting these common functions into a shared metrics utils file and using it from both sets of processors?

rust_snuba/src/processors/release_health_metrics.rs
```diff
@@ -0,0 +1,761 @@
...
+            generate_timeseries_id(from.org_id, from.project_id, from.metric_id, &from.tags);
+        let (tag_keys, tag_values): (Vec<_>, Vec<_>) = from.tags.into_iter().unzip();
+        let retention_days = enforce_retention(Some(from.retention_days), &config.env_config);
+
+        // The way the _should_process functionality works in the Python
+        // processor today is by matching against the InputType Enum
+        // Maybe we should look into porting that Enum into Rust?
+        match from.r#type.as_str() {
+            "s" => Ok(Some(MetricsRawRow {
+                use_case_id: from.use_case_id,
```
nikhars, line 121: A lot of the repetition between fields which are common between different message types can be abstracted away into a common struct. See [this](https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/generic_metrics.rs#L82-L107) for example. You can do something similar to make the code more concise.

nikhars (commented):
Good job. Added some comments.

rust_snuba/src/processors/release_health_metrics.rs
```diff
@@ -0,0 +1,761 @@
...
+    types::{InsertBatch, RowData},
+    KafkaMessageMetadata, ProcessorConfig,
+};
+
+use super::utils::enforce_retention;
+
+const GRANULARITY_TEN_SECONDS: u8 = 0;
+const GRANULARITY_ONE_MINUTE: u8 = 1;
+const GRANULARITY_ONE_HOUR: u8 = 2;
+const GRANULARITY_ONE_DAY: u8 = 3;
```
nikhars, line 19: These are probably not used in the release health pipeline. So you can get rid of them

john-z-yang: Fixed, thanks

rust_snuba/src/processors/release_health_metrics.rs
```diff
@@ -0,0 +1,761 @@
...
+    }
+}
+
+pub fn process_metrics_message(
+    payload: KafkaPayload,
+    metadata: KafkaMessageMetadata,
+    config: &ProcessorConfig,
+) -> anyhow::Result<InsertBatch> {
+    process_message(payload, metadata, config)
+}
```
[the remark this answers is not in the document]
john-z-yang: Sounds good, fixed

rust_snuba/src/processors/release_health_metrics.rs
```diff
@@ -0,0 +1,761 @@
...
+    timestamp: f64,
+    sentry_received_timestamp: Option<f64>,
+    tags: BTreeMap<String, u64>,
+    value: MetricValue,
+    retention_days: u16,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+enum MetricValue {
```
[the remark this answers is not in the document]
ayirr7: Applied this change, thanks

rust_snuba/src/processors/release_health_metrics.rs
lynnagara: I think there are probably a few ways. You could have a custom trait `ToBytes` or something, and a u64 and string implementation of that trait.

But I assumed the 2 functions were identical before, and given Nikhar's comments about maintenance mode it's probably better to keep things separate as they are now.

snuba/datasets/processors/metrics_bucket_processor.py
```diff
@@ -117,28 +114,12 @@ def process_message(
...
-
 def timestamp_to_bucket(timestamp: datetime, interval_seconds: int) -> datetime:
     time_seconds = timestamp.timestamp()
     out_seconds = interval_seconds * (time_seconds // interval_seconds)
     return datetime.fromtimestamp(out_seconds, timestamp.tzinfo)
+
+
+class PolymorphicMetricsProcessor(RustCompatProcessor):
+    def __init__(self) -> None:
+        super().__init__("PolymorphicMetricsProcessor")
```
lynnagara, line 125: Just to confirm - this change is just for development so we can run all of the python tests and will be reverted later?

rust_snuba/src/processors/release_health_metrics.rs
john-z-yang: > You could have a custom trait ToBytes or something, and a u64 and string implementation of that trait.

How do I do that exactly? I'm assuming the trait would look something like this:

```rust
trait ToBytes {
    fn generate(&self) -> ???; // What do I do here?
}
```

If it is
```rust
impl ToBytes for String {
    fn generate(&self) -> [u8];
}
```
Then wouldn't it require me to know the size of that String at compile time?

Otherwise with
```rust
impl ToBytes for u32 {
    fn generate(&self) -> &[u8];
}
```
It would require me to borrow the [u8] from somewhere?

snuba/datasets/processors/metrics_bucket_processor.py
john-z-yang: Yes, but do we not need to do this eventually if we want to bring this to prod?

rust_snuba/src/processors/release_health_metrics.rs
lynnagara: I see. Making it `Vec<u8>` seems like the easiest way to me then. But I agree it doesn't seem worth the trouble in this specific scenario.

snuba/datasets/processors/metrics_bucket_processor.py
lynnagara: Eventually yes. But only after we have migrated the consumers in prod. Doing it now before we have tested or migrated to Rust means the Python consumer would be calling the rust processor in prod today, and this path has never really been tried in production at all.

rust_snuba/src/processors/release_health_metrics.rs
john-z-yang: Thanks!

snuba/datasets/processors/metrics_bucket_processor.py
john-z-yang: Make sense, will remove it once all the tests pass!

rust_snuba/src/processors/release_health_metrics.rs
```diff
@@ -0,0 +1,706 @@
...
+            tags_value: tag_values,
+            materialization_version: 4,
+            timeseries_id,
+            partition: meta.partition,
+            offset: meta.offset,
+            ..match from.value {
+                MetricValue::Set(value) => MetricsRawRow {
+                    metric_type: "set".to_string(),
+                    set_values: Some(value),
+                    ..Default::default()
```
ayirr7, line 143: Does this work by putting in the default values for the `count_value` and `distribution_values` fields? Or does it basically not have those fields in the new message output?

rust_snuba/src/processors/release_health_metrics.rs
john-z-yang: Done, didn't use a common struct but did something similar,  please let me know if it looks good

rust_snuba/src/processors/release_health_metrics.rs
john-z-yang: The first, `Default::default()` puts `None` for `Option<T>`

snuba/datasets/processors/metrics_bucket_processor.py
untitaker: this change needs to be reverted

untitaker (changes requested):
just one comment, after that we can deploy into a new consumer group

untitaker:
there are still quite a few commented out lines of code that I don't know if we need them. the tests sure would be good to uncomment and fix.

snuba/datasets/processors/metrics_bucket_processor.py
john-z-yang: Done

john-z-yang:
> there are still quite a few commented out lines of code that I don't know if we need them. the tests sure would be good to uncomment and fix.

Ah, yeah those are from Nikhar's generic metrics processor. Will remove them and add some rust tests, we have been relying on the python tests via the `RustCompatProcessors` until now

untitaker:
approving because I 1) don't see anything that would break prod here 2) we can iron out everything else when running in no-write mode

up to you though if you want to test a bit more, I haven't reviewed the logic in great detail

nikhars (approved):

snuba/datasets/processors/metrics_bucket_processor.py
```diff
@@ -96,9 +96,12 @@ def process_message(
             "project_id": message["project_id"],
             "metric_id": message["metric_id"],
             "use_case_id": message.get("use_case_id", "sessions"),
-            "timestamp": timestamp,
+            "timestamp": int(timestamp.replace(tzinfo=timezone.utc).timestamp()),
```
ayirr7, line 99: @john-z-yang Should we revert this change? I thought Clickhouse expects a DateTime?

john-z-yang: Clickhouse client automatically converts datetime to ints, because that is how it is stored in the database itself
https://clickhouse.com/docs/en/integrations/python#write-formats
I prefer that we do not revert these changes because eventually we need to use the RustCompatProcessor and the tests would need to pass

PR #5424, opened 2024-01-19
workaround for multiple-insertion issue in metrics API tests

onewland:
The tests assume that data insertion runs once, but it actually runs once per test method in the test class. This works around it but we should probably understand how to better use fixtures for a cleaner solution

### DIFF
--- a/tests/test_metrics_api.py
+++ b/tests/test_metrics_api.py
@@ -44,6 +44,10 @@ def utc_yesterday_12_15() -> datetime:
     )
 
 
+have_generated_counters = False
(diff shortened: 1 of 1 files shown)


tests/test_metrics_api.py
```diff
@@ -44,6 +44,10 @@ def utc_yesterday_12_15() -> datetime:
     )
 
 
+have_generated_counters = False
```
untitaker, line 47: i'd put those on `self`

untitaker (approved):

PR #5428, opened 2024-01-19
feat(metrics-summaries): Handle incomplete metrics summaries

phacops:
Handle having missing fields or null values in metrics summaries.

This reflects what's in the schema better: https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-spans.v1.schema.json#L202-L223.

### DIFF
--- a/rust_snuba/src/processors/metrics_summaries.rs
+++ b/rust_snuba/src/processors/metrics_summaries.rs
@@ -68,7 +68,7 @@ struct FromSpanMessage {
     trace_id: Uuid,
 
     #[serde(default)]
-    #[warn(dead_code)]
+    #[allow(dead_code)]
     parent_span_id: String,
 }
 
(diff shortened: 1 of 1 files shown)


untitaker (approved):
do you want to replicate https://github.com/getsentry/snuba/pull/5402 for metrics-summaries? it will force you to align the json schema with the rust struct (assuming the schema is correct here)

untitaker (approved):

rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -158,4 +169,48 @@ mod tests {
         process_message(payload, meta, &ProcessorConfig::default())
             .expect("The message should be processed");
     }
+
+    #[test]
+    fn test_missing_values() {
+        let span = br#"{
```
untitaker, line 175: i think if you have a testcase like this it would be best in sentry-kafka-schemas, then it will be implicitly tested

PR #5435, opened 2024-01-19
feat(on-call): Add illegal aggregate function in conditions entity validator

enochtangg:
### Overview
Aggregation functions (e.g. `sum(duration)`) are being found in the WHERE clause of incoming queries to snuba. These should be put in the HAVING clause instead.  Example [sentry issue](https://sentry.sentry.io/issues/4875058423/?project=300688&query=start%3D2024-01-18T11%3A52%3A00%26end%3D2024-01-18T18%3A27%3A17%26groupStatsPeriod%3Dauto&referrer=alerts-related-issues-issue-stream).

This PR is responsible for adding a new built-in entity validator that checks if aggregation functions are being used in the conditions of a query. If an aggregation function expression is found, reject the query.

### Testing
Use runtime config `enable_illegal_aggregate_in_condition_validator` to enable validator.

### DIFF
--- a/snuba/datasets/pluggable_entity.py
+++ b/snuba/datasets/pluggable_entity.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from typing import Any, List, Mapping, Optional, Sequence
 
+from snuba import state
 from snuba.clickhouse.columns import Column
 from snuba.datasets.entities.entity_data_model import EntityColumnSet
 from snuba.datasets.entities.entity_key import EntityKey
@@ -27,6 +28,7 @@ from snuba.query.validation import FunctionCallValidator
 from snuba.query.validation.validators import (
     ColumnValidationMode,
     EntityContainsColumnsValidator,
+    IllegalAggregateInConditionValidator,
     QueryValidator,
 )
 from snuba.utils.schemas import SchemaModifiers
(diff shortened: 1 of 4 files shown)


snuba/datasets/pluggable_entity.py
```diff
@@ -65,13 +67,20 @@ class PluggableEntity(Entity):
...
                 self.validate_data_model or ColumnValidationMode.ERROR,
             )
         ]
+        enable_illegal_aggregate_validator = state.get_config(
+            "enable_illegal_aggregate_in_condition_validator", 0
+        )
+        assert enable_illegal_aggregate_validator is not None
+        if int(enable_illegal_aggregate_validator) == 1:
+            builtin_validators.append(IllegalAggregateInConditionValidator())
+        return builtin_validators
```
volokluev, line 83: I'd put the killswitch in the validator, not the entity code. That creates a separation of concerns where the entity is responsible for entity stuff and the processor is responsible for itself

snuba/query/validation/validators.py
```diff
@@ -434,3 +438,76 @@ def validate(self, query: Query, alias: Optional[str] = None) -> None:
...
+                function_name=Param(
+                    "aggregate",
+                    Or(self.aggregate_function_names),
+                ),
+                with_optionals=True,
+            ).match(expression)
+            if match is not None:
+                matches.append(match)
+
+            match = FunctionCallMatcher(
```
volokluev, line 482: As far as I understand, this will match on any boolean function. Why is it necessary?

tests/datasets/validation/test_illegal_aggregate_conditions_validation.py
```diff
@@ -0,0 +1,108 @@
...
+                            ),
+                        ),
+                    ),
+                    Literal(None, 42),
+                ),
+            ),
+        ),
+        id="aggregate function in nested condition 1",
+    ),
+    pytest.param(
```
volokluev, line 69: Let's do the negative case as well. Make sure it does not reject a query it's not supposed to

volokluev (commented):
Nice! one comment on my understanding and a test case we should include

snuba/datasets/pluggable_entity.py
enochtangg: Makes sense. Updated!

tests/datasets/validation/test_illegal_aggregate_conditions_validation.py
enochtangg: Added!

snuba/query/validation/validators.py
enochtangg: It matches any boolean function in order to traverse the conditions tree down to the aggregation function. Conditions are almost always nested and chained together with boolean conditions.

volokluev (approved):

PR #5444, opened 2024-01-23
fix(optimize): Reduce parallel optimize job cutoff time

nikhars:
When we introduced the parallel optimize job cutoff time, we made its value same as the overall optimize job cutoff time. As of April 2023, the optimize job would finish by 3am since the size of individual partitions was no more than 110GB. Today, the size of each partition is close to 180GB. With the bigger partitions, it takes longer for the optimize job to finish. Cap the amount of time the optimize job can run in parallel so that parallel optimizations are not running during peak hours.

Also, removed the jitter because I don't believe they are adding any value. Removing it would allow the parallel phase to run a bit longer during the starting phase of the job.

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -371,7 +371,7 @@ OPTIMIZE_MERGE_MIN_ELAPSED_CUTTOFF_TIME = 10 * 60  # 10 mins
 # merges larger than this will be considered large and will be waited on
 OPTIMIZE_MERGE_SIZE_CUTOFF = 50_000_000_000  # 50GB
 # Maximum jitter to add to the scheduling of threads of an optimize job
-OPTIMIZE_PARALLEL_MAX_JITTER_MINUTES = 30
+OPTIMIZE_PARALLEL_MAX_JITTER_MINUTES = 0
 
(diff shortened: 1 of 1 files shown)


onewland:
Is there a risk to this? That we'd have less finished merges after 6am tomorrow than the current behavior (even if that means lower load)?

fpacifici:
Have you considered running 3 jobs in the early hours ?

nikhars:
> Is there a risk to this? That we'd have less finished merges after 6am tomorrow than the current behavior (even if that means lower load)?

Yes there is a risk that it might happen. Few things to mitigate that risk.
1. Removed jitter in parallel phase so that gives us extra 30 minutes to optimize.
2. Start the optimize job 1 hour earlier so that the parallel phase gets to run for 1 hour extra.

nikhars:
> Have you considered running 3 jobs in the early hours ?

Yes, that is the next plan of action I have in mind. I would like to avoid staggering too many changes together. I prefer tuning some of the knobs we have today and observing how the system behaves. Once we find out system behavior, we can make further improvements.

dbanda (approved):

PR #5445, opened 2024-01-23
fix: Upgrading to Clickhouse 22.8

evanh:
In Clickhouse 22.8 and higher, if the query on a aggregate function -Merge doesn't have the same parameters as what was stored in the -State, the query will fail. Add a query processor that catches instances of this so that Sentry queries don't fail.

Also add an xfailing test for tuple conditions that is fixed in 23.8.


nikhars (approved):
🎉

PR #5448, opened 2024-01-24
feat(rust): Implement replacements

lynnagara:
- Message processors can be registered as one of 2 types - standard message processors than return `InsertBatch` only or replacement message processors that can return either InsertBatch or ReplacementData.
- There is a new `ProduceReplacements` strategy which is chained immediately after the transform step, and only for those processors which support replacement data. This strategy forwards inserts to the next step and produces the replacements. Subsequent steps in the pipeline are unchanged.
- There are currently no message processors that actually produce replacements since this is only relevant for the "errors" dataset


rust_snuba/src/consumer.rs
```diff
@@ -80,6 +80,8 @@ pub fn consumer_impl(
 
     // TODO: Support multiple storages
     assert_eq!(consumer_config.storages.len(), 1);
+
+    // TODO: Remove this once errors is implemented in Rust
```
untitaker, line 84: can't we remove it now?

rust_snuba/src/processors/mod.rs
```diff
@@ -61,6 +69,16 @@ pub fn get_cogs_label(processor_name: &str) -> Option<String> {
     }
 }
 
+pub fn get_processing_function_with_replacements(
```
untitaker, line 72: I think having a new kind of processor function here is tricky because it means that every test we have that tests all message processors generically has to be updated (such as `test_schemas` in this file)

is there no way to make all message processors return the same type?

rust_snuba/src/strategies/processor.rs
```diff
@@ -57,18 +130,24 @@ pub fn get_schema(schema_name: &str, enforce_schema: bool) -> Option<Arc<Schema>
...
 
 #[derive(Clone)]
-struct MessageProcessor {
+struct MessageProcessor<TResult: Clone, TNext: Clone> {
     schema: Option<Arc<Schema>>,
     enforce_schema: bool,
-    func: ProcessingFunction,
+    // Convert payload to either InsertBatch (or either insert or replacement for the errors dataset)
+    func:
+        fn(KafkaPayload, KafkaMessageMetadata, config: &ProcessorConfig) -> anyhow::Result<TResult>,
```
untitaker, line 138: why'd we get rid of the type alias here?

rust_snuba/src/consumer.rs
lynnagara: Since errors is not actually working properly yet, I figured it's best to leave it in

rust_snuba/src/processors/mod.rs
lynnagara: I tweaked things slightly. Processors still return two different types unfortunately, but I'm defining them in one spot now, and I updated `test_schemas` to test the two kinds of processors (with and without replacements). Let me know if you think this is an ok compromise.

rust_snuba/src/strategies/processor.rs
lynnagara: I changed it to a generic, as this can be either `ProcessingFunction` or `ProcessingFunctionWithReplacements` now

untitaker (approved):

PR #5452, opened 2024-01-24
perf: use `--release` when watching

anonrig:
It's useful to eliminate false positives when benchmarking and watching.


untitaker (commented):
if we want to do this then we should do it for `make build-rs-dev` as well, but i am somewhat cautious about the increased build time for regular snuba users

PR #5453, opened 2024-01-25
feat(metrics-summaries): Add segment-related columns to ease up querying

phacops:
As we've learned more about what we wanted out of the product, we need to adapt a bit what we store in `metrics_summaries`.

This PR will add useful IDs to be able to query spans more efficiently.

This was the proposal: https://www.notion.so/sentry/Metrics-Correlations-Data-Model-Proposal-32b0979b48094247bdb5c3e866d5730c .

### DIFF
--- a/snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
+++ b/snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
@@ -1,0 +1,55 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+storage_set_name = StorageSetKey.METRICS_SUMMARIES
+
+columns: List[Column[Modifiers]] = [
+    Column("segment_id", UInt(64)),
+    Column("duration_ms", UInt(32)),
+    Column("group", UInt(64)),
+    Column("is_segment", UInt(8)),
+]
+
+targets = [
+    (
+        OperationTarget.LOCAL,
+        "metrics_summaries_local",
+    ),
+    (
+        OperationTarget.DISTRIBUTED,
+        "metrics_summaries_dist",
+    ),
+]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=storage_set_name,
+                table_name=table_name,
+                column=column,
+                target=target,
+            )
+            for column in columns
+            for table_name, target in targets
+        ]
+
+    def backwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.DropColumn(
+                storage_set=storage_set_name,
+                table_name=table_name,
+                column_name=column.name,
+                target=target,
+            )
+            for column in columns
+            for target, table_name in targets
+        ]
--- a/snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
+++ b/snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
@@ -39,7 +39,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 target=target,
             )
             for column in columns
-            for table_name, target in targets
+            for target, table_name in targets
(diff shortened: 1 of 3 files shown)


snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
```diff
@@ -0,0 +1,55 @@
...
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+storage_set_name = StorageSetKey.METRICS_SUMMARIES
+
+columns: List[Column[Modifiers]] = [
+    Column("segment_id", UInt(64)),
+    Column("duration_ms", UInt(32)),
+    Column("group", UInt(64)),
```
iambriccardo, line 14: What is the `group`?

snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
```diff
@@ -0,0 +1,55 @@
...
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+storage_set_name = StorageSetKey.METRICS_SUMMARIES
+
+columns: List[Column[Modifiers]] = [
```
iambriccardo, line 11: Can we also add the `trace_id`?

nikhars:
For reference, you are adding 2 columns which will not compress well since they are practically UUIDs. Columns with UUIDs are the biggest columns on the spans table. Look at the uncompressed size in the image below which has been taken from 1 of the shards. group got compressed in the spans table since it is part of the order by key. In metrics summaries it would show the same behavior as segment_id i.e. would not compress. Wanted to share the insight around the cost of duplicating data in multiple tables. But ultimately its the products call whether to take the increased cost of duplicating data.

Screenshot 2024-01-25 at 12 58 16 PM

snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
phacops: `group` is the fingerprint of the span which we use in the `spans` primary key. It kind of looks like a transaction name but more specific and one group can map to multiple spans of the same sort (like the same query with different parameters).

snuba/snuba_migrations/metrics_summaries/0003_metrics_summaries_add_segment_id_duration_group_columns.py
phacops: It's already in there https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/metrics_summaries/0001_metrics_summaries_create_table.py#L18

phacops:
I understand the cost concern. We want to try the feature, we need these fields to try a different query pattern. We'll likely run that for the alpha and then see what happens. We can't make a decision based on "it might cost a lot" without even seeing how useful the feature is. We welcome ideas to make it more compact though.

On a side node, can't we compress the `group` as well on `metrics_summaries`, even if it's not part of the primary key?

nikhars:
I have an idea which might be worth exploring to see if it gives you enough performance gains.

The spans dataset is sorted by `groups`. What if you only added the `groups` column to the metric summaries dataset and not the `segment_id`, `trace_id` etc. Then you always make the query to get the required spans from the spans dataset with a filter clause on the `groups` column.

nikhars:
> I understand the cost concern. We want to try the feature, we need these fields to try a different query pattern. We'll likely run that for the alpha and then see what happens. We can't make a decision based on "it might cost a lot" without even seeing how useful the feature is. We welcome ideas to make it more compact though.
> 
> On a side node, can't we compress the `group` as well on `metrics_summaries`, even if it's not part of the primary key?

Sure, i will approve. if you wanted to compress `group` you would need to add that column to the order by key. Unfortunately that means complete table rewrite. And we don't support that today since it is a dangerous migration.

nikhars (approved):

nikhars (changes requested):
You won't be able to change the order by key like this.

phacops:
> You won't be able to change the order by key like this.

I'm not hoping to change it. Right now, this dataset exists on the `spans` cluster. We have a `metrics_summaries` cluster ready to be used and the dataset has not been created on it yet. So, the idea is to merge this with this modified order key so that, when the table is created on the new cluster, it has the correct order key.

Then, we'll migrate to the new cluster and will stop using the current dataset.

nikhars (approved):

PR #5455, opened 2024-01-25
fix(generic-metrics): Make timestamp in Rust processor optional to match schema

ayirr7:
There's a mismatch for `sentry_received_timestamp` against the relevant Kafka schema. All of the example JSON tests have the field though so I think that's why this wasn't caught. But it should be optional I believe.


untitaker:
Thinking about this more, where do those metrics come from that don't have this set? surely all metrics go through the indexer?

PR #5458, opened 2024-01-26
build: Use Rollup instead of Webpack

gggritso:
I did this as a proof-of-concept for another project and felt bad about throwing the code out. Opening this PR _just for posterity!_


gggritso:
🚮 and in case you're curious, no, Rollup did not build he project any faster than Webpack did!

PR #5472, opened 2024-01-30
ref(errors): Remove trace context from extra contexts

untitaker:
This data does not need to be stored twice.


untitaker:
i think we might actually use this for parent_span_id :(

PR #5473, opened 2024-01-30
fix(rate-limit): Remove dual read/write

evanh:
There was a plan to move the rate limit config keys to a new namespace, so they
could be semantically separated from other config keys. The start of that work
was to start dual writing config keys to the new namespace. However that
project got abandoned, and now the capacity management tool is handling this
anyways. Remove all the dual reads/writes.


volokluev (approved):
chefskiss

PR #5474, opened 2024-01-30
ref(ddm): Dual write to Sentry metrics as well as Datadog

evanh:
This PR wraps the Datadog and Sentry metrics into a single backend, so we can
start dogfooding Sentry metrics. Currently the Sentry portion is feature
flagged and has a sampling rate. 

Sentry also provides a way to specify the unit and record distributions, so add 
that to the abstract so we can start using it later.


snuba/utils/metrics/backends/dualwrite.py
```diff
@@ -0,0 +1,82 @@
...
+        name: str,
+        value: int | float,
+        tags: Tags | None = None,
+        unit: str | None = None,
+    ) -> None:
+        self.datadog.gauge(name, value, tags, unit)
+        if self._use_sentry():
+            self.sentry.gauge(name, value, tags, unit)
+
+    def timing(
```
untitaker, line 52: timings are distributions -- i don't think we need both. the implementation in the SDK is identical

evanh: They're not identical, because in the SDK they accept different units. The timing ones specifically only accept timing units, and distributions are more flexible.

untitaker: can we have a default impl for timing that uses distribution internally or the other way around? The impl really is identical in the SDK (the only thing that differs is type hints and return value) -- i get that you want to have different defaults for unit in each case

evanh: Why not both? Both exist in the Sentry SDK and in Datadog. Why not use the SDKs as intended?

untitaker: because in both SDKs they're literally just aliases and now you're forcing every backend to implement those aliases.

the purpose of `sentry_sdk.metrics.timing` is to provide a decorator and context manager to measure a codeblock -- we're not using that

evanh: > forcing every backend to implement those aliases.

This is true, but we are missing the `distribution` functionality from all our backends, so I would have to add that no matter what (timing is a subset of distribution). Why not leave timing in place, add distributions, and then our backends line up with both of the production SDKs (DD and Sentry).

> literally just aliases

If the aliases are already available, why not leverage them? Why in turn make timing _another_ alias on distribution? The DD and Sentry SDKs already provide that alias for us.

untitaker: > I would have to add that no matter what

you really don't, you could continue to use timing or rename it to distribution if you want to. timing is not a subset of distribution in master branch, it's just distribution with maybe an unfortunate (not-technically-correct) name. i don't see the point in adding more methods that do the same thing but only differ in their default for unit, but I don't feel too strongly about all of this so i'm approving

untitaker (approved):

dbanda:
Do we need to do something to blacklist the metrics consumer to avoid going into a loop or is this handled by s4s?

untitaker:
> Do we need to do something to blacklist the metrics consumer to avoid going into a loop or is this handled by s4s?

at this sample rate probably not -- but as we increase it I would override the sample rate to 0 for those particular deployments

untitaker:
to be clear, there's sample rate setting that can be overridden with envvars, I suggest to set that envvar for the particular consumers where we see this feedback loop effect, either set it back to 0 or just a lower value

PR #5477, opened 2024-01-31
ref(ddm): Dual write to Sentry metrics as well as Datadog

evanh:
This PR wraps the Datadog and Sentry metrics into a single backend, so we can
start dogfooding Sentry metrics. Currently the Sentry portion is feature
flagged and has a sampling rate.

Sentry also provides a way to specify the unit and record distributions, so add
that to the abstract so we can start using it later.

Recreated from  #5474, with bug fixes.

### DIFF
--- a/snuba/environment.py
+++ b/snuba/environment.py
@@ -87,6 +87,12 @@ def setup_sentry() -> None:
         release=os.getenv("SNUBA_RELEASE"),
         traces_sample_rate=settings.SENTRY_TRACE_SAMPLE_RATE,
         profiles_sample_rate=settings.SNUBA_PROFILES_SAMPLE_RATE,
+        _experiments={
+            # Turns on the metrics module
+            "enable_metrics": True,
+            # Enables sending of code locations for metrics
+            "metric_code_locations": True,
+        },
     )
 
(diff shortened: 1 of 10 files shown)


untitaker (approved):
what were the bugfixes?

evanh:
> what were the bugfixes?

Moving the `state` import inside the `_use_sentry` call. That was causing an import loop.

I plan to remove that check once we have things working. Just a quick failsafe for now.

untitaker:
highly concerning that this didn't fail any tests, not even sentry integration tests.

evanh:
> highly concerning that this didn't fail any tests, not even sentry integration tests.

I'm open to suggestions on this: we don't test our metrics integration, because it requires having a valid DD client. I did my best to test all the changes locally but I didn't think to check snuba admin.

untitaker:
> I'm open to suggestions on this: we don't test our metrics integration, because it requires having a valid DD client

in theory you can pass a bogus DSN and bogus statsd endpoint to datadog and sentry clients, and nothing bad should happen (just a bunch of logged network errors), so I'm not even sure we need to run tests with dummy backend

PR #5480, opened 2024-01-31
feat(metrics-summaries): Ingest new segment-related columns

phacops:
Once https://github.com/getsentry/snuba/pull/5453 is merged and clusters are migrated, we can start ingesting new fields.

### DIFF
--- a/snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
+++ b/snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
@@ -71,6 +71,9 @@ query_processors:
   - processor: UUIDColumnProcessor
     args:
       columns: [trace_id]
+  - processor: HexIntColumnProcessor
+    args:
+      columns: [span_id, segment_id, group]
   - processor: MappingOptimizer
     args:
       column_name: tags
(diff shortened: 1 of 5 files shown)


snuba/datasets/configuration/spans/storages/metrics_summaries.yaml
```diff
@@ -67,9 +71,6 @@ query_processors:
   - processor: UUIDColumnProcessor
     args:
       columns: [trace_id]
-  - processor: HexIntColumnProcessor
```
untitaker, line 70: was this change intentional? I'm not really sure, I think this changes the type visible on the query layer, so any query using span_id would've to be fixed.

phacops: This is in the storage configuration, so should only matter at write time. From what I understand, this will transform a string to a integer when the string is a hexadecimal value. In this case, we're already transforming the data in Rust so there's not need for that processor.

phacops: I thought this was only applicable at write time (and it already arrives as an integer), but it is also in charge of transforming the value back to an hexadecimal string.

I added the relevant columns instead of removing it.

evanh (approved):

nikhars (approved):

PR #5481, opened 2024-02-01
feat(capman): Referrer guard rail policy

volokluev:
To prevent a specific referrer from taking over all capacity, put a default limit on all referrers which can be overidden. 

Put it just on errors first to test it in production (that's the capacity we're most concerned with protecting)

### DIFF
--- a/snuba/query/allocation_policies/per_referrer.py
+++ b/snuba/query/allocation_policies/per_referrer.py
@@ -1,0 +1,137 @@
+from __future__ import annotations
+
+import logging
+
+from snuba.query.allocation_policies import (
+    AllocationPolicyConfig,
+    QueryResultOrError,
+    QuotaAllowance,
+)
+from snuba.query.allocation_policies.concurrent_rate_limit import (
+    BaseConcurrentRateLimitAllocationPolicy,
+)
+from snuba.redis import RedisClientKey, get_redis_client
+from snuba.state.rate_limit import RateLimitParameters
+
+DEFAULT_CONCURRENT_QUERIES_LIMIT = 22
+DEFAULT_PER_SECOND_QUERIES_LIMIT = 50
+
+rds = get_redis_client(RedisClientKey.RATE_LIMITER)
+
+logger = logging.getLogger("snuba.query.allocation_policy_per_referrer")
+
+_DEFAULT_MAX_THREADS = 10
+_PASS_THROUGH_REFERRERS = set(
+    [
+        "subscriptions_executor",
+    ]
+)
+
+
+class ReferrerGuardRailPolicy(BaseConcurrentRateLimitAllocationPolicy):
+    """
+    A policy to prevent runaway referrers from consuming too many queries. This concern is orthogonal to customer rate limits.
+
+    For example, a product team may push out a feature that sends 20 snuba queries every 5 seconds on the UI. In that case, that feature should break,
+    not all of clickhouse.
+
+    """
+
+    @property
+    def rate_limit_name(self) -> str:
+        return "referrer_guard_rail_policy"
+
+    def _additional_config_definitions(self) -> list[AllocationPolicyConfig]:
+        return super()._additional_config_definitions() + [
+            AllocationPolicyConfig(
+                name="default_concurrent_request_per_referrer",
+                description="""how many concurrent requests does a referrer get by default? This is set to a pretty high number.
+                If every referrer did this number of concurrent queries we would not have enough capacity
+                """,
+                value_type=int,
+                param_types={},
+                default=100,
+            ),
+            AllocationPolicyConfig(
+                name="referrer_concurrent_override",
+                description="""override the concurrent limit for a referrer""",
+                value_type=int,
+                param_types={"referrer": str},
+                default=-1,
+            ),
+            AllocationPolicyConfig(
+                name="referrer_max_threads_override",
+                description="""override the max_threads for a referrer, applies to every query made by that referrer""",
+                param_types={"referrer": str},
+                value_type=int,
+                default=-1,
+            ),
+        ]
+
+    def _get_max_threads(self, referrer: str) -> int:
+        thread_override = int(
+            self.get_config_value(
+                "referrer_max_threads_override", {"referrer": referrer}
+            )
+        )
+        return thread_override if thread_override != -1 else _DEFAULT_MAX_THREADS
+
+    def _get_concurrent_limit(self, referrer: str) -> int:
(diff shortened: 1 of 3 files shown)


snuba/query/allocation_policies/per_referrer.py
```diff
@@ -0,0 +1,123 @@
...
+    QuotaAllowance,
+)
+from snuba.query.allocation_policies.concurrent_rate_limit import (
+    BaseConcurrentRateLimitAllocationPolicy,
+)
+from snuba.redis import RedisClientKey, get_redis_client
+from snuba.state.rate_limit import RateLimitParameters
+
+DEFAULT_CONCURRENT_QUERIES_LIMIT = 22
+DEFAULT_PER_SECOND_QUERIES_LIMIT = 50
```
nikhars, line 17: These values don't seem to be getting used. Is that intentional?

nikhars (commented):
Where does this policy fall in terms of ordering amongst the different policies we have? Could you explain the hierarchy of config policies that would get applied when this new policy gets added. There is a referrer_project_override and a referrer_organization_override in the ConccurrentAllocationPolicy. How do those overrides play with this new policy.

snuba/query/allocation_policies/per_referrer.py
volokluev: nope, I'll remove those

volokluev:
> Where does this policy fall in terms of ordering amongst the different policies we have?

Policies are independent of each other. They run one after the other (all of them). If any policy rejects, the query is rejected, the lowest thread number among all the policies is chosen. The code is [here](https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L788) if you're curious


> Could you explain the hierarchy of config policies that would get applied when this new policy gets added. 

There is no hierarchy, they are independent. They all have to pass or the query does not go through

> There is a referrer_project_override and a referrer_organization_override in the ConccurrentAllocationPolicy. How do those overrides play with this new policy.

That policy limits specific customer,referrer combinations. If there's an override for that query's specific referrer and it's doing too many queries, the query overall will be rejected

snuba/query/allocation_policies/per_referrer.py
volokluev: Done

nikhars:
> @nikhars
> 
> > Where does this policy fall in terms of ordering amongst the different policies we have?
> 
> Policies are independent of each other. They run one after the other (all of them). If any policy rejects, the query is rejected, the lowest thread number among all the policies is chosen. The code is [here](https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L788) if you're curious
> 
> > Could you explain the hierarchy of config policies that would get applied when this new policy gets added.
> 
> There is no hierarchy, they are independent. They all have to pass or the query does not go through
> 
> > There is a referrer_project_override and a referrer_organization_override in the ConccurrentAllocationPolicy. How do those overrides play with this new policy.
> 
> That policy limits specific customer,referrer combinations. If there's an override for that query's specific referrer and it's doing too many queries, the query overall will be rejected

Ok makes sense. I am curious to know whether there is a specific reason for making a separate allocation policy altogether. Would it be possible to integrate the referrer based one in the existing ConcurrentAllocationPolicy? If not, can you share why?

As a user, I would get confused when I see those 2 separate allocation policies. If there is only a single concurrent allocation policy but different ways to override it (referrer, referrer_org, referrer_project etc.), i know exactly which one to choose. That would be more intuitive to me. Curious to see what you think about it.

volokluev:
I have spent many hours thinking about this and have considered exactly what you say. the main reason I stuck with this solution is the following: 

The policies address orthogonal concerns and so they are separate. The `ConcurrentRateLimitAllocationPolicy` which at this point is badly named and should probably be renamed to `CustomerConcurrentRateLimitAllocationPolicy` limits customer usage of clickhouse. 

The `ReferrerGuardRailPolicy` limits sentry developers' usage of clickhouse.  Its concern has nothing to do with what customers are doing and everything to do with how we (sentry) implemented our code. If I made those policies the same thing, the policy code would be much more complex while still having orthogonal concerns.

What you're describing is similar to what we had in the legacy system and those overrides were actually confusing as well ([doc](https://www.notion.so/sentry/Rethinking-Rate-Limits-107cf5674dfe47519e0eed076063ce9e?pvs=4#9a72fc52bc5a450fb25e7d03963da003))

nikhars:
> I have spent many hours thinking about this and have considered exactly what you say. the main reason I stuck with this solution is the following:
> 
> The policies address orthogonal concerns and so they are separate. The `ConcurrentRateLimitAllocationPolicy` which at this point is badly named and should probably be renamed to `CustomerConcurrentRateLimitAllocationPolicy` limits customer usage of clickhouse.
> 
> The `ReferrerGuardRailPolicy` limits sentry developers' usage of clickhouse. Its concern has nothing to do with what customers are doing and everything to do with how we (sentry) implemented our code. If I made those policies the same thing, the policy code would be much more complex while still having orthogonal concerns.
> 
> What you're describing is similar to what we had in the legacy system and those overrides were actually confusing as well ([doc](https://www.notion.so/sentry/Rethinking-Rate-Limits-107cf5674dfe47519e0eed076063ce9e?pvs=4#9a72fc52bc5a450fb25e7d03963da003))

I like the idea of the rename of the existing ConcurrentRateLimitAllocationPolicy like you suggested. That would make it intuitive which allocation policy to use for which use case. We can probably tackle that in a separate PR though.

nikhars (approved):
LGTM 💯

snuba/query/allocation_policies/per_referrer.py
```diff
@@ -0,0 +1,120 @@
...
+rds = get_redis_client(RedisClientKey.RATE_LIMITER)
+
+logger = logging.getLogger("snuba.query.allocation_policy_per_referrer")
+
+_DEFAULT_MAX_THREADS = 10
+
+
+class ReferrerGuardRailPolicy(BaseConcurrentRateLimitAllocationPolicy):
+    """
+    A policy to prevent runaway referrers from consuming too many queries. This concern is orthogonal to customer rate limits.
```
evanh, line 25: Something to address here: customers can absolutely trip this limit, if a single customer decides to abuse a specific endpoint.

So on the surface this doesn't seem orthogonal to customers at all. We have a separate policy to handle customer+referrer limits, but from this comment it isn't clear when to use one over the other.

evanh (approved):
CI still failing, and I left one request for clarification.

Looks good!

PR #5489, opened 2024-02-02
fix(capman): Fix invalid tenant ids exception in allocation policy

volokluev:
The concurrent allocation policy was raising an `AllocationPolicyViolation` in its application code in order to reject queries. Unfortunately those queries were not actually being rejected because the policy base class would fail open (as it should). 

https://sentry.sentry.io/issues/4456295342/?project=300688&query=is%3Aunresolved+AllocationPolicyViolation&referrer=issue-stream&statsPeriod=7d&stream_index=0

This PR:

* Creates a new exception `InvalidTenantsForAllocationPolicy` which allows a policy to raise early to reject a query. The base class handles this exception as a first class citizen
* Updates the concurrent policy to raise that exception instead
* Tests the new usecases


snuba/query/allocation_policies/__init__.py
```diff
@@ -741,6 +758,8 @@ def get_quota_allowance(
                 allowance = QuotaAllowance(True, self.max_threads, {})
             else:
                 allowance = self._get_quota_allowance(tenant_ids, query_id)
+        except InvalidTenantsForAllocationPolicy as e:
+            allowance = QuotaAllowance(False, 0, e.to_dict())  # type: ignore
```
evanh, line 762: Why the #type: ignore here?

volokluev: Removed. The serializable exception conforms to the type but mypy is too dumb to realize it

evanh (approved):

PR #5501, opened 2024-02-06
feat(mql): Parse the entity from the MRI

evanh:
This has two changes: firstly, resolve public names in queries during the
parsing process, to ensure that we always are dealing with MRIs further on in
the pipeline.

Second, use that restriction to fetch the Entity based on the MRI being passed
in. Our entities align with the metric types, so we can map the metric type of
the MRI to the entity on our end. This will need to be expanded when we start
supporting multi type queries.

This is the first step in not needing an `entity` field in the SDK at all,
which will simplify the SDK and the Sentry layer, and enable features like
using MQL strings directly.

### Legal Boilerplate


volokluev (commented):
The issue that I have with this is that it makes it harder to integrate the indexer into snuba. Right now there's the assumption that all the metric_name: mri mappings are known at parse time which is not true if the indexer resolution is being done in snuba. overall I think it's a step backwards.

evanh:
> The issue that I have with this is that it makes it harder to integrate the indexer into snuba. Right now there's the assumption that all the metric_name: mri mappings are known at parse time which is not true if the indexer resolution is being done in snuba. overall I think it's a step backwards.

I think you're right but maybe for the wrong reasons: the public name mapping is not actually an indexer check, it's hardcoded in Sentry, and passed in the indexer_mappings. Initially I was trying to do avoid a full iteration through the query to do the mapping in Sentry, but now I think that might be a price that just has to be paid, and we can guarantee that the Snuba API only accepts MRIs. Then the public name mapping doesn't need to happen in the parser. I'll open a PR for that.

volokluev (approved):
Approving this but codecov says that some of the error cases are not covered by tests and I think they should be. I trust you to do what's needed here

PR #5506, opened 2024-02-07
fix: xfail optimize test again :(

MeredithAnya:
We removed the `xfail` in https://github.com/getsentry/snuba/pull/5447 but it's still flaky sometimes, not sure exactly why but when it fails it blocks CI so...


untitaker (approved):
:(

PR #5515, opened 2024-02-09
feat(rust): Add a metric for invalid schema validations

phacops:


untitaker (approved):

untitaker:
I'm reverting this because it seems that we already record this metric in `_validate_schema`.

PR #5522, opened 2024-02-11
fix: recursion error during parsing

kylemumma:


snuba/query/parser/exceptions.py
kylemumma, line 1: my new PostProcessingError

snuba/query/snql/parser.py
kylemumma, line 1: move post processing into a try-catch block in order to raise our new post processing error if the recursion error occurs

snuba/request/validation.py
```diff
@@ -116,64 +118,24 @@ def build_request(
     with sentry_sdk.start_span(description="build_request", op="validate") as span:
         try:
             request_parts = schema.validate(body)
-            tenant_referrer = request_parts.attribution_info["tenant_ids"].get(
```
kylemumma, line 119: this is the code that gets extracted into methods

snuba/request/validation.py
```diff
@@ -116,64 +118,24 @@ def build_request(
...
-                "referrer"
-            )
-            if tenant_referrer != referrer:
-                metrics.increment(
-                    "referrer_mismatch",
-                    tags={
-                        "tenant_referrer": tenant_referrer or "none",
-                        "request_referrer": referrer,
-                    },
+            referrer = _get_referrer(request_parts, referrer)
```
kylemumma, line 121: these are the methods its extracted into

snuba/request/validation.py
```diff
@@ -215,3 +177,82 @@ def build_request(
...
+    if org_ids is not None and len(org_ids) == 1:
+        sentry_sdk.set_tag("snuba_org_id", org_ids.pop())
+
+    query_project_id = _get_project_id(query)
+    if query_project_id:
+        sentry_sdk.set_tag("snuba_project_id", query_project_id)
+
+    attribution_info = _get_attribution_info(request_parts, referrer, query_project_id)
+
+    return create_request(
```
kylemumma, line 252: new request factory function

snuba/request/validation.py
```diff
@@ -116,64 +118,24 @@ def build_request(
...
-            project_ids = get_object_ids_in_query_ast(query, "project_id")
-            query_project_id = None
-            if project_ids is not None and len(project_ids) == 1:
-                query_project_id = project_ids.pop()
-                sentry_sdk.set_tag("snuba_project_id", query_project_id)
-
-            org_ids = get_object_ids_in_query_ast(query, "org_id")
-            if org_ids is not None and len(org_ids) == 1:
-                sentry_sdk.set_tag("snuba_org_id", org_ids.pop())
+                query_metadata = get_snuba_query_metadata(request, dataset, timer)
```
kylemumma, line 133: new factory function

snuba/request/factory.py
kylemumma, line 1: the factory function for request just auto sets the id for you so you dont need to specify it

snuba/querylog/factory.py
kylemumma, line 1: the factory function for SnubaQueryMetadata makes it so you only have to provide a subset of the fields and it computes the derived ones for you

snuba/query/snql/parser.py
```diff
@@ -1508,40 +1508,46 @@ def parse_snql_query(
...
+
+        # XXX: Select the entity to be used for the query. This step is temporary. Eventually
+        # entity selection will be moved to Sentry and specified for all SnQL queries.
+        _post_process(query, [_select_entity_for_dataset(dataset)], settings)
+
+        # Validating
+        with sentry_sdk.start_span(op="validate", description="expression_validators"):
+            _post_process(query, VALIDATORS)
+        return query, snql_anonymized
+    except ParsingException:
```
kylemumma, line 1550: gotta keep these as they're expected and due to invalid queries

snuba/query/snql/parser.py
```diff
@@ -1508,40 +1508,46 @@ def parse_snql_query(
...
+        raise
+    except Exception:
```
kylemumma, line 1552: this would be an unexpected exception like our RecursionError

snuba/query/data_source/projects_finder.py
kylemumma, line 1: extracted ProjectsFinder to its own file

snuba/web/query.py
kylemumma, line 1: extracted ProjectsFinder to its own file

snuba/query/parser/exceptions.py
kylemumma, line 1: My new PostProcessingError

snuba/querylog/factory.py
```diff
@@ -0,0 +1,34 @@
...
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.factory import get_dataset_name
+from snuba.query.data_source.projects_finder import ProjectsFinder
+from snuba.query.logical import Query
+from snuba.querylog.query_metadata import SnubaQueryMetadata
+from snuba.request import Request
+from snuba.utils.metrics.timer import Timer
+
+
+def get_snuba_query_metadata(
```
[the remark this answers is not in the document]
kylemumma: @evanh No they aren't used outside of validation.py. Initially I put them in the same file as the constructor for their object because I thought since they are both used for object creation they should be located together. But then I saw somewhere else that factory.py file was used for factory function, so I made a factory.py to put it in.

kylemumma:
@evanh No I didn't write any tests, i'm not sure what I should write tests for and how

snuba/querylog/factory.py
enochtangg, line 12: Should this be `create_snuba_query_metadata()` instead?

snuba/request/validation.py
```diff
@@ -116,64 +118,24 @@ def build_request(
...
-                }
-                query_settings["referrer"] = referrer
-                # TODO: referrer probably doesn't need to be passed in, it should be from the body
-                settings_obj: Union[HTTPQuerySettings, SubscriptionQuerySettings]
-                # the parameters accept either `str` or `bool` but we pass in `str | bool`
-                settings_obj = settings_class(**query_settings)  # type: ignore
-            elif settings_class == SubscriptionQuerySettings:
-                settings_obj = settings_class(
-                    consistent=_consistent_override(True, referrer),
+            except PostProcessingError as exception:
```
enochtangg, line 127: Could we add a test to make sure the recursion query is being handled properly and that we're catching `PostProcessingError`?

enochtangg (commented):
Nice! A few small comments.

snuba/querylog/factory.py
kylemumma: done

snuba/request/validation.py
kylemumma: done

kylemumma:
added test and get_snuba_query_metadata -> create_snuba_query_metadata

snuba/querylog/factory.py
```diff
@@ -0,0 +1,34 @@
+from snuba.clickhouse.query_dsl.accessors import get_time_range
```
evanh, line 1: I don't think we need these factory files. I think these functions can live in the validation file.

evanh (approved):
One small change to the structure, but otherwise looks good.

enochtangg (approved):

snuba/request/factory.py
```diff
@@ -0,0 +1,26 @@
...
+
+from snuba.attribution.attribution_info import AttributionInfo
+from snuba.query.composite import CompositeQuery
+from snuba.query.data_source.simple import Entity
+from snuba.query.logical import Query
+from snuba.query.query_settings import QuerySettings
+from snuba.request import Request
+
+
+def create_request(
```
volokluev, line 12: did you need all this code just to not have to pass in a uuid?

snuba/querylog/factory.py
```diff
@@ -0,0 +1,34 @@
...
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.factory import get_dataset_name
+from snuba.query.data_source.projects_finder import ProjectsFinder
+from snuba.query.logical import Query
+from snuba.querylog.query_metadata import SnubaQueryMetadata
+from snuba.request import Request
+from snuba.utils.metrics.timer import Timer
+
+
+def create_snuba_query_metadata(
```
volokluev, line 12: Why not just make the SnubaQueryMetadata constructor generate these things? What purpose does the factory function serve?

volokluev (commented):
Overall looks good, just some questions about the refactor

snuba/querylog/factory.py
kylemumma: This seemed like a good option, I just explored the idea of using default values in the constructor for SnubaQueryMetadata to simulate constructor override. And just generate them like this if they are None.

It turns out that this doesn't seem viable to implement given that the class is @dataclass, I can't override __init__ and it ends up messing up typing in the whole class.

I dont really know enough about @dataclass to know if its safe to remove it and just define my own constructor.

snuba/request/factory.py
kylemumma: fair point, I changed it to just pass in the uuid

snuba/querylog/factory.py
kylemumma: My thinking was that the create_snuba_metadata_query function is related to constructing SnubaMetadataQuery objects, rather than doing validation, which is why I thought it made more sense to be located with the SnubaMetadataQuery files.

One other thought I had is that the same construction logic is repeated here https://github.com/getsentry/snuba/blob/master/snuba/web/query.py#L102
this is where I took the logic from, I didnt change it to use the factory new factory function, but I probably should in order to avoid repeated code.

If I keep create_snuba_metadata_query in factory.py or perhaps query_metadata.py it would be easy to use in both places. If I make it a private method in validation.py, I couldnt do this.

kylemumma: I made the change to remove the duplicate code, so you can see what i was talking about

snuba/querylog/factory.py
volokluev: it is safe to remove and provide your own constructor

snuba/request/validation.py
```diff
@@ -215,3 +179,83 @@ def build_request(
...
+
+def _get_attribution_info(
+    request_parts: RequestParts, referrer: str, query_project_id: int | None
+) -> AttributionInfo:
+    return AttributionInfo(
+        **update_attribution_info(request_parts, referrer, query_project_id)
+    )
+
+
+def _build_request(
```
volokluev, line 236: why did this function need to be pulled out of `build_request`? Does it have different concerns? Is it going to be reused somewhere else?

kylemumma: Yes it is reused, I pulled it out to prevent duplicate code. It is called in the except clause as well as in the normal body.

I saw build_request as doing:
1. from raw request, parse into query
2. from query build request. (this is _build_request)

Step 2 was already done in the normal case, build now it had to be done in case of exception as well. So I factored it out into a method and used the method in both cases.

kylemumma:
I got rid of create_snuba_metadata factory function and instead overrode the constructor for SnubaQueryMetadata to make more of the parameters optional. The benefit of this is that we can just use the constructor for creation and there isn't an extra factory method. The downside is that the constructor is a bit smelly now.

An alternative design I saw along the way was [this chatgpt sponsored design pattern](https://docs.google.com/document/d/1xmIalcgtRKal69u24g48J12Yd0ro5-RwuIEZZRywq2A/edit?usp=sharing). I didn't go with this pattern as it would require changing the existing creation of a SnubaMetadataQuery object. But it did seem like it could be a cleaner approach.

PR #5524, opened 2024-02-12
ref(metrics_summaries): Read new metrics summaries schema from new topic

phacops:
This is an effort to split metrics summaries to their own topic from Relay to not have to read every span to know if we need to insert something in ClickHouse.

For the transition, we'll:
- merge the Relay part to start producing on the new topic
- let the consumer read the spans topic until we see 0 summaries being inserted
- merge the Snuba configuration change for this consumer
- merge this PR and deploy

DB schema doesn't change.

Related PRs:
- Relay: https://github.com/getsentry/relay/pull/3045/
- Schema: https://github.com/getsentry/sentry-kafka-schemas/pull/212
- Ops: https://github.com/getsentry/ops/pull/9396


untitaker:
this migration plan has pretty severe impact on ingestion latency, but i don't know if that is acceptable or not. there also needs to be an ops PR to configure the consumer to read from the new topic, which needs to be deployed as part of step 3

evanh (commented):
@untitaker mentioned needing an Ops PR, is that PR opened already?

Also, a lot of the CI checks are failing, is that something that you are working to fix?

phacops:
> @untitaker mentioned needing an Ops PR, is that PR opened already?

Not yet, I'll go ahead and do it.

> Also, a lot of the CI checks are failing, is that something that you are working to fix?

I was waiting on the schema PR since tests would still have failed without it. On it now.

phacops:
> this migration plan has pretty severe impact on ingestion latency, but i don't know if that is acceptable or not. there also needs to be an ops PR to configure the consumer to read from the new topic, which needs to be deployed as part of step 3

I think it's acceptable since it's a simple way to do the migration. I could have created a second set of consumers, ask ops to deploy them before, then switch the topic, then do a bunch of clean up but it doesn't seem worth the 15 min latency we'll have if we follow the steps I outlined.

phacops:
> this migration plan has pretty severe impact on ingestion latency, but i don't know if that is acceptable or not. there also needs to be an ops PR to configure the consumer to read from the new topic, which needs to be deployed as part of step 3

Is that true? We specified the storage, which doesn't change, and it should look up the topic automatically in the config.

https://github.com/getsentry/ops/blob/master/k8s/services/snuba/_values.yaml#L181

phacops:
I think this is what @untitaker meant: https://github.com/getsentry/ops/pull/9396.

phacops:
The PR is good, at least on the Snuba-side. We have a bit of an issue with some tests in Sentry.

Given we're changing the consumer, tests in Sentry are failing. Now, we could remove the file we're testing in this PR, merge this, then merge https://github.com/getsentry/sentry/pull/65177 and things will be fixed. It has to be done on time to minimize blocking master.

I'm not sure how we can get around it without keeping the 2 version of the consumers otherwise.

untitaker:
are you able to construct a patch to sentry that works with both versions of snuba? This would be ideal for local developers as well where versions can be slightly out of sync

phacops:
@untitaker suggested to disable the tests in Sentry until the transitions is done.

https://github.com/getsentry/sentry/pull/65177/commits/918e996f30f549e2972b4935208680fe09147940

snuba/clickhouse/http.py
```diff
@@ -261,6 +261,7 @@ def join(self, timeout: Optional[float] = None) -> None:
             # https://github.com/yandex/ClickHouse/issues/6272 is available.
             content = response.data.decode("utf8")
             details = CLICKHOUSE_ERROR_RE.match(content)
+            print(content)
```
lynnagara, line 264: please remove

lynnagara (approved):
Approving to unblock the rollout since @untitaker is out today

PR #5525, opened 2024-02-12
ref: Update the snuba-sdk to the latest version

evanh:
The new version doesn't require the Entity in the SDK for Metrics, since the
Snuba endpoint parses that from the MRI automatically.


enochtangg (approved):
Looks good, just need tests to pass on all clickhouse versions.

PR #5528, opened 2024-02-12
change: Move curl from buildDep to runtimeDep

dmajere:


MeredithAnya:
@dmajere how come moving it to `runtimeDeps` helps? if they are both installed in https://github.com/getsentry/snuba/blob/master/Dockerfile#L35 am I missing something?

untitaker (commented):
what are we trying to fix here? curl should not be a runtime dep, unless you want it installed for healthchecks maybe?

untitaker:
> if they are both installed in [master/Dockerfile#L35](https://github.com/getsentry/snuba/blob/master/Dockerfile?rgh-link-date=2024-02-12T20%3A21%3A03Z#L35) am I missing something?

buildDep is uninstalled later again

MeredithAnya:
> what are we trying to fix here? curl should not be a runtime dep, unless you want it installed for healthchecks maybe?

@untitaker We are trying to run a k8 job that and seeing `/bin/sh: 2: curl: not found` in the logs when it completes the job - this is the ops pr for it https://github.com/getsentry/ops/pull/9134

dmajere:
@untitaker curl is needed in container for cron job setups, we need someway to stop envoy. Usually https://github.com/getsentry/ops/blob/0e4f6035874b395db03937c40bd058840584522b/single-tenant/k8s/snuba/cleanup-storage-cronjob.yaml#L47

untitaker (approved):

PR #5529, opened 2024-02-12
fix(gocd): remove explicit context flag from snuba-stable

bmckerry:
after https://github.com/getsentry/snuba/pull/5510 the `de` GoCD pipeline was still erroring, this removes a few more explicit `--context` flags from the stable pipeline


IanWoodard (approved):
Good catch

nikhars (approved):

PR #5544, opened 2024-02-14
add query_fetcher cli

MeredithAnya:
This is kind of a redo of some of the steps that I outlined in https://github.com/getsentry/snuba/pull/5374. But things are laid out a bit differently. 

**Fetcher**
This does the work of the following:
* Fetching the queries that we want to re-run from the prod `system.query_log`
* Saving those queries to the GCS bucket

This PR replaces part of the work of the original **Replayer** (the fetching and saving of queries) but that is now going to be updated in https://github.com/getsentry/snuba/pull/5554 to be in charge of replaying queries and saving the results from the re-run queries

### DIFF
--- a/snuba/cli/query_fetcher.py
+++ b/snuba/cli/query_fetcher.py
@@ -17,7 +17,7 @@ QueryToRerun = namedtuple("QueryToRerun", ["query_str", "query_id"])
 
 
 class QueriesFileSaver:
-    def __init__(self, gcs_bucket: Optional[GCSUploader] = None) -> None:
+    def __init__(self, gcs_bucket: Optional[str] = None) -> None:
         self.uploader = None
         if gcs_bucket:
             try:
@@ -39,7 +39,7 @@ class QueriesFileSaver:
         hour = date.hour
         return f"queries/{day}/{hour}/{table}"
 
-    def _full_path(self, filename) -> str:
+    def _full_path(self, filename: str) -> str:
         return f"/tmp/{filename}.csv"
 
     def _save_to_csv(self, filename: str, results: Sequence[QueryToRerun]) -> None:
@@ -51,14 +51,19 @@ class QueriesFileSaver:
         logger.info(f"File {self._full_path(filename)} saved")
 
     def _save_to_gcs(self, filename: str, blob_name: str) -> None:
-        self.uploader.upload_file(self._full_path(filename), blob_name)
+        if self.uploader:
+            self.uploader.upload_file(self._full_path(filename), blob_name)
 
-    def save(self, table, date, results) -> None:
+    def save(self, table: str, date: datetime, results: Sequence[QueryToRerun]) -> None:
+        """
+        First save the results to local csv file,
+        then upload the file to gcs bucket.
+        """
         filename = self._format_filename(table, date)
         self._save_to_csv(filename, results)
-        if self.uploader:
-            blob_name = self._format_blob_name(table, date)
-            self._save_to_gcs(filename, blob_name)
+
+        blob_name = self._format_blob_name(table, date)
+        self._save_to_gcs(filename, blob_name)
 
 
 def get_querylog_query(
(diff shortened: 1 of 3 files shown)


snuba/cli/query_fetcher.py
```diff
@@ -0,0 +1,193 @@
...
+@click.option(
+    "--database",
+    type=str,
+    default="default",
+    help="Name of the ClickHouse database",
+    required=True,
+)
+@click.option(
+    "--gcs-bucket",
+    # todo: only putting this default for now
```
MeredithAnya, line 126: need to remove this, had a default for local testing that i removed

snuba/cli/query_fetcher.py
```diff
@@ -0,0 +1,193 @@
...
+    querylog_host: str,
+    querylog_port: int,
+    window_hours: int,
+    tables: str,
+    database: str,
+    gcs_bucket: Optional[str],
+    log_level: Optional[str] = None,
+) -> None:
+    """
+    We don't need the results of the query because we'll
```
MeredithAnya, line 141: need to update this docstring - its from copy/pasta

snuba/cli/query_fetcher.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        filename = self._format_filename(table, date)
+        self._save_to_csv(filename, results)
+
+        blob_name = self._format_blob_name(table, date)
+        self._save_to_gcs(filename, blob_name)
+
+
+def get_querylog_query(
+    databases: list[str], tables: list[str], start: datetime, end: datetime
```
nikhars, line 70: Are we going to pass multiple databases and tables to this API? The typing suggests so. But the query below suggests its going to be only 1. If there are going to be multiple databases and tables then maybe change the query to use IN clause.

snuba/cli/query_fetcher.py
```diff
@@ -0,0 +1,210 @@
...
+        query_id,
+        query
+    FROM system.query_log
+    WHERE (query_kind = 'Select')
+    AND (type = 'QueryFinish')
+    AND (databases = {databases})
+    AND (tables = {tables})
+    AND (query_start_time >= toDateTime('{start_time}'))
+    AND (query_start_time <= toDateTime('{end_time}'))
+    LIMIT 100
```
nikhars, line 86: Any reason for the limit? What would happen in cases where there are more than 100 queries? I do not see any sort of pagination happening.

MeredithAnya: this was just supposed to be for testing, not sure if we want to have limits and handle pagination now

snuba/cli/query_fetcher.py
MeredithAnya: @nikhars This is maybe a confusing by product of the [clickhouse querylog](https://clickhouse.com/docs/en/operations/system-tables/query_log) - you have to query as an array of dbs or array of tables, but if you as for `WHERE tables in ['default.errors_dist', 'default.transactions_dist']` you will get queries that have both of those tables, not either. Same with databases.

I think it makes more sense for us to just query one table at a time, even if that means that some queries will run twice because they had multiple tables

snuba/cli/query_fetcher.py
nikhars: If we want to get all queries to get compared within the time range we should fix this. Either remove limit. But then I do not know if clickhouse would apply some limit. Or add in some sort of pagination with limit and offset. Probably easier to do. That way you ensure that you are able to verify all queries.

snuba/cli/query_fetcher.py
```diff
@@ -0,0 +1,210 @@
...
+    Assuming window_hours is 2, table is `test_table`, and
+    the job is running at 00:01:00 (random time in hour 0)
+
+        Queries fetched (rounded to the hour):
+        * 2hours ago - 1hour ago
+        * 1hour ago - now
+
+        will be saved as:
+        queries/2024_01_16/test_table_22.csv - first hour
+        queries/2024_01_16/test_table_22.csv - second hour
```
nikhars, line 160: Nit: Both of these are the same.

snuba/cli/query_fetcher.py
MeredithAnya: Removing the limit for now - might have to come back to this though

nikhars (approved):

PR #5548, opened 2024-02-15
fix(rust): Fix wrong metric name

untitaker:


untitaker:
this metric is already recorded in the same file

PR #5550, opened 2024-02-15
remove context from deploy scripts for real this time

bmckerry:
Followup to #5529 - DE region migrations were still failing, this should fix


IanWoodard (approved):
Looks good! Should we also update snuba-sns-test to make it consistent with the rest?

lynnagara (approved):

PR #5559, opened 2024-02-16
ref: Add spans to allocation policy for debugging

evanh:


volokluev (approved):

evanh:
> Are you debugging something specific?

Basically, would like to know if/how a specific transaction was throttled. We tag the transaction with `max_threads`, but that doesn't always explain why a transaction was throttled.

PR #5560, opened 2024-02-17
feat(gen-metrics): Support encoding in gen metrics processor

john-z-yang:
### Overview

Adds some serde logic to support new schema format for the `values` field.

Before

```json
{
  "values": [1, 2, 3]
}
```

Now

```json
{
  "values": {
    "format": "array",
    "data": [1, 2, 3]
  }
}
```

Both are supported

### DIFF
--- a/rust_snuba/src/processors/generic_metrics.rs
+++ b/rust_snuba/src/processors/generic_metrics.rs
@@ -2,8 +2,11 @@ use adler::Adler32;
 use anyhow::Context;
 use chrono::DateTime;
 use rust_arroyo::backends::kafka::types::KafkaPayload;
-use serde::{Deserialize, Serialize};
-use std::{collections::BTreeMap, vec};
+use serde::{
+    de::value::{MapAccessDeserializer, SeqAccessDeserializer},
+    Deserialize, Deserializer, Serialize,
+};
+use std::{collections::BTreeMap, marker::PhantomData, vec};
 
 use crate::{
     types::{CogsData, InsertBatch, RowData},
@@ -65,9 +68,9 @@ enum MetricValue {
     #[serde(rename = "c")]
     Counter(f64),
     #[serde(rename = "s")]
-    Set(Vec<u64>),
+    Set(EncodedSeries<u64>),
     #[serde(rename = "d")]
-    Distribution(Vec<f64>),
+    Distribution(EncodedSeries<f64>),
     #[serde(rename = "g")]
     Gauge {
         count: u64,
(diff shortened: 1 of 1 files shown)


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -78,6 +81,60 @@ enum MetricValue {
...
+        }
+    }
+}
+
+fn encoded_series_deserializer<'de, T, D>(deserializer: D) -> Result<EncodedSeries<T>, D::Error>
+where
+    T: Deserialize<'de>,
+    D: Deserializer<'de>,
+{
+    struct Visitor<U> {
```
Dav1dde, line 103: nit:

```suggestion
    struct Visitor<U>(PhantomData<U>);
```

Usually define these Visitors as a unit struct.

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -78,6 +81,60 @@ enum MetricValue {
...
+        fn visit_map<A>(self, map: A) -> Result<Self::Value, A::Error>
+        where
+            A: serde::de::MapAccess<'de>,
+        {
+            EncodedSeries::deserialize(MapAccessDeserializer::new(map))
+        }
+    }
+
+    deserializer.deserialize_any(Visitor {
+        phantom: PhantomData::<T>,
```
Dav1dde, line 134:
```suggestion
        phantom: PhantomData,
```

You don't need to specify the type here, Rust can infer it form the return type.

With the Visitor as a unit struct, it becomes `Visitor(PhantomData)`, quite short.

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -78,6 +81,60 @@ enum MetricValue {
...
+    }
+
+    impl<'de, U> serde::de::Visitor<'de> for Visitor<U>
+    where
+        U: Deserialize<'de>,
+    {
+        type Value = EncodedSeries<U>;
+
+        fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+            formatter.write_str("expected either old or new distribution value format")
```
Dav1dde, line 114: We opted to call it `legacy` in Relay instead of `old`

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -767,6 +839,62 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_distribution_processor_with_v1_distribution_message() {
```
Dav1dde, line 843: Maybe makes sense to replicate the test for a set as well

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -64,10 +67,10 @@ struct FromGenericMetricsMessage {
 enum MetricValue {
     #[serde(rename = "c")]
     Counter(f64),
-    #[serde(rename = "s")]
-    Set(Vec<u64>),
-    #[serde(rename = "d")]
-    Distribution(Vec<f64>),
+    #[serde(rename = "s", deserialize_with = "encoded_series_deserializer")]
```
Dav1dde, line 70: I think it would be good if the deserializer fn name expressed that it handles the compatible "legacy" format since `EncodedSeries` already implements `Deserialize`.

Maybe something along the lines of `encoded_series_legacy_compat_deserializer` (I am really bad at naming things)

Dav1dde (approved):
Nice!

Just a few tiny nits (feel free to ignore).

ayirr7:
Going to add a test to this

rust_snuba/src/processors/generic_metrics.rs
ayirr7: Yes, adding that

rust_snuba/src/processors/generic_metrics.rs
john-z-yang: Fixed, thank you

rust_snuba/src/processors/generic_metrics.rs
john-z-yang: That's really cool, thanks

rust_snuba/src/processors/generic_metrics.rs
john-z-yang: How do you feel about just `compat_deserializer`, as the reader could infer legacy and new from `compat`?

untitaker (approved):

PR #5566, opened 2024-02-19
ref(rust): Refactor errors schema

untitaker:
1. Create StringToIntDatetime which produces the correct jsonschema
   instead of using serde(deserialize_with) everywhere
2. Fix a bug in schema comparison where comparing a subschema (like
   ErrorData) would result in all references being broken.
3. Remove the remaining diff -- it should pass after merging https://github.com/getsentry/sentry-kafka-schemas/pull/219


rust_snuba/src/processors/utils.rs
```diff
@@ -31,3 +32,10 @@ where
     };
     Ok(seconds_since_epoch.timestamp() as u32)
 }
+
+#[derive(Debug, Deserialize, JsonSchema, Default, Serialize)]
+pub struct StringToIntDatetime(
+    #[serde(deserialize_with = "ensure_valid_datetime")]
+    #[schemars(with = "String")]
```
lynnagara, line 39: Is this how you initially generated the schema? How come we're checking this change in permanently?

untitaker: because there's a test that asserts the two schemas are the same (or at least compatible), so we have to be able to generate schemas

lynnagara (approved):

PR #5571, opened 2024-02-21
feat: Remove query splitters from the API

evanh:
After testing how much the splitters were being used, and the improvement they
were providing, splitters were only being used on a very small percentage
of queries (up to 5% on the highest dataset). After pausing the splitters,
there was almost no impact on the performance of queries. Only a single
referrer had a noticeable performance degradation, and that was not enough to
cause any actual impact on the queries (rate limiting, timeouts etc.)


enochtangg (approved):

asottile-sentry:
this is breaking sentry:

```
FAILED tests/sentry/eventstore/snuba/test_backend.py::SnubaEventStorageTest::test_get_adjacent_event_ids - AssertionError: assert None == ('4553539026681858', 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
```

PR #5579, opened 2024-02-21
feat: Create a `snuba health` command 

xurui-c:
[Create a `snuba health` command](https://getsentry.atlassian.net/browse/SNS-2629)

# Problem
It's not easy for SREs to run curl commands against a `snuba api` pod in production, so health checks cannot be easily done

# Solution
Added the `snuba health` subcommand.

# Testing
Added `snuba health` cli test

### Legal Boilerplate


volokluev (changes requested):
This will break stuff, the healthcheck endpoint is still used by kubernetes. If you remove it, kubernetes will never route traffic to a pod running this code

snuba/cli/health.py
```diff
@@ -0,0 +1,152 @@
...
+    UndefinedClickhouseCluster,
+)
+from snuba.datasets.factory import get_dataset, get_enabled_dataset_names
+from snuba.datasets.schemas.tables import TableSchema
+from snuba.datasets.storage import Storage
+from snuba.environment import setup_logging
+from snuba.web.views import check_down_file_exists, metrics
+
+
+@click.command()
```
volokluev, line 27: instead of removing the endpoint, the api and this command should share code. Pull out the healthcheck code into its own utility function and have both the cli and endpoint call that utility function

snuba/utils/health_info.py
```diff
@@ -0,0 +1,181 @@
...
+                return False
+
+            start_time: float = uwsgi.started_on
+            IS_SHUTTING_DOWN = m_time > start_time
+            return IS_SHUTTING_DOWN
+        except OSError:
+            return False
+
+
+def get_health_info(thorough: Union[bool, str]) -> Tuple[str, int, Dict[str, str]]:
```
volokluev, line 72: nit: this tuple type doesn't tells me little about what the function returns. a `dataclass` with named fields would be a better experience for me as a reader/user of the code

snuba/web/views.py
```diff
@@ -72,126 +56,27 @@
 from snuba.subscriptions.codecs import SubscriptionDataCodec
 from snuba.subscriptions.data import PartitionId
 from snuba.subscriptions.subscription import SubscriptionCreator, SubscriptionDeleter
+from snuba.utils.health_info import (
+    IS_SHUTTING_DOWN,
```
volokluev, line 60: I think this is going to cause a problem due to how python imports work. Do you know how this variable is used? Can you think of some ways that this code will change in functionality now that this variable has been exported somewhere else?

tests/cli/test_health.py
```diff
@@ -0,0 +1,27 @@
+from unittest import mock
+
+from click.testing import CliRunner
+
+from snuba.cli.health import health
+
+
+def test_check_health() -> None:
```
volokluev, line 8: nit: the test cases are really good but they should be separate test cases, not one big one

volokluev (commented):
Great job overall, the IS_SHUTTING_DOWN variable may be problematic though. Do you have a good understanding of how it works and what it's for?

snuba/cli/health.py
xurui-c: The bulk of the health checking logic is now in `health_info.py`'s `get_health_info` function. The cli and the endpoint functions calls `get_health_info`.

tests/cli/test_health.py
xurui-c: Also separated teset cases in `test_views.py`

snuba/web/views.py
xurui-c: Made `get_shutdown` and `_set_shutdown`

volokluev (approved):

PR #5581, opened 2024-02-22
Unrevert: feat: Remove query splitters from the API 

evanh:
This was reverted because it was causing a test failure in Sentry master.


enochtangg (approved):
LGTM when sentry tests pass again.

PR #5584, opened 2024-02-23
ref(ch-upgrades): update query_comparer

MeredithAnya:
Updating the comparer to use the file_manager and grab the files to compare from GCS. Corresponding ops PR https://github.com/getsentry/ops/pull/9471

What the comparer does:

1. Looks at the `results-` directories to see if we have pairs of results that are ready to be compared and puts them together as a `matched_pairs`. e.g.
    ```python
     [('results-21-8/2024_02_15/meredith_test_1.csv', 'results-22-8/2024_02_15/meredith_test_1.csv')]
     ```
2. Iterates through those pairs to actually compare the results from those files. Before it does so though, it checks that we haven't already done the comparison (we ignore that if `--override` is used). We can check in the `compared-data/` and `compared-perf/` directories
3. We compare the results and generate mismatches for the `data` and `perf` categories based on the result data. Thresholds of what should be considered mismatching are in the corresponding `_THRESHOLDS` dicts.
4. We save the comparison results for each category separately, as well as send the slack overview + csv files per category as well. We do calculate the % mismatches based on the total rows of the files. Adding percentages of both categories might exceed 100% since the same query_id could have a mismatch in the performance AND the data consistency.

### DIFF
--- a/snuba/cli/query_comparer.py
+++ b/snuba/cli/query_comparer.py
@@ -8,19 +8,15 @@ import structlog
 
 from snuba import settings
 from snuba.admin.notifications.slack.client import SlackClient
+from snuba.clickhouse.upgrades.comparisons import (
+    BlobGetter,
+    FileManager,
+    QueryMeasurementResult,
+)
 from snuba.environment import setup_logging, setup_sentry
+from snuba.utils.gcs import GCSUploader
 
 logger = structlog.get_logger().bind(module=__name__)
-
-
-@dataclass(frozen=True)
-class QueryMeasurement:
-    query_id: str
-    query_duration_ms: int
-    result_rows: int
-    result_bytes: int
-    read_rows: int
-    read_bytes: int
 
 
 @dataclass(frozen=True)
@@ -111,26 +107,14 @@ def write_querylog_comparison_results_to_csv(
 
 @click.command()
 @click.option(
-    "--base-file",
-    help="Clickhouse queries results from base version.",
-    required=True,
-)
-@click.option(
-    "--upgrade-file",
-    help="Clickhouse queries results from upgrade version.",
-    required=True,
-)
-@click.option(
-    "--table",
-    help="Clickhouse queries results from upgrade version.",
+    "--gcs-bucket",
+    help="Name of gcs bucket to save query files to",
     required=True,
 )
 @click.option("--log-level", help="Logging level to use.")
 def query_comparer(
     *,
-    base_file: str,
-    upgrade_file: str,
-    table: str,
+    gcs_bucket: str,
     log_level: Optional[str] = None,
 ) -> None:
     """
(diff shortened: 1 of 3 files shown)


snuba/cli/query_comparer.py
```diff
@@ -139,48 +123,55 @@ def query_comparer(
...
+            assert (
+                v1_result.query_id == v2_result.query_id
+            ), "Invalid query_ids: ids must match"
+            total_rows += 1
+            if v1_result == v2_result:
+                continue
+
+            mismatches.append(_create_mismatch(v1_result, v2_result))
+
+        _send_slack_report(total_rows, mismatches, file_format.table)
```
MeredithAnya, line 170: Still need to test that this part works

snuba/cli/query_comparer.py
```diff
@@ -1,136 +1,50 @@
...
-            writer.writerow(row)
+MismatchResultSet = Sequence[PerfMismatchResult | DataMismatchResult]
 
 
 @click.command()
 @click.option(
-    "--base-file",
-    help="Clickhouse queries results from base version.",
+    "--gcs-bucket",
+    help="Name of gcs bucket to save query files to",
```
nikhars, line 28:
```suggestion
    help="Name of gcs bucket to read query results data from",
```

snuba/cli/query_comparer.py
```diff
@@ -139,86 +53,186 @@ def query_comparer(
...
-        return QueryMeasurement(
-            query_id=row[0],
-            query_duration_ms=int(row[1]),
-            result_rows=int(row[2]),
-            result_bytes=int(row[3]),
-            read_rows=int(row[4]),
-            read_bytes=int(row[5]),
-        )
+    if len(result_prefixes) != 2:
+        logger.info("Not enough results to compare.")
```
nikhars, line 64: Would it be better to throw an exception here?

snuba/cli/query_comparer.py
```diff
@@ -139,86 +53,186 @@ def query_comparer(
...
+        return
 
-    mismatches = []
-    total_rows = 0
-    for v1_row, v2_row in itertools.zip_longest(base_reader, upgrade_reader):
-        if v1_row[0] == "query_id":
-            # csv header row
+    def get_matched_pairs() -> Sequence[Tuple[str, str]]:
```
nikhars, line 67: Can you document what this function does and what it returns. Similar to what you have in the PR description.

snuba/cli/query_comparer.py
```diff
@@ -139,86 +53,186 @@ def query_comparer(
...
+            save_comparisons(compare_type, data_mismatches)
+        else:
+            # no compare type == run both reports
+            save_comparisons("perf", perf_mismatches)
+            save_comparisons("data", data_mismatches)
+
+
+PERF_THRESHOLDS = {
+    "query_duration_ms": 0,
+    "read_rows": 0,
```
nikhars, line 145: Why have read_rows and read_bytes in perf thresholds?

snuba/cli/query_comparer.py
```diff
@@ -139,86 +53,186 @@ def query_comparer(
...
+    "result_bytes": 0,
+}
+
+
+def create_perf_mismatch(
+    v1_data: QueryMeasurementResult, v2_data: QueryMeasurementResult
+) -> Optional[PerfMismatchResult]:
+    d_ms_1 = v1_data.query_duration_ms
+    d_ms_2 = v2_data.query_duration_ms
+    d_ms_delta = d_ms_2 - d_ms_1
```
nikhars, line 160: The duration mismatch should be based on some percentage variance of the base query duration rather than a fixed perf threshold. Otherwise you are bound do get a lot of noise in your reports. Should we start off with a 10% variance to begin with and see what results we get back?

snuba/cli/query_comparer.py
```diff
@@ -139,86 +53,186 @@ def query_comparer(
...
+            d_ms_1=d_ms_1,
+            d_ms_2=d_ms_2,
+            rd_r_1=rd_r_1,
+            rd_r_2=rd_r_2,
+            rd_b_1=rd_b_1,
+            rd_b_2=rd_b_2,
+            d_ms_delta=d_ms_delta,
+            rd_r_delta=rd_r_delta,
+            rd_b_delta=rd_b_delta,
+        )
```
nikhars, line 187: Let's rename the fields to be more explicit so that the CSV headers are easier to read when analyzing them.

snuba/cli/query_comparer.py
MeredithAnya: do we care to track if a query is reading more data than before? that would probably be reflected in the duration I guess

snuba/cli/query_comparer.py
MeredithAnya: yeah I can do that

snuba/cli/query_comparer.py
MeredithAnya: @nikhars would you prefer to have the full names written out? `duration_ms_1`, `read_rows_1`, etc

snuba/cli/query_comparer.py
nikhars: for query performance, it shouldn't really matter as the first point of finding mismatches. you should only use query duration. If you want to drill down into mismatches, those additional fields may give us hints.

snuba/cli/query_comparer.py
nikhars: yeah that should work. what does `_1` and `_2` signify. Maybe expand on that as well

snuba/cli/query_comparer.py
MeredithAnya: added 10% but also added a check for a delta of 10ms so that really fast queries don't get included in the report if they differ slightly

nikhars (approved):

PR #5586, opened 2024-02-26
fix(query-replayer): sort the blob names so we replay in the same order

MeredithAnya:


nikhars (approved):

snuba/cli/query_replayer.py
```diff
@@ -214,19 +217,20 @@ def rerun_queries_for_blob(blob: str) -> Tuple[int, int]:
...
-                table=queries_file_format.table,
-                hour=queries_file_format.hour,
-            ),
+            result_file_format,
             replay_results,
         )
 
         if notify:
             # TODO: maybe use new specific channel id
-            filename = file_manager.filename_from_blob_name(blob_name)
```
MeredithAnya, line 229: this would have been getting the blob name of the queries to re-run, which is not what we want. We want the result data from the replayed queries

PR #5588, opened 2024-02-27
on-call(admin): Allow viewing migrations for product tools role

enochtangg:
Product owners should be able to view which migrations have been executed in production to get a better sense of what their datasets/tables look like in prod. This can help product developers validate whether or not their tables have been created in a new environment without having to traverse GoCD.

### DIFF
--- a/snuba/admin/iam_policy/iam_policy.json
+++ b/snuba/admin/iam_policy/iam_policy.json
@@ -14,7 +14,8 @@
     },
     {
       "members": [
-        "group:team-sns@sentry.io"
+        "group:team-sns@sentry.io",
+        "group:access-snuba-admin@sentry.io"
       ],
       "role": "roles/MigrationsReader"
     },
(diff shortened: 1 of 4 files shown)


volokluev (commented):
Would they be able to run them as well?

snuba/admin/iam_policy/iam_policy.json
```diff
@@ -14,7 +14,8 @@
     },
     {
       "members": [
-        "group:team-sns@sentry.io"
+        "group:team-sns@sentry.io",
+        "group:access-snuba-admin@sentry.io"
```
enochtangg, line 18: @volokluev I've added them to the `MigrationsReader` role. A user must have the `NonBlockingMigrationsExecutor` or `AllMigrationsExecutor` to execute a migration.

volokluev (approved):

PR #5592, opened 2024-02-27
ref(mql): Return single quantiles as a value instead of an array

evanh:
When fetching quantiles, the results would always be returned as an array, even
if the call only was fetching a single quantile (which is most of the cases).
This was both confusing to the frontend, and was causing errors with Formulas,
since a quantile would return an array, which generally can't be used in math.
(e.g. [1] / 1 is not valid).

When a single quantiles is being fetched, wrap it in arrayElement(..., 1) so
that it unwraps the array and returns a simple value.


snuba/query/mql/parser.py
```diff
@@ -996,10 +997,34 @@ def populate_query_from_mql_context(
     return query, mql_context
 
 
+def quantiles_to_quantile(
+    query: Union[CompositeQuery[QueryEntity], LogicalQuery]
+) -> None:
+    """
+    Changes quantiles(0.5)(...) to arrayElement(quantiles(0.5)(...), 1)
```
volokluev, line 1004: explaining why would be helpful (like you did in your PR comment)

volokluev (approved):

snuba/query/mql/parser.py
```diff
@@ -996,10 +997,36 @@ def populate_query_from_mql_context(
...
+) -> None:
+    """
+    Changes quantiles(0.5)(...) to arrayElement(quantiles(0.5)(...), 1). This is to simplify
+    the API (so that the arrays don't need to be unwrapped) and also avoids bugs where comparing
+    arrays of values to values cause typing errors (e.g. [1] / 1).
+    """
+
+    def transform(exp: Expression) -> Expression:
+        if isinstance(exp, CurriedFunctionCall):
+            if exp.internal_function.function_name in ("quantiles", "quantilesIf"):
```
iambriccardo, line 1011: Maybe we would like to expose the function as `quantile` and `quantileIf` to avoid breaking existing implementations that use `arrayElement(quantilesIf(...), 1)`.

PR #5602, opened 2024-02-28
fix(allocation-policy): Rename Concurrent allocation policy to CustomerConcurrent

ayirr7:


snuba/query/allocation_policies/concurrent_rate_limit.py
```diff
@@ -31,7 +31,7 @@
 )
 
 
-class BaseConcurrentRateLimitAllocationPolicy(AllocationPolicy):
```
volokluev, line 34: This does not need to be changed. The base is a concurrent limit policy, has nothing to do with customers

volokluev (commented):
1. The base class name does not need to change
2. Can you look through the existing concurrent policies to see if there are any overrides we need to preserve? The runtime config values are stored keyed off the [class name](https://github.com/getsentry/snuba/blob/master/snuba/query/allocation_policies/__init__.py#L622) so when we change the name we would need to make sure any overrides were carried over as well

snuba/query/allocation_policies/concurrent_rate_limit.py
ayirr7: oops! the find-replace changed that

PR #5603, opened 2024-02-28
feat: Run a separate billing consumer in devserver

lynnagara:
In order to align behavior across different environments, billing related messages will be produced to the billing topic in dev, which matches how it is done in prod today. This runs the billing consumer in devserver so we ingest them properly.

### DIFF
--- a/snuba/cli/devserver.py
+++ b/snuba/cli/devserver.py
@@ -66,6 +66,17 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
             ],
         ),
         (
+            "outcomes-billing-consumer",
+            [
+                "snuba",
+                "rust-consumer",
+                "--storage=outcomes_raw",
+                "--consumer-group=outcomes_billing_group",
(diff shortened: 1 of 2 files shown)


untitaker (commented):
we already have too many consumers running, and spawning the same exact consumer again doesn't make sense to me for dev. can't we override the outcomes_billing topic for development instead?

lynnagara:
> we already have too many consumers running, and spawning the same exact consumer again doesn't make sense to me for dev. can't we override the outcomes_billing topic for development instead?

We could but I think it's better for development to be closer to prod, and for the data to be on the correct topic, rather than merge them. What is the issue exactly with too many consumers in the devserver? And if there are "already too many consumers running", shouldn't that be solved independently in Snuba with a more generic solution that combines consumer logic, rather than scrutinising additional consumers and attempting to merge them in dev in a different way than production?

untitaker (approved):

PR #5604, opened 2024-02-28
fix(query): Coerce incorrect metrics dataset in query to generic metrics dataset

ayirr7:
There are generic metrics queries that are being labeled as `metrics` dataset. This affects different kinds of analysis that depend on a matching dataset and request body. 

I'm not exactly sure if this is the right entrypoint for this kind of check, but this was my guess.

### DIFF
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -16,7 +16,8 @@ from snuba.clickhouse.formatter.query import format_query
 from snuba.clickhouse.query import Query
 from snuba.clickhouse.query_inspector import TablesCollector
 from snuba.datasets.dataset import Dataset
-from snuba.datasets.factory import get_dataset_name
+from snuba.datasets.entities.entity_key import EntityKey
+from snuba.datasets.factory import get_dataset, get_dataset_name
 from snuba.query.composite import CompositeQuery
 from snuba.query.data_source.simple import Table
 from snuba.query.exceptions import QueryPlanException
@@ -61,7 +62,20 @@ def parse_and_run_query(
     """
     Runs a Snuba Query, then records the metadata about each split query that was run.
     """
-    # from_clause = request.query.get_from_clause()
+
+    entity_key = request.query.get_from_clause().key
+
+    if (
+        (entity_key == EntityKey("generic_metrics_distributions"))
+        or entity_key == EntityKey("generic_metrics_gauges")
+        or entity_key == EntityKey("generic_metrics_counters")
+        or entity_key == EntityKey("generic_metrics_sets")
+    ):
+
+        if get_dataset_name(dataset) == "metrics":
+            # add some logging
+            dataset = get_dataset("generic_metrics")
+
     query_metadata = SnubaQueryMetadata(request, get_dataset_name(dataset), timer)
 
     try:
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -17,10 +17,9 @@ from snuba.clickhouse.query import Query
 from snuba.clickhouse.query_inspector import TablesCollector
 from snuba.datasets.dataset import Dataset
 from snuba.datasets.entities.entity_key import EntityKey
-from snuba.datasets.entity import Entity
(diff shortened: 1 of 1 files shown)


snuba/web/query.py
```diff
@@ -61,7 +62,21 @@ def parse_and_run_query(
     """
     Runs a Snuba Query, then records the metadata about each split query that was run.
     """
-    # from_clause = request.query.get_from_clause()
+
+    entity = request.query.get_from_clause()
```
ayirr7, line 66: I don't know if this is the best/cleanest way to verify if a query is actually hitting generic metrics tables

snuba/web/query.py
```diff
@@ -61,7 +62,21 @@ def parse_and_run_query(
...
+
+    if (
+        entity.key == EntityKey("generic_metrics_distributions")
+        or entity.key == EntityKey("generic_metrics_gauges")
+        or entity.key == EntityKey("generic_metrics_counters")
+        or entity.key == EntityKey("generic_metrics_sets")
+    ):
+
+        if get_dataset_name(dataset) == "metrics":
+            # add some logging
```
ayirr7, line 77: I will add some form of logging so we can catch the cases where this happens

snuba/web/query.py
```diff
@@ -61,7 +62,21 @@ def parse_and_run_query(
     """
     Runs a Snuba Query, then records the metadata about each split query that was run.
     """
-    # from_clause = request.query.get_from_clause()
+
+    entity = request.query.get_from_clause()
+    assert isinstance(entity, Entity)
```
volokluev, line 67: don't do this, it can break subqueries

volokluev (commented):
Is this code that you plan to remove later or do you foresee it staying in the codebase for longer than a week?

ayirr7:
> Is this code that you plan to remove later or do you foresee it staying in the codebase for longer than a week?

The idea would be to keep it around. There are many generic metrics queries coming in with a request body and dataset that are different than what they should be, and I'd like to include a check to catch and rectify this. I don't know if this is the right entrypoint for that though.

snuba/web/query.py
ayirr7: Hm, okay. Good to know. So what is the best way to get the actual table or entity that a query is hitting, at this stage in the pipeline? Or perhaps this check shouldn't live here?

volokluev:
What is the analysis which requires this dataset to be accurate? That will help me understand where better to put it. 

Also how hard would it be to just change the queries upstream to query the right dataset? We could then put in a check to make sure the client is not querying entities that are not part of a dataset. 

We could also maybe just log the entity where it needs to be logged and base the analysis on that, given that the dataset wasn't really doing much here. 

> > Is this code that you plan to remove later or do you foresee it staying in the codebase for longer than a week?
> 
> The idea would be to keep it around. There are many generic metrics queries coming in with a request body and dataset that are different than what they should be, and I'd like to include a check to catch and rectify this. I don't know if this is the right entrypoint for that though.

ayirr7:
> What is the analysis which requires this dataset to be accurate?

This dataset needs to be accurate in order for COGS Clickhouse compute cost accounting. If we look at the logic in this [PR](https://github.com/getsentry/snuba/pull/4905/files#diff-aef5635194773d2f3f49cdcc99e4d7da491d14b341b226224620364b2159f521R114), we can see that the dataset for a query in the querylog is what we use as `app_feature` for COGS recording. `app_feature` denotes, within the COGS system, which product feature a particular resource is associated with. For example, due to the problem of incorrect dataset fields, we are recording generic metrics costs under a different product feature. 

I see that there's an `entity` in the `SnubaQueryMetadata` object; that might be enough to solve the problem. It appears to be extracted from the request body itself, so it may actually point to the right thing.

snuba/web/query.py
evanh: The from clause is not guaranteed to be an Entity.

```
from_clause = request.query.get_from_clause()
if isinstance(from_clause, Entity):
    # the rest of your logic
```

PR #5606, opened 2024-02-29
rust: log batch byte size

dbanda:
Log the bytes size of batches written to Clickhouse. This is useful for debugging.

### DIFF
--- a/rust_snuba/src/strategies/clickhouse.rs
+++ b/rust_snuba/src/strategies/clickhouse.rs
@@ -10,7 +10,7 @@ use rust_arroyo::processing::strategies::{
     CommitRequest, ProcessingStrategy, StrategyError, SubmitError,
 };
 use rust_arroyo::types::Message;
-use rust_arroyo::{counter, gauge, timer};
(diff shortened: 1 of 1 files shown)


rust_snuba/src/strategies/clickhouse.rs
```diff
@@ -73,6 +73,10 @@ impl TaskRunner<BytesInsertBatch, BytesInsertBatch, anyhow::Error> for Clickhous
                 timer!("insertions.batch_write_ms", elapsed);
             }
             counter!("insertions.batch_write_msgs", insert_batch.len() as i64);
+            gauge!(
```
untitaker, line 76: why a gauge? if there are multiple batches written within the same second, you won't be able to read min/max/avg, no?

untitaker (approved):

rust_snuba/src/strategies/clickhouse.rs
dbanda: ok changed it to a counter

anonrig (approved):

PR #5597, opened 2024-02-28
ref(gocd): snuba-query-replayer in gocd?

MeredithAnya:
I dont really know how any of this works lol

PR #5609, opened 2024-02-29
gocd: add s4s query replayer

dbanda:
follow up to #5597 

This adds a manual step that can be called to run the query replayer tool. Arguments to this step can be passed in via the `REPLAYER_ARGS`  environment variable in GoCD.

### DIFF
--- a/gocd/templates/bash/s4s-replay-queries.sh
+++ b/gocd/templates/bash/s4s-replay-queries.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+args=( $REPLAYER_ARGS )
+eval $(/devinfra/scripts/regions/project_env_vars.py --region="${SENTRY_REGION}")
+/devinfra/scripts/k8s/k8stunnel
+
+/devinfra/scripts/k8s/k8s-spawn-job.py \
+  --label-selector="service=${SNUBA_SERVICE_NAME}" \
+  --container-name="${SNUBA_SERVICE_NAME}" \
+  "snuba-migrate" \
+  "us.gcr.io/sentryio/snuba:${GO_REVISION_SNUBA_REPO}" \
+  -- \
+  snuba query-replayer "${args[@]}"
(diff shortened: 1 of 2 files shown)


gocd/templates/bash/s4s-replay-queries.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+args=( $REPLAYER_ARGS )
+eval $(/devinfra/scripts/regions/project_env_vars.py --region="${SENTRY_REGION}")
+/devinfra/scripts/k8s/k8stunnel
+
+/devinfra/scripts/k8s/k8s-spawn-job.py \
+  --label-selector="service=${SNUBA_SERVICE_NAME}" \
+  --container-name="${SNUBA_SERVICE_NAME}" \
+  "snuba-migrate" \
```
MeredithAnya, line 9: what's this for?

dbanda: good catch. copy paste error. fixed it

MeredithAnya:
@dbanda wondering how we can add the service account https://github.com/getsentry/ops/blob/master/single-tenant/k8s/snuba/_job.query-replayer.j2#L33 which is needed to have access to the clickhouse clusters and potentially the gcs bucket as well?

gocd/templates/clickhouse-query-replayer.jsonnet
```diff
@@ -0,0 +1,65 @@
...
+
+local pipeline_group = 'snuba';
+local pipeline_name = 'clickhouse-query-replayer';
+
+local generate_replay_job(component) =
+  {
+    environment_variables: {
+      SENTRY_REGION: 's4s',
+      SNUBA_SERVICE_NAME: 'snuba-admin',
+      GOOGLE_CLOUD_PROJECT: 'search-and-storage',
```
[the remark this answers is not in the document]
dbanda: I was assuming the buckets are kept in search-and-storage gcloud by default. We can override if we need to switch it. We will need to override the args param anyway.

dbanda:
> @dbanda wondering how we can add the service account https://github.com/getsentry/ops/blob/master/single-tenant/k8s/snuba/_job.query-replayer.j2#L33 which is needed to have access to the clickhouse clusters and potentially the gcs bucket as well?

@MeredithAnya An easy way I can think of is define pod with that spec in ops and we set its replica count to 0. The we replace `snuba-admin` here with the name of that pod spec.

dbanda:
https://github.com/getsentry/ops/pull/9543/files

MeredithAnya (approved):
I think this is good enough to test out

PR #5610, opened 2024-02-29
feat: filter by metric_id in select optimizer

kylemumma:


snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+        assert e.function_name == "in"
+        lhs = e.parameters[0]
+        rhs = e.parameters[1]
+        if isinstance(lhs, Column) and lhs.column_name == "metric_id":
+            assert isinstance(rhs, FunctionCall) and rhs.function_name == "array"
+            ids = set()
+            for metric_id in rhs.parameters:
+                assert isinstance(metric_id, Literal) and isinstance(
+                    metric_id.value, int
+                )
```
volokluev, line 138: A corner case I could foresee here is something like

```
in(1,2,3,4, sum(1, 2))
```
the example may be silly (people don't add metric ids) but overall it seems like here you will fail the query entirely if there's an `in` condition with a functioncall as a parameter even though the query is syntactically valid. Better thing to do in such a case would be to not apply the optimization.

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+            return self._get_domain_of_in(condside)
+
+    def _domain_union(self, e1: Domain, e2: Domain) -> Domain:
+        """Given 2 domains, returns the union"""
+        if e1 is None or e2 is None:
+            return None
+        else:
+            return e1.union(e2)
+
+    def _is_mid_cond(self, f: FunctionCall) -> bool:
```
volokluev, line 175: not sure I understand this terminology

volokluev (commented):
Good job overall. This is my first pass on this. As a reader I got lost in some undefined terminology (e.g. midcond and domain).

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+    def optimize(self, query: Query) -> None:
+        """Optimizes the given query by adding any metric_id filters in the select clause to the where clause"""
+        domain = self.get_domain_of_query(query)
+        filter = self.domain_to_query_filter(domain)
+        if filter is not None:
+            query.add_condition_to_ast(filter)
+
+    def get_domain_of_query(self, query: Query) -> Domain:
+        """
+        Returns the metric_id domain of the given query, where the metric_id domain is the set of all metric_id that are relevant to the query.
```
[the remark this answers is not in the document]
kylemumma: yes

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+            - case 2: any other function is the union of domains of its params
+                ex. sumIf1 + sumIf2 / sumIf3
+        """
+        if f.function_name[-2:] == "If":
+            predicate = f.parameters[1]
+            assert isinstance(predicate, FunctionCall)
+            return self._get_domain_of_predicate(predicate)
+        else:
+            if len(f.parameters) == 0:
+                return None
```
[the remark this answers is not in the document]
kylemumma: its defined as set[int] | None, because I have None representing "all metric id"

snuba/query/processors/logical/filter_in_select_optimizer.py
evanh: To add to this: your functions in general have a lot of `assert` statements. Generally we only have `assert` statements in our code to satisfy typing (and in tests). There is an assumption that an assert should never be thrown.

It's hard to follow the logic of your code to see that an assert statement is guaranteed to pass. So in this example, it seems like there are cases where the assertion would fail. Instead of throwing an assertion error, that case should be handled either by throwing a more descriptive error, or bailing out of the optimization (probably the latter).

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+        elif rres and self._contains_midcond(lhs):
+            return None
+
+        condside = lhs if lres else rhs
+        if condside.function_name == "equals":
+            return self._get_domain_of_equals(condside)
+        else:
+            return self._get_domain_of_in(condside)
+
+    def _domain_union(self, e1: Domain, e2: Domain) -> Domain:
```
evanh, line 168: This is another example of a function that is used only once and doesn't seem complicated enough to warrant a stand alone function.

snuba/query/processors/logical/filter_in_select_optimizer.py
evanh: This function would read better if it checked the positive instead of eliminating the negative.

`return f.function_name in ("equals", "in") and isinstance(f.parameters[0], Column) and f.parameters[0].column_name == "metric_id"`

evanh (commented):
Generally I think the logic of this is makes sense. Some stylistic nits and questions.

snuba/query/processors/logical/filter_in_select_optimizer.py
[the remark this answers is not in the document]
volokluev: That is not clear until you read the code for a while. A comment is very helpful if you're introducing new terminology

tests/query/processors/test_filter_in_select_optimizer.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+@pytest.mark.parametrize("logical_query, expected_domain", logical_test_cases)
+def test_get_domain_of_logical(logical_query: Query, expected_domain: set[int]):
+    assert optimizer.get_domain_of_query(logical_query) == expected_domain
+
+
+""" --- MQL TEST CASES --- """
+
+mql_queries = [
```
enochtangg, line 67: Can we validate with a test that the expected results of a query is still the same, with or without the processor?

enochtangg (commented):
What is the behaviour of this processor when a query's `SELECT` clause aggregates two different metric_ids separately? Can we write a test that captures that case?

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,188 @@
...
+        It is always acceptable to have the following clause in a query:
+            WHERE metric_id in metric_id_domain
+        """
+        for select_exp in query.get_selected_columns():
+            curr = self._get_domain_of_exp(select_exp.expression)
+            if curr is not None:
+                return curr
+        return None
+
+    def domain_to_query_filter(self, domain: Domain) -> Expression | None:
```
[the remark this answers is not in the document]
kylemumma: I see what you are saying, and could see why inlining small non-reused functions could improve readability. What are your thoughts on the following arguments in favor of keeping small functions separate:
* Clarity and Focus: Small, well-named functions can make the code easier to understand at a glance, as each function's name acts as a form of documentation.
* Maintainability: Even if a function is only used once now, future development might find new use cases for it. Having it as a separate function from the start makes it easier to reuse.

I also noticed in our codebase that many functions can tend to grow very large in size and take on many responsibilities at once, which made it hard to follow.

tests/query/processors/test_filter_in_select_optimizer.py
kylemumma: I had the same idea but I couldnt figure out an easy way to do this. After asking Evan about it, I think he didnt think it would be necessary. As a safeguard against breaking existing queries I put the optimizer behind a feature flag.

kylemumma: > What is the behaviour of this processor when a query's SELECT clause aggregates two different metric_ids separately? Can we write a test that captures that case?

You bring up a good point. Currently the optimizer assumes that if there is any conditional aggregate function in the query, it's domain is the domain of the entire query. 

Thus if I have a query like `SELECT sumIf(value, metric_id=10), sumIf(value, metric_id=11)`

It will compute the domain of the first sumIf as 10, be finished, and add `WHERE metric_id=10`
Thus the query will be broken.

snuba/query/processors/logical/filter_in_select_optimizer.py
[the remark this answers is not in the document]
volokluev: I personallly don't have strong feelings on this and I don't think Evan does either which is why it's a nit. I like the separated out functions but like with everything it's a balance. I don't think you need to change anything here personally.

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def process_mql_query(
+        self, query: LogicalQuery | CompositeQuery[QueryEntity]
+    ) -> None:
+        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=0)
+        if feat_flag:
+            try:
+                domain = self.get_domain_of_mql_query(query)
+            except ValueError:
+                domain = {}
```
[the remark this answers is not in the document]
kylemumma: also should I add an overall metric for how many queries get optimized, or log for which ones

kylemumma:
ready for review, i left a couple questions in the comment. I was hoping to leave the failing codecov as failing bc the part it wants me to cover I think is fine to leave uncovered.

snuba/query/processors/logical/filter_in_select_optimizer.py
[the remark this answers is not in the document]
evanh: I think it's OK for now. We can add that later if we need to debug something.

tests/query/processors/test_filter_in_select_optimizer.py
```diff
@@ -0,0 +1,156 @@
...
+        "status_code": 222222,
+        "transaction": 333333,
+    },
+    "limit": None,
+    "offset": None,
+}
+
+""" IMPORTANT STUFF """
+optimizer = FilterInSelectOptimizer()
+mql_test_cases: list[tuple[str, dict]] = []
```
evanh, line 42: Shouldn't this have something in it?

evanh (commented):
Looks good, just need to a few more tests.

tests/query/processors/test_filter_in_select_optimizer.py
kylemumma: sorry it actually gets initialized on line 67, ill move it up to make less confusing

kylemumma:
I moved the test case definitions up to make them more clear in the file

evanh (approved):

PR #5611, opened 2024-03-01
feat: bump-version action also bumps rust

lynnagara:
I always forget to bump the rust versions of our libraries since it's not covered by this action


untitaker (approved):

untitaker:
thanks, i wanted to do this but forgot

PR #5613, opened 2024-03-01
feat(capman): Split global and tenant specific configs into two tables

john-z-yang:
### Overview

As as user it's a little hard to identify which option is applied globally and which is applied to a specific tenant.
This change splits the 2 types of configurations into separate tables.

The UI element distinguishes whether or not a configuration is global or not by checking if the `params` field is empty, based on the assumption that a configuration is applied globally iff it does not have params.

Also adds some handy functions in tests to verify table contents.

Before
image

After
image

### DIFF
--- a/snuba/admin/static/capacity_management/allocation_policy.tsx
+++ b/snuba/admin/static/capacity_management/allocation_policy.tsx
@@ -144,6 +144,9 @@ function AllocationPolicyConfigs(props: {
               row_data.edit,
             ])}
           columnWidths={[3, 2, 5, 1, 1]}
+          customStyles={createCustomTableStyles({
+            headerStyle: { backgroundColor: getTableColor(policy.configs) },
+          })}
         />
         <p style={paragraphStyle}>
           These are the tenant specific configurations.
--- a/snuba/admin/static/capacity_management/allocation_policy.tsx
+++ b/snuba/admin/static/capacity_management/allocation_policy.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from "react";
 
-import { Table, createCustomTableStyles} from "../table";
+import { Table, createCustomTableStyles } from "../table";
 import { COLORS } from "../theme";
 import Client from "../api_client";
 import { AllocationPolicy, AllocationPolicyConfig } from "./types";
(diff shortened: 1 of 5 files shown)


snuba/admin/static/capacity_management/allocation_policy.tsx
```diff
@@ -135,12 +157,23 @@ function AllocationPolicyConfigs(props: {
...
+              row_data.description,
+              row_data.type,
+              row_data.edit,
+            ])}
           columnWidths={[3, 3, 2, 5, 1, 1]}
-          customStyles={createCustomTableStyles({headerStyle: {backgroundColor: getTableColor(policy.configs)}})}
-
+          customStyles={createCustomTableStyles({
+            headerStyle: { backgroundColor: getTableColor(policy.configs) },
+          })}
```
volokluev, line 176: You need this in the global table as well

snuba/admin/static/capacity_management/row_data.tsx
```diff
@@ -6,30 +6,46 @@ function getReadonlyRow(
...
-    config.type,
-    <Button
-      variant="outline-secondary"
-      onClick={() => edit()}
-      data-testid={config.name + "_edit"}
-    >
-      edit
-    </Button>,
-  ];
+  return {
```
volokluev, line 9: is this just prettier?

snuba/admin/static/capacity_management/allocation_policy.tsx
volokluev: otherwise you're gonna have one table changing colors and the other not

volokluev (commented):
Nice job, there's a small bug, once you fix it, should be fine

snuba/admin/static/capacity_management/row_data.tsx
john-z-yang: I think this makes it more readable, because this was a list before, so when feeding the result of this function into the <Table> element we need to remove an element by index. Making this an object shows explicitly what we information we are omitting.

snuba/admin/static/capacity_management/allocation_policy.tsx
john-z-yang: Good catch, thank you

snuba/admin/static/capacity_management/row_data.tsx
volokluev: ah I see

volokluev (approved):

PR #5615, opened 2024-03-01
fix(COGS): Fix app_feature for generic metrics Clickhouse compute accounting

ayirr7:
Uses the entity instead of the dataset for generic metrics -- this is a more reliable source for getting the `app_feature` from a query than dataset since it comes directly from the request body. 

I did notice that the entity in a `SnubaQueryMetadata` object is known only if the query is a `LogicalQuery`. Under which circumstances would it not be a `LogicalQuery`? If I understand correctly, the other kinds of queries would be nested/subqueries, and queries involving joins. I don't know how common these two cases are for generic metrics, but we can also fall back to dataset in these cases.

### DIFF
--- a/snuba/querylog/__init__.py
+++ b/snuba/querylog/__init__.py
@@ -111,13 +111,18 @@ def _record_cogs(
     if not profile or (bytes_scanned := profile.get("progress_bytes")) is None:
         return
 
+    # The dataset is usually a good proxy for app_feature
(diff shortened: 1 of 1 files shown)


evanh:
> Under which circumstances would it not be a LogicalQuery?

If the query is a JOIN/subquery then it will be a `CompositeQuery` class. In those cases there's no simple Entity to look at.

evanh (approved):

PR #5617, opened 2024-03-04
feat(generic-metrics): Add a killswitch to processor

ayirr7:
Still kind of a WIP. The initial idea was to use the header to grab the namespace/use case from there, but it turns out we don't have support for "getting" headers in rust-arroyo. Additionally, it would be tricky to decode/deserialize the header ourselves. Lastly, I think there's a possibility that we do not even forward header information into the Rust processors in Snuba -- but I would need to double check this. 

As a result, I tried to use an approach where we partially deserialize the payload to only grab the use case and effectively discard the other fields. We check against the use case to determine the killswitch. 

We need to add the actual runtime config, then I will change the name of the key from `key` to the actual key we want.


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -240,6 +246,17 @@ where
     T: Parse + Serialize,
 {
     let payload_bytes = payload.payload().context("Expected payload")?;
+    let use_case: MessageUseCase = serde_json::from_slice(payload_bytes)?;
+    let killswitch_config = get_str_config("key");
+
+    if killswitch_config.is_ok() {
+        if let Some(str) = killswitch_config.unwrap() {
```
anonrig, line 253: I think it will be more readable if we moved all these lines into from_str() kind of function in the new structure that's added

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -240,6 +246,17 @@ where
...
+            if use_case.use_case_id == str {
+                return Ok(InsertBatch::skip());
+            }
+        }
+    }
```
Dav1dde, line 258: Untested:

```suggestion
    if let Some(killswitch) = killswitch_config.ok().flatten() {
        let use_case: MessageUseCase = serde_json::from_slice(payload_bytes)?;
        if killswitch == use_case.use_case_id {
            return Ok(InsertBatch::skip());
        }
    }
```

1) Only pre-parse the message if you need to (there is a killswitch)
2) Rust is a lot more powerful in pattern matching and combinators, which means you basically never have to `unwrap()`

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -240,6 +246,17 @@ where
     T: Parse + Serialize,
 {
     let payload_bytes = payload.payload().context("Expected payload")?;
+    let use_case: MessageUseCase = serde_json::from_slice(payload_bytes)?;
+    let killswitch_config = get_str_config("key");
```
Dav1dde, line 250: How do you configure multiple use case killswitches, e.g. how do I killswitch `custom` and `transactions`?

Dav1dde (commented):
Just a a quick review about Rust things, I don't really have a good understanding of the Snuba code (yet).

Also maybe it makes sense to switch to an intermediate format like `serde_json::Value` instead of just passing around bytes, this means you can skip on deserializing multiple times.

rust_snuba/src/processors/generic_metrics.rs
john-z-yang, line 250: Is `key` the name of the runtime config? If so, we should name it to something that is more descriptive

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -64,6 +65,11 @@ struct FromGenericMetricsMessage {
     aggregation_option: Option<String>,
 }
 
+#[derive(Debug, Deserialize)]
+struct MessageUseCase {
+    use_case_id: String,
+}
```
john-z-yang, line 71: @Dav1dde I am not familiar with the serde create to know if this is "fast enough". I think you're the best person to comment on the approach here.

The killswitch will be inactive most of the time, so deserializing this struct needs to be fast enough to justify this additional step during normal operation. Otherwise I would prefer it if we just check the `use_case_id` field of the fully deserialized struct.

Dav1dde: It will definitely go through the entire data an additional time, it does not stop early, that being said it's also not fully deserializing the entire JSON.

What should work is:

```
    let msg: FromGenericMetricsMessage = if let Some(killswitch) = killswitch_config.ok().flatten() {
        let data: serde_json::Value = serde_json::from_slice(payload_bytes)?;
        let use_case_id = data.as_object().and_then(|obj| obj.get("use_case_id"));
        if Some(killswitch) == use_case.use_case_id {
            return Ok(InsertBatch::skip());
        }
        let msg = serde_json::from_value(data)?;
    } else {
        serde_json::from_slice(payload_bytes)?;
    }
```

But this is getting quite complicated, does having the killswitch provide us any other benefit over just having a DLQ?

rust_snuba/src/processors/generic_metrics.rs
ayirr7: That was a placeholder name, sorry! I think we would need to create the actual runtime config first

ayirr7: We'll use the Snuba runtime config convention, which just passes in a list as a string, so a killswitch config would look like "[use_case_1, use_case_2]"

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -243,6 +249,14 @@ impl Parse for CountersRawRow {
     }
 }
 
+fn should_use_killswitch(config: Result<Option<String>, Error>, use_case: &MessageUseCase) -> bool {
+    if let Some(killswitch) = config.ok().flatten() {
+        return killswitch.contains(use_case.use_case_id.as_str());
+    }
+
+    false
```
john-z-yang, line 257: Nit, maybe using
```rust
config.ok().flatten().map_or(false, |killswitch| {
    killswitch.contains(use_case.use_case_id.as_str())
})
```
is more ergonomic here

rust_snuba/src/processors/generic_metrics.rs
john-z-yang: This seems a little janky, but I'm guessing this is because we are getting these configs through python and redis is not really typed?

john-z-yang (approved):
This looks good to me, seems like the tests are testing the `should_use_killswitch`, so I recommend testing this locally. Made some minor comments but they shouldn't be blocking

PR #5627, opened 2024-03-06
feat(generic-metrics): Add metrics around encoding format type in processor

ayirr7:
These metrics will allow us to check which "branch" of the processor we're in, and validate the new encoded format is indeed being deserialized in the processor.

### DIFF
--- a/rust_snuba/src/processors/generic_metrics.rs
+++ b/rust_snuba/src/processors/generic_metrics.rs
@@ -13,7 +13,7 @@ use crate::{
     KafkaMessageMetadata, ProcessorConfig,
 };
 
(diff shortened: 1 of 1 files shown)


Swatinem (approved):

john-z-yang (approved):

Dav1dde (approved):

Dav1dde (approved):
We could use the same counter and tag it with the format.

```rs
counter!("generic_metrics.format_count", format = "legacy");
counter!("generic_metrics.format_count", format = "encoded");
```

john-z-yang (approved):

PR #5628, opened 2024-03-06
the default value of trace_id will be a randomly generated uuid inste…

xurui-c:
…ad of empty string

### Legal Boilerplate

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -504,7 +504,7 @@ def _raw_query(
     timer: Timer,
     # NOTE: This variable is a piece of state which is updated and used outside this function
     stats: MutableMapping[str, Any],
-    trace_id: Optional[str] = None,
+    trace_id: str = str(uuid.uuid4()),
     robust: bool = False,
 ) -> QueryResult:
     """
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import textwrap
+import uuid
 from dataclasses import replace
 from functools import partial
 from math import floor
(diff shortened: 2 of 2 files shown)


snuba/web/query.py
```diff
@@ -338,7 +339,7 @@ def _format_storage_query_and_run(
             stats=stats,
             query_metadata_list=query_metadata.query_list,
             query_settings={},
-            trace_id="",
+            trace_id=str(uuid.uuid4()),
```
volokluev, line 342: I think it would be better to use a known uuid as opposed to random one like 

```
uuid.UUID('00000000-0000-0000-0000-000000000000')
```

snuba/web/db_query.py
```diff
@@ -504,7 +504,7 @@ def _raw_query(
     timer: Timer,
     # NOTE: This variable is a piece of state which is updated and used outside this function
     stats: MutableMapping[str, Any],
-    trace_id: Optional[str] = None,
+    trace_id: str = str(uuid.uuid4()),
```
volokluev, line 507: why do you need this? also this will make it so the uuid that is calculated at import time is the default argument for every subsequent call.

xurui-c: If the `trace_id` isn't passed in here, I think we should populate it with a default non-null `trace_id`, so that if an error occurs and the QueryException is written to querylog, we won't have a null `trace_id`

volokluev: ok let's do the default all zeroes one by default then to cover our bases

volokluev (commented):
nit: you're reusing the same constant value in your code, it's good to define it somewhere like 

```
# describe what it's for
DEFAULT_EMPTY_TRACE_ID = "00000000-0000-0000-0000-000000000000"
```

I would recommend maybe `settings/__init__.py`

volokluev (approved):

untitaker (approved):

PR #5630, opened 2024-03-06
feat(admin): Absolute imports in snuba-admin

volokluev:
No more having to import `../../../../Something` now we can import from the root 

```ts
import Thing from "SnubaAdmin/Stuff/Thing";
```


john-z-yang (approved):
🫡

dbanda (approved):

PR #5632, opened 2024-03-06
fix(snql) Correctly parse tag columns with an entity alias

evanh:
If a query had a tag column with an entity alias, e.g. `events.tags[status]`,
the parser would choke and would throw an exception.

### DIFF
--- a/tests/test_snql_api.py
+++ b/tests/test_snql_api.py
@@ -1333,6 +1333,38 @@ class TestSnQLApi(BaseApiTest):
 
         assert response.status_code == 200
 
+    def test_tags_column_in_join(self) -> None:
+        response = self.post(
+            "/events/snql",
+            data=json.dumps(
+                {
+                    "dataset": "events",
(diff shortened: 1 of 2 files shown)


tests/test_snql_api.py
```diff
@@ -1333,6 +1333,38 @@ def test_tags_key_column(self) -> None:
...
+        response = self.post(
+            "/events/snql",
+            data=json.dumps(
+                {
+                    "dataset": "events",
+                    "query": f"""MATCH (events: events) -[attributes]-> (ga: group_attributes)
+                        SELECT count() AS `count`
+                        BY events.time
+                        WHERE events.timestamp >= toDateTime('{self.base_time.isoformat()}')
+                        AND ifNull(events.tags[service-class], '') != 'devtest'
```
volokluev, line 1346: so in this case service-class is a column on the entity?

evanh: No, `tags` is a column on the entity.

volokluev: ah, got it

volokluev (approved):

PR #5635, opened 2024-03-07
ref(codecov) Try out the failed test feature in Codecov

evanh:
Reference: https://docs.codecov.com/docs/test-result-ingestion-beta

Screenshot 2024-03-08 at 12 14 00 PM


evanh:
Note that I intentionally failed a test so we could test this.

enochtangg (approved):

PR #5636, opened 2024-03-07
ref(gocd): add comparer pipeline, consolidate script?

MeredithAnya:
Tried to make one script that could be used by all three pipelines:
* `fetcher` - https://github.com/getsentry/snuba/pull/5625
* `replayer` - https://github.com/getsentry/snuba/pull/5609
* `comparer`  - This PR includes that pipeline

Also adds `--try-deployments-and-statefulsets` from https://github.com/getsentry/devinfra-deployment-service/pull/534

### DIFF
--- a/gocd/templates/bash/s4s-clickhouse-queries.sh
+++ b/gocd/templates/bash/s4s-clickhouse-queries.sh
@@ -32,4 +32,4 @@ eval $(/devinfra/scripts/regions/project_env_vars.py --region="${SENTRY_REGION}"
   "snuba-query-${SNUBA_CMD_TYPE}" \
   "us.gcr.io/sentryio/snuba:${GO_REVISION_SNUBA_REPO}" \
   -- \
-  snuba "query-${SNUBA_CMD_TYPE} ${ARGS}"
+  snuba "query-${SNUBA_CMD_TYPE} ${ARGS[@]}"
--- a/gocd/templates/bash/s4s-clickhouse-queries.sh
+++ b/gocd/templates/bash/s4s-clickhouse-queries.sh
@@ -29,6 +29,7 @@ eval $(/devinfra/scripts/regions/project_env_vars.py --region="${SENTRY_REGION}"
 /devinfra/scripts/k8s/k8s-spawn-job.py \
   --label-selector="service=snuba-${SNUBA_SERVICE_NAME}" \
   --container-name="${SNUBA_SERVICE_NAME}" \
+  --try-deployments-and-statefulsets \
   "snuba-query-${SNUBA_CMD_TYPE}" \
   "us.gcr.io/sentryio/snuba:${GO_REVISION_SNUBA_REPO}" \
   -- \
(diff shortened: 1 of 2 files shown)


gocd/templates/clickhouse-query-comparer.jsonnet
```diff
@@ -0,0 +1,67 @@
...
+      SNUBA_CMD_TYPE: 'comparer',
+    },
+    elastic_profile_id: pipeline_group,
+    tasks: [
+      gocdtasks.script(importstr './bash/s4s-clickhouse-queries.sh'),
+    ],
+  };
+
+local pipeline = {
+  environment_variables: {
```
dbanda, line 19: You want all the arguments to be set in gocd under here. I'm assuming that would be QUERYLOG_HOST e.t.c

gocd/templates/clickhouse-query-comparer.jsonnet
```diff
@@ -0,0 +1,67 @@
+local gocdtasks = import 'github.com/getsentry/gocd-jsonnet/libs/gocd-tasks.libsonnet';
+
+local pipeline_group = 'snuba';
+local pipeline_name = 'clickhouse-query-fetcher';
+
+local generate_compare_job() =
```
dbanda, line 6: maybe this should be something like `generate_query_job()` since it no longer just compare.

gocd/templates/clickhouse-query-comparer.jsonnet
MeredithAnya: @dbanda I was only putting the args that are needed for the comparer here - since we don't need the other args

gocd/templates/bash/s4s-clickhouse-queries.sh
```diff
@@ -0,0 +1,36 @@
...
+
+SNUBA_SERVICE_NAME="query-${SNUBA_CMD_TYPE}-gocd"
+
+eval $(/devinfra/scripts/regions/project_env_vars.py --region="${SENTRY_REGION}")
+/devinfra/scripts/k8s/k8stunnel
+
+/devinfra/scripts/k8s/k8s-spawn-job.py \
+  --label-selector="service=snuba-${SNUBA_SERVICE_NAME}" \
+  --container-name="${SNUBA_SERVICE_NAME}" \
+  --try-deployments-and-statefulsets \
```
MeredithAnya, line 32: @dbanda did I add this flag correctly?

gocd/templates/clickhouse-query-comparer.jsonnet
MeredithAnya: I think whenever we make this generic we can, but this file is still specific for the comparer for now.

gocd/templates/bash/s4s-clickhouse-queries.sh
dbanda: looks good. i don't know if dev-infra has deployed it out yet

dbanda (approved):

PR #5637, opened 2024-03-08
Allows empty `trace_id`

xurui-c:
# fix: Handles empty `trace_id` in payload

[Setting a default `trace_id` upstream](https://github.com/getsentry/snuba/pull/5628) proves to be insufficient to solve the [UUID parsing error](https://sentry.sentry.io/issues/4927370476/?project=300688&query=start%3D2024-02-07T15%3A27%3A00%26end%3D2024-02-07T22%3A23%3A38%26groupStatsPeriod%3Dauto&referrer=alerts-related-issues-issue-stream). It is also not sustainable to find all the places where `trace_id` is set to empty or null since it does not protect against future cases. This pr addresses this issue by adding deserialization logic that populates the query's `trace_id` with `Uuid::nil()` if the payload's `trace_id` is an empty string or null.

### Legal Boilerplate

### DIFF
--- a/rust_snuba/src/processors/querylog.rs
+++ b/rust_snuba/src/processors/querylog.rs
@@ -17,7 +17,7 @@ pub fn process_message(
     _config: &ProcessorConfig,
 ) -> anyhow::Result<InsertBatch> {
     let payload_bytes = payload.payload().context("Expected payload")?;
-    let payload_json: Value = serde_json::from_slice(&payload_bytes)?;
+    let payload_json: Value = serde_json::from_slice(payload_bytes)?;
     let from: FromQuerylogMessage = payload_json.try_into()?;
 
     let querylog_msg = QuerylogMessage {
--- a/rust_snuba/src/processors/querylog.rs
+++ b/rust_snuba/src/processors/querylog.rs
@@ -9,8 +9,6 @@ use serde::{ser::Error, Deserialize, Serialize, Serializer};
 use serde_json::Value;
 use uuid::Uuid;
 
-use crate::processors::utils;
-
 use crate::types::{InsertBatch, KafkaMessageMetadata};
 
 pub fn process_message(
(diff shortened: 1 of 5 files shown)


Swatinem (commented):
Instead of implementing this as a `TryFrom`, you should rather change the annotations for the `trace_id` property, and make sure it has a different default.

volokluev (approved):

rust_snuba/src/processors/querylog.rs
```diff
@@ -170,6 +168,49 @@ struct QueryList {
...
+        let mut query_list = vec![];
+
+        let from_query_list: Vec<Value> =
+            serde_json::from_value(from["query_list"].clone()).unwrap();
+
+        for query in from_query_list {
+            let mut query_trace_id = Uuid::nil();
+            if query["trace_id"] != "" && !query["trace_id"].is_null() {
+                query_trace_id =
+                    Uuid::parse_str(query["trace_id"].to_string().get(1..33).unwrap())?;
```
Swatinem, line 189: The `.get(1..33).unwrap()` will panic (crash the consumer) on any string that is not the desired length.

Also, the lower bound is inclusive, the upper one is exclusive, so you are cutting away the first character here.

Swatinem (commented):
Fixing all of this upstream would surely be preferable.

But I was not talking about any upstream.

Rather, you are parsing into a generic `Value`, and then manually transforming that into a `FromQuerylogMessage`, which is quite inefficient on its own, not to mention all the `clone` calls inside that are inefficient as well.

You have a `#[derive(Deserialize)]` on the outer `FromQuerylogMessage`, and on the `FromQuery` which I believe is the main concern here which contains the `trace_id`.

My idea was to plan around with the type and the serde attributes of the `FromQuery::trace_id` field.

It should be possible to keep all the auto-generated and efficient deserialization code around, but just change the way that particular property is deserialized.

If I understand correctly, the problem is that instead of a serialized `Uuid`, upstream provides `null | "" | "serialized uuid"`, so you would have to deserialize that.

Unfortunately, the `null` means you *have to* use `Option<Something>` as the target type :-( And then there are some ways to continue on. You might as well deserialize to an `Option<String>` and then parse the uuid manually in a later step, similar to how you do it right now.

volokluev (approved):

untitaker (approved):

PR #5638, opened 2024-03-08
feat(gocd): Add snuba API canary health check

enochtangg:
### Overview
Now that we deploy snuba API canary in GoCD, we can validate that snuba is able to connect to all the appropriate ClickHouse clusters before deploying to SaaS. As we continue to make snuba more self-serve, this safeguard can help product developers stop bad deploys (incorrect storage names, wrong table names, etc.) from going out to production. Additionally, this check would stop developers from prematurely enabling a storage (via readiness_state), before the necessary migrations are executed and tables are created.

### Expected Behaviour
This is currently set to `dry_run=true` and should not stop deploys even if check fails. We should be able to see this new check in GoCD UI.

### DIFF
--- a/gocd/templates/bash/canary-ddog-health-check.sh
+++ b/gocd/templates/bash/canary-ddog-health-check.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+/devinfra/scripts/checks/datadog/monitor_status.py --dry-run=true \
+  140973101
+
+
+# Above monitor IDs 140973101 map to following monitors:
+# Snuba API Health Check is Failing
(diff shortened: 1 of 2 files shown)


volokluev (approved):
Is there a dev-infra PR to be reviewed with this?

enochtangg:
@volokluev nope, we're using an [existing script](https://github.com/getsentry/devinfra-deployment-service/blob/main/gocd_agent/scripts/checks/datadog/monitor_status.py) from the devinfra-deployment-service to accomplish this. I've asked dev-infra to take a look though.

IanWoodard (commented):
Forgot to leave a review (discussed on slack), lgtm

PR #5639, opened 2024-03-08
ref(gocd): use shared script query-fetcher

MeredithAnya:
Assuming the shared script in https://github.com/getsentry/snuba/pull/5636 works totally fine, I'm updating the `query-fetcher` to use that as well, and I can remove the script that I made for it previously


gocd/templates/bash/s4s-clickhouse-queries.sh
```diff
@@ -13,12 +13,12 @@ fi
 
 if [ "${SNUBA_CMD_TYPE}" == "replayer" ]
 then
-    ARGS=$COMPARER_ARGS
+    ARGS=$REPLAYER_ARGS_ARGS
 fi
 
 if [ "${SNUBA_CMD_TYPE}" == "comparer" ]
 then
-    ARGS=$REPLAYER_ARGS_ARGS
```
MeredithAnya, line 21: also had to fix this oopies

PR #5640, opened 2024-03-08
fix: Fix a bug in HexIntColumnProcessor that skipped array conditions

evanh:
The HexIntColumnProcessor was only working on IN conditions that had a tuple as
the argument. Modify it to also accept arrays.


volokluev (approved):
YESSSSSSS

image

PR #5652, opened 2024-03-13
fix: vscode testing ui

kylemumma:
removes unittest from vscode settings, since we only use pytest, this allows vscode testing ui to work


volokluev (approved):

dbanda (approved):

volokluev (approved):

nikhars (approved):
🎉

PR #5653, opened 2024-03-13
fix: update filter in select optimizer

kylemumma:
this relates to #5610
* feature flag defaults to on
* sentry capture_exception in the case of internal optimizer error (includes unsupported optimization form)
* added a test

update:
* modify test_mql_formula_queries to pass with the new optimizer
* extract domain -> condition conversion to its own method

### DIFF
--- a/snuba/query/processors/logical/filter_in_select_optimizer.py
+++ b/snuba/query/processors/logical/filter_in_select_optimizer.py
@@ -1,3 +1,5 @@
+import sentry_sdk
+
 from snuba.query.composite import CompositeQuery
 from snuba.query.conditions import binary_condition
 from snuba.query.data_source.simple import Entity as QueryEntity
@@ -43,11 +45,12 @@ class FilterInSelectOptimizer:
     def process_mql_query(
         self, query: LogicalQuery | CompositeQuery[QueryEntity]
     ) -> None:
-        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=0)
+        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=1)
         if feat_flag:
             try:
                 domain = self.get_domain_of_mql_query(query)
-            except ValueError:
+            except ValueError as err:
+                sentry_sdk.capture_exception(err)
                 domain = {}
 
             if domain:
--- a/snuba/query/processors/logical/filter_in_select_optimizer.py
+++ b/snuba/query/processors/logical/filter_in_select_optimizer.py
@@ -55,27 +55,47 @@ class FilterInSelectOptimizer:
 
             if domain:
                 # add domain to where clause
-                domain_filter = None
-                for key, value in domain.items():
-                    clause = binary_condition(
-                        "in",
-                        key,
-                        FunctionCall(
-                            alias=None,
-                            function_name="array",
-                            parameters=tuple(value),
-                        ),
-                    )
-                    if not domain_filter:
-                        domain_filter = clause
-                    else:
-                        domain_filter = binary_condition(
-                            "and",
-                            domain_filter,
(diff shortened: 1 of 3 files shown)


kylemumma:
**modify test_mql_formula_queries to pass with the new optimizer.**
The optimizer adds an extra condition to the where clause of many mql formula queries, this was causing these existing tests to failed, so I updated them accordingly. I have no idea why this didnt fail in the earlier feature PR.

** extract domain -> condition conversion to its own method **
When updating the tests (above), the conditions added to the query were in a random order that changed each time the test was run. This made it impossible to test ==. Thus I added sorting to the domain -> condition conversion step. I thought this made the block of code a little long and complex so I extracted into its own method.

tests/query/parser/test_formula_mql_query.py
```diff
@@ -259,7 +279,7 @@ def test_simple_formula() -> None:
             ),
         ],
         groupby=[time_expression],
-        condition=formula_condition,
+        condition=binary_condition("and", metric_id_in({123456}), formula_condition),
```
kylemumma, line 282: @evanh modifying this file made me think: should the parsing of queries be more distinct than the post processing step.

evanh: What do you mean by this?

tests/query/processors/test_filter_in_select_optimizer.py
```diff
@@ -151,6 +163,4 @@ def test_get_domain_of_mql(mql_query: str, expected_domain: set[int]) -> None:
     logical_query, _ = parse_mql_query(str(mql_query), mql_context, generic_metrics)
     assert isinstance(logical_query, Query)
     res = optimizer.get_domain_of_mql_query(logical_query)
```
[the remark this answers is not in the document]
evanh: The general solution to this problem would be to not use `parse_mql_query` at all, and instead have the input to your test be a full query object.

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -43,36 +45,57 @@ class FilterInSelectOptimizer:
...
         self, query: LogicalQuery | CompositeQuery[QueryEntity]
     ) -> None:
-        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=0)
+        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=1)
         if feat_flag:
             try:
                 domain = self.get_domain_of_mql_query(query)
-            except ValueError:
+            except ValueError as err:
+                sentry_sdk.capture_exception(err)
```
evanh, line 53: I am worried that this is going to trigger our Sentry alerts. We discussed maybe using `logger.warning` as an alternative, so these get sent to Sentry as warnings instead of full errors.

kylemumma: oh yes forgot about this, will change

tests/query/parser/test_formula_mql_query.py
kylemumma: here my thinking:
formula mql -> initial ast -> post processed ast

this file tests given a formula mql, is the expected post processed ast generated. this makes it s.t. adding the new post processor requires me to modify these tests.

another design could be: given a formula mql, make sure the correct initial ast is generated
and then: given and initial ast, make sure the correct post processed ast is generated

this isnt possible due to the fact that parsing and post processing are tightly coupled

not saying we should actually do this, just something I thought of when I was modifying the tests

kylemumma: no need to reply to this I will bring it up in self serve since its relevant and see if they have any opinions

PR #5655, opened 2024-03-14
feat: Add use_case_id index to generic metrics

evanh:
The use_case_id is being sent with every query, and in the cases where a
metric_id is not specified, can be used to very effectively reduce the number
of rows scanned.

The granularity is set a little higher than for the tags hash maps, since the
cardinality of use case ID is quite high and the data should be fairly clumpy
(metric_id should all be stored together and each metric_id has only one use
case ID).


snuba/snuba_migrations/generic_metrics/0029_add_use_case_id_index.py
```diff
@@ -0,0 +1,76 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddIndex(
+                storage_set=StorageSetKey.GENERIC_METRICS_COUNTERS,
+                table_name="generic_metric_counters_aggregated_local",
+                index_name="set_idx_use_case_id",
+                index_expression="use_case_id",
+                index_type="set",
```
evanh, line 17: I opted for a `set` type here since use case IDs are very low cardinality, and our hardcoded in our repos, so they can't change easily.

snuba/snuba_migrations/generic_metrics/0029_add_use_case_id_index.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddIndex(
+                storage_set=StorageSetKey.GENERIC_METRICS_COUNTERS,
+                table_name="generic_metric_counters_aggregated_local",
+                index_name="set_idx_use_case_id",
+                index_expression="use_case_id",
+                index_type="set(0)",
+                granularity=100,
```
nikhars, line 18: The granularity on the indexes is a multiple on the original granularity of the table. In this case, the table has a granularity of 2048. If you set this granularity to 100, this index's granularity would become (2048 * 100 = 204800). I think what you would want here is granularity of 1. So that the set only contains unique use cases for 2048 rows. Higher the number of rows for the index, the more useless it would become.

nikhars (approved):

PR #5657, opened 2024-03-14
ref(dev-exp): Add Interface for Query Pipeline Stage

enochtangg:
### Overview
This PR is responsible for introducing a simple interface for representing a single stage in the snuba query execution pipeline. Currently, the query pipeline today is comprised of many functions that are deeply nested and scattered along the codebase. For instance, the [execute_entity_processors()](https://github.com/getsentry/snuba/blob/7ddd86444086ae250120fe9a2d1b53a8b76991d0/snuba/pipeline/simple_pipeline.py#L47) function is an example of a processing function that is hidden deep within multiple function calls and requires the creation of multiple objects to execute (even though those objects are not used in the function and arguably irrelevant to the context of the function).  

The purpose of this class is to provide an organized and transparent interface to execute specific processing steps with clearly defined inputs and outputs. These stages are designed to be chained and/or swapped among each other to form a more flexible query pipeline with an expected result.

Note: The structure of this class is subjected to change and evolve. As we continue to organize the pipeline (specifically entity processing), the interface will probably change. This is meant to be an initial starting point.


snuba/pipeline/query_pipeline.py
```diff
@@ -76,3 +81,51 @@ def build_planner(
...
+
+
+@dataclass
+class QueryPipelineResult(Generic[Tout]):
+    """
+    A container to represent the result of a query pipeline stage.
+    """
+
+    data: Optional[Tout]
+    error: Optional[Exception]
```
kylemumma, line 131: Is there any tradeoff between this design and an alternative design that uses subclassing: 

class QueryPipelineResult  # abstract
class QueryPipelineData is a QueryPipelineResult
class QueryPipelineError is a QueryPipelineError

another thought is execute could return either a QueryPipelineResult or a QueryPipelineError

snuba/pipeline/query_pipeline.py
```diff
@@ -76,3 +81,51 @@ def build_planner(
         self, query: Query, settings: QuerySettings
     ) -> QueryPlanner[ClickhouseQueryPlan]:
         raise NotImplementedError
+
+
+class QueryPipelineStage(Generic[Tin, Tout]):
```
kylemumma, line 86: does it matter to make this class abstract?

snuba/pipeline/query_pipeline.py
```diff
@@ -76,3 +81,51 @@ def build_planner(
...
+        >>>    def _execute(self, input: QueryPipelineResult[LogicalQuery]) -> QueryPipelineResult[LogicalQuery]:
+        >>>         try:
+        >>>             result = my_complex_processing_function(input.data)
+        >>>             return QueryPipelineResult(result, None)
+        >>>         except Exception as e:
+        >>>             return QueryPipelineResult(None, e)
+    """
+
+    @abstractmethod
+    def _execute(self, input: QueryPipelineResult[Tin]) -> QueryPipelineResult[Tout]:
```
kylemumma, line 114: It seems to me that `execute` should be the only interface method we define since from a top down view its the only thing we need a stage to do, execute input to give output. If the user wants to define a private helper function for execute they can do it or not, it doesnt matter.

volokluev: This is indeed the only method a user would need to define. That method is wrapped by default to allow expansion of the capabilities of the interface

snuba/pipeline/query_pipeline.py
volokluev: I would maybe put a `__post_init__` here to make sure at least data OR error is set. They can't both be None. You could also use a `Union` type to denote that it can be either

snuba/pipeline/query_pipeline.py
```diff
@@ -76,3 +81,51 @@ def build_planner(
...
+
+    def execute(self, input: QueryPipelineResult[Tin]) -> QueryPipelineResult[Tout]:
+        if input.error:
+            # Forward the error to next stage of pipeline
+            return QueryPipelineResult(None, input.error)
+        return self._execute(input)
+
+
+@dataclass
+class QueryPipelineResult(Generic[Tout]):
```
volokluev, line 125: style nit: I would maybe use a different typevar here, like a `Generic[T]` or something to decouple the type definition from the pipeline itself. but I don't think it really matters that much

volokluev (approved):
good start, we can iterate as we actually do the untangling work.

snuba/pipeline/query_pipeline.py
enochtangg: Subclassing can also give us some flexibility if we wanted to add different attributes or methods to the result (serialization?). Since there is no real use-case for that at the moment, we could evolve this as see what it looks like in practice. Added type validator.

snuba/pipeline/query_pipeline.py
enochtangg: We should as it helps enforce a common interface. Added!

snuba/pipeline/query_pipeline.py
enochtangg: Added!

snuba/pipeline/query_pipeline.py
kylemumma: Ohh I see, that's cool.

PR #5660, opened 2024-03-15
chore: add observability to filter in select optimizer

kylemumma:
add some datadog and sentry logging


snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -45,9 +53,14 @@ def process_mql_query(
     ) -> None:
         feat_flag = get_int_config("enable_filter_in_select_optimizer", default=0)
         if feat_flag:
+            metrics.increment("kyles_optimizer_ran")
```
volokluev, line 56: you probably want to add something that is actually conditional on the optimizer doing something. This is too high up I think

kylemumma: I moved it down to the point where the condition of a query is actually modified

evanh (approved):
I approved this, but I would ask you to double check: Is there any metadata you want to add to the logger warning or the metric? Dataset, referrer etc.?

Another to think of it would be: if the metric/warning show a problem, what information would you want to have to debug that problem?

kylemumma:
> I approved this, but I would ask you to double check: Is there any metadata you want to add to the logger warning or the metric? Dataset, referrer etc.?
I was going to add dataset to the metric to make sure its only generic_metrics, and referrer just for extra info but I found the passing of information to the right level might take a bit of work so I left it out for now.

> Another to think of it would be: if the metric/warning show a problem, what information would you want to have to debug that problem?
I added the query to the logger warning message to help with debugging.

PR #5668, opened 2024-03-18
feat(devexp): Separate query processing from query execution stage

volokluev:
With this change, we can run two different query pipelines:

1. The old one (default in everything but tests)
2. The new one (using the new QueryPipelineStage interface)

The new query pipeline *does not* depend on the query runner to process the query and the query execution does not need anything except for the physical query (sort of, the `SnubaQueryMetadata` side effects are still there). 

The is mostly wrapping the old stuff with a new wrapper however there is a fundamental split in how this is executed now. 

Main new functionality added:
* Cluster to read from is now selected from the physical query (legacy pipeline uses the logical query)
    * this is necessary to query storages on their own

# Rollout plan

This is a delicate change and therefore requires precision. When this PR is merged, *nothing changes* except that we're passing the `storage_key` with the physical query (a pretty harmless change). 

Once deployed:

(In runtime config)
```
SET run_new_query_pipeline 0.01 # 1 % of queries go through new pipeline
```
... monitor sentry, datadog

Continue increasing the `run_new_query_pipeline` until we hit `1`

Once that woirks, we can remove all of the old pipeline in `web/query`


snuba/web/query.py
```diff
@@ -133,26 +135,87 @@ def _run_query_pipeline(
...
+                    entity.get_query_pipeline_builder().build_execution_pipeline(
+                        request, _query_runner
+                    )
+                )
+            else:
+                execution_pipeline = CompositeExecutionPipeline(
+                    input.data.query, input.data.query_settings, _query_runner
+                )
+
+            execution_pipeline.execute()
```
enochtangg, line 182: If I'm not mistaken, calling `execute()` on the execution pipeline will also call the query runner passed into the pipeline.

volokluev: yes but we passed in a query runner that only sets a variable. It does not run the query! This is little bit of magic that allows us to actually decouple execution from processing. This was not possible before because query splitters were a thing and used the runner multiple times

snuba/datasets/storage.py
```diff
@@ -89,6 +89,10 @@ def get_mandatory_condition_checkers(self) -> Sequence[ConditionChecker]:
         """
         return []
 
+    @abstractmethod
```
volokluev, line 92: Every storage should have a storage key. This was causing mypy issues and it is stupid

snuba/pipeline/stages/query_execution.py
```diff
@@ -0,0 +1,358 @@
...
+        return data_source.storage_key
+
+    def _visit_join(self, data_source: JoinClause[Table]) -> StorageKey:
+        return data_source.accept(StorageKeyJoinFinder())
+
+    def _visit_simple_query(self, data_source: ProcessableQuery[Table]) -> StorageKey:
+        return data_source.get_from_clause().storage_key
+
+    def _visit_composite_query(self, data_source: CompositeQuery[Table]) -> StorageKey:
+        return self.visit(data_source.get_from_clause())
```
volokluev, line 89: This is the only new functionality! Before we depended on the StorageQueryPlanBuilder to select the cluster for us, which gives us a `Reader`. No more! Now because of this we can get the cluster straight from the physical query.,

snuba/pipeline/stages/query_execution.py
```diff
@@ -0,0 +1,358 @@
...
+        if (
+            query_settings.get_turbo()
+            and not clickhouse_query.get_from_clause().sampling_rate
+        ):
+            clickhouse_query.set_from_clause(
+                replace(
+                    clickhouse_query.get_from_clause(),
+                    sampling_rate=snuba_settings.TURBO_SAMPLE_RATE,
+                )
+            )
```
volokluev, line 358: This is all copy-pasted from `snuba/web/query` to ensure that we are truly running two separate pipelines. open to just importing these functions but I like having the entire execution stage in one file. If this works, we can just remove all the duplicate code in `web/query.py`

snuba/pipeline/stages/query_processing.py
```diff
@@ -0,0 +1,45 @@
...
+    ) -> ClickhouseQuery | CompositeQuery[Table]:
+        _clickhouse_query = None
+
+        def _query_runner(
+            clickhouse_query: ClickhouseQuery | CompositeQuery[Table],
+            *args: Any,
+            **kwargs: str
+        ) -> None:
+            nonlocal _clickhouse_query
+            _clickhouse_query = clickhouse_query
```
volokluev, line 28: This is the key that let's us decouple the execution from the processing. we use the entire old processing pipline but we pass in a dummy runner which just lets us know what the query to run is.

snuba/settings/__init__.py
```diff
@@ -395,6 +395,9 @@ class RedisClusters(TypedDict):
 # to slice id
 LOGICAL_PARTITION_MAPPING: Mapping[str, Mapping[int, int]] = {}
 
+# Max query size that can be sent to clickhouse
+MAX_QUERY_SIZE_BYTES = 256 * 1024  # 256 KiB by default
```
volokluev, line 399: this was in `web/query.py` but should be a setting

snuba/settings/settings_test.py
```diff
@@ -68,3 +68,4 @@
     ]
 }
 VALIDATE_DATASET_YAMLS_ON_STARTUP = True
+USE_NEW_QUERY_PIPELINE_SAMPLE_RATE = 1
```
volokluev, line 71: run the new pipeline in tests

snuba/web/query.py
```diff
@@ -63,16 +94,28 @@ def parse_and_run_query(
     """
     # from_clause = request.query.get_from_clause()
     query_metadata = SnubaQueryMetadata(request, get_dataset_name(dataset), timer)
-
+    run_new_query_pipeline_rollout = state.get_float_config(
```
volokluev, line 97: here we run old and new in the same deploy and can easily switch using runtime config

snuba/query/data_source/simple.py
```diff
@@ -58,6 +59,7 @@ class Table(SimpleDataSource):
 
     table_name: str
     schema: PhysicalColumnSet
+    storage_key: StorageKey
```
volokluev, line 62: This allows us to get all the goodies on the storage. imo ideally we would just query from storages in the ClickhouseQuery (since all these things are properties of the storage). But I didn't want to change that in this PR as well

snuba/web/query.py
```diff
@@ -50,6 +55,32 @@
...
+        query_metadata=query_metadata,
+        robust=robust,
+        concurrent_queries_gauge=concurrent_queries_gauge,
+    ).execute(clickhouse_query)
+    if res.error:
+        raise res.error
+    elif res.data:
+        return res.data
+    # we should never get here
+    raise Exception("No result or data, very bad exception")
```
volokluev, line 81: Voila, the new pipeline. 

Pipeline chaining can probably be made to look nicer, but we can save that for another time

volokluev (commented):
Some context for reviewers

snuba/pipeline/stages/query_processing.py
```diff
@@ -0,0 +1,45 @@
...
+from snuba.datasets.plans.query_plan import QueryRunner
+from snuba.pipeline.composite import CompositeExecutionPipeline
+from snuba.pipeline.query_pipeline import QueryPipelineData, QueryPipelineStage
+from snuba.query.composite import CompositeQuery
+from snuba.query.data_source.simple import Table
+from snuba.query.logical import Query as LogicalQuery
+from snuba.request import Request
+
+
+class EntityAndStoragePipelineStage(
```
enochtangg, line 14: I know we eventually want to break this stage up into separate entity processing and storage processing stages. But could write a comment describing what this stage is responsible for now (e.g. execute entity logical processors, apply query translator, apply translation mappers, storage processors, etc.)?

enochtangg (commented):
Nice stuff! Two small things

snuba/pipeline/query_pipeline.py
```diff
@@ -111,15 +113,51 @@ class QueryPipelineStage(Generic[Tin, Tout]):
...
+        self, pipe_input: QueryPipelineError[Tin]
+    ) -> Union[Tout, Exception]:
+        """default behaviour is to just pass through to the next stage of the pipeline
+        Can be overridden to do something else"""
+        logging.exception(pipe_input.error)
+        return pipe_input.error
+
     @abstractmethod
-    def _execute(self, input: QueryPipelineResult[Tin]) -> QueryPipelineResult[Tout]:
+    def _process_data(self, pipe_input: QueryPipelineData[Tin]) -> Tout:
```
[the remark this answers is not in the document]
volokluev: I'm happy to change the name, what else could we call it?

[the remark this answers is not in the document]
enochtangg: Discussed offline 👍

snuba/pipeline/query_pipeline.py
```diff
@@ -111,15 +113,51 @@ class QueryPipelineStage(Generic[Tin, Tout]):
...
-    def execute(self, input: QueryPipelineResult[Tin]) -> QueryPipelineResult[Tout]:
-        if input.error:
-            # Forward the error to next stage of pipeline
-            return QueryPipelineResult(None, input.error)
-        return self._execute(input)
+    def execute(
+        self, pipe_input: QueryPipelineResult[Tin]
+    ) -> QueryPipelineResult[Tout]:
+        if pipe_input.error:
+            res = self._process_error(pipe_input.as_error())
```
evanh, line 132: It's very confusing that `pipe_input.error` is true, and then you process that error `as_error`, but the output is not necessarily an error, and in fact might be a valid result.

snuba/pipeline/stages/query_execution.py
```diff
@@ -0,0 +1,358 @@
...
+                reader=cluster.get_reader(),
+                cluster_name=cluster.get_clickhouse_cluster_name() or "",
+            )
+
+
+def _dry_run_query_runner(
+    clickhouse_query: ClickhouseQuery | CompositeQuery[Table],
+    cluster_name: str,
+) -> QueryResult:
+    # NOTE (Volo) : this is misleading behavior. If this runner is used with a split query,
```
evanh, line 147: Splits have been removed so this comment is out of date.

evanh (commented):
I think a nice way to test this would be to run both the new and old pipeline for every query, but the new one in dry run only. Then compare the output SQL from the new and old pipeline and log any differences.

snuba/pipeline/query_pipeline.py
volokluev: I can see why you say that but let me give a counter-example. When we report the results of a query to the querylog, we take in either the error or the query result, but in either case, the reporting is successful

enochtangg (approved):

snuba/pipeline/query_pipeline.py
evanh: That's true, what's confusing me is that there are also two classes: `QueryPipelineError` and `QueryPipelineResult`. Those types seem interchangeable? Somehow `pipe_input` is both? This makes sense if `QueryPipelineResult` contains an optional error/result, but it is confusing that the same object `res` can be either an error or a result. There's another layer of confusion that `_process_error`, which would seem to be exclusive to handling errors, can also sometimes output a non-error.

PR #5669, opened 2024-03-19
feat(metrics): Add a meta table to counters

evanh:
**Problem**

Queries that fetch all the metric IDs for a given org/project, and another query to fetch the tag values for a given tag key, take a long time because they are done on the main aggregate table. These queries don't actually care about the values, granularities etc., but are coming up against those problems.

**Solution**

This creates a table and a materialized view to populate it for storing meta information about metrics. This table is meant to satisfy queries that are trying to find metric_ids, tag keys and tag values, but are not interested in the values associated with the metrics. Since this table is discarding the values and has a very coarse granularity (weekly), it should store much less data than the main table, and make queries faster.

In theory this will eventually be done for all the metric types, but for now this is being used just for counters to test how well this solution actually solves the problems.

**Potential Impact**

The major point of risk is that writing to two materialized views (the existing aggregate table, and now this meta table) will take longer on insert and cause slowdowns/backlogs when inserting data. In that case the reverse migration can be run to remove the view and that will stop.

This will also start storing more data on the node, so that should be watched. In theory this should store much less data than the main table, so that impact should  be small, and will certainly be incremental. If this does become a problem, the migrations can be reversed.


snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
```diff
@@ -0,0 +1,71 @@
...
+                    tags.raw_value AS tag_value
+                GROUP BY
+                    org_id,
+                    project_id,
+                    use_case_id,
+                    metric_id,
+                    tag_key,
+                    tag_value,
+                    timestamp,
+                    retention_days
```
onewland, line 59: could we make this `max(retention_days, 90)` or whatever the correct syntax is? with the upcoming 180-day billing use case I don't want stuff sticking around that long here.

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
```diff
@@ -0,0 +1,71 @@
...
+                    metric_id,
+                    tag_key,
+                    tag_value,
+                    toStartOfWeek(timestamp) as timestamp,
+                    retention_days,
+                    sumState(count_value) as count
+                FROM generic_metric_counters_raw_local
+                ARRAY JOIN
+                    tags.key AS tag_key,
+                    tags.raw_value AS tag_value
```
onewland, line 50: I want to make sure that I understand how this ARRAY JOIN works and that we think it's the right behavior.

If I insert:
```
{org_id=1,project_id=1,use_case_id="custom",metric_id=2,
  tags.key=["foo1","foo2"],tags.value=["bar1","bar2"]}
{org_id=1,project_id=1,use_case_id="custom",metric_id=2,
  tags.key=["foo1","foo2"],tags.value=["bar2","bar3"]}
{org_id=1,project_id=1,use_case_id="custom",metric_id=2,
  tags.key=["foo1","foo2"],tags.value=["bar3","bar4"]}
```

what rows should be output? I assume something like (assuming all prior columns are equal)

```
tags_key,tag_value,count
foo1,bar1,1
foo2,bar2,1
foo1,bar2,1
foo2,bar3,1
foo1,bar3,1
foo2,bar4,1
```

my question would be, do we actually want to store unique counts for key=value pairs, or should we be counting the number of times a unique string shows up and some context e.g.:

```
user_value,type,count
foo1,key,3
foo2,key,3
bar1,value,1
bar2,value,2
bar3,value,2
bar4,value,1
```

This drops information about the relationship between tag keys and values but I could imagine it being a much more efficient encoding scheme for some kinds of input data.

onewland (commented):
added some thoughts, but not blocking

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
evanh: Why not? Wouldn't they also want (in theory) to autocomplete on their metrics and tag keys? This is meant so that meta data stays as long as the actual data stays (roughly).

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
evanh: > I assume something like (assuming all prior columns are equal)

You are right about this.

> This drops information about the relationship between tag keys and values
We need the relationship between keys and values, because one of the two autocomplete queries is "given this tag key, tell me it's values".

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
onewland: As I think about this, I'm not sure we should support autocomplete for all use cases by default. For example, storing all this data for escalating issues is a waste. Tag values are going to be issue group IDs and there's never any reason to autocomplete them.

I don't want to block you from validating this overall approach, but I think we should consider adding a boolean column to the raw table like `record_meta` and only materialize rows where record_meta = true. It's just wasted space for use cases that never have manual user interaction.

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
onewland: sounds good to me

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
nikhars: I agree that we should make recording information in the meta table configurable. Especially since there would be more things coming on the generic metrics pipeline like billing information where auto complete is not a requirement. We can reduce the cost of operating the generic metrics pipeline by restricting the feature to only the use cases which need them.

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
nikhars: > This drops information about the relationship between tag keys and values
We need the relationship between keys and values, because one of the two autocomplete queries is "given this tag key, tell me it's values".

Maybe I might be missing some context. Would it be better to have the relationship between a tag key and its tag values (whose relationship is 1 -> N) be represented like this in the aggregate table?
```
tag_keys int
tag_values Array(str)
```
Storing data this way would make sure that the number of rows in the aggregate table is restricted only by the number of unique metric_id, tag_keys combination. Compared to the current approach of unique metric_id, tag_keys, tag_values combination. 

To answer the question about `give me all tag values for a given metric and tag key` return the full array back to the end user.

snuba/snuba_migrations/generic_metrics/0031_counters_meta_table_mv.py
evanh: How do I make this configurable? I can add an exclusion to the materialized table query (`WHERE use_case_id NOT IN (...)` but is there a way to make that configurable?

evanh:
@nikhars @onewland I updated this to use `groupUniqArray` and also to use a `record_meta` column in the raw table.

nikhars (approved):

evanh:
Did some local testing just to make sure this worked:

First, used tests to insert some data to `generic_metric_counters_raw_local`.

Then, ran these queries:

```
CREATE TABLE generic_metric_counters_meta_aggregated_local
(
    `org_id` UInt64,
    `project_id` UInt64,
    `use_case_id` LowCardinality(String),
    `metric_id` UInt64,
    `tag_key` String,
    `timestamp` DateTime CODEC(DoubleDelta),
    `retention_days` UInt16,
    `tag_values` AggregateFunction(groupUniqArray, String),
    `count` AggregateFunction(sum, Float64)
)
ENGINE = AggregatingMergeTree
PARTITION BY (retention_days, toMonday(timestamp))
PRIMARY KEY (org_id, project_id, use_case_id, metric_id, tag_key, timestamp)
ORDER BY (org_id, project_id, use_case_id, metric_id, tag_key, timestamp)
TTL timestamp + toIntervalDay(retention_days)
SETTINGS index_granularity = 2048
```

Then I ran this query:

```
INSERT INTO generic_metric_counters_meta_aggregated_local (org_id, project_id, use_case_id, metric_id, tag_key, timestamp, retention_days, tag_values, count) SELECT
    org_id,
    project_id,
    use_case_id,
    metric_id,
    tag_key,
    toStartOfWeek(timestamp) AS timestamp,
    retention_days,
    groupUniqArrayState(tag_value) AS tag_values,
    sumState(count_value) AS count
FROM generic_metric_counters_raw_local
ARRAY JOIN
    tags.key AS tag_key,
    tags.raw_value AS tag_value
GROUP BY
    org_id,
    project_id,
    use_case_id,
    metric_id,
    tag_key,
    timestamp,
    retention_days
```

And then I ran these approximations on how the autocomplete could be run in prod:

```
SELECT
    org_id,
    project_id,
    metric_id,
    sumMerge(count)
FROM generic_metric_counters_meta_aggregated_local
WHERE (org_id = 4553686409543680) AND (project_id = 4553686409674753) AND (use_case_id = 'transactions')
GROUP BY
    org_id,
    project_id,
    metric_id

Query id: 6a24c6eb-e48e-425b-a561-30c59a8d71cb

┌───────────org_id─┬───────project_id─┬───────────metric_id─┬─sumMerge(count)─┐
│ 4553686409543680 │ 4553686409674753 │ 9223372036854775933 │             135 │
└──────────────────┴──────────────────┴─────────────────────┴─────────────────┘
```

```
SELECT
    org_id,
    project_id,
    metric_id,
    tag_key,
    groupUniqArrayMerge(tag_values),
    sumMerge(count)
FROM generic_metric_counters_meta_aggregated_local
WHERE (org_id = 4553686409543680) AND (project_id = 4553686409674753) AND (use_case_id = 'transactions') AND (metric_id = 9223372036854775933) AND (tag_key = '10000')
GROUP BY
    org_id,
    project_id,
    metric_id,
    tag_key

Query id: c25358ac-26ee-4a63-a9a5-97d5271590e8

┌───────────org_id─┬───────project_id─┬───────────metric_id─┬─tag_key─┬─groupUniqArrayMerge(tag_values)─┬─sumMerge(count)─┐
│ 4553686409543680 │ 4553686409674753 │ 9223372036854775933 │ 10000   │ ['200','500']                   │              45 │
└──────────────────┴──────────────────┴─────────────────────┴─────────┴─────────────────────────────────┴─────────────────┘
```

evanh:
Some more numbers:

I pulled two days of data to see how many rows this would actually create.

```
use_case_id,count()
escalating_issues,355870
transactions,270810
spans,283241
metric_stats,459
custom,7288
```
It's nice to see that excluding escalating issues would eliminate about half of the total potential rows.

PR #5674, opened 2024-03-20
chore(on-call): Default org_id in admin query tool

enochtangg:
Some queries fail on the admin query tool if the dataset it is querying contains the `ConcurrentRateLimitAllocationPolicy` because it requires `organization_id` to be set in the `tenant_ids`.


evanh (approved):

evanh:
Is there a test that could be added?

PR #5675, opened 2024-03-20
fix(DLQ): Remove DLQ limits in all Snuba consumers

ayirr7:
DLQ limit calculations can be unclear and can cause the consumer to crash. We can remove the limits to avoid this case.


lynnagara:
Can you update this in the rust consuemr too?

lynnagara (approved):

ayirr7:
Tested locally by running `snuba devserver` and sample DLQ-ing some messages.

PR #5676, opened 2024-03-21
feat: optimizer rewrite

kylemumma:
https://getsentry.atlassian.net/browse/SNS-2669
This PR should get all of riccardos broken tests passing. 

What motivated the rewrite?
* I needed to implement new functionality to get riccardos tests passing
* It was gunna be a lot of work to do
* After talking to enoch we thought of an alternative approach, this turned out to a better solution that would be faster to implement

## Major Changes:
* pretty much full optimizer rewrite
* optimizer feature flag on by default
* fixed breaking integration tests

### heres how it works now:
1. use a visitor to get all the conditional aggregates (ex. sumMergeIf) in the ast
2. `or` all the conditions in these functions, and add this to the where clause.

ex:
```
SELECT sumIf(val, metric_id in [1,2,3] and status=200),
       avgIf(val, metric_id=7), 
       maxIf(val, metric_id=5 and status=400)
WHERE org_id=1
...
```
will generate the following
```
(metric_id in [1,2,3] and status=200) or (metric_id=7) or (metric_id=5 and status=400)
```
and add it to the where clause
```
SELECT sumIf(val, metric_id in [1,2,3] and status=200),
       avgIf(val, metric_id=7), 
       maxIf(val, metric_id=5 and status=400)
WHERE org_id=1 and (
        (metric_id in [1,2,3] and status=200) or (metric_id=7) or (metric_id=5 and status=400)
    )
...
```

### DIFF
--- a/snuba/query/conditions.py
+++ b/snuba/query/conditions.py
@@ -155,7 +155,10 @@ def is_not_in_condition_pattern(lhs: Pattern[Expression]) -> FunctionCallPattern
 def binary_condition(
     function_name: str, lhs: Expression, rhs: Expression
 ) -> FunctionCall:
-    return FunctionCall(None, function_name, (lhs, rhs))
+    """This function is deprecated please use snuba.query.dsl.binary_condition"""
+    from snuba.query.dsl import binary_condition as dsl_binary_condition
+
+    return dsl_binary_condition(function_name, lhs, rhs)
 
 
 binary_condition_patterns = {
--- a/snuba/query/mql/parser.py
+++ b/snuba/query/mql/parser.py
@@ -39,7 +39,7 @@ from snuba.query.parser.exceptions import ParsingException
 from snuba.query.processors.logical.filter_in_select_optimizer import (
     FilterInSelectOptimizer,
 )
-from snuba.query.query_settings import QuerySettings
+from snuba.query.query_settings import HTTPQuerySettings, QuerySettings
 from snuba.query.snql.anonymize import format_snql_anonymized
 from snuba.query.snql.parser import (
     MAX_LIMIT,
@@ -1064,19 +1064,12 @@ def quantiles_to_quantile(
     query.transform_expressions(transform)
 
 
-def optimize_filter_in_select(
-    query: CompositeQuery[QueryEntity] | LogicalQuery,
-) -> None:
-    FilterInSelectOptimizer().process_mql_query(query)
-
-
 CustomProcessors = Sequence[
     Callable[[Union[CompositeQuery[QueryEntity], LogicalQuery]], None]
 ]
 
 MQL_POST_PROCESSORS: CustomProcessors = POST_PROCESSORS + [
     quantiles_to_quantile,
-    optimize_filter_in_select,
 ]
 
 
@@ -1115,6 +1108,12 @@ def parse_mql_query(
             MQL_POST_PROCESSORS,
             settings,
         )
+
+    with sentry_sdk.start_span(op="processor", description="filter_in_select_optimize"):
+        if settings is None:
+            FilterInSelectOptimizer().process_query(query, HTTPQuerySettings())
+        else:
+            FilterInSelectOptimizer().process_query(query, settings)
 
     # Custom processing to tweak the AST before validation
     with sentry_sdk.start_span(op="processor", description="custom_processing"):
--- a/snuba/query/mql/parser.py
+++ b/snuba/query/mql/parser.py
@@ -49,7 +49,7 @@ from snuba.query.snql.parser import (
     _replace_time_condition,
     _treeify_or_and_conditions,
 )
-from snuba.state import explain_meta
+from snuba.state import explain_meta, get_int_config
 from snuba.util import parse_datetime
 from snuba.utils.constants import GRANULARITIES_AVAILABLE
 
@@ -1109,11 +1109,16 @@ def parse_mql_query(
             settings,
         )
 
-    with sentry_sdk.start_span(op="processor", description="filter_in_select_optimize"):
-        if settings is None:
-            FilterInSelectOptimizer().process_query(query, HTTPQuerySettings())
-        else:
-            FilterInSelectOptimizer().process_query(query, settings)
+    # Filter in select optimizer
+    feat_flag = get_int_config("enable_filter_in_select_optimizer", default=1)
+    if feat_flag:
+        with sentry_sdk.start_span(
+            op="processor", description="filter_in_select_optimize"
+        ):
+            if settings is None:
+                FilterInSelectOptimizer().process_query(query, HTTPQuerySettings())
+            else:
+                FilterInSelectOptimizer().process_query(query, settings)
 
     # Custom processing to tweak the AST before validation
     with sentry_sdk.start_span(op="processor", description="custom_processing"):
--- a/snuba/query/mql/parser.py
+++ b/snuba/query/mql/parser.py
@@ -1067,7 +1067,7 @@ def quantiles_to_quantile(
 def optimize_filter_in_select(
     query: CompositeQuery[QueryEntity] | LogicalQuery,
 ) -> None:
-    FilterInSelectOptimizer().process_mql_query2(query)
+    FilterInSelectOptimizer().process_mql_query(query)
 
 
 CustomProcessors = Sequence[
--- a/snuba/query/mql/parser.py
+++ b/snuba/query/mql/parser.py
@@ -1067,7 +1067,7 @@ def quantiles_to_quantile(
 def optimize_filter_in_select(
     query: CompositeQuery[QueryEntity] | LogicalQuery,
 ) -> None:
-    FilterInSelectOptimizer().process_mql_query(query)
+    FilterInSelectOptimizer().process_mql_query2(query)
 
 
 CustomProcessors = Sequence[
--- a/snuba/query/processors/logical/filter_in_select_optimizer.py
+++ b/snuba/query/processors/logical/filter_in_select_optimizer.py
@@ -110,7 +110,7 @@ class FilterInSelectOptimizer:
                 new_condition = self.get_select_filter(query)
                 if new_condition is not None:
                     query.add_condition_to_ast(new_condition)
-                    metrics.increment("kyles_optimizer_optimized")
+                    metrics.increment("filter_in_select_optimizer_optimized")
             except Exception:
                 logger.warning(
                     "Failed during optimization", exc_info=True, extra={"query": query}
@@ -134,16 +134,10 @@ class FilterInSelectOptimizer:
         # find and grab all the conditional aggregate functions
         cond_agg_functions: list[FunctionCall | CurriedFunctionCall] = []
         for selected_exp in query.get_selected_columns():
-            exp = selected_exp.expression
-            found = exp.accept(FindConditionalAggregateFunctionsVisitor())
-            if len(found) > 0:
-                if len(cond_agg_functions) > 0:
-                    raise ValueError(
-                        "expected only one selected column to contain conditional aggregate functions but found multiple"
-                    )
-                else:
-                    cond_agg_functions = found
-
+            found = selected_exp.expression.accept(
+                FindConditionalAggregateFunctionsVisitor()
+            )
+            cond_agg_functions += found
         if len(cond_agg_functions) == 0:
             return None
 
@@ -152,11 +146,11 @@ class FilterInSelectOptimizer:
         for func in cond_agg_functions:
             if len(func.parameters) != 2:
                 raise ValueError(
-                    f"expected conditional functions to be of the form funcIf(val, condition), but was given a function with {len(func.parameters)} parameters"
+                    f"expected conditional function to be of the form funcIf(val, condition), but was given one with {len(func.parameters)} parameters"
                 )
             if not isinstance(func.parameters[1], FunctionCall):
                 raise ValueError(
-                    f"expected conditional functions to be of the form funcIf(val, condition), but was given a function whos condition is {type(func.parameters[1])} rather than FunctionCall"
+                    f"expected conditional function to be of the form funcIf(val, condition), but the condition is type {type(func.parameters[1])} rather than FunctionCall"
                 )
 
             if new_condition is None:
--- a/snuba/query/processors/logical/filter_in_select_optimizer.py
+++ b/snuba/query/processors/logical/filter_in_select_optimizer.py
@@ -157,6 +157,6 @@ class FilterInSelectOptimizer:
                 new_condition = deepcopy(func.parameters[1])
             else:
                 new_condition = binary_condition(
-                    "or", deepcopy(func.parameters[1]), new_condition
+                    "or", new_condition, deepcopy(func.parameters[1])
                 )
         return new_condition
--- a/snuba/query/processors/logical/filter_in_select_optimizer.py
+++ b/snuba/query/processors/logical/filter_in_select_optimizer.py
@@ -9,7 +9,6 @@ from snuba.query.expressions import (
     Argument,
     Column,
     CurriedFunctionCall,
-    Expression,
     ExpressionVisitor,
     FunctionCall,
     Lambda,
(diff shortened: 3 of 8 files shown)


tests/test_metrics_sdk_api.py
```diff
@@ -359,9 +358,7 @@ def test_raw_mql_string(self, test_dataset: str, tag_column: str) -> None:
 
         assert response.status_code == 200, data
         rows = data["data"]
-        assert len(rows) >= 180, rows
-
-        assert math.isnan(rows[0]["aggregate_value"])  # division by zero
+        assert len(rows) == 0
```
kylemumma, line 361: this test used to give a bunch of nan rows. With the optimizer it returns no rows.

My thought was the division with conditional aggregates that have no matching columns must cause NAN. With the optimizer all rows would be filtered out making this expected behavior.

kylemumma: i didnt investigate this too much but I thought since the results was all NaN before, having an empty result seems fine

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,141 @@
...
 
     def process_mql_query(
         self, query: LogicalQuery | CompositeQuery[QueryEntity]
     ) -> None:
-        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=0)
+        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=1)
         if feat_flag:
             try:
-                domain = self.get_domain_of_mql_query(query)
+                new_condition = self.get_select_filter(query)
```
kylemumma, line 89: I made `query -> condition` its own abstracted thing to use for testing

tests/query/processors/test_filter_in_select_optimizer.py
```diff
@@ -50,107 +59,241 @@ def subscriptable_reference(name: str, key: str) -> SubscriptableReference:
...
+def _or(lhs: FunctionCall, rhs: FunctionCall) -> FunctionCall:
+    return binary_condition("or", lhs, rhs)
+
+
+def _in(lhs: Expression, rhs: Expression) -> FunctionCall:
+    return binary_condition("in", lhs, rhs)
+
+
+expected_optimize_condition = {
+    "sum(`d:transactions/duration@millisecond`){status_code:200} / sum(`d:transactions/duration@millisecond`)": _or(
```
kylemumma, line 79: my unit tests verify: given an mql query, does the optimizer generate the expected condition

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
         if feat_flag:
             try:
-                domain = self.get_domain_of_mql_query(query)
-            except ValueError:
+                new_condition = self.get_select_filter(query)
+                if new_condition is not None:
+                    query.add_condition_to_ast(new_condition)
+                    metrics.increment("kyles_optimizer_optimized")
+            except Exception:
                 logger.warning(
```
kylemumma, line 94: should I keep logging like this, for context it will block deploys

tests/query/processors/test_filter_in_select_optimizer.py
```diff
@@ -50,107 +59,241 @@ def subscriptable_reference(name: str, key: str) -> SubscriptableReference:
...
-            Column("_snuba_metric_id", None, "metric_id"): {
-                Literal(None, 123456),
-            }
-        },
-    ),
-    (
-        "sum(`d:transactions/duration@millisecond`){status_code:200} / sum(`d:transactions/duration@second`)",
-        {
-            Column("_snuba_metric_id", None, "metric_id"): {
+def equals(lhs: Expression, rhs: Expression) -> FunctionCall:
```
evanh, line 62: There's a file called `dsl.py` that has a bunch of helper functions like this, I would add these there so other people can use them in the future.

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+    def visit_lambda(self, exp: Lambda) -> None:
+        return
+
+
 class FilterInSelectOptimizer:
     """
-    This optimizer takes queries that filter by metric_id in the select clause (via conditional aggregate functions),
-    and adds the equivalent metric_id filtering to the where clause. Example:
+    This optimizer lifts conditions in the select clause into the where clause,
+    this is
```
volokluev, line 70: useless line

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+            self._matches.append(exp)
+
+    def visit_argument(self, exp: Argument) -> None:
+        return
+
+    def visit_lambda(self, exp: Lambda) -> None:
+        return
+
+
 class FilterInSelectOptimizer:
```
volokluev, line 67: This class is a query processor it should extend the query processor class

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
-        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=0)
+        feat_flag = get_int_config("enable_filter_in_select_optimizer", default=1)
         if feat_flag:
             try:
-                domain = self.get_domain_of_mql_query(query)
-            except ValueError:
+                new_condition = self.get_select_filter(query)
+                if new_condition is not None:
+                    query.add_condition_to_ast(new_condition)
+                    metrics.increment("kyles_optimizer_optimized")
```
volokluev, line 92: unhelpful metric name

tests/query/processors/test_filter_in_select_optimizer.py
```diff
@@ -50,107 +59,241 @@ def subscriptable_reference(name: str, key: str) -> SubscriptableReference:
...
+def test_condition_generation(mql_query: str, expected_condition: FunctionCall) -> None:
     logical_query, _ = parse_mql_query(str(mql_query), mql_context, generic_metrics)
     assert isinstance(logical_query, Query)
-    res = optimizer.get_domain_of_mql_query(logical_query)
-    if res != expected_domain:
-        raise
-    assert res == expected_domain
+
+    opt = FilterInSelectOptimizer()
+    actual = opt.get_select_filter(logical_query)
```
volokluev, line 297: this test does not actually test that the query processor is applying the changes to the condition (which is its reason for being). The test should assert that behavior

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+        for selected_exp in query.get_selected_columns():
+            exp = selected_exp.expression
+            finder = FindConditionalAggregateFunctionsVisitor()
+            exp.accept(finder)
+            found = finder.get_matches()
+            if len(found) > 0:
+                if len(cond_agg_functions) > 0:
                     raise ValueError(
-                        "Was expecting only 1 select expression to contain condition aggregate but found multiple"
+                        "expected only one selected column to contain conditional aggregate functions but found multiple"
```
volokluev, line 123: This edge case is not clear to me why it exists, the error message is not helpful for the user. It's also not a tested case. Also if this is actually an invalid query, you should raise `snuba.query.exceptions.InvalidQueryException`

volokluev: Reading this more, I don't think this is invalid. If there are multiple `If` conditions, their metric ids should just be merged together

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
-        domains: list[Domain] = []
-        self._get_conditional_domains_helper(exp, domains)
-        return domains
+        # validate the functions, and lift their conditions into new_condition, return it
+        new_condition = None
+        for func in cond_agg_functions:
+            if len(func.parameters) != 2 or not isinstance(
+                func.parameters[1], FunctionCall
+            ):
+                raise ValueError("unexpected form of function")
```
volokluev, line 137: what were you expecting and what did you find?

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+    Search an expression for all conditional aggregate functions.
+
+    Might be able to use matchers.Pattern to extend this idea for general search but I was too lazy.
+    """
+
+    def __init__(self) -> None:
+        self._matches: list[FunctionCall | CurriedFunctionCall] = []
+
+    def get_matches(self) -> list[FunctionCall | CurriedFunctionCall]:
+        return deepcopy(self._matches)
```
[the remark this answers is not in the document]
kylemumma: My copying was done so that when I put it in WHERE, changing SELECT wouldnt modify the one in WHERE.

Volo helped me see that the copying shouldnt be done in the visitor, but later on when i copy the one from SELECT into WHERE.

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+        return
+
+    def visit_function_call(self, exp: FunctionCall) -> None:
+        if exp.function_name[-2:] == "If":
+            self._matches.append(exp)
+        else:
+            for param in exp.parameters:
+                param.accept(self)
+
+    def visit_curried_function_call(self, exp: CurriedFunctionCall) -> None:
```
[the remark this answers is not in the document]
kylemumma: I was assuming that there wont be any inside of a curried function. I will make this change to also look here.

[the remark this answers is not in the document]
kylemumma: In my testing I was getting curried function calls of this form
`quantilesIf(0.5)(val, cond)`
where
```
internal_function = quantilesIf(0.5)
parameters = (val, cond)
```

are you saying to search the parameters of the internal_function?

snuba/query/processors/logical/filter_in_select_optimizer.py
kylemumma: The main issue I ran into trying to do this is:
LogicalQueryProcessor defines input as a logical query, but the query I will be given by the parser is logical or composite

Smaller (i think) issue I ran into is:
My processor is being called with the _post_process function which wont pass in query settings as required by LogicalQueryProcessor

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+        if len(cond_agg_functions) == 0:
+            return None
 
-    def _get_conditional_domains(self, exp: Expression) -> list[Domain]:
-        domains: list[Domain] = []
-        self._get_conditional_domains_helper(exp, domains)
-        return domains
+        # validate the functions, and lift their conditions into new_condition, return it
+        new_condition = None
+        for func in cond_agg_functions:
```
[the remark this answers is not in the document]
kylemumma: No, there can be multiple

`divide(sumIf(...), avgIf(...))` would cause it to be `[sumIf(...), avgIf(...)]`

I improved the comment for the visitor, hopefully this makes it a bit more clear.

snuba/query/processors/logical/filter_in_select_optimizer.py
kylemumma: removed

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
+            if len(func.parameters) != 2 or not isinstance(
```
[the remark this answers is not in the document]
kylemumma: this check is for my assumption that conditional aggregate functions are of the form `funcIf(val, condition)`

I added clearer error descriptions for this

snuba/query/processors/logical/filter_in_select_optimizer.py
```diff
@@ -1,234 +1,145 @@
...
-        if isinstance(exp, FunctionCall):
-            # add domain of function call
-            if exp.function_name[-2:] == "If":
-                if len(exp.parameters) != 2 or not isinstance(
-                    exp.parameters[1], FunctionCall
-                ):
-                    raise ValueError("unexpected form of function aggregate")
-                domains.append(self._get_domain_of_predicate(exp.parameters[1]))
+            if new_condition is None:
+                new_condition = deepcopy(func.parameters[1])
```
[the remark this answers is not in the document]
kylemumma: This copy is so when the SELECT condition is copied into WHERE. If some piece of code changes one it wont unknowingly modify both.

I removed the copy in the visitor but kept this one.

snuba/query/processors/logical/filter_in_select_optimizer.py
kylemumma: We mentioned this optimizer being in snuba/query/processors/logical is a bit weird. Maybe I could move it to snuba/query/mql, any other ideas?

volokluev: how does this processor handle composite queries? Did you test for that?

snuba/query/processors/logical/filter_in_select_optimizer.py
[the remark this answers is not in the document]
evanh: No but it is possible to have something like `topK(10)(sumIf(x), avgIf(y))`. There is nothing in production like that at the moment, but would be supported in the future.

snuba/query/processors/logical/filter_in_select_optimizer.py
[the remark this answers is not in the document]
evanh: I don't think this is necessary, all the objects involved here are frozen dataclasses, so someone would have to work hard to modify them in place.

snuba/query/processors/logical/filter_in_select_optimizer.py
evanh: For this processor we can assume that it will never be used for Composite queries. It's meant to solve a very specific problem for MQL, and CompositeQuery support is out of scope.

So what you could do is simply do nothing if the query is a CompositeQuery. Then the typing will match.

tests/query/processors/test_filter_in_select_optimizer.py
kylemumma: tests changed to test: input_query -> expected_output_query

volokluev (approved):

snuba/query/processors/logical/filter_in_select_optimizer.py
kylemumma: made it logical processor

snuba/query/conditions.py
```diff
@@ -155,7 +155,10 @@ def is_not_in_condition_pattern(lhs: Pattern[Expression]) -> FunctionCallPattern
 def binary_condition(
     function_name: str, lhs: Expression, rhs: Expression
 ) -> FunctionCall:
-    return FunctionCall(None, function_name, (lhs, rhs))
+    """This function is deprecated please use snuba.query.dsl.binary_condition"""
+    from snuba.query.dsl import binary_condition as dsl_binary_condition
+
+    return dsl_binary_condition(function_name, lhs, rhs)
```
kylemumma, line 161: couldnt use this function in dsl.py bc of circular import with conditions.py. I thought this belonged in dsl.py anyways so I moved it.

snuba/query/mql/parser.py
```diff
@@ -1116,6 +1109,17 @@ def parse_mql_query(
...
         )
 
+    # Filter in select optimizer
+    feat_flag = get_int_config("enable_filter_in_select_optimizer", default=1)
+    if feat_flag:
+        with sentry_sdk.start_span(
+            op="processor", description="filter_in_select_optimize"
+        ):
+            if settings is None:
+                FilterInSelectOptimizer().process_query(query, HTTPQuerySettings())
```
kylemumma, line 1119: I would like to point out that in this context settings can be None, LogicalQueryProcessor interface requires settings not None, thus I had to create dummy object.

Since settings arent used by my processor it doesnt matter, but the fact that i have to create this dummy instead of None smells to me

tests/query/parser/test_formula_mql_query.py
kylemumma, line 1: I decided to just get these tests working rather than do a refactor that loosens coupling between parsing and post_processing/optimization. Maybe its something I could think about in the future.

Evan made an interesting point about end-to-end testing as a single source of truth, and the increased confidence it brings in the pipeline,
the tradeoff seems to be that this testing file has to change any time another stage is added. Im not sure what the right answer is.

snuba/query/processors/logical/filter_in_select_optimizer.py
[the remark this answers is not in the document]
kylemumma: for now i have left the feature out, will relook once this change is added

PR #5680, opened 2024-03-22
feat(meta): Add record_meta field to counters consumer

evanh:
This field will be used by the materialized view to determine whether the row
should be processed into the meta table.

Depends on https://github.com/getsentry/snuba/pull/5681

### DIFF
--- a/rust_snuba/src/processors/generic_metrics.rs
+++ b/rust_snuba/src/processors/generic_metrics.rs
@@ -213,7 +213,7 @@ impl Parse for CountersRawRow {
         let retention_days = enforce_retention(Some(from.retention_days), &config.env_config);
 
         let mut record_meta: Option<u8> = Some(1);
-        let blacklisted_use_case_ids = vec!["escalating_issues", "metric_stats"];
+        let blacklisted_use_case_ids = ["escalating_issues", "metric_stats"];
         if blacklisted_use_case_ids
             .iter()
             .any(|&u| u == from.use_case_id)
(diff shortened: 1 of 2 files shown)


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -210,6 +212,15 @@ impl Parse for CountersRawRow {
...
         let retention_days = enforce_retention(Some(from.retention_days), &config.env_config);
 
+        let mut record_meta: Option<u8> = Some(1);
+        let blacklisted_use_case_ids = ["escalating_issues", "metric_stats"];
+        if blacklisted_use_case_ids
+            .iter()
+            .any(|&u| u == from.use_case_id)
+        {
+            record_meta = Some(0);
+        }
```
nikhars, line 222: A match statement might be more appropriate here. Something along the lines of
```rust
let record_meta = match from.use_case_id.as_str() {
            "escalating_issues" => Some(0),
            "metric_stats" => Some(0),
            _ => Some(1),
        };
```

nikhars (commented):
Since this is an optimization, would it be better to do it via an allowlist rather than a blacklist? That way we avoid new use cases being recorded in the meta table by default. Something like escalating issues causing unnecessary overhead on the meta table.

evanh:
> Since this is an optimization, would it be better to do it via an allowlist rather than a blacklist?

I did it this way so that we can avoid a potential step in someone creating a new use case. This way, if a product person creates a new use case, they don't then have to take another step to get auto-complete, it just comes for free.

Swatinem (approved):

PR #5681, opened 2024-03-22
feat(metrics): Add a meta table to counters

evanh:
The original PR (https://github.com/getsentry/snuba/pull/5669) was merged with a bug (column wasn't added to the dist table. I reverted that PR before it could be deployed broken, and created this replacement.

This creates everything needed to test  storing meta information about
metrics. The new table is meant to satisfy queries that are trying to find
metric_ids, tag keys and tag values, but are not interested in the values
associated with the metrics.

In theory this will eventually be done for all the metric types, but for now
this is being used just for counters to test how well this solution actually
solves the problems.

This does three things:

- Add a column to the raw table that can be used to control which data
goes to the meta table
- Create a new table for aggregated meta information
- Create a materialized view to populate that table from the raw data


nikhars (approved):
Thanks for fixing

PR #5683, opened 2024-03-25
ref: Initial refactor to facilitate streaming writes to clickhouse

untitaker:
We want to eventually stop buffering writes to clickhouse, and instead hold one long-running connection for the lifetime of a batch, streaming rows slowly in. This is already how Python works.

This initial refactor creates a new Batch type that holds onto the clickhouse client internally.

### DIFF
--- a/rust_snuba/src/strategies/clickhouse/mod.rs
+++ b/rust_snuba/src/strategies/clickhouse/mod.rs
@@ -138,33 +138,3 @@ impl ProcessingStrategy<BytesInsertBatch> for ClickhouseWriterStep {
         self.inner.join(timeout)
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    #[tokio::test]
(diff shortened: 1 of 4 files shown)


Swatinem (approved):

lynnagara (approved):

rust_snuba/src/strategies/clickhouse/mod.rs
```diff
@@ -141,90 +137,3 @@ impl ProcessingStrategy<BytesInsertBatch> for ClickhouseWriterStep {
         self.inner.join(timeout)
     }
 }
-
-#[derive(Clone)]
```
lynnagara, line 145: Do we need a test that actually writes a batch to clickhouse?

untitaker: right now those tests get in the way IMO

PR #5685, opened 2024-03-25

Zylphrex:
```
```
This results in a sql like
```
```
```
```

### DIFF
--- a/snuba/query/processors/physical/mapping_optimizer.py
+++ b/snuba/query/processors/physical/mapping_optimizer.py
@@ -1,6 +1,6 @@
 import logging
 from enum import Enum
-from typing import Optional, Tuple
+from typing import Optional, Tuple, cast
 
 from snuba import environment
 from snuba.clickhouse.query import Query
@@ -13,6 +13,7 @@ from snuba.clickhouse.translators.snuba.mappers import (
 )
 from snuba.query.conditions import (
     BooleanFunctions,
+    ConditionFunctions,
     combine_and_conditions,
     combine_or_conditions,
     get_first_level_and_conditions,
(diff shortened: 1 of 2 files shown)


evanh (approved):

snuba/query/processors/physical/mapping_optimizer.py
```diff
@@ -265,7 +351,7 @@ def _get_condition_without_redundant_checks(
                     if tag_exist_match:
                         matched_tag_exists_conditions[condition_id] = tag_exist_match
                 if not tag_exist_match:
-                    eq_match = self.__optimizable_pattern.match(cond)
+                    eq_match = self.__equals_condition_pattern.match(cond)
```
Zylphrex, line 354: @volokluev would you happen to know if I need to implement this removing of redundant checks for `IN` conditions?

volokluev: you could but it's not strictly necessary. I'm not sure how often we get those cases with IN conditions. Definitely something that can be added later

Zylphrex: I think this is less common on the older datasets but more likely to happen with the spans dataset as we have `sentry_tags` which contains some more commonly used columns.

The example I ran into was with `environment`. For a 24h period, it read >48GiB of data, and after applying this optimization, I saw it was reduced to <24GiB of data. On 7 day periods, the query was already timing out. So this optimization should already be helpful

volokluev: oh yes absolutely. Merge the PR. I was saying that the redundant clause optimization is probably not going to be as applicable for IN clauses

PR #5691, opened 2024-03-26

Zylphrex:
Re-apply #5685

```
```
This results in a sql like
```
```
```
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -366,6 +366,7 @@ jobs:
               tests/sentry/search/events \
               tests/sentry/event_manager \
               tests/sentry/api/endpoints/test_organization_profiling_functions.py \
+              tests/snuba/api/endpoints/test_organization_events_stats_mep.py \
               tests/snuba/test_snql_snuba.py \
               tests/snuba/test_metrics_layer.py \
               -vv --cov . --cov-report="xml:.artifacts/snuba.coverage.xml"
--- a/snuba/query/processors/physical/mapping_optimizer.py
+++ b/snuba/query/processors/physical/mapping_optimizer.py
@@ -1,6 +1,6 @@
 import logging
 from enum import Enum
-from typing import Optional, Tuple
+from typing import Optional, Tuple, cast
 
 from snuba import environment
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -366,6 +366,7 @@ jobs:
               tests/sentry/search/events \
               tests/sentry/event_manager \
               tests/sentry/api/endpoints/test_organization_profiling_functions.py \
+              tests/snuba/api/endpoints/test_organization_events_stats_mep.py \
```
Zylphrex, line 369: A test in here was failing in the last attempt

snuba/query/processors/physical/mapping_optimizer.py
```diff
@@ -155,20 +168,43 @@ def __classify_combined_conditions(self, condition: Expression) -> ConditionClas
...
+            rhs = in_condition_match.expression("right_hand_side")
+            assert isinstance(rhs, FunctionExpr)
+
+            params = rhs.parameters
+            for param in params:
+                assert isinstance(param, LiteralExpr)
+
+                # tags[foo] IN array('') is not optimizable
+                if param.value == "":
+                    return ConditionClass.NOT_OPTIMIZABLE
```
Zylphrex, line 204: This was the cause of the test failure. Due to the way we query for tags, even if `foo` is not a tag on the row, it has a value of `''` and won't exist in the tags hash map.

To handle this, we just mark these cases are not optimizable. and run the query as is.

tests/query/processors/test_mapping_optimizer.py
```diff
@@ -278,6 +278,108 @@
...
+            condition=binary_condition(
+                ConditionFunctions.IN,
+                nested_expression("tags", "my_tag"),
+                FunctionCall(
+                    None,
+                    "array",
+                    (
+                        Literal(None, "a"),
+                        Literal(None, "b"),
+                        Literal(None, ""),
```
Zylphrex, line 363: new test with an empty value in the list to capture this failure case

volokluev (approved):

PR #5692, opened 2024-03-26

Zylphrex:
Re-apply #5685 and #5691

```
```
This results in a sql like
```
```
```
```


Zylphrex:
We reverted this due to master failing but we've identified that cause of the failures and have merged a fix to sentry in getsentry/sentry#67751 so this change should be good to go now.

PR #5694, opened 2024-03-27
ref: Move to streaming writes in Rust consumer

untitaker:
Refactor the rust consumer so that from within Reduce, the HTTP request is
opened and rows are directly written into the socket.

This requires splitting up the BytesInsertBatch type as it can be in multiple
"states" depending on the stage of the pipeline. It contains a HttpBatch before
writing to clickhouse, and contains none before committing offsets.

The easiest way that I found is to first make it generic over its internal row
data R, then slowly fixing up all the instances where the generic was missing.

In a future PR I would like to see:

* deleting `InsertBatch` entirely, and replacing all instances with `BytesInsertBatch`
* renaming `BytesInsertBatch` to `InsertBatch`
* splitting up `types.rs` into multiple files

### DIFF
--- a/rust_snuba/Cargo.lock
+++ b/rust_snuba/Cargo.lock
@@ -2094,8 +2094,8 @@ dependencies = [
 
 [[package]]
 name = "rust_arroyo"
-version = "2.16.4"
-source = "git+https://github.com/getsentry/arroyo#17dc8fe1af1b320ac483e0ed6110058fb92a98eb"
+version = "2.17.0"
+source = "git+https://github.com/getsentry/arroyo#4a2015c5702855c9484d250deccd50f50059cdd9"
 dependencies = [
  "chrono",
  "coarsetime",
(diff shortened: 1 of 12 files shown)


Swatinem (approved):
lgtm.

Though this very much reminds me of one of my earliest comments about the whole arroyo architecture:
- Defining things in reverse direction makes things very hard to read and understand
- The whole architecture would possibly be very well expressed as a `Stream` using combinators that might just spawn futures onto the runtime, and a couple of bounded channels in between to act as concurrency limiters, etc.

untitaker:
> The whole architecture would possibly be very well expressed as a Stream using combinators that might just spawn futures onto the runtime, and a couple of bounded channels in between to act as concurrency limiters, etc.

I have a prototype of an entire batching clickhouse consumer + streaming without arroyo, and I agree that arroyo's abstractions tend to get in the way here. However, I found rdkafka's streaming consumer to be basically useless as well. It's easier to un-design by removing some of the intermediate strategies and merging them into one.

williamdes:
I am wondering if this might have create an even worse issue on https://github.com/getsentry/self-hosted/issues/2876#issuecomment-2082298438
The logs are filled with `skipping write of` but ClickHouse stats show 14k/s inserts.

PR #5697, opened 2024-03-27
cleanup(devexp): remove old execution pipeline code

volokluev:
The new execution pipeline has been running in production without issue. We can now remove the old code

### DIFF
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -62,7 +62,7 @@ def parse_and_run_query(
     concurrent_queries_gauge: Optional[Gauge] = None,
 ) -> QueryResult:
     """
-    Runs a Snuba Query, then records the metadata about each split query that was run.
+    Processes, runs a Snuba Query, then records the metadata about the query that was run.
     """
     query_metadata = SnubaQueryMetadata(request, get_dataset_name(dataset), timer)
 
@@ -72,6 +72,7 @@ def parse_and_run_query(
         )
         if not request.query_settings.get_dry_run():
             record_query(request, timer, query_metadata, result)
+        _set_query_final(request, result.extra)
     except QueryException as error:
         _set_query_final(request, error.extra)
         record_query(request, timer, query_metadata, error)
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -23,8 +23,6 @@ logger = logging.getLogger("snuba.query")
 
 metrics = MetricsWrapper(environment.metrics, "api")
 
-MAX_QUERY_SIZE_BYTES = 256 * 1024  # 256 KiB by default
-
 
 def _run_query_pipeline(
     request: Request,
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -1,53 +1,23 @@
 from __future__ import annotations
 
-import copy
 import logging
-import random
-import textwrap
-from dataclasses import replace
-from functools import partial
-from math import floor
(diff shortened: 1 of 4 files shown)


snuba/web/query.py
```diff
@@ -105,61 +64,12 @@ def parse_and_run_query(
...
-                    old_sql = result.extra["sql"]
-                    if new_sql != old_sql:
-                        logger.warning(
-                            "New and old query pipeline sql doesn't match: Old: %s, New: %s",
-                            old_sql,
-                            new_sql,
-                        )
-                except Exception as e:
-                    logger.exception(e)
-        _set_query_final(request, result.extra)
```
kylemumma, line 162: why u delete this when its outside the if/else?

tests/web/test_query_size.py
```diff
@@ -1,16 +1,19 @@
 import pytest
 
-from snuba.web.query import MAX_QUERY_SIZE_BYTES, get_query_size_group
+from snuba import settings
+from snuba.pipeline.stages.query_execution import get_query_size_group
 
-TENTH_PLUS_ONE = int(MAX_QUERY_SIZE_BYTES / 10) + 1
+TENTH_PLUS_ONE = int(settings.MAX_QUERY_SIZE_BYTES / 10) + 1
```
kylemumma, line 6: the same `MAX_QUERY_SIZE_BYTES` is also in settings? I think you did this because you thought settings is a better place for it and I would agree. Why didnt you delete the one from query.py?

tests/test_api.py
```diff
@@ -2141,45 +2141,6 @@ def test_record_queries(self, record_query_mock: MagicMock) -> None:
         assert metadata["dataset"] == "events"
         assert metadata["request"]["referrer"] == "test"
 
-    @patch("snuba.web.query._run_query_pipeline")
-    @patch("snuba.web.query._run_new_query_pipeline")
-    def test_error_handler(
```
kylemumma, line 2146: what was the point of this test?

snuba/web/query.py
```diff
@@ -69,17 +39,6 @@ def _run_new_query_pipeline(
             data=request, query_settings=request.query_settings, timer=timer, error=None
         )
     )
-    if force_dry_run:
```
kylemumma, line 72: how come this goes away?

kylemumma:
assuming the tests pass it LGTM other than the `_set_query_final`

snuba/web/query.py
volokluev: that was only necessary for the dry run rollout. You QuerySettings set dry run here

tests/test_api.py
volokluev: it already exists in `test_snql_api` it's a duplicate

tests/web/test_query_size.py
volokluev: good point

snuba/web/query.py
kylemumma: ah I see, we still have dry run right, itll just get set somewhere else now that its rolled out?

volokluev: correct, it's a field in the query settings. We were just forcing it for the rollout

kylemumma (approved):
GREAT WORK VOLO, BEST CODE IVE READ ALL WEEK

enochtangg (approved):

PR #5698, opened 2024-03-27
ref: separate parse query from build request

kylemumma:
## major changes
`build_request` is now broken into 2 functions that are called sequentially: `parse_api_request` and `build_request`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -65,5 +65,10 @@
   // test discovery is sluggish and the UI around running
   // tests is often in your way and misclicked
   "python.testing.pytestEnabled": true,
-  "python.testing.pytestArgs": ["tests"]
+  "python.testing.pytestArgs": [
+    "tests"
+  ],
+  "mypy-type-checker.args": [
+    "--strict"
+  ]
 }
--- a/snuba/request/validation.py
+++ b/snuba/request/validation.py
@@ -105,47 +105,56 @@ def update_attribution_info(
     return attribution_info
 
 
-def build_request(
+def parse_request(
     body: Dict[str, Any],
-    parser: Parser,
     settings_class: Union[Type[HTTPQuerySettings], Type[SubscriptionQuerySettings]],
     schema: RequestSchema,
     dataset: Dataset,
     timer: Timer,
     referrer: str,
+    is_mql: bool,
     custom_processing: Optional[CustomProcessors] = None,
-) -> Request:
-    with sentry_sdk.start_span(description="build_request", op="validate") as span:
+) -> tuple[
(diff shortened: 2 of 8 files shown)


.vscode/settings.json
```diff
@@ -65,5 +65,10 @@
   // test discovery is sluggish and the UI around running
   // tests is often in your way and misclicked
   "python.testing.pytestEnabled": true,
-  "python.testing.pytestArgs": ["tests"]
+  "python.testing.pytestArgs": [
+    "tests"
+  ],
+  "mypy-type-checker.args": [
+    "--strict"
```
kylemumma, line 72: bc this is what it is in CI

snuba/subscriptions/data.py
kylemumma, line 1: here down is updating all existing calls to use the new interface

kylemumma, line 1: here down is updating all existing calls to `build_request` to use the interface

snuba/request/validation.py
```diff
@@ -105,25 +94,38 @@ def update_attribution_info(
...
-) -> Request:
-    with sentry_sdk.start_span(description="build_request", op="validate") as span:
+) -> tuple[
+    RequestParts,
+    HTTPQuerySettings | SubscriptionQuerySettings,
+    Query | CompositeQuery[Entity],
+    str,
+]:
+    """
+    Given a snuba-api request-body (among other things), parse it into a query ast and return the ast (among other things).
```
volokluev, line 113:
```
    Given a snuba-api request-body (among other things), parse it into a query ast and return the ast (among other things).
```

(among other things) is not a helpful comment and if you have to put it maybe it indicates that this interface does too many things

volokluev (commented):
2 questions for you:

1. What value did this add to the codebase?
2. Can we use the `QueryPipelineStage` interface for this part of the `QueryPipeline`?

evanh (commented):
Can you provide more context on why this split was done?

It also seems confusing because `build_request` still takes the raw query body as input. Why is this necessary if we call `parse_api_request` first?

snuba/request/validation.py
enochtangg: We should think about separating parsing and build request in the context of querying storages directly. In the future, when an incoming "storage query" reaches snuba, would we parse the query the same way we do with an "entity query"? If not, we should be able to swap out that specific stage of the pipeline that handles parsing with something else that works for storage queries. Same goes for `build_request` (maybe this doesn't need to change). It would be nice to use the `QueryPipelineStage` interface to enable this work.

kylemumma:
wontfix

PR #5700, opened 2024-03-28
chore(gocd): Move de region to be between s4s and us

IanWoodard:
This change moves `de` from a test region that is run separate from the rest of our deploys, to a production region that is run, in this case, after `s4s` and before `us`. Now that `de` officially has customers, it is a good time to move it.

### DIFF
--- a/gocd/templates/jsonnetfile.json
+++ b/gocd/templates/jsonnetfile.json
@@ -8,7 +8,7 @@
           "subdir": "libs"
         }
       },
-      "version": "v2.8"
+      "version": "v2.9.1"
     }
   ],
   "legacyImports": true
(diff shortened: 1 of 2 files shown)


nikhars (commented):
I am a bit of a noob with gocd. I only see a version change happening here. Is the actual change which does the reordering somewhere else? If yes, could you link to the original change where this happened?

IanWoodard:
> I am a bit of a noob with gocd. I only see a version change happening here. Is the actual change which does the reordering somewhere else? If yes, could you link to the original change where this happened?

We generate the "pipedream" pipelines using jsonnet so we can easily create/update multi-regional deployment pipelines. This generating is done by GoCD using the jsonnet defined in this repo. The library we use to define the core pipedream jsonnet can be found here: https://github.com/getsentry/gocd-jsonnet. The updates in the change we are pinning can be found here: https://github.com/getsentry/gocd-jsonnet/commit/4a103d77466c9c2d24df3fdfc8fc876bdceea135.

nikhars (approved):

PR #5702, opened 2024-03-28
feat: filter in select optimizer rewrite

kylemumma:
see https://github.com/getsentry/snuba/pull/5676


volokluev:
Why was this reverted?

kylemumma:
> Why was this reverted?

the deploy pipeline failed and i saw de connection fails go up so I reverted it, but it turns out neither were from my change.

volokluev (approved):

PR #5708, opened 2024-03-29
feat(replays): Materialize replays_local table

cmanallen:
Materializes the `replays_local` table.

We're materializing the table because the number of rows scanned is too large.  Queries are taking 10's of seconds re-aggregating the same replays.  Sentry has a 25-1 ratio of rows to replay_id.  Some of our larger customers have closer to 100-1 ratio.  The ratio is not related to the total number of replays ingested but large customers with large ratios are particularly effected.


snuba/snuba_migrations/replays/0019_add_materialization.py
```diff
@@ -0,0 +1,192 @@
...
+    Column("replay_id", UUID()),
+    # Columns ordered by column-name.
+    any_if_nullable_string("browser_name"),
+    any_if_nullable_string("browser_version"),
+    sum("count_dead_clicks"),
+    sum("count_errors"),
+    sum("count_infos"),
+    sum("count_rage_clicks"),
+    count_nullable("count_segments"),
+    sum("count_urls"),
```
MeredithAnya, line 162: @cmanallen I see that some of these columns are materialized already, so with this materialized table, are we now going to be doing materialization twice? would we need the materialized columns still if we have this table? maybe I don't quite know exactly from the product standpoint if both are needed.

cmanallen: @MeredithAnya We'll still need the materialized columns.  Some queries aren't possible with the materialized view so we have a tiered system where performance degrades as queries get more complex.

snuba/snuba_migrations/replays/0019_add_materialization.py
```diff
@@ -0,0 +1,192 @@
...
+            storage_set=StorageSetKey.REPLAYS,
+            order_by="(project_id, toStartOfHour(to_hour_timestamp), replay_id)",
+            partition_by="(retention_days, toMonday(to_hour_timestamp))",
+            settings={"index_granularity": "8192"},
+            ttl="to_hour_timestamp + toIntervalDay(retention_days)",
+        ),
+        target=operations.OperationTarget.LOCAL,
+    )
+
+    yield operations.AddIndex(
```
evanh, line 43: Why do you need this index if `replay_id` is already in the primary key?

evanh (approved):

snuba/snuba_migrations/replays/0019_add_materialization.py
cmanallen: @evanh Right because we can binary search for the replay-id.  I can remove it.

cmanallen:
@evanh Since you last reviewed I'm now aggregating by `retention_days`.  Should be fine.  Lmk if there's an issue with it.

snuba/snuba_migrations/replays/0019_add_materialization.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def forward_iter() -> Iterator[operations.SqlOperation]:
+    yield operations.CreateTable(
+        storage_set=StorageSetKey.REPLAYS,
+        table_name="replays_aggregated_local",
+        columns=columns,
+        engine=table_engines.AggregatingMergeTree(
+            storage_set=StorageSetKey.REPLAYS,
+            order_by="(project_id, toStartOfHour(to_hour_timestamp), replay_id)",
```
evanh, line 35: I think you might need `retention_days` in the `order_by`, but then you could add a `primary_key` attribute with just these three columns.

cmanallen: I think it needs to be in the primary key right?  In case the same replay has multiple retention periods we'd want the materialized rows to expire at the same time as the source table.

At the moment we only support one retention period but I figure its better to preserve this than not.

JoshFerge (approved):
makes sense to me!

evanh (approved):

snuba/snuba_migrations/replays/0019_add_materialization.py
evanh: That works too.

PR #5711, opened 2024-04-01
feat(kafka-control-plane): bump sentry-kafka-schemas and add topic test

lynnagara:
Validates that all topics are registered in sentry-kafka-schemas as this will be used as the source of truth for creating and managing topics in various environments

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,7 +23,7 @@ python-dateutil==2.8.2
 python-rapidjson==1.8
 redis==4.3.4
 sentry-arroyo==2.17.1
-sentry-kafka-schemas==0.1.65
+sentry-kafka-schemas==0.1.67
 sentry-redis-tools==0.3.0
 sentry-relay==0.8.44
 sentry-sdk==1.40.5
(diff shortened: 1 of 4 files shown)


rust_snuba/src/processors/snapshots/rust_snuba__processors__tests__schemas@snuba-queries-QuerylogProcessor-snuba-queries__1__snuba-queries-empty-trace-id.json.snap
```diff
@@ -0,0 +1,102 @@
+---
+source: src/processors/mod.rs
```
ayirr7, line 2: what's the reason for adding this snapshot for the querylog consumer?

ayirr7 (approved):

rust_snuba/src/processors/snapshots/rust_snuba__processors__tests__schemas@snuba-queries-QuerylogProcessor-snuba-queries__1__snuba-queries-empty-trace-id.json.snap
lynnagara: it was needed to make some tests pass -- i think because the schemas library was updatedd

PR #5706, opened 2024-03-29
feat(replay): viewed_by_id migration

aliu3ntry:

PR #5712, opened 2024-04-01
feat(replay): add ReplayViewedEvent to replay processor

aliu3ntry:
relates to https://github.com/getsentry/sentry/issues/64924 and https://github.com/getsentry/snuba/pull/5706


volokluev (approved):

rust_snuba/src/processors/replays.rs
```diff
@@ -357,6 +368,12 @@ struct ReplayEventLinkEvent {
     warning_id: Option<Uuid>,
 }
 
+#[derive(Debug, Deserialize)]
+struct ReplayViewedEvent {
+    user_id: u64, // not the same as the column in ReplayRow. This is the id of a Sentry user in postgres, belonging to the authorized user POSTing to the viewed-by endpoint
```
cmanallen, line 373: Rather than putting a code comment here let's name the field something more contextual so it doesn't require one.  We could call it `viewed_by_id` or `sentry_user_id`.  Naming it `user_id` will create confusion because of the `user_id` column.

rust_snuba/src/processors/replays.rs
```diff
@@ -418,9 +435,10 @@ pub struct ReplayRow {
     trace_ids: Vec<Uuid>,
     urls: Vec<String>,
     user_email: String,
-    user_id: String,
+    user_id: String, // provided by customer
```
cmanallen, line 438: This comment can be removed.  See previous review comment.

rust_snuba/src/processors/replays.rs
```diff
@@ -188,6 +188,15 @@ pub fn deserialize_message(
                 ..Default::default()
             }]
         }
+        ReplayPayload::ViewedEvent(event) => {
+            vec![ReplayRow {
+                replay_id: replay_message.replay_id,
+                viewed_by_id: event.user_id,
+                event_hash: Uuid::from_u64_pair(0, event.user_id), // for dedup
+                timestamp: event.timestamp as u32,
+                ..Default::default()
```
cmanallen, line 197: We definitely do not want to default the `project_id` field.  You can pull it from the outer scope.  Its defined on the `ReplayMessage` type.

rust_snuba/src/processors/replays.rs
```diff
@@ -188,6 +188,15 @@ pub fn deserialize_message(
                 ..Default::default()
             }]
         }
+        ReplayPayload::ViewedEvent(event) => {
+            vec![ReplayRow {
+                replay_id: replay_message.replay_id,
+                viewed_by_id: event.user_id,
+                event_hash: Uuid::from_u64_pair(0, event.user_id), // for dedup
```
cmanallen, line 195: Is this deterministic?

cmanallen (changes requested):
Needs test coverage.

rust_snuba/src/processors/replays.rs
aliu3ntry: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.from_u64_pair
Yup, it just concatenates the inputs

cmanallen (changes requested):

rust_snuba/src/processors/replays.rs
```diff
@@ -421,6 +439,7 @@ pub struct ReplayRow {
     user_id: String,
     user_name: String,
     user: String,
+    viewed_by_id: u64,
     warning_id: Uuid,
 }
```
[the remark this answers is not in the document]
cmanallen: This was added before you pushed test oof

rust_snuba/src/processors/replays.rs
```diff
@@ -357,6 +369,12 @@ struct ReplayEventLinkEvent {
     warning_id: Option<Uuid>,
 }
 
+#[derive(Debug, Deserialize)]
+struct ReplayViewedEvent {
+    sentry_user_id: u64,
```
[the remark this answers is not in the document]
aliu3ntry: Makes sense, I'll change it. The idea was to explicitly mark this as a sentry user id. But it could be inferred from the u64 type and I guess the user id comes from the request context (not user provided).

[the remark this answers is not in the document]
cmanallen: In this context, we don't know what this id is or where it comes from.  All we know is that the value of this field should be inserted into the column `viewed_by_id`.  The best name for that field is therefore `viewed_by_id`.  Think data first; systems second.  This is just a message.

cmanallen (changes requested):

cmanallen (approved):

PR #5714, opened 2024-04-01
perf(metrics): Use partial deserialization

nikhars:
The current behavior of the generic metrics processor is to deserialize the whole message and see whether the message needs to be processed or not. So when a counter consumer sees a distribution message, it spends a lot of time just deserializing the data and checking whether it should process it or not. This has led to a scaling model of the consumers where all consumers need to scale together in lock step.

This PR aims to change that. Instead of deserializing the whole payload, it will only perform 2 step deserialization. In the first step it deserializes only part of the message which is needed to perform the necessary checks. If the initial step determines that a full deserialization is necessary, then it would perform that. But if its not needed we return early.

### Tests
Modified the logic in the [parse](https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/generic_metrics.rs#L196) method by introducing a panic and seeing that it panic'd when a counters processor saw a set message(i.e. [this](https://github.com/getsentry/snuba/blob/master/rust_snuba/src/processors/generic_metrics.rs#L754) test). Then after the change was introduced, validated that the panic did not happen. While we can keep the behavior of panic, that didn't seem wise to do since it is in the ingestion pipeline. But the functionality is verified.


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -287,7 +311,13 @@ pub fn process_counter_message(
     _metadata: KafkaMessageMetadata,
     config: &ProcessorConfig,
 ) -> anyhow::Result<InsertBatch> {
-    process_message::<CountersRawRow>(payload, config)
+    let payload_bytes = payload.payload().context("Expected payload")?;
+    let msg: PartialMetricsMessage = serde_json::from_slice(payload_bytes)?;
```
evanh, line 315: I tried to do some research because I was curious but I couldn't find anything: how does this save compute cost?

nikhars: The idea is, if on a pipeline there are 1000 messages, lets say with an average length of 1 KB. All the pipelines today need to deserialize 1000 * 1KB = 1MB of data with full deserialization. If lets say out of those 1000 messages, a pipeline is only interested in 10 of those (because they belong to its type of processing), then in theory it only needs to deserialize 10 * 1KB = 10KB of its own data. Lets say the partial message is of size 10. And in addition it needs to deserialize 1000 messages * 10 size = 10KB of all data. So in total, the pipeline is deserializing 10KB + 10KB = 20KB of data. Deserializing 20KB should take considerably lesser CPU than deserializing 1MB of data.

nikhars: I will write a benchmark test to see if I can show the improvement

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -49,6 +49,30 @@ fn generate_timeseries_id(
     adler.checksum()
 }
 
+// MetricType should be used only with PartialMetricsMessage. Its purpose is to
+// be able to categorize the type of metric message we see without incurring the
+// full cost of deserializing the entire payload.
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(tag = "type")]
```
Swatinem, line 56: This `tag` along with the `flatten` below looks very weird, but I was checking the docs and it seems like this is the way.
Otherwise, we usually serialize such data-less enums using `FromStr` and `Display`, with some annotations of how to serialize them. Though I would say this approach here is fine as well.

rust_snuba/src/processors/generic_metrics.rs
Swatinem: I would also be very curious. I would assume that the whole JSON still needs to be lexed at least to make sure that the parser can properly skip over properties that it is not interested in.
The only thing you are potentially saving here is converting the unused fields into proper data types, potentially allocating and copying strings around, etc.

A benchmark would surely be nice here.

Swatinem (commented):
code lgtm, though I would be very curious about a benchmark here. In particular, you are at least double-lexing (maybe not *full* parsing) the json payload now, so this might as well be a pessimization.

Also, am I missing something? This is a workaround with a whole lot of complexity and possible a perf hit in the happy-path for something that should really be "don’t send a Distribution to the Counter" topic?

rust_snuba/src/processors/generic_metrics.rs
evanh: >  I would assume that the whole JSON still needs to be lexed at least to make sure that the parser can properly skip over properties that it is not interested in.

This is my concern too. Although maybe it's capable of bailing early if it finds the keys it's looking for? Anyways, this should be proved with a benchmark.

nikhars: Documented the results here https://www.notion.so/sentry/Benchmarking-performance-of-doing-partial-deserialization-in-snuba-generic-metrics-consumers-96a42d0039c84d018ce4dbdcfc15392a. We can clearly see an improvement with running the benchmark.

nikhars:
> code lgtm, though I would be very curious about a benchmark here. In particular, you are at least double-lexing (maybe not _full_ parsing) the json payload now, so this might as well be a pessimization.
> 
> Also, am I missing something? This is a workaround with a whole lot of complexity and possible a perf hit in the happy-path for something that should really be "don’t send a Distribution to the Counter" topic?

Yeah, the ideal world would be not sending a distribution type on the counter metric. But that might take sometime to materialize. So for project week, i was going to see if there is some short term solution to the scaling model of our consumers.

rust_snuba/src/processors/generic_metrics.rs
nikhars: Yeah this is definitely more nuanced than other fields which need serialization/deserialization

ayirr7:
I think this change overall makes sense due to promising benchmark results. Do we think it would make sense to also add some unit tests just to check the basic functionality of skipping non-relevant messages for the specific processors? 

Also, since there was some partial deserialization done in this PR https://github.com/getsentry/snuba/pull/5617 I think there may need to be a small rebase (mainly, I think, the use case needs to be added to your `PartialMetricsMessage` struct.

nikhars:
> I think this change overall makes sense due to promising benchmark results. Do we think it would make sense to also add some unit tests just to check the basic functionality of skipping non-relevant messages for the specific processors?
> 
> Also, since there was some partial deserialization done in this PR #5617 I think there may need to be a small rebase (mainly, I think, the use case needs to be added to your `PartialMetricsMessage` struct.)

I did run the tests as mentioned in the PR description. Would suggest reading that and seeing if that makes sense. If you have suggestions on how to unit test without changing the underlying behavior of rejecting message in the actual message processing function, please let me know. I could not think of an easier way to test the performance improvement.

I will rebase my change to account for your change. Thanks for the heads up

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -435,6 +471,11 @@ pub fn process_distribution_message(
     _metadata: KafkaMessageMetadata,
     config: &ProcessorConfig,
 ) -> anyhow::Result<InsertBatch> {
+    // We do not perform the optimization of parital deserialization for distribution
+    // type messages because these messages are the highest in terms of volume and the
+    // penalty of double deserialization is worse than deserializing all the messages.
```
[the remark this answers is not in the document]
nikhars: Yes. I have updated the original documentation with more cases. This is the relevant section https://www.notion.so/sentry/Benchmarking-performance-of-doing-partial-deserialization-in-snuba-generic-metrics-consumers-96a42d0039c84d018ce4dbdcfc15392a?pvs=4#59db255dfc8b4f91b1933665549926ef

ayirr7:
> > I think this change overall makes sense due to promising benchmark results. Do we think it would make sense to also add some unit tests just to check the basic functionality of skipping non-relevant messages for the specific processors?
> > Also, since there was some partial deserialization done in this PR #5617 I think there may need to be a small rebase (mainly, I think, the use case needs to be added to your `PartialMetricsMessage` struct.)
> 
> I did run the tests as mentioned in the PR description. Would suggest reading that and seeing if that makes sense. If you have suggestions on how to unit test without changing the underlying behavior of rejecting message in the actual message processing function, please let me know. I could not think of an easier way to test the performance improvement.
> 
> I will rebase my change to account for your change. Thanks for the heads up

Ah, maybe I can clarify what I meant for the unit test. I was just thinking that we can add a unit test that, for example, calls `process_counter_message` on a `DUMMY_GAUGE_MESSAGE` and then we should be able to assert that we get an empty/skipped `InsertBatch`. Just for basic functionality test if possible.

nikhars:
Closing in favor of https://github.com/getsentry/snuba/pull/5756 which gives much more benefit in terms of performance

PR #5715, opened 2024-04-02
feat(metrics): Create a table for finding meta tag values

evanh:
In the original meta table, having tag values in a groupUniqArray had very poor
query performance, driven by the slowness in aggregating those data sketches.

Create a new table that can just be used for finding tag values, that stores
the tag values as rows in the table.

This table also deliberately does not have org_id or use_case_id in it. In
theory any query for tag values has a project ID, metric ID and tag key in it.
This means the org ID and use case ID don't provide anything useful.

### DIFF
--- a/snuba/snuba_migrations/generic_metrics/0033_counters_meta_tag_values_table.py
+++ b/snuba/snuba_migrations/generic_metrics/0033_counters_meta_tag_values_table.py
@@ -1,0 +1,66 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import AggregateFunction, Column, DateTime, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+from snuba.utils.schemas import Float
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    granularity = "2048"
+    local_table_name = "generic_metric_counters_meta_tag_value_aggregated_local"
+    dist_table_name = "generic_metric_counters_meta_tag_value_aggregated_dist"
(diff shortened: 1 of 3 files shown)


enochtangg (commented):
Is my understanding correct that we will now have two tables, one for querying tag keys and one for querying tag values? Does this mean we will have two storages and two entities?

evanh:
> Is my understanding correct that we will now have two tables, one for querying tag keys and one for querying tag values?

Yes. 

>  Does this mean we will have two storages and two entities?

I'm not sure yet about storages/entities. Right now I'm just trying to find the right solution for storing the data in Clickhouse. Although in theory, yes multiple entities/storages.

snuba/snuba_migrations/generic_metrics/0033_counters_meta_tag_values_table.py
```diff
@@ -0,0 +1,66 @@
...
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+from snuba.utils.schemas import Float
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    granularity = "2048"
```
nikhars, line 13: We could start using a granularity of 8192 (the default of clickhouse). Lower granularity on tables which have lots of data maybe could come to bite us at some point later. I know the original tables have this lower granularity, but we can probably start fixing this on the new tables.

nikhars (approved):
Minor comment on the granularity used on the table. Though its not a blocker.

snuba/snuba_migrations/generic_metrics/0033_counters_meta_tag_values_table.py
evanh: I made a note of this. These tables are experimental anyways, so I can drop this table and remake it with better settings after we prove out the concept.

PR #5717, opened 2024-04-02
evanh/feat/use composite queries metrics formulas

evanh:
Initial draft of changes to support multi-type formulas.

This PR only has changes to the MQL parser, the rest of the pipeline doesn't work yet.

To see an example of a finished query, look at `tests/query/parser/test_formula_mql_query.py`.


snuba/query/data_source/join.py
```diff
@@ -160,15 +160,6 @@ def get_alias_node_map(self) -> Mapping[str, IndividualNode[TSimpleDataSource]]:
             **self.right_node.get_alias_node_map(),
         }
 
-    def __post_init__(self) -> None:
```
evanh, line 163: This was removed because of resolution: this check happens before resolving happens, so there are bunch of columns in the query that "don't exist".

snuba/query/mql/context_population.py
```diff
@@ -0,0 +1,185 @@
+from __future__ import annotations
```
evanh, line 1: Moved all of this logic to this file to simplify the parser code a bit. Also made it able to work for simple and formula queries.

tests/query/joins/test_metrics_subqueries.py
```diff
@@ -0,0 +1,260 @@
+from __future__ import annotations
```
evanh, line 1: Ignore this for now.

tests/query/joins/test_subqueries.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
evanh, line 1: Ignore this for now.

enochtangg (commented):
I know the pipeline isn't suppose to be functional yet, but something I came across when re-organizing the query pipeline is adding the storages set keys to `JOINABLE_STORAGE_SETS`  in order for composite query pipeline to work properly: https://github.com/getsentry/snuba/blob/8261937ccd1c4d0870538e446681be7b889f2041/snuba/clusters/storage_sets.py#L86

evanh:
Unnecessary now.

PR #5718, opened 2024-04-02
feat(capman): Long term rejection allocation policy

volokluev:
Currently we throttle on bytes_scanned but never reject. This new policy (which is only needed for errors atm) keeps track of a ten minute sliding window for all `(project|organization, referrer)` combinations. This makes it a much more memory heavy process but we have plenty of redis headroom atm.


snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -0,0 +1,259 @@
...
+
+# we don't limit the amount of bytes subscriptions can scan at this time
+_PASS_THROUGH_REFERRERS = set(
+    [
+        "subscriptions_executor",
+    ]
+)
+
+
+UNREASONABLY_LARGE_NUMBER_OF_BYTES_SCANNED_PER_QUERY = int(1e12)
```
evanh, line 34: This constant name is not long enough :)

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -0,0 +1,259 @@
...
+    def _are_tenant_ids_valid(
+        self, tenant_ids: dict[str, str | int]
+    ) -> tuple[bool, str]:
+        if self.is_cross_org_query(tenant_ids):
+            return True, "cross org query"
+        if tenant_ids.get("referrer") is None:
+            return False, "no referrer"
+        return True, ""
+
+    def _get_customer_tenant_key_and_value(
```
evanh, line 95: Why does this return a project id OR an org id? Wouldn't we want to check both levels?

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -0,0 +1,259 @@
...
+        query_id: str,
+        result_or_error: QueryResultOrError,
+    ) -> None:
+        ids_are_valid, why = self._are_tenant_ids_valid(tenant_ids)
+        if not ids_are_valid:
+            # we already logged the reason before the query
+            return
+        if self.is_cross_org_query(tenant_ids):
+            return
+        if result_or_error.error:
```
evanh, line 216: I'm not sure we should skip the update on an error. I wouldn't want someone to send huge queries that time out, and all those time outs don't count against the limit.

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
volokluev: To be honest because the ConcurrentLimitAllocationPolicy does the same thing. 

If there's a project_id in the tenant_ids that means one project is being queried so we consume the project quota.
If there's no project_id in the tenant_ids that means many projects are being queried, so we consume the organization quota

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
volokluev: That's a good call, how about a configurable timeout query bytes_scanned penalization?

enochtangg (approved):
LGTM, thanks volo.

PR #5720, opened 2024-04-03
chore(gocd): Removing old config arg

IanWoodard:
This is a no-op change that removes a temporary argument used while transitioning to a different approach for rollbacks a few months back.


IanWoodard:
> Why is the CI failing?

I believe GitHub was having issues at the time, should be fixed now

evanh (approved):

PR #5721, opened 2024-04-04
ref: Remove sessions dataset

lynnagara:
doesn't exist anymore. this change means it cannot be queried anymore, and can be made now since sentry tests no longer use this endpoint. entities + storages should be removed as a follow up.

this depends on https://github.com/getsentry/sentry/pull/68226


evanh:
Do you mind cleaning up the failing tests? Otherwise this is good! Can't wait to see this finally go away.

volokluev (approved):

PR #5729, opened 2024-04-05
bug(dev-exp): Create copy request object for new pipeline

enochtangg:
When testing the new pipeline, there exist difference between the SQL queries outputted from the old and new pipeline. This the `Request` object is passed into the old pipeline, then is passed into the new pipeline. This is incorrect since the old pipeline mutates this object (specifically the query). This PR is responsible for creating a deep copy of the request object and passing it into the new pipeline so that both inputs to the pipelines are the same.

### DIFF
--- a/snuba/web/query.py
+++ b/snuba/web/query.py
@@ -73,16 +73,8 @@ def _run_query_pipeline(
         )
         clickhouse_query = StorageProcessingStage().execute(clickhouse_query)
     else:
-        request_copy = copy.deepcopy(request)
-        clickhouse_query = EntityAndStoragePipelineStage().execute(
-            QueryPipelineResult(
-                data=request,
-                query_settings=request.query_settings,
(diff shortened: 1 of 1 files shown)


snuba/web/query.py
```diff
@@ -72,6 +73,7 @@ def _run_query_pipeline(
         )
         clickhouse_query = StorageProcessingStage().execute(clickhouse_query)
     else:
+        request_copy = copy.deepcopy(request)
```
evanh, line 76: Can you change this so the copy only happens if `try_new_query_pipeline` is true?

enochtangg: Good catch. Updated, but I had to wrap `EntityAndStoragePipelineStage` with two if-statements so it doesn't look pretty, but it's just temporary.

evanh (approved):

PR #5731, opened 2024-04-08
fix: parser, bug in pushdown filter 

kylemumma:
The following error was happening during MQL parsing:
https://sentry.sentry.io/issues/5082818756/events/489d5cf1818c4909a985bb0c27ad409c/?project=300688

it was being caused by queries that look like `((avg(d:transactions/millisecond) * 100.0) * 50.0){filter}`
When pushdown_filter is called, it recursively will get called on the `100.0`, which causes `AttributeError: 'float' object has no attribute 'formula'`

In order to fix this, we skip pushdown_filter on anything that isnt an InitialParseResult, this is already done in the code but only for the very top layer of recursion. My fix moves this logic into the recursive case in order to ensure this error wont happen at any layer of recursion.

My fix also adds a test case for the query that previously errored.

related links:
https://www.notion.so/sentry/ddm-parsing-error-880b03749fd04681be77356ea7b65adb

### DIFF
--- a/snuba/query/dsl.py
+++ b/snuba/query/dsl.py
@@ -1,9 +1,27 @@
 from typing import Optional, Sequence
 
-from snuba.query.expressions import Column, Expression, FunctionCall, Literal
+from snuba.query.expressions import (
+    Column,
+    Expression,
+    FunctionCall,
+    Literal,
+    SubscriptableReference,
+)
 
 # Add here functions (only stateless stuff) used to make the AST less
 # verbose to build.
+
+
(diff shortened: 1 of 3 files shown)


volokluev:
@kylemumma Can you summarize the issue and explain the fix in the PR decsription? As a reader, I shouldn't have to follow two links to understand what this PR is fixing. This is also an open source codebase and our open source users do not have a way to see that info

snuba/query/dsl.py
kylemumma, line 1: this addition isnt necessary, but I build this at one point while making this change and left it bc I think it will be useful in the future

evanh (approved):

PR #5733, opened 2024-04-09
fix(meta): Remove experimental meta tables

evanh:
These tables have some missing features:

- The meta table doesn't need tag values
- The meta table has the tag key as a String instead of an int
- Both tables should use a granularity of 8192

Instead of fixing these with modifications, drop the tables completetly and
recreate them with a correct schema.


volokluev (approved):
drop it like it's hot 🔥

Issue #5695, opened 2024-03-27
[EPIC] Improved meta query performance


evanh:
https://www.notion.so/sentry/Metrics-Autocomplete-Performance-d154d3a27d954847921f6b6be5f99d6a
```[tasklist]
### Tasks
- [x] Document potential solutions https://www.notion.so/sentry/Meta-Queries-Run-in-Snuba-a823dfea444b4573ba1838ff0c1ff985?pvs=4#e1f93027127942cda1b1b3e39f60d49e
- [ ] https://github.com/getsentry/snuba/pull/5681
- [ ] https://github.com/getsentry/snuba/pull/5680
- [ ] https://github.com/getsentry/snuba/pull/5687
- [x] Test query performance of V1 meta table https://www.notion.so/sentry/Meta-Queries-Run-in-Snuba-a823dfea444b4573ba1838ff0c1ff985?pvs=4#49138f36034e45768d30a43568bec0f1
- [ ] https://github.com/getsentry/snuba/issues/5696
- [ ] https://github.com/getsentry/snuba/pull/5715
- [ ] https://github.com/getsentry/snuba/issues/5723
- [ ] https://github.com/getsentry/snuba/pull/5716
- [ ] https://github.com/getsentry/snuba/pull/5726
- [ ] https://github.com/getsentry/snuba/pull/5733
- [ ] https://github.com/getsentry/snuba/pull/5734
- [ ] https://github.com/getsentry/snuba/pull/5735
- [ ] https://github.com/getsentry/snuba/pull/5746
- [ ] https://github.com/getsentry/snuba/pull/5747
- [ ] https://github.com/getsentry/snuba/pull/5748
- [ ] https://github.com/getsentry/snuba/pull/5749
- [ ] https://github.com/getsentry/snuba/pull/5759
- [ ] https://github.com/getsentry/snuba/pull/5760
- [ ] https://github.com/getsentry/snuba/pull/5769
- [ ] https://github.com/getsentry/snuba/pull/5784
- [ ] https://github.com/getsentry/snuba/pull/5797
```

PR #5734, opened 2024-04-09
feat(meta): Add updated versions of meta tables for counters

evanh:
These are the final versions of the meta tables that should be used going
forwards. Since the counters table already has the `record_meta` column added,
these tables can be created directly.

This creates two tables populated by materialized views. The first contains
just tag keys, to be used to look up metric IDs for a project, and tag keys for
a given metric ID.

The second table contains tag values, to be used for looking up tag values for
a given metric ID and tag key. This table drops org_id and use_case_id since
those are superfluous given a project_id + metric_id filter.

These tables were tested and researched as part of this epic: https://github.com/getsentry/snuba/issues/5695


volokluev (approved):
Great job on this Evan

PR #5735, opened 2024-04-09
feat(meta): Add record_meta columns to metrics tables

evanh:
This column is used to determine which data should be added to the meta tables,
particularly so that use cases that don't need meta data don't get added
unnecessarily.

This column was added to the counters table as part of the initial experiment,
so it's just the other tables that need it added.


nikhars (changes requested):
I wouldn't recommend adding columns to tables which are in different clickhouse clusters in a single migration. Please create separate migrations for the different tables and clusters.

We haven't run this pattern before. I would assume for good reason. But I do not remember the reasoning at the moment.

nikhars (approved):

PR #5737, opened 2024-04-10
ref(ch-upgrades): create dist tables functionality

MeredithAnya:
Each time we run the series of these pipelines (fetcher, replayer, comparer) we first have to bring up nodes from backups and have our test clusters to work with. Doing the backup will restore the local tables, along with the data, but usually the dist tables are not backed up (if we change our backup process maybe we can change this) but as of now SRE folks would have to manually create the dist tables.

This aims to save the schemas of the dist tables during the fetching process, and when replaying for the first time we should be downloading the schema and using it to create the tables.

I tested this locally with my own gcs bucket

### DIFF
--- a/snuba/cli/query_fetcher.py
+++ b/snuba/cli/query_fetcher.py
@@ -44,19 +44,6 @@ def get_credentials() -> Tuple[str, str]:
     return ("default", "")
 
 
-def get_table_names(connection: ClickhousePool) -> Sequence[str]:
-    name_results = connection.execute("SHOW TABLES").results
-    if not name_results:
-        raise Exception("No tables names found")
-    return [n[0] for n in name_results]
-
-
-def get_table_schema(connection: ClickhousePool, table: str) -> str:
-    return connection.execute(
-        f"SELECT create_table_query FROM system.tables WHERE name = '{table}'"
-    ).results
-
-
 @click.command()
 @click.option(
     "--querylog-host",
--- a/snuba/cli/query_fetcher.py
+++ b/snuba/cli/query_fetcher.py
@@ -9,6 +9,8 @@ from snuba.clickhouse.upgrades.comparisons import (
     FileFormat,
     FileManager,
     QueryInfoResult,
+    delete_local_file,
+    full_path,
 )
 from snuba.clusters.cluster import ClickhouseClientSettings
 from snuba.environment import setup_logging, setup_sentry
@@ -40,6 +42,19 @@ def get_querylog_query(
 def get_credentials() -> Tuple[str, str]:
     # TOOO don't hardcode credentials, use settings
     return ("default", "")
+
+
+def get_table_names(connection: ClickhousePool) -> Sequence[str]:
+    name_results = connection.execute("SHOW TABLES").results
+    if not name_results:
+        raise Exception("No tables names found")
+    return [n[0] for n in name_results]
+
+
+def get_table_schema(connection: ClickhousePool, table: str) -> str:
+    return connection.execute(
+        f"SELECT create_table_query FROM system.tables WHERE name = '{table}'"
+    ).results
 
 
 @click.command()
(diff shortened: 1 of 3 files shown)


snuba/cli/query_replayer.py
```diff
@@ -8,20 +9,25 @@
...
+    delete_local_file,
+    full_path,
 )
 from snuba.clusters.cluster import ClickhouseClientSettings
 from snuba.environment import setup_logging, setup_sentry
 from snuba.utils.gcs import GCSUploader
 
 logger = structlog.get_logger().bind(module=__name__)
 
+TABLE_DOESNT_EXIST = 60
```
nikhars, line 29: If this is a clickhouse error, i believe they should expose it. If yes, lets use that instead.

snuba/cli/query_replayer.py
```diff
@@ -170,27 +192,66 @@ def get_version() -> str:
             blob_getter.get_name_diffs(("queries/", f"{results_directory}/"))
         )
 
+    def create_table_from_err(err: ClickhouseError) -> None:
```
nikhars, line 195: Nit: I think a better interface for this method would be `create_table(table_name)`. The logic of decoding the error and figuring out the table name should be separate logic.

snuba/cli/query_replayer.py
```diff
@@ -115,6 +130,12 @@ def get_credentials(user: Optional[str], password: Optional[str]) -> Tuple[str,
     required=True,
     default=30,
 )
+@click.option(
+    "--cluster-name",
+    help="Option to send saved csv file to slack",
```
nikhars, line 135: Please fix the help message

snuba/cli/query_replayer.py
```diff
@@ -56,6 +62,15 @@ def get_credentials(user: Optional[str], password: Optional[str]) -> Tuple[str,
     return (user or "default", password or "")
 
 
+# Kind of hacky but the saved schemas will have one
+# of these clusters defined
+CLUSTERS = {
+    "snuba-st",
+    "snuba-generic-metrics",
+    "snuba-shared",
+}
```
nikhars, line 71: Could you elaborate why this is needed at all. Why can't the CLI arguments explicitly provide this information?

snuba/cli/query_fetcher.py
```diff
@@ -154,6 +156,30 @@ def get_table_names() -> Sequence[str]:
...
+        when running the replayer for the first time.
+
+        Only upload missing schemas, puts them in same bucket
+        under the following convention:
+
+        schemas/table_name.sql
+        """
+        filename = f"{table}.sql"
+        if uploader.blob_exists(f"schemas/{filename}"):
+            return
```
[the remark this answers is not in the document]
MeredithAnya: @nikhars it is also really easy to delete buckets and re-run commands if we need which is why i wasn't worried about overwriting

snuba/cli/query_replayer.py
MeredithAnya: I can add it as a CLI arg - was just trying to limit how many args there were, and trying to make it easier to not think about what the previous cluster name was, just to know the current one (e.g. `snuba-test`) but i can update

snuba/cli/query_replayer.py
MeredithAnya: ah yeah clickhouse driver has them in the `ErrorCodes`

snuba/cli/query_replayer.py
MeredithAnya: isolated the parsing of the error

snuba/cli/query_replayer.py
MeredithAnya: updated

snuba/cli/query_fetcher.py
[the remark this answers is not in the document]
MeredithAnya: updated anyway

snuba/cli/query_fetcher.py
```diff
@@ -164,6 +188,9 @@ def get_table_names() -> Sequence[str]:
     interval = timedelta(hours=1)
 
     for table in table_names:
+        # we'll use the table schema in the replayer, so
+        # save the create_table_query if not already saved
```
nikhars, line 192:
```suggestion
        # save the create_table_query
```

nikhars (approved):

PR #5746, opened 2024-04-10
fix(meta): Add record_meta field to metrics processing

evanh:
The `record_meta` field is used to decide whether a use case should have meta
information tracked for it. Meta information is used to search metric IDs, tag
keys and tag values, and not all use cases need that information in a
performant way. This was previously just used in counters as testing,  but now
move it out to be used by all the metric types.

Depends on:
- https://github.com/getsentry/snuba/pull/5735
- https://github.com/getsentry/snuba/pull/5759
- https://github.com/getsentry/snuba/pull/5760


rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -519,6 +532,11 @@ impl Parse for GaugesRawRow {
             granularities.push(GRANULARITY_TEN_SECONDS);
         }
         let retention_days = enforce_retention(Some(from.retention_days), &config.env_config);
+        let record_meta = match from.use_case_id.as_str() {
+            "escalating_issues" => Some(0),
+            "metric_stats" => Some(0),
+            _ => Some(1),
+        };
```
volokluev, line 539: Can we make this it's own function se there's less copy/pasting og knowledge?

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -218,11 +226,7 @@ impl Parse for CountersRawRow {
         }
         let retention_days = enforce_retention(Some(from.retention_days), &config.env_config);
 
-        let record_meta = match from.use_case_id.as_str() {
-            "escalating_issues" => Some(0),
-            "metric_stats" => Some(0),
-            _ => Some(1),
-        };
+        let record_meta = should_record_meta(from.use_case_id.clone());
```
nikhars, line 229: And then this can change to pass by reference. That way there isn't necessarily a need to clone() this
```suggestion
        let record_meta = should_record_meta(from.use_case_id.as_str());
```

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -169,6 +169,16 @@ struct CommonMetricFields {
     hr_retention_days: Option<u8>,
     #[serde(skip_serializing_if = "Option::is_none")]
     day_retention_days: Option<u8>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    record_meta: Option<u8>,
+}
+
+fn should_record_meta(use_case_id: String) -> Option<u8> {
```
nikhars, line 176: Mind adding some test cases for this function? I acknowledge that this code is trivial. But adding a test for it would prevent future regressions.

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -169,6 +169,16 @@ struct CommonMetricFields {
...
+    match use_case_id.as_str() {
```
nikhars, line 177: It doesn't seem like you need a String object here. You could do the following here
```suggestion
fn should_record_meta(use_case_id: &str) -> Option<u8> {
    match use_case_id {
```

PR #5747, opened 2024-04-10
feat(meta) Create sets meta tables

evanh:
This creates the local table, distributed table and materialized view for the
meta tables of the sets cluster.

Depends on https://github.com/getsentry/snuba/pull/5735
Depends on https://github.com/getsentry/snuba/pull/5797


snuba/snuba_migrations/generic_metrics/0036_sets_meta_tables.py
```diff
@@ -0,0 +1,182 @@
...
+                    ttl="timestamp + toIntervalDay(retention_days)",
+                ),
+                columns=self.meta_table_columns,
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=self.storage_set_key,
+                table_name=self.meta_dist_table_name,
+                engine=table_engines.Distributed(
+                    local_table_name=self.meta_local_table_name, sharding_key=None
```
volokluev, line 63: is the table name right here?

snuba/snuba_migrations/generic_metrics/0036_sets_meta_tables.py
```diff
@@ -0,0 +1,182 @@
...
+                    ttl="timestamp + toIntervalDay(retention_days)",
+                ),
+                columns=self.tag_value_table_columns,
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=self.storage_set_key,
+                table_name=self.tag_value_dist_table_name,
+                engine=table_engines.Distributed(
+                    local_table_name=self.tag_value_local_table_name, sharding_key=None
```
volokluev, line 115: here as well, I think you want `dist`

snuba/snuba_migrations/generic_metrics/0036_sets_meta_tables.py
evanh: This setting is telling the Distributed engine which local table to use as the storage node. You can see `table_name` above has the dist table in it.

snuba/snuba_migrations/generic_metrics/0036_sets_meta_tables.py
evanh: See above.

snuba/snuba_migrations/generic_metrics/0036_sets_meta_tables.py
volokluev: 🤦

volokluev (approved):

PR #5749, opened 2024-04-10
feat(meta): Create gauges meta tables

evanh:
This creates the local table, distributed table and materialized view for the
meta tables of the gauges cluster.

Depends on #5760 
Depends on https://github.com/getsentry/snuba/pull/5747


snuba/snuba_migrations/generic_metrics/0036_gauges_meta_tables.py
```diff
@@ -0,0 +1,182 @@
...
+        Column("project_id", UInt(64)),
+        Column("metric_id", UInt(64)),
+        Column("tag_key", UInt(64)),
+        Column("tag_value", String()),
+        Column("timestamp", DateTime(modifiers=Modifiers(codecs=["DoubleDelta"]))),
+        Column("retention_days", UInt(16)),
+        Column("count", AggregateFunction("sum", [Float(64)])),
+    ]
+
+    storage_set_key = StorageSetKey.GENERIC_METRICS_SETS
```
nikhars, line 41: Please fix the storage set key to gauges.

snuba/snuba_migrations/generic_metrics/0036_gauges_meta_tables.py
```diff
@@ -0,0 +1,182 @@
...
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+from snuba.utils.schemas import Float
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    granularity = "8192"
+    meta_view_name = "generic_metric_gauges_meta_mv"
+    meta_local_table_name = "generic_metric_gauges_meta_local"
```
nikhars, line 15: Not a blocker. But is it necessary to build 2 tables? Wouldn't the `meta_tag_values` table also be able to answer the same questions about give back the list of all metric id's and tag keys?

evanh: Not in it's current form, since I stripped `use_case_id` and `org_id` out of the table. Both of those would be needed to guarantee finding accurate `metric_id`.

Here was my thought process:

The table that satisfies finding metric IDs and tag keys needs `org_id` and `use_case_id`. However it doesn't need tag values, and tag values account for 98% of the size of that table (3.61GiB out of 3.66GiB).

The tag value table doesn't need `org_id` or `use_case_id`, since it is guaranteed to always have a `metric_id`. So by discarding those columns, I save a bunch of unnecessary duplicated storage.

So by having two tables, I can store the smallest amount of things, with as little duplicated data as possible.

nikhars (approved):

PR #5750, opened 2024-04-10
chore: Lower max query size check to 128KiB

enochtangg:
The actual max query size that can be sent to ClickHouse is `131535 bytes`. This is approximately `128.45KiB`. However, snuba has set `MAX_QUERY_SIZE_BYTES` to `256KiB`, As a result, the large queries are incorrectly passing the snuba check and reaching ClickHouse resulting in `ClickHouseError: max query size exceeded`. This PR is responsible for lowering that limit such that snuba catches these errors and returns the proper expection.

More info here: https://www.notion.so/sentry/ClickHouse-max-query-size-exceeded-error-a5851e643b3f435d9dcb1551dbcbbc09


volokluev (approved):
nice

PR #5755, opened 2024-04-10
chore(capman): set default bytes scanned limit for rejecting policy

volokluev:
This is the limit that is in production right now and it's much higher than what I put in originally. Putting it in the code so we don't lose this number. Making the policy active by default. (we don't need to enforce this in single tenant at this time


snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -38,7 +38,7 @@
     get_redis_client(RedisClientKey.RATE_LIMITER)
 )
 DEFAULT_OVERRIDE_LIMIT = -1
-DEFAULT_BYTES_SCANNED_LIMIT = 10000000
+DEFAULT_BYTES_SCANNED_LIMIT = 640000000000
```
nikhars, line 41: Reading the value is hard since its so big. How about this instead?
```suggestion
GIGABYTE = 10 ** 9
DEFAULT_BYTES_SCANNED_LIMIT = 640 * GIGABYTE
```

evanh (approved):

PR #5762, opened 2024-04-11
chore: create tests for each API pipeline stage

kylemumma:
Added a general case test for `EntityProcessingStage`, `StorageProcessingStage`, and `ExecutionStage`.

I decided not to test `EntityAndStoragePipelineStage` since it will be phased out soon.

### DIFF
--- a/tests/pipeline/test_entity_processing_stage.py
+++ b/tests/pipeline/test_entity_processing_stage.py
@@ -43,7 +43,7 @@ class NoopQueryProcessor(LogicalQueryProcessor):
         query.add_condition_to_ast(equals(literal(1), literal(1)))
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture()  # (scope="module")
 def mock_storage() -> Generator[ReadableTableStorage, None, None]:
     # setup
     storkey = StorageKey("mockstorage")
@@ -71,7 +71,7 @@ def mock_storage() -> Generator[ReadableTableStorage, None, None]:
     del _STORAGE_SET_CLUSTER_MAP[storage.get_storage_set_key()]
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture()  # (scope="module")
 def mock_entity(
     mock_storage: ReadableTableStorage,
 ) -> Generator[PluggableEntity, None, None]:
--- a/tests/pipeline/test_entity_processing_stage.py
+++ b/tests/pipeline/test_entity_processing_stage.py
@@ -1,0 +1,102 @@
+from datetime import datetime
+
+from snuba.attribution import get_app_id
+from snuba.attribution.attribution_info import AttributionInfo
+from snuba.clickhouse.query import Query
+from snuba.datasets.factory import get_dataset
+from snuba.datasets.storages.storage_key import StorageKey
+from snuba.pipeline.query_pipeline import QueryPipelineResult
+from snuba.pipeline.stages.query_processing import EntityProcessingStage
+from snuba.query import SelectedExpression
+from snuba.query.dsl import and_cond, binary_condition, column, equals, literal
+from snuba.query.expressions import Column, FunctionCall
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.query.snql.parser import parse_snql_query
+from snuba.request import Request
+from snuba.utils.metrics.timer import Timer
+
+
+def test_basic():
+    query_body = {
+        "query": (
+            "MATCH (metrics_distributions) "
+            "SELECT avg(granularity) "
+            "WHERE "
+            "timestamp >= toDateTime('2021-05-17 19:42:01') AND "
+            "timestamp < toDateTime('2021-05-17 23:42:01') AND "
+            "org_id = 1 AND "
+            "project_id = 1"
+        ),
+    }
+    metrics_dataset = get_dataset("metrics")
+    query, _ = parse_snql_query(query_body["query"], metrics_dataset)
+    query_settings = HTTPQuerySettings()
+    timer = Timer("test")
+    request = request = Request(
+        id="",
+        original_body=query_body,
+        query=query,
+        snql_anonymized="",
+        query_settings=query_settings,
+        attribution_info=AttributionInfo(
+            get_app_id("blah"), {"tenant_type": "tenant_id"}, "blah", None, None, None
+        ),
+    )
+    res = EntityProcessingStage().execute(
+        QueryPipelineResult(
+            data=request,
+            query_settings=request.query_settings,
+            timer=timer,
+            error=None,
+        )
+    )
+    assert res.data and not res.error
+    assert res.query_settings == query_settings and res.timer == timer
+    assert isinstance(res.data, Query)
+    actual = res.data
+    assert (
+        actual.get_arrayjoin() is None
+        and actual.get_prewhere_ast() is None
+        and actual.get_groupby() == []
+        and actual.get_having() is None
+        and actual.get_orderby() == []
+        and actual.get_limitby() is None
+        and actual.get_offset() == 0
+        and not actual.has_totals()
+        and actual.get_granularity() is None
+        and actual.get_limit() == 1000
+    )
+    assert actual.get_from_clause().storage_key == StorageKey.METRICS_DISTRIBUTIONS
+    expected_selected = [
+        SelectedExpression(
+            "avg(granularity)",
+            FunctionCall(
+                "_snuba_avg(granularity)",
+                "avg",
+                (Column("_snuba_granularity", None, "granularity"),),
+            ),
+        )
+    ]
+    assert actual.get_selected_columns() == expected_selected
+    expected_cond = and_cond(
+        equals(column("granularity"), literal(60)),
+        and_cond(
+            binary_condition(
+                "greaterOrEquals",
+                column("timestamp", alias="_snuba_timestamp"),
+                literal(datetime(2021, 5, 17, 19, 42, 1)),
+            ),
+            and_cond(
+                binary_condition(
+                    "less",
+                    column("timestamp", alias="_snuba_timestamp"),
+                    literal(datetime(2021, 5, 17, 23, 42, 1)),
+                ),
+                and_cond(
+                    equals(column("org_id", alias="_snuba_org_id"), literal(1)),
+                    equals(column("project_id", alias="_snuba_project_id"), literal(1)),
+                ),
+            ),
+        ),
+    )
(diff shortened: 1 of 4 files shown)


tests/pipeline/test_entity_processing_stage.py
```diff
@@ -0,0 +1,102 @@
...
+    res = EntityProcessingStage().execute(
+        QueryPipelineResult(
+            data=request,
+            query_settings=request.query_settings,
+            timer=timer,
+            error=None,
+        )
+    )
+    assert res.data and not res.error
+    assert res.query_settings == query_settings and res.timer == timer
```
kylemumma, line 54: i did this to make sure they were the same, but in hindsight maybe ill delete this

tests/pipeline/test_entity_processing_stage.py
```diff
@@ -0,0 +1,102 @@
...
+        and actual.get_groupby() == []
+        and actual.get_having() is None
+        and actual.get_orderby() == []
+        and actual.get_limitby() is None
+        and actual.get_offset() == 0
+        and not actual.has_totals()
+        and actual.get_granularity() is None
+        and actual.get_limit() == 1000
+    )
+    assert actual.get_from_clause().storage_key == StorageKey.METRICS_DISTRIBUTIONS
```
kylemumma, line 69: I did this because I didnt want to recreate the entire Table object, but im not sure if its necessary to make sure it exactly matches or not...

tests/pipeline/test_storage_processing_stage.py
```diff
@@ -0,0 +1,88 @@
...
+from snuba.utils.metrics.timer import Timer
+
+
+def test_basic():
+    ch_query = Query(
+        from_clause=Table(
+            "metrics_distributions_v2_local",
+            ColumnSet([]),
+            storage_key=StorageKey.METRICS_DISTRIBUTIONS,
+        ),
```
volokluev, line 23: These tests should not use existing storages/entities, they should be independent of what we already have defined

tests/pipeline/test_storage_processing_stage.py
```diff
@@ -0,0 +1,88 @@
...
+
+    res = StorageProcessingStage().execute(
+        QueryPipelineResult(
+            data=deepcopy(ch_query),
+            query_settings=settings,
+            timer=timer,
+            error=None,
+        )
+    )
+    assert res.data == Query(
```
volokluev, line 70: so as we look at this test, what did it actually test? Have we actually asserted that the pipeline works as we expect it to?

tests/pipeline/test_execution_stage.py
```diff
@@ -0,0 +1,98 @@
...
+                        equals(
+                            column("project_id", alias="_snuba_project_id"), literal(1)
+                        ),
+                    ),
+                ),
+            ),
+        ),
+        limit=1000,
+    )
+    res = ExecutionStage(
```
volokluev, line 79: This stage does more than return the query, there are side effects that our application depends on, those should be tested.

tests/pipeline/test_entity_processing_stage.py
volokluev, line 69: we should recreate the table object, and again, this should be using a set of entities/storages that are created for this test

volokluev (commented):
Good start, but these tests should be living/enforced documentation of what these stages do, right now, that is not what they are

tests/pipeline/test_storage_processing_stage.py
kylemumma: I didn't have a solid understanding of what the storage pipeline is supposed to do, I read through it carefully to understand what it is supposed to do better and modified my test accordingly

tests/pipeline/test_execution_stage.py
```diff
@@ -0,0 +1,98 @@
...
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.querylog.query_metadata import SnubaQueryMetadata
+from snuba.request import Request
+from snuba.utils.metrics.timer import Timer
+from snuba.utils.schemas import DateTime, UInt
+
+
+@pytest.mark.clickhouse_db
+@pytest.mark.redis_db
+def test_basic():
```
volokluev, line 25: the execution stage has a crucial step of writing to the querylog. We should make sure that happens in this test

volokluev: Also missing a test for dry run queries

tests/pipeline/test_execution_stage.py
```diff
@@ -0,0 +1,98 @@
...
+            ColumnSet(
+                [
+                    ("org_id", UInt(64)),
+                    ("project_id", UInt(64)),
+                    ("timestamp", DateTime()),
+                    ("granularity", UInt(32)),
+                ]
+            ),
+            storage_key=StorageKey.METRICS_DISTRIBUTIONS,
+        ),
```
volokluev, line 43: Given that this is a high level test, this should include an allocation policy. The execution stage executes allocation policies which is quite important.

tests/pipeline/test_execution_stage.py
```diff
@@ -0,0 +1,98 @@
...
+        limit=1000,
+    )
+    res = ExecutionStage(
+        attinfo,
+        query_metadata=SnubaQueryMetadata(
+            Request("", {}, ch_query, settings, attinfo, ""),
+            "blah",
+            timer,
+        ),
+    ).execute(
```
volokluev, line 86: While this is not a great thing, this is part of the stage at this time, we apply sampling in the execution stage right now, we should make sure that it does happen

https://github.com/getsentry/snuba/blob/14d84ada93e04f3f12f82a8a813d1d029ec627c7/snuba/pipeline/stages/query_execution.py#L292-L310

tests/pipeline/test_entity_processing_stage.py
kylemumma: done

tests/pipeline/test_execution_stage.py
volokluev: > it looks like writing to the querylog happens outside of the execution stage

ah okay that's good 😌 

still should add the dry run test though

kylemumma:
## major changes
### entity stage
* using new/fake/isolated entity/storage

### storage stage
* using new/fake/isolated entity/storage
* got rid of mocks, now I have a simple query processor that I validate gets run

### execution stage
* added test for dry_run
* added simple allocation policy and make sure its applied
* added test for turbo_sampling

tests/pipeline/test_entity_processing_stage.py
```diff
@@ -0,0 +1,127 @@
...
+    storage_set_key=storsetkey,
+    schema=TableSchema(
+        columns=colset,
+        local_table_name=f"{storkey.value}_local",
+        dist_table_name=f"{storkey.value}_dist",
+        storage_set_key=storsetkey,
+    ),
+    readiness_state=ReadinessState.COMPLETE,
+)
+_STORAGE_SET_CLUSTER_MAP[storage.get_storage_set_key()] = CLUSTERS[0]
```
volokluev, line 55: you should put this in a pytest fixture that gets undone after the test completes

tests/pipeline/test_entity_processing_stage.py
```diff
@@ -0,0 +1,127 @@
...
+entity = PluggableEntity(
+    entity_key=entkey,
+    storages=[EntityStorageConnection(storage, TranslationMappers())],
+    query_processors=[],
+    columns=colset.columns,
+    validators=[],
+    required_time_column="timestamp",
+    storage_selector=DefaultQueryStorageSelector(),
+)
+override_entity_map(entkey, entity)
```
volokluev, line 65: here as well

tests/pipeline/test_entity_processing_stage.py
```diff
@@ -0,0 +1,127 @@
...
+        dist_table_name=f"{storkey.value}_dist",
+        storage_set_key=storsetkey,
+    ),
+    readiness_state=ReadinessState.COMPLETE,
+)
+_STORAGE_SET_CLUSTER_MAP[storage.get_storage_set_key()] = CLUSTERS[0]
+entity = PluggableEntity(
+    entity_key=entkey,
+    storages=[EntityStorageConnection(storage, TranslationMappers())],
+    query_processors=[],
```
volokluev, line 59: why no query processors here?

tests/pipeline/test_execution_stage.py
```diff
@@ -0,0 +1,190 @@
...
+
+@pytest.mark.clickhouse_db
+@pytest.mark.redis_db
+def test_basic() -> None:
+    attinfo = AttributionInfo(
+        get_app_id("blah"), {"tenant_type": "tenant_id"}, "blah", None, None, None
+    )
+    settings = HTTPQuerySettings()
+    timer = Timer("test")
+    res = ExecutionStage(attinfo, query_metadata=mockmetadata).execute(
```
volokluev, line 116: this query_metadata object gets updated on execution, the variable local to the test and you should probably check what's in it after this is done

tests/pipeline/test_execution_stage.py
```diff
@@ -0,0 +1,190 @@
...
+            timer=timer,
+            error=None,
+        )
+    )
+    assert (
+        res.data
+        and len(res.data.result["data"]) == 1
+        and "avg(duration)" in res.data.result["data"][0]
+    )
+    assert policy.did_update
```
volokluev, line 129: if you're only using the policy in this test, it should be instantiated in this test, right now this test has global state which is a recipe for bugs

tests/pipeline/test_storage_processing_stage.py
```diff
@@ -0,0 +1,85 @@
...
+        dist_table_name="mockstoragedist",
+        storage_set_key=StorageSetKey("mockstorageset"),
+    ),
+    readiness_state=ReadinessState.COMPLETE,
+    query_processors=[MockQueryProcessor()],
+    mandatory_condition_checkers=[],
+    allocation_policies=[],
+)
+get_config_built_storages()[mock_storage.get_storage_key()] = mock_storage
+_STORAGE_SET_CLUSTER_MAP[mock_storage.get_storage_set_key()] = CLUSTERS[0]
```
volokluev, line 40: this like this should be done in fixtures

volokluev (commented):
Good overall, just needs less global state

kylemumma:
## major changes
* got rid of global state and replace for fixtures with proper setup & teardown
* added some validation to the metadata object in execution stage test
* added entity processor to entity stage test

volokluev (approved):

PR #5767, opened 2024-04-12
chore(on-call): Add metric for concurrent queries by referrer that violate policy

enochtangg:
This PR is responsible for adding a metric in the `ReferrerGuardRailPolicy` that tracks the number of concurrent queries for each referrer that violate that policy. To do this, the `_is_within_rate_limit()` function was split into two smaller functions in order to access the `rate_limit_stats`.


snuba/query/allocation_policies/concurrent_rate_limit.py
```diff
@@ -80,6 +81,11 @@ def _is_within_rate_limit(
             rate_limit_prefix,
             self.get_config_value("max_query_duration_s"),
         )
+        return rate_limit_stats
+
+    def _is_within_rate_limit(
```
volokluev, line 86: I think the easier thing would be to return the rate_limit_stats from the original _is_within_rate_limit_function, that way you only have to change the call site of the other policies, not the actual logic

snuba/query/allocation_policies/per_referrer.py
```diff
@@ -93,10 +93,23 @@ def _get_quota_allowance(
...
+        can_run, explanation = self._is_within_rate_limit(
+            rate_limit_stats,
+            rate_limit_params,
+        )
+        if not can_run:
+            self.metrics.increment(
+                "concurrent_query_exceeded_limit",
+                rate_limit_stats.concurrent,
+                tags={"referrer": referrer},
+            )
```
volokluev, line 112: This metric:

1. Should be a histogram (which in our case is a timing metric)
2. should be emitted regardless of whether the query can run
3. should be called something like "concurrent_queries_referrer"

We want to use this to basically say "how many concurrent queries is each referrer executing right now"  and max that number for every time slice.

snuba/query/allocation_policies/concurrent_rate_limit.py
enochtangg: To me, it doesn't seem semantically correct to have `_is_within_rate_limit` return `rate_limit_stats`. As the function name implies, it feels like it should just return `True/False` and the explanation of why it is or isn't within the rate limits. That's why I decided to change split them and add a `_get_rate_limit_stats` function. What do you think?

volokluev: I'm not too attached to my opinion but here's what I'll say:

Before this change, every policy would just have to call `_is_within_rate_limit` now every policy has to call `_get_rate_limit_stats` and `_is_within_rate_limit` 

You can't really call `_is_within_rate_limit` without getting the rate limit information first, these things will never be called separately. I'm thinking of the `rate_limit_stats` as another detailed piece of the explanation as to why the function returned True/False and that's why I consider it the more reasonable thing to surface that data alongside the explanation.

volokluev (approved):
Functionally the change is good but consider my reply to your comment about splitting out vs returning the stats

snuba/query/allocation_policies/concurrent_rate_limit.py
enochtangg: > You can't really call _is_within_rate_limit without getting the rate limit information first, these things will never be called separately. I'm thinking of the rate_limit_stats as another detailed piece of the explanation

Thats fair. Updated, thanks volo!

PR #5768, opened 2024-04-12
feat(trace): Add trace id to discover&transactions

wmak:
- We're going to be querying traces more and more as part of the changes around removing event details. This adds an index on trace id to discover & transactions
- I mostly followed along what happened in https://github.com/getsentry/snuba/pull/5558 but I don't really understand why the migration is on table_local, and not table_dist? Maybe I'm missing something


evanh:
> I don't really understand why the migration is on table_local, and not table_dist?

The dist table doesn't actually scan any data, it farms the queries out to the local tables, which scan the data, aggregate and send it back to the dist table. The dist table then aggregates the results from the local tables. So the index is on the local tables.

Something we'll need to look into is whether you need the index on the discover table if it's already on the transactions table. The discover table is a merge table, I don't know if adding indexes there works as expected. @nikhars would probably be the best person to answer that.

wmak:
@evanh I think according to docs we don't need one in discover they mention
https://clickhouse.com/docs/en/engines/table-engines/special/merge
> When reading, the indexes of tables that are actually being read are used, if they exist.

nikhars:
> @evanh I think according to docs we don't need one in discover they mention https://clickhouse.com/docs/en/engines/table-engines/special/merge
> 
> > When reading, the indexes of tables that are actually being read are used, if they exist.

That's correct. The merge table will not perform any table data scan of its own. So the index needs to be added only to the underlying tables.

nikhars (approved):

PR #5774, opened 2024-04-15
chore(on-call): Add CrossOrgQueryAllocationPolicy to errors

enochtangg:
This PR is responsible for adding the CrossOrgQueryAllocationPolicy to errors in order to give us better control over referrers that run across all orgs (e.g.: [backfill seer grouping records script](https://github.com/getsentry/sentry/pull/68466)).

### DIFF
--- a/snuba/datasets/configuration/events/storages/errors.yaml
+++ b/snuba/datasets/configuration/events/storages/errors.yaml
@@ -275,6 +275,10 @@ allocation_policies:
       default_config_overrides:
         is_enforced: 0
         is_active: 0
+      cross_org_referrer_limits:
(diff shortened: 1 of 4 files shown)


volokluev:
Since you're already here, you should add config for the `getsentry.tasks.backfill_grouping_records` referrer. Give them like 4 concurrent queries and two threads or something, we can raise it if they need it

volokluev:
See https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/storages/distributions.yaml#L121 for an example

volokluev (approved):
Add the referrer config and then we should be good

PR #5776, opened 2024-04-15
feat(devexp): add ability to parse simple storage queries

volokluev:
This is the most simple version of this PR:

Work that is needed to make this really work:

* Add `required_time_column` to storages so we can do the date alignment logic we use for entities
    * or we just don't care like what we're doing right now
* Add BasicFunctionsProcessor as a post_process step so it's entity independent


snuba/query/dsl.py
```diff
@@ -23,18 +47,6 @@ def literal(value: OptionalScalarType, alias: str | None = None) -> Literal:
     return Literal(alias, value)
 
 
-def snuba_tags_raw(indexer_mapping: int) -> SubscriptableReference:
```
volokluev, line 26: was not used

snuba/query/dsl.py
```diff
@@ -91,8 +103,12 @@ def binary_condition(
     return FunctionCall(None, function_name, (lhs, rhs))
 
 
-def equals(lhs: Expression, rhs: Expression) -> FunctionCall:
-    return binary_condition("equals", lhs, rhs)
+def equals(
+    lhs: Expression | OptionalScalarType, rhs: Expression | OptionalScalarType
+) -> FunctionCall:
```
volokluev, line 108: happy to change this but I found it useful to not have to specify literals

snuba/query/logical.py
```diff
@@ -81,3 +79,73 @@ def _transform_expressions_impl(
 
     def _transform_impl(self, visitor: ExpressionVisitor[Expression]) -> None:
         pass
+
+
+class StorageQuery(ProcessableQuery[Storage]):
```
volokluev, line 84: TODO: I think this type can be generalized

snuba/query/snql/parser.py
```diff
@@ -1532,8 +1617,6 @@ def parse_snql_query(
         with sentry_sdk.start_span(op="processor", description="time_based_processing"):
             _post_process(query, [_replace_time_condition], settings)
 
-        # XXX: Select the entity to be used for the query. This step is temporary. Eventually
```
volokluev, line 1535: RIP this comment

snuba/query/snql/parser.py
```diff
@@ -360,6 +397,14 @@ def visit_entity_name(self, node: Node, visited_children: Tuple[Any]) -> EntityK
         except Exception:
             raise ParsingException(f"{node.text} is not a valid entity name")
 
+    def visit_storage_name(
+        self, node: Node, visited_children: Tuple[Any]
+    ) -> StorageKey:
+        try:
+            return StorageKey(node.text)
+        except Exception:
+            raise ParsingException(f"{node.text} is not a valid Storage name")
```
kylemumma, line 406: unrelated to current PR, but one thought I have in general is that it might be nice to be able to parse StorageKey and EntityKey even if it doesnt exist. And then later on you get an error if you try to access one that doesnt exist. This would make using fake ones in testing easier.

snuba/query/snql/parser.py
```diff
@@ -319,8 +338,13 @@ def visit_match_clause(
...
             join_clause = build_join_clause(match)
             return join_clause
+        if not isinstance(match, (QueryEntity, QueryStorage)):
+            import pdb
+
+            pdb.set_trace()
+            print(match)
 
-        assert isinstance(match, QueryEntity)  # mypy
+        assert isinstance(match, (QueryEntity, QueryStorage))  # mypy
```
kylemumma, line 347: random comment that you can ignore: if you care to read a relevant excerpt of the pragmatic programmer about "assertive programming" https://www.informit.com/articles/article.aspx?p=2982114&seqNum=3

snuba/query/snql/parser.py
```diff
@@ -319,8 +338,13 @@ def visit_match_clause(
         ):
             join_clause = build_join_clause(match)
             return join_clause
+        if not isinstance(match, (QueryEntity, QueryStorage)):
+            import pdb
+
+            pdb.set_trace()
+            print(match)
```
kylemumma, line 345:
```suggestion
```

snuba/query/snql/parser.py
```diff
@@ -1225,17 +1279,17 @@ def mangle_column_value(exp: Expression) -> Expression:
...
         return replace(exp, column_name=f"{alias_prefix}{exp.column_name}")
-
     query.transform_expressions(mangle_aliases, skip_array_join=True)
 
     # Check if this query has a subquery. If it does, we need to mangle the column name as well
     # and keep track of what we mangled by updating the mappings in memory.
-    if isinstance(query.get_from_clause(), LogicalQuery):
+    if isinstance(query.get_from_clause(), (LogicalQuery, StorageQuery)):
         query.transform_expressions(mangle_column_value)
+    print(query)
```
kylemumma, line 1288:
```suggestion
```

kylemumma (commented):
My biggest concern would be the introduction of a new StorageQuery class rather than going directly to a ClickhouseQuery.

snuba/query/logical.py
volokluev: Good question. The main thing is that the `Table` datasaoure does not resemble the entity in any real way and it's only by the time we get to the `StorageProcessing` stage that we can process the clickhouse query. 

Ideally what I would do is remove `Table` as a datasource and probably remove `ClickhouseQuery` later.

kylemumma: Why does it matter that Table doesn't resemble Entity?

PR #5779, opened 2024-04-16
inc-715: rust consumer can stop processing messages at specific timestamp

lynnagara:


rust_snuba/src/strategies/processor.rs
```diff
@@ -34,7 +35,18 @@ pub fn make_rust_processor(
         partition: Partition,
         offset: u64,
         timestamp: DateTime<Utc>,
+        stop_at_timestamp: Option<i64>,
     ) -> anyhow::Result<Message<BytesInsertBatch<RowData>>> {
+        // Don't process any more messages
+        if let Some(stop) = stop_at_timestamp {
+            if stop < timestamp.timestamp() {
+                let payload = BytesInsertBatch::default();
```
fpacifici, line 43: I assume this is an empty batch.

rust_snuba/src/strategies/processor.rs
```diff
@@ -214,9 +244,21 @@ impl<TResult: Clone, TNext: Clone> MessageProcessor<TResult, TNext> {
             timestamp: msg.timestamp,
         };
 
+        if let Some(stop_timestamp) = self.stop_at_timestamp {
+            if metadata.timestamp.timestamp() > stop_timestamp {
+                anyhow::bail!("message is after stop timestamp");
+            }
+        }
+
```
fpacifici, line 252: I don't think I understand why this is needed. 
Isn't the condition already in`result_to_next_msg`

ayirr7:
Is it possible to have a unit test that checks that any of the processors returns an empty batch for a message that exceeds stop timestamp?

rust_snuba/src/strategies/processor.rs
ayirr7: +1

fpacifici (approved):

rust_snuba/src/strategies/processor.rs
lynnagara: sorry, this was an old iteration and i forgot to push, updated

ayirr7 (approved):

PR #5782, opened 2024-04-16
ref(spans): Remove unused module column

phacops:
This column is not populated and not used in the frontend anymore and should be removed.


evanh (commented):
Can you split the migration into a separate PR from the other changes (config/consumer)?

phacops:
> Can you split the migration into a separate PR from the other changes (config/consumer)?

So, what exactly do I need to separate here? Since the column is not nullable, wouldn't insertion fail if I change the consumer? So, I can leave a default empty value for that, but then, do I need to keep it in the storage config as well?

phacops:
Migration moved to https://github.com/getsentry/snuba/pull/5792.

phacops:
@evanh I added back to the processor the `module` field with an empty string value to be inserted into ClickHouse since the column is not nullable, inserts would fail if I didn't, right?

evanh:
> @evanh I added back to the processor the `module` field with an empty string value to be inserted into ClickHouse since the column is not nullable, inserts would fail if I didn't, right?

You added that here? If the column doesn't exist, why do you need to write to it?

PR #5784, opened 2024-04-16
feat(meta): Adjust partitioning/settings of counters meta tables

evanh:
There was an issue with the existing tables that caused them to have a very
high number of inactive parts. This change makes three adjustments to help with
that.

- Remove `retention_days` from the `PARTITION BY`. This will reduce the number
  of partitions that the table has, and thus the number of inactive parts
- Change the materialized view to use `toMonday` instead of `toStartOfWeek` so
  that the partition lines up with the data in the table. This should help
prevent writing new data into older partitions (which was happening  because
`toStartOfWeek` and `toMonday` don't line up.
- Since partitions are no longer broken up by retention days,
  add the `ttl_only_drop_parts` setting so that we drop rows within the
partitions based on retention.

Since it's not possible to modify the `PARTITION BY` of a table in place, drop
the existing tables and recreate them with the accurate `PARTITION BY`.

### DIFF
--- a/snuba/snuba_migrations/generic_metrics/0041_adjust_partitioning_meta_tables.py
+++ b/snuba/snuba_migrations/generic_metrics/0041_adjust_partitioning_meta_tables.py
@@ -1,0 +1,332 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import AggregateFunction, Column, DateTime, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+from snuba.utils.schemas import Float
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    granularity = "8192"
+    meta_view_name = "generic_metric_counters_meta_mv"
+    meta_local_table_name = "generic_metric_counters_meta_local"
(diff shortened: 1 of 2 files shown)


snuba/snuba_migrations/generic_metrics/0041_adjust_partitioning_meta_tables.py
```diff
@@ -0,0 +1,332 @@
...
+from snuba.migrations.operations import OperationTarget, SqlOperation
+from snuba.utils.schemas import Float
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+    granularity = "8192"
+    meta_view_name = "generic_metric_counters_meta_mv"
+    meta_local_table_name = "generic_metric_counters_meta_local"
+    meta_dist_table_name = "generic_metric_counters_meta_dist"
```
nikhars, line 16: Since you are dropping tables and recreating them immediately (with the same names), I am not sure how zookeeper would react to it. It is quite possible that it may barf because it sees in its metadata that a table with this name already exists.

A saner way to do this would be to create the mv's and aggregated tables with new names. That way you can completely avoid this problem.

Or alternatively, perform drop tables in 1 migration. And do create tables in another migration.

evanh: That makes sense. I was hoping I could get away with it. I updated to use new tables names.

nikhars (approved):

PR #5786, opened 2024-04-16
Add priority to group attributes table

snigdhas:
### Legal Boilerplate

### DIFF
--- a/snuba/snuba_migrations/group_attributes/0002_add_priority_to_group_attributes.py
+++ b/snuba/snuba_migrations/group_attributes/0002_add_priority_to_group_attributes.py
@@ -1,0 +1,44 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_local",
+                column=Column("group_priority", UInt(8)),
+                target=OperationTarget.LOCAL,
+                after="group_substatus",
+            ),
+            operations.AddColumn(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_dist",
+                column=Column("group_priority", UInt(8)),
+                target=OperationTarget.DISTRIBUTED,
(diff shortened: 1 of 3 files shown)


snuba/snuba_migrations/group_attributes/0002_add_priority_to_group_attributes.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_local",
+                column=Column("group_priority", UInt(8)),
```
scefali, line 17: @snigdhas can we just call it `priority`? I don't think there is a need to call it `group_priority` when it's on a group table. Also, should there be a default value?

snigdhas: sticking with the [existing pattern](https://github.com/getsentry/snuba/blob/92f2ad28d1828e5497f57852e4e2157c8e444e0b/snuba/snuba_migrations/group_attributes/0001_group_attributes.py#L13-L16) where columns from the group table are prefixed with `group_`. 

We _do_ default to Medium priority as the base case on ingestion but I don't believe we need a default here. Since it's a nullable column, it should be fine while we backfilled the rows and then going forwards it should be included in the snapshots.

scefali: @snigdhas fair enough, let's stick to conventions

wedamija (commented):
This looks good to me, do we add to the `group_attributes.yaml` and related files separately to the migration?

evanh:
> This looks good to me, do we add to the `group_attributes.yaml` and related files separately to the migration?

Yes that is a separate PR. The migration PR can only have the migration code in it.

snuba/snuba_migrations/group_attributes/0002_add_priority_to_group_attributes.py
evanh: > Since it's a nullable column

Your migration does not make this column Nullable. You need to add `Modifiers(nullable=True)` to your Column definition.

`UInt(8, Modifiers(nullable=True))`

evanh: > while we backfilled the rows

How are you planning to do this?

snigdhas: using `bulk_send_snapshot_values` in this [migration](https://github.com/getsentry/sentry/pull/69072)

evanh (approved):

PR #5788, opened 2024-04-17
feat(migrations): add discover local

dbanda:
Our migrations in SaaS don't have the discover local table. This causes a mismatch between discover migrations in SaaS and self-hosted/CI. The mismatch prevents us from running any new discover migrations. 

To fix this , this migration adds a `discover_local` table if it doesn't exist to reconcile SaaS with the self-hosted migrations. In SaaS the local table will not be used because we merge the distributed tables directly, but this will allow us to unblock discover migrations.

### DIFF
--- a/snuba/migrations/group_loader.py
+++ b/snuba/migrations/group_loader.py
@@ -136,6 +136,7 @@ class DiscoverLoader(DirectoryLoader):
             "0005_discover_fix_transaction_name",
             "0006_discover_add_trace_id",
             "0007_discover_add_span_id",
+            "0008_discover_fix_new_dist_table",
         ]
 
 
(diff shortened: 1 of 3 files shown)


dbanda:
That is correct. We have the migration but we didn't run it in SaaS.

evanh:
> That is correct. We have the migration but we didn't run it in SaaS.

But then why add the migration again? Why not run the migration in SaaS?

dbanda:
> > That is correct. We have the migration but we didn't run it in SaaS.
> 
> But then why add the migration again? Why not run the migration in SaaS?

The table has changed since then.

evanh (approved):

PR #5789, opened 2024-04-17
feat(migrations): replace discover_dist with discover_dist_new

dbanda:
This is a follow up to #5788 

In this migration we replace `discover_dist` table with `discover_dist_new` that references `discover_local`. This makes our SaaS environments and others consistent with the self-hosted one.

### DIFF
--- a/snuba/snuba_migrations/discover/0009_discover_fix_rename_dist_table.py
+++ b/snuba/snuba_migrations/discover/0009_discover_fix_rename_dist_table.py
@@ -19,13 +19,13 @@ class Migration(migration.ClickhouseNodeMigration):
         return [
             operations.RenameTable(
                 storage_set=StorageSetKey.DISCOVER,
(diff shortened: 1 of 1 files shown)


MeredithAnya:
@dbanda wait I thought self-hosted was the only thing we wanted to fix, since ST and SaaS actually have the same setup ?

dbanda:
> @dbanda wait I thought self-hosted was the only thing we wanted to fix, since ST and SaaS actually have the same setup ? Seems less disruptive to change self-hosted that to change SaaS and ST?

Closing this PR in favor of #5789 .  I think changing self-hosted will be harder because the convention we use is that each storage has local and dist tables. If we drop the local table, we will not be able to run snuba on a single node without adding in extra code to make exceptions for discover which seems quite messy. I opted to instead just add discover local in SaaS and ST. The table wont be used, but it will just be there to keep things consistent.

PR #5787, opened 2024-04-17
 feat(replays): add replay_id column to merged discover table

JoshFerge:
https://github.com/getsentry/snuba/pull/5777 was reverted, opening this up again, with more column processor changes for handling the replay_id column as a uuid

PR #5791, opened 2024-04-17
feat(replays): add replay_id uuid processors to config

JoshFerge:
split out from https://github.com/getsentry/snuba/pull/5787


snuba/datasets/configuration/transactions/storages/transactions.yaml
```diff
@@ -157,6 +157,7 @@ schema:
       },
       { name: app_start_type, type: String },
       { name: profile_id, type: UUID, args: { schema_modifiers: [nullable] } },
+      { name: replay_id, type: UUID, args: { schema_modifiers: [nullable] } },
```
aliu3ntry, line 160: why is this needed for transactions?

aliu3ntry: do you also need to add a replay_id column like this to discover.yaml?

snuba/datasets/configuration/discover/storages/discover.yaml
```diff
@@ -88,6 +88,7 @@ schema:
         args: { schema_modifiers: [nullable], size: 64 },
       },
       { name: deleted, type: UInt, args: { size: 8 } },
+      { name: trace_id, type: UUID, args: { schema_modifiers: [nullable] } }
```
aliu3ntry, line 91: why are we adding trace_id? Isn't this pr a followup to the migration where you added replay_id?

aliu3ntry: Do you also have to add a column to the entities/discover.yaml file?

snuba/datasets/configuration/transactions/storages/transactions.yaml
JoshFerge: because we want to query on this column which exists in the transactions dataset

snuba/datasets/configuration/discover/storages/discover.yaml
JoshFerge: thanks for the catch, should have been replay_id

evanh (approved):

shruthilayaj (approved):

PR #5792, opened 2024-04-17
ref(spans): Add a migration to remove the module column

phacops:
Remove unused column.


snuba/snuba_migrations/spans/0013_spans_remove_module_column.py
```diff
@@ -0,0 +1,38 @@
...
+
+table_name_prefix = "spans"
+column_name = "module"
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        target = operations.OperationTarget.LOCAL
```
dbanda, line 16: I'd avoid doing this since the enum value can change in future.

evanh (commented):
Can you fix the CI?

PR #5794, opened 2024-04-17
fix(self-hosted): respect the max_connections passed through settings

volokluev:
Follow up from #4988 , users noticed that this setting did not actually propagate to the cluster which makes sense because we did not use it

https://github.com/getsentry/snuba/pull/4988/files/29d6765a9bb94fe6cf4ff40575aca2774e710c2c#r1569493727

Update cluster instantiation to make sure we respect the setting

### DIFF
--- a/snuba/clusters/cluster.py
+++ b/snuba/clusters/cluster.py
@@ -195,6 +195,7 @@ class ConnectionCache:
 
 
 connection_cache = ConnectionCache()
+_DEFAULT_MAX_CONNECTIONS = 1
 
 
 class ClickhouseCluster(Cluster[ClickhouseWriterOptions]):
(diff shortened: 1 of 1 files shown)


evanh (approved):

ghlLarney:
@volokluev 
I read the code for creating connections with clickhouse. The problem is that there are separate connection classes being used for reads vs writes, which is fine.
For reads -> https://github.com/getsentry/snuba/blob/0b3bfd837cd87a101979c0a87622013d9d815dd7/snuba/clickhouse/native.py#L74
For writes -> https://github.com/getsentry/snuba/blob/0b3bfd837cd87a101979c0a87622013d9d815dd7/snuba/clickhouse/http.py#L289

The problem is in the HTTPBatchWriter, where **max_connections** and **block_connections** are not being respected. A probable fix would be something like this

`max_connections: int = 1,
        block_connections: bool = False,
    ):
        self.__pool: HTTPSConnectionPool | HTTPConnectionPool
        if secure:
            self.__pool = HTTPSConnectionPool(
                host,
                port,
                maxsize=max_connections,
                block=block_connections,
                ca_certs=ca_certs,
                cert_reqs="REQUIRED" if verify else "CERT_NONE",
            )
        else:
            self.__pool = HTTPConnectionPool(host, port, maxsize=max_connections, block=block_connections)`
            
  Can we re-open this, because people in other thread says that it's still not resolved completely.

PR #5797, opened 2024-04-18
fix(meta) Rename counter tables for consistency

evanh:
These were created with new names to avoid conflicting with the old tables when
the partition keys were changed. However for consistency's sake it would be
nice to have all the tables in the different clusters have the same naming
schema.


enochtangg (approved):

evanh:
This fails CI because the dist table name and local table name are the same locally and in CI. So when rolling back, the names collide and it breaks.

PR #5799, opened 2024-04-18
fix: Don't log this every time, it clogs the logs up

evanh:


enochtangg (approved):

snuba/utils/health_info.py
```diff
@@ -107,7 +107,6 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
         "healthcheck.latency", time.time() - start, tags={"thorough": str(thorough)}
     )
 
-    logger.info(json.dumps(body))
```
volokluev, line 110: This log was helpful in diagnosing an incident. Maybe don't log it if status is ok and the down file does not exist

PR #5802, opened 2024-04-19
fix(22.8) Add processor to handle 'allocator' bug in Clickhouse

evanh:
There is some bug in Clickhouse 22.8 that can cause errors like "Too large size
(13837309855095848992) passed to allocator. It indicates an error."

The issue was narrowed down to conditions on low cardinality string columns,
and specific types of data in those columns. This processor hacks those
conditions to be in a form that allows the query to run.

### DIFF
--- a/snuba/query/processors/logical/condition_simplifier_processor.py
+++ b/snuba/query/processors/logical/condition_simplifier_processor.py
@@ -42,7 +42,12 @@ class ConditionSimplifierProcessor(LogicalQueryProcessor):
                     exp.alias, "equals", (exp.parameters[0], rhs.parameters[0])
                 )
 
-            return FunctionCall(exp.alias, "has", (rhs, exp.parameters[0]))
+            # `has` requires an array, so convert everything to an array
+            return FunctionCall(
+                exp.alias,
+                "has",
+                (FunctionCall(None, "array", rhs.parameters), exp.parameters[0]),
+            )
 
         if state.get_int_config("use.condition.simplifier.processor", 1) == 0:
             return
--- a/snuba/query/processors/logical/condition_simplifier_processor.py
+++ b/snuba/query/processors/logical/condition_simplifier_processor.py
@@ -1,3 +1,4 @@
+from snuba import state
 from snuba.query.conditions import (
     combine_and_conditions,
     get_first_level_and_conditions,
@@ -30,7 +31,6 @@ class ConditionSimplifierProcessor(LogicalQueryProcessor):
                 return exp
 
             rhs = exp.parameters[1]
-            print(rhs)
             if not isinstance(rhs, FunctionCall) or rhs.function_name not in (
                 "tuple",
                 "array",
@@ -43,6 +43,9 @@ class ConditionSimplifierProcessor(LogicalQueryProcessor):
                 )
 
             return FunctionCall(exp.alias, "has", (rhs, exp.parameters[0]))
+
+        if state.get_int_config("use.condition.simplifier.processor", 1) == 0:
+            return
 
         condition = query.get_condition()
         if condition:
--- a/snuba/query/processors/logical/condition_simplifier_processor.py
+++ b/snuba/query/processors/logical/condition_simplifier_processor.py
@@ -1,0 +1,52 @@
+from snuba.query.conditions import (
+    combine_and_conditions,
+    get_first_level_and_conditions,
+    is_condition,
+)
+from snuba.query.expressions import Expression, FunctionCall
+from snuba.query.logical import Query
+from snuba.query.processors.logical import LogicalQueryProcessor
+from snuba.query.query_settings import QuerySettings
+
+
+class ConditionSimplifierProcessor(LogicalQueryProcessor):
+    """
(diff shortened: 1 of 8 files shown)


snuba/query/processors/logical/condition_simplifier_processor.py
```diff
@@ -0,0 +1,52 @@
...
+    e.g. release IN tuple('a', 'b', 'c')
+    will cause an error.
+    THis processor does two things: if the rhs is a single value, change the condition
+    to be equals(). Otherwise, flip the sides using the `has` operator.
+    So has(tuple('a', 'b', 'c'), release) instead of in(release, tuple('a', 'b', 'c'))
+    """
+
+    def process_query(self, query: Query, query_settings: QuerySettings) -> None:
+        def transform_cond(exp: Expression) -> Expression:
+            print("CHECKING", exp)
```
onewland, line 25: we want to remove this print, right?

evanh: What! Why???

snuba/query/processors/logical/condition_simplifier_processor.py
```diff
@@ -0,0 +1,51 @@
...
+    22.8 has some problems when dealing with certain conditions on strings.
+    Specifically, a condition with multiple values on a string column,
+    e.g. release IN tuple('a', 'b', 'c')
+    will cause an error.
+    THis processor does two things: if the rhs is a single value, change the condition
+    to be equals(). Otherwise, flip the sides using the `has` operator.
+    So has(tuple('a', 'b', 'c'), release) instead of in(release, tuple('a', 'b', 'c'))
+    """
+
+    def process_query(self, query: Query, query_settings: QuerySettings) -> None:
```
volokluev, line 23: do we want to put this behind a runtime config so we can turn on/off?

snuba/query/processors/logical/condition_simplifier_processor.py
```diff
@@ -0,0 +1,51 @@
...
+        def transform_cond(exp: Expression) -> Expression:
+            if not is_condition(exp):
+                return exp
+
+            assert isinstance(exp, FunctionCall)
+            if not exp.function_name == "in" or len(exp.parameters) != 2:
+                return exp
+
+            rhs = exp.parameters[1]
+            print(rhs)
```
volokluev, line 33: another print :D

evanh: It's a load bearing print.

volokluev (approved):
removing the print and feature flag would be good but I trust you with those

tests/test_discover_api.py
```diff
@@ -747,7 +747,7 @@ def test_transaction_group_ids(self) -> None:
                 entity="discover_transactions",
             ).data
         )
-
+        print("RESULT", result)
```
volokluev, line 750: 👀

tests/test_snql_api.py
```diff
@@ -1391,6 +1391,75 @@ def test_hexint_in_condition(self) -> None:
...
+        )
+
+    def test_discover_semver_bug(self) -> None:
+        response = self.post(
+            "/discover/snql",
+            data=json.dumps(
+                {
+                    "consistent": False,
+                    "turbo": False,
+                    "query": "MATCH (discover) SELECT event_id AS `id` WHERE release IN array('test@1.2.3+123', 'test2@1.2.4+124') AND timestamp >= toDateTime('2024-04-18T22:28:40.928000') AND timestamp < toDateTime('2024-04-19T22:28:40.306000') AND project_id IN array(4553863597588481) LIMIT 50",
```
volokluev, line 1445: @evanh this is the sentry test that's failing in a bizarre fashion:

```
'DB::Exception: First argument for function has must be an array.. 
```

sql:

```
SELECT (replaceAll(toString(event_id), '-', '') AS _snuba_event_id) FROM discover_local 
PREWHERE in((project_id AS _snuba_project_id), [4553863597588481]) 
WHERE equals(deleted, 0) AND has(['test@1.2.3+123', 'test2@1.2.4+124'], (release AS _snuba_release)) 
AND greaterOrEquals((timestamp AS _snuba_timestamp), toDateTime('2024-04-18T22:28:40', 'Universal')) 
AND less(_snuba_timestamp, toDateTime('2024-04-19T22:28:40', 'Universal')) LIMIT 50 OFFSET 0
```

evanh: Can you replace all the hardcoded values here with test values (project_id, org_id, timestamp etc.)?

volokluev: these are the values that are in the failing sentry test

volokluev: Just tried it, it fails the same way

evanh: This is very weird, because the SQL query works in production and when running it directly on Clickhouse.

evanh: Had a (hacky) idea: I just removed the optimizer from the discover entity. We don't actually need it there, just discover_transactions and transactions.

volokluev (approved):

PR #5804, opened 2024-04-22
fix(ci/cd): Remove broken monitor check

evanh:
The monitor that this was using no longer exists, which is causing the CD
pipeline to fail.

### Legal Boilerplate


volokluev:
was this monitor replaced by something? what was it checking?

enochtangg (approved):
nice

evanh:
> was this monitor replaced by something? what was it checking?

I don't know. I am trying to figure out what the monitor was doing, but in the meantime we need deploys to work again.

PR #5805, opened 2024-04-22
fix(22.8) Cast LowCardinality string columns to String

evanh:
In 22.8, if a LowCardinality(String) is queried, and it actually is not low
cardinality, it can cause an error in Clickhouse. To be safe, cast all
LowCardinality columns to regular String to avoid causing this issue.

So far this is only impacting Discover + transactions, so limit the change
there. This might decrease query performance for other queries, since
all LowCardinality strings will be treated as String.

### DIFF
--- a/snuba/query/processors/logical/low_cardinality_processor.py
+++ b/snuba/query/processors/logical/low_cardinality_processor.py
@@ -1,0 +1,49 @@
+from snuba import state
+from snuba.query.expressions import Column, Expression, FunctionCall, Literal
+from snuba.query.logical import Query
+from snuba.query.processors.logical import LogicalQueryProcessor
+from snuba.query.query_settings import QuerySettings
+
+
+class LowCardinalityProcessor(LogicalQueryProcessor):
+    """
+    22.8 has some problems when dealing with low cardinality string columns,
+    where the actual values in production are high cardinality. For example, in some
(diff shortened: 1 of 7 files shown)


evanh:
In theory this could replace the ConditionSimplifierProcessor, but I am leaving both in for now. We can use the config setting to see if we need both or not, and then remove anything that isn't necessary.

volokluev (approved):

snuba/query/processors/logical/low_cardinality_processor.py
```diff
@@ -0,0 +1,51 @@
...
+            ):
+                return exp
+
+            return FunctionCall(
+                exp.alias,
+                "cast",
+                (
+                    FunctionCall(
+                        None,
+                        "ifNull",
```
Zylphrex, line 41: This `ifNull` cast could be a source of bugs in the search. But given this is necessary for some queries to even run, let's go ahead with this and we can keep an eye out for what else may be impacted.

Zylphrex (approved):

snuba/query/processors/logical/low_cardinality_processor.py
evanh: We have another processor that should hopefully handle this: https://github.com/getsentry/snuba/blob/master/snuba/query/processors/physical/null_column_caster.py#L22

Issue #5722, opened 2024-04-04
Add support for new auth ENVs for ClickHouse auth


williamdes:
It's unclear in the documentation that ClickHouse should have no password on the default account.
See: https://github.com/getsentry/snuba/blob/7a5bd19982deb86af5694ce98a38c878211cba14/docs/source/getstarted.rst#requirements

https://github.com/search?q=repo%3Agetsentry%2Fsnuba%20CLICKHOUSE_HOST&type=code

It would be nice to have:
- CLICKHOUSE_USER
- CLICKHOUSE_PASSWORD

And maybe CLICKHOUSE_PORT

May be related to https://forum.sentry.io/t/secure-clickhouse-connection-with-on-premise-snuba/14919
And https://github.com/getsentry/snuba/pull/2033

enochtangg:
Hi @williamdes, thanks for raising this. We are looking to prioritizing this work.

onewland:
@williamdes can you be more clear about what you need or what specifically isn't working for you? We already use these environment variables in the default settings (if you map all datasets to one cluster):

https://github.com/getsentry/snuba/blob/7a5bd19982deb86af5694ce98a38c878211cba14/snuba/settings/__init__.py#L91-L92

williamdes:
Well, it seems like this was ignored. Maybe only by the rust workers?
I am not sure, but adding the envs in docker did not work and threw errors

onewland:
ok, that makes sense

xurui-c:
Hi William, do you mind telling me a little about how to reproduce the errors you saw? Was it running `sentry devservices up` and then `docker exec -it sentry_clickhouse clickhouse-client`

xurui-c:
Hi @williamdes , could you provide the errors you saw? Thanks!

williamdes:
I was running an external ClickHouse server and the errors are access denied auth errors in the CH server.
I have no logs for it now, but put a password 5o your default Clickhouse user and maybe create a new one too and start testing

andrsp:
i'm using dedicated clickhouse with set up non-default user for sentry (called `sentry`)
seems like new rust-consumers reporting in log on start with right config credentials but tries to connect to clickhouse with user `default` (which is not allowed in my installation) and crashes with something like this:
```
[...]

PR #5818, opened 2024-04-25
Enables Clickhouse Authorization

xurui-c:
https://github.com/getsentry/snuba/issues/5722

Canary: deploy to S4S before rolling out to US SaaS. This shouldn't affect affect production consumers because the default username and password are used.

### DIFF
--- a/rust_snuba/src/factory.rs
+++ b/rust_snuba/src/factory.rs
@@ -162,8 +162,8 @@ impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
             &self.storage_config.clickhouse_cluster.database,
             &self.clickhouse_concurrency,
             self.skip_write,
-            &self.env_config.clickhouse_user,
-            &self.env_config.clickhouse_password,
+            &self.storage_config.clickhouse_cluster.user,
+            &self.storage_config.clickhouse_cluster.password,
         );
 
         let accumulator = Arc::new(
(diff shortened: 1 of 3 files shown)


rust_snuba/src/factory.rs
```diff
@@ -162,6 +162,8 @@ impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
             &self.storage_config.clickhouse_cluster.database,
             &self.clickhouse_concurrency,
             self.skip_write,
+            &self.env_config.clickhouse_user,
+            &self.env_config.clickhouse_password,
```
MeredithAnya, line 166: why can't this be 

```rust
&self.storage_config.clickhouse_cluster.user,
&self.storage_config.clickhouse_cluster.password,
```

snuba/consumers/consumer_config.py
```diff
@@ -129,6 +131,8 @@ def _resolve_env_config() -> EnvConfig:
     lower_retention_days = settings.LOWER_RETENTION_DAYS
     valid_retention_days = list(settings.VALID_RETENTION_DAYS)
     record_cogs = settings.RECORD_COGS
+    clickhouse_user = str(settings.CLUSTERS[0].get("user"))
```
MeredithAnya, line 134: will we always have the same user and password for all clusters? can we use the `StorageConfig` instead of the `EnvConfig`?

MeredithAnya (approved):
@xurui-c looks good, lets maybe add in the description the plan for testing this out - will this be tested with one consumer first? in the canary environment? what happens if this breaks, can we easily revert? etc

williamdes (commented):
Awesome, thank you

PR #5820, opened 2024-04-25
cleanup(inc-723): always run LowCardinalityProcessor, remove ConditionSimpllifier

volokluev:
the `USE_CARDINALITY_CASTER` setting is no longer necessary. it was shipped to fix CI urgently. The real fix is in now. 

The `ConditionSimplifierProcessor` is also unnecessary now. We can remove it from the codebase and tests. The `LowCardinalityProcessor` takes care of this bug by itself


nikhars (approved):

snuba/query/processors/logical/low_cardinality_processor.py
```diff
@@ -41,12 +41,7 @@ def transform_expressions(exp: Expression) -> Expression:
                 ),
             )
 
-        if (
-            state.get_int_config(
-                "use.low.cardinality.processor", settings.USE_CARDINALITY_CASTER
-            )
-            == 0
-        ):
+        if state.get_int_config("use.low.cardinality.processor", 1) == 0:
```
Zylphrex, line 44: Enabling this by default caused some test failures last time that blocked sentry master, have we checked it won't happen again?

volokluev: The processor now works differently, the `ifNull(col, "")` is removed, instead it casts to a `Nullable(String)`. I ran some of the failed tests locally with the new processor and they passed

Zylphrex (approved):

PR #5824, opened 2024-04-26
feat(meta) Add configuration for meta tables in generic metrics

evanh:
This adds all the YAML files for the storages and entities required to query
the meta tables. Note that this is 4 copies of the same files, just replacing
the metric type in each one (counters/sets/gauges/distributions).

A required column for the meta table is the use case ID, but that is a string
column. This also slightly modifies the validator to allow string columns as
well as integer columns. The intent was to keep the strictness of the original
check (which specifically looks for ints), so this added a new similar check.

For reviewers: This is a big PR, but there are really only 4 new files added.
So just reviewing the files for counters would be sufficient.


snuba/datasets/configuration/generic_metrics/entities/counters_meta.yaml
```diff
@@ -0,0 +1,51 @@
+version: v1
```
volokluev, line 1: may be helpful to have a comment in these files explaining what the `meta` tables are for since the name could mean a lot of things

snuba/query/validation/validators.py
```diff
@@ -111,6 +118,16 @@ def validate(self, query: Query, alias: Optional[str] = None) -> None:
                 if not found:
                     missing.add(col)
 
+        missing = set()
```
volokluev, line 121: can you add a test for this new validator functionality specifically?

snuba/datasets/configuration/generic_metrics/storages/counters_meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+allocation_policies:
+  - name: ConcurrentRateLimitAllocationPolicy
+    args:
+      required_tenant_types:
+        - referrer
+        - organization_id
+        - project_id
+      default_config_overrides:
+        is_enforced: 0
```
volokluev, line 39: let's enforce these

snuba/datasets/configuration/generic_metrics/storages/counters_meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+        is_enforced: 1
+        throttled_thread_number: 1
+        org_limit_bytes_scanned: 100000
+  - name: ReferrerGuardRailPolicy
+    args:
+      required_tenant_types:
+        - referrer
+      default_config_overrides:
+        is_enforced: 0
+        is_active: 0
```
volokluev, line 55: Let's enforce these too

snuba/datasets/configuration/generic_metrics/storages/counters_meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+        is_enforced: 0
+  - name: BytesScannedWindowAllocationPolicy
+    args:
+      required_tenant_types:
+        - referrer
+        - organization_id
+      default_config_overrides:
+        is_enforced: 1
+        throttled_thread_number: 1
+        org_limit_bytes_scanned: 100000
```
volokluev, line 48: This number is probably too low

```suggestion
        org_limit_bytes_scanned: 10000000
```

volokluev (commented):
LGTM just some changes to allocation policy suggested and another test

enochtangg (commented):
LGTM. Not sure if you plan to do this in another PR, but you would also need to reference the new entities in the generic metrics dataset YAML.

nikhars (approved):
LGTM

PR #5826, opened 2024-04-26
feat(dev-exp): Move entity validators from parser to entity processing stage

enochtangg:
There exist certain validators in the parsing stage of the pipeline that are dependent on entities. These should be moved into the entity processing stage in order to enable storage direct query parsing.


tests/web/test_query_cache.py
```diff
@@ -29,6 +30,7 @@ def run_query() -> None:
         selected_columns=[
             SelectedExpression("event_id", Column("_snuba_event_id", None, "event_id")),
         ],
+        condition=in_condition(Column(None, None, "project_id"), [Literal(None, 123)]),
```
volokluev, line 33: why was this necessary?

tests/web/test_transform_names.py
```diff
@@ -75,6 +76,7 @@ def test_transform_column_names() -> None:
                 ),
             ),
         ],
+        condition=in_condition(Column(None, None, "project_id"), [Literal(None, 1)]),
```
volokluev, line 79: why was this necessary?

tests/pipeline/test_entity_processing_stage_composite.py
```diff
@@ -328,7 +328,7 @@
                 ),
                 SelectedExpression(
                     "_snuba_right",
-                    Column("_snuba_right", "groups", "right_col"),
+                    Column("_snuba_right", "groups", "status"),
```
volokluev, line 331: why was this necessary?

tests/clickhouse/query_dsl/test_project_id.py
```diff
@@ -176,17 +178,20 @@ def test_find_projects(
...
             assert isinstance(query, Query)
+            run_entity_validators(query, HTTPQuerySettings())
             identity_translate(query)
     else:
         request = json_to_snql(query_body, "events")
         request.validate()
         query, _ = parse_snql_query(str(request.query), events)
         assert isinstance(query, Query)
+        assert isinstance(query, Query)
+        run_entity_validators(query, HTTPQuerySettings())
```
volokluev, line 194: Is this necessary?

volokluev: I just ran the test without this, I don't think you need it

tests/pipeline/test_entity_processing_stage_composite.py
enochtangg: Now that entity validation (including column validators) takes places in the entity processing stage. The tests that call `_run_query()` directly can no longer get away with providing bad or missing column. As a result, we need to update these tests and make them valid.

tests/clickhouse/query_dsl/test_project_id.py
enochtangg: Sure we can remove it. I've added this here because I didn't want to change  what this test use to do (parse the query, run entity validators, then find the project_ids). The entity validator part contains `EntityRequiredColumnValidator` which ensures that a project_id was first passed in before running the project_id finder

volokluev (approved):
Thank you Enoch!

tests/clickhouse/query_dsl/test_project_id.py
enochtangg: Wait, sorry we need this here. There are two tests that are expected to fail because it does not provide project_id (e.g. https://github.com/getsentry/snuba/blob/6680821603eca90018d20f8d9f64203c2086c8c9/tests/clickhouse/query_dsl/test_project_id.py#L48-L56)

volokluev: ok but that's not really testing the component that is under test. what is being tested here is the project id finder.

PR #5833, opened 2024-04-29
chore(admin): Allow product tools role to use  snuba-explain tool

enochtangg:
Snuba admin users with the `ProductTools` role should be able to access and use snuba-explain tool.

### DIFF
--- a/tests/admin/test_authorization.py
+++ b/tests/admin/test_authorization.py
@@ -34,10 +34,12 @@ def test_product_tools_role(
     response = admin_api.get("/tools")
     assert response.status_code == 200
     data = json.loads(response.data)
+    print(data)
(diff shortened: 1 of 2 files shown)


tests/admin/test_authorization.py
```diff
@@ -34,10 +34,12 @@ def test_product_tools_role(
     response = admin_api.get("/tools")
     assert response.status_code == 200
     data = json.loads(response.data)
+    print(data)
```
volokluev, line 37:
```suggestion
```

volokluev (approved):
remove print, otherwise it's good

tests/admin/test_authorization.py
enochtangg: oops thanks

PR #5834, opened 2024-04-29
fix(cd): failed to get success from all checks before timeout

kylemumma:
This change will fix the bug in our CD pipeline where it fails bc "failed to get success from all checks before timeout."

It times out before CI finished, so the solution is to remove timeout from this stage of the pipeline. If CI actually times out on the github end, GOCD will see this failure and fail too (so its safe).


IanWoodard (approved):

volokluev (approved):
thanks kyle

PR #5835, opened 2024-04-30
feat(bucket-encoding): Switch base64 to base64 without padding

Dav1dde:
Relay PR: https://github.com/getsentry/relay/pull/3509

Switches Base64 to Base64 without padding to save some bytes for small values.


loewenheim (approved):

rust_snuba/src/processors/generic_metrics.rs
```diff
@@ -131,7 +132,7 @@ impl<T> EncodedSeries<T> {
             EncodedSeries::Base64 { data, .. } => {
                 counter!("generic_metrics.encoded_receive_count", 1, "format" => "base64");
 
-                let decoded_bytes = BASE64.decode(data.as_bytes())?;
+                let decoded_bytes = BASE64_NOPAD.decode(data.as_bytes())?;
```
nikhars, line 135: Are there already BASE64 encode values flowing in any of the environments like S4S, SaaS or any other ST? If yes, wouldn't it cause this to fail in those environments?

What would happen if you didn't change this [line](https://github.com/getsentry/snuba/pull/5835/files#diff-c3be0a0574331ae4a7d8fa91feb7068877e1190931b6e740ff493b18b3f2ca80L918) with the code already modified?

ayirr7: It looks like we've already reverted back to the prior iteration of the bucket format. So, in s4s, we have the array encoded format flowing (which looks like `"value": {"format": "array", "data":, [1,2,3]}`)

ayirr7: I can look into the test

ayirr7: The previous test cases would fail because they utilize padding. So that's why the test cases have been updated.

nikhars (approved):

ayirr7:
Gonna bump the Kafka schema in this PR and add the generated snapshots so the tests can pass

PR #5836, opened 2024-04-30
chore(devexp): cleanup dead code (again)

volokluev:
When I merged #5793 it conflicted with the new tests which were added and caused mypy issues. We should be able to merge this now


kylemumma (approved):
my gratitude for this change is immeasurable

PR #5838, opened 2024-04-30
chore(rust): Minor logging improvements

nikhars:
Remove a confusing log entry.


ayirr7 (approved):

williamdes:
I guess my logs will be happier empty, but debug level would make sense
The actual code after merge is still confusing, it shows a success without documenting a partial fail

Ref: https://github.com/getsentry/self-hosted/issues/2876#issuecomment-2082298438

PR #5839, opened 2024-04-30
chore(devexp): remove parser-based entity validation

volokluev:
#5826 moved this stage out of parsing, we no longer need it


enochtangg (approved):
thanks!

PR #5840, opened 2024-05-01
feat(devexp): add a default subscriptable mapper to the storage stage

volokluev:
In order to make storage queries work with nested columns, we need to build in default processing of queries that have expressions like:

`tags[something] = 'blah'`

this PR implements this feature. It also removes the SubscriptableMapper for metric_summaries to show that it's no longer necessary

Also add some syntactic sugar to the DSL to make the tests look nicer

### DIFF
--- a/snuba/query/dsl.py
+++ b/snuba/query/dsl.py
@@ -38,14 +38,18 @@ class _FunctionCall:
     def __init__(self, name: str) -> None:
         self.name = name
 
-    def __call__(self, *args: Expression | OptionalScalarType, **kwargs: str):
+    def _arg_to_literal_expr(self, arg: Expression | OptionalScalarType) -> Expression:
+        if isinstance(arg, Expression):
+            return arg
+        return Literal(None, arg)
+
+    def __call__(
+        self, *args: Expression | OptionalScalarType, **kwargs: str
+    ) -> FunctionCall:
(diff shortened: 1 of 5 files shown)


enochtangg (commented):
This would apply transform the expressions to a subscriptable mapper for all storages, is this expected?

snuba/query/dsl.py
```diff
@@ -13,6 +13,47 @@
 # verbose to build.
 
 
+class NestedColumn:
```
kylemumma, line 16: i feel like my mind has been expanded after reading these...

kylemumma (approved):
Hi Volo, it was interesting to see how you implemented this and compare it to how I was thinking about going about it. The DSL classes are so clever, I didn't even know that was possible. Impressive!

volokluev:
@enochtangg 

> This would apply transform the expressions to a subscriptable mapper for all storages, is this expected?

Correct! But it would not matter for any existing storage because they already apply the SubscriptableMapper through config in the entity stage. By the time the query gets to the storage stage, they won't have any `SubscriptableReference` expressions.

enochtangg (approved):
Cool thanks! I ask that question because I was wondering if we have datasets that purposely exclude the subscriptable mapper because it would break queries.

phacops (approved):

PR #5842, opened 2024-05-01
Prevents automatic logging of ClickHouse Authorization information

xurui-c:
We don't want consumers to log ClickHouse username and password, otherwise ClickHouse servers' security could be compromised.

### DIFF
--- a/rust_snuba/src/consumer.rs
+++ b/rust_snuba/src/consumer.rs
@@ -37,7 +37,6 @@ pub fn consumer(
     use_rust_processor: bool,
     enforce_schema: bool,
     max_poll_interval_ms: usize,
-    debug_mode: bool,
     python_max_queue_depth: Option<usize>,
     health_check_file: Option<&str>,
     stop_at_timestamp: Option<i64>,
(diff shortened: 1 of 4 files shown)


rust_snuba/src/consumer.rs
```diff
@@ -80,7 +83,26 @@ pub fn consumer_impl(
...
+            storage.clickhouse_cluster.http_port,
+            storage.clickhouse_cluster.database,
+        );
+        if debug_mode {
+            tracing::info!(
+                "ClickHouse username: {}, ClickHouse password: {}",
+                storage.clickhouse_cluster.user,
+                storage.clickhouse_cluster.password
+            );
+        }
```
volokluev, line 103: I would remove this alltogether

rust_snuba/src/consumer.rs
```diff
@@ -80,7 +83,26 @@ pub fn consumer_impl(
...
+        //tracing::info!(storage.message_processor, "Message Processor")
```
volokluev, line 104: don't leave comments in code :)

volokluev (commented):
personally I don't think the debug mode is necessary, leaves room for a security vulnerability

rust_snuba/src/consumer.rs
evanh, line 104: Is this supposed to be commented?

xurui-c: I forgot to remove this line all together

volokluev (approved):

anonrig (approved):

PR #5843, opened 2024-05-01
feat(performance): Allow the split by char optimization in projects

wmak:
- Pretty sure I'm doing a bunch of this wrong, opening this to hopefully get feedback on the right way to approach this
- I'm also worried about the parameters in the validator being `AnyExpression` but couldn't find a better pattern to use
- This lets the query run now, but i see my local snuba logging: "Query is missing mandatory columns" from the MandatoryConditionEnforcer
- More info on the motivation behind this change here: https://www.notion.so/sentry/Speeding-up-the-Trace-Query-4ba8a9f3c1b445eab6f748dbe2a1817d

### DIFF
--- a/snuba/query/conditions.py
+++ b/snuba/query/conditions.py
@@ -383,5 +383,18 @@ def build_match(
                     ),
                 ),
             ),
+            FunctionCallPattern(
+                Or([String(op) for op in array_ops]),
+                (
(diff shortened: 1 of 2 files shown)


MeredithAnya:
@wmak is this PR still relevant?

PR #5844, opened 2024-05-01
feat(dsl) Add a visitor that maps AST objects into DSL strings

evanh:
Most of the AST objects that exist in the tests have a DSL syntactic sugar
function to go with them. This code will take AST objects as input, and output
a string that is valid Python code for the corresponding DSL function.

This is a rough MVP of this, and could use some polishing/cleanup.


kylemumma (approved):
yay thanks evan!

snuba/query/dsl_mapper.py
```diff
@@ -37,14 +40,26 @@ def and_cond_repr(exp: Expression, visitor: ExpressionVisitor[str]) -> str:
     return f"and_cond({parameters})"
 
 
+or_cond_match = FunctionCallMatch(
+    StringMatch("or"),
+)
+
+
+def or_cond_repr(exp: Expression, visitor: ExpressionVisitor[str]) -> str:
```
volokluev, line 48: why do you need this if we can just use the functions helper and generate basically the same thing?

PR #5846, opened 2024-05-02
ref(cli): Remove skip-write flag

john-z-yang:
### Overview

Remove the `skip-write` flag as the rust consumers are no longer experimental


john-z-yang:
I'm not familiar with the HTTP batch code, but I am wondering if we need to deal with the reporting a little differently here
```rust
pub async fn finish(mut self) -> Result<bool, anyhow::Error> {
    self.flush_chunk()?;
    // finish stream
    drop(self.sender.take());
    if let Some(handle) = self.result_handle.take() {
        handle.await??;
        Ok(true)
    } else {
        Ok(false)
    }
}
```
https://github.com/getsentry/snuba/pull/5846/files#diff-6b1c376f3512244d03fe9f25c4aafcb7fa0335e5fa719d633e6e01a401993288L157-L162

I feel like `self.result_handle` might very rarely be `None` now that we always initialize it to `Some` real handler (because the skip_write option is gone). The only chance that it will be `None` will be somehow `finish` gets called twice, so maybe the else case should be an `anyhow::err!`. But I'm not sure if `finish` gets called twice is something we want to handle.

loewenheim (approved):

nikhars (approved):

john-z-yang:
I am a little hesitant to merge this because running the consumer locally is broken on master (getting an `{"error":"the strategy errored"}`). Going to wait for a fix to land there and verify my changes first.

lynnagara (approved):

hubertdeng123:
What happens when users try passing in 
```
rust-consumer ...--no-skip-write
```

Does this result in an error? I think it should be fine since self-hosted e2e tests passed, but just wanted to make sure
https://github.com/getsentry/self-hosted/blob/1c72fbe6120444ac232df0dae242cd5af40b38df/docker-compose.yml#L261

john-z-yang:
> What happens when users try passing in
> 
> ```
> rust-consumer ...--no-skip-write
> ```
> 
> Does this result in an error? I think it should be fine since self-hosted e2e tests passed, but just wanted to make sure https://github.com/getsentry/self-hosted/blob/1c72fbe6120444ac232df0dae242cd5af40b38df/docker-compose.yml#L261

Hey sorry, we're getting rid of that `--no-skip-write flag`, because by default it skips writes. So that flag is a no-op

PR #5857, opened 2024-05-03
ref(http): Don't send X-ClickHouse-Key header when no password is configured

john-z-yang:
### Overview

When a user has no password configured, sending a request with empty `X-ClickHouse-Key` header causes a 400.

Maybe the `clickhouse_password` field should have been an `Option<&str>` instead of a `&str`, in case people have an empty password configured, but I would like this to be fixed now.


MeredithAnya:
Just adding some context - 400's don't happen in ClickHouse 22.8, so this is just affecting self-hosted folks (who should be able to upgrade soon) but yeah

MeredithAnya (approved):

john-z-yang:
This doesn't seem necessary anymore as new versions of clickhouse handles this gracefully

PR #5860, opened 2024-05-03
logs(http): monitor clickhouse streaming write channel size

john-z-yang:
We're seeing occasionally very high memory usage for some consumers, especially when there's a network problem. This PR monitors the size of the channel to see if this is the root cause.


rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -1,15 +1,17 @@
...
 use tokio::runtime::Handle;
-use tokio::sync::mpsc::{unbounded_channel, UnboundedSender};
+use tokio::sync::mpsc::{channel, Sender};
 use tokio::task::JoinHandle;
-use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_stream::wrappers::ReceiverStream;
 
 use crate::types::RowData;
 
-const CLICKHOUSE_HTTP_CHUNK_SIZE: usize = 1_000_000;
```
untitaker, line 12: why change the chunk size?

john-z-yang: Ooops sorry, forgot to change it back when I was testing locally

untitaker (approved):

PR #5861, opened 2024-05-03
ref: parse_mql_query_body unit tests

kylemumma:
This is the first PR in an attempt to refactor the tests for the API parsing pipeline, [described here](https://docs.google.com/presentation/d/1hfZSlRGktm3EZowd8BHR3h461t3r-ZxRR7LnSnPz0Rs/edit?usp=sharing).

## Major Changes
1. The main purpose of this PR is the introduction of unit tests for `parse_mql_query_body`, the initial conversion from mql to AST. They were autogenerated from the existing tests in `test_formula_mql_query` and `test_mql_query`. I manually verified that all tests were copied.
2. There are two optional changes I left comments for.

## Next Steps
I was able to successfully autogenerate unit tests for the first stage of the parsing pipeline, I will use a similar approach to autogenerate tests for the rest, which will be follow up PRs. This should be quicker than the first since I have the code written to grab the values. Finally, after all the unit tests are made, I will delete most of the end-to-end parse tests.

### DIFF
--- a/snuba/datasets/dataset.py
+++ b/snuba/datasets/dataset.py
@@ -40,3 +40,9 @@ class Dataset:
 
     def get_all_entities(self) -> Sequence[Entity]:
         return [get_entity(entity_key) for entity_key in self.__all_entities]
+
+    def __repr__(self) -> str:
+        # import down here bc of circular import
+        from snuba.datasets.factory import get_dataset_name
+
+        return f"get_dataset({repr(get_dataset_name(self))})"
--- a/snuba/query/dsl_mapper.py
+++ b/snuba/query/dsl_mapper.py
@@ -6,6 +6,7 @@ from snuba.query.conditions import (
     get_first_level_and_conditions,
     get_first_level_or_conditions,
 )
+from snuba.query.data_source.simple import Entity
 from snuba.query.expressions import (
     Argument,
     Column,
@@ -235,8 +236,15 @@ def query_repr(query: LogicalQuery | ClickhouseQuery) -> str:
     condition = ast_repr(query.get_condition(), visitor)
     groupby = ast_repr(query.get_groupby(), visitor)
 
+    qfrom = query.get_from_clause()
+    if isinstance(qfrom, Entity):
+        key = qfrom.key
+        from_clause = f"Entity({key},get_entity({key}).get_data_model())"
+    else:
+        from_clause = "from_clause"
+
     return f"""Query(
-        from_clause=from_clause,
+        from_clause={from_clause},
         selected_columns={selected},
         array_join={arrayjoin},
         condition={condition},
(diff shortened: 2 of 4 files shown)


snuba/datasets/dataset.py
kylemumma, line 1: this adds a functioning repr to dataset, repr used to just be an object pointer thingy but not its actually functional (it just relies on get_dataset)

snuba/query/dsl_mapper.py
kylemumma, line 1: The `query_repr` function used to have a placeholder `from_clause` for all queries. This will modify it for queries whose from_clause is an Entity, into:
```
Entity(
  EntityKey.GENERIC_METRICS_DISTRIBUTIONS,
  get_entity(EntityKey.GENERIC_METRICS_DISTRIBUTIONS).get_data_model()
)
```
which is functional, just relies on the get_entity method

volokluev (approved):
wow these tests are so much more readable now

snuba/datasets/dataset.py
```diff
@@ -40,3 +40,9 @@ def __init__(self, *, all_entities: Sequence[EntityKey]) -> None:
 
     def get_all_entities(self) -> Sequence[Entity]:
         return [get_entity(entity_key) for entity_key in self.__all_entities]
+
+    def __repr__(self) -> str:
+        # import down here bc of circular import
+        from snuba.datasets.factory import get_dataset_name
+
+        return f"get_dataset({repr(get_dataset_name(self))})"
```
evanh, line 48: Why isn't this `repr(self)`? Why wrap it in `get_dataset`?

tests/query/parser/unit_tests/test_parse_mql_query_body.py
```diff
@@ -0,0 +1,1013 @@
...
+                )
+            ],
+            array_join=None,
+            condition=and_cond(
+                f.equals(
+                    column("metric_id"), literal("d:transactions/duration@millisecond")
+                ),
+                in_cond(column("dist"), f.tuple(literal("dist1"), literal("dist2"))),
+            ),
+            groupby=None,
```
evanh, line 915: Can you get rid of all these default parameters? That would shave a lot of lines off.

snuba/datasets/dataset.py
kylemumma: repr(self) would return an object pointer since its not a dataclass.
I experimented with `f"Dataset(all_entities={repr(self.__all_entities)})"` because of what you said, but I think this would mess up the subclasses. Which is why I prefer the current change. A better solution could be to make it a dataclass so it has proper repr, but that seems like too much work.

tests/query/parser/unit_tests/test_parse_mql_query_body.py
kylemumma: Yeah, this would probably be best done in `query_repr`

PR #5868, opened 2024-05-06
ref: Deprecate TOPIC_PARTITION_COUNTS setting

lynnagara:
Snuba should fetch the actual value from the broker on startup rather than rely on a hardcoded TOPIC_PARTITION_COUNTS settings. This will ensure that it always has the correct values and remove the manual process and associated drift in ops that is currently involved in keeping this up to date.

Depends on https://github.com/getsentry/sentry/pull/70478

This is a prerequisite to https://github.com/getsentry/ops/pull/10392 and https://github.com/getsentry/ops/pull/10451 which will enable clean up of topicctl configuration in ops.


snuba/datasets/table_storage.py
```diff
@@ -57,13 +60,17 @@ def get_physical_topic_name(self, slice_id: Optional[int] = None) -> str:
...
     def partitions_number(self) -> int:
-        return settings.TOPIC_PARTITION_COUNTS.get(self.__topic.value, 1)
-
-    @property
-    def replication_factor(self) -> int:
-        return 1
+        config = get_default_kafka_configuration(self.__topic, None)
+        client = AdminClient(config)
+        topic_name = self.get_physical_topic_name()
+        print(topic_name)
```
fpacifici, line 68: ?

lynnagara: removed now

snuba/datasets/table_storage.py
```diff
@@ -57,13 +60,16 @@ def get_physical_topic_name(self, slice_id: Optional[int] = None) -> str:
...
 
-    @property
+    @cached_property
     def partitions_number(self) -> int:
-        return settings.TOPIC_PARTITION_COUNTS.get(self.__topic.value, 1)
-
-    @property
-    def replication_factor(self) -> int:
-        return 1
+        config = get_default_kafka_configuration(self.__topic, None)
```
evanh, line 65: This connection already happens in `bootstrap.py`. Is there a way we could leverage that code, to avoid connecting twice on startup?

Also, there is a utility function that does this: https://github.com/getsentry/snuba/blob/master/snuba/consumers/utils.py#L17

Could we use that instead?

evanh (commented):
Out of curiousity, if the partition check fails for some reason, what would happen

snuba/datasets/table_storage.py
lynnagara: Ugggh I just noticed a pretty bad bug with that other function. It does not properly resolve the topic name if it is overridden to something else and assumes the default. So I guess we can't safely use it right now, but also... this probably needs to be fixed?

lynnagara:
> Out of curiousity, if the partition check fails for some reason, what would happen

The pod would crash if Kafka or the topic is unavailable for some reason. This applies to the subscription scheduler pod as well as the API pod if the subscription endpoint is called.

snuba/datasets/table_storage.py
evanh: Sounds like it should be fixed. Did you want to add it to this PR?

evanh (approved):

enochtangg:
> This applies to the subscription scheduler pod as well as the API pod if the subscription POST endpoint is called

Is there any way around this so that the API pod doesn't crash? This kinda reminds me of the old readiness health-check we had before where snuba API would crash if it isn't able to connect to ClickHouse. Ideally, the API still works for all other the other datasets even if one partition check fails for a single dataset.

lynnagara:
Discussed offline. Looks like exception handling and returning a proper response is already automatically handled by https://github.com/getsentry/snuba/blob/177e23eca6475d21c9156571339a336e2a518c29/snuba/web/views.py#L173-L194

lynnagara:
also needs https://github.com/getsentry/sentry/pull/70809

lynnagara:
sentry test is failing

PR #5869, opened 2024-05-06
fix(cd): deploy fails from ci timeout

kylemumma:
this bug was meant to be fixed by #5834 but wasnt actually, this PR should actually fix it now.

### DIFF
--- a/gocd/templates/bash/check-github.sh
+++ b/gocd/templates/bash/check-github.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 /devinfra/scripts/checks/githubactions/checkruns.py \
+  --timeout-mins 60 \
   getsentry/snuba \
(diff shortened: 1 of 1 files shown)


IanWoodard (approved):

john-z-yang (approved):

gocd/templates/bash/check-github.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 /devinfra/scripts/checks/githubactions/checkruns.py \
+  --timeout-mins 60 \
```
john-z-yang, line 4: What was the timeout before this?

IanWoodard: 30 minutes (the default value)

kylemumma: 30 i think

PR #5873, opened 2024-05-07
snuba devserver fix?

MeredithAnya:
We've been seeing some issues with running `snuba devserver` locally and the problem seems to happen when there aren't actually any messages being processed but we still send requests to clickhouse (with `0` as the body).

This change should prevent a request to clickhouse from ever being made if we don't have any rows to write. If the sender doesn't send any messages then the receiver will be empty - which is what we use to check in order to determine when to open the request.

### DIFF
--- a/rust_snuba/src/strategies/clickhouse/batch.rs
+++ b/rust_snuba/src/strategies/clickhouse/batch.rs
@@ -81,15 +81,17 @@ impl BatchFactory {
             let client = self.client.clone();
 
             let result_handle = self.handle.spawn(async move {
-                let res = client
-                    .post(&url)
-                    .query(&[("query", &query)])
-                    .body(reqwest::Body::wrap_stream(ReceiverStream::new(receiver)))
-                    .send()
-                    .await?;
+                if !receiver.is_empty() && !receiver.is_closed() {
+                    let res = client
+                        .post(&url)
+                        .query(&[("query", &query)])
+                        .body(reqwest::Body::wrap_stream(ReceiverStream::new(receiver)))
+                        .send()
+                        .await?;
 
-                if res.status() != reqwest::StatusCode::OK {
-                    anyhow::bail!("error writing to clickhouse: {}", res.text().await?);
+                    if res.status() != reqwest::StatusCode::OK {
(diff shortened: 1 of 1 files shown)


untitaker (approved):

rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -79,17 +80,33 @@ impl BatchFactory {
...
+                    sleep(Duration::from_millis(800)).await;
+                }
+
+                if receiver.is_closed() && receiver.is_empty() {
+                    // batch is finished and we never had any data to write
+                    // to clickhouse
+                    return Ok(());
+                }
+
+                if !receiver.is_empty() {
```
untitaker, line 95: is this branch necessary, can't it run unconditionally?

rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -79,17 +80,33 @@ impl BatchFactory {
...
-
-                if res.status() != reqwest::StatusCode::OK {
-                    anyhow::bail!("error writing to clickhouse: {}", res.text().await?);
+                while receiver.is_empty() && !receiver.is_closed() {
+                    // continously check on the receiver stream, only when it's
+                    // not empty do we write to clickhouse
+                    sleep(Duration::from_millis(800)).await;
+                }
+
+                if receiver.is_closed() && receiver.is_empty() {
```
untitaker, line 89: I think `if receiver.is_empty()` alone should suffice, because if it is empty but not closed, we should not have exited the while loop at all. but i might be mistaken

rust_snuba/src/strategies/clickhouse/batch.rs
MeredithAnya: leaving this as is since if it's closed and empty we don't need to send a request to clickhosue

volokluev (approved):

john-z-yang:
This is working for me, thanks @MeredithAnya!

Do we know why is only needed when running `devserver` but when we run the consumer by itself (with rust-consumer) it seems to work?

PR #5875, opened 2024-05-07
fix: Handle tag columns in low cardinality processor

evanh:
Discover will sometimes send filters on columns as tags instead of actual
columns, e.g. tags[environment]. This was causing the allocator bug to be hit
again on some queries. This change handles checking for this types of queries
and correctly casting them as well to avoid the allocator bugs.


volokluev:
Is this happening because the tag keys/values are being promoted to columns somewhere down the line? Can we fix this by putting the processor further down?

Also thank you for writing tests for the processor 🫶

evanh:
> Is this happening because the tag keys/values are being promoted to columns somewhere down the line

Yes, in the SQL query it uses the `environment` column: 
```
(environment AS `_snuba_tags[environment]`), (count() AS _snuba_count), _snuba_count FROM discover_dist PREWHERE in(`_snuba_tags[environment]`, ['cron', 'worker'])
```

However I believe that happens in the storage mapping, so I would have to rewrite the processor as a storage processor. I think this is a cleaner workaround for now.

volokluev (approved):

PR #5878, opened 2024-05-08
perf(ci): Build snuba image only once in CI

john-z-yang:
### Overview

The snuba GHA that runs tests seems to be waiting on a GHA that builds a snuba image. But for some reason those tests also rebuilds the image again. This PR makes some changes to the test jobs definition to reuse the image that previous steps create.
The PR also includes some commits where I upload bad tests to make sure the CI is indeed using the right image.

This appears to have lowered the overall CI time from ~31 minutes to ~22 minutes.
When re-running tests jobs due to flakes, each test jobs should be at least 8 minutes quicker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -276,15 +276,6 @@ jobs:
         run: |
           echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u $GITHUB_ACTOR --password-stdin
 
-      - name: Pull the image
-        env:
-          IMAGE_URL: ghcr.io/getsentry/snuba-ci:${{ github.sha }}
-        shell: bash
-        run: |
-          echo "We poll for the Docker image that the GCB/GHA build produces until it succeeds or this job times out."
-          echo "Polling for $IMAGE_URL"
-          timeout 20m bash -c 'until docker pull "$IMAGE_URL" 2>/dev/null; do sleep 10; done'
-
       - name: Checkout sentry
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -199,24 +199,18 @@ jobs:
...
-          tags: snuba-test
-          load: true
-          target: testing
-          cache-from: |
-            type=registry,ref=ghcr.io/getsentry/snuba-ci:${{ github.sha }}
-            type=registry,ref=ghcr.io/getsentry/snuba-ci:${{ needs.snuba-image.outputs.branch }}
-            type=registry,ref=ghcr.io/getsentry/snuba-ci:latest
+      - name: Pull the image
+        env:
+          IMAGE_URL: us-central1-docker.pkg.dev/sentryio/snuba/image:${{ github.event.pull_request.head.sha || github.sha }}
```
hubertdeng123, line 208: I think you may need to use `ghcr.io/getsentry/snuba-ci:${{ github.sha }}` instead, that seems to be where the image is being pushed in step `snuba-image`

john-z-yang:
Failed (as hoped)
image

john-z-yang:
Failed (as hoped)
image

.github/workflows/ci.yml
john-z-yang: Thanks, fixed!

volokluev (approved):
Well done

hubertdeng123 (approved):
looks good

PR #5870, opened 2024-05-07
feat(devexp): Parse and run storage queries

volokluev:
This PR implements querying storages directly at the snuba level. After it is merged, it can be used immediately to query any storage

### Change Summary
* Add a new class `LogicalDataSource` which can be either a Storage or an Entity
    * I wanted to make the parser product clickhouse queries directly but there is too much stuff unrelated to entities that happens in the parser so I kept them separate
    * Eventually I'd actually like to remove `ClickhouseQuery` from the codebase but not in this PR
* Change the snql parser and grammar to be able to parse storage queries directly
* Change the parsing portion of the code to be agnostic to Entity/Storage

### Outstanding tasks

- [x] write more tests for the entity processing stage functionality
- [x] Remove the useless EntityColumnSet class

PR #5879, opened 2024-05-08
ref(devexp): Make entities optional in dataset config

enochtangg:
Now that queries can be sent to the storage directly, creating an entity component alongside a dataset is optional. This PR is responsible for changing jsonschema to remove `entities` as a required field in the dataset definition.

Depends on: 
* https://github.com/getsentry/snuba/pull/5870


volokluev:
With querying storages directly we don't need datasets either so maybe it's better to just make a default dataset that can be used to query any storage/entity?

enochtangg:
>With querying storages directly we don't need datasets either so maybe it's better to just make a default dataset that can be used to query any storage/entity?

One thing I was thinking about making entities optional was "what do we need to do to get rid the dataset component entirely?". Queries are evaluated based on the entity and now storage anyways. However, I think there's a decent amount of product logic in the sentry repo that depend on the dataset. 

I like your idea though it'd take a product developer a few more steps if they decide to add an entity later on.

snuba/utils/health_info.py
```diff
@@ -122,15 +121,13 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
...
-            if storage.get_readiness_state().value in settings.SUPPORTED_STATES:
-                storages.append(storage)
-    return storages
+    filtered_storages: List[Storage] = []
+    all_storage_keys = get_all_storage_keys()
+    for storage_key in all_storage_keys:
+        storage = get_storage(storage_key)
+        if storage.get_readiness_state().value in settings.SUPPORTED_STATES:
+            filtered_storages.append(storage)
+    return filtered_storages
```
kylemumma, line 130: so it used to be doing: datasets -> entities -> storages
but now since not all storages are attached to entity, you changed it to use "get_all_storage_keys"?

kylemumma (commented):
Nice, this is what enables defining a storage-only dataset?

snuba/utils/health_info.py
enochtangg: That's correct. This health-check ensures that Snuba is able to access each table in ClickHouse via the storage definitions. Now that an entities are optional, we should just `get_all_storage_keys` directly rather than using the entities to obtain the storages.

kylemumma (approved):
LGTM but maybe Volo should also read it since he's working closely on this as well.... idk

PR #5882, opened 2024-05-08
sessions: drop the clickhouse tables

lynnagara:


volokluev (approved):

snuba/snuba_migrations/sessions/0002_sessions_aggregates.py
```diff
@@ -112,14 +104,6 @@ def forwards_dist(self) -> Sequence[operations.SqlOperation]:
...
-            for [column, after] in new_raw_columns
-        ] + [
-            operations.DropColumn(
-                StorageSetKey.SESSIONS, "sessions_hourly_dist", column.name
-            )
-            for [column, after] in new_dest_columns
-        ]
+        # Sessions is fully deprecated, we don't need to keep supporting this reverse migration
+        # as providing it makes test_reverse_idempotency_all fail
+        return []
```
lynnagara, line 109: I hope this is ok, I don't think we need to worry about this anymore since sessions is dead, and it makes the reverse idempotency test fail as later migrations haven't defined all the reverse steps.

klboke:
@lynnagara After sessions are deleted, is there any way to migrate the existing session data to Metrics? I haven't seen such an operation. The expectation is to allow users to smoothly transition to Metrics. Is there any documentation or resources on this?

PR #5883, opened 2024-05-09
automatically skip replacements for pre-determined projects

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2712

This PR:
- creates a Replacements tab on Snuba Admin to view the skipped projects and their expiry after which they will be processed again
- allows expiry to be configurable via runtime config
- allows the processing time range to be configurable via runtime config (this is mostly for testing on S4S)

Screenshot 2024-05-16 at 10 34 50 AM

### DIFF
--- a/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
+++ b/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
@@ -1,8 +1,5 @@
 import React, { useState, useEffect } from "react";
-import {
-  AutoReplacementsBypassProjectsData,
-  ExpiryWindow,
-} from "SnubaAdmin/auto_replacements_bypass_projects/types";
+import { AutoReplacementsBypassProjectsData } from "SnubaAdmin/auto_replacements_bypass_projects/types";
 
 import Client from "SnubaAdmin/api_client";
 
@@ -11,35 +8,13 @@ function AutoReplacementsBypassProjects(props: { api: Client }) {
     null
   );
 
-  const [expiryWindow, setExpiryWindow] = useState<ExpiryWindow | null>(null);
-
   useEffect(() => {
     props.api.getAutoReplacementsBypassProjects().then((res) => {
       setData(res);
     });
   }, []);
 
-  useEffect(() => {
-    props.api.getExpiryWindow().then((res) => {
-      setExpiryWindow(res);
-    });
-  }, []);
-
-  useEffect(() => {
-    props.api.getExpiryWindow().then((res) => {
-      setExpiryWindow(res);
-    });
-  }, []);
-
-  return (
-    <div>
-      {JSON.stringify(data)}
-      <div>
-        <h2>Expiration window:</h2>
-        <div>{JSON.stringify(expiryWindow)}</div>
-      </div>
-    </div>
-  );
+  return <div>{JSON.stringify(data)}</div>;
 }
 
 export default AutoReplacementsBypassProjects;
--- a/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
+++ b/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
@@ -1,0 +1,20 @@
+import React, { useState, useEffect } from "react";
+import { AutoReplacementsBypassProjectsData } from "SnubaAdmin/auto_replacements_bypass_projects/types";
+
+import Client from "SnubaAdmin/api_client";
+
+function AutoReplacementsBypassProjects(props: { api: Client }) {
+  const [data, setData] = useState<AutoReplacementsBypassProjectsData[] | null>(
+    null
+  );
+
+  useEffect(() => {
+    props.api.getAutoReplacementsBypassProjects().then((res) => {
+      setData(res);
+    });
+  }, []);
+
+  return <div>{JSON.stringify(data)}</div>;
+}
+
+export default AutoReplacementsBypassProjects;
--- a/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
+++ b/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
@@ -1,5 +1,8 @@
 import React, { useState, useEffect } from "react";
-import { AutoReplacementsBypassProjectsData } from "SnubaAdmin/auto_replacements_bypass_projects/types";
+import {
+  AutoReplacementsBypassProjectsData,
+  ExpiryWindow,
+} from "SnubaAdmin/auto_replacements_bypass_projects/types";
 
 import Client from "SnubaAdmin/api_client";
 
@@ -8,13 +11,35 @@ function AutoReplacementsBypassProjects(props: { api: Client }) {
     null
   );
 
+  const [expiryWindow, setExpiryWindow] = useState<ExpiryWindow | null>(null);
+
   useEffect(() => {
     props.api.getAutoReplacementsBypassProjects().then((res) => {
       setData(res);
     });
   }, []);
 
-  return <div>{JSON.stringify(data)}</div>;
+  useEffect(() => {
+    props.api.getExpiryWindow().then((res) => {
+      setExpiryWindow(res);
+    });
+  }, []);
+
+  useEffect(() => {
+    props.api.getExpiryWindow().then((res) => {
+      setExpiryWindow(res);
+    });
+  }, []);
+
+  return (
+    <div>
+      {JSON.stringify(data)}
+      <div>
+        <h2>Expiration window:</h2>
+        <div>{JSON.stringify(expiryWindow)}</div>
+      </div>
+    </div>
+  );
 }
 
 export default AutoReplacementsBypassProjects;
--- a/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
+++ b/snuba/admin/static/auto_replacements_bypass_projects/index.tsx
@@ -2,6 +2,7 @@ import React, { useState, useEffect } from "react";
 import { AutoReplacementsBypassProjectsData } from "SnubaAdmin/auto_replacements_bypass_projects/types";
 
 import Client from "SnubaAdmin/api_client";
+import { Table } from "../table";
 
 function AutoReplacementsBypassProjects(props: { api: Client }) {
   const [data, setData] = useState<AutoReplacementsBypassProjectsData[] | null>(
@@ -14,7 +15,24 @@ function AutoReplacementsBypassProjects(props: { api: Client }) {
     });
   }, []);
 
-  return <div>{JSON.stringify(data)}</div>;
+  if (!data) {
+    return null;
+  }
+
+  const rowData = data.map(({ projectID, expiry }) => {
+    return [projectID, expiry];
+  });
+
+  return (
+    <div>
+      <div>
+        <Table
+          headerData={["Project ID", "Expiration Time"]}
+          rowData={rowData}
+        />
+      </div>
+    </div>
+  );
 }
 
 export default AutoReplacementsBypassProjects;
--- a/snuba/replacer.py
+++ b/snuba/replacer.py
@@ -28,7 +28,7 @@ from arroyo.processing.strategies.abstract import (
 )
 from arroyo.types import BrokerValue, Commit, Message, Partition
 
-from snuba import settings, state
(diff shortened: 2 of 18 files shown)


snuba/state/__init__.py
```diff
@@ -315,6 +323,44 @@ def get_all_config_descriptions() -> Mapping[str, Optional[str]]:
         return {}
 
 
+def set_config_auto_replacements_bypass_projects(
```
volokluev, line 326: This functionality is for replacements, it should live with replacements

snuba/replacer.py
```diff
@@ -589,6 +591,12 @@ def _attempt_emitting_metric_for_projects_exceeding_limit(
         projects_exceeding_limit = (
             self.__processing_time_counter.get_projects_exceeding_limit()
         )
+        # this is for testing purely on S4S, will delete immediately after testing
+        if 2 not in projects_exceeding_limit and get_config(
```
xurui-c, line 595: It’s hard to determine how many replacement operations I have to manually perform to reach the threshold. One way to get around this is to always add S4S (project id = 2) to the list of offending projects, so effectively skipping all replacements for S4S. After verifying this works, I’ll have to immediately merge a PR that undoes this. What do people think?

snuba/utils/uwsgi.py
```diff
@@ -65,6 +65,7 @@ def run(module: str, bind: str, **kwargs: bool | int | str | None) -> NoReturn:
         "vacuum": True,
         "virtualenv": sys.prefix,
         "wsgi_env_behavior": "holy",
+        "py_autoreload": 1,
```
xurui-c, line 68: I'll move this to a different PR after we're done with code review

snuba/replacers/replacements_utils.py
```diff
@@ -0,0 +1,62 @@
...
+logger = logging.getLogger(__name__)
+
+
+from snuba.redis import RedisClientKey, get_redis_client
+
+redis_client = get_redis_client(RedisClientKey.REPLACEMENTS_STORE)
+config_auto_replacements_bypass_projects_hash = (
+    "snuba-config-auto-replacements-bypass-projects-hash"
+)
+EXPIRY_WINDOW = timedelta(minutes=5)
```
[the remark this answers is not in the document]
volokluev: Also it's easier if you store the number itself as a constant so

```
EXPIRY_WINDOW_MINUTES = 5


# in your code
timedelta(minutes=EXPIRY_WINDOW_MINUTES)
```

This makes it possible to set these variables via runtime config and also use this number in tests easier

tests/state/test_replacements_bypass_projects.py
```diff
@@ -0,0 +1,70 @@
...
+    get_config_auto_replacements_bypass_projects,
+    set_config_auto_replacements_bypass_projects,
+)
+
+
+@freeze_time("2024-5-13 09:00:00")
+class TestState:
+    start_test_time = datetime.now()
+    proj1_add_time = start_test_time
+    proj2_add_time = start_test_time + timedelta(minutes=2)
```
volokluev, line 17:
```suggestion
    proj2_add_time = start_test_time + timedelta(minutes=EXPIRY_WINDOW_MINUTES // 2)
```

The reader of this file shouldn't have to *know* that the EXPIRY_WINDOW is 5 minutes long and that 2 is less than 5. Just one project expires sooner than the other

volokluev (changes requested):
Looking much better, a few more things though

snuba/replacer.py
xurui-c: We decided to make `COUNTER_WINDOW_SIZE` configurable through the runtime config, so by setting `counter_window_size_minutes` = 0 in the S4S runtime config, we are effectively skipping every replacements project in S4S only

snuba/settings/__init__.py
```diff
@@ -381,8 +381,8 @@ class RedisClusters(TypedDict):
 ENTITY_CONFIG_FILES_GLOB = f"{CONFIG_FILES_PATH}/**/entities/*.yaml"
 DATASET_CONFIG_FILES_GLOB = f"{CONFIG_FILES_PATH}/**/dataset.yaml"
 
-# Counter utility class window size in minutes
-COUNTER_WINDOW_SIZE_MINUTES = 10
+# Counter utility class window size in minutes hash key
+COUNTER_WINDOW_SIZE_MINUTES_KEY = "counter_window_size_minutes"
```
volokluev, line 385: I don't think this needs to be in settings, are you planning on changing it per environment?

volokluev (commented):
Good overall, just a few more things

snuba/admin/static/auto_replacements_bypass_projects/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+    null
+  );
+
+  useEffect(() => {
+    props.api.getAutoReplacementsBypassProjects().then((res) => {
+      setData(res);
+    });
+  }, []);
+
+  return <div>{JSON.stringify(data)}</div>;
```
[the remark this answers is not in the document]
xurui-c: included - you're right it looks much better and more consistent

snuba/replacers/errors_replacer.py
```diff
@@ -203,8 +206,13 @@ def process_message(
...
         if processed is not None:
-            bypass_projects = get_config("replacements_bypass_projects", "[]")
-            projects = json.loads(cast(str, bypass_projects))
+            manual_bypass_projects = get_config("replacements_bypass_projects", "[]")
+            auto_bypass_projects = list(
+                get_config_auto_replacements_bypass_projects(datetime.now())
+            )
+            if manual_bypass_projects is not None:
+                auto_bypass_projects.extend(manual_bypass_projects)
+            projects = json.loads(cast(str, auto_bypass_projects))
```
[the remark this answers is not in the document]
xurui-c: I made an update to this part of the code - I didn't realize `replacements_bypass_projects` is passed in as a string even though it's a list, and I cleaned it up a little more.

> what happens if someone enters invalid json into the runtime config?
Good point - ideally the safety checks for the manual `replacements_bypass_projects` should've been included in its original PR, so I included a try block for now. This PR intends to use the existing logic for manual projects retrieval, so I think if we want to add validity checks during manual projects retrieval, I should do it in a separate PR. 

> also isn't auto_bypass_projects already a list? Do you mean manual_bypass_projects?
No, we need to combine auto and manual projects because all of them need to be skipped. I agree the naming before was a little confusing though, so I made the it clearer.

snuba/replacers/replacements_and_expiry.py
```diff
@@ -0,0 +1,76 @@
...
+
+redis_client = get_redis_client(RedisClientKey.REPLACEMENTS_STORE)
+config_auto_replacements_bypass_projects_hash = (
+    "snuba-config-auto-replacements-bypass-projects-hash"
+)
+
+REPLACEMENTS_EXPIRY_WINDOW_MINUTES_KEY = "replacements_expiry_window_minutes"
+
+
+def get_minutes(key: str, default: int) -> int:
```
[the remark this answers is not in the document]
xurui-c: > this function is not descriptively named. What are the minutes for?

We can either get the expiry minutes (ie project should be skipped for 5 minutes) or the counter window (which is the time window we use to calculate the threshold for determining if we should skip a project). I renamed it to `get_expiry_window_or_counter_window_size`. The main purpose of this function, though, is to get around the type checking, since in our case the default value (5 or 10) should be returned instead of `None` anyways.

> Also is there ever a time you would have the arguments be anything other than:
`get_minutes(REPLACEMENTS_EXPIRY_WINDOW_MINUTES_KEY, 5)`

Yes: https://github.com/getsentry/snuba/pull/5883/files#diff-44863b6f784105daa8ce60ea6f4c8db7cb46e122dec3e3c0c59d8ec3d96b1fbaR30

snuba/settings/__init__.py
xurui-c: No, I moved it

snuba/replacers/replacements_utils.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
xurui-c: Named it to replacements_and_expiry.py

snuba/replacers/errors_replacer.py
```diff
@@ -203,9 +207,17 @@ def process_message(
...
-            projects = json.loads(cast(str, bypass_projects))
-            if processed.get_project_id() in projects:
+            manual_bypass_projects = get_config("replacements_bypass_projects", "[]")
+            auto_bypass_projects = list(
+                get_config_auto_replacements_bypass_projects(datetime.now()).keys()
+            )
+            projects_to_skip = auto_bypass_projects
+            if manual_bypass_projects is not None:
+                try:
+                    projects_to_skip.extend(ast.literal_eval(manual_bypass_projects))
```
volokluev, line 217: You basically never want to `ast.literal_eval` code anywhere. `json.loads` will do just fine here

snuba/replacers/replacements_and_expiry.py
```diff
@@ -0,0 +1,78 @@
...
+config_auto_replacements_bypass_projects_hash = (
+    "snuba-config-auto-replacements-bypass-projects-hash"
+)
+
+REPLACEMENTS_EXPIRY_WINDOW_MINUTES_KEY = "replacements_expiry_window_minutes"
+
+
+def get_expiry_window_or_counter_window_size(key: str, default: int) -> int:
+    minutes = get_int_config(key=key, default=default)
+    return int(minutes) if minutes else default
```
volokluev, line 23: This function is not really abstracting much, it's also confusing because you use it within and outside this module for different reasons. It would make more sense to simply get this config where you actually need it rather than define a function which does not simplify anything

snuba/replacers/replacements_and_expiry.py
```diff
@@ -0,0 +1,78 @@
...
+
+    valid_projects = {}
+    pipeline = redis_client.pipeline()
+    for project_id in curr_projects:
+        if curr_projects[project_id] < curr_time:
+            pipeline.hdel(config_auto_replacements_bypass_projects_hash, project_id)
+        else:
+            valid_projects[project_id] = pipeline.hget(
+                config_auto_replacements_bypass_projects_hash, project_id
+            )
```
volokluev, line 75: You don't need to do this. You already have this data in the `curr_projects` hashmap, just add it to `valid_projects` and move on. I'm also not sure if the redis pipeline works this way. You won't get the data back until you actually execute

volokluev (commented):
So close! The security risk and the unnecessary `hget` are my main concerns

snuba/replacers/replacements_and_expiry.py
xurui-c: Makes sense. The other way I can think of currently to get around the type checking is using `assert` statements. Let me know what you think

volokluev: you can just use `int(get_int_config(key, 20))` if you're passing an int as a default to `get_int_config` there's no way you're not getting back an int

volokluev (approved):

tests/replacer/test_replacements_and_expiry.py
```diff
@@ -13,7 +14,7 @@
 
 
 @freeze_time("2024-5-13 09:00:00")
-class TestState:
+class TestState(unittest.TestCase):
```
volokluev, line 17: you don't need this

snuba/replacers/replacements_and_expiry.py
xurui-c: I agree that there's no way we get back `None` in practice, however directly casting it to `int` without either `assert` or the helper function approach results in this error: https://github.com/getsentry/snuba/actions/runs/9134800036/job/25121068644?pr=5883

tests/replacer/test_replacements_and_expiry.py
xurui-c: Sorry, I left it in when trying to fix the mock error

snuba/replacers/replacements_and_expiry.py
volokluev: then you can use `typing.cast` or `# type: ignore`

PR #5885, opened 2024-05-09
ref(api): parse pipeline tests, mql pipeline before treeify

kylemumma:
## Major Changes
* separate everything that comes before treeify in mql parse pipeline into its own stage
* add tests for this stage
* misc other things

PR #5886, opened 2024-05-09
ref(api): parse pipeline tests, snql pipeline before treeify

kylemumma:
## Major Changes
this is part 2, now all the "before treeify" is separate. depends on https://github.com/getsentry/snuba/pull/5885

* write tests for "before treeify in snql parse pipeline." No need to put into its own stage bc its just 1 method.
* small dsl and dsl mapper changes (and some random tests had to change bc of it)

I verified by hand that all the test cases from e2e are present.

## Next steps
* mql after treeify
* snql after treeify
* treeify unit tests
* delete e2e tests


kylemumma (commented):
for the diff you will have to shift+click changes from only 3 the commits after "rename." thats bc this one is dependent on #5885

snuba/query/dsl.py
kylemumma, line 1: previously you could provide more than 2 conditions to and it it would return:
and(c, and(c, and(c, c))) but this was causing issues. so now it will return a non-treeified and i.e. and and with more than 2 parameters.

tests/query/parser/unit_tests/test_parse_snql_query_initial.py
```diff
@@ -0,0 +1,971 @@
...
+            order_by=None,
+            limitby=None,
+            limit=1000,
+            offset=0,
+            totals=False,
+            granularity=None,
+        ),
+    ),
+    pytest.param(
+        "\n           MATCH (events)\n           SELECT test_func(release) AS test_func_alias,\n              event_id BY project_id, platform\n           WHERE timestamp >= toDateTime('2021-01-01T00:00:00') AND timestamp < toDateTime('2021-01-02T00:00:00') AND project_id = 1\n        ",
```
volokluev, line 40: Can you fix this up?

snuba/query/dsl.py
volokluev: * What are the issues it was causing?
* the docstring is now out of date.
* `and_cond` is now functionally equivalent to `f.and` so maybe it makes sense to just get rid of `and_cond` and `or_cond`?

volokluev (commented):
Just some questions and a request to fix up the test formatting.

snuba/query/dsl.py
kylemumma: The issue was:
Say you have FunctionCall(None, "and", (c1, c2, c3, c4, c5))
dsl_mapper turns it into and_cond(c1,c2,c3,c4,c5)
which aren't equal. 
An alternative approach would be to has dslmapper return FunctionCall rather than and_cond, but I had the opinion that and_cond should behave in a consistent well defined manner (as to replace the constructor) and I thought the treeify part was a little too. "black box" which is why I made this decision

I will update the docstring now

I agree with removing and_cond and or_cond in favor of f.and. But I think that they are used quite often and it would take some amount of time to make the change. My current focus is to get the bug fix out ASAP (bc i think matt is mad at me) which is why I would prefer to leave this refactoring for another time.

tests/query/parser/unit_tests/test_parse_snql_query_initial.py
kylemumma: do i have to 😢

volokluev: yes, there are not that many. A simple:

`s/\\n           / /g` should do

volokluev (approved):

tests/query/parser/unit_tests/test_parse_snql_query_initial.py
kylemumma: ok, you know your regex, very useful skill

PR #5887, opened 2024-05-09
fix(capman): make allocation policies work with joins

volokluev:
Queries using joins have up to this point not been subject to allocation policies. making them an abuse vector. Now that we have join queries serving in production, put the protection on them

### DIFF
--- a/tests/datasets/test_cdc_events.py
+++ b/tests/datasets/test_cdc_events.py
@@ -130,10 +130,12 @@ class TestCdcEvents(BaseApiTest):
                         "btime": self.base_time,
                         "ntime": self.next_time,
                     },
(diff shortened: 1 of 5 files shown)


tests/datasets/test_cdc_events.py
```diff
@@ -130,11 +130,13 @@ def test_groups_join(
                         "btime": self.base_time,
                         "ntime": self.next_time,
                     },
+                    "tenant_ids": {"organization_id": 1, "referrer": "abcd"},
```
nikhars, line 133: Not specific to your PR. But at some point of time we should probably just get rid of all CDC related code since it is unused.

nikhars (approved):

PR #5888, opened 2024-05-10
fix: Disallow column aliases in inner queries of subqueries

evanh:
When a column is aliased to something different in the inner subquery, it
breaks how aliasing works at the higher levels. It's also not necessary, since
the column can be referenced directly, and the outer query can alias the
expression however it likes in the results.

Also fixed some typing in the test file.


volokluev (approved):

evanh:
Out of date

PR #5891, opened 2024-05-10
chore(self-hosted): Remove dependency on snuba-image build step during e2e action

hubertdeng123:
Don't think this is needed, it just needs to wait on `build-on-snuba-pr` image to finish job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -387,7 +387,6 @@ jobs:
           curl -Os https://uploader.codecov.io/latest/linux/codecov && chmod +x codecov && ./codecov -t ${CODECOV_TOKEN}
 
   self-hosted-end-to-end:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -387,9 +387,8 @@ jobs:
           curl -Os https://uploader.codecov.io/latest/linux/codecov && chmod +x codecov && ./codecov -t ${CODECOV_TOKEN}
 
   self-hosted-end-to-end:
-    needs: [snuba-image]
     runs-on: ubuntu-latest
-    timeout-minutes: 25
```
hubertdeng123, line 392: I think we need to bump this timeout, since it takes a while to build the image

john-z-yang (approved):

PR #5895, opened 2024-05-12
feat(profiles): Add a dataset to store profile chunks

phacops:
This will add a new dataset for profile chunks in order to support our continuous profiling feature (receiving profile chunk by chunk and not one profile per transaction).

The SDK will start a profiler session, identified by `profiler_id`, will profile and send a chunk with containing this profile ID. It will also tag spans collected with this profiler ID.

Later on, to fetch a profile for a span, we'll receive a profile ID and start and stop timestamps for the span and, using this dataset, we'll query the chunk IDs necessary to assemble the profile for that span, with a query looking like this:
```sql
SELECT
    DISTINCT chunk_id
FROM profile_chunks
WHERE
    project_id = <project_id>
    AND profiler_id = <profiler_id>
    AND start_timestamp <= <span.start_timestamp>
    AND end_timestamp >= <span.end_timestamp>
ORDER BY start_timestamp;
```

Since all our queries will contain a date range on both `start_timestamp` and `end_timestamp`, I think it's useful to have them in the sort key. `chunk_id` appears there in order to make it work with a `ReplacingMergeTree` since this will guaranteed uniqueness of rows, even though having 2 chunks for the same profiler session and timestamps would be a bug.

We're using the `DateTime64` type to be able to store sub-millisecond precision and now have 2 different fields. I add support for that field in https://github.com/getsentry/snuba/pull/5896.

This PR (https://github.com/getsentry/ops/pull/10526) is related as it adds the necessary config for `StorageSetKey.PROFILE_CHUNKS` to the profiling cluster in every environment.


snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,85 @@
...
+        "end_timestamp",
+        DateTime64(
+            precision=6,
+            modifiers=Modifiers(codecs=["DoubleDelta"]),
+        ),
+    ),
+    Column("retention_days", UInt(16)),
+    Column("partition", UInt(16)),
+    Column("offset", UInt(64)),
+    Column("deleted", UInt(8)),
```
viglia, line 37: Q.Q.: what's `deleted` for?

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,85 @@
...
+
+UNKNOWN_SPAN_STATUS = 2
+
+columns: List[Column[Modifiers]] = [
+    Column("project_id", UInt(64)),
+    Column("profiler_id", UUID()),
+    Column("chunk_id", UUID()),
+    Column(
+        "start_timestamp",
+        DateTime64(
```
Zylphrex, line 22: Nice, no more adding the seconds and milliseconds portions together.

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,85 @@
...
+table_prefix = "profile_chunks"
+local_table_name = f"{table_prefix}_local"
+dist_table_name = f"{table_prefix}_dist"
+
+UNKNOWN_SPAN_STATUS = 2
+
+columns: List[Column[Modifiers]] = [
+    Column("project_id", UInt(64)),
+    Column("profiler_id", UUID()),
+    Column("chunk_id", UUID()),
```
Zylphrex, line 19: In relay, these 2 fields are typed as `String` with a comment that it should be UUIDs. Should we go back and validate that to ensure non UUID values don't make it here?

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,85 @@
...
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, cityHash64(profiler_id), start_timestamp, end_timestamp, cityHash64(chunk_id))",
```
Zylphrex, line 51: Will we have a processor that maps `profiler_id = ...` conditions into `cityHash64(profiler_id) = cityHash64(...)`? Without it, I believe queries won't fully take advantage of the order by for max performance.

Zylphrex (approved):
This looks good to me 👍

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,85 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import UUID, Column, DateTime64, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+storage_set = StorageSetKey.PROFILE_CHUNKS
```
volokluev, line 9: Are you putting this on the profiles cluster or a new one? If you're using the profiles one, you can use the PROFILES storage_set_key

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
volokluev: What is the purpose of hashing the profiler id?

volokluev: I'm also wondering if the `end_timestamp` is necessary? I can't imagine profiles being so far apart that you need both.

volokluev (commented):
A few questions

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,85 @@
...
+table_prefix = "profile_chunks"
+local_table_name = f"{table_prefix}_local"
+dist_table_name = f"{table_prefix}_dist"
+
+UNKNOWN_SPAN_STATUS = 2
```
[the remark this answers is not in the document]
phacops: Definitely unused.

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: For data retention reason. If a customer wants us to delete its data, we mark it as deleted.

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: I think we'll update the fields to be proper UUIDs so I'll keep the UUID columns here.

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: The purpose of hashing the `profiler_id` is for the sampling key (it needs to be an integer) but @Zylphrex and I already agreed it `chunk_id` might be a better sampling key anyway. I'll update to that and remove the hashing of `profiler_id`.

`end_timestamp` is not necessarily needed, I suppose I wanted to maximize the use of the index but maybe that's trying to do too much.

volokluev: I would leave end_timestamp out of it. The order_by key is used to whittle down how many parts you scan and it's likely you're not going to cut down on any parts. You'll just make the sorting key bigger. Tbh I don't know what the negative impact of that will be but I don't see a positive one either. 

If you're interested in an explanation of query processing:
https://docs.google.com/presentation/d/1-pwlFWx05sPqkv6WYn4TQZFWfLzw1wAoE8eykq_aL2s/edit#slide=id.g229f20e6452_0_892

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: https://github.com/getsentry/relay/pull/3588

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: Is it actually true?
```
snuba.migrations.groups.DuplicateStorageSetFoundInGroup: The storage_set=StorageSetKey.PROFILES is present in more than one MigrationGroup. Each storage_set should only reference one migration group.
```

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
```diff
@@ -0,0 +1,83 @@
...
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    order_by="(project_id, profiler_id, start_timestamp, end_timestamp, cityHash64(chunk_id))",
+                    partition_by="(retention_days, toStartOfDay(start_timestamp))",
+                    sample_by="cityHash64(chunk_id)",
+                    settings={"index_granularity": "8192"},
+                    storage_set=storage_set,
+                    ttl="toDateTime(end_timestamp) + toIntervalDay(retention_days)",
+                    version_column="deleted",
```
viglia, line 55: This is a follow-up to the previous [comment](https://github.com/getsentry/snuba/pull/5895/files#r1599004882) and maybe someone from @getsentry/sns can chime-in and answer this.

I'm looking at the clickhouse docs for `ReplacingMergeTree` and a couple of things caught my attention:

- `is_deleted` can only be enabled when `ver` is used
- `ENGINE = ReplacingMergeTree(eventTime, is_deleted)`

from what I can see here though: 
1) we're using the _deleted_ column as the _version_column_, and I don't think that's what we want? 
2) we're only setting the `VER` (version column), not the `is_deleted` since the current Snuba definition of the ReplacingMergeTree seems to neither accept nor set this column (see [here](https://github.com/getsentry/snuba/blob/master/snuba/migrations/table_engines.py#L126)).

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: I think it's best to keep them separate, these are 2 different features for 2 different products even if they share the same cluster and part of the name.

snuba/snuba_migrations/profile_chunks/0001_create_profile_chunks_table.py
phacops: This is correct with ClickHouse version >= 23.2.1 (https://github.com/ClickHouse/ClickHouse/commit/6526c2a8abbab6f027672903cf05e1537274f109). Before that, only the `ver` setting existed and I believe we used it like the `is_delete`. I believe the profiling cluster is still on 22.8.

Anyway, since then, they discovered the feature wasn't working well (https://github.com/ClickHouse/ClickHouse/issues/50346), removed it (https://github.com/ClickHouse/ClickHouse/pull/57932) and replaced it with (https://github.com/ClickHouse/ClickHouse/issues/54880).

snuba/migrations/groups.py
```diff
@@ -169,6 +172,11 @@ def __init__(
         storage_sets_keys={StorageSetKey.METRICS_SUMMARIES},
         readiness_state=ReadinessState.PARTIAL,
     ),
+    MigrationGroup.PROFILE_CHUNKS: _MigrationGroup(
+        loader=ProfileChunksLoader(),
+        storage_sets_keys={StorageSetKey.PROFILE_CHUNKS},
+        readiness_state=ReadinessState.PARTIAL,
```
enochtangg, line 178: Does the ClickHouse cluster already exist in SaaS (US + DE) and S4S? If the table is suppose to live the profiles cluster, has the `storage_set` been added there? If this isn't done yet, we should set the readiness state to `ReadinessState.LIMITED` for now so that GoCD doesn't try to run these migrations on a cluster that doesn't exist.

enochtangg (commented):
Left a comment

snuba/migrations/groups.py
phacops: The ClickHouse cluster we'd use exists in all environments (`us`, `de`, `s4s`, all STs).

I also have https://github.com/getsentry/ops/pull/10526 to configure it for `us` and `s4s`.

volokluev: +1

volokluev: which cluster are you using @phacops ?

phacops: The cluster dedicated to profiling.

volokluev (approved):

PR #5896, opened 2024-05-12
feat: Add support for DateTime64 column type

phacops:
We're relying a lot more on timestamps with sub-second precision for spans and continuous profiling and it will be handy to be able to store a timestamp in one column.


volokluev (approved):
Appreciate you coming in and grabbing a shovel Pierre. Thank you for your contribution

PR #5898, opened 2024-05-13
ref(api): parse pipeline tests, snql after treeify

kylemumma:
this is part 4 it depends on #5894 for the diff you will have to click only the last commit at the bottom

## Major Changes
* after treeify part of snql parse pipeline is not separated into its own stage
* there are tests for it

I verified these tests were good by tracking the expected number of tests, success and error. Feel free to skip these bullets unless you want to know how exactly I verified:
* the total number of e2e tests for snql was 19 success tests, 8 error tests.
* for stage 1 (before treeify) i had 23 success, 1 error, 3 lost (will talk more about this)
* hence for this stage I expect 19 success, 4 error which is what i got.

Noteably 3 tests from test_invalid_query got lost, this is bc they error in json_to_snql before ever calling parse_snql_query. So really... they arent parser tests. To account for this I will just keep these tests that were lost.

## Next Steps
* unit tests for treeify by hand
* delete e2e tests and some of the invalid_query ones.


snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,67 @@ def parse_snql_query(
...
+        res = PostProcessAndValidateQuery().execute(
+            QueryPipelineResult(
+                data=(query, dataset, settings, custom_processing),
+                error=None,
+                query_settings=dummy_settings,
+                timer=dummy_timer,
+            )
+        )
+        if res.error:
+            raise res.error
```
volokluev, line 1558: why raise and then recatch and raise again?

snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,67 @@ def parse_snql_query(
         with sentry_sdk.start_span(op="processor", description="treeify_conditions"):
             _post_process(query, [_treeify_or_and_conditions], settings)
 
+        # dummy settings and timer dont matter
+        dummy_settings = HTTPQuerySettings()
```
volokluev, line 1547: the query settings are passed into this function

snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,67 @@ def parse_snql_query(
...
+        dummy_timer = Timer("snql_pipeline")
```
volokluev, line 1548: This is not a dummy timer, it's just a timer

snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,67 @@ def parse_snql_query(
...
+        LogicalQuery | CompositeQuery[LogicalDataSource],
+    ]
+):
+    def _process_data(
+        self,
+        pipe_input: QueryPipelineData[
+            tuple[
+                LogicalQuery | CompositeQuery[LogicalDataSource],
+                Dataset,
+                QuerySettings | None,
```
volokluev, line 1584: you don't need QuerySettings here, thery're accessible in the pipeline stage

snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,67 @@ def parse_snql_query(
...
+                CustomProcessors | None,
```
volokluev, line 1585: I would leave CustomProcessors out of this stage, make a separate stage with them

volokluev (commented):
Pipeline is not being used properly

snuba/query/snql/parser.py
kylemumma: this also relates to the other comment. QueryPipelineResult expects QuerySettings, I have QuerySettings | None, which is why I pass it in and create the dummy one.

snuba/query/snql/parser.py
kylemumma: true... i just didnt think it actually gets used

snuba/query/snql/parser.py
volokluev: well then do `settings=query_settings or HTTPQuerySettings()`

snuba/query/snql/parser.py
kylemumma: Can I leave this for another time bc to factor it out into its own stage would require quite a few changes and I just want to get this done

snuba/query/snql/parser.py
kylemumma: done

snuba/query/snql/parser.py
volokluev: Then catch the treeify exception where it's happening and move the exception handling into the pipeline stage

snuba/query/snql/parser.py
volokluev: why would it require quite a few changes? custom processing is barely used in the codebase

snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,67 @@ def parse_snql_query(
         with sentry_sdk.start_span(op="processor", description="treeify_conditions"):
             _post_process(query, [_treeify_or_and_conditions], settings)
 
-        with sentry_sdk.start_span(op="processor", description="post_processors"):
-            _post_process(
-                query,
-                POST_PROCESSORS,
-                settings,
+        # dummy settings and timer dont matter
```
volokluev, line 1546:
```suggestion
```

snuba/query/snql/parser.py
kylemumma: im not sure how I would change the structure of the exception catching while maintaining the current behavior

snuba/query/snql/parser.py
```diff
@@ -1537,29 +1543,66 @@ def parse_snql_query(
...
-        return query
+
+        return res.data
     except InvalidQueryException:
         raise
     except Exception:
         raise PostProcessingError(query)
+
+
+class PostProcessAndValidateQuery(
```
volokluev, line 1575:
```suggestion
class PostProcessQuery(
```

volokluev (approved):
Good enough but the pipeline stage name is now no longer accurate

snuba/query/snql/parser.py
kylemumma: When I took custom processing out of the stage, this puts it before `replace_time_condition`, it turns out that it actually does things that need to come before. So this breaks things. I see my options as:

1. leave custom processing as part of the stage as it was
2. move custom processing before the stage instead of after to see it it works
3. split the stage into 3: post_proc1 -> custom_proc -> post_proc2

I would choose option 1, I see it being the most quick, easy, and safe change. With the tradeoff being the stages are less independent.

PR #5901, opened 2024-05-13
ref(api): parse pipeline tests, treeify unit tests and remove e2e

kylemumma:
this is the final stage of the test refactor, once this is merge its done. depends on #5898
## Major Changes
* add unit tests for treeify
* remove old e2e tests that have now been replaced
* add new e2e tests


snuba/query/snql/parser.py
```diff
@@ -1039,6 +1039,8 @@ def _treeify_or_and_conditions(
     codebase which assume `or` and `and` have two arguments
 
     Adding this post-process step is easier than changing the rest of the query pipeline
+
+    Note: does not apply to the conditions of a from_clause subquery (the nested one)
```
volokluev, line 1043: wait what? How?

kylemumma: see test_treeify_or_and_conditions.py, test_composite

volokluev: I see that you have written a test for it but I don't understand why this holds true

volokluev: After looking in deeper, I see that `transform_expressions` is not implemented for composite queries. You should explain that in the test that you wrote and maybe here as well. Otherwise someone is going to come to that code later, implement it and not understand why we don't transform composite subquery expressions but assume that it's just supposed to be this way. In reality this is an oversight

volokluev (approved):

PR #5908, opened 2024-05-14
fix(api): recursion bug fix experiment

kylemumma:
I have put the recursion bug fix behind a feature flag in order to see if it works in prod or breaks things

### DIFF
--- a/snuba/query/conditions.py
+++ b/snuba/query/conditions.py
@@ -286,7 +286,7 @@ def combine_and_conditions(conditions: Sequence[Expression]) -> Expression:
 
 def _combine_conditions(conditions: Sequence[Expression], function: str) -> Expression:
     rollout_percent = get_float_config(
-        "use_new_combine_conditions_percent", default=1.0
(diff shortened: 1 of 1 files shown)


snuba/query/conditions.py
```diff
@@ -283,6 +285,39 @@ def combine_and_conditions(conditions: Sequence[Expression]) -> Expression:
 
 
 def _combine_conditions(conditions: Sequence[Expression], function: str) -> Expression:
+    rollout_percent = get_float_config(
+        "use_new_combine_conditions_percent", default=0.0
+    )
+    assert isinstance(rollout_percent, float)
```
volokluev, line 291: if this assert fails it will break prod. Find a better way

volokluev, line 291:
```suggestion
    rollout_percent = rollout_percent if isinstance(rollout_percent, float) else 0.00
```

volokluev (approved):
Take out the assert and we're good

PR #5923, opened 2024-05-16
feat(profiles): Add a dataset to store profile chunks

phacops:
Same migration as in https://github.com/getsentry/snuba/pull/5895 which couldn't be applied due to permission problems with ClickHouse and was reverted because it blocked deploys. The permission issues are now fixed.


mcannizz (approved):
Same as previously approved PR https://github.com/getsentry/snuba/pull/5895.

PR #5930, opened 2024-05-16
chore(on-call): Do not count CANNOT_PARSE_DOMAIN_VALUE_FROM_STRING against SLO

enochtangg:
There are a bunch of queries that contain bad strings as and arg to the `toIPv4 ` clickhouse function. See [example issue](https://sentry.sentry.io/issues/5227106123/events/f6103be456d440779c4c43e2dd96a263/?project=1). The error code returned is `CANNOT_PARSE_DOMAIN_VALUE_FROM_STRING` and is currently count against our SLO. This should not be the case since the query is invalid.


volokluev:
Can we catch this in the query pipeline so we don't send it to clickhouse?

enochtangg:
> Can we catch this in the query pipeline so we don't send it to clickhouse?
@volokluev we probably could write some logic to catch it, but what benefit do we have from doing that?

volokluev:
We just don't waste clickhouse resources on queries that don't matter, also we can provide a more informative error message to the user

enochtangg:
That makes sense, it's possible we could weed out a bunch of queries from ever getting to Clickhouse. I'd also note that this error isn't the only invalid request we are catching. See the [error codes](https://github.com/getsentry/snuba/blob/6f09e4a24887cee279e2e1cf35b744e68c4a1bbc/snuba/querylog/query_metadata.py#L110-L124) that contain `RequestStatus.INVALID_REQUEST` (e.g. illegal columns, unknown functions, invalid uuid, etc.). If you'd like I can make a ticket for this and we could evaluate whether or not it's worth catching each of these cases and if it really does benefit ClickHouse?

volokluev (approved):

PR #5941, opened 2024-05-20
Metrics for replacements DD

xurui-c:
As part of the roll out plan of automatically skipping offending replacements projects, we want to create a DD dashboard to monitor the following metrics:
- number of projects skipped
- latency of adding skipped projects to redis
- latency of deleting expired projects from redis
- latency of getting skipped projects from redis

### DIFF
--- a/snuba/replacers/replacements_and_expiry.py
+++ b/snuba/replacers/replacements_and_expiry.py
@@ -50,6 +50,7 @@ def set_config_auto_replacements_bypass_projects(
             tags={},
         )
     except Exception as e:
+        metrics.increment("set_config_auto_replacements_bypass_projects_error")
         logger.exception(e)
 
 
@@ -69,6 +70,7 @@ def _retrieve_projects_from_redis() -> Mapping[int, datetime]:
         )
         return projects
     except Exception as e:
+        metrics.increment("retrieve_projects_from_redis_error")
         logger.exception(e)
         return {}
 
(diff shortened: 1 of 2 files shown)


snuba/replacers/replacements_and_expiry.py
```diff
@@ -8,8 +8,12 @@
 logger = logging.getLogger(__name__)
 
 
+from snuba import environment
 from snuba.redis import RedisClientKey, get_redis_client
 from snuba.state import get_int_config
+from snuba.utils.metrics.wrapper import MetricsWrapper
+
+metrics = MetricsWrapper(environment.metrics, "bucket_timer")
```
volokluev, line 16: Why is it called `bucket_timer`?

snuba/replacers/replacements_and_expiry.py
```diff
@@ -24,6 +28,7 @@ def set_config_auto_replacements_bypass_projects(
 ) -> None:
     try:
         projects_within_expiry = get_config_auto_replacements_bypass_projects(curr_time)
+        start = datetime.now()
```
volokluev, line 31: use `time.time()` instead

volokluev: that will give you the measurement in seconds already

snuba/replacers/replacements_and_expiry.py
xurui-c: Corrected to `replacements_and_expiry`

volokluev (approved):
also think it would be good to output a metric whenever anything fails

PR #5948, opened 2024-05-21
feat(capman): Short circuit allocation policy

volokluev:
Currently, all the policies of a storage run on every request. This can put undue load on redis when we already know the query is rejected. 

If one policy rejects from now on, just don't run any others

Also reorder the policies on errors so that the rejecting ones run first


untitaker (approved):

nikhars (approved):
🎉

PR #5952, opened 2024-05-21
number of projects skipped

xurui-c:
emits a DD metric for the number of projects skipped

### DIFF
--- a/snuba/replacers/errors_replacer.py
+++ b/snuba/replacers/errors_replacer.py
@@ -217,6 +217,7 @@ class ErrorsReplacer(ReplacerProcessor[Replacement]):
                 except Exception as e:
                     metrics.increment("errors_replacer_process_message_error")
                     logger.exception(e)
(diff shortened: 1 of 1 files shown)


snuba/replacers/errors_replacer.py
```diff
@@ -217,6 +217,7 @@ def process_message(
                 except Exception as e:
                     metrics.increment("errors_replacer_process_message_error")
                     logger.exception(e)
+            metrics.increment("projects_to_skip", value=len(projects_to_skip))
```
volokluev, line 220: is a counter metric what you want here?

xurui-c: I think so, since I want to exact number of projects that we are skipping

xurui-c: changed to `gauge`

volokluev (approved):

PR #5955, opened 2024-05-22
ref(async-inserts): allow async inserts as an option

MeredithAnya:
We are going to try out async inserts for specific S4S clusters to test it out and see the impacts of enabling it. This adds a cli arg that we can use to target consumers for specific clusters (e.g. https://github.com/getsentry/ops/pull/10687). I'm also adding a runtime config flag `async_inserts_allowed` so that we can turn it on and off as we test without having to put up code changes. 

When (and if) we move forward to enabling this on more clusters, we won't be able to use the runtime config in the same way since it would turn things on and off across the board, so this is really for testing purposes at this moment

Pre-req for https://github.com/getsentry/ops/pull/10687

### DIFF
--- a/rust_snuba/src/factory.rs
+++ b/rust_snuba/src/factory.rs
@@ -45,6 +45,7 @@ pub struct ConsumerStrategyFactory {
     pub clickhouse_concurrency: ConcurrencyConfig,
     pub commitlog_concurrency: ConcurrencyConfig,
     pub replacements_concurrency: ConcurrencyConfig,
+    pub async_inserts: bool,
     pub python_max_queue_depth: Option<usize>,
     pub use_rust_processor: bool,
     pub health_check_file: Option<String>,
@@ -55,7 +56,6 @@ pub struct ConsumerStrategyFactory {
     pub physical_topic_name: Topic,
     pub accountant_topic_config: config::TopicConfig,
     pub stop_at_timestamp: Option<i64>,
-    pub async_inserts: bool,
 }
 
 impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
--- a/rust_snuba/src/factory.rs
+++ b/rust_snuba/src/factory.rs
@@ -55,6 +55,7 @@ pub struct ConsumerStrategyFactory {
     pub physical_topic_name: Topic,
     pub accountant_topic_config: config::TopicConfig,
     pub stop_at_timestamp: Option<i64>,
+    pub async_inserts: bool,
 }
 
 impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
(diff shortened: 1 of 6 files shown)


rust_snuba/src/factory.rs
```diff
@@ -116,6 +116,7 @@ impl ProcessingStrategyFactory<KafkaPayload> for ConsumerStrategyFactory {
             &self.clickhouse_concurrency,
             &self.storage_config.clickhouse_cluster.user,
             &self.storage_config.clickhouse_cluster.password,
+            self.storage_config.clickhouse_cluster.async_inserts,
```
MeredithAnya, line 119: I was thinking for a minute this would actually be easier to have as a clickhouse config but it seems like a much more involved change than I thought so going to move this to `cli` arg

rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -48,7 +50,14 @@ impl BatchFactory {
             HeaderValue::from_str(database).unwrap(),
         );
 
-        let query_params = "load_balancing=in_order&insert_distributed_sync=1".to_string();
+        let mut query_params = String::new();
+        query_params.push_str("load_balancing=in_order&insert_distributed_sync=1");
+
+        let async_inserts_allowed = get_str_config("async_inserts_allowed").ok().flatten();
+        if async_inserts == true && async_inserts_allowed == Some(1.to_string()) {
```
MeredithAnya, line 57: I don't know if this is the right way to do this, but my goal is to have a runtime config that we can turn on and off for our testing purposes as well as a cli arg (`async_inserts`) for the specific cluster we want

This generic runtime config wouldn't really scale if we are trying to turn async inserts on and off for different clusters, but it's mainly for running the first experiments in a safe/easy way

untitaker: this looks fine, however i would suggest to only query the get_str_config when `async_inserts` is true

rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -48,7 +50,14 @@ impl BatchFactory {
             HeaderValue::from_str(database).unwrap(),
         );
 
-        let query_params = "load_balancing=in_order&insert_distributed_sync=1".to_string();
+        let mut query_params = String::new();
+        query_params.push_str("load_balancing=in_order&insert_distributed_sync=1");
+
+        let async_inserts_allowed = get_str_config("async_inserts_allowed").ok().flatten();
+        if async_inserts && async_inserts_allowed == Some(1.to_string()) {
```
untitaker, line 57:
```suggestion
        if async_inserts && async_inserts_allowed == Some("1") {
```

that should also work

MeredithAnya: I think I still need `to_string()` because it was expecting `String` instead of `&str`

untitaker (approved):

PR #5963, opened 2024-05-23
Debugging bucket timer

xurui-c:
As part of the debugging efforts on why [auto replacements project skipping](https://github.com/getsentry/snuba/pull/5883) isn't working as intended, we have a new hypothesis that the bucket timer algo we depend on isn't working as we expected. More specifically, we noticed (from logging) that [projects_exceeding_limit](https://github.com/getsentry/snuba/blob/76fd1fff03be960a5da216223ed809860fa0fd6b/snuba/replacer.py#L596) logs, but [projects_exceeding_time_limit](https://github.com/getsentry/snuba/blob/76fd1fff03be960a5da216223ed809860fa0fd6b/snuba/utils/bucket_timer.py#L95) does not. We add additional logging to confirm the control flow of our code

### DIFF
--- a/snuba/utils/bucket_timer.py
+++ b/snuba/utils/bucket_timer.py
@@ -49,6 +49,7 @@ class Counter:
         self.limit = self.counter_window_size * percentage
 
     def __trim_expired_buckets(self, now: datetime) -> None:
+        logger.debug("Entering __trim_expired_buckets method")
         current_minute = floor_minute(now)
         window_start = current_minute - self.counter_window_size
         new_buckets: Buckets = {}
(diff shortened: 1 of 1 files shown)


snuba/utils/bucket_timer.py
```diff
@@ -88,6 +88,7 @@ def get_projects_exceeding_limit(self) -> List[int]:
         now = datetime.now()
         self.__trim_expired_buckets(now)
         project_groups: dict[int, timedelta] = defaultdict(lambda: timedelta(seconds=0))
+        logger.debug("self.buckets_size: %s" % len(self.buckets))
```
volokluev, line 91: is this all that you want to instrument? It takes a while to deploy something

xurui-c: no I'm still drafting

snuba/utils/bucket_timer.py
```diff
@@ -75,6 +77,7 @@ def __add_to_bucket(
     def record_time_spent(
         self, project_id: int, start: datetime, end: datetime
     ) -> None:
+        logger.info("Entering record_time_spent method")
```
volokluev, line 80: these should be debug logs

volokluev (approved):

PR #5965, opened 2024-05-23
allow one replacement project to be skipped

xurui-c:
In the current replacements set up, if a project takes up more time processing replacements than what the threshold allows, it actually won't be skipped if it's the only project that exists. The intention was that it's fine for the single existing project to use up the entire replacements processing budget. Thus, when we are testing within S4S, even if we are going over the testing threshold (which is 0), we are not adding sentry-for-sentry to the list of projects to skip, since it's the only project that exists.

This PR uses another runtime config, `s4s_replacements_testing`, so that if it's true, we allow the single existing project to be skipped as long as it exceeds the threshold (0).

This is for S4S testing only and will be removed

### DIFF
--- a/snuba/utils/bucket_timer.py
+++ b/snuba/utils/bucket_timer.py
@@ -112,7 +112,10 @@ class Counter:
                 % (project_id, total_processing_time, self.limit)
             )
             if total_processing_time > self.limit and (
-                len(project_groups) > 1 or get_int_config("s4s_replacements_testing")
+                len(project_groups) > 1
+                or get_int_config(
+                    "allows_skipping_single_project_replacements", default=0
+                )
             ):
                 projects_exceeding_time_limit.append(project_id)
 
--- a/snuba/utils/bucket_timer.py
+++ b/snuba/utils/bucket_timer.py
@@ -111,7 +111,9 @@ class Counter:
                 "project_id: %s, total_processing_time: %s, limit: %s"
                 % (project_id, total_processing_time, self.limit)
             )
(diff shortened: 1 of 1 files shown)


snuba/utils/bucket_timer.py
```diff
@@ -111,7 +111,10 @@ def get_projects_exceeding_limit(self) -> List[int]:
                 "project_id: %s, total_processing_time: %s, limit: %s"
                 % (project_id, total_processing_time, self.limit)
             )
-            if total_processing_time > self.limit and len(project_groups) > 1:
+            if total_processing_time > self.limit and (
+                len(project_groups) > 1
+                or get_int_config("s4s_replacements_testing", default=0)
```
volokluev, line 116: I think a more descriptive name would be more apt like `allows_skipping_single_project_replacements` I'm also not even sure if this `len(project_groups) > 1` thing even makes sense or not but we can do that later

volokluev (approved):

snuba/utils/bucket_timer.py
xurui-c: I think it makes sense because if it's the only project that is doing replacements, why not just let it consume all the resources

volokluev: My first thought is:

If a project sent a massive replacements batch and taken up a whole bunch of resources to do it it blocks the queue for any other projects trying to do replacements. Just because this batch only has one project I'm not sure what that says about it taking too long. 

This is just a thought though, I could be misinformed about how the replacer works

PR #5975, opened 2024-05-27
fix(meta) Change the view to use LEFT ARRAY JOIN

evanh:
In the current implementation if a metric has no tags associated with it, it will get excluded from the meta table since the array join only selects rows that have values.

This change will cause rows to be created with an empty tag key (0) so that the metrics can still be searched for in the meta tables.

~TODO: This depends on https://github.com/getsentry/snuba/pull/5974 because I still need to validate that this
actually works correctly before this can be merged.~


evanh:
Validated these queries in snuba admin and they work as expected. There might need to be a follow up PR in the metrics layer to handle the `tag_key = 0` case.

john-z-yang (approved):

PR #5976, opened 2024-05-28
ref(subscriptions): Add query-based exception handling in scheduler 

ayirr7:


evanh (approved):

lynnagara:
Just curious how an invalid query got in there to begin with given that they are validated and run when created?

evanh:
> Just curious how an invalid query got in there to begin with given that they are validated and run when created?

This is something I've been trying to figure out and so far have had no luck. I'm hoping maybe we can pull some logs or recreate the request from Sentry in order to reproduce.

ayirr7:
@evanh I could limit it to gauges, but wouldn't we want the same behavior across datasets? Ideally, we never crashloop if we hit this condition?

lynnagara (changes requested):
I feel like it's important to address the real problem of how we allow completely malformed queries in here to begin with? My fear is skipping them in this way will allow broken queries to persist in the system and we will not even be aware of them.

evanh:
> I feel like it's important to address the real problem of how we allow completely malformed queries in here to begin with? My fear is skipping them in this way will allow broken queries to persist in the system and we will not even be aware of them.

I'm working with @shellmayr to deduce the main problem. However this is causing subscriptions scheduler to crash. @ayirr7 that's also why I changed your PR to only skip for gauges, which is the only known broken thing right now. Once we figure out and fix the cause, we can remove this.

@lynnagara I would like to merge this now, since it will stop the scheduler from crashing and make it easier to debug the problem (since we log extra info now).

lynnagara:
@evanh how will merging this and skipping the exception will help debug exactly?

also the error is `snuba.query.exceptions.InvalidQueryException: organization not found in metadata: {}`

surely that is already enough info to debug

PR #5978, opened 2024-05-28
Remove replacements logging that was previously used for roll out

xurui-c:
Previously, I added logging as part of rolling out [auto-skipping offending replacements projects](https://github.com/getsentry/snuba/pull/5883). After verifying that things work as expected within S4S, I'm ready to remove logging

### DIFF
--- a/snuba/utils/bucket_timer.py
+++ b/snuba/utils/bucket_timer.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 import typing
 from collections import defaultdict
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


onewland (approved):
looks fine, but I want to point out a couple things:

1. always let the logging framework do string interpolation so that you don't create new strings for messages you don't log.

```
    logger.debug(
        "project_id: %s, total_processing_time: %s, limit: %s"
        % (project_id, total_processing_time, self.limit)
    )
```
could be quite expensive if run many times in a tight loop, but if you passed the arguments as params to logger.debug, they only get interpolated if the system is configured to log at the debug level.

2. it's mostly fine to leave debug logs in. In production, they should not have a performance penalty, and we should not be emitting them (minimum log level should be info)

PR #5982, opened 2024-05-28
ref(subscriptions): Improve logging by adding metadata field

ayirr7:
The metadata field is where additional `organization` column data should be passed through for generic metrics subscriptions. Add this info to logging so we can better identify the problem at hand.


snuba/subscriptions/scheduler_processing_strategy.py
```diff
@@ -468,8 +468,9 @@ def submit(self, message: Message[CommittableTick]) -> None:
                     entity = task.task.subscription.data.entity
                     if get_entity_name(entity) == EntityKey.GENERIC_METRICS_GAUGES:
                         logger.warning(
-                            "Skipping malformed subscription query %r in scheduler",
+                            "Skipping malformed subscription query %r with metadata %r in scheduler",
                             task.task.subscription.data.query,
+                            task.task.subscription.data.metadata,
                         )
```
nikhars, line 474: Is there a reason this is happening only for the gauges subscription? I would find this information useful for other entities as well.

PR #5985, opened 2024-05-30
Emit metric for throttled queries

xurui-c:
Successful queries can still be throttled, but there is currently no way of surfacing this. This PR emits a metric for every query that executed successfully (but was throttled), so we can see it in DD


volokluev:
Not necessary

PR #5986, opened 2024-05-30
feat(issue-search): Add group_first_release_id to the GroupAttributes table

snigdhas:
Add a new column, `group_first_release_id` to the GroupAttributes table. This column will be backfilled as part of the issue search improvements, with data from `Group.first_release_id`. 

This column will help speed up searches for `firstRelease`. We'll still handle `firstRelease` queries using Postgres if the query includes the environment.

### DIFF
--- a/snuba/snuba_migrations/group_attributes/0003_add_first_release_id_to_group_attributes.py
+++ b/snuba/snuba_migrations/group_attributes/0003_add_first_release_id_to_group_attributes.py
@@ -1,0 +1,49 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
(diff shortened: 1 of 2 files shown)


vartec (approved):

wedamija (commented):
This looks good to me. `first_release_id` also exists on `GroupEnvironment`, are we also planning to replicate that somehow?

evanh (approved):

snuba/snuba_migrations/group_attributes/0003_add_first_release_id_to_group_attributes.py
```diff
@@ -0,0 +1,49 @@
...
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_local",
+                column=Column(
+                    "group_first_release_id", UInt(8, Modifiers(nullable=True))
```
evanh, line 19: Question, is this actually a full release ID? Is that really a UInt8 only? Aren't releases UUIDs?

snigdhas:
> This looks good to me. first_release_id also exists on GroupEnvironment, are we also planning to replicate that somehow?

We'll handle `firstRelease` queries using Postgres if the query includes the environment. Adding this column will still help improve queries that don't include an anv.

evanh (approved):

PR #5988, opened 2024-05-31
feat: autogenerate add migrations checkpoint 1

kylemumma:
The last PR #6025 was the UI. It took a path to a storage and returned the original and modified version of that storage, which it determined as: the modified one is whats in working directory and the original one is whats at head.

This PR takes those original and modified storage, and returns the corresponding migration operations to get from original to modified.

It adds this as an extension to the last PR.

The addition is `generate_migration_ops` which:
* given an old storage yaml and a new storage yaml,
* validates the diff between the storages is valid column addition,
* converts the diff into forward and backward SqlOperations
* returns them

## Next Step
take the forward and backward operation objects and generate python migration


.vscode/launch.json
```diff
@@ -1,6 +1,13 @@
 {
     "version": "0.2.0",
     "configurations": [
+        {
```
kylemumma, line 4: unrelated to current PR, this allows vscode users to run python files in the debugger

snuba/migrations/operations.py
```diff
@@ -82,6 +84,11 @@ def execute(self) -> None:
     def format_sql(self) -> str:
         raise NotImplementedError
 
+    def __eq__(self, other: Any) -> bool:
```
kylemumma, line 87: I needed __eq__ on operations for my tests, this seems like a reasonable implementation

tests/migrations/test_autogen_migration.py
```diff
@@ -0,0 +1,160 @@
...
+        "{ name: project_id, type: UInt, args: { size: 64 } }",
+        "{ name: timestamp, type: DateTime }",
+        "{ name: event_id, type: UUID }",
+    ]
+    new_cols = [
+        "{ name: project_id, type: UInt, args: { size: 64 } }",
+        "{ name: timestamp, type: DateTime }",
+        "{ name: newcol1, type: DateTime }",
+        "{ name: event_id, type: UUID }",
+        "{ name: newcol2, type: UInt, args: { schema_modifiers: [nullable], size: 8 } }",
```
volokluev, line 47: what happens if the newcol is not appended at the end of the columns array?

tests/migrations/test_autogen_migration.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def test_modify_column() -> None:
+    cols = [
+        "{ name: timestamp, type: DateTime }",
+    ]
+    new_cols = [
+        "{ name: timestamp, type: UUID }",
+    ]
+    with pytest.raises(ValueError):
```
volokluev, line 122: would be good to check that this fails for the reason you expect and that the error message is correct and instructive to the user

volokluev (commented):
I appreciate the incremental nature, how do you expect this to be used by snuba customers?

onewland:
> I appreciate the incremental nature, how do you expect this to be used by snuba customers?

Yeah I'm a bit confused as to how this will get used in practice. Can we have something like a fake/example transcript of how a user goes about adding a new column to a schema? You mention a UI in the PR, but I'm not sure I want to require users to run a web interface to use this feature

kylemumma:
@volokluev @onewland 
Sorry heres how I expect it to be used:
1. User adds a column to errors.yaml storage
2. User does something like `snuba migration autogenerate snuba/datasets/configuration/events/storages/errors.yaml`
3. (the above command is what I meant by UI) this will fetch the remote file from github, as well as the local file, and pass them in as strings to the functions implemented in this PR (as oldstorage and newstorage)

tests/migrations/test_autogen_migration.py
kylemumma: it will still work (see newcol1), i just tested it but I can add a test where 1 isnt at the end if you think it would be good.

volokluev:
@kylemumma I think the interface on how this will be used needs more thought. I recommend writing out a document considering multiple options and explaining your decision. Right now this is implementing from the inside out, I'd like you to implement from the outside in (interface first, then internals)

snuba/migrations/autogeneration/diff.py
```diff
@@ -0,0 +1,139 @@
...
+    oldstorage_cols = oldstorage_dict["schema"]["columns"]
+    newstorage_cols = newstorage_dict["schema"]["columns"]
+
+    colnames_old = set(e["name"] for e in oldstorage_cols)
+    colnames_new = set(e["name"] for e in newstorage_cols)
+    if not colnames_old.issubset(colnames_new):
+        return (False, "Column removal is not supported")
+
+    pold, pnew = 0, 0
+    while pold < len(oldstorage_cols) and pnew < len(newstorage_cols):
```
nachivrn, line 99: Kyle, can we just use the set data structure and set operations to determine if new columns are added.

nachivrn (commented):
Just had one question. Other than that code looks good!

snuba/migrations/autogeneration/diff.py
kylemumma: Using a set would lose the concept of ordering of columns. Currently the [AddColumn object](https://github.com/getsentry/snuba/blob/271f64e3c026bdb1595fdf2657176174f83ad61a/snuba/migrations/operations.py#L255) has a notion of column order from the "after" parameter. So I used this to make sure the order of the columns matters.

nachivrn (approved):

PR #5991, opened 2024-05-31
Include metrics for the number of queries resulting in cache hits, misses, and waits associated with datasets.

nachivrn:
Include metrics for the number of queries resulting in cache hits, misses, and waits associated with datasets

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -783,6 +783,12 @@ def db_query(
                 query_id=query_id,
                 result_or_error=QueryResultOrError(query_result=result, error=error),
             )
+        if stats.get("cache_hit"):
+            metrics.increment("cache_hit", tags={"dataset": dataset_name})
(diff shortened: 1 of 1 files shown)


snuba/web/db_query.py
```diff
@@ -783,6 +783,12 @@ def db_query(
                 query_id=query_id,
                 result_or_error=QueryResultOrError(query_result=result, error=error),
             )
+        if stats.get("cache_hit"):
+            metrics.increment("cache_hit", tags={"dataset": dataset_name})
+        elif stats.get("cache_wait"):
```
evanh, line 788: I don't think is correct, the code uses `is_duplicate`: https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L398

nachivrn: @evanh Thanks for catching this error. I will update

volokluev (approved):

PR #5992, opened 2024-06-01
SNS-2737: Making changes for a simple readthrough cache without queuing

nachivrn:
read-through cache has complex logic to manage different states and conditions, queuing incoming queries to prevent duplicate executions. However, internal analysis shows that this complex logic of queuing in cache provides minimal benefit for searches with the same query ID concurrently, as these concurrent searches are relatively infrequent compared to the total volume.  Therefore, the advantage of this complex logic does not justify the reduced maintainability of the code.

### DIFF
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -1,5 +1,6 @@
 import concurrent.futures
 import logging
+import random
 import uuid
 from concurrent.futures import ThreadPoolExecutor
 from typing import Callable, Optional
@@ -335,7 +336,12 @@ class RedisCache(Cache[TValue]):
 
         try:
             # set disable_lua_scripts to use the simple read-through cache without queueing.
-            if get_config("read_through_cache.disable_lua_scripts", 0):
+            sample_rate = get_config(
+                "read_through_cache.disable_lua_scripts_sample_rate", 0
+            )
+            assert sample_rate is not None
+            disable_lua_scripts = random.random() < float(sample_rate)
+            if disable_lua_scripts:
                 return self.__get_value_with_simple_readthrough(
                     key, function, record_cache_hit_type, timeout, timer
                 )
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -283,6 +283,31 @@ class RedisCache(Cache[TValue]):
         else:
             raise ValueError("unexpected result from script")
 
+    def __get_readthrough_cache(
+        self,
+        key: str,
+        function: Callable[[], TValue],
+        record_cache_hit_type: Callable[[int], None],
+    ) -> TValue:
+        result_key = self.__build_key(key)
+
+        value = self.__client.get(result_key)
+
+        if value is not None:
+            record_cache_hit_type(RESULT_VALUE)
+            return self.__codec.decode(value)
+        else:
+            # If the value does not exist in the cache, execute the function
+            # and store the result in the cache
+            value = function()
+            self.__client.set(
+                result_key,
+                self.__codec.encode(value),
+                ex=get_config("cache_expiry_sec", 1),
+            )
+            record_cache_hit_type(RESULT_EXECUTE)
+            return value
+
     def get_readthrough(
         self,
         key: str,
@@ -297,9 +322,15 @@ class RedisCache(Cache[TValue]):
             return function()
 
         try:
-            return self.__get_readthrough(
-                key, function, record_cache_hit_type, timeout, timer
-            )
+            # set disable lua scripts to use the simple read through cache without queueing.
+            if get_config("read_through_cache.disable_lua_scripts", 0):
+                return self.__get_readthrough_cache(
+                    key, function, record_cache_hit_type
+                )
+            else:
+                return self.__get_readthrough(
+                    key, function, record_cache_hit_type, timeout, timer
+                )
         except (ConnectionError, ReadOnlyError, RedisTimeoutError, ValueError):
             if settings.RAISE_ON_READTHROUGH_CACHE_REDIS_FAILURES:
                 raise
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -293,14 +293,14 @@ class RedisCache(Cache[TValue]):
     ) -> TValue:
         result_key = self.__build_key(key)
 
-        value = self.__client.get(result_key)
+        cached_value = self.__client.get(result_key)
         if timer is not None:
             timer.mark("cache_get")
         metric_tags = timer.tags if timer is not None else {}
 
-        if value is not None:
+        if cached_value is not None:
             record_cache_hit_type(RESULT_VALUE)
-            return self.__codec.decode(value)
+            return self.__codec.decode(cached_value)
         else:
             try:
                 value = self.__executor.submit(function).result(timeout)
--- a/snuba/state/cache/redis/backend.py
+++ b/snuba/state/cache/redis/backend.py
@@ -306,7 +306,7 @@ class RedisCache(Cache[TValue]):
             return self.__codec.decode(cached_value)
         else:
             try:
-                value = self.__executor.submit(function).result(timeout)
+                value = function()
                 self.__client.set(
                     result_key,
                     self.__codec.encode(value),
@@ -315,9 +315,6 @@ class RedisCache(Cache[TValue]):
                 record_cache_hit_type(RESULT_EXECUTE)
                 if timer is not None:
                     timer.mark("cache_set")
-            except concurrent.futures.TimeoutError as error:
-                metrics.increment("execute_timeout", tags=metric_tags)
-                raise TimeoutError("timed out while running query") from error
             except Exception as e:
                 metrics.increment("execute_error", tags=metric_tags)
                 raise e
@@ -341,8 +338,9 @@ class RedisCache(Cache[TValue]):
             sample_rate = get_config(
                 "read_through_cache.disable_lua_scripts_sample_rate", 0
             )
-            assert sample_rate is not None
-            disable_lua_scripts = random.random() < float(sample_rate)
+            disable_lua_scripts = sample_rate is not None and random.random() < float(
+                sample_rate
+            )
             if disable_lua_scripts:
                 return self.__get_value_with_simple_readthrough(
                     key, function, record_cache_hit_type, timeout, timer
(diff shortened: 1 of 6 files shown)


snuba/state/cache/redis/backend.py
```diff
@@ -297,9 +334,15 @@ def get_readthrough(
             return function()
 
         try:
-            return self.__get_readthrough(
-                key, function, record_cache_hit_type, timeout, timer
-            )
+            # set disable_lua_scripts to use the simple read-through cache without queueing.
+            if get_config("read_through_cache.disable_lua_scripts", 0):
```
evanh, line 338: I would change this to a percentage check, so you can roll out incrementally.

```
if random.random() < get_config(...):
```

evanh (commented):
Please provide a description of your changes in the PR, along with links to any relevant docs and context / impact of the changes.

Also, this isn't a sufficient change, you will also need to ensure that the `query_id` is being set correctly. When we send a query to Clickhouse, we also have to include a `query_id`. Clickhouse will reject queries with identical `query_id`s. Normally the `query_id` is a hash of the SQL string, and the readthrough cache ensures that duplicate `query_id`s won't get sent to Clickhouse. https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L307

However with this change, we might see duplicate `query_id` in Clickhouse. So you will need to ensure that the `query_id` is overridden while still maintaining the cache keys.

snuba/state/cache/redis/backend.py
```diff
@@ -283,6 +283,43 @@ def build_notify_queue_key(task_ident: str) -> str:
         else:
             raise ValueError("unexpected result from script")
 
+    def __get_value_with_simple_readthrough(
```
volokluev, line 286: how would the cache value ever be set here?

nachivrn:
@evanh Regarding the comment about queryId, it is set in the db_query file :
https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L387

When concurrent queries are received with the same queryId, due to a cache miss with the simple read-through cache, all concurrent queries will be directed to ClickHouse. One of the queries will be executed while the others will be rejected by ClickHouse. This situation is handled in db_query, which assigns a random value to queryId for the rejected queries.
https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L337

snuba/state/cache/redis/backend.py
nachivrn: I will update the code.

onewland:
> @evanh Regarding the comment about queryId, it is set in the db_query file : https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L387
> 
> When concurrent queries are received with the same queryId, due to a cache miss with the simple read-through cache, all concurrent queries will be directed to ClickHouse. One of the queries will be executed while the others will be rejected by ClickHouse. This situation is handled in db_query, which assigns a random value to queryId for the rejected queries. https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L337

Isn't this worse than randomizing the query ID before sending it to ClickHouse, though? It seems like there's a performance (and simplicity) penalty for two round-trips

evanh:
> Isn't this worse than randomizing the query ID before sending it to ClickHouse, though? It seems like there's a performance (and simplicity) penalty for two round-trips

I agree with Oliver here. The correct flow should be: check cache, and if it's a miss, randomize the query ID and send it to Clickhouse.

nachivrn:
@evanh @onewland If we get concurrent duplicate queries handled by different pods, there are challenges to keep track of duplicate queries. We may need to keep track in shared data structure in redis and also come up with a way to evict those entries.

nachivrn:
@evanh @onewland Thanks for your review comments. The focus of the Jira task was on simplifying the read-through cache by removing the queueing mechanism and adopting a straightforward approach, acknowledging that this introduces a performance penalty due to additional round-trips. The performance issue is recognized and can be addressed separately in a different PR. We will explore optimizations and monitor the system's performance after merging this PR.

nachivrn:
Notes: https://www.notion.so/sentry/Read-through-cache-simplification-85b308c25da748049fb280fdf32ea384

Rollout plan: https://www.notion.so/sentry/Experiment-1-Turn-off-read-through-cache-with-queuing-and-use-simple-read-through-cache-7bbd60f358874502b6045e34f0e69a5e

volokluev (commented):
You should have a test that tests this cache functionality at least one level higher, e.g. in test_db_query

snuba/state/cache/redis/backend.py
```diff
@@ -283,6 +283,43 @@ def build_notify_queue_key(task_ident: str) -> str:
...
+        if timer is not None:
+            timer.mark("cache_get")
+        metric_tags = timer.tags if timer is not None else {}
+
+        if cached_value is not None:
+            record_cache_hit_type(RESULT_VALUE)
+            return self.__codec.decode(cached_value)
+        else:
+            try:
+                value = self.__executor.submit(function).result(timeout)
```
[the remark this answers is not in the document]
nachivrn: I used the concurrent.futures module for function timeout.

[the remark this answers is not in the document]
volokluev: I don't think it's as necessary here because there's no notification queue. I think it's fine to just run the query synchronously. That's what we do now when we `short_circuit` the cache.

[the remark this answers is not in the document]
nachivrn: Ok. I will update to a synchronous call.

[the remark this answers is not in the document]
nachivrn: While debugging the code, I still opt to maintain the timeout. 

This is because setting the timeout value too high could result in the thread being held up, waiting for the completion of the clickhouse query.
https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/native.py#L382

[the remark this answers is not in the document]
evanh: Clickhouse will reject the query if it takes more than 30 seconds as well. So we shouldn't end up actually running a query longer than 30 seconds either way. I tend to agree with Volo: let's remove this added complexity. If we find that Clickhouse is holding threads longer than 30 seconds we can add it back.

snuba/state/cache/redis/backend.py
```diff
@@ -297,9 +337,20 @@ def get_readthrough(
             return function()
 
         try:
-            return self.__get_readthrough(
-                key, function, record_cache_hit_type, timeout, timer
+            # set disable_lua_scripts to use the simple read-through cache without queueing.
+            sample_rate = get_config(
+                "read_through_cache.disable_lua_scripts_sample_rate", 0
             )
+            assert sample_rate is not None
```
evanh, line 344: This should be an `if sample_rate is not None`. Otherwise someone could break queries simply by setting a bad value here (or having no value).

evanh (approved):
Couple comments, otherwise looks good.

snuba/state/cache/redis/backend.py
nachivrn: @evanh above line sets the defualt value.  
sample_rate = get_config(
    "read_through_cache.disable_lua_scripts_sample_rate", 0
)

snuba/state/cache/redis/backend.py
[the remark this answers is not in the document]
nachivrn: Ok. I will be updating code to synchronous call.

snuba/state/cache/redis/backend.py
nachivrn: Implemented fix.

tests/web/test_db_query.py
```diff
@@ -714,3 +720,118 @@ def test_db_query_ignore_consistent() -> None:
...
+            query_metadata_list=[],
+            formatted_query=formatted_query,
+            reader=reader,
+            timer=Timer("foo"),
+            stats={},
+            trace_id="trace_id",
+            robust=False,
+        )
+
+    assert result.extra["stats"]["lua_scripts_disabled"]
```
volokluev, line 765:
```suggestion
    assert result.extra["stats"]["cache_hit_simple"]
```

tests/web/test_db_query.py
```diff
@@ -714,3 +720,118 @@ def test_db_query_ignore_consistent() -> None:
...
+            if mock_data and cached_data:
+                if mock_data[0] is not None and cached_data[0] is not None:
+                    assert mock_data[0] == cached_data[0]
+    if mock_result["meta"] and cached_value.get("meta"):
+        mock_meta = mock_result.get("meta")
+        cached_meta = cached_value.get("meta")
+        if isinstance(mock_meta, list) and isinstance(cached_meta, list):
+            if mock_meta and cached_meta:
+                if mock_meta[0] is not None and cached_meta[0] is not None:
+                    assert mock_meta[0] == cached_meta[0]
```
volokluev, line 782: this test is doing and mocking too much. Why not just ran the same query twice and make sure there's a cache hit?

tests/web/test_db_query.py
```diff
@@ -714,3 +720,118 @@ def test_db_query_ignore_consistent() -> None:
...
+    query, storage, attribution_info = _build_test_query("count(distinct(project_id))")
+    state.set_config("disable_lua_randomize_query_id", disable_lua_randomize_query_id)
+    state.set_config(
+        "read_through_cache.disable_lua_scripts_sample_rate",
+        disable_lua_scripts_sample_rate,
+    )
+
+    query_settings = HTTPQuerySettings()
+    formatted_query = format_query(query)
+    reader = mock.MagicMock()
```
volokluev, line 807: why do you need to mock this?

nachivrn: Removed the mocking.

tests/web/test_db_query.py
nachivrn: Second test cases handles. I have removed this testcase

tests/web/test_db_query.py
```diff
@@ -714,3 +719,56 @@ def test_db_query_ignore_consistent() -> None:
...
+    )
+
+    query_settings = HTTPQuerySettings()
+    formatted_query = format_query(query)
+    reader = storage.get_cluster().get_reader()
+    clickhouse_query_settings: Dict[str, Any] = {}
+    stats: dict[str, Any] = {}
+
+    with mock.patch(
+        "snuba.web.db_query.get_query_cache_key", return_value="test_query_id"
```
volokluev, line 753: why does this need to be mocked? The cache key should be the same if the query is the same

nachivrn: Reason for mocking, is in the test query_id is not computed.

volokluev (approved):

PR #5998, opened 2024-06-04
ref(async-inserts): Add system queries

MeredithAnya:
These are some very basic queries that we can use (and modify) for our purposes when testing out async inserts. I would have liked to use the more robust queries that I found in https://clickhouse.com/blog/monitoring-asynchronous-data-inserts-in-clickhouse but it was taking too long to easily add these with the way the system queries is built right now.

So for now we have 

* `PartCreations`: This can let us see how many rows/the size each of our new parts has, with async inserts this should be much higher in theory than we have now since it should be buffering multiple inserts. 
* `AsyncInsertPendingFlushes`: This can give us a quick pulse on how many inserts are pending flushing in total
* `AsyncInsertFlushes`: This is like the query log, but for the async insert flush, and can be used in tandem with the query log by looking up the query_id, though this query is for successful inserts flushes
* `AsyncInsertFlushErrors`: Same use as above but specific to failed insert flushes

### DIFF
--- a/snuba/admin/clickhouse/predefined_system_queries.py
+++ b/snuba/admin/clickhouse/predefined_system_queries.py
@@ -132,3 +132,92 @@ class ColumnSizeOnDisk(SystemQuery):
         column
     ORDER BY size DESC;
     """
+
+
+class PartCreations(SystemQuery):
+    """
+    New parts created in the last 10 minutes. Replicas have event_type 'DownloadPart',
+    therefore we'll only see the new parts on the node that got the insert
+    """
+
+    sql = """
+    SELECT
+        hostName() AS node_name,
+        event_time,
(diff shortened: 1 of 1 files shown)


nikhars (approved):
Curious to know if there are any async insert metrics which would be useful for troubleshooting problems. If yes, it would be nice to have them on our clickhouse datadog dashboards.

MeredithAnya:
> Curious to know if there are any async insert metrics which would be useful for troubleshooting problems. If yes, it would be nice to have them on our clickhouse datadog dashboards.

Yeah was thinking about this too - I think I'll wait until the experiments to see what kind of metrics we can use/how turning on async inserts impacts the metrics we already have

PR #6005, opened 2024-06-05
fix: snuba admin queries check

MeredithAnya:
Some queries fail because we don't do `sql_query = " ".join(sql_query.split())` before we match on the regex, I believe this is due to the fact that we explicitly have `\s` in the regex, vs a newline


volokluev (commented):
what is this fixing and is there a test for it?

MeredithAnya:
> what is this fixing and is there a test for it?

Added a test, but some snuba admin queries just return `"Query is invalid"` even though they should be fine. One such example is the query used in the test I added. It wasn't matching properly before

volokluev (approved):

PR #6006, opened 2024-06-06
ref(gocd): Add a crashloop monitor to deploys

evanh:
Add a crashloop monitor to the S4S deploy to ensure that the changes at least keep the pods stable.

### DIFF
--- a/gocd/templates/bash/s4s-ddog-health-check.sh
+++ b/gocd/templates/bash/s4s-ddog-health-check.sh
@@ -1,1 +1,8 @@
 #!/bin/bash
+
+/devinfra/scripts/checks/datadog/monitor_status.py --dry-run=true \
+  146517092
+
+
+# Above monitor IDs map to following monitors respectively:
+# Snuba Pods Crashlooping in {{sentry_region.name}}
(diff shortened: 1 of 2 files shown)


gocd/templates/bash/s4s-ddog-health-check.sh
```diff
@@ -1 +1,8 @@
 #!/bin/bash
+
+/devinfra/scripts/checks/datadog/monitor_status.py --dry-run=true \
```
volokluev, line 3: what does `dry-run=true` do here?

volokluev (approved):

gocd/templates/bash/s4s-ddog-health-check.sh
evanh: Oh hmm it means that on failure it doesn't actually raise an alert. Which is weird, because I copied this from the SaaS health check script. So I guess we're not actually checking the monitors there either.

volokluev: should we... check it? 🤔

evanh: I'll look at the monitors involved and see. I think the answer is yes.

enochtangg (approved):

PR #6014, opened 2024-06-10
[CapMan] Add throttling to ReferrerGuardRail policy

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2793

The ReferrerGuardRail policy prevents referrers from sending too many queries by rejecting them once a referrer reaches a rejection threshold. I add a throttle threshold so the offending referrer will get throttled first before getting rejected. ReferrerGuardRailPolicy will half the threads when the number of concurrent queries goes over 50% of the rejection threshold for that referrer (e.g. subscriptions_executor rejects at 200 concurrent queries, we throttle at 100 queries)


snuba/query/allocation_policies/per_referrer.py
```diff
@@ -100,6 +100,12 @@ def _get_quota_allowance(
             query_id,
             rate_limit_params,
         )
+        assert (
+            rate_limit_params.concurrent_limit is not None
+        ), "concurrent_limit must be set"
+        num_threads = self._get_max_threads(referrer)
+        if rate_limit_stats.concurrent > rate_limit_params.concurrent_limit // 2:
+            num_threads = self._get_max_threads(referrer) // 2
```
volokluev, line 108: This ratio should be configurable rather than hardcoded

volokluev (commented):
good overall just one change

volokluev:
How are you planning to roll this out?

snuba/query/allocation_policies/per_referrer.py
```diff
@@ -63,6 +63,18 @@ def _additional_config_definitions(self) -> list[AllocationPolicyConfig]:
                 value_type=int,
                 default=-1,
             ),
+            AllocationPolicyConfig(
+                name="throttle_threshold",
+                description="The threshold at which we will decrease the number of threads (THROTTLED_THREADS) used to execute queries",
+                value_type=int,
+                default=50,
```
volokluev, line 70: avoid magic numbers, I know there's already one above here so let's fix that too

snuba/query/allocation_policies/per_referrer.py
```diff
@@ -100,6 +112,12 @@ def _get_quota_allowance(
             query_id,
             rate_limit_params,
         )
+        assert (
+            rate_limit_params.concurrent_limit is not None
+        ), "concurrent_limit must be set"
+        num_threads = self._get_max_threads(referrer)
+        if rate_limit_stats.concurrent > self.get_config_value("throttle_threshold"):
+            num_threads = self.get_config_value("throttled_threads")
```
volokluev, line 120: how does this work if the concurrent queries is overridden for the referrer?

xurui-c:
> How are you planning to roll this out?

In this PR, I set the throttle dividers to 1 so that the rejection threshold and the throttle threshold are the same, so effectively this PR will create the options on Snuba Admin without doing any throttling. Then, within S4S, I will adjust the throttle dividers. With this, I expect to see queries being throttled in DD under the metric `concurrent_queries_throttled`. I will change the throttle dividers to 0.5 afterwards.

snuba/query/allocation_policies/per_referrer.py
```diff
@@ -19,6 +19,11 @@
...
 
 _DEFAULT_MAX_THREADS = 10
+_DEFAULT_CONCURRENT_REQUEST_PER_REFERRER = 100
+_REFERRER_CONCURRENT_OVERRIDE = -1
+_REFERRER_MAX_THREADS_OVERRIDE = -1
+_REQUESTS_THROTTLE_DIVIDER = 1
+_THREADS_THROTTLE_DIVIDER = 1
 
 
 class ReferrerGuardRailPolicy(BaseConcurrentRateLimitAllocationPolicy):
```
onewland, line 29: I think this new behavior should be described in the doc comment below

snuba/query/allocation_policies/per_referrer.py
```diff
@@ -100,6 +117,18 @@ def _get_quota_allowance(
...
         )
+        assert (
+            rate_limit_params.concurrent_limit is not None
+        ), "concurrent_limit must be set"
+        num_threads = self._get_max_threads(referrer)
+        requests_throttle_threshold = self.get_config_value(
+            "default_concurrent_request_per_referrer"
+        ) // self.get_config_value("requests_throttle_divider")
+        if rate_limit_stats.concurrent > requests_throttle_threshold:
+            num_threads //= self.get_config_value("threads_throttle_divider")
```
onewland, line 128: I think it's less clean to mutate this variable, as opposed to setting it [once] conditionally

Additionally, if your divider is set to 2 and your max_threads is 1, then I think this logic might result in an error (`1 // 2` evaluates to 0 in python). However we want to handle this, there should be a test for it, or a guardrail that prevents the bad config from being used in production

onewland: I might suggest for simplicity that you just have a static throttled_thread_count

onewland (changes requested):
attached some suggestions for changes, but fundamentally the concept and approach seem sound

snuba/query/allocation_policies/per_referrer.py
xurui-c: I think the intention is to auto-adjust the warning or throttling threshold when the rejection threshold is adjusted (ie prevents human error like setting rejection threshold to 20 but warning threshold was left at 40). I used `max` to ensure we will always execute with at least 1 thread

volokluev (approved):

snuba/query/allocation_policies/per_referrer.py
volokluev: @onewland The static threshold becomes annoying in the case of overrides. Every time you override a referrer's concurrent queries you would also need to override it's throttle threshold.

onewland (approved):

PR #6025, opened 2024-06-12
feat: autogenerate add migration ui

kylemumma:
This PR is the interface users will use to generate migrations
```
snuba migrations generate path/to/storage
```
the code in this PR:
* get the storage file at that path
* gets the version of it at HEAD

I will use this to send to the diff functions that take the old and new storage as input and generate the actual migration

### DIFF
--- a/snuba/migrations/autogeneration/main.py
+++ b/snuba/migrations/autogeneration/main.py
@@ -1,15 +1,30 @@
 import os
 import subprocess
 
+import requests
+import yaml
 
-def generate(storage_path: str) -> None:
-    repo_path = subprocess.run(
-        ["git", "rev-parse", "--show-toplevel"], capture_output=True
-    ).stdout.decode("utf-8")
-    storage_path = os.path.abspath(os.path.expanduser(storage_path))
-    if not storage_path.startswith(repo_path):
-        raise ValueError("Storage path is not in the git repository")
 
-    rel_path = os.path.relpath(storage_path, repo_path)
-    old_storage = ""
-    new_storage = ""
+def generate(local_storage_path: str) -> tuple[str, str]:
+    local_repo_path = (
+        subprocess.run(["git", "rev-parse", "--show-toplevel"], capture_output=True)
+        .stdout.decode("utf-8")
+        .strip()
+    )
+    local_storage_path = os.path.abspath(os.path.expanduser(local_storage_path))
+    if not local_storage_path.startswith(local_repo_path):
+        raise ValueError(
+            f"Storage path '{local_storage_path}' is not in the git repository '{local_repo_path}'"
+        )
+
+    # get local storage
+    rel_storage_path = os.path.relpath(local_storage_path, local_repo_path)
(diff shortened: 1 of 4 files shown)


volokluev (approved):
nice this is actually pretty clean. helpful that snuba is open source :D

snuba/migrations/autogeneration/main.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # get local storage
+    rel_storage_path = os.path.relpath(local_storage_path, local_repo_path)
+    with open(os.path.join(local_repo_path, rel_storage_path), "r") as f:
+        new_storage = yaml.safe_load(f)
+
+    # get remote storage
+    REMOTE_SNUBA_REPO = "https://raw.githubusercontent.com/getsentry/snuba/master"
+    res = requests.get(f"{REMOTE_SNUBA_REPO}/{rel_storage_path}")
+    old_storage = yaml.safe_load(res.text)
```
onewland, line 28: Have you considered removing this remote git call and comparing whatever is unstaged with the local HEAD? This should be similar to your `subprocess.run` on line 10, with some slightly different arguments.

Some downsides of this approach, in order of descending concern to me:
- hard-codes a base branch, master, which might not be great when working on a new feature
- requires Github API to be up and for the user to have an active internet connection to run the command
- assumes Github API route remains the same over time, and doesn't require authentication
- assumes that we will always consider Github the canonical source of truth (unlikely to change, but I think the less we hard-code this dependency the better)

kylemumma: I considered using the local repo instead of remote. The reason I ended up deciding on remote was because I wanted "old storage" to be as close to what actually in clickhouse prod as possible. If users have their own local branch in some specific state this wouldnt be the case.

MeredithAnya: If we are using master it might get confusing if someone's local branch is not up to date. If someone else merged in a migration that you hadn't run yet, we wouldn't want your local migration to include those schema changes, right?

snuba/migrations/autogeneration/main.py
```diff
@@ -0,0 +1,37 @@
+import os
+import subprocess
+
+import requests
+import yaml
+
+
+def generate(local_storage_path: str) -> tuple[str, str]:
```
MeredithAnya, line 8: this only deals with migrations that are updating an existing schema yeah? Do we care about the initial migration for a storage, or are we not worry about that for now?

snuba/migrations/autogeneration/main.py
kylemumma: @MeredithAnya ok yeah I see what you're saying now... so do you agree with Oliver that comparing whatever is unstaged with the local HEAD might be a better approach?

snuba/migrations/autogeneration/main.py
kylemumma: For now we only care about updating existing. Initial migrations may be something we want to do in the future, but for now its scoped to just updates.

PR #6032, opened 2024-06-17
ref(spans): Create script for generating fake traces

evanh:
Create a script that generates fake traces for a set of projects.

All the traces for a single organization (1123) that spans 3 projects. Conceptually it is meant to
mimic a frontend, backend and service project.

The script can write to a file, output to stdout or produce to a kafka topic. It's all random, so
there are no guarantees of how many transactions/spans will actually appear in a given trace. There
is an option to run the script until a given number of spans are generated. Note this isn't exact,
the script will run until at least that many spans are generated.

### DIFF
--- a/scripts/send_spans.py
+++ b/scripts/send_spans.py
@@ -91,8 +91,15 @@ def create_span(
     duration_ms: int,
     start_timestamp: datetime.datetime,
     transaction_name: str,
+    add_datetimes: bool = False,
 ) -> Span:
     end_timestamp = start_timestamp + datetime.timedelta(milliseconds=duration_ms)
+    datetime_fields = {}
+    if add_datetimes:
+        datetime_fields = {
+            "start_datetime": start_timestamp.isoformat(),
+            "end_datetime": end_timestamp.isoformat(),
+        }
     return {
         "event_id": uuid.uuid4().hex,
         "organization_id": ORGANIZATION_ID,
@@ -123,6 +130,7 @@ def create_span(
                 "unit": "unit",
             },
         },
+        **datetime_fields,
     }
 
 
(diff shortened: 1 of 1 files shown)


untitaker (approved):
idk where this script is best put, but i think it could go eitehr into snuba or the sandbox. in the sandbox you have the advantage of building some adhoc deployment scripts too

scripts/send_spans.py
```diff
@@ -0,0 +1,427 @@
...
+            add_datetimes=add_datetimes,
+        )
+
+        to_split_or_not_to_split = random.random()
+        if span_count < 100 and to_split_or_not_to_split < 0.3:
+            subspans = []
+            subduration = 0
+            if project_id != 3 and to_split_or_not_to_split < 0.15:
+                if verbose:
+                    print(f"{prefix}T{project_id}")
```
volokluev, line 179: I feel like log level is a better way to do this

volokluev (approved):
just a few small comments

scripts/send_spans.py
```diff
@@ -0,0 +1,427 @@
...
+    ("tag1", partial(random.randint, 0, 10)),
+    ("tag2", lambda: str(uuid.uuid4())),
+    ("tag3", partial(random.randint, 0, 100000)),
+]
+
+Span = dict[str, Any]
+Producer = Callable[[list[Span], bool, bool], None]
+
+
+def custom_tags(project_id: int) -> dict[str, str]:
```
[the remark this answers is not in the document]
evanh: That's probably fair. I'll change it.

scripts/send_spans.py
evanh: I agree, I just stopped improving this basically.

PR #6034, opened 2024-06-17
Add additional information to policies

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2796

The end goal is to have a payload that looks like the following:
```
"quota_allowance": {
	"summary": {
		"threads_used": 2, # the actual amount of max_threads sent to clickhouse
		"throttled_by": {
			"policy": "BytesScannedRejectingPolicy",
			"throttle_threshold": 800,
			"quota_used": 900,
			"quota_units": "bytes",
			"suggestion": "scan fewer bytes n00b",
		},
		"rejected_by": {},
	}
	# policy details as before
	...
}
```
In order to accomplish this, this PR adds in the relevant information into policy definitions and the `QuotaAllowance` class, and will be followed up by a subsequent PR which will actually update the payload to the above.

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -764,7 +764,7 @@ class AllocationPolicy(ABC, metaclass=RegisteredClass):
                 },
             )
         if not self.is_enforced:
-            return QuotaAllowance(True, self.max_threads, {})
+            return QuotaAllowance(True, self.max_threads, {}, True)
         # make sure we always know which storage key we rejected a query from
         allowance.explanation["storage_key"] = str(self._storage_key)
         return allowance
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import os
+import sys
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field, replace
 from typing import Any, cast
@@ -96,6 +97,7 @@ class QuotaAllowance:
     # about what caused that action. Not currently well typed
     # because I don't know what exactly should go in it yet
     explanation: dict[str, JsonSerializable]
+    is_throttled: bool
 
     def to_dict(self) -> dict[str, Any]:
         return asdict(self)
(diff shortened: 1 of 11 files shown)


snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -314,3 +311,71 @@ def _update_quota_balance(
...
+            customer_tenant_value,
+        ) = self._get_customer_tenant_key_and_value(tenant_ids)
+        referrer = tenant_ids.get("referrer", "no_referrer")
+        if referrer in _PASS_THROUGH_REFERRERS:
+            return sys.maxsize
+        return self.__get_scan_limit(
+            customer_tenant_key, customer_tenant_value, referrer
+        )
+
+    def get_quota_used(self, tenant_ids: dict[str, str | int]) -> int:
```
xurui-c, line 337: @volokluev I'm not a fan of repeating this chunk of code

snuba/query/allocation_policies/__init__.py
```diff
@@ -799,6 +804,26 @@ def _update_quota_balance(
...
+        pass
+
+    @abstractmethod
+    def get_quota_units(self) -> str:
+        pass
+
+    @abstractmethod
+    def get_suggestion(self) -> str:
+        pass
+
```
volokluev, line 826: why is this necessary?

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -314,3 +311,71 @@ def _update_quota_balance(
...
+                    ],
+                ),
+            ]
+        )
+
+        used_quota = 0
+        for granted_quota in granted_quotas:
+            used_quota += granted_quota.granted
+
+        return used_quota
```
volokluev, line 375: Is this now running the calculation twice?

volokluev (commented):
It's not clear to me why you need to change anything besides the QuotaAllowance class

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
volokluev: so why are you doing it then?

xurui-c:
redesign

PR #6038, opened 2024-06-18
fix(docs): remove make command that no longer exists

IanWoodard:
When setting up my Snuba dev environment locally, I noticed that the make command in the docs no longer exists and has no specific replacement that I am aware of.


enochtangg (approved):
thanks!

PR #6041, opened 2024-06-18
Add additional information to policies

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2796

Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, the result of applying allocation policies in the internal API is simply accept/reject/throttle. In this PR, we are updating existing allocation policies to return additional metadata, most importantly to distinguish between queries that are accepted, rejected, or throttled.

Next steps: surface these new metadata fields in the [external API](https://getsentry.atlassian.net/browse/SNS-2797)
When all changes are finished, we should have a payload for Sentry that looks like:

```
"quota_allowance": {
	"summary": {
		"threads_used": 2, # the actual amount of max_threads sent to clickhouse
		"throttled_by": {
			"policy": "BytesScannedRejectingPolicy",
			"throttle_threshold": 800,
			"quota_used": 900,
			"quota_units": "bytes",
			"suggestion": "scan fewer bytes n00b",
		},
		"rejected_by": {},
	}
	# policy details as before
	...
}
```

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -25,6 +25,13 @@ CAPMAN_HASH = "capman"
 IS_ACTIVE = "is_active"
 IS_ENFORCED = "is_enforced"
 MAX_THREADS = "max_threads"
+NO_UNITS = "no_units"
+NO_SUGGESTION = "no_suggestion"
+CROSS_ORG_SUGGESTION = "cross org queries do not have limits"
+PASS_THROUGH_REFERRERS_SUGGESTION = (
+    "subscriptions currently do not undergo rate limiting in any way"
+)
+MAX_THRESHOLD = int(1e12)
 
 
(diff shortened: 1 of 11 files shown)


onewland (approved):
Looks pretty good to me. Some thoughts, that don't need to block merging:

1. It seems like `QuotaAllowance` is going to become more important for our users in the future. I think maybe it's important enough to have a doc comment explaining what the class is for (e.g. indicating the application of quotas against one request), and maybe a comment for each field that isn't totally self-explanatory
2. `QuotaAllowance.__eq__` is currently only using a few fields for comparison. I don't know if this matters at all, but `is_throttled` seems as relevant as `can_run` for comparing equality (assuming anybody actually does this ever)
3. I would consider centralizing the `suggestion`s into other constants, if tests are going to look for matches, that seems a bit cleaner to me than copying the string values around

PR #6043, opened 2024-06-19
ref(consumer): Add batch write timeout to Rust consumer

ayirr7:
Adds a new CLI arg to the Rust consumers, `batch_write_timeout_ms`. This parameter ends up in the batch writer (see `batch.rs`) where the timeout gets applied to the HTTP client that writes data to Clickhouse.  

The timeout would get applied to the entire time period of the connection + the time to receive a response from the server. 

The reason behind adding this timeout is to give us a configurable way to control the period of time that a consumer could _potentially_ stall after encountering transients Clickhouse connection errors. We have a current theory that such errors can affect specific partitions of a consumer, leading to slower-than-ideal recovery and a backlog on said partitions. Introducing a timeout could allow the partitions to recover more quickly and possibly prevent from such backlogs forming in the first place.

### DIFF
--- a/rust_snuba/src/strategies/clickhouse/batch.rs
+++ b/rust_snuba/src/strategies/clickhouse/batch.rs
@@ -381,13 +381,10 @@ mod tests {
     }
 
     #[test]
+    #[should_panic]
     fn test_write_no_time() {
         crate::testutils::initialize_python();
         let server = MockServer::start();
-        let mock = server.mock(|when, then| {
-            when.method(POST).path("/").body("{\"hello\": \"world\"}\n");
-            then.status(200).body("hi");
-        });
 
         let concurrency = ConcurrencyConfig::new(1);
         let factory = BatchFactory::new(
@@ -412,9 +409,7 @@ mod tests {
             ]))
             .unwrap();
 
-        assert!(concurrency.handle().block_on(batch.finish()).is_err());
-
-        mock.assert_hits(0);
+        concurrency.handle().block_on(batch.finish()).unwrap();
     }
 
     #[test]
--- a/rust_snuba/src/strategies/clickhouse/batch.rs
+++ b/rust_snuba/src/strategies/clickhouse/batch.rs
@@ -379,4 +379,77 @@ mod tests {
         // ensure there has not been any HTTP request
         mock.assert_hits(0);
     }
+
+    #[test]
+    fn test_write_no_time() {
+        crate::testutils::initialize_python();
+        let server = MockServer::start();
+        let mock = server.mock(|when, then| {
+            when.method(POST).path("/").body("{\"hello\": \"world\"}\n");
+            then.status(200).body("hi");
+        });
+
+        let concurrency = ConcurrencyConfig::new(1);
+        let factory = BatchFactory::new(
+            &server.host(),
+            server.port(),
+            "testtable",
+            "testdb",
+            &concurrency,
+            "default",
+            "",
+            true,
+            // pass in an unreasonably short timeout
+            // which prevents the client request from reaching Clickhouse
+            Some(Duration::from_millis(0)),
+        );
+
+        let mut batch = factory.new_batch();
+
+        batch
+            .write_rows(&RowData::from_encoded_rows(vec![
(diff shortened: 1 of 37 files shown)


rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -72,10 +73,18 @@ impl BatchFactory {
...
         let query = format!("INSERT INTO {table} FORMAT JSONEachRow");
 
-        let client = ClientBuilder::new()
-            .default_headers(headers)
-            .build()
-            .unwrap();
+        let client = if let Some(timeout_duration) = batch_write_timeout {
+            ClientBuilder::new()
+                .default_headers(headers)
+                .timeout(timeout_duration)
```
ayirr7, line 79: I saw some other options for timeout, e.g. https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout. Might be worth double checking which one we want to use?

untitaker: this one is fine. it applies to the entire lifecycle of the request.

however, I just realized, it also means that the timeout always needs to be higher than the max batch time...

untitaker (approved):

rust_snuba/src/strategies/clickhouse/batch.rs
untitaker: the timeout is for the entire lifecycle of the request, which is from when the batch is being created (=when the http request is being started and starts streaming) to when the batch is flushed (=when the last chunk is being written and we wait for and check the HTTP status code)

ayirr7:
Tested locally by starting up a consumer with `--batch-write-timeout-ms=0`, which triggered the timeout and caused the consumer to crash due to panic with the following error messages: 

```
{"timestamp":"2024-06-25T09:00:14.577750Z","level":"INFO","fields":{"message":"Caught error, terminating strategy"},"target":"rust_arroyo::processing"}
{"timestamp":"2024-06-25T09:00:14.578528Z","level":"INFO","fields":{"message":"Partitions to revoke: [Partition { topic: Topic(\"snuba-generic-metrics\"), index: 0 }]"},"target":"rust_arroyo::processing"}
{"timestamp":"2024-06-25T09:00:14.578653Z","level":"INFO","fields":{"message":"Joining strategy"},"target":"rust_arroyo::processing"}
{"timestamp":"2024-06-25T09:00:14.578771Z","level":"WARN","fields":{"message":"Timeout Some(0ns) reached while waiting for tasks to finish"},"target":"rust_arroyo::processing::strategies::reduce"}
{"timestamp":"2024-06-25T09:00:14.580411Z","level":"INFO","fields":{"message":"Partition revocation complete."},"target":"rust_arroyo::processing"}
{"timestamp":"2024-06-25T09:00:14.849284Z","level":"ERROR","fields":{"message":"Strategy(error sending request for url (http://127.0.0.1:8123/?load_balancing=in_order&insert_distributed_sync=1&query=INSERT+INTO+generic_metric_distributions_raw_local+FORMAT+JSONEachRow): operation timed out\n\nCaused by:\n    operation timed out\n\nStack backtrace:\n   0: std::backtrace_rs::backtrace::libunwind::trace\n             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/ ......
```

@untitaker If I understand correctly, this would be the desired behavior? Locally, the panic also puts the consumer in a bad state, where if I try to start it again, it crashes again (but this may just be due to some persistent local state).

rust_snuba/src/strategies/clickhouse/batch.rs
ayirr7: Added a constraint in `consumer.rs` to account for this. It makes sure that batch timeout >= batch time, if it is not, then the batch timeout passed in is silently discarded.

untitaker:
>  it crashes again without me sending any data through (but this may just be due to some persistent local state

it may have some unprocessed messages from the previous crash, as it didn't commit offsets

anyway, this behavior is fine, it now doesn't hang anymore

PR #6044, opened 2024-06-20
feat(capman): Replace BytesScannedWindow policy with BytesScannedRejectingPolicy for spans, replays

volokluev:
We are facing load issues on spans and replays due to excessive bytes scanned. Add the policy that has been successful on errors in mitigating this load and preventing abusive actors. 

Remove the BytesScannedWindow policy because it has not mitigated these load spikes and the BytesScannedRejectingPolicy also throttles now

### DIFF
--- a/snuba/datasets/configuration/replays/storages/replays.yaml
+++ b/snuba/datasets/configuration/replays/storages/replays.yaml
@@ -190,15 +190,6 @@ allocation_policies:
         - project_id
       default_config_overrides:
         is_enforced: 0
-  - name: BytesScannedWindowAllocationPolicy
-    args:
-      required_tenant_types:
-        - organization_id
-        - referrer
-      default_config_overrides:
-        is_enforced: 1
-        throttled_thread_number: 1
-        org_limit_bytes_scanned: 100000
   - name: ReferrerGuardRailPolicy
     args:
       required_tenant_types:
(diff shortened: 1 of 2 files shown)


snuba/datasets/configuration/replays/storages/replays.yaml
```diff
@@ -190,19 +190,19 @@ allocation_policies:
         - project_id
       default_config_overrides:
         is_enforced: 0
-  - name: BytesScannedWindowAllocationPolicy
```
ayirr7, line 193: For my understanding, what's the reason for removing this? Would it be conflicting with the `BytesScannedRejectingPolicy`?

volokluev: no it's just not really useful to have both of them. We're going to be deprecating the BytesScannedWindow policy 

https://www.notion.so/sentry/Increasing-visibility-into-the-Capacity-Management-System-c16b375ed2ee43c19cd2fa163fc93332?pvs=4#185824a1f34b4ff6b31fe1c6ea872aca

volokluev: Just thought I'd hit two birds with one stone. Especially since that policy did not prevent load spikes anyways

ayirr7: Makes sense, thanks!

ayirr7 (approved):

PR #6045, opened 2024-06-20
Relay query quota information to Sentry

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2797

Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, the result of applying allocation policies in the internal API is simply accept/reject/throttle. In a [previous PR](https://github.com/getsentry/snuba/pull/6041), we added internal API to keep track of additional metadata. In this PR, we relay the additional metadata to Sentry by adding them to the HTTP payload. It is then up to product engineers to use this information however they prefer to.

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import random
+import typing
 import uuid
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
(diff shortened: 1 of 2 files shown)


onewland (approved):
Looks good to me, are we planning to change snuba-sdk to convert this into span annotations of some kind or is this sufficient to surface this info in Sentry?

xurui-c:
The [next ticket](https://getsentry.atlassian.net/browse/SNS-2798) is to add to current spans

PR #6048, opened 2024-06-21
fix(group-attributes): Drop group_first_release_id column

snigdhas:
Dropping this column because the type is wrong. Previous PR to change the column type failed and had to be reverted.


ayirr7:
Can you double check that https://github.com/getsentry/snuba/pull/6039 was rolled out everywhere? Thanks!

snuba/migrations/0004_drop_first_release_id_from_group_attributes.py
```diff
@@ -0,0 +1,51 @@
...
+        ]
+
+    def backwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=StorageSetKey.GROUP_ATTRIBUTES,
+                table_name="group_attributes_dist",
+                column=Column(
+                    "group_first_release_id",
+                    UUID(Modifiers(nullable=True)),
```
ayirr7, line 36: Just checking here. Would UUID be the desired column type for the reverse operation (i.e. adding back the new column)? Or is UInt64 the correct type?

snigdhas: It sounds like the type change didn't work? so I assumed it wasn't safe to rollback to that

snigdhas:
> Can you double check that https://github.com/getsentry/snuba/pull/6039 was rolled out everywhere? Thanks!

Changes were rolled out in [this deploy](https://deploy.getsentry.net/go/pipelines/value_stream_map/deploy-snuba-us/587)

snuba/migrations/0004_drop_first_release_id_from_group_attributes.py
ayirr7: Ok, we can rollback to UUID. I suppose when we add back the column, we'll add it with the Uint64 type?

snigdhas: yeah, that's the plan!

ayirr7 (approved):

PR #6052, opened 2024-06-23
surface CapMan summary in case of AllocationPolicyViolations

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2798

Allocation policies are our mechanism for doing traffic management for Snuba queries. On Sentry issues webpage, the [metadata regarding allocation policies](https://github.com/getsentry/snuba/pull/6041) are not surfaced.

Screenshot 2024-06-22 at 8 02 53 PM

In this PR, we relay the metadata to Sentry issues webpage so that they are displayed in that same section.

### DIFF
--- a/snuba/query/allocation_policies/__init__.py
+++ b/snuba/query/allocation_policies/__init__.py
@@ -166,39 +166,32 @@ class AllocationPolicyViolations(SerializableException):
 
     @property
     def quota_allowance(self) -> dict[str, dict[str, Any]]:
-        return {
-            k: v
-            for k, v in self._quota_allowances_and_summary.items()
-            if k != "summary"
-        }
+        return cast(
+            dict[str, dict[str, Any]], self.extra_data.get("quota_allowances", {})
+        )
 
     @property
     def summary(self) -> dict[str, Any]:
-        return {
-            k: v
-            for k, v in self._quota_allowances_and_summary.items()
-            if k == "summary"
-        }
-
-    @property
-    def _quota_allowances_and_summary(self) -> dict[str, dict[str, Any]]:
-        return cast(
-            dict[str, dict[str, Any]],
-            self.extra_data.get("quota_allowances_and_summary", {}),
-        )
+        summary = self.extra_data.get("summary")
+        assert summary is not None
+        return {"summary": summary}
 
     @classmethod
     def from_args(
-        cls, quota_allowances_and_summary: dict[str, Any]
+        cls,
+        quota_allowances: dict[str, QuotaAllowance],
+        summary: dict[str, Any],
     ) -> "AllocationPolicyViolations":
-        obj = cls(
+        quota_allowances_dict = {
+            key: quota_allowance.to_dict()
+            for key, quota_allowance in quota_allowances.items()
+        }
+
+        return cls(
             "Query on could not be run due to allocation policies",
-            quota_allowances_and_summary={
-                key: quota_allowance
-                for key, quota_allowance in quota_allowances_and_summary.items()
-            },
+            quota_allowances=quota_allowances_dict,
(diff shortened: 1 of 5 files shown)


snuba/query/allocation_policies/__init__.py
```diff
@@ -158,29 +158,47 @@ class AllocationPolicyViolations(SerializableException):
...
     def quota_allowance(self) -> dict[str, dict[str, Any]]:
+        return {
+            k: v
+            for k, v in self._quota_allowances_and_summary.items()
+            if k != "summary"
+        }
+
+    @property
+    def summary(self) -> dict[str, Any]:
+        return {
```
evanh, line 177: Not sure why the loop is necessary to find a single key. 

```
summary = self._quote_allowances_and_summary.get("summary")
return {"summary": summary} if summary else {}
```

snuba/web/db_query.py
```diff
@@ -853,7 +853,7 @@ def _apply_allocation_policies_quota(
     Sets the resource quota in the query_settings object to the minimum of all available
     quota allowances from the given allocation policies.
     """
-    quota_allowances: dict[str, QuotaAllowance] = {}
```
evanh, line 856: Why was this changed?

evanh (commented):
There seems to be work done here to have two concepts (QuotaAllowance and Summary) in one dictionary. Why not have them be separate attributes?

xurui-c:
I originally wanted to abstract away the fact that summary is an addition to quota allowances, since the intention is that any time quota allowances is retrieved, a summary should come with it. I updated pr to be your approach since I believe it's cleaner and easier to understand

snuba/query/allocation_policies/__init__.py
```diff
@@ -170,16 +170,25 @@ def quota_allowance(self) -> dict[str, dict[str, Any]]:
             dict[str, dict[str, Any]], self.extra_data.get("quota_allowances", {})
         )
 
+    @property
+    def summary(self) -> dict[str, Any]:
+        return cast(dict[str, Any], self.extra_data.get("summary"))
```
evanh, line 175: Is there a case where the `summary` key won't be in the `extra_data`? If so, I don't think this typing holds true. In that case the function will return `None`.

xurui-c: No, there will always be a summary

xurui-c: I added an assertion

snuba/query/allocation_policies/__init__.py
```diff
@@ -158,28 +158,33 @@ class AllocationPolicyViolations(SerializableException):
     """
 
     def __str__(self) -> str:
-        return f"{self.message}, details: {self.violations}"
+        return f"{self.message}, info: {self.violations | self.summary}"
```
volokluev, line 161:
```suggestion
        return f"{self.message}, info: {"violations" self.violations, "summary": self.summary}"
```

snuba/query/allocation_policies/__init__.py
```diff
@@ -158,28 +158,33 @@ class AllocationPolicyViolations(SerializableException):
...
     def quota_allowance(self) -> dict[str, dict[str, Any]]:
         return cast(
             dict[str, dict[str, Any]], self.extra_data.get("quota_allowances", {})
         )
 
+    @property
+    def summary(self) -> dict[str, Any]:
+        summary = self.quota_allowance.get("summary")
+        assert summary is not None
+        return {"summary": summary}
```
volokluev, line 178:
```suggestion
       return self.quota_allowance.get("summary", {})
```

volokluev (approved):
just some small changes but otherwise looks good

PR #6054, opened 2024-06-24
feat(metrics_v2): Build entities and storages for metrics_v2

nikhars:
This PR builds the entities and storages for the next iteration of generic metrics which does not have the indexer. What I did is copy/paste the dataset, storages and entities from the generic_metrics configuration and created a new generic_metrics_v2 from it and modified fields according to what is needed without the indexer. This included
- Renaming metric_id to metric_mri and changing it from uint to String type
- Changing tag_key to String type
- Get rid of the granularity processor to avoid the granularity mappings. We can save the granularity as is in the table

It would be easier to review the diff in this [gist](https://gist.github.com/nikhars/3f9141e6b2f156fdb10c3437f41c9156#file-gistfile1-txt-L1669) rather than reviewing the full PR.


snuba/migrations/groups.py
```diff
@@ -177,6 +179,16 @@ def __init__(
         storage_sets_keys={StorageSetKey.PROFILE_CHUNKS},
         readiness_state=ReadinessState.PARTIAL,
     ),
+    MigrationGroup.GENERIC_METRICS_V2: _MigrationGroup(
```
enochtangg, line 182: I think it makes sense to have a storage_set per metric type. This gives us the ability to decide whether or we want 1-N clickhouse clusters for generic metrics. One implication I'd point out of this pattern (storage set per metric type in a single migration loader) is the inability to gradually roll out a new table in this migration group. For example, though maybe unlikely, say sentry decides we'd like to support a new metric type (RATE? HISTOGRAM?), if we create a new storage_set for this metric type and write the new table migration in this `GENERIC_METRICS_V2` loader, then it will automatically inherit its deployment state of this loader and go straight to production (SaaS, DE, ST, etc.). But to be clear, this is exactly how our current implementation of generic_metrics behaves today and we've managed this use-case with rolling out gauges in the past.

enochtangg (commented):
The general structure of these components makes sense to me. Had one comment.

snuba/migrations/groups.py
nikhars: That is a valid point. I had completely forgotten about this limitation of migrations. Will change the storage set to be different per metric type.

snuba/datasets/configuration/generic_metrics_v2/entities/gauges.yaml
```diff
@@ -0,0 +1,161 @@
...
+    },
+    {
+      name: count,
+      type: AggregateFunction,
+      args: { func: count, arg_types: [{ type: UInt, args: { size: 64 } }] },
+    },
+    {
+      name: last,
+      type: AggregateFunction,
+      args: { func: count, arg_types: [{ type: Float, args: { size: 64 } }] },
```
ayirr7, line 49: This was a typo I made on the original gauges entity. If I understand correctly, these args only affects the name of the aggregate function and name of types that's being passed around (not the actual Clickhouse aggregate function being used under the hood). So I think it should be

```
{
          func: argMax,
          arg_types: [{ type: Float, args: { size: 64 } }, { type: DateTime }],
}
```

snuba/datasets/configuration/generic_metrics_v2/entities/gauges.yaml
```diff
@@ -0,0 +1,161 @@
...
+    },
+    {
+      name: sum,
+      type: AggregateFunction,
+      args: { func: sum, arg_types: [{ type: Float, args: { size: 64 } }] },
+    },
+    {
+      name: count,
+      type: AggregateFunction,
+      args: { func: count, arg_types: [{ type: UInt, args: { size: 64 } }] },
```
ayirr7, line 44: Same thing here, this was a typo. Should be

`{ func: sum, arg_types: [{ type: UInt, args: { size: 64 } }] }`

since the aggregate count would be the sum of the individual counts sent by individual gauges.

PR #6064, opened 2024-06-25
dev(rate-limits): up concurrent rate limit for similarity backfill

JoshFerge:
we are simply doing point queries, so should be able to do a bit more concurrently without issues.

### DIFF
--- a/snuba/datasets/configuration/events/storages/errors_ro.yaml
+++ b/snuba/datasets/configuration/events/storages/errors_ro.yaml
@@ -264,8 +264,8 @@ allocation_policies:
         is_active: 1
       cross_org_referrer_limits:
         getsentry.tasks.backfill_grouping_records:
-          max_threads: 2
-          concurrent_limit: 4
+          max_threads: 10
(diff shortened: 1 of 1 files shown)


snuba/datasets/configuration/events/storages/errors_ro.yaml
```diff
@@ -264,8 +264,8 @@ allocation_policies:
         is_active: 1
       cross_org_referrer_limits:
         getsentry.tasks.backfill_grouping_records:
-          max_threads: 2
-          concurrent_limit: 4
+          max_threads: 10
```
volokluev, line 267: if you're doing point queries you don't need this to be this high

volokluev, line 267:
```suggestion
          max_threads: 5
```

JoshFerge: 👍🏼

enochtangg (approved):

PR #6067, opened 2024-06-25
fix(ratelimits) up rate limit for events for backfill as well

JoshFerge:
in this PR https://github.com/getsentry/snuba/pull/6064 we upped the rate limit for this referrer, but need to do it on the events dataet as well. i'm not actually sure how the errors_ro gets hit, but it seems our query in prod hits events.


JoshFerge:
Fixes https://sentry.sentry.io/issues/5536030538/?project=1&referrer=performance-related-issues-issue-stream&statsPeriod=1h

PR #6071, opened 2024-06-27
feat(replays): Add materialization YAML definitions

cmanallen:
Add entity and storage YAML files for `replays_aggregated` materialization.

Previous PR which added the migration: https://github.com/getsentry/snuba/pull/5708.


tests/datasets/test_entity_factory.py
```diff
@@ -29,6 +29,7 @@
     EntityKey.PROFILES,
     EntityKey.FUNCTIONS,
     EntityKey.REPLAYS,
+    EntityKey.REPLAYS_AGGREGATED,
```
aliu39, line 32: will this test against the clickhouse table?

aliu39 (approved):

tests/datasets/test_entity_factory.py
cmanallen: I have no idea what this test does.  But now that the YAML is defined the test requires it.

aliu39: kk just wondering because I def did not review every column of the schema lol

snuba/datasets/configuration/replays/entities/aggregated.yaml
```diff
@@ -0,0 +1,343 @@
...
+    {
+      name: user_name,
+      type: AggregateFunction,
+      args:
+        {
+          func: anyIf,
+          arg_types:
+            [
+              { type: String, args: { schema_modifiers: [nullable] } },
+              { type: UInt, args: { schema_modifiers: [nullable], size: 8 } },
```
aliu39, line 307: Is this for byte array? A single char?

cmanallen: Its the data-type which contains the result of the `-If` condition in the materialized view.

aliu39: Oh.. my bad I totally wasn't thinking correctly 😓

aliu39 (approved):

nikhars:
Mind adding some description to your PR? Without a PR description I have no idea what you are trying to achieve here. Like, I do not know whether the migrations have already been created or are not needed. If migrations are already created, can you link the PR in which they were created? Providing context helps the reviewer not look around for context by themselved.

cmanallen:
@nikhars Added.

snuba/datasets/configuration/replays/entities/aggregated.yaml
```diff
@@ -0,0 +1,343 @@
...
+            ],
+        },
+    },
+  ]
+required_time_column: to_hour_timestamp
+storages:
+  - storage: replays_aggregated
+    is_writable: true
```
nikhars, line 328: Don't believe this should be writable since the materialization is filling in the table.

snuba/migrations/parse_schema.py
```diff
@@ -46,7 +46,7 @@
     enum_val         = ~r"\d+"
     agg              = "AggregateFunction" open_paren space* agg_func space* comma space* agg_types space* close_paren
     agg_func         = ~r"[a-zA-Z0-9]+\([a-zA-Z0-9\,\.\s]+\)|[a-zA-Z0-9]+"
-    agg_types        = (primitive (space* comma space*)?)*
+    agg_types        = (type (space* comma space*)?)*
```
nikhars, line 49: Why does this need to change?

nikhars (commented):
Did you know about Snuba's capability to directly query storages instead of entities? It allows you to bypass creating entities altogether. Have some very rudimentary example here https://getsentry.github.io/snuba/language/snql.html#match.

Some additional documentation can be found here https://www.notion.so/sentry/Creating-a-Dataset-Locally-de21b3c9a14746baa9cb011e63328072?pvs=4#442278dca04a42159128040945287b58

An example dataset is here https://github.com/getsentry/snuba/blob/nikhars/feat/metrics-v2/snuba/datasets/configuration/profiles/storages/chunks.yaml

Your call whether to use that model or not. Wasn't sure whether you were aware or not.

snuba/migrations/parse_schema.py
cmanallen: The `Nullable()` wrapping.

cmanallen:
@nikhars If I remove the entity definition does anything need to change in my sentry code?  E.g. do I need to change this `Entity("replays_aggregated")` to something like `Storage("replays_aggregated")`?

cmanallen:
@nikhars It looks like the Sentry code needs to be updated.  `Entity` -> `Storage`.  This is what I found for the profile_chunks storage: `Storage(StorageKey.ProfileChunks.value)` 👍

nikhars (approved):
All my pending comments are non-blocking. Hence approving the PR

snuba/migrations/parse_schema.py
cmanallen: That would have been preferred.  At the time I didn't know how to accomplish that.  I'm guessing in the materialized view I'd do something like `if(isNull(column), a, b)`?  I don't really want to back fill for another 90 days.  Do you have familiarity with migrations from nullable to not nullable and the performance implications?

nikhars: >  I'm guessing in the materialized view I'd do something like if(isNull(column), a, b)?

Correct

> I don't really want to back fill for another 90 days. Do you have familiarity with migrations from nullable to not nullable and the performance implications?

That makes sense to me. I didn't know that the aggregated table is already being written to for such a long time. You won't be able to change a nullable column to non-nullable unfortunately. The only way I can think of is to  create new columns in aggregate tables which are non-nullable. And in the materialization write to both the old and new columns. And when you have both you can update the query processors to add a column mapper which maps old columns to new columns. But this is a heavy process. Might as well go with what you have now and revisit if you still have problems.

PR #6086, opened 2024-07-02
feat: create distributed table search_issue_v2 for SEARCH_ISSUES_TMP storage set

kylemumma:
this migration creates an up to date version of search_issues_dist_v2 on SEARCH_ISSUES_TMP storage set.

this storage set uses the same query node as errors/group_attributes in every region and will therefore enable joins between them and search issue.

## design decisions
* I decided to put it in the existing migration group for search_issues rather than creating a new one for search_issues_tmp. This is bc the migration implicitly assumes the existance of tables on the local node, which means search_issue migrations need to run first.
* In order to create the table I was originally just gunna copy the create table query from snuba admin and turn it into a migration. But converting between the CH query and the migration python object turned out to be too hard. So instead I copied all the existing migrations for search_issues_v2, and removed the local table parts.

## other thoughts
everything related to these storage set thingy require that search issue and group attribute are behind the same query node. but this is 100% defined in ops and nowhere in snuba. so the storage set abstraction is kind of insufficient. but it doesnt matter rn in practice, just thoughts.


MeredithAnya:
@kylemumma so what are the order of operations that are going to happen with this PR? 

1. Make query nodes aware of other storage nodes https://github.com/getsentry/ops/pull/11245/
2. Add snuba cluster config for storage set `search_issues_tmp` https://github.com/getsentry/ops/pull/11246
3. Merge/deploy this PR

Is the above correct, and if so is there anything to be concerned about between 2 and 3? If not, what needs to change/be added?

kylemumma:
@MeredithAnya Yes you're correct. (And I think the order of 1 and 2 is interchangeable, they just both need to happen before 3).

The only potential concern between 2 and 3 that I can think of would be the migration failing if it cant find the mapping properly. But I don't think that will happen and if it does it shouldnt be a big deal. Also since all the changes we are doing is additions and not modifying anything, I don't think there is much risk of messing anything up.

PR #6087, opened 2024-07-02
feat: create search_issues_dist_v2 on same query node as group_attributes

kylemumma:
This migration targets the new SEARCH_ISSUES_TMP storage set.
In OPS repo this storage set has a the same distributed cluster as group_attributes and local cluster of search_issues_stoage.
This migration will create-if-not-exists an up to date search_issues_dist_v2 on the distributed cluster (same as group attributes).

This will enable the larger goal of being able to run a join on search_issues and group_attributes on that query node.

## next steps
1. update storage-set-to-cluster mapping in ops to make search_issues go to group_attributes query
2. delete search_issues_tmp

## design choices
We make the decision to not include backwards ops in this migration.

The problem we ran into with the backwards ops is the table has the same name so the backwards ops for the new storage sets migration would override things for SEARCH_ISSUE storage set migration causing missing table error.

1 approach would be to try to use a new table name for the migration. This was explored and deemed would not work. This is because we need to do renaming and rerouting which occur in 2 separate repos, meaning could not happen without downtime. Thus the renaming would need to happen on both SEARCH_ISSUES and SEARCH_ISSUES_TMP storage set, but if this were done the same issue of naming would happen as initially.

Another is a CodeMigration, this is the only viable alternative we could think of. We determined that due to the temporary nature of this migration its ok to not have backwards ops. The consequences are this migration will need to be reversed "by hand" using ddl sql commands on the infra. This will not impact self hosted.


snuba/migrations/groups.py
```diff
@@ -177,6 +182,13 @@ def __init__(
         storage_sets_keys={StorageSetKey.PROFILE_CHUNKS},
         readiness_state=ReadinessState.PARTIAL,
     ),
+    MigrationGroup.SEARCH_ISSUES_TMP: _MigrationGroup(
+        loader=SearchIssuesTmpLoader(),
+        storage_sets_keys={
+            StorageSetKey.SEARCH_ISSUES_TMP,
+        },
+        readiness_state=ReadinessState.PARTIAL,
```
kylemumma, line 190: is this right?

MeredithAnya: I think you'd want `COMPLETE`?

MeredithAnya:
@kylemumma now that you added a new migration group, do we have guarantees that the search issues local table is created before the search issues tmp dist table is created?

kylemumma:
@MeredithAnya good point, we dont have the guarantee anymore, ill move it back into the same group

MeredithAnya (approved):

PR #6088, opened 2024-07-02
sets throttle policy correctly

xurui-c:
The throttle policy should be the policy with the smallest number of threads. The existing logic simply takes the last throttling policy in the `for-loop`

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -869,7 +869,6 @@ def _apply_allocation_policies_quota(
                 allowance = allocation_policy.get_quota_allowance(
                     attribution_info.tenant_ids, query_id
                 )
-                num_threads = min(num_threads, allowance.max_threads)
                 can_run &= allowance.can_run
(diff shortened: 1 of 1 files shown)


snuba/web/db_query.py
```diff
@@ -869,18 +869,19 @@ def _apply_allocation_policies_quota(
...
                     "quota_allowance",
                     quota_allowances[allocation_policy.config_key()],
                 )
                 if allowance.is_throttled:
-                    throttle_quota_and_policy = _QuotaAndPolicy(
-                        quota_allowance=allowance,
-                        policy_name=allocation_policy.config_key(),
-                    )
+                    if allowance.max_threads < num_threads:
+                        throttle_quota_and_policy = _QuotaAndPolicy(
```
ayirr7, line 880: Is this basically saying to set some sort of query throttling if we want to allocate a lower max threads ?

xurui-c: exactly

ayirr7 (approved):

PR #6092, opened 2024-07-08
ref(snuba-deletes): add DeletionSettings

MeredithAnya:
**context:**
Embarking on adding (lightweight) deletes to Snuba. The first storage we want to add this to is search issues, since right now you cannot delete via the Sentry application. 

The first step is to add some definitions on the storage that we will use in the delete pipeline (which will be a new endpoint we add later). 

**changes:**
This PR adds some deletion settings that are needed. they are:

* `is_enabled`: whether deletes are enabled for the storage. If no deletion settings are provided in the yaml file, we default to an empty deletion settings state (where `is_enabled` is `0`)
* `tables`: list of tables that we delete from. in some cases there will be more than one table to delete from - as is the case for storages that have a materialized view and where both the local and aggregated tables are queried from the product. In the empty deletion settings state tables is an empty list
* `max_rows_to_delete`: how many rows you can delete in one request. we want some limits to prevent excessive load on clickhouse. defaults to 1000 if deletions are enabled but `max_rows_to_delete` is not specified. in the empty deletion settings state `max_rows_to_delete` is `0`.

### DIFF
--- a/snuba/datasets/configuration/issues/storages/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/storages/search_issues.yaml
@@ -114,6 +114,12 @@ query_processors:
         contexts:
           trace.trace_id: trace_id
 
+deletion_settings:
+  is_enabled: 0
+  max_rows_to_delete: 1000
+  tables:
+    - search_issues_local_v2
+
 mandatory_condition_checkers:
   - condition: ProjectIdEnforcer
 
--- a/snuba/datasets/configuration/json_schema.py
+++ b/snuba/datasets/configuration/json_schema.py
@@ -538,6 +538,25 @@ ENTITY_JOIN_RELATIONSHIPS = {
     },
 }
 
+DELETION_SETTINGS_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "is_enabled": {
+            "type": "integer",
+        },
+        "tables": {
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "Names of the tables to delete from.",
+        },
+        "max_rows_to_delete": {
+            "type": "integer",
+        },
+    },
+    "required": ["is_enabled", "tables"],
+    "additionalProperties": False,
+}
+
 # Full schemas:
 
 V1_READABLE_STORAGE_SCHEMA = {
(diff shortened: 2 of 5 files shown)


snuba/datasets/configuration/storage_builder.py
```diff
@@ -78,22 +80,29 @@ def __build_readable_storage_kwargs(config: dict[str, Any]) -> dict[str, Any]:
...
-                **{
-                    **policy.get("args", {}),
-                    "storage_key": storage_key.value,
-                }
-            )
-            for policy in config[ALLOCATION_POLICIES]
-        ]
-        if ALLOCATION_POLICIES in config
-        else [],
+        ALLOCATION_POLICIES: (
```
MeredithAnya, line 93: this got re-formatted

vartec (approved):

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -114,6 +114,12 @@ query_processors:
         contexts:
           trace.trace_id: trace_id
 
+deletion_settings:
+  is_enabled: 0
```
xurui-c, line 118: should is_enabled = 1 for true?

snuba/datasets/configuration/json_schema.py
```diff
@@ -538,6 +538,25 @@ def registered_class_array_schema(
...
+            "type": "integer",
+        },
+    },
+    "required": ["is_enabled", "tables"],
+    "additionalProperties": False,
+}
+
 # Full schemas:
 
 V1_READABLE_STORAGE_SCHEMA = {
```
xurui-c, line 562: Should `deletion_settings` be added to `V1_WRITABLE_STORAGE_SCHEMA`?

snuba/datasets/storage.py
```diff
@@ -157,6 +160,9 @@ def get_mandatory_condition_checkers(self) -> Sequence[ConditionChecker]:
     def get_allocation_policies(self) -> list[AllocationPolicy]:
         return self.__allocation_policies or super().get_allocation_policies()
 
+    def get_deletion_settings(self) -> DeletionSettings:
+        return self.__deletion_settings
+
```
xurui-c, line 165: should we have this method for WritableTableStorage as well?

xurui-c (commented):
Left some quick queestions

snuba/datasets/configuration/issues/storages/search_issues.yaml
MeredithAnya: I was going to leave this as `0` until we were ready to test it out (and until we had the killswitch implemented)

xurui-c (approved):

PR #6095, opened 2024-07-08
ref(snuba-deletes): Skeleton for DeletionProcessor

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2819

context:
Embarking on adding (lightweight) deletes to Snuba. The first storage we want to add this to is search issues, since right now you cannot delete via the Sentry application.

The first step is to add some definitions on the storage that we will use in the delete pipeline (which will be a new endpoint we add later).

changes:
This PR sets up the skeleton for adding a DeletionProcessor: ColumnFilterProcessor. DeletionProcessors are responsible for validating that a query is a valid deletion query against the storage


snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -128,3 +128,10 @@ stream_loader:
   default_topic: generic-events
   commit_log_topic: snuba-generic-events-commit-log
   dlq_topic: snuba-dead-letter-generic-events
+
+deletion_processors:
+  - processor: ColumnFilterProcessor
+    args:
+      column_filters:
+        - project_id
+        - replay_id
```
MeredithAnya, line 137: this is for search issues so it won't be `replay_id` anymore

xurui-c: We will add occurrence_id

JoshFerge (commented):
nit: please title the pull request similar to the guidelines here: https://develop.sentry.dev/commit-messages/

snuba/query/processors/physical/__init__.py
```diff
@@ -41,6 +41,10 @@ def process_query(self, query: Query, query_settings: QuerySettings) -> None:
     def get_from_name(cls, name: str) -> "ClickhouseQueryProcessor":
         return cast("ClickhouseQueryProcessor", cls.class_from_name(name))
 
+    @classmethod
+    def is_valid_delete_query(self, query: Query) -> bool:
+        return True
```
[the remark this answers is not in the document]
MeredithAnya: we don't want a @classmethod here since the method depends on an instance of the class, not the class itself. I think actually it might be better to have this kind of check on the `Query` itself, something like `is_delete` but I'm not sure, for now I'd actually leave it out of this PR

snuba/query/processors/physical/column_filter_processor.py
```diff
@@ -0,0 +1,13 @@
+from typing import Sequence
+
+from snuba.clickhouse.query import Query
+from snuba.query.processors.physical import ClickhouseQueryProcessor
+from snuba.query.query_settings import QuerySettings
+
+
+class ColumnFilterProcessor(ClickhouseQueryProcessor):
+    def __init__(self, column_filters: Sequence[str]) -> None:
+        pass
```
MeredithAnya, line 10: probably want to at least set the column_filters 
`self.__column_filters = column_filters`

snuba/query/processors/physical/column_filter_processor.py
```diff
@@ -0,0 +1,13 @@
+from typing import Sequence
+
+from snuba.clickhouse.query import Query
+from snuba.query.processors.physical import ClickhouseQueryProcessor
+from snuba.query.query_settings import QuerySettings
+
+
+class ColumnFilterProcessor(ClickhouseQueryProcessor):
```
MeredithAnya, line 8: wanna add a docstring explaining this processor? doesn't have to be too long

MeredithAnya (approved):

JoshFerge (approved):

PR #6098, opened 2024-07-10
feat(replays): Set platform as empty string if platform is not provided

cmanallen:
Previously it was set to `javascript` as that was the only platform replay was available on.  With the addition of mobile replays this is no longer true.  The query in `sentry` already uses the `anyIf` operator and filters out empty values.

Related: https://github.com/getsentry/sentry/issues/74074

### DIFF
--- a/rust_snuba/src/processors/replays.rs
+++ b/rust_snuba/src/processors/replays.rs
@@ -53,7 +53,7 @@ pub fn deserialize_message(
                 event_hash: click.event_hash,
                 offset,
                 partition,
-                platform: "javascript".to_string(),
+                platform: "".to_string(),
                 project_id: replay_message.project_id,
                 replay_id: replay_message.replay_id,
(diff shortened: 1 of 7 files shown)


michellewzhang:
does this mean that if someone were to refresh the page on a mobile replay (referencing the video @romtsn sent), the platform in the tags panel might sometimes disappear (instead of changing to javascript)?

cmanallen:
@michellewzhang No if a non-empty value is set it will always be returned.  There should only be one non-empty value.  If there are two (in the case @romtsn identified) then one of the non-empty values will be chosen at "random".

michellewzhang (approved):

ayirr7 (approved):
Might also be a good idea to update any schema examples in the sentry-kafka-schemas repo if needed.

cmanallen:
@ayirr7 All good.  Platform is not provided on any of the affected events.

PR #6100, opened 2024-07-10
feat(snuba-deletes): ColumnFilterProcessor

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2820

In order to support lightweight deletes in Snuba, we add the `ColumnFilterProcessor`. The role of the ColumnFilterProcessor is to make sure that the columns in the query are exactly the columns listed in the storage definition. The processor should throw an exception if there are columns missing or if there are columns that are not included in the definition.

### DIFF
--- a/snuba/query/processors/physical/column_filter_processor.py
+++ b/snuba/query/processors/physical/column_filter_processor.py
@@ -12,7 +12,11 @@ class ColumnFilterProcessor(ClickhouseQueryProcessor):
     """
 
     def __init__(self, column_filters: Sequence[str]) -> None:
-        self.__column_filters = column_filters
+        self.__column_filters = set(column_filters)
 
     def process_query(self, query: Query, query_settings: QuerySettings) -> None:
-        pass
+        assert query.get_is_delete(), f"{query} is not a valid DELETE query"
+        column_names = [column.name for column in query.get_columns()]
+        assert (
+            set(column_names) == self.__column_filters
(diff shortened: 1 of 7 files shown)


snuba/query/processors/physical/column_filter_processor.py
```diff
@@ -12,7 +13,15 @@ class ColumnFilterProcessor(ClickhouseQueryProcessor):
...
     def process_query(self, query: Query, query_settings: QuerySettings) -> None:
-        pass
+        visitor = ColumnVisitor()
+        condition = query.get_condition()
+        assert (
+            condition is not None
+        ), "Not a valid DELETE query. Delete queries should be in the form of DELETE FROM [db.]table WHERE expr"
+        column_names = condition.accept(visitor)
+        assert (
+            column_names == self.__column_filters
```
volokluev, line 26: add the column filters in the message as well. Also, throw a proper exception. asserts can be [optimized away](https://docs.python.org/3/reference/simple_stmts.html#assert) depending on how python is running. Asserts should only be used to assert behavior that should never happen. This can happen

MeredithAnya: Yeah we can use `InvalidQueryException`. I also think we can simplify the error message: `f"Query columns {column_names} must match columns {self.__column_filters}"`

volokluev (approved):

PR #6104, opened 2024-07-15
chore: make snuba co-owners of datasets again

kylemumma:
There was a previous PR that made product teams CODEOWNERS for their snuba datasets. But that PR made it so snuba is no longer codeowners. This PR fixes this to make it so both product teams and snuba own the datasets.


volokluev (approved):
ty!

PR #6105, opened 2024-07-15
feat(auto-import): Enable default vscode auto-import for Snuba

xurui-c:
Enables auto-import for Snuba. Mirrors https://github.com/getsentry/sentry/pull/74197


kylemumma (approved):
i think its possible to do this user only: cmd+shift+p then "open user settings (json)"
im not familiar with this specific vscode setting. if you think it would be beneficial to everyone then this makes sense, but if you think its more of a personal preference than maybe personal user settings would be better?

PR #6108, opened 2024-07-16
fix: LowCardinality column bug w/ aliases

MeredithAnya:
Initial context in https://github.com/getsentry/snuba/pull/6107 Okie it might be the `LowCardinality` column not actually being low cardinality.

Casting the column seems to make it work 
>
```
SELECT
  countIf(
    equals((cast(op, 'String') AS _snuba_op), 'browser')
    OR equals(_snuba_op, 'resource.link')
  ) AS _snuba_matching_count
FROM spans_dist...
```


Zylphrex (approved):
Nice catch!

PR #6113, opened 2024-07-16
feat(snuba-deletes): DeleteRequest object

xurui-c:
In order to support lightweight deletes in Snuba, we will accept `DELETE` queries. Since they are much simpler than the other types of queries Snuba supports, we figured it'd be easier to have a separate `DeleteRequest` object to work off of.

### DIFF
--- a/snuba/request/__init__.py
+++ b/snuba/request/__init__.py
@@ -21,3 +21,9 @@ class Request:
     @property
     def referrer(self) -> str:
         return self.attribution_info.referrer
+
+
+@dataclass(frozen=True)
+class DeleteRequest:
+    storage: str
+    where_clause: str


MeredithAnya:
Actually okay what if you had 

```python
class DeleteRequest:
    id: str
    query: ProcessableQuery
```

The `ProcessableQuery` should cover the where clause (condition) and the storage since the from clause will have the storage

xurui-c:
I want a field to contain the query (or at least the where clause) in string form, since our native clickhouse driver takes in a string. ProcessableQuery does not have a serialize() method, so if we don't have the query string in `DeleteRequest`, I'd have to implement that. The first approach is easier for our purposes.

MeredithAnya (approved):

PR #6118, opened 2024-07-17
support max_bytes_before_external_group_by ClickHouse insert flag

onewland:
I'm trying to track down the source of some production errors when writing to generic-metrics-distributions, where ClickHouse is unable to allocate enough memory to handle the execution of the materialized views. The errors look something like:

```
Strategy(error writing to clickhouse: Code: 241. DB::Exception: Received from xxx:9000. DB::Exception: Memory limit (for query) exceeded: would use 9.81 GiB (attempt to allocate chunk of 1073741824 bytes), maximum: 9.31 GiB.: while executing 'ARRAY JOIN granularities :: 10 -> arrayJoin(granularities) UInt8 : 9': while pushing to view default.generic_metric_distributions_aggregation_mv_v3 (63e6c929-c73f-473e-83d5-9f58c48e904a). (MEMORY_LIMIT_EXCEEDED) (version 23.8.11.29.altinitystable (altinity build))


Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
             at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.82/src/error.rs:83:36
   1: rust_snuba::strategies::clickhouse::batch::BatchFactory::new_batch::{{closure}}
             at ./rust_snuba/src/strategies/clickhouse/batch.rs:135:21
   2: tokio::runtime::task::core::Core<T,S>::poll::{{closure}}
```

Based on some googling, it seems like https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#group-by-in-external-memory might be a parameter we want to adjust around a bit. Ultimately, this should probably be in storage config, but I think for testing it's easiest if we can adjust via CLI

### DIFF
--- a/rust_snuba/src/strategies/clickhouse/batch.rs
+++ b/rust_snuba/src/strategies/clickhouse/batch.rs
@@ -322,6 +322,42 @@ mod tests {
     }
 
     #[test]
+    fn test_write_with_external_groupby() {
+        crate::testutils::initialize_python();
+        let server = MockServer::start();
+        let mock = server.mock(|when, then| {
+            when.method(POST).path("/").body("{\"hello\": \"world\"}\n");
+            then.status(200).body("hi");
+        });
+
+        let concurrency = ConcurrencyConfig::new(1);
+        let factory = BatchFactory::new(
+            &server.host(),
+            server.port(),
+            "testtable",
+            "testdb",
+            &concurrency,
+            "default",
+            "",
+            true,
+            None,
+            Some(500_000),
+        );
+
+        let mut batch = factory.new_batch();
+
+        batch
+            .write_rows(&RowData::from_encoded_rows(vec![
+                br#"{"hello": "world"}"#.to_vec()
+            ]))
+            .unwrap();
+
+        concurrency.handle().block_on(batch.finish()).unwrap();
+
+        mock.assert();
+    }
+
+    #[test]
     fn test_empty_batch() {
         let server = MockServer::start();
         let mock = server.mock(|when, then| {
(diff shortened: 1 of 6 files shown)


rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -70,6 +71,12 @@ impl BatchFactory {
             }
         }
 
+        if max_bytes_before_external_group_by.is_some() {
+            let mut query_segment: String = "&max_bytes_before_external_group_by=".to_owned();
+            query_segment.push_str(&max_bytes_before_external_group_by.unwrap().to_string());
+            query_params.push_str(&query_segment)
+        }
```
john-z-yang, line 78: Nit: I recommend the following to avoid the `.unwrap` call
``` rust
if let Some(max_bytes_before_external_group_by) = max_bytes_before_external_group_by {
    query_segment.push_str(max_bytes_before_external_group_by);
    ...
}
```
Also, follow up question:
how come we are pushing the value in before the option name? Should the order be opposite where we push the option name first, then the value?

onewland: Taking your first suggestion into account.

For the second, I think the way push_str works this should work out to "&max_bytes_before_external_group_by=500000" (or whatever value in place of 500k), right?

john-z-yang: Ah yes you're right, I misread the code.

rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -331,6 +340,7 @@ mod tests {
             "",
             false,
             None,
+            None,
```
volokluev, line 343: maybe having one test where this number is set to something would be good

volokluev (approved):

PR #6121, opened 2024-07-17
fix: wait on system.mutations between hosts during migration execution

kylemumma:
## changes
Now when migrations are executed, the migration will not move on from 1 host to the next one until `select count(*) from system.mutations where is_done=0` is 0

## testing
Not sure how to do a comprehensive test of this but heres how i tested:

* in the case where no mutation in progress, this works. I successfully ran a normal migration locally.
* in the case of a mutation in progress I wasnt able to directly test, but I tested with is_done=1 to make sure the count(*) results were good (`[(13,)]`) so i think it should work

### DIFF
--- a/snuba/migrations/operations.py
+++ b/snuba/migrations/operations.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import time
 from abc import ABC, abstractmethod
 from enum import Enum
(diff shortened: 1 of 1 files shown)


nikhars (approved):
LGTM. Agree with Oliver's comment about the exception raised.

PR #6123, opened 2024-07-18
Add processor for new spans schema

colin-sentry:
We are working on a new schema for spans, to better handle arbitrary keys (hashing), improve the sort key, etc.

This listens to the same kafka topic as the existing consumer, but puts data into the new schema.

PR #6126, opened 2024-07-18
Define the models for the new Events Analytics Platform

colin-sentry:
Related to #6123, the actual migration to create the spans_v2 table


snuba/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,189 @@
...
+            Map(String(), String(), modifiers=Modifiers(codecs=["ZSTD(1)"])),
+        )
+        for i in range(50)
+    ]
+    + [
+        Column(
+            f"attr_num_{i}",
+            Map(String(), Float(64), modifiers=Modifiers(codecs=["ZSTD(1)"])),
+        )
+        for i in range(50)
```
evanh, line 70: I thought we were only doing 10 buckets for these?

snuba/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,189 @@
...
+        return [
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.CollapsingMergeTree(
+                    primary_key="(organization_id, _sort_timestamp, trace_id)",
+                    order_by="(organization_id, _sort_timestamp, trace_id, span_id)",
+                    sign_column="sign",
+                    partition_by="(toMonday(end_timestamp))",
```
evanh, line 157:
```suggestion
                    partition_by="(toMonday(_sort_timestamp))",
```

snuba/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,189 @@
...
+    + [
+        Column(
+            f"attr_bool_{i}",
+            Map(String(), Bool(), modifiers=Modifiers(codecs=["ZSTD(1)"])),
+        )
+        for i in range(10)
+    ]
+)
+
+index_create_ops: List[SqlOperation] = (
```
evanh, line 81: I have a suspicion we're going to want some kind of index on project_id and segment_name. Not sure if we want to add something for that now, or wait and add them later.

colin-sentry: i'm skeptical a DSI on project ID would do much, maybe hold off for now

snuba/clusters/storage_sets.py
```diff
@@ -17,6 +17,7 @@
     "FUNCTIONS": "functions",
     "SEARCH_ISSUES": "search_issues",
     "SPANS": "spans",
+    "SPANS_V2": "spans_v2",
```
phacops, line 20: I think we should make it clear it's a events platform dataset by using `eap_spans` instead of a generic v2.

snuba/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,190 @@
...
+
+columns: Sequence[Column[Modifiers]] = (
+    [
+        Column("organization_id", UInt(64)),
+        Column("project_id", UInt(64)),
+        Column("trace_id", UUID()),
+        Column("span_id", UInt(64)),
+        Column("parent_span_id", UInt(64, Modifiers(codecs=["ZSTD(1)"]))),
+        Column("segment_id", UInt(64, Modifiers(codecs=["ZSTD(1)"]))),
+        Column("segment_name", String(Modifiers(default="''", codecs=["ZSTD(1)"]))),
```
phacops, line 32: I don't think we should have a default here. The Rust struct will specify one anyway and we don't have default for other columns, so either we have default everywhere or not at all (I think not at all is better).

snuba/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,190 @@
...
+    ]
+)
+
+index_create_ops: Sequence[SqlOperation] = (
+    [
+        operations.AddIndex(
+            storage_set=storage_set_name,
+            table_name=local_table_name,
+            index_name="bf_trace_id",
+            index_expression="trace_id",
```
phacops, line 88: Do we still need this index since we're relying on the sort key for querying anyway?

colin-sentry: we aren't using trace TS yet, after that we can drop this index

snuba/snuba_migrations/spans_v2/0001_spans_v2.py
```diff
@@ -0,0 +1,189 @@
...
+        return [
+            operations.CreateTable(
+                storage_set=storage_set_name,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.CollapsingMergeTree(
+                    primary_key="(organization_id, _sort_timestamp, trace_id)",
+                    order_by="(organization_id, _sort_timestamp, trace_id, span_id)",
+                    sign_column="sign",
+                    partition_by="(toMonday(_sort_timestamp))",
```
phacops, line 158: I'd partition by day instead of week, ClickHouse will still be happy and this might prove useful for the mutability case (cc @untitaker )

untitaker: there's definitely downsides to producing this many partitions. I think at this point the decision is a bit arbitrary as I believe we may have to revisit it again anyway with very high likelihood.

phacops (approved):

untitaker (approved):

PR #6130, opened 2024-07-22
feat(cleanup): remove complicated readthrough cache

volokluev:
This cache is no longer being used in production. No reason to keep it around. If we really need it again, we have the git history


nachivrn (commented):
@volokluev Thanks! for working on this PR to remove the old code.

Question: Should we also remove the configuration for randomizing query id as part of this PR?https://github.com/getsentry/snuba/blob/71c9f236ee2be5ee49b58e73b000a506244e7380/snuba/web/db_query.py#L321

volokluev:
yes good call with that @nachivrn

evanh (approved):

PR #6132, opened 2024-07-22
ref(snuba-deletes): Basic delete API

MeredithAnya:
**context**
This is a simple delete API that allows for deleting data for specific storages that are flagged in, if the runtime config `storage_deletes_enabled` is enabled.

This is not production ready, but rather a way to test out the delete functionality in production, with multi-node clusters. We will be using this to test on search_issues data in S4S only as we iterate. 

**Payload**
This schema may change in the future, and be added to, but the basic structure is a `columns` mapping, which maps the column name to the values you filtering on to delete. This only constructs in or eq expressions, not gte or lte for now.

```json
 {
    "columns": {"occurrence_id": ["4c90c9d3-5a51-4bb7-90f1-913a4b1b758d"], "project_id": [3]},
    "tenant_ids": {"referrer": "api_referrer", "organization_id": 1},
}
```

**What the API has:**
* Killswitch
    * Use `storage_deletes_enabled` to turn off the feature (even if storages are enabled), default is off
* Basic functionality to turn a payload into an AST and formats the query
    * Handles adding ON CLUSTER for multi-node delete queries
* Implements `check_shutdown` which checks to see if any requests come in after shutdown

**What API is missing:**
* Proper validation and error handling
    * This includes the `MaxRowEnforcer` and incorporating the `ColumnFilterProcessor`
    * Error handling needs to also include responses to ON CLUSTER requests which seem to have a different response body
* More unit testing, mostly have tests on the endpoint call

### DIFF
--- a/snuba/clickhouse/formatter/query.py
+++ b/snuba/clickhouse/formatter/query.py
@@ -38,7 +38,7 @@ def format_query(query: FormattableQuery) -> FormattedQuery:
     This is the entry point for any type of query, whether simple or
     composite.
     """
-    if query.is_delete():
+    if isinstance(query, Query) and query.is_delete():
         return FormattedQuery(
             _format_delete_query_content(query, ClickhouseExpressionFormatter)
         )
@@ -156,10 +156,20 @@ def _format_delete_query_content(
                     query.get_from_clause()
                 ),
             ),
+            _format_on_cluster(query, formatter),
             _build_optional_string_node("WHERE", query.get_condition(), formatter),
         ]
         if v is not None
     ]
+
+
+def _format_on_cluster(
+    query: AbstractQuery, formatter: ExpressionVisitor[str]
+) -> StringNode:
+    on_cluster = query.get_on_cluster()
+    if on_cluster:
+        return StringNode(f"ON CLUSTER {on_cluster.accept(formatter)}")
+    return None
 
 
 def _format_select(
--- a/snuba/clickhouse/formatter/query.py
+++ b/snuba/clickhouse/formatter/query.py
@@ -38,7 +38,14 @@ def format_query(query: FormattableQuery) -> FormattedQuery:
     This is the entry point for any type of query, whether simple or
     composite.
     """
-    return FormattedQuery(_format_query_content(query, ClickhouseExpressionFormatter))
+    if query.is_delete():
+        return FormattedQuery(
+            _format_delete_query_content(query, ClickhouseExpressionFormatter)
+        )
+    else:
+        return FormattedQuery(
+            _format_query_content(query, ClickhouseExpressionFormatter)
+        )
 
 
 def format_query_anonymized(query: FormattableQuery) -> FormattedQuery:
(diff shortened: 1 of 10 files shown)


snuba/web/delete_query.py
```diff
@@ -0,0 +1,36 @@
+from snuba.clickhouse.formatter.query import format_query
+from snuba.clickhouse.query import Query
+from snuba.datasets.storages.storage_key import StorageKey
+from snuba.query.conditions import combine_and_conditions
+from snuba.query.data_source.simple import Table
+from snuba.query.dsl import column, equals, in_cond, literal, literals_tuple
+
+
+def construct_condition(body):
```
MeredithAnya, line 9: This currently will not handle the case of `gte` or `lte` conditions, like we would do with timestamp.

snuba/web/delete_query.py
```diff
@@ -0,0 +1,36 @@
...
+            literal_values = [literal(v) for v in values]
+            exp = in_cond(column(col), literals_tuple(literal_values))
+
+        and_conditions.append(exp)
+
+    return combine_and_conditions(and_conditions)
+
+
+def delete_query(storage: StorageKey, table: str, body):
+    query = Query(
```
MeredithAnya, line 24: @xurui-c I think we can call processors on this directly since it would be in the correct form. And also for the `MaxRowEnforcer` I think it should be possible to take the `from_clause` and `condition` and construct your own `Query` by adding your select, but I didn't test this out

snuba/clickhouse/formatter/query.py
```diff
@@ -73,6 +80,13 @@ def _visit_simple_query(
         self, data_source: ProcessableQuery[Table]
     ) -> FormattedSubQuery:
         assert isinstance(data_source, Query)
+        if data_source.is_delete():
```
MeredithAnya, line 83: I actually don't know that this is possible, we wouldn't have `DELETE FROM` as a subquery

snuba/web/views.py
```diff
@@ -290,6 +293,35 @@ def mql_dataset_query_view(*, dataset: Dataset, timer: Timer) -> Union[Response,
         assert False, "unexpected fallthrough"
 
 
+@application.route("/<storage:storage>/", methods=["DELETE"])
+@util.time_request("delete_query")
+def storage_delete(*, storage: Storage, timer: Timer) -> Union[Response, str]:
```
MeredithAnya, line 298: Still need to add a basic ratelimiter

snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -115,7 +115,7 @@ query_processors:
           trace.trace_id: trace_id
 
 deletion_settings:
-  is_enabled: 0
+  is_enabled: 1
```
xurui-c, line 118: are we ready to turn this on?

volokluev (approved):

snuba/web/delete_query.py
```diff
@@ -0,0 +1,49 @@
...
+        and_conditions.append(exp)
+
+    return combine_and_conditions(and_conditions)
+
+
+def delete_query(
+    storage: WritableTableStorage, table: str, body: Dict[str, Any]
+) -> Result:
+    cluster_name = storage.get_cluster().get_clickhouse_cluster_name()
+    on_cluster = literal(cluster_name) if cluster_name else None
```
xurui-c, line 34: In what situation would `storage.get_cluster()` not throw an `UndefinedClickhouseCluster` error, but `cluster_name` is still None?

MeredithAnya: @xurui-c if the cluster name isn't defined then we have a single node cluster, which means there is no replication going on (this is [what we have locally](https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L83)). In that case we don't want to add the `ON CLUSTER cluster_name` part of the delete query

PR #6138, opened 2024-07-23
feat(profiling): Write thread id to transactions table

Zylphrex:
In addition to the profiler id, sometimes, we need the thread id to be able to link to the right thread within the profile.

### DIFF
--- a/snuba/datasets/processors/transactions_processor.py
+++ b/snuba/datasets/processors/transactions_processor.py
@@ -421,6 +421,11 @@ class TransactionsMessageProcessor(DatasetMessageProcessor):
         transaction_ctx.pop("trace_id", None)
         transaction_ctx.pop("span_id", None)
 
+        transaction_data = transaction_ctx.pop("data", {})
+        if "thread.id" in transaction_data:
+            # The thread.id can be either a str/int. Make sure to always convert to a str.
+            transaction_ctx["thread_id"] = str(transaction_data["thread.id"])
+
         # The hash and exclusive_time is being stored in the spans columns
         # so there is no need to store it again in the context array.
         transaction_ctx.pop("hash", None)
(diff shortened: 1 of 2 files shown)


snuba/datasets/processors/transactions_processor.py
```diff
@@ -421,6 +421,11 @@ def _sanitize_contexts(
         transaction_ctx.pop("trace_id", None)
         transaction_ctx.pop("span_id", None)
 
+        transaction_data = transaction_ctx.pop("data", {})
```
nikhars, line 424: This is popping the full data. And then below you fill in only the transaction_ctx for thread_id. Rest of the `data` will not make it to storage. Does that imply that `data` was only being used for thread info and nothing else?

Zylphrex: Values inside `data` never made it to storage. When processing the context, only top level `int | float | str` (See [link](https://github.com/getsentry/snuba/blob/txiao/feat/write-thread-id-to-transactions-table/snuba/datasets/events_format.py#L95)) were taken and written to storage and `data` is guaranteed to be a dict from relay (See [link](https://github.com/getsentry/relay/blob/master/relay-event-schema/src/protocol/contexts/trace.rs#L142))

Zylphrex: I could use `.get()` instead if you think that's better but the reason I opted for `.pop()` here is to make it clear that no other fields are stored and that's expected.

nikhars: Makes sense. Could you add the explanation in the code so that context is right there when someone looks at the code a year from now.

nikhars (approved):
Approving since my comment is non-blocking.

PR #6140, opened 2024-07-24
fix(metrics_summaries): Handle null values for optional fields

phacops:


rust_snuba/src/processors/metrics_summaries.rs
```diff
@@ -158,6 +158,44 @@ mod tests {
             .expect("The message should be processed");
     }
 
+    #[test]
+    fn test_summary_with_only_min() {
```
shellmayr, line 162: This test does not use a message without min/max/sum as it is created by relay, or am I missing something here?

shellmayr (approved):

nikhars (approved):

evanh (approved):

PR #6142, opened 2024-07-24
feat(lwdeletes): Add MaxRowEnforcer

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2821

The cost of a lightweight delete operation depends on the number of matching rows in the WHERE clause and the current number of data parts. This operation will be most efficient when matching a small number of rows, and on wide parts (where the _row_exists column is stored in it’s own file) 

Because of the above, we want to limit the number of rows one deletes at a time. The MaxRowsEnforcer will query clickhouse to see how many rows we plan on deleting and if it crosses the max_rows_to_delete set for that storage we will reject the query.

### DIFF
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -50,3 +50,8 @@ class QueryPlanException(SerializableException):
     @classmethod
     def from_args(cls, exception_type: str, message: str) -> "QueryPlanException":
         return cls(exception_type=exception_type, message=message)
+
+
+class TooManyDeleteRowsException(SerializableException):
+    def __init__(self, message: str):
+        super().__init__(message)
--- a/snuba/web/delete_query.py
+++ b/snuba/web/delete_query.py
@@ -1,13 +1,18 @@
+import typing
 from typing import Any, Dict
 
 from snuba.clickhouse.columns import ColumnSet
 from snuba.clickhouse.formatter.query import format_query
 from snuba.clickhouse.query import Query
 from snuba.datasets.storage import WritableTableStorage
+from snuba.datasets.storages.factory import get_storage
+from snuba.datasets.storages.storage_key import StorageKey
+from snuba.query import SelectedExpression
 from snuba.query.conditions import combine_and_conditions
 from snuba.query.data_source.simple import Table
 from snuba.query.dsl import column, equals, in_cond, literal, literals_tuple
-from snuba.query.expressions import Expression
+from snuba.query.exceptions import TooManyDeleteRowsException
+from snuba.query.expressions import Expression, FunctionCall
 from snuba.reader import Result
 
 
@@ -27,6 +32,48 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
     return combine_and_conditions(and_conditions)
 
 
+def _get_rows_to_delete(
+    storage_key: StorageKey, select_query_to_count_rows: Query
+) -> int:
+    formatted_select_query_to_count_rows = format_query(select_query_to_count_rows)
+    select_query_results = (
+        get_storage(storage_key)
+        .get_cluster()
+        .get_reader()
+        .execute(formatted_select_query_to_count_rows)
+    )
+    return typing.cast(int, select_query_results["data"][0]["count"])
+
+
+def _enforce_max_rows(delete_query: Query) -> None:
+    """
+    The cost of a lightweight delete operation depends on the number of matching rows in the WHERE clause and the current number of data parts.
+    This operation will be most efficient when matching a small number of rows, **and on wide parts** (where the `_row_exists` column is stored
+    in its own file)
+
+    Because of the above, we want to limit the number of rows one deletes at a time. The `MaxRowsEnforcer` will query clickhouse to see how many
+      rows we plan on deleting and if it crosses the `max_rows_to_delete` set for that storage we will reject the query.
+    """
+    select_query_to_count_rows = Query(
+        selected_columns=[
+            SelectedExpression("count", FunctionCall("count", "count", ())),
+        ],
+        from_clause=delete_query.get_from_clause(),
+        condition=delete_query.get_condition(),
+    )
+    storage_key = delete_query.get_from_clause().storage_key
+    rows_to_delete = _get_rows_to_delete(
+        storage_key=storage_key, select_query_to_count_rows=select_query_to_count_rows
+    )
+    max_rows_allowed = (
+        get_storage(storage_key).get_deletion_settings().max_rows_to_delete
+    )
+    if rows_to_delete > max_rows_allowed:
+        raise TooManyDeleteRowsException(
+            f"Too many rows to delete ({rows_to_delete}), maximum allowed is {max_rows_allowed}"
+        )
+
+
 def delete_query(
     storage: WritableTableStorage, table: str, body: Dict[str, Any]
 ) -> Result:
@@ -44,6 +91,8 @@ def delete_query(
         on_cluster=on_cluster,
         is_delete=True,
     )
+    _enforce_max_rows(query)
+
     formatted_query = format_query(query)
     # TODO error handling and the lot
     return storage.get_cluster().get_deleter().execute(formatted_query)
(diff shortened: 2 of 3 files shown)


snuba/web/delete_query.py
```diff
@@ -27,23 +30,68 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+        storage.get_cluster().get_reader().execute(formatted_select_query_to_count_rows)
+    )
+    return typing.cast(int, select_query_results["data"][0]["count"])
+
+
+def _enforce_max_rows(
+    storage: WritableTableStorage,
+    from_clause: Table,
+    condition: Expression,
+    on_cluster: Literal | None,
```
volokluev, line 47: why would you ever need this variable?

snuba/web/delete_query.py
```diff
@@ -27,23 +30,68 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+    formatted_select_query_to_count_rows = format_query(select_query_to_count_rows)
+    select_query_results = (
+        storage.get_cluster().get_reader().execute(formatted_select_query_to_count_rows)
+    )
+    return typing.cast(int, select_query_results["data"][0]["count"])
+
+
+def _enforce_max_rows(
+    storage: WritableTableStorage,
+    from_clause: Table,
```
volokluev, line 45: you shouldn't need this, you have the storage key on the Table

snuba/web/delete_query.py
```diff
@@ -27,23 +30,68 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+) -> int:
+    formatted_select_query_to_count_rows = format_query(select_query_to_count_rows)
+    select_query_results = (
+        storage.get_cluster().get_reader().execute(formatted_select_query_to_count_rows)
+    )
+    return typing.cast(int, select_query_results["data"][0]["count"])
+
+
+def _enforce_max_rows(
+    storage: WritableTableStorage,
```
volokluev, line 44: you shouldn't need this, you have the storage key on the table

snuba/web/delete_query.py
```diff
@@ -27,23 +30,68 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
     storage: WritableTableStorage, table: str, body: Dict[str, Any]
 ) -> Result:
     cluster_name = storage.get_cluster().get_clickhouse_cluster_name()
     on_cluster = literal(cluster_name) if cluster_name else None
+    from_clause = Table(
+        table,
+        ColumnSet([]),
+        storage_key=storage.get_storage_key(),
+        # TODO: add allocation policies
+        allocation_policies=[],
```
volokluev, line 78: FYI: even if you add allocation policies here they still will not run because of how you're querying them

https://github.com/getsentry/snuba/pull/6142/files#diff-c3016e51e59d45df20ed861d6db6b02d99925816fd18b5cebe581a6c44e684dbR38

no action needed, just FYI

snuba/web/delete_query.py
```diff
@@ -27,23 +30,68 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+    storage: WritableTableStorage, select_query_to_count_rows: Query
+) -> int:
+    formatted_select_query_to_count_rows = format_query(select_query_to_count_rows)
+    select_query_results = (
+        storage.get_cluster().get_reader().execute(formatted_select_query_to_count_rows)
+    )
+    return typing.cast(int, select_query_results["data"][0]["count"])
+
+
+def _enforce_max_rows(
```
volokluev, line 43: in the architecture diagram this is a distinct component. This means that it should be tested on its own

MeredithAnya:
@xurui-c I think to me it makes more sense to have the row enforcer take a Query object, instead of the separate pieces. So it can take the the delete query and then get the attributes it needs

```python
Query(
    selected_columns=[
        SelectedExpression("count", FunctionCall("count", "count", ())),
    ],
    from_clause=delete_query.get_from_clause(),
    condition=delete_query.get_condition(),
)
```

Also you don't need the `on_cluster` attribute, that is specific to deletes because the query hits a local node. Select queries go through the dist node. You also don't need to specify `delete=False`, since that's the default

snuba/web/delete_query.py
```diff
@@ -27,23 +30,68 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
     return combine_and_conditions(and_conditions)
 
 
+def _get_rows_to_delete(
+    storage: WritableTableStorage, select_query_to_count_rows: Query
+) -> int:
+    formatted_select_query_to_count_rows = format_query(select_query_to_count_rows)
+    select_query_results = (
```
MeredithAnya, line 37: I wonder if it would be useful to use  `_format_storage_query_and_run` in here? I haven't looked at it too deeply but it would be nice to re-use error handling, retry, and logging logic  cc @volokluev

snuba/query/exceptions.py
```diff
@@ -50,3 +50,8 @@ def __init__(
     @classmethod
     def from_args(cls, exception_type: str, message: str) -> "QueryPlanException":
         return cls(exception_type=exception_type, message=message)
+
+
+class MaxRowsEnforcerException(SerializableException):
```
volokluev, line 55: This exception  name does not tell me anything except the component that threw it. Why would this exception be thrown?

snuba/web/delete_query.py
volokluev: I think this is fine as is. Unless there's something specific about the error handling that you want to call out

snuba/web/delete_query.py
```diff
@@ -27,23 +32,62 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+        get_storage(storage_key)
+        .get_cluster()
+        .get_reader()
+        .execute(formatted_select_query_to_count_rows)
+    )
+    return typing.cast(int, select_query_results["data"][0]["count"])
+
+
+def _enforce_max_rows(delete_query: Query) -> None:
```
volokluev, line 48: leave a short comment explaining what this is for and why. It doesn't have a return type so the user has to glean from the thrown errors why this was every implemented

tests/web/test_max_rows_enforcer.py
```diff
@@ -0,0 +1,51 @@
...
+    from_clause=from_clause,
+    condition=condition,
+    on_cluster=None,
+    is_delete=True,
+)
+
+
+@pytest.mark.clickhouse_db
+def test_max_row_enforcer_passes() -> None:
+    _enforce_max_rows(query)
```
MeredithAnya, line 39: This test doesn't assert anything, I think it might make more sense to have this function return true or false so we can assert that it passes.

snuba/web/delete_query.py
```diff
@@ -27,23 +32,62 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+    storage_key = delete_query.get_from_clause().storage_key
+    rows_to_delete = _get_rows_to_delete(
+        storage_key=storage_key, select_query_to_count_rows=select_query_to_count_rows
+    )
+    max_rows_allowed = (
+        get_storage(storage_key).get_deletion_settings().max_rows_to_delete
+    )
+    if rows_to_delete > max_rows_allowed:
+        raise MaxRowsEnforcerException(
+            f"Query wants to delete {rows_to_delete} rows, but the maximum allowed is {max_rows_allowed}"
```
MeredithAnya, line 65: I would just want this a little more explicit and wee shorter, something like

`Too many rows to delete (3000), maximum allowed is 1000`

snuba/query/exceptions.py
MeredithAnya: I think you could name it `TooManyDeleteRows` or something

snuba/web/delete_query.py
```diff
@@ -27,23 +32,70 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
...
+            f"Too many rows to delete ({rows_to_delete}), maximum allowed is {max_rows_allowed}"
+        )
+
+
 def delete_query(
     storage: WritableTableStorage, table: str, body: Dict[str, Any]
 ) -> Result:
     cluster_name = storage.get_cluster().get_clickhouse_cluster_name()
     on_cluster = literal(cluster_name) if cluster_name else None
+    from_clause = Table(
```
MeredithAnya, line 82: I'd also put this back in the Query, we don't need `from_clause` and `condition` to be defined separately imo

MeredithAnya (approved):

PR #6143, opened 2024-07-24
ref: deletes api

kylemumma:
## major changes
1. extract the logic that was inside of `views.py:storage_delete` and put it into a new function `delete_query.py:delete_from_storage`. 
Now whats in views.py is basically just responsible for handling the http request stuff, and the actual logic for deletes happens in `delete_from_storage`. This way my snuba admin UI can call `delete_from_storage` as its interface. The motivation for this change can be best described by a summary of my conversations with Evan and Filippo:

> **Filippo & Evan**: The api-endpoint-view-function (`storage_delete`) should be as lightweight as possible, only responsible for parsing the http request inputs and then passing them along to another function that actually handles the logic.

> **Kyle:** But why does it have to be this way? Can't I just use api endpoint itself as the interface that I send the snuba admin requests to. Then the internals don't matter.

> **Filippo & Evan:** You shouldn't use the api endpoint as the interface for snuba admin requests.

> **Evan:** What if we wanted to do deletes with RPC instead of http? What if we wanted to do them within python instead of http? The view function is only responsible for the transit layer, http, the actual logic of deletes is de-coupled and follows a separate interface. This way whether we want to use http, RPC, or python function call, its more flexible.

> **Kyle**: Oh, i see now.

2. change `body` input to `columns`
The previously defined input for our delete interface was `body` which was an arbitrary key-value mapping. I found this to be unclear on what the expected/required input is. I changed this to be the `columns` object which I think clarifies a bit what is required by the caller.

this change is part of [SNS-2827](https://getsentry.atlassian.net/browse/SNS-2827?atlOrigin=eyJpIjoiMzYxMjNmMGMxMjAxNDM2ZGFjZjAzM2ZjMWY1NTVhZGQiLCJwIjoiaiJ9)

[SNS-2827]: https://getsentry.atlassian.net/browse/SNS-2827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/snuba/web/delete_query.py
+++ b/snuba/web/delete_query.py
@@ -14,11 +14,12 @@ from snuba.query.dsl import column, equals, in_cond, literal, literals_tuple
 from snuba.query.exceptions import TooManyDeleteRowsException
 from snuba.query.expressions import Expression, FunctionCall
 from snuba.reader import Result
+from snuba.state import get_config
 
 
-def construct_condition(body: Dict[str, Any]) -> Expression:
+def construct_condition(columns: Dict[str, Any]) -> Expression:
     and_conditions = []
-    for col, values in body["columns"].items():
+    for col, values in columns.items():
         if len(values) == 1:
             exp = equals(column(col), literal(values[0]))
         else:
@@ -74,8 +75,25 @@ def _enforce_max_rows(delete_query: Query) -> None:
         )
 
 
-def delete_query(
-    storage: WritableTableStorage, table: str, body: Dict[str, Any]
+def delete_from_storage(
+    storage: WritableTableStorage, conditions: Dict[str, Any]
+) -> dict[str, Any]:
+    if get_config("storage_deletes_enabled", 0):
+        raise Exception("Deletes not enabled")
+
+    delete_settings = storage.get_deletion_settings()
+    if not delete_settings.is_enabled:
+        raise Exception(f"Deletes not enabled for {storage.get_storage_key().value}")
+
+    payload: dict[str, Any] = {}
+    for table in delete_settings.tables:
+        result = delete_from_table(storage, table, conditions)
+        payload[table] = {**result}
+    return payload
+
+
+def delete_from_table(
+    storage: WritableTableStorage, table: str, conditions: Dict[str, Any]
 ) -> Result:
     cluster_name = storage.get_cluster().get_clickhouse_cluster_name()
     on_cluster = literal(cluster_name) if cluster_name else None
@@ -87,7 +105,7 @@ def delete_query(
             # TODO: add allocation policies
             allocation_policies=[],
         ),
-        condition=construct_condition(body),
+        condition=construct_condition(conditions),
         on_cluster=on_cluster,
         is_delete=True,
     )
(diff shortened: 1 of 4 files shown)


snuba/web/delete_query.py
```diff
@@ -9,11 +9,12 @@
 from snuba.query.dsl import column, equals, in_cond, literal, literals_tuple
 from snuba.query.expressions import Expression
 from snuba.reader import Result
+from snuba.state import get_config
 
 
-def construct_condition(body: Dict[str, Any]) -> Expression:
+def construct_condition(columns: Dict[str, Any]) -> Expression:
```
xurui-c, line 15: I think we should change this to _construct_condition since it seems like a private method used solely for deletes

xurui-c, line 15: I think we should change this to _construct_condition since it seems like a private method used solely for deletes

snuba/web/views.py
```diff
@@ -302,26 +309,35 @@ def storage_delete(
...
+            }
+            which represents
+            WHERE id in (1,2,3) AND status='failed'
+            """
+            columns = body["columns"]
+        except Exception:
+            return make_response(
+                jsonify(
+                    {
+                        "error": "required input 'conditions' not present in body of request"
```
MeredithAnya, line 330: > "required input 'conditions' not present in body of request"

I'm assuming you wanted `'columns'` not `'conditions'`

snuba/web/delete_query.py
```diff
@@ -27,8 +28,25 @@ def construct_condition(body: Dict[str, Any]) -> Expression:
     return combine_and_conditions(and_conditions)
 
 
-def delete_query(
-    storage: WritableTableStorage, table: str, body: Dict[str, Any]
+def delete_from_storage(
+    storage: WritableTableStorage, conditions: Dict[str, Any]
```
MeredithAnya, line 32: I think we will want to pass through `tenant_ids` and other information aside from the `conditions` in the future because we want logging around where the delete queries are coming from, but for now this works

MeredithAnya (approved):

snuba/web/delete_query.py
MeredithAnya: yeah you could do that for `_delete_from_table` as well

MeredithAnya:
@kylemumma It actually might be good to add a test for a request with a bad payload

xurui-c:
BTW, which JIRA ticket does this address?

snuba/web/delete_query.py
kylemumma: ok will do!

snuba/web/views.py
kylemumma: yes good catch

kylemumma:
@MeredithAnya @xurui-c I made all the suggested changes. I added a new test case for a bad http request that doesnt include the columns field. And I added the relevant jira ticket to the PR description.

xurui-c (approved):

PR #6147, opened 2024-07-25
ref(lwdeletes): Integrate deletion processors into the pipeline

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2820

Now that the lightweight deletes pipeline/skeleton has been built, we integrate deletion processors into the pipeline

### DIFF
--- a/snuba/datasets/deletion_settings.py
+++ b/snuba/datasets/deletion_settings.py
@@ -1,11 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, MutableMapping, Optional, Sequence
-
-from snuba.query.query_settings import QuerySettings
-from snuba.state.quota import ResourceQuota
-from snuba.state.rate_limit import RateLimitParameters
+from typing import Sequence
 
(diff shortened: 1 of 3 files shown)


kylemumma (commented):
It seems like `DeletionQuerySettings` is just a dummy class that you made because `process_query` requires `QuerySettings`. But it seems that `DeletionSettings` doesnt actually have any effect on it. This makes me feel like the extra class isn't really necessary and just add unnecessary complexity? What if you got rid of `DeletionQuerySetting` and instead just pass in a dummy `QuerySettings` object instead?

MeredithAnya:
> It seems like `DeletionQuerySettings` is just a dummy class that you made because `process_query` requires `QuerySettings`. But it seems that `DeletionSettings` doesnt actually have any effect on it. This makes me feel like the extra class isn't really necessary and just add unnecessary complexity? What if you got rid of `DeletionQuerySetting` and instead just pass in a dummy `QuerySettings` object instead?

Yup I agree with @kylemumma here, we don't use the deletion settings in the processors so I think you can create a QuerySettings instance that you pass through

kylemumma (approved):

PR #6148, opened 2024-07-26
fix(eap-spans): Reduce number of buckets we use for attributes

phacops:
After some insertion tests, it turns out we're using too much memory to insert with 200 buckets, so reducing this now.


evanh (approved):

tests/migrations/test_runner.py
```diff
@@ -449,6 +449,9 @@ def test_no_schema_differences() -> None:
         if not isinstance(schema, TableSchema):
             continue
 
+        if storage_key == "events_analytics_platform":
+            print(schema)
```
john-z-yang, line 453: Is this intended?

PR #6149, opened 2024-07-26
fix(eap-spans): Change the nunber of buckets the consumer can use to reflect the schema

phacops:


evanh (commented):
Can we make this value a constant instead of hardcoding in multiple places?

phacops:
> Can we make this value a constant instead of hardcoding in multiple places?

I don't think I can since it's used in a macro.

evanh (approved):

PR #6152, opened 2024-07-26
feat(eap): Add attribute key/value lookup table

evanh:
Create a table that stores just the attribute keys and values that have been seen by an org. This
can be used for autocomplete or any other validation feature.

The table doesn't have any indexes on it at the moment. It also provides a count, to do simple
ranking if that's useful in the future.

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0002_spans_attributes_mv.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0002_spans_attributes_mv.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import Sequence
 
 from snuba.clickhouse.columns import AggregateFunction, Column, DateTime, String, UInt
@@ -57,7 +59,7 @@ class Migration(migration.ClickhouseNodeMigration):
     ]
 
     def forwards_ops(self) -> Sequence[SqlOperation]:
-        create_table_ops = [
+        create_table_ops: list[SqlOperation] = [
             operations.CreateTable(
                 storage_set=self.storage_set_key,
                 table_name=self.meta_local_table_name,
@@ -88,7 +90,7 @@ class Migration(migration.ClickhouseNodeMigration):
             ),
         ]
 
-        materialized_view_ops = []
+        materialized_view_ops: list[SqlOperation] = []
         for value_type in self.value_types:
             str_value = "attribute_value" if value_type == "str" else "NULL"
             num_value = "attribute_value" if value_type == "num" else "NULL"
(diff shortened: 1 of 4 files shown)


snuba/utils/constants.py
```diff
@@ -5,3 +5,6 @@
 
 #: Metrics granularities for which a materialized view exist, in ascending order
 GRANULARITIES_AVAILABLE = (10, 60, 60 * 60, 24 * 60 * 60)
+
+# Number of EAP Span buckets
```
volokluev, line 9: should probably have something here like explaining what it would mean to change this constant

snuba/snuba_migrations/events_analytics_platform/0002_spans_attributes_mv.py
```diff
@@ -0,0 +1,138 @@
...
+    organization_id,
+    attribute_key,
+    {attribute_value} AS attribute_value,
+    toMonday(start_timestamp) AS timestamp,
+    retention_days,
+    sumState(cast(1, 'UInt64')) AS count
+FROM eap_spans_local
+LEFT ARRAY JOIN
+    arrayConcat({key_columns}) AS attribute_key,
+    arrayConcat({value_columns}) AS attr_value
```
colin-sentry, line 23: why is this one attr_value when everything else is attribute_value?

snuba/snuba_migrations/events_analytics_platform/0002_spans_attributes_mv.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        create_table_ops: list[SqlOperation] = [
+            operations.CreateTable(
+                storage_set=self.storage_set_key,
+                table_name=self.meta_local_table_name,
+                engine=table_engines.AggregatingMergeTree(
+                    storage_set=self.storage_set_key,
+                    primary_key="(organization_id, attribute_key)",
+                    order_by="(organization_id, attribute_key, attribute_value, timestamp)",
```
colin-sentry, line 66: if timestamp is in the sorting key, nothing will ever be aggregated? what's the point of using aggregating at all?

snuba/utils/constants.py
colin-sentry: agree, also change the other places that use this constant to also use this one - see my PR for example https://github.com/getsentry/snuba/pull/6173/files

colin-sentry (changes requested):

snuba/snuba_migrations/events_analytics_platform/0002_spans_attributes_mv.py
evanh: Earlier in the function I do `{attribute_value} AS attribute_value,` and I didn't want to overload the alias.

snuba/snuba_migrations/events_analytics_platform/0002_spans_attributes_mv.py
evanh: The timestamp is rounded to the nearest Monday in the query, so it will aggregate an entire week of data.

Not sure if that answers your question at all.

colin-sentry (approved):

volokluev (approved):

PR #6153, opened 2024-07-26
fix(eap): Update number of columns in entity

colin-sentry:
We moved from 200 columns to 20, reflect that here


phacops (approved):

ayirr7:
What is this change for exactly?

ayirr7 (approved):

colin-sentry:
> What is this change for exactly?

The entity exposes a "attr_str[hello]" column, that gets mapped to one of the buckets by hashing hello - this number is the modulus, if it's wrong you get errors due to trying to access the wrong bucket

PR #6156, opened 2024-07-29
fix(events-analytics-platform): Fix sharding key, clickhouse does not like u128

colin-sentry:
This works locally at least:

```
8a035ef769f1 :) select cityHash64(reinterpretAsUInt128('550e8400-e29b-41d4-a716-446655440000'::UUID));

SELECT cityHash64(reinterpretAsUInt128(CAST('550e8400-e29b-41d4-a716-446655440000', 'UUID')))

 7067610571742207507
```

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0001_spans.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0001_spans.py
@@ -139,7 +139,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 columns=columns,
                 engine=table_engines.Distributed(
                     local_table_name=local_table_name,
-                    sharding_key="reinterpretAsUInt128(trace_id)",
+                    sharding_key="cityHash64(reinterpretAsUInt128(trace_id))",  # sharding keys must be at most 64 bits
                 ),
                 target=OperationTarget.DISTRIBUTED,
             ),


ayirr7 (approved):

ayirr7:
Actually, it might be a good idea to double check if you can modify the migration in-line like this. I'm not sure what's customary when we want to change the sharding key, maybe @MeredithAnya knows?

colin-sentry:
> Actually, it might be a good idea to double check if you can modify the migration in-line like this. I'm not sure what's customary when we want to change the sharding key, maybe @MeredithAnya knows?

we've done this before for this migration, will take care of it

MeredithAnya:
> > Actually, it might be a good idea to double check if you can modify the migration in-line like this. I'm not sure what's customary when we want to change the sharding key, maybe @MeredithAnya knows?
> 
> we've done this before for this migration, will take care of it

@colin-sentry when you changed it before, did you already have the tables created in prod? Because it looks like this migration has already been run now, in which case, you'd need a new migration.

onewland:
@colin-sentry Riya's intuition was correct, this is not the right way to fix a migration that made it to production. It should have been done in a follow-up migration

colin-sentry:
> @colin-sentry Riya's intuition was correct, this is not the right way to fix a migration that made it to production. It should have been done in a follow-up migration

wouldn't follow-up migration that drop table / create table have had the same issue with "table too big to drop" ?

colin-sentry:
I see, makes sense

PR #6157, opened 2024-07-29
feat: new page in snuba admin to do deletes

kylemumma:
This PR adds a page to snuba admin that we can use to test deletes in prod.

## major changes
1. create a front-end react interface for deletes
* `static/delete_tool/index.tsx` is the new react UI, it calls `api_client.tsx:runLightweightDelete` 
* `api_client.py:runLightweightDelete` sends the input via http request to `/delete` endpoint. (note: i think it would have been equally valid to just send the http request from index.tsx rather than having this separate function, but I did it like this bc everyone else in the codebase did)
* `data.tsx` adds the new page to the Nav bar
2. create a new `/delete` endpoint in snuba admin `admin/views.py:delete`
As said in bullet 1, the react UI sends the input to `/delete`. This is a new endpoint we created, it mostly just calls `delete_from_storage` (what Meredith implemented that actually does the delete)
3. set permissions for `/delete` endpoint and UI `tool_policies.py`
It just goes into AllTools by default which is the most restrictive role, only snuba team and people with access to everything can use it. I think this should be good.
4. small `delete_query.py` changes
* I slightly changed the input/output types of `delete_query.py:delete_from_storage` in order to make the interface more strict/clear.
* added more input verification for the `columns` input.

## considerations
This tool is pretty much just a layer on top of `delete_from_storage` meaning it can do anything that `delete_from_storage` can do. Hence it will be able to delete in any environment, and any dataset that has deletes enabled. Is this fine or should we add additional strictness to this interface, so that it can only be used for search_issues in s4s no matter what.

Screenshot 2024-07-30 at 1 10 29 PM

### DIFF
--- a/snuba/admin/static/delete_tool/index.tsx
+++ b/snuba/admin/static/delete_tool/index.tsx
@@ -19,9 +19,9 @@ function DeleteTool(props: { api: Client }) {
                         alert("expect columnConditions to be valid json but its not");
                         return;
                     }
-                    props.api.runLightweightDelete(storageName, conds).then(data => setResult(data)
-                    }
-                }>Submit</button>
+                    props.api.runLightweightDelete(storageName, conds).then(data => setResult(JSON.stringify(data)))
+                }
+            }>Submit</button>
             <p>latest result: {result}</p>
         </div>
       );
--- a/snuba/admin/static/delete_tool/index.tsx
+++ b/snuba/admin/static/delete_tool/index.tsx
@@ -1,0 +1,18 @@
+import Client from "SnubaAdmin/api_client";
+import React, { useEffect, useState } from "react"
+
+function DeleteTool(props: { api: Client }) {
+    const [inputValue, setInputValue] = useState('')
+    const [submittedVal, setSubmittedVal] = useState('')
+
+
+    return (
+        <div>
+            <input type="text" value={inputValue} onChange={(event) => setInputValue(event.target.value)} />
+            <button type="submit" onClick={(event) => setSubmittedVal(inputValue)}>Submit</button>
+            <p>sent: {submittedVal}</p>
+        </div>
+      );
+}
+
+export default DeleteTool
--- a/snuba/admin/static/delete_tool/index.tsx
+++ b/snuba/admin/static/delete_tool/index.tsx
@@ -2,15 +2,27 @@ import Client from "SnubaAdmin/api_client";
 import React, { useEffect, useState } from "react"
 
 function DeleteTool(props: { api: Client }) {
-    const [inputValue, setInputValue] = useState('')
-    const [submittedVal, setSubmittedVal] = useState('')
-
+    const [storageName, setStorageName] = useState('')
+    const [columnConditions, setColumnConditions] = useState('')
+    const [result, setResult] = useState('')
 
     return (
         <div>
-            <input type="text" value={inputValue} onChange={(event) => setInputValue(event.target.value)} />
-            <button type="submit" onClick={(event) => setSubmittedVal(inputValue)}>Submit</button>
-            <p>sent: {submittedVal}</p>
+            <input type="text" value={storageName} placeholder="storage name" onChange={(event) => setStorageName(event.target.value)} /><br/>
+            <textarea value={columnConditions} placeholder="column conditions" onChange={(event) => setColumnConditions(event.target.value)} /><br/>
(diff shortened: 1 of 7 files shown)


snuba/web/views.py
```diff
@@ -331,13 +331,15 @@ def storage_delete(
             )
 
         try:
-            payload = delete_from_storage(storage, columns)
+            results = delete_from_storage(storage, columns)
+        except ValueError as error:
+            return make_response(jsonify({"error": str(error)}), 400)
```
kylemumma, line 336: this catches the new `columns` validation added to `delete_from_storage`

xurui-c:
Not too important, but it'd be nice to have a screenshot of what your UI looks like, kinda like this: https://github.com/getsentry/snuba/pull/5883#issue-2286691038

kylemumma:
@xurui-c I added a screenshot to the description

snuba/web/delete_query.py
```diff
@@ -44,11 +47,20 @@ def delete_from_storage(
     if not delete_settings.is_enabled:
         raise Exception(f"Deletes not enabled for {storage.get_storage_key().value}")
 
-    payload: dict[str, Any] = {}
+    # assert isinstance(columns, dict[str, list[Any]])
+    columns_is_expected_type = isinstance(columns, dict) and all(
```
MeredithAnya, line 51: I would prefer to have a schema that we validate against similar to the `RequestSchema` logic in https://github.com/getsentry/snuba/blob/e7f134852bdd33fd6d22483926463facba8ef3e0/snuba/request/schema.py#L32

I can see if I can add something in that will work for that

snuba/web/delete_query.py
```diff
@@ -44,11 +47,20 @@ def delete_from_storage(
...
+        [isinstance(k, str) and isinstance(v, list) for k, v in columns.items()]
+    )
+    if not columns_is_expected_type:
+        raise ValueError(
+            "expected input 'columns' to be of type dict[str, list[Any]] but it was not"
```
MeredithAnya, line 56: This is kind of an odd response because the request is in json, so maybe some thing like "Invalid 'columns' object values: {columns.values()}' or something

I also think it could make sense to raise a `InvalidQueryException` instead

snuba/admin/static/delete_tool/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+function DeleteTool(props: { api: Client }) {
+    const [storageName, setStorageName] = useState('')
+    const [columnConditions, setColumnConditions] = useState('')
+    const [result, setResult] = useState('')
+
+    return (
+        <div>
+            <input type="text" value={storageName} placeholder="storage name" onChange={(event) => setStorageName(event.target.value)} /><br/>
+            <textarea value={columnConditions} placeholder="column conditions" onChange={(event) => setColumnConditions(event.target.value)} /><br/>
+            <button type="submit" onClick={
```
MeredithAnya, line 13: do you think you could pull out the `onClick` function here? instead of defining inline

snuba/admin/static/delete_tool/index.tsx
```diff
@@ -0,0 +1,39 @@
+import Client from "SnubaAdmin/api_client";
+import React, { useEffect, useState } from "react"
+
+function DeleteTool(props: { api: Client }) {
+    const [storageName, setStorageName] = useState('')
+    const [columnConditions, setColumnConditions] = useState('')
+    const [result, setResult] = useState('')
+
+    return (
+        <div>
```
MeredithAnya, line 10: also could you add a little description here about what the payload looks like ? That way someone could just copy paste it easily

snuba/web/delete_query.py
kylemumma: The reason I added this in is bc python doesnt have runtime type checking so I needed to add the logic for it myself.

What is the reasoning you would prefer to have it like the `RequestSchema` object? Is it so that the input validation is abstracted and not a part of the function?

snuba/admin/static/delete_tool/index.tsx
kylemumma: done

snuba/admin/static/delete_tool/index.tsx
kylemumma: done

MeredithAnya (approved):

PR #6158, opened 2024-07-29
ref(snuba-deletes): add with_span

MeredithAnya:
Just adds some sentry tracing for the delete flow


snuba/web/views.py
```diff
@@ -307,7 +307,6 @@ def storage_delete(
     *, storage: WritableTableStorage, timer: Timer
 ) -> Union[Response, str]:
     if http_request.method == "DELETE":
-
```
MeredithAnya, line 310: this space was bothering me so removing it

kylemumma (approved):

PR #6159, opened 2024-07-30
fix(eap-spans): Add sync to the downwards migration, mark dataset experimental

phacops:
We didn't configure the table the right way to take advantage of the storage policy `multidisk` we defined. We need to drop the table to re-create it.

This PR will turn the `events_analytics_platform` dataset into an experimental dataset and use the `SYNC` option to drop the table immediately and avoid problems with Zookeeper keeping some data by default for DR.

From https://clickhouse.com/docs/en/sql-reference/statements/drop
```
If the SYNC modifier is specified, the entity is dropped without delay.
```

We will set the default disk policy to `multidisk` on clusters that need it.

Plan to deploy this would be:
- Set the readiness state for eap_spans to experimental
- Deploy this change to every environment

Then, in `de` then `us`:
- Truncate the `eap_spans_local` table if there's data (so we can drop it)
- Run the reverse migration
- Change the default disk policy to `multidisk`
- Run the forward migration

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0001_spans.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0001_spans.py
@@ -1,3 +1,4 @@
+from os import environ
 from typing import List, Sequence
 
 from snuba.clusters.storage_sets import StorageSetKey
@@ -117,6 +118,10 @@ class Migration(migration.ClickhouseNodeMigration):
     blocking = False
 
     def forwards_ops(self) -> Sequence[SqlOperation]:
+        settings = {"index_granularity": "8192"}
+        if environ.get("SENTRY_REGION") in ["de", "us"]:
+            settings["storage_policy"] = "multidisk"
+
         res: List[SqlOperation] = [
             operations.CreateTable(
                 storage_set=storage_set_name,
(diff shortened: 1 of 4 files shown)


colin-sentry (approved):

snuba/snuba_migrations/events_analytics_platform/0001_spans.py
```diff
@@ -158,10 +158,12 @@ def backwards_ops(self) -> Sequence[SqlOperation]:
                 storage_set=storage_set_name,
                 table_name=local_table_name,
                 target=OperationTarget.LOCAL,
+                sync=True,
```
ayirr7, line 161: What do we plan to do after changing this migration in-line?

colin-sentry: should probably wait for @onewland - we'll have to drop all of the data again because we are changing the storage configuration of the clickhouse cluster

ayirr7:
Could we also maybe outline the plan for rolling out the changes in this PR in the description?

phacops:
- Set the readiness state for `eap_spans` to `experimental`
- Deploy this change to every environment

Then, in `de` then `us`:
- Run the reverse migration
- Change the default disk policy to `multidisk`
- Run the forward migration

ayirr7:
> DROP TABLE doesn't work if tables have any data

Thanks for calling that out, @onewland

phacops:
We found out truncating the table would still let ClickHouse takes into account the new default disk policy, so we will not need to drop the table.

PR #6160, opened 2024-07-30
chore(eap): Set client sample rate if it comes from upstream

colin-sentry:
We would like to re-multiply things if they are set from upstream.


evanh (approved):
I question why the sample rate is being sent as an attribute on the measurements, but otherwise this is fine.

PR #6161, opened 2024-07-30
ref(snuba-deletes): add DELETE_QUERY_SCHEMA

MeredithAnya:


snuba/web/delete_query.py
```diff
@@ -44,9 +46,14 @@ def delete_from_storage(
     if not delete_settings.is_enabled:
         raise Exception(f"Deletes not enabled for {storage.get_storage_key().value}")
 
+    def _get_columns() -> dict[str, Any]:
```
kylemumma, line 49: My concern with this here is that `delete_from_storage` is meant to be independent of HTTP. `web/views.py:storage_delete` is responsible for parsing the HTTP request and handing it to `delete_from_storage` so from this point HTTP is done. but with this change we are kind of designing it to interface with HTTP. 

I also feel as though it makes the expected input less clear as its a generic body.

My opinion would be that this schema validation should happen inside of `web/views.py:storage_delete` rather than here, since its specific to an HTTP request.

snuba/web/views.py
```diff
@@ -310,28 +310,12 @@ def storage_delete(
         check_shutdown({"storage": storage.get_storage_key()})
         body = parse_request_body(http_request)
         try:
-            """
```
kylemumma, line 313: putting the schema validation here would also make it more clear what the HTTP interface looks like.

AS-IS you would have to dig into the `delete_from_storage` to figure out the interface.

kylemumma (commented):
I think doing schema validation like this is an improvement and helps clarify and validate the interface that we are building. But my main concern with this is the HTTP abstraction line, this schema validation seems specific to the HTTP interface, and thus  I believe it should happen in `web/views.py:storage_delete` rather than `delete_from_storage`.

snuba/web/delete_query.py
MeredithAnya: > but with this change we are kind of designing it to interface with HTTP

Can you elaborate on this a bit?

My aim is to allow for more information, like tenant ids and debug info to make it through to the `delete_from_storage` entry point. Otherwise it will make it hard to get that info for allocation policies and the like. So to me, the interface has to allow for more than just `columns`.

MeredithAnya: @kylemumma i updated the PR to use the request parts, lemme know what you think of that instead

kylemumma: My opinion is that the most important thing is that we have a clearly defined interface, which in this case is the function signature. As far as my understanding goes, our current interface is the name of a storage, and a dictionary mapping column names to a list of values representing the AND statement. Because this is my understanding of what the interface is, thats why im so strongly opposed to the function looking different than that.

If we want to in the future add things like tenant id to the interface, I think we should add them once they are a part of the interface, but right now it seems they arent.

In the last iteration when we change this function from `dict[str, list[Any]]` to `dict[Any]` in my opinion it makes our interface less clearly defined and more confusing. In the current iteration.

If you believe that the interface for lightweight deletes is: a storage name and a RequestParts object, than I would support this change. But I think the question would be is that the interface?

I'll approve the PR so you can merge this version, or the last version you had, or we can get on a call to talk about this more. And then I will modify my PR based on what you go with. This is just my opinion on the matter.

kylemumma (approved):

snuba/web/delete_query.py
MeredithAnya: Just adding context since we chatted - will keep the columns signature for now, though it will change in upcoming PRs

PR #6164, opened 2024-07-31
feat(eap): Also read the data in span.data into the buckets

colin-sentry:
- Read span.data from relay and insert it into the eap_spans buckets
- Use type inference from the sent json to decide whether each value in `span.data` goes into the string buckets or the number buckets

### DIFF
--- a/rust_snuba/src/processors/eap_spans.rs
+++ b/rust_snuba/src/processors/eap_spans.rs
@@ -80,11 +80,6 @@ fn fnv_1a(input: &[u8]) -> u32 {
 
 impl From<FromSpanMessage> for EAPSpan {
     fn from(from: FromSpanMessage) -> EAPSpan {
-        let sentry_tags = from.sentry_tags.unwrap_or_default();
-        let tags = from.tags.unwrap_or_default();
-        let measurements = from.measurements.unwrap_or_default();
-        let data = from.data.unwrap_or_default();
-
         let mut res = Self {
             organization_id: from.organization_id,
             project_id: from.project_id,
@@ -97,7 +92,6 @@ impl From<FromSpanMessage> for EAPSpan {
             segment_id: from
                 .segment_id
                 .map_or(0, |s| u64::from_str_radix(&s, 16).unwrap_or(0)),
-            segment_name: sentry_tags.get("transaction").cloned().unwrap_or_default(),
             is_segment: from.is_segment,
             _sort_timestamp: (from.start_timestamp_ms / 1000) as u32,
             start_timestamp: (from.start_timestamp_precise * 1e6) as u64,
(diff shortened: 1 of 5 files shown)


rust_snuba/src/processors/eap_spans.rs
```diff
@@ -119,32 +115,71 @@ impl From<FromSpanMessage> for EAPSpan {
...
+            };
+
+            let mut insert_num = |k: String, v: f64| {
+                attr_num_buckets[(fnv_1a(k.as_bytes()) as usize) % attr_num_buckets.len()]
+                    .insert(k.clone(), v);
+            };
+
+            if let Some(sentry_tags) = from.sentry_tags {
+                sentry_tags.iter().for_each(|(k, v)| {
+                    if k == "transaction" {
```
phacops, line 137: Why don’t we get this outside of the loop since it’s a map?

rust_snuba/src/processors/eap_spans.rs
```diff
@@ -119,32 +115,71 @@ impl From<FromSpanMessage> for EAPSpan {
...
+
+            if let Some(tags) = from.tags {
+                tags.iter().for_each(|(k, v)| {
+                    insert_string(k.clone(), v.clone());
+                })
+            }
+
+            if let Some(measurements) = from.measurements {
+                measurements.iter().for_each(|(k, v)| {
+                    if k == "client_sample_rate" && v.value != 0.0 {
```
phacops, line 153: I think I would find this measurement specifically outside the loop as well.

rust_snuba/src/processors/eap_spans.rs
colin-sentry: otherwise it shows up twice - once in the bucket, and again normalized

rust_snuba/src/processors/eap_spans.rs
colin-sentry: ditto

rust_snuba/src/processors/eap_spans.rs
```diff
@@ -119,32 +115,71 @@ impl From<FromSpanMessage> for EAPSpan {
...
+                            serde_json::to_string(array).unwrap_or_default(),
+                        ),
+                        Value::Object(object) => insert_string(
+                            k.clone(),
+                            serde_json::to_string(object).unwrap_or_default(),
+                        ),
+                        Value::Number(number) => {
+                            insert_num(k.clone(), number.as_f64().unwrap_or_default())
+                        }
+                        Value::Bool(true) => insert_num(k.clone(), 1.0),
```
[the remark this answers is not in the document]
colin-sentry: I considered both, if you add it as a numeric value you can do `count` easily by doing `sum(val * sign)`

phacops (approved):

ayirr7 (approved):

PR #6166, opened 2024-07-31
fix: snuba deletes config default 

kylemumma:
fix bug in region specific delete feature flag


MeredithAnya (approved):

xurui-c:
Can we write a test where this exception would've been raised?

xurui-c (approved):

PR #6169, opened 2024-07-31
ref(eap): Add performance to EAP

evanh:


phacops (changes requested):
We're adding a team for EAP, let's not add performance for EAP.

https://github.com/getsentry/security-as-code/pull/901

phacops (commented):
Can we assign the `eap_spans.rs` file to the team as well please?

PR #6175, opened 2024-08-01
Make the number of threads to run the optimize cron job configurable

xurui-c:
The optimize job has been [failing](https://app.datadoghq.com/monitors/72460548?q=kube_job%3Asnuba-optimize-errors-tiger-mz-2-28703520%2Csentry_region%3Aus&event_id=7684526334976626842&source=monitor_notif&view=spans&from_ts=1722295368000&to_ts=1722296568000&eval_ts=1722296568000) due to not meeting the cutoff time. 

There's two aspects of the job scheduling:
- a period of time where the job runs in parallel, which puts more load on ClickHouse
- a period of time when the job optimizes one partition at a time. This lets it overlap with our peak load time without exhausting ClickHouse's resources

In this PR, we changed the parallelization period to run with a configurable number of threads, and this configuration is checked every time a partition finishes optimizing. This way, we can change the number of threads we use the parallelize to avoid overwhelming Clickhouse, before the cron job finishes.

### DIFF
--- a/snuba/cli/optimize.py
+++ b/snuba/cli/optimize.py
@@ -35,16 +35,17 @@ from snuba.redis import RedisClientKey, get_redis_client
 @click.option("--log-level", help="Logging level to use.")
 @click.option(
     "--parallel",
+    "default_parallel_threads",
     type=click.IntRange(1, 2),
     default=1,
-    help="Run parallel optimizations",
+    help="Default parallel threads",
 )
 def optimize(
     *,
     clickhouse_host: Optional[str],
     clickhouse_port: Optional[int],
     storage_name: str,
-    parallel: int,
+    default_parallel_threads: int,
     log_level: Optional[str] = None,
 ) -> None:
     from datetime import datetime
@@ -119,7 +120,7 @@ def optimize(
         clickhouse=connection,
         storage=storage,
         database=database,
-        parallel=parallel,
+        default_parallel_threads=default_parallel_threads,
         clickhouse_host=clickhouse_host,
         tracker=tracker,
         before=today,
--- a/snuba/clickhouse/optimize/optimize.py
+++ b/snuba/clickhouse/optimize/optimize.py
@@ -299,7 +299,7 @@ def optimize_partition_runner(
     scheduler = OptimizeScheduler(default_parallel_threads=default_parallel_threads)
 
     with ThreadPoolExecutor(max_workers=32) as executor:
-        futures: set[Future[Any]] = set()
+        pending_futures: set[Future[Any]] = set()
 
         partitions_to_optimize = deque(partitions)
         while partitions_to_optimize:
@@ -310,8 +310,10 @@ def optimize_partition_runner(
                 f"{schedule.cutoff_time} with {configured_num_threads} threads"
             )
 
-            while partitions_to_optimize and len(futures) < configured_num_threads:
-                futures.add(
+            while (
+                partitions_to_optimize and len(pending_futures) < configured_num_threads
+            ):
+                pending_futures.add(
                     executor.submit(
                         optimize_partitions,
                         clickhouse,
@@ -324,10 +326,10 @@ def optimize_partition_runner(
                     )
                 )
 
-            done, futures = concurrent.futures.wait(
-                futures, return_when=concurrent.futures.FIRST_COMPLETED
+            completed_futures, futures = concurrent.futures.wait(
+                pending_futures, return_when=concurrent.futures.FIRST_COMPLETED
             )
-            for future in done:
+            for future in completed_futures:
                 future.result()
 
 
--- a/snuba/clickhouse/optimize/optimize.py
+++ b/snuba/clickhouse/optimize/optimize.py
@@ -1,10 +1,12 @@
+import concurrent
+import concurrent.futures
 import multiprocessing
 import os
 import threading
 import time
 import typing
 from collections import deque
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timedelta
 from typing import Mapping, Optional, Sequence
 
@@ -50,6 +52,8 @@ logger = structlog.get_logger().bind(module=__name__)
 logger = structlog.wrap_logger(logger, processors=[thread_info_processor])
 
 metrics = MetricsWrapper(environment.metrics, "optimize")
+
+_OPTIMIZE_PARALLEL_THREADS_KEY = "optimize_parallel_threads"
 
 
 def _get_metrics_tags(table: str, clickhouse_host: Optional[str]) -> Mapping[str, str]:
@@ -281,11 +285,10 @@ class _ScheduledPartitionAndJiter:
     start_time_jitter_minutes: int | None
 
 
-def _get_partition_and_jitter_pairs(
+def _get_partitions_to_optimize(
     schedule: OptimizationSchedule,
 ) -> deque[_ScheduledPartitionAndJiter]:
-    scheduled_partitions_and_jitter_pairs: deque[_ScheduledPartitionAndJiter] = deque()
-    # eque[List[Union[str, Union[int, None]]]]
+    partitions_to_optimize: deque[_ScheduledPartitionAndJiter] = deque()
 
     max_partition_group_size = max(
         len(partition_group) for partition_group in schedule.partitions_groups
@@ -294,7 +297,7 @@ def _get_partition_and_jitter_pairs(
     for i in range(max_partition_group_size):
         for partition_group in schedule.partitions_groups:
             if i < len(partition_group):
-                scheduled_partitions_and_jitter_pairs.append(
+                partitions_to_optimize.append(
                     _ScheduledPartitionAndJiter(
                         partition_group[i],
                         schedule.start_time_jitter_minutes[i]
@@ -302,7 +305,7 @@ def _get_partition_and_jitter_pairs(
                         else None,
                     )
                 )
-    return scheduled_partitions_and_jitter_pairs
+    return partitions_to_optimize
 
 
 def optimize_partition_runner(
@@ -324,46 +327,71 @@ def optimize_partition_runner(
     1. There are no more partitions which need optimization.
     2. The final cutoff time is reached. In this case, the scheduler will
     raise an exception which would be propagated to the caller.
+
+    Details of execution flow:
+    1. start by reading configured_num_threads from the Snuba Admin runtime config
+    2. dispatches configured_num_threads threads to optimize configured_num_threads partitions (1 partition per thread)
+    3. as soon as one thread finishes, check configured_num_threads from runtime config again
+    4. if configured_num_threads > number of currently active threads, dispatch more threads
     """
 
     remaining_partitions = partitions
-    while remaining_partitions:
-        schedule = scheduler.get_next_schedule(remaining_partitions)
-        scheduled_partition_and_jitter_pairs = _get_partition_and_jitter_pairs(schedule)
+    with ThreadPoolExecutor(max_workers=32) as executor:
+        while remaining_partitions:
+            schedule = scheduler.get_next_schedule(remaining_partitions)
+            partitions_to_optimize = _get_partitions_to_optimize(schedule)
 
-        num_threads = len(schedule.partitions_groups)
-        configured_num_threads = get_config("optimize_parallel_threads")
-        if scheduler.get_is_running_parallel() and configured_num_threads:
-            num_threads = typing.cast(int, configured_num_threads)
-        logger.info(
-            f"Running schedule with cutoff time: "
-            f"{schedule.cutoff_time} with {num_threads} threads"
-        )
+            default_num_threads = len(schedule.partitions_groups)
+            configured_num_threads = typing.cast(
+                int, get_config(_OPTIMIZE_PARALLEL_THREADS_KEY, default_num_threads)
+            )
+            logger.info(
+                f"Running schedule with cutoff time: "
+                f"{schedule.cutoff_time} with {configured_num_threads} threads"
+            )
 
-        with ThreadPoolExecutor(max_workers=num_threads) as executor:
             futures = []
-
-            while scheduled_partition_and_jitter_pairs:
-                scheduled_partition_and_jitter = (
-                    scheduled_partition_and_jitter_pairs.popleft()
-                )
+            for _ in range(min(len(partitions_to_optimize), configured_num_threads)):
+                partition_to_optimize = partitions_to_optimize.popleft()
                 futures.append(
                     executor.submit(
                         optimize_partitions,
                         clickhouse,
                         database,
                         table,
-                        [scheduled_partition_and_jitter.partition],
+                        [partition_to_optimize.partition],
                         schedule.cutoff_time,
                         tracker,
                         clickhouse_host,
-                        scheduled_partition_and_jitter.start_time_jitter_minutes,
+                        partition_to_optimize.start_time_jitter_minutes,
                     )
                 )
 
-                if len(futures) >= num_threads:
-                    for future in as_completed(futures):
-                        futures.remove(future)
+            _, active_futures = concurrent.futures.wait(
+                futures, return_when=concurrent.futures.FIRST_COMPLETED
+            )
+
+            configured_num_threads = typing.cast(
+                int, get_config(_OPTIMIZE_PARALLEL_THREADS_KEY, default_num_threads)
+            )
+            for _ in range(configured_num_threads - len(active_futures)):
+                if not partitions_to_optimize:
+                    break
+
+                partition_to_optimize = partitions_to_optimize.popleft()
+                futures.append(
+                    executor.submit(
+                        optimize_partitions,
+                        clickhouse,
+                        database,
+                        table,
+                        [partition_to_optimize.partition],
+                        schedule.cutoff_time,
+                        tracker,
+                        clickhouse_host,
+                        None,
+                    )
+                )
 
         # If there are still partitions needing optimization then move on to the
         # next bucket with the partitions which still need optimization.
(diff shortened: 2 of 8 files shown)


snuba/clickhouse/optimize/optimize.py
```diff
@@ -287,41 +328,71 @@ def optimize_partition_runner(
     2. The final cutoff time is reached. In this case, the scheduler will
     raise an exception which would be propagated to the caller.
     """
+
     remaining_partitions = partitions
     while remaining_partitions:
         schedule = scheduler.get_next_schedule(remaining_partitions)
-        num_threads = len(schedule.partitions)
+        scheduled_partition_and_jitter_pairs = _get_partition_and_jitter_pairs(schedule)
```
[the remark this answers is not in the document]
xurui-c: > Also, how does this variable relate to `remaining_partitions`?
My understanding is that if there are any partitions from the current schedule that didn't finish optimizing, then they will get optimized again because `remaining_partitions` won't be empty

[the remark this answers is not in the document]
mcannizz: Do we need two while loops to get that done? Do we need to re-create the threadpool in one of those loops? I would have expected that we'd just create the threadpool once.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -267,6 +279,35 @@ def get_current_large_merges(
...
+            if i < len(partition_group):
+                scheduled_partitions_and_jitter_pairs.append(
+                    _ScheduledPartitionAndJiter(
+                        partition_group[i],
+                        schedule.start_time_jitter_minutes[i]
+                        if schedule.start_time_jitter_minutes
+                        else None,
+                    )
+                )
+    return scheduled_partitions_and_jitter_pairs
```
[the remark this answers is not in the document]
mcannizz: @xurui-c what do you think about this point? I don't think we've addressed it yet.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,42 +327,74 @@ def optimize_partition_runner(
...
+                        executor.submit(
+                            optimize_partitions,
+                            clickhouse,
+                            database,
+                            table,
+                            [scheduled_partition_and_jitter_pair.partition],
+                            schedule.cutoff_time,
+                            tracker,
+                            clickhouse_host,
+                            None,
```
xurui-c, line 395: @mcannizz This part of the code adds in futures that weren't part of the first `configured_num_threads` optimize jobs. I think by passing in `None` for `start_jitter`, we effectively get rid of jittering

snuba/clickhouse/optimize/optimize.py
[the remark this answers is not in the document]
xurui-c: https://github.com/getsentry/snuba/pull/6175#discussion_r1704713137

snuba/clickhouse/optimize/optimize.py
mcannizz: Ah, I see that now. Two things:
1. I don't think we can reasonably expect a reader of the code to figure out that intention as it is right now. The difference between the calls to `optimize_partitions` is quite easy to miss.
2. I think there's a simpler way to implement this that likely removes the need for `scheduled_partition_and_jitter_pairs` entirely.

On my second point, `optimize_partition_runner` is supposed to be responsible for scheduling and coordinating the work while `optimize_partitions` is supposed to actually optimize partitions. Jitter is a scheduling concern, so we should seek to implement it entirely in `optimize_partition_runner` if we can.

How can we do that? We could try something like this:
```
def jitter_start_time(start_jitter):
    if start_jitter:    
        logger.info(
            f"{threading.current_thread().name}: Jittering start time by"
            f" {start_jitter} minutes"
        )
        time.sleep(start_jitter * 60)
```

And then:
```
# Ramp up load on ClickHouse slowly, over OPTIMIZE_PARALLEL_MAX_JITTER_MINUTES
# by forcing each thread other than the first to wait before starting.
for jitter in schedule.start_time_jitter():
    futures.append(executor.submit(jitter_start_time, jitter)

while partitions_to_optimize:
    _, active_futures = concurrent.futures.wait(
        futures, return_when=concurrent.futures.FIRST_COMPLETED
    )

    # add futures that call optimize_partitions as before
    # ...
```

What do you think about that idea? Happy to discuss sync if you'd like.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -330,6 +385,80 @@ def optimize_partition_runner(
...
+    configured_num_threads: int,
+    schedule: OptimizationSchedule,
+    clickhouse: ClickhousePool,
+    database: str,
+    table: str,
+    tracker: OptimizedPartitionTracker,
+    clickhouse_host: str,
+    start_time_jitter_for_each_partition: Sequence[int],
+    is_initial_run: bool,
+) -> None:
```
mcannizz, line 442: I'm not sure if splitting this code out into a separate function makes things more readable, because there's an awful lot of parameters that need to be passed. Also, this function is a no-op when `is_initial_run` is `False`, so even if we want to keep this function separate we should get rid of this parameter and just not call this method outside of the initial run. I think we can easily do this by moving the call outside of the body of the `while remaining_partitions:` loop.

This is what I'm thinking, in psuedo-code:
```
with ThreadPoolExecutor(max_workers=32) as executor:
    futures = []
    # jitter start time
    for jitter in start_time_jitter_for_each_partition:
        futures.append(executor.submit(_jitter_start_time, jitter))

    partitions_to_optimize = deque(partitions)
    while partitions_to_optimize:
        while len(futures) < configured_num_threads:
            futures.append(executor.submit(_new_optimize_job(partitions_to_optimize.popleft())))
        _, futures = concurrent.futures.wait(futures, return_when=concurrent.futures.FIRST_COMPLETED)
```
Thoughts?

mcannizz (changes requested):

snuba/clickhouse/optimize/optimize.py
```diff
@@ -1,9 +1,14 @@
+import concurrent
+import concurrent.futures
```
xurui-c, line 2: -

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,48 +294,58 @@ def optimize_partition_runner(
...
-        for i in range(0, num_threads):
-            threads.append(
-                threading.Thread(
-                    target=optimize_partitions,
-                    args=(
+
+    with ThreadPoolExecutor(max_workers=32) as executor:
+        futures: set[Future[Any]] = set()
+        # jitter start time
+        for jitter in scheduler.get_start_time_jitter_for_each_partition():
```
xurui-c, line 308: @mcannizz 

Previously, I had jitter scheduling after determining the number of threads. This is because the number of jitter futures should be the same as the number of threads. However, the code's readability suffered. 

This current approach of scheduling jittering before determining the number of threads has the problem of scheduling either 1 or 2 jitter futures, since in `get_start_time_jitter_for_each_partition`, `self.__parallel` is either 1 or 2.

I think a quick and easy fix is to just make `get_start_time_jitter_for_each_partition` work with the number of threads instead of `self.__parallel`, so the function is more like `get_start_time_jitter_for_each_thread`.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -267,6 +278,22 @@ def get_current_large_merges(
...
+
+    max_partition_group_size = max(
+        len(partition_group) for partition_group in schedule.partitions_groups
+    )
+
+    for i in range(max_partition_group_size):
+        for partition_group in schedule.partitions_groups:
+            if i < len(partition_group):
+                partitions_to_optimize.append(partition_group[i])
+    return partitions_to_optimize
```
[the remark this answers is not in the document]
xurui-c: I went with approach 1, but the reason why I was zig-zaging instead of doing approach 2 is because i was trying to preserve [the order in which partitions are optimized](https://github.com/getsentry/snuba/blob/cc6f37c4513f1d5d2356d0e84033a94ccb08ee3e/snuba/clickhouse/optimize/optimize_scheduler.py#L77-L79)

My understanding is that if we had the `schedule.partitions_groups = [[1,2], [3,4,5], [6,7]]`, then before this PR, the order in which partitions should be optimized is
- partition 1, 3, 6 (simultaneously optimized)
- partition 2, 4, 7
- partition 5

so by zig-zaging, `partitions_to_optimize` becomes `[1,3,6,2,4,7,5]`

tests/clickhouse/optimize/test_optimize.py
```diff
@@ -244,6 +246,100 @@ def test_metrics_tags(
...
+
+class TestOptimizeFrequency:
+    @pytest.mark.clickhouse_db
+    @pytest.mark.redis_db
+    @pytest.mark.parametrize(
+        "storage_key, create_event_row_for_date, current_time",
+        test_data,
+    )
+    @patch("snuba.clickhouse.optimize.optimize.optimize_partitions")
+    def test_optimize_partitions_is_called_the_same_number_of_times_as_number_of_partitions(
```
mcannizz, line 281: I was running this test in the debugger to better understand it, and I found that when I paused it for a while to think, the test would fail. Here's why I think that is happening:
1. By doing two inserts on a series of dates, the test creates a situation where ClickHouse has 10 partitions with 2 parts in each.
2. Left to its own devices, ClickHouse will merge the parts together, until it has 10 partitions with 1 part each.
3. If ClickHouse completes its background merge process before the test calls `optimize.get_partitions_to_optimize`, there's no optimization necessary and the test fails.

A test with a race condition like this will be flaky in CI so we need to fix it. I have two ideas on how we can do that:
1. This test asserts that if we ask `optimize_partition_runner` to schedule the work needed to optimize 10 partitions, then `optimize_partitions` is called 10 times. Can we just skip inserting all this data into the database and instead just pass `optimize_partition_runner` a list of 10 fake partitions?
2. If that doesn't work, we can tell ClickHouse to not do any background merges by adding this code to the start of the test:
```
clickhouse.execute(f"SYSTEM STOP MERGES {database}.{table}")
```

What do you think? As an aside, I suspect that this problem is why `test_optimize` is expected to flake sometimes. Can we fix it too? Probably the `SYSTEM` command is the only way to fix that one.

tests/clickhouse/optimize/test_optimize.py
```diff
@@ -244,6 +246,100 @@ def test_metrics_tags(
...
+            port=9000,
+            database=database,
+            table=table,
+            expire_time=datetime.now() + timedelta(minutes=10),
+        )
+        partitions = optimize.get_partitions_to_optimize(
+            clickhouse, storage, database, table
+        )
+
+        tracker.update_all_partitions([part.name for part in partitions])
```
mcannizz, line 324: If I comment out this line, the test still passes. Do we need it? I see that this line causes some state to be written to redis, but I don't understand how that state fits in with the logic we changed. Could you explain that?

snuba/clickhouse/optimize/optimize.py
mcannizz: I'm not sure I understand the problem. Can you explain it a bit more? It would be helpful to understand what behavior we expect and what actual behavior we get with this code.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,48 +294,58 @@ def optimize_partition_runner(
...
+            )
+            logger.info(
+                f"Running schedule with cutoff time: "
+                f"{schedule.cutoff_time} with {configured_num_threads} threads"
+            )
+
+            while partitions_to_optimize and len(futures) < configured_num_threads:
+                futures.add(
+                    executor.submit(
+                        optimize_partitions,
```
mcannizz, line 326: What happens if `optimize_partitions` fails?

mcannizz (dismissed):

snuba/cli/optimize.py
```diff
@@ -35,16 +35,17 @@
 @click.option("--log-level", help="Logging level to use.")
 @click.option(
     "--parallel",
+    "default_parallel_threads",
     type=click.IntRange(1, 2),
```
xurui-c, line 39: For now I kept this because I don't want the cron job to start with too many threads

xurui-c, line 39: For now I kept this because I don't want the cron job to start with too many threads

xurui-c, line 39: For now I kept this because I don't want the cron job to start with too many threads

xurui-c, line 39: For now I kept this because I don't want the cron job to start with too many threads

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,48 +289,58 @@ def optimize_partition_runner(
...
-                threading.Thread(
-                    target=optimize_partitions,
-                    args=(
+    scheduler = OptimizeScheduler(default_parallel_threads=default_parallel_threads)
+
+    with ThreadPoolExecutor(max_workers=32) as executor:
+        futures: set[Future[Any]] = set()
+        # jitter start time
+        for jitter in scheduler.get_start_time_jitter():
+            futures.add(executor.submit(_jitter_start_time, jitter))
```
nikhars, line 305: Since we are touching this code, lets get rid of the jitter. It was originally introduced with the theory that concurrent reads/writes happening for same sized partitions at the end of the optimization run would overload IOPS. But in practice that was not the case. So we can cleanup the jitter code.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,48 +289,58 @@ def optimize_partition_runner(
...
+                    )
                 )
-            )
 
-            threads[i].start()
+            done, futures = concurrent.futures.wait(
+                futures, return_when=concurrent.futures.FIRST_COMPLETED
+            )
+            for future in done:
+                future.result()
```
nikhars, line 334: Should you also remove the future from the futures set so that [this](https://github.com/getsentry/snuba/pull/6175/files#diff-fb97ede04634458aaa8ca52ac172447bbd893822cc750d7765f1272f4d1a7a0fR316) while loop continues in the next iteration?

xurui-c: My understanding is that the `futures` set is reassigned by `concurrent.futures.FIRRST_COMPLETED` to only contain futures that haven't finished

nikhars: That makes sense. I feel from a code readability perspective the following changes would help

1. Rename futures to pending_futures.
2. Rename done on [this](https://github.com/getsentry/snuba/pull/6175/files#diff-fb97ede04634458aaa8ca52ac172447bbd893822cc750d7765f1272f4d1a7a0fR327) line to completed_futures.

nikhars (approved):
Left a minor comment for better code readability. Other than that looks good to me.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,48 +289,48 @@ def optimize_partition_runner(
...
-                threading.Thread(
-                    target=optimize_partitions,
-                    args=(
+    scheduler = OptimizeScheduler(default_parallel_threads=default_parallel_threads)
+
+    with ThreadPoolExecutor(max_workers=32) as executor:
+        pending_futures: set[Future[Any]] = set()
+
+        partitions_to_optimize = deque(partitions)
+        while partitions_to_optimize:
```
onkar, line 305: I wonder if we can simplify two loops into just one:
```
pending_futures: set[Future[Any]] = set()
partitions_to_optimize = deque(partitions)
configured_num_threads = get_num_threads(default_parallel_threads)
while partitions_to_optimize and len(pending_futures) < configured_num_threads:
    schedule = scheduler.get_next_schedule(partitions_to_optimize)
    logger.info(...)
    pending_futures.add(...)
    completed_futures, pending_futures = wait(pending_futures,return_when=FIRST_COMPLETED)
    
    for future in completed_futures:
        future.result()
```

In fact, I think the two conditions in the `while` loop are testing for the same thing, so we should be able to even simplify it further to just one condition like `while partitions_to_optimize`:

snuba/clickhouse/optimize/optimize_scheduler.py
```diff
@@ -79,66 +79,50 @@ def subdivide_partitions(
...
-        sorted_partitions = OptimizeScheduler.sort_partitions(partitions)
+        sorted_partitions = OptimizeScheduler._sort_partitions(partitions)
         output: MutableSequence[Sequence[str]] = []
 
         for i in range(number_of_subdivisions):
             output.append(sorted_partitions[i::number_of_subdivisions])
 
         return output
 
-    def start_time_jitter(self) -> Sequence[int]:
```
onkar, line 90: Was this removed because this method is unused?

onkar (approved):
LGTM overall, well done! I added a few comments on simplifying the implementation. Let me know if taking in those suggestions make sense and the code still works as expected.

snuba/clickhouse/optimize/optimize.py
```diff
@@ -286,48 +289,48 @@ def optimize_partition_runner(
...
+                partitions_to_optimize and len(pending_futures) < configured_num_threads
+            ):
+                pending_futures.add(
+                    executor.submit(
+                        optimize_partitions,
                         clickhouse,
                         database,
                         table,
-                        schedule.partitions[i],
+                        [partitions_to_optimize.popleft()],
```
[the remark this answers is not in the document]
xurui-c: The existing `optimize_partitions` [definition](https://github.com/getsentry/snuba/blob/d703a97004cc671a5c8e37fdc0de46af800de080/snuba/clickhouse/optimize/optimize.py#L274) takes in a Sequence

snuba/clickhouse/optimize/optimize.py
xurui-c: If we simplify it to a single loop like that, then you are only checking the configured number of threads once, before the `while` loop. We want to adjust the number of threads on the fly.

> I think the two conditions in the while loop are testing for the same thing, so we should be able to even simplify it further to just one condition like while partitions_to_optimize

if we take out `len(pending_futures) < configured_num_threads`, then we can potentially schedule more partitions to optimize than the number of threads. The goal is to never schedule more partitions than the number of configured threads

snuba/clickhouse/optimize/optimize_scheduler.py
xurui-c: yep: https://github.com/getsentry/snuba/pull/6175#discussion_r1735378797

PR #6180, opened 2024-08-06
ref(snuba-deletes): introduce delete allocation policies

MeredithAnya:
**context**
Right now we have a simple ratelimiter that limits deletes to `1` concurrent query at a time for the whole endpoint https://github.com/getsentry/snuba/blob/ea2f8e716ea519ae173e7a6833da5d2bf16859f2/snuba/web/views.py#L305

This is good enough for testing out ourselves, but beyond that we wanted to add a more robust ratelimiter that allowed for more granular levers that we have for other endpoints. Product should also be ratelimiting on their end though, this is not supposed to be in lieu of that. 

**delete allocation policies**
We can reuse the allocation policy framework for the delete query pipeline. I've made these policies separate to the other allocation policies definitions because we should be adjusting the limits separately; deletes should not have as high thresholds as select queries because the load those queries cause will be higher, even though they are 'lightweight' (lightweight compared to heavy mutations).

**`DeleteConcurrentRateLimitAllocationPolicy`**
I made a separate class because I want the redis key to be different than the `ConcurrentRateLimitAllocationPolicy` so that we don't clobber ratelimits

This PR just adds the set up for the allocation policies, and does nothing with them (aside from adding they to the `Query` object. The rest of the logic will be in https://github.com/getsentry/snuba/pull/6181

### DIFF
--- a/snuba/datasets/configuration/issues/storages/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/storages/search_issues.yaml
@@ -89,6 +89,15 @@ allocation_policies:
         is_enforced: 0
         is_active: 0
 
+delete_allocation_policies:
+  - name: DeleteConcurrentRateLimitAllocationPolicy
+    args:
+      required_tenant_types:
+        - referrer
+        - project_id
+      default_config_overrides:
+        is_enforced: 0
+
 query_processors:
   - processor: TableRateLimit
   - processor: MappingOptimizer
--- a/snuba/datasets/configuration/json_schema.py
+++ b/snuba/datasets/configuration/json_schema.py
@@ -593,6 +593,7 @@ V1_READABLE_STORAGE_SCHEMA = {
         "deletion_processors": DELETION_PROCESSORS_SCHEMA,
         "mandatory_condition_checkers": STORAGE_MANDATORY_CONDITION_CHECKERS_SCHEMA,
         "allocation_policies": STORAGE_ALLOCATION_POLICIES_SCHEMA,
+        "delete_allocation_policies": STORAGE_ALLOCATION_POLICIES_SCHEMA,
         "required_time_column": {
             "type": ["string", "null"],
             "description": "The name of the required time column specifed in schema",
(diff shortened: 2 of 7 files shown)


snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -89,6 +89,15 @@ allocation_policies:
         is_enforced: 0
         is_active: 0
 
+delete_allocation_policies:
+  - name: DeleteConcurrentRateLimitAllocationPolicy
+    args:
+      required_tenant_types:
+        - referrer
+        - project_id
```
armenzg, line 97: Would the org be also needed or valuable to have?

volokluev (commented):
If you're changing the yaml configuration please add a test for it, otherwise this looks good

kylemumma:
I've never worked with allocation policies before so i'm not super comfortable with this code but as far as I can tell it LGTM.

1 question I have: why did you put `DeleteConcurrentRateLimitAllocationPolicy` under `delete_allocation_policies` rather than under the existing `allocation_policies`

The reason im not approving is bc maybe you should add a test based on Volos comment? Is he referring to the addition of `delete_allocation_policies` in the yaml?

kylemumma (commented):
I've never worked with allocation policies before so i'm not super comfortable with this code but as far as I can tell it LGTM.

1 question I have: why did you put `DeleteConcurrentRateLimitAllocationPolicy` under `delete_allocation_policies` rather than under the existing `allocation_policies`

The reason im not approving is bc maybe you should add a test based on Volos comment? Is he referring to the addition of `delete_allocation_policies` in the yaml?

MeredithAnya:
@kylemumma 

> 1 question I have: why did you put `DeleteConcurrentRateLimitAllocationPolicy` under `delete_allocation_policies` rather than under the existing `allocation_policies`

I don't want the delete rate limits to run in the main query pipeline, this is because we will be enforcing different, and more strict limits. I didn't want the logic of getting the allocation policies have to know if the policy is for deletes or not, rather that pipeline is in charge of either getting the select allocation policies or the delete allocation policies. Does that clear things up?

snuba/datasets/configuration/issues/storages/search_issues.yaml
MeredithAnya: yeah I can add it in

kylemumma:
> @kylemumma
> 
> > 1 question I have: why did you put `DeleteConcurrentRateLimitAllocationPolicy` under `delete_allocation_policies` rather than under the existing `allocation_policies`
> 
> I don't want the delete rate limits to run in the main query pipeline, this is because we will be enforcing different, and more strict limits. I didn't want the logic of getting the allocation policies have to know if the policy is for deletes or not, rather that pipeline is in charge of either getting the select allocation policies or the delete allocation policies. Does that clear things up?

@MeredithAnya Yeah that makes sense, thanks for the explanation

kylemumma (approved):

PR #6181, opened 2024-08-06
ref(snuba-deletes): use allocation policies

MeredithAnya:
**context**
We have the configurations for the delete allocation policies but they aren't actually being used in the delete pipeline yet.
Some more info in the original PR that added the configuration for the delete allocation policies https://github.com/getsentry/snuba/pull/6180. 

This reuses some of the functionality from the regular select query pipeline. The delete concurrent policy we defined will not be enforced yet, so we will still be relying on the simple rate limiter  https://github.com/getsentry/snuba/blob/ea2f8e716ea519ae173e7a6833da5d2bf16859f2/snuba/web/views.py#L305
until we are ready to enforce the allocation policy. We can also add another delete policy once we verify the delete allocation policy pipeline is working as expected.

**next steps**
- [ ] add delete allocation policies to the snuba admin UI so that we can change values https://github.com/getsentry/snuba/pull/6248
- [ ] test out the delete allocation policy to verify it
- [ ] add a delete referrer guard policy

### DIFF
--- a/snuba/web/delete_query.py
+++ b/snuba/web/delete_query.py
@@ -1,6 +1,9 @@
 import typing
-from typing import Any, Dict
+import uuid
+from typing import Any, Dict, Mapping, MutableMapping
 
+from snuba.attribution import get_app_id
+from snuba.attribution.attribution_info import AttributionInfo
 from snuba.clickhouse.columns import ColumnSet
 from snuba.clickhouse.formatter.query import format_query
 from snuba.clickhouse.query import Query
@@ -8,6 +11,10 @@ from snuba.datasets.storage import WritableTableStorage
 from snuba.datasets.storages.factory import get_storage, get_writable_storage
 from snuba.datasets.storages.storage_key import StorageKey
 from snuba.query import SelectedExpression
+from snuba.query.allocation_policies import (
+    AllocationPolicyViolations,
+    QueryResultOrError,
+)
 from snuba.query.conditions import combine_and_conditions
 from snuba.query.data_source.simple import Table
 from snuba.query.dsl import column, equals, in_cond, literal, literals_tuple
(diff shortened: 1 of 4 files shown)


snuba/web/delete_query.py
```diff
@@ -137,17 +160,75 @@ def _delete_from_table(
...
+            )
+            if result
+            else None
+        )
+        result_or_error = QueryResultOrError(query_result=query_result, error=error)
+        for allocation_policy in allocation_policies:
+            allocation_policy.update_quota_balance(
+                tenant_ids=attribution_info.tenant_ids,
+                query_id=query_id,
+                result_or_error=result_or_error,
```
volokluev, line 225: This should be in its own function

snuba/web/delete_query.py
```diff
@@ -136,48 +195,88 @@ def get_new_from_clause() -> Table:
...
-    formatted_query = format_query(query)
-    # TODO error handling and the lot
-    return storage.get_cluster().get_deleter().execute(formatted_query)
+    try:
+        _apply_allocation_policies_quota(
+            query_settings,
+            attribution_info,
+            formatted_query,
+            stats,
+            allocation_policies,
```
volokluev, line 238: You're using the `_get_allocation_policies` function here which will get you the query policies. Aren't you using different policies for deleting?

MeredithAnya: @volokluev shouldn't that function get the policies off the query? I assumed https://github.com/getsentry/snuba/pull/6181/files/b8b6fd6c3dbdb65e411abb524a272e3a78393c1e#diff-c3016e51e59d45df20ed861d6db6b02d99925816fd18b5cebe581a6c44e684dbR97 is where I set the policies, and that should be populating the delete ones

volokluev: that's a really roundabout way to do that. It's (somewhat) necessary in the execution pipeline but not here. You have the storage object. just get the delete policies from the storage

MeredithAnya: @volokluev I still pulled out getting the allocation policies out of that function in order to be able to mock it in tests

PR #6184, opened 2024-08-07
feat(eap): rough protobuf timeseries API

volokluev:


requirements.txt
```diff
@@ -11,6 +11,7 @@ Flask==2.2.5
 google-cloud-storage==2.18.0
 googleapis-common-protos==1.63.2
 google-api-core==2.19.1
+proto-plus==1.24.0
```
colin-sentry, line 14: this is already in this file

snuba/protobufs/base_messages.proto
```diff
@@ -0,0 +1,67 @@
...
+  LTE = 4;
+  GTE = 5;
+  LIKE = 6;
+  NOT_LIKE = 7;
+  IN = 8;
+  NOT_IN = 9;
+}
+
+
+message PentityFilter{
```
evanh, line 28: At a certain point this will just be a Protobuf version of the objects we have in the Snuba SDK, right?

volokluev: More restrictive. For example we won't allow arbitrary functions here like we do in the snuba SDK

volokluev: I'm actually going to remove these filters and use the ones Colin proposed

https://github.com/getsentry/snuba/blob/master/snuba/protobufs/Filters.proto

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans.yaml
```diff
@@ -54,7 +57,13 @@ storages:
 storage_selector:
   selector: DefaultQueryStorageSelector
 
-query_processors: []
+query_processors:
```
evanh, line 60: Can we get rid of this? I don't like this processor and would prefer it go away. It does a lot more than provide a `time` alias that this entity doesn't need.

snuba/web/rpc/timeseries.py
```diff
@@ -0,0 +1,68 @@
...
+
+TimeSeriesResponse = Any
+
+
+def _agg_type_to_ch_function(agg_type: AggregationType) -> str:
+    lookup = {
+        AggregationType.AVG: "avg",
+        AggregationType.MIN: "min",
+        AggregationType.MAX: "max",
+        AggregationType.P50: "quantiles(0.5)",
```
evanh, line 27: Can you change this to `quantile` instead? Otherwise you'll get an array in the results.

evanh (commented):
I notice that we are not trying to solve how functions can be used in this API, which is fine to start. However how to handle functions has been a difficult problem in all our APIs, and it's one that this API will need to solve as well.

I think figuring that out earlier will be time well spent, because functions are almost certainly going to be something we need to support. For example, your filters won't support `has(tags.key, '...') = 1`, or `fromUTCTimestamp(timestamp_ms) > '...'`.

volokluev:
This PR is no longer necessary

snuba/web/rpc/timeseries.py
volokluev: Thank you!

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans.yaml
volokluev: The TimeSeriesProcessor has been pretty helpful in making these queries easy to write, will on the performance team asked for it as well. What is the concern with it?

PR #6185, opened 2024-08-07
feat(generic-metrics): Bump sentry kafka schema ver

john-z-yang:
Bump the version so we can have sampling weight in the generic metrics messages


john-z-yang:
On second thought, it probably better to include this change in the consumer change

PR #6189, opened 2024-08-08
Increase the default throttle/warning threshold for allocation policies

xurui-c:
Allocation policies are our mechanism for doing traffic management for Snuba queries. Currently, we see a lot of warnings (Warning: Query from referrer ... is throttled) on Sentry Issues. This is because ~17% of queries to Snuba are throttled by capacity management/allocation policies. This is an overkill since we actually have a lot of Clickhouse capacity.

This PR increases the default throttle/warning threshold to be higher than half of the rejection threshold. I will also increase the thresholds in Snuba Admin.


snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -102,14 +102,14 @@ def _additional_config_definitions(self) -> list[AllocationPolicyConfig]:
...
-                int,
+                float,
                 DEFAULT_BYTES_THROTTLE_DIVIDER,
             ),
             AllocationPolicyConfig(
                 "threads_throttle_divider",
                 "max threads divided by this number is the number of threads we use to execute queries for a throttled (project_id|organization_id, referrer)",
                 int,
-                DEFAULT_BYTES_THROTTLE_DIVIDER,
+                DEFAULT_THREADS_THROTTLE_DIVIDER,
```
xurui-c, line 112: this fixes an existing bug

MeredithAnya (commented):
let's update the test as well https://github.com/getsentry/snuba/blob/5e70b76567296a7f1c3a91c9a65d9e71c17bea90/tests/query/allocation_policies/test_bytes_scanned_rejecting_policy.py#L115

onewland:
IMO the title for this PR (and merge commit message, when it comes time for that) should be changed. 

This doesn't "get rid of sentry warnings" for throttled queries, it adjusts the threshold in the hopes of reducing the quantity of throttles and, downstream of that, the quantity of warnings we emit.

PR #6190, opened 2024-08-08
ref(eap): Change sampling_weight to a UInt

evanh:
All of the Clickhouse -Weighted quantile functions require a UInt as the weight.
Change the column in spans to be a `UInt64` so it can be used directly in
-Weighted functions, without needing to do any type conversion.


MeredithAnya (commented):
@evanh I think you are missing the `target` field on the operations, also does this just need to be changed for the local tables?

enochtangg (commented):
Should we be defaulting `1` here?

nikhars (approved):

enochtangg (approved):

PR #6192, opened 2024-08-12
Sets throttle policy correctly

xurui-c:
Redo of https://github.com/getsentry/snuba/pull/6088 because it got reverted


volokluev (approved):

volokluev (dismissed):
Can you add a test for this behavior?

onewland:
Can you describe the incorrect behavior and the corrected behavior?

tests/web/test_db_query.py
```diff
@@ -478,6 +482,132 @@ def test_db_query_fail() -> None:
...
+
+        def _update_quota_balance(
+            self,
+            tenant_ids: dict[str, str | int],
+            query_id: str,
+            result_or_error: QueryResultOrError,
+        ) -> None:
+            nonlocal update_called
+            update_called = True
+            return
```
onewland, line 549: would it be a little more readable if the output was an argument to the constructor here and you just had 1 class (with two instances)? Or does the logic require them to actually be two separate classes

onewland (approved):
thanks for fixing the explanation plus adding a test.

note: please clean up the "title" of the PR before merge to be a little clearer. maybe something like: "fix(capman): expose correct throttle policy in query result metadata when multiple policies present"

tests/web/test_db_query.py
xurui-c: What do you mean by "output was an argument to the constructor"?

The logic I'm testing here is to have 2 different policies. In reality they could be `ReferrerGuardRailPolicy` and `BytesScannedRejectingPolicy`, which are separate classes.

snuba/web/db_query.py
```diff
@@ -875,18 +875,19 @@ def _apply_allocation_policies_quota(
...
                 span.set_data(
                     "quota_allowance",
                     quota_allowances[allocation_policy.config_key()],
                 )
                 if allowance.is_throttled:
-                    throttle_quota_and_policy = _QuotaAndPolicy(
-                        quota_allowance=allowance,
-                        policy_name=allocation_policy.config_key(),
-                    )
+                    if allowance.max_threads < num_threads:
```
onkar, line 885: These two conditions can be `and`ed. For readability, I would rename `num_threads` to `min_threads_across_policies` or a shorter version of it. `num_threads` is a bit abstract and does not really call out the intent of this change.

tests/web/test_db_query.py
```diff
@@ -478,6 +482,132 @@ def test_db_query_fail() -> None:
...
```
onkar, line 518: nit: Is this `return` necessary?

tests/web/test_db_query.py
onkar, line 549: ditto, is this `return` necessary?

onkar (approved):
LGTM, thanks for making this change. Left a few comments around readability.

tests/web/test_db_query.py
onewland: I'm imagining there's a class called `MockAllocationPolicy` and it looks something like

```python
class MockAllocationPolicy(AllocationPolicy):
   def __init__(quota_allowance_value):
      self.__quota_allowance = quota_allowance_value
  
   def _get_quota_allowance(self, tenant_ids: dict[str, str | int], query_id: str):
     return self.__quota_allowance
```

and you instantiate two instances of that rather than having two subclasses that do essentially the same thing with a slightly different faked return.

That said, looking at the code a bit further, I think the class names do have semantic value here and you really do need two separate subclasses.

onkar: We could have `MockAllocationPolicy` inherit from `AllocationPolicy` and then `MockAllocationPolicy` implement the currently duplicate functionality. Then inherit `ThrottleAllocationPolicy1` and `ThrottleAllocationPolicy2` from `MockAllocationPolicy` and override `_get_quota_allowance` for things like `max_threads` in each Throttle class. That should reduce duplication, increase readability and still have 2 classes like you want.

tests/web/test_db_query.py
xurui-c: no it's not, I removed it

tests/web/test_db_query.py
xurui-c: Thank you Onkar and Oliver! I've made the changes.

onkar (approved):
Looks great, @xurui-c! Thanks for addressing all review comments 👍

PR #6196, opened 2024-08-13
feat(EAP) Add Make directive that creates proper import paths

colin-sentry:


snuba/protobufs/Filters.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
```
volokluev, line 1: I like these filter definitions, very elegant

volokluev (approved):

PR #6198, opened 2024-08-13
ref(snuba-deletes): add ColumnValidator

MeredithAnya:
We have some [schema validation](https://github.com/getsentry/snuba/blob/d4b164c06710e15bed33d868102305adf8fe495c/snuba/query/schema.py#L38) on delete api, but because the columns depend on the storage, we can't do that level of type validation when parsing the payload. The only required column we have is the `project_id` (which is debatable on whether we should have that there but anyway)

If the column `occurrence_id` is an uuid and someone sends:

```json
{ "columns": {"occurrence_id": ["hello"]}
```
we'll get a clickhouse error when trying to run the delete.

**ColumnValidator**
This validator will check to make sure that the type defined in the schema matches the values passed through in the api. I've only implemented a handful of types and I don't expect that we will need many more, at least for now.

### DIFF
--- a/snuba/utils/schemas.py
+++ b/snuba/utils/schemas.py
@@ -21,6 +21,7 @@ from typing import (
 
 from snuba.clickhouse.escaping import escape_identifier
 from snuba.utils.constants import NESTED_COL_EXPR_RE
+from snuba.utils.serializable_exception import SerializableException
 
 
 class TypeModifier(ABC):
@@ -707,39 +708,49 @@ class Enum(ColumnType[TModifiers]):
         return Enum(self.values)
 
 
+class InvalidColumnType(SerializableException):
+    pass
+
+
 class ColumnValidator:
     def __init__(self, column_set: ColumnSet):
         self._column_set = column_set
 
     def validate(self, column_name: str, values: Sequence[AnyType]) -> None:
         expected_type = self._column_set[column_name].type
-        val_func: Optional[Callable[[AnyType], None]]
+        is_valid_func: Optional[Callable[[AnyType], None]]
         match expected_type:
             case UUID():
-                val_func = self._validate_uuid
+                is_valid_func = self._valid_uuid
             case Int():
-                val_func = self._validate_int
+                is_valid_func = self._valid_int
             case UInt():
-                val_func = self._validate_int
+                is_valid_func = self._valid_int
             case Float():
-                val_func = self._validate_float
+                is_valid_func = self._valid_float
             case String():
-                val_func = self._validate_string
+                is_valid_func = self._valid_string
             case _:
-                raise NotImplementedError(
-                    f"ColumnValidator not implement for: {expected_type}"
-                )
+                raise InvalidColumnType(f"No validator for type: {expected_type}")
         for val in values:
-            val_func(val)
+            if is_valid_func(val):
+                continue
+            raise InvalidColumnType(
+                f"Invalid value {val} for column type {expected_type}"
+            )
 
-    def _validate_uuid(self, value: str) -> None:
-        assert uuid.UUID(str(value))
+    def _valid_uuid(self, value: str) -> bool:
+        try:
+            uuid.UUID(str(value))
+            return True
+        except ValueError:
+            return False
 
-    def _validate_int(self, value: int) -> None:
-        assert isinstance(value, int)
+    def _valid_int(self, value: int) -> bool:
+        return isinstance(value, int)
 
-    def _validate_float(self, value: float) -> None:
-        assert isinstance(value, float)
+    def _valid_float(self, value: float) -> bool:
+        return isinstance(value, float)
 
-    def _validate_string(self, value: str) -> None:
-        assert isinstance(value, str)
+    def _valid_string(self, value: str) -> bool:
+        return isinstance(value, str)
--- a/snuba/utils/schemas.py
+++ b/snuba/utils/schemas.py
@@ -718,7 +718,7 @@ class ColumnValidator:
 
     def validate(self, column_name: str, values: Sequence[AnyType]) -> None:
         expected_type = self._column_set[column_name].type
-        is_valid_func: Optional[Callable[[AnyType], None]]
+        is_valid_func: Optional[Callable[[AnyType], bool]]
         match expected_type:
             case UUID():
                 is_valid_func = self._valid_uuid
--- a/snuba/utils/schemas.py
+++ b/snuba/utils/schemas.py
@@ -725,7 +725,7 @@ class ColumnValidator:
             case Int():
                 is_valid_func = self._valid_int
             case UInt():
-                is_valid_func = self._valid_int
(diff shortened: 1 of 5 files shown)


snuba/utils/schemas.py
```diff
@@ -702,3 +703,39 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+            case Int():
+                val_func = self._validate_int
+            case UInt():
+                val_func = self._validate_int
+            case Float():
+                val_func = self._validate_float
+            case String():
+                val_func = self._validate_string
+            case _:
+                raise Exception("no type found")
```
MeredithAnya, line 727: gonna add a better error message here

snuba/utils/schemas.py
```diff
@@ -702,3 +705,41 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+                val_func = self._validate_string
+            case _:
+                raise NotImplementedError(
+                    f"ColumnValidator not implement for: {expected_type}"
+                )
+        for val in values:
+            val_func(val)
+
+    def _validate_uuid(self, value: str) -> None:
+        assert uuid.UUID(str(value))
```
kylemumma, line 736: is it fine for these to be asserts or should they be exceptions raised?

kylemumma (approved):
The only addition to this PR I think is missing is we probably need to add logic to the API endpoint to send failed validations as a 400. As-is it will be a 500 error right?

snuba/utils/schemas.py
```diff
@@ -702,3 +705,41 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
...
+    def validate(self, column_name: str, values: Sequence[AnyType]) -> None:
+        expected_type = self._column_set[column_name].type
+        val_func: Optional[Callable[[AnyType], None]]
+        match expected_type:
+            case UUID():
+                val_func = self._validate_uuid
+            case Int():
+                val_func = self._validate_int
+            case UInt():
+                val_func = self._validate_int
```
onkar, line 723: I am trying to understand the support of unsigned integers in the `schemas.py` - https://github.com/getsentry/snuba/blob/d4b164c06710e15bed33d868102305adf8fe495c/snuba/utils/schemas.py#L539 but it looks virtually similar to integer. IMO, validating unsigned integers in the same way as integers is most probably wrong. One way I can think of is to use `ctypes` - https://docs.python.org/3/library/ctypes.html#ctypes.c_uint but curious to know what you think.

tests/utils/test_columns_validator.py
```diff
@@ -0,0 +1,56 @@
...
+        Column("date_param", Date()),
+    ]
+)
+
+
+@pytest.mark.parametrize(
+    "column_name, values, is_valid",
+    [
+        # valid column types
+        pytest.param("uint_param", [1, 2, 3, 4], True),
```
onkar, line 34: If a column's type is unsigned integer and if user passes say -17 to it, we should disallow that because unsigned integers do not have negative values. Probably good to add a test case around that.

onkar (changes requested):
Mostly looks good. Added a comment around unsigned integer support. I don't know if this type must be supported but the current implementation appears to be wrong.

MeredithAnya:
> Mostly looks good. Added a comment around unsigned integer support. I don't know if this type must be supported but the current implementation appears to be wrong.

@onkar Yeah I was thinking about this, I was treating them same under some assumptions that we'd never actually be sending negative values for these, more that people might send the wrong type (a string instead of an int) but I should take this into account. 

I don't think this has to be super comprehensive in the validation, was thinking `assert isinstance(value, int) and value > 0` would be enough, thoughts?

snuba/utils/schemas.py
MeredithAnya: I actually might change these all to return a bool and then if it's false in the main validate function i'll raise a invalid column type exception

snuba/web/delete_query.py
```diff
@@ -148,8 +151,13 @@ def _delete_from_table(
         on_cluster=on_cluster,
         is_delete=True,
     )
-    _enforce_max_rows(query)
 
+    columns = storage.get_schema().get_columns()
+    column_validator = ColumnValidator(columns)
+    for col, values in conditions.items():
+        column_validator.validate(col, values)
```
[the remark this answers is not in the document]
MeredithAnya: I'm gonna change the `NotImplementedError` to be the same kind of exception as the invalid so I can re-raise here as an Invalid query error

onkar:
> > Mostly looks good. Added a comment around unsigned integer support. I don't know if this type must be supported but the current implementation appears to be wrong.
> 
> @onkar Yeah I was thinking about this, I was treating them same under some assumptions that we'd never actually be sending negative values for these, more that people might send the wrong type (a string instead of an int) but I should take this into account.
> 
> I don't think this has to be super comprehensive in the validation, was thinking `assert isinstance(value, int) and value > 0` would be enough, thoughts?

A couple of things:

1. Do we really need to support unsigned int data type? If we must, then it _may_ involve handling intricacies around the range. Python does not support it probably because almost all of the things can be done with plain integers.
2. Your validation might be sufficient. I did this using ctypes:
```
>>> import ctypes
>>> numPositive, numNegative = 100, -100
>>> numUInt = ctypes.c_uint(numPositive)
>>> numInt = ctypes.c_uint(numNegative)
>>> numUInt.value == numPositive
True
>>> numInt.value == numNegative
False 
```
An unsigned int will preserve its value through ctypes conversion while a signed integer won't.

MeredithAnya:
@onkar 

> 1. Do we really need to support unsigned int data type?

we use the [UInt type](https://github.com/getsentry/snuba/blob/d580e8e6afe8f9bcda47765650aca06bb649d588/snuba/utils/schemas.py#L539) in a lot of the schema definitions so it should be part of this validator. I actually think we only use `Int` in the new eap schemas just from a quick cmd f

However, my motivation around this validator is mostly around the UUID. I think if we actually see more errors from clickhouse about column types mismatch we can make this validator more robust, but I just wanted something that prevented the most obvious of errors

snuba/utils/schemas.py
```diff
@@ -702,3 +706,54 @@ def set_modifiers(self, modifiers: Optional[TModifiers]) -> Enum[TModifiers]:
 
     def get_raw(self) -> Enum[TModifiers]:
         return Enum(self.values)
+
+
+class InvalidColumnType(SerializableException):
```
xurui-c, line 711: Very minor nit: maybe `InvalidColumnTypeException` is clearer, but not important

xurui-c (approved):

onkar (approved):
LGTM 👍 , thanks for addressing comments and providing explanation!

kylemumma (approved):
LGTM

PR #6199, opened 2024-08-13
ref(snuba-deletes) don't run delete queries when 0 rows need to be deleted

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2842

Currently, if there are no rows that match the criteria specified by the delete query, the code will still go through the deletion pipeline (and any post-processing). This is unnecessary. Instead, let’s short-circuit and throw an error if there are no rows to delete.

### DIFF
--- a/snuba/query/exceptions.py
+++ b/snuba/query/exceptions.py
@@ -55,3 +55,7 @@ class QueryPlanException(SerializableException):
 class TooManyDeleteRowsException(SerializableException):
     def __init__(self, message: str):
         super().__init__(message)
+
+
+class NoRowsToDeleteException(SerializableException):
+    pass
--- a/snuba/web/delete_query.py
+++ b/snuba/web/delete_query.py
@@ -11,7 +11,7 @@ from snuba.query import SelectedExpression
 from snuba.query.conditions import combine_and_conditions
 from snuba.query.data_source.simple import Table
 from snuba.query.dsl import column, equals, in_cond, literal, literals_tuple
-from snuba.query.exceptions import TooManyDeleteRowsException
+from snuba.query.exceptions import NoRowsToDeleteException, TooManyDeleteRowsException
 from snuba.query.expressions import Expression, FunctionCall
 from snuba.query.query_settings import HTTPQuerySettings
 from snuba.reader import Result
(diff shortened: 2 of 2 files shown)


snuba/web/delete_query.py
```diff
@@ -147,7 +147,11 @@ def _delete_from_table(
         on_cluster=on_cluster,
         is_delete=True,
     )
-    _enforce_max_rows(query)
+    try:
+        _enforce_max_rows(query)
+    except NoRowsToDeleteException as e:
```
kylemumma, line 152: I think you forgot to raise this exception in `_enforce_max_rows`

kylemumma (changes requested):
I think you forgot to raise the new exception you created.

My other feedback would be that maybe we dont need to add the "extra message" to the result we can just stop early and return the normal notion of success, the user doesnt need to know that we stopped early, to them it can just look like it went fine. Maybe ask @MeredithAnya what she thinks about this.

snuba/query/exceptions.py
```diff
@@ -55,3 +55,8 @@ def from_args(cls, exception_type: str, message: str) -> "QueryPlanException":
 class TooManyDeleteRowsException(SerializableException):
     def __init__(self, message: str):
         super().__init__(message)
+
+
+class NoRowsToDeleteException(SerializableException):
+    def __init__(self) -> None:
+        super().__init__("No rows to delete")
```
MeredithAnya, line 62: I don't think we need the message here, I would just remove the `__init__`

snuba/web/delete_query.py
```diff
@@ -147,7 +147,11 @@ def _delete_from_table(
...
+        result: Result = {"extra_message": e.message}
```
MeredithAnya, line 153: Also I don't think we need the `extra_message` here either I think you can make a `Result` object that just has empty data like `{"data": [], "meta": []}`, the reader result is kind of weird anyway because deleting data doesn't return how many rows you deleted or anything

snuba/reader.py
```diff
@@ -31,6 +31,7 @@
         "totals": Row,
         "profile": Optional[Dict[str, Any]],
         "trace_output": str,
+        "extra_message": str,
```
kylemumma, line 34:
```suggestion
```

i think you should not add this

MeredithAnya (approved):

PR #6200, opened 2024-08-13
feat(eap): Add a proto definition for the aggregate bucket endpoint

colin-sentry:
Instead of a big complicated endpoint that returns multiple buckets, just let people query a single bucket at a time for now.


colin-sentry:
closed in favor of https://github.com/getsentry/snuba/pull/6202

PR #6201, opened 2024-08-13
feat(gen-metrics): Add sampling to gen metrics entity

john-z-yang:


enochtangg (approved):
LGTM, could we also add a small `gauges` unit test for sanity

PR #6204, opened 2024-08-14
feat(eap): stub API for aggregate bucket request

volokluev:
Providing an API to hit which will take the expected payload


colin-sentry (approved):

xurui-c (approved):

onewland:
Looking at the protobuf code at https://github.com/getsentry/snuba/tree/master/snuba/protobufs, is it automated how we build these models from the .proto files? Are things going to break if somebody changes a protobuf and forgets to run `make protos`? Could we make that a pre-commit check

Does `make develop` run `make protos`?

volokluev:
@onewland this is for a POC. your concerns are valid but we'll be moving this to the sentry protobuf repo once we have the prototype working

PR #6206, opened 2024-08-14
fix(eap-spans): Drop ineffective indices to speed up insertion

phacops:
Inserting data in this cluster is slow and having many indices is one of the reason. These indices we added were not as effective as we thought so it's harmless to remove them and will probably help insertion.

### DIFF
--- a/snuba/migrations/operations.py
+++ b/snuba/migrations/operations.py
@@ -550,7 +550,7 @@ class DropIndices(SqlOperation):
         self.__indices = indices
 
     def format_sql(self) -> str:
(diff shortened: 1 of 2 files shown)


volokluev (approved):

asottile-sentry:
this is breaking sentry

```
ValueError: Duplicate migration number for the following files:
    snuba/snuba_migrations/events_analytics_platform/0005_drop_attribute_key_project_id_indexes.py
    snuba/snuba_migrations/events_analytics_platform/0005_remove_attribute_mv.py
```

PR #6212, opened 2024-08-15
fix(eap-spans): Split migration to drop project_id index only

phacops:


snuba/snuba_migrations/events_analytics_platform/0006_drop_project_id_index.py
```diff
@@ -0,0 +1,35 @@
...
+            operations.AddIndex(
+                storage_set=StorageSetKey.EVENTS_ANALYTICS_PLATFORM,
+                table_name="eap_spans_local",
+                index_name="bf_project_id",
+                index_expression="project_id",
+                index_type="bloom_filter",
+                granularity=1,
+                target=operations.OperationTarget.LOCAL,
+            ),
+        ]
```
volokluev, line 35: what if the reverse was a noop?

snuba/snuba_migrations/events_analytics_platform/0007_drop_attribute_key_bucket_0_indexes.py
```diff
@@ -0,0 +1,31 @@
+from typing import Sequence
```
volokluev, line 1: can we leave this for another PR? I just want to try one index deletion first

evanh (approved):

PR #6213, opened 2024-08-15
fix(eap-spans): Drop indexes on attribute keys bucket 0

phacops:


phacops:
We won't be running this index removal, it takes too long.

phacops:
Added a way to run asynchronously.

evanh (approved):

PR #6216, opened 2024-08-15
feat(eap): Mutations consumer MVP

ayirr7:
Rolling out the mutations consumer that has general pattern of INSERT INTO .. SELECT FROM into Clickhouse, in order to simultaneously insert updated and cancelled/negated rows into Clickhouse to reflect a single mutation. 

This PR includes:
- The Rust consumer entrypoint
- The strategies that parse mutations and extract the updates to specific span attributes
- The strategies that map updates to attributes of old data already in the table 
- The actual write to Clickhouse


untitaker (approved):

snuba/cli/rust_consumer.py
```diff
@@ -140,6 +140,14 @@
     default=False,
     help="Enable async inserts for ClickHouse",
 )
+@click.option(
+    "--allow-mutability",
```
untitaker, line 144: probably rename this to something else (or ideally, really make a separate command

rust_snuba/src/consumer.rs
```diff
@@ -228,33 +232,63 @@ pub fn consumer_impl(
...
+        let mut_factory = MutConsumerStrategyFactory {
+            storage_config: first_storage,
+            env_config,
+            logical_topic_name,
+            max_batch_size,
+            max_batch_time,
+            processing_concurrency: ConcurrencyConfig::new(concurrency),
+            clickhouse_concurrency: ConcurrencyConfig::new(clickhouse_concurrency),
+            commitlog_concurrency: ConcurrencyConfig::new(2),
+            replacements_concurrency: ConcurrencyConfig::new(4),
```
untitaker, line 247: a few parameters like this do not have to be passed at all to the strategy, but i can clean that up in a followup pr

untitaker (changes requested):

rust_snuba/src/consumer.rs
```diff
@@ -228,33 +232,63 @@ pub fn consumer_impl(
...
+            async_inserts,
+            python_max_queue_depth,
+            use_rust_processor,
+            health_check_file: health_check_file.map(ToOwned::to_owned),
+            enforce_schema,
+            commit_log_producer,
+            replacements_config,
+            physical_consumer_group: consumer_group.to_owned(),
+            physical_topic_name: Topic::new(&consumer_config.raw_topic.physical_topic_name),
```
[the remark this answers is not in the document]
ayirr7: thanks for catching this

rust_snuba/src/consumer.rs
ayirr7: I removed those 2 concurrency configs as well as replacement config

snuba/cli/rust_consumer.py
ayirr7: separate command meaning a different entrypoint than rust_consumer?

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    for i in 0..20 {
+        attr_columns.push_str(&format!(",attr_num_{i} Map(String, Float64)"));
+        attr_combined_columns.push_str(&format!(
+            ",mapUpdate(old_data.attr_num_{i}, new_data.attr_num_{i}) AS attr_num_{i}"
+        ));
+    }
+
+    let mut body = format!(
```
untitaker, line 111: TODO: Write cancellation rows

snuba/cli/rust_consumer.py
untitaker: I would say in the short term we should rename this flag to something better, long term we need a completely separate CLI

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+fn format_query(table: &str, batch: &MutationBatch) -> Vec<u8> {
+    let mut attr_columns = String::new();
+    // attr_combined_columns is intentionally ordered the same as the clickhouse schema.
+    // INSERT INTO .. SELECT FROM .. matches up columns by position only, and ignores names.
+    // subqueries don't have this property.
+    let mut attr_combined_columns = String::new();
+    for i in 0..20 {
```
colin-sentry, line 96: 20 should be a constant somewhere

snuba/cli/rust_consumer.py
ayirr7: Renamed it, let me know what you think?

rust_snuba/src/arroyo_utils.rs
```diff
@@ -0,0 +1,17 @@
+use rust_arroyo::processing::strategies::InvalidMessage;
+/// Some helper functions that work around Arroyo's ergonomics, and should eventually make it into
```
ayirr7, line 2: @untitaker curious why this logic was added/ported to Snuba and maybe not Arroyo?

rust_snuba/src/consumer.rs
ayirr7: Removed some other stuff, let me know if you think there's more

rust_snuba/src/arroyo_utils.rs
untitaker: @ayirr7 it's a bunch of things that should eventually be moved to arroyo, but might require further thought. I don't think this function that I put here is generic enough to be put into arroyo as-is, but something *like this* should exist in arroyo.

In this case I'd like arroyo to have some way to generate InvalidMesage from a Message more easily, but I can't come up with a good design right now. I think this function is not one.

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -0,0 +1,158 @@
...
+    for i in 0..ATTRS_SHARD_FACTOR {
+        attr_columns.push_str(&format!(",attr_num_{i} Map(String, Float64)"));
+        attr_combined_columns.push_str(&format!(
+            ",mapUpdate(old_data.attr_num_{i}, new_data.attr_num_{i}) AS attr_num_{i}"
+        ));
+    }
+
+    let mut body = format!(
+        "
+        INSERT INTO {table}
```
ayirr7, line 113: @untitaker 2 thoughts on this query:

1. We can rewrite this to avoid JOIN by using the WITH context. We would need to look into the performance differences. 

2. What if we tried a UNION to get both the cancellation and the update in this one query? (SELECT ... UNION SELECT ...)

I'm also wondering if there's a way to do something like INSERT INTO ... VALUES (SELECT ...) (SELECT ...) so we can have 2 insertions back to back? 

Just some thoughts -- I can try these out on some local data

untitaker: I think WITH conflicts directly with input() in my tests. I was also trying to use WITH to make the cancellation rows happen (using both WITH and UNION ALL). I'm not sure how it would help to avoid the join?

If you have a specific query let's discuss 1:1 or just push it, if it works already.

> so we can have 2 insertions back to back?

then we should just merge those MutationBatches at JSON level, I don't think the query needs to adapt to accomodate larger batches.

ayirr7:
From local testing:

- `OPTIMIZE TABLE ... FINAL` for some reason clears out the table. I inserted 1 "normal" row and then inserted 1 "updated" row. It seems that running optimize somehow cancelled out the two, even though they had the same sign (+1).
-

untitaker:
@ayirr7 with the test command I gave you, it sends duplicate rows, and the rows contain timestamps that are immediately subjected to TTL.

rust_snuba/src/consumer.rs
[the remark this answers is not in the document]
untitaker: we decided to use `--raw-events-topic` to work around this issue for now

ayirr7:
I have published the new topic and its schema as follows: https://github.com/getsentry/sentry-kafka-schemas/pull/333. Will bump in Snuba tomorrow. 

As I see it, we'll have to modify the `process_message` function; we'll have to do something a bit more complicated than the current

```
let parsed: MutationMessage = serde_json::from_slice(payload)?;
```
As the payload will not be directly deserialized into the `MutationMessage` struct.

rust_snuba/src/processors/eap_spans.rs
```diff
@@ -82,19 +134,21 @@ fn fnv_1a(input: &[u8]) -> u32 {
 impl From<FromSpanMessage> for EAPSpan {
     fn from(from: FromSpanMessage) -> EAPSpan {
         let mut res = Self {
-            organization_id: from.organization_id,
+            primary_key: PrimaryKey {
```
ayirr7, line 137: @untitaker I was curious why exactly we added this here? Is it just so that we have some consistency between the mutations and EAP spans schema?

untitaker:
going to try and get this merged even though the query is not performant yet -- otherwise we'll keep fighting merge conflicts

untitaker (approved):

PR #6218, opened 2024-08-16
bug(mql): Fix WITH TOTALS for MQL queries

enochtangg:
### Overview
This PR is responsible for fixing a bug with the `totals` flag in MQL. When the `totals` flag is set to true, snuba will insert a `WITH TOTALS` modifier next to the `GROUP BY` in the query. Read more about WITH TOTALS [here](https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#with-totals-modifier). However, if the query does not contain a `GROUP BY` then the modifier will not be added to the query. As a result, the flag is ignored and CH does not return a totals row. When this happens, the query AST still has `totals` set to true, and during [result transformation](https://github.com/getsentry/snuba/blob/fc11af17f7558e56a4139d061f47bef0369de0b7/snuba/clickhouse/native.py#L472), snuba pops the last element of the data array and assumes it is the totals row (even though it isn't).

To fix this, in MQL parsing, we should only set and respect the totals flag only if there exists a group by. If not `WITH TOTALS` cannot be applied anyways, and therefore we should not set the flag.

### DIFF
--- a/snuba/query/mql/parser_supported_join.py
+++ b/snuba/query/mql/parser_supported_join.py
@@ -1301,7 +1301,6 @@ def populate_query_from_mql_context(
         )
         query.add_condition_to_ast(context_condition)
 
-        query.set_totals(with_totals)
         if orderby:
             query.set_ast_orderby([orderby])
 
@@ -1314,6 +1313,10 @@ def populate_query_from_mql_context(
                 query.set_ast_groupby(list(groupby) + [selected_time.expression])
             else:
                 query.set_ast_groupby([selected_time.expression])
+
+        if query.get_groupby():
+            # Only set WITH TOTALS if there is a group by.
+            query.set_totals(with_totals)
 
     if isinstance(query, CompositeQuery):
 
--- a/snuba/query/mql/parser_supported_join.py
+++ b/snuba/query/mql/parser_supported_join.py
@@ -1389,7 +1389,7 @@ def populate_query_from_mql_context(
             # ensure we correctly join the subqueries. The column names will be the same for all the
             # subqueries, so we just need to map all the table aliases.
             add_time_join_keys(join_clause)
-        elif query.has_totals() and no_groupby_or_one_sided_groupby:
+        elif with_totals and no_groupby_or_one_sided_groupby:
             # If formula query has no interval and no group by or a onesided groupby, but has totals, we need to convert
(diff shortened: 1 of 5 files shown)


evanh (approved):

snuba/query/mql/parser_supported_join.py
```diff
@@ -1315,6 +1314,10 @@ def populate_query_from_mql_context(
             else:
                 query.set_ast_groupby([selected_time.expression])
 
+        if query.get_groupby():
+            # Only set WITH TOTALS if there is a group by.
```
onkar, line 1318: Is this same issue applicable to `HAVING` clause? The [documentation](https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#with-totals-modifier) says that the behavior in case of `HAVING` depends on `totals_mode` setting and we have a few occurrences of `HAVING` clause in snuba code base.

tests/query/parser/test_parser.py
```diff
@@ -40,7 +40,7 @@ def test_mql() -> None:
             "orderby": "ASC",
             "granularity": 60,
             "interval": None,
-            "with_totals": "True",
+            "with_totals": "False",
```
onkar, line 43: These tests are setting `with_totals` to `False`. Do we need to add tests in `group by` to verify and assert that `with_totals` is set to `True`?

onkar (approved):
LGTM 👍 added a couple of questions, thanks!

snuba/query/mql/parser_supported_join.py
enochtangg: The `HAVING` clause would not apply here since MQL doesn't support it

tests/query/parser/test_parser.py
enochtangg: There are a few implicit tests that check for this in `test_metrics_mql_api.py`. I added a more explicit test here: https://github.com/getsentry/snuba/pull/6218/commits/316cdc623caee95be82de73b903e558db3d5e05b

onkar: Thanks for the clarification!

snuba/query/mql/parser_supported_join.py
onkar: Thanks for the clarification!

PR #6220, opened 2024-08-16
Added a few missing details on Snuba development environment README page

onkar:
I ran into issues regarding python 3.8.11 version and rust while setting up the local development environment using https://getsentry.github.io/snuba/contributing/environment.html. This PR fixes the documentation and makes it more consistent with what is being used in the team today.


onewland (approved):
nice, looks like a good set of improvements

PR #6221, opened 2024-08-19
ref(admin): Show raw trace output in the Admin tool

evanh:
Have an option to show the raw trace output in the Tracing tool, as well as the ability to copy just
the raw tracing output instead of the full JSON response.

raw-output


mcannizz (approved):

onkar (approved):
That was fast! LGTM 👍

PR #6226, opened 2024-08-20
ref(admin): Refactor formatted output for trace queries

evanh:
We have a better understanding now of the type of information that is important in a query trace.
Expose just that information in the backend, so it can be shown in the frontend.

Also update the frontend to show this new view summarized view. Make use of Mantine components
to organize things cleanly.

formatted


evanh:
Note this can't be merged yet, it will break the admin frontend.

onkar (approved):
LGTM 👍

snuba/admin/clickhouse/tracing.py
```diff
@@ -66,115 +29,10 @@ def run_query_and_get_trace(storage_name: str, query: str) -> TraceOutput:
     query_result = connection.execute(
         query=query, capture_trace=True, with_column_types=True
     )
-    formatted_trace_output = format_trace_output(query_result.trace_output)
+    summarized_trace_output = summarize_trace_output(query_result.trace_output)
     return TraceOutput(
         trace_output=query_result.trace_output,
-        formatted_trace_output=formatted_trace_output,
+        summarized_trace_output=summarized_trace_output,
```
onkar, line 35: Wondering if this will break the frontend [here](https://github.com/getsentry/snuba/blob/master/snuba/admin/static/tracing/query_display.tsx#L57)  now that you changed the key of the `TraceOutput` object.

onkar: Nevermind, just read your comment - https://github.com/getsentry/snuba/pull/6226#issuecomment-2299621214

PR #6229, opened 2024-08-21
Gather ClickHouse's profile event data in Snuba admin tracing tool

onkar:
ClickHouse's profile events have a lot of useful data that can be found [here](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/ProfileEvents.cpp). The query passed to tracing tool is appended with `SETTINGS log_profile_events=1` so that ClickHouse makes an entry in the `system.query_log` table. Then another query is made to scan `system.query_log` to get the profile event data. 

Finally, the Snuba admin frontend is modified to accommodate the extra information returned.

Testing done:
Screenshot 2024-08-23 at 12 14 28 PM

### DIFF
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -24,6 +24,10 @@ class InvalidStorageError(SerializableException):
     pass
 
 
+class ClickHouseTimeoutError(SerializableException):
+    pass
+
+
 def is_valid_node(
     host: str, port: int, cluster: ClickhouseCluster, storage_name: str
 ) -> bool:
--- a/snuba/admin/clickhouse/common.py
+++ b/snuba/admin/clickhouse/common.py
@@ -24,7 +24,7 @@ class InvalidStorageError(SerializableException):
     pass
 
 
-class ClickHouseTimeoutError(SerializableException):
+class ClickhouseTimeoutError(SerializableException):
     pass
 
 
--- a/snuba/admin/clickhouse/tracing.py
+++ b/snuba/admin/clickhouse/tracing.py
@@ -21,7 +21,7 @@ class TraceOutput:
     summarized_trace_output: TracingSummary
     cols: list[tuple[str, str]]
     num_rows_result: int
-    profile_events_results: dict[dict[str, str]]
+    profile_events_results: dict[dict[str, str], Any]
     profile_events_meta: list[Any]
     profile_events_profile: Dict[str, int]
 
--- a/snuba/admin/clickhouse/tracing.py
+++ b/snuba/admin/clickhouse/tracing.py
@@ -15,12 +15,20 @@ from snuba.clusters.cluster import ClickhouseClientSettings
 
 
 @dataclass
+class QueryTraceData:
+    host: str
+    port: int
+    query_id: str
+    node_name: str
+
+
+@dataclass
 class TraceOutput:
     trace_output: str
     summarized_trace_output: TracingSummary
     cols: list[tuple[str, str]]
     num_rows_result: int
-    profile_events_results: dict[dict[str, str], Any]
+    profile_events_results: dict[str, Any]
     profile_events_meta: list[Any]
     profile_events_profile: dict[str, int]
 
--- a/snuba/admin/clickhouse/tracing.py
+++ b/snuba/admin/clickhouse/tracing.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import re
 from dataclasses import dataclass
+from typing import Any, Dict, List
 
 from snuba.admin.clickhouse.common import (
     get_ro_query_node_connection,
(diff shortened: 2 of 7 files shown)


snuba/admin/clickhouse/tracing.py
```diff
@@ -19,6 +20,9 @@ class TraceOutput:
     summarized_trace_output: TracingSummary
     cols: list[tuple[str, str]]
     num_rows_result: int
+    profile_events_results: dict[dict[str, str], Any]
+    profile_events_meta: list[Any]
+    profile_events_profile: Dict[str, int]
```
evanh, line 25:
```suggestion
    profile_events_profile: dict[str, int]
```

snuba/admin/static/tracing/index.tsx
```diff
@@ -110,11 +117,25 @@ function TracingQueries(props: { api: Client }) {
...
+      profileEventRows.push(v.rows[0]);
+    }
+
     return (
       <ol style={collapsibleStyle} key={title + "-root"}>
+        <h4>Profile Events Output</h4>
+        {profileEventRows.map((line, index) => {
+          return (
+            <li key={title + index}>
+              <span>{line}</span>
```
evanh, line 134:
```suggestion
              <Text>{line}</Text>
```

Mantine provides a Text component which is easier to style.

snuba/admin/static/tracing/index.tsx
```diff
@@ -110,11 +117,25 @@ function TracingQueries(props: { api: Client }) {
...
     const parsedLines: Array<string> = value.split(/\n/);
 
+    const profileEventRows: Array<string> = [];
+    for (const [k, v] of Object.entries(profileEventResults)) {
+      profileEventRows.push(v.rows[0]);
+    }
+
     return (
       <ol style={collapsibleStyle} key={title + "-root"}>
+        <h4>Profile Events Output</h4>
```
evanh, line 130:
```suggestion
        <Title order={4}>Profile Events Output</Title>
```

Mantine provides this component which is easier to style.

snuba/admin/static/tracing/index.tsx
```diff
@@ -110,11 +117,25 @@ function TracingQueries(props: { api: Client }) {
...
+        {profileEventRows.map((line, index) => {
+          return (
+            <li key={title + index}>
+              <span>{line}</span>
+            </li>
+          );
+        })}
+        <h4>Trace Output</h4>
```
evanh, line 138:
```suggestion
        <Title order={4}>Trace Output</Title>
```

snuba/admin/views.py
```diff
@@ -441,8 +442,58 @@ def clickhouse_trace_query() -> Response:
         )
 
     try:
-        result = run_query_and_get_trace(storage, raw_sql)
-        return make_response(jsonify(asdict(result)), 200)
+        if not raw_sql.endswith("SETTINGS log_profile_events=1"):
```
evanh, line 445: You will have to handle the case where the input query already has a settings clause. E.g.

```
SELECT ...
FROM ...
SETTINGS max_threads=1
```

snuba/admin/views.py
```diff
@@ -441,8 +442,58 @@ def clickhouse_trace_query() -> Response:
...
+                # There is a race between the original query to trace finishing and the system query log populating.
+                # Sleep between the query executions.
+                system_query_result = run_system_query_on_host_with_sql(
+                    host_port_query_id_node["host"],
+                    int(host_port_query_id_node["port"]),
+                    storage,
+                    sql,
+                )
+                if not system_query_result.results:
+                    time.sleep(1)
```
evanh, line 467: I might do 0.1 here or maybe 0.5.

snuba/admin/views.py
```diff
@@ -441,8 +442,58 @@ def clickhouse_trace_query() -> Response:
...
+                    counter += 1
+                else:
+                    break
+
+            assert (
```
evanh, line 472: Can you handle this error a little better? I think right now this will end up returning a generic 500 error to the frontend with the message `Must get ProfileEvents!`. Maybe return a timeout error or something like that?

snuba/admin/clickhouse/tracing.py
onkar: Done.

snuba/admin/static/tracing/index.tsx
onkar: Done.

snuba/admin/views.py
onkar: Done.

snuba/admin/views.py
onkar: The smaller sleep intervals in milliseconds do not work. I tried that before sending out the PR but no data was returned in milliseconds of sleep. I had to increase the duration to seconds. And even then sometimes it took a few seconds to get the result. Dropping this suggestion.

snuba/admin/views.py
```diff
@@ -441,8 +442,65 @@ def clickhouse_trace_query() -> Response:
...
+            if columns:
+                res = {}
+                res["column_names"] = [name for name, _ in columns]
+                res["rows"] = []
+                for query_result in system_query_result.results:
+                    if query_result[0]:
+                        res["rows"].append(
+                            json.dumps(query_result[0])
+                            .replace("{", "")
+                            .replace("}", "")
```
onewland, line 498: this seems a little odd to me. We want to use `json.dumps` to generate a string that ultimately isn't JSON?

snuba/admin/clickhouse/common.py
```diff
@@ -24,6 +24,10 @@ class InvalidStorageError(SerializableException):
     pass
 
 
+class ClickHouseTimeoutError(SerializableException):
```
onewland, line 27: styling nit: we generally don't capitalize the H in ClickHouse in variable names e.g.

```suggestion
class ClickhouseTimeoutError(SerializableException):
```

snuba/admin/views.py
```diff
@@ -462,13 +520,53 @@ def clickhouse_trace_query() -> Response:
...
+            result.append(
+                {
+                    "host": local_nodes[0].get("host")
+                    if local_nodes
+                    else query_node.get("host"),  # type: ignore
+                    "port": local_nodes[0].get("port")
+                    if local_nodes
+                    else query_node.get("port"),  # type: ignore
+                    "query_id": query_summary.query_id,
+                    "node_name": node_name,
```
onewland, line 564: I'm not so concerned about this in /admin/ code but since you know all the field names in advance, this feels more like a dataclass than a dictionary use case.

snuba/admin/views.py
```diff
@@ -462,13 +520,53 @@ def clickhouse_trace_query() -> Response:
...
+                }
+            )
```
onewland, line 566: maybe it's not possible here, and I'm less picky about admin code, but I'd generally prefer that we build a value (like `result`) in a single list comprehension rather than appending a mutable accumulator

snuba/admin/views.py
```diff
@@ -441,8 +442,65 @@ def clickhouse_trace_query() -> Response:
...
+                )
+                if not system_query_result.results:
+                    time.sleep(1)
+                    counter += 1
+                else:
+                    break
+
+            if system_query_result is None or len(system_query_result.results) == 0:
+                raise ClickHouseTimeoutError(
+                    "Timed out while executing: {}".format(sql)
```
onewland, line 481: nit: I think this error message could be misleading. ClickHouse is serving these queries without timing out; we're just waiting for a table to get filled and that isn't happening (probably because query log isn't enabled or something). What you're timing out for is waiting for a query log row to be inserted.

I call this a nit because nothing else uses your `ClickHouseTimeoutError` class but it would be more important if that thing had wider use.

onewland (approved):
left some stylistic notes, mostly not blocking. 

I do think ClickHouseTimeoutError should be renamed before merge, but if you do that I'm fine with these changes going out (and I don't feel a need to re-review after that)

snuba/admin/views.py
onkar: Fair point. I converted this to a dataclass now.

snuba/admin/views.py
onkar: It's possible to use list comprehension and I incorporated this suggestion now.

snuba/admin/views.py
onkar: Thanks for catching this. I was probably mucking around while passing a string to the admin frontend and make that parsing work and this hack remained in there. I removed it now.

PR #6230, opened 2024-08-21
feat: CodeMirror

gggritso:
Replaces the text area in most query input with a proper CodeMirror editor.

- monospace
- syntax highlighting
- `localStorage` sync

**e.g.,**
Screenshot 2024-08-21 at 5 23 03 PM


volokluev:
OMGGGGG

snuba/admin/static/query_editor.tsx
```diff
@@ -48,7 +49,14 @@ function QueryEditor(props: {
...
 }) {
   const [query, setQuery] = useState<string>("");
-  const [queryTemplate, setQueryTemplate] = useState<string>("");
+  const hash = window.location.hash;
+
+  // Namespace the storage by the hash, which corresponds to the screen
+  const [queryTemplate, setQueryTemplate] = useLocalStorage<string>({
+    key: `${hash}-query-editor-query`,
+    defaultValue: "",
+  });
```
nikhars, line 58: Could you elaborate a bit on the need for this part of the code? More specifically, why do you use the hash of the window location?

Also, is there a reason you decided to use `useLocalStorage` here instead of `useState`?

nikhars (approved):
Left a minor comment for my own understanding. But that should not block from merging

snuba/admin/static/query_editor.tsx
gggritso: `useLocalStorage` saves the query to [Web Storage](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API) so if you refresh the window, you still have your query! A few devs have complained of lost work if they refresh the browser window, this prevents that

As for the hash, the hash makes it possible to store the queries separately per page. So, if you go to "ClickHouse Tracing", you have another query there separate from for example the "System Queries" screen. Without a hash, they share the same storage. Using the window hash is the easiest way to do this, since each page uses a different hash (e.g., `/#snuba-explain` vs. `/#tracing` and so on)

PR #6231, opened 2024-08-22
feat(snuba-admin): Add query history

wmak:
- Adds history to queries so that even when you refresh your last 5 queries will still be on the page

https://github.com/user-attachments/assets/7c053617-952f-42a4-9f7f-cfae071b0595

### DIFF
--- a/snuba/admin/static/cardinality_analyzer/query_display.tsx
+++ b/snuba/admin/static/cardinality_analyzer/query_display.tsx
@@ -4,6 +4,7 @@ import { Collapse } from "SnubaAdmin/collapse";
 import { CSV } from "SnubaAdmin/cardinality_analyzer/CSV";
 import QueryEditor from "SnubaAdmin/query_editor";
 import ExecuteButton from "SnubaAdmin/utils/execute_button";
+import { getRecentHistory, setRecentHistory } from "SnubaAdmin/query_history";
 
 import {
   CardinalityQueryRequest,
@@ -17,6 +18,7 @@ enum ClipboardFormats {
 }
 type QueryState = Partial<CardinalityQueryRequest>;
 
+const HISTORY_KEY = "cardinality_analyzer";
 function QueryDisplay(props: {
   api: Client;
   resultDataPopulator: (queryResult: CardinalityQueryResult) => JSX.Element;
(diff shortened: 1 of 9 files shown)


snuba/admin/static/cardinality_analyzer/query_display.tsx
```diff
@@ -25,7 +27,7 @@ function QueryDisplay(props: {
   const [query, setQuery] = useState<QueryState>({});
   const [queryResultHistory, setCardinalityQueryResultHistory] = useState<
     CardinalityQueryResult[]
-  >([]);
+  >(getRecentHistory(HISTORY_KEY));
```
nikhars, line 30: Nit: I wonder if we could use `window.location.href` instead of defining different HISTORY_KEY in different files. I see that the href is different on different pages

```
On tracing tag
console.log(window.location.href)
https://snuba-admin.getsentry.net/#tracing

On system queries tab
console.log(window.location.href)
https://snuba-admin.getsentry.net/#system-queries

On cardinality analyzer tab
console.log(window.location.href)
https://snuba-admin.getsentry.net/#cardinality-analyzer
```

nikhars (commented):
Left a minor comment. But overall looks good. Thanks for improving snuba admin

snuba/admin/static/cardinality_analyzer/query_display.tsx
wmak: I think this would cause really unexpected behaviour if we ever changed the urls (suddenly losing history). And the current behaviour of defining a key means you can be explicit about it and reuse results between pages if there's ever a use-case for that

nikhars (approved):

snuba/admin/static/tests/query_history.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+        {"something": 2},
+      ]);
+    });
+    it("different keys don't conflict", () => {
+      const TEST_KEY1 = "test3"
+      const TEST_KEY2 = "test4"
+      setRecentHistory(TEST_KEY1, {"something": "foo"});
+      expect(getRecentHistory(TEST_KEY2)).toStrictEqual([]);
+    });
+  });
```
onkar, line 41: Is it possible to add a test where `setRecentHistory()` raises exception and test confirms that `storageKey` is removed successfully from `localStorage`?

onkar (approved):
LGTM, learned a few things through this PR. Left an ask of an additional unit test.

snuba/admin/static/tests/query_history.spec.tsx
wmak: Not easily, i think it'll add a lot of complexity to test that unfortunately. If that becomes a problem maybe its something we can revisit in the future

PR #6238, opened 2024-08-22
feat(admin): create an api to overwrite a snuba migration state

volokluev:
Yes this is kind of a hack but asking OPS to run these commands every time we mess up a migration is driving me and @phacops insane

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -243,8 +243,16 @@ def force_overwrite_migration_status(
         logger.error(err, exc_info=True)
         return make_response(jsonify({"error": "Group not found"}), 400)
 
-    migration_group = MigrationGroup(group)
     runner.force_overwrite_status(migration_group, migration_id, Status(new_status))
+    user = request.headers.get(USER_HEADER_KEY)
+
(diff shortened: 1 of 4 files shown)


phacops (approved):

snuba/admin/views.py
```diff
@@ -228,6 +229,26 @@ def reverse_migration(group: str, migration_id: str) -> Response:
     )
 
 
+@application.route(
+    "/migrations/<group>/overwrite/<migration_id>/status/<new_status>",
+    methods=["POST"],
+)
+@check_tool_perms(tools=[AdminTools.MIGRATIONS])
```
onewland, line 236: is this as restrictive a permission as we can get? because this seems more dangerous than just being able to run predefined scripts.

onewland (commented):
we should log this somewhere visibly (should probably show up in #feed-sns and the audit log)

snuba/admin/views.py
volokluev: ugh so boring

onewland (approved):

PR #6248, opened 2024-08-28
ref(admin): include delete policies in capacity mgmt

MeredithAnya:
Adds delete allocation policies to show up in snuba admin per storage along side the other allocation policies.
> Screenshot 2024-08-28 at 3 54 24 PM

<br/>

If there are no delete query allocation policies listed it will look like the following:
> Screenshot 2024-08-28 at 3 54 42 PM

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -74,8 +74,8 @@ from snuba.migrations.connect import check_for_inactive_replicas
 from snuba.migrations.errors import InactiveClickhouseReplica, MigrationError
 from snuba.migrations.groups import MigrationGroup, get_group_readiness_state
 from snuba.migrations.runner import MigrationKey, Runner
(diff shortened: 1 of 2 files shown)


volokluev:
I would like to see a visual differentiation in the UI for delete policies and query poicies given that they are not the same thing.

MeredithAnya:
> I would like to see a visual differentiation in the UI for delete policies and query poicies given that they are not the same thing.

I do think it would be useful to see all the allocation policies (delete and select queries alike) for a storage together since they all affect capacity management. That being said I think a simple header for each category (delete query, select query) could separate the policies.

@volokluev Is that inline with what you were expecting?

volokluev:
> That being said I think a simple header for each category (delete query, select query) could separate the policies.

Yup that's good enough

volokluev (approved):

PR #6249, opened 2024-08-28
feat(eap): Add an RPC for span samples, refactor a lot

colin-sentry:
This lays the foundation for a lot of the shared snuba RPC logic, including evaluating column names and filters.

### DIFF
--- a/snuba/web/rpc/common.py
+++ b/snuba/web/rpc/common.py
@@ -10,11 +10,11 @@ from sentry_protos.snuba.v1alpha.trace_item_filter_pb2 import (
     TraceItemFilter,
 )
 
-from snuba.query import Expression, Query
+from snuba.query import Query
 from snuba.query.conditions import combine_and_conditions, combine_or_conditions
 from snuba.query.dsl import Functions as f
 from snuba.query.dsl import and_cond, column, in_cond, literal, literals_array, or_cond
-from snuba.query.expressions import FunctionCall, SubscriptableReference
+from snuba.query.expressions import Expression, FunctionCall, SubscriptableReference
 from snuba.web.rpc.exceptions import BadSnubaRPCRequestException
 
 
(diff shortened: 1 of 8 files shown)


phacops (approved):

snuba/web/rpc/common.py
```diff
@@ -0,0 +1,281 @@
...
+            column("project_id"),
+            literals_array(
+                None, [literal(k) for k in context.project_ids_to_names.keys()]
+            ),
+            literals_array(
+                None, [literal(v) for v in context.project_ids_to_names.values()]
+            ),
+            literal("unknown"),
+            alias="project_name",
+        )
```
volokluev, line 104: Why are we doing this? What's the problem with querying by project id?

snuba/web/rpc/common.py
```diff
@@ -0,0 +1,281 @@
...
+        return f.CAST(
+            SubscriptableReference(
+                alias=attr_key.name,
+                column=column("attr_num"),
+                key=literal(attr_key.name),
+            ),
+            "Int64",
+        )
+    if attr_key.type == AttributeKey.Type.TYPE_BOOLEAN:
+        return f.CAST(
```
volokluev, line 153: I'm seeing a lot of casts in this function, why are they necessary?

snuba/web/rpc/common.py
colin-sentry: people want to do things like order by, which wouldn't make sense if it was ordering by project id

snuba/web/rpc/common.py
colin-sentry: if you request a boolean, this returns a boolean? it'd be weird if you requested a boolean and got a number back

snuba/web/rpc/common.py
volokluev: People are trying to order by project name specifically?

colin-sentry: that's what product said

colin-sentry: there's a table that can contain project name, and if they click the carat they would like it to actually order by project name - this logic currently lives in the sentry backend, we can probably optimize it by making a `Join` table for project_name <> project_id, since that seems like it'll come up again in the future

snuba/web/rpc/common.py
```diff
@@ -0,0 +1,256 @@
...
+            return combine_or_conditions(exp.parameters)
+        else:
+            return exp
+
+    query.transform_expressions(transform)
+
+
+# These are the columns which aren't stored in attr_str_ nor attr_num_ in clickhouse
+NORMALIZED_COLUMNS: Final[Mapping[str, AttributeKey.Type.ValueType]] = {
+    "project_name": AttributeKey.Type.TYPE_STRING,  # extra special column, derived from a context sent with the request
```
volokluev, line 48:
```suggestion
```

volokluev (approved):

PR #6251, opened 2024-08-29
Snuba admin sudo mode for system commands

fpacifici:
There has been a lot of requirests for prodction access and SRE tickets
to execute clickhouse system commands that were not supported in the
snuba admin.

This includes commands like: SYSTEM and ALTER.

These commands do not expose customer data. 
They were not exposed at the beginning because they are as safe as
playing with an armed nuclear warhead.

Now snuba has RBAC and an audit log. The Sns team can consider revisiting
this decision.

This PR implements the sudo mode for the system queries
tool in the snuba admin:
- a new parameter is passed to http://127.0.0.1:1219/run_clickhouse_system_query. Which is called `sudo` 
- if the parameter is true the user is trying to execute a query in sudo mode
- If the user has the required role, the system would accept SYSTEM and ALTER statements
- SYSTEM KILL and SYSTEM SHUTDOWN are still not allowed as the servers cannot be restarted without external intervention.
- The role is assigned to sns and ops so far.
- if the parameter is false and a system query is attempted, the behavior is the same as before.
- Using sudo mode writes an entry in the audit log.

The UI (my web design abilities stopped improving in 1997 on Internet Explorer 4)
Screenshot 2024-08-28 at 11 57 39 PM

### DIFF
--- a/snuba/admin/auth_roles.py
+++ b/snuba/admin/auth_roles.py
@@ -82,6 +82,7 @@ TOOL_RESOURCES = {
     "cardinality-analyzer": ToolResource("cardinality-analyzer"),
     "production-queries": ToolResource("production-queries"),
     "system-queries": ToolResource("system-queries"),
+    "sudo-system-queries": ToolResource("system-queries"),
     "clickhouse-migrations": ToolResource("clickhouse-migrations"),
     "snuba-explain": ToolResource("snuba-explain"),
     "all": ToolResource("all"),
@@ -107,6 +108,14 @@ class ExecuteNonBlockingAction(MigrationAction):
 
 
 class ExecuteNoneAction(MigrationAction):
+    pass
+
+
+class ExecuteSudoSystemQuery(ToolAction):
+    """
+    Allow specific users to run system queries with elevated permissions
+    """
+
     pass
 
 
@@ -173,6 +182,10 @@ ROLES = {
         name="AllMigrationsExecutor",
         actions={ExecuteAllAction(list(MIGRATIONS_RESOURCES.values()))},
     ),
+    "ClickhouseAdmin": Role(
+        name="clickhouse-admin",
+        actions={ExecuteSudoSystemQuery([TOOL_RESOURCES["sudo-system-queries"]])},
+    ),
 }
 
 DEFAULT_ROLES = [
@@ -184,3 +197,4 @@ DEFAULT_ROLES = [
 
 if settings.TESTING or settings.DEBUG:
     DEFAULT_ROLES.append(ROLES["AllTools"])
+    DEFAULT_ROLES.append(ROLES["ClickhouseAdmin"])
--- a/snuba/admin/clickhouse/system_queries.py
+++ b/snuba/admin/clickhouse/system_queries.py
@@ -2,14 +2,25 @@ import re
 
 from clickhouse_driver.errors import ErrorCodes
 
-from snuba.admin.clickhouse.common import InvalidCustomQuery, get_ro_node_connection
+from snuba.admin.auth_roles import ExecuteSudoSystemQuery
+from snuba.admin.clickhouse.common import (
+    InvalidCustomQuery,
+    get_ro_node_connection,
+    get_sudo_node_connection,
+)
+from snuba.admin.user import AdminUser
 from snuba.clickhouse.errors import ClickhouseError
 from snuba.clickhouse.native import ClickhouseResult
 from snuba.clusters.cluster import ClickhouseClientSettings
+from snuba.utils.serializable_exception import SerializableException
+
+
+class UnauthorizedForSudo(SerializableException):
+    pass
 
 
 def _run_sql_query_on_host(
-    clickhouse_host: str, clickhouse_port: int, storage_name: str, sql: str
+    clickhouse_host: str, clickhouse_port: int, storage_name: str, sql: str, sudo: bool
 ) -> ClickhouseResult:
     """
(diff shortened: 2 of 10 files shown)


snuba/admin/views.py
```diff
@@ -390,20 +393,26 @@ def clickhouse_system_query() -> Response:
...
     try:
-        result = run_system_query_on_host_with_sql(host, port, storage, raw_sql)
+        result = run_system_query_on_host_with_sql(
+            host, port, storage, raw_sql, sudo_mode, g.user
+        )
         rows = []
         rows, columns = cast(List[List[str]], result.results), result.meta
 
-        if columns:
+        if columns is not None:
```
fpacifici, line 407: I am not sure about this change. Specifically I don't know if testing `if columns` rather than nullability was intentional. If yes what was the scenario ?

snuba/admin/auth_roles.py
```diff
@@ -173,6 +182,10 @@ def generate_tool_test_role(tool: str) -> Role:
         name="AllMigrationsExecutor",
         actions={ExecuteAllAction(list(MIGRATIONS_RESOURCES.values()))},
     ),
+    "ClickhouseAdmin": Role(
+        name="clickhouse-admin",
+        actions={ExecuteSudoSystemQuery([TOOL_RESOURCES["sudo-system-queries"]])},
+    ),
```
fpacifici, line 188: This role does not need specify resources as of today. It could be extended in order to specify which clusters to use. Should I just remove the ROLL_RESOURCE ?

onewland:
Looks fine to me but I think the test failure is legitimate: https://github.com/getsentry/snuba/actions/runs/10610316363/job/29407963144?pr=6251#step:6:2099

fpacifici:
> Looks fine to me but I think the test failure is legitimate: https://github.com/getsentry/snuba/actions/runs/10610316363/job/29407963144?pr=6251#step:6:2099

Right. I thought I had ran that test.

snuba/admin/clickhouse/system_queries.py
```diff
@@ -102,21 +145,66 @@ def is_query_describe(sql_query: str) -> bool:
...
+    match = ALTER_QUERY_RE.match(sql_query)
+    return True if match else False
+
+
+def _can_user_sudo(user: AdminUser) -> bool:
+    for role in user.roles:
+        for action in role.actions:
+            if isinstance(action, ExecuteSudoSystemQuery):
+                return True
+    return False
```
onkar, line 171: nit: Something like this will be more pythonic.

```
return any(isinstance(action, ExecuteSudoSystemQuery) for role in user.roles for action in role.actions)
```

snuba/admin/views.py
onkar: Can't speak for the intention but I think `if columns` would cover both cases like `columns is not None` and `len(columns) != 0`. AFAIK, `columns` is most likely a `tuple`. Probably that it why it was added. My suggestion is to keep the original line as is to avoid surprises.

snuba/admin/views.py
```diff
@@ -472,6 +481,8 @@ def clickhouse_trace_query() -> Response:
                     int(query_trace_data.port),
                     storage,
                     sql,
+                    False,
+                    g.user,
```
onkar, line 485: Ditto here about the naming of `g.user`

snuba/admin/views.py
```diff
@@ -395,20 +398,26 @@ def clickhouse_system_query() -> Response:
...
         storage = req["storage"]
         raw_sql = req["sql"]
+        sudo_mode = req.get("sudo", False)
     except KeyError:
         return make_response(jsonify({"error": "Invalid request"}), 400)
 
     try:
-        result = run_system_query_on_host_with_sql(host, port, storage, raw_sql)
+        result = run_system_query_on_host_with_sql(
+            host, port, storage, raw_sql, sudo_mode, g.user
```
onkar, line 407: I am having trouble understanding what `g.user` means, can we rename it to something more readable?

onkar (commented):
LGTM overall, added a few comments and nits.

snuba/admin/views.py
fpacifici: The original would not work as, for system statements the columns list is empty.  Which makes sense as the response is not a table.

That's why I was wondering if there was a reason for raising an error in case of empty columns list. I cannot think about one.

snuba/admin/views.py
fpacifici: `g` in flask is the special variable provided by the framework to store request scoped data. 
So I cannot rename it. It is provided by the framework. https://flask.palletsprojects.com/en/2.3.x/api/#flask.g

snuba/admin/views.py
fpacifici: See https://github.com/getsentry/snuba/pull/6251#discussion_r1736826390

snuba/admin/views.py
onkar: Thanks for the link. I didn't know it existed.

onkar (approved):
LGTM, thanks for addressing review comments.

snuba/admin/views.py
volokluev: The result for the system queries tool often is a table:

```
SELECT * FROM system.mutations WHERE is_done=0
```

Usually when we're using this endpoint we are indeed returning a table

fpacifici: So, changing should not be an issue,  right as it was not implemented this way "to" prevent non table results, but "because" results were tables ?

PR #6256, opened 2024-08-29
More comprehensive payload from CapMan

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2852

Allocation policies are our mechanism for doing traffic management for Snuba queries. This PR adds more fields to the payload returned by Snuba, so we can filter more efficiently in Sentry traces

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -847,6 +847,26 @@ def _add_quota_info(
             quota_info["throttle_threshold"] = quota_allowance.throttle_threshold
 
 
+def _populate_query_status(
+    summary: dict[str, Any],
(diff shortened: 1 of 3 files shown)


onewland (approved):
Seems fine, but just so I understand:

Is the point here to make it so that "is_throttled" is always found in the same place in the payload? Since under individual policies you have to crawl the whole tree to figure out which ones exist

xurui-c:
What do you mean by "same place"?

The goal is more generic. If the prod dev wants to filter out queries that are rejected, throttled, or successful, they can just do `allocation_policy.is_rejected:True` or `allocation_policy.is_throtteld:True` or `allocation_policy.is_successful:True`.

Issue #5823, opened 2024-04-26
CODEOWNERS file is broken


asottile-sentry:
currently codeowners is not being considered due to errors in the file: https://github.com/getsentry/snuba/blob/master/CODEOWNERS

PR #6259, opened 2024-09-02
Fix Snuba CODEOWNERS

onkar:
Resolves issue #5823

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -19,7 +19,7 @@
 /snuba/snuba_migrations/events-analytics-platform @getsentry/events-analytics-platform @getsentry/owners-snuba
 
 # rust_snuba
-/rust_snuba @getsentry/owners-snuba @getsentry/processing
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -19,19 +19,19 @@
 /snuba/snuba_migrations/events-analytics-platform @getsentry/events-analytics-platform @getsentry/owners-snuba
 
 # rust_snuba
-/rust_snuba @getsentry/owners-snuba @getsentry/processing
```
onkar, line 22: I could not find `@getsentry/processing` team but `@getsentry/product-owners-processing` [team](https://github.com/orgs/getsentry/teams/product-owners-processing) exists. Should I keep `product-owners-processing` here?

onewland: I think dropping this line is fine (since fallback is owners-snuba). Processing was helping us with Rust when we had less experience with it, but that team has been folded into ingest. They don't need special approval access anymore.

PR #6261, opened 2024-09-03
Emits DD metric for queries statuses

xurui-c:
We currently emit these metrics on a per-policy level, which means some queries can be counted multiple times for multiple policies. This PR adds metrics so we only count each query once. This will allow us to get a more accurate sense of the % of queries that are successful/throttled/rejected

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -936,7 +936,22 @@ def _apply_allocation_policies_quota(
         stats["quota_allowance"]["summary"] = summary
 
         if not can_run:
+            metrics.increment(
+                "rejected_query",
+                tags={"storage_key": allocation_policies[0].storage_key.value},
+            )
             raise AllocationPolicyViolations.from_args(stats["quota_allowance"])
+
+        if throttle_quota_and_policy is not None:
+            metrics.increment(
+                "throttled_query",
+                tags={"storage_key": allocation_policies[0].storage_key.value},
+            )
+        else:
+            metrics.increment(
+                "successful_query",
+                tags={"storage_key": allocation_policies[0].storage_key.value},
+            )
         # Before allocation policies were a thing, the query pipeline would apply
         # thread limits in a query processor. That is not necessary if there
         # is an allocation_policy in place but nobody has removed that code yet.


snuba/web/db_query.py
```diff
@@ -936,7 +936,34 @@ def _apply_allocation_policies_quota(
         stats["quota_allowance"]["summary"] = summary
 
         if not can_run:
+            metrics.increment(
+                "rejected_query",
+                tags={
+                    "referrer": str(
+                        attribution_info.tenant_ids.get("referrer", "no_referrer")
```
onewland, line 943: referrer seems kind of high-cardinality as a metric tag. do we use it elsewhere?

xurui-c: Yes: https://github.com/getsentry/snuba/blob/74e80f24835f42707d5b1b23ca3c74949957e8a0/snuba/query/allocation_policies/__init__.py#L791

side note: `db_request_rejected` and `db_request_throttled` are the metrics that I mentioned in the PR description that can be counted multiple times for a single query

xurui-c: Now that I think about it, since `db_request_rejected` and `db_request_throttled` already has the referrer tag, I don't need them here

snuba/web/db_query.py
```diff
@@ -936,7 +936,19 @@ def _apply_allocation_policies_quota(
         stats["quota_allowance"]["summary"] = summary
 
         if not can_run:
+            metrics.increment(
+                "rejected_query",
+            )
```
volokluev, line 941: you're not tagging with anything? You should at least tag with dataset or storage key, otherwise this metric is not that useful

xurui-c: I'll add the tags, but why isn't a metric like this useful to get a volume % breakdown of successful/rejected/throttled queries?

volokluev: It is useful for that! And if you think about how snuba is used and how queries are throttled, it's all based on the actual data that is being accessed. If you roll up all the throttled queries together then you won't know how many queries were throttled on errors or replays. It will all just be one number and it won't really be helpful. You can still sum across all tags to get the volume % breakdown of course

volokluev (approved):

onewland (approved):

PR #6262, opened 2024-09-04
feat(eap): Re-add the span attributes meta table

colin-sentry:
Pierre and I had a call with Altinity, they suggested adding an attributes table (like the one Evan had proposed in https://github.com/getsentry/snuba/pull/6152)

The big difference is adding `trace_id UUID` to the destination of the materialized view, which altinity says can improve query performance with a subquery:
```
SELECT * FROM eap_spans_local WHERE organization_id=1234 AND (_sort_timestamp, trace_id) IN (
    SELECT any(_sort_timestamp) as ts, trace_id FROM spans_str_attrs WHERE (organization_id, attr_key, attr_val) IN ((1234, 'key_1', 'value_1'), (1234, 'key_2', 'value_2'))
    GROUP BY trace_id 
    HAVING groupBitOr(attr_key = 'key_1' AND attr_val = 'value_1') AND groupBitOr(attr_key = 'key_2' AND attr_val = 'value_2') AND max_duration > 1000
    ORDER BY timestamp DESC LIMIT 1000
)
```


snuba/snuba_migrations/generic_metrics/0059_distributions_aggregated_add_weighted_columns.py
```diff
@@ -21,7 +21,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 AggregateFunction(
                     "quantilesTDigestWeighted(0.5, 0.75, 0.9, 0.95, 0.99)",
                     [Float(64), UInt(64)],
-                    MigrationModifiers(codecs=["ZSTD(1)"]),
+                    modifiers=MigrationModifiers(codecs=["ZSTD(1)"]),
```
volokluev, line 24: was this change necessary? Was there a bug in it?

snuba/utils/schemas.py
```diff
@@ -466,11 +466,13 @@ def __init__(
         self,
         func: str,
         arg_types: Sequence[ColumnType[TModifiers]],
+        simple: bool = False,  # SimpleAggregateFunction is faster for things like sum
```
volokluev, line 469: The [list of SimpleAggregateFunction](https://clickhouse.com/docs/en/sql-reference/data-types/simpleaggregatefunction)(s) is quite small, why not either:

*  just detect those functions here and set `simple` automatically
* Create a SimpleAggregateFunction class so that it maps to the clickhouse version

Right now this requires that the user of the class know what `SimpleAggregateFunction` is AND it is hiding it behind an ambiguous flag called `simple`

colin-sentry: I can set it by default in another PR, was scared of side effects

snuba/snuba_migrations/events_analytics_platform/0011_span_attribute_table.py
```diff
@@ -0,0 +1,126 @@
...
+    project_id,
+    trace_id,
+    attrs.1 as attr_key,
+    attrs.2 as attr_value,
+    toStartOfDay(_sort_timestamp) AS timestamp,
+    retention_days,
+    1 AS count
+FROM eap_spans_local
+LEFT ARRAY JOIN
+    arrayConcat({",".join(f"CAST(attr_str_{n}, 'Array(Tuple(String, String))')" for n in range(ATTRIBUTE_BUCKETS))}) AS attrs
```
volokluev, line 91: What is the thinking behind arrayConcat vs how Evan did it?

https://github.com/getsentry/snuba/pull/6152/files#diff-8174280b297958e24d650d9555da2a5026239db01793cf1ab6be4146ef504bbdR94-R99

volokluev (approved):
Just one question about the transformation but otherwise this looks good. Matviews are fairly easy to drop anyways

snuba/snuba_migrations/events_analytics_platform/0011_span_attribute_table.py
colin-sentry: Evan still did arrayconcat ([L22 of that file](https://github.com/getsentry/snuba/pull/6152/files#diff-8174280b297958e24d650d9555da2a5026239db01793cf1ab6be4146ef504bbdR22)), this just does it with one less copy since the tuples can be shared as-is, instead of being split and re-joined into a list

PR #6263, opened 2024-09-04
Fix snuba admin trace query execution

nachivrn:
Snuba admin changes to append ProfileEvents to the trace query output were timing out after 10 seconds and not returning the trace query output. The fix involves updating the Snuba admin trace query execution to wait 60 seconds for retrieving ProfileEvents from both storage and query nodes where the query was executed. If the ProfileEvents aren't available within that time, it will return just the trace query output.

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -520,19 +520,20 @@ def clickhouse_trace_query() -> Response:
                 else:
                     break
 
-            query_trace.profile_events_meta.append(system_query_result.meta)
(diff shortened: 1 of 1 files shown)


volokluev (approved):

onkar:
Do you have to fix this unit test https://github.com/getsentry/snuba/blob/master/tests/admin/test_api.py#L262? Also, plesae make sure that the UI works with this change if you have not done it already.

PR #6268, opened 2024-09-05
Fix snuba admin's query tracing to connect to right storage and query nodes

onkar:
Snuba admin's tracing tool shows query trace output. In addition to that, it will now show the profile events data. This PR parses the trace output and builds a `dict` of node name to query id. Then system query is executed on those nodes to get the profile events for the corresponding query id.

Before connecting to the nodes, a socket connection is attempted to see if the hostname resolves. If it does not, then the node is assumed to be `127.0.0.1`. This is required because while running locally, snuba admin process will connect to clickhouse container. But the container's hostname does not resolve, defaulting to `127.0.0.1 `as the hostname.

To make sure hostnames in the CI jobs resolve, the CI docker-compose file is updated.


snuba/admin/views.py
```diff
@@ -520,41 +553,37 @@ def clickhouse_trace_query() -> Response:
             400,
         )
     except ClickhouseError as err:
+        logger.error(err, exc_info=True)
```
onkar, line 556: We should probably add logger.error() in all except blocks in this file. I found it quite useful in debugging multiple failures.

docker-compose.gcb.yml
```diff
@@ -17,7 +17,7 @@ x-test-common: &test-common
     - ".artifacts:/.artifacts"
   environment:
     SNUBA_SETTINGS: "$SNUBA_SETTINGS"
-    CLICKHOUSE_HOST: clickhouse
+    CLICKHOUSE_HOST: clickhouse.dev.ci
```
asottile-sentry, line 20: `.ci` is a *real* TLD (I know because I own a few of them!) -- please pick something that isn't real or use the example domain (`.local` I believe is the standard answer for this)

asottile-sentry (dismissed):
and actually this does not seem like a good idea at all.  what is the problem you're solving?   having everyone edit `/etc/hosts` on their mac is not a good solution and is going to cause endless pain for developers

onkar:
> and actually this does not seem like a good idea at all. what is the problem you're solving? having everyone edit `/etc/hosts` on their mac is not a good solution and is going to cause endless pain for developers

@asottile-sentry thanks for the review. Let's take step back to the larger problem. I will edit the description if this explanation makes sense.

At Sentry, snuba developers use a tool called snuba admin. It has a query tracing section that allows snuba developers to _trace_ how a query is executed in clickhouse. We wanted to enhance that tool to also show clickhouse's [profile events](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/ProfileEvents.cpp). Without this enhancement, snuba developers have to do these manual steps to get profile events:

1. Manually copy query ids from the raw trace output.
2. Navigate to system query dashboard.
3. Type in a different query to get profile events and replace query-id (one at a time) with the strings gathered in step 1.
4. As if this isn't enough, choose various hosts from the dropdown in systems query dashboard and repeat step 3.

You can see that these manual steps are error-prone and there is scope to automate them.

The query trace output has a filed called `summarized_trace_output`. Here is a sample of it:
```
{
    "module": "snuba.admin.views",
    "event": "summarized_trace_output = TracingSummary(query_summaries={'ebaf1a40d262': QuerySummary(node_name='ebaf1a40d262', is_distributed=True, query_id='fc3e8017-51ce-47c0-bb0f-209470fb70c8', execute_summaries=[ExecuteSummary(rows_read=1, memory_size='4.01 KiB', seconds=0.007365, rows_per_second=135.77732518669382, bytes_per_second='544.17 KiB')], select_summaries=None, index_summaries=None, stream_summaries=None, aggregation_summaries=None, sorting_summaries=None)})",
    "severity": "info",
    "user_ip": "127.0.0.1",
    "endpoint": "clickhouse_trace_query",
    "timestamp": "2024-09-05T03:47:13.227537Z"
}
```
The `node_name='ebaf1a40d262'` is the container id/hostname of the clickhouse container that snuba admin tool wants to connect to run system queries to get profile events. It has these ports forwarded:
`127.0.0.1:8123->8123/tcp, 127.0.0.1:9000->9000/tcp, 127.0.0.1:9009->9009/tcp`. In order to reliably connect to this host, we need it to have a proper name that we control (not a container id that docker runtime chooses). In order to resolve that hostname, we need entry in `/etc/hosts` file. Without name resolution, running system query gives an error like `"Host clickhouse.dev.local and port 9000 are not valid."`

To be clear, not every developer needs this entry in `/etc/hosts`. Only those that want to fix issues/enhance snuba admin while running it locally. In the CI pipeline, the configuration is different and so I made changes to the CI clickhouse hostname.

asottile-sentry:
wouldn't it be simpler, and a better experience, to write a tool which does all of the manual translation automatically (including replacing with the appropriate localhost / loopback addresses? -- we already utilize `host.docker.internal` for example)

snuba/settings/__init__.py
```diff
@@ -82,7 +82,7 @@
 
 CLUSTERS: Sequence[Mapping[str, Any]] = [
     {
-        "host": os.environ.get("CLICKHOUSE_HOST", "127.0.0.1"),
+        "host": os.environ.get("CLICKHOUSE_HOST", "clickhouse.dev.local"),
```
untitaker, line 85: if this line is reverted, won't it be no longer necessary to add this name to `/etc/hosts`?

onkar: As discussed on our call, I will replace `"clickhouse.dev.local"` with `"localhost"` and we _won't_ need edits to `/etc/hosts`. I will also update https://github.com/getsentry/sentry/pull/77125 to set the hostname to `localhost`. That should resolve your and @asottile-sentry's concerns.

asottile-sentry: `localhost` is problematic as it can mean v6 `::0` on some machines and we don't run dualstack (in fact we run v4-only).  we previously removed all references to `localhost` specifically because it broke some sentry developer setups

onkar: @asottile-sentry Hmm, how do you suggest we handle this then? The problem is that snuba admin is not run as a docker container when developing locally, it is spawned as a process by running `snuba admin` command. This is because snuba developers want to use the latest snuba on their Mac as opposed to the nightly build generated snuba (or whatever is in `master` branch). This means that in order to connect `snuba admin` (not in container) to clickhouse container locally, we need hostname of clickhouse that is recognized outside of docker. But it seems like neither `localhost` nor `clickhouse.dev.local` with additional path in `/etc/hosts` is a viable solution.
We don't have this problem in production because snuba admin is packaged in the `sentry_snuba` container.

onkar:
> wouldn't it be simpler, and a better experience, to write a tool which does all of the manual translation automatically (including replacing with the appropriate localhost / loopback addresses? -- we already utilize `host.docker.internal` for example)

I resolved the issues based on the feedback in our call. No network settings need to be changed to make this work now.

untitaker (approved):

snuba/admin/views.py
```diff
@@ -498,14 +500,46 @@ def clickhouse_trace_query() -> Response:
...
             sql = profile_events_raw_sql.format(query_trace_data.query_id)
             logger.info(
-                "Profile event gathering host: {}, port = {}, storage = {}, sql = {}, g.user = {}".format(
+                "Gathering profile event using host: {}, port = {}, storage = {}, sql = {}, g.user = {}".format(
                     query_trace_data.host, query_trace_data.port, storage, sql, g.user
                 )
             )
-            # TODO: Onkar to add the profile event logic later.
+            system_query_result, counter = None, 0
+            while counter < 60:
```
nachivrn, line 511: lets change to 30 seconds. API server timesout in 30 seconds.

nachivrn (approved):

docker-compose.gcb.yml
```diff
@@ -85,6 +92,9 @@ services:
     depends_on:
       - zookeeper
     image: "${CLICKHOUSE_IMAGE:-ghcr.io/getsentry/image-mirror-altinity-clickhouse-server:23.3.19.33.altinitystable}"
+    hostname: clickhouse-query.local
+    extra_hosts:
+      - "clickhouse-query.local:127.0.0.1"  # Add entry to /etc/hosts file
```
mcannizz, line 97: I suggest rewording the comment to explain why this is desirable, rather than restating what the code does.

mcannizz (approved):

PR #6273, opened 2024-09-06
Update docker entrypoint to run heaptrack 

ayirr7:
Updating this to run heaptrack if the corresponding environment variable is set. Otherwise just run the Snuba process by default.


ayirr7:
We might want to try running `heaptrack` at the same time of launching the snuba process like this:

```
if [[ $* == *--heaptrack* ]]; then
  for arg do
    shift
    [ "$arg" = "--heaptrack" ] && continue
    set -- "$@" "$arg"
  done

  exec "heaptrack $*"
else
  exec "$@"
fi
```

But this doesn't seem to work, it errors out with `FileNotFoundError: [Errno 2] No such file or directory: '/usr/src/snuba/snuba/cli/snuba.py` indicating that heaptrack is being passed a CLI arg to Snuba when it should be `heaptrack snuba ...` instead.

docker_entrypoint.sh
```diff
@@ -18,4 +18,8 @@ else
   printf "\n${help_result}"
 fi
 
+if [[ -n "${ENABLE_HEAPTRACK}" ]]; then
```
ayirr7, line 21: The idea here is to only run heaptrack if the string value for the env var is NOT null

asottile-sentry, line 21:
```suggestion
if [ -n "${ENABLE_HEAPTRACK:-}" ]; then
```

avoid unbound variables

(arguably this shell script should have `set -u` and `set -o pipefail` as well but that's out of scope for this patch)

PR #6274, opened 2024-09-06
cleanup(capman): remove legacy table rate limits

volokluev:
Before the capacity management framework we would apply rate limits as query processors in the query pipeline. That is no longer necessary. The TableRateLimit was the last holdout of these legacy rate limits but it is no longer necessary for these reasons:

1. The existing policies cover all the necessary rate limiting and throttling under normal operations
2. When we lose a clickhouse replica or the cluster capacity diminished drastically, it is the clickhouse nodes themselves which reject the query, not the table rate limit. 

Functionally, this table rate limit is wasted code and CPU cycles


onewland (approved):

onkar (approved):
Thanks for the cleanup. Was it cathartic? Granted that such deletions should be carefully done but I wonder if we can leverage sentry-autofix-experimental (see https://github.com/getsentry/snuba/pull/6275) to do such cleanups in the future.

PR #6276, opened 2024-09-09
Fixing Snuba Admin trace UI error.

nachivrn:
When the execute summaries field in the response payload is empty, the trace UI code attempts to access the value and throws an error. The fix adds a null check before retrieving the value.


onkar (approved):
Can you share the query where the summary response is empty?

PR #6222, opened 2024-08-19
ref(admin): Standardize Execute Query buttons

evanh:
This removes a lot of duplicate logic around handling errors and ensures that every page has a
proper loading indicator.

execute-button

PR #6280, opened 2024-09-09
bug(admin): Fix invalid query error alerting in snuba admin

enochtangg:
There was a bug in https://github.com/getsentry/snuba/pull/6222 which incorrectly parsed the `err` string in `snuba/admin/static/utils/execute_button.tsx`. The PR fixes that to ensure that query errors are presented back to the user in admin.

### DIFF
--- a/snuba/admin/static/cardinality_analyzer/query_display.tsx
+++ b/snuba/admin/static/cardinality_analyzer/query_display.tsx
@@ -65,7 +65,7 @@ function QueryDisplay(props: {
       .executeCardinalityQuery(query as CardinalityQueryRequest)
       .then((result) => {
         result.input_query = query.sql || "<Input Query>";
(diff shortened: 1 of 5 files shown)


snuba/admin/static/cardinality_analyzer/query_display.tsx
```diff
@@ -65,7 +65,7 @@ function QueryDisplay(props: {
       .executeCardinalityQuery(query as CardinalityQueryRequest)
       .then((result) => {
         result.input_query = query.sql || "<Input Query>";
-        setRecentHistory(HISTORY_KEY, result)
+        setRecentHistory(HISTORY_KEY, result);
```
volokluev, line 68: damn we really need a linter here 😅

volokluev (approved):

onkar (approved):

PR #6281, opened 2024-09-09
feat(job-runner): create a new `snuba jobs` command

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2868

Snuba runs migrations automatically on deploy. Clickhouse migrations are operating on a large distributed system and things can go wrong which require manual intervention. Migration repair efforts have the goal of allowing SnS to run DDL operations on Clickhouse without SRE intervention

The `Snuba Jobs` command is the first step towards allowing us to trigger jobs from Snuba Admin.

Executing dry run queries:
Screenshot 2024-09-10 at 2 11 23 PM

Executing non dry run queries:
Screenshot 2024-09-10 at 2 11 59 PM


volokluev:
how would this command be triggered from snuba admin?

snuba/cli/jobs.py
```diff
@@ -0,0 +1,21 @@
...
+    is_flag=True,
+    default=True,
+)
+@click.argument("pairs", nargs=-1)
+def jobs(*, dry_run: bool, pairs: Tuple[str, ...]) -> None:
+    kwargs = {}
+    for pair in pairs:
+        k, v = pair.split("=")
+        kwargs[k] = v
+    job_to_run = ToyJob(dry_run, **kwargs)
```
onewland, line 20: I think you might want to include the logic in this PR for taking in the job type as an argument and resolving it to a class. e.g. the command should be `snuba jobs run ToyJob arg1=val1`. This resolution could be done in a follow-up but I think it's the next step anyway

snuba/cli/jobs.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@click.command()
+@click.option(
+    "--dry_run",
+    is_flag=True,
+    default=True,
+)
+@click.argument("pairs", nargs=-1)
+def jobs(*, dry_run: bool, pairs: Tuple[str, ...]) -> None:
```
onewland, line 15: I think what you're doing here is a subcommand like `snuba jobs run`

I could imagine, for example, there also being e.g. a `snuba jobs list` or `snuba jobs status`

onewland (commented):
I think we're on the same page. I do feel that it should be `snuba jobs run`, not just `snuba jobs` though

onewland:
> how would this command be triggered from snuba admin?

The way I had suggested to Rachel that we approach this is to create a headless/testing version of running jobs first, and then wire the logic into snuba admin later. As long as it's not wired to snuba admin, then we don't have to worry about e.g. people running dangerous jobs without permission in production. But obviously the decisions we make should make this easy to support in snuba admin after.

snuba/manual_jobs/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Job(ABC, metaclass=RegisteredClass):
+    def __init__(self, dry_run: bool, **kwargs) -> None:
+        self.dry_run = dry_run
+        for k, v in kwargs.items():
+            setattr(self, k, v)
+
+    @abstractmethod
+    def _build_query(self) -> str:
+        pass
```
onewland, line 14: I don't think this should be a method on the abstract class (I can imagine some jobs may not result in ClickHouse queries)

snuba/cli/jobs.py
```diff
@@ -0,0 +1,32 @@
...
+    "--dry_run",
+    default=True,
+)
+@click.argument("pairs", nargs=-1)
+def run(*, job_name: str, dry_run: bool, pairs: Tuple[str, ...]) -> None:
+    job_type_class = Job.class_from_name(job_name)
+    if job_type_class is None:
+        raise click.ClickException(
+            f"Job does not exist. Did you make a file {job_name}.py yet?"
+        )
```
onewland, line 25: I think this and line 31 should be combined into a class that lives in snuba/manual_jobs that is something like

`JobLoader.get_job_instance(class_name, **kwargs)`

because that will be useful for both snuba admin and this command. Then, this file should only contain logic related to parsing stuff from the command line

onewland (commented):
mostly looks good if tests are passing, would just ask for the one refactor in my other comment

snuba/manual_jobs/__init__.py
```diff
@@ -0,0 +1,29 @@
+import os
+from abc import ABC, abstractmethod
+from typing import Any, cast
+
+from snuba.utils.registered_class import RegisteredClass, import_submodules_in_directory
+
+
+class Job(ABC, metaclass=RegisteredClass):
```
onewland, line 8: I don't think the base class/interface belongs in __init__ here. 

Let's separate this out into snuba/manual_jobs/base.py or snuba/manual_jobs/job.py or something. In general `__init__` should not contain much (I do think it's reasonable enough for it to contain the import statement below, though).

snuba/manual_jobs/ToyJob.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any
```
onewland, line 1: filename here should be toy_job.py, I think, to match python conventions

snuba/manual_jobs/__init__.py
onewland: I guess we use this pattern with `ClickhouseQueryProcessor`. Maybe this can stay

onewland: yeah, we can ignore this for now

snuba/manual_jobs/job_loader.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class JobLoader:
+    @staticmethod
+    def get_job_instance(class_name: str, dry_run: bool, **kwargs: Any) -> "Job":
+        job_type_class = Job.class_from_name(class_name)
+        if job_type_class is None:
+            raise click.ClickException(
+                f"Job does not exist. Did you make a file {class_name}.py yet?"
+            )
```
onewland, line 15: nit: this should be replaced with a non-Click exception (since this method may not be called from the CLI `snuba jobs` in the future). But we've gone through a number of rounds on this and I think it can wait for a later iteration

onewland (approved):

PR #6284, opened 2024-09-10
feat(eap): make mapContains work with EAP dataset

colin-sentry:
This hashes the key before it's sent over to clickhouse.


tests/query/processors/test_hash_bucket_functions_processor.py
```diff
@@ -184,6 +184,36 @@
...
+            selected_columns=[
+                SelectedExpression(
+                    "unrelated",
+                    Column(None, None, "column2"),
+                ),
+            ],
+            condition=binary_condition(
+                "or",
+                f.mapContains(column("attr_str_2"), literal("blah"), alias="x"),
+                f.mapContains(column("attr_strz"), literal("blah"), alias="z"),
```
nachivrn, line 213: Is the test expecting that attr_strz column in the map should not be modified?

nachivrn (approved):

PR #6285, opened 2024-09-10
bug(query): Run entity validators in composite query pipeline

enochtangg:
This PR contains a fix for this sentry [issue](https://sentry.sentry.io/issues/5825035976/?project=300688&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=4&utc=true). In the composite query pipeline, snuba is not running entity validators. As a result, join queries with missing columns (e.g. mql formula queries with unresolved tag keys) are reaching ClickHouse and affecting our SLO. Instead, the entity validator should catch these invalid queries and fail accordingly.

### DIFF
--- a/snuba/pipeline/stages/query_processing.py
+++ b/snuba/pipeline/stages/query_processing.py
@@ -42,6 +42,7 @@ class EntityProcessingStage(
         elif isinstance(query, CompositeQuery):
             # if we were not able to translate the storage query earlier and we got to this point, this is
             # definitely a composite entity query
+            run_entity_validators(query, pipe_input.query_settings)
             return translate_composite_query(
                 cast(CompositeQuery[Entity], query),
                 pipe_input.query_settings,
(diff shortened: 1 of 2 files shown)


volokluev (approved):
nice

snuba/pipeline/stages/query_processing.py
```diff
@@ -34,10 +34,10 @@ def _process_data(
         if translated_storage_query:
             return translated_storage_query
 
+        run_entity_validators(cast(EntityQuery, query), pipe_input.query_settings)
```
onkar, line 37: Before this change, entity validation was done only if the query was `LogicalQuery` and from clause had an `Entity`. But now with this change, that check is ignored and the validation is done outside of the `isinstance()` check. Does the validation work for `CompositeQuery`?

PR #6288, opened 2024-09-11
feat(job-runner): enable `snuba jobs` command to run from JSON manifest

xurui-c:
https://getsentry.atlassian.net/browse/SNS-2872

The SnS engineer will make JSON manifests in the ops repo. This PR enables us to pass in the file path of a JSON manifest file to the `Snuba Jobs` command

Next step will be validation within ManifestReader

Screenshot 2024-09-16 at 4 01 41 PM


snuba/cli/jobs.py
```diff
@@ -1,28 +1,54 @@
...
+@click.option("--job_id")
 @click.option(
     "--dry_run",
     default=True,
 )
-@click.argument("pairs", nargs=-1)
-def run(*, job_name: str, dry_run: bool, pairs: Tuple[str, ...]) -> None:
+def run_from_manifest(*, json_manifest: str, job_id: str, dry_run: bool) -> None:
+    job_specs = ManifestReader.read(json_manifest)
+    for job_spec in job_specs:
```
xurui-c, line 34: Maybe ManifestReader.read can return a map of job id to job spec

snuba/manual_jobs/job_loader.py
```diff
@@ -1,15 +1,20 @@
...
 class JobLoader:
     @staticmethod
-    def get_job_instance(class_name: str, dry_run: bool, **kwargs: Any) -> "Job":
-        job_type_class = Job.class_from_name(class_name)
+    def get_job_instance(job_spec: JobSpec, dry_run: bool) -> "Job":
+        job_type_class = Job.class_from_name(job_spec.job_type)
         if job_type_class is None:
             raise Exception(
-                f"Job does not exist. Did you make a file {class_name}.py yet?"
+                f"Job does not exist. Did you make a file {job_spec.job_type}.py yet?"
```
xurui-c, line 17: will make a snake case to camel case later

snuba/manual_jobs/manifest_reader.py
```diff
@@ -1,18 +1,30 @@
...
         with open(os.path.join(local_root, filename)) as stream:
             contents = simplejson.loads(stream.read())
-            assert isinstance(contents, Sequence)
-            return contents
+
+            job_specs = []
+            for content in contents:
+                content_dict = dict(zip(content[::2], content[1::2]))
+                job_spec = JobSpec(
+                    job_id=typing.cast(str, content_dict.get("id")),
```
onewland, line 21: I think it might make sense to do `typing.assert_type` here so we reject bad input (instead of just satisfying mypy) https://docs.python.org/3/library/typing.html#typing.assert_type

snuba/cli/jobs.py
onewland: I'd be fine with that

onewland (commented):
I think this is looking good so far. Is it meant to be a draft PR?

snuba/manual_jobs/toy_job.py
```diff
@@ -16,4 +18,10 @@ def _build_query(self) -> str:
             return "not dry run query"
 
     def execute(self) -> None:
-        click.echo("executing query `" + self._build_query() + "`")
+        click.echo(
```
onewland, line 21: nit that doesn't block merge: we shouldn't use `click` anything in this namespace because that's specific to CLI usage, and in prod we expect this to be invoked by snuba admin (let's logger.info in a later PR)

snuba/manual_jobs/manifest_reader.py
```diff
@@ -1,18 +1,34 @@
...
-from typing import Any, Sequence
+from typing import Mapping
 
 import simplejson
 
+from snuba.manual_jobs import JobSpec
 
-class _ManifestReader:
+
+class ManifestReader:
```
onewland, line 9: is there a reason that you want to make `ManifestReader` "public"? I'd prefer to expose as little as possible.

not a blocker for merge

onewland (approved):

snuba/manual_jobs/manifest_reader.py
xurui-c: at one point I was using it outside, but then that got deleted and i forgot to change it back to private

nikhars (changes requested):
This PR is changing 30k lines of code. Are you sure you are committing only the changes relevant to the PR?

PR #6289, opened 2024-09-11
Bump Arroyo to 2.17.6

ayirr7:
Bumped the Python requirements, not sure if anything needs to be changed on the Rust side because of https://github.com/getsentry/snuba/blob/master/rust_snuba/Cargo.toml#L34

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ pytest-watch==4.2.0
 python-dateutil==2.8.2
 python-rapidjson==1.8
 redis==4.3.4
-sentry-arroyo==2.17.1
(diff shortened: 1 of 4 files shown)


untitaker:
you also need to run `cargo update -p`. also, you need to wait for getsentry/pypi to finish before CI can pass

onkar (commented):
Did we do any testing to make sure that the new version works as expected? Please share the testing details in the PR.

untitaker:
@onkar the changes in arroyo are not significant enough to cause concern like this. CI should be enough.

untitaker (approved):
thank you for your contribution!

PR #6291, opened 2024-09-11
fix: store event ID (it is not in tags)

colin-sentry:
For some reason event ID is sent normalized and not in any _tags fields, pull it out explicitly.


rust_snuba/src/processors/eap_spans.rs
```diff
@@ -133,6 +133,9 @@ impl From<FromSpanMessage> for EAPSpan {
                 attr_num_buckets[(fnv_1a(k.as_bytes()) as usize) % attr_num_buckets.len()]
                     .insert(k.clone(), v);
             };
+            if let Some(event_id) = from.event_id {
+                insert_string("sentry.event_id".into(), event_id.to_string())
```
phacops, line 137: Can you remove dashes from it?

phacops (changes requested):

mcannizz:
What's the motivation behind this change?

PR #6292, opened 2024-09-12
feat(eap): add virtual column support

volokluev:
Add support for virtual columns to allow for sorting by project name or semver release (as opposed to release id).

### DIFF
--- a/snuba/web/rpc/common.py
+++ b/snuba/web/rpc/common.py
@@ -187,9 +187,6 @@ def apply_virtual_columns(
             return expression
         context = mapped_column_to_context.get(str(expression.key.value))
         attr_str = NestedColumn("attr_str")
-        import pdb
-
-        pdb.set_trace()
         if context:
             if context.from_column_name in NORMALIZED_COLUMNS:
                 return f.transform(
--- a/snuba/web/rpc/common.py
+++ b/snuba/web/rpc/common.py
@@ -175,7 +175,7 @@ def apply_virtual_columns(
         if not isinstance(expression, SubscriptableReference):
             return expression
 
-        if not expression.column.column_name != "attr_str":
+        if expression.column.column_name != "attr_str":
             return expression
         context = mapped_column_to_context.get(str(expression.key.value))
         if context:
--- a/snuba/web/rpc/common.py
+++ b/snuba/web/rpc/common.py
@@ -1,4 +1,4 @@
-from typing import Final, Mapping, Set
+from typing import Final, Mapping, Sequence, Set
 
 from sentry_protos.snuba.v1alpha.request_common_pb2 import RequestMeta
 from sentry_protos.snuba.v1alpha.trace_item_attribute_pb2 import (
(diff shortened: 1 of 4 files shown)


mcannizz:
Does this feature work for only the `span_samples_query` or do other operations like `timeseries_query` also support it?

snuba/web/rpc/common.py
```diff
@@ -130,6 +141,83 @@ def attribute_key_to_expression(attr_key: AttributeKey) -> Expression:
...
+        # virtual columns will show up as `attr_str[virtual_column_name]` or `attr_num[virtual_column_name]`
+        if not isinstance(expression, SubscriptableReference):
+            return expression
+
+        if not expression.column != "attr_str":
+            return expression
+        context = mapped_column_to_context.get(str(expression.key.value))
+        attr_str = NestedColumn("attr_str")
+        if context:
+            if context.from_column_name in NORMALIZED_COLUMNS:
```
colin-sentry, line 191: instead of special casing this, can't you do like:

`f.transform(f.CAST(attribute_key_to_expression(request_name, String), 'String', ...)`

snuba/web/rpc/common.py
```diff
@@ -130,6 +141,83 @@ def attribute_key_to_expression(attr_key: AttributeKey) -> Expression:
...
+
+        contexts:
+            [   {from_column_name: project_id, to_column_name: project_name, value_map: {1: "sentry", 2: "snuba"}} ]
+
+
+        Query will be transformed into:
+
+        SELECT
+        -- see the project name column transformed and the value mapping injected
+          transform( CAST( project_id, 'String'), array( '1', '2'), array( 'sentry', 'snuba'), 'unknown') AS `project_name`,
```
colin-sentry, line 168: also I think cast should go in other way, if you request a numeric column it'll be better performance if we cast '1' instead of project_id

snuba/web/rpc/common.py
volokluev: yes good point

volokluev:
@mcannizz How would virtual columns work in a timeseries? I don't think I understand

snuba/web/rpc/common.py
volokluev: I agree with the performance being better if we cast `'1'` rather than `project_id`. I was doing this to keep the logic and interface simpler

mcannizz:
> @mcannizz How would virtual columns work in a timeseries? I don't think I understand

Suppose you wanted to count events over time by project -- one time series per project -- and you wanted to sort those by project name. Would we support that? (I'm asking to understand the scope of the feature as implemented, not assert a requirement.)

volokluev:
I see, right now that is not supported here and the protobuf interface for timeseries doesn't support it either

colin-sentry (approved):
would be nice if test had multiple spans to make sure order_by actually worked, but it's probably fine

tests/web/rpc/test_span_samples.py
```diff
@@ -231,3 +232,103 @@ def test_booleans_and_number_compares(self, setup_teardown: Any) -> None:
...
+        assert [
+            dict((k, x.results[k].val_str) for k in x.results)
+            for x in response.span_samples
+        ] == [
+            {
+                "project_name": "sentry",
+                "sentry.sdk.name": "sentry.python.django",
+                "release_version": "4.2.0.69",
+            }
+            for _ in range(60)
```
nachivrn, line 289: Question: Is the first 60 response sample going to return project_name sentry?

nachivrn (approved):

PR #6293, opened 2024-09-12
feat(profiling): add a function_examples dataset

viglia:
For both transaction-based and continuous profiling, we'd like to add a way to provide examples of profiles which generated some metrics. 

The idea is, when we're visualizing profile functions metrics, the user, for example while browsing the slowest functions, would be able to click on them and see a list of (clickable) profiles associated with those values. 

For transaction-based profiling we used to store the example along with the metrics directly in the `functions` dataset.

Since we have deprecated the `functions` dataset and moved our metrics to the generic metrics dataset, we now need a solution for storing these examples, both for the transaction-based and the continuous profiling.

This PR would add a new dataset called `functions_examples` that serves this purpose.

Trying to keep the PR small. _Processor_ and _consumer_ will be added in separate _PRs_


snuba/snuba_migrations/functions_summaries/0001_functions_summaries_create_table.py
```diff
@@ -0,0 +1,81 @@
...
+storage_set_name = StorageSetKey.FUNCTIONS_SUMMARIES
+local_table_name = "functions_summaries_local"
+dist_table_name = "functions_summaries_dist"
+
+columns: List[Column[Modifiers]] = [
+    Column("project_id", UInt(64)),
+    # profile_id is nullable since this will only be used by transaction-based profiling
+    Column("profile_id", UUID(Modifiers(nullable=True, low_cardinality=False))),
+    # profiler_id is nullable since this will only be used by continuous profiling
+    Column("profiler_id", UUID(Modifiers(nullable=True, low_cardinality=False))),
```
phacops, line 19: A nullable is a bit like storing 2 columns where one is a boolean `is_null`. Maybe we can have 1 column with the UUID and a boolean `is_continuous` instead.

snuba/snuba_migrations/functions_summaries/0001_functions_summaries_create_table.py
```diff
@@ -0,0 +1,81 @@
...
+    Column("transaction_name", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("fingerprint", UInt(64)),
+    Column("name", String()),
+    Column("package", String()),
+    Column("thread_id", String()),
+    Column("min", Float(64)),
+    Column("max", Float(64)),
+    Column("sum", Float(64)),
+    Column("count", UInt(64)),
+    Column("end_timestamp", DateTime()),
```
phacops, line 30: In order to fetch a profile for continuous profiling, we need a start and end timestamp. I believe this is to match an example of a function to a profile, right? I think we might need to record another timestamp then.

snuba/snuba_migrations/functions_summaries/0001_functions_summaries_create_table.py
```diff
@@ -0,0 +1,80 @@
...
+from snuba.migrations.operations import OperationTarget, SqlOperation
+from snuba.utils.schemas import Float
+
+storage_set_name = StorageSetKey.FUNCTIONS_SUMMARIES
+local_table_name = "functions_summaries_local"
+dist_table_name = "functions_summaries_dist"
+
+columns: List[Column[Modifiers]] = [
+    Column("project_id", UInt(64)),
+    Column("profile_id", UUID(Modifiers(nullable=True, low_cardinality=False))),
```
phacops, line 16: Then we don't need the `nullable` since it'd be filled with either `profiler_id` or `profile_id` depending if it's continuous.

viglia: yep, forgot to remove!

evanh (commented):
Can you link to the design doc for this?

viglia:
> Can you link to the design doc for this?

@evanh this is the Notion [doc](https://www.notion.so/sentry/Functions-Summaries-Dataset-f05137b960b048e2b45731445fb02c01) I'm using to keep track of this work.

I've explicitly avoided talking about schema details in the notion doc and rather have the technical details discussed here as feedback so that it's coupled with this PR for future reference.

PR #6299, opened 2024-09-12
Update migrations list command to show migrations that no longer exist in the codebase

### LINKED ISSUE #2159: migrations list does not show migrations not in list
### Environment

We are running an older version of Sentry, self-hosted on Kubernetes. After a botched upgrade, we had issues with out Snuba Transactions Consumer. The first symptoms where a `OFFSET_OUT_OF_RANGE` error. And after clearing the queue we got a 
```
Snuba.clickhouse.errors.ClickhouseWriterError: [49] Block structure mismatch in RemoteBlockOutputStream stream: different types:
```

Thats all, nothing about what types and such. We decided to check on the logs of ClickHouse and found 2 logs with contain the queried types and the actual types in the DB with a mis match. 

So we thought at first that we might have missed some migrations and ran `snuba migrations list`. It showed all migrations as done. Not able to place this, we looked at all the versions of everything and noticed the the actual schema was added in a newer version then we where running. 

What happens is that the available migrations are queried from this code list: https://github.com/getsentry/snuba/blob/4a60b5dd19bcaa303f983a6199e957f4f3a15ea1/snuba/migrations/groups.py#L106 

It then checks if those ID's are in the migration table and checks it's status. However, this does not account for a downgrade, as the migration might not exist anymore in the code, but could still be applied in the DB. 

### Steps to Reproduce

1. Update from an older version to a newer version
2. Run schema migrations
3. Downgrade without properly undoing the migrations
4. run `snuba migrations list`

You will now not see the "newer" migrations

### Expected Result

- When running `snuba migrations list`
- Query ClickHouse for all it's Migration ID's
- Show the (non-existing in code) migration in the output of the command. 

### Actual Result

You think everything is up to date, but actually migrations are applied Snuba does not know about

davidtsuk:
Fixes #2159

Running `snuba migrations list` will now show migrations that are in clickhouse but no longer in the codebase (e.g. because snuba was downgraded).

# Testing
Deleted migration in codebase to check that it shows up in the list.
image

### DIFF
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -149,7 +149,7 @@ class Runner:
             migration_groups = get_active_migration_groups()
 
         migration_status = self._get_migration_status(migration_groups)
-        clickhouse_group_migrations = {}
+        clickhouse_group_migrations: MutableMapping[MigrationGroup, List[str]] = {}
         for group, migration_id in migration_status.keys():
             clickhouse_group_migrations.setdefault(group, []).append(migration_id)
 
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -176,7 +176,7 @@ class Runner:
 
             if include_nonexistent:
                 non_existing_migrations = clickhouse_group_migrations.get(
-                    group, []
+                    group, set()
                 ).difference(set(migration_ids))
                 for migration_id in non_existing_migrations:
                     migration_key = MigrationKey(group, migration_id)
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -64,6 +64,7 @@ class MigrationDetails(NamedTuple):
     migration_id: str
     status: Status
     blocking: bool
+    exists: bool
 
 
 class Runner:
@@ -148,6 +149,9 @@ class Runner:
             migration_groups = get_active_migration_groups()
 
         migration_status = self._get_migration_status(migration_groups)
+        clickhouse_group_migrations = {}
+        for group, migration_id in migration_status.keys():
+            clickhouse_group_migrations.setdefault(group, []).append(migration_id)
 
         def get_status(migration_key: MigrationKey) -> Status:
             return migration_status.get(migration_key, Status.NOT_STARTED)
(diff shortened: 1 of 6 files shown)


snuba/migrations/runner.py
```diff
@@ -148,6 +149,9 @@ def show_all(
             migration_groups = get_active_migration_groups()
 
         migration_status = self._get_migration_status(migration_groups)
+        clickhouse_group_migrations: MutableMapping[MigrationGroup, List[str]] = {}
+        for group, migration_id in migration_status.keys():
+            clickhouse_group_migrations.setdefault(group, []).append(migration_id)
```
onkar, line 154: nit: Would `defaultdict()` be a better choice here?
```
clickhouse_group_migrations = defaultdict(set)
for group, migration_id in migration_status.keys():
    clickhouse_group_migrations[group].add(migration_id)
```

snuba/migrations/runner.py
```diff
@@ -156,12 +160,27 @@ def get_status(migration_key: MigrationKey) -> Status:
             group_migrations: List[MigrationDetails] = []
             group_loader = get_group_loader(group)
 
-            for migration_id in group_loader.get_migrations():
+            migration_ids = group_loader.get_migrations()
```
onkar, line 163: I'd probably convert this to a `set` instead of doing it later:
```
migration_ids = set(group_leader.get_migrations())
```

onkar (commented):
Can you add unit test with covering the case of completed but non-existent migrations? 

Also, is such a migration configuration possible? `MigrationDetails("0001", Status.COMPLETED, True, False)` where migration status is completed and it's non-existent but still blocked?

snuba/migrations/runner.py
davidtsuk: The reason I didn't do this is to maintain the original order of `get_migrations()`. A set is unordered so there is no guarantee that the order will stay the same.

snuba/migrations/runner.py
davidtsuk: This is probably better yeah

davidtsuk:
> Also, is such a migration configuration possible? `MigrationDetails("0001", Status.COMPLETED, True, False)` where migration status is completed and it's non-existent but still blocked?
I'm not sure if this is possible (maybe if a migration file is deleted while the migration is still being processed).

onkar (approved):
LGTM 👍 thanks for addressing the comments.

PR #6302, opened 2024-09-13
fix: EAP timeseries order, add weight to calculations

colin-sentry:
The old tests used the same number a bunch of times, so there was a bug where the results were not ordered by time.

This makes the tests a lot more robust and implements weight in the aggregate functions

### DIFF
--- a/snuba/web/rpc/timeseries.py
+++ b/snuba/web/rpc/timeseries.py
@@ -5,17 +5,18 @@ from sentry_protos.snuba.v1alpha.endpoint_aggregate_bucket_pb2 import (
     AggregateBucketRequest,
     AggregateBucketResponse,
 )
+from sentry_protos.snuba.v1alpha.trace_item_attribute_pb2 import AttributeKey
 
 from snuba.attribution.appid import AppID
 from snuba.attribution.attribution_info import AttributionInfo
 from snuba.datasets.entities.entity_key import EntityKey
 from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.pluggable_dataset import PluggableDataset
-from snuba.query import SelectedExpression
+from snuba.query import OrderBy, OrderByDirection, SelectedExpression
 from snuba.query.data_source.simple import Entity
 from snuba.query.dsl import CurriedFunctions as cf
 from snuba.query.dsl import Functions as f
-from snuba.query.dsl import column
+from snuba.query.dsl import column, literal
 from snuba.query.expressions import Expression
 from snuba.query.logical import Query
 from snuba.query.query_settings import HTTPQuerySettings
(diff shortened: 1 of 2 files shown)


snuba/web/rpc/timeseries.py
```diff
@@ -34,19 +36,43 @@
...
+            )
+    sampling_weight_expr = column("sampling_weight_2")
+    sign_expr = column("sign")
+    sampling_weight_times_sign = f.multiply(sampling_weight_expr, sign_expr)
+
     if request.aggregate == AggregateBucketRequest.FUNCTION_SUM:
-        return f.sum(key_col, alias="sum")
-    if request.aggregate == AggregateBucketRequest.FUNCTION_AVERAGE:
-        return f.avg(key_col, alias="avg")
+        return f.sum(f.multiply(key_expr, sampling_weight_times_sign), alias="sum")
```
wmak, line 55: Does this make all the functions the weighted ones by default using `sampling_weight_2`? Isn't that going to cause some confusion since that column is empty for older data right now

snuba/web/rpc/timeseries.py
```diff
@@ -34,19 +36,43 @@
...
+        )
     if request.aggregate == AggregateBucketRequest.FUNCTION_P50:
-        return cf.quantile(0.5)(key_col, alias="p50")
+        return cf.quantileTDigestWeighted(0.5)(
+            key_expr, sampling_weight_expr, alias="p50"
+        )
     if request.aggregate == AggregateBucketRequest.FUNCTION_P95:
-        return cf.quantile(0.95)(key_col, alias="p90")
+        return cf.quantileTDigestWeighted(0.95)(
+            key_expr, sampling_weight_expr, alias="p90"
```
wmak, line 70: hmm is this a typo
```suggestion
            key_expr, sampling_weight_expr, alias="p95"
```

snuba/web/rpc/timeseries.py
```diff
@@ -34,19 +36,43 @@
...
+        )
     if request.aggregate == AggregateBucketRequest.FUNCTION_P99:
-        return cf.quantile(0.99)(key_col, alias="p95")
+        return cf.quantileTDigestWeighted(0.99)(
+            key_expr, sampling_weight_expr, alias="p95"
```
wmak, line 74: ditto here
```suggestion
            key_expr, sampling_weight_expr, alias="p99"
```

wmak (commented):
I'm nervous about enabling weight by default, but maybe i'm just missing context there
Some minor typos (i think) otherwise lgtm

wmak (approved):

phacops (approved):

phacops:
I think we'll need a way to show a summary of the extrapolation: average sampling rate, variance, total count of samples and total extrapolated count.

PR #6303, opened 2024-09-13
Revert "metric(consumer): Add a metric to track the size of individual spans"

ayirr7:
Reverts getsentry/snuba#6300


mcannizz:
Why is a revert necessary?

ayirr7:
@mcannizz Not necessary, opened just in case it affects the consumer's processing

PR #6304, opened 2024-09-13
feat(eap): implement backend for "what tag values exist" RPC

colin-sentry:
If you have spans with tags like "hello": "world" and "hello": "blah", you'd like to be able to say 'what values are possible for key="hello"?'

This endpoint implements that functionality.

### DIFF
--- a/snuba/web/rpc/trace_item_values.py
+++ b/snuba/web/rpc/trace_item_values.py
@@ -1,0 +1,111 @@
+import uuid
+from datetime import datetime
+
+from google.protobuf.json_format import MessageToDict
+from sentry_protos.snuba.v1alpha.endpoint_tags_list_pb2 import (
+    AttributeValuesRequest,
+    AttributeValuesResponse,
+)
+
+from snuba.attribution.appid import AppID
+from snuba.attribution.attribution_info import AttributionInfo
+from snuba.datasets.entities.entity_key import EntityKey
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.pluggable_dataset import PluggableDataset
+from snuba.query import OrderBy, OrderByDirection, SelectedExpression
+from snuba.query.data_source.simple import Entity
+from snuba.query.dsl import Functions as f
+from snuba.query.dsl import column, literal, literals_array
+from snuba.query.logical import Query
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.request import Request as SnubaRequest
+from snuba.utils.metrics.timer import Timer
+from snuba.web.query import run_query
+from snuba.web.rpc.common import base_conditions_and, treeify_or_and_conditions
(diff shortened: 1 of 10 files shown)


snuba/web/rpc/trace_item_values.py
```diff
@@ -0,0 +1,111 @@
...
+            f.multiSearchAny(
+                column("attr_value"),
+                literals_array(None, [literal(request.value_substring_match)]),
+            ),
+        ),
+        groupby=[column("attr_value", alias="attr_value")],
+        order_by=[
+            OrderBy(direction=OrderByDirection.ASC, expression=column("attr_value"))
+        ],
+        limit=request.limit,
```
Zylphrex, line 70: If I'm reading this query right, you might get more performance by including the organization id and attr key in the order by

snuba/web/rpc/trace_item_values.py
```diff
@@ -0,0 +1,111 @@
...
+        ],
+        condition=base_conditions_and(
+            request.meta,
+            f.equals(column("attr_key"), literal(request.name)),
+            # multiSearchAny has special treatment with ngram bloom filters
+            # https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree#functions-support
+            f.multiSearchAny(
+                column("attr_value"),
+                literals_array(None, [literal(request.value_substring_match)]),
+            ),
```
Zylphrex, line 64: Not sure what happens when `request.value_substring_match` is empty but when that happens we should just return any result. Perhaps this condition can be optional

colin-sentry: if it's empty this condition is a no-op since every string contains the empty string

snuba/web/rpc/trace_item_values.py
colin-sentry: those are both in the WHERE clause, does it actually still matter?

wmak (approved):
lgtm

wmak:
oh hm, we should follow up with Tony when he's back re: https://github.com/getsentry/snuba/pull/6304#discussion_r1759347097

phacops (approved):

phacops (approved):

snuba/web/rpc/trace_item_values.py
```diff
@@ -0,0 +1,115 @@
+import uuid
```
volokluev, line 1: file naming: `trace_item_attribute_values.py` otherwise it's misleading

tests/web/rpc/test_trace_item_values.py
```diff
@@ -0,0 +1,127 @@
+import uuid
```
volokluev, line 1: file naming here too

volokluev (approved):
approved but please rename the files

PR #6279, opened 2024-09-09
Specify UTC timezone argument in datetime.now() and fix unit tests

onkar:
While running unit tests locally using `make test`, I saw these failures:
```
=========================================================================================================== short test summary info ===========================================================================================================
FAILED tests/clickhouse/optimize/test_optimize.py::test_optimize_partitions_raises_exception_with_cutoff_time - Failed: DID NOT RAISE <class 'snuba.clickhouse.optimize.optimize_scheduler.OptimizedSchedulerTimeout'>
FAILED tests/clickhouse/optimize/test_optimize_scheduler.py::test_get_next_schedule[non parallel] - assert OptimizationSchedule(partitions_groups=[["(90,'2022-06-27')",\n                                         "(90,'2022-06-20')",\n   ...
FAILED tests/clickhouse/optimize/test_optimize_scheduler.py::test_get_next_schedule[parallel before final cutoff] - assert OptimizationSchedule(partitions_groups=[["(90,'2022-03-28')",\n                                         "(90,'202...
FAILED tests/clickhouse/optimize/test_optimize_scheduler.py::test_get_next_schedule_raises_exception - Failed: DID NOT RAISE <class 'snuba.clickhouse.optimize.optimize_scheduler.OptimizedSchedulerTimeout'>
FAILED tests/subscriptions/test_scheduler_consumer.py::test_tick_time_shift - AssertionError: assert Tick(partition=0, offsets=Interval(lower=0, upper=1), timestamps=Interval(lower=86400.0, upper=172800.0)) == Tick(partition=0, offsets=...
========================================================================== 5 failed, 2601 passed, 4 skipped, 2 deselected, 1 xfailed, 6 xpassed in 430.69s (0:07:10) ==========================================================================
```

The reason for failure is that some of these tests use [time_machine's context manager](https://github.com/adamchainz/time-machine?tab=readme-ov-file#context-manager), which expects the time in UTC. When these tests are run locally in PST, there is a difference of 7 hours that never triggers the cutoff time expiry. 
```
>>> from datetime import datetime, timezone
>>> datetime.now()
datetime.datetime(2024, 9, 9, 11, 51, 15, 279890)
>>> datetime.now(timezone.utc)
datetime.datetime(2024, 9, 9, 18, 51, 26, 40580, tzinfo=datetime.timezone.utc)
```

The fix is to use `datetime`'s `timezone.utc` instead of `datetime.now()`. `Datetime` also used to have `datetime.utcnow()` but it has been deprecated.

[...]

PR #6313, opened 2024-09-17
Cleanup some settings from settings_test.py

onkar:
During the investigation of https://github.com/getsentry/snuba/pull/6279, I found that:
- `OPTIMIZE_JOB_CUTOFF_TIME` is overridden to `24` and it was causing test failures.
- `OPTIMIZE_PARALLEL_MAX_JITTER_MINUTES` does not seem to be used anywhere in the code.

With the change, less no. of local tests fail. The failure number is non-zero because https://github.com/getsentry/snuba/pull/6279 was reverted yesterday . I would like to verify if this small change passes through the CI and stays stable in prod.

### DIFF
--- a/snuba/settings/settings_test.py
+++ b/snuba/settings/settings_test.py
@@ -39,11 +39,6 @@ REPLACER_PROCESSING_TIMEOUT_THRESHOLD = 0  # ms
 # Set enforce retention to true for tests
 ENFORCE_RETENTION = True
 
-# Ignore optimize job cut off time for tests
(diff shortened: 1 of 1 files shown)


untitaker (approved):
it's not logical to me that setting the jitter to any value other than 0 will improve test stability

onkar:
> it's not logical to me that setting the jitter to any value other than 0 will improve test stability

Thanks for the review. The test failures I was referring to was in the optimize code path, which is kind of complicated to fix.

PR #6316, opened 2024-09-17
feat(eap): rewrite the meta tables to have lower cardinality

colin-sentry:
We tried to improve performance of other queries with a subselect, but got disappointing results.
This change:
- Drops trace_id from both tables
- Removes attr_value from the number tables (we'd never want to autocomplete numbers)
- Adds the normalized columns (name, segment_name, service, duration_ms, exclusive_time_ms) to both meta tables

### DIFF
--- a/snuba/datasets/configuration/events_analytics_platform/entities/spans_num_attrs.yaml
+++ b/snuba/datasets/configuration/events_analytics_platform/entities/spans_num_attrs.yaml
@@ -7,8 +7,24 @@ schema:
     { name: organization_id, type: UInt, args: { size: 64 } },
     { name: project_id, type: UInt, args: { size: 64 } },
     { name: attr_key, type: String },
-    { name: attr_min_value, type: Float, args: { size: 64 } },
-    { name: attr_max_value, type: Float, args: { size: 64 } },
+    {
+      name: attr_min_value,
+      type: SimpleAggregateFunction,
+      args:
+        {
+          func: "min",
+          arg_types: [ { type: Float, args: { size: 64 } } ],
+        },
+    },
+    {
+      name: attr_max_value,
+      type: SimpleAggregateFunction,
+      args:
+        {
+          func: "max",
+          arg_types: [ { type: Float, args: { size: 64 } } ],
+        },
+    },
     { name: timestamp, type: DateTime },
     { name: retention_days, type: UInt, args: { size: 16 } },
     {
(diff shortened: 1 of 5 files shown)


nikhars (dismissed):
The migration as its currently written would fail via CI/CD because there is a limit on how much data a table can have before it is allowed to be dropped. The setting is this https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#max_table_size_to_drop

The setting in the ops repo is here https://github.com/getsentry/ops/blob/master/salt/states/clickhouse/files/config.xml#L16

You will need to work with a SRE to manually perform the drop table/truncate table before merging the PR.

colin-sentry:
> The migration as its currently written would fail via CI/CD because there is a limit on how much data a table can have before it is allowed to be dropped. The setting is this https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#max_table_size_to_drop
> 
> The setting in the ops repo is here https://github.com/getsentry/ops/blob/master/salt/states/clickhouse/files/config.xml#L16
> 
> You will need to work with a SRE to manually perform the drop table/truncate table before merging the PR.

There is an explicit truncate before dropping, is that not enough?

nikhars:
> > The migration as its currently written would fail via CI/CD because there is a limit on how much data a table can have before it is allowed to be dropped. The setting is this https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#max_table_size_to_drop
> > The setting in the ops repo is here https://github.com/getsentry/ops/blob/master/salt/states/clickhouse/files/config.xml#L16
> > You will need to work with a SRE to manually perform the drop table/truncate table before merging the PR.
> 
> There is an explicit truncate before dropping, is that not enough?

Nope, it won't work. From the docs linked to clickhouse docs above

> If the size of a [MergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree) table exceeds max_table_size_to_drop (in bytes), you can’t delete it using a [DROP](https://clickhouse.com/docs/en/sql-reference/statements/drop) query or [TRUNCATE](https://clickhouse.com/docs/en/sql-reference/statements/truncate) query.

colin-sentry:
Opened ops request: https://getsentry.atlassian.net/servicedesk/customer/portal/18/INF-274

nikhars:
> Opened ops request: https://getsentry.atlassian.net/servicedesk/customer/portal/18/INF-274

Thank you. Please respond on this PR when the ops request has been completed. I will remove my blocking review then.

colin-sentry:
> > Opened ops request: https://getsentry.atlassian.net/servicedesk/customer/portal/18/INF-274
> 
> Thank you. Please respond on this PR when the ops request has been completed. I will remove my blocking review then.

Is done, you can dismiss

phacops (approved):

PR #6320, opened 2024-09-18
feat(job-runner): enable snuba jobs command to run from JSON manifest

xurui-c:
https://github.com/getsentry/snuba/pull/6288

### DIFF
--- a/snuba/cli/jobs.py
+++ b/snuba/cli/jobs.py
@@ -1,8 +1,12 @@
-from typing import Tuple
+from typing import Any, MutableMapping, Tuple
 
 import click
 
+from snuba.manual_jobs import JobSpec
 from snuba.manual_jobs.job_loader import JobLoader
+from snuba.manual_jobs.manifest_reader import _ManifestReader
+
+JOB_SPECIFICATION_ERROR_MSG = "Missing job type and/or job id"
 
 
 @click.group()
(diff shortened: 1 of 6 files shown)


onewland (approved):

snuba/cli/jobs.py
```diff
@@ -11,18 +15,41 @@ def jobs() -> None:
...
 
-    kwargs = {}
+
+def _parse_params(pairs: Tuple[str, ...]) -> MutableMapping[Any, Any]:
+    params = {}
     for pair in pairs:
         k, v = pair.split("=")
-        kwargs[k] = v
+        params[k] = v
+    return params
```
onkar, line 38: nit: We can use dict comprehension here. Something like this:
`return {k: v for k, v in (pair.split("=") for pair in pairs)}`

snuba/manual_jobs/job_loader.py
```diff
@@ -1,15 +1,15 @@
...
 class JobLoader:
     @staticmethod
-    def get_job_instance(class_name: str, dry_run: bool, **kwargs: Any) -> "Job":
-        job_type_class = Job.class_from_name(class_name)
+    def get_job_instance(job_spec: JobSpec, dry_run: bool) -> "Job":
+        job_type_class = Job.class_from_name(job_spec.job_type)
         if job_type_class is None:
             raise Exception(
-                f"Job does not exist. Did you make a file {class_name}.py yet?"
+                f"Job does not exist. Did you make a file {job_spec.job_type}.py yet?"
```
onkar, line 12: Does it make sense to create a custom exception instead of the base `Exception`?

snuba/manual_jobs/toy_job.py
```diff
@@ -1,13 +1,14 @@
-from typing import Any
-
-import click
-
-from snuba.manual_jobs import Job
+from snuba.cli import logger
```
onkar, line 1: nit: Not a big deal at this point but do you think we should have a separate logger for `manual_jobs`? The [second answer](https://stackoverflow.com/questions/15727420/using-logging-in-multiple-modules) to this question might be something we want to have, or we can follow the pattern from other modules.

onkar (approved):
LGTM overall, left a few nits.

PR #6326, opened 2024-09-19
chore(eap): add a migration to namespace sentry tags

colin-sentry:
We need to be able to disambiguate things sentry sets (like name) from things user set (like `setAttribute("name")`) - this facilitates that in the materialized view

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0015_span_attribute_table_namespaced.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0015_span_attribute_table_namespaced.py
@@ -1,0 +1,142 @@
+from __future__ import annotations
+
+from typing import Sequence
+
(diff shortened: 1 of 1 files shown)


snuba/snuba_migrations/events_analytics_platform/0015_span_attribute_table_namespaced.py
```diff
@@ -0,0 +1,142 @@
...
+    project_id,
+    attrs.1,
+    attrs.2,
+    timestamp,
+    retention_days
+""",
+            ),
+        ]
+
+    def backwards_ops(self) -> Sequence[SqlOperation]:
```
phacops, line 130: Backwards ops is not exactly restoring anything. Should it just be a no-op?

phacops (approved):

PR #6331, opened 2024-09-20
(snuba-deletes): storage_deletes_enabled default enabled

MeredithAnya:
Makes the default `storage_deletes_enabled=1`. 

Sentry tests should be testing against this feature. In production all regions except S4S have this set to `0`. Additionally, only the `search_issues` dataset has deletes enabled in the storage configuration.

### DIFF
--- a/tests/test_search_issues_api.py
+++ b/tests/test_search_issues_api.py
@@ -198,7 +198,6 @@ class TestSearchIssuesSnQLApi(SimpleAPITest, BaseApiTest, ConfigurationTest):
         assert "'query' is a required property" in res.get_json()["error"]["message"]
 
         # test for invalid column types
-        set_config("storage_deletes_enabled", 1)
         res = self.app.delete(
             "/search_issues/",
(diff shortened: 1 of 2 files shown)


tests/test_search_issues_api.py
```diff
@@ -198,7 +198,6 @@ def test_bad_delete(self) -> None:
         assert "'query' is a required property" in res.get_json()["error"]["message"]
 
         # test for invalid column types
-        set_config("storage_deletes_enabled", 1)
```
armenzg, line 201: Would it make sense to remove these lines?
https://github.com/getsentry/snuba/blob/6fe79f0997f435c08bc4d0170ebf3ea3763c3ccd/tests/test_search_issues_api.py#L158-L164

MeredithAnya: I had left them in just to test that the feature flagging works

armenzg (approved):

onkar (approved):

PR #6332, opened 2024-09-20
feat(job-runner): track job status

xurui-c:
Track job status to prevent double executions

Making a few changes here:
1. Make `JobSpec` the main way to call run_job and run_job the main interface for executing manual jobs
2. Introduce a new Redis cluster to store job status
3. for `run_job`, acquire a lock via a redis key or abort the attempted execution
4. for `run_job`, also abort execution if the job_id contains a non-null/"not started" status
5. new command to fetch job status given a job ID

JIRA: https://getsentry.atlassian.net/browse/SNS-2869


tests/jobs/test_job.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+from snuba.manual_jobs import FINISHED, NOT_STARTED, JobSpec, get_job_status
+from snuba.manual_jobs.toy_job import ToyJob
+
+JOB_ID = "abc1234"
+test_job_spec = JobSpec(job_id=JOB_ID, job_type="ToyJob")
+
+# @pytest.fixture(autouse=True)
```
xurui-c, line 9: it turns out the state of the redis client does not persist outside, so after inserting ['abc1234', 'not started'] during init of ToyJob, redis_client.hkeys would show nothing inserted in `test_job_status_changes_to_finished`

snuba/manual_jobs/__init__.py
```diff
@@ -23,9 +31,15 @@ def __init__(self, job_spec: JobSpec, dry_run: bool) -> None:
         if job_spec.params:
             for k, v in job_spec.params.items():
                 setattr(self, k, v)
+        self._set_status(NOT_STARTED)
 
-    @abstractmethod
     def execute(self) -> None:
+        self._set_status(RUNNING)
+        self._execute()
+        self._set_status(FINISHED)
```
onewland, line 39: I'm thinking that we might want status tracking to live outside `Job`.I think of Job as more of an interface for users to implement, rather than something that contains logic for running jobs itself (I also think things can get confusing quickly when you rely on inheritance in Python). 

What if we had something like a JobRunner static class or module that exposed utilities that would orchestrate the job run, and manage status based on the completely separate Job class? I think that produces a looser coupling.

snuba/manual_jobs/runner.py
```diff
@@ -34,7 +63,69 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
...
+    have_lock = _acquire_job_lock(job_spec.job_id)
+    if not have_lock:
+        raise JobLockedException(job_spec.job_id)
+
+    current_job_status = _set_job_status(job_spec.job_id, JobStatus.NOT_STARTED)
+
+    job_to_run = _JobLoader.get_job_instance(job_spec, dry_run)
+
+    try:
+        if not dry_run:
```
onewland, line 119: I'm starting to feel like `dry_run` may not be worth the effort to support. I think it won't be relevant in a lot of situations and it generates a fair amount of branching

xurui-c: sgtm

snuba/manual_jobs/runner.py
```diff
@@ -34,7 +63,69 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
...
-    return _read_from_path(jobs_filename)
+    return _read_manifest_from_path(manifest_filename)
+
+
+def run_job(job_spec: JobSpec, dry_run: bool) -> JobStatus:
+    current_job_status = get_job_status(job_spec.job_id)
+    if current_job_status is not None and current_job_status != JobStatus.NOT_STARTED:
+        raise JobStatusException(job_id=job_spec.job_id, status=current_job_status)
+
+    have_lock = _acquire_job_lock(job_spec.job_id)
```
onkar, line 110: I feel like this could be written in a much cleaner way using a context manager `RedisContext()` that allows us to do things like:
```
with RedisContext() as redis_context:
    # do job things..
```
Currently, it is too verbose and branching could be avoided. With context manager, that could be abstracted cleanly.

snuba/manual_jobs/runner.py
```diff
@@ -34,7 +63,69 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
...
+    _redis_client.set(name=_job_status_key(job_id), value=status.value)
+    return status
+
+
+def get_job_status(job_id: str) -> Optional[JobStatus]:
+    redis_status = _redis_client.get(name=_job_status_key(job_id))
+    if redis_status is None:
+        return redis_status
+    else:
+        return JobStatus(redis_status.decode("utf-8"))
```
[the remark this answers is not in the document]
onewland: redis_status returns a byte array, not a string. For comparison to the enum values, we need a string

snuba/manual_jobs/runner.py
```diff
@@ -34,7 +63,69 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
...
+def _acquire_job_lock(job_id: str) -> bool:
+    return bool(
+        _redis_client.set(
+            name=_job_lock_key(job_id), value=1, nx=True, ex=(24 * 60 * 60)
+        )
+    )
+
+
+def _release_job_lock(job_id: str) -> None:
+    _redis_client.delete(_job_lock_key(job_id))
```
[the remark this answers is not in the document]
onewland: I think that's overkill for the level of synchronization we're looking for here. We're basically just looking to prevent mistakes of e.g. a single engineer clicking twice because they think the first click did nothing. We don't have to worry about 1000 users trying to run the same job concurrently.

snuba/manual_jobs/runner.py
```diff
@@ -34,7 +63,69 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
...
+
+
+def _set_job_status(job_id: str, status: JobStatus) -> JobStatus:
+    _redis_client.set(name=_job_status_key(job_id), value=status.value)
+    return status
```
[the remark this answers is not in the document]
onewland: it's just to make the chaining on lines 114, 120, and 126 eaier. e.g.  instead of

```
current_job_status = JobStatus.FINISHED
_set_job_status(job_spec.job_id, JobStatus.FINISHED)
```

you have

```
```

if you prefer the verbose way, I can change it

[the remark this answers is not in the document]
onewland: I don't think there's much value to returning the result of set(). If anything, I'd probably throw an exception if set fails

snuba/manual_jobs/runner.py
```diff
@@ -34,7 +63,69 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
     return job_spec
 
 
+def _job_lock_key(job_id: str) -> str:
+    return f"snuba:manual_jobs:{job_id}:lock"
+
+
+def _job_status_key(job_id: str) -> str:
```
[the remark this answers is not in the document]
onewland: I don't think the verb "get" is explanatory. If you wanted to do `_construct_job_status_key()` or `_build_job_status_key()` I'd be fine with that.

A meta-comment, though: I'm not sure how much we should discuss the names of private (leading underscore) methods. They shouldn't ever be exported outside the namespace and by nature they're subject to change and go away.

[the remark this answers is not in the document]
onkar: > I don't think the verb "get" is explanatory. If you wanted to do _construct_job_status_key() or _build_job_status_key() I'd be fine with that.

I am okay with this naming.

snuba/manual_jobs/runner.py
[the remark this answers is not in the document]
onkar: Yes, the failure case made me question the return value. Probably better to throw the exception if `set()` fails, otherwise return the `status`.

snuba/manual_jobs/runner.py
onewland: Can you explain how branching would be avoided? Is there something you expect redis_context to give access to that will make `_run_job` cleaner? If all it handles is lock/release lock I'm not seeing the value of another layer of abstraction

snuba/manual_jobs/runner.py
onkar: Are we going to do the `dry_run` cleanup in this PR or in a follow-up PR? I don't think it's very useful, tbh. Maybe we can add it later if folks really need it. Thoughts?

onewland: I'd rather isolate it in a follow-up PR, since removing dry_run will touch a few things not related to status tracking

PR #6334, opened 2024-09-23
feat: do not clobber existing overrides with new overrides

colin-sentry:
We want to optimize things in EAP by setting clickhouse overrides per-endpoint, this facilitates that

### DIFF
--- a/tests/query/processors/test_clickhouse_settings_override.py
+++ b/tests/query/processors/test_clickhouse_settings_override.py
@@ -52,7 +52,11 @@ def test_apply_clickhouse_settings(
             )
         ],
     )
+
     settings = HTTPQuerySettings()
+    settings.set_clickhouse_settings({"use_query_cache": "true"})
 
     ClickhouseSettingsOverride(clickhouse_settings).process_query(query, settings)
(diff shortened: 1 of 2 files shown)


tests/query/processors/test_clickhouse_settings_override.py
```diff
@@ -52,7 +52,11 @@ def test_apply_clickhouse_settings(
...
     )
+
     settings = HTTPQuerySettings()
+    settings.set_clickhouse_settings({"use_query_cache": "true"})
 
     ClickhouseSettingsOverride(clickhouse_settings).process_query(query, settings)
-    assert settings.get_clickhouse_settings() == clickhouse_settings
+    expected = dict(clickhouse_settings)
+    expected["use_query_cache"] = "true"
+    assert settings.get_clickhouse_settings() == expected
```
volokluev, line 62: The way you have modified this test is confusing to read through. Please write another test case to test the specific  behavior you are modifying.

colin-sentry: done

evanh (approved):

PR #6335, opened 2024-09-23
ref(snuba-jobs): refactor manifest loading a bit

onewland:
A class didn't do much here, and I wanted to remove a dependency on a private thing. I also think `snuba.manual_jobs.runner.list_available` makes more sense than `snuba.manual_jobs.manifest_reader.read_jobs_manifest`.


snuba/manual_jobs/runner.py
```diff
@@ -0,0 +1,40 @@
...
+        contents = simplejson.loads(stream.read())
+
+        job_specs = {}
+        for content in contents:
+            job_spec = _job_spec_from_entry(content)
+            job_specs[job_spec.job_id] = job_spec
+        return job_specs
+
+
+def _job_spec_from_entry(content: Any) -> JobSpec:
```
xurui-c, line 22: can we do _get_job_spec

xurui-c (approved):

snuba/manual_jobs/runner.py
onewland: how about `_extract` or `_build`? get specifically just seems kind of vague to me

xurui-c: youre right, build is bettere than get

PR #6336, opened 2024-09-23
feat(observability): record failed queries in querylog

volokluev:
We've been having trouble debugging issues that failed on parsing. Record them in the querylog so they can be examined


onewland:
do we have a sense that this is fairly rare? I'm fine with this as a quick way to get visibility but it seems like this could screw up calculations based on counts if you don't know to filter out queries that errored out.

evanh (approved):

volokluev:
> do we have a sense that this is fairly rare? I'm fine with this as a quick way to get visibility but it seems like this could screw up calculations based on counts if the user doesn't know to filter out queries that errored out.

@onewland which calculations are you concerned about?

PR #6337, opened 2024-09-24
feat(job-runner): connect admin to snuba manual jobs

onewland:
For now, this view is read-only

In this changeset, we:
- create a new admin tool setting for viewing jobs
- add a new nav section for viewing jobs
- create a new page for viewing jobs, that loads the job specs from disk

Screenshot:
Screenshot 2024-09-24 at 7 38 38 AM


snuba/admin/static/api_client.tsx
```diff
@@ -106,9 +106,10 @@ interface Client {
     storage_name: string,
     column_conditions: object
   ) => Promise<Response>;
+  listJobSpecs: () => Promise<JobSpecMap>;
 }
 
-function Client() {
+function Client(): Client {
```
onewland, line 112: this enforces that we define every method in the interface, which seems good to me

snuba/admin/static/index.tsx
```diff
@@ -34,7 +34,7 @@ client.getSettings().then((settings) => {
       tracesSampleRate: settings.tracesSampleRate,
       // Profiles
       profilesSampleRate: settings.profilesSampleRate,
-      tracePropagationTargets: settings.tracePropagationTargets,
+      tracePropagationTargets: settings.tracePropagationTargets ?? undefined,
```
onewland, line 37: not sure why this wasn't broken before, but the typing was definitely wrong

snuba/admin/static/data.tsx
```diff
@@ -15,6 +15,7 @@ import MQLQueries from "SnubaAdmin/mql_queries";
 import SnubaExplain from "SnubaAdmin/snuba_explain";
 import Welcome from "SnubaAdmin/welcome";
 import DeleteTool from "SnubaAdmin/delete_tool";
+import ViewCustomJobs from "./manual_jobs";
```
onkar, line 18: Can we follow the same import pattern as other imports? i.e. from `SnubaAdmin` dir?

onkar (approved):
LGTM

PR #6341, opened 2024-09-24
Surface storage key for capacity management

xurui-c:
Upon getting a request to increase the thresholds for a customer, one of the things we need to figure out is which storage. This PR surfaces the `StorageKey` to allow easier querying using SnubaAdmin's QueryLog

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -756,7 +756,7 @@ def _add_quota_info(
         quota_info["quota_used"] = quota_allowance.quota_used
         quota_info["quota_unit"] = quota_allowance.quota_unit
         quota_info["suggestion"] = quota_allowance.suggestion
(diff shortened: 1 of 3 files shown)


volokluev:
The storage key is already in the explanation of every quota allowance:

https://github.com/getsentry/snuba/blob/master/snuba/query/allocation_policies/__init__.py#L816

What does this PR give you on top of that?

xurui-c:
Because I want to have it in the summary field to avoid having to look at each policy of a rejected/throttled query to see which storage to navigate to on Snuba Admin. Also, I want to `SELECT JSON_QUERY(arrayJoin(clickhouse_queries.stats), '$.quota_allowance.summary.rejection_storage_key')`

volokluev (approved):

PR #6344, opened 2024-09-25
ref(eap): Clean up the mutations interface

untitaker:
* span_id should be a string
* _sort_timestamp is no longer directly exposed, instead users provide
  start_timestamp_ms and it gets converted using the same logic used in
  the ingestion consumer

### DIFF
--- a/rust_snuba/src/mutations/parser.rs
+++ b/rust_snuba/src/mutations/parser.rs
@@ -13,7 +13,7 @@ use sentry::SentryFutureExt;
 use serde::Deserialize;
 
 use crate::arroyo_utils::invalid_message_err;
-use crate::processors::eap_spans::PrimaryKey;
+use crate::processors::eap_spans::{FromPrimaryKey, PrimaryKey};
 
 #[derive(Debug, Default, Deserialize, JsonSchema)]
 pub(crate) struct Update {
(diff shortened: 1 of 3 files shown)


ayirr7:
Thanks for cleaning this up

ayirr7 (approved):

onkar (approved):

ayirr7:
We need to bump the sentry-kafka-schemas in this too, right?

evanh (approved):

untitaker:
there is some discussion on https://github.com/getsentry/sentry-kafka-schemas/pull/335 that raises doubts about whether it's better to expose _sort_timestamp or start_timestamp_ms

IMO we will probably end up with two kinds of mutations in practice:

1. the ones that react to data on the ingestion topic, or mutations that come from relay or the SDK itself (might not make a difference to the mutations platform, not sure yet)
2. the ones that query clickhouse periodically and emit mutations based on that. In this case `_sort_timestamp` can be exposed as an opaque attribute that is forwarded from query results to the mutations interface.

For the first group: `_sort_timestamp` doesn't exist on the ingestion topic. So either they have to perform the computation for _sort_timestamp themselves, or they have to forward their start_timestamp_ms with the assumption baked in that this is somehow indirectly part of the primary key.

The other option is that we remove the attribute entirely, and find some other way to fetch this attribute based on `(organization_id, trace_id, span_id)` from within the mutations consumer.

phacops:
In either of those cases would I expose `_sort_timestamp`. We need to expose and make people use `start_timestamp` and let `_sort_timestamp` be the internal timestamp the platform use but not users.

PR #6346, opened 2024-09-25
test(eap): Testing framework for running mutations on Clickhouse

ayirr7:
A follow-up to #6216 

More rigorous testing for the main query run on the mutations consumer. We need to be able to catch any correctness issues with the query in CI itself.


rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -170,6 +170,76 @@ fn format_query(table: &str, batch: &MutationBatch) -> Vec<Vec<u8>> {
...
 
+struct ClickhouseTestClient {
+    url: String,
+    table: String,
+    client: Client,
+}
+
+impl ClickhouseTestClient {
+    pub async fn new(table: String) -> anyhow::Result<Self> {
+        // hardcoding local Clickhouse settings
```
untitaker, line 181: i think we want to make those environment variables with these defaults. however, i think the 9000 port is also wrong, you want 8123

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -170,6 +170,76 @@ fn format_query(table: &str, batch: &MutationBatch) -> Vec<Vec<u8>> {
...
+        let hostname = "127.0.0.1";
+        let http_port = 9000;
+        let url = format!("http://{hostname}:{http_port}");
+
+        let client = reqwest::Client::new();
+        let test_table = format!("{table}_test_local\n");
```
untitaker, line 187: this should contain a UUID or otherwise random value. we will have many tests that use this client, and `cargo test` runs them in parallel. if the tests were run sequentially, it would not be necessary to create a separate table per test at all

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -170,6 +170,76 @@ fn format_query(table: &str, batch: &MutationBatch) -> Vec<Vec<u8>> {
...
+
+        let insert =
+        format!("INSERT INTO {table} (organization_id, _sort_timestamp, trace_id, span_id, sign) VALUES ({organization_id}, {_sort_timestamp}, \'{trace_id}\', {span_id}, 1)\n").into_bytes();
+
+        self.client.post(&self.url).body(insert).send().await?;
+
+        Ok(())
+    }
+
+    pub async fn drop_table(&self) -> anyhow::Result<()> {
```
untitaker, line 233: do `impl Drop for ClickhouseTestClient`, and inside `fn drop()`, run `tokio::block_on(self.drop_table())`. then this function does not need to be called explicitly from the test, the table will just disappear when the test is over

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -170,6 +170,76 @@ fn format_query(table: &str, batch: &MutationBatch) -> Vec<Vec<u8>> {
...
+                .body(q)
+                .send()
+                .await?
+                .error_for_status()?;
+        }
+
+        Ok(())
+    }
+
+    pub async fn insert_data(&self, primary_key: PrimaryKey) -> anyhow::Result<()> {
```
untitaker, line 217: I don't think this abstraction is necessary, I believe it's easier (or at least more flexible) to write the entire query yourself in the test, and invoke `run_query`

rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -170,6 +170,76 @@ fn format_query(table: &str, batch: &MutationBatch) -> Vec<Vec<u8>> {
...
+            table: test_table,
+            client: client,
+        })
+    }
+
+    // actually run a query in the table
+    pub async fn run_query(&self, queries: Vec<Vec<u8>>) -> anyhow::Result<()> {
+        // run the vec of queries against Clickhouse
+        for q in queries {
+            self.client
```
untitaker, line 206: this needs to have session_id as an attribute, otherwise the temporary table will disappear

rust_snuba/src/mutations/clickhouse.rs
ayirr7: oh yeah i was testing this out and forgot to revert. 8123 is correct

rust_snuba/src/mutations/clickhouse.rs
ayirr7: Oh, I didn't realize that `cargo test` would run them in parallel. Thanks for calling that out

ayirr7:
I've addressed the comments locally. However I think I'm still running into some issues with the core functionality of the test (I described them offline)

rust_snuba/src/mutations/clickhouse.rs
ayirr7: Done

rust_snuba/src/mutations/clickhouse.rs
ayirr7: Removed

rust_snuba/src/mutations/clickhouse.rs
ayirr7: I tried this and for some reason it's causing the test to time out (hang) altogether, i.e:

`test mutations::clickhouse::tests::test_simple_mutation has been running for over 60 seconds`

I also tried a different variation and ran into `Cannot start a runtime from within a runtime.` I can dig deeper into this to understand what would work, but it seems a manual drop could also do? Let me know if you find another approach that works -- for supporting an asynchronous drop function

untitaker: Ok let's do it manually for now, we can fix it later

untitaker (approved):

untitaker:
1. moved ClickhouseTestClient into the `mod tests{}`, then you don't have to use `allow(dead_code)` as it is only compiled when the tests are being compiled (`cfg(test)`)
2. removed some comments
3. fixed the CI issue -- in CI, the clickhouse is available on `clickhouse`, not `localhost`: https://github.com/getsentry/snuba/blob/bc31d9544380581d77309087a0e000b2e60408d4/docker-compose.gcb.yml#L20

so i added a check for that environment variable

the mutations change looks good. i think the PR is ready to merge

rust_snuba/src/mutations/clickhouse.rs
ayirr7: I agree, I'd like to fix it so that I don't have to manually call drop. I'll keep looking into this on the side and if I find a solution that works I'll fix it.

PR #6347, opened 2024-09-25
fix(allocation-policies): delete policies to admin re-do

MeredithAnya:
The UI was supposed to get added in https://github.com/getsentry/snuba/pull/6248, not sure what happened here. 

Also had to fix the endpoint too - I think I did a bad merge conflict resolution or something

Screenshot 2024-09-26 at 12 05 44 PM

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -939,14 +939,6 @@ def get_allocation_policy_configs(storage_key: str) -> Response:
     add_policy_data(policies, "select")
     add_policy_data(delete_policies, "delete")
 
-    data = [
-        {
-            "policy_name": policy.config_key(),
-            "configs": policy.get_current_configs(),
-            "optional_config_definitions": policy.get_optional_config_definitions_json(),
-        }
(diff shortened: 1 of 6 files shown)


snuba/admin/views.py
kylemumma, line 1: im confused what this deletion is for?

MeredithAnya: this was a result of a bad merge conflict - it overwrites the logic that adds the `query_type`. it was supposed to be deleted before

onkar:
Can you attach the screenshot of how this change looks like in the UI? Just want to make sure it does the right thing in like in https://github.com/getsentry/snuba/pull/6248.

MeredithAnya:
> Can you attach the screenshot of how this change looks like in the UI? Just want to make sure it does the right thing like in #6248.

Deletes look like (there wont be a header if no delete allocation policies exist): 

Screenshot 2024-09-26 at 12 06 44 PM

kylemumma (approved):

PR #6351, opened 2024-09-26
init

kylemumma:


snuba/web/delete_query.py
```diff
@@ -38,11 +39,16 @@ class DeletesNotEnabledError(Exception):
...
+class TooManyOngoingMutationsError(Exception):
+    pass
+
+
 @with_span()
 def delete_from_storage(
     storage: WritableTableStorage,
     columns: Dict[str, list[Any]],
     attribution_info: Mapping[str, Any],
+    max_ongoing_mutations: int = 5,
```
volokluev, line 51: this number should be a runtime config option

snuba/web/delete_query.py
```diff
@@ -73,6 +84,33 @@ def delete_from_storage(
...
+    else:
+        query = f"""
+SELECT max(cnt)
+FROM (
+    SELECT hostname() as host, count() as cnt
+    FROM clusterAllReplicas('{cluster.get_clickhouse_cluster_name()}', 'system', mutations)
+    WHERE table IN ({", ".join(map(repr, delete_settings.tables))}) AND is_done=0
+    GROUP BY host
+    )
+"""
```
volokluev, line 105: Make this piece that determines the amount of mutations its own function. Mock it in a test. then you can test this functionality

snuba/web/delete_query.py
MeredithAnya: but can still use 5 as the default for the config

snuba/web/delete_query.py
```diff
@@ -73,6 +84,33 @@ def delete_from_storage(
...
+    num_ongoing_mutations = (
+        cluster.get_query_connection(ClickhouseClientSettings.QUERY)
+        .execute(query)
+        .results[0][0]
+    )
+    if num_ongoing_mutations > max_ongoing_mutations:
+        raise TooManyOngoingMutationsError()
```
onkar, line 112: Should this exception be raised with a message saying how many mutations are ongoing? Possibly log the exception with relevant details like how many concurrent mutations are allowed etc.?

snuba/web/delete_query.py
```diff
@@ -73,6 +84,33 @@ def delete_from_storage(
...
+    FROM system.mutations
+    WHERE table IN ({", ".join(map(repr, delete_settings.tables))}) AND is_done=0
+"""
+    else:
+        query = f"""
+SELECT max(cnt)
+FROM (
+    SELECT hostname() as host, count() as cnt
+    FROM clusterAllReplicas('{cluster.get_clickhouse_cluster_name()}', 'system', mutations)
+    WHERE table IN ({", ".join(map(repr, delete_settings.tables))}) AND is_done=0
```
MeredithAnya, line 102: This will combine the counts for the tables right? Can we group by table and check the max of the results? So the max of the max for the cluster query, and it would be the max of the single node

MeredithAnya: but if there are two tables for a node it will combine them

```
SELECT count(), table
FROM system.mutations
WHERE table in ('outcomes_raw_local', 'outcomes_hourly_local')
GROUP BY table
```
Screenshot 2024-09-27 at 2 29 02 PM

If you don't group by table, the total count is 19

This is for one node. This won't be a problem for the search issues table, but I'm just saying if we ever need to delete from multiple tables it will combine the counts

kylemumma:
* I made MAX_ONGOING_MUTATIONS_FOR_DELETE a runtime config option
* I added how many mutation are ongoing and the max allowed to the error message
* I added a test (it doesnt rly test the sql but it tests the other logic. I wouldnt know how to test the sql bc it would involve fake data in clickhouse)

snuba/web/delete_query.py
kylemumma: Oh I see what you mean now, fixed.

MeredithAnya (approved):

PR #6352, opened 2024-09-26
feat: add a way of running EAP queries async with a thread pool

colin-sentry:
This is a performance optimization that removes most of the overhead of running multiple concurrent clickhouse queries.

Benchmarked against `SNUBA_SETTINGS=test pytest -vv tests/web/rpc/test_timeseries_api.py::TestTimeSeriesApi::test_query_caching`

before:
`took 688ms to run timeseries_query`

after:
`took 102ms to run timeseries_query`

This is a request that doesn't hit query cache at all and does ~25 requests.

### DIFF
--- a/snuba/web/rpc/common/eap_execute.py
+++ b/snuba/web/rpc/common/eap_execute.py
@@ -1,0 +1,84 @@
+import uuid
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import Any, Literal, MutableMapping, Optional
+
+from snuba import settings
+from snuba.attribution import AppID
+from snuba.attribution.attribution_info import AttributionInfo
+from snuba.datasets.entities.entity_key import EntityKey
+from snuba.datasets.entities.factory import get_entity
+from snuba.datasets.pluggable_dataset import PluggableDataset
+from snuba.query.data_source.simple import Entity
+from snuba.query.logical import Query
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.request import Request
+from snuba.utils.metrics.timer import Timer
+from snuba.web import QueryResult
+from snuba.web.query import run_query
+
+eap_executor = ThreadPoolExecutor(max_workers=settings.CLICKHOUSE_MAX_POOL_SIZE)
+
+
+def run_eap_query(
+    dataset: Literal["eap_spans", "spans_str_attrs", "spans_num_attrs"],
+    query: Query,
+    original_body: dict[str, Any],
+    referrer: str,
+    organization_id: int,
+    parent_api: str,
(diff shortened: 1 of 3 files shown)


snuba/web/rpc/common/eap_transform.py
```diff
@@ -0,0 +1,65 @@
...
+
+    if exp.column_name == "timestamp":
+        return Column(
+            table_name=exp.table_name, column_name="_sort_timestamp", alias=exp.alias
+        )
+
+    return exp
+
+
+def replace_selected_column(
```
colin-sentry, line 27: @volokluev is there a way of doing this more nicely?

snuba/web/rpc/common/eap_execute.py
```diff
@@ -0,0 +1,51 @@
...
+    query: Query,
+    clickhouse_settings: Optional[Mapping[str, Any]] = None,
+    storage_name: str = "eap_spans",
+) -> ClickhouseQueryResult:
+    storage_key = StorageKey(storage_name)
+    storage = get_storage(storage_key)
+    data_source = storage.get_schema().get_data_source()
+    assert isinstance(data_source, TableSource)
+    cluster = storage.get_cluster()
+    clickhouse_query = identity_translate(query)
```
colin-sentry, line 25: @volokluev not sure if this is the best translate, it seems to be used elsewhere in prod but it says it's only for tests?

snuba/web/rpc/common/eap_execute.py
```diff
@@ -0,0 +1,51 @@
...
+    clickhouse_query.set_from_clause(
+        Table(
+            table_name=data_source.get_table_name(),
+            schema=data_source.get_columns(),
+            storage_key=storage_key,
+        )
+    )
+    formatted_query = format_query(clickhouse_query)
+    reader = cluster.get_reader()
+    return reader.execute(formatted_query, clickhouse_settings, robust=True)
```
volokluev, line 35: this should be wrapped in the allocation policies. You can see an example of it here:

https://github.com/getsentry/snuba/blob/master/snuba/web/db_query.py#L600

volokluev (changes requested):
I'm not approving this without capacity management

snuba/web/rpc/common/eap_transform.py
volokluev: what is this for?

snuba/web/rpc/common/eap_execute.py
volokluev: this is fine, feel free to remove that comment too

volokluev (approved):

PR #6360, opened 2024-09-30
fix: Broken tests in CI

ayirr7:
The test appears to be failing because it is now the end of month and the tests rely on things like (TODAY + 1). Changes base time so that we never exceed the number of days in the month.

### DIFF
--- a/tests/web/rpc/v1alpha/test_trace_item_attribute_values.py
+++ b/tests/web/rpc/v1alpha/test_trace_item_attribute_values.py
@@ -16,7 +16,7 @@ from tests.base import BaseApiTest
 from tests.helpers import write_raw_unprocessed_events
 
 BASE_TIME = datetime.utcnow().replace(minute=0, second=0, microsecond=0) - timedelta(
-    minutes=180
+    days=2, minutes=180
 )
 COMMON_META = RequestMeta(
     project_ids=[1, 2, 3],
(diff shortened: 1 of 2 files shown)


tests/web/rpc/v1alpha/test_trace_item_attribute_values.py
```diff
@@ -16,7 +16,7 @@
 from tests.helpers import write_raw_unprocessed_events
 
 BASE_TIME = datetime.utcnow().replace(minute=0, second=0, microsecond=0) - timedelta(
-    minutes=180
+    days=2, minutes=180
```
colin-sentry, line 19: I think this will just push the problem to the 2nd, where BASE_TIME will be the 30th again, and then base_time.days + 1 will be the 31st

ayirr7: thanks!

colin-sentry (approved):

onkar:
`datetime.utcnow()` is deprecated, perhaps replace that as well with something like `datetime.now(timezone.utc)`?

onkar (approved):
LGTM, thanks for addressing the feedback.

enochtangg (approved):

colin-sentry (approved):

PR #6365, opened 2024-09-30
ref(ci): test_run_optimize_with_ongoing_merges as xfail

MeredithAnya:
This is constantly failing around 5pm PST or later and blocks PRs and deploys. We should get a better long term fix, but this is to unblock deploys.


ayirr7 (approved):

ayirr7:
I noticed another one that's failing on my PR: `test_get_next_schedule`

PR #6368, opened 2024-10-01
Revert "Revert "fix(gocd): Add snuba-eap-mutations consumer to US""

untitaker:
Reverts getsentry/snuba#6364

In order to roll out #6366 , we need to... deploy the eap consumer

i wish we could just exclude it from our deployment health checks instead, will check that next


ayirr7:
> just exclude it from our deployment health checks instead

+1 to this

PR #6371, opened 2024-10-02
Querylog should record the correct `bytes_scanned`

xurui-c:
Following up on https://github.com/getsentry/snuba/pull/4714, let's make that change for the Querlog as well.

Currently, [this Snuba producer](https://github.com/getsentry/snuba/blob/ffcfdb9241960e879af9904b06f57ba3387be8f2/snuba/state/__init__.py#L352) produces a bunch of information about queries to the Querylog topic, and the Querylog consumer is responsible for deserializing all that information. We should be deserializing the `progress_bytes` instead of `bytes`.


rust_snuba/src/processors/mod.rs
```diff
@@ -159,7 +159,8 @@ mod tests {
                 settings.set_description(std::str::from_utf8(example.payload()).unwrap());
                 let _guard = settings.bind_to_scope();
 
-                let payload = KafkaPayload::new(None, None, Some(example.payload().to_vec()));
+                let payload: KafkaPayload =
```
xurui-c, line 162: this just got reformatted

evanh (approved):

rust_snuba/src/processors/querylog.rs
```diff
@@ -225,6 +226,7 @@ impl TryFrom<Vec<FromQuery>> for QueryList {
             array_join_columns.push(q.profile.array_join_cols);
             let result_profile = q.result_profile.unwrap_or_default();
             bytes_scanned.push(result_profile.bytes);
```
nachivrn, line 228: profile bytes can we removed. We need to store progress bytes in query log.

rust_snuba/src/processors/querylog.rs
```diff
@@ -597,6 +601,7 @@ mod tests {
                 },
                 "result_profile": {
                   "bytes": 1305,
+                  "progress_bytes": 0,
```
nachivrn, line 604: NIT: Payload can be updated to have a non zero positive integer.

PR #6375, opened 2024-10-02
ref(eap) Remove old attributes tables

evanh:
The original meta attribute tables were replaced with slightly different versions.
Remove the old tables.

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0016_remove_old_span_attribute_table.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0016_remove_old_span_attribute_table.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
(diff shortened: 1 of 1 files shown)


colin-sentry:
isn't this in migration 5 already?

snuba/snuba_migrations/events_analytics_platform/0016_remove_old_span_attribute_table.py
```diff
@@ -0,0 +1,72 @@
...
+                target=OperationTarget.LOCAL,
+            ),
+            operations.DropTable(
+                storage_set=self.storage_set_key,
+                table_name=self.meta_dist_table_name,
+                target=OperationTarget.DISTRIBUTED,
+            ),
+        ]
+
+    def backwards_ops(self) -> Sequence[SqlOperation]:
```
volokluev, line 42: we probably shouldn't have a backwards ops. We don't want to try and recreate this table if the migration fails

colin-sentry (approved):

PR #6378, opened 2024-10-02
feat(replay): add arrayJoin translation_mappers for tags.key and value

aliu39:
Needed for https://github.com/getsentry/sentry/issues/78531. Sentry tagstore makes this query:

```
MATCH (replays SAMPLE 1.0) SELECT count() AS `count` BY
tags_key WHERE timestamp >= toDateTime('2024-08-12T00:00:00') AND
timestamp < toDateTime('2024-08-13T23:59:59') AND project_id IN
tuple(1) AND project_id IN tuple(1) ORDER BY count DESC LIMIT 1000
```

And right now we're getting http response `'Validation failed for entity replays: Tag keys (tags_key) not resolved'`.

Uses [entities/search_issues.yaml](https://github.com/getsentry/snuba/blob/2544b8e87d9693bc8ffe9feadf5b6809e1cf71d8/snuba/datasets/configuration/issues/entities/search_issues.yaml#L1-L1) and [storages/search_issues.yaml](https://github.com/getsentry/snuba/blob/2544b8e87d9693bc8ffe9feadf5b6809e1cf71d8/snuba/datasets/configuration/issues/storages/search_issues.yaml#L1-L1) as references. I don't think we _need_ the additions to storage `query_processors`, but they may come in handy.

### DIFF
--- a/snuba/datasets/configuration/replays/entities/replays.yaml
+++ b/snuba/datasets/configuration/replays/entities/replays.yaml
@@ -147,6 +147,20 @@ required_time_column: timestamp
 storages:
   - storage: replays
     is_writable: true
+    translation_mappers:
+      columns:
+        - mapper: ColumnToFunctionOnColumn
+          args:
+            from_table_name: null
+            from_col_name: tags_key
+            to_function_name: arrayJoin
+            to_function_column: tags.key
+        - mapper: ColumnToFunctionOnColumn
+          args:
+            from_table_name: null
+            from_col_name: tags_value
+            to_function_name: arrayJoin
+            to_function_column: tags.value
 
 storage_selector:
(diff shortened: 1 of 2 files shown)


cmanallen:
@aliu39 Have you tested these changes locally?

aliu39:
> @aliu39 Have you tested these changes locally?
> 
> Edit: if you run this branch locally you can use it when running tests in the test suite. Create a new Sentry branch taking advantage of these changes.
@cmanallen 
Yes, I ran the test in https://github.com/getsentry/sentry/pull/78534 and it works! Also ran the /replays/ test suite as a regression test.

cmanallen (approved):

PR #6379, opened 2024-10-03
fix: Fix SyntaxWarning in Python 3.12

evanh:
When an escape character is used in a string, Python 3.12 will now show a SyntaxWarning.

See here: https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes

This was causing mypy to fail to run. Fix this bug across the codebase.

Also exclude the rust_snuba files (which suffer from this problem) since they are autogenerated.


volokluev (approved):
thanks evan

PR #6380, opened 2024-10-03
cleanup(job-runner): remove dry_run

onewland:
Removing the dry_run option as we don't see it providing a ton of value, but it creates a lot of branching that makes the code less readable


onkar (approved):
Looks cleaner!

PR #6381, opened 2024-10-03
chore(admin): Allow product tools role to use querylog tool

xurui-c:
Because otherwise an SnS engineer will have to manually run it every time we have a capman issue


davidtsuk (approved):

kylemumma (approved):

onewland:
do you need to attach this to the route in views.py?

PR #6383, opened 2024-10-03
fix(eap): Fix syntax error in migration

phacops:
This syntax error prevents us from running migrations from scratch.


phacops:
Seems like the tests started to pass on the other PR, this is not needed anymore.

PR #6386, opened 2024-10-03
feat(database-clusters): Add server settings for each node and additional data

davidtsuk:
Added server settings data for each node as well as additional columns in the main table.

image


snuba/admin/static/api_client.tsx
```diff
@@ -39,7 +39,7 @@ import { AllocationPolicy } from "SnubaAdmin/capacity_management/types";
 
 import { ReplayInstruction, Topic } from "SnubaAdmin/dead_letter_queue/types";
 import { AutoReplacementsBypassProjectsData } from "SnubaAdmin/auto_replacements_bypass_projects/types";
-import { ClickhouseNodeInfo } from "./database_clusters/types";
+import { ClickhouseNodeInfo, ClickhouseSystemSetting } from "./database_clusters/types";
```
onkar, line 42: Please `import` from absolute path and not relative path.

snuba/admin/static/database_clusters/index.tsx
```diff
@@ -2,17 +2,28 @@ import { MantineReactTable, MRT_ColumnDef, useMantineReactTable } from "mantine-
 import React, { useMemo } from "react";
 import { useEffect, useState } from "react";
 import Client from "SnubaAdmin/api_client";
-import { ClickhouseNodeInfo } from "./types";
+import { ClickhouseNodeInfo, ClickhouseSystemSetting } from "./types";
```
onkar, line 5: Ditto comment on import from relative path.

snuba/admin/views.py
```diff
@@ -1277,3 +1277,23 @@ def execute_job(job_id: str) -> Response:
...
+def clickhouse_system_settings() -> Response:
+    try:
+        host = request.args.get("host")
+        port = request.args.get("port")
+        storage = request.args.get("storage")
+        if not all([host, port, storage]):
+            return make_response(
+                jsonify({"error": "Host, port, and storage are required"}), 400
+            )
+
```
onkar, line 1293: nit: I wonder if these lines need to be under the `try` block.

onkar (approved):
LGTM mostly, added a few comments.

onewland:
can you attach a screenshot for this? this is all accessible from system queries today so I'm just trying to figure out why this is superior

davidtsuk:
> can you attach a screenshot for this? this is all accessible from system queries today so I'm just trying to figure out why this is superior

I've added a screenshot. All this information is accessible from system queries, but the purpose of this tab is to display that information in a nice and convenient way that can be easily accessed. The other advantage is that all the nodes are shown, so people can get an overview of our clusters without having to query individual nodes.

snuba/admin/views.py
davidtsuk: I don't think they need to be, I'll pull them out

onewland:
> > can you attach a screenshot for this? this is all accessible from system queries today so I'm just trying to figure out why this is superior
> 
> I've added a screenshot. All this information is accessible from system queries, but the purpose of this tab is to display that information in a nice and convenient way that can be easily accessed (you can click on a node and see the system settings, and sort/filter/search different fields). The other advantage is that all the nodes are shown, so people can get an overview of our clusters without having to query individual nodes.

how many entries are in that table? if it's >10 or 20, should we potentially modify the query to only return results where `changed=1`? That seems like the more interesting set of settings

PR #6389, opened 2024-10-04
feat(capman): remove legacy cap on threads

volokluev:
This is causing us to not be able to raise the thread limit above 10, remove it since all storages have allocation policies now


onkar (approved):
Is there a `max_threads` attribute in `get_resource_quota()`? If so, should that be removed as well?

volokluev:
> Is there a max_threads attribute in get_resource_quota()? If so, should that be removed as well

It's used to set the clickhouse max threads later so I can't remove it just yet

PR #6396, opened 2024-10-07
feat: Add EAP span subscriptions for devserver

shruthilayaj:
To enable dev work on EAP alerts, this PR starts an alerts scheduler and executor for EAP span alerts in "separate scheduler and executor mode" with snuba devserver.
It adds dataset configuration for span alerts to work (including enabling commit logs for EAP spans).


requirements.txt
```diff
@@ -28,7 +28,7 @@ python-dateutil==2.8.2
 python-rapidjson==1.8
 redis==4.3.4
 sentry-arroyo==2.17.6
-sentry-kafka-schemas==0.1.112
+sentry-kafka-schemas==0.1.113
```
shruthilayaj, line 31: includes this change https://github.com/getsentry/sentry-kafka-schemas/pull/337/

snuba/cli/devserver.py
```diff
@@ -170,6 +170,30 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                     "--auto-offset-reset=latest",
                 ],
             ),
+            (
```
untitaker, line 173: there is an else-branch where you should add the same consumer again. normally in devserver, SEPARATE_SCHEDULER_EXECUTOR_SUBSCRIPTIONS_DEV is False, and all subscription executors are running in a single consumer (for lower resource usage)

shruthilayaj: added https://github.com/getsentry/snuba/pull/6396/commits/d8035005090d523bc427f6d871042d6e5189b575

untitaker (approved):

PR #6398, opened 2024-10-08
feat(profiling): add new columns to the raw_functions table

viglia:
These will be used to later create a new _materialized view_ that will support both the _transaction-based_ and the _continuous_ profiling.


snuba/snuba_migrations/functions/0003_add_new_columns_to_raw_functions.py
```diff
@@ -0,0 +1,86 @@
...
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+
+
+@dataclass(frozen=True)
+class NewColumn:
+    column: Column[Modifiers]
+    after: str | None
```
evanh, line 14: Why would `after` ever be `None`?

snuba/snuba_migrations/functions/0003_add_new_columns_to_raw_functions.py
```diff
@@ -0,0 +1,86 @@
...
+
+from snuba.clickhouse.columns import Column, DateTime64, String
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget
+
+
+@dataclass(frozen=True)
+class NewColumn:
```
evanh, line 12: Is a class really necessary here? I feel like an array of tuples would work just as well.

viglia: strictly necessary? probably not, but I wanted to be consistent with the other migrations in this same dataset: https://github.com/getsentry/snuba/blob/master/snuba/snuba_migrations/functions/0002_add_new_columns_to_raw_functions.py#L11-L14

snuba/snuba_migrations/functions/0003_add_new_columns_to_raw_functions.py
viglia: If there are columns for which we don't need to specify it.

The `after` attribute of the `AddColumn` class after all has it as an optional type (`Optional[str]`) 
https://github.com/getsentry/snuba/blob/master/snuba/migrations/operations.py#L286-L291

evanh (approved):

snuba/snuba_migrations/functions/0003_add_new_columns_to_raw_functions.py
```diff
@@ -0,0 +1,86 @@
...
+    column: Column[Modifiers]
+    after: str | None
+
+
+new_columns: Sequence[NewColumn] = [
```
Zylphrex, line 17: I think this is missing the thread id

Zylphrex (approved):

PR #6401, opened 2024-10-08
feat(eap): Port TraceItemTable endpoint to v1

volokluev:
Implement the V1 TraceItemTable endpoint and bring its functionality on par with the v1alpha SpanSamples endpoint.

Functionality missing from this PR (to be implemented in another):

* PageToken semantics
* Aggregations in the table views


snuba/web/rpc/v1alpha/common.py
```diff
@@ -0,0 +1,353 @@
+from datetime import datetime, timedelta
```
volokluev, line 1: This file is just copied over, there is nothing new here. It was necessary to not have v1alpha/v1 protobuf types conflict (even though they are not meaningfully different

tests/web/rpc/v1/test_endpoint_trace_item_table.py
```diff
@@ -0,0 +1,425 @@
+import random
```
volokluev, line 1: This is basically a copy over from the test_span_samples tests. all the same cases are covered

evanh (approved):

PR #6402, opened 2024-10-09
feat(rpc): handle exceptions by returning a proto and non-200

colin-sentry:
Fixes https://github.com/getsentry/eap-planning/issues/60 (on the snuba side, at least)


snuba/web/views.py
```diff
@@ -280,11 +284,37 @@ def unqualified_query_view(*, timer: Timer) -> Union[Response, str, WerkzeugResp
...
+    except (RPCRequestException, QueryException) as e:
+        exc_proto = convert_rpc_exception_to_proto(e)
+        return Response(exc_proto.SerializeToString(), status=exc_proto.code)
+    except Exception as e:
+        return convert_rpc_exception_to_proto(
+            RPCRequestException(
+                status_code=500,
+                message=f"internal error occurred while executing this RPC call: {e}",
+            )
+        )
```
volokluev, line 317: I would like to have as little logic in this layer as possible and as much as possible moved into the `RPCEndpoint` class. This is doing too much in the flask layer

snuba/web/rpc/common/exceptions.py
```diff
@@ -0,0 +1,31 @@
...
+
+class RPCRequestException(Exception):
+    status_code: int
+
+    def __init__(self, status_code: int, message: str):
+        self.status_code = status_code
+        super().__init__(message)
+
+
+class BadSnubaRPCRequestException(RPCRequestException):
```
volokluev, line 16: Bad/Good are too general. Since it's a 400 error by default it means it's a user error. Let's communicate that
```suggestion
class InvalidSnubaRPCRequestException(RPCRequestException):
```

colin-sentry: 400 is called 'badrequest' semantically

volokluev: fair nuff

snuba/web/rpc/__init__.py
```diff
@@ -101,5 +112,44 @@ def _after_execute(
 }
 
 
-for version, module_path in _TO_IMPORT.items():
-    import_submodules_in_directory(module_path, f"snuba.web.rpc.{version}")
+for v, module_path in _TO_IMPORT.items():
+    import_submodules_in_directory(module_path, f"snuba.web.rpc.{v}")
+
+
+def run_rpc_handler(name: str, version: str, data: bytes) -> Response:
```
volokluev, line 119: write a test for this please

colin-sentry: i added a test for the exception being returned - was there something else you had in mind?

snuba/web/rpc/__init__.py
```diff
@@ -101,5 +112,44 @@ def _after_execute(
 }
 
 
-for version, module_path in _TO_IMPORT.items():
-    import_submodules_in_directory(module_path, f"snuba.web.rpc.{version}")
+for v, module_path in _TO_IMPORT.items():
+    import_submodules_in_directory(module_path, f"snuba.web.rpc.{v}")
```
volokluev, line 116: why the change lol

snuba/web/rpc/__init__.py
colin-sentry: the actual happy path is already tested by the _simple tests

snuba/web/rpc/__init__.py
colin-sentry: mypy fails because "version" is in the global scope here, is easier to just rename

colin-sentry: for some reason python hoists loop variables into the global scope, annoying

snuba/web/rpc/__init__.py
volokluev: What's untested:

1. failed deserialization
2. the endpoint throwing an exception on execution and this function returning an ErrorProto

I know there are endpoint tests but this will also be used by snuba-admin and I would like this to be tested as its own unit, without the flask app as an intermediary

PR #6406, opened 2024-10-10
feat(admin): Show data in clickhouse tracing tool

untitaker:
Show query results, but remove everything that is not a number. That's
practically getting rid of all PII. Discussed beforehand with Evan and
Matt.

We need this to evaluate query correctness in the context of EAP
mutability. How do we know if our percentiles are correct?

But also more generally: If you're comparing two queries for
performance, and the output is wrong in one of them, how are you going
to notice that your entire benchmark is garbage?

image


volokluev (approved):
Can you add a test for `scrub_row`? Otherwise this is brilliant, thanks for doing this

PR #6408, opened 2024-10-10
feat(api): Add a clickhouse connection sanity check to api healthcheck

john-z-yang:
### Overview

This is a follow up item for inc-894, where a API pod landed on a bad node with no clickhouse connection. Because it was not configured with a thorough healthcheck, the pod stayed and served 5XX to all requests.

Adds a connection check to see if any clickhouse query node is operable. Fails the healthcheck if all query nodes are inoperable, as it likely indicates connection error with the host, so k8 can rotate the pod. Renames the previous check_clickhouse function to `check_all_tables_present` to distinguish the 2.

### DIFF
--- a/snuba/utils/health_info.py
+++ b/snuba/utils/health_info.py
@@ -13,6 +13,7 @@ from snuba import environment, settings
 from snuba.clickhouse.errors import ClickhouseError
 from snuba.clusters.cluster import (
     ClickhouseClientSettings,
+    ClickhouseCluster,
     ConnectionId,
     UndefinedClickhouseCluster,
 )
@@ -84,7 +85,11 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
         "thorough": str(thorough),
     }
 
-    clickhouse_health = check_clickhouse(metric_tags=metric_tags) if thorough else True
+    clickhouse_health = (
+        check_all_tables_present(metric_tags=metric_tags)
+        if thorough
+        else sanity_check_clickhouse_connections()
+    )
     metric_tags["clickhouse_ok"] = str(clickhouse_health)
 
     body: Mapping[str, Union[str, bool]]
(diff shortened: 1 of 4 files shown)


snuba/utils/health_info.py
```diff
@@ -120,17 +125,52 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
...
+    """
+    setup_logging(None)
+    logger = logging.getLogger("snuba.health")
+
+    storages: List[Storage] = []
+
+    try:
+        filter_checked_storages(storages)
+    except KeyError:
+        pass
```
john-z-yang, line 149: I'm not entirely sure if I should use the checked_storages function to find clickhouse connections. Please let me know if there's a better idea.

xurui-c: I think this is correct

xurui-c (approved):

snuba/utils/health_info.py
```diff
@@ -120,17 +125,52 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
...
 
 
-def check_clickhouse(metric_tags: dict[str, Any] | None = None) -> bool:
+def sanity_check_clickhouse_connections() -> bool:
+    """
+    Check if at least a sinlge clickhouse query node is operable,
+    returns True if so, False otherwise.
+    """
+    setup_logging(None)
+    logger = logging.getLogger("snuba.health")
```
onkar, line 142: Any reason why this logger setup cannot be done outside this function?

snuba/utils/health_info.py
```diff
@@ -120,17 +125,52 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
...
+    for storage in storages:
+        try:
+            unique_clusters[
+                storage.get_cluster().get_connection_id()
+            ] = storage.get_cluster()
+        except UndefinedClickhouseCluster as err:
+            logger.error(err)
+            continue
+
+    for _, cluster in unique_clusters.items():
```
onkar, line 162: nit: `unique_clusters.values()` can be used since we anyways do not need keys.

onkar (approved):
LGTM. A couple of questions:

1. Is this the liveness probe solution we discussed during the post mortem of inc-894? If so, can we call out the issue and resolution in the PR description?
2. In addition to the query node health, why are we not also checking the storage node health? Or is that already done somewhere else?

snuba/utils/health_info.py
john-z-yang: Nice catch, thanks!

snuba/utils/health_info.py
```diff
@@ -120,17 +125,52 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
...
         storage = get_storage(storage_key)
         if storage.get_readiness_state().value in settings.SUPPORTED_STATES:
             filtered_storages.append(storage)
-    return filtered_storages
 
 
-def check_clickhouse(metric_tags: dict[str, Any] | None = None) -> bool:
+def sanity_check_clickhouse_connections() -> bool:
+    """
+    Check if at least a sinlge clickhouse query node is operable,
```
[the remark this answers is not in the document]
john-z-yang: Fixed, thank you

snuba/utils/health_info.py
john-z-yang: I'm not entirely sure. I'm just following the pattern of the other functions. Updated my branch to see if this passes CI, if so, I will remove it.

john-z-yang:
> LGTM. A couple of questions:
> 
> 1. Is this the liveness probe solution we discussed during the post mortem of inc-894? If so, can we call out the incident number and resolution in the PR description?
> 2. In addition to the query node health, why are we not also checking the storage node health? Or is that already done somewhere else?

1. Yes this is the follow up for inc-894. I didn't reference the inc number in a pr because this is a public open source repo and incidents are internal within the company, but I updated the description.
2. This is technically done by the ingestion consumers, they're the only ones directly communicating with those nodes. If they are not able to perform insert they will exit with a bad status code, forcing k8s to rotate them. Checking them here would not be appropriate here because we still want to serve request to existing data even if ingestion stopped.

enochtangg (approved):

PR #6409, opened 2024-10-10
feat(eap): Implement aggregations in V1 Table endpoint

volokluev:
This endpoint implements aggregations in the v1 table endpoint. This should be sufficient for building any span table views in the product.

To come:

* Support for extrapolation. This PR does not handle sampling extrapolation, that will be done in a subsequent PR


snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -179,7 +203,9 @@ def request_class(cls) -> Type[TraceItemTableRequest]:
         return TraceItemTableRequest
 
     def execute(self, in_msg: TraceItemTableRequest) -> TraceItemTableResponse:
+        in_msg = _apply_labels_to_columns(in_msg)
         snuba_request = _build_snuba_request(in_msg)
+        print(snuba_request.query)
```
evanh, line 208: 💀

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -82,6 +91,9 @@ def _build_query(request: TraceItemTableRequest) -> Query:
             trace_item_filters_to_expression(request.filter),
         ),
         order_by=_convert_order_by(request.order_by),
+        groupby=[
+            attribute_key_to_expression(attr_key) for attr_key in request.group_by
```
evanh, line 95: Will all the non-aggregations in the selected columns also be added to the group by? If not, this code should handle mismatches between selected columns and the groupby.

volokluev: yes good point

snuba/web/rpc/common/common.py
```diff
@@ -64,6 +76,32 @@ def transform(exp: Expression) -> Expression:
     query.transform_expressions(transform)
 
 
+def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
```
volokluev, line 79: @Zylphrex says:

looking this PR, the implementation of all the aggregations on the attribute columns is wrong.
for duration, which is present on every span, it's fine to directly apply the aggregate function over it like avg(duration) but for an attribute, that's wrong because clickhouse uses a default of 0 for spans that dont have that attribute
so avg(attr_num_15[foo]) needs to be to be translated to something like avgIf(attr_num_15[foo], mapContains(attr_num_15, 'foo')) to get the correct aggregation

volokluev: @Zylphrex I created a query processor to do this transformation so as not to pollute the RPC code too much. I also added a test for the case you outlined to make sure it worked as expected

evanh: Is this something that needs to be toggleable? I would think sometimes you would want to take into account the missing spans as well?

volokluev: can you name an actual case?

Zylphrex: I would expect this to behave like nulls vs 0s where nulls should be skipped during aggregation and 0s should be included during aggregation.

evanh: If there's no case to make it toggleable then let's not. Just wanted to know.

Zylphrex (approved):

evanh (approved):

PR #6410, opened 2024-10-10
fix(eap): remove flask stuff from rpc

volokluev:
In #6402 I asked to put as [little of the error handling logic into the flask layer as possible](https://github.com/getsentry/snuba/pull/6402#discussion_r1794202135) but it went too far in the other direction and bled the flask object into the RPC code (which was not the intention).

This PR removes the flask `Response` object out of the RPC code

### DIFF
--- a/snuba/web/rpc/__init__.py
+++ b/snuba/web/rpc/__init__.py
@@ -140,7 +140,7 @@ def run_rpc_handler(
         )
 
     try:
-        return endpoint.execute(deserialized_protobuf)
+        return cast(ProtobufMessage, endpoint.execute(deserialized_protobuf))
     except (RPCRequestException, QueryException) as e:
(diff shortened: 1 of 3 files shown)


nachivrn (approved):

snuba/web/rpc/__init__.py
```diff
@@ -116,40 +116,37 @@ def _after_execute(
     import_submodules_in_directory(module_path, f"snuba.web.rpc.{v}")
 
 
-def run_rpc_handler(name: str, version: str, data: bytes) -> Response:
+def run_rpc_handler(
+    name: str, version: str, data: bytes
+) -> ProtobufMessage | ErrorProto:
```
colin-sentry, line 121: nit: errorproto is a subclass of protobufmessage

colin-sentry (approved):

PR #6414, opened 2024-10-11
fix(rpc): Correct config key for rpc endpoints

Zylphrex:
The config key should be based on the protobuf classes instead of the endpoint.

### DIFF
--- a/snuba/web/rpc/__init__.py
+++ b/snuba/web/rpc/__init__.py
@@ -43,7 +43,9 @@ class RPCEndpoint(Generic[Tin, Tout], metaclass=RegisteredClass):
 
     @classmethod
     def config_key(cls) -> str:
-        return f"{cls.__name__}__{cls.version()}"
+        request_class = cls.request_class().__name__
+        request_version = cls.version()
+        return f"{request_class}__{request_version}"
 
     @property
     def metrics(self) -> MetricsWrapper:
--- a/tests/web/rpc/test_base.py
+++ b/tests/web/rpc/test_base.py
@@ -2,7 +2,6 @@ import time
 from typing import Type
 
 import pytest
-from google.protobuf.message import Message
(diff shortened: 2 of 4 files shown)


volokluev:
@Zylphrex 

How strongly do you feel that this should be the case? I'm not super attached to it but also don't necessarily feel like updating all the tests. The naming convention follows the `endpoint_*` file naming conventions stated [here](https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/snuba/README.md#conventions)

Zylphrex:
> @Zylphrex
> 
> How strongly do you feel that this should be the case? I'm not super attached to it but also don't necessarily feel like updating all the tests. The current naming convention follows the `endpoint_*` file naming conventions stated [here](https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/snuba/README.md#conventions)

I'm not attached to it but the names of the classes are now public interface as it determines the endpoint. So if we were to use the class names in snuba to decide the endpoint paths, we now have to
1. ensure they do not change
2. we now need to hard code a mapping from the rpc name to the snuba class name within the sentry code base for every rpc call as we can not rely on the rpc class to tell us which endpoint it needs to call

volokluev:
> we now need to hard code a mapping from the rpc name to the snuba class name within the sentry code base for every rpc call as we can not rely on the rpc class to tell us which endpoint it needs to call

@Zylphrex I see what you're saying. Here is what I will say. In gRPC, the definition of the RPC and request/response payload are separate ([example](https://grpc.io/docs/what-is-grpc/core-concepts/#service-definition)) because they are separate things. When we switch to gRPC you'd have to do this anyways. What you are proposing would make that transition more difficult

Zylphrex:
Closing this in favour of helper functions in sentry that maps to each rpc endpoint

PR #6416, opened 2024-10-15
feat: add support for Tuple column data type

viglia:
This PR add support for the Tuple Clickhouse data type in Snuba.

See clickhouse docs: https://clickhouse.com/docs/en/sql-reference/data-types/tuple


evanh (commented):
This looks fine, could you add some tests? Like to `test_columns_validator.py`?

viglia:
> seems fine, any background on why we need the tuple column?

@onewland yes, we're going to create a new `functions` table (let's call it v2), that is going to replace the existing one to support both continuous profiling and transaction-based profiling.

In this case the `example` and the `worst` column won't hold a simple `profile_id` anymore but rather a tuple `(profile_id, thread_id, start_timestamp, end_timestamp)`.

evanh (approved):

PR #6418, opened 2024-10-15
fix(migrations): Only check inactive replicas for clusters migrations are being run on

davidtsuk:
Fixes [SNS-2888](https://getsentry.atlassian.net/browse/SNS-2888)

[SNS-2888]: https://getsentry.atlassian.net/browse/SNS-2888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snuba/migrations/connect.py
```diff
@@ -157,27 +157,17 @@ def _get_all_nodes_for_storage(
...
-                storage_key
-            )
-            storage = get_storage(storage_key)
-            cluster = storage.get_cluster()
-        except UndefinedClickhouseCluster:
-            continue
-
-        for node in (*local_nodes, *distributed_nodes, query_node):
+    for cluster in clusters:
+        for node in cluster.get_local_nodes():
```
onkar, line 167: It seems like previously we were traversing local nodes and query node but now we are only looping over local nodes on the cluster. Any reason to skip query node?

davidtsuk: The system.replicas table is only available for replicated tables and query nodes use distributed tables which won't have any data in that table.

snuba/migrations/runner.py
```diff
@@ -279,6 +291,10 @@ def run_migration(
 
         migration_group, migration_id = migration_key
 
+        check_for_inactive_replicas(
+            get_clickhouse_clusters_for_migration_group(migration_group)
+        )
+
```
volokluev, line 297: I'm pretty sure you only need this check in `_run_migration_impl`. Right now you are repeating the same check three times, which will make the codebase more brittle over time. Try to find a place where this can be specified once.

volokluev (commented):
How did you test that this works?

onkar (approved):
LGTM

snuba/migrations/runner.py
davidtsuk: I could do this, but I will still need to add this check to `_reverse_migration_impl` as well. It will also end up sending more queries to clickhouse as it will query all local nodes in a cluster for each migration. If you're okay with that, I'll move it.

davidtsuk:
> How did you test that this works?

I modified the existing test to ensure the check function still works as intended. I've also tried reversing a migration and then running it again.

snuba/migrations/runner.py
davidtsuk: I moved it since it does simplify the code

volokluev (approved):

PR #6419, opened 2024-10-15
ref(subscriptions): Introduce abstract SubscriptionData class

shruthilayaj:
This is a refactor to prepare for RPC subscriptions.
1. Renames SubscriptionData to SnQLSubscriptionData ([commit](https://github.com/getsentry/snuba/pull/6419/commits/be2a040c63bcc14f67782947d15e5a43d452ced6), easiest reviewed on its own, and then the rest).
2. Introduces an abstract SubscriptionData class that SnQLSubscriptionData & eventually RPCSubscriptionData will inherit from (something like this, although [this](https://github.com/getsentry/snuba/pull/6434/files) is not ready for review).

The idea is that the RPCSubscription class will store a serialized proto instead of `query` in SnQLSubscriptionData.


snuba/subscriptions/data.py
```diff
@@ -74,20 +74,53 @@ def from_string(cls, value: str) -> SubscriptionIdentifier:
         return cls(PartitionId(int(partition)), UUID(uuid))
 
 
-@dataclass(frozen=True)
-class SubscriptionData:
-    """
-    Represents the state of a subscription.
-    """
-
+@dataclass(frozen=True, kw_only=True)
```
shruthilayaj, line 77: Adding the `kw_only` arg so we can specify a default for tenant_ids in the base class here

>If true (the default value is False), then all fields will be marked as keyword-only. If a field is marked as keyword-only, then the only affect is that the __init__() parameter generated from a keyword-only field must be specified with a keyword when __init__() is called.

snuba/subscriptions/data.py
```diff
@@ -74,20 +76,56 @@ def from_string(cls, value: str) -> SubscriptionIdentifier:
...
     project_id: int
     resolution_sec: int
     time_window_sec: int
     entity: Entity
-    query: str
     metadata: Mapping[str, Any]
     tenant_ids: Mapping[str, Any] = field(default_factory=lambda: dict())
 
+
+TRequest = TypeVar("TRequest")
```
shruthilayaj, line 89: Generic return type for `build_request` since it's going to be different for RPCs

volokluev:
Question: why is inheritance necessary for this subscription data class?

snuba/subscriptions/codecs.py
```diff
@@ -10,29 +10,29 @@
...
-    SubscriptionData,
     SubscriptionIdentifier,
     SubscriptionTaskResult,
     SubscriptionWithMetadata,
 )
 from snuba.utils.codecs import Codec, Encoder
 
 
-class SubscriptionDataCodec(Codec[bytes, SubscriptionData]):
+class SubscriptionDataCodec(Codec[bytes, SnQLSubscriptionData]):
```
evanh, line 22: Shouldn't this still take a `SubscriptionData` instead of `SnQLSubscriptionData`? Once you add the RPC type, that's what will need to happen here.

snuba/subscriptions/partitioner.py
```diff
@@ -2,12 +2,12 @@
 from binascii import crc32
 
 from snuba.datasets.table_storage import KafkaTopicSpec
-from snuba.subscriptions.data import PartitionId, SubscriptionData
```
evanh, line 5: Why can't this still use `SubscriptionData`?

snuba/subscriptions/codecs.py
shruthilayaj: Yes, you're right! I was doing union types for some reason when I introduce the RPC class, this absolutely did not cross my mind 😅

shruthilayaj: addressed in https://github.com/getsentry/snuba/pull/6419/commits/97a0360ffed6c0c10981c776611eeafea55147f1

snuba/subscriptions/codecs.py
```diff
@@ -32,7 +33,7 @@ def decode(self, value: bytes) -> SubscriptionData:
         except json.JSONDecodeError:
             raise InvalidQueryException("Invalid JSON")
 
-        return SubscriptionData.from_dict(data, self.entity_key)
+        return SnQLSubscriptionData.from_dict(data, self.entity_key)
```
evanh, line 36: Shouldn't this be just `SubscriptionData`?

shruthilayaj: `SubscriptionData` is the base abstract class now

evanh (approved):

PR #6420, opened 2024-10-16
feat(profiling): add columns to existing materialized functions table and create a new view (functions_v2_local)

viglia:
This PR add a 3 new columns to the existing materialized functions table and create a new view (`functions_v2_local`) that read data with `materialization_version=1` and write to the materialized functions table accordingly.

Tested E2E with a bunch of transactions-based profiles:

```
Query:
select argMaxMerge(worst_v2) from functions_mv_local fml where fml.worst_v2 is not NULL LIMIT 1;

Result:
argMaxMerge(worst_v2)                                |
-----------------------------------------------------+
('cad66e4f-e5eb-4f25-a110-4f5e0e596db8','',NULL,NULL)|


Query:
select groupUniqArrayMerge(5)(examples_v2) from functions_mv_local fml where fml.worst_v2 is not NULL LIMIT 1;

Result:
groupUniqArrayMerge(5)(examples_v2)                                                                                                                                      |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
['('be7b50ec-90fd-4e91-864c-d0986b67ac47'','','','NULL)','('cad66e4f-e5eb-4f25-a110-4f5e0e596db8'','','','NULL)','('b756c0b8-8bc1-4447-85e8-374f6f31a4a6'','','','NULL)']|
```


viglia (commented):
-

snuba/snuba_migrations/functions/0004_functions_v2.py
```diff
@@ -0,0 +1,197 @@
...
+            operations.CreateTable(
+                storage_set=self.storage_set,
+                table_name=self.local_functions_table,
+                columns=columns,
+                engine=table_engines.AggregatingMergeTree(
+                    storage_set=self.storage_set,
+                    order_by="(project_id, timestamp, transaction_name, fingerprint, name, package, is_application, profiling_type, platform, environment, release, retention_days)",
+                    primary_key="(project_id, timestamp, transaction_name, fingerprint)",
+                    partition_by="(retention_days, toMonday(timestamp))",
+                    settings={"index_granularity": "2048", "allow_nullable_key": 1},
```
viglia, line 114: @Zylphrex I've carried over this specific `index_granularity` settings from the previous `functions` mv.

Can you remind me why this was picked specifically instead of the default granule size of `8012`? Did you ran some experiments on existing functions data and it turned out the queries were too slow with `8012`?

Zylphrex: I'm not sure, it may have just a copy paste from another table.

Zylphrex (commented):
I'm not sure this is what we want to do. This will create a new table in clickhouse that we have to dual write to for 90 days before switching. What we originally discussed was to add columns to the existing table so we can dual write to the new columns until we have all the data.

viglia:
> I'm not sure this is what we want to do. This will create a new table in clickhouse that we have to dual write to for 90 days before switching. What we originally discussed was to add columns to the existing table so we can dual write to the new columns until we have all the data.

@Zylphrex I've refactored with the changes we discussed.

evanh (commented):
Couple things:

Is there already a materialized view writing to those tables? If so, the old view will still be writing to the tables. Is that intended?

Can you add tests for the schema changes?

viglia:
> Couple things:
> 
> Is there already a materialized view writing to those tables? If so, the old view will still be writing to the tables. Is that intended?

Yes, this is intended. The older one is only reading data with `materialized_version=0`, the newer one is reading data with `materialized_version=1`.

> Can you add tests for the schema changes?

aren't migration changes already covered from all the tests running (migrations) in CI?

evanh (approved):

snuba/snuba_migrations/functions/0004_functions_v2.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
+                storage_set=self.storage_set,
+                table_name=table_name,
+                column=new_column.column,
+                after=new_column.after,
+                target=target,
+            )
```
Zylphrex, line 143: What's this? Dont most if not all these columns already exist on the table?

viglia: @Zylphrex I'm only adding 3 missing columns to the existing materialized table (namely `worst_v2`, `example_v2`, `profiling_type`). You can see I'm only adding columns from _new_columns_ `column=new_column.column`

All the other columns that I've listed, it's because I have to specify them in the new `view` (`columns=all_columns`), but I'm not adding all of those again in the materialized functions table.

Zylphrex: ahh i see 👍

snuba/snuba_migrations/functions/0004_functions_v2.py
```diff
@@ -0,0 +1,232 @@
...
+                countState(arrayJoin(durations) AS duration) AS count,
+                quantilesState(0.5, 0.75, 0.9, 0.95, 0.99)(duration) AS percentiles,
+                minState(duration) AS min,
+                maxState(duration) AS max,
+                avgState(duration) AS avg,
+                sumState(duration) AS sum,
+                argMaxState(profile_id, duration) as worst,
+                argMaxState((profile_id, thread_id, start_timestamp, end_timestamp), duration) as worst_v2,
+                groupUniqArrayState(5)(profile_id) as examples,
+                groupUniqArrayState(5)((profile_id, thread_id, start_timestamp, end_timestamp)) as examples_v2
```
Zylphrex, line 211: Hm... I just realized that in order to truly drop the existing columns `worst` and `examples`, we'll need yet another migration where we create another materialized view but without those 2 aggregations.

viglia: Yes, we can think about that in ~90 days.

For now it's handy to have both since depending on the time-range we might still need to access examples for which we only have data in the older `worst` and `examples` columns.

PR #6423, opened 2024-10-16
ref(eap-mutations): Use session-check=1

untitaker:
We are seeing this error in prod:

```
Strategy(bad response while inserting mutation, status: 404 Not Found, response body: Ok("Code: 60. DB::Exception: Table default.new_data does not exist. (UNKNOWN_TABLE) (version 23.8.11.29.altinitystable (altinity build))\n")
```

Let's see if the session exists at all by using session_check=1. Maybe
it timed out or got dropped by some load balancer.


rust_snuba/src/mutations/clickhouse.rs
```diff
@@ -62,16 +62,34 @@ impl ClickhouseWriter {
...
+                // https://clickhouse.com/docs/en/interfaces/http#http_response_codes_caveats
+                // TODO: port to main consumer
+                ("wait_end_of_query", "1"),
+            ]);
+
+            if session_check {
+                // ensure that we are in a session. if some load balancer drops the querystring, we
+                // want to know
+                request = request.query(&[("session_check", "1")]);
+            }
```
onkar, line 88: Will this `if` block prevent enabling session check for the 0th query request? `session_check` is set to `false` to start with.

untitaker: yep that's the intent. the session is upserted on the first request

evanh (approved):

PR #6427, opened 2024-10-16
feat(job-runner): async job functionality

xurui-c:
https://getsentry.atlassian.net/jira/software/c/projects/SNS/boards/147?selectedIssue=SNS-2871

At times we will want to execute long-running ClickHouse jobs that don’t fit neatly into the migrations system; either they only run in a single environment, or we don’t want to block later migrations on them finishing. This PR supports users to create job types that can run asynchronously.

### DIFF
--- a/snuba/manual_jobs/__init__.py
+++ b/snuba/manual_jobs/__init__.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Any, MutableMapping, Optional, cast
 
-from snuba.manual_jobs.redis import _set_job_is_async, _set_job_type
+from snuba.manual_jobs.redis import _set_job_type
 from snuba.utils.registered_class import RegisteredClass, import_submodules_in_directory
 
 logger = logging.getLogger("snuba.manual_jobs")
@@ -44,10 +44,6 @@ class Job(ABC, metaclass=RegisteredClass):
     def __init__(self, job_spec: JobSpec) -> None:
         self.job_spec = job_spec
         self.is_async = job_spec.is_async
-        _set_job_is_async(
-            job_spec.job_id,
-            job_spec.is_async if job_spec.is_async is not None else False,
-        )
         _set_job_type(job_spec.job_id, job_spec.job_type)
         if job_spec.params:
             for k, v in job_spec.params.items():
@@ -67,7 +63,7 @@ class Job(ABC, metaclass=RegisteredClass):
 
     @classmethod
     def async_finished_check(cls, job_id: str) -> bool:
-        return True
+        raise NotImplementedError
 
 
 import_submodules_in_directory(
--- a/snuba/manual_jobs/__init__.py
+++ b/snuba/manual_jobs/__init__.py
@@ -2,19 +2,12 @@ import logging
 import os
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from enum import StrEnum
 from typing import Any, MutableMapping, Optional, cast
 
+from snuba.manual_jobs.redis import _set_job_is_async, _set_job_type
 from snuba.utils.registered_class import RegisteredClass, import_submodules_in_directory
 
 logger = logging.getLogger("snuba.manual_jobs")
-
-
-class JobStatus(StrEnum):
-    RUNNING = "running"
-    FINISHED = "finished"
-    NOT_STARTED = "not_started"
-    FAILED = "failed"
 
 
 class JobLogger(ABC):
@@ -43,12 +36,19 @@ class JobLogger(ABC):
 class JobSpec:
     job_id: str
     job_type: str
+    is_async: Optional[bool] = False
     params: Optional[MutableMapping[Any, Any]] = None
 
 
 class Job(ABC, metaclass=RegisteredClass):
     def __init__(self, job_spec: JobSpec) -> None:
         self.job_spec = job_spec
+        self.is_async = job_spec.is_async
+        _set_job_is_async(
+            job_spec.job_id,
+            job_spec.is_async if job_spec.is_async is not None else False,
+        )
+        _set_job_type(job_spec.job_id, job_spec.job_type)
         if job_spec.params:
             for k, v in job_spec.params.items():
                 setattr(self, k, v)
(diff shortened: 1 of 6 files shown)


snuba/manual_jobs/job_status.py
```diff
@@ -0,0 +1,9 @@
+from enum import StrEnum
```
xurui-c, line 1: job_status.py was created to avoid circular dependency

snuba/manual_jobs/runner.py
```diff
@@ -84,10 +98,12 @@ def list_job_specs_with_status(
     manifest_filename: str = MANIFEST_FILENAME,
 ) -> Mapping[str, Mapping[str, Union[JobSpec, JobStatus]]]:
     specs = list_job_specs(manifest_filename)
-    job_ids = specs.keys()
+    job_ids = list(specs.keys())
     statuses = _get_job_status_multi(
         [_build_job_status_key(job_id) for job_id in job_ids]
     )
+    for i in range(len(job_ids)):
```
xurui-c, line 105: I know this for loop defeats the purpose of mget used in `_get_job_status_multi`, so I'm also not opposed to write `_get_is_job_async_multi` and `_get_job_types_multi`

snuba/manual_jobs/runner.py
```diff
@@ -65,13 +68,24 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
     return job_spec
 
 
+def _update_job_status_if_async(job_id: str, job_status: JobStatus) -> JobStatus:
+    if _redis_client.get(
+        name=_build_is_job_async_key(job_id)
+    ) is True and Job.get_from_name(_get_job_type(job_id)).async_finished_check(job_id):
```
xurui-c, line 74: I should put this in a method _get_is_job_async()

snuba/manual_jobs/__init__.py
```diff
@@ -43,12 +36,19 @@ def error(self, line: str) -> None:
...
 
 
 class Job(ABC, metaclass=RegisteredClass):
     def __init__(self, job_spec: JobSpec) -> None:
         self.job_spec = job_spec
+        self.is_async = job_spec.is_async
+        _set_job_is_async(
+            job_spec.job_id,
+            job_spec.is_async if job_spec.is_async is not None else False,
+        )
```
onewland, line 50: do we need to set a special field to track whether a job is async?

could we just detect that by looking at a job status and seeing if it's in `async_running_background`?

snuba/manual_jobs/job_status.py
onewland: where were you having a circular import?

xurui-c: `redis.py` imports `__init__.py/JobStatus`, `__init__.py` imports `redis.py?_set_job_type`

snuba/manual_jobs/__init__.py
xurui-c: you are right

snuba/manual_jobs/runner.py
```diff
@@ -65,13 +67,24 @@ def _build_job_spec_from_entry(content: Any) -> JobSpec:
...
     redis_status = _redis_client.get(name=_build_job_status_key(job_id))
-    return (
+    status = _update_job_status_if_async(
+        job_id,
         JobStatus(redis_status.decode("utf-8"))
         if redis_status
-        else JobStatus.NOT_STARTED
+        else JobStatus.NOT_STARTED,
     )
+    return status
```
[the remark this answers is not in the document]
xurui-c: What additional benefits will logging provide, if it's already going to be displayed on Snuba Admin?

tests/manual_jobs/test_job_statuses.py
```diff
@@ -0,0 +1,27 @@
+import pytest
```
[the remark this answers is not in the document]
xurui-c: Those are already in `test_job.py`. I was thinking of moving them to `test_job_statuses.py` in a subsequent PR, since it falls outside the scope of async jobs

snuba/manual_jobs/runner.py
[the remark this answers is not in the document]
onkar: Ah, if this is going to be return to snuba admin, then we are okay.

snuba/manual_jobs/runner.py
onkar: What would the implementations of `_get_is_job_async_multi` and `_get_job_types_multi` look like?

onewland (commented):
I think the overall logic is fine but we need a test that triggers the transition of async job from `ASYNC_RUNNING_BACKGROUND` to `FINISHED`.

snuba/manual_jobs/__init__.py
```diff
@@ -65,6 +61,10 @@ def config_key(cls) -> str:
     def get_from_name(cls, name: str) -> "Job":
         return cast("Job", cls.class_from_name(name))
 
+    @classmethod
+    def async_finished_check(cls, job_id: str) -> bool:
+        return True
```
onewland, line 66: I think this should raise an exception (`NotImplemented` or something) so that we don't accidentally have jobs transitioning from ASYNC_RUNNING_BACKGROUND to FINISHED because somebody forgets to override

snuba/manual_jobs/runner.py
xurui-c: We got rid of `[get_is_job_async](https://github.com/getsentry/snuba/pull/6427#discussion_r1809149842)`, so I won't be implementing `_get_is_job_async_multi`. The implementation of `_get_job_types_multi` will just mirror `_get_job_types_multi` and `_get_job_status_multi`

tests/manual_jobs/test_job_statuses.py
```diff
@@ -0,0 +1,37 @@
...
+    run_job(async_job_spec)
+    assert get_job_status(JOB_ID) == JobStatus.ASYNC_RUNNING_BACKGROUND
+
+
+@pytest.mark.redis_db
+def test_job_status_changes_from_async_running_background_to_finished() -> None:
+    assert get_job_status(JOB_ID) == JobStatus.NOT_STARTED
+    run_job(async_job_spec)
+    with patch.object(AsyncJob, "async_finished_check", return_value=True):
+        assert get_job_status(JOB_ID) == JobStatus.FINISHED
```
onewland, line 37: sorry to be annoying, but could we add a third assertion here to verify this job passes through  `ASYNC_RUNNING_BACKGROUND` before reaching `FINISHED`?

onewland (approved):
Looks fine if the test faillure is just a flake

snuba/manual_jobs/runner.py
xurui-c: ^ will revisit this in subsequent PR

PR #6430, opened 2024-10-16
ref(deletes): remove simple ratelimit

MeredithAnya:
We have allocation policies set in place now, so we can remove the simple rate-limiter.


onkar:
Are there any unit tests etc. that need to be removed once this simple rate limiter is removed?

kylemumma (approved):

onkar (approved):

PR #6431, opened 2024-10-16
fix(devserver): Metric alert consumers follow the right group

shruthilayaj:
@edwardgou-sentry found a bug with how we were spinning up alerts schedulers for metric alerts.
They were not following the right consumer group, they need to follow the group producing the commit logs.


snuba/cli/devserver.py
```diff
@@ -337,7 +337,7 @@ def devserver(*, bootstrap: bool, workers: bool) -> None:
                             "subscriptions-scheduler",
                             "--entity=generic_metrics_distributions",
                             "--consumer-group=snuba-generic-metrics-distributions-subscriptions-schedulers",
-                            "--followed-consumer-group=snuba-generic-metrics-distributions-consumers",
+                            "--followed-consumer-group=snuba-gen-metrics-distributions-consumers",
```
shruthilayaj, line 340: For example, https://github.com/getsentry/snuba/pull/6431/files#diff-27ffd2f11e990878aec700a2b0460871300553cc59992f0403c6782bdaf0763aR269

volokluev (approved):

PR #6434, opened 2024-10-17
draft(subscriptions): Introduce RPCSubscriptionData class

shruthilayaj:
Draft, I want to make sure this is will do what I want it to do


snuba/subscriptions/data.py
```diff
@@ -111,10 +120,77 @@ def to_dict(self) -> Mapping[str, Any]:
...
 
 
+@dataclass(frozen=True, kw_only=True)
+class RPCSubscriptionData(SubscriptionData):
+    """
+    Represents the state of an RPC subscription.
+    """
+
+    table_request: str
+
```
volokluev, line 130: 1. the table_request is bytes, not a string
2. we should be storing the name of the endpoint and the version of it. The API will evolve and we need to be able to change the model along with it

snuba/subscriptions/data.py
```diff
@@ -111,10 +136,109 @@ def to_dict(self) -> Mapping[str, Any]:
...
+    def build_request(
```
volokluev, line 147: returning a snuba `Request` object here does not make sense. The RPC implementations will build whatever requests are necessary and execute them

PR #6436, opened 2024-10-17
ref(eap-mutations): Implement synchronizer

untitaker:
Induce a hardcoded delay so that the eap-mutations consumer will always
lag behind the eap-spans consumer.

This PR (and most of my work on the mutations consumer) makes a point to not use custom strategies, as that's too much
boilerplate. Rather, we should improve the arroyo API. For this PR, this
arroyo PR is necessary: https://github.com/getsentry/arroyo/pull/388


evanh (approved):
Question: presumably this is going to cause a bunch of errors to get thrown, at least when it is first deployed, since none of the messages coming in currently have been delayed.

Is there a concern with that at all? Or is it just something that will blip for a bit and then stabilize?

untitaker:
those errors are not logged, just causing backpressure (paused state in arroyo) arroyo will emit a warning at most every 5 seconds for when the consumer is seeing an abnormal/constant amount of backpressure, so that might cause some noise in sentry. but we already have an issue for that, so it should group nicely.

rust_snuba/src/mutations/synchronize.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl Synchronizer {
+    pub fn process_message<T>(
+        &mut self,
+        message: Message<T>,
+    ) -> Result<Message<T>, SubmitError<T>> {
+        if let Some(ts) = message.timestamp() {
+            if Utc::now() - ts < self.min_delay {
+                sleep(Duration::from_secs(10));
+                return Err(MessageRejected { message }.into());
```
ayirr7, line 27: Can you explain why we return this Err here?

rust_snuba/src/mutations/factory.rs
```diff
@@ -97,6 +101,12 @@ impl ProcessingStrategyFactory<KafkaPayload> for MutConsumerStrategyFactory {
             Some("parse"),
         );
 
+        let mut synchronizer = Synchronizer {
+            min_delay: TimeDelta::days(1),
```
ayirr7, line 105: Any reason why we specifically chose 1 day here?

rust_snuba/src/mutations/synchronize.rs
ayirr7: Nvm, ignore

evanh (approved):

ayirr7 (approved):

rust_snuba/src/mutations/synchronize.rs
```diff
@@ -0,0 +1,33 @@
...
+            }
```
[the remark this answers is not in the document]
untitaker: added a code comment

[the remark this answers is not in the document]
evanh: That would cause the consumer to loop constantly (due to some idiosyncrasies in arroyo) , the sleep prevents the CPU idling at 100%.

onkar (approved):

PR #6440, opened 2024-10-17
ref(bulk-deletes): bulk delete query path and producer

MeredithAnya:
**context**
The current pipeline we have for deleting data (via lightweight deletes) is an API endpoint that executes the query directly. `DELETE` requests are technically synchronous at the moment due to the ClickHouse version we are on (the `lightweight_delete_mutations_sync` parameter exist in newer version).

**problem**
We'd like to keep the number of concurrent `DELETE` requests to a minimum since they add an increased load to ClickHouse. Though the number of rows being deleted does have an effect on load, even a delete that deletes a few rows can have a significant enough effect because it's still a mutation (under the hood the `DELETE FROM` becomes an `UPDATE`). 

**batched deletes**
In order to scale we are adding a consumer that will batch up DELETE queries so that we have fewer mutations, even though the number of rows deleted/mutated will be higher.

The new pipeline is a bit different and has some new components:
- [ ] `DELETE` api endpoint will now take a request and produce it to a topic (This PR)
    * The topics are storage specific, ex. `LW_DELETIONS_SEARCH_ISSUES = "snuba-lw-deletions-search-issues"`
- [ ] Consumer(s) will be added to consume from the topic that corresponds to the storage the consumer is for, right now this will just be search issues

**other notable changes (in this pr)**
* addition of `allowed_columns` in the deletion settings, this will eventually replace the `ColumnFilterProcessor` once deletes are fully on the bulk delete endpoint
* `rows_to_delete` is added to the delete query payload so that we can batch based on number of rows to be delete instead of batching on number of messages

### DIFF
--- a/snuba/datasets/deletion_settings.py
+++ b/snuba/datasets/deletion_settings.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Optional, Sequence
+from typing import Sequence
 
 MAX_ROWS_TO_DELETE_DEFAULT = 100000
 
@@ -10,5 +10,5 @@ MAX_ROWS_TO_DELETE_DEFAULT = 100000
 class DeletionSettings:
     is_enabled: int
     tables: Sequence[str]
-    allowed_columns: Optional[Sequence[str]] = field(default_factory=list)
+    allowed_columns: Sequence[str] = field(default_factory=list)
     max_rows_to_delete: int = MAX_ROWS_TO_DELETE_DEFAULT
--- a/snuba/datasets/deletion_settings.py
+++ b/snuba/datasets/deletion_settings.py
@@ -10,4 +10,5 @@ MAX_ROWS_TO_DELETE_DEFAULT = 1000
 class DeletionSettings:
     is_enabled: int
     tables: Sequence[str]
+    allowed_columns: Sequence[str]
     max_rows_to_delete: int = MAX_ROWS_TO_DELETE_DEFAULT
--- a/snuba/datasets/deletion_settings.py
+++ b/snuba/datasets/deletion_settings.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
-from typing import Sequence
+from dataclasses import dataclass, field
+from typing import Optional, Sequence
 
-MAX_ROWS_TO_DELETE_DEFAULT = 1000
+MAX_ROWS_TO_DELETE_DEFAULT = 100000
 
 
 @dataclass
 class DeletionSettings:
     is_enabled: int
     tables: Sequence[str]
-    allowed_columns: Sequence[str]
+    allowed_columns: Optional[Sequence[str]] = field(default_factory=list)
     max_rows_to_delete: int = MAX_ROWS_TO_DELETE_DEFAULT
--- a/snuba/web/bulk_delete_query.py
+++ b/snuba/web/bulk_delete_query.py
@@ -57,7 +57,6 @@ class InvalidStorageTopic(Exception):
 def _get_kafka_producer(topic: Topic) -> Producer:
     producer = PRODUCER_MAP.get(topic.value)
     if not producer:
-
         producer = Producer(
             build_kafka_producer_configuration(
                 topic=topic,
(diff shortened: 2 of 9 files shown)


snuba/datasets/deletion_settings.py
```diff
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Sequence
 
-MAX_ROWS_TO_DELETE_DEFAULT = 1000
+MAX_ROWS_TO_DELETE_DEFAULT = 100000
```
armenzg, line 6: What happens if an issue has more than these many rows?

snuba/web/bulk_delete_query.py
```diff
@@ -0,0 +1,183 @@
...
+    delete_settings = storage.get_deletion_settings()
+    if not delete_settings.is_enabled:
+        raise DeletesNotEnabledError(
+            f"Deletes not enabled for {storage.get_storage_key().value}"
+        )
+
+    if list(conditions.keys()).sort() != list(delete_settings.allowed_columns).sort():
+        raise InvalidQueryException(
+            f"Invalid Columns to filter by, must be in {delete_settings.allowed_columns}"
+        )
```
onkar, line 110: To avoid sorting lists explicitly, can we do this?

```
difference =  set(conditions.keys()) - set(delete_settings.allowed_columns)
if difference != set():
    raise InvalidQueryException(
        f"Invalid Columns {difference} to filter by, must be in {delete_settings.allowed_columns}"
    )
```

snuba/web/bulk_delete_query.py
```diff
@@ -0,0 +1,183 @@
...
+def construct_or_conditions(conditions: Sequence[Dict[str, Any]]) -> Expression:
+    """
+    Combines multiple AND conditions: (equals(project_id, 1) AND in(group_id, (2, 3, 4, 5))
+    into OR conditions for a bulk delete
+    """
+    or_conditions = []
+    for cond in conditions:
+        and_condition = _construct_condition(cond)
+        or_conditions.append(and_condition)
+    return combine_or_conditions(or_conditions)
```
onkar, line 183:
```
return combine_or_conditions([_construct_condition(cond) for cond in conditions])
```

onkar (commented):
Left comments. Please add unit tests since the existing coverage does not seem sufficient.

snuba/datasets/deletion_settings.py
kylemumma: its looks like this function https://github.com/getsentry/snuba/blob/55398e54eaa9d96c4ce6a3eeb6816e41195cb975/snuba/web/delete_query.py#L137
will raise an exception and the delete will fail.

@MeredithAnya reading this function I realize it could cause an issue. Since we enforce max rows and fail in delete_from_table. if there are multiple tables the first one delete could succeed, but the second has too many rows and fails causing inconsistency. I think we need to change it so we check max rows on all tables before we start deleting on any.

snuba/web/bulk_delete_query.py
```diff
@@ -0,0 +1,183 @@
...
+    rows_to_delete: int
+    storage_name: str
+    conditions: ConditionsType
+    tenant_ids: Mapping[str, str | int]
+
+
+delete_kfk: Producer | None = None
+
+
+def _delete_kafka_producer() -> Producer:
```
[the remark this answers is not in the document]
kylemumma: I agree

snuba/datasets/deletion_settings.py
kylemumma: I will hold off implementing the bug fix until this PR is merged since it would cause a merge conflict. I am happy to take on this change though.

snuba/web/bulk_delete_query.py
```diff
@@ -0,0 +1,183 @@
...
+        producer.poll(0)  # trigger queued delivery callbacks
+        producer.produce(
+            settings.KAFKA_TOPIC_MAP.get(
+                Topic.LW_DELETIONS.value, Topic.LW_DELETIONS.value
+            ),
+            data.encode("utf-8"),
+            on_delivery=_record_query_delivery_callback,
+            # TODO(should we add headers? or key)
+        )
+    except Exception as ex:
```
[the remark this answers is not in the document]
MeredithAnya: @onkar I wondering if for now this is okay to keep whatever the error we want to know about it if we had a problem producing to the topic.

onkar (approved):

PR #6442, opened 2024-10-22
feat(eap-spans): Add a new schema for spans

phacops:
- Change the partition key to include `_sort_timestamp` so we can use `ttl_only_drop_parts` and not go through TTLDelete merges
    - This forces us to recreate a new table
- Fix `sample_weight` type
- Change codecs as suggested in our call with Altinity
- Change `duration_ms` and `exclusive_time_ms` to be microseconds stored into a `UInt64` column

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0016_spans_v2.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0016_spans_v2.py
@@ -41,12 +41,16 @@ columns: List[Column[Modifiers]] = [
         DateTime64(6, modifiers=Modifiers(codecs=["DoubleDelta", "ZSTD(1)"])),
     ),
     Column(
-        "duration_ms",
-        UInt(32, modifiers=Modifiers(codecs=["T64", "ZSTD(1)"])),
+        "duration_micro",
+        UInt(64, modifiers=Modifiers(codecs=["T64", "ZSTD(1)"])),
+    ),
+    Column(
+        "exclusive_time_micro",
+        UInt(64, modifiers=Modifiers(codecs=["T64", "ZSTD(1)"])),
     ),
     Column(
         "retention_days",
-        UInt(16, modifiers=Modifiers(codecs=["DoubleDelta", "ZSTD(1)"])),
+        UInt(16, modifiers=Modifiers(codecs=["T64", "ZSTD(1)"])),
     ),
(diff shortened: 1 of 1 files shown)


colin-sentry (changes requested):
this all seems reasonable to me

colin-sentry (approved):
this all seems reasonable to me

evanh (changes requested):
Why are we adding start_timestamp to the sort key? And why is it necessary for the parts to drop?

evanh (commented):
As far as I can tell all this PR really does is remove the `_sort_timestamp` column? Is that accurate? Since _sort_timestamp is already mapped to the start_timestamp?

Note this will require changing the consumers, rebuilding the materialized views, changing mutations consumer and also the query API.

snuba/snuba_migrations/events_analytics_platform/0016_spans_v2.py
```diff
@@ -0,0 +1,138 @@
...
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("service", String(Modifiers(codecs=["ZSTD(1)"]))),
+    Column("trace_id", UUID()),
+    Column("span_id", UInt(64)),
+    Column("parent_span_id", UInt(64, Modifiers(codecs=["ZSTD(1)"]))),
+    Column("segment_id", UInt(64, Modifiers(codecs=["ZSTD(1)"]))),
+    Column("segment_name", String(Modifiers(codecs=["ZSTD(1)"]))),
+    Column("is_segment", UInt(8, Modifiers(codecs=["LZ4"]))),
+    Column("_sort_timestamp", DateTime(Modifiers(codecs=["DoubleDelta", "ZSTD(1)"]))),
```
evanh, line 34: Maybe call this `start_timestamp` and then we can have `start/end_timestamp_precise`? Unless we still think we might change what we use as the sort key?

phacops: I don't think we'll change what we use in the sort key but maybe, instead of rounding it to the second, we'll round it to something else?

I think it's still a good idea to keep indicating it's really only meant to be used as a sort timestamp and not really treat it as the real span timestamp.

evanh (approved):

PR #6443, opened 2024-10-22
feat: Add debug information support to API Endpoints

nachivrn:
This commit adds support for including debug information in the API Endpoints when requested. The changes include:

1. Creating a new `extract_response_meta` function in `snuba/web/rpc/common/debug_info.py` to generate debug information.
2. Updating `EndpointTraceItemTable.execute` in `snuba/web/rpc/v1/endpoint_trace_item_table.py` to use the new function and include debug info in the response.
3. Adding tests in `tests/web/rpc/test_debug_info.py` to verify the presence and absence of debug info based on the request.

These changes allow clients to receive detailed query information when debugging is enabled, improving troubleshooting capabilities.


nachivrn:
Updating the other API endpoints to include debug information.

nachivrn:
https://github.com/getsentry/sentry-protos/pull/51

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -107,9 +108,10 @@ def _build_query(request: TraceItemTableRequest) -> Query:
 
 def _build_snuba_request(
     request: TraceItemTableRequest,
+    request_id: str,
```
volokluev, line 111: this is in the request already no?

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -235,14 +237,19 @@ def _execute(self, in_msg: TraceItemTableRequest) -> TraceItemTableResponse:
         in_msg = _apply_labels_to_columns(in_msg)
         _validate_select_and_groupby(in_msg)
         _validate_order_by(in_msg)
-        snuba_request = _build_snuba_request(in_msg)
+        request_id = in_msg.meta.request_id or str(uuid.uuid4())
```
volokluev, line 240: I would just enrich the object itself rather than adding a special function parameter

tests/web/rpc/test_debug_info.py
```diff
@@ -0,0 +1,67 @@
+import uuid
```
volokluev, line 1: put this file in the `v1` folder

volokluev (commented):
looks good overall, just a few tweaks

snuba/web/rpc/common/debug_info.py
```diff
@@ -0,0 +1,63 @@
...
+
+def extract_response_meta(
+    request_id: str, debug: bool, query_results: List[QueryResult], timers: List[Timer]
+) -> ResponseMeta:
+    query_info = []
+
+    if debug:
+        for query_result, timer in zip(query_results, timers):
+            stats = query_result.extra.get("stats", {}) or {}
+            profile = query_result.result.get("profile", {}) or {}
```
onkar, line 23: Why is the `or {}` part needed?

snuba/web/rpc/common/debug_info.py
```diff
@@ -0,0 +1,63 @@
...
+            )
+
+            query_stats = QueryStats(
+                rows_read=stats.get("result_rows", 0),
+                columns_read=stats.get("result_cols", 0),
+                blocks=profile.get("blocks", 0),
+                progress_bytes=profile.get("progress_bytes", 0),
+                max_threads=stats.get("quota_allowance", {})
+                .get("summary", {})
+                .get("threads_used"),
```
onkar, line 40: Should the default value of `max_threads` be `0` (or a valid number) in case `stats.get("quota_allowance").get("summary").get("threads_used")` is not found / resolves to `None`? It looks like as is this will resolve to `None`.

nachivrn: In Protobuf, the handling of None and 0 for numeric fields differs from standard Python behavior. If nested get() calls result in None, the max_threads field in the Protobuf message is automatically set to 0.

snuba/web/rpc/common/debug_info.py
nachivrn: Mypy infers that the result might be `None`. By adding `or {}`, you're ensuring that the result is always a dictionary. Could you clarify the change you're requesting here? I'm not sure I understand the intent behind your question. Do you have an alternative solution?

volokluev (approved):

snuba/web/rpc/common/debug_info.py
```diff
@@ -0,0 +1,65 @@
...
+    ResponseMeta,
+    TimingMarks,
+)
+
+from snuba.utils.metrics.timer import Timer
+from snuba.web import QueryResult
+
+
+def extract_response_meta(
+    request_id: str, debug: bool, query_results: List[QueryResult], timers: List[Timer]
```
evanh, line 16: Why are `query_results` and `timers` lists? When would there be more than one of each?

evanh (approved):

snuba/web/rpc/common/debug_info.py
nachivrn: Composite query

snuba/web/rpc/common/debug_info.py
onkar: I thought `query_result.extra.get("stats", {})` should automatically handle the default dictionary case. Was curious to know why the extra `or` is needed.

onkar (approved):

PR #6445, opened 2024-10-22
feat(job-runner): reorganize tests

xurui-c:
Moved code around so we have `test_job_statuses.py` and `test_job_lock.py`


onkar (approved):
LGTM, thanks for addressing the follow-up!

PR #6447, opened 2024-10-23
feat(job-runner): record job's start timestamp to support timeouts

xurui-c:
store a record of when a job started executing so that we can support timeouts

### DIFF
--- a/snuba/manual_jobs/runner.py
+++ b/snuba/manual_jobs/runner.py
@@ -124,17 +124,14 @@ def run_job(job_spec: JobSpec) -> JobStatus:
 
     job_to_run = _JobLoader.get_job_instance(job_spec)
 
+    running_status_type = (
+        JobStatus.ASYNC_RUNNING_BACKGROUND if job_spec.is_async else JobStatus.RUNNING
+    )
     try:
-        if job_spec.is_async:
-            current_job_status = _set_job_status(
-                job_spec.job_id, JobStatus.ASYNC_RUNNING_BACKGROUND
-            )
-            _record_start_time(job_spec.job_id)
-            job_to_run.execute(job_logger)
-        else:
-            current_job_status = _set_job_status(job_spec.job_id, JobStatus.RUNNING)
-            _record_start_time(job_spec.job_id)
-            job_to_run.execute(job_logger)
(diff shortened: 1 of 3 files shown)


snuba/manual_jobs/runner.py
```diff
@@ -128,9 +129,11 @@ def run_job(job_spec: JobSpec) -> JobStatus:
...
                 job_spec.job_id, JobStatus.ASYNC_RUNNING_BACKGROUND
             )
+            _record_start_time(job_spec.job_id)
             job_to_run.execute(job_logger)
         else:
             current_job_status = _set_job_status(job_spec.job_id, JobStatus.RUNNING)
+            _record_start_time(job_spec.job_id)
             job_to_run.execute(job_logger)
             current_job_status = _set_job_status(job_spec.job_id, JobStatus.FINISHED)
             job_logger.info("[runner] job execution finished")
```
onkar, line 139: We are doing pretty much the same things in the `if` and `else` blocks here. The difference is just the args passed. Can this be simplified?:
```
job_running = JobStatus.ASYNC_RUNNING_BACKGROUND if job_spec.is_async else JobStatus.RUNNING
try:
    current_job_status = _set_job_status(job_spec.job_id, job_running)
    _record_start_time(job_spec.job_id)
    job_to_run.execute(job_logger)
    if not job_spec.is_async:
        # Finish current job if no exception is raised and it is synchronous.
        job_logger.info("[runner] job execution finished")
except BaseException:
    ...
```

onkar (approved):
LGTM

snuba/manual_jobs/runner.py
xurui-c: This approach is much cleaner. Thank you Onkar!

kylemumma (approved):
nice

tests/manual_jobs/test_record_job_start.py
```diff
@@ -0,0 +1,23 @@
+from unittest.mock import patch
```
onewland, line 1: In the future, I think a method this narrow doesn't need its own test file, but it's fine for now

onewland (approved):

PR #6450, opened 2024-10-24
ref(querylog): Produce request ID consistently

untitaker:
There are two ways to stringify a UUID: `str(uuid)`, and `uuid.hex` -- they produce different output, the first one is hyphenated, the second one isn't.

It seems we use both at the same time, and the consumer can only deal with the latter.

But also, sometimes we do this: `str(uuid.uuid4)`

That always produces something like `'<function uuid4 at 0x100c39620>'`,
which is total garbage but since everything is string-typed, mypy
doesn't complain.

Refactor everything to use UUID types and only convert to string on
serialization.

### DIFF
--- a/snuba/web/rpc/v1/endpoint_trace_item_table.py
+++ b/snuba/web/rpc/v1/endpoint_trace_item_table.py
@@ -108,7 +108,7 @@ def _build_query(request: TraceItemTableRequest) -> Query:
 
 def _build_snuba_request(request: TraceItemTableRequest) -> SnubaRequest:
     return SnubaRequest(
-        id=request.meta.request_id,
+        id=uuid.UUID(request.meta.request_id),
         original_body=MessageToDict(request),
         query=_build_query(request),
         query_settings=HTTPQuerySettings(),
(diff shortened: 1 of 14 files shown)


evanh (approved):

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -108,7 +108,7 @@ def _build_query(request: TraceItemTableRequest) -> Query:
 
 def _build_snuba_request(request: TraceItemTableRequest) -> SnubaRequest:
     return SnubaRequest(
-        id=request.meta.request_id,
+        id=uuid.UUID(request.meta.request_id),
```
nachivrn, line 111: Why aren't we using the request_id provided by the client in the request? If the client supplies a request_id, we should use it to track the request from start to finish.
Additionally, why do we want to change the id to be of type UUID instead of a str? Are we planning to use it for comparisons or other library functions?

volokluev (approved):

snuba/web/rpc/v1/endpoint_trace_item_table.py
nachivrn: I initially thought we were not using request_id sent from the client. 
it looks good.

onkar (approved):

PR #6456, opened 2024-10-25
Enhance query tracing and debug information on RPC Endpoints

nachivrn:
These changes provide more detailed query profiling and debugging capabilities, allowing for better performance analysis and troubleshooting of RPC Endpoints.

- Update sentry-protos to version 0.1.30
- Optimize Clickhouse settings handling for trace capture
- Extend ResponseMeta extraction to include clickhouse trace logs
- Add support for enabling trace logs in TraceItemTable endpoint


snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -107,11 +107,18 @@ def _build_query(request: TraceItemTableRequest) -> Query:
 
 
 def _build_snuba_request(request: TraceItemTableRequest) -> SnubaRequest:
+    query_settings = HTTPQuerySettings()
+
+    if request.meta.enable_trace_logs:
+        query_settings.set_clickhouse_settings(
+            {"send_logs_level": "trace", "log_profile_events": 1}
+        )
+
```
nachivrn, line 116: Refactoring: To implement this functionality across all RPC endpoints, I will create a common method or function that sets up the query settings with trace logs enabled.

nachivrn:
https://github.com/getsentry/sentry-protos/pull/54

snuba/clickhouse/native.py
```diff
@@ -166,12 +166,13 @@ def execute(
...
+                        capture_trace or settings.get("send_logs_level") == "trace"
+                    )
+
                     if capture_trace:
-                        settings = (
-                            {**settings, "send_logs_level": "trace"}
-                            if settings
-                            else {"send_logs_level": "trace"}
-                        )
+                        settings["send_logs_level"] = "trace"
```
onkar, line 175: Can you elaborate on the intent here? If I read this correctly, key `"send_logs_level"` is set to value `"trace"` either when the function argument `capture_trace` is `True` or `settings` already had key `"send_logs_level"` with value `"trace"`. This is confusing. It appears that the only time you want `settings["send_logs_level"] = "trace"` is when `capture_trace` is `True`. So why not just say -
```
if capture_trace:
    settings["send_logs_level"] = "trace"
```

nachivrn: The capture_trace argument is not directly accessible at the RPC Endpoint, as the request must traverse the entire query pipeline before reaching the execute method. We use query settings to propagate trace-capturing information through the pipeline to address this limitation.

To fully understand the rationale behind this implementation, it’s essential to consider the entire lifecycle of an API invocation through the query pipeline. This process includes multiple stages of query processing and optimization, culminating in the execution phase.

I recommend reviewing this flow to fully grasp the context and necessity of the current logic. While your proposed simplification would work if the execute method were directly invoked with capture_trace set to True, it will not work in other case because the execute method in the ClickhousePool class are not directly available from the RPC endpoint. The RPC endpoint and the Clickhouse execution are at different layers of the application.

evanh (approved):

PR #6463, opened 2024-10-28
fix(admin): Allow OPTIMIZE queries in sudo mode

evanh:
This allows OPTIMIZE queries to be run in the snuba admin when using sudo
mode. This can be useful to potentially resolve tables/partitions in bad
states.

### DIFF
--- a/tests/admin/test_system_queries.py
+++ b/tests/admin/test_system_queries.py
@@ -8,6 +8,7 @@ from snuba.admin.clickhouse.common import InvalidCustomQuery
 from snuba.admin.clickhouse.system_queries import (
     UnauthorizedForSudo,
     is_query_alter,
+    is_query_optimize,
     is_query_select,
     is_system_command,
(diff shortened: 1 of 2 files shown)


tests/admin/test_system_queries.py
```diff
@@ -86,10 +87,15 @@ def test_is_query_select() -> None:
         ("system KILL", False),
         ("ALTER TABLE eap_spans_local_merge DROP PARTITION '1970-01-01'", True),
         ("CREATE TABLE eap_spans_local_merge (all my fieds)", True),
+        ("OPTIMIZE TABLE eap_spans_local", True),
```
nachivrn, line 90: Niit: Should we consider adding test case for 
Multiple spaces between words ("OPTIMIZE    TABLE    my_table")
Mixed case variations ("optimize table my_table")

nachivrn (approved):

PR #6466, opened 2024-10-29
chore(deps): bump braces from 3.0.2 to 3.0.3 in /snuba/admin

evanh:
Bumps braces from 3.0.2 to 3.0.3.

https://github.com/getsentry/snuba/security/dependabot/53

### DIFF
--- a/snuba/admin/package.json
+++ b/snuba/admin/package.json
@@ -26,6 +26,7 @@
     "@tiptap/starter-kit": "^2.0.3",
     "@types/react": "^18.0.20",
     "@types/react-dom": "^18.2.6",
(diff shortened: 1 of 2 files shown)


nachivrn (commented):
Question: Is the yarn.lock file changes need to be part of git

nachivrn (approved):

evanh:
> Question: Is the yarn.lock file changes need to be part of git

Yes, that's what control the frontend dependencies. Normally when you update a package, you run yarn add package@version, and that will update the yarn.lock file. Then the deployment uses the yarn file to download the appropriate packages.

PR #6469, opened 2024-10-29
chore(deps): bump openssl from 0.10.64 to 0.10.68 in /rust_snuba

evanh:
https://github.com/getsentry/snuba/security/dependabot/58

### DIFF
--- a/rust_snuba/Cargo.lock
+++ b/rust_snuba/Cargo.lock
@@ -2191,9 +2191,9 @@ checksum = "0ab1bc2a289d34bd04a330323ac98a1b4bc82c9d9fcb1e66b63caa84da26b575"
 
 [[package]]
 name = "openssl"
-version = "0.10.64"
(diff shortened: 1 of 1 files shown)


rust_snuba/Cargo.lock
```diff
@@ -2191,9 +2191,9 @@ checksum = "0ab1bc2a289d34bd04a330323ac98a1b4bc82c9d9fcb1e66b63caa84da26b575"
 
 [[package]]
 name = "openssl"
-version = "0.10.64"
+version = "0.10.68"
```
onkar, line 2194: This version is 0.10.68 while the commit message says 0.10.64. Is this intended?

onkar (approved):

rust_snuba/Cargo.lock
evanh: Fixed

PR #6475, opened 2024-10-29
feat(eap): Timeseries V1 RPC

volokluev:
This PR implements the basics of the timeseries API.

Supported:

* Aggregations on all attributes
* filters
* groupby
* zerofilling nonexistent data

To come in future PRs:

* Sample count
* extrapolation

### DIFF
--- a/snuba/web/rpc/v1/endpoint_time_series.py
+++ b/snuba/web/rpc/v1/endpoint_time_series.py
@@ -1,0 +1,24 @@
+from typing import Type
+from snuba.web.rpc import RPCEndpoint
+from sentry_protos.snuba.v1.endpoint_time_series_pb2 import TimeSeriesRequest, TimeSeriesResponse, DataPoint
+from sentry_protos.snuba.v1.request_common_pb2 import RequestMeta
+
+
+
+
+class EndpointTimeSeries(RPCEndpoint[TimeSeriesRequest, TimeSeriesResponse]):
+
+    @classmethod
+    def version(cls):
+        return "v1"
+
+    @classmethod
+    def request_class(cls) -> Type[TimeSeriesRequest]:
+        return TimeSeriesRequest
+
+    @classmethod
+    def response_class(cls) -> Type[TimeSeriesResponse]:
+        return TimeSeriesResponse
+
+    def _execute(self, in_msg: TimeSeriesRequest) -> TimeSeriesResponse:
+        raise NotImplementedError
--- a/snuba/web/rpc/v1/endpoint_time_series.py
+++ b/snuba/web/rpc/v1/endpoint_time_series.py
@@ -65,7 +65,6 @@ def _convert_result_timeseries(
                 result_timeseries[(group_by_key, col_name)] = TimeSeries(
                     group_by_attributes=group_by_map, label=col_name
                 )
-                # TODO: convert row time to timestamp
             result_timeseries_timestamp_to_row[(group_by_key, col_name)][
                 int(datetime.fromisoformat(row["time"]).timestamp())
             ] = row
--- a/snuba/web/rpc/v1/endpoint_time_series.py
+++ b/snuba/web/rpc/v1/endpoint_time_series.py
@@ -1,3 +1,4 @@
+import math
 import uuid
 from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -0,0 +1,203 @@
...
+    # create a mapping for each timeseries of timestamp: row to fill data points not returned in the query
+    result_timeseries_timestamp_to_row: defaultdict[
+        tuple[str, str], dict[int, Dict[str, Any]]
+    ] = defaultdict(dict)
+    query_duration = (
+        request.meta.end_timestamp.seconds - request.meta.start_timestamp.seconds
+    )
+    time_buckets = [
+        Timestamp(seconds=(request.meta.start_timestamp.seconds) + secs)
+        for secs in range(0, query_duration, request.granularity_secs)
```
evanh, line 61: What happens if the granularity doesn't line up with the duration? E.g. a granularity of 61 seconds with a 10 minute window?

Do we need extra validation to ensure that the result timestamps will line up with these generated buckets?

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -0,0 +1,203 @@
...
+    ]
+
+    for row in data:
+        group_by_map = {}
+
+        for col_name, col_value in row.items():
+            if col_name in group_by_labels:
+                group_by_map[col_name] = col_value
+
+        group_by_key = "|".join([f"{k},{v}" for k, v in group_by_map.items()])
```
evanh, line 71: This is a performance nit, but I might make this a tuple instead of a string.

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -0,0 +1,203 @@
...
+        SelectedExpression(
+            name=attr_key.name, expression=attribute_key_to_expression(attr_key)
+        )
+        for attr_key in request.group_by
+    ]
+
+    res = Query(
+        from_clause=entity,
+        selected_columns=[
+            SelectedExpression(name="time", expression=column("time", alias="time")),
```
evanh, line 125: I would rather not have things depend on the TimeSeriesProcessor, but that is a personal opinion. I've been trying to deprecate that processor for a while.

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -0,0 +1,203 @@
...
+    setup_trace_query_settings,
+)
+
+
+def _convert_result_timeseries(
+    request: TimeSeriesRequest, data: list[Dict[str, Any]]
+) -> Iterable[TimeSeries]:
+    # to convert the results, need to know which were the groupby columns and which ones
+    # were aggregations
+    aggregation_labels = set([agg.label for agg in request.aggregations])
```
evanh, line 46: Is it valid to have a request with duplicate labels?

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -0,0 +1,203 @@
...
+        for col_name in aggregation_labels:
+            if not result_timeseries.get((group_by_key, col_name), None):
+                result_timeseries[(group_by_key, col_name)] = TimeSeries(
+                    group_by_attributes=group_by_map,
+                    label=col_name,
+                    buckets=time_buckets,
+                )
+            result_timeseries_timestamp_to_row[(group_by_key, col_name)][
+                int(datetime.fromisoformat(row["time"]).timestamp())
+            ] = row
```
kylemumma, line 81: what is group_by_map and group_by_key? Im not sure what you're doing here.

kylemumma (approved):
I got a bit confused trying to understand `_convert_result_timeseries` but other than that it looks good

snuba/web/rpc/v1/endpoint_time_series.py
volokluev: good point, met me enforce that that does not happen

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -0,0 +1,203 @@
...
+    def request_class(cls) -> Type[TimeSeriesRequest]:
+        return TimeSeriesRequest
+
+    @classmethod
+    def response_class(cls) -> Type[TimeSeriesResponse]:
+        return TimeSeriesResponse
+
+    def _execute(self, in_msg: TimeSeriesRequest) -> TimeSeriesResponse:
+        # TODO: Move this to base
+        in_msg.meta.request_id = getattr(in_msg.meta, "request_id", None) or str(
```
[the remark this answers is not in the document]
volokluev: yeah pretty much

snuba/web/rpc/v1/endpoint_time_series.py
volokluev: I don't know how much faster that will actually make it as the tuple will still need to be hashed eventually

snuba/web/rpc/v1/endpoint_time_series.py
volokluev: I added an explanation, lemme know if it makes sense now

snuba/web/rpc/v1/endpoint_time_series.py
volokluev: This is done now

snuba/web/rpc/v1/endpoint_time_series.py
volokluev: Good point, I added some logic to handle this and tests for it as well

evanh (approved):
LGTM.

PR #6477, opened 2024-10-30
feat(eap): AttributeNames "autocomplete" v1 API

kylemumma:
This PR implements the endpoint for TraceItemAttributeNames. For context, please see [this diagram](https://www.notion.so/sentry/Qu-EAP-Querying-Events-Analytics-Platform-76629acc212b4ac486c63e7979f16076?pvs=4#66feb9128c2b4752a7921e47c0cb20fc) "attribute names" in green at the bottom. Also see [this ticket](https://github.com/getsentry/projects/issues/293).

This API allows you to query for the names of all attributes attached to spans for a given project and dates. The interface definition is here https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto#L12-L28

## implementation
Its implemented in the following way at a high-level
1. receive a request from the user
2. convert that request into a `SnubaRequest`. This is an existing object we use for snuba requests, and is directly compatible to go through the existing snuba query pipeline. So thats what we do with it once its converted.
3. Convert the `QueryResult` into the proper result. Once the query pipeline gives us our result we convert it to the proper format our users expect and return it.

### DIFF
--- a/snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
+++ b/snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
@@ -6,6 +6,7 @@ from sentry_protos.snuba.v1.endpoint_trace_item_attributes_pb2 import (
     TraceItemAttributeNamesRequest,
     TraceItemAttributeNamesResponse,
 )
+from sentry_protos.snuba.v1.request_common_pb2 import PageToken
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import AttributeKey
 
 from snuba.attribution.appid import AppID
@@ -15,6 +16,7 @@ from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.pluggable_dataset import PluggableDataset
 from snuba.query import OrderBy, OrderByDirection, SelectedExpression
 from snuba.query.data_source.simple import Entity
+from snuba.query.dsl import Functions as f
 from snuba.query.dsl import column
 from snuba.query.logical import Query
 from snuba.query.query_settings import HTTPQuerySettings
@@ -24,6 +26,7 @@ from snuba.web import QueryResult
 from snuba.web.query import run_query
 from snuba.web.rpc import RPCEndpoint
 from snuba.web.rpc.common.common import base_conditions_and, treeify_or_and_conditions
+from snuba.web.rpc.common.debug_info import extract_response_meta
 from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
 
 
@@ -51,7 +54,6 @@ def convert_to_snuba_request(req: TraceItemAttributeNamesRequest) -> SnubaReques
         raise BadSnubaRPCRequestException(f"Unknown attribute type: {req.type}")
 
     # truncate_request_meta_to_day(request.meta)
-    # TODO: value substring filter
     query = Query(
         from_clause=entity,
         selected_columns=[
@@ -60,7 +62,9 @@ def convert_to_snuba_request(req: TraceItemAttributeNamesRequest) -> SnubaReques
                 expression=column("attr_key"),
             ),
         ],
-        condition=base_conditions_and(req.meta),
+        condition=base_conditions_and(
+            req.meta, f.like(column("attr_key"), f"%{req.value_substring_match}%")
+        ),
         order_by=[
             OrderBy(
                 direction=OrderByDirection.ASC, expression=column("organization_id")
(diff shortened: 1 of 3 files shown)


snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
```diff
@@ -0,0 +1,137 @@
...
+                direction=OrderByDirection.ASC, expression=column("organization_id")
+            ),
+            OrderBy(direction=OrderByDirection.ASC, expression=column("attr_key")),
+        ],
+        groupby=[
+            column("organization_id", alias="organization_id"),
+            column("attr_key", alias="attr_key"),
+        ],
+        limit=req.limit,
+        offset=req.offset,
```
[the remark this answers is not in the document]
kylemumma: https://github.com/getsentry/sentry-protos/pull/57/

snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
```diff
@@ -0,0 +1,165 @@
...
+from snuba.query.query_settings import HTTPQuerySettings
+from snuba.reader import Row
+from snuba.request import Request as SnubaRequest
+from snuba.web import QueryResult
+from snuba.web.query import run_query
+from snuba.web.rpc import RPCEndpoint
+from snuba.web.rpc.common.common import base_conditions_and, treeify_or_and_conditions
+from snuba.web.rpc.common.debug_info import extract_response_meta
+from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
+
```
nachivrn, line 31: Do we need to have a logger to capture

1. Debug-level information about incoming requests
2. Warning-level logs for limit violations
3. Info-level logs for request processing
4. Error-level logs for query failures
5. Debug-level logs for successful query completion

snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
```diff
@@ -0,0 +1,165 @@
...
+
+def convert_to_snuba_request(req: TraceItemAttributeNamesRequest) -> SnubaRequest:
+    if req.limit > 1000:
```
nachivrn, line 34: Should we consider extracting into named constants like MAX_LIMIT = 1000

nachivrn (commented):
Overall, looks good with test coverage. 

I have a question.  Do we need to add validation that would help prevent performance issues that could arise from very short search strings (like "a" or "b")

tests/web/rpc/v1/test_endpoint_trace_item_attribute_names.py
```diff
@@ -0,0 +1,342 @@
...
+                ),
+            ),
+            limit=1000,
+            type=AttributeKey.Type.TYPE_STRING,
+            value_substring_match="",
+        )
+        res = EndpointTraceItemAttributeNames().execute(req)
+
+        expected = []
+        for i in range(30):
```
volokluev, line 100: where is this magic number coming from?

volokluev (commented):
looks good overall, just some questions

snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
kylemumma: done

snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
```diff
@@ -0,0 +1,165 @@
...
+        AttributeKey.Type.TYPE_INT,
+        AttributeKey.Type.TYPE_BOOLEAN,
+    ):
+        entity = Entity(
+            key=EntityKey("spans_num_attrs"),
+            schema=get_entity(EntityKey("spans_num_attrs")).get_data_model(),
+            sample=None,
+        )
+    else:
+        raise BadSnubaRPCRequestException(f"Unknown attribute type: {req.type}")
```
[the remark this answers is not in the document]
kylemumma: done

tests/web/rpc/v1/test_endpoint_trace_item_attribute_names.py
kylemumma: I introduced some constants like `TOTAL_GENERATED_SPANS` to get rid of the magic numbers

kylemumma (commented):
Updated based on PR feedback.
* There was concern over potential performance implication of very short search strings (like "a" or "b"), and it was suggested to set a min length for them. For now we will not do this, and come back to it if any issues arise in the future.
* I will create a follow up PR to implement page_token.filter_offset
* I will explore the idea of adding logging or observability metrics to the endpoint as a follow up

volokluev (approved):

nachivrn (approved):

PR #6481, opened 2024-10-30
fix(EAP): fix bug with unassigned labels

volokluev:
protobuf sets the label as an empty string by default which is truthy in python. hence we were not actually bubbling up labels correctly. this fixes the bug

### DIFF
--- a/tests/web/rpc/v1/test_endpoint_trace_item_table.py
+++ b/tests/web/rpc/v1/test_endpoint_trace_item_table.py
@@ -628,7 +628,7 @@ class TestTraceItemTable(BaseApiTest):
         with pytest.raises(BadSnubaRPCRequestException):
             EndpointTraceItemTable().execute(message)
 
-    def test_order_by_aggregation(self, setup_teardown) -> None:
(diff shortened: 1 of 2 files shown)


tests/web/rpc/v1/test_endpoint_trace_item_table.py
```diff
@@ -717,3 +729,25 @@ def test_aggregation_on_attribute_column(self) -> None:
         response = EndpointTraceItemTable().execute(message)
         measurement_avg = [v.val_float for v in response.column_values[0].results][0]
         assert measurement_avg == 420
+
+
+class TestUtils:
+    def test_apply_labels_to_columns(self) -> None:
+        message = TraceItemTableRequest(
+            columns=[
```
nachivrn, line 737: Should there be test that will verify how the _apply_labels_to_columns function handles both empty string and None value cases for labels?

volokluev: Done.

None is technically not possible in protobuf but in python land it is. I check for both now

nachivrn (approved):

PR #6485, opened 2024-10-31
feat(snuba-admin): Use AST and Query Tree generated by EXPLAIN query for system query validation

davidtsuk:
Instead of using regex for validating system queries, we can use AST/Query Tree output from CH EXPLAIN to check for dangerous queries. This approach is more maintainable and in the future when the new CH analyzer is no longer experimental, we can switch fully to using the Query Tree.

### DIFF
--- a/snuba/admin/clickhouse/system_queries.py
+++ b/snuba/admin/clickhouse/system_queries.py
@@ -49,23 +49,6 @@ def _run_sql_query_on_host(
     return query_result
 
 
-SYSTEM_QUERY_RE = re.compile(
-    r"""
-        ^ # Start
-        (SELECT|select)
-        \s
-        (?P<select_statement>[\w\s\',()*+\-\/:]+|\*)
-        \s
-        (FROM|from)
-        \s
-        system.[a-z_]+
-        (?P<extra>\s[\w\s,=()*+<>'%"\-\/:]+)?
-        ;? # Optional semicolon
-        $ # End
-    """,
-    re.VERBOSE,
-)
-
 DESCRIBE_QUERY_RE = re.compile(
     r"""
         ^ # Start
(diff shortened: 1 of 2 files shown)


snuba/admin/clickhouse/system_queries.py
```diff
@@ -129,13 +112,62 @@ def _run_sql_query_on_host(
...
+    """
+    Validation based on Query Tree and AST to ensure the query is a valid select query.
+    """
+    explain_ast_query = f"EXPLAIN AST {sql_query}"
+    disallowed_ast_nodes = ["AlterQuery", "AlterCommand", "DropQuery", "InsertQuery"]
+    explain_ast_result = _run_sql_query_on_host(
+        clickhouse_host, clickhouse_port, storage_name, explain_ast_query, False
+    )
+
+    print(explain_ast_result.results)
```
nachivrn, line 160: Debug print statement can be removed

snuba/admin/clickhouse/system_queries.py
```diff
@@ -200,19 +232,30 @@ def run_system_query_on_host_with_sql(
...
+        clickhouse_host: str,
+        clickhouse_port: int,
+        storage_name: str,
+        system_query_sql: str,
+        sudo_mode: bool,
+    ) -> bool:
+        if not sudo_mode and is_valid_system_query(
+            clickhouse_host, clickhouse_port, storage_name, system_query_sql
+        ):
+            return True
```
nachivrn, line 245: Should we return specific error messages from validation functions. Raising InvalidCustomQuery with detailed error messages

nachivrn (commented):
Overall, this is a solid improvement to the codebase, replacing regex-based validation with a more robust AST-based approach. Thanks David!

snuba/admin/clickhouse/system_queries.py
davidtsuk: I think we should do that, but it's kind of hard to have detailed error messages when we don't know what kind of query the user is trying to write.

tests/admin/test_system_queries.py
```diff
@@ -98,8 +125,6 @@ def test_sudo_queries(sudo_query: str, expected: bool) -> None:
         or is_query_alter(sudo_query)
         or is_query_optimize(sudo_query)
     ) == expected
```
kylemumma, line 127: should you be passing this into `is_valid_query` as well to ensure it gives the correct output?

snuba/admin/clickhouse/system_queries.py
kylemumma: I agree that this would be nice to have. Then if users enter an invalid query they could quickly understand why its invalid.

snuba/admin/clickhouse/system_queries.py
```diff
@@ -129,13 +112,62 @@ def _run_sql_query_on_host(
...
+
+
+def is_valid_system_query(
+    clickhouse_host: str, clickhouse_port: int, storage_name: str, sql_query: str
+) -> bool:
+    """
+    Validation based on Query Tree and AST to ensure the query is a valid select query.
+    """
+    explain_ast_query = f"EXPLAIN AST {sql_query}"
+    disallowed_ast_nodes = ["AlterQuery", "AlterCommand", "DropQuery", "InsertQuery"]
```
kylemumma, line 155: Would it be better to use allowed nodes rather than disallowed nodes? I worry if a new node type is ever made it could be an issue.

kylemumma (approved):
Good PR definitely an improvement like Nachi said

snuba/admin/clickhouse/system_queries.py
davidtsuk: This disallowed node check is just an extra layer of protection, we already use a readonly mode when querying clickhouse which provides the main level of protection. If I made a list of allowed nodes, it would be much longer.

snuba/admin/clickhouse/system_queries.py
```diff
@@ -129,13 +112,60 @@ def _run_sql_query_on_host(
...
+) -> bool:
     """
-    Simple validation to ensure query is a select command
+    Run the EXPLAIN QUERY TREE on the given sql_query and check that the only tables
+    in the query are system tables.
     """
-    sql_query = " ".join(sql_query.split())
-    match = SYSTEM_QUERY_RE.match(sql_query)
-    return True if match else False
+    sql_query = sql_query[:-1] if sql_query.endswith(";") else sql_query
```
nachivrn, line 125: Nit: Consider using sql_query.strip().rstrip(';') to handle this more robustly to  removes any leading or trailing whitespace.

PR #6486, opened 2024-10-31
feat(eap-spans): Respect the decision to ingest or not in EAP

phacops:
https://github.com/getsentry/sentry/pull/80064
https://github.com/getsentry/relay/pull/4206


volokluev (approved):

volokluev:
would you need to update `sentry_kafka_schemas`?

phacops:
> would you need to update `sentry_kafka_schemas`?

We allow additional fields, and given this is a temporary one, mostly for the LA/GA, I'm tempted not to change the schema.

PR #6488, opened 2024-11-01
fix: Fix tagging with new SDK version

evanh:
The new version of the sentry-sdk checks for existing spans in a different way.
Fix this so that data is still tagged correctly.


volokluev (approved):
Thank you for the diligence

PR #6490, opened 2024-11-01
fix(vscode): update deprecated vscode workspace settings

kylemumma:
Changes
1. The following vscode settings
```
"python.linting.mypyEnabled": true,
"python.linting.flake8Enabled": true,
"python.formatting.provider": "black",
```
have been deprecated in favor of using the extensions, [see here](https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions#linting-and-formatting-settings-deprecation). I don't think they even do anything anymore. So I removed them and added their corresponding extensions to the workspace extension recommendations.

2. `python.pythonPath` settings has been deprecated and I'm pretty sure it doesnt do anything anymore. `python.defaultInterpreterPath` should be used instead so I replaced it with that. See here https://devblogs.microsoft.com/python/python-in-visual-studio-code-july-2021-release/#selecting-a-python-interpreter-no-longer-modifies-workspace-settings

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,7 @@
+{
+    "recommendations": [
+        "ms-python.black-formatter",
+        "ms-python.mypy-type-checker",
+        "ms-python.flake8"
+    ]
+}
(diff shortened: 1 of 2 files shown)


davidtsuk (approved):
LGTM

PR #6493, opened 2024-11-01
fix(eap): fix table aliasing

volokluev:
Using the wrong value for selected expression name, also aggregations sometimes don't have a label, we can just inherit the column name


wmak (approved):
Thanks!

phacops (approved):

PR #6494, opened 2024-11-01
feat(devservices): Remove devservices from requirements.txt

hubertdeng123:
Now that we have a clear update command for devservices that bumps the version in place, let's get rid of this. It removes ambiguity between using venv installs and global binary installs through PATH.

Installation instructions here:
https://github.com/getsentry/devservices?tab=readme-ov-file#installation

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,6 @@ click==8.1.3
 clickhouse-driver==0.2.6
 confluent-kafka==2.3.0
 datadog==0.21.0
-devservices==0.0.4
 flake8==5.0.4
 Flask==2.2.5
 google-cloud-storage==2.18.0


evanh:
@hubertdeng123 Can you explain this change more? The README you linked doesn't provide an update in place command, and suggests we have this dependency in our requirements.

IanWoodard (approved):

hubertdeng123:
@evanh Thanks for catching, the readme is now updated to include the `devservices update` command. We wanted to discourage people from installing the tool through requirements.txt since it is too reliant on a python venv. We want the devservices tool to be used across any repo, even when the codebase is not written in python.

hubertdeng123:
Talked as a team, going to close this in favor of supporting both global and virtual environment installs

PR #6505, opened 2024-11-04
Consolidate migrations label into labeler workflow

onkar:
Even after merging https://github.com/getsentry/snuba/pull/6504, the labeler stage is failing when dependabot makes the changes. This is because we had two `labeler.yml` files. The deleted file just adds a label `migrations` when some specific files are touched. This PR combines that functionality into existing labeler workflow so that we have one `labeler.yml` file.


onkar:
Will open a new PR

PR #6506, opened 2024-11-04
Fix labeler files to make v5 work correctly

onkar:
### Legal Boilerplate


volokluev (approved):

onkar:
Closing this because moving the labeler to v5 is kind of a mess. https://github.com/actions/labeler says that it has introduced breaking changes and v5 is not backward compatible. Additionally, the documentation is not clear on what exactly needs to be done to upgrade labeler to v5.

PR #6510, opened 2024-11-05
ref(deletes): bulk delete consumer

MeredithAnya:
**context**:
The work for bulk deleting in snuba has so far included the following:
* Added the kafka schemas - https://github.com/getsentry/sentry-kafka-schemas/pull/347
* Adding the new endpoint including producing to the topic - https://github.com/getsentry/snuba/pull/6440
* Created the topic in production for all env - https://github.com/getsentry/ops/pull/12711

**what's left**:
Now that we have the topics created we can finish up the consumer side.
- [ ] Add the consumer logic to snuba (This PR)
- [ ] Add the consumer deployment to S4S region in the ops repository
- [ ] Set up datadog alerts/metrics and other observability

**this PR**:
It's a bit of a larger PR but it can be reviewed in a couple sections:
* The main consumer logic and strategy
    * This has the logic to create the strategy factory for the consumer and composes the strategy steps. The `strategy.py` file has the details for actually executing the delete query
* The `batching.py` file - I have an [arroyo PR](https://github.com/getsentry/arroyo/pull/390) that makes this file obsolete but in the meantime I don't think it needs to block this PR since it will be easy to remove after
* The formatters are going to be the only code that someone will need to write in the future when deploying the deletions consumer for a different storage. How one formats the conditions for the `DELETE` query is up to that logic.


MeredithAnya:
For the batching file I added, I've also added it into arroyo itself: https://github.com/getsentry/arroyo/pull/390 though I don't think that should be a blocker for this PR. Just was something I thought would clean up the duplicate work - if folks think it makes sense to add to arroyo

snuba/cli/lw_deletions_consumer.py
```diff
@@ -0,0 +1,174 @@
...
+DEFAULT_DELETIONS_MAX_BATCH_TIME_MS = 60000 * 2
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.option(
+    "--consumer-group",
+    help="Consumer group use for consuming the deletion topic.",
+    default="lw-deletions-consumer",
```
MeredithAnya, line 39: hmm I don't think a default makes sense actually

snuba/lw_deletions/formatters.py
```diff
@@ -0,0 +1,82 @@
...
+
+        and_conditions: list[ConditionsType] = []
+        for and_condition in mapping.values():
+            and_conditions.append(
+                {
+                    "project_id": [and_condition.project_id],
+                    "group_id": and_condition.group_ids,
+                }
+            )
+        return and_conditions
```
onkar, line 77: Could use list comprehension here to make it more pythonic.

snuba/lw_deletions/strategy.py
```diff
@@ -0,0 +1,158 @@
...
+        self.__next_step = next_step
+        self.__storage = storage
+        self.__cluster_name = self.__storage.get_cluster().get_clickhouse_cluster_name()
+        self.__tables = storage.get_deletion_settings().tables
+        self.__formatter: Formatter = formatter
+
+    def poll(self) -> None:
+        self.__next_step.poll()
+
+    def submit(self, message: Message[ValuesBatch[KafkaPayload]]) -> None:
```
onkar, line 53: Should we log something in this function?

snuba/lw_deletions/strategy.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+
+    def _execute_delete(self, conditions: Sequence[ConditionsType]) -> None:
+        self._check_ongoing_mutations()
+        query_settings = HTTPQuerySettings()
+        for table in self.__tables:
+            query = construct_query(
+                self.__storage, table, construct_or_conditions(conditions)
+            )
+            print("query", query)
```
onkar, line 83: Stray debug log?

snuba/lw_deletions/strategy.py
```diff
@@ -0,0 +1,158 @@
...
+    def submit(self, message: Message[ValuesBatch[KafkaPayload]]) -> None:
+        decode_messages = [
+            rapidjson.loads(m.payload.value) for m in message.value.payload
+        ]
+        conditions = self.__formatter.format(decode_messages)
+
+        try:
+            self._execute_delete(conditions)
+        except TooManyOngoingMutationsError:
+            raise MessageRejected
```
onkar, line 62: What is the intended behavior when too many mutations are ongoing and a message is rejected? Probably better to clarify this behavior in a comment.

onkar (commented):
Looks good overall. Left a few comments and questions.

snuba/lw_deletions/strategy.py
MeredithAnya: @onkar raising `MessageRejected` is the way in which we pause the consumer when we want to apply back pressure. Comes from https://github.com/getsentry/arroyo/blob/a8545c79bf81d78ff432addf9c8ad737b81082a7/arroyo/processing/processor.py#L417

But I can add a comment clarifying that this will cause back pressure

snuba/lw_deletions/strategy.py
MeredithAnya: oops yes

snuba/lw_deletions/batching.py
```diff
@@ -0,0 +1,157 @@
...
+    def append(self, message: BaseValue[TPayload]) -> None:
+        """
+        Instead of increasing the buffer size based on the number
+        of messages, we use the `rows_to_delete` attribute in the
+        message payload so we can batch by the number of rows we
+        want to delete.
+        """
+        self._buffer = self.accumulator(self._buffer, message)
+        if self.increment_by:
+            buffer_increment = self.increment_by(message)
```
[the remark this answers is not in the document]
MeredithAnya: the `increment_by` is a function that returns an integer. and that function can do anything based on the message. The function we used is created here: https://github.com/getsentry/snuba/pull/6510/files#diff-9f2e56ee48e901305aa0d40ccb0a705f36a0c7ab07f032fd6360046a965310c5R121 and then passed to the strategy

snuba/lw_deletions/batching.py
```diff
@@ -0,0 +1,157 @@
...
+    ):
+        self.accumulator = accumulator
+        self.initial_value = initial_value
+        self.max_batch_size = max_batch_size
+        self.max_batch_time = max_batch_time
+        self.increment_by = increment_by
+
+        self._buffer = initial_value()
+        self._buffer_size = 0
+        self._buffer_until = time.time() + max_batch_time
```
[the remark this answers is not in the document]
MeredithAnya: This is copied from the arroyo code - and ultimately will be deleting this file entirely so Im just going to keep as is since it's consistent with arroyo

snuba/cli/lw_deletions_consumer.py
```diff
@@ -0,0 +1,174 @@
...
+    "--auto-offset-reset",
+    default="earliest",
+    type=click.Choice(["error", "earliest", "latest"]),
+    help="Kafka consumer auto offset reset.",
+)
+@click.option(
+    "--no-strict-offset-reset",
+    is_flag=True,
+    help="Forces the kafka consumer auto offset reset.",
+)
```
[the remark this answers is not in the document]
MeredithAnya: This is an arroyo specific configuration, more details in the consumer doc strings https://github.com/getsentry/arroyo/blob/a8545c79bf81d78ff432addf9c8ad737b81082a7/arroyo/backends/kafka/consumer.py#L169 but basically if this is set to False, it means we pass True to the `strict_offset_reset` to arroyo and if we do that, then the auto offset reset is set to `"error"` https://github.com/getsentry/arroyo/blob/a8545c79bf81d78ff432addf9c8ad737b81082a7/arroyo/backends/kafka/consumer.py#L209

snuba/cli/lw_deletions_consumer.py
```diff
@@ -0,0 +1,174 @@
...
+            kafka_params=KafkaParameters(
+                group_id=consumer_group,
+                auto_offset_reset=auto_offset_reset,
+                strict_offset_reset=not no_strict_offset_reset,
+                queued_max_messages_kbytes=queued_max_messages_kbytes,
+                queued_min_messages=queued_min_messages,
+            ),
+            processing_params=ProcessingParameters(None, None, None),
+            max_batch_size=max_rows_batch_size,
+            max_batch_time_ms=max_batch_time_ms,
```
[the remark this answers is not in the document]
MeredithAnya: So yeah this is a little confusing but it's due to historical reason of how the consumer builder works (it gets passed in `build_streaming_strategy_factory`) . I thought about writing my own consumer builder just for deletes but that felt like re-doing a lot of the work already done in the consumer builder. 

I think we could definitely improve the consumer builder code but I didn't want to overhaul that in my PR.

snuba/consumers/consumer_builder.py
```diff
@@ -375,6 +375,15 @@ def build_dlq_consumer(
             dlq_policy,
         )
 
+    def build_lw_deletions_consumer(
+        self, strategy_factory: ProcessingStrategyFactory[KafkaPayload]
+    ) -> StreamProcessor[KafkaPayload]:
+        return self.__build_consumer(
+            strategy_factory,
+            self.raw_topic,
+            self.__build_default_dlq_policy(),
```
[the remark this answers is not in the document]
MeredithAnya: Since we don't pass in a `dlq_topic` into the consumer config there is now `dlq_policy` and this doesn't do anything. we can add a dlq later if we want

snuba/cli/lw_deletions_consumer.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if consumer is not None:
+            consumer.signal_shutdown()
+
+    signal.signal(signal.SIGINT, handler)
+    signal.signal(signal.SIGTERM, handler)
+
+    topic = STORAGE_TOPIC[storage_name]
+
+    while not shutdown_requested:
```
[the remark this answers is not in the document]
MeredithAnya: There is logging in the arroyo stream processor, and the rest of the metrics will come from the specific strategy steps that we've added in this PR. We can always add more logging if there is something specific we don't get from these initial logs/metrics

snuba/lw_deletions/formatters.py
MeredithAnya: updated the formatter to be simpler

snuba/lw_deletions/formatters.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@dataclass
+class SearchIssueCondition:
+    project_id: int
+    group_ids: list[int] = []
+
+    def add_group_ids(self, group_ids: Sequence[int]) -> None:
+        for group_id in group_ids:
+            self.group_ids.append(group_id)
```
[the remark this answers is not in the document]
MeredithAnya: updated the formatter to be simpler

onkar (approved):

PR #6511, opened 2024-11-05
ref(deletes): add ConditionsType

MeredithAnya:
Adds better typing to the delete pipeline as well separates some code out from https://github.com/getsentry/snuba/pull/6510. To avoid circular imports, it's easier to have `bulk_delete_query` file almost import from `delete_query`, and never have `delete_query` import from `bulk_delete_query` (where the ConditionsType was before)


kylemumma (approved):
nice

PR #6512, opened 2024-11-05
fix(job-runner): handle the case of no job runs

xurui-c:
If there are no jobs, we don't need to fetch statuses from redis. This fixes the bug where `_redis_client.mget` errors because the keys are empty


snuba/manual_jobs/redis.py
```diff
@@ -72,6 +72,8 @@ def _get_job_types_multi(job_ids_keys: Sequence[str]) -> List[str]:
 
 
 def _get_job_status_multi(job_ids_keys: Sequence[str]) -> List[JobStatus]:
+    if len(job_ids_keys) == 0:
+        return []
```
onewland, line 76: nit: add a newline under this `return []` or make it `return [] if len(job_ids_keys) == 0` (with a newline before the next return)

onewland (approved):

PR #6514, opened 2024-11-05
feat(dev): manage python through devenv

joshuarli:
this adds devenv to manage python + `.venv`

future improvements:
- auto-install rust
- auto-install node+yarn for snuba admin

PR #6515, opened 2024-11-05
devenv: add colima for devservices

joshuarli:
devenv and colima are required for https://github.com/getsentry/devservices

i have a followup pr https://github.com/getsentry/snuba/pull/6514 which will require devenv in `.envrc` and adds the ability to manage the `.venv`


joshuarli:
ok colima will probably be moved to devservices

PR #6536, opened 2024-11-07
feat(eap): Add support for extrapolation

davidtsuk:
Added support for EXTRAPOLATION_MODE_SAMPLE_WEIGHTED


snuba/query/processors/logical/optional_attribute_aggregation.py
```diff
@@ -37,66 +37,69 @@ def __init__(
...
-        def is_attribute_expression(exp_parameters: tuple[Expression, ...]) -> bool:
-            if len(exp_parameters) != 1:
-                return False
-            func_param = exp_parameters[0]
-            if not isinstance(func_param, SubscriptableReference):
-                return False
-            if func_param.column.column_name in self._attribute_column_names:
-                return True
-            return False
+        def find_subscriptable_reference(exp: Expression) -> SubscriptableReference:
```
volokluev, line 40: why do you need to do this? transform expressions will already iterate through every expression in the query

davidtsuk: this is to handle cases like `sum(multiply(multiply(...)))`

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table_extrapolation.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+BASE_TIME = datetime.utcnow().replace(minute=0, second=0, microsecond=0) - timedelta(
+    minutes=180
+)
+
+
+@pytest.mark.clickhouse_db
+@pytest.mark.redis_db
+class TestTraceItemTableWithExtrapolation(BaseApiTest):
```
volokluev, line 116: This test should test all the other extrapolated aggregations that you implemented, how do you know they work?

Also the math here is the part that is unclear so you should take the time to explain that. Why should the sum equal 155? If you come back to this test in 4 months, how much reverse engineering are you going to have to do?

volokluev (commented):
Implementation looks good but let's add more testing

davidtsuk:
@volokluev I added the other aggregations to the test, but testing the quantiles is a bit difficult since they are inaccurate for small datasets so I only tested p90 which appears to be accurate.
From the Clickhouse Docs:
image

volokluev (approved):

PR #6545, opened 2024-11-08
[CapMan visibility] rejected queries are ran with 0 threads

xurui-c:
Snuba payload should correctly reflect that rejected queries are ran with 0 threads


volokluev (commented):
no test?

onkar (approved):

PR #6560, opened 2024-11-11
ref(deletes): add use_bulk_deletes runtime config

MeredithAnya:
Adding a temporary runtime config (`use_bulk_deletes`) to allow for delete queries to go through the bulk delete pipeline, until we are ready to fully move over to using it. 

I choose a runtime config because I didn't want sentry to have to be aware of which implementation was used (bulk deletes or not). 

We should only enable this when we have the consumers running in S4S

### DIFF
--- a/snuba/web/views.py
+++ b/snuba/web/views.py
@@ -58,7 +58,7 @@ from snuba.query.query_settings import HTTPQuerySettings
 from snuba.redis import all_redis_clients
 from snuba.request.exceptions import InvalidJsonRequestException, JsonDecodeException
 from snuba.request.schema import RequestSchema
-from snuba.state import get_config
+from snuba.state import get_int_config
 from snuba.state.rate_limit import RateLimitExceeded
 from snuba.subscriptions.codecs import SubscriptionDataCodec
 from snuba.subscriptions.data import PartitionId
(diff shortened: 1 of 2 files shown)


snuba/web/views.py
```diff
@@ -330,11 +332,18 @@ def storage_delete(
         try:
             schema = RequestSchema.build(HTTPQuerySettings, is_delete=True)
             request_parts = schema.validate(body)
-            payload = delete_from_storage(
-                storage,
-                request_parts.query["query"]["columns"],
-                request_parts.attribution_info,
-            )
+            if get_config("use_bulk_deletes", False):
```
onkar, line 335: The default is `False` but shouldn't we set `use_bulk_delete` to `False` using `set_config()` in this PR and flip it to `True` once we are ready to use bulk delete in s4s? Also, the note on https://github.com/getsentry/snuba/blob/ce292773fabd4f8a49970944d794868230655cd3/snuba/state/__init__.py#L190 says that `get_config()` is supposed to be deprecated. Does it make sense to add a function in `snuba/state` that returns `bool` and use it here?

onkar (approved):
LGTM overall, added a couple of questions.

snuba/web/views.py
MeredithAnya: oh didn't see that comment - I'll update

and I can set the default to be `0` and use `get_int_config`

PR #6561, opened 2024-11-11
feat(consumers): rust consumers quantized rebalance

volokluev:
Add the ability to do quantized rebalancing for consumers, toggleable by runtime config.

### What is quantized rebalancing?

Rebalancing events are synchronized to the tick of the clock. Controlled by the runtime config: `quantized_rebalance_consumer_group_delay_secs__{consumer_group}`

Let's say: 

`SET  quantized_rebalance_consumer_group_delay_secs__spans=15` 

* When a consumer starts up: delay subscribing to a topic until `timestamp % 15 == 0`
* when a consumer gets a shutdown signal (ctrl-c): delay leaving the group until `timestamp % 15 == 0`

### What else is in this PR?

This PR adds the ability to read from redis config directly. This was mostly done because I couldn't get the python bindings to work on my machine, it will still fall back to the old implementation if there is an error in the direct implementation


rust_snuba/src/consumer.rs
```diff
@@ -286,10 +292,21 @@ pub fn consumer_impl(
...
 
-    ctrlc::set_handler(move || {
-        handle.signal_shutdown();
-    })
-    .expect("Error setting Ctrl-C handler");
+    match rebalance_delay_secs {
+        Some(secs) => {
+            ctrlc::set_handler(move || {
+                rebalancing::delay_kafka_rebalance(secs);
+                handle.signal_shutdown();
```
john-z-yang, line 299: Am I correct in assuming that this closure that is passed in runs in a separate thread, so the main loop is not blocked while it is sleeping before signaling shutdown?

john-z-yang:
This seems reasonable to me, do we plan to have quantized rebalancing on the other end? I.e. Making the consumer wait a little bit before joining the consumer group

rust_snuba/src/consumer.rs
john-z-yang, line 299: I'm curious if arroyo let's us to the following
-

volokluev: From the documentation it seems to be the case

https://docs.rs/ctrlc/latest/ctrlc/fn.set_handler.html

```
Register signal handler for Ctrl-C.

Starts a new dedicated signal handling thread. Should only be called once, typically at the start of your program.
```

volokluev:
@john-z-yang it does wait before joining, see

https://github.com/getsentry/snuba/pull/6561/files#diff-03eab290aeb86b49a5a13b44a3a48fd7cae4efdc4ae908e4e92309f8241aaac9R238-R242

(and I tested that this happens locally as well)

rust_snuba/src/rebalancing.rs
```diff
@@ -0,0 +1,81 @@
...
+    ) {
+        Ok(delay_secs) => match delay_secs {
+            Some(secs) => match secs.parse() {
+                Ok(v) => Some(v),
+                Err(_) => None,
+            },
+            None => None,
+        },
+        Err(_) => None,
+    }
```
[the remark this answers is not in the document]
volokluev: nice, today I learned about `.ok` :D thank you john

john-z-yang (approved):

rust_snuba/src/consumer.rs
john-z-yang: Nice!

PR #6564, opened 2024-11-12
feat(consumers): rust consumer runtime config

volokluev:
Our current setup is accessing runtime config by FFI'ing into the python module. This is not only ugly and a performance hit, but also makes local testing development difficult because the FFI semantics don't seem to be the same across machines. (For example, I can't get the `test_python` and `test_runtime_config` tests to work on my machine.

The fewer python dependencies we have the better

### Do not merge this yet, ops repo needs to be changed to pass the redis params in via env vars

### DIFF
--- a/rust_snuba/Cargo.lock
+++ b/rust_snuba/Cargo.lock
@@ -125,6 +125,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


untitaker:
not opposed to landing this if it makes local development simpler (was unaware of those issues), but we would have to commit to snuba's redis being a stable API boundary. or maybe we port runtime config to Rust, and call it from Python?

what was the error you got locally for those tests?

volokluev:
I got a `No module found: snuba` error

untitaker:
Ah, it's possible that `cargo` didn't run in the virtualenv? can debug next week maybe

PR #6571, opened 2024-11-18
feat(subscriptions): Add a create subscription rpc endpoint

shruthilayaj:
This PR adds the CreateSubscription RPC endpoint. To create an RPC subscription we need to:
1. Create an RPC subscription object from the CreateSubscription protobuf
2. Serialize the request (to put it into redis)
3. Validate it
4. Build it & run it before saving

It builds and runs subscription request outside the RPCSubscriptionData class
for now so that we don't have to update the entire subscription pipeline to use both
SnubaRequest and TimeseriesRequest types.

### DIFF
--- a/snuba/subscriptions/data.py
+++ b/snuba/subscriptions/data.py
@@ -312,21 +312,6 @@ class SnQLSubscriptionData(SubscriptionData):
                 "At least one Entity must have a timestamp column for subscriptions"
             )
 
-    def validate(self) -> None:
-        if self.time_window_sec < 60:
-            raise InvalidSubscriptionError(
-                "Time window must be greater than or equal to 1 minute"
-            )
-        elif self.time_window_sec > 60 * 60 * 24:
-            raise InvalidSubscriptionError(
-                "Time window must be less than or equal to 24 hours"
-            )
-
-        if self.resolution_sec < 60:
-            raise InvalidSubscriptionError(
-                "Resolution must be greater than or equal to 1 minute"
-            )
-
     def build_request(
         self,
         dataset: Dataset,
--- a/snuba/subscriptions/data.py
+++ b/snuba/subscriptions/data.py
@@ -25,6 +25,7 @@ from sentry_protos.snuba.v1.endpoint_create_subscription_pb2 import (
     CreateSubscriptionRequest,
 )
 from sentry_protos.snuba.v1.endpoint_time_series_pb2 import TimeSeriesRequest
+from sentry_protos.snuba.v1.trace_item_attribute_pb2 import ExtrapolationMode
 
 from snuba.datasets.dataset import Dataset
 from snuba.datasets.entities.entity_key import EntityKey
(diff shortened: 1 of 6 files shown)


tests/web/rpc/v1/test_create_subscription.py
```diff
@@ -0,0 +1,124 @@
...
+            redis_client.hgetall("subscriptions:eap_spans:0").items()
+        )[0]
+        subscription_request = stored_subscription_data[1]
+        subscription_data = json.loads(subscription_request.decode("utf-8"))
+
+        assert "time_series_request" in subscription_data
+        assert subscription_data["time_window"] == 300
+        assert subscription_data["resolution"] == 60
+        assert subscription_data["request_name"] == "TimeSeriesRequest"
+        assert subscription_data["request_version"] == "v1"
```
volokluev, line 82: you should actually try and deserialize the request here

tests/web/rpc/v1/test_create_subscription.py
```diff
@@ -0,0 +1,124 @@
...
+            "/rpc/CreateSubscriptionRequest/v1", data=message.SerializeToString()
+        )
+        assert response.status_code == 200
+        response_class = CreateSubscriptionResponse()
+        response_class.ParseFromString(response.data)
+        assert response_class.subscription_id
+
+        redis_client = get_redis_client(RedisClientKey.SUBSCRIPTION_STORE)
+        stored_subscription_data = list(
+            redis_client.hgetall("subscriptions:eap_spans:0").items()
```
volokluev, line 73: ```"subscriptions:eap_spans:0"```

where did this string come from?

snuba/subscriptions/data.py
```diff
@@ -83,9 +101,20 @@ class SubscriptionData(ABC):
...
+        if self.time_window_sec < 60:
+            raise InvalidSubscriptionError(
+                "Time window must be greater than or equal to 1 minute"
+            )
+        elif self.time_window_sec > 60 * 60 * 24:
+            raise InvalidSubscriptionError(
+                "Time window must be less than or equal to 24 hours"
+            )
+
+        if self.resolution_sec < 60:
```
volokluev, line 114: also validate that there is not groupby in the request too

shruthilayaj: This validation function runs for both SnQL and RPCs, I'll be updating the RPC validation to be more robust
https://github.com/getsentry/snuba/pull/6571/files#diff-8e3672e3fdccd3ce17c850cd47dfe0b41f53feabd7dcf550c30077fe3f7e31fbR154-R158
I'm planning to do that in a follow up, if that's okay. I want to validate there are no group bys, orderbys, only a single aggregation, single project id, a bunch of stuff

shruthilayaj: Ended up adding it here! I expected to add more validation, but there wasn't much else I could think of at the moment https://github.com/getsentry/snuba/pull/6571/commits/83f1693dcb4fe17cf3c46fb8f7c4027ec222939b

volokluev (approved):

PR #6580, opened 2024-11-19
fix: convert max_batch_time seconds

MeredithAnya:
`max_batch_time` in the batch step calls for seconds, not milliseconds.


xurui-c (approved):

snuba/lw_deletions/strategy.py
```diff
@@ -169,7 +169,7 @@ def create_with_partitions(
     ) -> ProcessingStrategy[KafkaPayload]:
         batch_step = BatchStepCustom(
             max_batch_size=self.max_batch_size,
-            max_batch_time=self.max_batch_time_ms,
+            max_batch_time=(self.max_batch_time_ms / 1000),
```
onkar, line 172: Since max_batch_time is an int, should this be:
`max_batch_time=(self.max_batch_time_ms // 1000),`

onkar (approved):
LGTM, a minor comment.

PR #6581, opened 2024-11-19
RPC handles same name attributes

xurui-c:
Fixes https://github.com/getsentry/projects/issues/400

### DIFF
--- a/snuba/web/rpc/common/common.py
+++ b/snuba/web/rpc/common/common.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timedelta
-from typing import Final, Mapping, Optional, Sequence, Set
+from typing import Final, Mapping, Sequence, Set
 
 from sentry_protos.snuba.v1.request_common_pb2 import RequestMeta
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import (
@@ -194,15 +194,15 @@ TIMESTAMP_COLUMNS: Final[Set[str]] = {
 }
 
 
-def attribute_key_to_expression(
-    attr_key: AttributeKey, alias: Optional[str] = None
-) -> Expression:
+def attribute_key_to_expression(attr_key: AttributeKey) -> Expression:
+    def _build_label_mapping_key(attr_key: AttributeKey) -> str:
+        return attr_key.name + " " + AttributeKey.Type.Name(attr_key.type)
+
     if attr_key.type == AttributeKey.Type.TYPE_UNSPECIFIED:
         raise BadSnubaRPCRequestException(
             f"attribute key {attr_key.name} must have a type specified"
         )
-    if alias is None:
-        alias = attr_key.name
+    alias = _build_label_mapping_key(attr_key)
 
     if attr_key.name == "sentry.trace_id":
         if attr_key.type == AttributeKey.Type.TYPE_STRING:
--- a/snuba/web/rpc/common/common.py
+++ b/snuba/web/rpc/common/common.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timedelta
-from typing import Final, Mapping, Sequence, Set
+from typing import Final, Mapping, Optional, Sequence, Set
 
 from sentry_protos.snuba.v1.request_common_pb2 import RequestMeta
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import (
@@ -194,12 +194,15 @@ TIMESTAMP_COLUMNS: Final[Set[str]] = {
 }
 
 
-def attribute_key_to_expression(attr_key: AttributeKey) -> Expression:
+def attribute_key_to_expression(
+    attr_key: AttributeKey, alias: Optional[str] = None
+) -> Expression:
     if attr_key.type == AttributeKey.Type.TYPE_UNSPECIFIED:
         raise BadSnubaRPCRequestException(
             f"attribute key {attr_key.name} must have a type specified"
         )
-    alias = attr_key.name
+    if alias is None:
+        alias = attr_key.name
 
     if attr_key.name == "sentry.trace_id":
         if attr_key.type == AttributeKey.Type.TYPE_STRING:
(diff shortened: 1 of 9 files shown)


xurui-c:
This is what CH looks like after generating the message:
Screenshot 2024-11-19 at 11 46 53 PM

This is what `columns` look like in the RPC request:
`{
      "key": {
        "type": "TYPE_INT",
        "name": "foo"
      },
      "label": "tags[foo,number]"
    },
    {
      "key": {
        "type": "TYPE_STRING",
        "name": "foo"
      },
      "label": "tags[foo,string]"
    },
    {
      "key": {
        "type": "TYPE_STRING",
        "name": "foo"
      },
      "label": "tags[foo]"
    },`
    
after processing the first column (`label`=`tags[foo,number]`), the query pipeline caches the name `foo` in `ParsingContext()` [here](https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/formatter/expression.py#L56). So when processing the second column (`label`=`tags[foo,string]`) and the third column (`label`=`tags[foo]`), the query pipeline sees that the `name` (or `alias` in this part of the codebase) is the same (`foo`), so it will use the cached result [here](https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/formatter/expression.py#L48-L54).

What this caching alias does is turn the CH query into this: `SELECT (name AS `sentry.name`), **(CAST(arrayElement(attr_num_3, 'foo'), 'Int64') AS foo), foo, foo**, (lower(hex(span_id)) AS `sentry.span_id`), (transform(CAST((project_id AS `sentry.project_id`), 'String'), ['4555075531898880'], ['bar'], 'unknown') AS `project.name`) FROM eap_spans_2_local WHERE in(project_id, [1]) AND equals(organization_id, 1) AND less(_sort_timestamp, toDateTime(1732107600)) AND greaterOrEquals(_sort_timestamp, toDateTime(1732104000)) AND true GROUP BY `sentry.name`, foo, foo, foo, `sentry.span_id`, `project.name` ORDER BY `sentry.name` ASC LIMIT 10000 OFFSET 0`. So the second and third `foo` columns (which should've been string `"five"`) are cached to be the same as `(CAST(arrayElement(attr_num_3, 'foo'), 'Int64') AS foo)`, which is integer `5`.

Potential solutions:
1. What if we cache at the end of a query, instead of as we process each column? I tried that in a [previous commit](https://github.com/getsentry/snuba/pull/6581/commits/7f5b6e94f87942ed1255b1389cb6603c23c1a516) of this PR, and made the CH query look like this: `SELECT (name AS `sentry.name`), (CAST(arrayElement(attr_num_3, 'foo'), 'Int64') AS foo), (arrayElement(attr_str_3, 'foo') AS foo), (arrayElement(attr_str_3, 'foo') AS foo), (lower(hex(span_id)) AS `sentry.span_id`), (transform(CAST((project_id AS `sentry.project_id`), 'String'), ['4555075531898880'], ['bar'], 'unknown') AS `project.name`) FROM eap_spans_2_local WHERE in(project_id, [1]) AND equals(organization_id, 1) AND less(_sort_timestamp, toDateTime(1732107600)) AND greaterOrEquals(_sort_timestamp, toDateTime(1732104000)) AND true GROUP BY `sentry.name`, foo, foo, foo, `sentry.span_id`, `project.name` ORDER BY `sentry.name` ASC LIMIT 10000 OFFSET 0`. But, clickhouse does not allow the same alias (`foo`). The error is `snuba.clickhouse.errors.ClickhouseError: DB::Exception: Different expressions with the same alias foo:
attr_str_3['foo'] AS foo
and
CAST(attr_num_3['foo'], 'Int64') AS foo
: While processing attr_str_3['foo'] AS foo. Stack trace:`
2. We don't allow attributes with the same name but different types
3. Change the query pipeline so that we use the [label](https://github.com/getsentry/sentry-protos/blob/4f931bdaebdf482adf0d5030c46e52a91f2461e7/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto#L31) as the alias in the CH query

volokluev:
Great job reproducing this @xurui-c 

I think what would make sense is to have the expression be generated by us [here](https://github.com/getsentry/snuba/blob/master/snuba/web/rpc/common/common.py#L202) to avoid alias collisions. 

And the SelectedExpression alias should have the label the client assigned in the API call. Does that make sense?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -842,6 +842,111 @@ def test_cast_bug(self, setup_teardown: Any) -> None:
         # just ensuring it doesnt raise an exception
         EndpointTraceItemTable().execute(err_msg)
 
+    def test_rachel(self) -> None:
```
xurui-c, line 845: need to change name of test

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -65,6 +65,10 @@ def _convert_order_by(
     return res
 
 
+def _build_label_mapping_key(column_key: AttributeKey) -> str:
+    return column_key.name + " " + AttributeKey.Type.Name(column_key.type)
```
volokluev, line 69: I think you should just do this in `attribute_key_to_expression`

volokluev: and it should be the default behavior

snuba/web/rpc/common/common.py
```diff
@@ -195,11 +195,14 @@ def get_subscriptable_field(field: Expression) -> SubscriptableReference | None:
 
 
 def attribute_key_to_expression(attr_key: AttributeKey) -> Expression:
+    def _build_label_mapping_key(attr_key: AttributeKey) -> str:
+        return attr_key.name + " " + AttributeKey.Type.Name(attr_key.type)
```
volokluev, line 199: tip: don't use spaces as delimiters in software systems. `__` usually is a better choice

volokluev (approved):
almost there, just change the delimiter and then you are good to merge. Also @davidtsuk does this mess up your extrapolation aliasing business?

snuba/web/rpc/common/common.py
xurui-c: https://github.com/getsentry/snuba/pull/6592

PR #6586, opened 2024-11-20
test(runtime-config): Add some tests to make sure data is cached

john-z-yang:
### Overview

We ran into an incident where we accidentally did not store runtime config into our cache and retrieved it from redis every time the get_runtime_config is called. This pr tries to add a couple of tests to assert the caching behaviour.

### DIFF
--- a/rust_snuba/src/runtime_config.rs
+++ b/rust_snuba/src/runtime_config.rs
@@ -40,12 +40,56 @@ pub fn get_str_config(key: &str) -> Result<Option<String>, Error> {
 
 #[cfg(test)]
 mod tests {
+
     use super::*;
 
     #[test]
     fn test_runtime_config() {
         crate::testutils::initialize_python();
-        let config = get_str_config("test");
+        let config = get_str_config("key0");
         assert_eq!(config.unwrap(), None);
+        CONFIG.write().clear();
+    }
+
+    #[test]
+    fn test_runtime_config_is_stored_in_cache() {
+        crate::testutils::initialize_python();
+        assert!(!CONFIG.read().contains_key("key1"));
+        let config = get_str_config("key1");
+        assert_eq!(config.unwrap(), None);
+        assert!(CONFIG.read().contains_key("key1"));
+        CONFIG.write().clear();
+    }
+
+    #[test]
+    fn test_runtime_config_retrieves_from_cache_within_deadline() {
+        {
+            CONFIG.write().insert(
+                "key2".to_string(),
+                (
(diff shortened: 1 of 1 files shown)


volokluev:
Good start but what we really want to check is that `get_str_config` actually stores a value in cache if it retrieves it

rust_snuba/src/runtime_config.rs
```diff
@@ -48,4 +50,39 @@ mod tests {
...
+        CONFIG.write().clear();
+    }
+
+    #[test]
+    fn test_runtime_config_invalidates_cache_outside_deadline() {
+        CONFIG.write().insert(
+            "key".to_string(),
+            (
+                Some("value".to_string()),
+                Deadline::new(Duration::from_secs(0)),
```
john-z-yang, line 82: I wanted to do Deadline::new(Duration::from_secs(1)) here, and then do a sleep but because we're using coarse time, the cache doesn't seem like it's invalidated?

john-z-yang:
@volokluev Let me know how `test_runtime_config_is_stored_in_cache` looks. I prefer not to reaching into python and set a vale there

rust_snuba/src/runtime_config.rs
```diff
@@ -40,12 +40,56 @@ pub fn get_str_config(key: &str) -> Result<Option<String>, Error> {
...
+                "key3".to_string(),
+                (
+                    Some("value3".to_string()),
+                    Deadline::new(Duration::from_secs(0)),
+                ),
+            );
+        }
+        let config = get_str_config("key3");
         assert_eq!(config.unwrap(), None);
+        CONFIG.write().clear();
```
john-z-yang, line 93: I wanted to set this deadline to maybe 1 sec from now, and sleep for 1 sec to retrieve it, like so
```rust
#[test]
    fn test_runtime_config_invalidates_cache_outside_deadline() {
        {
            CONFIG.write().insert(
                "key3".to_string(),
                (
                    Some("value3".to_string()),
                    Deadline::new(Duration::from_secs(1)),
                ),
            );
        }
        thread::sleep(Duration::from_secs(900));
        let config = get_str_config("key3");
        assert_eq!(config.unwrap(), None);
        CONFIG.write().clear();
    }
```
But for some reason this fails, and the debugger shows that `coarsetime::Instant::recent()` does not appear to be advancing

untitaker: these tests will run concurrently -- the easiest way to get around this issue is to run all tests in one function.

PR #6587, opened 2024-11-20
feat(devservices): Add restart settings and use external devservices network

hubertdeng123:
This automatically restarts the containers defined in the config file when they crash, unless explicitly stopped. Also uses devservices as an external network to avoid race conditions involving creating/removing networks.


IanWoodard (approved):

davidtsuk (approved):
LGTM

PR #6590, opened 2024-11-21
chore(metrics-summaries): Remove metrics summaries consumer code

phacops:
Related PR to stop the consumers in prod https://github.com/getsentry/ops/pull/12968.

Since we removed all the code querying the dataset already, it's OK to remove consumer + dataset code at the same time since it's intertwined.

https://github.com/getsentry/team-ingest/issues/544


Zylphrex (approved):
🔥

snuba/clusters/storage_sets.py
```diff
@@ -19,7 +19,6 @@
     "SPANS": "spans",
     "EVENTS_ANALYTICS_PLATFORM": "events_analytics_platform",
     "GROUP_ATTRIBUTES": "group_attributes",
-    "METRICS_SUMMARIES": "metrics_summaries",
```
MeredithAnya, line 22: @phacops don't you need to remove the clickhouse cluster definitions from ops (https://github.com/getsentry/ops/blob/b37ad515a21d028102dd8c9aeb8f9e751ac4b22c/k8s/services/snuba/region_overrides/us/default.yaml#L281) before you remove this from snuba?

phacops: Good point, you're right.

phacops: https://github.com/getsentry/ops/pull/12987

phacops: I think when we add a new dataset, we configure the cluster and associate the dataset to the cluster even if there's no code associated with it in order to have the migrations go to the right cluster right away.

It would make sense to go the reverse order to remove the dataset, first the code, then the config in ops.

MeredithAnya (approved):
We are going to remove this code changes here first before removing the related ops configuration.

PR #6593, opened 2024-11-21
fix(eap): fix bug where we pass non str group by mapping to timeseries

davidtsuk:
Fixes https://github.com/getsentry/projects/issues/426
TimeSeries expects a mapping from str to str, so we have to convert the column value to a string first.


volokluev (commented):
test?

davidtsuk:
@volokluev done

volokluev (approved):

PR #6594, opened 2024-11-22
fix(eap-api): align time buckets correctly

xurui-c:
Fixes https://github.com/getsentry/projects/issues/364 by making start_timestamp divisible by granularity. This makes it so that our timebuckets line up properly.

### DIFF
--- a/snuba/web/rpc/v1/endpoint_time_series.py
+++ b/snuba/web/rpc/v1/endpoint_time_series.py
@@ -65,6 +65,10 @@ _VALID_GRANULARITY_SECS = set(
 _MAX_BUCKETS_IN_REQUEST = 1000
 
 
+def _rewind(start_timestamp: int, granularity: int) -> int:
+    return (start_timestamp // granularity) * granularity
+
+
 def _convert_result_timeseries(
     request: TimeSeriesRequest, data: list[Dict[str, Any]]
 ) -> Iterable[TimeSeries]:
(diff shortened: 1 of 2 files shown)


snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -128,11 +132,15 @@ def _convert_result_timeseries(
...
+    start_timestamp_seconds = _rewind(
+        request.meta.start_timestamp.seconds, request.granularity_secs
+    )
+
     query_duration = (
         request.meta.end_timestamp.seconds - request.meta.start_timestamp.seconds
     )
     time_buckets = [
-        Timestamp(seconds=(request.meta.start_timestamp.seconds) + secs)
+        Timestamp(seconds=(start_timestamp_seconds) + secs)
```
shruthilayaj, line 143: This will drop the last time bucket unless the request timestamp is perfectly aligned to begin with. Here's an example, say my start timestamp was **10:53:30** 
and end timestamp was 5 min later, **10:58:30** 
I request it with 1 minute granularity, a `GROUP BY toDateTime(multiply(intDiv(toUInt32(_sort_timestamp), 60), 60), 'Universal')` would return the following time buckets in the query results:

10:53
10:54
10:55
10:56
10:57
10:58

But when we generate the time buckets here, we'd drop the last bucket since we're rewinding and only generating buckets based on the query duration: starts at 10:53, incrementing by one min for 5 min, so ends at 10:57.
So maybe we should always generate the next time bucket here also?

PR #6595, opened 2024-11-22
feat(consumers): Quantized rebalancing in Rust

untitaker:
reapply https://github.com/getsentry/snuba/pull/6561 but remove the direct redis call. thanks to volo for writing basically all of the code (with tests!)

personal opinion: redis is not a great API boundary to have (there's very few things in redis that help you abstract over the raw datastructure), and using it directly from two codebases would be like accessing the same postgres from multiple microservices.

it seems to work fine on localhost. for future reference, I did:

```
make watch-rust-snuba
# then make my changes
# then go to snuba admin and set the setting for consumer group lol4
RUST_LOG=info snuba rust-consumer --log-level info --storage eap_spans --consumer-group lol4 --concurrency 2 --use-rust-processor --max-batch-size 1000000 --max-batch-time-ms 100 --auto-offset-reset latest --no-strict-offset-reset
```

watch the log output

### DIFF
--- a/rust_snuba/src/runtime_config.rs
+++ b/rust_snuba/src/runtime_config.rs
@@ -1,9 +1,7 @@
 use anyhow::Error;
 use parking_lot::RwLock;
 use pyo3::prelude::{PyModule, Python};
-use redis::Commands;
 use std::collections::BTreeMap;
-use std::collections::HashMap;
 use std::time::Duration;
 
 use rust_arroyo::timer;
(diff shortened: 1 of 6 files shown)


rust_snuba/src/runtime_config.rs
```diff
@@ -9,6 +9,15 @@ use rust_arroyo::utils::timing::Deadline;
 
 static CONFIG: RwLock<BTreeMap<String, (Option<String>, Deadline)>> = RwLock::new(BTreeMap::new());
 
+#[cfg(test)]
+pub fn patch_str_config_for_test(key: &str, value: Option<&str>) {
```
volokluev, line 13: elegant, nice job

volokluev (approved):

PR #6597, opened 2024-11-22
feat(eap): Extract profile id for eap spans

Zylphrex:
We need to make sure we extract profile id for eap spans as well.


Zylphrex:
> add to the test plz

done

volokluev (approved):

PR #6601, opened 2024-11-25
feat(profiler): Update formatting of Heaptrack output file

ayirr7:
Since Heaptrack performs continuous profiling of the Rust consumer, we will want to track the profiler's output for various profiling runs that happen. A profiling run happens for every consumer process that is started.  

Adds timestamp and ensures Heaptrack writes to a formatted file for each profiling run. This will also make analysis easier as we will know which file corresponds to which date and time.

### DIFF
--- a/docker_entrypoint.sh
+++ b/docker_entrypoint.sh
@@ -19,7 +19,8 @@ else
 fi
 
 if [ -n "${ENABLE_HEAPTRACK:-}" ]; then
-  set -- heaptrack "$@"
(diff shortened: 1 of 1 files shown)


volokluev:
That sounds like it can grow very high

ayirr7:
@volokluev It wouldn't be every second. It would be 1 file per profiling run -- and Heaptrack will write a new file for every consumer process that is started. 

So maybe easier to think of it as 1 per pod startup.

volokluev (approved):

IanWoodard (approved):

PR #6604, opened 2024-11-26
Remove hierarchical hashes from code base

armenzg:
Hierarchical grouping was deprecated back in September.

See project: https://github.com/orgs/getsentry/projects/242

The failing test depends on this PR getting merged first: https://github.com/getsentry/sentry-kafka-schemas/pull/353


untitaker (approved):
I think, assuming that CI passes, this can be rolled out. However, we should follow up with a migration to remove this col from prod, eventually.

PR #6605, opened 2024-11-27
perf(final): add flag to skip cross-partition merging on FINAL

onewland:
The errors table is partitioned by a key that never changes (timestamp of the event). Given that, we can pass [this ClickHouse query setting](https://clickhouse.com/docs/en/guides/replacing-merge-tree#exploiting-partitions-with-replacingmergetree) and hopefully see some reduced load/performance improvement on queries for groups/projects/etc. that have had replacements run on them. Based on the docs, the worst case seems like no improvement

### DIFF
--- a/snuba/query/processors/physical/replaced_groups.py
+++ b/snuba/query/processors/physical/replaced_groups.py
@@ -129,6 +129,15 @@ class PostReplacementConsistencyEnforcer(ClickhouseQueryProcessor):
                     )
                 )
 
+        if (
+            set_final
+            and self.__replacer_state_name is not None
+            and self.__replacer_state_name == ReplacerState.ERRORS
+        ):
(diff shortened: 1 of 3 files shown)


evanh (approved):

onkar:
@onewland can you provide details on the testing? There's unit test but how do we make sure that the setting works as expected on the larger dataset?

onewland:
> @onewland can you provide details on the testing? There's unit test but how do we make sure that the setting works as expected on the larger dataset?


I don't see any way to meaningfully test this. We can verify that it works locally but it wouldn't change my confidence much about whether it will work in SaaS.

Note: this is a read-side query setting change, so it can be reverted trivially and future queries won't be affected.

onkar (approved):

PR #6608, opened 2024-12-03
chore: Upgrade sentry-kafka-schemas

armenzg:
This is to include [this change](https://github.com/getsentry/sentry-kafka-schemas/pull/353) and to unblock #6604.


armenzg:
I have cherry-picked the change and put it in https://github.com/getsentry/snuba/pull/6604

PR #6609, opened 2024-12-03
feat(eap): Rolling time buckets

volokluev:
Fixes https://github.com/getsentry/projects/issues/364

Allows bucketing at arbitrary time intervals by taking the difference between the start time and the timestamp as opposed to what the TimeSeriesProcessor did which was 

`toDateTime(multiply(intDiv(toUInt32(_sort_timestamp), 300), 300), 'Universal')`

This would always make it land on the clean interval of `09:00, 09:05, 09:10 ...`


shruthilayaj (approved):
👍

xurui-c (approved):

PR #6611, opened 2024-12-04
chore(rpc-subscriptions): Remove timestamp rounding

shruthilayaj:
With this change https://github.com/getsentry/snuba/pull/6609 it's no longer 
required to round the timestamps to get back a single time bucket.

### DIFF
--- a/tests/subscriptions/__init__.py
+++ b/tests/subscriptions/__init__.py
@@ -174,7 +174,10 @@ class BaseSubscriptionTest:
             gen_span_message(self.base_time + timedelta(minutes=tick))
             for tick in range(self.minutes)
         ]
-        write_raw_unprocessed_events(spans_storage, messages)
+        extra_messages = [
+            gen_span_message(self.base_time - timedelta(hours=4)) for _ in range(2)
+        ]
+        write_raw_unprocessed_events(spans_storage, extra_messages + messages)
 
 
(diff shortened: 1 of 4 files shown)


tests/subscriptions/test_data.py
```diff
@@ -124,7 +124,7 @@
                         ),
                     ],
                 ),
-                time_window_secs=3600,
+                time_window_secs=10800,
```
shruthilayaj, line 127: The `BaseSubscriptionTest` class that the test inherits from adds events to the previous hour. This test worked before because we were rounding timestamp to the nearest granularity. 

The larger time window will catch the events I expect to catch.

tests/subscriptions/__init__.py
```diff
@@ -174,7 +174,10 @@ def setup_teardown(self, clickhouse_db: None) -> None:
             gen_span_message(self.base_time + timedelta(minutes=tick))
             for tick in range(self.minutes)
         ]
-        write_raw_unprocessed_events(spans_storage, messages)
+        extra_messages = [
+            gen_span_message(self.base_time - timedelta(hours=4)) for _ in range(2)
+        ]
```
shruthilayaj, line 179: adding a couple extra messages outside the time bounds of the subscription execution so I know those don't get included in the count

volokluev (approved):

PR #6613, opened 2024-12-04
feat(eap-api): Use signs in count, average, and sum calculations

davidtsuk:
Closes https://github.com/getsentry/eap-planning/issues/105


volokluev (approved):

kylemumma:
What is the motivation behind this PR?

davidtsuk:
> What is the motivation behind this PR?

We are using a ReplacingMergeTree which stores a sign representing whether a row was added or deleted. In order to compute counts, sums, and averages correctly we need to account for the sign in the calculation.

onkar:
@davidtsuk should we add unit tests covering the sign representation?

davidtsuk:
> @davidtsuk should we add unit tests covering the sign representation?

We aren't actually using negative signs yet, so we can't add any tests yet for that. Since the existing tests pass, I think it's okay to wait until we start using negative signs.

PR #6614, opened 2024-12-04
chore: Bump rust-arroyo to published crate

untitaker:


onkar:
@untitaker the last run failed but I re-ran the pipeline and it passes now. Guess you are free to merge this now.

onkar (approved):
Should this commit message be "Replace rust_arroyo with sentry_arroyo"?

untitaker:
I don't think it makes it more understandable. The crate got renamed but in fact it is a new version of the same codebase.

PR #6618, opened 2024-12-05
Implement filter offset for attribute keys API

xurui-c:
https://github.com/getsentry/projects/issues/376

Need to do this for attribute values API next


tests/web/rpc/v1/test_endpoint_trace_item_attribute_names.py
```diff
@@ -212,26 +215,53 @@ def test_with_page_token(self) -> None:
...
+        # grab 10 at a time until we get them all
+        done = 0
+        page_token = PageToken(
+            filter_offset=TraceItemFilter(
+                comparison_filter=ComparisonFilter(
+                    key=AttributeKey(type=AttributeKey.TYPE_STRING, name="attr_key"),
+                    op=ComparisonFilter.OP_GREATER_THAN,
+                    value=AttributeValue(val_str=""),
+                )
+            )
```
volokluev, line 236: you should be able to pass in a null page token

volokluev (approved):

PR #6621, opened 2024-12-06
chore(discover): Add trace response code to context array

shruthilayaj:
Related to https://github.com/getsentry/sentry/issues/81418
Looks like some SDKs don't send a response context which
is where we currently read http.status code from. This information
is also found in the data dict on the trace context in most SDKs,
so adding promoting that to the contexts array so users can
query/aggregate on this data.


shruthilayaj:
Going to follow up with the SDK team first since this is happening because the response context isn't being attached to the event.

PR #6635, opened 2024-12-10
chore(api): Do not log healthcheck error if downfile exists

untitaker:
Kubernetes tells the old snuba pods to mark itself as unhealthy, so that
envoy stops sending traffic to it.

Snuba reports itself as unhealthy to Sentry.

Gocd checks Sentry for new errors.

Gocd stalls the deployment because there are new errors.

### DIFF
--- a/snuba/utils/health_info.py
+++ b/snuba/utils/health_info.py
@@ -107,7 +107,8 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
     payload = json.dumps(body)
     if status != 200:
         metrics.increment("healthcheck_failed", tags=metric_tags)
-        logger.error(f"Snuba health check failed! Tags: {metric_tags}")
+        if not down_file_exists:
+            logger.error(f"Snuba health check failed! Tags: {metric_tags}")
 
     if status != 200 or down_file_exists:
         logger.info(payload)
--- a/snuba/utils/health_info.py
+++ b/snuba/utils/health_info.py
@@ -107,8 +107,10 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
     payload = json.dumps(body)
     if status != 200:
         metrics.increment("healthcheck_failed", tags=metric_tags)
-        if not down_file_exists:
-            logger.error(f"Snuba health check failed! Tags: {metric_tags}")
+        if down_file_exists:
+            logger.error("Snuba health check failed! Tags: %s", metric_tags)
+        else:
+            logger.info("Snuba health check failed! Tags: %s", metric_tags)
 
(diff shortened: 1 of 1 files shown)


snuba/utils/health_info.py
```diff
@@ -107,7 +107,8 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
     payload = json.dumps(body)
     if status != 200:
         metrics.increment("healthcheck_failed", tags=metric_tags)
-        logger.error(f"Snuba health check failed! Tags: {metric_tags}")
+        if not down_file_exists:
+            logger.error(f"Snuba health check failed! Tags: {metric_tags}")
```
volokluev, line 111: I would just info log it if the down file exists, it's useful log info, just not an error

untitaker:
I found this additional health_envoy check. If we already check the downfile for envoy this way, what other services need to react to the downfile? Could they use the same endpoint instead? https://github.com/getsentry/snuba/blob/120c358c88f313d999619ee1055c7afa33b44418/snuba/web/views.py#L212-L213

snuba/utils/health_info.py
```diff
@@ -107,7 +107,10 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
     payload = json.dumps(body)
     if status != 200:
         metrics.increment("healthcheck_failed", tags=metric_tags)
-        logger.error(f"Snuba health check failed! Tags: {metric_tags}")
+        if down_file_exists:
+            logger.error("Snuba health check failed! Tags: %s", metric_tags)
```
untitaker, line 111: Moving away from f-strings to logger-native string formatting. The reason we got a new issue at all is because we were interpolating the string before logging. Now we only get one issue instead of N for every possible tag combination.

snuba/utils/health_info.py
```diff
@@ -107,7 +107,10 @@ def get_health_info(thorough: Union[bool, str]) -> HealthInfo:
...
+        else:
+            logger.info("Snuba health check failed! Tags: %s", metric_tags)
 
     if status != 200 or down_file_exists:
         logger.info(payload)
```
MeredithAnya, line 116: @untitaker do we need this log then? if you added the check above?

snuba/utils/health_info.py
evanh: This works because then it's not a "new" issue in Sentry, is that correct?

evanh (approved):

snuba/utils/health_info.py
untitaker: yes that's the goal. now it's grouping by the literal string `"Snuba health check failed! Tags: %s"`

PR #6644, opened 2024-12-10
ref(lw-deletes): enforce ratelimiter

MeredithAnya:
Realized that the allocation policy was active but not enforced, so making it enforced here.

### DIFF
--- a/snuba/datasets/configuration/issues/storages/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/storages/search_issues.yaml
@@ -96,7 +96,7 @@ delete_allocation_policies:
         - referrer
         - project_id
       default_config_overrides:
-        is_enforced: 0
+        is_enforced: 1
 
 query_processors:
   - processor: MappingOptimizer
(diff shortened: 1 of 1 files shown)


snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -124,7 +124,7 @@ query_processors:
 
 deletion_settings:
   is_enabled: 1
-  max_rows_to_delete: 100000
+  max_rows_to_delete: 2000000
```
onkar, line 127: Is this change intentional? Are we saying that since rate limiter is enforced, we are increasing the default no. of rows to be deleted for search issues?

onkar (approved):

kylemumma (approved):

snuba/datasets/configuration/issues/storages/search_issues.yaml
MeredithAnya: this is intentional, it's to match the value passed in ops https://github.com/getsentry/ops/blob/27d30b56d219c1655d2fac9361f0d0a6eb4c708c/k8s/services/snuba/_values.yaml#L558

PR #6648, opened 2024-12-10
fix(scripts): Always create local tables before MV

untitaker:


john-z-yang (approved):

scripts/copy_tables.py
```diff
@@ -131,6 +131,9 @@ def copy_tables(
     if not tables:
         tables = [result[0] for result in source_client.execute("SHOW TABLES")]
 
+    # create local tables before MVs
+    tables.sort(key=lambda x: "_local" in x)
```
untitaker, line 135:
```suggestion
    tables.sort(key=lambda x: "_local" not in x)
```

false comes before true. so local tables need to evaluate to false.

PR #6649, opened 2024-12-10
hotfix(inc-984): Add manual job to scrub IPs from spans

volokluev:


snuba/manual_jobs/scrub_ips_from_spans.py
```diff
@@ -0,0 +1,49 @@
...
+        start_datetime = self._start_datetime.isoformat()
+        end_datetime = self._end_datetime.isoformat()
+        on_cluster = f"ON CLUSTER '{cluster_name}'" if cluster_name else ""
+        return f"""ALTER TABLE spans_local
+{on_cluster}
+UPDATE `sentry_tags.value` = arrayMap((k, v) -> if(k = 'user.ip', 'scrubbed', v), `sentry_tags.key`, `sentry_tags.value`)
+WHERE project_id IN [{project_ids}]
+AND has(`sentry_tags.key`, 'user.ip')
+AND end_timestamp > toDateTime('{start_datetime}')
+AND end_timestamp <= toDateTime('{end_datetime}')"""
```
onewland, line 33: I think `connection.execute` will do some interpolation for you, though I don't know if it's sophisticated enough to translate the project_ids array into a ClickHouse SQL array: 

See https://github.com/getsentry/snuba/blob/17a19ad9d020d25f847dd3fabd94f8383519a666/snuba/manual_jobs/update_migration_status.py#L78-L91 for an example.

Not a deal-killer, but I think it would be nice if we could get the custom interpolation out of here.

snuba/manual_jobs/scrub_ips_from_spans.py
```diff
@@ -0,0 +1,49 @@
...
+        connection = cluster.get_node_connection(
+            ClickhouseClientSettings.CLEANUP, storage_node
+        )
+        if not cluster.is_single_node():
+            cluster_name = cluster.get_clickhouse_cluster_name()
+        else:
+            cluster_name = None
+        query = self._get_query(cluster_name)
+        logger.info("Executing query: {query}")
+        result = connection.execute(query=query, settings={"mutations_sync": 0})
```
onewland, line 47: mutations_sync=0 means that this will probably return immediately (and there's no async check on this job).

That's fine but just want to make sure it's an explicit decision you made

snuba/manual_jobs/scrub_ips_from_spans.py
```diff
@@ -0,0 +1,49 @@
...
+WHERE project_id IN [{project_ids}]
+AND has(`sentry_tags.key`, 'user.ip')
+AND end_timestamp > toDateTime('{start_datetime}')
+AND end_timestamp <= toDateTime('{end_datetime}')"""
+
+    def execute(self, logger: JobLogger) -> None:
+        cluster = get_cluster(StorageSetKey.SPANS)
+        storage_node = cluster.get_local_nodes()[0]
+        connection = cluster.get_node_connection(
+            ClickhouseClientSettings.CLEANUP, storage_node
```
onewland, line 39: not disagreeing with the choice, but why use the `CLEANUP` setting?

snuba/manual_jobs/scrub_ips_from_spans.py
```diff
@@ -0,0 +1,49 @@
...
+        self._project_ids = params["project_ids"]
+        self._start_datetime = datetime.fromisoformat(params["start_datetime"])
+        self._end_datetime = datetime.fromisoformat(params["end_datetime"])
+
+    def _get_query(self, cluster_name: str | None) -> str:
+        project_ids = ",".join([str(p) for p in self._project_ids])
+        start_datetime = self._start_datetime.isoformat()
+        end_datetime = self._end_datetime.isoformat()
+        on_cluster = f"ON CLUSTER '{cluster_name}'" if cluster_name else ""
+        return f"""ALTER TABLE spans_local
```
evanh, line 27: This is going to update every single part in the table.

- Does the cluster even have the resources for that (memory/disk space etc.)
- Are there parts that are so big that they can't be merged anymore?

untitaker (changes requested):

onewland (approved):
Seems fine at a high level, I would definitely run this with timestamp boundaries so that only a single partition is affected by any update (I might even start with an hour window, then increase to a day, etc.)

snuba/manual_jobs/scrub_ips_from_spans.py
volokluev: yup it is an explicit design decision, this job can take days

snuba/manual_jobs/scrub_ips_from_spans.py
onewland: on the subject of future improvements: this could be more general by making `'user.ip'` a parameter, and then the job is "scrub arbitrary field from spans" (which could be further generalized if we took the cluster as a parameter).

snuba/manual_jobs/scrub_ips_from_spans.py
onewland: ok, but we could have an "async check" method on job that checks if `system.mutations` is empty, which would check when you load the snuba admin jobs page and leave it in "running" if it's not done or "finished" if the mutation finishes

fine if you don't want to deal with that, though

snuba/manual_jobs/scrub_ips_from_spans.py
volokluev: seemed appropriate, is there a different one I should use?

onewland: seems fine, I don't see any special query settings attached to it (which would be the only problem)

snuba/manual_jobs/scrub_ips_from_spans.py
onewland: I don't think `ALTER TABLE UPDATE` cares about max_merge_size, I think it may just rewrite the part in the same way OPTIMIZE ignores https://clickhouse.com/docs/en/operations/settings/merge-tree-settings#max-bytes-to-merge-at-max-space-in-pool

snuba/manual_jobs/scrub_ips_from_spans.py
volokluev: Yeah I thought about generalizing but It's not especially complicated to do later

untitaker (approved):

snuba/manual_jobs/scrub_ips_from_spans.py
volokluev: Actually scratch what I said, it's 510 GBs (I was looking at EAP spans by mistake)

PR #6652, opened 2024-12-11
fix: run sentry tests when RPC changes

colin-sentry:
This forces relevant sentry tests to run when we change RPC files.


wmak (commented):
Could we add the stats and alerts tests to the list as well

phacops (approved):

PR #6655, opened 2024-12-11
refactor: Remove hardcoded spans dataset references

colin-sentry:
Preparing for when we add other datatypes to EAP, we have to remove hardcoded references to the spans table and its structure

### DIFF
--- a/snuba/web/rpc/common/common.py
+++ b/snuba/web/rpc/common/common.py
@@ -1,30 +1,14 @@
 from datetime import datetime, timedelta
-from typing import Final, Mapping, Sequence, Set
 
-from sentry_protos.snuba.v1.request_common_pb2 import RequestMeta
-from sentry_protos.snuba.v1.trace_item_attribute_pb2 import (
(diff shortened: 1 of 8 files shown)


snuba/web/rpc/common/common.py
```diff
@@ -72,282 +56,6 @@ def transform(exp: Expression) -> Expression:
...
-}
-
-TIMESTAMP_COLUMNS: Final[Set[str]] = {
-    "sentry.timestamp",
-    "sentry.start_timestamp",
-    "sentry.end_timestamp",
-}
-
-
-def attribute_key_to_expression(attr_key: AttributeKey) -> Expression:
```
colin-sentry, line 103: this is the big problem with the current approach, the mapping of RPC keys to snuba expressions is different per-dataset, and we stuff it into this global function here.

phacops:
I'd rather see your approach based on entities, with every entity having the same shape but with different storage underneath and the required amount of column mappers.

PR #6657, opened 2024-12-11
feat(eap): add default value to virtual column

davidtsuk:
Closes https://github.com/getsentry/projects/issues/488


volokluev (approved):

snuba/web/rpc/common/common.py
kylemumma, line 1: should the comment on line 198 where it says `default` change

kylemumma (approved):

PR #6665, opened 2024-12-12
feat: Add LiteralRewriter query processor

untitaker:


snuba/query/processors/physical/literal_rewriter.py
```diff
@@ -0,0 +1,66 @@
...
+        if project_ids_setting_key is None:
+            self.__project_ids = None
+        else:
+            self.__project_ids = getattr(settings, project_ids_setting_key)
+
+    def process_query(self, query: Query, query_settings: QuerySettings) -> None:
+        project_ids = get_object_ids_in_query_ast(query, self.__project_column)
+
+        if self.__project_ids is not None:
+            if not project_ids & self.__project_ids:
```
davidtsuk, line 49: I'm guessing you didn't mean to use a bitwise & here?

onewland (commented):
~~ would this handle `group by tags["myTag"]` ? should it?~~ read it over again, this is handled

snuba/query/processors/physical/literal_rewriter.py
untitaker: sets overload that operator to mean intersect

snuba/query/processors/physical/literal_rewriter.py
```diff
@@ -0,0 +1,66 @@
...
+                return
+
+        def transform_literal(exp: Expression) -> Expression:
+            if not isinstance(exp, Literal):
+                return exp
+
+            if not isinstance(exp.value, str):
+                return exp
+
+            try:
```
evanh, line 59: Why not do `if exp.value in self.__literal_mappings`?

evanh (commented):
You'll need to add this processor to the spans config as well.

onewland (approved):
Looks good to me, though I think a person who has dealt with QueryProcessors more recently may be better able to spot a problem

snuba/query/processors/physical/literal_rewriter.py
```diff
@@ -0,0 +1,66 @@
...
+from snuba.query.expressions import Expression, Literal
+from snuba.query.processors.physical import ClickhouseQueryProcessor
+from snuba.query.query_settings import QuerySettings
+
+
+class LiteralRewriter(ClickhouseQueryProcessor):
+    """
+    Removes any string literals by value and replaces them with another value,
+    for the given (optional) project IDs. If no project IDs are provided, all
+    projects will be matched.
```
volokluev, line 14: should be no projects match if none provided

snuba/query/processors/physical/literal_rewriter.py
davidtsuk: oh so the list of project ids in settings needs to be a set?

enochtangg (commented):
So if the new tag we're rewritting to does not exist, the query will fail. Is that intended?

untitaker:
> So if the new tag we're rewritting to does not exist, the query will fail. Is that intended?

the intent is that the query behaves as if the tag didn't exist.

evanh (approved):

volokluev (approved):

evanh (approved):

kylemumma (approved):
makes sense to me

davidtsuk (approved):

snuba/query/processors/physical/literal_rewriter.py
untitaker: I screwed up a bit here, fixed

untitaker:
we decided not to do this and focus on hard removal

PR #6667, opened 2024-12-12
Implement filter offset for attribute values API

xurui-c:
https://github.com/getsentry/eap-planning/issues/115

Already did it for attribute names API, doing it for attribute values API in this PR. Also sorted attribute values alphabetically in response so filter offset can be implemented

### DIFF
--- a/snuba/web/rpc/v1/trace_item_attribute_values.py
+++ b/snuba/web/rpc/v1/trace_item_attribute_values.py
@@ -7,6 +7,11 @@ from sentry_protos.snuba.v1.endpoint_trace_item_attributes_pb2 import (
     TraceItemAttributeValuesResponse,
 )
 from sentry_protos.snuba.v1.request_common_pb2 import PageToken
+from sentry_protos.snuba.v1.trace_item_attribute_pb2 import AttributeKey, AttributeValue
+from sentry_protos.snuba.v1.trace_item_filter_pb2 import (
+    ComparisonFilter,
+    TraceItemFilter,
+)
 
 from snuba.attribution.appid import AppID
 from snuba.attribution.attribution_info import AttributionInfo
@@ -17,6 +22,7 @@ from snuba.query import OrderBy, OrderByDirection, SelectedExpression
 from snuba.query.data_source.simple import Entity
 from snuba.query.dsl import Functions as f
 from snuba.query.dsl import column, literal, literals_array
+from snuba.query.expressions import Expression
 from snuba.query.logical import Query
 from snuba.query.query_settings import HTTPQuerySettings
 from snuba.request import Request as SnubaRequest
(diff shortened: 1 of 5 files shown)


snuba/web/rpc/v1/trace_item_attribute_values.py
```diff
@@ -50,26 +96,8 @@ def _build_query(request: TraceItemAttributeValuesRequest) -> Query:
...
-            ),
-        )
-        if request.value_substring_match is not None
-        else base_conditions_and(
-            request.meta,
-            f.equals(column("attr_key"), literal(request.key.name)),
-        ),
+        condition=_build_base_conditions_and(request),
         order_by=[
-            OrderBy(
```
kylemumma, line 69: why did you remove this order by?

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
```diff
@@ -147,15 +146,34 @@ def test_page_token(self, setup_teardown: Any) -> None:
         assert expected_values == []
 
     def test_filter_offset_is_not_implemented_right_now(self) -> None:
```
kylemumma, line 148: can you change the name of the test to reflect its new behavior

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
```diff
@@ -147,15 +146,34 @@ def test_page_token(self, setup_teardown: Any) -> None:
...
+        while done < total_number_of_values:
+            req = TraceItemAttributeValuesRequest(
+                meta=COMMON_META,
+                limit=at_a_time,
+                key=AttributeKey(name="tag1", type=AttributeKey.TYPE_STRING),
+                value_substring_match="",
+                page_token=page_token,
+            )
+            res = AttributeValuesRequest().execute(req)
+            page_token = res.page_token
```
kylemumma, line 172: can you please add an assert to make sure that this is using `filter_offset` as opposed to `offset`

kylemumma (commented):
My main concern is that your test could be using `offset` by default rather than `filter_offset`

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
volokluev: that breaks backwards compatibility. I would prefer not to do that

kylemumma: > Kyle Mumma
>wym
>
>Volo Kluev
>The page token can be an filter offset or a number. If we assert that it's a filter offset, we break the >client that could have been using the numbers
>
>Kyle Mumma
>if no page_token is provided by the user, we should default to filter_offset
>otherwise we should continue using the type they provided
>
>Volo Kluev
>If that's what's happening then all good
>
>Kyle Mumma
>Since the test begins by not providing any page_token, it should default to filter_offset so it should >be ok to assert that
>
>Volo Kluev
>:+1:

snuba/web/rpc/v1/trace_item_attribute_values.py
xurui-c: addressed in Slack

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
```diff
@@ -146,16 +145,36 @@ def test_page_token(self, setup_teardown: Any) -> None:
...
+            assert isinstance(page_token, PageToken)
```
kylemumma, line 173: basically this page_token could be an offset ex: `10000` or a filter_offset ex `attr_name > ...`

I am worried that `page_token` is by default a offset rather than filter_offset, so this test might not actually be testing your implementation of filter_offset

I would want an assert something like `assert res.WhichOneOf("page_token") == "filter_offset"` so we know its using the filter_offset and not offset

Hope that makes sense please ping me if it didnt or you disagree

kylemumma (approved):
LGTM assuming the tests pass

kylemumma (approved):

PR #6671, opened 2024-12-13
feat(eap): Add a GetTraces endpoint

phacops:
This is a possible implementation of the new FindTraces endpoint.

https://github.com/getsentry/sentry-protos/pull/71

Closes https://github.com/getsentry/eap-planning/issues/122.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -46,4 +46,4 @@ sqlparse==0.5.0
 google-api-python-client==2.88.0
 sentry-usage-accountant==0.0.11
 freezegun==1.2.2
-sentry-protos @ git+ssh://git@github.com/getsentry/sentry-protos.git@pierre/eap-find-traces#subdirectory=py
+sentry-protos @ git+https://github.com/getsentry/sentry-protos@pierre/eap-find-traces#subdirectory=py
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,6 +29,7 @@ python-rapidjson==1.8
 redis==4.5.4
 sentry-arroyo==2.19.3
 sentry-kafka-schemas==0.1.122
+sentry-protos==0.1.41
 sentry-redis-tools==0.3.0
 sentry-relay==0.9.2
 sentry-sdk==2.18.0
@@ -46,4 +47,3 @@ sqlparse==0.5.0
 google-api-python-client==2.88.0
 sentry-usage-accountant==0.0.11
 freezegun==1.2.2
--e file:///Users/phacops/code/sentry-protos/py#sentry-protos
--- a/requirements.txt
+++ b/requirements.txt
@@ -46,4 +46,4 @@ sqlparse==0.5.0
 google-api-python-client==2.88.0
 sentry-usage-accountant==0.0.11
 freezegun==1.2.2
-sentry-protos==0.1.40
+sentry-protos @ git+ssh://git@github.com/getsentry/sentry-protos.git@pierre/eap-find-traces#subdirectory=py
--- a/requirements.txt
+++ b/requirements.txt
@@ -46,4 +46,4 @@ sqlparse==0.5.0
 google-api-python-client==2.88.0
 sentry-usage-accountant==0.0.11
 freezegun==1.2.2
-sentry-protos @ git+https://github.com/getsentry/sentry-protos@pierre/eap-find-traces#subdirectory=py
+-e file:///Users/phacops/code/sentry-protos/py#sentry-protos
--- a/snuba/web/rpc/v1/endpoint_get_traces.py
+++ b/snuba/web/rpc/v1/endpoint_get_traces.py
@@ -6,20 +6,21 @@ from google.protobuf.json_format import MessageToDict
 from sentry_protos.snuba.v1.endpoint_get_traces_pb2 import (
     GetTracesRequest,
     GetTracesResponse,
-    TraceColumn,
+    TraceAttribute,
 )
-from sentry_protos.snuba.v1.request_common_pb2 import PageToken
+from sentry_protos.snuba.v1.request_common_pb2 import PageToken, TraceItemName
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import AttributeKey, AttributeValue
+from sentry_protos.snuba.v1.trace_item_filter_pb2 import TraceItemFilter
 
 from snuba.attribution.appid import AppID
 from snuba.attribution.attribution_info import AttributionInfo
 from snuba.datasets.entities.entity_key import EntityKey
 from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.pluggable_dataset import PluggableDataset
-from snuba.query import OrderBy, OrderByDirection, SelectedExpression
+from snuba.query import LimitBy, OrderBy, OrderByDirection, SelectedExpression
 from snuba.query.data_source.simple import Entity
 from snuba.query.dsl import Functions as f
-from snuba.query.dsl import column
+from snuba.query.dsl import and_cond, column, in_cond, literal, literals_array
 from snuba.query.expressions import Expression
 from snuba.query.logical import Query
 from snuba.query.query_settings import HTTPQuerySettings
@@ -28,6 +29,8 @@ from snuba.web.query import run_query
 from snuba.web.rpc import RPCEndpoint
 from snuba.web.rpc.common.common import (
     base_conditions_and,
+    project_id_and_org_conditions,
+    timestamp_in_range_condition,
     trace_item_filters_to_expression,
     treeify_or_and_conditions,
 )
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -46,4 +46,4 @@ sqlparse==0.5.0
 google-api-python-client==2.88.0
 sentry-usage-accountant==0.0.11
 freezegun==1.2.2
-sentry-protos==0.1.40
+sentry-protos @ git+https://github.com/getsentry/sentry-protos@pierre/eap-find-traces#subdirectory=py
```
Zylphrex, line 49: I assume this will be bumped once getsentry/sentry-protos#71 is merged + released?

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,289 @@
...
+    return PageToken(offset=request.page_token.offset + num_rows)
+
+
+def _validate_order_by(in_msg: GetTracesRequest) -> None:
+    order_by_cols = set([ob.column.name for ob in in_msg.order_by])
+    selected_columns = set([c.name for c in in_msg.columns])
+    if not order_by_cols.issubset(selected_columns):
+        raise BadSnubaRPCRequestException(
+            f"Ordered by columns {order_by_cols} not selected: {selected_columns}"
+        )
```
Zylphrex, line 250: Is the intention here to allow ordering by any of the selected columns? The 2 main ones we'd like to have will be by timestamp (newest/oldest trace) and by duration (shortest/longest trace) though this second one is harder to define clearly as the trace duration is as it's becoming closer to a session depending on the platform.

I'm not opposed to just supporting ordering by timestamp first and we can add more later once we more clearly define them.

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,289 @@
...
+            f.min(column("start_timestamp")),
+            _TYPES_TO_CLICKHOUSE[trace_column.type],
+            alias=_COLUMN_TO_NAME[trace_column.name],
+        )
+    if trace_column.name == TraceColumn.Name.NAME_ROOT_SPAN_NAME:
+        return f.anyIf(
+            column("name"),
+            f.equals(column("is_segment"), True),
+            alias=_COLUMN_TO_NAME[trace_column.name],
+        )
```
Zylphrex, line 93: This has a chance of returning `null` due to our data model. Can the first option we add be the name of the first span which is guaranteed to exist and we add the rest as needed?

So something like `argMin(name, start_timestamp)`

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,289 @@
...
+        key=EntityKey("eap_spans"),
+        schema=get_entity(EntityKey("eap_spans")).get_data_model(),
+        sample=None,
+    )
+    res = Query(
+        from_clause=entity,
+        selected_columns=selected_columns,
+        condition=base_conditions_and(
+            request.meta,
+        ),
```
Zylphrex, line 148: This isn't applying the user conditions. Wouldn't that result in a full scan of the table across all orgs? Or am I missing something?

phacops: Yes, this is a mistake.

snuba/web/rpc/v1/endpoint_get_traces.py
phacops: We'll have to modify the protobuf files to add `FIRST_SEEN_SPAN_NAME` and actually do this properly.

For now, I'll just make the behavior of `ROOT_SPAN_NAME` to be `FIRST_SEEN_SPAN_NAME` and we'll iterate.

snuba/web/rpc/v1/endpoint_get_traces.py
phacops: Yes, that's the intention. We can have a default sort to timestamp and you'll see later if you use custom sorts.

tests/web/rpc/v1/test_endpoint_get_traces.py
```diff
@@ -0,0 +1,448 @@
...
+                            value=AttributeValue(
+                                val_str=trace_id,
+                            ),
+                        ),
+                    ],
+                )
+                for trace_id in sorted(_TRACE_IDS)
+            ],
+            page_token=PageToken(offset=len(_TRACE_IDS)),
+            meta=ResponseMeta(request_id="be3123b3-2e5d-4eb9-bb48-f38eaa9e8480"),
```
kylemumma, line 225: maybe extract `be3123b3-...` into a variable since it is also used in the request

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,405 @@
...
+            page_token=_get_page_token(in_msg, traces),
+            meta=response_meta,
+        )
+
+    def _list_trace_ids(
+        self,
+        request: GetTracesRequest,
+    ) -> dict[str, int]:
+        # Find first span filter.
+        # TODO: support more than one filter.
```
kylemumma, line 278: should you throw `NotImplementedError` if `len(request.filters) > 1` ?

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,405 @@
...
+            [self._timer],
+        )
+
+        # Get a dict of trace IDs and timestamps.
+        trace_ids = self._list_trace_ids(request=in_msg)
+        if len(trace_ids) == 0:
+            return GetTracesResponse(meta=response_meta)
+
+        # Get metadata for those traces.
+        traces = self._get_metadata_for_traces(request=in_msg, trace_ids=trace_ids)
```
kylemumma, line 266: why do you first query for trace_ids, and then do another query for the attributes, rather than doing them together in 1 query?

kylemumma (approved):
lgtm

snuba/web/rpc/v1/endpoint_get_traces.py
phacops: For performance and because we can't compute all attributes in one query. It's much faster to select a list of trace IDS and timestamps matching the conditions and then to compute the appropriate attributes on the list.

Right now, the UI will query all the attributes at once, and I can't compute `FILTERED_ITEM_COUNT` and `TOTAL_ITEM_COUNT` in one query since, in order to count the appropriate items matching some conditions, I can't pass them in the `WHERE` clause, I need to use a `countIf` without a `WHERE`.

We could optimize later on by not doing 2 queries depending on the attributes that are requested.

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,405 @@
...
+            condition=base_conditions_and(
+                request.meta,
+                trace_item_filters_expression,
+            ),
+            order_by=[
+                OrderBy(
+                    direction=OrderByDirection.DESC,
+                    expression=column("_sort_timestamp"),
+                ),
+            ],
```
Zylphrex, line 336: This completely ignores the order by specified by the request. I'm not sure how you'd implement the rest but I'm fine if you just validate it to only allow sorting by timestamp for now.

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,418 @@
...
+            trace_ids[row["trace_id"]] = row["_sort_timestamp"]
+        return trace_ids
+
+    def _get_metadata_for_traces(
+        self,
+        request: GetTracesRequest,
+        trace_ids: dict[str, int],
+    ) -> list[GetTracesResponse.Trace]:
+        trace_item_filters_expression = trace_item_filters_to_expression(
+            request.filters[0].filter if len(request.filters) > 0 else TraceItemFilter()
```
Zylphrex, line 359: did you want to wrap this in `_select_supported_filters` as well? not necessary since the first one already validate it but lets you get rid of this inline ternary

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,418 @@
...
+    extract_response_meta,
+    setup_trace_query_settings,
+)
+from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
+
+_DEFAULT_ROW_LIMIT = 10_000
+
+_ATTRIBUTES: dict[
+    TraceAttribute.Key.ValueType,
+    tuple[str, AttributeKey.Type.ValueType],
```
Zylphrex, line 48: Nit: a named tuple/data class/typed dict would be a little cleaner here so you don't have to access it by index

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,418 @@
...
+        timestamps = trace_ids.values()
+        query = Query(
+            from_clause=entity,
+            selected_columns=selected_columns,
+            condition=and_cond(
+                project_id_and_org_conditions(request.meta),
+                timestamp_in_range_condition(
+                    min(timestamps) - 2 * 3600,
+                    max(timestamps) + 2 * 3600,
+                ),
```
Zylphrex, line 388: Be nice if the buffer is modifiable via an option of some sort in prod without a deploy but I'm not sure how this works in snuba.

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -0,0 +1,418 @@
...
+            ),
+            groupby=[
+                _attribute_to_expression(
+                    TraceAttribute(
+                        key=TraceAttribute.Key.KEY_TRACE_ID,
+                    ),
+                ),
+            ],
+            order_by=_convert_order_by(request.order_by),
```
Zylphrex, line 407: Applying the order by at this point isn't very helpful as it's only sorting the trace ids results from the first request. But i guess it'd look sorted?

I'm inclined to lean towards just
1. restrict sorts to just timestamp
2. then ensure the final results are sorted by timestamp by sorting it again in clickhouse or python

and not worry about other attributes for now

Zylphrex (approved):
Mostly nitpicks but looks good to me overall.

onkar (approved):

PR #6677, opened 2024-12-13
ref(lw-deletes): add project_id killswitch and some logging

MeredithAnya:
This PR has a couple improvements and additions to the lw delete pipeline, the main one being the killswitch for the project. 

Most storages should have the `project_id` as one of the allowed columns so the assumption right now is that they do and so the killswitch will work for that case. If we add more storages in the future that don't have that, we can update the killswitch logic. The config will be based on the storage:

`lw_deletes_killswitch_search_issues`

The killswitch is added at the API level because the moment it's produced to the consumer it's batched with other messages, most likely with other `project_id`'s. Another option would be to add another strategy step prior to the batch step that filters based on the killswitch, but I felt like not producing to the topic at all made more sense right now.

### DIFF
--- a/snuba/datasets/configuration/issues/storages/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/storages/search_issues.yaml
@@ -94,6 +94,7 @@ delete_allocation_policies:
     args:
       required_tenant_types:
         - referrer
+        - project_id
       default_config_overrides:
         is_enforced: 1
 
--- a/snuba/datasets/configuration/issues/storages/search_issues.yaml
+++ b/snuba/datasets/configuration/issues/storages/search_issues.yaml
@@ -94,7 +94,6 @@ delete_allocation_policies:
     args:
       required_tenant_types:
         - referrer
-        - project_id
       default_config_overrides:
         is_enforced: 1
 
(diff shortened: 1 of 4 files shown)


snuba/datasets/configuration/issues/storages/search_issues.yaml
```diff
@@ -94,7 +94,6 @@ delete_allocation_policies:
     args:
       required_tenant_types:
         - referrer
-        - project_id
```
MeredithAnya, line 97: This is intentional - now that we have the batching approach, the allocation policy won't have a `project_id` because it will be (potentially) multiple `project_id`s, so we can ratelimit based on that. This allocation policy will work similar to the subscriptions executor where there is one referrer for those queries. Though the referrer below will be based on storage.

snuba/lw_deletions/strategy.py
```diff
@@ -62,19 +65,25 @@ def submit(self, message: Message[ValuesBatch[KafkaPayload]]) -> None:
...
+                raise MessageRejected
 
         self.__next_step.submit(message)
 
     def _get_attribute_info(self) -> AttributionInfo:
         return AttributionInfo(
             app_id=AppID("lw-deletes"),
             tenant_ids={},
-            referrer="lw-deletes",
+            referrer=f"lw-deletes-{self.__storage_name}",
```
MeredithAnya, line 86: There are no tenant ids passed from the batch of messages, just the referrer which I've added to be by storage so that we can change the allocation policy per storage

MeredithAnya: Although now that I think about it, the policy is defined on the storage... so this might just be redundant

MeredithAnya: yeah this is redundant, the policies are done by the storage

armenzg (approved):

onkar (approved):

PR #6678, opened 2024-12-16
feat(eap): Use weighted average instead of simple average for calculating average sampling rate

davidtsuk:
Also updating the count column to use sign
Closes https://github.com/getsentry/eap-planning/issues/133

### DIFF
--- a/snuba/web/rpc/common/aggregation.py
+++ b/snuba/web/rpc/common/aggregation.py
@@ -252,8 +252,11 @@ def _get_count_column_alias(aggregation: AttributeAggregation) -> str:
 
 
 def get_count_column(aggregation: AttributeAggregation) -> Expression:
-    field = attribute_key_to_expression(aggregation.key)
-    return f.count(field, alias=_get_count_column_alias(aggregation))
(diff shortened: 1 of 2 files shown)


snuba/web/rpc/common/aggregation.py
```diff
@@ -233,9 +233,10 @@ def get_average_sample_rate_column(aggregation: AttributeAggregation) -> Express
         referenced_column=aggregation.label,
         metadata={},
     ).to_alias()
-    return f.avgIf(
-        f.divide(literal(1), sampling_weight_column),
-        get_field_existence_expression(aggregation),
+    field = attribute_key_to_expression(aggregation.key)
+    return f.divide(
+        f.count(field),
```
colin-sentry, line 238: shouldn't this be

sumIf(sign, exists_expression) / sumIf(sign * sampling_weight, exists_expression)

colin-sentry: also should probably have a test!

kylemumma (approved):

PR #6679, opened 2024-12-16
ref(lw-deletes): concurrent allocation policy requires org id

MeredithAnya:
Ran into `InvalidTenantsForAllocationPolicy` because I wasn't passing in a `project_id` or an `organization_id`

Used `project_id` since that is what I had specified in the storage definition https://github.com/getsentry/snuba/blob/90a2d471ec64dafe282b406452f5791ed5893123/snuba/datasets/configuration/issues/storages/search_issues.yaml#L97

FIXES [SNUBA-6H4](https://sentry.sentry.io/issues/6147044083/?project=300688)

### DIFF
--- a/snuba/lw_deletions/strategy.py
+++ b/snuba/lw_deletions/strategy.py
@@ -74,7 +74,7 @@ class FormatQuery(ProcessingStrategy[ValuesBatch[KafkaPayload]]):
         return AttributionInfo(
             app_id=AppID("lw-deletes"),
             # concurrent allocation policies requires project or org id
-            tenant_ids={"organization_id": 1},
+            tenant_ids={"project_id": 1},
             referrer="lw-deletes",
             team=None,
             feature=None,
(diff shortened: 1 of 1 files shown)


snuba/lw_deletions/strategy.py
```diff
@@ -73,7 +73,8 @@ def submit(self, message: Message[ValuesBatch[KafkaPayload]]) -> None:
     def _get_attribute_info(self) -> AttributionInfo:
         return AttributionInfo(
             app_id=AppID("lw-deletes"),
-            tenant_ids={},
+            # concurrent allocation policies requires project or org id
+            tenant_ids={"project_id": 1},
```
armenzg, line 77: Why 1?

armenzg (approved):

snuba/lw_deletions/strategy.py
MeredithAnya: we are really just ratelimiting on the referrer, and so keeping the project hardcoded works here. Its the same thing we do in the subscriptions executor subscription. I can't pass in user project ids because the query will contain multiple projects and multiple organizations. And furthermore what we really want to limit is the number of concurrent mutations on the system

onkar (approved):

PR #6680, opened 2024-12-16
fix(admin): Allow special characters in SYSTEM/OPTIMIZE queries

evanh:
These queries sometimes have to reference table names or clusters with quotes or other special
characters.


volokluev (approved):

untitaker (approved):
add `\`` too?

PR #6681, opened 2024-12-16
fix(eap): Fix divide by 0 errors caused when the sample count is 0

davidtsuk:
Fixes https://github.com/getsentry/eap-planning/issues/132


onkar (approved):
Should there be a unit test to assert the `None` response if there is no data to extrapolate i.e. `sample_count` is `0`.

PR #6686, opened 2024-12-18
feat(uptime): snuba table and configuration for new dataset

JoshFerge:
Context: https://github.com/getsentry/team-uptime/issues/23

migration PR: https://github.com/getsentry/snuba/pull/6690


rust_snuba/src/processors/uptime_monitor_checks.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+pub fn deserialize_message(
+    payload: &[u8],
+    partition: u16,
+    offset: u64,
+) -> anyhow::Result<(Vec<UptimeMonitorCheckRow>, f64)> {
+    let monitor_message: UptimeMonitorCheckMessage = serde_json::from_slice(payload)?;
+
+    let rows = vec![UptimeMonitorCheckRow {
```
JoshFerge, line 29: relevant upstream code: https://github.com/getsentry/uptime-checker/blob/e120f3ff6856d5f47cabb883efa11fae03c8c7b6/src/types/result.rs#L61

need to tweak this consumer before turning on.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    # do i actually need primary key to be different than sorting key?
+                    primary_key="(organization_id, project_id, uptime_subscription_id, _sort_timestamp, uptime_check_id, trace_id)",
+                    order_by="(organization_id, project_id, uptime_subscription_id, _sort_timestamp, uptime_check_id, trace_id)",
```
JoshFerge, line 58: for our query pattern, it probably makes sense to include project_id and subscription_id in the order_by key. and uptime_check_id corresponds to event_id.

also noting that when we update uptime with regions, this may need to switch to uptime_region_check_id for this. (trace_id makes these unique anyways in the meantime)

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import AddIndicesData, OperationTarget, SqlOperation
+
+storage_set = StorageSetKey.UPTIME_MONITOR_CHECKS
+table_prefix = "uptime_monitor_checks"
+local_table_name = f"{table_prefix}_local"
+dist_table_name = f"{table_prefix}_dist"
+
+## what about all the fancy codecs? do we need those?
```
JoshFerge, line 14: should i use all the codecs that the spans table uses? or does it matter?

JoshFerge:
should i split this PR up? what would be the right way to do that? could i leave the processor off the config to start?

scripts/load_uptime_checks.py
evanpurkhiser, line 1: What's the script for?

JoshFerge: i'm using it to load in data locally to test row size, compression ratio, and querying the data using EAP API.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("uptime_subscription_id", UInt(64)),
```
JoshFerge, line 19: what about  uptime_subscription_region_id or project_uptime_subscription_id?

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+    Column("uptime_check_id", UUID()),
+    Column("scheduled_check_time", DateTime()),
+    Column("timestamp", DateTime()),
+    Column("_sort_timestamp", DateTime()),
+    Column("duration", UInt(64)),
+    Column("region_id", UInt(16, Modifiers(nullable=True))),
```
JoshFerge, line 25: maybe this should be region slug? or need to increase to UInt 64? TBD.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    # do i actually need primary key to be different than sorting key?
+                    primary_key="(organization_id, project_id, uptime_subscription_id, _sort_timestamp, uptime_check_id, trace_id)",
```
JoshFerge, line 57: do we actually need trace_id here?

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+                    order_by="(organization_id, project_id, uptime_subscription_id, _sort_timestamp, uptime_check_id, trace_id)",
+                    partition_by="(toMonday(_sort_timestamp))",
```
JoshFerge, line 59: why only timestamp? should we also include org_id/project_id?

snuba/utils/streams/topics.py
```diff
@@ -45,6 +45,7 @@ class Topic(Enum):
     PROFILE_CHUNKS = "snuba-profile-chunks"
 
     REPLAYEVENTS = "ingest-replay-events"
+    UPTIME_MONITOR_CHECKS = "snuba-uptime-monitor-checks"
```
JoshFerge, line 48: change this to uptime results

JoshFerge: need to actually figure out if we need to enrich stuff from postgres, or we modify uptime checker

phacops (changes requested):
You should put the `storage` config in the EAP and you won't need an entity since you're going to use the EAP RPC.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
phacops: It should be partition by `(retention_days, toMonday(timestamp))` so we can drop the part at once.

As for not putting the organization_id or project_id in there, it's to not have hundreds of thousands of partitions.
```A merge only works for data parts that have the same value for the partitioning expression. This means you shouldn’t make overly granular partitions (more than about a thousand partitions). Otherwise, the SELECT query performs poorly because of an unreasonably large number of files in the file system and open file descriptors.
```
https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/custom-partitioning-key

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
phacops: The EAP RPC has a way to query by `trace_id`. If you're not putting a `trace_id` somewhere, the requests on this endpoint will be slower.

Also, you don't need to set the sort key and the primary key. What we usually want is the primary key to be as small as possible (it's in RAM) and there's no need to be overly specific since, at some point, there will be so little data compared to what needs to be scanned.

In your case, I would put the timestamp right after `project_id` because, after this, how many `uptime_subscription_id` will you have? if it's something like under 8k, it might fit into one part and scanning the primary key then the part is not faster than just the part.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
phacops: Are we going to have that many regions? Perhaps a LowCardinality(String), not null (make it a `global` region instead of `null`) is better and with the automatic compression (`ZSTD(1)` by default), this will be a very small column.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
phacops: Start with no codec and we'll see where we need them. By default, everything is `ZSTD(1)`.

phacops:
> should i split this PR up? what would be the right way to do that?

You need to split the PRs as Snuba won't let you add a migration with other code. Put the migration in another PR and keep the configuration and the processors here.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
JoshFerge: yeah, we're also discussing on the uptime team side if we want to just use a slug everywhere for the region. will update this.

phacops: `Nullable` makes ClickHouse maintains 2 columns instead of 1 with one column just saying if it's null or not. It's a bit wasteful. If we can avoid it, we should.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
JoshFerge: >The EAP RPC has a way to query by trace_id. If you're not putting a trace_id somewhere, the requests on this endpoint will be slower.

i've added a bloom filter index in the migration.

>Also, you don't need to set the sort key and the primary key. What we usually want is the primary key to be as small as possible (it's in RAM) and there's no need to be overly specific since, at some point, there will be so little data compared to what needs to be scanned.

thank you. modified.

JoshFerge:
> > should i split this PR up? what would be the right way to do that?
> 
> You need to split the PRs as Snuba won't let you add a migration with other code. Put the migration in another PR and keep the configuration and the processors here.

migration moved to https://github.com/getsentry/snuba/pull/6690/f

JoshFerge:
> You should put the `storage` config in the EAP and you won't need an entity since you're going to use the EAP RPC.

done.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
JoshFerge: got it. modified. thanks!

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
JoshFerge: 👍

snuba/migrations/groups.py
```diff
@@ -178,6 +181,11 @@ def __init__(
         storage_sets_keys={StorageSetKey.PROFILE_CHUNKS},
         readiness_state=ReadinessState.PARTIAL,
     ),
+    MigrationGroup.UPTIME_MONITOR_CHECKS: _MigrationGroup(
```
JoshFerge, line 184: TODO: determine if we should change the `StorageSetKey` here

JoshFerge:
are there any further tests i need to add? do i need to add a test in python similar to https://github.com/getsentry/snuba/blob/2313872f001fbe6241f2f60c4cdfd846aaf0635a/tests/datasets/test_spans_processor.py#L1?

tests/admin/dead_letter_queue/test_dlq.py
```diff
@@ -2,4 +2,4 @@
 
 
 def test_dlq() -> None:
-    assert len(get_dlq_topics()) == 9
```
untitaker, line 5: Feel free to delete this test, IMO that's just annoying to deal with as you add new datasets.

JoshFerge:
closing as this wasn't in the right order

PR #6690, opened 2024-12-18
feat(uptime): add initial table migration

JoshFerge:
broken out from https://github.com/getsentry/snuba/pull/6686

context: https://github.com/getsentry/team-uptime/issues/23

### DIFF
--- a/snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
+++ b/snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
@@ -1,0 +1,96 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import UUID, Column, DateTime, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import AddIndicesData, OperationTarget, SqlOperation
+
+storage_set = StorageSetKey.UPTIME_MONITOR_CHECKS
+table_prefix = "uptime_monitor_checks"
+local_table_name = f"{table_prefix}_local"
+dist_table_name = f"{table_prefix}_dist"
+
+## what about all the fancy codecs? do we need those?
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("uptime_subscription_id", UInt(64)),
+    Column("uptime_check_id", UUID()),
+    Column("scheduled_check_time", DateTime()),
+    Column("timestamp", DateTime()),
+    Column("_sort_timestamp", DateTime()),
+    Column("duration", UInt(64)),
+    Column("region_id", UInt(16, Modifiers(nullable=True))),
+    Column("check_status", String(Modifiers(low_cardinality=True))),
+    Column(
+        "check_status_reason",
+        String(Modifiers(nullable=True, low_cardinality=True)),
+    ),
+    Column("http_status_code", UInt(16)),
+    Column("trace_id", UUID()),
+    Column("retention_days", UInt(16)),
+]
+
+indices: Sequence[AddIndicesData] = [
+    AddIndicesData(
+        name="bf_trace_id",
+        expression="trace_id",
+        type="bloom_filter",
+        granularity=1,
+    )
+]
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    # do i actually need primary key to be different than sorting key?
+                    primary_key="(organization_id, project_id, _sort_timestamp, uptime_check_id)",
+                    order_by="(organization_id, project_id, uptime_subscription_id, _sort_timestamp, uptime_check_id)",
+                    partition_by="(retention_days, toMonday(_sort_timestamp))",
+                    settings={"index_granularity": "8192"},
+                    storage_set=storage_set,
+                    ttl="_sort_timestamp + toIntervalDay(retention_days)",
+                ),
+                target=OperationTarget.LOCAL,
+            ),
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=dist_table_name,
+                columns=columns,
+                engine=table_engines.Distributed(
+                    local_table_name=local_table_name,
+                    sharding_key="cityHash64(reinterpretAsUInt128(trace_id))",
+                ),
+                target=OperationTarget.DISTRIBUTED,
+            ),
+            operations.AddIndices(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                indices=indices,
+                target=OperationTarget.LOCAL,
+            ),
+        ]
+
+    def backwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.DropTable(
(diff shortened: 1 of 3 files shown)


snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,96 @@
...
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("uptime_subscription_id", UInt(64)),
+    Column("uptime_check_id", UUID()),
+    Column("scheduled_check_time", DateTime()),
+    Column("timestamp", DateTime()),
+    Column("_sort_timestamp", DateTime()),
+    Column("duration", UInt(64)),
+    Column("region_id", UInt(16, Modifiers(nullable=True))),
```
JoshFerge, line 25: TODO: make non-nullable / a low cardinality string

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,95 @@
...
+## what about all the fancy codecs? do we need those?
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("uptime_subscription_id", UInt(64)),
+    Column("uptime_check_id", UUID()),
+    Column("scheduled_check_time", DateTime()),
+    Column("timestamp", DateTime()),
+    Column("_sort_timestamp", DateTime()),
```
phacops, line 23: You don't need to have this column, your `timestamp` will be what you use. This is a compromise we made in the spans schema but it shouldn't be copied.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,95 @@
...
+    Column("duration", UInt(64)),
```
phacops, line 24: Indicate the unit in the field name.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    primary_key="(organization_id, project_id, _sort_timestamp, uptime_check_id)",
+                    order_by="(organization_id, project_id, uptime_subscription_id, _sort_timestamp, uptime_check_id)",
```
phacops, line 57: The sort key needs to contain the primary key, so `uptime_subscription_id` needs to be removed here.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,95 @@
...
+    Column(
+        "check_status_reason",
+        String(Modifiers(nullable=True, low_cardinality=True)),
+    ),
+    Column("http_status_code", UInt(16)),
+    Column("trace_id", UUID()),
+    Column("retention_days", UInt(16)),
+]
+
+indices: Sequence[AddIndicesData] = [
```
phacops, line 36: I wouldn't start with adding indices already. If you don't want to put the `trace_id` in the sort key, fine, but this might not be necessary.

JoshFerge: i've added the trace_id to the sort key / primary key

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,79 @@
+from typing import List, Sequence
+
+from snuba.clickhouse.columns import UUID, Column, DateTime, String, UInt
+from snuba.clusters.storage_sets import StorageSetKey
+from snuba.migrations import migration, operations, table_engines
+from snuba.migrations.columns import MigrationModifiers as Modifiers
+from snuba.migrations.operations import OperationTarget, SqlOperation
+
+storage_set = StorageSetKey.UPTIME_MONITOR_CHECKS
```
phacops, line 9: I think you need to use the EAP storage set.

You can keep the migration group as uptime monitor checks.

JoshFerge:
````
The storage_set=StorageSetKey.EVENTS_ANALYTICS_PLATFORM is present in more than one MigrationGroup. Each storage_set should only reference one migration group
````

doesn't seem to like that

File: snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+table_prefix = "uptime_monitor_checks"
+local_table_name = f"{table_prefix}_local"
+dist_table_name = f"{table_prefix}_dist"
+
+## what about all the fancy codecs? do we need those?
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("uptime_subscription_id", UInt(64)),
```
JoshFerge, line 20: NOTE: this should be a UUID.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def forwards_ops(self) -> Sequence[SqlOperation]:
+        return [
+            operations.CreateTable(
+                storage_set=storage_set,
+                table_name=local_table_name,
+                columns=columns,
+                engine=table_engines.ReplacingMergeTree(
+                    primary_key="(organization_id, project_id, toDateTime(timestamp), uptime_check_id, trace_id)",
+                    order_by="(organization_id, project_id, toDateTime(timestamp), uptime_check_id, trace_id)",
```
phacops, line 48: Is there only 1 check per trace? You might something more specific here otherwise the ReplacingMergeTree will merge 2 rows with the same sort key.
```
The engine differs from MergeTree in that it removes duplicate entries with the same sorting key value (ORDER BY table section, not PRIMARY KEY).
```
https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree

If you're not going to update values or if you don't care about duplicates, you could just use a MergeTree.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
phacops: Eh, I was wrong, you can't keep it.

You should move the migration to the events_analytics_platform directory. The benefit is your migration will be run automatically on the right clusters since the dataset is configured everywhere.

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
JoshFerge: only one check per trace. duplicates would not be ideal

snuba/snuba_migrations/uptime_monitor_checks/0001_uptime_monitor_checks.py
JoshFerge: 👍 will do so shortly

JoshFerge:
closing as the migration needs to happen first.

PR #6691, opened 2024-12-18
inc984: scrub `user` from eap_spans

xurui-c:


snuba/manual_jobs/scrub_users_from_eap_spans.py
```diff
@@ -0,0 +1,59 @@
...
+        on_cluster = f"ON CLUSTER '{cluster_name}'" if cluster_name else ""
+        return f"""ALTER TABLE eap_spans_2_local
+{on_cluster}
+UPDATE `attr_str_2` = mapApply((k, v) -> (k, if(k = 'user' AND startsWith(v, '{_IP_PREFIX}'), concat(
+                    '{_IP_PREFIX}',
+                    if(
+                        isIPv4String(substring(v, 4)) OR isIPv6String(substring(v, 4)),
+                        '{_SCRUBBED}',
+                        substring(v, 4)
+                    )
```
davidtsuk, line 38: This performs a string substring and concat even when we don't have to scrub anything, might be more efficient to pull out that condition into the outer if. Something like
```
mapApply((k, v) -> (k, if(k = 'user' AND startsWith(v, '{_IP_PREFIX}') AND (isIPv4String(substring(v, 4)) OR isIPv6String(substring(v, 4))), '{_IP_PREFIX + _SCRUBBED}', v)
```

snuba/manual_jobs/scrub_users_from_eap_spans.py
```diff
@@ -0,0 +1,59 @@
...
+                ), v)), `attr_str_2`)
```
davidtsuk, line 39: are we only scrubbing from `attr_str_2`?

xurui-c: I thought `user` only exists in `attr_str_2`, like how `user.ip` only in `attr_str_1`. Are you seeing `user` elsewhere?

kylemumma (approved):
LGTM

1. do you have to make an ops repo change to actually run this job?
2. have you ensured this wont alter so many rows that it would cause too much load on the cluster?

snuba/manual_jobs/scrub_users_from_eap_spans.py
kylemumma: do u need to worry about scrubbing from `spans_str_attr` aka `spans_str_attrs_3_local` table
https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/events_analytics_platform/storages/spans_str_attrs.yaml

xurui-c:
@kylemumma 
1. No, the Snuba Jobs Systems handles that
2. the typical workflow is run it for 2 hours and monitor CH, before running the entire thing. FWIW, it was fine for user.ip on eap spans

volokluev (approved):

PR #6694, opened 2024-12-19
984: scrub eap_spans_str_attrs

kylemumma:
this job deletes all attributes from the `spans_str_attrs` storage if 
1. its name is`sentry.user` and its an ip 
2. or if its name is `sentry.user.ip`

https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/events_analytics_platform/storages/spans_str_attrs.yaml

### DIFF
--- a/snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
+++ b/snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
@@ -28,7 +28,15 @@ class ScrubUserFromEAPSpansStrAttrs(Job):
         on_cluster = f"ON CLUSTER '{cluster_name}'" if cluster_name else ""
         return f"""ALTER TABLE spans_str_attrs_3_local
 {on_cluster}
-DELETE WHERE (attr_key = 'sentry.user.ip' OR attr_key = 'sentry.user')
+DELETE WHERE
+(
+    attr_key = 'sentry.user.ip' OR
+    (
+        attr_key = 'sentry.user' AND
+        startsWith(attr_value, 'ip:') AND
+        (isIPv4String(substring(attr_value, 4)) OR isIPv6String(substring(attr_value, 4)))
+    )
+)
 AND organization_id IN [{organization_ids}]
 AND timestamp >= toDateTime('{start_datetime}')
 AND timestamp < toDateTime('{end_datetime}')"""
@@ -45,7 +53,7 @@ AND timestamp < toDateTime('{end_datetime}')"""
             cluster_name = None
         query = self._get_query(cluster_name)
         logger.info("Executing query: {query}")
-        result = connection.execute(query=query, settings={"mutations_sync": 2})
+        result = connection.execute(query=query, settings={"mutations_sync": 0})
 
         logger.info("complete")
         logger.info(repr(result))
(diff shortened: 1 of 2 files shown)


snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
```diff
@@ -0,0 +1,51 @@
...
+        self._end_datetime = datetime.fromisoformat(params["end_datetime"])
+
+    def _get_query(self, cluster_name: str | None) -> str:
+        organization_ids = ",".join([str(p) for p in self._organization_ids])
+        start_datetime = self._start_datetime.isoformat()
+        end_datetime = self._end_datetime.isoformat()
+        on_cluster = f"ON CLUSTER '{cluster_name}'" if cluster_name else ""
+        return f"""ALTER TABLE spans_str_attrs_3_local
+{on_cluster}
+DELETE WHERE (attr_key = 'sentry.user.ip' OR attr_key = 'sentry.user')
```
xurui-c, line 31: I think we are scrubbing right? 'scrubbed' for user.ip and 'ip:scrubbed` for user

volokluev: you can delete too, it's file

volokluev, line 31: we only want to delete the `sentry.user` if it's an Ip address, not in all cases

snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
```diff
@@ -0,0 +1,51 @@
...
+        connection = cluster.get_node_connection(
+            ClickhouseClientSettings.CLEANUP, storage_node
+        )
+        if not cluster.is_single_node():
+            cluster_name = cluster.get_clickhouse_cluster_name()
+        else:
+            cluster_name = None
+        query = self._get_query(cluster_name)
+        logger.info("Executing query: {query}")
+        result = connection.execute(query=query, settings={"mutations_sync": 2})
```
volokluev, line 48: This will not finish on time, it has to be asynchronous
```suggestion
        result = connection.execute(query=query, settings={"mutations_sync": 0})
```

snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
kylemumma: it wont let me `alter table update` bc the columns are in the primary key or something, i have to do a delete

snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
kylemumma: done

snuba/manual_jobs/scrub_users_from_eap_spans_str_attrs.py
kylemumma: done

kylemumma:
1. I added a real test to make sure it only deletes what we want
2. I made the changes volo requested

volokluev (approved):
once you fix test failures, lgtm

PR #6697, opened 2024-12-20
feat(uptime): storage configuration and and consumer

JoshFerge:


JoshFerge:
depends on https://github.com/getsentry/sentry-kafka-schemas/pull/359 being merged and us bumping the kafka schema

JoshFerge:
needs https://github.com/getsentry/snuba/pull/6717 but otherwise should be good to go

rust_snuba/src/processors/uptime_monitor_checks.rs
```diff
@@ -0,0 +1,157 @@
...
+        uptime_check_id: monitor_message.guid,
+        scheduled_check_time: monitor_message.scheduled_check_time_ms,
+        timestamp: monitor_message.actual_check_time_ms,
+        duration: monitor_message.duration_ms,
+        region_slug: Some("global".to_string()),
+        check_status: monitor_message.status,
+        check_status_reason: monitor_message.status_reason.map(|r| r.r#type),
+        http_status_code: monitor_message
+            .request_info
+            .unwrap()
```
phacops, line 43: It panics if `unwrap()` fails here.

Add the `Default` trait to your `RequestInfo` structure and use `unwrap_or_default()` for both.

rust_snuba/src/processors/uptime_monitor_checks.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+#[derive(Debug, Deserialize)]
+pub struct RequestInfo {
+    pub http_status_code: Option<u16>,
+}
+#[derive(Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub struct CheckStatusReason {
+    /// The type of the status reason
+    pub r#type: String,
```
phacops, line 81: I'm sure it works but damn that's ugly. What happened to naming it `ty` and putting an `#[serde(rename = "type")]`...

phacops: Also, it looks like it's a very static reason, we could benefit from having it be a `'a &str` instead of `String`.

rust_snuba/src/processors/uptime_monitor_checks.rs
```diff
@@ -0,0 +1,157 @@
...
+    let rows = vec![UptimeMonitorCheckRow {
+        organization_id: 1,
+        project_id: 1,
+        environment: monitor_message.environment,
+        uptime_subscription_id: monitor_message.subscription_id,
+        uptime_check_id: monitor_message.guid,
+        scheduled_check_time: monitor_message.scheduled_check_time_ms,
+        timestamp: monitor_message.actual_check_time_ms,
+        duration: monitor_message.duration_ms,
+        region_slug: Some("global".to_string()),
```
phacops, line 38: If we're going to set a default, we can remove the `Option` then.

snuba/datasets/configuration/events_analytics_platform/storages/uptime_monitor_checks.yaml
```diff
@@ -0,0 +1,61 @@
...
+storage:
+  key: uptime_monitor_checks
+  set_key: uptime_monitor_checks
+readiness_state: limited
+schema:
+  columns:
+    [
+      { name: organization_id, type: UInt, args: { size: 64 } },
+      { name: project_id, type: UInt, args: { size: 64 } },
+      { name: environment, type: String, args: { schema_modifiers: [nullable, low_cardinality] } },
```
volokluev, line 13: this should not be low cardinality, it's set by a customer, you can't guarantee it

JoshFerge: environment is set to low cardinality on some other datasets, was just copying based on that. i can remove it if you'd like though though https://github.com/getsentry/snuba/blob/4c7d8d90fa0186a3b2de9039a79a3196878ed079/snuba/datasets/configuration/issues/entities/search_issues.yaml#L25 https://github.com/getsentry/snuba/blob/4c7d8d90fa0186a3b2de9039a79a3196878ed079/snuba/snuba_migrations/events/0011_rebuild_errors.py#L24

rust_snuba/src/processors/uptime_monitor_checks.rs
phacops: Not that big of a deal, saves on memory. We do it a lot in Relay too.

https://github.com/getsentry/relay/blob/master/relay-server/src/services/store.rs#L1255

snuba/clusters/storage_sets.py
```diff
@@ -20,6 +20,7 @@
     "EVENTS_ANALYTICS_PLATFORM": "events_analytics_platform",
     "GROUP_ATTRIBUTES": "group_attributes",
     "PROFILE_CHUNKS": "profile_chunks",
+    "UPTIME_RESULTS": "uptime_monitor_checks",
```
phacops, line 23: You don't need this. The migration belongs to the `events_analytics_platform` storage set already. That also means removing migration related changes in this PR (in `group_loader.py`, `groups.py` and the directory in `snuba_migrations`).

I believe it hasn't been run because it's prefixed with `0020_` when the previous one is `0018_`, and yes, that's my fault since I reverted `0019_`, but as soon as you fixed that, the migration will run everywhere the `events_analytics_platform` is available.

snuba/datasets/configuration/events_analytics_platform/storages/uptime_monitor_checks.yaml
```diff
@@ -0,0 +1,61 @@
+version: v1
+kind: writable_storage
+name: uptime_monitor_checks
+storage:
+  key: uptime_monitor_checks
+  set_key: uptime_monitor_checks
```
phacops, line 6: I think it should belong to the `events_analytics_platform` since migrations are also link to the `events_analytics_platform` storage set.

I suppose it's not a MUST but since the migrations are already in the `events_analytics_platform` set, I'd keep things together.

rust_snuba/src/processors/uptime_monitor_checks.rs
JoshFerge: went ahead and converted all `String`s to `&str`s for all the strings! let me know if what i did makes sense, not super familiar with lifetimes in rust.

snuba/clusters/storage_sets.py
JoshFerge: i'll update the migration in a separate PR! thanks for the callout.

snuba/datasets/configuration/events_analytics_platform/storages/uptime_monitor_checks.yaml
JoshFerge: i've gone ahead and removed the errant uptime_monitor_checks config stuff.

snuba/clusters/storage_sets.py
JoshFerge: i've also removed the migration related changes as specified. thanks.

phacops (approved):

PR #6698, opened 2024-12-20
chore(eap): write script to send scrubbed data into a gcs bucket

davidtsuk:
Closes https://github.com/getsentry/eap-planning/issues/139

### DIFF
--- a/snuba/manual_jobs/extract_span_data.py
+++ b/snuba/manual_jobs/extract_span_data.py
@@ -1,5 +1,3 @@
-from datetime import datetime
-
 from snuba.clusters.cluster import ClickhouseClientSettings, get_cluster
 from snuba.clusters.storage_sets import StorageSetKey
 from snuba.manual_jobs import Job, JobLogger, JobSpec
@@ -10,11 +8,10 @@ class ExtractSpanData(Job):
         super().__init__(job_spec)
 
     def _generate_spans_query(self):
-        # Columns that should not be hashed (numeric and date types)
-        numeric_columns = {
-            "organization_id",
-            "project_id",
+        # Columns that should not be scrubbed
+        unscrubbed_columns = {
             "span_id",
+            "trace_id",
             "parent_span_id",
             "segment_id",
             "is_segment",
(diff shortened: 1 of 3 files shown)


snuba/manual_jobs/extract_span_data.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def execute(self, logger: JobLogger) -> None:
+        cluster = get_cluster(StorageSetKey.EVENTS_ANALYTICS_PLATFORM)
+        connection = cluster.get_query_connection(ClickhouseClientSettings.QUERY)
+
+        current_time = datetime.now().strftime("%Y%m%d_%H%M%S")
+        file_name = f"scrubbed_spans_data_{current_time}.csv.gz"
+
+        query = f"""
+        INSERT INTO FUNCTION gcs('https://storage.googleapis.com/{self.gcp_bucket_name}/{file_name}',
```
phacops, line 90: Dump that into a directory with the date of the dump.

You can probably make that one variable to be called later in the log.

snuba/manual_jobs/extract_span_data.py
```diff
@@ -0,0 +1,102 @@
...
+        WHERE _sort_timestamp BETWEEN toDateTime('{self.start_timestamp}') AND toDateTime('{self.end_timestamp}')
+        AND organization_id IN {self.organization_ids}
+        LIMIT {self.limit}
+        """
+
+        return query
+
+    def execute(self, logger: JobLogger) -> None:
+        cluster = get_cluster(StorageSetKey.EVENTS_ANALYTICS_PLATFORM)
+        connection = cluster.get_query_connection(ClickhouseClientSettings.QUERY)
```
phacops, line 84: Let's do one query that does an `INSERT INTO <gcs_bucket> (SELECT ... FROM eap_spans_2_local)` instead.

And make sure it's running asynchronous so this script can stop with success.

phacops (changes requested):

snuba/manual_jobs/extract_span_data.py
davidtsuk: > Let's do one query that does an INSERT INTO <gcs_bucket> (SELECT ... FROM eap_spans_2_local) instead.
How is this different from what I already have?

davidtsuk: I was looking into using those settings, but when I tested them out, they didn't work for inserts into a gcs bucket. It seems like they're mainly intended for inserts into actual clickhouse tables.

snuba/manual_jobs/extract_span_data.py
```diff
@@ -0,0 +1,121 @@
...
+            "organization_ids",
+            "start_timestamp",
+            "end_timestamp",
+            "table_name",
+            "limit",
+            "gcp_bucket_name",
+            "output_file_path",
+        ]
+        for param in required_params:
+            assert param in params
```
onkar, line 25: I think you can use Counter() to compare two unordered lists.
```
Python 3.11.8 (main, Aug 16 2024, 10:27:21) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import Counter
>>> l1 = ["one", "two", "three"]
>>> l2 = ["two", "one", "three"]
>>> assert Counter(l1) == Counter(l2), "Lists must be same!"
>>> l1.append("four")
>>> assert Counter(l1) == Counter(l2), "Lists must be same!"
  File "<stdin>", line 1, in <module>
AssertionError: Lists must be same!
```
Also, consider adding a message along with `assert` to make debugging easier.

onkar (approved):
LGTM, left a minor comment.

phacops (approved):

PR #6700, opened 2024-12-23
ref(admin): Switch to esbuild

untitaker:
webpack takes ~more than a minute~ 30s, esbuild takes 2 seconds. esbuild has
reasonable defaults wrt building css imports, webpack needs two
additional plugins for that.

esbuild has no file-watching support, but node has. since the build is
now fast, we can just do a full rebuild when any file changes.

there are other, newer, cooler runtimes (bun, rome, biome, idk), but
esbuild has not changed in the past 2 years and is the one I know.


.envrc
```diff
@@ -19,4 +19,6 @@ else
     unset PS1
 fi
 
+export PATH="$PWD/snuba/admin/node_modules/.bin/:$PATH"
```
untitaker, line 22: not necessary but being able to run `jest` directly is nice

onkar (approved):

PR #6701, opened 2024-12-27
feat(eap): Start decoupling EAP entities at the entity layer

colin-sentry:
There's currently a big chunk of code in `common.py` that maps column accesses to the correct place.

However, it's all hard-coded to the spans table right now.

We would like to add new non-span entity types to the EAP RPCs, and the easiest place to do that is at the entity layer.

This essentially hides the 'real columns' behind a well-known set of columns (`organization_id`, `attr_str`, `attr_f64`, `attr_i64`, ...) which will be shared across all EAP entities.

When we add new entities, we can specify in the entity YAML what maps where in a way that the RPC doesn't have to know what type of entity is being processed.

Works towards solving https://github.com/getsentry/eap-planning/issues/126

### DIFF
--- a/snuba/clickhouse/translators/snuba/mappers.py
+++ b/snuba/clickhouse/translators/snuba/mappers.py
@@ -263,7 +263,7 @@ class SubscriptableHashBucketMapper(SubscriptableReferenceMapper):
             return None
 
         bucket_idx = fnv_1a(key.value.encode("utf-8")) % ATTRIBUTE_BUCKETS
-        expr = arrayElement(
+        expr: Expression = arrayElement(
             None,
             ColumnExpr(None, self.to_col_table, f"{self.to_col_name}_{bucket_idx}"),
             key,
--- a/snuba/clickhouse/translators/snuba/mappers.py
+++ b/snuba/clickhouse/translators/snuba/mappers.py
@@ -25,8 +25,6 @@ from snuba.query.matchers import (
     Param,
     String,
 )
-from snuba.utils.constants import ATTRIBUTE_BUCKETS
-from snuba.utils.hashes import fnv_1a
 
 
 # This is a workaround for a mypy bug, found here: https://github.com/python/mypy/issues/5374
@@ -227,41 +225,6 @@ class SubscriptableMapper(SubscriptableReferenceMapper):
             )
         else:
             return None
-
-
-@dataclass(frozen=True)
-class SubscriptableHashBucketMapper(SubscriptableReferenceMapper):
-    """
-    Maps a key into the appropriate bucket by hashing the key. For example, hello[test] might go to attr_str_22['test']
-    """
-
-    from_column_table: Optional[str]
-    from_column_name: str
-    to_col_table: Optional[str]
-    to_col_name: str
-
-    def attempt_map(
-        self,
-        expression: SubscriptableReference,
-        children_translator: SnubaClickhouseStrictTranslator,
-    ) -> Optional[FunctionCallExpr]:
-        if (
-            expression.column.table_name != self.from_column_table
-            or expression.column.column_name != self.from_column_name
-        ):
-            return None
-        key = expression.key.accept(children_translator)
-        if not isinstance(key, LiteralExpr):
-            return None
-        if not isinstance(key.value, str):
-            return None
-
-        bucket_idx = fnv_1a(key.value.encode("utf-8")) % ATTRIBUTE_BUCKETS
-        return arrayElement(
-            expression.alias,
-            ColumnExpr(None, self.to_col_table, f"{self.to_col_name}_{bucket_idx}"),
-            key,
-        )
 
 
 @dataclass(frozen=True)
--- a/snuba/clickhouse/translators/snuba/mappers.py
+++ b/snuba/clickhouse/translators/snuba/mappers.py
@@ -9,8 +9,7 @@ from snuba.clickhouse.translators.snuba.allowed import (
     SubscriptableReferenceMapper,
     ValidColumnMappings,
 )
-from snuba.query.dsl import Functions as f
-from snuba.query.dsl import arrayElement, column
+from snuba.query.dsl import arrayElement, column, literal
 from snuba.query.expressions import Column as ColumnExpr
 from snuba.query.expressions import CurriedFunctionCall, Expression
 from snuba.query.expressions import FunctionCall as FunctionCallExpr
@@ -268,24 +267,30 @@ class SubscriptableHashBucketMapper(SubscriptableReferenceMapper):
             and key.value in self.normalized_columns
             and self.data_type
         ):
-            return f.CAST(
-                column(self.normalized_columns[key.value]),
-                self.data_type,
-                alias=expression.alias,
+            return FunctionCallExpr(
+                expression.alias,
+                "CAST",
+                (
+                    column(self.normalized_columns[key.value]),
+                    literal(self.data_type),
+                ),
             )
 
         bucket_idx = fnv_1a(key.value.encode("utf-8")) % ATTRIBUTE_BUCKETS
         if self.data_type:
-            return f.CAST(
-                arrayElement(
-                    None,
-                    ColumnExpr(
-                        None, self.to_col_table, f"{self.to_col_name}_{bucket_idx}"
+            return FunctionCallExpr(
+                expression.alias,
+                "CAST",
+                (
+                    arrayElement(
+                        None,
+                        ColumnExpr(
+                            None, self.to_col_table, f"{self.to_col_name}_{bucket_idx}"
+                        ),
+                        key,
                     ),
-                    key,
+                    literal(self.data_type),
                 ),
-                self.data_type,
-                alias=expression.alias,
             )
         else:
             return arrayElement(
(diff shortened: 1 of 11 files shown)


snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+    # these are the required columns for an 'RPC entity' that can be used by EAP RPCs
+    { name: organization_id, type: UInt, args: { size: 64 } },
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: time, type: DateTime }, # used by TimeSeriesProcessor
+    { name: timestamp, type: DateTime }, # mapped to _sort_timestamp
+    { name: retention_days, type: UInt, args: { size: 16 } },
+    { name: sampling_factor, type: Float, args: { size: 64 } },
+    { name: sampling_weight, type: UInt, args: { size: 64 } },
+    { name: sign, type: Int, args: { size: 8 } },
```
phacops, line 28: I don't think this should be a required column as we're not using `CollapsingMergeTree` everywhere.

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,124 @@
...
+    { name: end_timestamp, type: DateTime64, args: { precision: 6 } },
+    { name: duration_ms, type: Float, args: { size: 64 } },
+    { name: exclusive_time_ms, type: Float, args: { size: 64 } },
+    { name: name, type: String },
+
+    # these are the required columns for an 'RPC entity' that can be used by EAP RPCs
+    { name: organization_id, type: UInt, args: { size: 64 } },
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: time, type: DateTime }, # used by TimeSeriesProcessor
+    { name: timestamp, type: DateTime }, # mapped to _sort_timestamp
```
phacops, line 24: Is there a way we do not have this duplication? I think one timestamp is enough for the required fields.

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,124 @@
+version: v1
+kind: entity
+name: eap_spans_rpc
+
+schema:
+  [
+    { name: service, type: String },
+    { name: trace_id, type: UUID },
```
phacops, line 8: This should be a required field given the API we have.

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,124 @@
...
+    { name: span_id, type: UInt, args: { size: 64 } },
```
phacops, line 9: We should have some sort of a `event_id`, as a string, required in for the RPC. Having it as a string would let us pass any sort of UUIDs or span IDs.

phacops (commented):
We also shoul

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,124 @@
...
+    { name: project_id, type: UInt, args: { size: 64 } },
+    { name: time, type: DateTime }, # used by TimeSeriesProcessor
+    { name: timestamp, type: DateTime }, # mapped to _sort_timestamp
+    { name: retention_days, type: UInt, args: { size: 16 } },
+    { name: sampling_factor, type: Float, args: { size: 64 } },
+    { name: sampling_weight, type: UInt, args: { size: 64 } },
+    { name: sign, type: Int, args: { size: 8 } },
+    { name: attr_str, type: Map, args: { key: { type: String }, value: { type: String } } },
+    { name: attr_f64, type: Map, args: { key: { type: String }, value: { type: Float, args: { size: 64 } } } },
+    { name: attr_i64, type: Map, args: { key: { type: String }, value: { type: Int, args: { size: 64 } } } },
```
phacops, line 31: Let's leave things that don't exist yet out of it.

tests/clickhouse/translators/snuba/test_translation.py
```diff
@@ -146,6 +146,59 @@ def test_hash_bucket_tag_translation() -> None:
     )
 
 
+def test_hash_bucket_normalized() -> None:
```
onkar, line 149: Can you also add unit tests for when:

1. `normalized_columns` is specified but `data_type` is _not_ specified? Based on the comment above IMO, this should raise some exception or error condition.
2. Only `data_type` is specified but `normalized_columns` is _not_ specified. This should ideally just do the casting.
3. Neither `normalized_columns` nor `data_type` is specified. This should return `arrayElement(...)`

onkar (commented):
Thanks for working on this. I left a few comments.

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
colin-sentry: You can query integers with the RPC layer already, this just exposes the ability to query something as an integer

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
colin-sentry: We can delete these once all the references are removed, let's keep `time` and `sign` (etc) for now, since they are actively being used - later we can delete them instead of doing one massive PR

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
colin-sentry: How about uint128? String would have a big performance hit

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,124 @@
...
+  - processor: HashBucketFunctionTransformer
+    args:
+      hash_bucket_name_mapping:
+        attr_str: attr_str
+        attr_f64: attr_num
+        attr_i64: attr_num
+  - processor: OptionalAttributeAggregationTransformer
+    args:
+      attribute_column_names:
+        - attr_num
```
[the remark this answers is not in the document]
colin-sentry: it would actually be attr_f64 in this case, but fixed regardless

tests/clickhouse/translators/snuba/test_translation.py
colin-sentry: 1 - fixed by making data_type required
2 - test_hash_bucket_tag_translation does this now
3 - irrelevant since data_type is required now

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
colin-sentry: Let's add that later, would require another net-new processor

snuba/clickhouse/translators/snuba/mappers.py
```diff
@@ -229,41 +227,6 @@ def attempt_map(
             return None
 
 
-@dataclass(frozen=True)
-class SubscriptableHashBucketMapper(SubscriptableReferenceMapper):
```
colin-sentry, line 233: This gets moved into a query processor (the same one which handles mapKeys and mapContains)

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans.yaml
```diff
@@ -88,11 +74,16 @@ query_processors:
       curried_aggregation_names:
         - quantile
         - quantileTDigestWeighted
-  - processor: HashBucketFunctionTransformer
```
colin-sentry, line 91: There was a really annoying order of operations, where query processors needed to know what bucket things would end up in, but that was done at the storage level.

By merging all of the processors which need to know the actual bucket-level information to a single one at the end, the pipeline is a lot more understandable and has less chance for bugs

snuba/datasets/configuration/events_analytics_platform/entities/eap_spans_rpc.yaml
```diff
@@ -0,0 +1,140 @@
...
+          args:
+            from_table_name: null
+            from_col_name: timestamp
+            to_table_name: null
+            to_col_name: _sort_timestamp
+
+storage_selector:
+  selector: DefaultQueryStorageSelector
+
+query_processors:
```
colin-sentry, line 49: I rewrote a lot of this, there were starting to be conflicting edge cases where, e.g., `sum(sentry.duration_ms)` should become `sum(duration_ms)` but the very similar `sum(attr_i64[blah])` should become `sumIf(CAST(attr_num_2[blah], 'Integer'), mapContains(attr_num_2, 'blah'))`

colin-sentry:
EAP team will be finishing this PR

PR #6704, opened 2024-12-30
feat(job-runner): surface errors to Snuba admin

xurui-c:
Currently, errors are surfaced to GCP, and Snuba Admin just displays server error. I want to surface the actual error message to Snuba Admin as well

### DIFF
--- a/snuba/manual_jobs/runner.py
+++ b/snuba/manual_jobs/runner.py
@@ -134,7 +134,7 @@ def run_job(job_spec: JobSpec) -> JobStatus:
         if not job_spec.is_async:
             current_job_status = _set_job_status(job_spec.job_id, JobStatus.FINISHED)
             job_logger.info("[runner] job execution finished")
-    except BaseException as e:
+    except Exception as e:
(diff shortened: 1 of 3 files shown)


evanh (approved):

snuba/manual_jobs/runner.py
```diff
@@ -134,10 +134,11 @@ def run_job(job_spec: JobSpec) -> JobStatus:
         if not job_spec.is_async:
             current_job_status = _set_job_status(job_spec.job_id, JobStatus.FINISHED)
             job_logger.info("[runner] job execution finished")
-    except BaseException as e:
+    except Exception as e:
```
onkar, line 137: Why did this exception type need change? AFAIK, `BaseException` is used to catch all exceptions including the ones needed to terminate the program, while `Exception` is used to catch exceptions that do not terminate the program.

onkar (approved):

snuba/manual_jobs/runner.py
xurui-c: My mistake, fixed!

PR #6711, opened 2025-01-02
ref(consumers): Remove ErrorsProcessor

untitaker:
This processor has been implemented in
rust_snuba/src/processors/errors.rs for quite a while now, and deployed in US region since then. The python one should no longer be needed.

* Fix a user-facing bug where HTTP referers were no longer searchable, as the `http_referer` column would always hold an empty string. The new behavior is that any HTTP header like `Referer` or `Referrer` would be used.
* Apparently the rust processor uses different hyphenization in `contexts[trace_id]` than the python one. nobody noticed this as we use a top-level trace_id column.


evanh (approved):
Nice to see more code deleted!

untitaker:
now blocked on: https://github.com/getsentry/sentry/pull/82934

PR #6712, opened 2025-01-02
chore(inc-984): Add basic PII detector

untitaker:
In INC-984 we had an incident where IP addresses were copied into a new
unscrubbed field. It took us too long to detect that issue.

This metric could be used in combination with anomaly detection to
detect such faulty rollouts more quickly, but not prevent them.

TBD is who actually gets those alerts and how we should react to them.


onkar (approved):

enochtangg (approved):
Some small questions/comments. I haven't been following INC-984 too closely. 

1. This metric looks like a good additional signal to have. But I'd imagine it'd be difficult to distinguish between a faulty rollout and maybe more/new orgs sending us many more events (with the prevent storing IP option off)?

2. Are we adding this to all processors?

untitaker:
> This metric looks like a good additional signal to have. But I'd imagine it'd be difficult to distinguish between a faulty rollout and maybe more/new orgs sending us many more events (with the prevent storing IP option off)?

my feeling right now is that I'd like to take the %age of "Events with PII", and install anomaly detection for that. new orgs signup shouldn't change the %age dramatically.

> Are we adding this to all processors?

yep, it's enabled unconditionally for all processors. if this turns out to be too expensive I can add sampling. I don't want to do it per-dataset because then we'll have potential monitoring gaps again.

PR #6713, opened 2025-01-02
feat(flags): Update errors processor to index feature flag data

cmanallen:
Updates the errors processor to extract the flags context and set it on the flags column.

DACI: https://www.notion.so/sentry/DACI-Search-for-Errors-by-Feature-Flag-1288b10e4b5d8014976dcbfbf335bc37?pvs=4#1288b10e4b5d805eb476d313eecae009
Sentry-kafka-schemas changes: https://github.com/getsentry/sentry-kafka-schemas/pull/364
Depends on: https://github.com/getsentry/snuba/pull/6723
Closes: https://github.com/getsentry/team-replay/issues/520


tests/datasets/test_errors_processor.py
```diff
@@ -377,8 +379,15 @@ def build_result(self, meta: KafkaMessageMetadata) -> Mapping[str, Any]:
...
             "transaction_name": "",
             "num_processing_errors": len(self.errors) if self.errors is not None else 0,
+            "features.key": [],
+            "features.value": [],
         }
 
+        if self.features:
+            for feature in self.features:
+                expected_result["features.key"].append(feature["key"])
+                expected_result["features.value"].append(str(feature["value"]).lower())
```
evanh, line 389: Where does the `.lower()` happen in the processor?

evanh (approved):

tests/datasets/test_errors_processor.py
cmanallen: Unicodify serializes booleans the way JSON does (e.g. `true`).  Calling `str` in python serializes booleans with the starting letter uppercased.  The correct way to handle this is to use `json.dumps` instead of `str`.  I'll fix it.

rust_snuba/src/processors/errors.rs
```diff
@@ -563,6 +589,19 @@ impl ErrorRow {
             replay_id = Some(rid)
         }
 
+        // Split feature keys and values into two vectors if the context could be parsed.
+        let mut features_key = Vec::new();
+        let mut features_value = Vec::new();
+
+        if let Some(FeatureContextEnum::Typed(ctx)) = from_context.features {
+            ctx.values.into_iter().for_each(|i| {
```
untitaker, line 597:
```suggestion
            for i in ctx.values {
```

rust_snuba/src/processors/errors.rs
```diff
@@ -183,6 +185,26 @@ struct TraceContext {
     other: GenericContext,
 }
 
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(untagged)]
+enum FeatureContextEnum {
+    Typed(FeatureContext),
+    #[allow(warnings)]
+    Untyped(Value),
```
untitaker, line 193: I don't think this is a reasonable schema. every time there is any schema incompatibility within the feature context, we drop it entirely. why are there sometimes untyped values in this context?

why is `allow(warnings)` needed?

cmanallen: At least one customer sends their feature flags on the context using the same key as us but not the same format.  So I didn't want to fail de-serialization for them.  Allow warnings is used because we never access the inner value and it produces a lint error.

untitaker: I suggest to handle this case in Relay. It would null out or rename the context, and already has a lot of facilities for partial deserialization/partial errors.

cmanallen: @untitaker Whats the purpose of doing this in Relay? Why should we not deserialize in this way in Snuba?

untitaker: think of not only snuba as downstream consumers of the ingestion topic, but also sbc or the UI. they'd all have to handle these compatibility issues if relay didn't iron them out at the start.

i suggest to either:

* drop user-defined contexts if they don't match the schema in relay
* rename them within relay to another key

either way you will be able to have a stricter schema in the ingestion topic and then in snuba, and the complexity for handling those different user inputs is contained in relay instead of {sbc, snuba, ...}

untitaker (approved):

PR #6714, opened 2025-01-02
fix: bug in AttributeNames RPC endpoint, filter offset

kylemumma:
There is an index out of range error with filter offset when the attribute results are empty. This PR solves this bug.
The new test `test_empty_results` failed before (due to the bug), and now passes

here is the sentry issue it fixes: https://sentry.sentry.io/issues/6189483746/?project=300688&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=1h&stream_index=8

and it also closes this ticket https://github.com/getsentry/eap-planning/issues/138


xurui-c (approved):

onkar (approved):
LGTM. Please verify if this fix also addresses https://github.com/getsentry/eap-planning/issues/140.

PR #6715, opened 2025-01-02
fix(eap): Ensure reliabilities are returned more consistently

Zylphrex:
When extrapolating, if there's no data, we skip returning reliabilities. This is problematic as we want to line up each reliability value with the corresponding result so those arrays must be the same length. This fills it with `RELIABILITY_UNSPECIFIED`.

### DIFF
--- a/snuba/web/rpc/v1/endpoint_time_series.py
+++ b/snuba/web/rpc/v1/endpoint_time_series.py
@@ -29,7 +29,7 @@ from snuba.request import Request as SnubaRequest
 from snuba.web.query import run_query
 from snuba.web.rpc import RPCEndpoint
 from snuba.web.rpc.common.aggregation import (
-    ExtrapolationMeta,
+    ExtrapolationContext,
     aggregation_to_expression,
     get_average_sample_rate_column,
     get_confidence_interval_column,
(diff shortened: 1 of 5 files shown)


snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -170,16 +170,23 @@ def _convert_result_timeseries(
...
-                extrapolation_meta = ExtrapolationMeta.from_row(
-                    row_data, timeseries.label
+                extrapolation_context = ExtrapolationContext.from_row(
+                    timeseries.label, row_data
                 )
-                if extrapolation_meta is not None:
+                if (
+                    # This isn't quite right but all non extrapolated aggregates
+                    # are assumed to be present.
+                    not extrapolation_context.is_extrapolated
```
Zylphrex, line 179: I didn't want to change too much existing behaviour in this PR but for whatever reason, when using a non aggregate in a timeseries query, we always consider as `data_present=True`.

kylemumma (approved):
I liked your use of comments and your code was easy to read

snuba/web/rpc/v1/endpoint_time_series.py
Zylphrex: Yeah, I didn't want to change too many behaviours in this one PR so I kept this behaviour for now but it does feels like a bug.

onkar: @kylemumma can you look into this and file a ticket on eap-planning to address this? I am not following what the bug is really but if either of you have an idea of the issue, we should track it and even fix it. Let's not leave this unaddressed.

Zylphrex: https://github.com/getsentry/eap-planning/issues/144

kylemumma: thanks tony

kylemumma:
This PR addresses this issue https://github.com/getsentry/eap-planning/issues/142

PR #6719, opened 2025-01-07
chore(deps): Bump Python to 3.11.11

beninabox:
There are a number of security vulnerabilities which are present in the Python 3.11.8 image, so bumped to the latest version of 3.11 which is 3.11.11.

Fixes #6718

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #6575, opened 2024-11-19
feat: Secure clickhouse connections

untitaker:
***this pr is a mirror of #6459 due to issues with CI***

### Description
This pull request introduces SSL/TLS support for ClickHouse connections in the Snuba project. The changes include new CLI options for enabling secure connections, updates to the ClickhousePool and HTTPBatchWriter classes, and corresponding configuration options in settings and tests.

### Changes Overview
1. **CLI Options**:
   - Introduced new CLI options for enabling secure connections to ClickHouse:
     - `--clickhouse-secure`: If true, an encrypted connection will be used.
     - `--clickhouse-ca-certs`: An optional path to certificates directory.
     - `--clickhouse-verify`: Verify ClickHouse SSL cert.

2. **Class Updates**:
   - Modified `ClickhousePool`, `HTTPBatchWriter`, and other relevant classes to support SSL/TLS connections.
   - Updated constructors and methods to accept and handle SSL/TLS parameters.

3. **Configuration**:
   - Added SSL/TLS configuration options in settings and tests.
   - Updated environment variables to support SSL/TLS settings.

4. **Testing**:
   - Included SSL/TLS configuration in test cases.
   - Updated existing tests to ensure compatibility with SSL/TLS options.

### Detailed Changes
- **snuba/cli/cleanup.py**: Added new CLI options for secure ClickHouse connections.
- **snuba/cli/migrations.py**: Added new CLI options for secure ClickHouse connections.
- **snuba/cli/optimize.py**: Added new CLI options for secure ClickHouse connections.
- **snuba/clickhouse/http.py**: Modified `HTTPBatchWriter` to support SSL/TLS connections.
- **snuba/clickhouse/native.py**: Updated `ClickhousePool` to handle SSL/TLS parameters.
- **snuba/clusters/cluster.py**: Updated `ClickhouseCluster` to include SSL/TLS configuration.
- **snuba/migrations/runner.py**: Added SSL/TLS parameters to migration runner.
- **snuba/settings/__init__.py**: Added SSL/TLS configuration options.
- **snuba/settings/settings_distributed.py**: Added SSL/TLS configuration options.
- **tests/clusters/fake_cluster.py**: Updated `FakeClickhouseCluster` to include SSL/TLS parameters.
- **tests/clusters/test_cluster.py**: Updated tests to include SSL/TLS configuration.
- **tests/conftest.py**: Updated test setup to include SSL/TLS configuration.
[...]

PR #6721, opened 2025-01-07
fix(ci): Enable third-party contributions

untitaker:
we have had two external contributions recently:

* #6575
* #6719 

neither of the contributors can iterate properly on their changes because our CI is actually busted for PRs launched from forks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,13 +160,13 @@ jobs:
           load: true
           target: testing
           tags: |
-            ghcr.io/${{ github.repository_owner }}/snuba-ci:${{ github.sha }}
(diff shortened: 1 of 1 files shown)


untitaker:
Should probably adopt https://docs.docker.com/build/ci/github-actions/share-image-jobs/ instead of using USERNAME/snuba-ci

onkar (approved):

onkar:
I noticed that the artifacts are published in `/tmp` directory. Is that intentional? Or do we see a need to inspect that `.tar` file in the future and in this case, it should be saved somewhere other than `/tmp`?

untitaker:
@onkar I was mostly following the linked docker tutorial, I think saving to /tmp is just an intermediate step because github artifacts can only load/save on the filesystem.

PR #6726, opened 2025-01-08
fix(eap): fix bug with data_present

davidtsuk:
Fixes https://github.com/getsentry/eap-planning/issues/144

## Additional Context
When we perform aggregations over attributes, the function is converted into function_nameIf (e.g. countIf) which returns a default value if no rows match the condition. This means that there is no way to distinguish between an aggregate result that is 0 because it's for example a sum of values that add up to 0, and a result that is 0 because no values matched the given condition. To deal with this, we compute the number of events being aggregated even when we aren't extrapolating so we can determine if data was present or not.

### DIFF
--- a/snuba/web/rpc/common/aggregation.py
+++ b/snuba/web/rpc/common/aggregation.py
@@ -67,8 +67,8 @@ class ExtrapolationContext(ABC):
         value = row_data[column_label]
 
         confidence_interval = None
-        average_sample_rate = None
-        sample_count = None
+        average_sample_rate = 0
+        sample_count = 0
 
         percentile = 0.0
         granularity = 0.0
@@ -113,7 +113,7 @@ class ExtrapolationContext(ABC):
                 value=value,
                 confidence_interval=None,
                 average_sample_rate=0,
-                sample_count=sample_count if sample_count is not None else 0,
+                sample_count=sample_count,
             )
 
         if is_percentile:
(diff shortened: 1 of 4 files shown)


snuba/web/rpc/common/aggregation.py
```diff
@@ -46,7 +46,7 @@ class ExtrapolationContext(ABC):
     sample_count: int
 
     @property
-    def extrapolated_data_present(self) -> bool:
+    def data_present(self) -> bool:
```
onkar, line 49: nit:  should this be renamed for `is_data_present` since it returns an answer to this question?

snuba/web/rpc/common/aggregation.py
```diff
@@ -108,16 +108,12 @@ def from_row(
...
-            or average_sample_rate is None
-            or sample_count is None
-        ):
+        if confidence_interval is None:
             return GenericExtrapolationContext(
                 value=value,
                 confidence_interval=None,
                 average_sample_rate=0,
-                sample_count=0,
+                sample_count=sample_count,
```
onkar, line 116: I am wondering if this `return` can be collapsed with the `return` on L130 for better readability. All we are checking is if `confidence_interval` is `None` and if so, we pass `None`.

onkar (approved):
LGTM

snuba/web/rpc/common/aggregation.py
davidtsuk: yeah you are right, it can be collapsed, nice catch

PR #6734, opened 2025-01-09
fix(eap): use TDigestWeighted instead of TDigest for perncetile confidence calculation

shellmayr:
In order to calculate the correct percentiles, we need to use the weighted TDigest - the current implementation uses the unweighted TDigest to calculate percentiles from all samples, but when the percentile estimate calculation is using the TDigest in the weighted variant, the system will basically mark everything as unreliable except from outliers. 

Contributes to https://github.com/getsentry/projects/issues/530


davidtsuk (approved):
LGTM, nice catch!

PR #6739, opened 2025-01-10
deps: replace python-jose with pyjwt

mdtro:
python-jose has some unaddressed security vulnerabilities, with the latest release from 2021.
This PR swaps python-jose with PyJWT.

This will resolve the following dependabot alerts:
- https://github.com/getsentry/snuba/security/dependabot/50
- https://github.com/getsentry/snuba/security/dependabot/51


onkar (approved):
Thanks for working on this! Last time I attempted it by just bumping the version, it did not work.

PR #6745, opened 2025-01-10
[wip] divide partitions based on day

onewland:
### Legal Boilerplate

### DIFF
--- a/snuba/clickhouse/optimize/optimize.py
+++ b/snuba/clickhouse/optimize/optimize.py
@@ -72,7 +72,7 @@ def run_optimize(
     table = schema.get_local_table_name()
     database = storage.get_cluster().get_database()
 
-    partitions = get_partitions_to_optimize(
+    partitions = get_partitions_from_clickhouse(
         clickhouse, storage, database, table, before
     )
     partition_names = [partition.name for partition in partitions]
@@ -95,14 +95,17 @@ def run_optimize_cron_job(
     clickhouse_host: str,
     tracker: OptimizedPartitionTracker,
     before: Optional[datetime] = None,
+    divide_partitions_count: int = 0,
 ) -> int:
     """
     The sophisticated form of running an optimize final on a storage.
 
-    The sophistication include:
+    The sophistication includes:
     1. Being able to run multiple optimize final jobs concurrently.
     2. Being able to set time boundaries for different phases of optimization.
     3. Being able to work with stateful optimization states.
+    4. Splitting partitions up into `divide_partitions_count` groups so that each
+       optimize cron job only optimizes a subset of partitions.
     """
     start = time.time()
     schema = storage.get_schema()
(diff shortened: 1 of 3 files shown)


snuba/clickhouse/optimize/optimize.py
```diff
@@ -157,7 +170,7 @@ def run_optimize_cron_job(
     return len(partitions_to_optimize)
 
 
-def get_partitions_to_optimize(
+def get_partitions_from_clickhouse(
```
onewland, line 173: `OptimizedPartitionTracker` has a method on it called `get_partitions_to_optimize`, and I thought it was confusing that the two (quite different) methods had the same name

tests/clickhouse/optimize/test_optimize_scheduler.py
```diff
@@ -132,7 +132,7 @@ def test_subdivide_partitions(
     )
 
 
-last_midnight = (datetime.now() + timedelta(minutes=10)).replace(
+last_midnight = (datetime.now(UTC) + timedelta(minutes=10)).replace(
```
onewland, line 135: this fixes a test that was broken separately from these changes (e.g. this fails if run in a non-UTC timezone on the master branch)

snuba/clickhouse/optimize/optimize.py
```diff
@@ -400,3 +414,25 @@ def is_busy_merging(clickhouse: ClickhousePool, database: str, table: str) -> bo
...
+    """
+    Determines if a partition should be optimized today based on the partition name
+    and the current day of year.
+    """
+    if divide_partitions_count <= 0:
+        return True
+
+    return (
+        _hash_partition(partition_name) % divide_partitions_count
+        == datetime.now().timetuple().tm_yday % divide_partitions_count
```
onewland, line 437: this could cause inconsistency at year boundaries (two even days in a row), if we wanted to change the input day to something like "days since epoch (unix timestamp / 86400)"

snuba/clickhouse/optimize/optimize.py
volokluev: Can you add some documentation as to what this function is actually supposed to do? It's fairly long and it makes some design decisions, but it's not obvious at a glance for what purpose

snuba/clickhouse/optimize/optimize.py
```diff
@@ -400,3 +414,25 @@ def is_busy_merging(clickhouse: ClickhousePool, database: str, table: str) -> bo
...
+
+
+def should_optimize_partition_today(
+    partition_name: str, divide_partitions_count: int
+) -> bool:
+    """
+    Determines if a partition should be optimized today based on the partition name
+    and the current day of year.
+    """
+    if divide_partitions_count <= 0:
```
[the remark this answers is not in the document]
onewland: yeah, I think I'll get rid of the <1 case everywhere because 1 is the more clear default, and only in some cases will we override to 2

snuba/clickhouse/optimize/optimize.py
onewland: added a doc comment, let me know if you think it makes sense

snuba/clickhouse/optimize/optimize.py
```diff
@@ -123,9 +127,19 @@ def run_optimize_cron_job(
     except NoOptimizedStateException:
         # We don't have any recorded state of partitions needing optimization
         # for today. So we need to build it.
-        partitions = get_partitions_to_optimize(
+        partitions = get_partitions_from_clickhouse(
             clickhouse, storage, database, table, before
         )
+
+        if divide_partitions_count > 0:
```
nikhars, line 134: Nit: This will always be true since the default value is 1. Probably skip this check or change it to `> 1`

nikhars (approved):

PR #6750, opened 2025-01-10
fix(scripts): Grab everything in the path

phacops:
Input  looks like this. We need to grab the path inside the string.
```
CREATE TABLE somedb.sometable (column TYPE, ...) 
ENGINE = MergeTree('/clickhouse/tables/clustername/{shard}/default/tablename', '{replica}', sign)
```
Instead of trying to grab specific classes of characters (after all, we can't know if our table names have capitals, numbers or else), we'll simply grab everything between the two single quotes.

### DIFF
--- a/scripts/copy_tables.py
+++ b/scripts/copy_tables.py
@@ -21,9 +21,9 @@ def _get_client(
 
 
 def get_regex_match(curr_create_table_statement: str) -> str:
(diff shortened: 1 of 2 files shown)


volokluev:
should probably add a test

volokluev (approved):

PR #6756, opened 2025-01-12
fix(settings): specify VALID_RETENTION_DAYS for self-hosted

aldy505:
Fixes https://github.com/getsentry/self-hosted/issues/3421

I saw from the rust_snuba code, that we have this: https://github.com/getsentry/snuba/blob/acfbe9a31880ecfe15888386bf40a2afe2b16088/rust_snuba/src/processors/utils.rs#L10-L20

But if we follow the default value, it will always defaults to `30`. Whereas not everyone on the self-hosted instance uses value greater than 30. My own instance is set to 14.
https://github.com/getsentry/snuba/blob/acfbe9a31880ecfe15888386bf40a2afe2b16088/snuba/settings/__init__.py#L251

### Legal Boilerplate


aldy505:
> can we move the fetching of env vars into `settings/__init__.py`? I don't think there's a reason to expose those env vars for self-hosted only.

@untitaker Will it break your production servers? If you don't need to change anything on your prod configs, then I'll do that.

aldy505:
Also, can we merge this before the next release window (which is tomorrow)?

untitaker:
let's just keep it for now as that seems to be an established pattern.

PR #6758, opened 2025-01-13
ref(uptime): rebuild uptime storage table

JoshFerge:
after testing and implementing regions, we want to make a few changes to https://github.com/getsentry/snuba/issues/6696

- [x] use `scheduled_check_time` instead of `timestamp`, in order_by as there's a small chance we could have duplicates, and `scheduled_check_time` will allow us to de-duplicate.
- [x] rename `region_slug` -> `region`, as that's what we went with in upstream schema

after this is merged, we can modify our strorage to use the new table, and then delete the old table. no data is being read by the UI at the moment

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0022_uptime_monitors_init_v2.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0022_uptime_monitors_init_v2.py
@@ -17,6 +17,7 @@ columns: List[Column[Modifiers]] = [
     Column("project_id", UInt(64)),
     Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
     Column("uptime_subscription_id", UUID()),
+    Column("uptime_check_id", UUID()),
     Column("scheduled_check_time", DateTime64(3)),  # millisecond precision
     Column("timestamp", DateTime64(3)),  # millisecond precision
     Column("duration_ms", UInt(64)),
@@ -39,7 +40,7 @@ class Migration(migration.ClickhouseNodeMigration):
                 table_name=local_table_name,
                 columns=columns,
                 engine=table_engines.ReplacingMergeTree(
-                    order_by="(organization_id, project_id, toDateTime(scheduled_check_time), trace_id)",
+                    order_by="(organization_id, project_id, toDateTime(scheduled_check_time), trace_id, uptime_check_id)",
                     partition_by="(retention_days, toMonday(timestamp))",
                     settings={"index_granularity": "8192"},
                     storage_set=storage_set,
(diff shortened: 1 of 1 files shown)


wedamija (commented):
I don't think we should get rid of the uptime_check_id. It happens that we're storing the same thing now, but I don't think it's a good idea to assume they'll always be the same. Other than that this looks good!

JoshFerge:
> I don't think we should get rid of the uptime_check_id. It happens that we're storing the same thing now, but I don't think it's a good idea to assume they'll always be the same. Other than that this looks good!

updated.

snuba/snuba_migrations/events_analytics_platform/0022_uptime_monitors_init_v2.py
```diff
@@ -0,0 +1,75 @@
...
+dist_table_name = f"{table_prefix}_dist"
+
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("environment", String(Modifiers(nullable=True, low_cardinality=True))),
+    Column("uptime_subscription_id", UUID()),
+    Column("uptime_check_id", UUID()),
+    Column("scheduled_check_time", DateTime64(3)),  # millisecond precision
+    Column("timestamp", DateTime64(3)),  # millisecond precision
```
phacops, line 22: Is it really useful to have 2 timestamps that could be slightly different and store them with millisecond precision?

wedamija: One is when it was scheduled to happen, the other is when actually happened. I suppose it wouldn't affect users much if it was at second precision

phacops:
>  use scheduled_check_time instead of timestamp, in order_by as there's a small chance we could have duplicates, and scheduled_check_time will allow us to de-duplicate.
You can also not use a `ReplacingMergeTree` if you want to keep all rows.

JoshFerge:
> > use scheduled_check_time instead of timestamp, in order_by as there's a small chance we could have duplicates, and scheduled_check_time will allow us to de-duplicate.
> 
> You can also not use a `ReplacingMergeTree` if you want to keep all rows.

in this case we want the behavior a `ReplacingMergeTree` gives us

snuba/snuba_migrations/events_analytics_platform/0022_uptime_monitors_init_v2.py
JoshFerge: i'll update the `scheduled_check_time` to be second precision, and keep timestamp as is if that makes sense to you both, as it may end up being used in the UI / our customers could find ms precision useful if they're correlating the failed check with other things.

wedamija: Yeah, `scheduled_check_time` will never have ms anyway, so that makes sense

phacops (approved):

PR #6760, opened 2025-01-13
fix(optimize): make tests and optimize use UTC timestamps

onewland:
There are different expectations in the tests and application of optimize for whether UTC or local TZ times are being used. Fixing them to UTC fixes unreliable tests (at least those failing in my environment and @xurui-c )


nikhars (approved):

volokluev (approved):
🙏

PR #6762, opened 2025-01-13
ref(uptime): use region instead of region_slug

JoshFerge:
since we named it `region` upstream and in the table, we can just call it so here. https://github.com/getsentry/snuba/pull/6758


JoshFerge:
note that we've turned off message production, so these consumers are currently idling. should not result in errors in prod from this

wedamija (approved):

evanh (approved):

PR #6763, opened 2025-01-14
fix: snuba admin system query error messaging

kylemumma:
this fixes error messaging in the system queries tool of snuba admin. as a follow up i will see if i can expand this new error messaging to the rest of the tools.

heres how it used to look
Screenshot 2025-01-13 at 4 20 59 PM

and heres how it looks now
Screenshot 2025-01-14 at 3 05 30 PM


onkar (approved):
LGTM, is it possible to format the error text so that easier to read?

kylemumma:
@onkar done

PR #6764, opened 2025-01-14
fix(eap): remove effective sample size from confidence calculation for averages

shellmayr:
After careful consideration, this PR removes the effective sample size and replaces it with the sample size itself. The reason is that for very large sample sizes, low sample rates and highly heterogeneous sample rates, this actually causes issues in the confidence interval calculations ([see Notion for more details](https://www.notion.so/sentry/2024-12-18-Confidence-Evaluation-15e8b10e4b5d80aa89eff6579b704ca2?pvs=4)). 
This change also has a positive effect on the stability of the calculations, as there were some issues with NULL results that are not present with this version of the formula, probably caused by some divide-by-zero issues in the query for some buckets.

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
@@ -508,53 +508,47 @@ def get_confidence_interval_column(
         ),
         # confidence interval = Z * \sqrt{\frac{N * (\sum_{i=1}^n w_ix_i^2 - \frac{(\sum_{i=1}^n w_ix_i)^2}{N})}{n * (N-1) c* (N-1)}}
         #          ┌────────────────────────────┐
-        #          │                  ₙ
-        #          │                  ⎲
-        #          │      ₙ         ( ⎳  wᵢxᵢ)²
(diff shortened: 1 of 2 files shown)


volokluev:
Can you add a test?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -506,55 +506,49 @@ def get_confidence_interval_column(
...
-                        column(_get_count_column_alias(aggregation)),
-                        f.multiply(
-                            f.minus(column(f"{alias}_N"), literal(1)),
-                            f.minus(column(f"{alias}_N"), literal(1)),
+                        f.sumIf(
+                            sampling_weight_column,
+                            get_field_existence_expression(aggregation),
+                            alias=f"{alias}_N",
                         ),
+                        f.minus(column(f"{alias}_N"), literal(1)),
```
davidtsuk, line 551: this doesn't appear to correspond to the formula above. This would be N*(N-1) while the updated formula is n*N.

shellmayr: Great spot - should be corrected now!

PR #6765, opened 2025-01-14
fix(admin): Show errors in system queries results

volokluev:
https://github.com/user-attachments/assets/ced18488-83d4-4717-8dfc-7d7ab0594265


volokluev:
Dup of https://github.com/getsentry/snuba/pull/6763

PR #6766, opened 2025-01-14
Bump arroyo version

volokluev:
Bring in rebalancing and logging fixes

### Legal Boilerplate

### DIFF
--- a/rust_snuba/Cargo.toml
+++ b/rust_snuba/Cargo.toml
@@ -53,7 +53,7 @@ data-encoding = "2.5.0"
 zstd = "0.12.3"
 serde_with = "3.8.1"
 seq-macro = "0.3"
-sentry_arroyo = "2.19.5"
+sentry_arroyo = "2.19.11"
 regex = "1.11.1"
 
 


rust_snuba/Cargo.toml
```diff
@@ -53,7 +53,7 @@ data-encoding = "2.5.0"
 zstd = "0.12.3"
 serde_with = "3.8.1"
 seq-macro = "0.3"
-sentry_arroyo = "2.19.5"
+sentry_arroyo = "2.19.11"
```
ayirr7, line 56: I think the lockfile should be automatically be updated, not sure why it's not here. 

Maybe we can look at other PRs to see how they did it

volokluev: I think it's because it fails to build. This is because Arroyo introduced a breaking change without updating the major version

ayirr7: Just saw the conversation offline, thanks for finding it

untitaker (approved):

ayirr7 (approved):

PR #6770, opened 2025-01-14
Use GitHub Container Registry instead of Google Container Registry for snuba builds

onkar:
Currently, the snuba image is stored in Google Container Registry instead of the Github Container Registry. This puts limits on the external contributor's change passing through the CI. It specifically fails in self-hosted CI check stage with the error:

```
Run echo "We poll for the Docker image that the GCB/GHA build produces until it succeeds or this job times out."
  echo "We poll for the Docker image that the GCB/GHA build produces until it succeeds or this job times out."
  echo "Polling for $IMAGE_URL"
  [ -n "$(docker images -q $IMAGE_URL)" ] || timeout 20m bash -c 'until docker pull "$IMAGE_URL" 2>/dev/null; do sleep 10; done'
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    IMAGE_URL: us-central1-docker.pkg.dev/sentryio/snuba/image:f15ec56c42edcdc542b31dae438e6e0d0e2bb133
We poll for the Docker image that the GCB/GHA build produces until it succeeds or this job times out.
Polling for us-central1-docker.pkg.dev/sentryio/snuba/image:f15ec56c42edcdc542b31dae438e6e0d0e2bb133
Error: Process completed with exit code 124.
```

This PR moves the registry to Github Container Registry.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -438,7 +438,7 @@ jobs:
         uses: getsentry/self-hosted@master
         with:
           project_name: snuba
-          image_url: us-central1-docker.pkg.dev/sentryio/snuba/image:${{ github.event.pull_request.head.sha || github.sha }}
+          image_url: ghcr.io/getsentry/snuba:${{ github.event.pull_request.head.sha || github.sha }}
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


onkar:
Closing this as a dup of https://github.com/getsentry/snuba/pull/6800.

PR #6776, opened 2025-01-15
feat: implement aggregate filters on EndpointTraceItemTable

kylemumma:
This PR closes [this ticket](https://github.com/getsentry/eap-planning/issues/143)

It implements support for [this field in TraceItemTableRequest](https://github.com/getsentry/sentry-protos/blob/dc01de301c524ce75cea38750f1d5c3e6538cfcf/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto#L40)

Aggregation filters can now be specified like [this one](https://github.com/getsentry/snuba/pull/6776/files#diff-95d7f61d81f5bafda1267a2fb8d7ea3b9368b23e34dc2d24de38fb40532f9a1dR1105-R1117)

```
AggregationFilter(
    comparison_filter=AggregationComparisonFilter(
        aggregation=AttributeAggregation(
            aggregate=Function.FUNCTION_SUM,
            key=AttributeKey(
                type=AttributeKey.TYPE_FLOAT, name="my.float.field"
            ),
        ),
        op=AggregationComparisonFilter.OP_GREATER_THAN,
        val=350,
    )
)
```
which translates to
`HAVING sum(my.float.field) > 350`

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
@@ -7,6 +7,10 @@ from dataclasses import dataclass
 from functools import cached_property
 from typing import Any, Dict, List, Optional
 
+from sentry_protos.snuba.v1.endpoint_trace_item_table_pb2 import (
+    AggregationComparisonFilter,
+    AggregationFilter,
+)
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import (
     AttributeAggregation,
     ExtrapolationMode,
@@ -16,7 +20,7 @@ from sentry_protos.snuba.v1.trace_item_attribute_pb2 import (
 
 from snuba.query.dsl import CurriedFunctions as cf
 from snuba.query.dsl import Functions as f
-from snuba.query.dsl import column
+from snuba.query.dsl import and_cond, column, or_cond
 from snuba.query.expressions import (
     CurriedFunctionCall,
     Expression,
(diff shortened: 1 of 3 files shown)


volokluev (approved):
well done

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -648,3 +652,44 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
         )
 
     return agg_func_expr
+
+
+def aggregation_filter_to_expression(agg_filter: AggregationFilter) -> Expression:
```
davidtsuk, line 657: Since this function is only used for the trace item table endpoint, I think it would be to put it in resolver_trace_item_table.py

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -648,3 +652,44 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
...
+            if op_expr is None:
+                raise BadSnubaRPCRequestException(
+                    f"Unsupported aggregation filter op: {AggregationComparisonFilter.Op.Name(agg_filter.comparison_filter.op)}"
+                )
+            return op_expr(
+                aggregation_to_expression(agg_filter.comparison_filter.aggregation),
+                agg_filter.comparison_filter.val,
+            )
+        case "and_filter":
+            return and_cond(
```
davidtsuk, line 679: if and_filter contains less than 2 filters, this won't work since the `and` function in clickhouse requires at least 2 arguments. Even though it doesn't make sense to pass in less than 2 filters into an and function, we should handle those cases anyway as the API allows it.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -648,3 +652,44 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
...
+                *(
+                    aggregation_filter_to_expression(x)
+                    for x in agg_filter.and_filter.filters
+                )
+            )
+        case "or_filter":
+            return or_cond(
```
davidtsuk, line 686: same issue as with and_filter, this won't work if there are less than 2 filters

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -1040,6 +1044,332 @@ def test_table_with_group_by_columns_without_aggregation(
...
+            TraceItemColumnValues(
+                attribute_name="sum(my.float.field)",
+                results=[
+                    AttributeValue(val_float=1214.4),
+                    AttributeValue(val_float=3036),
+                ],
+            ),
+        ]
+
+        # now try with an or filter
```
davidtsuk, line 1248: I think the or filter should be a separate test

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -648,3 +652,44 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
...
+        AggregationComparisonFilter.OP_EQUALS: f.equals,
+        AggregationComparisonFilter.OP_NOT_EQUALS: f.notEquals,
+    }
+
+    match agg_filter.WhichOneof("value"):
+        case "comparison_filter":
+            op_expr = op_to_expr.get(agg_filter.comparison_filter.op)
+            if op_expr is None:
+                raise BadSnubaRPCRequestException(
+                    f"Unsupported aggregation filter op: {AggregationComparisonFilter.Op.Name(agg_filter.comparison_filter.op)}"
```
davidtsuk, line 672: nit: `agg_filter.comparison_filter.op` can be replaced with `op_expr` in the format string

davidtsuk (changes requested):
Overall, looks good! Let's also add some tests with different aggregations (not just sum).

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
kylemumma: I don't really want to make it a separate test bc theres so much boilerplate in the test setup i think its fine to keep it in one

phacops (approved):

PR #6777, opened 2025-01-15
chore(trace-item-types): throw an error if no trace item type is specified

colin-sentry:
https://github.com/getsentry/sentry/pull/83530 made the sentry backend send trace_item_type with every request, so now we can get rid of this unspecified => span assumption


volokluev (approved):

onkar:
@colin-sentry Can you add test that verifies that `BadSnubaRPCRequestException` is raised when no trace item type is specified.

colin-sentry:
> @colin-sentry Can you add test that verifies that `BadSnubaRPCRequestException` is raised when no trace item type is specified.

done

PR #6779, opened 2025-01-16
fix(admin): make profile events optional for trace queries

volokluev:
Sometimes the system.query_log table doesn't exist in the environment. Right now that breaks the entire tool. Gathering profile events can also be quite slow and for long queries it can time out. Allow it to be optional in the UI

### DIFF
--- a/snuba/admin/static/tracing/query_display.tsx
+++ b/snuba/admin/static/tracing/query_display.tsx
@@ -107,10 +107,24 @@ function QueryDisplay(props: {
             options={storages}
           />
         </div>
-        <ExecuteButton
-          onClick={executeQuery}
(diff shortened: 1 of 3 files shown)


MeredithAnya:
> Sometimes the system.query_log table doesn't exist in the environment.

When would this happen? Did we purposefully not want the system query log ?

MeredithAnya (approved):
I don't think my comments are blocking but I am surprised that the `system.query_log` wouldn't exist in some environments

volokluev:
@MeredithAnya gathering profile events can be slow (a few seconds), and sometimes we just want the trace output and the scrubbed query result, this would speed that up

PR #6780, opened 2025-01-16
feat(eap-rpc): Add a GetTrace endpoint

phacops:


kylemumma (approved):
from the protobuf definition
https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/snuba/v1/endpoint_get_trace.proto#L9-L12
> you can pass a list of attributes to return for each item type.

it seems like this is unsupported in the current implementation, should we notify the users in the response somehow if they attempt to do it?

> Each item will return every attributes by default

This implementation seems to always return only timestamp and trace_id, which contradicts this comment in the protobuf definition. Which behavior do we want?

How come we need `endpoint_get_trace` is `endpoint_get_traces` not sufficient?

phacops:
> How come we need endpoint_get_trace is endpoint_get_traces not sufficient?
GetTrace is meant to return one trace with items of that traces.
GetTraces is meant to return a list of trace-level info, not item-level info of that trace.

PR #6782, opened 2025-01-16
fix(eap-api): Snuba receives and returns doubles, while still supporting floats

xurui-c:
Update plan:

1. Sentry can receive doubles and floats
2. Snuba can receive (from Sentry) and return (to Sentry) doubles and floats. In the case of aggregation, replace float with double. This is fine because Sentry handles the case of receiving doubles already
3. Sentry asks for doubles only

This PR addresses step 2. Step 1's PR is here: https://github.com/getsentry/sentry/pull/83566

For every test that involved floats, I added a corresponding test with doubles. It ensures backward compatibility by verifying that Snuba can handle both floats and doubles

### DIFF
--- a/snuba/web/rpc/v1/endpoint_get_traces.py
+++ b/snuba/web/rpc/v1/endpoint_get_traces.py
@@ -44,6 +44,13 @@ from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
 _DEFAULT_ROW_LIMIT = 10_000
 _BUFFER_WINDOW = 2 * 3600  # 2 hours
 
+
+def _convert_key_to_support_doubles_and_floats_for_backward_compat(
+    key: TraceAttribute.Key.ValueType,
+) -> TraceAttribute.Key.ValueType:
+    return TraceAttribute.Key.ValueType(-1 * key)
+
+
 _ATTRIBUTES: dict[
     TraceAttribute.Key.ValueType,
     tuple[str, AttributeKey.Type.ValueType],
@@ -69,6 +76,16 @@ _ATTRIBUTES: dict[
         AttributeKey.Type.TYPE_STRING,
     ),
 }
+# for every AttributeKey of TYPE_FLOAT a user may add during the backward compat period, this adds the TYPE_DOUBLE equivalent
+_attributes_backward_compat = dict()
+for k in _ATTRIBUTES:
+    v = _ATTRIBUTES[k]
+    if v[1] == AttributeKey.Type.TYPE_FLOAT:
+        _attributes_backward_compat[
+            _convert_key_to_support_doubles_and_floats_for_backward_compat(k)
+        ] = (v[0], AttributeKey.Type.TYPE_DOUBLE)
+_ATTRIBUTES.update(_attributes_backward_compat)
+
 _TYPES_TO_CLICKHOUSE: dict[
     AttributeKey.Type.ValueType,
     tuple[str, Callable[[Any], AttributeValue]],
@@ -84,6 +101,10 @@ _TYPES_TO_CLICKHOUSE: dict[
     AttributeKey.Type.TYPE_FLOAT: (
         "Float64",
         lambda x: AttributeValue(val_float=float(x)),
+    ),
+    AttributeKey.Type.TYPE_DOUBLE: (
+        "Float64",
+        lambda x: AttributeValue(val_double=float(x)),
     ),
 }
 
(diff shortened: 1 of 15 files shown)


tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -555,6 +556,39 @@ def test_table_with_aggregates(self, setup_teardown: Any) -> None:
             limit=5,
         )
         response = EndpointTraceItemTable().execute(message)
+
```
xurui-c, line 559: im going to remove these

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -69,6 +76,16 @@
         AttributeKey.Type.TYPE_STRING,
     ),
 }
+# for every AttributeKey of TYPE_FLOAT a user may add during the backward compat period, this adds the TYPE_DOUBLE equivalent
```
phacops, line 79: We're not letting users choose the type they want in this API. I would remove all the compatibility and just change the timestamp type to `TYPE_DOUBLE`.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -37,6 +37,8 @@
 
 CUSTOM_COLUMN_PREFIX = "__snuba_custom_column__"
 
+_FLOATING_POINT_PRECISION = 8
```
phacops, line 40: A `Float32` would provide us with at least 8 digits of precision already I think, we should extend that to 12 if we're going to support `Float64`.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -612,7 +614,11 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
     alias = aggregation.label if aggregation.label else None
     alias_dict = {"alias": alias} if alias else {}
     function_map: dict[Function.ValueType, CurriedFunctionCall | FunctionCall] = {
-        Function.FUNCTION_SUM: f.sum(f.multiply(field, sign_column), **alias_dict),
+        Function.FUNCTION_SUM: f.round(
+            f.sum(f.multiply(field, sign_column)),
+            _FLOATING_POINT_PRECISION,
+            **alias_dict,
+        ),
```
phacops, line 621: Any reason not to round other functions like the percentiles?

phacops (changes requested):

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -612,7 +614,11 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
     alias = aggregation.label if aggregation.label else None
     alias_dict = {"alias": alias} if alias else {}
     function_map: dict[Function.ValueType, CurriedFunctionCall | FunctionCall] = {
-        Function.FUNCTION_SUM: f.sum(f.multiply(field, sign_column), **alias_dict),
+        Function.FUNCTION_SUM: f.round(
```
davidtsuk, line 617: why do we need to round?

kylemumma:
Hi I don't really understand what exactly you're trying to do so im finding it hard to review. Do you mind adding more context behind the PR to the description?

phacops (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
```diff
@@ -612,23 +614,42 @@ def aggregation_to_expression(aggregation: AttributeAggregation) -> Expression:
...
             **alias_dict,
         ),
-        Function.FUNCTION_P50: cf.quantile(0.5)(field, **alias_dict),
-        Function.FUNCTION_P75: cf.quantile(0.75)(field, **alias_dict),
-        Function.FUNCTION_P90: cf.quantile(0.9)(field, **alias_dict),
-        Function.FUNCTION_P95: cf.quantile(0.95)(field, **alias_dict),
-        Function.FUNCTION_P99: cf.quantile(0.99)(field, **alias_dict),
-        Function.FUNCTION_AVG: f.avg(field, **alias_dict),
+        Function.FUNCTION_P50: f.round(
+            cf.quantile(0.5)(field), _FLOATING_POINT_PRECISION, **alias_dict
```
volokluev, line 636: why do you need to add this _FLOATING_POINT_PRECISION argument?

volokluev (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/common/aggregation.py
xurui-c: We still have floating point precision errors :(

PR #6784, opened 2025-01-16
feat(devservices): Use new devservices for sentry integration tests

hubertdeng123:


IanWoodard (approved):

IanWoodard (commented):
Do we need to set the `USE_NEW_DEVSERVICES` env var?

hubertdeng123:
> Do we need to set the USE_NEW_DEVSERVICES env var?
No, that's only needed when relay is brought up for tests

IanWoodard (approved):

onkar (approved):

PR #6788, opened 2025-01-17
fix(errors): Support writer options in the rust consumer

evanh:
Currently the rust consumer doesn't load the writer options field in the config. This is needed by errors
to  insert fields that don't exist in the tables yet.

Errors is the first user of this, to allow populating a new Nested column without actually writing it yet.


volokluev:
Does the rust processor read this setting?

davidtsuk (approved):

evanh:
> Does the rust processor read this setting?

@volokluev this was a good point. I updated the PR to do this.

rust_snuba/src/strategies/clickhouse/batch.rs
```diff
@@ -69,6 +72,10 @@ impl BatchFactory {
                 query_params.push_str("&async_insert=1&wait_for_async_insert=1");
             }
         }
+        for (key, value) in writer_options {
+            query_params.push_str(&format!("&{key}={value}"));
+        }
+        println!("query_params: {}", query_params);
```
volokluev, line 78: 🔪

untitaker (approved):

Issue #6751, opened 2025-01-11
Allow copying CSV results from ClickHouse system table through snuba admin


onkar:
### Environment

Snuba admin

### Steps to Reproduce

Snuba admin does not provide a way to copy data from ClickHouse system tables. Fix it to do that.

### Expected Result

Snuba admin has a way (like exporting results) that allows to copy data from system tables.

### Actual Result

There is no way to copy ClickHouse system table data.

PR #6789, opened 2025-01-17
feat(admin): Add a standardized component for copying query results

evanh:
At lot of components offer the ability to copy the results of a run query, but each had a slightly
different way of providing that functionality. That meant that not all functionality across the
components.

Add a standard component for these buttons. This will also ensure that system queries now have a
CSV copy results.

Screenshot 2025-01-17 at 3 41 35 PM

Resolves https://github.com/getsentry/snuba/issues/6751


volokluev (approved):
beautiful

onkar (approved):
Thanks for adding this support!

PR #6791, opened 2025-01-20
feat(eap): adapt sum confidence interval formula in accordance with avg

shellmayr:
As `sum = N * average`, the confidence interval of the sum should also be equivalent to N times the confidence interval of the average. Therefore, we can re-use the formula from averages here (of course amended by this multiplier), to end up with consistent results for these two aggregates.


davidtsuk (approved):
LGTM

PR #6796, opened 2025-01-20
ci: Fix self-hosted test runs

BYK:


onkar:
@BYK I took a stab at addressing somewhat similar problem in https://github.com/getsentry/snuba/pull/6770. Are you trying to solve the same issue in this PR?

BYK:
@onkar sort of. I'd say they are orthogonal. There are also many things to improve and optimize for Snuba's CI pipeline. For instance I ran the CI 3-4 times in this PR without _any_ changes to Snuba itself and every single time it build the Docker image from scratch taking north of 5 minutes. Clearly whatever caching we think in place is not working at all.

Anyway, I think you can merge this and keep building your patch on top of it. This PR is specifically concerned with the self-hosted e2e job. The only reason I'm switching to `ghcr.io` is because that seems to be the thing we are building in the previous image step, not the GCR image.

untitaker (approved):

untitaker:
in #6798 i tested that this PR does indeed fix third-party PRs. I think #6770 can be closed. Also, self-hosted e2e is now a required status check as per request from BYK. as failure blocks self-hosted releases

PR #6798, opened 2025-01-21
testing thirdparty ci

untitaker:


untitaker:
context: after https://github.com/getsentry/snuba/pull/6796 has been merged, @BYK wanted us to make self-hosted e2e required tests. I think the blocker for making that required is to ensure it works for third-party contributions and absence of flakiness. Let's see how it does. Note that this PR is opened from a branch in my own fork of snuba.

untitaker:
seems to run OK. `self-hosted-end-to-end` is now a required status check

PR #6799, opened 2025-01-21
fix(uptime): duration is nullable

JoshFerge:
Fixes SNUBA-6R7

Wish there was a way for snuba to auto generate a schema based on the [kafka schema type.](https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/snuba-uptime-results.v1.schema.json) missed this when typing the message.

### DIFF
--- a/rust_snuba/src/processors/uptime_monitor_checks.rs
+++ b/rust_snuba/src/processors/uptime_monitor_checks.rs
@@ -29,7 +29,7 @@ pub fn deserialize_message(payload: &[u8]) -> anyhow::Result<(Vec<UptimeMonitorC
         uptime_check_id: monitor_message.guid,
         scheduled_check_time: monitor_message.scheduled_check_time_ms,
         timestamp: monitor_message.actual_check_time_ms,
-        duration_ms: monitor_message.duration_ms,
+        duration_ms: monitor_message.duration_ms.unwrap_or(0),
         region: monitor_message.region.unwrap_or_default(),
         check_status: monitor_message.status,
         check_status_reason: monitor_message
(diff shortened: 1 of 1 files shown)


rust_snuba/src/processors/uptime_monitor_checks.rs
```diff
@@ -146,4 +146,50 @@ mod tests {
...
+            "project_id": 1,
+            "retention_days": 30,
+            "region": "global",
+            "environment": "prod",
+            "subscription_id": "123e4567-e89b-12d3-a456-426614174000",
+            "guid": "550e8400-e29b-41d4-a716-446655440000",
+            "scheduled_check_time_ms": 1702659277,
+            "actual_check_time_ms": 1702659277,
+            "duration_ms": null,
+            "status": "ok",
```
evanpurkhiser, line 163: I think duration is only null for misses, should we change this to a miss to better reflect when duration would be null?

https://github.com/getsentry/sentry-kafka-schemas/blob/fddec4d390658d8177d46da6de1176f2e4534f77/schemas/uptime-results.v1.schema.json#L98

JoshFerge: i don't think it really matters for the sake of the test, but sure.

evanpurkhiser: I think it's always good to try and make tests reflect reality as much as possible (within reason), since it can really help to avoid confusion.

JoshFerge: updated!

evanh (approved):

volokluev (approved):

evanpurkhiser (approved):

PR #6800, opened 2025-01-21
ref(ci): Refactor self-hosted e2e and dockerhub steps to not use GCR image

untitaker:
we have three builds:

* snuba-ci GHCR (ci.yml) -- this includes test dependencies
* snuba GHCR (image.yml)
* snuba GCR (cloudbuild.yml)

the latter two images are identical, but it's too much hassle to poll GCR and re-publish, or pull one image from another, so we just build it twice for now

then:

* snuba GCR is re-published to dockerhub (happens in ci.yml by polling GCR)
* snuba GCR is used to run self-hosted e2e tests (this has now been changed)

now, to the changes in this PR:

* self-hosted e2e would like to run on the same image as dockerhub.
  since https://github.com/getsentry/snuba/pull/6796 those two images
  have diverged since one is using snuba-ci (GHCR) and the other GCR. this
  re-unifies them on top of snuba (GHCR)

* publish-to-dockerhub was polling GCR. instead, it could just wait for
  the job result of GHCR


aldy505:
Dupe? https://github.com/getsentry/snuba/pull/6770

onkar:
> Dupe? #6770

Closed #6770

untitaker:
@aldy505 oh, that would be nice, to not build the same image twice. we didn't think it was worth the time to figure out how to do that and were fine with building the same image twice. I think cloudbuild.yml can become a noop and we can publish from image.yml. if you want to port that from relay go ahead.

curious @Dav1dde did you have to do anything in terms of secret mgmt to make this work?

Dav1dde:
> curious @Dav1dde did you have to do anything in terms of secret mgmt to make this work?

Yes. ([Internal PR](https://github.com/getsentry/security-as-code/pull/1187))

If you add snuba to that list, you can authenticate with the artifact registry and push the image from GHA, similar to what I updated in Symbolicator and Relay has been doing for a long time, making the cloudbuild obsolete.

Both repos also contain two different ways to build a multiarch arm64/amd64 container, maybe one of these methods you can use to easily adapt for snuba as well.

Relay builds two binaries and builds the multiarch image in a single step using the (cross-)compiled binaries. Symbolicator merges two independent images into a single multiarch image (easier to use if you have a complicated Dockerfile).

Dav1dde:
Yeah, that's what symbolicator does now, I just took the existing Dockerfile and added a native ARM64 build, then merges the images together. Ideally I'd like to do it differently, but that was just the easiest way to make it work.

aldy505:
Building the arm64 image on the regular amd64 will take a long time, though. 

Bobby (EM of devinfra) said to me personally that Sentry had some problem with github runner exhaustion, due to using too many runners.

.github/workflows/image.yml
```diff
@@ -38,3 +38,63 @@ jobs:
...
+          docker push getsentry/snuba:${GITHUB_SHA}
+          # second, the short sha of the commit
+          docker tag ${IMAGE_URL} getsentry/snuba:${SHORT_SHA}
+          docker push getsentry/snuba:${SHORT_SHA}
+          # finally, nightly
+          docker tag ${IMAGE_URL} getsentry/snuba:nightly
+          docker push getsentry/snuba:nightly
+
+  self-hosted-end-to-end:
+    needs: build
```
evanh, line 90: Should this be "publish-to-dockerhub"?

evanh (approved):

PR #6804, opened 2025-01-22
Remove CDC from EVENTS migration group

volokluev:
CDC no longer exists:

https://github.com/getsentry/ops/pull/10334/files

### DIFF
--- a/snuba/migrations/groups.py
+++ b/snuba/migrations/groups.py
@@ -4,6 +4,7 @@ from typing import Dict, Set
 from snuba.clusters.storage_sets import StorageSetKey
 from snuba.datasets.readiness_state import ReadinessState
 from snuba.migrations.group_loader import (
+    CDCLoader,
     DiscoverLoader,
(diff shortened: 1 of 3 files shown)


ayirr7 (approved):

MeredithAnya:
wanna also remove the references in the settings files like https://github.com/getsentry/snuba/blob/caddadc9f8007a8e61cb507ebe9452d04c68a1fa/snuba/settings/__init__.py#L100

snuba/migrations/groups.py
```diff
@@ -82,12 +83,16 @@ def __init__(
         storage_sets_keys={StorageSetKey.MIGRATIONS},
         readiness_state=ReadinessState.COMPLETE,
     ),
+    MigrationGroup.CDC: _MigrationGroup(
+        loader=SystemLoader(),
```
ayirr7, line 87: We might need to define a different loader here

volokluev: you're so right bestie

PR #6814, opened 2025-01-24
[high cardinality queries] track OOM on trace item table and time series endpoints

xurui-c:
https://github.com/getsentry/projects/issues/463

### DIFF
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -509,17 +509,6 @@ def _raw_query(
             error_code=error_code,
             triggered_rate_limiter=str(trigger_rate_limiter),
         )
-        if error_code == 241:
-            raise QueryException.from_args(
-                cause.__class__.__name__,
-                str(cause),
-                {
-                    "stats": stats,
-                    "sql": sql,
-                    "experiments": clickhouse_query.get_experiments(),
-                    "code": 241,
-                },
-            ) from cause
         raise QueryException.from_args(
             # This exception needs to have the message of the cause in it for sentry
             # to pick it up properly
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -509,6 +509,17 @@ def _raw_query(
             error_code=error_code,
             triggered_rate_limiter=str(trigger_rate_limiter),
         )
+        if error_code == 241:
+            raise QueryException.from_args(
+                cause.__class__.__name__,
+                str(cause),
+                {
+                    "stats": stats,
+                    "sql": sql,
+                    "experiments": clickhouse_query.get_experiments(),
+                    "code": 241,
+                },
+            ) from cause
         raise QueryException.from_args(
             # This exception needs to have the message of the cause in it for sentry
             # to pick it up properly
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -183,11 +183,6 @@ def execute_query(
         robust=robust,
     )
 
-
-    print("readerrrr", reader)
-
-    print("resultttt", result)
-
     timer.mark("execute")
     stats.update(
         {
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -183,6 +183,11 @@ def execute_query(
         robust=robust,
     )
 
+
+    print("readerrrr", reader)
+
+    print("resultttt", result)
+
     timer.mark("execute")
     stats.update(
         {
--- a/snuba/web/rpc/__init__.py
+++ b/snuba/web/rpc/__init__.py
@@ -198,7 +198,12 @@ def run_rpc_handler(
 
     try:
         return cast(ProtobufMessage, endpoint.execute(deserialized_protobuf))
-    except (RPCRequestException, QueryException) as e:
+    except RPCRequestException as e:
+        return convert_rpc_exception_to_proto(e)
+    except QueryException as e:
+        if "code" in e.extra_data and e.extra["code"] == 241:
+            endpoint.metrics.increment("OOM_query")
+            sentry_sdk.capture_exception(e)
         return convert_rpc_exception_to_proto(e)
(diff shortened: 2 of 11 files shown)


snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -104,4 +109,10 @@ def _execute(self, in_msg: TraceItemTableRequest) -> TraceItemTableResponse:
                 "This endpoint requires meta.trace_item_type to be set (are you requesting spans? logs?)"
             )
         resolver = self.get_resolver(in_msg.meta.trace_item_type)
-        return resolver.resolve(in_msg)
+        try:
+            return resolver.resolve(in_msg)
+        except Exception as e:
+            if "DB::Exception: Memory limit (for query) exceeded" in str(e):
```
onkar, line 115: Should code `241` be used instead of the string matching? That way if ClickHouse changes the exception message, it will still work.

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -105,4 +111,12 @@ def _execute(self, in_msg: TimeSeriesRequest) -> TimeSeriesResponse:
...
             )
         resolver = self.get_resolver(in_msg.meta.trace_item_type)
-        return resolver.resolve(in_msg)
+        try:
+            return resolver.resolve(in_msg)
+        except QueryException as e:
+            if e.extra[
+                "code"
+            ] == 241 or "DB::Exception: Memory limit (for query) exceeded" in str(e):
+                metrics.increment("endpoint_trace_item_table_OOM")
```
volokluev, line 120: You should instead use the `metrics` property included in the class

https://github.com/getsentry/snuba/blob/master/snuba/web/rpc/__init__.py#L105-L111

`self.metrics.increment("oom_query")`

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -104,4 +110,12 @@ def _execute(self, in_msg: TraceItemTableRequest) -> TraceItemTableResponse:
...
-        return resolver.resolve(in_msg)
+        try:
+            return resolver.resolve(in_msg)
+        except QueryException as e:
+            if e.extra[
+                "code"
+            ] == 241 or "DB::Exception: Memory limit (for query) exceeded" in str(e):
+                metrics.increment("endpoint_trace_item_table_OOM")
+                sentry_sdk.capture_exception(e)
+            raise BadSnubaRPCRequestException(str(e))
```
volokluev, line 121: I would actually consider putting this in the RPC base class. I think it's generic enough to apply to all endpoints. I'll leave that up to you though

snuba/web/rpc/v1/endpoint_time_series.py
```diff
@@ -105,4 +111,12 @@ def _execute(self, in_msg: TimeSeriesRequest) -> TimeSeriesResponse:
                 "This endpoint requires meta.trace_item_type to be set (are you requesting spans? logs?)"
             )
         resolver = self.get_resolver(in_msg.meta.trace_item_type)
-        return resolver.resolve(in_msg)
+        try:
+            return resolver.resolve(in_msg)
+        except QueryException as e:
+            if e.extra[
+                "code"
+            ] == 241 or "DB::Exception: Memory limit (for query) exceeded" in str(e):
```
phacops, line 119: Do we really need to match on the code and the string? Or do we need to look for the full message as opposed to just `Memory limit`?

snuba/web/rpc/v1/endpoint_trace_item_table.py
phacops, line 121: I think that's something we'd want for all endpoints, so I'd put that in the base handler.

snuba/web/rpc/v1/endpoint_time_series.py
xurui-c: We can do either. I'm kind of making the assumption here that [ClickHouse's OOM error will always be 241](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/ErrorCodes.cpp#L208). As an extra measure I'm also doing string matching with "DB::EXception: Memory..."

I'm not string matching for the full message since the full message can be [extremely long and differs for every error](https://github.com/getsentry/eap-planning/issues/111#:~:text=%7B%20%22code%22%3A%20500%2C%20%22message,bin/clickhouse%5Cn%22%20%7D), but it seems like "DB::Exception: Memory limit (for query) exceeded" is the same in every OOM error

snuba/web/db_query.py
```diff
@@ -509,6 +509,17 @@ def _raw_query(
...
+            raise QueryException.from_args(
+                cause.__class__.__name__,
+                str(cause),
+                {
+                    "stats": stats,
+                    "sql": sql,
+                    "experiments": clickhouse_query.get_experiments(),
+                    "code": 241,
+                },
+            ) from cause
```
phacops, line 522: Do we have to make it a specific case? Why not set the code anyway for all?

snuba/web/rpc/__init__.py
```diff
@@ -198,7 +198,12 @@ def run_rpc_handler(
 
     try:
         return cast(ProtobufMessage, endpoint.execute(deserialized_protobuf))
-    except (RPCRequestException, QueryException) as e:
+    except RPCRequestException as e:
+        return convert_rpc_exception_to_proto(e)
+    except QueryException as e:
+        if "error_code" in e.extra["stats"] and e.extra["stats"]["error_code"] == 241:
+            endpoint.metrics.increment("OOM_query")
+            sentry_sdk.capture_exception(e)
```
volokluev, line 206: this should be in the RPC class

volokluev (approved):

phacops (approved):

PR #6817, opened 2025-01-24
feat: eap EndpointTraceItemTable spare aggregate attribute transformer

kylemumma:
this pr addresses this ticket for EndpointTraceItemTable. https://github.com/getsentry/eap-planning/issues/158

It makes it so when we have a query like
```
SELECT sum(wing.count)
...
GROUP BY animal_type
```
it will add
```
SELECT sum(wing.count)
WHERE hasAttribute(wing.count)
...
GROUP BY animal_type
```

## Major change
The major change in this PR is the addition of `SparseAggregateAttributeTransformer` this is a class responsible for transforming the original `TraceItemTableRequest` into the new one with the additional filter. This query transformation step is then added as a step in the resolver that all requests go through.

I think it should be sufficient to only do EndpointTraceItemTable and not EndpointTimeSeries since timeseries wont graph the 0 data but ill ask tony to make sure. If timeseries is needed it should only be 1-2 days of extra work.


tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -2101,6 +2136,94 @@ def test_bad_aggregation_filter(self, setup_teardown: Any) -> None:
         with pytest.raises(BadSnubaRPCRequestException):
             EndpointTraceItemTable().execute(message)
 
+    def test_sparse_aggregate(self, setup_teardown: Any) -> None:
```
kylemumma, line 2139: this test ensures desired behavior as described in this ticket https://github.com/getsentry/eap-planning/issues/158
that there arent the 0 rows

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -131,6 +131,41 @@ def gen_message(
     }
 
 
+def write_eap_span(
```
kylemumma, line 134: this function is unrelated to the PR, I added it to make `gen_message` easier to use in my tests

tests/web/rpc/v1/visitors/test_sparse_aggregate_attribute_transformer.py
kylemumma, line 1: these are unit tests for the transformer

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -2101,6 +2136,94 @@ def test_bad_aggregation_filter(self, setup_teardown: Any) -> None:
...
+            ),
+            columns=[
+                Column(
+                    key=AttributeKey(type=AttributeKey.TYPE_STRING, name="animal_type")
+                ),
+                Column(
+                    aggregation=AttributeAggregation(
+                        aggregate=Function.FUNCTION_SUM,
+                        key=AttributeKey(
+                            type=AttributeKey.TYPE_FLOAT, name="wing.count"
```
xurui-c, line 2190: Let's use AttributeKey.TYPE_DOUBLE

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -2101,6 +2136,94 @@ def test_bad_aggregation_filter(self, setup_teardown: Any) -> None:
...
+                attribute_name="animal_type",
+                results=[
+                    AttributeValue(val_str="bird"),
+                    AttributeValue(val_str="chicken"),
+                ],
+            ),
+            TraceItemColumnValues(
+                attribute_name="sum(wing.count)",
+                results=[
+                    AttributeValue(val_float=20),
```
xurui-c, line 2221: Can we use val_double instead of val_float please

snuba/web/rpc/v1/visitors/sparse_aggregate_attribute_transformer.py
```diff
@@ -0,0 +1,61 @@
...
+        chicken       |   12
+
+    why? because the columns for chicken and bird don't have the attribute "wing.count"
+    but by default it sets it to 0 when the attribute is not present.
+
+    This class tranforms TraceItemTableRequest adding a filter `hasAttribute("wing.count")`
+    for all attributes that are used in aggregate functions. (this ensures the correct behavior as
+    described above)
+
+    It does not modify the given request, it returns a new request with the filter added.
```
xurui-c, line 33: why?

kylemumma: I just decided to do it this way arbitrarily, I think it could go either way

davidtsuk (commented):
I have the following suggestion:
When we do aggregations we already transform them into `aggregate_functionIf` which has the needed conditions. The problem is that the `If` combinator in clickhouse returns the default value (0) if no rows match the condition. One way to address this is to add the `orNull` combinator to the aggregation so that it returns Null when no rows match the condition. We could then decide what to do with the null values depending on the behaviour we want. This would also be the first part of addressing https://github.com/getsentry/eap-planning/issues/169.

This would also ensure we don't add additional overhead from copying the request and adding conditions that we already add to the aggregations.

onkar:
> One way to address this is to add the `orNull` combinator to the aggregation so that it returns Null when no rows match the condition. We could then decide what to do with the null values depending on the behaviour we want. This would also be the first part of addressing [getsentry/eap-planning#169](https://github.com/getsentry/eap-planning/issues/169).

@davidtsuk I don't think I follow it. https://github.com/getsentry/eap-planning/issues/169 requires us to distinguish between empty and null, whereas you seem to suggest that we should return null when no rows match the condition (i.e. empty result). Isn't that contradictory to what's expected in the ticket?

davidtsuk:
> @davidtsuk I don't think I follow it. [getsentry/eap-planning#169](https://github.com/getsentry/eap-planning/issues/169) requires us to distinguish between empty and null, whereas you seem to suggest that we should return null when no rows match the condition (i.e. empty result). Isn't that contradictory to what's expected in the ticket?
@onkar Our protobuf definition currently doesn't allow us to distinguish between a null result and a default result (i.e. empty). The point I was trying to make is that we should distinguish between them because a default value (e.g. 0) can actually represent a result (e.g. if you add -2 + 2) so we cannot use empty/default values to represent the case where no rows were aggregated.

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table_extrapolation.py
```diff
@@ -553,42 +552,36 @@ def test_count_reliability_with_group_by_backward_compat(self) -> None:
...
-        assert measurement_tags == ["foo", "bar"]
+        assert measurement_tags == ["foo"]
 
         measurement_sums = [v.val_double for v in response.column_values[1].results]
         measurement_reliabilities = [v for v in response.column_values[1].reliabilities]
-        assert measurement_sums == [sum(range(5)), 0]
+        assert measurement_sums == [sum(range(5))]
         assert measurement_reliabilities == [
             Reliability.RELIABILITY_LOW,
-            Reliability.RELIABILITY_UNSPECIFIED,
```
davidtsuk, line 563: why have these changed?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -2166,6 +2201,112 @@ def make_request(page_token: PageToken) -> TraceItemTableRequest:
             time = datetime.fromisoformat(val.val_str)
             assert time > last_timestamp
 
+    def test_sparse_aggregate(self, setup_teardown: Any) -> None:
+        """
+        This test ensures that when aggregates are done, groups that dont have the attribute being
```
davidtsuk, line 2206: the comment seems to be outdated now that the test changed

snuba/web/rpc/v1/visitors/sparse_aggregate_attribute_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+                    filters=[
+                        TraceItemFilter(exists_filter=ExistsFilter(key=key))
+                        for key in agg_keys
+                    ]
+                )
+            )
+            # combine the new filters with the existing one
+            if self.req.HasField("filter"):
+                new_filter = TraceItemFilter(
+                    and_filter=AndFilter(filters=[self.req.filter, filter_to_add])
```
davidtsuk, line 62: shouldn't this be an `OrFilter` as we discussed?

davidtsuk: nvm, I was looking at the wrong thing

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table_extrapolation.py
kylemumma: if you look to the top of the test, the messages with `bar` have no `custom_measurement` https://github.com/getsentry/snuba/pull/6817/files/cb324ebce1e4392d60bc65138a8ae15d0af652b9#diff-440653fbc080c88c781f59384bcd3722c93ec37540b50595da42233e67449c1aR478-R481

it used to be the case that those rows would still be included in the results, but due to the new query processor that adds the `hasAttribute(custom_measurement)` filter, they are no longer part of the result

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
kylemumma: i think its still relevant

davidtsuk (approved):

phacops (approved):

PR #6825, opened 2025-01-25
ci: provide arm64 image build

aldy505:
Effort for https://github.com/getsentry/self-hosted/issues/1585

Build result for arm64: https://github.com/getsentry/snuba/actions/runs/12960880084/job/36155350053?pr=6825

### Legal Boilerplate


aldy505:
cc @untitaker @BYK @hubertdeng123

.github/workflows/image.yml
```diff
@@ -4,9 +4,19 @@ on:
...
-      image_tag: ${{ steps.get_image_tag.outputs.image_tag }}
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+    runs-on: |-
+      ${{fromJson('{
+        "amd64": "ubuntu-22.04",
+        "arm64": "ubuntu-22.04-arm"
+      }')[matrix.arch] }}
```
BYK, line 14: 🤯

BYK:
Yeah, I cannot really review this. Will leave to @untitaker

.github/workflows/image.yml
untitaker: if you want to see some real crazy usage of fromJson, check out what cargo-dist does. it seems that one can generate matrices programmatically, put them into job output and then reference them here.

.github/workflows/image.yml
```diff
@@ -41,27 +49,44 @@ jobs:
...
+        run: docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+
+      - name: Assemble Sha Image
+        run: |
+          docker buildx imagetools create -t "ghcr.io/getsentry/snuba:${{ github.sha }}" \
+            "ghcr.io/getsentry/snuba:arm64-${{ github.sha }}" \
+            "ghcr.io/getsentry/snuba:amd64-${{ github.sha }}"
+
+      - name: Assemble Latest Image
+        if: github.ref_name == 'main'
```
untitaker, line 74: our branch is called master (still)

.github/workflows/image.yml
```diff
@@ -4,9 +4,19 @@ on:
...
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+    runs-on: |-
+      ${{fromJson('{
+        "amd64": "ubuntu-22.04",
+        "arm64": "ubuntu-22.04-arm"
+      }')[matrix.arch] }}
+    env:
+      IMG_VERSIONED: ghcr.io/getsentry/vroom:${{ matrix.arch }}-${{ github.sha }}
```
untitaker, line 16: copypaste error, we're not vroom

.github/workflows/image.yml
```diff
@@ -41,27 +49,44 @@ jobs:
...
+      - name: Assemble Sha Image
+        run: |
+          docker buildx imagetools create -t "ghcr.io/getsentry/snuba:${{ github.sha }}" \
+            "ghcr.io/getsentry/snuba:arm64-${{ github.sha }}" \
+            "ghcr.io/getsentry/snuba:amd64-${{ github.sha }}"
+
+      - name: Assemble Latest Image
+        if: github.ref_name == 'main'
+        run: |
+          docker buildx imagetools create -t "ghcr.io/getsentry/snuba:latest" \
```
untitaker, line 76: are you 100% sure that `imagetools create` pushes to the registry? though i don't see a `--push` flag avail

Dav1dde: Yes it does, this is [taken from Symbolicator](https://github.com/getsentry/symbolicator/blob/master/.github/workflows/image.yml#L171-L182) and that's how it pushes its images.

.github/workflows/image.yml
Dav1dde: Just to note, this is absolutely not necessary here, I did it this way to not have the runner show up in the generated job name in CI.

aldy505: > Just to note, this is absolutely not necessary here, I did it this way to not have the runner show up in the generated job name in CI.

No no, it's useful. We need to use the correct GHA runner with the corresponding arch name (amd64/arm64)

Dav1dde: > No no, it's useful.

It's not, you can just put it in the matrix. The purpose of this is simply to not have the runner name in the job name.

aldy505: > > No no, it's useful.
> 
> 
> 
> It's not, you can just put it in the matrix. The purpose of this is simply to not have the runner name in the job name.

Now it makes sense lol.

.github/workflows/image.yml
```diff
@@ -2,67 +2,120 @@ on:
...
+
 jobs:
   build:
-    outputs:
-      image_tag: ${{ steps.get_image_tag.outputs.image_tag }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-24.04
+            platform: amd64
```
onkar, line 12: Can we run amd64 images natively on arm64 without using emulation?

onkar (commented):
Looks good overall, added a question.

.github/workflows/image.yml
aldy505: No, tried it before, it sucked.

aldy505:
Can we merge this before EOD?

PR #6831, opened 2025-01-27
fix(eap-api): `equal`, `not equal`, `in`, `not in` ignoring case

xurui-c:
https://github.com/getsentry/eap-planning/issues/127

The ticket asks for ignoring case for the values, not the keys themselves
1. equality on keys: "span.op" = "SPAN.OP" <-- we don't support this
2. equality on values: if span.op is pageload, then "pageload" = "PAGELOAD" <-- we support this

### DIFF
--- a/snuba/web/rpc/common/common.py
+++ b/snuba/web/rpc/common/common.py
@@ -291,7 +291,11 @@ def trace_item_filters_to_expression(item_filter: TraceItemFilter) -> Expression
             case "val_bool":
                 v_expression: Expression = literal(v.val_bool)
             case "val_str":
-                v_expression = literal(v.val_str)
+                v_expression = (
+                    literal(v.val_str.lower())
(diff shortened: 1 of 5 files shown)


kylemumma (commented):
Wouldnt this fail with something like `"HELLO" = "HELLO"` because it will be converted to `"HELLO" = "hello"`?

xurui-c:
I thought from the ticket description including `lower()` means that lowercase "hello" is properly stored on ClickHouse, and then the query will contain "HELLO" and `comparison_filter=ComparisonFilter(..., ignore_case=True)`, so the "HELLO" will be lower case "hello".

kylemumma (commented):
the logic LGTM but I think it would be good to add a check that throws an error if they try to use ignore_case with non-string types

kylemumma (approved):

PR #6835, opened 2025-01-29
chore(eap): Trim time range if going over 30 days

phacops:
We need to trim the time range for the time being since we do receive queries longer than 30 days which we do not want to allow right now.

We trim to the more recent data since our product always queries for the more recent data.

We trim instead of returning an error to not break part of the UI where we can't fix querying for 90 days of data.

This is temporary and will be removed before GA.

### DIFF
--- a/snuba/web/rpc/__init__.py
+++ b/snuba/web/rpc/__init__.py
@@ -131,7 +131,7 @@ class RPCEndpoint(Generic[Tin, Tout], metaclass=RegisteredClass):
             if (end - start).days > MAXIMUM_TIME_RANGE_IN_DAYS:
                 timestamp = Timestamp()
                 timestamp.FromDatetime(end - timedelta(days=MAXIMUM_TIME_RANGE_IN_DAYS))
-                in_msg.meta.start_timestamp.CopyFrom(timestamp)  # type: ignore
+                in_msg.meta.start_timestamp.CopyFrom(timestamp)
         try:
             out = self._execute(in_msg)
         except Exception as e:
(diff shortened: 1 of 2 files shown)


snuba/web/rpc/__init__.py
```diff
@@ -117,6 +121,17 @@ def execute(self, in_msg: Tin) -> Tout:
...
+        if (
+            hasattr(in_msg, "meta")
+            and hasattr(in_msg.meta, "start_timestamp")
+            and hasattr(in_msg.meta, "end_timestamp")
+        ):
+            start = in_msg.meta.start_timestamp.ToDatetime()
+            end = in_msg.meta.end_timestamp.ToDatetime()
+            if (end - start).days > MAXIMUM_TIME_RANGE_IN_DAYS:
+                timestamp = Timestamp()
+                timestamp.FromDatetime(end - timedelta(days=MAXIMUM_TIME_RANGE_IN_DAYS))
```
onewland, line 133: I have no opposition to this, but I'm just curious if there's a reason we're starting at the `end_timestamp` when we trim vs. the `start_timestamp`

phacops:
It's more interesting to get the last 30 days of data instead of the first 30 days and that's why I keep `end_timestamp` constant and just adjust from there. The time ranges the product has are "Last 90 days", "Last 30 days", etc.

And I chose to trim instead of sending an error because dashboards could be sending a request and it would break them instead of making them functional but not returning the full dataset.

This will end up being cleaned up or GA when we have sampling in storage, it's temporary.

PR #6841, opened 2025-01-30
chore: refactor eap request transformation

kylemumma:
I decided that it would make sense to put request transformation step at a different spot, not logical difference just for clarity.

Previously it was at the beginning of `req -> query` translation. I think it made sense to go before the build query step.

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
@@ -132,9 +132,6 @@ def _convert_order_by(
 
 
 def _build_query(request: TraceItemTableRequest) -> Query:
-    # this is similar to the query processor step of the snql pipeline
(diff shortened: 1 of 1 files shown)


phacops (approved):

onkar (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
```diff
@@ -297,12 +294,22 @@ def _get_page_token(
     return PageToken(offset=request.page_token.offset + num_rows)
 
 
+def _transform_request(request: TraceItemTableRequest) -> TraceItemTableRequest:
```
volokluev, line 297: does it make sense to put this in the endpoint code rather than the resolver?

kylemumma: yes that makes sense, and I think would be a better place for it. thanks for the suggestion!

volokluev (approved):

PR #6843, opened 2025-01-30
feat(eap): Tag metrics with time buckets and referrer

phacops:
Add additional tags to the metrics to understand the performance of the endpoints better.

### DIFF
--- a/snuba/utils/metrics/timer.py
+++ b/snuba/utils/metrics/timer.py
@@ -31,6 +31,9 @@ class Timer:
     def __diff_ms(self, start: float, end: float) -> int:
         return int((end - start) * 1000)
 
+    def set_tags(self, tags: Mapping[str, str]) -> None:
+        self.__tags.update(tags)
+
     def get_duration_group(self) -> str:
         if self.__data is None:
             return "unknown"
--- a/snuba/web/rpc/__init__.py
+++ b/snuba/web/rpc/__init__.py
@@ -1,4 +1,5 @@
 import os
+from bisect import bisect_left
 from datetime import timedelta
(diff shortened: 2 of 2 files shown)


snuba/utils/metrics/timer.py
```diff
@@ -31,6 +31,9 @@ def mark(self, name: str) -> None:
     def __diff_ms(self, start: float, end: float) -> int:
         return int((end - start) * 1000)
 
+    def set_tags(self, tags: Mapping[str, str]) -> None:
```
volokluev, line 34: this is more of an "update_tags" function, it doesn't override existing ones

volokluev (commented):
You should test that this works as expected

snuba/web/rpc/__init__.py
```diff
@@ -1,4 +1,5 @@
 import os
+from bisect import bisect_left
```
onewland, line 2: didn't know this existed

snuba/web/rpc/__init__.py
```diff
@@ -66,6 +67,9 @@ def resolve(self, in_msg: Tin) -> Tout:
         raise NotImplementedError
 
 
+_TIME_PERIOD_BUCKETS = [1, 7, 14, 30, 90]
```
onewland, line 70:
```suggestion
_TIME_PERIOD_DAYS_BUCKETS = [1, 7, 14, 30, 90]
```

I assume the unit here is days?

onewland (approved):
seems fine to me

volokluev:
The RPC base class has a file for tests, if we're going to base our SLOs on this functionality, it should have a test to make sure it does what's expected

PR #6844, opened 2025-01-30
feat: formula queries in EndpointTraceItemTable

kylemumma:
This PR implements support for formulas in the TraceItemTable endpoint. It is relevant to this ticket https://github.com/orgs/getsentry/projects/284/views/1?pane=issue&itemId=85243940&issue=getsentry%7Ceap-planning%7C27
It enables queries such as `sum(my_attribute) / count(my_attribute)` which werent possible before

## Major changes
* the column to expression conversion logic used to take place [here](https://github.com/getsentry/snuba/pull/6844/files#diff-e1e06d7f875a7c2870cc11bc4301dd6ab9fba73263c76260452c0b3176f66110L141-L187) I extracted this logic into 2 separate new functions: `_get_reliability_context_columns` and `_column_to_expression`. (i.e. the logic stayed the same but its not inside these functions)
* I then extended `_column_to_expression` to support formulas https://github.com/getsentry/snuba/pull/6844/files#diff-e1e06d7f875a7c2870cc11bc4301dd6ab9fba73263c76260452c0b3176f66110R180-R191
* I also had to extend the existing result conversion and order by logic to support formulas https://github.com/getsentry/snuba/pull/6844/files#diff-e1e06d7f875a7c2870cc11bc4301dd6ab9fba73263c76260452c0b3176f66110R287-R288

## Testing
I wrote 3 new tests for this feature: 
* one that tests a simple formula on aggregates `sum(my_attribute) / count(my_attribute)`
* one that is the same as above uses extrapolation, to ensure formulas work with extrapolation.
* one that tests formulas on attributes without aggregation `my_attribute + my_other_attribute`

## design decisions
* reliabilities dont work with formulas, if you do a formula on extrapolated aggregates you will not get any reliability information back. This decision was made because of the increased complexity it would add to support. If needed we can implement support for this in the future.
* I realized while implementing this that formulas using constants are not supported such as `my_attribute * 10` if we need support for this it must be implemented as a follow up. and will require further modification of our protobuf grammar.
* it only supports spans not uptime or logs

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,7 +29,7 @@ python-rapidjson==1.8
 redis==4.5.4
 sentry-arroyo==2.19.12
 sentry-kafka-schemas==0.1.129
-sentry-protos==0.1.57
+sentry-protos==0.1.58
 sentry-redis-tools==0.3.0
 sentry-relay==0.9.5
 sentry-sdk==2.18.0
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
@@ -298,7 +298,7 @@ class ResolverTimeSeriesEAPSpans(ResolverTimeSeries):
         return TraceItemType.TRACE_ITEM_TYPE_SPAN
 
     def resolve(self, in_msg: TimeSeriesRequest) -> TimeSeriesResponse:
-        if in_msg.HasField("expressions"):
+        if len(in_msg.expressions) > 0:
             raise BadSnubaRPCRequestException("expressions field not yet implemented")
 
         snuba_request = _build_snuba_request(in_msg)
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
@@ -37,6 +37,7 @@ from snuba.web.rpc.common.debug_info import (
     extract_response_meta,
     setup_trace_query_settings,
 )
+from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
 from snuba.web.rpc.v1.resolvers import ResolverTimeSeries
 from snuba.web.rpc.v1.resolvers.R_eap_spans.common.aggregation import (
     ExtrapolationContext,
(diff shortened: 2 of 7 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
kylemumma, line 1: if people try to use formulas in Timeseries Endpoint it returns a not-implemented-error. I will implement support for formulas in timeseries endpoint as my next PR.

requirements.txt
```diff
@@ -29,7 +29,7 @@ python-rapidjson==1.8
 redis==4.5.4
 sentry-arroyo==2.19.12
 sentry-kafka-schemas==0.1.129
-sentry-protos==0.1.55
+sentry-protos==0.1.58
```
kylemumma, line 32: new protobuf definition with support for formulas, see https://github.com/getsentry/sentry-protos/pull/105

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
kylemumma, line 1: if people try to use formulas in Timeseries Endpoint it returns a not-implemented-error. I will implement support for formulas in timeseries endpoint as my next PR.

volokluev:
So this implements formulas only for spans. How much work would it be to support it for all trace item types

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
```diff
@@ -125,9 +126,87 @@ def _convert_order_by(
...
+        )
+        context_columns.append(
+            SelectedExpression(name=count_column.alias, expression=count_column)
+        )
+        return context_columns
+    return []
+
+
+def _formula_to_expression(formula: Column.BinaryFormula) -> Expression:
+    op_to_expr = {
```
phacops, line 177: nit: this can be a constant somewhere.

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
```diff
@@ -125,9 +126,87 @@ def _convert_order_by(
                     expression=aggregation_to_expression(x.column.aggregation),
                 )
             )
+        elif x.column.HasField("formula"):
+            res.append(
+                OrderBy(
+                    direction=direction,
+                    expression=_formula_to_expression(x.column.formula),
```
phacops, line 133: Wouldn't we be able to set the expression to be the label of the expression instead of the whole formula again?

kylemumma:
> So this implements formulas only for spans. How much work would it be to support it for all trace item types

It would be non-trivial to extend this PR to support all data types, we will leave it as span only

phacops (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
kylemumma: i tried, it didnt work

PR #6850, opened 2025-02-03
feat(eap): Add a table to store items

phacops:
Most of our events are a bunch of IDs and some attributes and we're looking for a way to easily support such a schema to add different event types and query them more easily.

This will add a table implementing this generic schema.

Resolves https://github.com/getsentry/eap-planning/issues/174


kylemumma (approved):

snuba/snuba_migrations/events_analytics_platform/0024_items.py
```diff
@@ -0,0 +1,141 @@
...
+local_table_name = "eap_items_1_local"
+dist_table_name = "eap_items_1_dist"
+num_attr_buckets = 40
+
+columns: List[Column[Modifiers]] = [
+    Column("organization_id", UInt(64)),
+    Column("project_id", UInt(64)),
+    Column("item_type", UInt(8)),
+    Column("timestamp", DateTime(Modifiers(codecs=["DoubleDelta", "ZSTD(1)"]))),
+    Column("trace_id", UInt(128)),
```
colin-sentry, line 19: why not uuid?

onewland:
Question: is the intention that this will be inserted to on the same storage nodes as eap_spans? 

Not trying to block this change but we want to keep an eye on adding too much insert/merge load onto any single node in ClickHouse. By default, each table is going to do background merging, fetching, mutations and the more tables you have the more you increase the likelihood of tables starving each other for resources

phacops:
> is the intention that this will be inserted to on the same storage nodes as eap_spans?

Yes, it is, and yes, it adds load on the storage node. We're monitoring this, it's manageable for now.

onkar (approved):

snuba/snuba_migrations/events_analytics_platform/0024_items.py
```diff
@@ -0,0 +1,145 @@
...
+    ),
+    Column(
+        "attributes_bool",
+        Map(
+            String(),
+            Bool(),
+        ),
+    ),
+    Column(
+        "attributes_int64",
```
colin-sentry, line 44: just a minor nit, but this naming is confusing - you'll have both `attributes_int64` and `attributes_string_16` and `attributes_float64_16` in the same table, i think just doing `attributes_int` would be clearer

colin-sentry (approved):

PR #6853, opened 2025-02-04
fix(eap-api): GetTracesEndpoint returns all attributes if none are qu…

xurui-c:
Currently, if the user doesn't specify attributes to the `GetTracesEndpoint`, we just don't return anything; however, the desired behavior is to return all the attributes in `_ATTRIBUTES`


volokluev (approved):

phacops:
I don't think we need to return all attributes for this endpoint. They are all discoverable since they are described in the `enum` in the protobuf.

PR #6855, opened 2025-02-04
feat(flags): Update snuba grammar to support flag queries

cmanallen:


evanh (commented):
Are you going to add something to the SnQL SDK as well?

Also, can you please add a test to `test_snql_api.py` that does an end to end test?

cmanallen:
@evanh If I need to.  I'm unaware of the full scope of this thing.  Let me know what else I need to do.

cmanallen:
@evanh It does use the legacy endpoints.  I am interested in updating the SDK but I don't see anything specific to tags in there (other than the MQL grammar).

PR #6856, opened 2025-02-04
fix(eap-api): return normalized fields for GetTraceEndpoint

xurui-c:
Currently, if the user doesn't specify attributes to the GetTraceEndpoint, we won't return normalized fields; however, we should

https://github.com/getsentry/eap-planning/issues/187

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
@@ -88,7 +88,6 @@ def _build_query(request: GetTraceRequest) -> Query:
             ),
         ]
         columns_to_exclude = ["retention_days", "sign", "attr_str", "attr_num"]
-        # print("bruhhhh", get_entity(EntityKey("eap_spans")).get_data_model().columns)
         selected_columns.extend(
             [
                 SelectedExpression(
@@ -98,8 +97,6 @@ def _build_query(request: GetTraceRequest) -> Query:
                 if col.name not in columns_to_exclude
             ]
         )
-
-        print("bruhh", selected_columns)
 
     entity = Entity(
         key=EntityKey("eap_spans"),
@@ -182,14 +179,24 @@ def _value_to_attribute(key: str, value: Any) -> tuple[AttributeKey, AttributeVa
                 val_double=value,
             ),
         )
-    elif isinstance(value, str) or isinstance(value, datetime):
+    elif isinstance(value, str):
         return (
             AttributeKey(
                 name=key,
                 type=AttributeKey.Type.TYPE_STRING,
             ),
             AttributeValue(
-                val_str=str(value),
+                val_str=value,
+            ),
+        )
+    elif isinstance(value, datetime):
+        return (
+            AttributeKey(
+                name=key,
+                type=AttributeKey.Type.TYPE_DOUBLE,
+            ),
+            AttributeValue(
+                val_double=value.timestamp(),
             ),
         )
     else:
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
@@ -86,6 +86,30 @@ def _build_query(request: GetTraceRequest) -> Query:
                 ),
             ),
         ]
+        selected_columns.extend(
+            [
+                SelectedExpression(name=col_name, expression=column(col_name))
+                for col_name in [
+                    "organization_id",
+                    "project_id",
+                    "service",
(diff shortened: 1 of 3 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
```diff
@@ -86,6 +87,32 @@ def _build_query(request: GetTraceRequest) -> Query:
...
+                    "segment_name",
+                    "is_segment",
+                    "_sort_timestamp",
+                    "start_timestamp",
+                    "end_timestamp",
+                    "duration_micro",
+                    "exclusive_time_micro",
+                    "name",
+                    "sampling_factor",
+                    "sampling_weight",
```
volokluev, line 112: where is this list coming from?

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
```diff
@@ -168,14 +195,14 @@ def _value_to_attribute(key: str, value: Any) -> tuple[AttributeKey, AttributeVa
...
-    elif isinstance(value, str):
+    elif isinstance(value, str) or isinstance(value, datetime):
         return (
             AttributeKey(
                 name=key,
                 type=AttributeKey.Type.TYPE_STRING,
             ),
             AttributeValue(
-                val_str=value,
+                val_str=str(value),
```
volokluev, line 205: timestamps should be converted to the protobuf TimeStamp object, not utc strings

tests/web/rpc/v1/test_endpoint_get_trace.py
```diff
@@ -35,6 +35,25 @@
...
+    "is_segment",
+    "_sort_timestamp",
+    "start_timestamp",
+    "end_timestamp",
+    "duration_micro",
+    "exclusive_time_micro",
+    "name",
+    "sampling_factor",
+    "sampling_weight",
+]
```
volokluev, line 56: The fact that you re copying this list in the test and in the code should give you pause

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
volokluev: 1. You are not querying the storage, you are querying the entity
2. If it's coming from the entity, you can just get the entity `get_entity(EntityKey('eap_spans')` here rather than duplicating its column names

tests/web/rpc/v1/test_endpoint_get_trace.py
```diff
@@ -35,6 +37,18 @@
 ) - timedelta(minutes=180)
 _SPAN_COUNT = 120
 _REQUEST_ID = uuid.uuid4().hex
+_NORMALIZED_FIELDS_TO_EXCLUDE = [
+    "retention_days",
+    "sign",
+    "attr_str",
+    "attr_num",
+    "timestamp",
```
xurui-c, line 45: I added "timestamp" because https://github.com/getsentry/snuba/blob/master/snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py#L192

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
phacops: I'd suggest to return timestamps as unix timestamp for now since we don't have a date time type in the RPC yet.

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_get_trace.py
```diff
@@ -86,6 +87,19 @@ def _build_query(request: GetTraceRequest) -> Query:
                 ),
             ),
         ]
+        columns_to_exclude = ["retention_days", "sign", "attr_str", "attr_num"]
```
phacops, line 90: You want to add `span_id` and `timestamp` in there as well as we're already fetching them.

tests/web/rpc/v1/test_endpoint_get_trace.py
```diff
@@ -35,6 +37,19 @@
...
+    "attr_str",
+    "attr_num",
+    "span_id",
+    "timestamp",
+]
+_NORMALIZED_FIELDS_TO_INCLUDE = [
+    col.name
+    for col in get_entity(EntityKey("eap_spans")).get_data_model().columns
+    if col.name not in _NORMALIZED_FIELDS_TO_EXCLUDE
+]
```
volokluev, line 52: you are repeating the same logic in your functionality and in your test, it should only be specified once

volokluev (approved):

PR #6857, opened 2025-02-05
chore: bump sentry-kafka-schemas

JoshFerge:
primarily for new uptime failure types added in https://github.com/getsentry/sentry-kafka-schemas/pull/377

- [x] some small tweaks to our dockerfile and rust dependencies 
- [x] some snapshot diffs changed

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -13,7 +13,7 @@ steps:
         "-f",
         "./Dockerfile",
       ]
-    timeout: 12000s
+    timeout: 2400s
 
   # Pull docker image again, so we can get build info.
   # https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/212#issuecomment-1478828752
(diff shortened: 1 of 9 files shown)


evanpurkhiser (approved):

cloudbuild.yaml
```diff
@@ -13,7 +13,7 @@ steps:
         "-f",
         "./Dockerfile",
       ]
-    timeout: 1200s
+    timeout: 2400s
```
onewland, line 16: can we back out changes to this file and see if the build still succeeds? I think this is more "power" than we need if we're not totally building the image from scratch

JoshFerge: i'm just trying to bump a package version and have spent way too much time on this already 😭 

but sure.

volokluev (approved):

IanWoodard (approved):

PR #6870, opened 2025-02-10
feat(eap-api): support conditional aggregations in SELECT

xurui-c:
https://github.com/getsentry/eap-planning/issues/166

We now support conditional aggregations for EndpointTraceItemTable. Next is time series


snuba/web/db_query.py
```diff
@@ -176,6 +176,8 @@ def execute_query(
     # Apply clickhouse query setting overrides
     clickhouse_query_settings.update(query_settings.get_clickhouse_settings())
 
+    print("formatted_queryyyy", formatted_query)
```
xurui-c, line 179: The formatted query is SELECT (if(mapContains(attr_str_9, 'kylestag'), arrayElement(attr_str_9, 'kylestag'), NULL) AS kylestag_TYPE_STRING), (round(sumIfOrNull((if(mapContains(attr_num_17, 'my.float.field'), arrayElement(attr_num_17, 'my.float.field'), NULL) AS `my.float.field_TYPE_DOUBLE`), mapContains(attr_num_17, 'my.float.field') **AND mapContains(attr_str_9, 'kylestag'))**, 9) AS `sum(my.float.field)`) FROM eap_spans_2_local WHERE in(project_id, [1, 2, 3]) AND equals(organization_id, 1) AND less(_sort_timestamp, toDateTime(1739239200)) AND greaterOrEquals(_sort_timestamp, toDateTime(1739235600)) AND mapContains(attr_num_17, 'my.float.field') GROUP BY kylestag_TYPE_STRING ORDER BY kylestag_TYPE_STRING ASC LIMIT 10000 OFFSET 0

the bolded part should've done the If in sumIf, I'm having trouble understanding why the result still included is_null

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -1755,6 +1758,91 @@ def test_aggregation_filter_basic(self, setup_teardown: Any) -> None:
...
+        print("responseee", response)
+        breakpoint()
+        assert response.column_values == [
+            TraceItemColumnValues(
+                attribute_name="kylestag",
+                results=[
+                    AttributeValue(val_str="val1"),
+                    AttributeValue(val_str="val2"),
+                    AttributeValue(val_str="val3"),
+                    AttributeValue(is_null=True),
```
xurui-c, line 1833: This is populated by all the rows that don't contain kylestag

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -1755,6 +1758,91 @@ def test_aggregation_filter_basic(self, setup_teardown: Any) -> None:
...
+                ],
+            ),
+            TraceItemColumnValues(
+                attribute_name="sum(my.float.field)",
+                results=[
+                    AttributeValue(is_null=True),
+                    AttributeValue(val_double=404.8),
+                    AttributeValue(is_null=True),
```
xurui-c, line 1841: these rows are populated by `get_field_existence_expression`, which is false for non-matching rows. Then the orNull suffix in `aggregation_to_expression` makes it return null

snuba/web/db_query.py
davidtsuk, line 179: remove

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -43,7 +46,13 @@ def _validate_select_and_groupby(in_msg: TraceItemTableRequest) -> None:
         [c.key.name for c in in_msg.columns if c.HasField("key")]
     )
     grouped_by_columns = set([c.name for c in in_msg.group_by])
-    aggregation_present = any([c for c in in_msg.columns if c.HasField("aggregation")])
+    aggregation_present = any(
+        [
+            c
+            for c in in_msg.columns
+            if (c.HasField("aggregation") or c.HasField("conditional_aggregation"))
```
davidtsuk, line 53: nit: i don't think the brackets are needed here

davidtsuk (commented):
I think you also need to update the confidence interval calculations to include the condition. Also, could you please add a test for extrapolated conditional aggregates

snuba/web/db_query.py
kylemumma, line 179: did you forget to delete this?

kylemumma (approved):
I think you should clean up those print statements laying around but other than that LGTM assuming the tests pass. Also thanks for including a PR description.

volokluev (changes requested):
From the changes that I see in sentry-protos, you have recreated the aggregation struct with the condition field that you need (we spoke about this):

https://github.com/getsentry/sentry-protos/pull/109/files#diff-4b3d0e1ccced0b672ff3b41746de66f106ab7239b75c2bcc10048ec4c0017a9bR9-R13


But in this code here, what you are doing is creating branches for both cases, this makes the code more fragmented. 

What you should be doing instead, is converting the old (soon to be deprecated) object into the new `AttributeConditionalAggregation` object in the very beginning and then doing all of your operations from there

xurui-c:
There's still some of the fragmented code in `aggregation.py` because it's still being used by the TimeSeries endpoint. I'll get rid of those in a subsequent PR for time series

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -1756,6 +1760,191 @@ def test_aggregation_filter_basic(self, setup_teardown: Any) -> None:
...
+                    AttributeValue(is_null=True),
+                ],
+            ),
+        ]
+
+    def test_rachel(self) -> None:
```
xurui-c, line 1845: ill change the naming of this

tests/web/rpc/v1/visitors/test_sparse_aggregate_attribute_transformer.py
```diff
@@ -76,6 +78,7 @@ def test_basic() -> None:
     )
     transformed = SparseAggregateAttributeTransformer(req).transform()
     # filter was set properly
+    print(transformed.filter)
```
xurui-c, line 81: will delete

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -260,36 +278,49 @@ def get_attribute_confidence_interval_alias(
...
-            get_field_existence_expression(field),
+            and_cond(get_field_existence_expression(field), condition_in_aggregation),
         ),
         alias=alias,
     )
 
 
-def _get_count_column_alias(aggregation: AttributeAggregation) -> str:
+def _get_count_column_alias(
+    aggregation: AttributeAggregation | AttributeConditionalAggregation,
```
volokluev, line 305: if you've transformed everything up front, when would you still have an `AttributeAggregation` in the arguments here?

xurui-c: time series endpoint

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -80,6 +87,48 @@ def _transform_request(request: TraceItemTableRequest) -> TraceItemTableRequest:
     return SparseAggregateAttributeTransformer(request).transform()
 
 
+def convert_to_conditional_aggregation(in_msg: TraceItemTableRequest) -> None:
```
volokluev, line 90: How different would this function be for the TimeSeries endpoint?

volokluev: 1. leave a comment explaining why this is being done. It's not immediately clear unless you are the user
2. Test this function independently. It has clearly defined inputs and outputs

volokluev (commented):
Overall looks good! My main concern is just testing that request processor you wrote independently

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -2595,12 +2785,73 @@ def test_nonexistent_attribute(setup_teardown: Any) -> None:
         ]
 
 
+def _build_sum_attribute_aggregation_column_with_name(name: str) -> Column:
```
volokluev, line 2788: by which i mean this should be in its own file

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -80,6 +87,52 @@ def _transform_request(request: TraceItemTableRequest) -> TraceItemTableRequest:
...
+    """
+    This function adds the equivalent conditional aggregation for every aggregation in each Column or AggregationFilter. We don't add the filter field so outside code logic will set the default condition to true. The purpose of this function is to "transform" every AttributeAggregation to AttributeConditionalAggregation in order to avoid code fragmentation
+    """
+
+    def _add_conditional_aggregation(
```
volokluev, line 95: see if you can replace in all the cases so you don't have the split behavior with Column and AggregationComparsionFilter

snuba/web/rpc/v1/endpoint_trace_item_table.py
```diff
@@ -80,6 +87,52 @@ def _transform_request(request: TraceItemTableRequest) -> TraceItemTableRequest:
     return SparseAggregateAttributeTransformer(request).transform()
 
 
+def convert_to_conditional_aggregation(in_msg: TraceItemTableRequest) -> None:
+    """
+    This function adds the equivalent conditional aggregation for every aggregation in each Column or AggregationFilter. We don't add the filter field so outside code logic will set the default condition to true. The purpose of this function is to "transform" every AttributeAggregation to AttributeConditionalAggregation in order to avoid code fragmentation
```
volokluev, line 92: Explain historical reasons here, that's the important part. Think of yourself as a person new to this code

snuba/web/rpc/v1/endpoint_trace_item_table.py
xurui-c: replace is done with `input.ClearField("aggregation")`, but I still need the separate `isinstance(input, Column)` and `isinstance(input, AggregationFilter)` because of mypy. Also I feel like it makes `_convert` more readable(?)

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
xurui-c: test_conditional_aggregation.py

volokluev (approved):
much better

PR #6875, opened 2025-02-11
migrate(smart-autocomplete): Add a hashmap optimization + BF index to smart autocomplete table

volokluev:
Using a similar trick as the tags hashmap optimization, store a hash of all the keys as an array so we can look up the existence of an attribute much faster (no need for string compares), also add a bloom filter index onto that column as well.

With this trick we can add the following statement to a prewhere to speed up the query:

```
has(_str_attr_keys_hash_map, cityHash64('sentry.op'))
```

### DIFF
--- a/snuba/datasets/storages/tags_hash_map.py
+++ b/snuba/datasets/storages/tags_hash_map.py
@@ -33,6 +33,10 @@ SENTRY_TAGS_HASH_MAP_COLUMN = (
 )
 
 
+def get_array_vals_hash(col_name: str) -> str:
+    return f"arrayMap(k -> cityHash64(k), {col_name})"
+
+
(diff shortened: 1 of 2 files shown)


kylemumma (approved):
What does a row in your table represent? Im having a hard time understanding what a single `attrs_string` array represents?

What is the bloom filter index for?

volokluev:
@kylemumma don't approve stuff you don't understand :D

> What does a single `attrs_string` represent?

It represents all the string attributes in an event in one array. 

> What is the bloom filter index for?

The bloom filter index is a [data skipping index](https://clickhouse.com/docs/en/optimize/skipping-indexes). If a block does not contain a specified key, it will not be scanned

PR #6887, opened 2025-02-14
feat(uptime): Ability to ingest and query in_incident

evanpurkhiser:
Depends on GH-6886


phacops (approved):

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table_uptime_checks.py
```diff
@@ -52,6 +52,7 @@ def gen_message(
         },
         "http_status_code": 200,
         "trace_id": _TRACE_ID,
+        "incident_status": False,
```
phacops, line 55: This should be an integer.

wedamija (approved):

asottile-sentry:
reverting: this appears to be breaking a bunch of uptime tests in `sentry`

PR #6894, opened 2025-02-17
docs: Fix docs for running sentry tests

jan-auer:
The new `devservices` command does not have an `--exclude` option.


untitaker (approved):

jan-auer:
There may be a proper replacement, but I haven't found it in `devservices`. We can update the docs again.

PR #6905, opened 2025-02-19
chore(eap): Add an example to calculate the HTTP response rate

phacops:
This is an example of an implementation of the current `http_response_rate(5)` function we use in Insights.


volokluev (approved):
nice

PR #6906, opened 2025-02-20
feat(eap-api): support conditional aggregation for TimeSeriesEndpoint

xurui-c:
https://github.com/getsentry/eap-planning/issues/193

### DIFF
--- a/snuba/web/rpc/proto_component.py
+++ b/snuba/web/rpc/proto_component.py
@@ -28,7 +28,19 @@ class ColumnWrapper(ProtoWrapper):
         self.column = column
 
     def accept(self, visitor: ProtoVisitor) -> None:
-        visitor.visit_column(self)
+        visitor.visit(self)
+
+        if self.column.HasField("formula"):
+            ColumnWrapper(self.column.formula.left).accept(visitor)
+            ColumnWrapper(self.column.formula.right).accept(visitor)
+
+
+class AggregationComparisonFilterWrapper(ProtoWrapper):
+    def __init__(self, aggregation_comparison_filter: AggregationComparisonFilter):
+        self.aggregation_comparison_filter = aggregation_comparison_filter
+
+    def accept(self, visitor):
+        visitor.visit(self)
 
 
 class AggregationFilterWrapper(ProtoWrapper):
@@ -36,7 +48,18 @@ class AggregationFilterWrapper(ProtoWrapper):
         self.aggregation_filter = aggregation_filter
 
     def accept(self, visitor: ProtoVisitor) -> None:
-        visitor.visit_aggregation_filter(self)
+        if self.aggregation_filter.HasField("comparison_filter"):
+            AggregationComparisonFilterWrapper(
+                self.aggregation_filter.comparison_filter
+            ).accept(visitor)
+
+        if self.aggregation_filter.HasField("and_filter"):
+            for agg_filter in self.aggregation_filter.and_filter.filters:
+                AggregationFilterWrapper(agg_filter).accept(visitor)
+
+        if self.aggregation_filter.HasField("or_filter"):
+            for agg_filter in self.aggregation_filter.or_filter.filters:
+                AggregationFilterWrapper(agg_filter).accept(visitor)
 
 
 class TimeSeriesExpressionWrapper(ProtoWrapper):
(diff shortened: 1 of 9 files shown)


tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -237,6 +240,109 @@ def test_fails_for_logs(self) -> None:
             error.ParseFromString(response.data)
         assert response.status_code == 400, (error.message, error.details)
 
+    def test_rachel(self) -> None:
```
kylemumma, line 243: can you give this a better name

kylemumma (commented):
lgtm other than the test name.

did you ensure that this wont break the other resolvers that arent uptime and eap_spans?

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -239,6 +249,75 @@ def from_alias(alias: str) -> "CustomColumnInformation":
...
+
+    For messages that don't have conditional aggregation, this function replaces the aggregation with a conditional aggregation,
+    where the filter is null, and every field is the same. This allows code elsewhere to set the default condition to always
+    be true.
+
+    The reason we do this "transformation" is to avoid code fragmentation down the line, where we constantly have to check
+    if the request contains `AttributeAggregation` or `AttributeConditionalAggregation`
+    """
+
+    def _add_conditional_aggregation(
```
mcannizz, line 267: If I'm following this correctly, this function does two things:
1. It traverses the tree represented by `TraceItemTableRequest` and `TimeSeriesRequest`.
2. When it finds an aggregation, it converts that aggregation to a conditional one.

I suspect the code would be easier to read if we separated these two mostly unrelated concerns. The visitor pattern is one way to achieve this (I'm happy to explain this if you're unfamiliar with the visitor pattern idea).

mcannizz (commented):
Addressing my comments is optional.

snuba/web/rpc/proto_component.py
```diff
@@ -0,0 +1,150 @@
...
+
+class TimeSeriesRequestWrapper(ProtoWrapper):
+    def accept(self, visitor: ProtoVisitor) -> None:
+        visitor.visit_TimeSeriesRequestWrapper(self)
+        time_series_request = self.underlying_proto
+        for expression in time_series_request.expressions:
+            TimeSeriesExpressionWrapper(expression).accept(visitor)
+
+
+class ProtoVisitor(ABC):
```
volokluev, line 96: There's an important thing to note here in the comments for the user:

1. How do they use this class? You're doing some metaprogramming here, provide an example
2. With how this is implemented, it is important to understand that if the visitor traverses the structure, it will be traversed again by the wrapper object. The user should be given guidance to not traverse the object and have the visitor only work on the object it is visiting

snuba/web/rpc/proto_component.py
```diff
@@ -0,0 +1,150 @@
...
+
+class AggregationToConditionalAggregationVisitor(ProtoVisitor):
+    """
+    We support aggregation, but now we want to support conditional aggregation for the insights team, which only aggregates
+    if the field satisfies the condition: https://clickhouse.com/docs/en/sql-reference/aggregate-functions/combinators#-if.
+    For messages that don't have conditional aggregation, this function replaces the aggregation with a conditional aggregation,
+    where the filter is null, and every field is the same. This allows code elsewhere to set the default condition to always
+    be true.
+    The reason we do this "transformation" is to avoid code fragmentation down the line, where we constantly have to check
+    if the request contains `AttributeAggregation` or `AttributeConditionalAggregation`
```
volokluev, line 132: This is not communicating the important parts of why this transformation exists. It can be summed up as:

Originally, the protobuf request did not support conditional aggregations. When they were added, the file structure did not allow adding a TraceItemFilter to the `AttributeAggregation` message without breaking backwards compatibility. A new `AttributeConditionalAggregation` message was added whose fields are a superset of of `AttributeAggregation`. `AttributeAggregation` was deprecated.

This code transforms every `AttributeAggregation` to an `AttributeConditionalAggregation` to be able to serve  queries before the deprecation, and have the rest of the query pipeline only concerned with one kind of aggregation object.

snuba/web/rpc/proto_visitor.py
```diff
@@ -94,6 +100,39 @@ def accept(self, visitor: ProtoVisitor) -> None:
...
+    2. the wrappers that wrap around the protobuf objects. Think of these as the "nodes" of an AST. Responsible for traversal
+
+    I put the responsibility of traversing the wrapper structure onto the wrapper itself. This is a valid approach according
+    to Design Patterns: Elements of Reusable Object-Oriented Software page 339: "We can
+    put responsibility for traversal in any of three places: in the object structure,
+    in the visitor, or in a separate iterator object (see Iterator (257)).
+    Often the object structure is responsible for iteration. A collection will simply
+    iterate over its elements, calling the Accept operation on each. A composite
+    will commonly traverse itself by having each Accept operation traverse the
+    element's children and call Accept on each of them recursively."
```
volokluev, line 115: lmao

volokluev (approved):

PR #6910, opened 2025-02-25
ref(ourlogs): Add integration test for ourlogs to snuba

k-fish:
### Summary
This integration should be added to the CI tests so it doesn't break on master in the future. 

see: https://github.com/getsentry/snuba/pull/6903


colin-sentry (approved):
lgtm, need approval from @phacops

phacops (approved):

PR #6914, opened 2025-02-25
feat(ourlogs): Add item type check to resolver

colin-sentry:
Now that trace_item_table is the union of both logs and spans, send item_type in the `WHERE` clause


snuba/web/rpc/v1/resolvers/R_ourlogs/resolver_trace_item_table.py
```diff
@@ -68,14 +70,14 @@ def _build_query(request: TraceItemTableRequest) -> Query:
     )
 
     selected_columns = []
-    for column in request.columns:
-        if column.HasField("key"):
-            key_col = attribute_key_to_expression(column.key)
+    for col in request.columns:
+        if col.HasField("key"):
+            key_col = attribute_key_to_expression(col.key)
```
volokluev, line 75: why the rename?

colin-sentry: python hoists the `column` out of the loop scope to shadow the import, makes pypy mad

phacops (approved):

PR #6923, opened 2025-03-03
fix(eap): Update accuracy formulas for sum, count, and avg

jan-auer:
### Legal Boilerplate

### DIFF
--- a/snuba/web/rpc/v1/resolvers/common/aggregation.py
+++ b/snuba/web/rpc/v1/resolvers/common/aggregation.py
@@ -478,7 +478,7 @@ def get_confidence_interval_column(
     condition_in_aggregation = _get_condition_in_aggregation(aggregation)
 
     function_map_confidence_interval = {
-        # confidence interval = Z \cdot \sqrt{-log{(\frac{\sum_{i=1}^n \frac{1}{w_i}}{n})} \cdot \sum_{i=1}^n w_i^2 - w_i}
+        # confidence interval = Z \cdot \sqrt{\sum_{i=1}^n w_i^2 - w_i}
         #        ┌─────────────────────────┐
         #        │      ₙ
         #        │      ⎲  1
@@ -491,23 +491,20 @@ def get_confidence_interval_column(
         Function.FUNCTION_COUNT: f.multiply(
             z_value,
             f.sqrt(
-                f.multiply(
-                    f.negate(f.log(get_average_sample_rate_column(aggregation))),
-                    f.sumIf(
-                        f.minus(
-                            f.multiply(sampling_weight_column, sampling_weight_column),
-                            sampling_weight_column,
-                        ),
-                        and_cond(
-                            get_field_existence_expression(field),
-                            condition_in_aggregation,
-                        ),
+                f.sumIf(
+                    f.minus(
+                        f.multiply(sampling_weight_column, sampling_weight_column),
+                        sampling_weight_column,
                     ),
-                )
+                    and_cond(
+                        get_field_existence_expression(field),
+                        condition_in_aggregation,
+                    ),
+                ),
             ),
             **alias_dict,
         ),
-        # confidence interval = N * Z * \sqrt{\frac{\sum_{i=1}^n w_ix_i^2 - \frac{(\sum_{i=1}^n w_ix_i)^2}{N}}{n * N}}
+        # confidence interval = Z \cdot \sqrt{\sum_{i=1}^n x_i^2 \cdot (w_i^2 - w_i)}
         #              ┌────────────────────────────┐
         #              │              ₙ
         #              │              ⎲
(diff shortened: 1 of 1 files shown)


snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -478,7 +478,7 @@ def get_confidence_interval_column(
     condition_in_aggregation = _get_condition_in_aggregation(aggregation)
 
     function_map_confidence_interval = {
-        # confidence interval = Z \cdot \sqrt{-log{(\frac{\sum_{i=1}^n \frac{1}{w_i}}{n})} \cdot \sum_{i=1}^n w_i^2 - w_i}
+        # confidence interval = Z \cdot \sqrt{\sum_{i=1}^n w_i^2 - w_i}
         #        ┌─────────────────────────┐
```
jan-auer, line 482: The ASCII-rendering for all of the formulas still needs to be updated.

volokluev:
I see in this that there are a bunch of tests where reliabilities that were previously considered low are now considered high. What would be considered a low probability. Can you add a test for that?

snuba/query/dsl.py
```diff
@@ -182,6 +182,10 @@ def divide(
     return FunctionCall(alias, "divide", (lhs, rhs))
 
 
+def abs(lhs: Expression, alias: Optional[str] = None) -> FunctionCall:
```
volokluev, line 185: I don't think you need this

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -463,182 +463,147 @@ def get_extrapolated_function(
...
+) -> Expression:
+    """
+    confidence interval = Z \cdot \sqrt{\sum_{i=1}^n x_i^2 \cdot (w_i^2 - w_i)}
+
+          ┌─────────────────┐
+          │ ₙ
+       ╲  │ ⎲   2    2
+    Z ⋅ ╲ │ ⎳  xᵢ ⋅(wᵢ - wᵢ)
+         ╲│ⁱ⁼¹
+    """
```
volokluev, line 514: In the PR description, you said this was using the [Horvitz-Thompson estimator](https://en.wikipedia.org/wiki/Horvitz%E2%80%93Thompson_estimator) method. I don't think I understand how this formula maps to the method

volokluev:
Also, I see that we are changing the formulas but it's not clear to the reader of this PR why we are doing it or on what basis we think these estimation formulas are better. Did you do some validation numpy? Is there a plan for a formal verification process in production?

snuba/query/dsl.py
jan-auer: Do you mean I should emit `FunctionCall("abs", ...)` directly into the place where I'm using this?

snuba/web/rpc/v1/resolvers/common/aggregation.py
jan-auer: Ah yes, I'll update that description and make it less cryptic 👍

jan-auer:
Thanks, @volokluev. I've expanded the PR description now to include more rationale. The changes were motivated through a review of the formulas, noticing that we were using invalid approaches. I've validated the formulas locally on synthetic test data, but haven't gone through a larger effort given the prior validation in the industry. 

In the product, we collect analytics, though they only make sense to look at once these fixes are in.

snuba/query/dsl.py
jan-auer: Indeed, thank you!

snuba/web/rpc/v1/resolvers/common/aggregation.py
jan-auer: To answer your question, page 9 / formula 9 of [the paper](https://www.stat.cmu.edu/~brian/905-2008/papers/Horvitz-Thompson-1952-jasa.pdf) is where this comes from. There's a great [blog post](https://blog.cloudflare.com/how-we-make-sense-of-too-much-data/#extracting-value-from-downsampled-data) from cloudflare that elaborates on the same approach.

I've reconsidered and don't think we should go beyond stating which estimator and variance method we use. However, we can put more reasoning in our [math repo](https://github.com/getsentry/extrapolation-math/tree/main) (I've not updated that yet) and then link it, WDYT?

volokluev (approved):

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -470,190 +470,185 @@ def get_extrapolated_function(
...
+         ╲│ⁱ⁼¹
+
+    where w_i is the sampling weight for the i-th event and n is the number of
+    events. This is based on the Horvitz-Thompson estimator for totals of
+    weighted samples:
+     1. Since we're counting, the value of each event is 1, which removes the
+        value from the formula.
+     2. The sampling weight is the inverse of the inclusion probability
+        (sampling rate).
+     3. Samples undergo independent Bernulli trials, which means the term for
```
shellmayr, line 495:
```suggestion
     3. Samples undergo independent Bernoulli trials, which means the term for
```

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -470,190 +470,185 @@ def get_extrapolated_function(
...
-                        ),
-                    ),
-                )
+                    f.plus(expr_sum, expr_sum_err),
+                    f.minus(expr_count, expr_count_err),
+                ),
+                f.divide(
+                    f.minus(column(f"{alias}__sum"), column(f"{alias}__sum_err")),
+                    f.plus(column(f"{alias}__count"), column(f"{alias}__count_err")),
+                ),
```
shellmayr, line 624: Can we add a test to make sure this doesn't throw when the count is zero?

PR #6928, opened 2025-03-03
Run rustup toolchain install explicitly

volokluev:
rustup will no longer automatically install the active toolchain if it is not installed. To ensure its installation, run `rustup toolchain install` with no arguments.

https://blog.rust-lang.org/2025/03/02/Rustup-1.28.0.html

In the cloudbuild logs we see this happening which is breaking our builds

```
info: default host triple is x86_64-unknown-linux-gnu
info: skipping toolchain installation
```


Dockerfile
```diff
@@ -87,6 +87,7 @@ COPY ./rust_snuba/Cargo.toml ./rust_snuba/Cargo.toml
 COPY ./rust_snuba/rust-toolchain.toml ./rust_snuba/rust-toolchain.toml
 COPY ./rust_snuba/Cargo.lock ./rust_snuba/Cargo.lock
 COPY ./scripts/rust-dummy-build.sh ./scripts/rust-dummy-build.sh
+RUN set -ex; rustup toolchain install
```
phacops, line 90: You need to add `stable` here.

volokluev: not quite

phacops (approved):

PR #6929, opened 2025-03-04
feat(capman): Instead of flatly rejecting heavy projects, limit their bytes scanned on the query level

volokluev:
Sometimes large customers complain that they get throttled because someone scans all their data and then they're locked out for 10 minutes. Not a great experience

To ease the pain, still allow their queries to go through but give them a short leash using clickhouses `max_bytes_to_read` setting

By default this will not be turned on until a human is around to monitor it

### DIFF
--- a/snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
+++ b/snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
@@ -111,6 +111,18 @@ class BytesScannedRejectingPolicy(AllocationPolicy):
                 int,
                 DEFAULT_THREADS_THROTTLE_DIVIDER,
             ),
+            AllocationPolicyConfig(
+                "limit_bytes_instead_of_rejecting",
+                "instead of rejecting a query, limit its bytes with max_bytes_to_read on clickhouse",
+                int,
+                0,
+            ),
+            AllocationPolicyConfig(
+                "max_bytes_to_read_scan_limit_divider",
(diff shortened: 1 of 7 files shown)


snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
```diff
@@ -230,32 +242,72 @@ def _get_quota_allowance(
...
+                    can_run=True,
+                    max_threads=max(
+                        1,
+                        self.max_threads
+                        // self.get_config_value("threads_throttle_divider"),
+                    ),
+                    max_bytes_to_read=max_bytes_to_read,
+                    explanation=explanation,
+                    is_throttled=True,
+                    throttle_threshold=throttle_threshold,
```
xurui-c, line 276: should `throttle_threshold = max_bytes_to_read`?

evanh (approved):
What's the purpose of the `max_bytes_to_read_scan_limit_divider`? Is that a toggle to grow/shrink how many bytes a particular query is limited too?

volokluev:
@evanh 

> Is that a toggle to grow/shrink how many bytes a particular query is limited too?

yup exactly. It's some ratio to the scan limit. 

@enochtangg 

RE: separate policy: This policy can still reject if the user query goes over the bytes scan limit and the accounting mechanism is the same (bytes scanned over a 10 minute window). That's why I put it in this policy

snuba/query/allocation_policies/bytes_scanned_rejecting_policy.py
volokluev: no because that's not the throttle threshold. The throttle threshold is how many bytes the customer scanned to get to the point that `max_bytes_to_read` was set

PR #6931, opened 2025-03-04
feat(ch) Test 23.8.16 clickhouse version

evanh:
The replays cluster needs at least this version to fix a segfault issue.

### DIFF
--- a/devservices/config.yml
+++ b/devservices/config.yml
@@ -25,7 +25,7 @@ x-sentry-service-config:
 
 services:
   clickhouse:
-    image: ghcr.io/getsentry/image-mirror-altinity-clickhouse-server:23.8.11.29.altinitystable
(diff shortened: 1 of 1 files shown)


onkar (approved):
LGTM, @dmajere confirmed that he performed the upgrade of ClickHouse to this version in production. 

One question: is the intention to this only for replay cluster for now and see how that works out? Trying to understand how do we do that conditionally in the code.

onkar:
>  Trying to understand how do we do that conditionally in the code.

Never mind, saw that https://github.com/getsentry/ops/pull/14327 does this for replay.

onkar:
This incident reoccured on ClickHouse version 23.8.16 on replays cluster. It means that the upgrade to 23.8.16 does not solve the issue. Closing this PR.

PR #6932, opened 2025-03-04
fix(eap): Remove column processor preventing the use of an index

phacops:
This column will transform the `trace_id` expression to:
```
(cast(replaceAll(toString(trace_id), '-', ''), 'String') AS trace_id)
```

As a result, when this is in the `PREWHERE`, it will not use the bloom filter created for the `trace_id`. Disabling this query processor will make it use the index.


volokluev (approved):
As long as the tests are fixed lgtm

asottile-sentry:
breaking sentry tests: https://sentry.sentry.io/issues/6359598701/?referrer=alerts-related-issues-issue-stream

PR #6933, opened 2025-03-04
bug: cant use group by with formulas

kylemumma:
fixes https://github.com/getsentry/eap-planning/issues/201


kylemumma:
added a test that ensures group by works as expected with formulas. the test was failing with the bug before the fix.

xurui-c (approved):

onkar (approved):

davidtsuk (approved):

PR #6934, opened 2025-03-05
feat(eap): use eap_items in timeseries endpoint

davidtsuk:
This PR updates the timeseries endpoint for the eap_spans resolver to use the new eap_items table when the following conditions are met:
- the `use_eap_items_table` runtime config is set to `1`
- the start time of the query is greater than `use_eap_items_table_start_timestamp_seconds` (measured in UTC seconds)

The tests have also been updated to been run once on the eap_spans table and once on the eap_items table.

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -53,7 +53,9 @@ NORMALIZED_COLUMNS_EAP_ITEMS: Final[
         AttributeKey.Type.TYPE_INT,
         AttributeKey.Type.TYPE_STRING,
     ],
-    "sentry.trace_id": [AttributeKey.Type.TYPE_STRING],
+    "sentry.trace_id": [
+        AttributeKey.Type.TYPE_STRING
+    ],  # this gets converted from a uuid to a string in a storage processor
     "sentry.item_id": [AttributeKey.Type.TYPE_STRING],
 }
 
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -138,6 +138,11 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
                 "Nullable(Int64)",
                 alias=alias,
             )
+        return SubscriptableReference(
+            column=column(PROTO_TYPE_TO_ATTRIBUTE_COLUMN[attr_key.type]),
+            key=literal(attr_key.name),
+            alias=alias,
+        )
 
     raise BadSnubaRPCRequestException(
         f"Attribute {attr_key.name} has an unknown type: {AttributeKey.Type.Name(attr_key.type)}"
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -1,10 +1,12 @@
-from typing import Final, Mapping, Sequence, Set
+from typing import Any, Final, Mapping, Sequence, Set
 
+from sentry_protos.snuba.v1.request_common_pb2 import RequestMeta
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import (
     AttributeKey,
     VirtualColumnContext,
 )
 
+from snuba import state
 from snuba.query import Query
 from snuba.query.dsl import Functions as f
 from snuba.query.dsl import column, literal, literals_array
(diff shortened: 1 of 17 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -38,6 +40,160 @@
     "sentry.end_timestamp",
 }
 
+COLLUMN_PREFIX: str = "sentry."
```
volokluev, line 43: mispelled

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -38,6 +40,160 @@
...
+
+    mapped_column_to_context = {c.to_column_name: c for c in virtual_column_contexts}
+
+    def transform_expressions(expression: Expression) -> Expression:
+        # virtual columns will show up as `attr_str[virtual_column_name]` or `attr_num[virtual_column_name]`
+        if not isinstance(expression, SubscriptableReference):
+            return expression
+
+        if expression.column.column_name != "attributes_string":
+            return expression
```
volokluev, line 172: nice

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
```diff
@@ -240,25 +267,37 @@ def _proto_expression_to_ast_expression(expr: ProtoExpression) -> Expression:
 
 def _build_query(request: TimeSeriesRequest) -> Query:
     # TODO: This is hardcoded still
```
volokluev, line 269: you can remove this comment I think

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -303,13 +304,14 @@ def _get_count_column_alias(
 
 def get_count_column(
     aggregation: AttributeAggregation | AttributeConditionalAggregation,
+    attribute_key_to_expression: Callable[[Any], Expression],
```
volokluev, line 307: nit: is the typing for this right? It's not really `[Any] -> Expression` is it?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -183,9 +183,11 @@ def write_eap_span(
 @pytest.fixture(autouse=False)
 def setup_teardown(clickhouse_db: None, redis_db: None) -> None:
     spans_storage = get_storage(StorageKey("eap_spans"))
+    items_storage = get_storage(StorageKey("eap_items"))
     start = BASE_TIME
     messages = [gen_message(start - timedelta(minutes=i)) for i in range(120)]
     write_raw_unprocessed_events(spans_storage, messages)  # type: ignore
+    write_raw_unprocessed_events(items_storage, messages)  # type: ignore
```
volokluev, line 190: I would extract the double writing to its own function and change that across the board rather than writing the double write everywhere

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -38,6 +40,160 @@
...
+            key=literal(attr_key.name),
+            alias=alias,
+        )
+
+    raise BadSnubaRPCRequestException(
+        f"Attribute {attr_key.name} has an unknown type: {AttributeKey.Type.Name(attr_key.type)}"
+    )
+
+
+def apply_virtual_columns_eap_items(
```
volokluev, line 129: where does this function get called?

volokluev:
Oh one more thing so we can compare performance:

Can you add something that lets us force using the eap_spans table from the client side so we can compare performance?

Maybe a special prefix in the referrer or something

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -38,6 +40,160 @@
...
+    "sentry.timestamp": [
+        AttributeKey.Type.TYPE_FLOAT,
+        AttributeKey.Type.TYPE_DOUBLE,
+        AttributeKey.Type.TYPE_INT,
+        AttributeKey.Type.TYPE_STRING,
+    ],
+    "sentry.trace_id": [
+        AttributeKey.Type.TYPE_STRING
+    ],  # this gets converted from a uuid to a string in a storage processor
+    "sentry.item_id": [AttributeKey.Type.TYPE_STRING],
```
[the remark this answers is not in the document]
davidtsuk: it's being used in the `attribute_key_to_expression_eap_items` function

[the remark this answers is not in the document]
volokluev: I meant that it's used here too but not explicitly i.e.

`/s/"sentry.organization_id"/f"{PREFIX}organization_id"`

just a nit

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: it doesn't get called yet but when I update the trace item table endpoint, it will get used

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -77,6 +77,9 @@
 
 
 def use_eap_items_table(request_meta: RequestMeta) -> bool:
+    if request_meta.referrer.startswith("force_use_eap_items_table"):
```
volokluev, line 80: I want to force using the OLD table

think about this: I run a timeseries query for some ingested timeframe, if `use_eap_items_table` is turned on, there's no way to for me to compare that query against the old table.

volokluev (approved):

PR #6935, opened 2025-03-05
fix(eap): Also compute reliability for negative values

jan-auer:
### Legal Boilerplate

### DIFF
--- a/snuba/web/rpc/v1/resolvers/common/aggregation.py
+++ b/snuba/web/rpc/v1/resolvers/common/aggregation.py
@@ -36,7 +36,7 @@ sampling_weight_column = column("sampling_weight")
 z_value = 1.96
 
 PERCENTILE_PRECISION = 100000
-CONFIDENCE_INTERVAL_THRESHOLD = 1.5
+CONFIDENCE_INTERVAL_THRESHOLD = 0.5
 
 CUSTOM_COLUMN_PREFIX = "__snuba_custom_column__"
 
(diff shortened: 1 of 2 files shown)


shellmayr (approved):

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -151,7 +151,7 @@ def reliability(self) -> Reliability.ValueType:
             return Reliability.RELIABILITY_UNSPECIFIED
 
         relative_confidence = (
-            (self.value + self.confidence_interval) / self.value
+            abs(self.confidence_interval / self.value)
             if self.value != 0
             else float("inf")
```
jan-auer, line 156: This default is unfortunate - it means that values of `0` will always report as low confidence in the product. For now, I would suggest to special case this to `0` if `self.confidence_interval == 0`. In that case, at least for 100% sample rates, we will report high confidence. Then we can work with product to investigate a better approach.

davidtsuk: That default was added to avoid divide by 0 errors. I think infinity is a natural default here as the limit of ci / value as value approaches 0 is infinity. Changing it to 0 won't fix the issue as dividing by a small number will still result in a high relative confidence (a property of relative errors). For the case with 100% sample rates, I think it's logical to return 0 when both the confidence interval and the value is 0.

PR #6936, opened 2025-03-05
fix(ourlogs): Logs trace item table should support offset

k-fish:
### Summary
We need to be able to paginate, this adds offset similar to the eap spans trace items table.

Fixes https://github.com/getsentry/sentry/pull/86397


colin-sentry (approved):
we should switch to page token pagination once sentry backend supports it

PR #6953, opened 2025-03-07
feat: drop item_attrs mv

kylemumma:
this drops the item_attrs mv for now bc it was causing inserted rows to go up a lot.

i make backwards ops the whole table again bc i feel like if i make no backwards ops it wont let me bc its not symmetric

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0035_drop_item_attrs.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0035_drop_item_attrs.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+from typing import Sequence
+
+from snuba.clusters.storage_sets import StorageSetKey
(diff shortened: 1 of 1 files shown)


snuba/snuba_migrations/events_analytics_platform/0035_drop_item_attrs.py
```diff
@@ -0,0 +1,112 @@
...
+            ),
+            operations.DropTable(
+                storage_set=self.storage_set_key,
+                table_name=self.local_table,
+                target=OperationTarget.LOCAL,
+            ),
+            operations.DropTable(
+                storage_set=self.storage_set_key,
+                table_name=self.dist_table,
+                target=OperationTarget.DISTRIBUTED,
```
volokluev, line 110: you don't need to drop these in this migration, just the MV will be fine

volokluev (approved):

PR #6956, opened 2025-03-10
feat: new mv for item_attrs

kylemumma:
this recreates the materialized view from https://github.com/getsentry/snuba/pull/6943 but that one had to be reverted bc too many rows were being inserted. this should solve the issue by adding a group by.

Resolves https://github.com/getsentry/eap-planning/issues/194


volokluev:
How did you validate that this would fix the issue?

volokluev (dismissed):

volokluev (approved):
We ran this in production

PR #6958, opened 2025-03-11
fix(eap-items): swap description and raw_description

davidtsuk:


phacops (changes requested):
I think we should keep the name `description` for the raw description and use `normalized_description` elsewhere.

phacops:
I think you also need to remove the alias in https://github.com/getsentry/snuba/blob/master/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py#L79 otherwise we'll select a non-existent attribute.

phacops (dismissed):

phacops (approved):

PR #6961, opened 2025-03-11
fix(eap-items): add attribute mappings for backwards compatibility

davidtsuk:
The `eap_spans` table top level columns that were used to store certain attributes like `sentry.description`. Since `eap_items` doesn't have those columns, we need we map the old attribute names to the new attribute names.

We use `attributes_float` for all attribute types other than string to allow us to use the bucketing optimization


snuba/settings/settings_docker.py
```diff
@@ -17,3 +17,5 @@
 DOGSTATSD_PORT = env("DOGSTATSD_PORT")
 
 SENTRY_DSN = env("SENTRY_DSN")
+
+USE_EAP_ITEMS_TABLE = True
```
onkar, line 21: Is there a reason to not read this through the environment like other variables?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -68,22 +68,28 @@
 }
 
 PROTO_TYPE_TO_ATTRIBUTE_COLUMN: Final[Mapping[AttributeKey.Type.ValueType, str]] = {
-    AttributeKey.Type.TYPE_INT: "attributes_int",
+    AttributeKey.Type.TYPE_INT: "attributes_float",
     AttributeKey.Type.TYPE_STRING: "attributes_string",
     AttributeKey.Type.TYPE_DOUBLE: "attributes_float",
     AttributeKey.Type.TYPE_FLOAT: "attributes_float",
-    AttributeKey.Type.TYPE_BOOLEAN: "attributes_bool",
+    AttributeKey.Type.TYPE_BOOLEAN: "attributes_float",
```
onkar, line 75: With this change integer and boolean attribute key types are considered as float. Does it make sense to add a comment and explain why is this a right thing to do?

onkar (approved):
LGTM, added a couple of questions.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: this is to make sure we are using the bucketing in those maps, just an optimization

snuba/settings/settings_docker.py
davidtsuk: I didn't see a reason to do that, this is just a simple setting

PR #6965, opened 2025-03-11
fix(admin): fix log profile events nonsense

volokluev:
we were passing in settings in maybe the dumbest way possible. Stop doing that

### DIFF
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -485,7 +485,6 @@ def clickhouse_trace_query() -> Response:
     try_gather_profile_events = req.get("gather_profile_events", True)
     try:
         settings = {}
-        # Append 'log_profile_events=1' with/without the SETTINGS clause to the incoming query.
         if try_gather_profile_events:
             settings["log_profile_events"] = 1
 
--- a/snuba/admin/views.py
+++ b/snuba/admin/views.py
@@ -482,21 +482,16 @@ def clickhouse_trace_query() -> Response:
             ),
             400,
         )
-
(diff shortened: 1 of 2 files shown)


onkar (approved):
As the person who originally introduced this bug, I take the blame.

snuba/admin/views.py
```diff
@@ -482,21 +482,16 @@ def clickhouse_trace_query() -> Response:
             ),
             400,
         )
-
+    try_gather_profile_events = req.get("gather_profile_events", True)
     try:
+        settings = {}
         # Append 'log_profile_events=1' with/without the SETTINGS clause to the incoming query.
```
untitaker, line 488: comment doesn't fully make sense anymore

untitaker (approved):

snuba/admin/views.py
```diff
@@ -482,21 +482,15 @@ def clickhouse_trace_query() -> Response:
...
-            if "log_profile_events=1" not in remaining:
-                raw_sql += ", log_profile_events=1"
-        else:
-            raw_sql += " SETTINGS log_profile_events=1"
+        settings = {}
+        if try_gather_profile_events:
+            settings["log_profile_events"] = 1
 
-        query_trace = run_query_and_get_trace(storage, raw_sql)
+        query_trace = run_query_and_get_trace(storage, raw_sql, settings)
```
colin-sentry, line 491: will this make it so we can't explicitly add `SETTINGS` anymore to trace queries?

volokluev: no

Issue #6964, opened 2025-03-11
Support multiple reverse migrations


mwarkentin:
There are times when we may want to completely reverse a full group of migrations before re-migrating, especially while setting up a new region or working on a new cluster. 

The current `snuba migrations reverse` supports only reversing a single migration by passing in the group and migration id. Ideally we could specify the migration to reverse to (eg. if we're on migration 0009 and run `snuba migrations reverse --group foo --migration-id 0005` it would undo 0009, 0008, 0007, and 0006 (or maybe 0005 too, not sure..))

mwarkentin:
For some idea, if we did something similar to Django's migration system (which I'm pretty sure snuba migrations were based on initially): https://docs.djangoproject.com/en/5.1/topics/migrations/#reversing-migrations

They support using the same `migrate` command to go forwards and backwards, which makes more sense when thinking about "reverse to this point" IMO.

MeredithAnya:
Addressed in https://github.com/getsentry/snuba/pull/6966

PR #6966, opened 2025-03-12
ref(migrations): add revert command

MeredithAnya:
Addresses https://github.com/getsentry/snuba/issues/6964

**New Usage**

Revert all migrations
```
snuba migrations revert --force
```

Revert all migrations, including the `MigrationGroup.SYSTEM` (aka `migrations_local/dist` table(s))
```
snuba migrations revert --include-system --force
```

Revert all migrations for a group
```
snuba migrations revert --group profile_chunks --force
```

Fake reverting all migrations for a group
```
snuba migrations revert --group profile_chunks --fake --force
```

**Notes**
* The `--include-system` is only relevant for when one wants to revert all migrations, it doesn't do anything when you revert with a specific group
* These new commands shouldn't be added to snuba admin as they are not something we really want to do in production, but more to add flexibility when setting up clusters/regions for the first time

### DIFF
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -133,7 +133,7 @@ def migrate(
     "--readiness-state",
     multiple=True,
     type=click.Choice([r.value for r in ReadinessState], case_sensitive=False),
-    default=None,
+    default=(),
 )
 @click.argument("through", default="all")
 @click.option("--force", is_flag=True)
@@ -178,9 +178,11 @@ def revert(
     try:
         if group:
             migration_group = MigrationGroup(group)
+        elif through != "all":
+            raise click.ClickException(
+                "A migration group must be specified when 'through' is not 'all'."
+            )
         else:
-            if through != "all":
-                raise click.ClickException("Need migration group")
             migration_group = None
         runner.reverse_all(
             through=through,
--- a/snuba/cli/migrations.py
+++ b/snuba/cli/migrations.py
@@ -127,6 +127,76 @@ def migrate(
 
 
 @migrations.command()
+@click.option("-g", "--group", default=None)
+@click.option(
+    "-r",
+    "--readiness-state",
+    multiple=True,
+    type=click.Choice([r.value for r in ReadinessState], case_sensitive=False),
+    default=None,
+)
+@click.argument("through", default="all")
+@click.option("--force", is_flag=True)
+@click.option("--fake", is_flag=True)
+@click.option("--include-system", is_flag=True)
+@click.option(
+    "--log-level", help="Logging level to use.", type=click.Choice(LOG_LEVELS)
+)
+def revert(
+    group: Optional[str],
+    readiness_state: Optional[Sequence[str]],
+    through: str,
+    force: bool,
+    fake: bool,
+    include_system: bool,
+    log_level: Optional[str] = None,
+) -> None:
+    """
+    If group is specified, reverse all the migrations for a group.
+    If no group is specified, reverses all migrations for all groups.
+        * by default SYSTEM migrations are NOT reversed
+        * use --include-system to also reverse SYSTEM migrations
+
+    --force is required
+    """
+
+    readiness_states = (
+        [ReadinessState(state) for state in readiness_state]
+        if readiness_state
+        else None
+    )
+
+    setup_logging(log_level)
+    clusters_to_check = (
+        get_clusters_for_readiness_states(readiness_states, CLUSTERS)
+        if readiness_states
+        else CLUSTERS
+    )
+    check_clickhouse_connections(clusters_to_check)
+    runner = Runner()
+
+    try:
+        if group:
+            migration_group = MigrationGroup(group)
+        else:
+            if through != "all":
+                raise click.ClickException("Need migration group")
+            migration_group = None
+        runner.reverse_all(
+            through=through,
+            force=force,
+            fake=fake,
+            include_system=include_system,
+            group=migration_group,
+            readiness_states=readiness_states,
+        )
+    except MigrationError as e:
+        raise click.ClickException(str(e))
+
+    click.echo("Finished reversing migrations")
+
+
+@migrations.command()
 @click.option("--group", required=True, help="Migration group")
 @click.option("--migration-id", required=True, help="Migration ID")
 @click.option("--force", is_flag=True)
--- a/snuba/migrations/runner.py
+++ b/snuba/migrations/runner.py
@@ -384,6 +384,67 @@ class Runner:
         else:
             self._reverse_migration_impl(migration_key)
 
+    def reverse_all(
+        self,
+        *,
+        through: str = "all",
+        fake: bool = False,
+        force: bool = False,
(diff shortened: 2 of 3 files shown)


snuba/cli/migrations.py
```diff
@@ -128,6 +128,76 @@ def migrate(
...
 
 
+@migrations.command()
+@click.option("-g", "--group", default=None)
+@click.option(
+    "-r",
+    "--readiness-state",
+    multiple=True,
+    type=click.Choice([r.value for r in ReadinessState], case_sensitive=False),
+    default=None,
```
onkar, line 138: According to [this](https://click.palletsprojects.com/en/stable/options/#multiple-options), when `multiple=True`, then default value must be a list or a tuple. I'd set `default=()`.

snuba/cli/migrations.py
```diff
@@ -128,6 +128,76 @@ def migrate(
...
+        * use --include-system to also reverse SYSTEM migrations
+
+    --force is required
+    """
+
+    readiness_states = (
+        [ReadinessState(state) for state in readiness_state]
+        if readiness_state
+        else None
+    )
```
onkar, line 169: This will be changed to
```
readiness_states = [ReadinessState(state) for state in readiness_state] if readiness_state else None
```

snuba/cli/migrations.py
```diff
@@ -128,6 +128,76 @@ def migrate(
     click.echo("Finished running migrations")
 
 
+@migrations.command()
+@click.option("-g", "--group", default=None)
+@click.option(
+    "-r",
+    "--readiness-state",
+    multiple=True,
+    type=click.Choice([r.value for r in ReadinessState], case_sensitive=False),
```
onkar, line 137: To support case insensitivity on [ReadinessState] enum (https://github.com/getsentry/snuba/blob/c636a661b5b3a61aa17e6e63896117572835b7f4/snuba/datasets/readiness_state.py#L6), will we need conversion to upper case?

snuba/cli/migrations.py
```diff
@@ -128,6 +128,76 @@ def migrate(
...
+    )
+    check_clickhouse_connections(clusters_to_check)
+    runner = Runner()
+
+    try:
+        if group:
+            migration_group = MigrationGroup(group)
+        else:
+            if through != "all":
+                raise click.ClickException("Need migration group")
```
onkar, line 185: This can be more descriptive:
```
raise click.ClickException("A migration group must be specified when 'through' is not 'all'.")
```

snuba/cli/migrations.py
```diff
@@ -128,6 +128,76 @@ def migrate(
...
+            migration_group = None
```
onkar, line 186: This can be simplified to avoid extra indentation:
```
try:
    if group:
        migration_group = MigrationGroup(group)
    elif through != "all":
        raise click.ClickException("A migration group must be specified when 'through' is not 'all'.")
    else:
        migration_group = None
```

snuba/migrations/runner.py
```diff
@@ -390,6 +390,67 @@ def get_status(migration_key: MigrationKey) -> Status:
...
+
+        def exact_migration_exists(through: str) -> bool:
+            migration_ids = [
+                key.migration_id
+                for key in completed_migrations
+                if key.migration_id.startswith(through)
+            ]
+            if len(migration_ids) == 1:
+                return True
+            return False
```
onkar, line 436: Should this be renamed to `exact_one_migration_exists`?
```
def exact_one_migration_exists(through: str) -> bool:
    return sum(1 for key in completed_migrations if key.migration_id.startswith(through)) == 1
```

snuba/migrations/runner.py
```diff
@@ -390,6 +390,67 @@ def get_status(migration_key: MigrationKey) -> Status:
...
+        completed_migrations = self._get_completed_migrations(groups)
+
+        if readiness_states:
+            completed_migrations = [
+                m
+                for m in completed_migrations
+                if get_group_readiness_state(m.group) in readiness_states
+            ]
+
+        use_through = False if through == "all" else True
```
onkar, line 426: This could be more concise:
```
use_through = through != "all"
```

snuba/migrations/runner.py
```diff
@@ -390,6 +390,67 @@ def get_status(migration_key: MigrationKey) -> Status:
...
+            ]
+            if len(migration_ids) == 1:
+                return True
+            return False
+
+        if use_through and not exact_migration_exists(through):
+            raise MigrationError(f"No exact match for: {through}")
+
+        if not force:
+            raise MigrationError("Requires force to reverse migrations")
```
onkar, line 442: Should this check be enforced earlier in the logic? Maybe before `exact_migration_exists`?

snuba/migrations/runner.py
```diff
@@ -494,6 +555,34 @@ def get_status(migration_key: MigrationKey) -> Status:
...
+    def _get_completed_migrations(
+        self, groups: Sequence[MigrationGroup]
+    ) -> List[MigrationKey]:
+        """
+        Get a list of completed migrations for a list of groups
+        """
+        migration_status = self._get_migration_status()
+
+        def get_status(migration_key: MigrationKey) -> Status:
+            return migration_status.get(migration_key, Status.NOT_STARTED)
```
onkar, line 567: Not sure why this function is needed.  Why don't we just do:
```
migration_status.get(migration_key, Status.NOT_STARTED)
```
instead of adding this function call overhead?

snuba/migrations/runner.py
```diff
@@ -494,6 +555,34 @@ def get_status(migration_key: MigrationKey) -> Status:
...
+            group_loader = get_group_loader(group)
+            for migration_id in group_loader.get_migrations():
+                migration_key = MigrationKey(group, migration_id)
+                status = get_status(migration_key)
+                if status == Status.IN_PROGRESS:
+                    # can't reverse migrations if one is stuck pending
+                    raise MigrationInProgress(str(migration_key))
+                if status == Status.NOT_STARTED:
+                    continue
+                elif status == Status.COMPLETED:
```
onkar, line 580: the `continue` exits the loop already, so `elif` is unnecessary and can be replaced by `if`.

snuba/cli/migrations.py
MeredithAnya: @onkar I believe that the choice options take care of this https://click.palletsprojects.com/en/stable/options/#choice-options

snuba/cli/migrations.py
MeredithAnya: mine is getting auto formatted to this, I think its probably okay to keep

snuba/migrations/runner.py
MeredithAnya: I think the naming is okay imo, but I will simplify this a bit since we can just return `len(migration_ids) == 1`

snuba/migrations/runner.py
MeredithAnya: updated this a bit to check for another case

onkar (approved):
LGTM

PR #6969, opened 2025-03-14
feat(eap-items): use eap items in all endpoints except get trace

davidtsuk:
This is a continuation of the work required to migrate to from `eap_spans` to `eap_items`

### DIFF
--- a/snuba/web/rpc/v1/endpoint_get_traces.py
+++ b/snuba/web/rpc/v1/endpoint_get_traces.py
@@ -9,7 +9,11 @@ from sentry_protos.snuba.v1.endpoint_get_traces_pb2 import (
     GetTracesResponse,
     TraceAttribute,
 )
-from sentry_protos.snuba.v1.request_common_pb2 import PageToken, TraceItemType
+from sentry_protos.snuba.v1.request_common_pb2 import (
+    PageToken,
+    RequestMeta,
+    TraceItemType,
+)
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import AttributeKey, AttributeValue
 from sentry_protos.snuba.v1.trace_item_filter_pb2 import TraceItemFilter
 
@@ -42,6 +46,8 @@ from snuba.web.rpc.common.debug_info import (
 from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
 from snuba.web.rpc.v1.resolvers.R_eap_spans.common.common import (
     attribute_key_to_expression,
+    attribute_key_to_expression_eap_items,
+    use_eap_items_table,
 )
 
 _DEFAULT_ROW_LIMIT = 10_000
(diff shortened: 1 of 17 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_stats.py
```diff
@@ -100,23 +102,47 @@ def _build_attr_distribution_snuba_request(
...
         function_name="mapConcat",
-        parameters=tuple(column(f"attr_str_{i}") for i in range(ATTRIBUTE_BUCKETS)),
+        parameters=tuple(
+            column(
+                f"attributes_string_{i}"
+                if use_eap_items_table(in_msg.meta)
+                else f"attr_str_{i}"
+            )
+            for i in range(
+                40 if use_eap_items_table(in_msg.meta) else ATTRIBUTE_BUCKETS
```
volokluev, line 128: 40 should be a constant as well

tests/web/rpc/v1/test_endpoint_get_traces.py
```diff
@@ -533,7 +534,7 @@ def test_with_data_and_aggregated_fields_all_keys(
                             key=TraceAttribute.Key.KEY_EARLIEST_FRONTEND_SPAN_DURATION_MS,
                             type=AttributeKey.TYPE_INT,
                             value=AttributeValue(
-                                val_int=1000,
+                                val_int=152,
```
volokluev, line 537: why the change?

volokluev (approved):
looks good overall, just a few nits

tests/web/rpc/v1/test_endpoint_get_traces.py
davidtsuk: In `eap_spans` we calculated the duration ourselves by doing `end_timestamp - start_timestamp`. In `eap_items`, we store `duration_ms` directly from the message, so this change is to account for that. However, I just realized we don't store the `duration_ms` we get in the message in `eap_spans`, so I'll have to update the test to make sure duration_ms matches the real duration.

mcannizz:
Will this change have any impact on the users of EAP's RPC layer? If so, what?

snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -364,19 +445,30 @@ def _list_trace_ids(
                 ),
             ),
             SelectedExpression(
-                name="_sort_timestamp",
+                name="timestamp",
```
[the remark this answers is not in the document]
davidtsuk: yeah, I'm just changing the alias here so that it's more generic, the actual expression will still use `_sort_timestamp`.

davidtsuk:
> Will this change have any impact on the users of EAP's RPC layer? If so, what?

There should be no impact as this change is meant to be backwards compatible. There will be a performance boost once we turn on eap_items since the table has a more optimized sort key.

PR #6944, opened 2025-03-06
feat: storage for new item_attrs table

kylemumma:
this is the storage definition for #6943
will merge after #6943

PR #6970, opened 2025-03-14
feat(autocomplete): autocomplete impl, no materialized views

volokluev:
After realizing that how we we have been doing autocomplete for values in EAP is unscalable (#6944 ), changing approaches to be more approximate but much faster.


snuba/web/rpc/v1/legacy/trace_item_attribute_names.py
```diff
@@ -0,0 +1,107 @@
+import uuid
```
volokluev, line 1: nothing changes in this file from the eap_spans implementation. Once the double write is done, this file can be safely deleted

snuba/query/processors/logical/hash_bucket_functions.py
```diff
@@ -101,3 +100,58 @@ def transform_map_contains_and_array_element_expression(
...
+                return exp
+
+            # TODO: also support hasAll
+            if exp.function_name not in ("has",):
+                return exp
+
+            key = exp.parameters[1]
+            if not isinstance(key, Literal) or not isinstance(key.value, str):
+                return exp
+            bucket_idx = fnv_1a(key.value.encode("utf-8")) % self.num_attribute_buckets
```
onkar, line 138: Can `self.num_attribute_buckets` be `0`? If yes, we should handle it to avoid divide by zero issue.

snuba/web/rpc/v1/trace_item_attribute_values.py
```diff
@@ -19,97 +19,114 @@
...
-    res = Query(
+    attr_value = attribute_key_to_expression_eap_items(request.key)
+    assert attr_value.alias
+    inner_query = Query(
         from_clause=entity,
+        selected_columns=[
+            SelectedExpression(name=attr_value.alias, expression=attr_value)
+        ],
+        condition=_build_conditions(request),
+        offset=request.page_token.offset,
```
onkar, line 104: The `offset` is calculated in the `inner_query` and also on L120 in calculating `res`. Is this intentional? Wondering if the pagination is needed at both inner and outer levels and so should inner `offset` be removed?

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
```diff
@@ -202,7 +203,34 @@ def test_empty_results(self) -> None:
...
             ),
+            key=AttributeKey(name="tag1", type=AttributeKey.TYPE_STRING),
             value_substring_match="this_definitely_doesnt_exist_93710",
         )
         res = AttributeValuesRequest().execute(req)
         assert res.values == []
+
+
+@pytest.mark.clickhouse_db
+@pytest.mark.redis_db
```
onkar, line 214: `redis_db` fixture seems unused. Can be removed.

onkar (approved):
LGTM

snuba/query/processors/logical/hash_bucket_functions.py
volokluev: no it cannot

snuba/web/rpc/v1/trace_item_attribute_values.py
volokluev: offset shouldn't be here at all actually, good catch

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
volokluev: yes it is, it's how the runtime config works

PR #6971, opened 2025-03-14
fix(eap-items): add a floating point sampling weight column

davidtsuk:
Storing the sampling weight as an int produces errors of up to 0.3 on the sampling weight due to rounding. We should store it as a float instead.

### DIFF
--- a/snuba/snuba_migrations/events_analytics_platform/0038_eap_items_add_float_sampling_weight.py
+++ b/snuba/snuba_migrations/events_analytics_platform/0038_eap_items_add_float_sampling_weight.py
@@ -1,0 +1,55 @@
+from typing import Sequence
+
+from snuba.clickhouse.columns import Column, Float
+from snuba.clusters.storage_sets import StorageSetKey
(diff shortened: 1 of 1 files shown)


phacops:
I just want to point out we went the opposite way earlier in the project (https://github.com/getsentry/snuba/pull/6190) because all `-Weighted` functions seemed to require `UInt64`. So, double check that it works.

snuba/snuba_migrations/events_analytics_platform/0038_eap_items_add_float_sampling_weight.py
```diff
@@ -0,0 +1,55 @@
...
+local_table_name = "eap_items_1_local"
+dist_table_name = "eap_items_1_dist"
+
+
+class Migration(migration.ClickhouseNodeMigration):
+    blocking = False
+
+    def forwards_ops(self) -> Sequence[operations.SqlOperation]:
+        return [
+            operations.AddColumn(
```
volokluev, line 19: materialized views should be updated as well

PR #6972, opened 2025-03-14
fix(eap-items): write to new float sampling_weight column

davidtsuk:
don't merge until https://github.com/getsentry/snuba/pull/6971 is deployed


onkar:
Can you fill the description section to describe why this PR is required?

davidtsuk:
closing since the weighted quantiles function in clickhouse doesn't support float weights

PR #6977, opened 2025-03-18
fix(eap-items): fix a failing test

davidtsuk:
The test was failing because it was using local time instead of UTC time


volokluev (approved):

colin-sentry (approved):
good catch

PR #6978, opened 2025-03-18
feat(autocomplete): Implement key autocomplete for eap items

volokluev:
Requires https://github.com/getsentry/snuba/pull/6970 to be merged 
Requires https://github.com/getsentry/sentry/pull/87511 to be merged

Resolves https://github.com/getsentry/eap-planning/issues/204

Implements the key autocomplete without the intersecting attributes filter, no longer supports attribute pagination:

See summarized decision here:

https://www.notion.so/sentry/Deprecating-pagination-in-autocomplete-APIs-1b98b10e4b5d80ab832ff700d49f15c3


snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
```diff
@@ -65,6 +78,21 @@ def visit_TraceItemFilterWrapper(
             self.keys.add(trace_item_filter.comparison_filter.key.name)
 
 
+def _backwards_compatible_mapping_expr() -> Expression:
+    map_elems = []
+    for key, val in ATTRIBUTE_MAPPINGS.items():
+        map_elems.append(key)
+        map_elems.append(val)
```
onkar, line 85: Are these `map_elems` intended to be used anywhere?

snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
```diff
@@ -300,7 +338,18 @@ def t(row: Row) -> TraceItemAttributeNamesResponse.Attribute:
             name=attr_name, type=getattr(AttributeKey.Type, attr_type)
         )
 
-    return list(map(t, query_res.result.get("data", [])))
+    data = query_res.result.get("data", [])
```
onkar, line 341: nit: If `"data"` is not present in `query_res.result`, then we get an empty list, which is further extended but not persisted back to the `result`. Should it be persisted?

onkar (commented):
LGTM, added a couple of nits.

onkar (approved):

snuba/web/rpc/v1/endpoint_trace_item_attribute_names.py
volokluev: good catch

snuba/web/rpc/v1/trace_item_attribute_values.py
```diff
@@ -166,6 +166,12 @@ def _execute(
         self, in_msg: TraceItemAttributeValuesRequest
     ) -> TraceItemAttributeValuesResponse:
         in_msg.limit = in_msg.limit or 1000
+        # HACK (Volo): for backwards compatibility. eap_spans used to store this value,
+        # eap_items does not
+        if in_msg.key == "sentry.service":
```
colin-sentry, line 171: I don't think sentry.service is used anywhere on the sentry side, probably safe to remove

PR #6979, opened 2025-03-19
feat(sampling-in-storage): preflight mode

xurui-c:
Implements preflight mode according to https://github.com/getsentry/eap-planning/issues/212

https://github.com/getsentry/eap-planning/issues/217

NOTE: DEPENDS ON https://github.com/getsentry/snuba/pull/6981


snuba/datasets/entities/storage_selectors/eap_items.py
```diff
@@ -0,0 +1,47 @@
...
+            return EntityStorageConnection(
+                storage=get_storage(StorageKey.EAP_ITEMS),
+                translation_mappers=storage_connections[0].translation_mappers,
+            )
+
+        else:
+            eap_items_downsample_config = load_configuration_data(
+                f"{Path(__file__).parent.parent.parent.as_posix()}/configuration/events_analytics_platform/storages/eap_items.yaml",
+                STORAGE_VALIDATORS,
+            )
```
onkar, line 37: Is there a safer way to load this config? I [looked up](https://github.com/search?q=repo%3Agetsentry%2Fsnuba+load_configuration_data&type=code) how is this function called in other places in snuba and the parent traversal looks a bit odd. If `__file__ ` is not accessible, then `Path(__file__)` will raise an error.

snuba/datasets/plans/storage_processing.py
```diff
@@ -73,13 +76,32 @@ def check_storage_readiness(storage: ReadableStorage) -> None:
             )
 
 
+def _is_downsampled_storage_key(storage_key: str) -> bool:
+    return bool(re.match(r"^StorageKey\.EAP_ITEMS_DOWNSAMPLE_\d+$", storage_key))
```
onkar, line 80: I wonder if this conversion of `StorageKey` to a `str` is necessary. What if we do this:

```
def _is_downsampled_storage_key(storage_key: StorageKey) -> bool:
    return storage_key.value.startswith("eap_items_downsample_")

```

snuba/datasets/plans/storage_processing.py
```diff
@@ -73,13 +76,32 @@ def check_storage_readiness(storage: ReadableStorage) -> None:
...
+
+
 def build_best_plan(
     physical_query: Union[Query, ProcessableQuery[Table]],
     settings: QuerySettings,
     post_processors: Sequence[ClickhouseQueryProcessor] = [],
 ) -> ClickhouseQueryPlan:
     storage_key = StorageKeyFinder().visit(physical_query)
-    storage = get_storage(storage_key)
+    if _is_downsampled_storage_key(str(storage_key)):
```
onkar, line 101: And pass `storage_key` as is here, instead of its `str` representation.

snuba/datasets/plans/storage_processing.py
```diff
@@ -73,13 +76,32 @@ def check_storage_readiness(storage: ReadableStorage) -> None:
...
+def _get_corresponding_table(storage_key: str) -> str:
+    downsampling_factor = re.search(
+        r"StorageKey\.EAP_ITEMS_DOWNSAMPLE_(\d+)", storage_key
+    )
+    assert downsampling_factor is not None
+    return (
+        f"eap_items_1_downsample_{downsampling_factor.group(1)}_local"
+        if get_cluster(StorageSetKey.EVENTS_ANALYTICS_PLATFORM).is_single_node()
+        else f"eap_items_1_downsample_{downsampling_factor.group(1)}_dist"
+    )
```
onkar, line 92: If the `str` conversion is unnecessary, you should be able to do this:
```
def _get_corresponding_table(storage_key: StorageKey) -> str:
    match = re.search(r"eap_items_downsample_(\d+)", storage_key.value)
    assert match is not None, f"Invalid downsampled storage key: {storage_key}"
    
    factor = match.group(1)
    cluster = get_cluster(StorageSetKey.EVENTS_ANALYTICS_PLATFORM)
    
    return f"eap_items_1_downsample_{factor}_{'local' if cluster.is_single_node() else 'dist'}"
```

onkar (commented):
I know this is a draft PR, but left a few comments.

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
xurui-c, line 1: please ignore reviewing this file until https://github.com/getsentry/snuba/pull/6981 is merged

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -45,6 +47,14 @@
 from tests.conftest import SnubaSetConfig
 from tests.helpers import write_raw_unprocessed_events
 
+SEED = 331
+RANDOM_GENERATOR = random.Random(SEED)
+
+
+def seeded_uuid4() -> uuid.UUID:
+    print("helpppp", type(uuid.UUID(int=RANDOM_GENERATOR.getrandbits(128))))
```
xurui-c, line 55: will delete this

snuba/datasets/entities/storage_selectors/eap_items.py
volokluev: I agree, this is not a good way to do this. You can achieve the same result with:

```python3

class DynamicStorage(ReadableTableStorage):
    def __init__(self, source_storage, new_schema):
        self._source_storage = source_storage
        self._new_schema = new_schema

    __getattr__(self, name):
       return getattr(self._source_storage, name)

    def get_schema(self):
        return self._new_schema

from copy import deepcopy
storage = get_storage(StorageKey.EAP_ITEMS)
cur_schema = storage.get_schema()
new_schema = TableSchema(
        columns=cur_schema.get_columns()
        local_table_name="YOURLOCALTABLENAME",
        dist_table_name="YOURDISTTABLENAME",
        storage_set_key=cur_schema.get_storage_set_key(),
)
return DynamicStorage(storage, new_schema)
```

snuba/datasets/plans/storage_processing.py
```diff
@@ -145,6 +172,21 @@ def apply_storage_processors(
...
+                table_name=_get_corresponding_table(storage_key),
+                schema=original_table.schema,
+                storage_key=original_table.storage_key,
+                allocation_policies=original_table.allocation_policies,
+                final=original_table.final,
+                sampling_rate=original_table.sampling_rate,
+                mandatory_conditions=original_table.mandatory_conditions,
+            )
+        )
+
```
volokluev, line 189: Why is there modifications in the core query pipeline for this one storage selector?

snuba/query/query_settings.py
```diff
@@ -133,6 +134,17 @@ def get_asynchronous(self) -> bool:
...
+    """
+    Tiers indicate which storage tier to direct the query to. This is used by the TimeSeriesEndpoint and the TraceItemTableEndpoint
+    in EAP.
+    """
+
+    def set_tier(self, tier: int) -> None:
+        self.__tier = tier
+
+    def get_tier(self) -> int:
+        return self.__tier
```
volokluev, line 146: naming: `set_sampling_tier`, `get_sampling_tier`

volokluev: also these should be enums, not raw integers

snuba/downsampled_storage_tiers.py
xurui-c, line 1: I think this file can't be in `snuba/web/rpc` because it's being imported by `query_settings.py`

I initially put it in `snuba/web/rpc/v1/resolvers/R_eap_spans.common/` but I got the following error:
```
ImportError while loading conftest '/Users/rachelchen/snuba/tests/conftest.py'.
tests/conftest.py:15: in <module>
    from snuba.datasets.factory import reset_dataset_factory
snuba/datasets/factory.py:9: in <module>
    from snuba.datasets.configuration.dataset_builder import build_dataset_from_config
snuba/datasets/configuration/dataset_builder.py:6: in <module>
    from snuba.datasets.pluggable_dataset import PluggableDataset
snuba/datasets/pluggable_dataset.py:5: in <module>
    from snuba.datasets.dataset import Dataset
snuba/datasets/dataset.py:6: in <module>
    from snuba.datasets.entities.factory import get_entity
snuba/datasets/entities/factory.py:9: in <module>
    from snuba.datasets.configuration.entity_builder import build_entity_from_config
snuba/datasets/configuration/entity_builder.py:16: in <module>
    from snuba.datasets.configuration.utils import parse_columns
snuba/datasets/configuration/utils.py:20: in <module>
    from snuba.query.processors.physical import ClickhouseQueryProcessor
snuba/query/processors/physical/__init__.py:8: in <module>
    from snuba.query.query_settings import QuerySettings
snuba/query/query_settings.py:5: in <module>
    from snuba.web.rpc.v1.resolvers.R_eap_spans.common.downsampled_storage_tiers import Tier
snuba/web/rpc/__init__.py:242: in <module>
    import_submodules_in_directory(module_path, f"snuba.web.rpc.{v}")
snuba/utils/registered_class.py:179: in import_submodules_in_directory
    imported_modules.append(importlib.import_module(module_str))
snuba/web/rpc/v1alpha/__init__.py:3: in <module>
    from snuba.web.rpc.v1alpha.timeseries.timeseries import AggregateBucketRequest
snuba/web/rpc/v1alpha/timeseries/timeseries.py:18: in <module>
    from snuba.web.rpc.common.eap_execute import run_eap_query_async
snuba/web/rpc/common/eap_execute.py:9: in <module>
    from snuba.datasets.entities.factory import get_entity
E   ImportError: cannot import name 'get_entity'
```

snuba/web/db_query.py
```diff
@@ -176,6 +176,8 @@ def execute_query(
     # Apply clickhouse query setting overrides
     clickhouse_query_settings.update(query_settings.get_clickhouse_settings())
 
+    print("formatted_queryyy", formatted_query)
```
xurui-c, line 179: will delete

volokluev (commented):
Looks good overall, let's just remove the seeded RNGs and just test that:

1. the preflight mode returns fewer queries
2. the response metadata is updated correctly

snuba/datasets/entities/storage_selectors/eap_items.py
```diff
@@ -0,0 +1,27 @@
...
+        query: Query,
+        query_settings: QuerySettings,
+        storage_connections: Sequence[EntityStorageConnection],
+    ) -> EntityStorageConnection:
+        assert isinstance(query_settings, HTTPQuerySettings)
+        tier = query_settings.get_sampling_tier().value
+        if tier == 1:
+            storage_key = StorageKey.EAP_ITEMS
+        else:
+            storage_key = getattr(StorageKey, f"EAP_ITEMS_DOWNSAMPLE_{tier}")
```
onkar, line 23: nit: I am not following why we need to do raw integer comparison here. Can we not use the enum like this:
```
tier = query_settings.get_sampling_tier()

if tier == Tier.TIER_1:
    storage_key = StorageKey.EAP_ITEMS
else:
    storage_key = getattr(StorageKey, f"EAP_ITEMS_DOWNSAMPLE_{tier.value}")
```

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -104,6 +109,17 @@
 }
 
 
+def add_tier_to_query_settings(
```
onkar, line 112: nit: naming could be changed to something more specific like `set_sampling_tier_512` to be explicit. Also, I see that this function is called conditionally, so that condition can be handled inside this function instead of outside:

```
def set_sampling_tier_512(
    request: TimeSeriesRequest | TraceItemTableRequest,
    query_settings: HTTPQuerySettings,
) -> None:
    if not request.meta.HasField("downsampled_storage_config"):
        return  # No downsampled storage config, so do nothing

    if (
        request.meta.downsampled_storage_config.mode
        == DownsampledStorageConfig.MODE_PREFLIGHT
    ):
        query_settings.set_sampling_tier(Tier.TIER_512)
```

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_time_series.py
```diff
@@ -362,6 +363,9 @@ def _build_snuba_request(request: TimeSeriesRequest) -> SnubaRequest:
         setup_trace_query_settings() if request.meta.debug else HTTPQuerySettings()
     )
 
+    if request.meta.HasField("downsampled_storage_config"):
+        add_tier_to_query_settings(request, query_settings)
+
```
onkar, line 368: This and other occurrences become:
```
set_sampling_tier_512(request, query_settings)
```

onkar (approved):
LGTM mostly, added a few nits. Didn't look into the test failures, but are those happening because of the use of `random`? Maybe better to simplify to not use `random` if that is the cause of the failure.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
xurui-c: If the mode isn't preflight we shouldn't add tier 512 to query settings. the behavior of the function depends on the mode so I didn't want to specify which tier we're setting in the name

snuba/web/rpc/common/debug_info.py
```diff
@@ -18,11 +20,35 @@ def extract_response_meta(
...
+
+    if extract_sampling_tier:
+        assert (
+            len(query_results) == 1
+        ), "we can only extract 1 result at a time for EndpointTimeSeres and EndpointTraceItemTable"
+        sampling_tier = (
+            query_results[0].extra.get("stats", {}).get("sampling_tier", Tier.TIER_1)
+        )
+        downsampled_storage_meta = DownsampledStorageMeta(
+            tier=getattr(
```
xurui-c, line 37: I could get rid of `extract_sampling_tier`, but then the function would have to check if `sampling_tier` is set and make implicit assumption that len(query_results) == 1..idk

volokluev: just go through all of them and pick the highest one?

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
```diff
@@ -362,18 +362,22 @@ def trace_item_type(cls) -> TraceItemType.ValueType:
...
-        snuba_request = _build_snuba_request(in_msg)
+        query_settings = construct_query_settings(in_msg)
+
+        snuba_request = _build_snuba_request(in_msg, query_settings)
         res = run_query(
             dataset=PluggableDataset(name="eap", all_entities=[]),
             request=snuba_request,
             timer=self._timer,
         )
+        add_sampling_tier_to_query_stats(res, query_settings)
```
volokluev, line 373: may be better to do this in `db_query` so you don't have to repeat yourself

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -1277,3 +1281,52 @@ def use_eap_items_table(
...
+        granularity_secs = 15
+        query_duration = granularity_secs * 1
+        store_spans_timeseries(
+            BASE_TIME,
+            1,
+            query_duration,
+            metrics=[DummyMetric("test_preflight_metric", get_value=lambda x: 1)],
+        )
+
+        preflight_message = TimeSeriesRequest(
```
volokluev, line 1296: this test should run:

1. The non-sampled request
2. the sampled request

and then compare that (2) has fewer elements than 1 by at least a factor of 100

volokluev (commented):
mostly good., just a few comments

davidtsuk (approved):
lgtm

PR #6980, opened 2025-03-19
Revert "feat(eap-items): use eap items in all endpoints except get trace"

davidtsuk:
Reverts getsentry/snuba#6969
reverting cause it's breaking sentry CI

### DIFF
--- a/snuba/datasets/configuration/events_analytics_platform/storages/eap_items.yaml
+++ b/snuba/datasets/configuration/events_analytics_platform/storages/eap_items.yaml
@@ -219,10 +219,6 @@ query_processors:
   - processor: UUIDColumnProcessor
     args:
       columns: [trace_id]
-  - processor: HexIntColumnProcessor
-    args:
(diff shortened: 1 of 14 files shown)


xurui-c (approved):

mcannizz:
@davidtsuk why did https://github.com/getsentry/snuba/pull/6969 break Sentry CI?

davidtsuk:
> @davidtsuk why did #6969 break Sentry CI?

a test that we weren't running failed on the sentry CI and I'm currently investigating the failure and why the test didn't run on our CI.

mcannizz:
@davidtsuk which test broke? I would have expected no test failures because https://github.com/getsentry/snuba/pull/6969 was intended to be backwards compatible, so let's figure out what went wrong.

davidtsuk:
@mcannizz so it turns out we weren't running the tests that broke in our CI, so I've added them to our CI now and fixed the backwards compatibility

PR #6981, opened 2025-03-19
feat(eap-items): use eap items in all endpoints except get trace

davidtsuk:
This is a continuation of the work required to migrate to from `eap_spans` to `eap_items`

### DIFF
--- a/snuba/query/processors/physical/hexint_column_processor.py
+++ b/snuba/query/processors/physical/hexint_column_processor.py
@@ -15,14 +15,17 @@ from snuba.query.processors.physical.type_converters import (
 
 
 class HexIntColumnProcessor(BaseTypeConverter):
-    def __init__(self, columns: Set[str]) -> None:
+    def __init__(self, columns: Set[str], size: int = 16) -> None:
         super().__init__(columns, optimize_ordering=True)
-        self._size = 16
+        self._size = size
 
     def _translate_literal(self, exp: Literal) -> Literal:
         try:
             assert isinstance(exp.value, str)
-            return Literal(alias=exp.alias, value=int(exp.value, self._size))
+            # 128 bit integers in clickhouse need to be referenced as strings
+            if self._size == 32:
+                return Literal(alias=exp.alias, value=str(int(exp.value, 16)))
+            return Literal(alias=exp.alias, value=int(exp.value, 16))
         except (AssertionError, ValueError):
             raise ColumnTypeError("Invalid hexint", should_report=False)
 
--- a/snuba/query/processors/physical/hexint_column_processor.py
+++ b/snuba/query/processors/physical/hexint_column_processor.py
@@ -16,6 +16,9 @@ from snuba.query.processors.physical.type_converters import (
 
 class HexIntColumnProcessor(BaseTypeConverter):
     def __init__(self, columns: Set[str], size: int = 16) -> None:
+        """
+        size is the number of characters in the hex string representation of the integer (e.g. 32 for 128 bit integers)
+        """
         super().__init__(columns, optimize_ordering=True)
         self._size = size
 
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -40,6 +40,21 @@ TIMESTAMP_COLUMNS: Final[Set[str]] = {
     "sentry.end_timestamp",
 }
 
+ATTRIBUTES_TO_EXCLUDE_IN_EAP_ITEMS: Final[Set[str]] = {
+    "sentry.raw_description",
+    "sentry.transaction",
+    "sentry.start_timestamp_precise",
+    "sentry.end_timestamp_precise",
+    "sentry.duration_ms",
+    "sentry.event_id",
+    "sentry.exclusive_time_ms",
+    "sentry.is_segment",
+    "sentry.parent_span_id",
+    "sentry.profile_id",
+    "sentry.received",
+    "sentry.segment_id",
+}
+
 COLUMN_PREFIX: str = "sentry."
 
 NORMALIZED_COLUMNS_EAP_ITEMS: Final[
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -155,6 +155,11 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
             )
 
         # To maintain backwards compatibility with the old span_id column, we only need the last 16 characters of the item_id
+        # In eap_items, they're just integers, there's no issue fitting a 64 bit integer in a 128 bit integer.
+        # The problem is that this is just our internal representation, but when a user interacts with span_id through EAP, it's treated as a hex string with 16 characters.
+        # However, a 128 bit integer cannot be represented as a 16 character hex string (16 characters can at most represent a 64 bit integer). Hence,
+        # by default we represent item_id as a 32 character hex string. Since the user expects 16 characters and we know that a span_id will currently never use the full 128 bits,
+        # it's safe to get rid of the first 16 characters since we know those will just be padding.
         if attr_key.name == "sentry.span_id":
             return f.right(column(attr_name[len(COLUMN_PREFIX) :]), 16, alias=alias)
 
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -128,6 +128,10 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
                 f"Attribute {attr_key.name} must be one of [{formatted_attribute_types}], got {AttributeKey.Type.Name(attr_key.type)}"
             )
 
+        # To maintain backwards compatibility with the old span_id column, we only need the last 16 characters of the item_id
+        if attr_key.name == "sentry.span_id":
+            return f.right(column(attr_name[len(COLUMN_PREFIX) :]), 16, alias=alias)
+
         return f.CAST(
             column(attr_name[len(COLUMN_PREFIX) :]),
             PROTO_TYPE_TO_CLICKHOUSE_TYPE[attr_key.type],
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -57,6 +57,7 @@ NORMALIZED_COLUMNS_EAP_ITEMS: Final[
         AttributeKey.Type.TYPE_STRING
     ],  # this gets converted from a uuid to a string in a storage processor
     f"{COLUMN_PREFIX}item_id": [AttributeKey.Type.TYPE_STRING],
+    f"{COLUMN_PREFIX}sampling_weight": [AttributeKey.Type.TYPE_DOUBLE],
 }
 
 PROTO_TYPE_TO_CLICKHOUSE_TYPE: Final[Mapping[AttributeKey.Type.ValueType, str]] = {
@@ -112,26 +113,26 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
             f"attribute key {attr_key.name} must have a type specified"
         )
 
-    if attr_key.name in NORMALIZED_COLUMNS_EAP_ITEMS:
-        if attr_key.type not in NORMALIZED_COLUMNS_EAP_ITEMS[attr_key.name]:
+    attr_name = attr_key.name
+    if attr_key.name in ATTRIBUTE_MAPPINGS:
+        attr_name = ATTRIBUTE_MAPPINGS[attr_key.name]
+
+    if attr_name in NORMALIZED_COLUMNS_EAP_ITEMS:
+        if attr_key.type not in NORMALIZED_COLUMNS_EAP_ITEMS[attr_name]:
             formatted_attribute_types = ", ".join(
-                map(AttributeKey.Type.Name, NORMALIZED_COLUMNS_EAP_ITEMS[attr_key.name])
+                map(AttributeKey.Type.Name, NORMALIZED_COLUMNS_EAP_ITEMS[attr_name])
             )
             raise BadSnubaRPCRequestException(
                 f"Attribute {attr_key.name} must be one of [{formatted_attribute_types}], got {AttributeKey.Type.Name(attr_key.type)}"
             )
 
         return f.CAST(
-            column(attr_key.name[len(COLUMN_PREFIX) :]),
+            column(attr_name[len(COLUMN_PREFIX) :]),
             PROTO_TYPE_TO_CLICKHOUSE_TYPE[attr_key.type],
             alias=alias,
         )
 
     if attr_key.type in PROTO_TYPE_TO_ATTRIBUTE_COLUMN:
-        attr_name = attr_key.name
-        if attr_key.name in ATTRIBUTE_MAPPINGS:
-            attr_name = ATTRIBUTE_MAPPINGS[attr_key.name]
-
         if attr_key.type == AttributeKey.Type.TYPE_BOOLEAN:
             return f.CAST(
                 SubscriptableReference(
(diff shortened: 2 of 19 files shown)


volokluev:
What was the change that fixed the sentry CI breakage? Is that test running in CI now?

davidtsuk:
> What was the change that fixed the sentry CI breakage? Is that test running in CI now?

I added the test to CI

davidtsuk:
@volokluev @mcannizz I added the test to the CI and made it backwards compatible now. The problem was that the test wasn't expecting to receive the new attributes we were adding to the attributes columns, which were previously stored in their own columns. I have excluded them from the results and added a test to ensure that we return the same number of attributes as before.

snuba/query/processors/physical/hexint_column_processor.py
```diff
@@ -15,14 +15,17 @@
...
     def _translate_literal(self, exp: Literal) -> Literal:
         try:
             assert isinstance(exp.value, str)
-            return Literal(alias=exp.alias, value=int(exp.value, self._size))
+            # 128 bit integers in clickhouse need to be referenced as strings
+            if self._size == 32:
+                return Literal(alias=exp.alias, value=str(int(exp.value, 16)))
+            return Literal(alias=exp.alias, value=int(exp.value, 16))
         except (AssertionError, ValueError):
             raise ColumnTypeError("Invalid hexint", should_report=False)
```
onkar, line 30: Should this be:
```
try:
    parsed_value = int(exp.value, 16)
    return Literal(alias=exp.alias, value=str(parsed_value) if self._size == 32 else parsed_value)
except ValueError:
    raise ColumnTypeError("Invalid hexint", should_report=False)
```
Currently, the literal can be sometimes a `str` or an `int`. Does it make sense to standardize it to one type?

snuba/web/rpc/v1/resolvers/R_eap_spans/resolver_trace_item_table.py
```diff
@@ -220,12 +257,18 @@ def _column_to_expression(column: Column) -> Expression:
...
+            key=EntityKey("eap_items"),
+            schema=get_entity(EntityKey("eap_items")).get_data_model(),
+            sample=None,
+        )
+    else:
+        entity = Entity(
+            key=EntityKey("eap_spans"),
+            schema=get_entity(EntityKey("eap_spans")).get_data_model(),
+            sample=None,
+        )
```
xurui-c, line 271: nit, nonblocking
this part is the same as in `resolver_time_series.py` right? can we consolidate?

snuba/query/processors/physical/hexint_column_processor.py
```diff
@@ -15,14 +15,17 @@
 
 
 class HexIntColumnProcessor(BaseTypeConverter):
-    def __init__(self, columns: Set[str]) -> None:
+    def __init__(self, columns: Set[str], size: int = 16) -> None:
         super().__init__(columns, optimize_ordering=True)
-        self._size = 16
+        self._size = size
```
volokluev, line 20: you should specify what "size" means, not very clear just looking at the signature or variable name

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -40,6 +40,21 @@
...
+    "sentry.duration_ms",
+    "sentry.event_id",
+    "sentry.exclusive_time_ms",
+    "sentry.is_segment",
+    "sentry.parent_span_id",
+    "sentry.profile_id",
+    "sentry.received",
+    "sentry.segment_id",
+}
+
```
volokluev, line 57: can you add a comment to explain what this is for?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -57,6 +72,7 @@
         AttributeKey.Type.TYPE_STRING
     ],  # this gets converted from a uuid to a string in a storage processor
     f"{COLUMN_PREFIX}item_id": [AttributeKey.Type.TYPE_STRING],
+    f"{COLUMN_PREFIX}sampling_weight": [AttributeKey.Type.TYPE_DOUBLE],
```
volokluev, line 75: sampling weight is a double?

onkar (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -80,6 +96,8 @@
     "sentry.description": "sentry.normalized_description",
     "sentry.span_id": "sentry.item_id",
     "sentry.segment_name": "sentry.transaction",
+    "sentry.start_timestamp": "sentry.start_timestamp_precise",
+    "sentry.end_timestamp": "sentry.end_timestamp_precise",
```
volokluev, line 100: can you add a comment explaining what these are as well?

volokluev: I may need to use them in my work too but I need to understand how they're meant to be used

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -112,26 +130,30 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
             f"attribute key {attr_key.name} must have a type specified"
         )
 
-    if attr_key.name in NORMALIZED_COLUMNS_EAP_ITEMS:
-        if attr_key.type not in NORMALIZED_COLUMNS_EAP_ITEMS[attr_key.name]:
+    attr_name = attr_key.name
+    if attr_key.name in ATTRIBUTE_MAPPINGS:
+        attr_name = ATTRIBUTE_MAPPINGS[attr_key.name]
```
volokluev, line 135: nit: 
```suggestion
    attr_name = ATTRIBUTE_MAPPINGS.get(attr_key.name, attr_key.name)
```

snuba/query/processors/physical/hexint_column_processor.py
onkar: Sorry, the code is wrong. I meant should it be converted to either a `str` or an `int`?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -112,26 +130,30 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
...
+                map(AttributeKey.Type.Name, NORMALIZED_COLUMNS_EAP_ITEMS[attr_name])
             )
             raise BadSnubaRPCRequestException(
                 f"Attribute {attr_key.name} must be one of [{formatted_attribute_types}], got {AttributeKey.Type.Name(attr_key.type)}"
             )
 
+        # To maintain backwards compatibility with the old span_id column, we only need the last 16 characters of the item_id
+        if attr_key.name == "sentry.span_id":
+            return f.right(column(attr_name[len(COLUMN_PREFIX) :]), 16, alias=alias)
+
```
volokluev, line 149: what is the backwards incompatible behavior that happens if you send back the full span_id?I imagine that this may actually break the client which would use this span id to look something up

tests/web/rpc/v1/test_endpoint_get_traces.py
```diff
@@ -62,7 +63,7 @@ def gen_message(
         timestamp += random.random()
     return {
         "description": span_name,
-        "duration_ms": 152,
+        "duration_ms": 1000,
```
volokluev, line 66: why the change

volokluev (commented):
some questions

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: span_id is a UInt64 while item_id is a UInt128, so in order to maintain backwards compatibility, we only care about the last 8 bytes for span_id

for example, imagine a query that looks like select * where span_id = '<8 byte hex string>'. this wouldn't work if we replaced span_id with item_id

tests/web/rpc/v1/test_endpoint_get_traces.py
davidtsuk: I answered this last time:

> In eap_spans we calculated the duration ourselves by doing end_timestamp - start_timestamp. In eap_items, we store duration_ms directly from the message, so this change is to account for that. However, I just realized we don't store the duration_ms we get in the message in eap_spans, so I'll have to update the test to make sure duration_ms matches the real duration.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
volokluev: so when we store a span in eap_items, they all have the form:

`00000000f123dbee`

?

tests/web/rpc/v1/test_endpoint_get_traces.py
volokluev: cool

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: this is what we did with eap_spans, so I kept it the same

volokluev:
In the runtime config, are you able to turn this on for one org at a time or is it an all or nothing thing?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: In eap_items, they're just integers, there's no issue fitting a 64 bit integer in a 128 bit integer. The problem is that this is just our internal representation, but when a user interacts with span_id through EAP, it's treated as a hex string with 16 characters. However, a 128 bit integer cannot be represented as a 16 character hex string (16 characters can at most represent a 64 bit integer). Hence, by default we represent item_id as a 32 character hex string. Since the user expects 16 characters and we know that a span_id will currently never use the full 128 bits, it's safe to get rid of the first 16 characters since we know those will just be padding.

davidtsuk:
> In the runtime config, are you able to turn this on for one org at a time or is it an all or nothing thing?

I was planning to add that option in a separate PR, but since I'm already addressing other comments I'll just do it in this PR.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
volokluev: ok that's helpful context. I think your above comment can be put into a code comment above.

Also is there a test in snuba that reads the span_id with this truncation mechanism?

snuba/query/processors/physical/hexint_column_processor.py
davidtsuk: added a comment

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: done

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: added a comment, let me know if it's unclear

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -108,13 +111,24 @@ def use_eap_items_table(request_meta: RequestMeta) -> bool:
     if settings.USE_EAP_ITEMS_TABLE:
         return True
 
+    use_eap_items_orgs = state.get_config("use_eap_items_orgs")
+    if use_eap_items_orgs:
+        use_eap_items_orgs = map(int, use_eap_items_orgs.strip("[]").split(","))
+
```
volokluev, line 117: someone mistypes an input in runtime config they can take down the entire API

davidtsuk: good point

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: I believe all the snuba tests that use `span_id` assume it's 16 characters. For example: https://github.com/getsentry/snuba/blob/9ba51e7db6d9850a237e583dc6731047f86dbc46/tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py#L443

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -108,13 +111,24 @@ def use_eap_items_table(request_meta: RequestMeta) -> bool:
...
     if (
         state.get_config("use_eap_items_table", False)
         and use_eap_items_table_start_timestamp_seconds is not None
     ):
+        if (
+            use_eap_items_orgs
+            and request_meta.organization_id not in use_eap_items_orgs
+        ):
+            return False
+
```
volokluev, line 131: this logic is a bit weird. You'd have to eventually put in every org id to turn it on for everyone. I think instead it should be:

1. `use_eap_items_table_all` - turned on for everyone
2. `turned_on_for_org`-  `request_meta.organization_id in use_eap_items_orgs`
3. `within_items_timerange` - aka ` request_meta.start_timestamp.seconds  >= use_eap_items_table_start_timestamp_seconds` 

`return within_items_timerange and (use_eap_items_table_all or turned_on_for_org)`

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
volokluev: but this test is not searching by span_id. What I'm saying is that if we write a 64bit span id to the eap_items table, and we search by that 8 character hex string, we should get that span back. There should be a test for that

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -155,6 +155,11 @@ def attribute_key_to_expression_eap_items(attr_key: AttributeKey) -> Expression:
             )
 
         # To maintain backwards compatibility with the old span_id column, we only need the last 16 characters of the item_id
+        # In eap_items, they're just integers, there's no issue fitting a 64 bit integer in a 128 bit integer.
+        # The problem is that this is just our internal representation, but when a user interacts with span_id through EAP, it's treated as a hex string with 16 characters.
+        # However, a 128 bit integer cannot be represented as a 16 character hex string (16 characters can at most represent a 64 bit integer). Hence,
+        # by default we represent item_id as a 32 character hex string. Since the user expects 16 characters and we know that a span_id will currently never use the full 128 bits,
+        # it's safe to get rid of the first 16 characters since we know those will just be padding.
         if attr_key.name == "sentry.span_id":
             return f.right(column(attr_name[len(COLUMN_PREFIX) :]), 16, alias=alias)
```
volokluev, line 164: also here, we're using `attr_key.name` and in the expression it's `attr_name[len(COLUMN_PREFIX) :]` should those be the same?

davidtsuk: They're intentionally not the same because one of them refers to the external attr_key name and the other refers to the internal name. I realize now the naming is probably confusing, so I'll update it.

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
```diff
@@ -111,24 +111,32 @@ def use_eap_items_table(request_meta: RequestMeta) -> bool:
...
+        or request_meta.organization_id in eap_items_enabled_orgs
+    )
 
     use_eap_items_table_start_timestamp_seconds = state.get_int_config(
         "use_eap_items_table_start_timestamp_seconds"
     )
 
     if (
-        state.get_config("use_eap_items_table", False)
+        state.get_int_config("use_eap_items_table", 0)
```
volokluev, line 136: nit: is this config still necessary?

volokluev (approved):
great job

snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
davidtsuk: just for extra safety i guess

volokluev: up to you if you wanna keep it, I probably wouldn't

PR #6982, opened 2025-03-19
feat: Enable Spotlight in dev mode by default

BYK:
This is a follow up to getsentry/sentry#87295. Can be disabled by setting `SENTRY_SPOTLIGHT=0` env variable


untitaker (approved):

untitaker:
is this to enable additional functionality for the toolbar in sentry? we don't have a web interface in snuba except snuba admin and I don't think that one even has transaction/perf instrumentation

PR #6984, opened 2025-03-20
ref(dev): change default log level to info

JoshFerge:
makes the default log level of the devserver to be `INFO` instead of `DEBUG` which makes the amount of logs in the devserver much more manageable. includes an environment variable option to use `DEBUG` instead by setting `DEBUG=1` in the environment.


untitaker (approved):
how about adding a `--log-level` cli arg to devserver that forwards its value to each subprocess?

untitaker (approved):

PR #6985, opened 2025-03-20
feat(subscriptions): subscription scheduler followed consumer group override

onewland:
This is a proof-of-concept, validated locally, that we can switch between followed consumer groups in a subscription scheduler without restarting.

We want this in order to run eap_spans subscriptions from the eap_items commit log without having to switch subscription scheduler and executor topics or run a parallel processing pipeline. These are fed from the same topic (`snuba-spans`) and so the commit offsets/timestamps should be comparable.

Some subscriptions results will likely be lost in the cutover, meaning maybe a few seconds without subscriptions executing on eap_spans. This should not be noticeable/important for a pre-GA product

Testing "script":
- created a subscription on eap_spans
- ran `snuba rust-consumer --storage eap_items --consumer-group snuba-items-consumers`
- ran `snuba subscriptions-scheduler --entity eap_spans --followed-consumer-group snuba-spans-consumers` (mimicking current production config)
- inserted some data with scripts/send_spans.py
- **verified that no offsets were committed by the scheduler, and no subscriptions were scheduled** (as desired, the override isn't set yet)
- shut down the `snuba-items-consumers` group consumer
- ran `snuba rust-consumer --storage eap_spans --consumer-group snuba-spans-consumers`
- inserted some data with scripts/send_spans.py
- **verified that the subscription was scheduled**
- set the run-time config to override following snuba-spans-consumers: 
  -   `snuba config set subscriptions_override_consumer_group_snuba-spans-consumers snuba-items-consumers`
- inserted some data with scripts/send_spans.py
- **verified that no subscriptions were scheduled** (once the run-time config is enabled, it should be ignoring commits from the snuba-spans-consumer group in favor of the override snuba-items-consumer. at this point the snuba-items-consumer is not running)
- **verified that we were hitting the override code path via logs**
- shut down the `snuba-spans-consumers` group consumer
- brought up the `snuba-items-consumers` group consumer
- inserted some data with scripts/send_spans.py
- **verified that the subscription was scheduled**

### DIFF
--- a/snuba/subscriptions/scheduler_consumer.py
+++ b/snuba/subscriptions/scheduler_consumer.py
@@ -17,6 +17,7 @@ from snuba.datasets.entities.entity_key import EntityKey
 from snuba.datasets.entities.factory import get_entity
 from snuba.datasets.table_storage import KafkaTopicSpec
 from snuba.redis import RedisClientKey, get_redis_client
+from snuba.state import get_str_config
 from snuba.subscriptions.data import PartitionId
 from snuba.subscriptions.scheduler import SubscriptionScheduler
 from snuba.subscriptions.scheduler_processing_strategy import (
@@ -109,6 +110,16 @@ class CommitLogTickConsumer(Consumer[Tick]):
         self.__synchronization_timestamp = synchronization_timestamp
         self.__time_shift = time_shift.total_seconds() if time_shift is not None else 0
 
+    def followed_group_dynamic(self) -> str:
+        override_group = get_str_config(
+            f"subscriptions_override_consumer_group_{self.__followed_consumer_group}"
+        )
+
+        if override_group is not None:
+            logger.info("override_group tripped")
+            return override_group
+        return self.__followed_consumer_group
+
     def subscribe(
         self,
         topics: Sequence[Topic],
@@ -143,7 +154,7 @@ class CommitLogTickConsumer(Consumer[Tick]):
             )
             return None
 
-        if commit.group != self.__followed_consumer_group:
+        if commit.group != self.followed_group_dynamic():
             return None
 
         previous_message = self.__previous_messages.get(commit.partition)
--- a/snuba/subscriptions/scheduler_consumer.py
+++ b/snuba/subscriptions/scheduler_consumer.py
@@ -116,7 +116,10 @@ class CommitLogTickConsumer(Consumer[Tick]):
         )
 
         if override_group is not None:
-            logger.info("override_group tripped")
+            self.__metrics.increment(
+                "followed_group_override",
+                tags={"original_group": self.__followed_consumer_group},
+            )
             return override_group
         return self.__followed_consumer_group
 
@@ -155,6 +158,10 @@ class CommitLogTickConsumer(Consumer[Tick]):
             return None
 
         if commit.group != self.followed_group_dynamic():
+            self.__metrics.increment(
+                "followed_group_mismatch",
+                tags={"group": commit.group},
+            )
             return None
 
         previous_message = self.__previous_messages.get(commit.partition)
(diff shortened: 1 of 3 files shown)


snuba/subscriptions/scheduler_consumer.py
```diff
@@ -109,6 +110,16 @@ def __init__(
         self.__synchronization_timestamp = synchronization_timestamp
         self.__time_shift = time_shift.total_seconds() if time_shift is not None else 0
 
+    def followed_group_dynamic(self) -> str:
+        override_group = get_str_config(
+            f"subscriptions_override_consumer_group_{self.__followed_consumer_group}"
+        )
+
+        if override_group is not None:
+            logger.info("override_group tripped")
```
onewland, line 119: todo: this should emit a metric

snuba/subscriptions/scheduler_consumer.py
```diff
@@ -143,7 +154,7 @@ def poll(self, timeout: Optional[float] = None) -> Optional[BrokerValue[Tick]]:
             )
             return None
 
-        if commit.group != self.__followed_consumer_group:
+        if commit.group != self.followed_group_dynamic():
```
onewland, line 157: todo: this should emit a metric (inside-the-if case)

onewland:
> The test procedure looks good, only thing I would test which looks like it hasn't been tested is:
> 
> ...
> Or you could yolo it if you feel confident

Tested that both consumers writing at once does what we'd expect (if override is not set, only the original topic is used to commit. else, only the override topic is used)

onkar (approved):

lynnagara:
> This is a proof-of-concept, validated locally, that we can switch between followed consumer groups in a subscription scheduler without restarting.

What's the motivation for this given that restart is a pretty quick and painless thing to do?

Just my 2c, but i think providing multiple mechanisms for the same thing (topic override) is confusing. I think that adding runtime config to the mix of how topic overrides happen can be a little confusing. Topics and overrides are usually managed via CLI args and settings, and I'm not sure it's worth adding a different mechanism.

fpacifici:
Why not just starting a second scheduler that re-consumes the commit log looking for a different consumer group and then shutting down the old one? 
To my knowledge the Sentry result consumer deduplicates alerts (you may want to reach out to the issue team to confirm). Also the executor skips alerts that are older than a certain time so you will reach the end of the topic quickly.

onewland:
> Why not just starting a second scheduler that re-consumes the commit log looking for a different consumer group and then shutting down the old one? To my knowledge the Sentry result consumer deduplicates alerts (you may want to reach out to the issue team to confirm). Also the executor skips alerts that are older than a certain time so you will reach the end of the topic quickly.

I spoke to @wedamija  about this, and I think that the Sentry results consumer deduplicates by timestamp, and I think the generated subscription timestamps could differ between the two consumer groups due to the random jitter added into executions.

onewland:
> > This is a proof-of-concept, validated locally, that we can switch between followed consumer groups in a subscription scheduler without restarting.
> 
> What's the motivation for this given that restart is a pretty quick and painless thing to do?
> 
> Just my 2c, but i think providing multiple mechanisms for the same thing (topic override) is confusing. I think that adding runtime config to the mix of how topic overrides happen can be a little confusing. Topics and overrides are usually managed via CLI args and settings, and I'm not sure it's worth adding a different mechanism.

Yeah, I think that's the right call. I can:
1. deploy the small snuba change in this PR to make eap-items-consumer write to the eap-spans-commit-log
2. deploy a k8s change to make the scheduler follow the other consumer group 

and then we don't have to mess with any application logic. I think I got turned around thinking about deploy order but it's not that complicated

onewland:
Superseded by https://github.com/getsentry/snuba/pull/6990 and https://github.com/getsentry/ops/pull/14552

PR #6990, opened 2025-03-21
feat(eap-spans): push eap-items commits to the commit log

onewland:
We want this in order to run eap_spans subscriptions from the eap_items commit log without having to switch subscription scheduler and executor topics or run a parallel processing pipeline. These are fed from the same topic (snuba-spans) and so the commit offsets/timestamps should be comparable.

This change starts pushing eap_items commits to the spans commit log topic. The scheduler consumer has logic to ignore these messages until an ops configuration change is made. We add a metric here to make sure that path is hit

### DIFF
--- a/snuba/datasets/configuration/events_analytics_platform/storages/eap_items.yaml
+++ b/snuba/datasets/configuration/events_analytics_platform/storages/eap_items.yaml
@@ -245,3 +245,9 @@ mandatory_condition_checkers:
 stream_loader:
   processor: EAPItemsProcessor
   default_topic: snuba-spans
+  commit_log_topic: snuba-eap-spans-commit-log
+  subscription_scheduler_mode: global
+  subscription_synchronization_timestamp: orig_message_ts
+  subscription_scheduled_topic: scheduled-subscriptions-eap-spans
+  subscription_result_topic: eap-spans-subscription-results
+  subscription_delay_seconds: 60
--- a/snuba/subscriptions/scheduler_consumer.py
+++ b/snuba/subscriptions/scheduler_consumer.py
@@ -144,6 +144,10 @@ class CommitLogTickConsumer(Consumer[Tick]):
             return None
 
         if commit.group != self.__followed_consumer_group:
+            self.__metrics.increment(
+                "followed_group_mismatch",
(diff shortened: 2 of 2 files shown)


untitaker:
> We want this in order to run eap_spans subscriptions from the eap_items commit log without having to switch subscription scheduler and executor topics or run a parallel processing pipeline. These are fed from the same topic (snuba-spans) and so the commit offsets/timestamps should be comparable.

so eap-items consumer is inserting into eap-items table, but the subscription scheduler still reads from the spans table?

that defeats the purpose of the commit log. yes, it's containing offsets from the same topic, but that commit-log for eap-spans is about what the spans table contains. with your setup:

* if the eap-items consumer starts to lag, subscriptions are starting to slow down even though the eap-spans consumer did already insert the necessary data, causing unnecessary outage on alerts
* if the eap-spans consumer starts to lag, subscriptions will still see recent commit log from eap-items consumer and start querying the eap-spans table even though it contains no recent data.

onewland:
> > We want this in order to run eap_spans subscriptions from the eap_items commit log without having to switch subscription scheduler and executor topics or run a parallel processing pipeline. These are fed from the same topic (snuba-spans) and so the commit offsets/timestamps should be comparable.
> 
> so eap-items consumer is inserting into eap-items table, but the subscription executor reads from the spans table while being scheduled based on the commit log from eap-items consumer?
> 
> if i understood that right, that defeats the purpose of the commit log. yes, it's containing offsets from the same topic, but that commit-log for eap-spans is about what the spans table contains. with your setup:
> 
> * if the eap-items consumer starts to lag, subscriptions are starting to slow down even though the eap-spans consumer did already insert the necessary data, causing unnecessary outage on alerts
> * if the eap-spans consumer starts to lag, subscriptions will still see recent commit log from eap-items consumer and start querying the eap-spans table even though it contains no recent data.

High-level, items and spans contain the exact same data with a different sort key. This is one step in the process of transitioning from spans to items.

It's true that there's some weird inconsistency stuff happening here for now, but it should be resolved quite soon by switching the queried storage to items (but we're trying to make these changes in a way that's invisible to external clients)

xurui-c (approved):

PR #6991, opened 2025-03-23
feat(sampling-in-storage): best effort mode

xurui-c:
giving my best effort to the best effort mode

Implements best effort mode according to https://github.com/getsentry/eap-planning/issues/212. Also:
1. added functionality to Timer to get the duration between 2 marks before "finishing"
2. refactored build_snuba_request (because it's the same between the time series and trace item table endpoints) to take in build_request (which is different between the endpoints)

https://github.com/getsentry/eap-planning/issues/218

### DIFF
--- a/snuba/utils/metrics/timer.py
+++ b/snuba/utils/metrics/timer.py
@@ -49,6 +49,21 @@ class Timer:
 
         return duration_group
 
+    def get_duration_between_marks(self, start_mark: str, end_mark: str) -> float:
+        start_mark_duration = -1
+        end_mark_duration = -1
+        for name, timestamp in self.__marks:
+            if name == start_mark:
+                start_mark_duration = self.__diff_ms(self.__marks[0][1], timestamp)
+            if name == end_mark:
+                end_mark_duration = self.__diff_ms(self.__marks[0][1], timestamp)
+
+        assert (
+            start_mark_duration != -1 and end_mark_duration != -1
+        ), "Please pass in mark names that exist"
+
+        return end_mark_duration - start_mark_duration
+
     def finish(self) -> TimerData:
         if self.__data is None:
             start = self.__marks[0][1]
--- a/snuba/utils/metrics/timer.py
+++ b/snuba/utils/metrics/timer.py
@@ -52,10 +52,10 @@ class Timer:
     def get_duration_between_marks(self, start_mark: str, end_mark: str) -> float:
         start_mark_duration = -1
         end_mark_duration = -1
-        for name, timestamp in self.__marks:
-            if name == start_mark:
+        for mark, timestamp in self.__marks:
+            if mark == start_mark:
                 start_mark_duration = self.__diff_ms(self.__marks[0][1], timestamp)
-            if name == end_mark:
+            if mark == end_mark:
                 end_mark_duration = self.__diff_ms(self.__marks[0][1], timestamp)
 
         assert (
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -177,6 +177,12 @@ def execute_query(
     # Apply clickhouse query setting overrides
     clickhouse_query_settings.update(query_settings.get_clickhouse_settings())
 
+    if (
+        isinstance(query_settings, HTTPQuerySettings)
+        and query_settings.get_record_query_duration()
+    ):
+        timer.mark("right_before_execute")
+
     result = reader.execute(
         formatted_query,
         clickhouse_query_settings,
--- a/snuba/web/db_query.py
+++ b/snuba/web/db_query.py
@@ -177,11 +177,7 @@ def execute_query(
     # Apply clickhouse query setting overrides
     clickhouse_query_settings.update(query_settings.get_clickhouse_settings())
 
-    if (
-        isinstance(query_settings, HTTPQuerySettings)
-        and query_settings.get_record_query_duration()
-    ):
-        timer.mark("right_before_execute")
+    timer.mark("right_before_execute")
 
     result = reader.execute(
         formatted_query,
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -6,6 +6,7 @@ from sentry_protos.snuba.v1.downsampled_storage_pb2 import DownsampledStorageCon
 from sentry_protos.snuba.v1.endpoint_time_series_pb2 import TimeSeriesRequest
 from sentry_protos.snuba.v1.endpoint_trace_item_table_pb2 import TraceItemTableRequest
 
+from snuba import state
 from snuba.attribution import AppID
 from snuba.attribution.attribution_info import AttributionInfo
 from snuba.datasets.pluggable_dataset import PluggableDataset
@@ -20,15 +21,24 @@ from snuba.web.query import run_query
 T = TypeVar("T", TimeSeriesRequest, TraceItemTableRequest)
 
 
-SENTRY_TIMEOUT = 30000  # 30s = 30000ms
-ERROR_BUDGET = 5000  # 5s
-
 DOWNSAMPLING_TIER_MULTIPLIERS = {
     Tier.TIER_512: 1,
     Tier.TIER_64: 8,
     Tier.TIER_8: 64,
     Tier.TIER_1: 512,
 }
+
+
+def _get_sentry_timeout_ms() -> int:
+    return cast(
+        int, state.get_int_config("sampling_in_storage_sentry_timeout", default=30000)
+    )  # 30s = 30000ms
+
+
+def _get_error_budget() -> int:
+    return cast(
+        int, state.get_int_config("sampling_in_storage_error_budget", default=5000)
+    )  # 5s = 5000ms
 
 
 def _get_query_duration(timer) -> int:
@@ -44,7 +54,10 @@ def _get_target_tier(timer: Timer) -> Tier:
             most_downsampled_query_duration_ms
             * cast(int, DOWNSAMPLING_TIER_MULTIPLIERS.get(tier))
         )
-        if estimated_query_duration_to_this_tier <= SENTRY_TIMEOUT - ERROR_BUDGET:
+        if (
+            estimated_query_duration_to_this_tier
+            <= _get_sentry_timeout_ms() - _get_error_budget()
+        ):
             target_tier = tier
     return target_tier
 
@@ -59,7 +72,10 @@ def _is_best_effort_mode(in_msg: T) -> bool:
 
 def _enough_time_budget_to_at_least_run_next_tier(timer: Timer) -> bool:
     most_downsampled_query_duration_ms = _get_query_duration(timer)
-    return most_downsampled_query_duration_ms * 9 < SENTRY_TIMEOUT - ERROR_BUDGET
+    return (
+        most_downsampled_query_duration_ms * 9
+        < _get_sentry_timeout_ms() - _get_error_budget()
+    )
 
 
 def build_snuba_request(
@@ -113,7 +129,8 @@ def run_query_to_correct_tier(
         timer
     ):
         query_settings.push_clickhouse_setting(
-            "max_execution_time", (SENTRY_TIMEOUT - ERROR_BUDGET) / 1000
+            "max_execution_time",
+            (_get_sentry_timeout_ms() - _get_error_budget()) / 1000,
         )
         query_settings.push_clickhouse_setting("timeout_overflow_mode", "break")
         query_settings.set_sampling_tier(_get_target_tier(timer))
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -23,7 +23,7 @@ T = TypeVar("T", TimeSeriesRequest, TraceItemTableRequest)
 SENTRY_TIMEOUT = 30000  # 30s = 30000ms
 ERROR_BUDGET = 5000  # 5s
 
-TIER_MULTIPLIERS = {
+DOWNSAMPLING_TIER_MULTIPLIERS = {
     Tier.TIER_512: 1,
     Tier.TIER_64: 8,
     Tier.TIER_8: 64,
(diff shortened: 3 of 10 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,123 @@
...
+    return SnubaRequest(
+        id=uuid.UUID(request.meta.request_id),
+        original_body=MessageToDict(request),
+        query=build_query(request),
+        query_settings=query_settings,
+        attribution_info=AttributionInfo(
+            referrer=request.meta.referrer,
+            team="eap",
+            feature="eap",
+            tenant_ids={
```
xurui-c, line 69: I noticed this function was the same between time series and trace item table endpoints so I refactored it this way

snuba/web/db_query.py
```diff
@@ -177,6 +177,12 @@ def execute_query(
     # Apply clickhouse query setting overrides
     clickhouse_query_settings.update(query_settings.get_clickhouse_settings())
 
+    if (
+        isinstance(query_settings, HTTPQuerySettings)
+        and query_settings.get_record_query_duration()
```
xurui-c, line 182: I decided to mark the start of query execution here so we include query processing in our time

volokluev: no need to do this via setting, just do it every time

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,127 @@
...
+
+SENTRY_TIMEOUT = 30000  # 30s = 30000ms
+ERROR_BUDGET = 5000  # 5s
+
+TIER_MULTIPLIERS = {
+    Tier.TIER_512: 1,
+    Tier.TIER_64: 8,
+    Tier.TIER_8: 64,
+    Tier.TIER_1: 512,
+}
```
volokluev: this is already the enum value, no need to repeat it

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,127 @@
...
+
+    if is_best_effort_mode(in_msg):
+        query_settings.push_clickhouse_setting(
+            "max_execution_time", SENTRY_TIMEOUT / 1000 - ERROR_BUDGET
+        )
+        query_settings.push_clickhouse_setting("timeout_overflow_mode", "break")
+        query_settings.set_sampling_tier(_get_target_tier(timer))
+        query_settings.set_record_query_duration(
+            False
+        )  # doesn't make a functional diff
```
volokluev, line 116: what about the case where so much data is scanned in the lowest tier that we have no budget left?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,127 @@
...
+from snuba.request import Request as SnubaRequest
+from snuba.utils.metrics.timer import Timer
+from snuba.web import QueryResult
+from snuba.web.query import run_query
+
+T = TypeVar("T", TimeSeriesRequest, TraceItemTableRequest)
+
+
+SENTRY_TIMEOUT = 30000  # 30s = 30000ms
+ERROR_BUDGET = 5000  # 5s
```
volokluev, line 24: you should make this runtime configurable so you can change it without redeploying

volokluev (commented):
good overall, just a few things

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,122 @@
...
+
+    res = run_query(
+        dataset=PluggableDataset(name="eap", all_entities=[]),
+        request=request_to_most_downsampled_tier,
+        timer=timer,
+    )
+
+    if is_best_effort_mode(in_msg):
+        query_settings.push_clickhouse_setting(
+            "max_execution_time", SENTRY_TIMEOUT / 1000 - ERROR_BUDGET
```
mcannizz, line 107: What are the units of `max_execution_time`? Milliseconds, seconds, something else?

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -1362,3 +1363,89 @@ def test_preflight(self) -> None:
...
+        )
+
+        if best_effort_response.result_timeseries == []:
+            best_effort_metric_sum = 0.0
+        else:
+            best_effort_metric_sum = (
+                best_effort_response.result_timeseries[0].data_points[0].data
+            )
+
+        assert (
```
mcannizz, line 1441: I don't follow what this assertion is testing. Can you explain in a code comment?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3173,3 +3174,76 @@ def test_preflight(self, setup_teardown: Any) -> None:
...
+            ],
+        )
+
+        mock_get_duration_between_marks.return_value = 3124.0
+
+        best_effort_response = EndpointTraceItemTable().execute(best_effort_message)
+        non_downsampled_tier_response = EndpointTraceItemTable().execute(
+            message_to_non_downsampled_tier
+        )
+        assert (
```
mcannizz, line 3240: I don't follow what this assertion is testing. Can you explain in a code comment?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
xurui-c: seconds, fixed to (sentry timeout - error budget) / 1000

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
volokluev: I see

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,140 @@
...
+
+    target_tier = Tier.TIER_NO_TIER
+    for tier in sorted(Tier, reverse=True)[:-1]:
+        estimated_query_duration_to_this_tier = (
+            most_downsampled_query_duration_ms
+            * cast(int, DOWNSAMPLING_TIER_MULTIPLIERS.get(tier))
+        )
+        if estimated_query_duration_to_this_tier <= _get_time_budget():
+            target_tier = tier
+    return target_tier
```
onkar, line 57: If the condition does not get hit, then the `target_tier` will remain as `Tier.TIER_NO_TIER`:
```
if estimated_query_duration_to_this_tier <= _get_time_budget():
    target_tier = tier
```
Is that going to be a problem or defaulting to `TIER_1` is a better option?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,140 @@
...
+
+        request_to_target_tier = build_snuba_request(
+            in_msg, query_settings, build_query
+        )
+
+        res = run_query(
+            dataset=PluggableDataset(name="eap", all_entities=[]),
+            request=request_to_target_tier,
+            timer=timer,
+        )
```
onkar, line 138: - Instead of overwriting `res`, will it be better to store this in `second_res` and return `second_res` immediately from here? If this query fails, we would be overwriting the results of the first query in the most downsampled tier. In that failure case, it's probably better to return the first result.
- Should exception handling be done around `run_query()`?

Thoughts?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,140 @@
...
-            == DownsampledStorageConfig.MODE_PREFLIGHT
-        ):
-            query_settings.set_sampling_tier(Tier.TIER_512)
-
-    return query_settings
+
+
+def _enough_time_budget_to_at_least_run_next_tier(timer: Timer) -> bool:
+    most_downsampled_query_duration_ms = _get_query_duration(timer)
+    return most_downsampled_query_duration_ms * 9 < _get_time_budget()
```
onkar, line 70: Can you please add a comment explaining why `most_downsampled_query_duration_ms` is multiplied by 9?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
xurui-c: what type of error would happen with second_res but not res?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
xurui-c: defaulted to TIER_512

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
xurui-c: refactored code

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
xurui-c: For first bullet point: if the second query fails, we should properly raise that instead of masking it by returning res. Note that the second query will not fail due to timing out due to setting timeout_overflow_mode

For second bullet point: what additional exception handling are you thinking of in addition to the ones that run_query already does?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -1,23 +1,140 @@
...
+    query_settings: HTTPQuerySettings,
+    timer: Timer,
+    build_query: Callable[[T], Query],
+) -> QueryResult:
+    if not in_msg.meta.HasField("downsampled_storage_config"):
+        return run_query(
+            dataset=PluggableDataset(name="eap", all_entities=[]),
+            request=build_snuba_request(in_msg, query_settings, build_query),
+            timer=timer,
+        )
```
volokluev, line 116: You should time the estimation query and emit that as a metric so we know how much time this is taking up. You could probably use the timer class for it that is passed in here

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -57,6 +59,15 @@ def _get_target_tier(timer: Timer) -> Tier:
...
             target_tier = tier
+
+    timer.mark("sampling_in_storage_finished_estimation")
+    metrics_backend.timing(
+        "sampling_in_storage_estimation_duration",
+        timer.get_duration_between_marks(
+            "sampling_in_storage_start_estimation",
+            "sampling_in_storage_finished_estimation",
+        ),
+    )
```
volokluev, line 70: I meant to run the estimation query, this should be pretty fast

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -1362,3 +1363,90 @@ def test_preflight(self) -> None:
...
+                start_timestamp=Timestamp(seconds=int(BASE_TIME.timestamp())),
+                end_timestamp=Timestamp(
+                    seconds=int(BASE_TIME.timestamp() + query_duration)
+                ),
+                trace_item_type=TraceItemType.TRACE_ITEM_TYPE_SPAN,
+            ),
+            aggregations=aggregations,
+            granularity_secs=granularity_secs,
+        )
+        # this forces the query to route to tier 64
```
volokluev, line 1424: would be good to explain why to the reader

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3173,3 +3174,81 @@ def test_preflight(self, setup_teardown: Any) -> None:
...
+            ],
+        )
+        # this forces the query to route to tier 64
+        mock_get_duration_between_marks.return_value = 2777.0
+        best_effort_response = EndpointTraceItemTable().execute(best_effort_message)
+        non_downsampled_tier_response = EndpointTraceItemTable().execute(
+            message_to_non_downsampled_tier
+        )
+
+        breakpoint()
```
volokluev, line 3241: 👀

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
onkar: > what additional exception handling are you thinking of in addition to the ones that run_query already does?

My bad. I did not see that run_query actual does it. Feel free to drop this suggestion.


> For first bullet point: if the second query fails, we should properly raise that instead of masking it by returning res. Note that the second query will not fail due to timing out due to setting timeout_overflow_mode

I see.

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3173,3 +3174,81 @@ def test_preflight(self, setup_teardown: Any) -> None:
...
+
+        # tier 1's results should be 3600, so tier 64's results should be around 3600 / 64 (give or take due to random sampling)
```
volokluev, line 3243: shouldn't extrapolation still make them equal-ish?

snuba/utils/metrics/timer.py
```diff
@@ -49,6 +49,21 @@ def get_duration_group(self) -> str:
 
         return duration_group
 
+    def get_duration_between_marks(self, start_mark: str, end_mark: str) -> float:
```
davidtsuk, line 52: what is a mark?

onewland:
Is this sampling algorithm described somewhere in the codebase not touched by this PR? I think this is important enough logic that there should be a doc comment somewhere that explains the logic of how a tier is selected.

xurui-c:
@onewland https://www.notion.so/Sampling-in-Storage-Implementation-Details-1ae8b10e4b5d808ab722ff17187a950d?pvs=4

volokluev (approved):
once the tests pass, the rest of my comments are optional

onkar (approved):

PR #6993, opened 2025-03-24
Revert "feat(eap-spans): push eap-items commits to the commit log"

onewland:
Reverts getsentry/snuba#6990

This change is good but we're not ready for it yet


volokluev (approved):

onewland:
Meant to revert an ops branch, this one is actually what we want now 🙈 . See https://github.com/getsentry/ops/pull/14575.

Issue #6962, opened 2025-03-11
snuba migrations succeed with misconfigured clusters


mwarkentin:
### Environment

itty-bitty

### Steps to Reproduce

1. Misconfigured `snuba`'s cluster configuration with mismatching cluster names
2. Ran migrations

### Expected Result

Migrations either succeed and create the table on the proper hosts, or fail if there are no hosts to run against.

### Actual Result

Migrations "succeed" however several `dist` tables are missing entirely.

### Notes

> my guess is something like when we grab the nodes for the clusters and iterate there were just no nodes? 

https://github.com/getsentry/snuba/blob/04cd24e74715ee843caca37f539e9b7027979039/snuba/migrations/operations.py#L75

https://github.com/getsentry/snuba/blob/04cd24e74715ee843caca37f539e9b7027979039/snuba/clusters/cluster.py#L424

mwarkentin:
This also breaks reverse migrations (I would've expected the initial migration(s) to fail in a similarly loud fashion, tbh):

```
snuba@snuba-api-production-7cbbb7755b-9jbsq:/usr/src/snuba$ snuba migrations reverse --group search_issues --migration-id 0009_add_message --force
2025-03-11 20:37:50,039 Initializing Snuba...
2025-03-11 20:37:51,274 Snuba initialization took 1.2353908939985558s
{"module": "snuba.migrations.runner", "event": "Reversing migration: 0009_add_message", "severity": "info", "timestamp": "2025-03-11T20:37:51.341072Z"}
{"module": "snuba.migrations.operations", "event": "Executing op: ALTER TABLE search_issues_dist_v...", "severity": "info", "timestamp": "2025-03-11T20:37:51.349652Z"}
{"module": "snuba.migrations.operations", "event": "Executing on distributed node: snuba-query-1-1:9000", "severity": "info", "timestamp": "2025-03-11T20:37:51.349832Z"}
[...]

PR #6999, opened 2025-03-25
fix(migrations): handle missing nodes

MeredithAnya:
Addresses https://github.com/getsentry/snuba/issues/6962

A migration should only be successful if it actually runs the sql operations on the nodes for the cluster - if there are no nodes, it probably means the cluster or migration itself wasn't set up right

The exception to this is when migrations are run in single node mode (which is the setup locally) - in that case it is expected to not have any nodes for the distributed target operations


volokluev (approved):

onkar (approved):

mwarkentin:
Thank you!

PR #7001, opened 2025-03-25
fix(eap): fix bug with virtual columns

davidtsuk:
Closes https://github.com/getsentry/eap-planning/issues/226

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/common.py
@@ -272,13 +272,13 @@ def apply_virtual_columns_eap_items(
             attribute_expression = attribute_key_to_expression_eap_items(
                 AttributeKey(
                     name=context.from_column_name,
-                    type=NORMALIZED_COLUMNS.get(
-                        context.from_column_name, AttributeKey.TYPE_STRING
-                    ),
+                    type=NORMALIZED_COLUMNS_EAP_ITEMS.get(
+                        context.from_column_name, [AttributeKey.TYPE_STRING]
+                    )[0],
                 )
             )
             return f.transform(
-                f.CAST(attribute_expression, "String"),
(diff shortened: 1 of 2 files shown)


onkar:
There's code duplication in the two `transform_expressions()` methods in `common.py`. Can we just make one common function and remove the duplication? Something like this:

```
def transform_expressions(expression: Expression) -> Expression:
    # Ensure the expression is of the correct type
    if not isinstance(expression, SubscriptableReference):
        return expression

    # Determine which column and mapping to use
    column_mappings = {
        "attr_str": (NORMALIZED_COLUMNS, attribute_key_to_expression),
        "attributes_string": (NORMALIZED_COLUMNS_EAP_ITEMS, attribute_key_to_expression_eap_items),
    }

    if expression.column.column_name not in column_mappings:
        return expression  # If the column doesn't match, return as is

    column_mapping, expression_func = column_mappings[expression.column.column_name]
    context = mapped_column_to_context.get(str(expression.key.value))

    if context:
        attribute_expression = expression_func(
            AttributeKey(
                name=context.from_column_name,
                type=column_mapping.get(context.from_column_name, AttributeKey.TYPE_STRING)
                if isinstance(column_mapping, dict)
                else column_mapping.get(context.from_column_name, [AttributeKey.TYPE_STRING])[0],
            )
        )

        return f.transform(
            f.CAST(f.ifNull(attribute_expression, literal("")), "String"),
            literals_array(None, [literal(k) for k in context.value_map.keys()]),
            literals_array(None, [literal(v) for v in context.value_map.values()]),
            literal(
                context.default_value if context.default_value != "" else "unknown"
            ),
            alias=context.to_column_name,
        )

    return expression
```

davidtsuk:
@onkar The code duplication there is intentional since we're planning to delete eap_spans soon and it will make it easier to just delete that chunk of code rather than trying to figure out what belongs to what.

PR #7003, opened 2025-03-26
chore(eap): use eap_items by default

davidtsuk:
Now that we've turned on eap_items for all endpoints, we can turn it on by default


wmak (approved):

onkar (approved):

volokluev (dismissed):

volokluev:
Would the time based slicing still work in this case? if this variable is set, will it still route to eap_spans for queries that are before march 12th?

davidtsuk:
> Would the time based slicing still work in this case? if this variable is set, will it still route to eap_spans for queries that are before march 12th?

you're right, this setting takes precedence over the runtime config

snuba/settings/__init__.py
```diff
@@ -218,7 +218,8 @@ class RedisClusters(TypedDict):
 BULK_CLICKHOUSE_BUFFER = 10000
 BULK_BINARY_LOAD_CHUNK = 2**22  # 4 MB
 
-USE_EAP_ITEMS_TABLE = os.environ.get("USE_EAP_ITEMS_TABLE", False)
+USE_EAP_ITEMS_TABLE = os.environ.get("USE_EAP_ITEMS_TABLE", True)
+USE_EAP_ITEMS_TABLE_START_TIMESTAMP_SECONDS = 1741762800
```
volokluev, line 222: nit: good to leave a comment for the reader telling them what this date actually is so they don't have to convert it themselves. Also good to clarify when this setting can be removed

volokluev (approved):

PR #7004, opened 2025-03-26
bug(sampling-in-storage): fix marks for timer

xurui-c:


onkar:
@xurui-c should we add a test that checks if the relevant code path is getting exercised?

snuba/utils/metrics/timer.py
```diff
@@ -59,8 +59,11 @@ def get_duration_between_marks(self, start_mark: str, end_mark: str) -> float:
                 end_mark_duration = self.__diff_ms(self.__marks[0][1], timestamp)
 
         assert (
-            start_mark_duration != -1 and end_mark_duration != -1
-        ), "Please pass in mark names that exist"
+            start_mark_duration != -1
+        ), f"Please pass in timer marks that exist: missing {start_mark}"
+        assert (
```
davidtsuk, line 64: Should this throw an error instead of asserting? I don't think we should assert something that might fail due to user input.

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3350,3 +3346,39 @@ def test_best_effort_route_to_tier_64(
                 tier=DownsampledStorageMeta.SelectedTier.SELECTED_TIER_64
             )
         )
+
+    def test_best_effort_end_to_end(self, setup_teardown: Any) -> None:
```
davidtsuk, line 3350: is this test supposed to have an assert?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3276,7 +3276,7 @@ def test_preflight(self, setup_teardown: Any) -> None:
 
     @patch.object(Timer, "get_duration_between_marks")
     def test_best_effort_route_to_tier_64(
-        self, mock_get_duration_between_marks: MagicMock
+        self, mock_get_duration_between_marks: MagicMock, setup_teardown: Any
```
kylemumma, line 3279: why did u change this test?

kylemumma, line 3279: why did you change this test?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3350,3 +3346,43 @@ def test_best_effort_route_to_tier_64(
                 tier=DownsampledStorageMeta.SelectedTier.SELECTED_TIER_64
             )
         )
+
+    def test_best_effort_end_to_end(self) -> None:
```
kylemumma, line 3350: why did you add this test?

kylemumma (commented):
Did you make a test that reproduces the bug from the ticket? and then passes after your change?

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
xurui-c: this is the test that reproduces the bug

kylemumma (approved):

PR #7005, opened 2025-03-27
feat(item-details): Add support for span data to trace-item-details

wmak:
- This updates trace-item-details so it can be used to query for span data as well
- Most of the code is just directly copied from snuba/web/rpc/v1/resolvers/R_ourlogs/resolver_trace_item_details.py


snuba/web/rpc/v1/endpoint_trace_item_details.py
```diff
@@ -1,14 +1,186 @@
...
+    is_logs = request.meta.trace_item_type == TraceItemType.TRACE_ITEM_TYPE_LOG
+
+    return SnubaRequest(
+        id=uuid.UUID(request.meta.request_id),
+        original_body=MessageToDict(request),
+        query=_build_query(request),
+        query_settings=query_settings,
+        attribution_info=AttributionInfo(
+            referrer=request.meta.referrer,
+            team="ourlogs" if is_logs else "eap",
```
wmak, line 121: I'm guessing what to put here, not sure what's the expected values

colin-sentry: you can just use "eap" everywhere here

wmak: Even for logs data?

wmak: Also what about `parent_api`

snuba/web/rpc/v1/endpoint_trace_item_details.py
```diff
@@ -1,14 +1,186 @@
...
+            feature="ourlogs" if is_logs else "eap",
+            tenant_ids={
+                "organization_id": request.meta.organization_id,
+                "referrer": request.meta.referrer,
+            },
+            app_id=AppID("eap"),
+            parent_api="ourlog_trace_item_table" if is_logs else "eap_trace_item_table",
```
colin-sentry, line 128: same here

snuba/web/rpc/v1/endpoint_trace_item_details.py
```diff
@@ -22,13 +194,6 @@ def version(cls) -> str:
     def request_class(cls) -> Type[TraceItemDetailsRequest]:
         return TraceItemDetailsRequest
 
-    def get_resolver(
```
colin-sentry, line 25: delete the other stuff too - all references to ResolverTraceItemDetails

colin-sentry (changes requested):

snuba/web/rpc/v1/endpoint_trace_item_details.py
colin-sentry: also add a test that it works for logs

colin-sentry (approved):

snuba/web/rpc/v1/endpoint_trace_item_details.py
```diff
@@ -1,14 +1,185 @@
...
+                ),
+            ),
+            SelectedExpression(
+                "attributes_float",
+                f.mapConcat(
+                    *[column(f"attributes_float_{n}") for n in range(_BUCKET_COUNT)],
+                    alias="attributes_float",
+                ),
+            ),
+            SelectedExpression(
```
davidtsuk, line 83: this will result in duplicates since we double write ints and bools to attributes_float

snuba/web/rpc/v1/endpoint_trace_item_details.py
```diff
@@ -1,14 +1,185 @@
...
+)
+from snuba.web.rpc.common.debug_info import (
+    extract_response_meta,
+    setup_trace_query_settings,
+)
+from snuba.web.rpc.common.exceptions import (
+    BadSnubaRPCRequestException,
+    RPCRequestException,
+)
+from snuba.web.rpc.v1.resolvers.R_ourlogs.common.attribute_key_to_expression import (
```
davidtsuk, line 42: this function might cause issues, the one we use for eap_items is in R_eap_spans.common

snuba/web/rpc/v1/endpoint_trace_item_details.py
colin-sentry: I think it's fine to have duplicates, users could explicitly send a float and a bool with the same name, and if we removed the duplicate it would be confusing for them.

snuba/web/rpc/v1/endpoint_trace_item_details.py
davidtsuk: we are planning to merge them into one, but I think it's fine to switch between the two dynamically for now

PR #7009, opened 2025-03-28
feat: modify the behavior of equality functions on null values

kylemumma:
this addresses https://github.com/getsentry/eap-planning/issues/200

We redefine the default behavior of some functions in the case of the column being null:
* equals
old: `column = value`
new: `column = value OR (isNull(column) and isNull(value))`
explanation: since rhs can be null or non-null we have`col=null`,`col=7` where 7 represents a non-null value. Consider the case where `col` is null, 
`col=null <-> null=null` evaluates to false
`col=7 <-> null=7` evaluates to false
so only the behavior of the first one have to change.  `col=v` is now redefined as `col=v OR isNull(col) and isNull(v)`. similar logic was used to determine the modifications to the other functions so I will leave it out.
* not equals
old: `column != value`
new `column != value OR (isNull(column) and not isNull(val))`

### DIFF
--- a/snuba/web/rpc/common/common.py
+++ b/snuba/web/rpc/common/common.py
@@ -217,18 +217,28 @@ def trace_item_filters_to_expression(
 
         if op == ComparisonFilter.OP_EQUALS:
             _check_non_string_values_cannot_ignore_case(item_filter.comparison_filter)
-            return (
+            expr = (
                 f.equals(f.lower(k_expression), f.lower(v_expression))
                 if item_filter.comparison_filter.ignore_case
                 else f.equals(k_expression, v_expression)
             )
+            # we redefine the way equals works for nulls to be more intuitive
+            expr_with_null = or_cond(
+                expr, and_cond(f.isNull(k_expression), f.isNull(v_expression))
+            )
+            return expr_with_null
         if op == ComparisonFilter.OP_NOT_EQUALS:
             _check_non_string_values_cannot_ignore_case(item_filter.comparison_filter)
-            return (
+            expr = (
                 f.notEquals(f.lower(k_expression), f.lower(v_expression))
                 if item_filter.comparison_filter.ignore_case
                 else f.notEquals(k_expression, v_expression)
             )
+            # we redefine the way not equals works for nulls to be more intuitive
+            expr_with_null = or_cond(
+                expr, and_cond(f.isNull(k_expression), not_cond(f.isNull(v_expression)))
+            )
+            return expr_with_null
         if op == ComparisonFilter.OP_LIKE:
             if k.type != AttributeKey.Type.TYPE_STRING:
                 raise BadSnubaRPCRequestException(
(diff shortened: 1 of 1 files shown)


snuba/web/rpc/common/common.py
```diff
@@ -215,20 +216,39 @@ def trace_item_filters_to_expression(
                         f"translation of AttributeValue type {default} is not implemented"
                     )
 
+        set_config("use_new_null_comparison", 1)
+        use_new_null_comparison = (bool)(
+            get_int_config("use_new_null_comparison", default=0)
+        )
```
onkar: Question: do we usually use feature flags in such situations? I am wondering what would break if we do not use this feature flag? We could simplify the `if use_new_null_comparison` branching if this feature flag is not required.

tests/web/rpc/v1/test_R_ourlogs/test_logs_expression_converters.py
```diff
@@ -78,55 +80,65 @@ def test_attr_buckets(self) -> None:
                 (column(col), literal("z")),
             )
 
+    @pytest.mark.redis_db
     def test_trace_item_filters_to_expression(self) -> None:
```
kylemumma, line 84: this test was explicitly testing how filters are translated. since the translation was updated, I had to update this test.

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3386,3 +3387,56 @@ def test_best_effort_end_to_end(self) -> None:
             response.meta.downsampled_storage_meta.tier
             != DownsampledStorageMeta.SELECTED_TIER_UNSPECIFIED
         )
+
+    @pytest.mark.redis_db
+    def test_non_existant_attribute_filter(self) -> None:
```
kylemumma, line 3392: this one tests the new behavior

snuba/web/rpc/common/common.py
kylemumma: The reason I am choosing to use a feature flag is because this change modifies the filter translation logic for all queries to all endpoints. Meaning if there is a bug it has the potential to break a lot of queries which I think is high risk. Using a feature flag makes the change more safe and logic will be cleaned up later once the feature flag is removed.

Going forward I would prefer if you wait until I mark the PR ready for review before reviewing. Typically if I have it in draft mode its not ready for review.

snuba/web/rpc/common/common.py
```diff
@@ -215,20 +216,38 @@ def trace_item_filters_to_expression(
...
+
         if op == ComparisonFilter.OP_EQUALS:
             _check_non_string_values_cannot_ignore_case(item_filter.comparison_filter)
-            return (
+            expr = (
                 f.equals(f.lower(k_expression), f.lower(v_expression))
                 if item_filter.comparison_filter.ignore_case
                 else f.equals(k_expression, v_expression)
             )
+            # we redefine the way equals works for nulls to be more intuitive
```
volokluev, line 230: This comment is not helpful. In what way does equals work for nulls?

kylemumma: updated comment

onkar (approved):

PR #7010, opened 2025-03-28
hotfix(inc-1098): Increase use 'break' overflow mode for replays

volokluev:
The `group_by_overflow_mode=any` setting is causing the cluster to segfault (oopsie!)

As a bandaid, use `break`


cmanallen:
@volokluev ClickHouse bug?  We can use a different overflow mode (if that helps).

onewland (approved):

MeredithAnya (approved):

onkar (approved):

PR #7011, opened 2025-03-31
perf(sampling-in-storage): add perf metrics

xurui-c:
record the following things for each query:
1. The time that it took to run the query on the most downsampled tier
2. difference between estimated time and actual time it took to run the query
3. what tier the query was run on

exceptions:
preflight mode queries will not (2) or (3)recorded

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -63,6 +63,11 @@ def _get_target_tier(
             most_downsampled_query_duration_ms
             * cast(int, DOWNSAMPLING_TIER_MULTIPLIERS.get(tier))
         )
+        metrics_backend.timing(
+            "sampling_in_storage_estimated_query_duration",
+            estimated_query_duration_to_this_tier,
+            tags={"referrer": referrer, "tier": str(tier)},
+        )
         if (
             estimated_query_duration_to_this_tier
             <= _get_time_budget() - most_downsampled_query_duration_ms
@@ -83,6 +88,16 @@ def _is_best_effort_mode(in_msg: T) -> bool:
         in_msg.meta.HasField("downsampled_storage_config")
         and in_msg.meta.downsampled_storage_config.mode
         == DownsampledStorageConfig.MODE_BEST_EFFORT
+    )
+
+
+def _record_actual_query_duration(
+    metrics_backend: MetricsBackend, timer: Timer, tags: Dict[str, str]
+) -> None:
+    metrics_backend.timing(
+        "sampling_in_storage_actual_query_duration",
+        _get_query_duration(timer),
+        tags=tags,
     )
 
 
@@ -167,9 +182,9 @@ def run_query_to_correct_tier(
         )
 
         if target_tier == _get_most_downsampled_tier():
-            metrics_backend.timing(
-                "sampling_in_storage_difference_between_estimated_and_true_query_duration",
-                0,
+            _record_actual_query_duration(
+                metrics_backend,
+                timer,
                 tags={"referrer": referrer, "tier": str(target_tier)},
             )
             return res
@@ -185,8 +200,13 @@ def run_query_to_correct_tier(
             request=request_to_target_tier,
             timer=timer,
         )
+        _record_actual_query_duration(
+            metrics_backend,
+            timer,
+            tags={"referrer": referrer, "tier": str(target_tier)},
+        )
         metrics_backend.timing(
-            "sampling_in_storage_difference_between_estimated_and_true_query_duration",
+            "sampling_in_storage_estimation_error",
             estimated_target_tier_query_duration - _get_query_duration(timer),
             tags={"referrer": referrer, "tier": str(target_tier)},
         )
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -1,5 +1,5 @@
 import uuid
-from typing import Any, Callable, Dict, TypeVar, cast
+from typing import Callable, Dict, TypeVar, cast
 
 from google.protobuf.json_format import MessageToDict
 from sentry_protos.snuba.v1.downsampled_storage_pb2 import DownsampledStorageConfig
@@ -61,9 +61,9 @@ def _get_target_tier(
             * cast(int, DOWNSAMPLING_TIER_MULTIPLIERS.get(tier))
         )
         metrics_backend.timing(
-            f"sampling_in_storage_estimated_query_duration_to_TIER_{tier}",
+            "sampling_in_storage_estimated_query_duration",
             estimated_query_duration_to_this_tier,
-            tags={"referrer": referrer},
+            tags={"referrer": referrer, "tier": str(tier)},
         )
         if (
             estimated_query_duration_to_this_tier
@@ -71,8 +71,10 @@ def _get_target_tier(
         ):
             target_tier = tier
 
-    metrics_backend.gauge(
-        "sampling_in_storage_routed_tier", target_tier, tags={"referrer": referrer}
+    metrics_backend.timing(
+        "sampling_in_storage_routed_tier",
+        target_tier,
+        tags={"referrer": referrer, "tier": str(target_tier)},
     )
     return target_tier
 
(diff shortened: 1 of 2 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -54,12 +61,20 @@ def _get_target_tier(timer: Timer) -> Tier:
...
+            estimated_query_duration_to_this_tier,
+            tags={"referrer": referrer},
+        )
         if (
             estimated_query_duration_to_this_tier
             <= _get_time_budget() - most_downsampled_query_duration_ms
         ):
             target_tier = tier
 
+    metrics_backend.gauge(
```
xurui-c, line 75: is gauge the right metric for this?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -94,33 +113,34 @@ def build_snuba_request(
...
-            timer.get_duration_between_marks(
-                start_mark,
-                end_mark,
-            ),
-        )
+    metrics_backend.timing(
+        "sampling_in_storage_query_duration_from_most_downsampled_tier",
+        _get_query_duration(timer),
+        tags={"referrer": referrer},
+    )
```
xurui-c, line 130: The time that it took to run the query on the most downsampled tier

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -45,21 +50,35 @@ def _get_query_duration(timer: Timer) -> float:
...
     for tier in sorted(Tier, reverse=True)[:-1]:
         estimated_query_duration_to_this_tier = (
             most_downsampled_query_duration_ms
             * cast(int, DOWNSAMPLING_TIER_MULTIPLIERS.get(tier))
         )
+        metrics_backend.timing(
+            f"sampling_in_storage_estimated_query_duration_to_TIER_{tier}",
+            estimated_query_duration_to_this_tier,
+            tags={"referrer": referrer},
+        )
```
xurui-c, line 72: What the predicted time to run is on every tier

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -45,21 +50,35 @@ def _get_query_duration(timer: Timer) -> float:
...
         if (
             estimated_query_duration_to_this_tier
             <= _get_time_budget() - most_downsampled_query_duration_ms
         ):
             target_tier = tier
 
+    metrics_backend.gauge(
+        "sampling_in_storage_routed_tier", target_tier, tags={"referrer": referrer}
```
xurui-c, line 80: what tier the query was run on

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -94,33 +113,34 @@ def build_snuba_request(
...
 
 
+def _record_actual_query_duration(
+    metrics_backend: MetricsBackend, timer: Timer, tags: Dict[str, Any]
+) -> None:
+    metrics_backend.timing(
+        "sampling_in_storage_actual_query_duration",
+        _get_query_duration(timer),
+        tags=tags,
+    )
```
xurui-c, line 141: the actual time it took to run the query

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
volokluev: I would use a `timing` metric so that you can have distributions

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -31,6 +31,11 @@
 }
 
 
+def _get_referrer(in_msg: T) -> str:
+    # if the referrer isn't set, this would return ""
+    return in_msg.meta.referrer
```
volokluev, line 36: why does this need to be its own function

xurui-c: it doesnt

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -157,9 +173,14 @@ def run_query_to_correct_tier(
...
-        target_tier = _get_target_tier(timer)
-
-        if target_tier == Tier.TIER_512:
+        target_tier = _get_target_tier(timer, metrics_backend, referrer)
+
+        if target_tier == _get_most_downsampled_tier():
+            _record_actual_query_duration(
+                metrics_backend,
+                timer,
+                tags={"referrer": referrer, "tier": str(target_tier)},
```
xurui-c, line 182: this is the same as `sampling_in_storage_query_duration_from_most_downsampled_tier`, I emitted it again to be consistent (`sampling_in_storage_actual_query_duration` isn't just missing a tier in its tags)

xurui-c: but if we don't see it as necessary i can just remove it

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
volokluev: This is not the most important metric to capture. What is important is that for each model estimation, you record the *delta* between the estimation and the actual time it took

volokluev: The expression you're proposing will be measuring a different thing than what I am proposing. Can you figure out why? It's a pretty important distinction

xurui-c: bc the one i proposed isn't specific to a single query

volokluev: correct, and do you understand why that's important?

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -157,9 +162,16 @@ def run_query_to_correct_tier(
...
-        target_tier = _get_target_tier(timer)
+        target_tier, estimated_target_tier_query_duration = _get_target_tier(
+            timer, metrics_backend, referrer
+        )
 
-        if target_tier == Tier.TIER_512:
+        if target_tier == _get_most_downsampled_tier():
+            metrics_backend.timing(
+                "sampling_in_storage_difference_between_estimated_and_true_query_duration",
+                0,
```
volokluev, line 172: 0 ?

volokluev: I would say in this case don't record it at all, it'll just mess up the reporting. in this case the estimate was not used

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -173,5 +185,10 @@ def run_query_to_correct_tier(
             request=request_to_target_tier,
             timer=timer,
         )
+        metrics_backend.timing(
+            "sampling_in_storage_difference_between_estimated_and_true_query_duration",
```
volokluev, line 189: bit long of a name, `sampling_in_storage_estimation_delta`

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -45,22 +45,42 @@ def _get_query_duration(timer: Timer) -> float:
...
             <= _get_time_budget() - most_downsampled_query_duration_ms
         ):
             target_tier = tier
+            estimated_target_tier_query_duration = estimated_query_duration_to_this_tier
 
-    return target_tier
+    metrics_backend.timing(
+        "sampling_in_storage_routed_tier",
+        target_tier,
+        tags={"referrer": referrer, "tier": str(target_tier)},
```
volokluev, line 81: this tier tag makes no sense, it's the value and the tag?

volokluev (approved):

PR #7012, opened 2025-04-01
fix(eap): Floating point error when calculating duration

Zylphrex:
Doing the subtraction first here is prone to floating point errors. So perform the multiplication first followed by the subtraction after it's been converted to an u64 to minimize the impact.


phacops (changes requested):
You likely want to change this in `eap_items` too since that's the consumer inserting spans now.

PR #7013, opened 2025-04-01
fix(eap): Exclude standalone spans from the GetTraces results

phacops:
Standalone spans are not yet visible in the trace view. Since the GetTraces endpoint will still consider them as part of the results, we might end up in a situation where we list a trace with only a standalone span and link to the trace view which would be empty.

This will exclude those standalone spans from the results in GetTraces, preventing us from linking to empty trace views.

### DIFF
--- a/rust_snuba/.cargo/config.toml
+++ b/rust_snuba/.cargo/config.toml
@@ -1,0 +1,3 @@
+[env]
+# Workaround for https://github.com/confluentinc/librdkafka/pull/5012
+CMAKE_POLICY_VERSION_MINIMUM = "3.10"


snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -475,6 +487,17 @@ def _list_trace_ids(
...
+                    f.equals(
+                        generate_expression_from_attribute_key(
+                            AttributeKey(
+                                name="sentry.segment_id",
+                                type=AttributeKey.Type.TYPE_STRING,
+                            ),
+                        ),
+                        literal("00"),
+                    ),
+                ),
```
volokluev, line 500: leave a comment explaining what this does plz

volokluev (approved):

davidtsuk:
are we going to remove this once standalone spans are visible in the trace view?

phacops:
> are we going to remove this once standalone spans are visible in the trace view?

Yes, we'll remove this once we have support for this in the trace view.

mcannizz:
What is a standalone span?

phacops:
> What is a standalone span?

It's a span not attached to a segment, used for reporting some web vitals from the SDKs. It was made this way since we were on AM2 at the time, wanted to report those metrics but not to count as a transaction, so this was sent as a standalone span.

Since then, a lot of things happened, between AM3 and us trying to support spans and metrics in a different way, we'd likely not make the same choices, but we need to handle those until we clean up the situation.

PR #7016, opened 2025-04-01
ci: workaround for rdkafka build error

davidtsuk:
Trying the fix from https://github.com/getsentry/relay/pull/4632


davidtsuk:
looks like it doesn't work

PR #7017, opened 2025-04-01
ci: workaround for cmake issue

davidtsuk:
inspired by https://github.com/getsentry/relay/pull/4632


phacops (approved):

onkar (approved):

phacops:
I think I'd prefer to use the same fix as Relay and create a `rust_snuba/.cargo/config.toml` file with the same config.

PR #7019, opened 2025-04-01
feat(sampling-in-storage): remove dependency on Timer

xurui-c:
https://github.com/getsentry/eap-planning/issues/237

Reason for this bug: if the query gets cached and the cached result is fetched, then the timer never got the "right_before_execute" and "execute" marks.

Fix: just use the elapsed field: https://github.com/mymarilyn/clickhouse-driver/blob/8a4e7c5b99b532df2b015651d893a6f36288a22c/clickhouse_driver/result.py#L143

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -42,7 +42,7 @@ def _get_time_budget() -> float:
 
 
 def _get_query_duration_ms(res: QueryResult) -> float:
-    return cast(float, res.result.get("profile", {}).get("elapsed", 0) * 1000)
+    return cast(float, res.result.get("profile", {}).get("elapsed", 0) * 1000)  # type: ignore
 
 
 def _get_most_downsampled_tier() -> Tier:
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -41,8 +41,8 @@ def _get_time_budget() -> float:
     return sentry_timeout_ms - error_budget_ms
 
 
-def _get_query_duration(timer: Timer) -> float:
-    return timer.get_duration_between_marks("right_before_execute", "execute")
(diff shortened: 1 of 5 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -41,18 +41,18 @@ def _get_time_budget() -> float:
     return sentry_timeout_ms - error_budget_ms
 
 
-def _get_query_duration(timer: Timer) -> float:
-    return timer.get_duration_between_marks("right_before_execute", "execute")
+def _get_query_duration_ms(res: QueryResult) -> float:
+    return cast(float, res.result.get("profile", {}).get("elapsed", 0) * 1000)  # type: ignore
```
volokluev, line 45: if the cache is hit and there is no time spent on clickhouse, how useful is the elapsed time for deciding query routing?

xurui-c: I think if the cache is hit, then the cache would return the result of the previous same query, which has the correct elapsed time: https://github.com/getsentry/snuba/blob/6f03d26b17c14e488b8e2ff89f789ed5fbd6b816/snuba/state/cache/redis/backend.py#L66-L81

volokluev: Did you verify this or are you assuming?

xurui-c: i saw that the elapsed time is the same on two calls (manual inspection) and I wrote a test test_cached_query_results_is_used_in_routing

tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3527,3 +3529,56 @@ def test_non_existant_attribute_filter(self) -> None:
                 ],
             ),
         ]
+
+    def test_cached_query_results_is_used_in_routing(self) -> None:
```
volokluev, line 3533: I don't think you need to keep this test in here. Thank you for doing it to make sure that the caching keeps the `elapsed` field intact. 

I would recommend removing this test because it's too dependent on internal details of the endpoint

PR #7023, opened 2025-04-02
fix(eap): do not count invalid requests against the SLO

volokluev:


kylemumma:
Do we still wanna capture it in sentry when its a `BadSnubaRPCRequestException`?

volokluev:
> Do we still wanna capture it in sentry when its a BadSnubaRPCRequestException?

I think so, we can exclude it from alerting if we want but it's useful info

onkar (approved):

PR #7027, opened 2025-04-02
fix(sampling-in-storage): make tests not flakey

xurui-c:
Bounds are too tight, let's loosen them. As long as the count is reasonably away from 3600/512 (Tier 512) and 3600/8 (Tier 8) we're good

### DIFF
--- a/tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
+++ b/tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
@@ -3373,9 +3373,9 @@ class TestTraceItemTableEAPItems(TestTraceItemTable):
 
             # tier 1's results should be 3600, so tier 64's results should be around 3600 / 64 (give or take due to random sampling)
             assert (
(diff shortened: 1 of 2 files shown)


tests/web/rpc/v1/test_endpoint_trace_item_table/test_endpoint_trace_item_table.py
```diff
@@ -3373,9 +3373,9 @@ def test_best_effort_route_to_tier_64(self) -> None:
 
             # tier 1's results should be 3600, so tier 64's results should be around 3600 / 64 (give or take due to random sampling)
             assert (
-                len(non_downsampled_tier_response.column_values[0].results) / 90
+                len(non_downsampled_tier_response.column_values[0].results) / 200
                 <= len(best_effort_response.column_values[0].results)
-                <= len(non_downsampled_tier_response.column_values[0].results) / 40
+                <= len(non_downsampled_tier_response.column_values[0].results) / 16
```
volokluev, line 3378: the comment and the above numbers don't match, you should explain why

volokluev (approved):

onkar (approved):

PR #7024, opened 2025-04-02
chore: combine spans and logs resolver into eap_items resolver for trace_item_table

kylemumma:
the easiest entrypoint to reviewing this PR would probably be `R_eap_spans/resolver_trace_item_table.py::resolve` and `R_ourlogs/resolver_trace_item_table.py::resolve`

# Major Changes
* copy/paste the eap_spans resolver into a new class called eap_items resolver
* eap_spans now uses eap_items resolver for its logic
* extend the new eap_items resolver to support what logs needs
* switch logs to use the new eap_items resolver as well

The major change, switch the logs resolver is behind a feature flag, so if anything breaks it will go back to the old one. I ran CI with feature flag forced off, and on, to make sure it passes with both.

There are a few changes to eap_spans resolver, which arent behind feature flag but they are pretty minor so i think its fine to not put them behind a flag.

There is a case with `attribute_to_expression` where `sentry.timestamp` used to have the alias `sentry.timestamp_TYPE_STRING` in the old logs but now will have alias `sentry.timestamp` but since integration tests pass I will assume its ok.

PR #7028, opened 2025-04-02
Filter spans in TraceItemTable and TimeSeries endpoints

volokluev:
Repeat #7026 but this time the sentry tests should pass


xurui-c (approved):

kylemumma:
wont this cause any query on these endpoints that goes to `eap_spans` to fail since it doesnt have `item_type` column? could you just put `f.equals(snuba_column("item_type"), request.meta.trace_item_type)` behind `if use_eap_items_table`?

https://github.com/getsentry/snuba/pull/7024/files#diff-e1e06d7f875a7c2870cc11bc4301dd6ab9fba73263c76260452c0b3176f66110R285-R296

volokluev:
yeah good point kyle

phacops (approved):

PR #7032, opened 2025-04-03
chore: retire feature flag use_new_null_comparison 

kylemumma:
I will wait to merge this until monday. this has been enabled in production for a while now so im getting rid of the feature flag.


volokluev (approved):
I don't think there's a need to wait to merge this

PR #7033, opened 2025-04-03
Handle uint overflow for segment id in replays processor

volokluev:
This is causing too much noise on replays consumer.

### DIFF
--- a/rust_snuba/src/processors/replays.rs
+++ b/rust_snuba/src/processors/replays.rs
@@ -155,7 +155,6 @@ pub fn deserialize_message(
                 retention_days: replay_message.retention_days,
                 sdk_name: event.sdk.name.unwrap_or_default(),
                 sdk_version: event.sdk.version.unwrap_or_default(),
-                segment_id: segment_id,
                 timestamp: event.timestamp as u32,
                 trace_ids: event.trace_ids.unwrap_or_default(),
                 urls: event.urls.unwrap_or_default(),
@@ -163,6 +162,7 @@ pub fn deserialize_message(
                 user_email: event.user.email.unwrap_or_default(),
                 user_id: user_id.unwrap_or_default(),
                 user_name: event.user.username.unwrap_or_default(),
+                segment_id,
                 title,
                 tags_key,
                 tags_value: tags_value
--- a/rust_snuba/src/processors/replays.rs
+++ b/rust_snuba/src/processors/replays.rs
@@ -115,6 +115,17 @@ pub fn deserialize_message(
             let error_sample_rate = event.contexts.replay.error_sample_rate.unwrap_or(-1.0);
             let session_sample_rate = event.contexts.replay.session_sample_rate.unwrap_or(-1.0);
 
+            let segment_id = match event.segment_id {
+                Some(s_id) => {
+                    if s_id <= u16::MAX.into() {
+                        Some(s_id as u16)
+                    } else {
+                        None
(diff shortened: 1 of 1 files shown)


MeredithAnya (approved):

rust_snuba/src/processors/replays.rs
```diff
@@ -144,14 +155,14 @@ pub fn deserialize_message(
...
                 sdk_version: event.sdk.version.unwrap_or_default(),
-                segment_id: event.segment_id,
                 timestamp: event.timestamp as u32,
                 trace_ids: event.trace_ids.unwrap_or_default(),
                 urls: event.urls.unwrap_or_default(),
                 user,
                 user_email: event.user.email.unwrap_or_default(),
                 user_id: user_id.unwrap_or_default(),
                 user_name: event.user.username.unwrap_or_default(),
+                segment_id,
```
phacops, line 165: There's always a Rust 1-liner in case you need it.
```suggestion
+                segment_id: event.segment_id.filter(|s| *s <= (u16::MAX as u32)).map_or(None, |s| Some(s as u16)),
```

phacops (approved):

rust_snuba/src/processors/replays.rs
```diff
@@ -115,6 +115,17 @@ pub fn deserialize_message(
...
+            let segment_id = match event.segment_id {
+                Some(s_id) => {
+                    if s_id <= u16::MAX.into() {
+                        Some(s_id as u16)
+                    } else {
+                        None
+                    }
+                }
+                None => None,
+            };
```
onkar, line 127:
```
let segment_id = event.segment_id.and_then(|s_id| {
    if s_id <= u16::MAX as u32 {
        Some(s_id as u16)
    } else {
        None
    }
});
```

The` into()` conversion is unnecessary because `u16::MAX` is already a `u16`. `and_then` simplifies the option handling.

rust_snuba/src/processors/replays.rs
```diff
@@ -469,6 +480,77 @@ mod tests {
...
+                "segment_id": null,
+                "start_time": 100,
+                "type": "replay_event"
+            }}"#
+        );
+
+        let (rows, _) = deserialize_message(data.as_bytes(), 0, 0).unwrap();
+        let replay_row = rows.first().unwrap();
+        assert_eq!(replay_row.segment_id, None);
+    }
```
onkar, line 552:
```
let valid_segment_id = u16::MAX as u32;  // Maximum valid value
    let payload_valid = payload.replace("4206942069", &valid_segment_id.to_string());
    let data_valid = format!(
        r#"{{
            "payload": {payload_valid:?},
            "project_id": 1,
            "replay_id": "048aa04be40243948eb3b57089c519ee",
            "retention_days": 30,
            "segment_id": null,
            "start_time": 100,
            "type": "replay_event"
        }}"#
    );

    let (rows_valid, _) = deserialize_message(data_valid.as_bytes(), 0, 0).unwrap();
    let replay_row_valid = rows_valid.first().unwrap();

    // segment_id should be correctly parsed when it's within u16 range
    assert_eq!(
        replay_row_valid.segment_id,
        Some(u16::MAX),
        "segment_id should be stored when within u16 range"
    );
```

nit: Consider adding a test when the segment id is set to the maximum value.

onkar (approved):
LGTM, suggested a couple of things.

PR #7040, opened 2025-04-07
fix(eap): Use trace_id field in TraceItemDetails

phacops:
Use the now required `trace_id` field in this request so the `trace_id` is correctly promoted to the `PREWHERE` clause.
```
SELECT (toUnixTimestamp(timestamp) AS timestamp), (leftPad(lower(hex(item_id)), 32, '0') AS hex_item_id), (replaceAll(toString(trace_id), '-', '') AS trace_id), organization_id, project_id, item_type, (mapConcat(attributes_string_0, ...) AS attributes_string), (mapConcat(attributes_float_0, ...) AS attributes_float), attributes_int, attributes_bool FROM eap_items_1_local PREWHERE equals(trace_id, 'f16dd70d-80d4-47f5-ae7d-6e2048f9ea4d') WHERE in(project_id, [1]) AND equals(organization_id, 1) AND less(timestamp, toDateTime(1744054948)) AND greaterOrEquals(timestamp, toDateTime(1744041600)) AND equals(item_type, 1) AND equals(item_id, '276171113345555592925493239168611803141') AND true LIMIT 1 OFFSET 0
```

### DIFF
--- a/snuba/web/rpc/v1/endpoint_trace_item_details.py
+++ b/snuba/web/rpc/v1/endpoint_trace_item_details.py
@@ -216,6 +216,22 @@ class EndpointTraceItemDetails(
             raise BadSnubaRPCRequestException(
                 "This endpoint requires meta.trace_item_type to be set (are you requesting spans? logs?)"
             )
+        if in_msg.item_id == "":
+            raise BadSnubaRPCRequestException(
+                "This endpoint requires item_id to be set."
+            )
+        if in_msg.trace_id == "":
+            raise BadSnubaRPCRequestException(
+                "This endpoint requires trace_id to be set."
+            )
+        else:
+            try:
+                _ = uuid.UUID(in_msg.trace_id)
+            except ValueError:
+                raise BadSnubaRPCRequestException(
+                    "This endpoint requires trace_id to be a valid UUID."
+                )
+
         snuba_request = _build_snuba_request(in_msg)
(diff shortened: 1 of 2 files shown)


wmak (approved):

davidtsuk (approved):

snuba/web/rpc/v1/endpoint_trace_item_details.py
```diff
@@ -101,6 +101,10 @@ def _build_query(request: TraceItemDetailsRequest) -> Query:
                 column("item_id"),
                 literal(request.item_id),
             ),
+            f.equals(
+                column("trace_id"),
+                literal(request.trace_id),
```
colin-sentry, line 106: `literal(str(uuid.UUID(request.trace_id))),`

also [merge this first
](https://github.com/getsentry/sentry/pull/88956)

colin-sentry: it doesn't actually improve performance in its current form because of an unrelated issue

volokluev: protobuf fields are all optional. you should only add this clause if that field is actually set

volokluev: that's fine too

volokluev (approved):

PR #7042, opened 2025-04-07
feat: remove logs resolver, endpoint itemstats

kylemumma:
we create a unified eap_items resolver that is now used for both spans and logs.

# major changes
* extracted eap_spans resolver into eap_items resolver
* made eap_spans use this
* also made logs use this
* added a test to make sure the endpoint works for logs now


tests/web/rpc/v1/test_endpoint_trace_item_stats_logs.py
```diff
@@ -0,0 +1,75 @@
...
+            ),
+            stats_types=[
+                StatsType(
+                    attribute_distributions=AttributeDistributionsRequest(
+                        max_buckets=100, max_attributes=1000
+                    )
+                )
+            ],
+        )
+        EndpointTraceItemStats().execute(message)
```
mcannizz, line 75: This test would be more useful if it made some assertions about the response.

volokluev (approved):
Ditto to what Matt said but approving given you do that

tests/web/rpc/v1/test_endpoint_trace_item_stats_logs.py
kylemumma: added assertions about the response

PR #7043, opened 2025-04-07
feat(sourcemaps) - Add `symbolicated_in_app` column on errors

yuvmen:
As part of the  [Sourcemapped Frames Search project](https://www.notion.so/sentry/Sourcemapped-Frames-Search-Tech-Spec-15f8b10e4b5d803fa561cbeb5d2691a3) added a new column on errors indicating if all in_app
 frames are symbolicated for an error.
This flag now arrives as an optional field on events from Sentry and so we simply populate it from the event.

Moved the migration to a PR which will be merged before this one: #7044


rust_snuba/Cargo.toml
```diff
@@ -32,7 +32,7 @@ procspawn = { version = "1.0.0", features = ["json"] }
 pyo3 = { version = "0.18.1", features = ["chrono"] }
 reqwest = { version = "0.11.11", features = ["stream"] }
 sentry = { version = "0.32.0", features = ["anyhow", "tracing"] }
-sentry-kafka-schemas = "1.1.0"
+sentry-kafka-schemas = "1.1.6"
```
volokluev, line 35: don't ship the migration and the processor change in the same PR. migration first, then processor

yuvmen: oh I see, but that means basically moving the migration file and `errors.yaml` change to a separate PR and push that in first right? What you are commenting on here is the Kafka schema bump and that I can do along with the change in `errors.rs` right?

snuba/snuba_migrations/events/0026_add_symbolicated_in_app_column.py
```diff
@@ -0,0 +1,77 @@
...
+        return [
+            AddColumn(
+                storage_set=StorageSetKey.EVENTS,
+                table_name="errors_local",
+                column=Column(
+                    "symbolicated_in_app",
+                    schemas.UInt(
+                        8,
+                        modifiers=MigrationModifiers(
+                            nullable=True,
```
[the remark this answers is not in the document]
yuvmen: I actually want this field to be nullable, it will often be null (basically except for javascript events at the moment)

PR #7049, opened 2025-04-08
fix(sampling-in-storage): fix increment metric

xurui-c:
Right now all the span attributes is saying target tier is 1

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -143,13 +143,12 @@ def _get_target_tier(
                     bytes_scanned_limit,
                 )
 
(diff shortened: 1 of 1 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -143,13 +143,12 @@ def _get_target_tier(
...
                 )
 
-        _record_value_in_span_and_DD(
-            span,
-            metrics_backend.increment,
-            "target_tier",
+        metrics_backend.increment(
+            _SAMPLING_IN_STORAGE_PREFIX + "target_tier",
             1,
             {"referrer": referrer, "tier": str(tier)},
```
mcannizz, line 149: What is the value of `tier` here? How is it different from `target_tier`?

volokluev (approved):

PR #7050, opened 2025-04-08
fix(sampling-in-storage): fix increment metric

xurui-c:
to figure out: are we over or under estimating? 

my understandig is that gauge is the authoritative value


mcannizz:
What's the difference between this PR and https://github.com/getsentry/snuba/pull/7049?

xurui-c:
@mcannizz https://github.com/getsentry/snuba/pull/7051

PR #7051, opened 2025-04-08
perf(sampling-in-storage): emit raw estimation error

xurui-c:
to figure out: are we over or under estimating?

my understandig is that gauge is the authoritative value

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -296,15 +296,23 @@ def run_query_to_correct_tier(
                     request=request_to_target_tier,
                     timer=timer,
                 )
+
+                estimation_error = (
+                    estimated_target_tier_query_bytes_scanned
+                    - _get_query_bytes_scanned(res)
+                )
                 _record_value_in_span_and_DD(
                     span,
                     metrics_backend.distribution,
                     "estimation_error_percentage",
-                    abs(
-                        estimated_target_tier_query_bytes_scanned
-                        - _get_query_bytes_scanned(res)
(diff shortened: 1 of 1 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -296,15 +296,23 @@ def run_query_to_correct_tier(
...
-                        estimated_target_tier_query_bytes_scanned
-                        - _get_query_bytes_scanned(res)
-                    )
-                    / _get_query_bytes_scanned(res),
+                    abs(estimation_error) / _get_query_bytes_scanned(res),
+                    {"referrer": referrer, "tier": str(target_tier)},
+                )
+                _record_value_in_span_and_DD(
+                    span,
+                    metrics_backend.gauge,
```
mcannizz, line 313: Why is gauge the right metric type for this?

xurui-c: I want to emit the authoritative value, or the snapshot of the estimation error, which could be negative or positive

volokluev (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
mcannizz: There will be multiple requests during a metric time bucket, so using a gauge will give you the error for one of those requests and ignore all the others. Is that what you intend? I'm not sure what you mean by authoritative value.

mcannizz:
@xurui-c I think the suspect issue that Sentry reported about this PR is worth a look - we may have a real bug here.

PR #7054, opened 2025-04-09
feat(eap): add a sample count threshold for percentile reliability

davidtsuk:
Adding back the sample count threshold only for percentile reliability.


snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -191,6 +192,9 @@ def reliability(self) -> Reliability.ValueType:
             abs(max_err / self.value) if self.value != 0 else float("inf")
         )
 
+        if self.sample_count <= PERCENTILE_SAMPLE_COUNT_THRESHOLD:
+            return Reliability.RELIABILITY_LOW
+
```
volokluev, line 197: could probably move this up top so you don't have to do all the other calculations

volokluev (approved):

PR #7055, opened 2025-04-09
fix(sampling-in-storage): query scans 0 bytes

xurui-c:
There are 2 possibilities that can cause a 0 div error like [this one](https://sentry.sentry.io/issues/6532558122/?project=300688&query=firstRelease%3A7498331448cc0b2572a944654f36df17308a3a4b%20issue.type%3Aerror%20%21level%3Ainfo%20%21server_name%3A%2Aeap%2A&referrer=issue-stream&stream_index=0)
1. `ClickhouseProfile` never gets filled, which shouldnt happen, so I just error if it does
2. The query actually scanned 0 bytes, in which case I'd want to see what the result from CH looks like

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -1,5 +1,6 @@
 import uuid
-from typing import Callable, Dict, Optional, Tuple, TypeAlias, TypeVar, Union, cast
+from dataclasses import asdict
+from typing import Any, Callable, Dict, Optional, Tuple, TypeAlias, TypeVar, Union, cast
 
 import sentry_sdk
 from google.protobuf.json_format import MessageToDict
@@ -21,7 +22,6 @@ from snuba.request import Request as SnubaRequest
 from snuba.utils.metrics import MetricsBackend
 from snuba.utils.metrics.timer import Timer
 from snuba.utils.metrics.util import with_span
-from snuba.utils.serializable_exception import SerializableException
 from snuba.web import QueryResult
 from snuba.web.query import run_query
 
@@ -29,11 +29,6 @@ T = TypeVar("T", TimeSeriesRequest, TraceItemTableRequest)
 MetricsBackendType: TypeAlias = Callable[
     [str, Union[int, float], Optional[Dict[str, str]], Optional[str]], None
 ]
-
-
-class QueryResultMissingInfoException(SerializableException):
-    def __init__(self, info: str, res: QueryResult):
-        super().__init__(f"Query result missing {info}: {res}")
 
 
 DOWNSAMPLING_TIER_MULTIPLIERS: Dict[Tier, int] = {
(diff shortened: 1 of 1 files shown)


snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -68,7 +74,14 @@ def _get_bytes_scanned_limit() -> int:
 
 
 def _get_query_bytes_scanned(res: QueryResult) -> int:
-    return cast(int, res.result.get("profile", {}).get("progress_bytes", 0))  # type: ignore
+    profile_dict = res.result.get("profile")
+    if profile_dict is None:
+        raise QueryResultMissingInfoException("profile", res)
```
volokluev, line 79: this will fail a query for a user, don't do this, log it as an error and sentry will pick it up

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -68,7 +74,14 @@ def _get_bytes_scanned_limit() -> int:
...
+
+    progress_bytes = profile_dict.get("progress_bytes")
+    if progress_bytes is None:
+        raise QueryResultMissingInfoException("progress_bytes", res)
```
volokluev, line 83: same here

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
volokluev: you could even add the result to the span (minus the `data` key which you should not include

kylemumma (dismissed):
lgtm

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -298,15 +311,28 @@ def run_query_to_correct_tier(
...
+                    - actual_query_bytes_scanned
                 )
                 _record_value_in_span_and_DD(
                     span,
                     metrics_backend.distribution,
                     "estimation_error_percentage",
-                    abs(estimation_error) / _get_query_bytes_scanned(res),
+                    abs(estimation_error) / actual_query_bytes_scanned
+                    if actual_query_bytes_scanned != 0
+                    else abs(estimation_error),
```
volokluev, line 335: if the acutal_query_bytes_scanned is 0 then it's meaningless to record the estimation_error_percentage. Just skip doing this in that case

kylemumma:
oh yeah I agree with Volo. Failure in error estimation shouldn't fail the users query

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
kylemumma: why is it meaningless? it still seems relevant to me

volokluev: this metric is recording the estimation error as a fraction of the actual bytes scanned. If no bytes were scanned (which is an error condition), and instead you record the estimation error in bytes (which again is wrong because you don't *know* what the estimation error was), all this does is pollute the metric

kylemumma (commented):
I agree with Volo. Failure in error estimation shouldn't fail the users query

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -67,8 +68,30 @@ def _get_bytes_scanned_limit() -> int:
...
+    if progress_bytes is not None:
+        return cast(int, progress_bytes)
+
+    error_type = (
+        "QueryResult_contains_no_profile"
+        if profile_dict is None
+        else "QueryResult_contains_no_progress_bytes"
+    )
+    res_without_user_data = exclude_user_data_from_res(res)
+    sentry_sdk.capture_message(f"{error_type}: {res_without_user_data}", level="error")
```
xurui-c, line 92: this might cause multilpe sentry errors to be emitted for one in_msg, but it's fine

volokluev (approved):

PR #7058, opened 2025-04-10
fix(sampling-in-storage): fix division by zero in estimation metric

mjq:
I'm seeing queries to EAP fail with a `division by zero` error (see [issue in Sentry](https://sentry.sentry.io/issues/6532558122/)) introduced in #7051.

`_get_query_bytes_scanned` can return 0:
https://github.com/getsentry/snuba/blob/04609965f16b12e83b411ad0828a44bdcf7ec33a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py#L70-L71
This throws when used as a divisor in the `estimation_error_percentage` metric. This change makes emitting the metric conditional on bytes scanned being > 0, avoiding the division otherwise.

Note that I don't know if 0 is ever an expected value here - there might be an underlying issue as well.

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -302,13 +302,14 @@ def run_query_to_correct_tier(
                     estimated_target_tier_query_bytes_scanned
                     - _get_query_bytes_scanned(res)
                 )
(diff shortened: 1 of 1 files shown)


davidtsuk:
I think this PR already fixes the issue: https://github.com/getsentry/snuba/pull/7055

mjq:
@davidtsuk Yes looks like it. Thanks for the pointer, closing this one 👍

PR #7061, opened 2025-04-10
perf(sampling-in-storage): track timeout queries

xurui-c:


snuba/web/rpc/__init__.py
```diff
@@ -190,6 +210,9 @@ def __extract_request_tags(self, in_msg: Tin) -> dict[str, str]:
         if hasattr(meta, "referrer"):
             tags["referrer"] = meta.referrer
 
+        if self._uses_storage_routing(in_msg):
+            tags["storage_routing_mode"] = meta.downsampled_storage_config.mode
```
volokluev, line 214: this would be an integer value right? Can you make it the string name?

snuba/web/rpc/__init__.py
```diff
@@ -40,6 +41,12 @@
 ]
 _BUCKETS_COUNT = len(_TIME_PERIOD_HOURS_BUCKETS)
 
+_STORAGE_ROUTING_MODE_MAP = {
+    0: "MODE_UNSPECIFIED",
+    1: "MODE_PREFLIGHT",
+    2: "MODE_BEST_EFFORT",
+}
+
```
volokluev, line 49: don't do this there's a `name()` method on enum types

snuba/web/rpc/__init__.py
```diff
@@ -190,6 +223,11 @@ def __extract_request_tags(self, in_msg: Tin) -> dict[str, str]:
         if hasattr(meta, "referrer"):
             tags["referrer"] = meta.referrer
 
+        if self._uses_storage_routing(in_msg):
+            tags["storage_routing_mode"] = _STORAGE_ROUTING_MODE_MAP[
+                meta.downsampled_storage_config.mode
+            ]
+
```
volokluev, line 230: you're still using it here

snuba/web/rpc/__init__.py
volokluev: remove this

volokluev (approved):

PR #7065, opened 2025-04-11
Set timeout_overflow_mode to break on all tier 1 queries

volokluev:


phacops (approved):

snuba/datasets/configuration/events_analytics_platform/storages/eap_items.yaml
```diff
@@ -231,13 +231,15 @@ query_processors:
   - processor: TupleUnaliaser
   - processor: ClickhouseSettingsOverride
     args:
+      overwrite_existing: false
```
xurui-c, line 234: I don't know how likely this situation is, but, what if in the future we want to override clickhouse settings for specific customers? For example, shorter max_execution_time for some customers.

PR #7066, opened 2025-04-11
feat(sampling-in-storage): most downsampled tier is Tier 64, also refactored

xurui-c:


volokluev:
also... what?

xurui-c:
@volokluev also refactored

phacops (approved):

phacops:
> also... what?

Based on manual QA, we see tier512 being a small amount of data for small orgs which doesn't provide a good way to extrapolate (hence why we also bump the threshold).

On top of that, we noticed the data we show until we display the result of the higher tier query is meaningfully different.

Based on this, bumping the minimal tier would make us query more data in hopes of better accuracy when estimating what higher tier to hit and also present less of a difference in the shape of the data.

xurui-c:
closing in favor of https://github.com/getsentry/snuba/pull/7067

PR #7067, opened 2025-04-12
feat(sampling-in-storage): most downsampled tier is now tier 64

xurui-c:
Based on manual QA, we see tier512 being a small amount of data for small orgs which doesn't provide a good way to extrapolate (hence why we also bump the threshold).

On top of that, we noticed the data we show until we display the result of the higher tier query is meaningfully different.

Based on this, bumping the minimal tier would make us query more data in hopes of better accuracy when estimating what higher tier to hit and also present less of a difference in the shape of the data.

https://github.com/getsentry/snuba/pull/7066#issuecomment-2798106164

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
@@ -99,7 +99,14 @@ def _get_query_duration_ms(res: QueryResult) -> float:
 
 
 def _get_most_downsampled_tier() -> Tier:
-    return sorted(Tier, reverse=True)[0]
+    return sorted(Tier, reverse=True)[1]
+
+
+def _get_multiplier(tier: Tier) -> int:
+    return int(
+        DOWNSAMPLING_TIER_MULTIPLIERS[tier]
+        / DOWNSAMPLING_TIER_MULTIPLIERS[_get_most_downsampled_tier()]
+    )
 
 
 def _record_value_in_span_and_DD(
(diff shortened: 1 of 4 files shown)


tests/web/rpc/v1/test_sampling_in_storage_util.py
```diff
@@ -47,9 +46,9 @@ def test_get_target_tier(
         target_tier, estimated_target_tier_query_bytes_scanned = _get_target_tier(
             res, metrics_mock, DOESNT_MATTER_STR, timer
         )
+        print(estimated_target_tier_query_bytes_scanned)
```
phacops, line 49: Let's remove this.

phacops (approved):

snuba/web/rpc/v1/resolvers/R_eap_spans/common/sampling_in_storage_util.py
```diff
@@ -132,16 +139,14 @@ def _get_target_tier(
 
         target_tier = _get_most_downsampled_tier()
         estimated_target_tier_bytes_scanned = (
-            most_downsampled_query_bytes_scanned
-            * cast(int, DOWNSAMPLING_TIER_MULTIPLIERS.get(target_tier))
+            most_downsampled_query_bytes_scanned * _get_multiplier(target_tier)
```
mcannizz, line 142: It seems like it may be possible for this function to return the 512 tier since that tier is still in the list of tiers from which this function picks its return value. Maybe this doesn't happen in practice because of how the logic works, but it's not immediately obvious.

We could just remove the 512 tier from that list, but I'd imagine we considered and rejected that approach. What's the reason?

PR #7073, opened 2025-04-14
feat(eap): use sampling factor in aggregations

davidtsuk:
Closes https://github.com/getsentry/eap-planning/issues/239

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -223,7 +223,7 @@ USE_EAP_ITEMS_TABLE = bool(os.environ.get("USE_EAP_ITEMS_TABLE", True))
 # Represents 12AM PST March 12, 2025. We can remove this setting once 30 days have passed since this date.
 USE_EAP_ITEMS_TABLE_START_TIMESTAMP_SECONDS = 1741762800
 
-# Represents 10AM PST April 8, 2025. We can remove this setting once 90 days have passed since this date.
+# Represents 10AM PST April 8, 2025 which is the date we started writing the sampling factor. We can remove this setting once 90 days have passed since this date.
 USE_SAMPLING_FACTOR_TIMESTAMP_SECONDS = 1744131600
 
 # Processor/Writer Options
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -223,6 +223,9 @@ USE_EAP_ITEMS_TABLE = bool(os.environ.get("USE_EAP_ITEMS_TABLE", True))
 # Represents 12AM PST March 12, 2025. We can remove this setting once 30 days have passed since this date.
 USE_EAP_ITEMS_TABLE_START_TIMESTAMP_SECONDS = 1741762800
 
+# Represents 10AM PST April 8, 2025. We can remove this setting once 90 days have passed since this date.
+USE_SAMPLING_FACTOR_TIMESTAMP_SECONDS = 1744131600
+
 # Processor/Writer Options
 
 
--- a/snuba/web/rpc/common/common.py
+++ b/snuba/web/rpc/common/common.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timedelta
-from typing import Callable
+from typing import Callable, cast
 
 from sentry_protos.snuba.v1.request_common_pb2 import RequestMeta
 from sentry_protos.snuba.v1.trace_item_attribute_pb2 import AttributeKey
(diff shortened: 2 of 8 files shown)


snuba/settings/__init__.py
```diff
@@ -223,6 +223,9 @@ class RedisClusters(TypedDict):
 # Represents 12AM PST March 12, 2025. We can remove this setting once 30 days have passed since this date.
 USE_EAP_ITEMS_TABLE_START_TIMESTAMP_SECONDS = 1741762800
 
+# Represents 10AM PST April 8, 2025. We can remove this setting once 90 days have passed since this date.
```
volokluev, line 226: might as well say what that date is

snuba/web/rpc/common/common.py
```diff
@@ -59,6 +60,20 @@ def truncate_request_meta_to_day(meta: RequestMeta) -> None:
     meta.end_timestamp.seconds = int(end_timestamp.timestamp())
 
 
+def use_sampling_factor(meta: RequestMeta) -> bool:
```
volokluev, line 63: give a short explanation as to why this exists

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -29,12 +29,17 @@
...
 sampling_weight_column = column("sampling_weight")
+sampling_factor_column = column("sampling_factor")
 
 Z_VALUE_P95 = 1.96  # Z value for 95% confidence interval is 1.96 which comes from the normal distribution z score.
 Z_VALUE_P975 = 2.24  # Z value for 97.5% confidence interval used for the avg() CI
 
 PERCENTILE_PRECISION = 100000
 PERCENTILE_SAMPLE_COUNT_THRESHOLD = 50
+PERCENTILE_CORRECTION_FACTOR = (
+    1000  # We multiply the sampling weight by this factor to reduce rounding error
```
volokluev, line 40: could use more detailed explanation here as to why. Your document has good context you can include here

snuba/web/rpc/common/common.py
```diff
@@ -59,6 +60,20 @@ def truncate_request_meta_to_day(meta: RequestMeta) -> None:
...
+    use_sampling_factor_timestamp_seconds = cast(
+        int,
+        state.get_int_config(
+            "use_sampling_factor_timestamp_seconds",
+            settings.USE_SAMPLING_FACTOR_TIMESTAMP_SECONDS,
+        ),
```
volokluev, line 69: can this return None?

volokluev: if the user mistypes a runtime config, will the API still work?

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -369,9 +381,16 @@ def _get_possible_percentiles_expression(
         aggregation, {"granularity": str(granularity), "width": str(width)}
     )
     alias_dict = {"alias": alias} if alias else {}
+    sampling_weight = (
+        f.divide(1, sampling_factor_column)
+        if use_sampling_factor
+        else sampling_weight_column
+    )
```
volokluev, line 388: you're repeating this line over and over, might be good to put it into its own function

snuba/web/rpc/v1/resolvers/common/aggregation.py
```diff
@@ -412,51 +437,70 @@ def get_extrapolated_function(
...
+                    get_field_existence_expression(field), condition_in_aggregation
+                ),
+            ),
             **alias_dict,
         ),
         Function.FUNCTION_P50: cf.quantileTDigestWeightedIfOrNull(0.5)(
             field,
-            sampling_weight_column,
+            f.cast(
+                f.round(f.multiply(sampling_weight, PERCENTILE_CORRECTION_FACTOR)),
```
volokluev, line 465: this one as well

volokluev (commented):
Just a few notes. Also, all the tests should be using this functionality already right?

snuba/web/rpc/common/common.py
davidtsuk: It will just use the default, so it should still work. In fact, we don't even really need to use a runtime config here, but it gives us a way to turn this off if needed by setting it to some date far in the future.

snuba/web/rpc/v1/resolvers/common/aggregation.py
volokluev: you made it its own function but never used it lol

davidtsuk: oh oops

volokluev (approved):

PR #7077, opened 2025-04-15
Fix: Fixed another hardcoded http url schema in clickhouse connection string

LimeBeck:
Fixed another hardcoded http url schema in clickhouse connection string

P.S.: Honestly, this part should be somewhere in clickhouse utility functions, like `create_connection_url`, but I don't know where to place it

### Legal Boilerplate


untitaker (approved):
this consumer is currently not used in prod and is unlikely to be, so there's no strong need to refactor or fix

untitaker:
pls fix linters, otherwise lgtm

phacops:
This consumer was removed.

PR #7088, opened 2025-04-16
feat: eap traceitemtable, enable default values for formulas

kylemumma:
enables users to specify a default value for formulas. the default is applied if the formula ever evaluates to null. resolves this ticket https://github.com/orgs/getsentry/projects/284/views/1?pane=issue&itemId=104071496&issue=getsentry%7Ceap-planning%7C232

still needs to be done for timeseries

### DIFF
--- a/snuba/web/rpc/v1/resolvers/R_eap_items/resolver_trace_item_table.py
+++ b/snuba/web/rpc/v1/resolvers/R_eap_items/resolver_trace_item_table.py
@@ -215,10 +215,21 @@ def _get_reliability_context_columns(
 def _formula_to_expression(
     formula: Column.BinaryFormula, request_meta: RequestMeta
 ) -> Expression:
-    return OP_TO_EXPR[formula.op](
+    formula_expr = OP_TO_EXPR[formula.op](
         _column_to_expression(formula.left, request_meta),
         _column_to_expression(formula.right, request_meta),
     )
+    match formula.WhichOneof("default_value"):
+        case None:
+            return formula_expr
+        case "default_value_double":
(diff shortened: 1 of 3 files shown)


phacops (approved):

snuba/web/rpc/v1/resolvers/R_eap_items/resolver_trace_item_table.py
```diff
@@ -215,10 +215,21 @@ def _get_reliability_context_columns(
...
+        case None:
+            return formula_expr
+        case "default_value_double":
+            return f.coalesce(formula_expr, formula.default_value_double)
+        case "default_value_int64":
+            return f.coalesce(formula_expr, formula.default_value_int64)
+        case default:
+            raise BadSnubaRPCRequestException(
+                f"Unknown default_value in formula. Expected default_value_double or default_value_int64 but got {default}"
+            )
```
onkar, line 232: A nitpick but seems like the standard way to handle default is to use `_` and not bind the value to `default`:

```
match formula.WhichOneof("default_value"):
    case None:
        return formula_expr
    case "default_value_double":
        return f.coalesce(formula_expr, formula.default_value_double)
    case "default_value_int64":
        return f.coalesce(formula_expr, formula.default_value_int64)
    case _:
        raise BadSnubaRPCRequestException(
            f"Unknown default_value in formula. Expected default_value_double or default_value_int64 but got {formula.WhichOneof('default_value')}"
        )
```
https://stackoverflow.com/questions/68804209/how-to-do-an-else-default-in-match-case

onkar (approved):

PR #7092, opened 2025-04-17
fix: Add a cli flag for the quantized rebalance delay

evanh:
Add a cli flag for the quantized rebalance delay so that it doesn't have to be set via the config.


volokluev (commented):
would this do anything for rust consumers?

evanh:
> would this do anything for rust consumers?

The rust consumer already has this CLI option: https://github.com/getsentry/snuba/blob/master/snuba/cli/rust_consumer.py#L188

MeredithAnya (approved):

PR #7094, opened 2025-04-17
perf(sampling-in-storage): detect when we hit timeout_overflow_mode

xurui-c:


snuba/web/rpc/v1/resolvers/R_eap_items/routing_strategies/linear_bytes_scanned_storage_routing.py
```diff
@@ -263,8 +263,16 @@ def _emit_estimation_error_info(
 
     def _output_metrics(self, routing_context: RoutingContext) -> None:
         assert routing_context.query_result
+        target_tier = routing_context.query_settings.get_sampling_tier()
+        query_duration_ms = self._get_query_duration_ms(routing_context.query_result)
+        if query_duration_ms >= 25000:
```
volokluev, line 268: this time budget should be a runtime config, defaulted in settings

snuba/web/rpc/v1/resolvers/R_eap_items/routing_strategies/linear_bytes_scanned_storage_routing.py
```diff
@@ -263,8 +263,16 @@ def _emit_estimation_error_info(
 
     def _output_metrics(self, routing_context: RoutingContext) -> None:
         assert routing_context.query_result
+        target_tier = routing_context.query_settings.get_sampling_tier()
+        query_duration_ms = self._get_query_duration_ms(routing_context.query_result)
+        if query_duration_ms >= self._get_time_budget_ms():
```
volokluev, line 268: I would subtract 500ms from this because I don't think the timeout overflow mode = break is that precise

volokluev (approved):

PR #7096, opened 2025-04-17
feat(sampling_in_storage): Log routing decisions to querylog

volokluev:
Right now we don't have a good way to drill down and understand our routing decisions. spans do an okay job but are subject to sampling. This PR:

* creates a hacky entry in the querylog for every routing decision we make, it abuses the `clickhouse_queries.stats` field which accepts arbitrary JSON and the `dataset=storage_routing` field to make the requests separate
* All metrics outputted by the routing strategy are added to the payload
* Adds tests for this functionality

### DIFF
--- a/snuba/settings/__init__.py
+++ b/snuba/settings/__init__.py
@@ -287,6 +287,12 @@ PRETTY_FORMAT_EXPRESSIONS = os.environ.get("PRETTY_FORMAT_EXPRESSIONS", "1") == 
 # situation eventually)
 RAISE_ON_ALLOCATION_POLICY_FAILURES = False
 
+# By default, routing strategies won't block requests from going through in a production
+# environment to not cause incidents unnecessarily. If something goes wrong with routing strategy
+# code, the request will still be able to go through (but it will create a dangerous
+# situation eventually)
+RAISE_ON_ROUTING_STRATEGY_FAILURES = False
+
 # By default, the readthrough cache won't block requests from going through in a production
 # environment to not cause incidents unnecessarily. If something goes wrong with redis or the readthrough cache
 # the request will still be able to go through as if the cache did not exist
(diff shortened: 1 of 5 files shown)


onewland:
Would it be correct to say that with this approach every RPC-initiated query makes two querylog entries, one for the actual ClickHouse SQL execution and the other an entry in `storage_routing`?

Is there a way to correlate the original query + its routing decision? Or are we just looking for aggregate patterns?

volokluev:
> Would it be correct to say that with this approach every RPC-initiated query makes two querylog entries, one for the actual ClickHouse SQL execution and the other an entry in storage_routing?

correct

> Is there a way to correlate the original query + its routing decision? Or are we just looking for aggregate patterns?

Yes: https://github.com/getsentry/snuba/pull/7096/files#diff-a82dffbe17acd3169b7c95e9d58c0c2f6944795c6a41067dce6a42ca44ba3005R202 

the source message is also saved in both entries

xurui-c:
is the metric (bytes scanned, rows scanned, etc) on the most downsampled tier recorded?

xurui-c (approved):

volokluev:
> is the metric (bytes scanned, rows scanned, etc) on the most downsampled tier recorded?

yes, you'll notice that I'm adding all the logged metrics to the routing context's extra info as well

https://github.com/getsentry/snuba/pull/7096/files#diff-d11a1f27525c7199e98450db0d93e1fa6cdd4205b748686e1f091e334d947b78R199-R203

volokluev:
You'll see this is tested [here](https://github.com/getsentry/snuba/pull/7096/files#diff-a82dffbe17acd3169b7c95e9d58c0c2f6944795c6a41067dce6a42ca44ba3005R195-R199)

PR #7097, opened 2025-04-18
[wip] job(eap): rewrite eap_spans subscriptions to be based on eap_items

onewland:
before we can delete the EAP_SPANS entity key, we need subscriptions to be based off EAP_ITEMS. In our largest environment there are only about 50 of these so I think this can run in a few seconds


volokluev:
This is only redis which is a cache. The subscriptions are actually populated from sentry

onewland:
> This is only redis which is a cache. The subscriptions are actually populated from sentry

I can update the subscription data in Sentry but I think calling this a cache is a little misleading as only manual operations cause this to get reloaded from Postgres; this is essentially a long-lived persistent store.

onewland:
To your point, though, we should also be fixing the upstream to make items subscriptions (or give up on deprecating EAP_SPANS the entity)

onewland:
we're gonna go with a different approach here

PR #7101, opened 2025-04-18
cleanup(eap): remove some, but not all, eap_spans stuff

onewland:
Leaving the storage + entity definitions intact so we don't have to deal with subscriptions immediately, but this should effectively remove all query paths to eap_spans.


tests/web/rpc/v1/test_endpoint_get_trace.py
```diff
@@ -364,18 +363,3 @@ def test_with_specific_attributes(self, setup_teardown: Any) -> None:
             ],
         )
         assert MessageToDict(response) == MessageToDict(expected_response)
-
-
-@pytest.mark.clickhouse_db
```
volokluev, line 369: I think these tests double write to eap_spans, you can remove that now as well

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -1315,120 +1314,6 @@ def test_literal(self) -> None:
             expected_formula_timeseries,
         ]
 
-
-class TestUtils:
```
volokluev, line 1319: I don't think this should be removed

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -1315,120 +1314,6 @@ def test_literal(self) -> None:
...
-        )
-
-        _validate_time_buckets(message)
-        # add another bucket to fit into granularity_secs
-        assert message.meta.end_timestamp.seconds == int(BASE_TIME.timestamp()) + 75
-
-
-@pytest.mark.clickhouse_db
-@pytest.mark.redis_db
-class TestTimeSeriesApiEAPItems(TestTimeSeriesApi):
```
volokluev, line 1420: this one can be removed but also remove the double write

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
```diff
@@ -149,32 +148,6 @@ def test_empty_results(self) -> None:
...
-    # (straight from the source table) without significant performance degradation
-    # as of 2025-03-14, this functionality has also not been used. As such for now, it's better to
-    # ship an autocomplete that is mostly functional minus this paginatin feature. Instead we rely on the
-    # user to add more characters in the value_substring_match
-    def test_page_token(self, setup_teardown: Any) -> None:
-        pass
-
-    def test_page_token_filter_offset(self) -> None:
-        pass
-
```
volokluev, line 177: You'll have to remove or skip these tests from the original test class

volokluev (commented):
nice work, there's also the code in:
https://github.com/getsentry/snuba/tree/master/snuba/web/rpc/v1/legacy

not sure if that's on your radar. please remove the double write though and we should be all good

tests/datasets/entities/storage_selectors/test_eap_items.py
```diff
@@ -10,7 +10,7 @@
 
 EAP_ITEMS_ENTITY = Entity(
     key=EntityKey("eap_items"),
-    schema=get_entity(EntityKey("eap_spans")).get_data_model(),
+    schema=get_entity(EntityKey("eap_items")).get_data_model(),
```
kylemumma, line 13: why did this have to change when eap items is already in use?

kylemumma, line 13: why did this have to change?

kylemumma, line 13: why did this have to change

kylemumma (commented):
LGTM once you address volos comments

tests/web/rpc/v1/test_trace_item_attribute_values_v1.py
onewland: I think these don't override any actual test. 

```
➜  snuba git:(cleanup/remove-eap-spans-from-queries-but-keep-key) ✗ git --no-pager grep -c 'test_page_token'
➜  snuba git:(cleanup/remove-eap-spans-from-queries-but-keep-key) ✗ git --no-pager grep -c 'test_page_token_filter_offset'
```

they're probably referencing something we already deleted

tests/datasets/entities/storage_selectors/test_eap_items.py
onewland: i'm not sure it did, but it seems to me like it was just wrong to assume that schema should come from eap_spans

onewland:
> nice work, there's also the code in: https://github.com/getsentry/snuba/tree/master/snuba/web/rpc/v1/legacy
> 
> not sure if that's on your radar. please remove the double write though and we should be all good

double writes removed, I'll get these additional files in a follow-up PR. this one is already pretty big

kylemumma (approved):

PR #7108, opened 2025-04-22
feat(eap): store span kind in EAP Items

mjq:
Spans have a `kind` field that is [sent over Kafka](https://github.com/getsentry/relay/blob/888d9f54854d1bf49ce7b82d6658e8baf614d9ba/relay-server/src/services/store.rs#L1427) but not yet stored in the EAP Items table. Write it to the `kind` key, to follow the intended search key for this field.

Fixes OPE-1.

### DIFF
--- a/rust_snuba/src/processors/eap_items.rs
+++ b/rust_snuba/src/processors/eap_items.rs
@@ -246,6 +246,10 @@ impl From<FromSpanMessage> for EAPItem {
             res.attributes
                 .insert_bool("sentry.is_segment".to_string(), from.is_segment);
 
+            if let Some(kind) = from.kind {
+                res.attributes.insert_str("kind".to_string(), kind);
+            }
+
             if let Some(parent_span_id) = from.parent_span_id {
                 res.attributes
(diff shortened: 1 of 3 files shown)


mjq:
Updated schema en route to fixing the CI failure: see https://github.com/getsentry/sentry-kafka-schemas/pull/400

phacops (changes requested):
This is something that should be set in the producer.

Having the items consumer be the one that set these attributes is a pattern we want to stop as we transition to the generic items pipeline.

mjq:
@phacops I agree, but it wasn't really possible until the items consumer is in and I didn't want to wait for it. Thanks for the pointer to https://github.com/getsentry/snuba/pull/7122 though, maybe that means the items consumer will be ready sooner rather than later? In that case I'm okay to wait

phacops:
I don't think you have to wait for anything, you can put it in the producer (in Relay) and have access to it.

At the same time, I think you should also put the rest of the attributes we promote in Relay as well.

PR #7111, opened 2025-04-22
fix(eap): Improve performance by using a DISTINCT instead of a LIMIT BY

phacops:
Using `LIMIT BY` was creating performance issues, using `DISTINCT` doesn't.

We need to remove the optimization with the `timestamp` values but since we have a `trace_id` bloom filter index, it shouldn't have too much impact as long as this query uses the index.

### DIFF
--- a/snuba/web/rpc/v1/endpoint_get_traces.py
+++ b/snuba/web/rpc/v1/endpoint_get_traces.py
@@ -510,7 +510,7 @@ class EndpointGetTraces(RPCEndpoint[GetTracesRequest, GetTracesResponse]):
         )
         trace_ids: list[str] = []
         for row in results.result.get("data", []):
-            trace_ids.append(row["trace_id"])
+            trace_ids.append(list(row.values())[0])
         return trace_ids
 
     def _get_metadata_for_traces(
--- a/snuba/web/rpc/v1/endpoint_get_traces.py
+++ b/snuba/web/rpc/v1/endpoint_get_traces.py
@@ -43,7 +43,6 @@ from snuba.web.rpc import RPCEndpoint
 from snuba.web.rpc.common.common import (
     base_conditions_and,
     project_id_and_org_conditions,
-    timestamp_in_range_condition,
     trace_item_filters_to_expression,
     treeify_or_and_conditions,
 )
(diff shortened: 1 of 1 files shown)


snuba/web/rpc/v1/endpoint_get_traces.py
```diff
@@ -447,22 +446,8 @@ def _list_trace_ids(
         selected_columns: list[SelectedExpression] = [
             SelectedExpression(
                 name="trace_id",
-                expression=f.cast(
+                expression=f.distinct(
                     column("trace_id"),
-                    "String",
-                    alias="trace_id",
```
colin-sentry, line 453: still needs alias

colin-sentry: also isn't _BUFFER_WINDOW unused now?

phacops: I removed the need for an alias, otherwise Snuba generates a query with parenthesis and ClickHouse rejects the use of `distinct` as a function, otherwise it doesn't.

colin-sentry: it should still work with parenthesis?

```
918234ee834b :) select distinct(item_id) as blah from eap_items_1_local limit 10;

SELECT DISTINCT item_id AS blah
FROM eap_items_1_local
LIMIT 10

Query id: a18e7a76-4548-412a-b843-939efea5b66b

┌──────────────────────────────────blah─┐
│ 2109481693716072850015838260209320467 │
│ 2109481693732989345846189931518514808 │
└───────────────────────────────────────┘
┌─────────────────blah─┐
│  9504858063562667929 │
│ 11614359868351650071 │
│ 12255089683557749555 │
│ 12478751934349190228 │
│ 13513421960033050652 │
│ 13686746155332883608 │
│  9383199407994008684 │
│ 10514034753536212285 │
└──────────────────────┘
```

phacops: Snuba generates parenthesis around the distinct + alias. I'm not sure how we get around this.
```
Query:
select (distinct(item_id) as blah) from eap_items_1_local limit 10;


Received exception from server:
Code: 46. DB::Exception: Unknown function distinct: While processing distinct(item_id) AS blah. (UNKNOWN_FUNCTION) (version 23.8.11.29.altinitystable (altinity build))
```

volokluev (approved):

PR #7112, opened 2025-04-22
feat: throw error if duplicate labels are provided in RPC endpoints

kylemumma:
this makes it timeseries and itemtable endpoints throw BadSnubaRPCException if there are duplicate labels in the request. previously there was no error and it was just causing unexpected results for users.

see: https://github.com/getsentry/eap-planning/issues/242

this resolves https://github.com/getsentry/eap-planning/issues/244

### DIFF
--- a/tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
+++ b/tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
@@ -40,7 +40,6 @@ from sentry_protos.snuba.v1.trace_item_filter_pb2 import (
 
 from snuba.datasets.storages.factory import get_storage
 from snuba.datasets.storages.storage_key import StorageKey
-from snuba.query.parser.exceptions import AliasShadowingException
 from snuba.web import QueryException
 from snuba.web.rpc import RPCEndpoint
 from snuba.web.rpc.common.exceptions import BadSnubaRPCRequestException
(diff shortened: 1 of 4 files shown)


volokluev (commented):
The validation of labels should be done on the request object, not the snuba query. You're relying too much on an internal detail of how we represent queries rather than validating the user input directly and telling the use exactly what is wrong

kylemumma:
Its not so simple to do the validation of duplicate labels on the request objects. Its a much simpler solution to do it this way, and personally I think its a better one. Heres why its not simple:
1. The main reason is because in practice labels in the request object are frequently duplicated, but it doesnt lead to errors later in the pipeline. Heres an example
```
expression {
   label = "sum"
   aggregations {
      label = "sum"
      ...
   }
}
```
this would throw duplicate error but in reality doesnt cause any issues w duplicate aliases later on, due to how labels get translated to alias (which is pretty much a black box).

I could enforce duplicate detection in this way but it would break existing snuba tests, existing sentry integration tests, and likely existing queries used by product.

2. The current visitor pattern for traversing request objects is very bad. I did this implementation and the visitor objects are very messy and hard to work with. the solution can be done but its longer and more complex and hacky and the visitors should be redesigned.


So I see what you're saying and in theory I agree, but in practice I think this solution is better.

tests/web/rpc/v1/test_endpoint_time_series/test_endpoint_time_series.py
```diff
@@ -1638,3 +1639,156 @@ def test_best_effort_end_to_end(self) -> None:
             response.meta.downsampled_storage_meta.tier
             != DownsampledStorageMeta.SELECTED_TIER_UNSPECIFIED
         )
+
+    def test_duplicate_labels(self) -> None:
```
volokluev, line 1643: this test should be smaller, just bare essentials to show the behavior

PR #7115, opened 2025-04-22
feat(errors) - Add column `timestamp_ms` on errors

yuvmen:
New field will be used to allow better time precision when querying errors


volokluev:
how do you plan to use this column? Just so you know, you're gonna have to still provide timestamp conditions alongside timestamp_ms otherwise the filtering won't work. Just want to make sure that you're aware of that

yuvmen:
hey, yea I am aware, we are going to use it to do like a secondary ordering, not replace the original `timestamp` field.
heres the tech spec for full context:
https://www.notion.so/sentry/Tech-Spec-Increase-Error-Timestamp-Precision-1d78b10e4b5d801590b3e11d6ac6d84c#1d78b10e4b5d81f8aa87d63bc6f8e684

volokluev (approved):
LGTM but our CI considers the yamls a code change, you have to separate them out to merge

PR #7116, opened 2025-04-23
fix(eap): Add condition on item type as we only want to consider spans for GetTraces

phacops:


Zylphrex (approved):
Can confirm the behaviour is still as expected

onkar (approved):

